PR #985, opened 2026-05-05
feat(manifest)!: Replace FetchEntries with iterator approach

hdnpth:
`FetchEntries` materializes every manifest entry into a `[]ManifestEntry` before returning, so a caller pays the full per-manifest entry slice even when it only needs to filter, count, or break out on first match. On tables with thousands of entries per manifest this shows up as avoidable memory pressure.

### DIFF
--- a/cmd/iceberg/output.go
+++ b/cmd/iceberg/output.go
@@ -150,14 +150,20 @@ func (t textOutput) Files(tbl *table.Table, history bool) {
 			snapshotTree = append(snapshotTree, pterm.LeveledListItem{
 				Level: 1, Text: "Manifest: " + m.FilePath(),
 			})
+			var iterErr error
 			for e, err := range m.Entries(afs, false) {
 				if err != nil {
-					t.Error(err)
-					os.Exit(1)
+					iterErr = err
+
+					break
 				}
 				snapshotTree = append(snapshotTree, pterm.LeveledListItem{
 					Level: 2, Text: "Datafile: " + e.DataFile().FilePath(),
 				})
+			}
+			if iterErr != nil {
+				t.Error(iterErr)
+				os.Exit(1)
 			}
 		}
 	}
--- a/cmd/iceberg/output.go
+++ b/cmd/iceberg/output.go
@@ -150,12 +150,11 @@ func (t textOutput) Files(tbl *table.Table, history bool) {
 			snapshotTree = append(snapshotTree, pterm.LeveledListItem{
 				Level: 1, Text: "Manifest: " + m.FilePath(),
 			})
-			datafiles, err := m.FetchEntries(afs, false)
-			if err != nil {
-				t.Error(err)
-				os.Exit(1)
-			}
-			for _, e := range datafiles {
+			for e, err := range m.Entries(afs, false) {
+				if err != nil {
+					t.Error(err)
+					os.Exit(1)
+				}
 				snapshotTree = append(snapshotTree, pterm.LeveledListItem{
 					Level: 2, Text: "Datafile: " + e.DataFile().FilePath(),
 				})
--- a/manifest.go
+++ b/manifest.go
@@ -349,12 +349,17 @@ func (m *manifestFile) Entries(fs iceio.IO, discardDeleted bool) iter.Seq2[Manif
 
 			return
 		}
+		aborted := false
 		defer func() {
-			_ = f.Close()
+			if cerr := f.Close(); cerr != nil && !aborted {
+				yield(nil, cerr)
+			}
 		}()
 
 		for entry, err := range iterManifest(m, f, discardDeleted) {
 			if !yield(entry, err) {
+				aborted = true
+
 				return
 			}
 		}
@@ -790,8 +795,11 @@ func iterManifest(m ManifestFile, f io.Reader, discardDeleted bool) iter.Seq2[Ma
 
 			return
 		}
+		aborted := false
 		defer func() {
-			_ = manifestReader.Close()
+			if cerr := manifestReader.Close(); cerr != nil && !aborted {
+				yield(nil, cerr)
+			}
 		}()
 
 		for {
@@ -800,7 +808,9 @@ func iterManifest(m ManifestFile, f io.Reader, discardDeleted bool) iter.Seq2[Ma
 				if errors.Is(err, io.EOF) {
 					return
 				}
-				yield(nil, err)
+				if !yield(nil, err) {
+					aborted = true
+				}
 
 				return
 			}
@@ -808,6 +818,8 @@ func iterManifest(m ManifestFile, f io.Reader, discardDeleted bool) iter.Seq2[Ma
 				continue
 			}
 			if !yield(entry, nil) {
+				aborted = true
+
 				return
 			}
 		}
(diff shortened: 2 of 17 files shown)


manifest.go
```diff
@@ -463,6 +481,13 @@ type ManifestFile interface {
 	HasAddedFiles() bool
 	// HasExistingFiles returns true if ExistingDataFiles > 0 or if it was null.
 	HasExistingFiles() bool
+	// Entries streams the manifest entries from the manifest file using
+	// the provided file system IO interface. Entries that have been
+	// marked as deleted are skipped if discardDeleted is true.
+	//
+	// Prefer Entries over FetchEntries when walking large manifests
+	// since it avoids loading every entry into memory at once.
+	Entries(fs iceio.IO, discardDeleted bool) iter.Seq2[ManifestEntry, error]
```
laskoviymishka, line 495: Adding `Entries` to the public `ManifestFile` interface is a breaking change for any external implementer or mock — they'll fail to compile against the next module bump. That's probably acceptable for a 0.x module, but I'd want it called out: at minimum a release-note entry, plus a `// Deprecated:` line on `FetchEntries` if the intent is to phase it out.

If we'd rather keep the interface stable, `Entries` could live as a method on the concrete `*manifestFile` only, or as a free helper `iceberg.IterManifestEntries(m, fs, discardDeleted)`. What's the preferred direction here?

manifest.go
```diff
@@ -339,8 +340,35 @@ func (m *manifestFile) FirstRowID() *int64 { return m.FirstRowIDValue }
...
+func (m *manifestFile) Entries(fs iceio.IO, discardDeleted bool) iter.Seq2[ManifestEntry, error] {
+	return func(yield func(ManifestEntry, error) bool) {
+		f, err := fs.Open(m.FilePath())
+		if err != nil {
+			yield(nil, err)
+
+			return
+		}
+		defer func() {
+			_ = f.Close()
```
laskoviymishka: The streaming path drops Close errors here (`_ = f.Close()`), while `FetchEntries` below uses `internal.CheckedClose` to surface them via the named return. For object-store IO wrappers that flush metrics or surface deferred errors via Close, that's a real signal we'd lose as callers migrate from one to the other.

Could we either yield the close error as a final synthetic `(nil, cerr)` pair, or at minimum call out the asymmetry in both godocs? Same applies to `_ = manifestReader.Close()` inside `IterManifest`.

manifest.go
```diff
@@ -751,33 +776,54 @@ func (c *ManifestReader) ReadEntry() (ManifestEntry, error) {
 	return tmp, nil
 }
 
+// IterManifest returns an iterator that streams manifest entries from
+// the provided reader without buffering them. If discardDeleted is true,
+// entries whose status is "deleted" are skipped.
+func IterManifest(m ManifestFile, f io.Reader, discardDeleted bool) iter.Seq2[ManifestEntry, error] {
```
laskoviymishka: `IterManifest` ends up with one in-tree caller (`ReadManifest`), and the method form `Entries` already owns the file lifecycle for everyone else. Exporting it locks in a third public way to read entries (alongside `Entries` and `FetchEntries`) without an obvious external use case — and the doc-comment doesn't pin down the contract (single-shot? error yields once and stops? close errors swallowed?).

Could we lower-case to `iterManifest` unless there's a known consumer that needs the bare-`io.Reader` form? If it stays exported, worth spelling out those guarantees in the godoc.

cmd/iceberg/output.go
```diff
@@ -150,12 +150,11 @@ func (t textOutput) Files(tbl *table.Table, history bool) {
...
-			datafiles, err := m.FetchEntries(afs, false)
-			if err != nil {
-				t.Error(err)
-				os.Exit(1)
-			}
-			for _, e := range datafiles {
+			for e, err := range m.Entries(afs, false) {
+				if err != nil {
+					t.Error(err)
+					os.Exit(1)
```
laskoviymishka: With the slice-based `FetchEntries` the file was already closed before this branch ran. Now the file is open for the duration of iteration, so `os.Exit(1)` here skips the deferred `f.Close()` inside `Entries`. Cosmetic for a short-lived CLI but a new leak shape worth fixing while we're here — break out of the loop and exit after the range, or refactor `Files` to return an error.

table/scanner.go
```diff
@@ -153,13 +153,12 @@ func GetPartitionRecord(dataFile iceberg.DataFile, partitionType *iceberg.Struct
 func openManifest(io io.IO, manifest iceberg.ManifestFile,
 	partitionFilter, metricsEval func(iceberg.DataFile) (bool, error),
 ) ([]iceberg.ManifestEntry, error) {
-	entries, err := manifest.FetchEntries(io, true)
-	if err != nil {
-		return nil, err
-	}
+	var out []iceberg.ManifestEntry
```
laskoviymishka: We lost the `make([]iceberg.ManifestEntry, 0, len(entries))` preallocation here. `openManifest` is on the scan hot path, so for wide manifests this trades one fixed allocation for several `append` regrowths.

Worth pre-sizing from the manifest counts — `int(manifest.AddedDataFiles()) + int(manifest.ExistingDataFiles())` is a defensible upper bound since `discardDeleted=true`.

laskoviymishka (changes requested):

manifest.go
zeroshade: I think it's fine to deprecate and add the new method since we're 0.x as long as we make sure to call it out in the next release (i.e. change the title of this to be `fix(manifest)!:` so that i gets called out as a breaking change).

Eventually we're going to want to overhaul the public API before we do a 1.0 release anyways.

manifest.go
zeroshade: I vote we yield the close error if it's non-nil

manifest.go
zeroshade: agreed, there's no reason to export both this AND have the `Entries` method on the manifest file itself. What's the purpose of this vs just opening the manifest directly to get a `ManifestFile` and then just calling `Entries(fs, discard)` on it?

manifest.go
hdnpth: Do we actually should care about the error upon Close here?

hdnpth: I do not like approach with returning synthetic pair, because it is so uncanonical

zeroshade: i would say that we should *only* return the pair if the error from close is non-nil. If the error is nil, then no need to yield the synthetic pair (follow the canonical scenario of just letting the function end to indicate we're done)

zeroshade: i would say that we should *only* return the pair if the error from close is non-nil. If the error is nil, then no need to yield the synthetic pair (follow the canonical scenario of just letting the function end to indicate we're done)

laskoviymishka (approved):
Nice job, :shipit: .

I diged into the close-error `yield`-from-`defer` pattern too, and it looks correct under Go’s range-over-func behavior. The asks from the previous round are addressed properly, the breaking-change note is clear, and the capacity hints look safe.

This mergeable now. 

A few follow-ups worth tracking rather than holding this PR:

* Add a bit more contract docs on `Entries`: iteration stops on first error, a final `(nil, closeErr)` can be yielded, and early break is safe.
* Add tests for early break + close, mid-stream error, and close-error-as-final-pair.
* Finish migrating the remaining `snapshot_producers.go` callers that still go through `fetchManifestEntry` into a slice.

Tiny nit: the V1 capacity-hint clamp is harmless, but the comment slightly overstates the behavior when only one side is `-1`, but this is not worth blocking.

PR #1017, opened 2026-05-05
feat(schema): reject complex type defaults #1011 

### LINKED ISSUE #1011: feat(schema): reject non-single-value-parseable defaults on complex types
Parent: #589

v3 spec: `initial-default` / `write-default` on struct/list/map fields must be null or a single-value-parseable JSON value. Java enforces this in `SchemaParser`; iceberg-go has rejection rules for `UnknownType` defaults but does not gate complex-type defaults the same way — non-null defaults on struct / list / map fields are silently accepted.

Walk the schema during validation; for every nested-type field with a non-null default, validate that the default parses cleanly under the type's encoding rules (a JSON object for struct, JSON array for list, JSON object-of-pairs for map). Mirror the existing `UnknownType` default rejection's error-message style for consistency.

Tests cover struct, list, and map separately (each a clear failure case + a passing null-default case), plus a primitive case to confirm primitive defaults still pass through unchanged.

tanmayrauth:
Validates that initial-default / write-default on struct, list, and map fields match the expected JSON shape (object for struct/map, array for list). Walks the full schema tree including nested fields within list/map elements. Mirrors the existing unknownTypeValidator pattern.

Closes #1011

### DIFF
--- a/table/metadata_builder_internal_test.go
+++ b/table/metadata_builder_internal_test.go
@@ -1837,3 +1837,147 @@ func TestUnknownTypeValidation(t *testing.T) {
 		require.ErrorContains(t, err, "must be optional")
 	})
 }
+
+func TestComplexTypeDefaultValidation(t *testing.T) {
+	t.Run("InvalidStructInitialDefault", func(t *testing.T) {
+		schema := iceberg.NewSchema(1,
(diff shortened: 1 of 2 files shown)


laskoviymishka (approved):
LGTM

PR #1020, opened 2026-05-05
feat(table): support zstd metadata compression codec

### LINKED ISSUE #1005: feat(table): support zstd metadata compression codec
Parent: #589

The v3 spec allows zstd-compressed `metadata.json`. iceberg-go's metadata-location codec layer handles `gzip` and `none` and returns `unsupported write metadata compression codec` for anything else. Add zstd on both read (decompress when reading a `.metadata.json.zstd`) and write (compress when `write.metadata.compression-codec=zstd`).

`klauspost/compress/zstd` is already a transitive dependency via parquet, so no new go.mod entry. Add a new `MetadataCompressionCodecZstd` constant alongside the existing gzip / none, and thread it through both directions of the codec helper.

Tests: round-trip a zstd-compressed metadata.json end-to-end; golden-fixture test against a Java-produced zstd metadata.json pins the exact byte format.

tanmayrauth:
Adds MetadataCompressionCodecZstd alongside gzip/none and wires it through both read (decompress .metadata.json.zstd) and write (write.metadata.compression-codec=zstd) paths. Uses  klauspost/compress/zstd already present as a transitive dependency.

Closes #1005


zeroshade:
Does the Iceberg spec explicitly state what compression codecs are allow for metadata.json that has zstd part of it?

zeroshade:
I'm trying to add comment reviews but GitHub is failng at this...

Essentially, we should create a more generic helper for this since the only difference between the gzip and zstd routes are essentially `gzip.NewReader` vs `zstd.NewReader`

tanmayrauth:
The spec's write.metadata.compression-codec formally lists only gzip and none, but both PyIceberg and the Java implementation already support zstd as an extension I've followed the same pattern.    For the generic helper point, I've refactored it: extracted a newDecompressor(r io.Reader, codec string) (io.ReadCloser, error) that handles both codecs behind a unified io.ReadCloser  interface, so adding a future codec is just one case in a switch.

catalog/internal/utils.go
```diff
@@ -55,10 +56,19 @@ func WriteTableMetadata(metadata table.Metadata, fs icebergio.WriteFileIO, loc s
...
 		writer = compressWriter
 		defer internal.CheckedClose(compressWriter, &err)
+	case table.MetadataCompressionCodecZstd:
+		enc, zErr := zstd.NewWriter(out)
+		if zErr != nil {
+			return zErr
+		}
+		compressWriter = enc
+		writer = compressWriter
+		defer internal.CheckedClose(compressWriter, &err)
```
zeroshade: we don't need the `compressWriter` intermediate here.

we can just use `defer internal.CheckedClose(enc, &err)` right? `compressWriter` doesn't need to be pre-declared

zeroshade (changes requested):

catalog/internal/utils.go
tanmayrauth: Good catch, removed the intermediate variable. Also cleaned up the gzip case the same way while I was at it.

tanmayrauth:
@zeroshade  the PR has been updated.

zeroshade (approved):

PR #1021, opened 2026-05-05
feat(table): assign row IDs on write for v3 row lineage

### LINKED ISSUE #998: feat(table): assign row IDs on write for v3 row lineage
Parent: #589

The metadata side of row lineage landed in #659 / #728 / #735 / #741 (closed #727). What's missing is the write-side: iceberg-go writes data files but doesn't populate per-row `_row_id` for new appends.

Per spec: the writer assigns each new row a row ID starting at the current `next-row-id`, the manifest entry's `first_row_id` is the smallest assigned ID, the manifest list aggregates, and metadata `next-row-id` advances by the total row count at commit time.

Two questions worth nailing down in this thread before the code goes in:

- **Where row-id assignment happens** — at the writer (per-task pre-claim of a range) or at commit time (post-process the staged files). Per-task pre-claim is simpler under sequential commits but needs care under retries. Java assigns at the writer using a pre-claimed range; mirror unless we have a reason to diverge.
- **How parallel writers coordinate** — each writer claims its range at task setup; commit ordering settles the actual `first_row_id` per file.

Touches: `table/internal/parquet_files.go` (writer), `table/snapshot_producers.go` (commit-time `next-row-id` advance), `table/metadata.go` (consistency check).

Tests: scan after a v3 append asserts `_row_id` increments contiguously across the new files; metadata `next-row-id` advances by exactly the row count of the snapshot. Spec: [Iceberg row lineage](https://iceberg.apache.org/spec/#row-lineage). Java parity references: [apache/iceberg#15187](https://github.com/apache/iceberg/pull/15187), [#15508](https://github.com/apache/iceberg/pull/15508).

tanmayrauth:
Assigns per-row _row_id during v3 appends using a pre-claimed range starting at next-row-id. Sets manifest entry first_row_id to the smallest assigned ID and advances metadata next-row-id by total  row count at commit time. Mirrors Java's approach.

Closes: #998


manifest.go
```diff
@@ -2187,6 +2187,19 @@ func (b *DataFileBuilder) FirstRowID(id int64) *DataFileBuilder {
 	return b
 }
 
+// SetDataFileFirstRowID sets the first_row_id on an existing DataFile.
+// This is used by the snapshot producer to assign row IDs at write time for v3 tables.
+// Returns false if the DataFile implementation does not support mutation.
+func SetDataFileFirstRowID(df DataFile, id int64) bool {
```
laskoviymishka: I'd reconsider exposing this as a top-level mutator. The function type-asserts to the unexported `*dataFile`, returns `false` silently for any other implementation, and aliases the caller's pointer — so a `DataFile` the caller still holds gets its `FirstRowIDField` mutated as a side effect. That's a surprising contract for a public API, and the `bool` return treats what is really an invariant (`*dataFile` is the only impl in this package) as a soft optional.

Could we either fold this into `DataFileBuilder` (build a fresh `*dataFile` with `FirstRowID` set, no in-place mutation) or keep the helper unexported within the package? wdyt?

table/snapshot_producers.go
```diff
@@ -733,7 +733,20 @@ func (sp *snapshotProducer) manifestProducer(content iceberg.ManifestContent, fi
 		}
 		defer internal.CheckedClose(wr, &err)
 
+		// For v3 data manifests, assign first_row_id to each data file.
+		// Each file claims a contiguous range of row IDs starting from NextRowID.
+		assignRowIDs := sp.txn.meta.formatVersion >= 3 && content == iceberg.ManifestContentData
```
laskoviymishka: I think the bigger question is whether we should be assigning per-file `first_row_id` at all on ADDED entries. The v3 spec ("First Row ID Inheritance") says ADDED data files carry null `first_row_id`; the manifest-list writer assigns the manifest-level value, and readers inherit per-file IDs by summing preceding `record_count`s. Java's `MergingSnapshotProducer.appendManifest` explicitly rejects inputs where `manifest.firstRowId() != null`.

iceberg-go already implements both halves of the inheritance model — list writer at `manifest.go:1419-1425`, reader at `manifest.go:734-742`. After this PR, manifests we write would carry non-null per-file IDs that Java's `appendManifest` would refuse to replay, and the per-file values become a second source of truth alongside the manifest-level one.

Is there a reason we're diverging from inheritance here? Or could we drop the per-file assignment and rely on what's already wired up?

table/snapshot_producers.go
```diff
@@ -733,7 +733,20 @@ func (sp *snapshotProducer) manifestProducer(content iceberg.ManifestContent, fi
...
+		nextRowID := sp.txn.meta.NextRowID()
```
laskoviymishka: Even if we keep the per-file assignment, this read happens once at producer-build time and the values get baked into the data-manifest avro. On OCC retry, `rebuildSnapshotUpdates` (`table/table.go`) only rewrites the manifest *list*, not the data manifest entries — so under contention the per-file IDs stay at the stale `NextRowID` from the first attempt while metadata's `NextRowID` has moved.

`validateAndUpdateRowLineage` (`table/metadata.go:474-478`) would then reject the retry deterministically. The recently-landed manifest-list rebuild was designed assuming data manifests can be reused as-is on retry; pre-assigning IDs breaks that invariant.

Worth a v3 OCC retry test either way — the existing `occ_scenario_test.go` patterns should fit. Could we either move the assignment to commit time (after retries settle) or scrap pre-assignment in favor of inheritance?

laskoviymishka (changes requested):
Thanks for picking this up, row lineage is tricky, and getting `first_row_id` plumbed through the write path is *hard*.

I’d hold this version before merge. My main concern is the assignment model. My read of the v3 spec is that added data files should carry `null` `first_row_id`, then the manifest-list writer assigns the manifest-level first row ID, and readers inherit per-file IDs from record counts. `iceberg-go` already has that inheritance machinery, and Java rejects append manifests that come in with pre-assigned row IDs. So I’m worried manifests written by this PR may not round-trip cleanly through Java’s append path.

A few related concerns:

* Pre-assigning IDs during producer build seems fragile with OCC retries. If `NextRowID` advances on retry, the data manifest may still contain stale per-file IDs.
* `mergeAppend` can rewrite manifests, but I don’t think it preserves the manifest-level `FirstRowIDValue`, so bin-packing could make the counters drift.
* “Closes #998” feels a bit too broad. The issue also asks for actual `_row_id` materialization in Parquet, which is not part of this PR.

Before merge, I’d like to align on the intended model: are we intentionally diverging from the spec/Java inheritance approach, and if yes, how should iceberg-go → Java replay work?

If we keep pre-assignment, I think we also need to fix the OCC retry path and avoid exposing `SetDataFileFirstRowID` as a public in-place mutator if possible. Happy to take another pass once that direction is settled.

table/snapshot_producers.go
tanmayrauth: Yeah, this was a real correctness issue. The manifest-list rebuild on retry intentionally treats data manifests as immutable, and baking in per-file IDs at producer-build time breaks that invariant. On contention, stale IDs would get rejected by validateAndUpdateRowLineage deterministically,  making retries useless.

table/snapshot_producers.go
tanmayrauth: You're correct, and I should have read the spec more carefully here. I went back and re-read the "First Row ID Inheritance" section - ADDED files carry null, manifest-list writer assigns the manifest-level value, readers derive per-file from record_count sums. We already implement both halves of that  (list writer at manifest.go:1508, reader at manifest.go:785). So this was just duplicating existing machinery while also creating a second source of truth that would break Java replay.

manifest.go
tanmayrauth: Yeah, you're right - this was a hack. Type-asserting to the unexported concrete type and silently returning false is not a contract anyone should have to reason about. I've removed it entirely in the latest push since the per-file assignment approach is gone now.

Issue #830, opened 2026-04-03
feat: concurrent writer conflict detection (OCC + retry)


laskoviymishka:
### Feature Request / Improvement

Parent: #829 (v2 spec completion)

iceberg-go currently has a single OCC guard: `AssertRefSnapshotID("main", baseSnapshotID)` sent to the catalog (#672). This detects that *someone* else committed, but provides no semantic analysis of *what* changed or whether the conflict is resolvable. There is no retry loop — any conflict is a terminal error.

The Java implementation has full conflict detection with per-operation validation, isolation levels, and retry-with-rebase.

Rust (iceberg-rust) also hasn't implemented this yet ([apache/iceberg-rust#348](https://github.com/apache/iceberg-rust/issues/348)).

### What's missing
**Semantic conflict validation** — when a concurrent commit is detected, check whether the changes actually conflict:
- `FastAppend`: always safe, no validation needed
- `RowDelta` (#789) with equality deletes (#809, #823): conflict if concurrent append added data files in covered partitions (serializable isolation)
- `Overwrite`: conflict if concurrent commit added files matching the overwrite predicate
- `RewriteFiles` (compaction): conflict if concurrent delete files target the files being rewritten
**Retry loop** — on `ErrCommitFailed`, reload metadata, validate conflicts, re-apply against new HEAD, retry with exponential backoff. The `commit.retry.num-retries` property is already in test fixtures but never read.
**Isolation levels** — `write.delete.isolation-level` property (`serializable` vs `snapshot`), matching Java semantics.

### Context
This is needed for any multi-writer workload — concurrent CDC consumers, parallel ETL jobs, or compaction running alongside ingest. Without this, concurrent RowDelta commits can produce incorrect results if equality delete files miss newly appended data.

mzzz-zzm:
Great progress on #911 and #912! I've been working on the "refresh-and-replay" piece
that #912 explicitly defers, and I ran into six bugs that will all fire once
`num-retries` is raised above zero. Documenting them here before opening a PR, since
#912 describes this as a follow-up.

---

### Context

I tested a `rebase()` implementation against AWS S3 Tables (a real managed Iceberg REST
catalog backed by S3). The test spins up two goroutines writing to the same table
concurrently and expects both to eventually succeed via the OCC retry loop. Six bugs
prevent that from working. Four are OCC retry correctness issues; two are data races
exposed by concurrent manifest writing.

None of these are visible today with the default `num-retries=0`, but they will all
[...]

PR #1022, opened 2026-05-06
feat(table): skip validator registration for no-op producers

hectar-glitches:
## Summary

This PR gates validator registration on a new `needsValidation() bool` method on the `producerImpl` interface:

- `fastAppendFiles` returns `false`: appends are commutative and need no conflict checks
- `mergeAppendFiles` inherits `false` via embedding
- `overwriteFiles` returns `true`: real conflict validation required
- `snapshotProducer.commit()` skips registration when `needsValidation()` is false

The no-op `validate()` methods are kept as belt-and-suspenders.

Behavior is locked with two unit tests asserting that fast-append and merge-append commits leave `txn.validators` empty.

Parent: #830

### DIFF
--- a/table/snapshot_producers.go
+++ b/table/snapshot_producers.go
@@ -54,6 +54,8 @@ type producerImpl interface {
 	// producers that are safe against concurrent appends (fast-append
 	// and merge-append).
 	validate(cc *conflictContext) error
+
+	needsValidation() bool
 }
 
 func newManifestFileName(num int, commit uuid.UUID) string {
@@ -110,6 +112,8 @@ func (fa *fastAppendFiles) deletedEntries(_ context.Context) ([]iceberg.Manifest
 func (fa *fastAppendFiles) validate(_ *conflictContext) error {
 	return nil
 }
+
+func (fa *fastAppendFiles) needsValidation() bool { return false }
 
 type overwriteFiles struct {
 	base *snapshotProducer
@@ -322,6 +326,8 @@ func (of *overwriteFiles) deletedEntries(ctx context.Context) ([]iceberg.Manifes
 
 	return finalResult, nil
 }
+
+func (of *overwriteFiles) needsValidation() bool { return true }
 
 type manifestMergeManager struct {
 	targetSizeBytes int
@@ -897,7 +903,7 @@ func (sp *snapshotProducer) commit(ctx context.Context) (_ []Update, _ []Require
 	// pay only the no-op-closure cost. Nil producerImpl is possible
 	// only in unit tests that exercise commit() directly on a bare
 	// snapshotProducer; guard to keep those tests green.
-	if impl := sp.producerImpl; impl != nil {
+	if impl := sp.producerImpl; impl != nil && impl.needsValidation() {
 		sp.txn.validators = append(sp.txn.validators, func(cc *conflictContext) error {
 			return impl.validate(cc)
 		})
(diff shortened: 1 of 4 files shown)


table/snapshot_producers.go
```diff
@@ -54,6 +54,8 @@ type producerImpl interface {
 	// producers that are safe against concurrent appends (fast-append
 	// and merge-append).
 	validate(cc *conflictContext) error
+
+	needsValidation() bool
```
laskoviymishka, line 62: Worth a one-line doc on the contract here? The neighbour `validate` has a 7-line block (lines 50-56), and `needsValidation` is the gate that decides whether `validate` even runs — defining "return false iff validate is unconditionally a no-op" pins the invariant at the declaration. Without it, someone could add real work to a producer's `validate()` later without flipping the predicate and silently bypass the check.

table/snapshot_producers.go
```diff
@@ -111,6 +113,8 @@ func (fa *fastAppendFiles) validate(_ *conflictContext) error {
 	return nil
 }
 
+func (fa *fastAppendFiles) needsValidation() bool { return false }
```
laskoviymishka, line 120: Tiny robustness ask while we're here: I'd consider adding an explicit `func (m *mergeAppendFiles) needsValidation() bool { return false }` next to the `mergeAppendFiles` declaration so the choice is local to that type rather than inherited via Go method promotion from the embedded `fastAppendFiles`. Right now nothing at the `mergeAppendFiles` site signals the policy, and a future producer copy-pasting the embed pattern would silently inherit `false` with no compiler nudge. wdyt?

table/snapshot_producers.go
```diff
@@ -897,7 +903,7 @@ func (sp *snapshotProducer) commit(ctx context.Context) (_ []Update, _ []Require
 	// pay only the no-op-closure cost. Nil producerImpl is possible
 	// only in unit tests that exercise commit() directly on a bare
 	// snapshotProducer; guard to keep those tests green.
-	if impl := sp.producerImpl; impl != nil {
+	if impl := sp.producerImpl; impl != nil && impl.needsValidation() {
```
laskoviymishka, line 911: Comment block just above is now stale — it still says fast/merge-append `pay only the no-op-closure cost`, but with the new gate they pay nothing. I'd freshen the rationale while we're here, e.g. "Producers that are commutative against concurrent appends (fast-append, merge-append) opt out via `needsValidation() == false` and skip registration entirely. The nil guard remains for unit tests that drive `commit()` on a bare `snapshotProducer`."

laskoviymishka (approved):
:shipit: 
Thanks for contribution!

Skipping the no-op closure for fast/merge append matches Java’s `BaseFastAppend` / `MergeAppend`, which also don’t register `validate()` on the append path.

No correctness or spec concerns from my side. 

A few nits before merging @hectar-glitches :

* The comment above the gated `if` in `commit()` looks a bit stale now — “pay only the no-op-closure cost” no longer quite describes the new behavior.
* A one-line doc comment on `needsValidation()` would make the interface contract clearer for future readers.
* `mergeAppendFiles.needsValidation` is inherited through Go method promotion. That works, but an explicit override could make the intent easier to see locally.
* If we keep relying on promotion, `TestMergeAppendFiles_InheritsFastAppendValidator` could also assert `require.False(t, ma.needsValidation())`.

Overall LGTM from my side.

hectar-glitches:
I'll get on those nit-picks

Issue #696, opened 2026-01-23
Enormous dependency tree — consider "drivers"?


brandur:
### Feature Request / Improvement

Given an empty project with Iceberg-go as its only dependency:

``` sh
cat main.go
```
```go
package main

import (
        _ "github.com/apache/iceberg-go"
        _ "github.com/apache/iceberg-go/catalog"
        _ "github.com/apache/iceberg-go/catalog/rest"
        _ "github.com/apache/iceberg-go/io"
        _ "github.com/apache/iceberg-go/table"
)

func main() {
}
```

```sh
head go.mod
```
```
module github.com/brandur/iceberg-go-test

go 1.25.6

require github.com/apache/iceberg-go v0.4.1-0.20260123180146-3ce3b4bb51c0

require (
        atomicgo.dev/cursor v0.2.0 // indirect
        atomicgo.dev/keyboard v0.2.9 // indirect
        atomicgo.dev/schedule v0.1.0 // indirect
```

Iceberg-go adds ~1s of compile overhead (on my commodity laptop). This might not be the end of the world, but you end up soaking it up on every test run compile and test run, which is fairly painful for any projects using Iceberg-go:

```
$ go test .
ok      github.com/brandur/iceberg-go-test      0.781s [no tests to run]
```

The project has about 130 dependencies. Manually inspecting the list, a big number of them are from the three major cloud providers (AWS, Azure, Google). Being able to jettison those would probably help a lot.

An approach that we took in one of our projects (River) is to use a "driver" system wherein you have sub-modules for major dependencies so users can pull them in piece by piece. e.g.

https://github.com/riverqueue/river/tree/master/riverdriver

It's far from perfect as Go's multi-module handling leaves a lot to be desired, but it works reasonably well, and a similar approach would probably help Iceberg-go in the longer run. Thoughts?

alessandro-nori:
Hello @brandur ! Thanks for looking into it, I'm also very interested in this feature.
I'm also available to work on this in case you don't have enough bandwidth

laskoviymishka:
This got a bit staled, with #980 landing need to finish go-mod split.

alessandro-nori:
I haven't started looking into splitting the go.mod but I should have some time to work on it this week

alessandro-nori:
Before continuing with the split of go modules, I identified two things that need to be addressed first:
- in-memory IO is used in multiple tests so I think we should provide an implementation that does not depend on gocloud
[...]

PR #1025, opened 2026-05-06
feat(io): add native in-memory FS

alessandro-nori:
## Summary

- Adds `MemFS`: an in-memory `IO` implementation with no external dependencies
- `MemFS` implements `WriteFileIO` and `ListableIO`
- Moves mem tests from `io/gocloud` to `io/` and removes the now-unnecessary gocloud import from `table/snapshot_producers_test.go`

Part of #696.


zeroshade (approved):
LGTM, this is definitely a nice way to remove a dependency for a small bit of complexity

PR #1030, opened 2026-05-06
fix(manifest): allow ManifestListWriter to reference lower-version manifests

### LINKED ISSUE #1029: bug: ManifestListWriter rejects lower-version manifests, blocking v1->v2 table upgrades
### Apache Iceberg version

main (development)

### Problem

`ManifestListWriter.AddManifests` (`manifest.go:1471-1475`) requires every input manifest to match the writer's version exactly:

```go
case 2, 3:
    for _, file := range files {
        if file.Version() != m.version {
            return fmt.Errorf("%w: ManifestListWriter only supports version %d manifest files", ErrInvalidArgument, m.version)
        }
        ...
```

The Iceberg spec, however, allows a v2 manifest list to reference v1 manifest files — the format version of an individual manifest is independent of the manifest list's version, so that a v1 table can be upgraded to v2 without rewriting historical manifests. Java's [`ManifestListWriter`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestListWriter.java) supports this; `iceberg-go` does not.

### Symptom

For tables that originated as v1, any commit after the table metadata is upgraded to v2 fails with:

```
invalid argument: ManifestListWriter only supports version 2 manifest files
```

The failure is permanent. Setting `format-version=2` on a v1 table is a metadata-only change — existing v1 manifest files on disk are never rewritten and continue to surface in `snapshotProducer.manifests()` (`table/snapshot_producers.go:713-721`, where `sp.existingManifests()` is called) on every subsequent commit. The condition affects any consumer performing a v1→v2 upgrade, whether via `Transaction.UpgradeFormatVersion` or via a metadata-only upgrade performed by another engine (Spark, Hive) before `iceberg-go` writes to the table.

### Root cause

The strict version gate predates the spec's mixed-version inheritance rules. The in-memory `manifestFile` produced from a v1 input — via `manifestFileV1.toFile()` (`manifest.go:181-237`) or via `NewManifestFile(1, ...)` — already carries the inheritance values mandated by the spec for v1 manifests inside a v2+ list:

- `Content = ManifestContentData`
- `SeqNumber = initialSequenceNumber = 0`
- `MinSeqNumber = initialSequenceNumber = 0`

So it can be encoded directly against the v2 (or v3) entry schema. The only thing standing in the way is the version check itself.

A v3 manifest inside a v2 list, however, is *not* representable: the v2 entry schema has no `first_row_id` column, so accepting it would silently drop data. That direction must remain rejected.

### Fix

Relax the gate to `file.Version() > m.version`:

- v2 manifest list: accepts v1 and v2 manifests
- v3 manifest list: accepts v1, v2, and v3 manifests
- v2 list still rejects v3 manifests; v1 list still rejects v2/v3 manifests

No conversion code is required. The existing `manifestFile` representation is already version-agnostic, and the writer's avro encoder picks the schema from `m.version`, so v1 inputs are encoded against the v2/v3 schema with the correct inherited values. The existing `first_row_id` assignment at `manifest.go:1480-1486` (data manifests with `FirstRowIDValue == nil`) already covers v1 manifests passed into v3 lists.

### Relation to existing issues

- #416 / #826: `format-version` `strconv.Atoi("")` fix at the manifest list and manifest file layers — separate symptom from the same v1→v2 upgrade flow; already resolved.
- #827: added `Transaction.UpgradeFormatVersion` — the upgrade entry point exercised by the failure path described above.

Jeffail:
## Problem

`ManifestListWriter.AddManifests` requires every input manifest file's version to match the writer's version exactly, which contradicts the Iceberg spec: a v2 manifest list must be able to reference v1 manifest files so that v1 tables can be upgraded without rewriting historical manifests. Java's [`ManifestListWriter`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestListWriter.java) handles this; `iceberg-go` does not.

The practical symptom for downstream users is that any commit to a table whose metadata has been upgraded from v1 to v2 — whether via `Transaction.UpgradeFormatVersion` or out-of-band by another engine — fails with `invalid argument: ManifestListWriter only supports version 2 manifest files`. The v1 manifest files written before the upgrade are never rewritten and continue to surface via `snapshotProducer.existingManifests()` on every subsequent commit, so the failure is permanent.

## Fix

`manifest.go:1410-1414`: replace the `file.Version() != m.version` exact-match gate with `file.Version() > m.version`. Newer-than-writer inputs are still rejected — the v2 entry schema has no place for v3 fields such as `first_row_id`, and accepting them would silently drop data. Lower-than-writer inputs are accepted because the in-memory `manifestFile` produced from a v1 input (via `manifestFileV1.toFile()` or `NewManifestFile(1, ...)`) already carries the spec's inheritance values — `Content = data` and `SeqNumber = MinSeqNumber = 0` — so it can be encoded directly against the v2/v3 entry schema. The existing `first_row_id` assignment for data manifests with `FirstRowIDValue == nil` covers v1 inputs in v3 lists without further changes.

## Testing

- `TestV2ManifestListAcceptsV1Manifests` — round-trips a v1 manifest through a v2 manifest list and asserts that the decoded entry has `content = data` and `sequence_number = min_sequence_number = 0` per spec inheritance.
- `TestV3ManifestListAcceptsV1AndV2Manifests` — writes a v1 data manifest and a v2 delete manifest into a v3 list; asserts inheritance for the v1 entry, asserts that `first_row_id` is assigned to the v1 data manifest and is left unset on the v2 delete manifest (data-only assignment per the existing v3 writer rules).
- `TestV2ManifestListRejectsV3Manifests` — confirms the downgrade direction is still blocked.
- `TestWriteManifestListClosesWriterOnError` — existing test updated to drive its `AddManifests` failure path through a v3-in-v2 input (still rejected) instead of v1-in-v2.

Confirmed failing on `main` before the fix and passing after. `go test ./...` passes.

Fixes #1029

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1470,8 +1470,18 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 
 	case 2, 3:
 		for _, file := range files {
-			if file.Version() != m.version {
-				return fmt.Errorf("%w: ManifestListWriter only supports version %d manifest files", ErrInvalidArgument, m.version)
+			// Per the Iceberg spec a v2 manifest list may reference v1 manifest
+			// files (and a v3 list may reference v1 or v2 manifests) so that a
+			// table can be upgraded without rewriting historical manifests. The
+			// in-memory ManifestFile produced for v1 inputs already carries the
+			// inheritance values mandated by the spec — Content=data and
+			// SeqNumber/MinSeqNumber=0 — so it can be encoded directly against
+			// the v2/v3 entry schema. Newer-than-writer inputs are rejected
+			// because the v2 schema cannot represent v3 fields such as
+			// first_row_id.
+			if file.Version() > m.version {
+				return fmt.Errorf("%w: manifest list v%d cannot reference v%d manifest files",
+					ErrInvalidArgument, m.version, file.Version())
 			}
 
 			wrapped := *(file.(*manifestFile))
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -1970,6 +1970,9 @@ func (w *limitedWriter) Write(p []byte) (int, error) {
 }
 
 func (m *ManifestTestSuite) TestWriteManifestListClosesWriterOnError() {
+	// A v2 manifest list cannot reference v3 manifests because the v2 entry
+	// schema has no first_row_id column; this gives us a deterministic
+	// AddManifests failure to assert that Close still flushes.
 	seqNum := int64(7)
 	var header bytes.Buffer
 	writer, err := NewManifestListWriterV2(&header, snapshotID, seqNum, nil)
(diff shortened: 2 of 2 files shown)


zeroshade:
Can you rebase with main? That'll fix the failing CI

zeroshade (approved):

PR #867, opened 2026-04-09
fix: use OpReplace instead of OpOverwrite in ReplaceDataFiles and ReplaceFiles

Bahtya:
## Summary

Fixes #841 (parent #832)

Changes `ReplaceDataFiles`, `ReplaceDataFilesWithDataFiles`, and `ReplaceFiles` to use `OpReplace` instead of `OpOverwrite` when creating snapshot updates.

## Problem

Per the Iceberg spec, `REPLACE` is the correct operation when data content is equivalent but reorganized into different files (e.g., compaction). The three replace methods were unconditionally using `OpOverwrite` despite a TODO comment acknowledging this was incorrect.

## Changes

- `table/transaction.go`: Changed `OpOverwrite` → `OpReplace` in three locations:
  - `ReplaceDataFiles` (line ~418)
  - `ReplaceDataFilesWithDataFiles` (line ~713)
  - `ReplaceFiles` (line ~826)
- Removed the TODO comment at `ReplaceDataFiles` that acknowledged the incorrect operation type
- `table/replace_files_test.go`: Updated `TestReplaceFiles_DataAndDeleteFiles` to assert `OpReplace`

## Testing

All existing tests pass:
```
=== RUN   TestReplaceFiles_DataAndDeleteFiles
--- PASS
=== RUN   TestReplaceFiles_DelegatesToReplaceDataFilesWhenNoDeleteFiles
--- PASS
=== RUN   TestReplaceFiles_ValidationErrors
--- PASS
```

Issue #841, opened 2026-04-03
fix: ReplaceDataFiles should use OpReplace instead of OpOverwrite


laskoviymishka:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Parent: #832

`ReplaceDataFiles` at `transaction.go:349` has a TODO noting the operation should be `OpReplace` (data is reorganized, not changed) but unconditionally uses `OpOverwrite` because there is no validation. Per the Iceberg spec, `REPLACE` means the data content is equivalent — just reorganized into different files. This is the correct operation for compaction.

tanmayrauth:
@laskoviymishka  @kevinjqliu  Can I work on this issue?

laskoviymishka:
There is already a pr for that

tanmayrauth:
@laskoviymishka  can I pick this issue and resolve the review comments?

laskoviymishka:
I think it's better to give some time to @Bahtya to resolve it and have a chance to land this PR on his own.
If you looking to stuff to pickup, i think https://github.com/apache/iceberg-go/issues/831 this one could be a good candidate.

If you are willing to contribute more - plz reach me in slack, i don't file all issue here, but there is a plenty of work left in general

tanmayrauth:
Sure, I will ping you on slack.

tanmayrauth:
@zeroshade I think we can close this one

PR #1033, opened 2026-05-06
feat(table): add RewriteFiles snapshot-op builder

twuebi:
Decompose the atomic rewrite-snapshot of `Transaction.RewriteDataFiles` which assumes in-process compaction to enable out-of-process distributed compaction, by exposing a builder similar to what iceberg-java has with `Table.newRewrite()`.

Mirrors iceberg-java's RewriteFiles (Table.newRewrite()), sitting beside RowDelta. Suppresses the overwrite producer's default validator and queues a rewrite-specific conflict validator internally, so RewriteDataFiles and external callers (distributed compaction coordinators) commit the same shape through one indivisible operation.

Transaction.NewRewrite(snapshotProps) returns *RewriteFiles with DeleteFile / AddDataFile / Apply / Commit. ExecuteCompactionGroup is the worker-side read+write step, returning a plain-data CompactionGroupResult.

RewriteDataFiles drives the builder internally. Partial-progress mode now commits per group inside the loop, fixing a latent bug where a mid-loop failure left staged groups uncovered by the rewrite validator.

Tag rewrite snapshots OpReplace instead of OpOverwrite (matching iceberg-java's BaseRewriteFiles.operation()) — gated on cfg.rewriteSemantics so generic ReplaceFiles callers, where row content can change, keep OpOverwrite. This also addresses #841 and should address the comments made in #867 since only the unexported `cfg.rewriteSemantics` makes OpOverwrite -> OpReplace and `RewriteFiles.Commit()` is the only caller of `withRewriteSemantics()`.

Black-box coverage: equivalence with RewriteDataFiles, safe pos-delete expunge, concurrent eq-delete rejection under refresh-and-replay.

### DIFF
--- a/table/rewrite_data_files.go
+++ b/table/rewrite_data_files.go
@@ -20,6 +20,7 @@ package table
 import (
 	"context"
 	"fmt"
+	"maps"
 
 	"github.com/apache/iceberg-go"
 )
@@ -178,6 +179,9 @@ func WithCompactionTargetFileSize(size int64) CompactionGroupOption {
 // WithCompactionScanConcurrency sets the scan concurrency used when
 // reading the group's tasks. Forwarded to [Table.Scan] as
 // [WitMaxConcurrency]. Zero (the default) means runtime.GOMAXPROCS.
+//
+// TODO: the [WitMaxConcurrency] link enshrines a pre-existing typo
+// (missing `h`). Update this reference when that symbol is renamed.
 func WithCompactionScanConcurrency(n int) CompactionGroupOption {
 	return func(c *compactionGroupConfig) {
 		c.scanConcurrency = n
@@ -325,12 +329,21 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 }
 
 // rewriteDataFilesPartial stages each group as its own rewrite
-// snapshot via a fresh [RewriteFiles] builder. Each builder commits
-// independently inside the loop, so a mid-loop write failure leaves
-// already-staged groups in the transaction and the catalog still
-// receives them at [Transaction.Commit] time.
+// snapshot via [Transaction.ReplaceFiles] directly. Per-group staging
+// lets a mid-loop write failure leave already-staged groups on the
+// transaction; the catalog still receives them at
+// [Transaction.Commit] time.
+//
+// Validator registration is coalesced: a single [rewriteValidator]
+// covering every rewritten path across all groups is registered once,
+// after the loop, instead of one per group. The transaction's
+// validator list otherwise grows linearly with the group count, and
+// each entry independently walks the concurrent-snapshot set on
+// refresh-replay — the union walk subsumes them.
 func (t *Transaction) rewriteDataFilesPartial(ctx context.Context, groups []CompactionTaskGroup, opts RewriteDataFilesOptions) (*RewriteResult, error) {
 	result := &RewriteResult{}
+	props := maps.Clone(opts.SnapshotProps)
+	var allRewritten []string
 
 	for _, group := range groups {
 		if err := ctx.Err(); err != nil {
@@ -350,11 +363,19 @@ func (t *Transaction) rewriteDataFilesPartial(ctx context.Context, groups []Comp
 			continue
 		}
 
-		if err := t.NewRewrite(opts.SnapshotProps).Apply(gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes).Commit(ctx); err != nil {
+		if err := t.ReplaceFiles(ctx, gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes,
+			props, withRewriteSemantics()); err != nil {
 			return result, fmt.Errorf("commit compaction group %q: %w", group.PartitionKey, err)
 		}
 
+		for _, f := range gr.OldDataFiles {
+			allRewritten = append(allRewritten, f.FilePath())
+		}
 		accumulateGroupMetrics(result, gr)
+	}
+
+	if len(allRewritten) > 0 {
+		t.addValidator(rewriteValidator(allRewritten))
 	}
 
 	return result, nil
@@ -396,6 +417,20 @@ func rewriteValidator(rewrittenPaths []string) conflictValidatorFunc {
 // Equality deletes are decided by [compaction.DecideDeadEqualityDeletes]
 // (which needs partition-wide visibility, not just the task scope).
 // Deletion vectors will be handled when DV read support lands.
+//
+// Caller contract: every data file referenced by a returned pos-delete
+// must be in the caller's rewrite set across the entire commit.
+// This function only sees one group's tasks, but a pos-delete file
+// can reference data files across multiple groups (the planner
+// bin-packs within a partition via [compaction.Config.PlanCompaction]
+// and skips files via MinInputFiles). If a pos-delete is reported safe
+// by one group but references a still-live data file in another group
+// — or a file the planner skipped — committing only this group's
+// rewrite would orphan the still-live data file's deletes. Coordinators
+// that aggregate multiple groups into one rewrite snapshot are
+// responsible for re-checking against the full set of rewritten paths,
+// or for moving this computation leader-side once worker outputs have
+// aggregated.
 //
 // [ExecuteCompactionGroup] calls this internally to populate
 // [CompactionGroupResult.SafePosDeletes]. It is kept exported for
--- a/table/rewrite_data_files.go
+++ b/table/rewrite_data_files.go
@@ -113,18 +113,24 @@ type CompactionGroupResult struct {
 // RewriteDataFilesOptions bundles the per-rewrite knobs for
 // [Transaction.RewriteDataFiles].
 type RewriteDataFilesOptions struct {
-	// PartialProgress, when true, stages each group via its own
-	// rewrite snapshot inside the loop so work survives a mid-loop
-	// write failure. When false (the default), all groups are
-	// committed in a single atomic rewrite snapshot.
+	// PartialProgress, when true, stages each group as its own
+	// rewrite snapshot inside the loop so a mid-loop write failure
+	// leaves the already-completed groups staged on this transaction
+	// (the in-memory transaction can be discarded by group rather
+	// than wholesale). When false (the default), every group lands in
+	// a single atomic rewrite snapshot.
 	//
-	// In both modes the final catalog commit happens once at
-	// [Transaction.Commit] time. True per-group durability (matching
-	// Java's behavior) requires committing separate transactions per
-	// group, which is left to the caller.
+	// In both modes the catalog commit happens once at
+	// [Transaction.Commit] time, so a process crash mid-loop loses
+	// every staged group regardless of this flag. Callers who need
+	// true per-group catalog durability (matching Java's behavior)
+	// should drive [Transaction.NewRewrite] themselves and commit a
+	// fresh transaction per group.
 	PartialProgress bool
 
 	// SnapshotProps are added to the rewrite snapshot's summary.
+	// In partial-progress mode the same properties land on every
+	// per-group snapshot rather than being summed or split.
 	SnapshotProps iceberg.Properties
 
 	// ExtraDeleteFilesToRemove are delete files (typically equality
--- a/table/rewrite_data_files.go
+++ b/table/rewrite_data_files.go
@@ -137,6 +137,40 @@ type RewriteDataFilesOptions struct {
 	// applied are removed automatically and do NOT need to be passed
 	// in here.
 	ExtraDeleteFilesToRemove []iceberg.DataFile
+
+	// GroupOptions are forwarded to every [ExecuteCompactionGroup]
+	// call to tune the per-group read+write pipeline (target file
+	// size, scan concurrency). See the With* helpers returning
+	// [CompactionGroupOption].
+	GroupOptions []CompactionGroupOption
+}
+
+// CompactionGroupOption configures a single [ExecuteCompactionGroup]
+// call. Use the With* helpers to construct values.
+type CompactionGroupOption func(*compactionGroupConfig)
+
+type compactionGroupConfig struct {
+	targetFileSize  int64
+	scanConcurrency int
+}
+
+// WithCompactionTargetFileSize sets the size target for output files
+// written by [ExecuteCompactionGroup]. Forwarded to [WriteRecords] as
+// [WithTargetFileSize]. Zero (the default) means inherit the table's
+// `write.target-file-size-bytes` property.
+func WithCompactionTargetFileSize(size int64) CompactionGroupOption {
+	return func(c *compactionGroupConfig) {
+		c.targetFileSize = size
+	}
+}
+
+// WithCompactionScanConcurrency sets the scan concurrency used when
+// reading the group's tasks. Forwarded to [Table.Scan] as
+// [WitMaxConcurrency]. Zero (the default) means runtime.GOMAXPROCS.
+func WithCompactionScanConcurrency(n int) CompactionGroupOption {
+	return func(c *compactionGroupConfig) {
+		c.scanConcurrency = n
+	}
 }
 
 // RewriteDataFiles compacts the given groups by reading data with
@@ -179,7 +213,7 @@ func (t *Transaction) RewriteDataFiles(ctx context.Context, groups []CompactionT
 			continue
 		}
 
-		gr, err := ExecuteCompactionGroup(ctx, t.tbl, group)
+		gr, err := ExecuteCompactionGroup(ctx, t.tbl, group, opts.GroupOptions...)
 		if err != nil {
 			return result, err
 		}
@@ -221,24 +255,42 @@ func (t *Transaction) RewriteDataFiles(ctx context.Context, groups []CompactionT
 //
 // In-process callers should prefer [Transaction.RewriteDataFiles],
 // which drives this and the commit step in one call.
-func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTaskGroup) (CompactionGroupResult, error) {
+//
+// Tunables are exposed via [CompactionGroupOption]. The clustered
+// write path is always used (a compaction group is single-partition
+// by construction so its read stream is trivially clustered).
+func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTaskGroup, opts ...CompactionGroupOption) (CompactionGroupResult, error) {
 	if len(group.Tasks) == 0 {
 		return CompactionGroupResult{PartitionKey: group.PartitionKey}, nil
 	}
 
-	scan := tbl.Scan()
-	arrowSchema, records, err := scan.ReadTasks(ctx, group.Tasks)
+	cfg := compactionGroupConfig{}
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+
+	var scanOpts []ScanOption
+	if cfg.scanConcurrency > 0 {
+		scanOpts = append(scanOpts, WitMaxConcurrency(cfg.scanConcurrency))
+	}
+
+	arrowSchema, records, err := tbl.Scan(scanOpts...).ReadTasks(ctx, group.Tasks)
 	if err != nil {
 		return CompactionGroupResult{}, fmt.Errorf("read tasks for compaction group %q: %w", group.PartitionKey, err)
 	}
 
 	// Each compaction group is single-partition by construction, so the
 	// read stream is trivially clustered and we can use the clustered writer.
+	writeOpts := []WriteRecordOption{WithClusteredWrite()}
+	if cfg.targetFileSize > 0 {
+		writeOpts = append(writeOpts, WithTargetFileSize(cfg.targetFileSize))
+	}
+
 	var (
 		newFiles   []iceberg.DataFile
 		bytesAfter int64
 	)
-	for df, err := range WriteRecords(ctx, tbl, arrowSchema, records, WithClusteredWrite()) {
+	for df, err := range WriteRecords(ctx, tbl, arrowSchema, records, writeOpts...) {
 		if err != nil {
 			return CompactionGroupResult{}, fmt.Errorf("write compacted files for group %q: %w", group.PartitionKey, err)
 		}
@@ -278,7 +330,7 @@ func (t *Transaction) rewriteDataFilesPartial(ctx context.Context, groups []Comp
 			continue
 		}
 
-		gr, err := ExecuteCompactionGroup(ctx, t.tbl, group)
+		gr, err := ExecuteCompactionGroup(ctx, t.tbl, group, opts.GroupOptions...)
 		if err != nil {
 			return result, err
 		}
--- a/table/rewrite_data_files.go
+++ b/table/rewrite_data_files.go
@@ -58,7 +58,8 @@ type RewriteResult struct {
 // import between table and table/compaction.
 //
 // Use [compaction.Config.PlanCompaction] to produce groups, then convert
-// [compaction.Group] → [CompactionTaskGroup] to call [Transaction.RewriteDataFiles].
+// [compaction.Group] → [CompactionTaskGroup] to call
+// [Transaction.RewriteDataFiles] or [ExecuteCompactionGroup].
 type CompactionTaskGroup struct {
 	// PartitionKey is an opaque grouping key for display/logging.
 	PartitionKey string
@@ -70,16 +71,55 @@ type CompactionTaskGroup struct {
 	TotalSizeBytes int64
 }
 
+// CompactionGroupResult is the per-group output of a compaction
+// worker: the new files written, the old files being replaced, and
+// the position delete files safe to expunge in the rewrite snapshot.
+//
+// A distributed coordinator aggregates results from N workers and
+// applies them to a [RewriteFiles] builder via [RewriteFiles.Apply]
+// to commit a single atomic snapshot. Each field is plain data
+// ([]iceberg.DataFile values plus scalars) — callers serialize the
+// contained DataFiles across process boundaries themselves; the
+// typical pattern is to have the worker write a manifest containing
+// the new files and ship the manifest path to the coordinator, which
+// re-reads it.
+type CompactionGroupResult struct {
+	// PartitionKey mirrors [CompactionTaskGroup.PartitionKey] for
+	// display/logging on the coordinator.
+	PartitionKey string
+
+	// OldDataFiles are the data files this group replaces.
+	OldDataFiles []iceberg.DataFile
+
+	// NewDataFiles are the consolidated outputs the worker wrote.
+	NewDataFiles []iceberg.DataFile
+
+	// SafePosDeletes are position-delete files referenced by tasks in
+	// this group whose target data file is being rewritten, computed
+	// via [CollectSafePositionDeletes]. They are safe to expunge in
+	// the rewrite snapshot.
+	SafePosDeletes []iceberg.DataFile
+
+	// BytesBefore is [CompactionTaskGroup.TotalSizeBytes] passed
+	// through, recorded so the coordinator can roll up metrics
+	// without re-reading the plan.
+	BytesBefore int64
+
+	// BytesAfter is the sum of [iceberg.DataFile.FileSizeBytes] across
+	// NewDataFiles.
+	BytesAfter int64
+}
+
 // RewriteDataFilesOptions bundles the per-rewrite knobs for
-// Transaction.RewriteDataFiles.
+// [Transaction.RewriteDataFiles].
 type RewriteDataFilesOptions struct {
-	// PartialProgress, when true, stages each group via ReplaceFiles
-	// inside the loop so work survives a mid-loop write failure. When
-	// false (the default), all groups are committed in a single atomic
-	// snapshot.
+	// PartialProgress, when true, stages each group via its own
+	// rewrite snapshot inside the loop so work survives a mid-loop
+	// write failure. When false (the default), all groups are
+	// committed in a single atomic rewrite snapshot.
 	//
 	// In both modes the final catalog commit happens once at
-	// Transaction.Commit() time. True per-group durability (matching
+	// [Transaction.Commit] time. True per-group durability (matching
 	// Java's behavior) requires committing separate transactions per
 	// group, which is left to the caller.
 	PartialProgress bool
@@ -89,9 +129,8 @@ type RewriteDataFilesOptions struct {
 
 	// ExtraDeleteFilesToRemove are delete files (typically equality
 	// deletes that are dead after the rewrite) that the caller wants
-	// expunged in the same snapshot as the rewrite. The executor
-	// passes them through to ReplaceFiles unchanged. Honored only
-	// when PartialProgress is false.
+	// expunged in the same snapshot as the rewrite. Honored only when
+	// PartialProgress is false.
 	//
 	// Use [compaction.CollectDeadEqualityDeletes] to compute this list
 	// from the current snapshot. Position delete files that are fully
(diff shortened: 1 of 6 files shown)


twuebi:
@laskoviymishka this touches compaction and exposes an interface which should enable distributed compaction implementations

table/rewrite_files.go
```diff
@@ -0,0 +1,168 @@
...
+
+// Commit stages the rewrite snapshot on the underlying transaction.
+// The catalog commit happens once, later, at [Transaction.Commit] time.
+//
+// Returns an error if the builder has no file changes, if any
+// [RewriteFiles.AddDataFile] input is not a data file, or if the underlying
+// [Transaction.ReplaceFiles] call fails.
+func (r *RewriteFiles) Commit(ctx context.Context) error {
+	if len(r.dataFilesToDelete) == 0 && len(r.dataFilesToAdd) == 0 && len(r.deleteFilesToRemove) == 0 {
+		return errors.New("rewrite must have at least one file change")
```
laskoviymishka: I'd wrap this with a sentinel rather than `errors.New` so callers can branch on it. `ErrInvalidOperation` already exists in this package and is what `ReplaceFiles` returns for the equivalent shape (`cannot replace files in a table without an existing snapshot`). A bare string here breaks `errors.Is` for anyone driving the builder programmatically.

```go
return fmt.Errorf("%w: rewrite must have at least one file change", ErrInvalidOperation)
```

table/rewrite_files.go
```diff
@@ -0,0 +1,168 @@
...
+	return r
+}
+
+// Commit stages the rewrite snapshot on the underlying transaction.
+// The catalog commit happens once, later, at [Transaction.Commit] time.
+//
+// Returns an error if the builder has no file changes, if any
+// [RewriteFiles.AddDataFile] input is not a data file, or if the underlying
+// [Transaction.ReplaceFiles] call fails.
+func (r *RewriteFiles) Commit(ctx context.Context) error {
```
laskoviymishka, line 205: I think `Commit` should be single-shot. As written, calling it twice re-appends a `rewriteValidator` to `r.txn.validators` and re-stages a `ReplaceFiles` against the (now mutated) transaction — Java's `BaseRewriteFiles` rejects this. A small `committed bool` guard plus an explicit error keeps the failure mode clean. wdyt?

table/rewrite_files.go
```diff
@@ -0,0 +1,168 @@
...
+		r.deleteFilesToRemove = append(r.deleteFilesToRemove, df)
+	}
+
+	return r
+}
+
+// AddDataFile queues a new data file. Adding delete files is not yet
+// supported by the underlying snapshot machinery; passing a
+// pos/eq-delete here is reported at [RewriteFiles.Commit].
+func (r *RewriteFiles) AddDataFile(df iceberg.DataFile) *RewriteFiles {
```
laskoviymishka, line 129: The validation deferral here interacts awkwardly with `Apply`: a worker that mistakenly puts a delete file into `adds` gets it queued silently, and Commit reports `index N` — but N is opaque to a caller that fed the builder via `Apply` (they never saw the queue order). I'd either reject at insertion time (return `*RewriteFiles` with a stashed error, drained at Commit), or at minimum log the file path prominently. The current message has the path but the index framing implies the caller chose it.

While we're here — content-type routing in `DeleteFile` is permissive (anything non-`EntryContentData` lands in `deleteFilesToRemove`). For symmetry I'd at least assert it's pos/eq deletes there, otherwise a future content type silently slips through.

table/rewrite_files.go
```diff
@@ -0,0 +1,168 @@
...
+// deletion vectors) is not yet supported; [RewriteFiles.AddDataFile]
+// rejects pos/eq-delete inputs at commit time. Add the support to
+// the underlying [Transaction.ReplaceFiles] before lifting that
+// restriction.
+type RewriteFiles struct {
+	txn                 *Transaction
+	dataFilesToDelete   []iceberg.DataFile
+	dataFilesToAdd      []iceberg.DataFile
+	deleteFilesToRemove []iceberg.DataFile
+	snapshotProps       iceberg.Properties
```
laskoviymishka, line 71: `snapshotProps` is captured by reference. If a caller mutates the map between `NewRewrite` and `Commit` (or across multiple builders sharing the same map, as in `rewriteDataFilesPartial`), the staged snapshot summary changes. A `maps.Clone` here, or a doc note that the map must not be mutated after `NewRewrite` returns, would close the foot-gun.

table/rewrite_files.go
```diff
@@ -0,0 +1,168 @@
...
+// Compared to a raw [Transaction.ReplaceFiles] call, the builder
+// owns the rewrite-specific isolation contract internally:
+//
+//   - The overwrite producer's default isolation validator is suppressed
+//     (concurrent appends into rewritten partitions are allowed; this
+//     is the defining behavior of a rewrite).
+//   - A rewrite-specific conflict validator is registered so concurrent
+//     pos/eq-delete files targeting any rewritten data file are
+//     rejected pre-flight at [Transaction.Commit] time. Pos-delete
+//     conflict detection requires v3 manifests (which carry the
```
laskoviymishka: Minor accuracy on the V3 claim. `referenced_data_file` (field id 143) was added in V2 and is *optional* there; it became *required* in V3 only for deletion-vector deletes. So the pos-delete validator branch fires whenever the writer populated the column, regardless of format version — it just happens that V2 pos-delete writers commonly leave it empty, which is the real reason the conservative V2 fallback exists. I'd rephrase as "requires that the concurrent pos-delete writer populated the `referenced_data_file` column (always set under V3 deletion vectors; often unset in V2 pos-deletes)" so a reader doesn't conclude V2 + populated `referenced_data_file` is unsupported.

table/transaction.go
```diff
@@ -765,8 +768,13 @@ func (t *Transaction) ReplaceDataFilesWithDataFiles(ctx context.Context, filesTo
 		}
 	}
 
+	op := OpOverwrite
```
laskoviymishka, line 783: Tagging the snapshot `replace` is the right move for parity with Java's `BaseRewriteFiles.operation()`, but I'd want to check that the overwrite producer's summary writer still emits the right keys under this tag. Java's `RewriteFiles` snapshots carry `added-data-files`, `deleted-data-files`, `removed-position-delete-files` (and the byte counterparts) — if our `mergeOverwrite` path keys any summary entry on `OpOverwrite` specifically, the flip to `OpReplace` will silently drop it.

Worth adding an assertion in one of the rewrite tests against `snap.Summary.Properties` — at minimum that `added-data-files`, `deleted-data-files`, `total-records`, `total-data-files` are populated. PyIceberg readers parse these; a missing field on `OpReplace` would be a cross-client interop regression.

table/rewrite_data_files.go
```diff
@@ -70,16 +71,55 @@ type CompactionTaskGroup struct {
...
 // RewriteDataFilesOptions bundles the per-rewrite knobs for
-// Transaction.RewriteDataFiles.
+// [Transaction.RewriteDataFiles].
 type RewriteDataFilesOptions struct {
-	// PartialProgress, when true, stages each group via ReplaceFiles
-	// inside the loop so work survives a mid-loop write failure. When
-	// false (the default), all groups are committed in a single atomic
-	// snapshot.
+	// PartialProgress, when true, stages each group via its own
+	// rewrite snapshot inside the loop so work survives a mid-loop
```
laskoviymishka: "so work survives a mid-loop write failure" reads as if there's catalog durability per group, but the catalog commit is still single, at `Transaction.Commit`. A process crash mid-loop loses everything just like the non-partial path. The disclaimer four lines down ("True per-group durability ... requires committing separate transactions per group") effectively contradicts the lead.

With the new `RewriteFiles` builder, the per-group-transaction pattern is now expressible by callers — I'd update the doc to say partial mode "stages each group as its own snapshot in the transaction so the in-memory transaction can be discarded by group rather than wholesale, and point at the `NewRewrite` builder for callers who want true per-group catalog durability".

table/rewrite_data_files.go
```diff
@@ -142,91 +213,151 @@ func (t *Transaction) RewriteDataFiles(ctx context.Context, groups []CompactionT
...
-		if len(opts.ExtraDeleteFilesToRemove) > 0 {
-			allOldDeletes = append(allOldDeletes, opts.ExtraDeleteFilesToRemove...)
-			result.RemovedEqualityDeleteFiles += len(opts.ExtraDeleteFilesToRemove)
+		if len(gr.OldDataFiles) == 0 && len(gr.NewDataFiles) == 0 {
+			continue
 		}
 
-		if err := t.ReplaceFiles(ctx, allOldData, allNewData, allOldDeletes, opts.SnapshotProps, withRewriteSemantics()); err != nil {
-			return result, fmt.Errorf("commit compaction: %w", err)
+		if err := t.NewRewrite(opts.SnapshotProps).Apply(gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes).Commit(ctx); err != nil {
```
laskoviymishka: Each loop iteration constructs `t.NewRewrite(opts.SnapshotProps)` with the same `opts.SnapshotProps` map. So whatever caller-supplied properties are in there land on every per-group snapshot summary. Most use cases (free-form properties like an actor or dag-run id) are fine with this, but if the caller pre-populates anything that's meant to be unique per snapshot they'll get duplicates. Worth a one-line doc note on `RewriteDataFilesOptions.SnapshotProps` that in partial mode it's applied to every snapshot, not summed.

table/rewrite_files_test.go
```diff
@@ -0,0 +1,442 @@
...
+	require.Error(t, err, "ReplaceFiles fails because the dummy files are not in the empty table")
+	// Any of these messages confirms we got past the builder's own
+	// validation and into ReplaceFiles' membership checks — i.e., the
+	// pos-delete was routed to deleteFilesToRemove and the data file
+	// to dataFilesToDelete.
+	msg := err.Error()
+	assert.True(t,
+		strings.Contains(msg, "do not belong to the table") ||
+			strings.Contains(msg, "without an existing snapshot"),
+		"got %q — expected the underlying ReplaceFiles membership/snapshot error", msg)
```
laskoviymishka: This assertion would still pass if `DeleteFile` mis-routed both files to `dataFilesToDelete` (the resulting `ReplaceFiles` would also error with `do not belong to the table`). The substring check proves we got into `ReplaceFiles`, not that the pos-delete went into the right slice.

A tighter shape: stage two builders, one with the pos-delete and one with the data file, each with a benign add. Each should produce a different error class — `do not belong to the table` (data slice, membership check) vs. the delete-files-not-in-table message. wdyt?

table/rewrite_files_test.go
```diff
@@ -0,0 +1,442 @@
...
+	require.NoError(t, rd.Commit(t.Context()))
+	_, err = peerTxn.Commit(t.Context())
+	require.NoError(t, err, "peer commit advances the catalog so the leader's first attempt fails")
+
+	beforeLeader := cat.attempts.Load()
+	_, err = leaderTxn.Commit(t.Context())
+	require.Error(t, err)
+	assert.ErrorIs(t, err, table.ErrConflictingDeleteFiles,
+		"refresh-and-replay must detect the concurrent equality-delete during a rewrite")
+	assert.Equal(t, int32(1), cat.attempts.Load()-beforeLeader,
```
laskoviymishka, line 634: The `attempts.Load() - beforeLeader == 1` math is load-bearing for the assertion's claim ("validator rejects on retry before re-issuing CommitTable") but takes a second read to follow — the count includes the leader's first attempt that fails on the stale `AssertRefSnapshotID`, then the validator runs on the refresh path and short-circuits before a second `CommitTable`. A line of comment naming both attempts (or naming the delta as "only the stale-assertion attempt landed; the retry never reached CommitTable") would help the next reader.

laskoviymishka (changes requested):
Direction looks good overall — the `RewriteFiles` builder is a clean lift from Java’s `Table.newRewrite()`, the per-group commit fix in partial-progress mode addresses a real case, and the `withRewriteSemantics` gating reads well. The distributed-equivalence and concurrent-eq-delete tests are also a good shape.

I’d hold this before merge on a few things:

* The `OpReplace` flip looks right for Java parity, but I’d like to lock down the snapshot summary contract. Java rewrite snapshots include keys like `added-data-files`, `deleted-data-files`, `removed-position-delete-files`, and the byte counters. If any of our summary-writing code still branches on `OpOverwrite`, this could silently drop fields that Java/PyIceberg readers expect. A summary-properties assertion in one rewrite test would cover this.

* The builder should probably be single-use. Calling `Commit` twice currently re-appends the rewrite validator and re-stages `ReplaceFiles`; Java rejects this pattern. Also, the empty-rewrite error should wrap `ErrInvalidOperation` so callers can branch on it.

* The partial-progress docs read a bit contradictory: one part says work survives a mid-loop failure, but another says true per-group durability requires separate transactions. Now that `NewRewrite` makes the per-group transaction pattern possible, the doc should point users there.

A few concrete asks before merge:

* wrap the empty-rewrite error with `ErrInvalidOperation`
* guard `RewriteFiles.Commit` against double-call
* add one rewrite test that asserts summary properties under `OpReplace`
* tighten the V3 manifest wording in the `RewriteFiles` type doc — `referenced_data_file` is V2-optional, V3-required for DVs
* reshape `TestRewriteFiles_DeleteFile_RoutesByContentType` so it checks behavior that would actually fail if the two slices were misrouted

twuebi:
@laskoviymishka, addressed your comments

table/rewrite_files.go
```diff
@@ -0,0 +1,213 @@
...
+	if r.committed {
+		return fmt.Errorf("%w: RewriteFiles.Commit already called on this builder", ErrInvalidOperation)
+	}
+	r.committed = true
+
+	if r.err != nil {
+		return r.err
+	}
+	if len(r.dataFilesToDelete) == 0 && len(r.dataFilesToAdd) == 0 && len(r.deleteFilesToRemove) == 0 {
+		return fmt.Errorf("%w: rewrite must have at least one file change", ErrInvalidOperation)
```
laskoviymishka, line 215: Empty-deletes-with-adds silently downgrades to `OpAppend`. When `dataFilesToDelete == 0 && dataFilesToAdd > 0 && deleteFilesToRemove == 0`, the guard at L196 passes, the call reaches `ReplaceFiles`, which delegates to `ReplaceDataFilesWithDataFiles` (no deletes to remove), which delegates to `AddDataFiles` (no files to delete). `AddDataFiles` is an `OpAppend` producer and never consults `cfg.rewriteSemantics`, so the snapshot is tagged `append` and no rewrite validator is registered — even though `withRewriteSemantics()` is set on the option chain.

A distributed coordinator that has a group with only new files (e.g., a worker that wrote replacement files but the delete side is empty because nothing was actually replaced) gets a silently wrong snapshot op and no isolation against concurrent writers. Worth rejecting explicitly:

```go
if len(r.dataFilesToDelete) == 0 && len(r.dataFilesToAdd) > 0 {
    return fmt.Errorf("%w: rewrite must delete at least one data file when adding data files", ErrInvalidOperation)
}
```

(A pure delete-files expunge with no data movement — `deleteFilesToRemove > 0` only — is still a legitimate `OpReplace` and would pass.)

table/rewrite_files.go
```diff
@@ -0,0 +1,213 @@
...
+	if err := r.txn.ReplaceFiles(ctx, r.dataFilesToDelete, r.dataFilesToAdd, r.deleteFilesToRemove, r.snapshotProps, withRewriteSemantics()); err != nil {
+		return err
+	}
+
+	if len(r.dataFilesToDelete) > 0 {
+		rewritten := make([]string, 0, len(r.dataFilesToDelete))
+		for _, df := range r.dataFilesToDelete {
+			rewritten = append(rewritten, df.FilePath())
+		}
+		r.txn.validators = append(r.txn.validators, rewriteValidator(rewritten))
```
laskoviymishka: `r.txn.validators = append(...)` mutates a `Transaction` field without holding `t.mx`. `Transaction.Commit` takes `t.mx` and then reads `t.validators` under the lock. The doc-comment on `RewriteFiles` (L46-49) explicitly endorses the fanout pattern of multiple builders feeding the same transaction — that is exactly the surface where two goroutines could race on this append.

`RowDelta.Commit` has the same convention break, so fixing both at once via a tiny helper would be cleanest:

```go
// in transaction.go
func (t *Transaction) addValidator(v conflictValidatorFunc) {
    t.mx.Lock()
    defer t.mx.Unlock()
    t.validators = append(t.validators, v)
}
```

Then `RewriteFiles.Commit` and `RowDelta.Commit` both use `r.txn.addValidator(...)`. `go test -race` would catch the existing race today under any concurrent driver.

table/rewrite_files.go
```diff
@@ -0,0 +1,213 @@
...
+//
+// The typical distributed-coordinator pattern is one [RewriteFiles]
+// builder + one Apply call per worker result + one Commit:
+//
+//	rewrite := leaderTxn.NewRewrite(snapshotProps)
+//	for _, gr := range workerResults {
+//	    rewrite.Apply(gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes)
+//	}
+//	if err := rewrite.Commit(ctx); err != nil { ... }
+func (r *RewriteFiles) Apply(deletes, adds, safeDeletes []iceberg.DataFile) *RewriteFiles {
```
laskoviymishka, line 158: The three-slice shape is a structural footgun for distributed coordinators. `deletes`, `adds`, and `safeDeletes` are all `[]iceberg.DataFile`; `deletes` and `safeDeletes` both route through `DeleteFile()` and are dispatched by content type, so a caller transposing slot 1 and slot 3 is silent unless the dispatch-by-content-type catches it. The documented coordinator pattern (L153-158) hands `gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes` in exactly the same slot order — three positional same-typed slices that happen to map 1:1 to three same-typed struct fields is the kind of glue that breaks silently under refactor.

A non-breaking option: add a typed sibling that takes the result directly:

```go
func (r *RewriteFiles) ApplyResult(gr CompactionGroupResult) *RewriteFiles {
    return r.Apply(gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes)
}
```

At minimum the param could be renamed `safePosDeletes` to match the source field name, and the godoc could warn that slot 1 and slot 3 are not interchangeable despite both routing through `DeleteFile`.

table/rewrite_data_files.go
```diff
@@ -70,34 +71,112 @@ type CompactionTaskGroup struct {
...
+type compactionGroupConfig struct {
+	targetFileSize  int64
+	scanConcurrency int
+}
+
+// WithCompactionTargetFileSize sets the size target for output files
+// written by [ExecuteCompactionGroup]. Forwarded to [WriteRecords] as
+// [WithTargetFileSize]. Zero (the default) means inherit the table's
+// `write.target-file-size-bytes` property.
+func WithCompactionTargetFileSize(size int64) CompactionGroupOption {
```
laskoviymishka, line 169: Negative values are silently ignored. The consumer at L291 checks `cfg.targetFileSize > 0` before applying it, so passing `-1` (or any negative) sets the field, fails the `> 0` gate, and silently falls through to the table-property default — but the doc on L163-166 only mentions zero as the inherit sentinel. A coordinator computing `targetFileSize = baseSize - overhead` could produce a negative under certain inputs and silently get the default instead of an error.

`WitMaxConcurrency` (table.go:692) already handles this — it returns `noopOption` for `n <= 0`. Mirror that:

```go
func WithCompactionTargetFileSize(size int64) CompactionGroupOption {
    if size <= 0 {
        return func(*compactionGroupConfig) {}
    }
    return func(c *compactionGroupConfig) { c.targetFileSize = size }
}
```

table/rewrite_data_files.go
```diff
@@ -70,34 +71,112 @@ type CompactionTaskGroup struct {
...
+	return func(c *compactionGroupConfig) {
+		c.targetFileSize = size
+	}
+}
+
+// WithCompactionScanConcurrency sets the scan concurrency used when
+// reading the group's tasks. Forwarded to [Table.Scan] as
+// [WitMaxConcurrency]. Zero (the default) means runtime.GOMAXPROCS.
```
laskoviymishka, line 181: Nit: this godoc link enshrines the pre-existing `WitMaxConcurrency` typo (missing `h`) in a second exported symbol's documentation, making the eventual rename a touch noisier. Either leave a TODO referencing the typo or wait for the rename PR to land before adding this link. Not blocking.

table/rewrite_data_files.go
```diff
@@ -242,19 +379,24 @@ func rewriteValidator(rewrittenPaths []string) conflictValidatorFunc {
...
 // Equality deletes are decided by [compaction.DecideDeadEqualityDeletes]
 // (which needs partition-wide visibility, not just the task scope).
 // Deletion vectors will be handled when DV read support lands.
-func collectSafePositionDeletes(tasks []FileScanTask) []iceberg.DataFile {
+//
+// [ExecuteCompactionGroup] calls this internally to populate
+// [CompactionGroupResult.SafePosDeletes]. It is kept exported for
+// custom workers that want the spec-shaped predicate without taking
+// the rest of [ExecuteCompactionGroup]'s read+write pipeline.
+func CollectSafePositionDeletes(tasks []FileScanTask) []iceberg.DataFile {
```
laskoviymishka, line 439: Predicate caveat: this collects every pos-delete file matched to *any* task in the group, but it doesn't check that *every* data file the pos-delete references is in the rewrite set. Within a partition the planner produces multiple groups via bin-packing (`compaction.go` `packer.PackEnd`) and skips files via `MinInputFiles` — so a pos-delete file that targets data files in groups 1 and 2 (or in group 1 + a skipped file) will get marked safe by group 1's pass. After group 1's commit, the still-live data files lose their deletes.

A second-order consequence now that this is public API: if a pos-delete spans two groups, both groups' `SafePosDeletes` will contain it; when the coordinator calls `Apply` twice on the same `RewriteFiles`, `deleteFilesToRemove` ends up with duplicate paths and `ReplaceFiles` rejects with "delete file paths must be unique." That at least makes the bug self-reporting in the multi-Apply case, but the single-group case is still silent data loss.

This is pre-existing as `collectSafePositionDeletes` and not a regression; but it's worth at minimum a doc note pinning the contract ("caller must verify every referenced data file is in the rewrite set") since the function is now public, and ideally a follow-up that takes a `rewrittenPaths` set or moves the computation leader-side after worker outputs aggregate.

table/rewrite_data_files.go
```diff
@@ -142,91 +219,151 @@ func (t *Transaction) RewriteDataFiles(ctx context.Context, groups []CompactionT
...
-		if len(opts.ExtraDeleteFilesToRemove) > 0 {
-			allOldDeletes = append(allOldDeletes, opts.ExtraDeleteFilesToRemove...)
-			result.RemovedEqualityDeleteFiles += len(opts.ExtraDeleteFilesToRemove)
+		if len(gr.OldDataFiles) == 0 && len(gr.NewDataFiles) == 0 {
+			continue
 		}
 
-		if err := t.ReplaceFiles(ctx, allOldData, allNewData, allOldDeletes, opts.SnapshotProps, withRewriteSemantics()); err != nil {
-			return result, fmt.Errorf("commit compaction: %w", err)
+		if err := t.NewRewrite(opts.SnapshotProps).Apply(gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes).Commit(ctx); err != nil {
```
laskoviymishka: Partial-progress accumulates one rewrite validator per group on the transaction. After N groups, the transaction has N validators, and at `Transaction.Commit` time (refresh-replay) each independently walks the same concurrent-snapshot set looking for deletes against its own path subset — O(N · |concurrent_snapshots|) walks where the conflict surface is really the union and one walk would do.

Not a correctness issue, but partial-progress is the path callers reach for when groups are many (the docstring frames it that way), so the perf cliff scales with the very dimension this mode is for. Cheap fix is to coalesce: collect all rewritten paths across groups and register one `rewriteValidator(allRewritten)` at the end of `rewriteDataFilesPartial`, skipping the per-group validator append in this path.

table/rewrite_data_files_test.go
```diff
@@ -243,6 +243,122 @@ func TestRewriteDataFiles_EmptyPlan(t *testing.T) {
...
+
+	tasks, err := tbl.Scan().PlanFiles(t.Context())
+	require.NoError(t, err)
+
+	plan, err := defaultTestCompactionCfg.PlanCompaction(tasks)
+	require.NoError(t, err)
+	require.NotEmpty(t, plan.Groups)
+
+	groups := toTaskGroups(plan.Groups)
+	require.Len(t, groups, 1, "test assumes a single group; tighten plan if this changes")
```
laskoviymishka: Brittle: this test exists to verify `WithCompactionTargetFileSize` is forwarded, but it couples that assertion to the planner producing exactly one group. If `MinInputFiles`, bin-packing parameters, or default file sizes ever shift, this fails for an unrelated reason.

`TestRewriteDataFiles_PartialRewritePreservesEqDelete` (L583-589) constructs the group manually, which decouples option-forwarding from planner behavior — same pattern would work here.

table/rewrite_data_files_test.go
```diff
@@ -243,6 +243,122 @@ func TestRewriteDataFiles_EmptyPlan(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NotEmpty(t, plan.Groups)
+
+	groups := toTaskGroups(plan.Groups)
+
+	got, err := table.ExecuteCompactionGroup(t.Context(), tbl, groups[0],
+		table.WithCompactionScanConcurrency(1))
+	require.NoError(t, err)
+	assert.NotEmpty(t, got.NewDataFiles)
+	assert.NotEmpty(t, got.OldDataFiles)
```
laskoviymishka: This test is effectively tautological — its only assertions are `NotEmpty(NewDataFiles)` and `NotEmpty(OldDataFiles)`, both of which would pass if `WithCompactionScanConcurrency` were a complete no-op. The test docstring acknowledges "we can't easily observe scan parallelism from the result," but in its current shape a regression that silently drops the option would not be caught.

At minimum, run a second `ExecuteCompactionGroup` without the option and assert the two results have the same `len(NewDataFiles)` / `len(OldDataFiles)` — so a regression that breaks the scan path when concurrency is set produces divergent output.

table/rewrite_data_files_test.go
```diff
@@ -243,6 +243,122 @@ func TestRewriteDataFiles_EmptyPlan(t *testing.T) {
...
+	plan, err := defaultTestCompactionCfg.PlanCompaction(tasks)
+	require.NoError(t, err)
+
+	tx := tbl.NewTransaction()
+	result, err := tx.RewriteDataFiles(t.Context(), toTaskGroups(plan.Groups), table.RewriteDataFilesOptions{
+		GroupOptions: []table.CompactionGroupOption{
+			table.WithCompactionTargetFileSize(1),
+		},
+	})
+	require.NoError(t, err)
```
laskoviymishka, line 370: Test doesn't commit the transaction. `result.AddedDataFiles > 1` is an in-process counter that's incremented before any `ReplaceFiles` is staged, so the option-forwarding claim is correct — but a regression that breaks `ReplaceFiles` for tiny-target-size rewrites would slip through.

Adding `committed, err := tx.Commit(t.Context()); require.NoError(t, err)` plus a row-count assertion against the committed snapshot would harden this against future regressions without much extra setup.

table/rewrite_files.go
```diff
@@ -0,0 +1,213 @@
...
+// files being rewritten and is therefore safe to expunge).
+//
+// Any other content type stages an error that is returned from the
+// next [RewriteFiles.Commit] call.
+func (r *RewriteFiles) DeleteFile(df iceberg.DataFile) *RewriteFiles {
+	if r.err != nil {
+		return r
+	}
+
+	switch df.ContentType() {
```
laskoviymishka, line 110: Nit: `df.ContentType()` is called without a nil-check. `iceberg.DataFile` is an interface, so a nil argument panics with a nil-pointer-dereference instead of returning a clean `ErrInvalidOperation`. `ReplaceFiles` at transaction.go:822/834 guards nil explicitly. Distributed callers deserialize data files from a wire format, so a wire-decode bug producing a nil entry would crash the coordinator rather than fail cleanly. Same applies at `AddDataFile` L129. Easy fix:

```go
if df == nil {
    r.err = fmt.Errorf("%w: nil data file", ErrInvalidOperation)
    return r
}
```

laskoviymishka (changes requested):
Iter2 looks good. The main issues from the last pass are addressed cleanly: single-shot `Commit` guard, validator ordering after `ReplaceFiles`, `maps.Clone` on `snapshotProps`, `ErrInvalidOperation` wrapping, `DeleteFile` switch/default, tighter routing test, cross-client summary key assertions, V3/V2 doc fix, partial-progress doc rewrite, and the `attempts.Load` comment. I resolved the relevant threads.

Two small pre-merge fixes left:

1. `Commit` with adds and no deletes still routes through `AddDataFiles`, producing `OpAppend` with no validator — so `withRewriteSemantics()` is dead on that branch. I’d reject that case.
2. `WithCompactionTargetFileSize` silently ignores negative values. I’d mirror `WithMaxConcurrency`: `<= 0` means no-op.

Three tests I’d also harden in the same pass:

1. `TestRewriteDataFiles_GroupOptionsForwarded` should commit the txn.
2. `TestExecuteCompactionGroup_ScanConcurrencyForwarded` should assert the actual forwarded value, not just non-empty.
3. `TestExecuteCompactionGroup_TargetFileSizeForwarded` is a bit brittle because it depends on `require.Len(groups, 1)`.

The rest I’d track as follow-ups, not merge blockers:

* `CollectSafePositionDeletes` has a known edge case when a position-delete spans groups or references files skipped by `MinInputFiles`. Since this PR promotes it to public API, the contract should be documented.
* `r.txn.validators` append is racy with `Transaction.Commit`; `RowDelta` seems to have the same issue. A small `Transaction.addValidator(v)` helper would fix both.
* Partial progress accumulates one validator per group, so retries do O(N) validator walks where one validator would probably be enough.
* `Apply(deletes, adds, safeDeletes)` is easy to misuse because slot 1 ↔ 3 transposition is silent. An additive `ApplyResult(CompactionGroupResult)` overload would read much cleaner.
* Java-parity follow-ups: conservative eq-delete branch, missing `validateFromSnapshot` analog, and no `WriteOption` pass-through from `Commit`.

Test gaps worth follow-up: concurrent append into rewritten partition succeeds, V3 position-delete branch, multi-`Apply` aggregation with a spanning position-delete, and spec evolution between worker and coordinator.

Net: with the two small fixes and the test hardenings, I’d merge.

twuebi:
@laskoviymishka addressed the blockers and also already most of the follow ups, they were rather small changes so I decided not to defer them. Missing follow ups are the Java-parity and test-gap follow-ups

laskoviymishka (approved):
Great job!  Approving :shipit: 

A few nice follow-ups to file while this surface is still fresh — many of these are test-only and could be good first issues:

* multi-`Apply` spanning position-delete test, which should expose the `CollectSafePositionDeletes` predicate gap
* V3 position-delete validator-branch test
* “concurrent append into rewritten partition succeeds” test — this is the core rewrite-vs-overwrite property; today we only cover the negative case
* spec-evolution-between-worker-and-coordinator test
* Java summary-key round-trip test
* direct `ApplyResult` test
* `seedDataFiles` helper to dedupe the repeated “AddFiles + Commit N small files” setup

A few non-test follow-ups are probably worth keeping closer:

* rename `safeDeletes` → `safePosDeletes` for symmetry with `CompactionGroupResult.SafePosDeletes`
* pass `WriteOption` through from `RewriteFiles.Commit`
* add a future `validateFromSnapshot(int64)` analog for distributed coordinators
* eventually fix `CollectSafePositionDeletes` by recomputing over the full rewrite set on the leader side
* tighten the conservative eq-delete branch for Java parity

Happy to file the issues myself if helpful.

Issue #798, opened 2026-03-17
Iceberg hadoop catalog support


juanli16:
### Feature Request / Improvement

iceberg-go currently lacks the support for hadoop catalog: https://iceberg.apache.org/javadoc/nightly/org/apache/iceberg/hadoop/HadoopCatalog.html

Is this on your roadmap or would you want contribution to add support for this?

tanmayrauth:
@laskoviymishka  I am planning to pick this is up. What do you think about the below plan? 


### Proposal for Hadoop Catalog Support                                                                                                                                                                                                                    
                                                            
 1: Scaffold + path helpers                                                                                                                                                                                                   
  Add the Hadoop type constant to catalog/catalog.go and create the catalog/hadoop/ package with the Catalog struct, NewCatalog constructor, and init() self-registration. CatalogType() returns catalog.Hadoop and all other  interface methods stub out with a "not yet implemented" error, with a compile-time var _ catalog.Catalog = (*Catalog)(nil) check. This PR also includes all path helpers — namespaceToPath, tableToPath, metadataDir, defaultTableLocation, metadataFilePath, and versionHintPath — along with the version regex v([0-9]+)(?:\.gz)?\.metadata\.json and isTableDir(path) which checks for matching metadata files. Tests cover registration via   catalog.Load, path construction, regex matching, and isTableDir true/false cases.

2: Version-hint management                                                                                                                                                                                                   
  Implement readVersionHint (reads the integer from version-hint.text, returns 0 on any error), writeVersionHint (writes to a temp file then renames, best-effort with errors logged but not returned), and findVersion with the full three-tier fallback: read hint and validate the file exists, scan forward from the hinted version until no file is found, and if there's no hint at all, list the metadata directory to find the max version then scan forward. Tests cover round-trip read/write, missing hint, corrupt hint, stale hint (v3 exists but hint says 2), scan-forward behavior, directory listing fallback, and the no-metadata-files case.                              
                                                            
3: Namespace operations                                                                                                                                                                                                      
[...]

PR #1034, opened 2026-05-07
feat(catalog): hadoop cli integration

tanmayrauth:
[7: CLI integration](https://github.com/apache/iceberg-go/issues/798#issuecomment-4320784323)
Added a case catalog.Hadoop: branch to the catalog switch in cmd/iceberg/main.go, wire the cfg.Warehouse property, and add the blank import for catalog/hadoop.

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -33,6 +33,8 @@ import (
 
 	"github.com/apache/iceberg-go"
 	"github.com/apache/iceberg-go/catalog"
+	"github.com/apache/iceberg-go/catalog/internal"
+	icebergio "github.com/apache/iceberg-go/io"
 	"github.com/apache/iceberg-go/table"
 	"github.com/google/uuid"
 )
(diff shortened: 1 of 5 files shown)


catalog/hadoop/hadoop.go
```diff
@@ -296,12 +298,208 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
...
+		return false, nil
+	}
+
+	return isTableDir(c.tableToPath(ident)), nil
+}
+
+// writeMetadataLocationKey is the Java Iceberg property that allows relocating
+// metadata to a custom path. The Hadoop catalog forbids this because it derives
+// metadata paths from the table identifier.
+const writeMetadataLocationKey = "write.metadata.location"
```
zeroshade: should this be a constant at the catalog package level or in table properties instead of being in the hadoop package?

zeroshade (changes requested):
add updates to the README for the support please

catalog/hadoop/hadoop.go
tanmayrauth: Makes sense, originally scoped it to the hadoop package since it was only used there for the rejection check, but table/properties.go is the right home given that's where all the other write.metadata.* keys live. Moved it there as a public WriteMetadataLocationKey constant.

zeroshade (approved):

PR #1035, opened 2026-05-07
feat(arrow): map unknown type to null

### LINKED ISSUE #1003: feat(arrow): map UnknownType to plain Arrow null and back
Parent: #589

iceberg-go's iceberg-to-arrow converter wraps `UnknownType` in an Opaque extension type around `arrow.Null`. pyiceberg maps it to plain `pa.null()` instead. Mirror pyiceberg's posture so the cross-client wire format on Arrow records matches.

Two pieces:

1. Forward (iceberg → arrow): drop the `OpaqueType` wrapper in the `convertToArrow.VisitUnknown` visitor; emit plain `arrow.Null`.
2. Reverse (arrow → iceberg): no case for `*arrow.NullType` in the `convertToIceberg` switch over Arrow types. Add it; map to `iceberg.PrimitiveTypes.Unknown`.

Tests cover round-trip in both directions. Watch for any callers that depend on the Opaque wrapper's extension metadata; if they exist, this is a small breaking change worth calling out in the PR.

Herrtian:
Closes #1003.

Maps Iceberg unknown fields to plain Arrow null fields, and maps Arrow null fields back to Iceberg unknown.

Checked with `go test ./table -run TestArrow`.

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -315,6 +315,8 @@ func (c convertToIceberg) Primitive(dt arrow.DataType) (result iceberg.NestedFie
 		return c.Primitive(dt.Encoded())
 	case *arrow.BooleanType:
 		result.Type = iceberg.PrimitiveTypes.Bool
+	case *arrow.NullType:
+		result.Type = iceberg.PrimitiveTypes.Unknown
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Good call

Issue #997, opened 2026-05-05
feat(table): deletion-vector writer


laskoviymishka:
Parent: #589

Depends on #866 (DV bitmap reader landing first to settle the codec). v3 prefers DVs over Parquet position-delete files. Iceberg-go would be the first non-Java client with a DV writer — pyiceberg writes Parquet position-deletes today, and iceberg-rust hasn't shipped this either.

Shape:

```go
// New file: table/internal/dv_writer.go
type DVWriter struct {
    fs    io.IO
    blobs []puffin.Blob
}

func (w *DVWriter) Add(dataFilePath string, positions []int64) error
func (w *DVWriter) Flush(ctx context.Context, location string) (DataFile, error)
```

Serializes a Roaring 64-bit bitmap → Puffin blob (`blob_type=deletion-vector-v1`) → manifest entry with `content=1, file_format=PUFFIN, ReferencedDataFile, ContentOffset, ContentSizeInBytes`. Commits through the existing `RowDelta.AddDeletes()` API so the rest of the producer stack is unchanged.

Add a table property `write.delete.format=position|dv`. On v3 the default flips to `dv`; v2 stays on `position`. When `dv`, the existing position-delete writer in `table/internal/parquet_files.go` is bypassed in favor of `DVWriter`.

Scope is large enough that it should land across multiple PRs — roaring serialization + writer skeleton, then producer wiring + property gating, then cross-client tests is one reasonable split. Discussion of the breakdown is welcome in this thread before any code lands.

Spec: [Iceberg deletion vectors](https://iceberg.apache.org/spec/#deletion-vectors), [Puffin format](https://iceberg.apache.org/puffin-spec/), [RoaringBitmap serialization](https://github.com/RoaringBitmap/RoaringFormatSpec). Cross-client coverage: write a DV via iceberg-go, read back by Java/pyiceberg, assert filtered rows match.

tanmayrauth:
@laskoviymishka  can I pick this up ? Here's my thinking on the implementation. 

                                                                                         
                  
  Serialization                                                                                                                                                                                                                          
                  
We have DeserializeDV() but no inverse. I'd add a SerializeDV(bitmap *RoaringPositionBitmap) ([]byte, error) in table/dv/deletion_vector.go that produces the spec format: [length(4)] [magic(4)] [bitmap bytes] [crc32(4)].   Straightforward mirror of the deserialization logic we already have.
                                                                                                                                                                                                                                         
[...]

PR #1037, opened 2026-05-08
feat(table): warn when writing Parquet position-deletes on v3

### LINKED ISSUE #1007: feat(table): warn when writing Parquet position-deletes on v3
Parent: #589

v3 prefers deletion vectors over Parquet position-delete files. Java logs a warning when the older format is used (Java precedent: [apache/iceberg#12048](https://github.com/apache/iceberg/pull/12048)); iceberg-go silently writes the deprecated path.

Emit a single `slog.Warn` per writer with a stable, deduped message naming the table and pointing users at deletion vectors. The dedup matters because a partitioned-fanout writer would otherwise log once per partition. Mirror the existing `slog`-based warnings (the silent-data-affecting-behavior ones added in #811 are a good template).

Test: wire a custom `slog.Handler` into the writer, run a v3 + position-delete write, assert exactly one warning record with the stable message; assert zero warnings on v2.

Becomes obsolete once a DV writer (#997) makes DV the v3 default — but useful in the meantime to guide users away from the deprecated path.

laskoviymishka:
Closes #1007.

V3 prefers deletion vectors over Parquet position-delete files (Iceberg spec, [apache/iceberg#12048](https://github.com/apache/iceberg/pull/12048)). Java logs a warning when the legacy format is used; iceberg-go currently writes it silently. This adds a single `slog.Warn` at the entry of the position-delete writer when the table is v3, naming the table location so users can identify which table is on the deprecated path.

Once the DV writer (#997) lands and flips the default on v3, this warning becomes a backstop for users who explicitly opt back into Parquet position-deletes via `write.delete.format=position`.

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1607,6 +1607,13 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string, 
 		}
 	}
 
+	// V3 prefers deletion vectors over Parquet position-delete files; warn so users
+	// migrate when DV-write support lands. See apache/iceberg#12048.
+	if latestMetadata.Version() >= 3 {
+		slog.Warn("writing Parquet position-delete file on a v3 table; deletion vectors are preferred",
+			"table", latestMetadata.Location())
+	}
+
 	if args.writeUUID == nil {
 		u := uuid.Must(uuid.NewRandom())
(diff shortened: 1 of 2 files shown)


laskoviymishka:
Closing — pushed for review only, will reopen when ready.

Issue #1038, opened 2026-05-08
bug(manifest): v2 manifests silently drop data_file.distinct_counts (Avro schema gap)


mzzz-zzm:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

iceberg-go writes v2 manifests that silently omit `data_file.distinct_counts`. The Iceberg v2 spec lists `distinct_counts` (field id 111) as a writable optional field on `data_file`, but it is not declared in any of our per-version Avro record schemas, so the encoder never serializes it. Discovered while implementing #1001 (v3 omits the same field, on purpose).

### Affected versions

Writer paths for **v1 and v2** manifests on `main` (`c62af3c`) and earlier. v3 is unaffected — the field is deprecated on v3.

### Root cause

`dataFile.DistinctCounts` carries the Avro tag `avro:"distinct_counts"` ([manifest.go#L1802](https://github.com/apache/iceberg-go/blob/main/manifest.go#L1802)), but the per-version record schemas in [`internal/avro_schemas.go`](https://github.com/apache/iceberg-go/blob/main/internal/avro_schemas.go) do not include a `distinct_counts` field declaration:

- `data_file_v1` (line 238) — no `distinct_counts`
- `data_file_v2` (line 276) — no `distinct_counts`
- `data_file_v3` (line 360) — no `distinct_counts` (intentional, per #1001)

`hamba/avro` writes only fields declared in the schema, so the Go-side pointer is dropped at encode time on every version.

### Reproduction

Drop the test below into `manifest_test.go` and run against `main`. It fails today; it will pass once `data_file_v2` declares `distinct_counts`.

```go
// TestWriteManifestV2KeepsDistinctCounts is a regression guard that v2
// manifest writers preserve data_file.distinct_counts (id 111) per the
// Iceberg v2 spec.
func (m *ManifestTestSuite) TestWriteManifestV2KeepsDistinctCounts() {
    partitionSpec := NewPartitionSpecID(0)
    snapshotID := int64(12345678)
    seqNum := int64(9876)

    dataFileBuilder, err := NewDataFileBuilder(
        partitionSpec,
        EntryContentData,
        "s3://bucket/ns/table/data/distinct.parquet",
        ParquetFile,
        map[int]any{},
        map[int]string{},
        map[int]int{},
        10,
        10*1000,
    )
    m.Require().NoError(err)
    dataFileBuilder.DistinctValueCounts(map[int]int64{1: 42})

    var buf bytes.Buffer
[...]

Issue #589, opened 2025-10-06
feat: v3 support tracking issue


twuebi:
### Feature Request / Improvement

Hi all, let's use this issue to track our progress on adding v3 support to iceberg-go. We can base ourselves on this pyiceberg issue: https://github.com/apache/iceberg-python/issues/1818

- [ ]  New DataTypes
  - [ ]  **UnknownType**
  - [ ]  **Timestamp(Tz)NanoType**
  - [ ]  **VariantType**
  - [ ]  **Geometry**/**Geography**

- [ ]  Default values

- [ ]  Multi-arg transforms
     - Reading of the metadata is being worked on in #585 
     -  Supporting multi-arg transforms still needs to happen

- [ ]  Row-level lineage

- [ ]  Deletion vectors
     - [ ]  Read support
     - [ ]  Write support

dwilson1988:
Great, I'll be taking a look here shortly. Looks like a good start, I just need to wrap my head around the arrow-go, geoarrow interplay with this.

PR #1039, opened 2026-05-08
feat(manifest): omit data_file.distinct_counts on v3 manifests

### LINKED ISSUE #1001: feat(manifest): stop writing distinct_counts on v3 manifests
Parent: #589

The v3 spec deprecates `distinct_counts` on `data_file` (Java precedent: [apache/iceberg#12182](https://github.com/apache/iceberg/pull/12182)). iceberg-go still writes the field unconditionally — `DataFileBuilder.DistinctValueCounts` always sets distinct counts on the underlying Avro record, regardless of format version.

Gate the write so v3 manifests omit the field while v2 still emits it. The struct field can stay (Avro tolerates absent optional fields); the change is on the writer side.

Round-trip test: write a v3 manifest with distinct counts supplied, read it back through `ReadManifest`, assert the distinct counts are absent; same test on v2 should still see the values present.

mzzz-zzm:
Fixes #1001.

## Summary

Add a defensive guard in `v3writerImpl.prepareEntry` that clears the Avro-facing `dataFile.DistinctCounts` pointer on entries we own, so v3 manifests cannot emit `data_file.distinct_counts` (deprecated by the v3 spec; Java parity: [apache/iceberg#12182](https://github.com/apache/iceberg/pull/12182)).

## Scope decision

Per the discussion on #1001 (reviewer approved option **(c) Defensive-only**), this PR ships the minimal v3 gate and defers the v2 schema gap to a separate issue.

**Why minimal**: while implementing this, I found that `data_file.distinct_counts` is **already** absent from the on-disk Avro for every manifest version on `main` — the per-version record schemas in [`internal/avro_schemas.go`](https://github.com/apache/iceberg-go/blob/main/internal/avro_schemas.go) (`data_file_v1` L238, `data_file_v2` L276, `data_file_v3` L360) do not declare a `distinct_counts` field, so the `hamba/avro` encoder drops the Go-side pointer at encode time on every version. The intended v3 outcome is therefore already in effect, but the v2 path is **non-compliant** (v2 spec still requires the field). Tracked separately as #1038.

The guard added here is **forward-compatibility insurance**: when #1038 is fixed by adding `distinct_counts` back to `data_file_v2`, the v3 path will stay compliant with no further code changes.

The `dataFile` struct, `DistinctValueCounts()` getter, and `DataFileBuilder.DistinctValueCounts` setter are intentionally preserved — the change is writer-side only.

## Test

`TestWriteManifestV3OmitsDistinctCounts` writes a v3 manifest with non-empty distinct counts supplied via `DataFileBuilder.DistinctValueCounts`, round-trips it through `ReadManifest`, and asserts the read side observes no distinct counts.

A v2 "keeps distinct counts" test was deliberately **not** added to this PR: against today's broken behavior it would either lie (assert `Empty` is correct) or fail CI (assert `{1:42}`). The v2 round-trip test belongs in the same commit that fixes the v2 schema (see #1038), where it can flip from failing to passing alongside the schema change.

## Verification

- `make test` — all packages green.
- `make lint` — clean.
- DCO sign-off present.

## Related

- Fixes #1001
- Discovered side-issue: #1038 (v2 schema gap)
- Java precedent: [apache/iceberg#12182](https://github.com/apache/iceberg/pull/12182)
- Parent v3 tracker: #589

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -944,6 +944,16 @@ func (v3writerImpl) prepareEntry(entry *manifestEntry, snapshotID int64) (Manife
 		}
 	}
 
+	// v3 spec deprecates data_file.distinct_counts (Java parity:
+	// apache/iceberg#12182). prepareEntry takes ownership of the entry's data
+	// file and clears the Avro-facing pointer; the cached distinctCntMap and
+	// DistinctValueCounts() getter are intentionally preserved so in-process
+	// readers keep their view. Best-effort: only the in-tree *dataFile is
+	// cleared; third-party DataFile impls bypass this guard.
+	if df, ok := entry.DataFile().(*dataFile); ok {
+		df.DistinctCounts = nil
+	}
+
 	return entry, nil
 }
 
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -2115,6 +2115,53 @@ func (m *ManifestTestSuite) TestManifestRoundTripSortOrderID() {
 	m.Equal(expectedSortOrderID, *got)
 }
 
+// TestWriteManifestV3OmitsDistinctCounts verifies the v3 writer clears
+// data_file.distinct_counts (deprecated by v3 spec; Java parity:
+// apache/iceberg#12182). v2 round-trip will be added with #1038.
+func (m *ManifestTestSuite) TestWriteManifestV3OmitsDistinctCounts() {
+	partitionSpec := NewPartitionSpecID(0)
+	snapshotID := int64(1)
+	seqNum := int64(1)
+
+	dataFileBuilder, err := NewDataFileBuilder(
+		partitionSpec,
+		EntryContentData,
+		"s3://bucket/ns/table/data/distinct.parquet",
+		ParquetFile,
+		map[int]any{},
+		map[int]string{},
+		map[int]int{},
+		1,
+		1,
+	)
+	m.Require().NoError(err)
+	dataFileBuilder.DistinctValueCounts(map[int]int64{1: 42})
+
+	entry := NewManifestEntry(
+		EntryStatusADDED,
+		&snapshotID,
+		&seqNum, &seqNum,
+		dataFileBuilder.Build(),
+	)
+
+	var buf bytes.Buffer
+	_, err = WriteManifest(
+		"s3://bucket/ns/table/metadata/distinct.avro", &buf, 3,
+		partitionSpec,
+		NewSchema(
+			0,
+			NestedField{ID: 1, Name: "id", Type: Int64Type{}, Required: true},
+		),
+		snapshotID,
+		[]ManifestEntry{entry},
+	)
+	m.Require().NoError(err)
+
+	df, ok := entry.DataFile().(*dataFile)
+	m.Require().True(ok)
+	m.Nil(df.DistinctCounts, "v3 writer must clear DistinctCounts on the entry's *dataFile")
+}
+
 func (m *ManifestTestSuite) TestWriteManifestClosesWriterOnEntryError() {
 	partitionSpec := NewPartitionSpecID(1,
 		PartitionField{FieldID: 1000, SourceIDs: []int{1}, Name: "VendorID", Transform: IdentityTransform{}},


manifest_test.go
```diff
@@ -2115,6 +2115,67 @@ func (m *ManifestTestSuite) TestManifestRoundTripSortOrderID() {
...
+// spec; parity with Java apache/iceberg#12182). Today this passes because none
+// of the per-version Avro record schemas in internal/avro_schemas.go declare
+// a distinct_counts field, so the field is dropped at encode time on every
+// version. The v3writerImpl.prepareEntry guard that clears DistinctCounts is
+// forward-compatibility insurance: if the v2 schema is later fixed to declare
+// distinct_counts (the v2 spec requires it; tracked separately), this test
+// keeps the v3 path compliant without further code changes.
+func (m *ManifestTestSuite) TestWriteManifestV3OmitsDistinctCounts() {
+	got := m.writeManifestWithDistinctCounts(3)
+	m.Empty(got, "v3 manifests must omit distinct_counts")
```
laskoviymishka: This assertion is true on `main` today without the guard — the per-version Avro schemas don't declare `distinct_counts`, so the round-trip is empty either way. You called this out in the PR body, but it means `m.Empty(got)` proves the schema, not the guard.

I'd add a second assertion that pins the actual code path this PR introduces: the in-memory `*dataFile` has its `DistinctCounts` cleared post-write. Something like:

```go
entry := NewManifestEntry(EntryStatusADDED, &snapshotID, &seqNum, &seqNum, dataFileBuilder.Build())
_, err := WriteManifest(... []ManifestEntry{entry})
m.Require().NoError(err)
m.Nil(entry.DataFile().(*dataFile).DistinctCounts, "v3 writer must clear DistinctCounts on the entry's *dataFile")
```

That actually flips red if the `df.DistinctCounts = nil` line is removed, which is the contract we want. The current round-trip assertion can stay as a redundant on-disk check, or be deferred to land alongside #1038 when it stops being vacuous.

manifest.go
```diff
@@ -944,6 +944,19 @@ func (v3writerImpl) prepareEntry(entry *manifestEntry, snapshotID int64) (Manife
...
+	// apache/iceberg#12182). Today none of the per-version Avro schemas in
+	// internal/avro_schemas.go declare distinct_counts, so the field is
+	// already dropped on encode for every version. This guard is
+	// forward-compatibility insurance: if the v2 schema is later fixed to
+	// declare distinct_counts (v2 spec requires it; tracked separately), the
+	// v3 path stays compliant by clearing the Avro-facing pointer here. The
+	// cached distinctCntMap and DistinctValueCounts() getter are left alone
+	// so in-memory readers in the same process keep their view.
+	if df, ok := entry.DataFile().(*dataFile); ok {
+		df.DistinctCounts = nil
```
laskoviymishka, line 954: Worth flagging: this mutates the caller's `*dataFile` in place. `DataFileBuilder.Build()` returns `b.d` directly (same pointer), so after `WriteManifest` returns, the caller's reference now has `DistinctCounts == nil` even though they explicitly set it via `DistinctValueCounts(...)`.

Also applies to the rewrite-existing path in `table/snapshot_producers.go:215` (`wr.Existing(entry)`): we mutate entries that came from `ReadManifest` and may be referenced elsewhere in the same process.

I'd either clone the data file before clearing, or add a sentence to the comment that `prepareEntry` is documented to take ownership of the entry's data file. wdyt?

manifest.go
```diff
@@ -944,6 +944,19 @@ func (v3writerImpl) prepareEntry(entry *manifestEntry, snapshotID int64) (Manife
...
+	// v3 spec deprecates data_file.distinct_counts (parity with Java
+	// apache/iceberg#12182). Today none of the per-version Avro schemas in
+	// internal/avro_schemas.go declare distinct_counts, so the field is
+	// already dropped on encode for every version. This guard is
+	// forward-compatibility insurance: if the v2 schema is later fixed to
+	// declare distinct_counts (v2 spec requires it; tracked separately), the
+	// v3 path stays compliant by clearing the Avro-facing pointer here. The
+	// cached distinctCntMap and DistinctValueCounts() getter are left alone
+	// so in-memory readers in the same process keep their view.
+	if df, ok := entry.DataFile().(*dataFile); ok {
```
laskoviymishka, line 953: The type assertion silently no-ops for any third-party `DataFile` implementation. `DataFile` is a public interface, so external callers can pass their own type and the guard becomes a no-op for them.

Probably fine to defer (the framing is "defensive insurance" anyway), but I'd add one line to the comment so future readers don't think this is a hard gate: `// Best-effort — only clears the Avro-facing pointer for the in-tree *dataFile; third-party DataFile impls bypass this guard.`

manifest.go
```diff
@@ -944,6 +944,19 @@ func (v3writerImpl) prepareEntry(entry *manifestEntry, snapshotID int64) (Manife
...
 
+	// v3 spec deprecates data_file.distinct_counts (parity with Java
+	// apache/iceberg#12182). Today none of the per-version Avro schemas in
+	// internal/avro_schemas.go declare distinct_counts, so the field is
+	// already dropped on encode for every version. This guard is
+	// forward-compatibility insurance: if the v2 schema is later fixed to
+	// declare distinct_counts (v2 spec requires it; tracked separately), the
+	// v3 path stays compliant by clearing the Avro-facing pointer here. The
+	// cached distinctCntMap and DistinctValueCounts() getter are left alone
+	// so in-memory readers in the same process keep their view.
```
laskoviymishka: This 8-line comment is essentially the PR body. The reference to `internal/avro_schemas.go` will rot (line numbers, file path) and the schema-already-drops-it framing is context for the review, not for future readers of the file.

I'd condense to two lines:

```go
// v3 spec deprecates data_file.distinct_counts (Java parity: apache/iceberg#12182).
// Defensive: clears the Avro-facing pointer; cached distinctCntMap and DistinctValueCounts() getter
// are intentionally preserved so in-process readers keep their view.
```

Link #1038 in the commit message rather than the source if you want a forward pointer.

manifest_test.go
```diff
@@ -2115,6 +2115,67 @@ func (m *ManifestTestSuite) TestManifestRoundTripSortOrderID() {
...
+
+// TestWriteManifestV3OmitsDistinctCounts verifies that v3 manifest writers do
+// not surface data_file.distinct_counts on the read side (deprecated by the v3
+// spec; parity with Java apache/iceberg#12182). Today this passes because none
+// of the per-version Avro record schemas in internal/avro_schemas.go declare
+// a distinct_counts field, so the field is dropped at encode time on every
+// version. The v3writerImpl.prepareEntry guard that clears DistinctCounts is
+// forward-compatibility insurance: if the v2 schema is later fixed to declare
+// distinct_counts (the v2 spec requires it; tracked separately), this test
+// keeps the v3 path compliant without further code changes.
```
laskoviymishka: Same as the manifest.go comment — most of this duplicates the PR body, and a chunk of it ("the v2 schema is later fixed... this test keeps the v3 path compliant") is only meaningful relative to #1038.

I'd trim to one sentence:

```go
// v3 manifests must not surface data_file.distinct_counts (deprecated by v3 spec;
// Java parity: apache/iceberg#12182). v2 round-trip will be added with #1038.
```

manifest_test.go
```diff
@@ -2115,6 +2115,67 @@ func (m *ManifestTestSuite) TestManifestRoundTripSortOrderID() {
 	m.Equal(expectedSortOrderID, *got)
 }
 
+// writeManifestWithDistinctCounts builds a single-entry manifest at the given
+// version with distinct_counts populated for one column, round-trips it through
+// WriteManifest + ReadManifest, and returns the distinct counts observed by
+// the reader.
+func (m *ManifestTestSuite) writeManifestWithDistinctCounts(version int) map[int]int64 {
```
laskoviymishka: The `version int` parameter is dead — only `3` is ever passed, and the v2 case the helper anticipates is correctly deferred to #1038.

I'd either inline this into `TestWriteManifestV3OmitsDistinctCounts` (it's only ~20 lines, only one call site) or leave the helper as-is and add the v2 caller in the #1038 PR. Building the abstraction now without a second user reads as premature.

While we're here: `snapshotID := 12345678`, `seqNum := 9876`, `recordCount: 10`, `fileSize: 10*1000` — none of those values are checked. A literal `1` everywhere would be clearer noise.

laskoviymishka (changes requested):
Right direction, two things before merg.

Test is vacuous against current main — distinct_counts isn't in any schema yet, so m.Empty(got) passes either way. Need an in-memory check on `*dataFile.DistinctCounts` post-write to actually exercise the guard.

Guard mutates the caller's data file (`Build()` returns the live pointer, rewrite path in `snapshot_producers.go` reuses entries). Clone or document ownership. Comment/docstring/helper nits inline.

manifest_test.go
mzzz-zzm: Fixed. The test now asserts the new code path directly:

```go
df, ok := entry.DataFile().(*dataFile)
m.Require().True(ok)
m.Nil(df.DistinctCounts, "v3 writer must clear DistinctCounts on the entry's *dataFile")

manifest.go
mzzz-zzm: Fixed by documenting the ownership choice in the source comment rather than cloning. The updated comment now states that `prepareEntry` takes ownership of the entry's data file and clears only the Avro-facing pointer while preserving the cached `distinctCntMap` / `DistinctValueCounts()` view.

I kept the mutation approach because this PR is intentionally a narrow writer-path guard, and the commit message now calls out the ownership behavior explicitly as durable context.

manifest_test.go
mzzz-zzm: Fixed. The test comment is now short and only states the behavior under test plus the Java parity reference / #1038 note. The PR-body-style explanation was removed.

manifest_test.go
mzzz-zzm: Fixed. I removed the helper and inlined the setup into `TestWriteManifestV3OmitsDistinctCounts`, since v3 is the only caller in this PR. I also changed the unused fixture values to `1` so they read as filler rather than meaningful test data.

mzzz-zzm:
The integration-test job failed on the `container trino is unhealthy` flake during `make integration-setup` (Trino's container health check missed the compose `--wait` window — `docker compose up` returned before Trino's JVM finished loading plugins). Same fingerprint hit unrelated branches in recent runs that later passed on retry, e.g. runs 25516186598, 25441863075, 25420611326 — all from PRs that have since merged or moved forward.

The PR diff is comments + a single test assertion change in `manifest_test.go` — nothing that can affect the docker-compose recipe or the Trino container.

Could a committer please re-run the failed `integration-test` job on run 25588914120? Thanks!

manifest.go
mzzz-zzm: Fixed. The comment now calls this out explicitly:

```go
Best-effort: only the in-tree *dataFile is cleared; third-party DataFile impls bypass this guard.

laskoviymishka (approved):
LGTM!

PR #1040, opened 2026-05-08
feat(table): warn when writing Parquet position-deletes on v3

### LINKED ISSUE #1007: feat(table): warn when writing Parquet position-deletes on v3

laskoviymishka:
Closes #1007.

V3 prefers deletion vectors over Parquet position-delete files (apache/iceberg#12048). Java logs a warning when the legacy path is used; iceberg-go silently writes the deprecated path. This emits a single slog.Warn at the entry of positionDeleteRecordsToDataFiles when the table is v3+, naming the table location so users can identify which table is on the deprecated path.

Once the DV writer (#997) lands and flips the default on v3, this becomes a backstop for users who explicitly opt back into Parquet position-deletes via write.delete.format=position.

The warning fires at writer entry, before partition fanout, so a partitioned write logs once total — not once per partition. The partitioned regression test yields two batches (one per partition path) because positionDeletePartitionedFanoutWriter.processBatch reads only the first row's file_path and routes the entire batch to one partition, so a single batch with two rows would not actually exercise the fanout. Verified by temporarily moving the warn into processBatch and confirming the test fails with count==2.

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1607,6 +1607,15 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string, 
 		}
 	}
 
+	// V3 and later prefer deletion vectors over Parquet position-delete files;
+	// warn so users migrate when DV-write support lands. The check is `>= 3`
+	// rather than `== 3` so the warning carries forward to v4+ without churn.
+	// See apache/iceberg#12048.
+	if latestMetadata.Version() >= 3 {
+		slog.Warn("writing Parquet position-delete file on a v3 table; use deletion vectors instead",
+			"table_location", latestMetadata.Location())
+	}
+
 	if args.writeUUID == nil {
 		u := uuid.Must(uuid.NewRandom())
 		args.writeUUID = &u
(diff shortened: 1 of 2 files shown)


table/arrow_utils.go
```diff
@@ -1607,6 +1607,15 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string,
 		}
 	}
 
+	// V3 and later prefer deletion vectors over Parquet position-delete files;
+	// warn so users migrate when DV-write support lands. The check is `>= 3`
+	// rather than `== 3` so the warning carries forward to v4+ without churn.
+	// See apache/iceberg#12048.
+	if latestMetadata.Version() >= 3 {
+		slog.Warn("writing Parquet position-delete file on a v3 table; use deletion vectors instead",
```
zeroshade: maybe change the wording to be "prefer deletion vectors" rather than "use deletion vectors"?

zeroshade (approved):

PR #1041, opened 2026-05-08
test(puffin): golden round-trip a deletion-vector-v1 blob

### LINKED ISSUE #1008: test(puffin): golden round-trip Java-produced deletion-vector-v1 blob
Parent: #589

The Puffin reader and writer (#676) have round-trip tests against fixtures iceberg-go produces itself. There is no test that exercises a Java-produced `.puffin` file containing a `deletion-vector-v1` blob — the cross-client wire-format guarantee is unpinned.

Generate a small fixture (a Java- or pyiceberg-produced `.puffin` with one DV blob), check it into `puffin/testdata/` alongside the existing canonical Java fixtures, then read it via the existing Puffin reader and assert blob count, blob type (`deletion-vector-v1`), magic header, and raw payload bytes against an expected snapshot.

Independent of #866 — uses raw bytes, no roaring codec required. Pins the wire format ahead of #866 / #996 / #997 landing.

laskoviymishka:
Pins the puffin envelope shape around a deletion-vector-v1 blob without depending on the in-flight roaring decoder (#866). The fixture is two files:

  - puffin/testdata/deletion-vector-v1-payload.bin — a Java-produced 64-bit roaring DV payload lifted directly from apache/iceberg test resources (small-alternating-values-position-index.bin, 50 bytes; bitmap encodes positions 1, 3, 5, 7, 9).

  - puffin/testdata/deletion-vector-v1.puffin — the same payload wrapped by puffin.Writer with the spec-canonical metadata (snapshot-id=-1, sequence-number=-1, deletion-vector-v1 blob type, referenced-data-file and cardinality properties).

The test asserts both layers. Reader side: blob count, type, spec-mandated invariants, properties, and that ReadBlob round-trips the inner payload byte-for-byte equal to the standalone Java fixture. Writer side: regenerates the envelope in-memory and asserts byte equality against the on-disk fixture. Without the writer-side check the test was reader-only round-trip and writer drift would silently calcify into the next regeneration.

Honest framing: this is a Go-writer wire-format pin with a Java- equivalent inner payload, not a strong Java cross-impl pin. The basic envelope shape is cross-checked by TestWriterBitIdenticalWithJava, but that test does not exercise empty Fields arrays or multi-key blob Properties — both of which this fixture relies on — and JSON key ordering of blob Properties is encoder-defined.

A regen test gated on REGEN_FIXTURES=1 reproduces the .puffin from the inner payload and self-validates by reading the blob back before overwriting the on-disk file.

Closes #1008.


puffin/testdata/README.md
```diff
@@ -17,5 +17,44 @@ specific language governing permissions and limitations
...
+is encoder-defined. The property values
+(`referenced-data-file=data/test.parquet`, `cardinality=5`,
+`created-by="iceberg-go test fixture"`) are fixture choices, not bytes
+inherited from any specific Java-emitted file.
+
+To regenerate after a deliberate puffin-format change:
+
+```
+REGEN_FIXTURES=1 go test ./puffin/ -run TestRegenerateDeletionVectorPuffinFixture
+```
```
zeroshade, line 53: the standard way of doing this for Go would be for it to be a `go generate` task instead

zeroshade (changes requested):
Can we configure the regen as a go:generate instead?

puffin/testdata/README.md
laskoviymishka: Yes, that's make sense, will do

zeroshade (approved):

Issue #1052, opened 2026-05-10
test(table): end-to-end deletion-vector scan over a real Parquet data file


laskoviymishka:
Parent: #589, follow-up to #996.

The scanner DV read path is unit-tested at the loader level: `TestReadAllDeletionVectors` builds a puffin file in-process, drives `readAllDeletionVectors` directly, and asserts the returned `perFilePosDeletes` map shape. What's not yet covered is the end-to-end claim that DV-deleted row positions are actually absent from `GetRecords` output for a real Parquet data file.

This is the most likely failure mode after a future refactor of the `processPositionalDeletes` pipeline: a change there would be type-correct (the compile-time signature assertion in `dv_scanner_read_test.go` pins the type) but semantically wrong, and no current test would catch it.

Scope: write a small Parquet file with a handful of rows, pair it with a hand-built DV puffin blob marking a subset of positions deleted, construct enough v3 table metadata to drive a scan via the public `TableScan` API, then assert the returned Arrow record batches contain exactly the surviving rows. The fixture scaffolding overlaps with what a future cross-client DV write/read test would need, so this can be folded into that work when it materializes.

Gated on #996 merging to main.

Issue #1051, opened 2026-05-10
fix(table): release Arrow chunks on GetRecords error paths


laskoviymishka:
Parent: #589.

`readAllDeleteFiles` returns a `perFilePosDeletes` map of `*arrow.Chunked` values. The cleanup `defer` in `createIterator` releases those chunks only when the returned iterator runs to exhaustion. On any error return from `GetRecords` between the loaders and the iterator construction — for example, a failing `readAllEqualityDeleteFiles` — the entire `deletesPerFile` map is discarded with `nil, nil, err` and no `Release()` is ever called. Arrow allocations are not freed by GC alone.

`readAllDeletionVectors` (added by #996) merges its results into the same map. Spec-compliant DV bitmaps can be substantially larger than typical Parquet position-delete files, so the per-incident leaked surface is now meaningfully bigger than the pre-existing Parquet-only leak.

Extract a `releasePerFilePosDeletes(m perFilePosDeletes)` helper and call it on every early-return path in `GetRecords`. `readAllDeletionVectors` already performs the same cleanup on its own internal error path — that block is a good template for the helper signature.

Issue #1050, opened 2026-05-10
feat(table): match Java's DV planning validations in PlanFiles


laskoviymishka:
Parent: #589.

`PlanFiles` builds `dvIndex` by iterating DV manifest entries and grouping by `ReferencedDataFile()`. Two gaps relative to Java's `DeleteFileIndex`:

The first is a missing sequence-number guard. The spec says a DV applies to a data file only when the data file's `data_sequence_number` is less than or equal to the DV's `data_sequence_number`. Java's `DeleteFileIndex.findDV` enforces this with a `ValidationException`. The Go `dvIndex` build skips the check, so a stale DV from a prior epoch paired with a newer data file is silently applied — a silent over-deletion path that only triggers on malformed manifests, but the guard is cheap.

The second is that multiple DVs per data file are silently unioned. Java's `Builder.add` errors with `ValidationException("Can't index multiple DVs for %s")`. The Go `dvIndex` is `map[string][]iceberg.DataFile`, so any number of entries are accepted per path. The scanner's `readAllDeletionVectors` defensively rejects this at read time as of #996, but the canonical home for the check is planning — that way callers inspecting `FileScanTask.DeletionVectorFiles` directly see only validated state.

One PR naturally: both checks live in the same `dvIndex` construction loop. The loop was last edited by #996, so this work text-conflicts with that PR's pos-delete suppression block — rebase if both are in flight together.

tanmayrauth:
@laskoviymishka  can I pick this up?

Issue #1049, opened 2026-05-10
feat(table/dv): plumb cardinality validation through ReadDV


laskoviymishka:
Parent: #589.

`dv.ReadDV` currently passes `-1` to `DeserializeDV`, skipping cardinality validation entirely. The inline comment notes that `dvFile.Count()` defaults to `0` and would incorrectly reject valid DVs — true, but the puffin blob's properties carry `cardinality` as a string property (the spec-mandated source of truth), which `ReadDV` doesn't read. The result is that a truncated bitmap whose CRC happens to validate is accepted, even though the spec gives us the means to catch it.

Two design choices:

1. Read the `cardinality` blob property in `ReadDV` and pass it to `DeserializeDV` — most spec-faithful, since `cardinality` is what the writer explicitly attaches.
2. Use `dvFile.Count()` when it's positive — relies on the manifest entry's `record_count`, which Avro permits to default to zero.

Option 1 is cleaner: the property is always present on a spec-compliant blob and unambiguously means "deleted-row count for this DV". `DeserializeDV` already supports the validation path via its `expectedCardinality` parameter.

Touches `table/dv/deletion_vector.go` only.

PR #1053, opened 2026-05-10
feat(table): scanner deletion-vector read path

### LINKED ISSUE #996: feat(table): wire deletion-vector reader into the scanner
Parent: #589

Depends on #866 (DV bitmap reader). PR #855 already classifies DV files in scan planning behind an `ErrNotImplemented` guard, and #866 adds the `RoaringPositionBitmap` codec + Puffin `deletion-vector-v1` blob support. This issue is the integration: wire `ReadDV` into `readAllDeleteFiles` in `table/arrow_scanner.go` so DV positions merge into the existing position-delete pipeline. The rest of the scanner doesn't need to know whether a delete came from a Parquet position-delete file or a Puffin DV — both surface as positions.

Remove the `ErrNotImplemented` guard added in #855 once the merge works. A mixed-case test belongs here too: same data file with both a DV and a Parquet position-delete — Java's behavior is "DV wins"; mirror that.

End-to-end coverage: write data → write a puffin DV referencing those positions → scan → assert deleted rows are filtered. Spec: [Iceberg deletion vectors](https://iceberg.apache.org/spec/#deletion-vectors).

laskoviymishka:
Closes #996. Drops the ErrNotImplemented guard in GetRecords and wires DVs into the same per-file delete map Parquet pos-deletes already feed — recordsFromTask sees one set[int64] either way. Dedup is keyed by referenced data file (one puffin file can carry multiple blobs); distinct DVs for the same data file are rejected. PlanFiles also stops pairing pos-delete files with a data file when a DV applies, per spec.

Tests build the puffin in-process so this isn't blocked on #1041. 

Adjacent gaps deferred to follow-ups:
- #1048 (puffin DV header check)
- #1049 (ReadDV cardinality)
- #1050 (PlanFiles DV seq + multi-DV reject)
- #1051 (GetRecords arrow leak)
- #1052 (end-to-end DV scan test).

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -33,6 +33,7 @@ import (
 	"github.com/apache/iceberg-go"
 	iceinternal "github.com/apache/iceberg-go/internal"
 	iceio "github.com/apache/iceberg-go/io"
+	"github.com/apache/iceberg-go/table/dv"
 	"github.com/apache/iceberg-go/table/internal"
 	"github.com/apache/iceberg-go/table/substrait"
 	"github.com/substrait-io/substrait-go/v8/expr"
@@ -70,11 +71,8 @@ func releasePerFilePosDeletes(deletesPerFile perFilePosDeletes) {
 }
 
 func readAllDeleteFiles(ctx context.Context, fs iceio.IO, tasks []FileScanTask, concurrency int) (perFilePosDeletes, error) {
-	var (
-		deletesPerFile = make(perFilePosDeletes)
-		uniqueDeletes  = make(map[string]iceberg.DataFile)
-		err            error
-	)
+	deletesPerFile := make(perFilePosDeletes)
+	uniqueDeletes := make(map[string]iceberg.DataFile)
 
 	for _, t := range tasks {
 		for _, d := range t.DeleteFiles {
@@ -92,24 +90,33 @@ func readAllDeleteFiles(ctx context.Context, fs iceio.IO, tasks []FileScanTask, 
 		return deletesPerFile, nil
 	}
 
-	g, ctx := errgroup.WithContext(ctx)
+	g, gctx := errgroup.WithContext(ctx)
 	g.SetLimit(concurrency)
 
 	perFileChan := make(chan map[string]*arrow.Chunked, concurrency)
 	go func() {
+		// Inner g.Wait() gates the deferred channel close so a late worker
+		// can't send on a closed channel. Outer g.Wait() below collects the
+		// error — no cross-goroutine shared err variable.
 		defer close(perFileChan)
 		for _, v := range uniqueDeletes {
 			g.Go(func() error {
-				deletes, err := readDeletes(ctx, fs, v)
-				if deletes != nil {
-					perFileChan <- deletes
+				deletes, err := readDeletes(gctx, fs, v)
+				if err != nil {
+					return err
 				}
-
-				return err
+				if deletes == nil {
+					return nil
+				}
+				select {
+				case perFileChan <- deletes:
+					return nil
+				case <-gctx.Done():
+					return gctx.Err()
+				}
 			})
 		}
-
-		err = g.Wait()
+		_ = g.Wait()
 	}()
 
 	for deletes := range perFileChan {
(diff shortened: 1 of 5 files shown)


zeroshade (changes requested):
Would it be more efficient to instead convert the roaring bitmap into a standard bitmap, then use it as an arrow Boolean array to pass into the Filter function? Rather than computing the indices and using Take?

laskoviymishka:
Refactored this.

`readAllDeletionVectors` now returns `map[string]*dv.RoaringPositionBitmap`, so we no longer materialize DVs into `int64` positions.

I also added a new `filterByDeletionVector` pipeline step. It builds a per-batch Arrow boolean keep-mask using `Contains`, then applies `compute.Filter`.

The bitmap now threads through `recordBatchesFromTasksAndDeletes → recordsFromTask` as a separate map alongside `perFilePosDeletes`, so the two sources don’t need to share a representation.

table/dv/roaring_bitmap.go
```diff
@@ -91,6 +95,37 @@ func (b *RoaringPositionBitmap) Cardinality() int64 {
...
 
+// Iter yields the set positions in ascending order. Lazy — preferred over
+// ToArray when the caller can consume positions one-at-a-time (e.g. feeding
+// an Arrow builder), since a DV bitmap can hold millions of positions.
+//
+// The bitmap must not be modified while an iterator is running: the closure
+// captures b by pointer and the underlying roaring.Bitmap library does not
+// support concurrent Set/Iterator. Single-goroutine consumption is the only
+// supported pattern.
+func (b *RoaringPositionBitmap) Iter() iter.Seq[uint64] {
```
zeroshade: With the shift to using bitmaps directly, do we still need this? Is it unused? Same for ToArray?

table/arrow_scanner.go
```diff
@@ -193,6 +323,44 @@ func processPositionalDeletes(ctx context.Context, deletes set[int64]) recProces
...
+	return func(r arrow.RecordBatch) (arrow.RecordBatch, error) {
+		defer r.Release()
+
+		currentIdx := nextIdx
+		nextIdx += r.NumRows()
+
+		maskBuilder := array.NewBooleanBuilder(mem)
+		defer maskBuilder.Release()
+		maskBuilder.Reserve(int(r.NumRows()))
+		for i := int64(0); i < r.NumRows(); i++ {
```
zeroshade: Is there no more efficient way to get a standard bitmap from a roaring bitmap? Looks like you could call ToDense on the roaring bitmap once to convert it to a regular bitmap and create an array.Boolean, then just use slices of that array (`array.NewSlice(bm, currentIdx, currentIdx+r.NumRows())`) to pass into the Call to filter. 

Would simplify this and would reduce the allocations

table/arrow_scanner.go
```diff
@@ -193,6 +323,44 @@ func processPositionalDeletes(ctx context.Context, deletes set[int64]) recProces
...
+			// mask[i] = keep row i? → row i of the batch is at absolute
+			// position currentIdx+i in the source file; keep if NOT in
+			// the deletion vector.
+			maskBuilder.Append(!bitmap.Contains(uint64(currentIdx + i)))
+		}
+		mask := maskBuilder.NewBooleanArray()
+		defer mask.Release()
+
+		out, err := compute.Filter(ctx, compute.NewDatumWithoutOwning(r),
```
zeroshade: Can simplify with https://pkg.go.dev/github.com/apache/arrow-go/v18@v18.6.0/arrow/compute#FilterRecordBatch?

zeroshade (changes requested):
Almost there, just another couple nitpicks.

table/dv/roaring_bitmap.go
laskoviymishka: fixed, plus add some helper + tests

table/arrow_scanner.go
laskoviymishka: good catch, fixe

table/arrow_scanner.go
laskoviymishka: done

table/dv/roaring_bitmap.go
```diff
@@ -160,7 +161,59 @@ func DeserializeRoaringPositionBitmap(data []byte) (*RoaringPositionBitmap, erro
...
+// (e.g. when the file row count is below a stale upper bound).
+//
+// Bucket-key arithmetic is exact: each 32-bit bucket covers exactly 2^32
+// positions, i.e. 2^29 bytes, so per-bucket byte offsets are always
+// 8-byte-aligned and the inverted dense words land cleanly without bit shifts.
+func (b *RoaringPositionBitmap) KeepMaskBytes(length int64) []byte {
+	if length <= 0 {
+		return nil
+	}
+	out := make([]byte, (length+7)/8)
```
zeroshade: use https://pkg.go.dev/github.com/apache/arrow-go/v18@v18.6.0/arrow/bitutil#BytesForBits :smile:

table/dv/roaring_bitmap.go
```diff
@@ -160,7 +161,59 @@ func DeserializeRoaringPositionBitmap(data []byte) (*RoaringPositionBitmap, erro
...
+	for i := range out {
+		out[i] = 0xFF
+	}
```
zeroshade: use https://pkg.go.dev/github.com/apache/arrow-go/v18@v18.6.0/arrow/memory#Set which will leverage SIMD for this :smile:

table/dv/roaring_bitmap.go
```diff
@@ -160,7 +161,59 @@ func DeserializeRoaringPositionBitmap(data []byte) (*RoaringPositionBitmap, erro
...
+	// Clear bits past length-1 in the trailing byte so the mask is exactly
+	// `length` bits wide — Arrow's Boolean array honors the length field but
+	// callers inspecting the raw bytes shouldn't see phantom keep bits.
+	if extra := length & 7; extra != 0 {
+		out[len(out)-1] &= byte(1<<extra - 1)
+	}
```
zeroshade: use https://pkg.go.dev/github.com/apache/arrow-go/v18@v18.6.0/arrow/bitutil#SetBitsTo ?

table/dv/roaring_bitmap.go
```diff
@@ -160,7 +161,59 @@ func DeserializeRoaringPositionBitmap(data []byte) (*RoaringPositionBitmap, erro
...
+
+	for key, bm := range b.bitmaps {
+		bucketByteBase := int64(key) << 29 // (key << 32) bits / 8
+		if bucketByteBase >= int64(len(out)) {
+			continue
+		}
+		dense := bm.ToDense() // []uint64, LE bit-packed
```
zeroshade: is it actually LE bit-packed? The docs of the roaring bitmap package doesn't say that, so I would assume it's native-endian

table/dv/roaring_bitmap.go
```diff
@@ -160,7 +161,59 @@ func DeserializeRoaringPositionBitmap(data []byte) (*RoaringPositionBitmap, erro
...
+		for i, word := range dense {
```
zeroshade: Could this be simplified by using https://github.com/apache/arrow-go/blob/v18.6.0/arrow/bitutil/bitmaps.go#L330 and `PutNextWord(^word)` ?

zeroshade (changes requested):

zeroshade (approved):

PR #1055, opened 2026-05-10
feat(puffin): validate deletion-vector blob header invariants on read

### LINKED ISSUE #1048: feat(puffin): validate deletion-vector blob header invariants on read
Parent: #589.

Puffin spec `deletion-vector-v1` requires `snapshot-id` and `sequence-number` on blob metadata to both be `-1`. The writer in `puffin.Writer.AddBlob` enforces this on the way out. The reader's `validateBlobs` does not enforce the same invariant on the way in, so a malformed DV puffin written by a third-party tool would be accepted and its blob bytes would flow into `dv.DeserializeDV` without complaint. The bitmap CRC still catches byte-level corruption, but the metadata invariant is silently violated.

Extend `validateBlobs` with a per-blob check: when `Type == BlobTypeDeletionVector`, both `SnapshotID == -1` and `SequenceNumber == -1`, else return a clear error. One unit test with a hand-built puffin carrying the wrong values is sufficient; the envelope-shape fixtures used by `TestDeletionVectorPuffinWireFormat` are a good starting point.

Touches `puffin/` only — no dependency on any other deletion-vector work in flight.

laskoviymishka:
Closes #1048. Spec requires `deletion-vector-v1` blobs to carry snapshot-id and sequence-number set to -1. The writer enforces this on the way out; validateBlobs now mirrors the check on the way in, so a malformed DV puffin written by a third-party tool is rejected before its bytes reach dv.DeserializeDV.

Test byte-patches a non-DV blob to claim it's a deletion-vector and asserts the reader rejects both invariant violations; uses a same- length placeholder type so the footer-payload-size trailer stays accurate.

### DIFF
--- a/puffin/dv_header_validation_test.go
+++ b/puffin/dv_header_validation_test.go
@@ -1,0 +1,135 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package puffin_test
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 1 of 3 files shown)


zeroshade:
The spec requires them to be -1? Is this because the puffin file itself has the snapshot id and sequence number at that level instead?

zeroshade (approved):

laskoviymishka:
Yep, spec mandates -1 for `deletion-vector-v1`. A DV is bound to a data file (via `referenced-data-file`), not to a snapshot — its applicable seq number comes from the v3 manifest entry, not the blob. Other blob types like theta sketches do use those fields (they're per-snapshot stats), but for DVs Java's `BaseDVFileWriter` writes -1/-1 and `DeleteFileIndex` reads the seq number off the manifest. Forcing -1 keeps readers from accidentally picking up the wrong field.

PR #1059, opened 2026-05-10
feat(table): emit null current-snapshot-id for empty v3 metadata

### LINKED ISSUE #1002: feat(table): emit null current-snapshot-id for empty v3 metadata
Parent: #589

v3 metadata expects `"current-snapshot-id": null` when no snapshot exists, not the field omitted entirely (Java precedent: [apache/iceberg#12335](https://github.com/apache/iceberg/pull/12335)). The current marshaller uses an `omitempty` JSON tag on the `*int64` `CurrentSnapshotID` field, so when it's nil the key is dropped.

For v3, emit explicit `null` instead. Cleanest shape is a custom `MarshalJSON` on the v3 metadata struct that writes the `current-snapshot-id` key with a literal `null` when the pointer is nil. v1/v2 omitempty behavior should stay (existing test fixtures use `-1` for the v1/v2 sentinel).

Golden-fixture round-trip test against a Java-produced empty v3 `metadata.json` pins the output shape.

laskoviymishka:
Closes #1002. v3 spec requires `"current-snapshot-id": null` for empty tables, not the field omitted entirely (Java apache/iceberg#12335). omitempty on the *int64 field was dropping the key when nil; metadataV3 now has a MarshalJSON that re-injects the key with explicit null after the snapshots field — Java's declaration order — while preserving v1/v2 omitempty behaviour, since v1/v2 fixtures rely on the -1 sentinel.

The injection walks the marshaled JSON via json.Decoder so the surrounding field order is preserved (a map round-trip would resort alphabetically and break byte-level diffs against Java-produced metadata.json).

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2076,6 +2076,41 @@ func (m *metadataV3) UnmarshalJSON(b []byte) error {
 	return m.validate()
 }
 
+// MarshalJSON serializes v3 metadata with the spec-mandated emission of
+// `current-snapshot-id: null` for empty tables, rather than dropping the
+// key via the underlying `omitempty` tag (Java reference behaviour per
+// apache/iceberg#12335). The override re-declares `CurrentSnapshotID`
+// without omitempty in an outer struct that embeds an alias of
+// `metadataV3`; Go's encoding/json resolves the field-name collision in
+// favour of the shallower (outer) declaration, so the value is always
+// emitted — null when nil, the id otherwise.
+//
+// v1/v2 keep the omitempty path since their fixtures use the -1 sentinel
+// for "no current snapshot".
+//
+// `last-partition-id` is required for v2+ per spec — the existing parse-
+// time validation enforces this on read; the symmetric check below
+// rejects writes of malformed in-memory state (builder/parser paths
+// always populate it, so this only fires when code constructs metadataV3
+// directly and skips the builder).
+func (m *metadataV3) MarshalJSON() ([]byte, error) {
+	if m.LastPartitionID == nil {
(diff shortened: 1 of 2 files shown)


table/metadata.go
```diff
@@ -2006,6 +2007,145 @@ func (m *metadataV3) UnmarshalJSON(b []byte) error {
...
+// time validation enforces this on read; the symmetric check below
+// rejects writes of malformed in-memory state (builder/parser paths
+// always populate it, so this only fires when code constructs metadataV3
+// directly and skips the builder).
+func (m *metadataV3) MarshalJSON() ([]byte, error) {
+	if m.LastPartitionID == nil {
+		return nil, fmt.Errorf("%w: last-partition-id must be set for v3 metadata", ErrInvalidMetadata)
+	}
+
+	type Alias metadataV3
```
zeroshade, line 2103: Rather than decoding and injecting, the preferred way to do this would be for Alias to embed metadataV3 and then overwrite the field definition to remove the omitempty etc. That would be more efficient and avoid having to decode what we just encoded. 

i.e. something like

```go
alias := &struct {
    *metadataV3
    CurrentSnapshotID *int64 `json:"current-snapshot-id"`
}{ m, m.CurrentSnapshotID }
```

That should do what you're looking for without needing the extra decode/encode/injection etc.

You can see this pattern in other MarshalJSON functions we have defined

zeroshade (changes requested):

table/metadata.go
laskoviymishka: make sense, will do.

zeroshade (approved):
Just need to resolve the conflicts. LGTM

PR #1060, opened 2026-05-10
feat(puffin): enforce required DV blob properties on write

### LINKED ISSUE #1057: feat(puffin): enforce required DV blob properties on write
Parent: #589, follows #1049.

`puffin.Writer.AddBlob` already enforces the spec invariants that `snapshot-id` and `sequence-number` on a `deletion-vector-v1` blob must both be `-1`. The other spec-mandated DV invariants — that `properties` carry `cardinality` and `referenced-data-file` — are not enforced. A Go DV writer that forgets to attach either can produce non-conformant blobs the reader silently accepts via the spec-deviation tolerance path landed in #1049.

Extend the existing DV branch in `AddBlob`: when `input.Type == BlobTypeDeletionVector`, require `input.Properties["cardinality"] != ""` and `input.Properties["referenced-data-file"] != ""`. Surface a clear error for each missing property. A couple of unit subtests in `puffin_test.go`'s existing DV-error coverage close the loop.

Symmetric with the reader-side check landed in #1048 (`validateBlobs`). Closes the gap that prompted the reader-side `slog.Warn` fallback in #1049 — once writers can't omit the property, the warning path stops being reachable from in-tree writes and only fires for cross-impl reads of files written by other tools.

laskoviymishka:
Closes #1057. Puffin spec requires `deletion-vector-v1` blobs to carry both `cardinality` and `referenced-data-file` properties. AddBlob already enforces the snapshot-id/sequence-number invariants for DV blobs; extend the same branch to reject blobs missing either property.

Closes the gap that prompted the reader-side slog.Warn fallback in #1049 — once writers can't omit the property, the warning path stays unreachable from in-tree writes and only fires on cross-impl reads of files written by third-party tools.

Two existing test fixtures that wrote DV blobs without the cardinality property are updated: TestRoundTrip in puffin_test.go and writePuffinWithDVBlob in dv's deletion_vector_test.go.

### DIFF
--- a/puffin/puffin_writer.go
+++ b/puffin/puffin_writer.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"strconv"
 
 	"github.com/apache/iceberg-go"
 )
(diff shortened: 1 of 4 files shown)


puffin/puffin_writer.go
```diff
@@ -137,6 +138,28 @@ func (w *Writer) AddBlob(input BlobMetadataInput, data []byte) (BlobMetadata, er
...
+		// them today, so we hold the line here to keep Go-emitted files
+		// always conformant.
+		if input.Properties["cardinality"] == "" {
+			return BlobMetadata{}, errors.New("puffin: deletion-vector-v1 requires a cardinality property")
+		}
+		// Reject non-numeric or negative values at write time too —
+		// otherwise a writer could emit "cardinality": "abc" that the
+		// reader hard-rejects when it parses the property. Symmetric
+		// with strconv.ParseInt + non-negative check on the read side.
+		card, err := strconv.ParseInt(input.Properties["cardinality"], 10, 64)
```
zeroshade: Could get two birds with one stone by using `ParseUint` instead which would error on a negative value. So you could just do something like

```go
if _, err := strconv.ParseUint(input.Properties["cardinality"], 10, 64); err != nil {
        // got something other than a non-negative integer
}
```

zeroshade (changes requested):

puffin/puffin_writer.go
laskoviymishka: Good catch, refactored this to `ParseUint`.

For `"-1"`, the leading `-` now fails as invalid syntax before any value is parsed, so non-numeric and negative values go through the same error path. No separate negative check needed.

I also cleaned up `dv/deletion_vector_test.go`: the old reader-side “negative cardinality rejected” and “malformed property rejected” subtests can’t build their fixtures anymore, because the writer now rejects them at `AddBlob` time. I removed those and left a pointer to the writer-side coverage in `puffin_test.go`.

The reader-side checks in `blobCardinality` still stay, since we can still hit malformed files written by third-party writers.

zeroshade (approved):

PR #1061, opened 2026-05-11
feat(io): add ManagedIdentityCredential support for Azure

alessandro-nori:
## Changes
- Add `AdlsManagedIdentityEnabled` (`adls.auth.managed-identity.enabled`) property to opt into `ManagedIdentityCredential` instead of `DefaultAzureCredential`
- Add `AdlsManagedIdentityClientID` (`adls.auth.managed-identity.client-id`) property for user-assigned managed identities

## Motivation
`DefaultAzureCredential` sets a short timeout on its first managed identity attempt to avoid long hangs during local development. In production, this causes "managed identity timed out" errors when the app requests a token before the hosting environment is ready. Using `ManagedIdentityCredential` directly avoids this timeout.

### DIFF
--- a/io/azure.go
+++ b/io/azure.go
@@ -38,12 +38,14 @@ var adlsURIPattern = regexp.MustCompile(`^(abfss?|wasbs?)://([^/?#]+)(.*)?$`)
 
 // Constants for Azure configuration options
 const (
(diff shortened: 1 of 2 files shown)


alessandro-nori:
Closing — recreating from updated main.

PR #1062, opened 2026-05-11
feat(io): add ManagedIdentityCredential support for Azure

alessandro-nori:
## Changes
- Add `ADLSManagedIdentityEnabled` (`adls.auth.managed-identity.enabled`) property to opt into `ManagedIdentityCredential` instead of `DefaultAzureCredential`
- Add `ADLSManagedIdentityClientID` (`adls.auth.managed-identity.client-id`) property for user-assigned managed identities

## Motivation
`DefaultAzureCredential` sets a short timeout on its first managed identity attempt. In production, this causes frequent "managed identity timed out" errors when the app requests a token. Using `ManagedIdentityCredential` directly avoids this timeout.

Reference: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/TROUBLESHOOTING.md#troubleshoot-defaultazurecredential-authentication-issues

### DIFF
--- a/io/config.go
+++ b/io/config.go
@@ -43,12 +43,14 @@ const (
 
 // Constants for Azure configuration options
 const (
-	ADLSSasTokenPrefix         = "adls.sas-token."
-	ADLSConnectionStringPrefix = "adls.connection-string."
-	ADLSSharedKeyAccountName   = "adls.auth.shared-key.account.name"
-	ADLSSharedKeyAccountKey    = "adls.auth.shared-key.account.key"
-	ADLSEndpoint               = "adls.endpoint"
-	ADLSProtocol               = "adls.protocol"
+	ADLSSasTokenPrefix          = "adls.sas-token."
+	ADLSConnectionStringPrefix  = "adls.connection-string."
+	ADLSSharedKeyAccountName    = "adls.auth.shared-key.account.name"
+	ADLSSharedKeyAccountKey     = "adls.auth.shared-key.account.key"
+	ADLSEndpoint                = "adls.endpoint"
+	ADLSProtocol                = "adls.protocol"
+	ADLSManagedIdentityEnabled  = "adls.auth.managed-identity.enabled"
+	ADLSManagedIdentityClientID = "adls.auth.managed-identity.client-id"
 
 	// Not in use yet
 	// ADLSReadBlockSize          = "adls.read.block-size-bytes"
--- a/io/config.go
+++ b/io/config.go
@@ -43,14 +43,14 @@ const (
 
 // Constants for Azure configuration options
 const (
-	ADLSSasTokenPrefix          = "adls.sas-token."
-	ADLSConnectionStringPrefix  = "adls.connection-string."
-	ADLSSharedKeyAccountName    = "adls.auth.shared-key.account.name"
-	ADLSSharedKeyAccountKey     = "adls.auth.shared-key.account.key"
-	ADLSEndpoint                = "adls.endpoint"
-	ADLSProtocol                = "adls.protocol"
-	ADLSManagedIdentityEnabled  = "adls.auth.managed-identity.enabled"
-	ADLSManagedIdentityClientID = "adls.auth.managed-identity.client-id"
+	ADLSSasTokenPrefix         = "adls.sas-token."
+	ADLSConnectionStringPrefix = "adls.connection-string."
+	ADLSSharedKeyAccountName   = "adls.auth.shared-key.account.name"
+	ADLSSharedKeyAccountKey    = "adls.auth.shared-key.account.key"
+	ADLSClientID               = "adls.client-id"
+	ADLSEndpoint               = "adls.endpoint"
+	ADLSProtocol               = "adls.protocol"
+	ADLSManagedIdentityEnabled = "adls.auth.managed-identity.enabled"
 
 	// Not in use yet
 	// ADLSReadBlockSize          = "adls.read.block-size-bytes"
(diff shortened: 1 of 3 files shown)


zeroshade:
@alessandro-nori is this ready for review? Or still a draft?

alessandro-nori:
> @alessandro-nori is this ready for review? Or still a draft?

@zeroshade I didn't have time to test it yet but I'm planning to do it before the end of the week

io/config.go
```diff
@@ -43,12 +43,14 @@ const (
...
-	ADLSEndpoint               = "adls.endpoint"
-	ADLSProtocol               = "adls.protocol"
+	ADLSSasTokenPrefix          = "adls.sas-token."
+	ADLSConnectionStringPrefix  = "adls.connection-string."
+	ADLSSharedKeyAccountName    = "adls.auth.shared-key.account.name"
+	ADLSSharedKeyAccountKey     = "adls.auth.shared-key.account.key"
+	ADLSEndpoint                = "adls.endpoint"
+	ADLSProtocol                = "adls.protocol"
+	ADLSManagedIdentityEnabled  = "adls.auth.managed-identity.enabled"
+	ADLSManagedIdentityClientID = "adls.auth.managed-identity.client-id"
```
laskoviymishka: These property names don't exist in Java's `AzureProperties`, PyIceberg's `pyiceberg/io/__init__.py`, or iceberg-rust's `azdls.rs` — they're Go-only. A REST catalog vending creds from Java will never emit them, so this branch is effectively unreachable in any cross-client deployment.

For the client-id case, both PyIceberg and Rust already use `adls.client-id`. I'd reuse that key rather than introduce `adls.auth.managed-identity.client-id` as a second name for the same field.

io/gocloud/azure.go
```diff
@@ -159,6 +159,28 @@ func createAzureBucket(ctx context.Context, parsed *url.URL, props map[string]st
 		if err != nil {
 			return nil, fmt.Errorf("failed container.NewClientFromConnectionString: %w", err)
 		}
+	} else if props[io.ADLSManagedIdentityEnabled] == "true" {
```
laskoviymishka, line 162: Since `azidentity` v1.8, setting `AZURE_TOKEN_CREDENTIALS=ManagedIdentityCredential` makes `DefaultAzureCredential` use only the MI credential in its chain (with `dac: false`), and `AZURE_CLIENT_ID` already drives user-assigned identity selection. That's runtime-equivalent to this branch, with no new code and no Go-only properties.

If the production case is "on AKS, skip the IMDS-probe slowness of DAC", the env-var path already solves it. If the real driver is per-table identity isolation in the same process, keeping the code is justified — but I'd want that motivation stated explicitly. wdyt?

io/gocloud/azure_test.go
```diff
@@ -73,6 +73,26 @@ func TestCreateAzureBucketDefaultCredentialEmptyBucketName(t *testing.T) {
...
+	bucket, err := createAzureBucket(ctx, parsedURL, map[string]string{
+		"adls.auth.managed-identity.enabled": "true",
+	})
+	assert.NoError(t, err)
+	assert.NotNil(t, bucket)
+
+	iter := bucket.List(nil)
+	_, err = iter.Next(ctx)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "ManagedIdentityCredential",
```
laskoviymishka, line 92: `"ManagedIdentityCredential"` appears in errors from both `ManagedIdentityCredential` and `DefaultAzureCredential` — DAC contains MI in its chain and propagates the credential name through. So this test (and its `TestCreateAzureBucketDefaultCredentialCalled` neighbour) passes whether the new branch was taken or not.

The cleanest fix is a small test seam — accept an optional credential constructor in `createAzureBucket` so the test can inject a stub and assert on the type. Short of that, at minimum assert the error does **not** contain `"DefaultAzureCredential"` so the two tests can't both pass on the same code path.

laskoviymishka (changes requested):
Thanks for bringing this in!

Overall look good, but couple things needed before merge:

- The new property names are Go-only: Java's AzureProperties, PyIceberg, and iceberg-rust don't define them, and PyIceberg/Rust already use `adls.client-id` for user-assigned identity. And since azidentity v1.8, `AZURE_TOKEN_CREDENTIALS=ManagedIdentityCredential` is runtime-equivalent to this new branch, so the code path may not need to exist at all. Worth settling that before locking in public constants.
- the new test doesn't actually distinguish MI from DAC (the credential name appears in both errors), so it passes regardless of which branch ran.

Once this is adressed - happy to merge.

io/config.go
alessandro-nori: > A REST catalog vending creds from Java will never emit them, so this branch is effectively unreachable in any cross-client deployment

These properties are not used in the context of credentials vending but when the client has already access to the container, so it's not true that the branch is unreachable.

Ok for the `client-id` key

io/gocloud/azure.go
alessandro-nori: I didn't know, I'll test the env variable in our prod environment. If that works, I can close the PR.

But the concern about the short timeout is the same

alessandro-nori: The motivation is explained in the PR description `DefaultAzureCredential sets a short timeout on its first managed identity attempt` and as suggested by the Azure troubleshooting [documentation](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/TROUBLESHOOTING.md#troubleshoot-defaultazurecredential-authentication-issues) we should use `ManagedIdentityCredential` directly in production

> Use [ManagedIdentityCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#ManagedIdentityCredential) directly, at least in production. It doesn't set a timeout on its authentication attempts.

laskoviymishka (approved):
🚢 👍

alessandro-nori:
@laskoviymishka I think the Github actions didn't run because GitHub is having an outage 
https://www.githubstatus.com/

Issue #957, opened 2026-04-29
[Proposal] feat(cli): maintenance commands - tracking issue


tanmayrauth:
### Feature Request / Improvement

Design doc of the proposal: [[link to doc](https://docs.google.com/document/d/1PHX2GAxUTuVL9rfizWuzKG5QFt0D8Mi-k76ziFsHjf4/edit?usp=sharing)]

## Summary                                                                                                                                                                                                                                                                                                      
                                                            
  Tracking issue for adding maintenance commands to the `iceberg` CLI.                                                                                                                                                                                                                 
                                                            
  This would make iceberg-go the first Iceberg implementation with a maintenance CLI,  Spark uses `CALL` procedures, pyiceberg's CLI is read-only.                                                                                                                                                                
                                                            
  **11 commands** across 6 work items:                                                                                                                                                                                                                                                                            
                                                            
  ## Plan                                                                                                                                                                                                                                                                                                         
                                                            
  ### 1. Shared infrastructure                                                                                                                                                                                                                                                                                    
  - [x] Duration parser (`7d` → `168h`, days only — weeks/months ambiguous)
  - [x] `--yes` confirmation prompt (TTY-aware; errors in non-interactive mode without `--yes`)                                                                                                                                                                                                                   
  - [x] Dry-run output contract (`[DRY RUN]` prefix in text, `"dry_run": true` in JSON, exit 0)                                                                                                                                                                                                                   
  - [x] `--output json|text` support for all new commands                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                  
  ### 2. Read-only commands `good-first-issue`                                                                                                                                                                                                                                                                    
  - [x] `info TABLE`                                                                                                                                                                                                                                                                                              
  - [x] `snapshots TABLE`                                                                                                                                                                                                                                                                                         
  - [x] `refs TABLE [--type REFTYPE]`                                                                                                                                                                                                                                                                             
  - [x] `partition-stats TABLE [--snapshot-id ID] [--all]`                                                                                                                                                                                                                                                        
  - [x] `schema TABLE --show-defaults`                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                  
  ### 3. Snapshot maintenance                                                                                                                                                                                                                                                                                     
  - [x] `expire-snapshots TABLE [--older-than DURATION] [--retain-last N] [--dry-run] [--yes]`                                                                                                                                                                                                                    
[...]

PR #1065, opened 2026-05-11
feat(cli): add partition-stats command

tanmayrauth:
Add `iceberg partition-stats TABLE_ID` to list partition statistics files with snapshot ID, path, and size. Supports --snapshot-id filter and --all flag.

Related: #957
Depends On: #1073

### DIFF
--- a/cmd/iceberg/partition_stats.go
+++ b/cmd/iceberg/partition_stats.go
@@ -19,17 +19,87 @@ package main
 
 import (
 	"context"
-	"errors"
+	"encoding/json"
 	"os"
+	"strconv"
 
 	"github.com/apache/iceberg-go/catalog"
 	"github.com/apache/iceberg-go/table"
+	"github.com/pterm/pterm"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/iceberg/main.go
```diff
@@ -286,6 +287,9 @@ func main() {
 		runProperties(ctx, output, cat, args.Properties)
 	case args.Compact != nil:
 		runCompact(ctx, output, cat, args.Compact)
+	case args.PartStats != nil:
```
laskoviymishka, line 317: this actually make this per file decomposition quite hard to iterate independently, can we decouple each command from main.go?

laskoviymishka (commented):
All PR-s are touching main.go, it would be painfull to merge, need either decouple them from main.go or squash into one signle PR.

cmd/iceberg/main.go
tanmayrauth: Makes sense, I will check to decouple them.

tanmayrauth:
@laskoviymishka  @zeroshade  I've rebased the branch, now all the PRs are independent changes.

laskoviymishka (approved):

PR #1067, opened 2026-05-11
feat(cli): add --show-defaults flag to schema command

tanmayrauth:
Enhance `iceberg schema TABLE_ID --show-defaults` to display initial-default and write-default values alongside field definitions.

Related: #957
Depends On: #1073


laskoviymishka (approved):
LGTM

PR #1068, opened 2026-05-11
feat(cli): add branch create and tag create commands

tanmayrauth:
Add `iceberg branch create TABLE_ID NAME` and `iceberg tag create TABLE_ID NAME` with snapshot ref retention options.

Related: #957
Depends On: #1073

### DIFF
--- a/cmd/iceberg/branch_tag.go
+++ b/cmd/iceberg/branch_tag.go
@@ -47,7 +47,22 @@ func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cm
 	tbl := loadTable(ctx, output, cat, cmd.TableID)
 	meta := tbl.Metadata()
 
+	for name := range meta.Refs() {
+		if name == cmd.BranchName {
+			output.Error(fmt.Errorf("ref %q already exists", cmd.BranchName))
+			os.Exit(1)
+		}
+	}
+
 	snapshotID := resolveSnapshotID(output, tbl, cmd.SnapshotID)
+
+	if err := confirmAction(
+		fmt.Sprintf("Create branch %q on %s at snapshot %d?", cmd.BranchName, tableIDString(tbl), snapshotID),
+		cmd.Yes,
+	); err != nil {
+		output.Error(err)
+		os.Exit(1)
+	}
 
 	var maxRefAgeMs int64
 	if cmd.MaxRefAge != "" {
@@ -78,26 +93,55 @@ func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cm
 
 	update := table.NewSetSnapshotRefUpdate(cmd.BranchName, snapshotID, table.BranchRef,
 		maxRefAgeMs, maxSnapshotAgeMs, minSnapshotsToKeep)
-	reqs := []table.Requirement{table.AssertTableUUID(meta.TableUUID())}
+	reqs := []table.Requirement{
+		table.AssertTableUUID(meta.TableUUID()),
+		table.AssertRefSnapshotID(cmd.BranchName, nil),
+	}
 
 	if _, _, err := cat.CommitTable(ctx, tbl.Identifier(), reqs, []table.Update{update}); err != nil {
-		output.Error(fmt.Errorf("create branch failed: %w", err))
+		output.Error(fmt.Errorf("failed to create branch: %w", err))
 		os.Exit(1)
 	}
 
-	output.RefCreated(RefCreatedResult{
+	result := RefCreatedResult{
 		Table:      tableIDString(tbl),
 		RefName:    cmd.BranchName,
 		RefType:    string(table.BranchRef),
 		SnapshotID: snapshotID,
-	})
+	}
+	if maxRefAgeMs > 0 {
+		result.MaxRefAgeMs = &maxRefAgeMs
+	}
+	if maxSnapshotAgeMs > 0 {
+		result.MaxSnapshotAgeMs = &maxSnapshotAgeMs
+	}
+	if minSnapshotsToKeep > 0 {
+		result.MinSnapshotsToKeep = &minSnapshotsToKeep
+	}
+
+	output.RefCreated(result)
 }
 
 func runTagCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *TagCreateCmd) {
 	tbl := loadTable(ctx, output, cat, cmd.TableID)
 	meta := tbl.Metadata()
 
+	for name := range meta.Refs() {
+		if name == cmd.TagName {
+			output.Error(fmt.Errorf("ref %q already exists", cmd.TagName))
+			os.Exit(1)
+		}
+	}
+
 	snapshotID := resolveSnapshotID(output, tbl, cmd.SnapshotID)
+
+	if err := confirmAction(
+		fmt.Sprintf("Create tag %q on %s at snapshot %d?", cmd.TagName, tableIDString(tbl), snapshotID),
+		cmd.Yes,
+	); err != nil {
+		output.Error(err)
+		os.Exit(1)
+	}
 
 	var maxRefAgeMs int64
 	if cmd.MaxRefAge != "" {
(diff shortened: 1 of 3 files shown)


cmd/iceberg/branch_tag.go
```diff
@@ -19,21 +19,135 @@ package main
...
+	}
+
+	var minSnapshotsToKeep int
+	if cmd.MinSnapshotsToKeep != nil {
+		minSnapshotsToKeep = *cmd.MinSnapshotsToKeep
+	}
+
+	update := table.NewSetSnapshotRefUpdate(cmd.BranchName, snapshotID, table.BranchRef,
+		maxRefAgeMs, maxSnapshotAgeMs, minSnapshotsToKeep)
+	reqs := []table.Requirement{table.AssertTableUUID(meta.TableUUID())}
```
laskoviymishka: Requirements slice is just `AssertTableUUID`, so `branch create` silently overwrites an existing ref — `iceberg branch create db.t main --snapshot-id 999` rewrites `main`. Two-part fix:

1. Pre-flight check `meta.Refs()` for `cmd.BranchName` and error out with `ref %q already exists` if present.
2. Add `table.AssertRefSnapshotID(cmd.BranchName, nil)` to `reqs` so REST catalogs reject concurrent creates server-side too.

Java's `UpdateSnapshotReferencesOperation.createBranch` uses `Preconditions.checkArgument(existingRef == null, ...)`; PyIceberg's `create_branch` emits `AssertRefSnapshotId(snapshot_id=None, ref=ref_name)`. Same applies to `runTagCreate` below (line 115).

cmd/iceberg/branch_tag.go
```diff
@@ -19,21 +19,135 @@ package main
...
+
+	if _, _, err := cat.CommitTable(ctx, tbl.Identifier(), reqs, []table.Update{update}); err != nil {
+		output.Error(fmt.Errorf("create branch failed: %w", err))
```
laskoviymishka: `"create branch failed"` (and `"create tag failed"` at line 118) inverts the verb/object order used everywhere else in this package — `"failed to create table"`, `"failed to parse properties"`, etc. Suggested: `fmt.Errorf("failed to create branch: %w", err)` for grep consistency.

cmd/iceberg/branch_tag.go
```diff
@@ -19,21 +19,135 @@ package main
...
+func runTag(ctx context.Context, output Output, cat catalog.Catalog, cmd *TagCmd) {
+	switch {
+	case cmd.Create != nil:
+		runTagCreate(ctx, output, cat, cmd.Create)
+	}
 }
 
-func (textOutput) RefCreated(_ RefCreatedResult) {}
-func (jsonOutput) RefCreated(_ RefCreatedResult) {}
+func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *BranchCreateCmd) {
```
laskoviymishka, line 46: `BranchCreateCmd` and `TagCreateCmd` are the only write commands in the series without a `Yes bool` / `confirmAction` gate. Once the existence check above is added the destructive surface shrinks, but the pattern asymmetry will surprise the next contributor. Add `Yes bool \`help:"skip confirmation prompt"\`` to both structs in `maintenance.go` and call `confirmAction("Create branch %s on %s?", cmd.Yes)` before `CommitTable`.

cmd/iceberg/branch_tag.go
```diff
@@ -19,21 +19,135 @@ package main
...
+		Table:      tableIDString(tbl),
+		RefName:    cmd.TagName,
+		RefType:    string(table.TagRef),
+		SnapshotID: snapshotID,
+	})
+}
+
+func resolveSnapshotID(output Output, tbl *table.Table, explicit *int64) int64 {
+	if explicit != nil {
+		return *explicit
```
laskoviymishka, line 189: When `explicit != nil` the value is returned without checking it exists in the metadata. The catalog will reject it eventually, but the resulting error is a wrapped REST failure rather than the clear `snapshot %d not found` you'd get pre-flight. `meta.SnapshotByID` is already used elsewhere in this file; one extra check here is consistent with how the rollback PR (#1071) handles the same case.

cmd/iceberg/branch_tag.go
```diff
@@ -19,21 +19,135 @@ package main
...
+	if _, _, err := cat.CommitTable(ctx, tbl.Identifier(), reqs, []table.Update{update}); err != nil {
+		output.Error(fmt.Errorf("create branch failed: %w", err))
+		os.Exit(1)
+	}
+
+	output.RefCreated(RefCreatedResult{
+		Table:      tableIDString(tbl),
+		RefName:    cmd.BranchName,
+		RefType:    string(table.BranchRef),
+		SnapshotID: snapshotID,
```
laskoviymishka, line 110: `RefCreatedResult` (defined in `maintenance.go`) emits only `table`, `ref_name`, `ref_type`, `snapshot_id` — not the retention values that were committed. A script consuming the JSON has no way to verify what `max_ref_age_ms`, `max_snapshot_age_ms`, `min_snapshots_to_keep` actually landed. Echo them back in the result struct (with `omitempty` so the tag path stays clean).

cmd/iceberg/branch_tag_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	output := buf.String()
+	assert.Contains(t, output, "Created tag")
+	assert.Contains(t, output, "v1.0")
+	assert.Contains(t, output, "3000")
+}
+
+func TestJSONOutputRefCreated(t *testing.T) {
+	oldStdout := os.Stdout
+	r, w, _ := os.Pipe()
```
laskoviymishka: `r, w, _ := os.Pipe()` drops the error. If pipe creation fails, `w` is nil and `os.Stdout = w` corrupts the process for subsequent tests. Suggested:

```go
r, w, err := os.Pipe()
require.NoError(t, err)
oldStdout := os.Stdout
os.Stdout = w
t.Cleanup(func() { w.Close(); os.Stdout = oldStdout })
```

cmd/iceberg/branch_tag_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	tbl := table.New([]string{"db", "tbl"}, meta, "", nil, nil)
+
+	result := resolveSnapshotID(textOutput{}, tbl, nil)
+	assert.Equal(t, int64(5000), result)
+}
+
+func TestTextOutputRefCreated(t *testing.T) {
+	var buf bytes.Buffer
+	pterm.SetDefaultOutput(&buf)
```
laskoviymishka, line 80: `pterm.SetDefaultOutput(&buf)` mutates package-global state without restore (same at line 102). Sequential today, data-race on parallel tomorrow. Add `t.Cleanup(func() { pterm.SetDefaultOutput(os.Stderr); pterm.EnableColor() })`.

laskoviymishka (changes requested):
This one still needs a bit of work.

My main concern is that `branch create` and `tag create` currently behave more like upserts than creates. They both call `cat.CommitTable` with `NewSetSnapshotRefUpdate` and only `AssertTableUUID`, so something like:

```bash
iceberg branch create db.t main --snapshot-id 999
```

can silently rewrite `main`.

Java rejects existing refs before creating, and PyIceberg also sends a server-side requirement so concurrent creates fail. I’d suggest matching that pattern:

* pre-flight check `meta.Refs()` for a clear local error
* add `table.AssertRefSnapshotID(name, nil)` to the requirements so REST catalogs enforce create-only semantics under races

Related: these are the only write commands in the series without a `--yes` / `confirmAction` gate. Since `confirmAction` already exists from `maintenance.go`, I’d add it here too for consistency.

The tag retention flags look correct in the CLI, so the remaining spec issue is library-internal: tags shouldn’t support `min-snapshots-to-keep` / `max-snapshot-age-ms`. That can be a follow-up guard in `table/metadata.go`.

Small nits:

* error style: prefer `failed to create branch` over `create branch failed`, to match the rest of the package
* `RefCreatedResult` JSON omits the committed retention values, so scripts can’t verify what actually landed
* `resolveSnapshotID` should pre-flight that an explicit snapshot ID actually exists

Test coverage looks solid for the current surface. Docker REST catalog integration is still the shared gap for the broader series.

tanmayrauth:
I've addressed the below points:


  1. Create-only semantics — Added pre-flight check against meta.Refs() for a clear local error, plus   AssertRefSnapshotID(name, nil) in the requirements slice for server-side race protection. Matches Java/PyIceberg behavior.
  2. Confirmation gate — Added --yes to both BranchCreateCmd and TagCreateCmd, with confirmAction called before CommitTable.                                                                                                                                                                                                
  3. Error style — Fixed to "failed to create branch" / "failed to create tag".                                                                                                                                                                                                                                             
  4. Snapshot ID validation — resolveSnapshotID now calls meta.SnapshotByID and errors with snapshot %d not found when the explicit ID doesn't exist.                                                                                                                                                                       
  5. Retention in JSON result — RefCreatedResult now includes max_ref_age_ms, max_snapshot_age_ms, min_snapshots_to_keep (with omitempty).                                                                                                                                                                                  
  6. Test fixes — os.Pipe() error is checked with require.NoError, pterm global state restored via t.Cleanup.

laskoviymishka (approved):

PR #1069, opened 2026-05-11
feat(table): gate v3-only metadata fields on sub-v3 reads (#1006)

### LINKED ISSUE #1006: feat(table): gate v3-only metadata fields on sub-v3 reads
Parent: #589

v3-only fields like `next-row-id` and `encryption-keys` are silently accepted when they appear in a v1 or v2 `metadata.json`. This masks corrupt fixtures and cross-version bugs. Java rejects with a descriptive error per field; iceberg-go should match.

Branch on `format-version` during metadata parse and error per forbidden field with a message naming the field and the format version it requires, e.g. `"v3-only field 'next-row-id' present in v2 metadata"`. The set of v3-only fields today is `next-row-id` and `encryption-keys`; if more land later, the same gate covers them.

One unit test per gated field — load a v2 metadata.json with the v3-only field injected, assert a clear error.

tanmayrauth:
Reject next-row-id and encryption-keys during v1/v2 metadata parsing with a descriptive error naming the field and required format version.

Closes: #1006

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1369,14 +1369,15 @@ type v3FieldCheck struct {
 }
 
 func rejectV3OnlyFields(version int, checks ...v3FieldCheck) error {
+	var errs []error
 	for _, c := range checks {
 		if c.present {
-			return fmt.Errorf("%w: v3-only field '%s' present in v%d metadata",
-				ErrInvalidMetadataFormatVersion, c.name, version)
+			errs = append(errs, fmt.Errorf("%w: v3-only field '%s' present in v%d metadata",
+				ErrInvalidMetadataFormatVersion, c.name, version))
 		}
 	}
 
-	return nil
+	return errors.Join(errs...)
 }
 
 // ParseMetadata parses json metadata provided by the passed in reader,
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1363,15 +1363,17 @@ var (
 	ErrPartitionSpecNotFound        = errors.New("partition spec not found")
 )
 
-func rejectV3OnlyFields(version int, nextRowID *int64, encryptionKeys []EncryptionKey) error {
-	if nextRowID != nil {
-		return fmt.Errorf("%w: v3-only field 'next-row-id' present in v%d metadata",
-			ErrInvalidMetadata, version)
-	}
+type v3FieldCheck struct {
+	name    string
+	present bool
+}
 
-	if len(encryptionKeys) > 0 {
-		return fmt.Errorf("%w: v3-only field 'encryption-keys' present in v%d metadata",
-			ErrInvalidMetadata, version)
+func rejectV3OnlyFields(version int, checks ...v3FieldCheck) error {
+	for _, c := range checks {
+		if c.present {
+			return fmt.Errorf("%w: v3-only field '%s' present in v%d metadata",
+				ErrInvalidMetadataFormatVersion, c.name, version)
+		}
 	}
 
 	return nil
@@ -1920,7 +1922,10 @@ func (m *metadataV1) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	if err := rejectV3OnlyFields(1, aux.NextRowID, aux.EncryptionKeyList); err != nil {
+	if err := rejectV3OnlyFields(aux.FormatVersion,
+		v3FieldCheck{"next-row-id", aux.NextRowID != nil},
+		v3FieldCheck{"encryption-keys", len(aux.EncryptionKeyList) > 0},
+	); err != nil {
 		return err
 	}
 
@@ -1990,7 +1995,10 @@ func (m *metadataV2) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	if err := rejectV3OnlyFields(2, aux.NextRowID, aux.EncryptionKeyList); err != nil {
+	if err := rejectV3OnlyFields(aux.FormatVersion,
+		v3FieldCheck{"next-row-id", aux.NextRowID != nil},
+		v3FieldCheck{"encryption-keys", len(aux.EncryptionKeyList) > 0},
+	); err != nil {
 		return err
 	}
 
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1363,6 +1363,20 @@ var (
 	ErrPartitionSpecNotFound        = errors.New("partition spec not found")
 )
 
+func rejectV3OnlyFields(version int, nextRowID *int64, encryptionKeys []EncryptionKey) error {
+	if nextRowID != nil {
+		return fmt.Errorf("%w: v3-only field 'next-row-id' present in v%d metadata",
+			ErrInvalidMetadata, version)
+	}
+
+	if len(encryptionKeys) > 0 {
+		return fmt.Errorf("%w: v3-only field 'encryption-keys' present in v%d metadata",
+			ErrInvalidMetadata, version)
+	}
+
+	return nil
+}
+
 // ParseMetadata parses json metadata provided by the passed in reader,
 // returning an error if one is encountered.
 func ParseMetadata(r io.Reader) (Metadata, error) {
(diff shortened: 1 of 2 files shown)


tanmayrauth:

table/metadata.go
```diff
@@ -1324,6 +1324,20 @@ var (
 	ErrPartitionSpecNotFound        = errors.New("partition spec not found")
 )
 
+func rejectV3OnlyFields(version int, nextRowID *int64, encryptionKeys []EncryptionKey) error {
```
laskoviymishka: Two concerns stacked here.

First, `version` is only used in the error message and callers pass `1` and `2` as literals — no validation, no typed constant. Either drop the parameter (the field name already carries the diagnostic weight) or read `aux.FormatVersion` to avoid the literal-drift risk.

Second, this signature won't scale. V3 has more fields landing (row-lineage tracking, etc.) and each new gate means adding a parameter, updating both call sites, and adding four new tests. A variadic `rejectV3OnlyField(version, name, present)` called once per field, or a slice of `{name, present}` checks, keeps the signature stable as V3 grows.

table/metadata.go
```diff
@@ -1867,6 +1881,10 @@ func (m *metadataV1) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
+	if err := rejectV3OnlyFields(1, aux.NextRowID, aux.EncryptionKeyList); err != nil {
```
laskoviymishka: This is the main concern, and it applies symmetrically to the v2 call site at line 1954.

Java's `TableMetadataParser.fromJson` doesn't error on these fields in v1/v2 — it reads `next-row-id` only when `formatVersion >= 3` (defaulting to `INITIAL_ROW_ID` otherwise) and has no version guard on `encryption-keys` at all. PyIceberg silently drops unknown fields via Pydantic's model boundary; iceberg-rust does the same via serde. A downgrade tool that produces v2 metadata with a populated `encryption-keys` is readable by every other client and rejected only by Go.

If the intent here is genuinely to be stricter than the reference (worth confirming in #1006 discussion), at minimum the sentinel should be `ErrInvalidMetadataFormatVersion` (or a new dedicated `ErrV3FieldInPreV3Metadata`) rather than `ErrInvalidMetadata` — callers `errors.Is`-ing on `ErrInvalidMetadata` today expect "corrupt file," not "forward-compat version mismatch."

Related: `last-sequence-number` is a v2+ field on the shared `commonMetadata` and is silently ignored on v1 reads today. If the project keeps strict-rejection here, the same logic should apply there for consistency — otherwise the gate isn't principled. (Probably a follow-up rather than blocking.)

table/metadata_internal_test.go
```diff
@@ -1854,3 +1854,91 @@ func TestZstdGoldenFixture(t *testing.T) {
...
+		"last-partition-id": 1000,
+		"sort-orders": [{"order-id": 0, "fields": []}],
+		"default-sort-order-id": 0,
+		"encryption-keys": [{"key-id": "k1", "encrypted-key-metadata": "abc123"}]
+	}`
+
+	_, err := ParseMetadataBytes([]byte(data))
+	require.ErrorIs(t, err, ErrInvalidMetadata)
+	assert.ErrorContains(t, err, "v3-only field 'encryption-keys' present in v2 metadata")
+}
```
laskoviymishka: Four structurally identical functions with copy-pasted JSON bodies; the only variation is version, field name, and expected substring. Collapse into one table-driven `TestRejectV3OnlyFields(t *testing.T)` with subtests and a small JSON-builder helper — each new V3 field becomes a one-line table entry, which matters more once the helper grows (see comment on metadata.go:1327).

Also missing two cases worth adding:
1. A positive `TestV3AcceptsV3OnlyFields` confirming v3 metadata containing `encryption-keys` parses without error — `ExampleTableMetadataV3` doesn't cover this, so the "v3 is OK" path is currently untested.
2. A `TestV1AcceptsEmptyEncryptionKeys` covering `encryption-keys: []` — the `len(...) > 0` guard is specifically designed for that boundary but nothing exercises it, so a refactor to `encryptionKeys != nil` would silently break interop without test failure.

laskoviymishka (changes requested):
The goal makes sense, catching v3-only fields leaking into v1/v2 metadata is a real correctness concern.

But I’m a bit skeptical about the strict reject here, because it seems to diverge from the other Iceberg clients. Java reads `next-row-id` only when `formatVersion >= 3` and otherwise defaults it. `encryption-keys` does not seem to have a version guard. PyIceberg drops unknown fields through Pydantic, and iceberg-rust does the same through serde.

So with this change, Go becomes the only client that errors on these fields. For example, a v3 → v2 downgrade with stray `encryption-keys` would still be readable by Java / Python / Rust, but unreadable by Go. That may be the right call, but I think it’s worth an explicit project-level decision before we make Go stricter than the reference behavior.

If we do want to be stricter, I’d at least use `ErrInvalidMetadataFormatVersion` rather than `ErrInvalidMetadata`. Callers usually read `ErrInvalidMetadata` as “corrupt file,” while this is more “field does not match format version.”

Design-wise, `rejectV3OnlyFields` taking each field as a typed parameter feels like it won’t scale well. V3 has more fields coming, and every new gate would require changing the helper signature, both call sites, and adding more flat tests. A variadic / slice-based helper would be easier to extend.

The tests could also be simpler as one table-driven subtest, and I think we’re missing a positive v3 case for `encryption-keys`.

tanmayrauth:
Adressed the below things:

                                                                                                                                                                                                                                                                                                                            
  - Switched to ErrInvalidMetadataFormatVersion — agreed that ErrInvalidMetadata implies corruption
  - Refactored to variadic ...v3FieldCheck so new fields are a one-line addition
  - Call sites now read aux.FormatVersion instead of literal 1/2                                                                                                                                                                                                                                                            
  - Tests collapsed into a single table-driven subtest, added the v3+encryption-keys positive case and the empty encryption-keys: [] boundary case                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                            
On the strictness question : the spec explicitly scopes these fields to v3, so my thinking was that surfacing the mismatch early (with a clear, distinguishable error) helps consumers catch misconfigured metadata closer to the source. In the v3→v2 downgrade scenario, ideally the downgrade tool would strip v3  fields - silently accepting them downstream could mask that gap for Go consumers.   What do you think?

tanmayrauth:
@laskoviymishka  does this PR looks fine now?

table/metadata.go
```diff
@@ -1363,6 +1363,22 @@ var (
...
+type v3FieldCheck struct {
+	name    string
+	present bool
+}
+
+func rejectV3OnlyFields(version int, checks ...v3FieldCheck) error {
+	for _, c := range checks {
+		if c.present {
+			return fmt.Errorf("%w: v3-only field '%s' present in v%d metadata",
+				ErrInvalidMetadataFormatVersion, c.name, version)
```
zeroshade: should we use `errors.Join` so that we can report *all* the fields we're rejecting at once instead of piece-meal?

tanmayrauth: Done. I changed rejectV3OnlyFields to collect all violations and report them via errors.Join instead of returning on the first one. Added a test case ("reports all rejected fields") that verifies both next-row-id and  encryption-keys are mentioned in the error when both are present simultaneously.

laskoviymishka (approved):
Looks good to me now. I’d like to track two follow-ups, neither blocking.

The bigger one is cross-client behavior. I read through `TableMetadataParser.java` while reviewing this, and Java doesn’t reject these fields: `next-row-id` is only read when `formatVersion >= 3` and otherwise defaults to `INITIAL_ROW_ID`; `encryption-keys` has no version guard at all. PyIceberg and iceberg-rust also silently drop unknown fields. So Go is the only client hard-erroring here.

That doesn’t hurt today because v3 isn’t widely deployed, but once downgrade tooling exists, a v2 file with stray `encryption-keys` would read everywhere except Go. It’s worth deciding before v3 GA whether we want to keep that strictness. I think it’s defensible — catching mis-versioned metadata early has real value — but if we keep it, a short code comment explaining the intentional divergence would help future readers.

Smaller one: `last-sequence-number` is v2+ on the shared `commonMetadata`, but is silently accepted on v1 reads today. If we’re gating v3-only fields on sub-v3 reads, the same logic naturally extends to v2-only fields on v1. 

Worth a follow-up issue to keep the gate principled.

PR #1070, opened 2026-05-11
feat(table): port remaining v3 metadata-upgrade validation rules

### LINKED ISSUE #1013: feat(table): port remaining v3 metadata-upgrade validation rules
Parent: #589

Java [apache/iceberg#10861](https://github.com/apache/iceberg/pull/10861) added a ruleset validating v1→v3 / v2→v3 upgrades — format-version monotonically increases, required v3 fields appear, last-column-id is preserved, and so on. iceberg-go has a subset of these checks today; this issue tracks porting the remainder.

Step one is enumerating Java's rules (a structured walk through `MetadataValidators` or equivalent) into a checklist in this issue body, then auditing each against iceberg-go's existing validation surface. Step two is adding the missing rules with a unit test per rule (each rule has a clear failure case in the Java suite that ports cleanly).

Scope is decomposable: the audit can land first as a checklist update on this issue, then individual rules can ship as small follow-up PRs without one big monolithic change.

tanmayrauth:
Add missing validation rules to SetFormatVersion that mirror Java's TableMetadata upgrade checks:

- Initialize lastSequenceNumber to 0 when upgrading from v1 to v2+
- Initialize nextRowID to 0 when upgrading to v3
- Assign UUID if not already set when upgrading to v2+
- Validate lastColumnId is not less than the highest field ID in schemas

Validation runs before state mutation to prevent leaving the builder in a half-upgraded state on failure.

Closes #1013

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -708,8 +708,47 @@ func (b *MetadataBuilder) SetFormatVersion(formatVersion int) error {
 		return nil
 	}
 
+	if err := b.validateLastColumnID(); err != nil {
+		return err
+	}
+
+	previousVersion := b.formatVersion
 	b.updates = append(b.updates, NewUpgradeFormatVersionUpdate(formatVersion))
 	b.formatVersion = formatVersion
+
+	if previousVersion < 2 && formatVersion >= 2 {
+		if b.uuid == (uuid.UUID{}) {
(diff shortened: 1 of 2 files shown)


tanmayrauth:

laskoviymishka (approved):
nice a clean!

PR #1075, opened 2026-05-12
feat: add wire codec helpers for DataFile and FileScanTask

twuebi:
Adds on-the-wire encoding for DataFile and FileScanTask to enable distributed compaction. With #1033, we decomposed compaction into a coordinator and worker portion. This enables plumbing them together over the wire. FileScanTask need to go coordinator -> worker (data file, delete files, scan range + v3 row lineage) and DataFile from worker -> coordinator collecting them for the commit.

iceberg.EncodeDataFile / DecodeDataFile reuse the manifest-entry Avro encoding so the bytes on the wire are the same bytes a manifest carries. The dataFile struct's avro tags remain the single source of truth. Adding a field to DataFile extends what the helpers transport, with no wire mirror to keep in sync. Encoding is non-mutating and thread-safe: a fresh *dataFile is cloned via reflection over avro tags, so the source is untouched.

table.EncodeFileScanTask / DecodeFileScanTask layer on top: each embedded DataFile is iceberg-encoded, then wrapped alongside the scan range and v3 row lineage in a small Avro envelope.

Design notes:
- The receiver supplies (spec, schema, version) out of band. Both sides in the distributed-compaction design already hold table metadata, and the per-(specID, version) avro schema is cached. Happy to switch to a self-describing payload if preferred.
- distinct_counts round-trips on v1/v2 (thanks to #1044, which this PR is rebased on); v3 omits it per spec (deprecated in v3, apache/iceberg#12182). v3 callers that need it must transport it separately.
- Reflection runs once per encode and is not in a hot path I can see. Happy to precompute the avro field index at init if preferred.
- An anonymous `var _ = FileScanTask{...}` literal next to the codec is a compile-time drift guard: adding/retyping/reordering a FileScanTask field breaks the build, forcing a deliberate call on whether it must cross the wire.

Format versions 1, 2, and 3 are supported. No change to existing manifest read/write paths.

Tests: round-trip across v1/v2/v3 with fully populated DataFiles, foreign-impl rejection, partition-data idempotence, and FileScanTask shape including v3 row lineage.

cc @laskoviymishka, this continues building blocks for distributed compaction


zeroshade:
Should this be moved into the compaction sub package of table instead of being in the table package? 

We need to get better at utilizing sun packages to improve the API surface of each package and prevent them from getting too large.

twuebi:
@zeroshade I was considering a `wire` package somewhere, this is not just relevant for compaction I think?

laskoviymishka:
I think `wire` package is a bit odd name, but i'm also believe movign this in separate package is a good idea.

twuebi:
moved to a `codec/` package, @laskoviymishka

data_file_codec.go
```diff
@@ -0,0 +1,219 @@
...
+	return out
+}
+
+type dataFileFieldMaps struct {
+	nameToID      map[string]int
+	idToType      map[int]string
+	idToFixedSize map[int]int
+}
+
+type dataFileSchemaCacheKey struct {
```
laskoviymishka, line 231: This cache key collides across tables. specID is per-table — `InitialPartitionSpecID = 0` means essentially every table created via `NewPartitionSpec` has specID 0. Two such tables with different partition column types share the same key in this process-global `sync.Map`: whichever encodes first wins, and the second table silently gets back the first table's Avro partition shape. Bytes round-trip but `Partition()` comes out mis-typed (e.g. `int64` decoded as `int32`) or gets dropped. The doc on `manifestEntrySchemaFor` even acknowledges the dependency — *"apart from how it shapes the partition struct"* — but the key omits the thing that shapes the struct.

Including schema id handles the realistic case:

```go
type dataFileSchemaCacheKey struct {
    specID   int
    schemaID int
    version  int
}
```

A stricter alternative is to fingerprint `spec.PartitionType(schema)`, but schema id is usually enough in practice. `TestManifestEntrySchemaForCaches` needs reworking too — as written it locks in the collision behavior. wdyt?

codec/data_file.go
```diff
@@ -0,0 +1,87 @@
...
+// EncodeDataFile is non-mutating and safe to call concurrently with
+// any other reader or encoder of the same DataFile.
+//
+// distinct_counts round-trips on v1 and v2. The v3 manifest-entry
+// schema omits the field (deprecated in the v3 spec, see
+// apache/iceberg#12182), so it does not survive encode→decode on v3 —
+// callers on v3 that need distinct counts must transport them
+// separately.
+func EncodeDataFile(df iceberg.DataFile, spec iceberg.PartitionSpec, schema *iceberg.Schema, version int) ([]byte, error) {
+	m, ok := df.(avroEntryMarshaler)
```
laskoviymishka: `iceberg.DataFile` is a public interface and external packages do implement it (test doubles, alternate file stores). Rejecting them with "requires the iceberg package's DataFile implementation" names the constraint but not how to satisfy it — the caller has no path forward short of giving up the interface.

I'd either expose the marshaler interface publicly so external impls can satisfy it, or narrow the signature to a concrete type that only this package can produce. Hidden runtime constraints on a public interface tend to bite people who skim the signature and don't read the doc.

codec/file_scan_task_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+			require.Len(t, decoded.DeleteFiles, len(original.DeleteFiles))
+			for i := range original.DeleteFiles {
+				require.Equal(t, original.DeleteFiles[i].FilePath(), decoded.DeleteFiles[i].FilePath())
+			}
+			require.Len(t, decoded.EqualityDeleteFiles, len(original.EqualityDeleteFiles))
+			for i := range original.EqualityDeleteFiles {
+				require.Equal(t, original.EqualityDeleteFiles[i].FilePath(), decoded.EqualityDeleteFiles[i].FilePath())
+			}
+			require.Len(t, decoded.DeletionVectorFiles, len(original.DeletionVectorFiles))
```
laskoviymishka, line 54: This assertion is `0 == 0` — `fullyPopulatedFileScanTask` never sets `DeletionVectorFiles`. DV files are the v3 first-class delete mechanism, so the whole encode/decode path for them is currently unexercised. Any regression here passes silently.

I'd add at least one DV file to the v3 fixture (`EntryContentPosDeletes` + `ReferencedDataFile`) and assert `FilePath()` / `ReferencedDataFile()` equality, mirroring the other two delete-file loops above.

data_file_codec.go
```diff
@@ -0,0 +1,219 @@
...
+// avro-tagged fields. Internal state (sync.Once, lazy-init caches,
+// specID, the field-id lookup maps) is intentionally left at zero
+// values because the avro encoder reads only the avro-tagged fields.
+//
+// Using reflection over the tag set means a new avro-tagged field
+// upstream is auto-copied without an update here — the dataFile struct
+// remains the single source of truth for the wire shape. It also
+// sidesteps the go-vet copies-lock warning that would fire on a
+// struct-literal copy of *dataFile (it embeds sync.Once).
+func cloneDataFileAvroFields(src *dataFile) *dataFile {
```
laskoviymishka, line 188: This is a shallow copy of pointer-typed fields. Things like `ColSizes *[]colMap[int, int64]` get their pointer copied — clone and source point at the same backing slice. `TestMarshalAvroEntryDoesNotMutate` only checks `PartitionData` (which is replaced wholesale before encoding), so the column-stats fields aren't covered. The thread-safety claim then rests on `twmb/avro` never writing to the slices it's handed during encode — fragile, since that's library-internal.

Either document the no-mutation guarantee as covering `PartitionData` only, or extend the test to snapshot the pointer-typed fields before/after a concurrent encode and assert equality. Also worth confirming `TestEncodeDataFileConcurrent` runs under `-race` in CI — without it, byte equality across goroutines doesn't actually validate the absence of a data race.

data_file_codec.go
```diff
@@ -0,0 +1,219 @@
...
+//
+// UnmarshalAvroDataFileEntry is the low-level avro primitive used by
+// the [github.com/apache/iceberg-go/codec] package; callers performing
+// cross-process transport should use that package's high-level API
+// rather than calling this function directly.
+//
+// The returned DataFile carries the partition spec id and the field-id
+// lookup tables, so Partition() and the stats accessors return id-keyed
+// maps as if the file had been read from a manifest.
+func UnmarshalAvroDataFileEntry(data []byte, spec PartitionSpec, schema *Schema, version int) (DataFile, error) {
```
laskoviymishka: The encode side is `MarshalAvroEntry` (method on the unexported `*dataFile`); the decode side is exported as a top-level function — and the doc explicitly tells callers not to use it directly. If we don't want callers to reach this, I'd unexport it and have `codec.DecodeDataFile` use it via an internal bridge. Cleaner than exporting a function whose own doc says don't call it.

codec/data_file_test.go
```diff
@@ -0,0 +1,257 @@
...
+		ValueCounts(map[int]int64{1: 1024, 2: 1024}).
+		NullValueCounts(map[int]int64{1: 0, 2: 4}).
+		NaNValueCounts(map[int]int64{1: 0, 2: 0}).
+		LowerBoundValues(map[int][]byte{1: {0x01}, 2: []byte("a")}).
+		UpperBoundValues(map[int][]byte{1: {0xff}, 2: []byte("z")}).
+		SplitOffsets([]int64{0, 4096}).
+		SortOrderID(0).
+		KeyMetadata([]byte("kms-key-1"))
+	if version < 3 {
+		builder.DistinctValueCounts(map[int]int64{1: 64, 2: 128})
```
laskoviymishka, line 193: `distinct_counts` (field 111) is marked "Deprecated. Do not write." in the spec for all versions, not just v3. Encoded bytes never go to a manifest file here, so it's not a correctness bug — but framing v1/v2 distinct-count round-trip as a feature in this test invites future callers to set it on new DataFiles. I'd reframe as a read-compatibility artifact: counts already present on a DataFile (from an old manifest read) survive the round trip, but new builder usage shouldn't set them.

data_file_codec.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return df, nil
+}
+
+// newEncodeEntry returns the right manifest-entry shape for the schema
+// version: v1's manifest_entry has a non-nullable snapshot_id and uses
+// [fallbackManifestEntry], v2/v3 use [manifestEntry] with nullable
+// pointers.
+func newEncodeEntry(version int, df *dataFile) any {
+	if version == 1 {
```
laskoviymishka, line 150: The v1 entry is built with `Snapshot: 0` (the zero value, since `fallbackManifestEntry.Snapshot` is a non-nullable `int64`). That's fine for the wire-transport use case here, but it means the bytes are not usable as a standalone v1 manifest entry — anyone who feeds them through a regular manifest reader gets snapshot id 0. Worth a doc note on `EncodeDataFile`: v1 bytes are only meaningful via `DecodeDataFile` and carry no snapshot context.

codec/data_file_test.go
```diff
@@ -0,0 +1,257 @@
...
+	require.Equal(t, want.SpecID(), got.SpecID())
+	if version < 3 {
+		require.Equal(t, want.DistinctValueCounts(), got.DistinctValueCounts())
+	} else {
+		require.Empty(t, got.DistinctValueCounts(),
+			"v3 manifest-entry schema omits distinct_counts (deprecated in spec); "+
+				"see internal/avro_schemas.go data_file_v3")
+	}
+	if version >= 2 {
+		require.Equal(t, want.ContentType(), got.ContentType())
```
laskoviymishka: `content` is an avro-tagged field on `dataFile` and lives in the v1 encoded bytes, but `ContentType()` is only asserted on `version >= 2`. A regression in v1 content-type encoding would slip past. Moving the assertion outside the gate is a one-line fix.

codec/file_scan_task.go
```diff
@@ -0,0 +1,191 @@
...
+// Each carried DataFile is encoded with [EncodeDataFile] and wrapped in
+// a small record that also carries the scan range and v3 row lineage.
+// The (spec, schema, version) triple must match what [DecodeFileScanTask]
+// is given on the receiver.
+func EncodeFileScanTask(task table.FileScanTask, spec iceberg.PartitionSpec, schema *iceberg.Schema, version int) ([]byte, error) {
+	fileBytes, err := EncodeDataFile(task.File, spec, schema, version)
+	if err != nil {
+		return nil, fmt.Errorf("file: %w", err)
+	}
+	del, err := encodeDataFileSlice(task.DeleteFiles, spec, schema, version)
```
laskoviymishka, line 49: All three delete-file kinds get encoded with the caller-supplied `(spec, schema, version)`. After partition evolution, a position-delete file can carry a different specID than the data file. Most deployments won't see this, but worth either reading specID off each `DataFile` and looking it up against table metadata, or at minimum a comment calling out the assumption.

codec/file_scan_task.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/apache/iceberg-go/table"
+	"github.com/twmb/avro"
+)
+
+// EncodeFileScanTask encodes a FileScanTask for cross-process transport.
+// Each carried DataFile is encoded with [EncodeDataFile] and wrapped in
+// a small record that also carries the scan range and v3 row lineage.
+// The (spec, schema, version) triple must match what [DecodeFileScanTask]
+// is given on the receiver.
+func EncodeFileScanTask(task table.FileScanTask, spec iceberg.PartitionSpec, schema *iceberg.Schema, version int) ([]byte, error) {
```
laskoviymishka, line 41: `EncodeFileScanTask` and `DecodeFileScanTask` don't validate `version` before delegating, so an invalid value surfaces as `"file: iceberg: MarshalAvroEntry: unsupported format version 0"` — readable but the function name in the error chain doesn't match the caller's entry point. The same `if version < 1 || version > 3` guard at the top of both, with a `codec:`-prefixed error, keeps things consistent with the rest of the package.

laskoviymishka (changes requested):
The codec itself looks well scoped, and the docs are clear about what is self-describing vs what still needs out-of-band context like spec, schema, or version. The `FileScanTask` struct-conversion drift guard is also a neat trick.

I’d hold before merging though. 

Main blocker:

`dataFileSchemaCache` in `data_file_codec.go` is a process-global `sync.Map` keyed by `(specID, version)`. But `specID` is per-table and is usually `0` for tables created via `NewPartitionSpec`.

So two tables can both have `specID=0`, but different partition column types. In that case they collide in the cache: the first encoder populates it, and the second one silently gets the wrong Avro shape back. Then partition values can come out mis-typed or get dropped.

That feels like a happy-path bug for the main codec use case: coordinator / worker handling multiple tables. Including the schema ID in the cache key should fix it.

Two more things I’d fix in this PR rather than follow up:

1. `EncodeDataFile` type-asserts to an unexported `avroEntryMarshaler` and rejects anything else with “requires the iceberg package’s DataFile implementation”. But `iceberg.DataFile` is public, so external implementations, including test doubles, have no way to work with this API. I’d either expose the marshaler interface or accept a concrete type.

2. The `FileScanTask` round-trip test checks `DeletionVectorFiles` length, but the fixture never actually has DV files. Since DV files are the v3 first-class delete mechanism, this encoding path should be covered before merge.

Smaller comments left inline.

One extra thing to confirm: does `TestEncodeDataFileConcurrent` run under `-race` in CI? Without race mode, byte equality across goroutines doesn’t really prove the thread-safety claim.

Once these are addressed, happy to take another pass.

data_file_codec.go
twuebi: Tests for `/codec` now run under `-race`, decided to make it per package opt in to not blow up CI time too much.

data_file_codec.go
twuebi: added the bridge

codec/file_scan_task_test.go
twuebi: added a delete file

data_file_codec.go
twuebi: went for the stricter variant and added a test

codec/data_file.go
```diff
@@ -0,0 +1,93 @@
...
+// DecodeDataFile decodes bytes produced by [EncodeDataFile] back into a
+// DataFile. The (spec, schema, version) triple must match the encoder;
+// passing a different spec or version yields a decode error or silently
+// mis-typed partition values.
+//
+// The returned DataFile carries the partition spec id and the field-id
+// lookup tables, so Partition() and the stats accessors return id-keyed
+// maps as if the file had been read from a manifest.
+func DecodeDataFile(data []byte, spec iceberg.PartitionSpec, schema *iceberg.Schema, version int) (iceberg.DataFile, error) {
+	res, err := datafileavro.Unmarshal(data, spec, schema, version)
```
laskoviymishka, line 101: Two robustness items at this call site. First, `datafileavro.Unmarshal` is a bare package variable; if a test binary ever imports `codec` without indirectly triggering `iceberg.init()`, the call panics with a nil function pointer and no diagnostics. A `if datafileavro.Unmarshal == nil` guard would make the init-order dependency explicit. Second, `res.(iceberg.DataFile)` on line 92 is unchecked — the two-value form with an `fmt.Errorf("codec: bridge returned unexpected type %T", res)` keeps the failure path informative if the bridge's return contract ever changes. wdyt?

data_file_codec.go
```diff
@@ -0,0 +1,260 @@
...
+	"sync"
+
+	"github.com/apache/iceberg-go/internal"
+	"github.com/apache/iceberg-go/internal/datafileavro"
+	"github.com/twmb/avro"
+)
+
+func init() {
+	datafileavro.Unmarshal = func(data []byte, spec, schema any, version int) (any, error) {
+		return unmarshalAvroDataFileEntry(data, spec.(PartitionSpec), schema.(*Schema), version)
```
laskoviymishka: Same defensive point on the writer side of the bridge: `spec.(PartitionSpec)` and `schema.(*Schema)` are unguarded. With `any`-typed parameters there's no compile-time enforcement; any future caller passing mismatched types turns into a panic with no context. Checked assertions + a wrapped error keeps the failure observable.

codec/file_scan_task.go
```diff
@@ -0,0 +1,205 @@
...
+
+func init() {
+	s, err := avro.Parse(fileScanTaskSchemaJSON)
+	if err != nil {
+		panic("codec: fileScanTaskSchema invalid: " + err.Error())
+	}
+	fileScanTaskSchema = s
+}
+
+func encodeDataFileSlice(files []iceberg.DataFile, spec iceberg.PartitionSpec, schema *iceberg.Schema, version int) ([][]byte, error) {
```
laskoviymishka, line 175: The cross-spec point from round 1 didn't fully land. The doc on `EncodeFileScanTask` tells callers to group delete files by per-file specID before calling — but `table.FileScanTask` doesn't expose the per-file specs in a form that supports doing that grouping at the call site without reaching back into table metadata that isn't part of the codec API. A one-liner in `encodeDataFileSlice` (`if df.SpecID() != spec.ID() { return nil, fmt.Errorf(...) }`) converts the silent partition-data corruption path into an explicit error and matches the contract the doc already promises. wdyt?

codec/data_file.go
```diff
@@ -0,0 +1,93 @@
...
+// is written as 0 by the iceberg implementation. v1 bytes are not
+// usable as a standalone manifest entry — they only round-trip via
+// [DecodeDataFile].
+//
+// distinct_counts (field 111) is deprecated in the spec for all
+// versions. Already-set values round-trip on v1 and v2 as a
+// read-compatibility artifact; v3 omits the field entirely
+// (apache/iceberg#12182). New DataFiles should not set distinct
+// counts.
+func EncodeDataFile(df iceberg.DataFile, spec iceberg.PartitionSpec, schema *iceberg.Schema, version int) ([]byte, error) {
```
laskoviymishka, line 70: Two parallel exports for the same operation: `iceberg.EncodeDataFile` in the root package, `codec.EncodeDataFile` here. The wrapper delegates to `MarshalAvroEntry` and surfaces errors like `"iceberg: MarshalAvroEntry: unsupported format version 0"` — exposing the internal symbol to `codec` callers, while `EncodeFileScanTask` in the same package uses `"codec: EncodeFileScanTask: ..."`. Picking one canonical entry point and aligning the error prefix would tidy this.

codec/file_scan_task_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"testing"
+
+	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/codec"
+	"github.com/apache/iceberg-go/table"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEncodeDecodeFileScanTaskRoundTrip(t *testing.T) {
+	for _, version := range []int{2, 3} {
```
laskoviymishka: v1 isn't in the round-trip range here. The v1 encode path uses `fallbackManifestEntry` with a non-nullable `snapshot_id` written as 0 — the most divergent branch in the pipeline, currently no integration coverage at the `FileScanTask` envelope level. Adding `1` to the version slice with a positional-deletes-only fixture would pin it.

data_file_codec.go
```diff
@@ -0,0 +1,260 @@
...
+var dataFileSchemaCache sync.Map
+
+// manifestEntrySchemaFor returns the cached avro schema and partition
+// field-id lookups for the given partition type and format version.
+// The cache key uses [StructType.String] as a structural fingerprint
+// of the partition type, so two specs that produce different partition
+// column types cache under different keys even when they share an id.
+func manifestEntrySchemaFor(spec PartitionSpec, schema *Schema, version int) (*avro.Schema, dataFileFieldMaps, error) {
+	partType := spec.PartitionType(schema)
+	key := dataFileSchemaCacheKey{partTypeFingerprint: partType.String(), version: version}
```
laskoviymishka: Nit on the cache key: `StructType.String()` appends field doc strings into the fingerprint. Two structurally identical partition specs with different documentation produce equivalent-but-distinct Avro schema entries — unbounded fragmentation in metadata-rich environments. The Avro shape from `partitionTypeToAvroSchema` itself would be a tighter key, but happy to leave as a follow-up.

data_file_codec.go
```diff
@@ -0,0 +1,260 @@
...
+	"fmt"
+	"reflect"
+	"sync"
+
+	"github.com/apache/iceberg-go/internal"
+	"github.com/apache/iceberg-go/internal/datafileavro"
+	"github.com/twmb/avro"
+)
+
+func init() {
```
laskoviymishka, line 37: Nit: a once-set guard would catch a future split of init responsibility from silently overwriting the bridge. `if datafileavro.Unmarshal != nil { panic("iceberg: datafileavro.Unmarshal already set") }` is the standard pattern for these hooks — fails loudly at startup instead of letting the second writer win.

laskoviymishka (changes requested):
Nice! Almost there, a few items left, mostly around defensive coding at the codec boundary.

Let's close this smallish items (mostly one liners):
 
- Three unchecked type assertions / one nil-pointer at the bridge that turn caller mistakes into panics instead of errors. Comments inline at codec/data_file.go and data_file_codec.go.
- Cross-spec delete file validation — the doc workaround from round 1 requires per-file spec lookup that the codec API doesn't surface, so let's catch the mismatch instead of letting partition data silently corrupt. One-liner suggestion inline.

If you up to - its nice to have:

- iceberg.EncodeDataFile and codec.EncodeDataFile are both public, with slightly inconsistent error prefixes. Pick one canonical entry point.
  
Also - worth to filling couple of follow-ups:

- v1 path missing from the FileScanTask round-trip
- Cache fingerprint includes field doc strings (cache fragmentation)
- Once-set guard on the bridge writer

twuebi:
@laskoviymishka addressed comments & also the follow-ups already in this PR

codec/data_file.go
twuebi: kept the `iceberg:` prefix on `MarshalAvroEntry` since it's the primitive on `*dataFile`, codec validates version upfront so the `codec.EncodeDataFile` path never surfaces it, and a direct caller of `MarshalAvroEntry` is in iceberg-land.

data_file_codec.go
```diff
@@ -0,0 +1,298 @@
...
+
+	return c
+}
+
+// SetSchemaCacheSize resizes the manifest-entry schema cache used by
+// the DataFile codec. The default capacity is sized for a few thousand
+// active partition specs; long-running consumers with larger working
+// sets (e.g. a compaction service touching many tables) should raise
+// it. Existing entries are preserved on grow; on shrink, least-recently
+// used entries are evicted down to the new size. Not safe to call
```
laskoviymishka, line 257: the underlying `lru/v2` cache acquires its internal mutex on `Resize` the same way it does on `Get`/`Add`, so calling `SetSchemaCacheSize` concurrently with ongoing codec operations is actually safe. A reader following this warning would either add external serialization that isn't needed, or avoid resizing at runtime entirely. 

I'd just drop the sentence or flip it to "Safe to call concurrently; the underlying cache is mutex-protected.", can be done after merge.

codec/file_scan_task.go
```diff
@@ -0,0 +1,208 @@
...
+// file's SpecID is validated and a mismatch returns an error. After
+// partition evolution, delete files may have been written under a
+// different partition spec than the data file; the caller is
+// responsible for partitioning the FileScanTask by per-file specID and
+// calling EncodeFileScanTask once per group.
+func EncodeFileScanTask(task table.FileScanTask, spec iceberg.PartitionSpec, schema *iceberg.Schema, version int) ([]byte, error) {
+	if version < 1 || version > 3 {
+		return nil, fmt.Errorf("codec: EncodeFileScanTask: unsupported format version %d", version)
+	}
+	fileBytes, err := EncodeDataFile(task.File, spec, schema, version)
```
laskoviymishka, line 45: small asymmetry: `encodeDataFileSlice` checks `int(f.SpecID()) != spec.ID()` for every delete / equality / DV file, but `task.File` goes through unchecked. A transposed `(spec, task)` pair would silently mis-encode the primary file's partition values while the delete files cleanly error out. A one-liner mirroring the slice path above this call closes the asymmetry, but fine to do it later.

codec/file_scan_task.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("file: %w", err)
```
laskoviymishka, line 47: these wraps strip the function-name prefix (`"file: ..."`, `"delete files: ..."`, `"decode: ..."`) while `EncodeDataFile` / `DecodeDataFile` consistently use `"codec: EncodeDataFile: ..."`. Final error chain ends up like `delete files: entry 0: codec: EncodeDataFile: ...` with no top-level marker for which envelope produced it. Worth aligning next time you're touching this file.

laskoviymishka (approved):
Looks good to me. All the questions covered: the bridge now has the once-set guard and checked type assertions, `encodeDataFileSlice` has explicit cross-spec validation, the schema cache fingerprint uses the Avro shape, v1 is covered in the `FileScanTask` round-trip, and the public `iceberg.AvroEntryMarshaler` replaces the unexported workaround.

Two small nits, both fine as follow-up PRs:

1. `SetSchemaCacheSize` godoc says it’s not safe to call concurrently with codec operations, but `lru/v2.Resize` takes the same mutex as `Get` / `Add`, so it should actually be safe. A false warning here may make callers add unnecessary synchronization.

2. `EncodeFileScanTask` validates `SpecID` on every delete / equality / DV file, but not on `task.File`. A mismatched `(spec, task)` pair could silently mis-encode the primary file’s partition values, while the delete files correctly error out. A one-line check before `EncodeDataFile(task.File, ...)` would mirror the slice path.

:shipit:

PR #1076, opened 2026-05-13
Skip DELETED tombstones in DeleteOrphanFiles' referenced set

dhananjaykrutika:
DeleteOrphanFiles builds "referenced files" set by walking each live snapshot's manifests, fetching every entry with discardDeleted=false, and unconditionally marking entry as referenced. This treats DELETED-status entries (tombstones) the same as ADDED/EXISTING entries, so any data file that had ever been overwritten or deleted in a copy-on-write operation remains referenced by the surviving snapshot's tombstone manifest. After every snapshot that actually held the file live is expired, the file is still reachable and orphan cleanup skips it.

This is a deviation from iceberg-java and iceberg-python implementations.

Switch manifest.Entries() to use discardDeleted=true so tombstones are excluded before they reach the reference-collection loop.

### DIFF
--- a/table/orphan_cleanup.go
+++ b/table/orphan_cleanup.go
@@ -268,7 +268,13 @@ func (t Table) getReferencedFiles(fs iceio.IO) (map[string]bool, error) {
 		for _, manifest := range manifestFiles {
 			referenced[manifest.FilePath()] = true
 
-			for entry, err := range manifest.Entries(fs, false) {
+			// discardDeleted=true: skip DELETED-status entries when
+			// computing the reachable file set. A DELETED entry is
+			// not live in this snapshot and should not pin the file
+			// against orphan cleanup once the snapshot that
+			// originally held it live has been expired.
+			// This matches iceberg-java and pyiceberg behavior.
+			for entry, err := range manifest.Entries(fs, true) {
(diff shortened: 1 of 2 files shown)


twuebi (approved):

laskoviymishka (approved):
Nice catch, and clean fix.

The `false` → `true` flip lines `getReferencedFiles` up with Java’s `ManifestReader.liveEntries()`, and the new test reproduces the bug end-to-end through `Append → Overwrite → Expire`. Happy to merge.

Two optional follow-ups, both non-blocking:

* The `discardDeleted=true` flip also applies to delete-file manifests, `ManifestContentDeletes`. That looks semantically right — a `DELETED` entry there means a positional/equality delete file was superseded — but there’s no test for that path. Worth adding a subtest before someone refactors this loop later.

* `inMemoryCatalog.LoadTable` returns `(nil, nil)`. Safe today since `CommitTable` never retries, but the helper name is generic and it’ll probably get reused. The commit retry path would nil-deref `fresh.metadata`. Worth returning a real `*Table`, or an explicit `errors.New("not implemented")`, so future reuse fails loudly instead of panicking inside commit machinery.

PR #1078, opened 2026-05-13
fix(gocloud): support cross-bucket IO for write.metadata.path

cassio-paesleme:
## Summary

The blob `FileIO` implementation assumes all file operations target a single S3 bucket (the warehouse bucket opened at init time). When Iceberg's `write.metadata.path` table property points to a different bucket (e.g. a dedicated versioned metadata bucket), `defaultKeyExtractor` strips the wrong bucket prefix and the file lands in the warehouse bucket under a mangled key. Readers following the absolute S3 URI in `metadata.json` get a 404.

**Concrete failure mode**: Setting `write.metadata.path = s3://metadata-bucket/db/table/` causes manifest lists (`snap-*.avro`) and manifest files (`*-m*.avro`) to be written to `s3://warehouse-bucket/metadata-bucket/db/table/snap-*.avro` instead of `s3://metadata-bucket/db/table/snap-*.avro`. The `metadata.json` records the correct URI, but the bytes are in the wrong place.

## Changes

- Add `resolveBucket()` to `blobFileIO` which parses the full S3 URI and routes to the correct bucket
- Primary bucket (warehouse) uses the fast path with no map lookup
- Secondary buckets are opened lazily via a `BucketOpener` callback and cached with `sync.RWMutex`
- Update `Open`, `NewWriter`, `WriteFile`, `Remove`, and `DeleteFiles` to use `resolveBucket`
- Wire S3 scheme registration to pass a `BucketOpener` that reuses the same AWS config
- Backward compatible: callers without a `BucketOpener` get the same legacy behavior

## Test plan

- [x] `TestMultiBucketWriteAndRead` - writes to two memblob buckets via different S3 URIs, verifies files land in the correct bucket and can be read back
- [x] `TestMultiBucketDelete` - verifies `Remove` and `DeleteFiles` route to the correct bucket
- [x] `TestMultiBucketOpenerCaching` - verifies the opener is called once per bucket name
- [x] `TestMultiBucketFallbackWithoutOpener` - verifies backward compatibility when no opener is set
- [x] All existing tests pass unchanged

### DIFF
--- a/io/gocloud/blob.go
+++ b/io/gocloud/blob.go
@@ -124,6 +124,7 @@ func (bfs *blobFileIO) resolveBucket(path string) (*blob.Bucket, string, error) 
 	if !hasScheme {
 		// No scheme: treat as a key in the primary bucket (legacy behavior).
 		key, err := bfs.keyExtractor(path)
+
 		return bfs.Bucket, key, err
 	}
 
@@ -142,6 +143,7 @@ func (bfs *blobFileIO) resolveBucket(path string) (*blob.Bucket, string, error) 
 		// No opener configured: fall back to primary bucket with legacy key extraction.
 		// This preserves backward compatibility for callers that don't set a BucketOpener.
 		key, err := bfs.keyExtractor(path)
+
 		return bfs.Bucket, key, err
 	}
 
@@ -163,6 +165,7 @@ func (bfs *blobFileIO) resolveBucket(path string) (*blob.Bucket, string, error) 
 	if existing, ok := bfs.buckets[bucketName]; ok {
 		bfs.mu.Unlock()
 		_ = b.Close()
+
 		return existing, key, nil
 	}
 	bfs.buckets[bucketName] = b
--- a/io/gocloud/blob.go
+++ b/io/gocloud/blob.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	icebergio "github.com/apache/iceberg-go/io"
 	"gocloud.dev/blob"
@@ -91,8 +92,20 @@ func defaultKeyExtractor(bucketName string) KeyExtractor {
 	}
 }
 
+// BucketOpener creates a new blob.Bucket for the given bucket name.
+// Used to lazily open secondary buckets when a URI references a bucket
+// different from the primary one (e.g. write.metadata.path pointing to
+// a separate metadata bucket).
+type BucketOpener func(ctx context.Context, bucketName string) (*blob.Bucket, error)
+
 type blobFileIO struct {
 	*blob.Bucket
+
+	primaryBucket string
+	bucketOpener  BucketOpener
+
+	mu      sync.RWMutex
+	buckets map[string]*blob.Bucket
 
 	keyExtractor KeyExtractor
 	ctx          context.Context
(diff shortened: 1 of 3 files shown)


io/gocloud/blob.go
```diff
@@ -102,23 +115,81 @@ type blobFileIO struct {
...
-		return nil, &fs.PathError{Op: "open", Path: path, Err: fs.ErrInvalid}
+	if !fs.ValidPath(key) {
+		return nil, &fs.PathError{Op: "open", Path: key, Err: fs.ErrInvalid}
 	}
 
-	key, name := path, filepath.Base(path)
+	name := filepath.Base(key)
 
-	r, err := bfs.NewReader(bfs.ctx, key, nil)
+	r, err := bucket.NewReader(bfs.ctx, key, nil)
```
laskoviymishka, line 192: I think there's a subtle issue — `Open` now routes through the resolved `bucket` here, but `blobOpenFile.ReadAt` (line 49) still calls `f.b.NewRangeReader(...)`, which dispatches to the embedded *primary* bucket. So any cross-bucket file opened here fails on the first `ReadAt` (404, or worse if a same-named key exists in the primary), and that's the path Parquet column readers and the puffin/DV reader actually take. `TestMultiBucketWriteAndRead` passes only because `io.ReadAll` uses `Read`.

Could we thread `bucket` into `blobOpenFile` and have `ReadAt` use `f.bucket.NewRangeReader(...)`? A `ReadAt`-specific test would help lock it in. wdyt?

io/gocloud/blob.go
```diff
@@ -91,9 +92,21 @@ func defaultKeyExtractor(bucketName string) KeyExtractor {
...
+type BucketOpener func(ctx context.Context, bucketName string) (*blob.Bucket, error)
+
 type blobFileIO struct {
 	*blob.Bucket
 
+	primaryBucket string
+	bucketOpener  BucketOpener
+
+	mu      sync.RWMutex
+	buckets map[string]*blob.Bucket
```
laskoviymishka, line 108: These secondary buckets are never closed — `blobFileIO` only inherits `Close()` from the embedded primary. Each cached `*blob.Bucket` holds its own HTTP transport / connection pool, so a long-running process that touches many distinct metadata buckets will leak transports indefinitely.

I'd add a `Close()` on `blobFileIO` that walks `bfs.buckets` under the lock and closes each before closing the primary. wdyt?

io/gocloud/blob.go
```diff
@@ -159,37 +228,53 @@ func (bfs *blobFileIO) WriteFile(name string, content []byte) error {
...
+		Bucket:        bucket,
+		primaryBucket: primaryBucket,
+		bucketOpener:  opener,
+		buckets:       make(map[string]*blob.Bucket),
+		keyExtractor:  defaultKeyExtractor(primaryBucket),
+		ctx:           ctx,
+	}
 }
 
 func (bfs *blobFileIO) WalkDir(root string, fn fs.WalkDirFunc) error {
```
laskoviymishka, line 280: `WalkDir` is the one IO entry point that doesn't go through `resolveBucket` — it still hardcodes `fs.WalkDir(bfs.Bucket, ...)`. So if `root` names a secondary bucket (e.g., orphan-file enumeration on `write.metadata.path`), this silently walks the primary warehouse bucket and returns the wrong listing.

Could we route this through `resolveBucket` too? Probably worth a small test covering the cross-bucket root.

io/gocloud/register.go
```diff
@@ -38,7 +39,13 @@ func registerS3Schemes() {
...
 		}
 
-		return createBlobFS(ctx, bucket, defaultKeyExtractor(parsed.Host)), nil
+		opener := func(ctx context.Context, bucketName string) (*blob.Bucket, error) {
+			u := &url.URL{Scheme: parsed.Scheme, Host: bucketName}
+
+			return createS3Bucket(ctx, u, props)
+		}
+
+		return createMultiBucketBlobFS(ctx, bucket, parsed.Host, opener), nil
```
laskoviymishka, line 48: The fix lands for S3 here, but `registerGCSScheme` (line 64) and `registerAzureSchemes` (line 76) still call `createBlobFS`, so a GCS or ADLS user with `write.metadata.path` on a separate bucket still hits the mangled-key bug we're fixing. PyIceberg and Java handle cross-bucket GCS transparently, so tables written by them become unreadable from Go in that layout.

Could we either add `BucketOpener` impls for GCS/Azure here, or call out the S3-only scope in the PR description with a follow-up tracked? wdyt?

laskoviymishka (changes requested):
Thanks for picking this up — `write.metadata.path` on a separate bucket is a real production gap. The routing shape with `resolveBucket` + per-bucket cache + `BucketOpener` callback also looks like the right direction.

I’d hold before merging though.

Main concern: the read path still doesn’t look fixed. `blobOpenFile.ReadAt` still goes to the primary bucket, so any cross-bucket Parquet / Avro / Puffin read will fail on the first range request. The new `TestMultiBucketWriteAndRead` doesn’t catch this because `io.ReadAll` uses `Read`, not `ReadAt`.

A few other things I’d sort out before merge:

* `WalkDir` still hardcodes the primary bucket, so orphan-file enumeration over `write.metadata.path` is still wrong.
* Cached secondary buckets are never closed, so long-running processes can leak HTTP transports.
* The fix is S3-only. GCS and Azure still call `createBlobFS`, so they keep the original mangled-key bug. I’d either extend the fix to those backends or call out the S3-only scope clearly and file a follow-up.

One smaller thing I wouldn’t block on: `TestMultiBucketFallbackWithoutOpener` currently asserts the broken legacy behavior as a contract. Worth either expecting an error, or renaming the test to make that intent explicit.

Once those are sorted, happy to take another pass.

io/gocloud/blob.go
zeroshade, line 192: Blocker: `Open` resolves the correct bucket for the initial reader, but the returned `blobOpenFile` still stores `b: bfs`, and `blobOpenFile.ReadAt` uses `f.b.NewRangeReader`. Range reads for Parquet/Avro/Puffin files in a secondary bucket will still go through the primary bucket and can 404 or read the wrong object. Please store the resolved bucket on `blobOpenFile`, use it in `ReadAt`, and add a cross-bucket `ReadAt` test.

io/gocloud/blob.go
```diff
@@ -127,27 +198,25 @@ func (bfs *blobFileIO) Open(path string) (icebergio.File, error) {
...
 func (bfs *blobFileIO) Remove(name string) error {
-	var err error
-	name, err = bfs.preprocess(name)
+	bucket, key, err := bfs.resolveBucket(name)
 	if err != nil {
 		return &fs.PathError{Op: "remove", Path: name, Err: err}
 	}
 
-	return bfs.Delete(bfs.ctx, name)
+	return bucket.Delete(bfs.ctx, key)
```
zeroshade, line 206: Blocker: returning `bucket.Delete(...)` directly drops the existing `gcerrors.NotFound` to `fs.ErrNotExist` mapping for `Remove`. That changes the FileIO not-found semantics. Please preserve the mapping after routing to the resolved bucket.

io/gocloud/blob.go
```diff
@@ -102,23 +115,81 @@ type blobFileIO struct {
 	newRangeReader func(ctx context.Context, key string, offset, length int64) (io.ReadCloser, error)
 }
 
+// resolveBucket parses path into a bucket handle and object key. If the URI
+// references the primary bucket (or has no scheme), the primary bucket is
+// returned without any additional allocation. URIs that reference a different
+// bucket are resolved via BucketOpener and cached for reuse.
+func (bfs *blobFileIO) resolveBucket(path string) (*blob.Bucket, string, error) {
+	_, after, hasScheme := strings.Cut(path, "://")
```
zeroshade, line 123: Blocker: bucket resolution splits on `://` and then routes only by bucket name, ignoring the URI scheme. That means a `gs://metadata-bucket/key` passed to an S3 FileIO can be treated as an S3 bucket, which conflicts with the strict scheme/authority validation merged in #1259. Please parse and validate scheme plus authority, preferably by reusing the object-location parser from #1259, or route through the global IO registry by full URI.

io/gocloud/blob.go
zeroshade, line 280: High: `WalkDir` still walks `bfs.Bucket`, so secondary-bucket metadata paths are not routed through `resolveBucket`. Cross-bucket support is also currently S3-only, and cached secondary buckets are never closed. Please route `WalkDir` through the resolved bucket and add a `Close` path that closes cached bucket handles.

io/gocloud/blob_test.go
```diff
@@ -379,3 +382,192 @@ func TestBlobFileIODeleteFilesEmpty(t *testing.T) {
...
+	// in the primary bucket under a mangled key (legacy behavior).
+	wfs := bfs.(icebergio.WriteFileIO)
+	require.NoError(t, wfs.WriteFile(
+		"s3://other-bucket/path/file.txt", []byte("hello")))
+
+	// The file lands in the primary bucket with the full "other-bucket/path/file.txt" key
+	// because TrimPrefix("other-bucket/...", "warehouse/") is a no-op.
+	exists, err := bucket.Exists(ctx, "other-bucket/path/file.txt")
+	require.NoError(t, err)
+	assert.True(t, exists, "without opener, cross-bucket URI should fall back to primary bucket with mangled key")
```
zeroshade, line 572: Medium: this test locks in the unsafe fallback where `s3://other-bucket/...` is written into the primary bucket. With #1259 merged, this should be rejected as an unsupported authority unless an explicit cross-bucket opener is configured. Please update the expected behavior instead of codifying the fallback.

zeroshade (changes requested):
Thanks for working through the cross-bucket FileIO path. I think this needs to be redesigned before merge. PR #1259 has now merged to main with stricter bucket/authority plus scheme validation for gocloud FileIO, while this PR introduces S3-specific fallback/routing semantics that go in the opposite direction. Please rebase and reconcile this with #1259 and with the BlobFileIO changes from #1350 before continuing.

zeroshade:
**🛠️ Maintainer triage note for @cassio-paesleme** · by `zeroshade` · 2026-08-11 UTC

PR #1081, opened 2026-05-15
fix(table): map nanosecond Arrow timestamps to v3 TimestampNs/TimestampTzNs types

### LINKED ISSUE #695: 'ns' timestamp precision not supported

I’m trying to write Arrow records into an Iceberg table.

In my Arrow records, start_time_unix_nano and time_unix_nano are in nanoseconds. I defined the Iceberg schema like this:

```
	schema := iceberg.NewSchemaWithIdentifiers(
		1,        // schema-id
		[]int{1}, // identifier fields: id
		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64, Required: true},
		iceberg.NestedField{ID: 2, Name: "parent_id", Type: iceberg.PrimitiveTypes.Int32, Required: true},
		//iceberg.NestedField{ID: 3, Name: "start_time_unix_nano", Type: iceberg.PrimitiveTypes.TimestampTz, Required: true}, // microsecond precision
		iceberg.NestedField{ID: 3, Name: "start_time_unix_nano", Type: iceberg.PrimitiveTypes.TimestampNs, Required: true}, // nanosecond precision
		iceberg.NestedField{ID: 4, Name: "time_unix_nano", Type: iceberg.PrimitiveTypes.TimestampNs, Required: true},       // nanosecond precision
		iceberg.NestedField{ID: 5, Name: "int_value", Type: iceberg.PrimitiveTypes.Int64, Required: false},
		iceberg.NestedField{ID: 6, Name: "double_value", Type: iceberg.PrimitiveTypes.Float64, Required: false},
		iceberg.NestedField{ID: 7, Name: "flags", Type: iceberg.PrimitiveTypes.Int64, Required: false},
	)
```

However, I get this error:

`"error":"error encountered during file writing: error encountered during arrow schema visitor: type error: 'ns' timestamp precision not supported"`

Does this mean TimestampNs is not fully supported yet, or am I missing something?

jacobmarble:
`table/Transaction.AddFiles()` and `table/WriteRecords()` panic with 'ns' timestamp precision not supported (#695). I've managed to work around this with `table/Transaction.AddDataFiles()`.

The fix is pretty simple, just map the new type in `table/ArrowTypeToIceberg()`. I've left the downcast path `downcastTimestamp=true` unchanged.

**Behavior change for v1/v2 callers:** Previously `UTC-tz` and `empty-tz` nanosecond timestamps with `downcastTimestamp=false` failed immediately at the conversion step with 'ns' timestamp precision not supported. They will now produce a `TimestampTzNs`/`TimestampNs` Iceberg type and fail later at `minFormatVersionForType` if the table is format version 1 or 2. The failure is preserved — only the site moves.

Closes #695 (again)
Helps #589


table/arrow_utils_test.go
```diff
@@ -76,11 +76,11 @@ func TestArrowToIceberg(t *testing.T) {
 		{arrow.FixedWidthTypes.Timestamp_s, iceberg.PrimitiveTypes.TimestampTz, false, ""},
 		{arrow.FixedWidthTypes.Timestamp_ms, iceberg.PrimitiveTypes.TimestampTz, false, ""},
 		{arrow.FixedWidthTypes.Timestamp_us, iceberg.PrimitiveTypes.TimestampTz, true, ""},
-		{arrow.FixedWidthTypes.Timestamp_ns, nil, false, "'ns' timestamp precision not supported"},
+		{arrow.FixedWidthTypes.Timestamp_ns, iceberg.PrimitiveTypes.TimestampTzNs, true, ""},
 		{&arrow.TimestampType{Unit: arrow.Second}, iceberg.PrimitiveTypes.Timestamp, false, ""},
 		{&arrow.TimestampType{Unit: arrow.Millisecond}, iceberg.PrimitiveTypes.Timestamp, false, ""},
 		{&arrow.TimestampType{Unit: arrow.Microsecond}, iceberg.PrimitiveTypes.Timestamp, true, ""},
-		{&arrow.TimestampType{Unit: arrow.Nanosecond}, nil, false, "'ns' timestamp precision not supported"},
+		{&arrow.TimestampType{Unit: arrow.Nanosecond}, iceberg.PrimitiveTypes.TimestampNs, true, ""},
```
laskoviymishka, line 83: while we're here, could we also add a non-UTC tz nanosecond row to lock in the new panic branch? something like:

```go
{&arrow.TimestampType{Unit: arrow.Nanosecond, TimeZone: "US/Pacific"}, nil, false, "unsupported arrow type for conversion - timestamp[ns, tz=US/Pacific]"},
```

right under the existing `us, tz=US/Pacific` panic row. the loop always calls `ArrowTypeToIceberg(tt.dt, false)` (downcastTimestamp=false), so this exercises the third sub-case of the new branch — the panic at line 348 — which the two added rows don't cover. small enough to fold into this PR, wdyt?

table/arrow_utils.go
```diff
@@ -342,10 +342,19 @@ func (c convertToIceberg) Primitive(dt arrow.DataType) (result iceberg.NestedFie
...
+		if dt.Unit == arrow.Nanosecond && !c.downcastTimestamp {
+			if slices.Contains(utcAliases, dt.TimeZone) {
+				result.Type = iceberg.PrimitiveTypes.TimestampTzNs
+			} else if dt.TimeZone == "" {
+				result.Type = iceberg.PrimitiveTypes.TimestampNs
+			} else {
+				panic(fmt.Errorf("%w: unsupported arrow type for conversion - %s", iceberg.ErrInvalidSchema, dt))
 			}
+
+			return result
```
laskoviymishka: tiny readability nit while we're here — `if dt.Unit == arrow.Nanosecond` shows up twice in a row with an early `return` in between. could fold into one block to keep the two ns paths adjacent:

```go
if dt.Unit == arrow.Nanosecond {
    if !c.downcastTimestamp {
        switch {
        case slices.Contains(utcAliases, dt.TimeZone):
            result.Type = iceberg.PrimitiveTypes.TimestampTzNs
        case dt.TimeZone == "":
            result.Type = iceberg.PrimitiveTypes.TimestampNs
        default:
            panic(fmt.Errorf("%w: unsupported arrow type for conversion - %s", iceberg.ErrInvalidSchema, dt))
        }
        return result
    }
    slog.Warn("downcasting nanosecond timestamp to microsecond, precision loss may occur")
}
```

equivalent semantics, take it or leave it.

table/arrow_utils.go
```diff
@@ -342,10 +342,19 @@ func (c convertToIceberg) Primitive(dt arrow.DataType) (result iceberg.NestedFie
...
 	case *arrow.TimestampType:
-		if dt.Unit == arrow.Nanosecond {
-			if !c.downcastTimestamp {
-				panic(fmt.Errorf("%w: 'ns' timestamp precision not supported", iceberg.ErrType))
+		if dt.Unit == arrow.Nanosecond && !c.downcastTimestamp {
+			if slices.Contains(utcAliases, dt.TimeZone) {
+				result.Type = iceberg.PrimitiveTypes.TimestampTzNs
+			} else if dt.TimeZone == "" {
+				result.Type = iceberg.PrimitiveTypes.TimestampNs
+			} else {
```
laskoviymishka: worth a line in the PR body that UTC-tz and empty-tz nanoseconds with `downcastTimestamp=false` used to panic (`'ns' timestamp precision not supported`) and now map to `TimestampTzNs` / `TimestampNs`. correct behavior for v3, but anyone bisecting #695 down the line will appreciate the explicit note that this is a behavior change for v1/v2 callers — they'll now produce a v3 schema and trip on `minFormatVersionForType` instead of failing at conversion. small enough to fold in here.

laskoviymishka (changes requested):
Nice — this lines up the v3 nanosecond timestamp mapping with the spec.

With `downcastTimestamp=false`, UTC and empty-tz nanoseconds now map to `TimestampTzNs` / `TimestampNs`. Non-UTC time zones still panic, which feels right — silently dropping the zone through the downcast path would be worse.

Three small things I’d fold into this PR rather than split out:

* One missing test row: nanosecond + non-UTC tz. The two new rows cover the UTC-alias and empty-tz paths, but the panic branch around line 348 is not pinned yet. Should be easy to add next to the existing `us, tz=US/Pacific` row.
* The two back-to-back `if dt.Unit == arrow.Nanosecond` checks read a bit awkwardly. Optional readability cleanup — take it or leave it.
* The PR body should mention the behavior change: UTC-tz / empty-tz nanoseconds with `downcastTimestamp=false` used to panic, and now produce v3 ns types. That is correct for v3, but worth surfacing in release notes. v1/v2 callers will now fail later at `minFormatVersionForType`, not during conversion.

All small enough to land here. Once they’re in, happy to merge

table/arrow_utils.go
jacobmarble: b667a98

table/arrow_utils_test.go
jacobmarble: aeca112

table/arrow_utils.go
jacobmarble: Done!

jacobmarble:
Thanks for your quick review @laskoviymishka

laskoviymishka (approved):

Issue #1086, opened 2026-05-17
discussion: align v1/v2 metadata parsing with Java/PyIceberg on v3-only fields


laskoviymishka:
Follow-up from #1069, open question.

#1069 hard-errors on `next-row-id` / `encryption-keys` in v1/v2 metadata. None of the reference clients do this — Java reads `next-row-id` only when `formatVersion >= 3` and defaults otherwise, has no version guard on `encryption-keys` at all; PyIceberg and iceberg-rust both drop unknown fields. So we're the only one that rejects here.

Doesn't bite today because v3 isn't widely deployed. Once downgrade tooling shows up, a v2 file with stray v3 fields reads everywhere except Go.

Strict surface has value (mis-versioned metadata is by definition wrong, and it catches our own writer if it ever leaks v3 fields), but forward-compat is also a spec goal. If we keep it strict, a code comment near `rejectV3OnlyFields` capturing the intentional divergence would help future readers not read it as a parity bug.

tanmayrauth:
Agreed, we should relax and match the other clients.                                                                                                                                                                                  
                                                                                                                                                                                                                                         
  Iceberg tables get read/written by multiple clients in the same pipeline. If any tool leaves a stray v3 field in v2 metadata, Java/PyIceberg/Rust all read it fine. Go becomes the only client that hard-fails. That's a production outage for Go users with no actual data integrity issue (the fields sit unused on commonMetadata and are never exposed via sub-v3 accessors).                                                                                          
                  
  Catching our own writer leaking v3 fields into v2 is valuable, but that belongs in writer-side tests - not reader-side validation that penalizes every consumer.                                                                       
                  
  Plan:                                                                                                                                                                                                                                  
  - Remove rejectV3OnlyFields and the call sites in v1/v2 unmarshal paths.
  - Add a short comment noting we intentionally match the lenient behavior of other clients, linking this issue.                                                                                                                         
                                                                                                                
  This also makes the generalized rejectFieldsBeyondVersion follow-up moot.

laskoviymishka:
Agree, but I think this is a bit bigger than just iceberg-go.

It feels worth starting a proper discussion on `dev@iceberg.apache.org`, and probably bringing the engine side into it too. Trino / Spark / Flink readers are where forward-compat issues will actually show up in production.

Right now the spec is silent on reader behavior for newer-version fields, and each client has picked its own default. We should probably pin that down before v4 lands.

tanmayrauth:
[...]

PR #1088, opened 2026-05-17
feat(table): gate v2-only metadata fields on v1 reads

### LINKED ISSUE #1087: feat(table): gate v2-only metadata fields on v1 reads
Follow-up from #1069.

#1069 gates v3-only fields on v1/v2 reads. By the same logic, `last-sequence-number` is v2+ and silently accepted on v1 today — the gate only catches one direction. Extend the existing helper to cover v2-only fields on v1 (or revisit if we end up loosening the v3 gate per #1086).

laskoviymishka:
Closes #1087.

`#1069` added `rejectV3OnlyFields` to catch v3-only fields in pre-v3 metadata.

This PR applies the same idea to `last-sequence-number`: it’s a v2+ field on the shared `commonMetadata` struct, but today we silently accept it when reading v1 metadata.

To keep the check from becoming one-off, this generalizes the helper to `rejectFieldsBeyondVersion(currentVersion, fields...)`, where each field includes the format version it was introduced in. `last-sequence-number` is then wired into the v1 read path. For v2 reads, `next-row-id` / `encryption-keys` go through the same generalized helper, with no behavior change.

One caveat: this depends on the discussion in `#1086`. If we decide to relax the v3-only gate and silently ignore forward-looking fields for compatibility, then this PR becomes moot and should be revisited together with the strictness introduced in `#1069`.

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1363,17 +1363,18 @@ var (
 	ErrPartitionSpecNotFound        = errors.New("partition spec not found")
 )
 
-type v3FieldCheck struct {
-	name    string
-	present bool
+type versionScopedField struct {
+	name       string
+	introduced int
+	present    bool
 }
 
-func rejectV3OnlyFields(version int, checks ...v3FieldCheck) error {
+func rejectFieldsBeyondVersion(currentVersion int, fields ...versionScopedField) error {
 	var errs []error
-	for _, c := range checks {
-		if c.present {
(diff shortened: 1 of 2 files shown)


laskoviymishka:
@tanmayrauth PTAL

tanmayrauth:
I replied back in #1086, lmk what you think?

zeroshade (approved):

PR #1089, opened 2026-05-17
fix(table): derive position-delete partition paths from table schema

### LINKED ISSUE #1082: Partitioned position-delete writer derives partition paths from delete schema
### Apache Iceberg version

main (development)

### Please describe the bug

`positionDeletePartitionedFanoutWriter` appears to derive partition paths using the positional-delete schema rather than the table schema.

In `positionDeletePartitionedFanoutWriter.partitionPath`, the partition type and path are currently built from `p.schema`:

```go
data := newPartitionRecord(partitionContext.partitionData, spec.PartitionType(p.schema))
return spec.PartitionToPath(data, p.schema), nil
```

The writer initializes `p.schema` to `iceberg.PositionalDeleteSchema`. For partition specs based on table data columns, those source fields are not present in the positional-delete schema. `PartitionSpec.PartitionType` skips missing source fields, so the derived partition type can be empty and `PartitionToPath` can return an empty path.

That means different target data-file partitions can collapse to the same rolling writer key. Since `writerFactory.getOrCreateRollingDataWriter` reuses writers by partition path, the delete file can be opened with the first partition's metadata and then receive position-delete rows for later target files from different partitions.

For example, for a table partitioned by `day`:

| Target data file | Real partition |
| --- | --- |
| `file-a.parquet` | `day=2026-05-01` |
| `file-b.parquet` | `day=2026-05-02` |

If the position-delete writer derives the partition path from `iceberg.PositionalDeleteSchema`, both can produce the same empty partition path. The rolling writer can then reuse the writer opened for `day=2026-05-01` while writing deletes that target `day=2026-05-02`.

The expected behavior is that the positional-delete file schema is still used for the delete file contents, but the table schema is used when deriving the table partition path from the target data file's partition data.

This looks related to, but distinct from, #767, which fixed nondeterministic partition path ordering. This issue is about using the delete schema instead of the table schema when deriving the path.

I have contributed to iceberg-python before, but this would be my first iceberg-go PR. I would like to work on a fix if this diagnosis sounds right.

laskoviymishka:
The partitioned position-delete writer was building partition paths from `iceberg.PositionalDeleteSchema`, which only contains `file_path` and `pos`.

That breaks for partition specs based on real table columns. `PartitionSpec.PartitionType` silently skips fields whose source IDs are not present in the schema, so those partitions collapse to an empty struct and `PartitionToPath` returns `""`.

The downstream effect is pretty bad: `writerFactory.getOrCreateRollingDataWriter` keys writers by that path, so every partition ends up sharing the same rolling writer. Since the writer keeps the first partition’s `partitionData`, the resulting delete file is tagged as belonging to the first partition while containing position deletes for files from multiple partitions.

This PR fixes that by deriving the partition path from `metadata.CurrentSchema()`, matching what the data-side `partitionedFanoutWriter` already does.

The actual file contents schema is unchanged: position-delete files still use `PositionalDeleteSchema`, which is passed separately to the writer factory via `withFactoryFileSchema`.

Fixes #1082.

### DIFF
--- a/table/pos_delete_partitioned_fanout_writer.go
+++ b/table/pos_delete_partitioned_fanout_writer.go
@@ -34,19 +34,17 @@ import (
 type positionDeletePartitionedFanoutWriter struct {
 	partitionContextByFilePath map[string]partitionContext
 	metadata                   Metadata
-	schema                     *iceberg.Schema
 	itr                        iter.Seq2[arrow.RecordBatch, error]
 	writerFactory              *writerFactory
 	concurrentDataFileWriter   *concurrentDataFileWriter
 }
 
 // newPositionDeletePartitionedFanoutWriter creates a new PartitionedFanoutWriter with the specified
-// partition specification, schema, and record iterator.
+// metadata, partition context, and record iterator.
 func newPositionDeletePartitionedFanoutWriter(metadata Metadata, concurrentWriter *concurrentDataFileWriter, partitionContextByFilePath map[string]partitionContext, itr iter.Seq2[arrow.RecordBatch, error], writerFactory *writerFactory) *positionDeletePartitionedFanoutWriter {
 	return &positionDeletePartitionedFanoutWriter{
 		partitionContextByFilePath: partitionContextByFilePath,
 		metadata:                   metadata,
-		schema:                     iceberg.PositionalDeleteSchema,
 		itr:                        itr,
 		writerFactory:              writerFactory,
 		concurrentDataFileWriter:   concurrentWriter,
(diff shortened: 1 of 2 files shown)


tanmayrauth:
LGTM, fix is correct and well-tested. Thanks!

fallintoplace:
Thanks Andrei @laskoviymishka!

I approved this PR.

fallintoplace (approved):

tanmayrauth (approved):

zeroshade (approved):

PR #1094, opened 2026-05-19
test(table): geospatial test fixtures

### LINKED ISSUE #1090: test(table) Geospatial Test Fixtures

Parent: https://github.com/apache/iceberg-go/issues/989

As mentioned in https://github.com/apache/iceberg-go/issues/989#issuecomment-4479477750 There is a need for `Building WKB sample tables, Java / PyIceberg round-trip test data, and GeoArrow conformance fixtures`.  According to my interpretation this would essentially mean generating small test cases for a variety of standard geoarrow data and edge cases generated by multiple different clients to ensure the go implementation works across the iceberg ecosystem.

I am happy to work on contributing this as I've done GeoArrow work in the past so that seems like a good intro to the codebase. I have created a new issue here since I did not see an existing one to track this and didn't want to clutter the main geo issue.

I have a few questions before I begin if you don't mind @laskoviymishka :

1. In which directory would these fixtures be placed in the iceberg-go codebase? 
    - I am assuming these live in this repo but if they also go anywhere else please let me know
3. Do you have preferences regarding the format / metadata of test data    
   - Since I am relatively new to iceberg, wasn't clear if putting a small iceberg table directly in the git repo would be too heavy. Is there a better way to do this?
   - Is there a specific package manager you prefer for Python/Java, etc. for the purposes of pinning the script which generates the test data (assuming you want that?)

C-Loftus:
- Add geospatial test fixtures that can be used for validating geospatial logic performs as expected. All fixtures come from the [parquet-testing](https://github.com/apache/parquet-testing/tree/master/data/geospatial) repository. I was previously unaware this repo existed but I think this is the right data to use since it prevents us from needing to write custom pyarrow code to generate them. 
    - I put all generation logic in a shell script since it's a bit more succinct than a custom script in go and I don't imagine it will need to be run very often given the last update of that test data was over a year ago.

- Note these fixtures use all the **native parquet geometry type** _not_ the geoarrow encoding. I am pretty sure this is what we want, but if I am misinterpreting something let me know. My understanding is that geoarrow is not used in the native parquet type (but from reading other PRs like it seems like geoarrow-go is being used for some of the parse logic, not the storage encoding?) There is more data here https://github.com/geoarrow/geoarrow-data if we need more, some of which is `_native.parquet` but since it does include a lot of geoparquet files with the geoarrow encoding, I did not use any of these.

- I have not added any parse tests for this since my understanding is that the initial geospatial PRs here have not been merged yet. Given these are standard parquet test files, it can be assumed they will need to be able to be parsed and conform to the spec.

Closes #1090


table/testdata/geo/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Geopatial Parquet Fixtures from [parquet-testing](https://github.com/apache/parquet-testing)
```
laskoviymishka: Nit — "Geopatial" → "Geospatial" (spelled correctly two lines down).

table/testdata/geo/gen_fixtures.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
laskoviymishka, line 1: Could we wire this up via `//go:generate` to match the existing convention in the repo? `puffin/gen_dv_fixture.go` + `puffin/dv_golden_test.go` set the pattern for fixture generators, and `exprs.go` does it for stringer — both run via `go generate ./...`.

Doesn't have to be a Go program — `//go:generate bash gen_fixtures.sh` is totally fine and keeps the shell script as-is. The win is discoverability: contributors find fixture regeneration through `go generate`, not by spelunking through `table/testdata/geo/` looking for a script.

table/testdata/geo/gen_fixtures.sh
```diff
@@ -0,0 +1,62 @@
...
+START_MARKER="<!-- BEGIN GENERATED FIXTURE LINKS -->"
+
+JSON="$(curl -s "$API_URL")"
+
+# Download all the parquet fixtures, not the associated config files
+# or generation scripts
+printf '%s\n' "$JSON" \
+  | jq -r '.[] | .download_url' \
+  | grep '\.parquet$' \
+  | xargs -n 1 wget -N
```
laskoviymishka: Three small portability nits in the script that'll bite the next contributor:

- No `cd "$(dirname "$0")"` at the top — running from anywhere else drops files in the caller's cwd
- `wget` / `wget -N` isn't installed on macOS by default and is GNU-only; `curl -fLO --time-cond` is the portable equivalent
- `curl -s` on line 30 swallows GitHub API rate-limit errors silently — `$JSON` becomes an error envelope, jq returns nothing, and the README gets rewritten with an empty fixture list under `set -e`. `curl -fsS` lets the failure propagate.

laskoviymishka (changes requested):
Fixtures look good to me — pinned to a `parquet-testing` SHA, ASLv2 on both sides, deterministic contents. I’m fine landing them before the parser.

One ask before merge: can we wire regeneration through `//go:generate`? `puffin/gen_dv_fixture.go` + `puffin/dv_golden_test.go` is the closest precedent. No need to rewrite the generator in Go, just `//go:generate bash gen_fixtures.sh` is fine. This just makes fixture refresh discoverable via `go generate ./...`.

I left a couple small portability nits inline, plus one README typo: `Geopatial` → `Geospatial`.

Thanks again for picking this up!

table/testdata/geo/README.md
C-Loftus: Fixed, thanks for catching

table/testdata/geo/gen_fixtures.sh
C-Loftus: > No cd "$(dirname "$0")" at the top — running from anywhere else drops files in the caller's cwd

Fixed

> wget / wget -N isn't installed on macOS by default and is GNU-only; curl -fLO --time-cond is the portable equivalent

Changed to curl. `--time-cond` has some issues where it didn't like the input from jq. I could do some transformations on the URL but figured it is easiest to omit. i.e. best to prioritize clarity of the script since the parquet files are so small anyways and presumably will not be regenerated often. If you are ok with more complexity this could be changed though. 


> curl -s on line 30 swallows GitHub API rate-limit errors silently — $JSON becomes an error envelope, jq returns nothing, and the README gets rewritten with an empty fixture list under set -e. curl -fsS lets the failure propagate.

Fixed

table/testdata/geo/gen_fixtures.sh
```diff
@@ -59,4 +62,6 @@ rm -f "${README}.bak"
 
   echo
   echo "<!-- END GENERATED FIXTURE LINKS -->"
-} >> "$README"
\ No newline at end of file
+} >> "$README"
+
+echo "Finished fetching parquet fixtures with geo data from parquet-testing repo"
```
C-Loftus, line 67: Added a little message here so it is clear what happened when a user runs go generate

table/gen_geo_fixtures.go
```diff
@@ -0,0 +1,22 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+// Run the following command to generate the parquet fixtures with
+// geo data for testing purposes:  go generate ./table/...
+//go:generate bash testdata/geo/gen_fixtures.sh
```
C-Loftus, line 22: I put this in the table package since there is not a geo package. Realistically once the geo package is merged and stabilized it may make more sense to put it there. Once that is merged feel free to ping me and I am happy to switch. Just didn't want to create an empty package for no reason that could cause merge conflicts with other open PRs

C-Loftus:
Should be all set for review again now, thanks for feedback!

table/gen_geo_fixtures.go
laskoviymishka: Yes, totally make sense as follow up later

laskoviymishka (approved):
:shipit: , nice work!

PR #1095, opened 2026-05-19
Fix Azure credential test that fails when az is logged in

kevinjqliu:
`TestCreateAzureBucketDefaultCredentialCalled` fails when Azure credentials are present on the machine (e.g., via `az login`).

When `DefaultAzureCredential` succeeds in creating a client, the `List` operation returns an Azure service error (e.g., `AuthorizationPermissionMismatch`) that does not contain "DefaultAzureCredential". The assertion was checking for a string that can never appear in this code path.

Removed the incorrect `assert.Contains` on the List error. The test still validates that `DefaultAzureCredential` is used (via the creation error path or non-nil bucket assertion).

### DIFF
--- a/io/gocloud/azure_test.go
+++ b/io/gocloud/azure_test.go
@@ -46,14 +46,11 @@ func TestCreateAzureBucketDefaultCredentialCalled(t *testing.T) {
 	// If bucket creation succeeds, verify we can't actually use it without proper auth
 	assert.NotNil(t, bucket, "Bucket should be created when DefaultAzureCredential client creation succeeds")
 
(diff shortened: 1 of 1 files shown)


io/gocloud/azure_test.go
laskoviymishka, line 28: nit: test name still about `DefaultCredential`

laskoviymishka (approved):

io/gocloud/azure_test.go
kevinjqliu: the name is still accurate, we're still testing that the path is still using `DefaultAzureCredential`

kevinjqliu:
thanks for the review @laskoviymishka

PR #1098, opened 2026-05-19
fix(table): handle missing strict metrics conservatively

### LINKED ISSUE #1091: Strict metrics evaluator treats missing null counts as no nulls
### Apache Iceberg version

main (development)

### Please describe the bug

I noticed a case where `strictMetricsEval` may be too optimistic when null counts are missing for a nullable field.

The strict evaluator loads metrics directly from the data file:

```go
ev.valueCounts, ev.nullCounts = file.ValueCounts(), file.NullValueCounts()
ev.nanCounts = file.NaNValueCounts()
ev.lowerBounds, ev.upperBounds = file.LowerBoundValues(), file.UpperBoundValues()
```

For range/equality predicates, it first checks whether the field may contain nulls:

```go
if m.canContainNulls(fieldID) || m.canContainNans(fieldID) {
    return rowsMightNotMatch
}
```

But `canContainNulls` returns false when the field is missing from `NullValueCounts()`:

```go
func (m *strictMetricsEval) canContainNulls(fieldID int) bool {
    cnt, exists := m.nullCounts[fieldID]

    return exists && cnt > 0
}
```

For nullable fields, I think a missing null-count entry should be treated as unknown rather than as zero nulls.

Example:

```text
predicate: id > 0

file metrics:
  lower_bounds[id] = 1
  upper_bounds[id] = 2
  null_value_counts[id] missing

possible rows:
  1
  2
  NULL
```

The lower bound proves that all non-null values are greater than zero, but it does not prove that there are no nulls. In this case, strict evaluation should not conclude that every row matches.

This matters because filtered overwrite/delete uses the strict evaluator to decide whether a file can be deleted entirely or needs a partial rewrite:

```go
if strict {
    localDelete = append(localDelete, df)
} else {
    localRewrite = append(localRewrite, df)
}
```

If a nullable file is missing null counts but has bounds, it can be classified as a whole-file delete even though null rows should survive the filter.

I’d be happy to work on a regression test/fix if this sounds right.

fallintoplace:
## Summary

- treat missing null counts for optional fields as unknown during strict range/equality evaluation
- treat missing NaN counts for float/double fields as unknown in the same strict checks
- add regression coverage for nullable fields and float fields with bounds present but missing metrics

## Why

The strict metrics evaluator is used to decide when a file is guaranteed to match a predicate. Missing metrics should not be interpreted as zero. For optional fields, an absent null count means the file may contain nulls. For float/double fields, an absent NaN count means the file may contain NaNs.

Fixes #1091.

## Testing

- `go test ./table -run TestEvaluators`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/table/evaluators.go
+++ b/table/evaluators.go
@@ -1586,16 +1586,15 @@ func (m *strictMetricsEval) canContainNans(fieldID int) bool {
 func (m *strictMetricsEval) mayContainNans(field iceberg.NestedField) bool {
 	switch field.Type.(type) {
 	case iceberg.Float32Type, iceberg.Float64Type:
+		cnt, exists := m.nanCounts[field.ID]
+		if !exists {
+			return true
+		}
+
+		return cnt > 0
 	default:
 		return false
 	}
-
-	cnt, exists := m.nanCounts[field.ID]
-	if !exists {
-		return true
-	}
-
-	return cnt > 0
 }
 
 // literalToPhysBytes converts an Iceberg literal to the physical byte
--- a/table/evaluators.go
+++ b/table/evaluators.go
@@ -1308,7 +1308,7 @@ func (m *strictMetricsEval) VisitLess(t iceberg.BoundTerm, lit iceberg.Literal) 
 	field := t.Ref().Field()
 	fieldID := field.ID
 
-	if m.canContainNulls(fieldID) || m.canContainNans(fieldID) {
+	if m.mayContainNulls(field) || m.mayContainNans(field) {
 		return rowsMightNotMatch
 	}
 
@@ -1330,7 +1330,7 @@ func (m *strictMetricsEval) VisitLessEqual(t iceberg.BoundTerm, lit iceberg.Lite
 	field := t.Ref().Field()
 	fieldID := field.ID
 
-	if m.canContainNulls(fieldID) || m.canContainNans(fieldID) {
+	if m.mayContainNulls(field) || m.mayContainNans(field) {
 		return rowsMightNotMatch
 	}
 
@@ -1352,7 +1352,7 @@ func (m *strictMetricsEval) VisitGreater(t iceberg.BoundTerm, lit iceberg.Litera
 	field := t.Ref().Field()
 	fieldID := field.ID
 
-	if m.canContainNulls(fieldID) || m.canContainNans(fieldID) {
+	if m.mayContainNulls(field) || m.mayContainNans(field) {
 		return rowsMightNotMatch
 	}
 
@@ -1379,7 +1379,7 @@ func (m *strictMetricsEval) VisitGreaterEqual(t iceberg.BoundTerm, lit iceberg.L
 	field := t.Ref().Field()
 	fieldID := field.ID
 
-	if m.canContainNulls(fieldID) || m.canContainNans(fieldID) {
+	if m.mayContainNulls(field) || m.mayContainNans(field) {
 		return rowsMightNotMatch
 	}
 
@@ -1406,7 +1406,7 @@ func (m *strictMetricsEval) VisitEqual(t iceberg.BoundTerm, lit iceberg.Literal)
 	field := t.Ref().Field()
 	fieldID := field.ID
 
-	if m.canContainNulls(fieldID) || m.canContainNans(fieldID) {
+	if m.mayContainNulls(field) || m.mayContainNans(field) {
 		return rowsMightNotMatch
 	}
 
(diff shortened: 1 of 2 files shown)


table/evaluators.go
```diff
@@ -1568,12 +1568,36 @@ func (m *strictMetricsEval) canContainNulls(fieldID int) bool {
 	return exists && cnt > 0
 }
 
+func (m *strictMetricsEval) mayContainNulls(field iceberg.NestedField) bool {
+	cnt, exists := m.nullCounts[field.ID]
+	if !exists {
+		return !field.Required
```
laskoviymishka, line 1574: `VisitNotEqual` and `VisitNotIn` still call `canContainNulls(fieldID) || canContainNans(fieldID)` — same absent-count short-circuit this PR is replacing everywhere else. Under their inverted return semantics (`rowsMustMatch` instead of `rowsMightNotMatch`), the bug class is identical: for an optional field with no per-key null count, `canContainNulls` returns `false`, the guard short-circuits, and the bounds check can wrongly conclude all rows satisfy `col != x` when nulls may actually be present. Both methods already have `field := t.Ref().Field()` in scope, so this is the same two-line substitution to `m.mayContainNulls(field) || m.mayContainNans(field)`. Worth folding in here to fully close #1091.

table/evaluators.go
```diff
@@ -1568,12 +1568,36 @@ func (m *strictMetricsEval) canContainNulls(fieldID int) bool {
...
 
 	return exists && cnt > 0
 }
 
+func (m *strictMetricsEval) mayContainNans(field iceberg.NestedField) bool {
+	switch field.Type.(type) {
+	case iceberg.Float32Type, iceberg.Float64Type:
+	default:
+		return false
+	}
```
laskoviymishka, line 1597: The empty `case iceberg.Float32Type, iceberg.Float64Type:` arm with `default: return false` below is valid but unusual — a reader pauses to confirm the float arm is intentionally empty-and-fall-through rather than an oversight. A one-line comment in the arm, or inverting to a positive guard, would save the double-take.

table/evaluators_test.go
```diff
@@ -2346,6 +2346,87 @@ func (suite *StrictMetricsTestSuite) TestMissingStats() {
 	}
 }
 
+func (suite *StrictMetricsTestSuite) TestMissingNullCounts() {
```
laskoviymishka, line 2349: Both new tests exercise only `GreaterThan`. The six migrated visitors share the same early guard so coverage is defensible, but one `EqualTo` + one `In` case (or a comment noting `GreaterThan` is representative) would make the regression intent explicit. Once `VisitNotEqual` / `VisitNotIn` are addressed, a case there closes the loop.

laskoviymishka (changes requested):
Nice work! 

The conservative-default fix looks right: missing null / NaN counts should mean “unknown”, not “zero”. That matches Java’s behavior, and the new helpers + `nullCounts` tweak are a good shape.

One thing before merge: the same bug still seems to exist in `VisitNotEqual` and `VisitNotIn`, which still use the old `canContainNulls` / `canContainNans`.

Could you either fold those two in here, or open a follow-up issue and note why they’re deferred? Not a big change, just the same correctness rule in the sibling branches, so should be easy to push forward.

Overall looks good, but let's tighten the last bits before merging.

table/evaluators.go
tanmayrauth: +1 on the VisitNotEqual/VisitNotIn gap, but I don't think mayContainNulls/mayContainNans is the right fix there. 

The Java reference uses containsNullsOnly/containsNaNsOnly (not canContainNulls) and has no  separate null/NaN guard beyond that — it falls directly through to bounds checks.                                                                                                                              
                                                                                                                                                                                                                 
  The Go code incorrectly uses canContainNulls (any nulls → rowsMustMatch) where it should use containsNullsOnly (all nulls → rowsMustMatch). The helpers already exist at evaluators.go:637. Probably best as a  separate fix given the different semantics from the six visitors this PR addresses.

fallintoplace:
Thanks both, that distinction makes sense.

I pushed `d3627bc` for the pieces that still belong in this PR:
- cleaned up `mayContainNans` so the float/double path is explicit
- added `EqualTo` and `IsIn` coverage for the missing-null-count guard

I left `VisitNotEqual` / `VisitNotIn` unchanged here after Tanmay’s @tanmayrauth note, since those branches need the Java-style `containsNullsOnly` / `containsNaNsOnly` behavior rather than the `mayContain*` helper used by the range/equality/IN paths. I opened #1103 for that follow-up so this PR stays scoped to the missing-metric conservative handling.

Could you take another look when you get a chance? @laskoviymishka

laskoviymishka (approved):
Looks good now!

Issue #999, opened 2026-05-05
feat(table): preserve row lineage through copy-on-write and merge-on-read


laskoviymishka:
Parent: #589

Depends on #998. Once new rows get `_row_id`s on first write, the next piece is preserving lineage through rewrites:

- **Copy-on-write:** when a row is rewritten into a new file, its `_row_id` is preserved and `_last_updated_sequence_number` is set to the new snapshot's sequence number.
- **Merge-on-read:** position-delete or DV records the row, but if the same row reappears in an equality-delete-induced rewrite, lineage must thread through.

Tests: scan after a CoW overwrite shows preserved `_row_id`s with bumped `_last_updated_sequence_number`; scan after a MoR delete + reinsert shows new `_row_id` for reinserted rows (because they are logically new).

Spec: [Iceberg row lineage](https://iceberg.apache.org/spec/#row-lineage).

tanmayrauth:
Here's my thinking on the three open questions. Cross-referenced Java's implementation of Spark/Flink write paths.                                                                                                  
                                                                                                                                                                                                                                         
  ---             
1. Where the rewrite path reads source _row_id                                                                                                                                                                                         
                                                                                                                                                                                                                                         
I'd go with scanner projection rather than re-reading source Parquet separately.
                                                                                                                                                                                                                                        
  During a rewrite, we add _row_id and _last_updated_sequence_number to the scan's projected schema. The scanner already synthesizes these from file metadata (first_row_id + position). The writer then stores the values explicitly in  output Parquet instead of leaving them null.                                                                                                                                                                                           
                                                                                                                                                                                                                                         
  Java does the same, SparkTable adds lineage columns to the read schema when isTableRewrite=true. The reader transparently handles both cases: files that already have an explicit _row_id column (from a previous rewrite) use the  stored value; files without it get synthesis. Our synthesizeRowLineageColumns already implements this dual path.
                                                                                                                                                                                                                                         
  The alternative (re-reading source Parquet outside the scanner) would add a second I/O pass, duplicate synthesis logic, and break on multi-rewrite files that already have explicit _row_id. I don't see a reason to diverge from Java  here.
                                                                                                                                                                                                                                         
  I'm thinking a WithPreserveRowLineage() writer optio, the change lives entirely at the scan+write layer, snapshot producer stays unaware.                                                                                            
   
  ---                                                                                                                                                                                                                                    
2. next-row-id accounting for rewrites
                                                                                                                                                                                                                                         
[...]

PR #1099, opened 2026-05-20
feat(table): preserve row lineage through copy-on-write rewrites

tanmayrauth:
Parent: #999                                                                                                                                                                                                   
                                                                                                                                                                                                                 
  Preserves `_row_id` through CoW rewrites and compaction on v3 tables. The scanner synthesizes row IDs from file metadata, the filter is applied post-synthesis to preserve correct positions, and the writer  stores `_row_id` explicitly in output Parquet. `_last_updated_sequence_number` is left null to inherit from the new file's `data_sequence_number`.                                                             
                                                                                                                                                                                                                  
  Also documents that `next-row-id` accounting intentionally overcounts (matching Java's `ManifestListWriter.V4Writer`) and removes the old `Projection()` synthesis helpers replaced by the explicit  `WithRowLineage()` scan option.
                                                                                                                                                                                                                 
  ## Test plan                                                                                                                                                                                                   
   
  - `TestCoWRewritePreservesRowID` : verifies surviving rows keep original `_row_id` after a CoW delete                                                                                                          
  - `TestCoWRewriteRowIDNextRowIDAccounting` : validates next-row-id advances correctly


tanmayrauth:

metadata_columns.go
```diff
@@ -59,3 +59,12 @@ func LastUpdatedSequenceNumber() NestedField {
 func IsMetadataColumn(fieldID int) bool {
 	return fieldID == RowIDFieldID || fieldID == LastUpdatedSequenceNumberFieldID
 }
+
+// SchemaWithRowID returns a new schema that appends the _row_id metadata column
+// to the given schema's fields. Used when reading source files during a CoW rewrite
+// so that row identity is preserved in the output.
+func SchemaWithRowID(s *Schema) *Schema {
+	fields := append(s.Fields(), RowID())
```
laskoviymishka: Two issues here that bite in practice.

First, if the input schema already has a field with ID 1 (e.g. a user column named `id`), `NewSchemaWithIdentifiers` will reject the duplicate and we panic at runtime. We can't assume user schemas avoid 1, 2, 3.

Second, `append(schema.Fields(), ...)` aliases the original schema's backing array when there's spare cap. A later mutation of either schema can corrupt the other. I'd `slices.Clone(schema.Fields())` first, and either skip-or-error when one of the row-lineage IDs collides instead of panicking. wdyt?

table/transaction.go
```diff
@@ -1532,6 +1532,26 @@ func (t *Transaction) rewriteSingleFile(ctx context.Context, fs io.IO, originalF
 		Length: originalFile.FileSizeBytes(),
 	}
 
+	// Preserve row lineage for v3 tables: include _row_id in the scan
+	// projection so it is synthesized (or read from existing explicit values)
+	// and written into the output file. _last_updated_sequence_number is left
+	// out — it will be null in the output and inherited from the new file's
+	// data_sequence_number at read time, which satisfies the spec requirement
```
laskoviymishka: The spec says `last-updated-sequence-number` on a data file is the sequence number of the snapshot in which the row was last updated. Leaving it null on rewrite and inheriting from the new file's `data_sequence_number` conflates "physically rewritten" with "logically updated" and breaks downstream readers that use this for CDC / incremental scans.

For a pure compaction (no row changes), we should preserve the original per-row `_last_updated_sequence_number` from the source files. Otherwise every compaction looks like a full table update to consumers.

table/rewrite_data_files.go
```diff
@@ -289,6 +289,11 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 		scanOpts = append(scanOpts, WitMaxConcurrency(cfg.scanConcurrency))
 	}
 
+	preserveLineage := tbl.metadata.Version() >= 3 && hasRowLineage(group.Tasks)
```
laskoviymishka: Mixing row-lineage and non-row-lineage source files in a single rewrite group is going to produce a file where some rows have `_row_id` / `_last_updated_sequence_number` and others don't, which violates the per-file invariant. `hasRowLineage` returning true if *any* task has lineage means the lineage path runs for the whole group regardless of the mix.

I'd partition the input files by file-level `has_row_lineage` and rewrite each partition separately, or skip non-lineage files when the table has row lineage enabled. Silently merging them feels like a correctness bug waiting to surface.

table/table.go
```diff
@@ -816,6 +816,14 @@ func WithOptions(opts iceberg.Properties) ScanOption {
 	}
 }
 
+// WithRowLineage adds _row_id to the scan projection so that row identity is
```
laskoviymishka: The doc says this is "ignored" on non-v3 tables, but the implementation sets `scan.includeRowLineage = true` unconditionally and `SchemaWithRowID` then runs in `Projection()` regardless of format version. On a v1/v2 table that ends up projecting a metadata column the table doesn't support.

Either the doc is wrong or the guard is missing. I'd add an explicit format-version check (either error or true no-op) and make the doc match the behavior.

table/table.go
```diff
@@ -816,6 +816,14 @@ func WithOptions(opts iceberg.Properties) ScanOption {
...
+// preserved through rewrites. Only meaningful for v3 tables; ignored otherwise.
+func WithRowLineage() ScanOption {
```
laskoviymishka, line 824: The name `WithRowLineage` suggests it controls the whole row-lineage feature, but it only adds `_row_id` to the projection — `_last_updated_sequence_number` is governed elsewhere. That's confusing for callers who'd expect one switch to mean one thing.

Could we either rename to scope it (e.g. `WithRowID`) or extend it to also project `_last_updated_sequence_number`? wdyt?

table/scanner.go
```diff
@@ -261,26 +260,22 @@ func (scan *Scan) Projection() (*iceberg.Schema, error) {
...
-	synthesisMeta := synthesizeMeta(selectedFieldsMeta, schemaMeta)
-	if len(synthesisMeta) > 0 && curVersion >= minFormatVersionRowLineage {
-
-		// synthesis path
-		removedMetaSlice, missingMetaFields := removeMetadataFromSelectedFields(scan.selectedFields, synthesisMeta)
-		sch, err := curSchema.Select(scan.caseSensitive, removedMetaSlice...)
+		schema = curSchema
+	} else {
+		var err error
+		schema, err = curSchema.Select(scan.caseSensitive, scan.selectedFields...)
```
laskoviymishka: This loses the metadata-column escape hatch that the old synthesis path supported. `Select("_row_id")` (or `Select("_last_updated_sequence_number")`) now goes straight to `curSchema.Select` and fails because the user schema doesn't contain those names — the only way to get those columns is `WithRowLineage()`, which is a separate API.

If that's intentional, worth a CHANGELOG note. If not, we should intercept the known metadata column names here before the schema lookup, the way the removed `synthesizeMeta`/`removeMetadataFromSelectedFields` did.

table/transaction.go
```diff
@@ -1788,3 +1843,40 @@ func (s *StagedTable) Refresh(ctx context.Context) (*Table, error) {
 func (s *StagedTable) Scan(opts ...ScanOption) *Scan {
 	panic(fmt.Errorf("%w: cannot scan a staged table", ErrInvalidOperation))
 }
+
+// filterRecordBatch applies an Iceberg filter to a record batch, returning
+// only matching rows. The filter is provided unbound and is bound against the
+// given schema internally. Used for post-synthesis row filtering in CoW rewrites
+// where row lineage must be synthesized before filtering.
+func filterRecordBatch(ctx context.Context, rec arrow.RecordBatch, filter iceberg.BooleanExpression, schema *iceberg.Schema, caseSensitive bool) (arrow.RecordBatch, error) {
```
laskoviymishka: Two things in `filterRecordBatch`:

This is called once per batch from the iterator loop, but `BindExpr` and `substrait.ConvertExpr` depend only on the (schema, filter) pair, not the batch. We're paying that cost N times for no reason — hoist the bind + convert out and pass them in, or memoize on first call.

Also, the `AlwaysFalse` branch returns `array.NewRecordBatch(emptySchema, nil, 0)` with no columns at all — downstream code that does `rec.Column(i)` will panic. The empty record should have the right schema *and* empty arrays for each field.

table/transaction.go
```diff
@@ -1542,11 +1562,19 @@ func (t *Transaction) rewriteSingleFile(ctx context.Context, fs io.IO, originalF
 		return nil, fmt.Errorf("failed to build metadata: %w", err)
 	}
 
+	// When preserving row lineage, omit the row filter from the scanner so that
+	// _row_id is synthesized using the true file-level row positions. The filter
+	// is applied below in the record iterator.
+	scanFilter := iceberg.BooleanExpression(iceberg.AlwaysTrue{})
```
laskoviymishka: Passing `AlwaysTrue` here disables both manifest-level and file-level pushdown for the rewrite scan when row lineage is on. For a selective delete on a wide partition that's a lot of unnecessary IO — we read every file in scope just to filter post-synthesis.

Worst case this is unavoidable because `_row_id` synthesis needs original positions, but the comment treats it as obviously fine. At minimum I'd leave a TODO calling out the cost, ideally we'd still apply file-level (not row-level) skipping based on `boundFilter`'s residuals.

table/rewrite_data_files.go
```diff
@@ -328,6 +336,18 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 	}, nil
 }
 
+// hasRowLineage returns true if any task in the group has a non-nil FirstRowID,
```
laskoviymishka: The comment says "indicating the table has v3 row lineage data to preserve", but the body checks per-task `FirstRowID` — a per-file property, not a table property. A v3 table can legitimately have a mix of tasks where some have `FirstRowID` and some don't (legacy files), so "any task has it" doesn't really mean "the table has lineage".

I'd either rename to `anyTaskHasRowLineage` (and fix the comment to match), or — better — gate on `tbl.metadata.Version() >= 3` at the call site and rewrite this to mean "all tasks have lineage" so we don't trip the mixed-lineage bug above.

table/write_records.go
```diff
@@ -94,6 +95,16 @@ func WithClusteredWrite() WriteRecordOption {
 	}
 }
 
+// WithPreserveRowLineage sets the output file schema to include _row_id so that
+// row identity is preserved through rewrites and compactions on v3 tables. The
+// input records must already contain the _row_id column (e.g. from a scan that
+// projected lineage columns).
+func WithPreserveRowLineage(schema *iceberg.Schema) WriteRecordOption {
```
laskoviymishka, line 110: `WithPreserveRowLineage` is accepted on any table — there's no check that the table actually has v3 row lineage enabled, or that the input records contain `_row_id`. A caller who sets it on a v2 table or feeds records without the lineage column gets silently-wrong output rather than an error.

I'd validate at option-application time (or inside `WriteRecords` once we have the `tbl`): if the table format is < 3, error. If the input Arrow schema is missing `_row_id`, error.

table/row_lineage_rewrite_test.go
```diff
@@ -0,0 +1,188 @@
...
+	filter := iceberg.EqualTo(iceberg.Reference("id"), int64(20))
+	tbl, err = tbl.Delete(ctx, filter, nil)
+	require.NoError(t, err)
+
+	// next-row-id advances by the rewritten manifest's total row count (2 surviving
+	// rows), even though those rows preserve old IDs. This "wastes" ID space but
+	// doesn't violate uniqueness — actual row IDs come from the explicit Parquet
+	// column, not the global counter.
+	nextRowID := tbl.Metadata().NextRowID()
+	assert.GreaterOrEqual(t, nextRowID, int64(5),
```
laskoviymishka: `GreaterOrEqual(..., int64(5))` makes the assertion vacuous in the loose direction — it passes even if next-row-id leapt to some huge value because of a bug elsewhere. And the comment above says the expected accounting is "original + rewritten = 3 + 2 = 5", so we actually know the expected value.

I'd assert `Equal(t, int64(5), nextRowID)` (or whatever the exact post-rewrite expectation is). As written, this test would still pass if next-row-id accounting drifted in either direction.

table/snapshot_producers.go
```diff
@@ -917,6 +917,11 @@ func (sp *snapshotProducer) commit(ctx context.Context) (_ []Update, _ []Require
 			return nil, nil, err
 		}
 		if writer.NextRowID() != nil {
+			// addedRows counts ALL rows in new manifests (existing + added), even
+			// for rewrites where survivors preserve old _row_id values. This
+			// "wastes" ID space but doesn't violate uniqueness: actual row IDs come
+			// from the explicit Parquet column, not the global counter. Java's
+			// ManifestListWriter.V4Writer uses the same accounting.
```
laskoviymishka: `ManifestListWriter.V4Writer` looks like a typo — Java's manifest-list writer for row lineage is `V3Writer` (the v3 format-version writer); there is no V4 manifest. Unless there's a v4 draft I'm missing, the comment should read `V3Writer`.

Misleading provenance comments are extra costly because the next reader trying to verify parity will hunt for a class that doesn't exist.

laskoviymishka (changes requested):
Nice work getting `_row_id` preservation wired through both CoW deletes and `ExecuteCompactionGroup`. The explicit write + read-back path feels like the right shape for v3, and the fresh-table happy-path test is a good start.

I’d still hold this before merge though, a few things need tightening.

Biggest one: `_last_updated_sequence_number`. The comment in `rewriteSingleFile` says leaving it null makes CoW rewrites bump it to the new commit seq. I think the v3 spec says the opposite for surviving unmodified rows: copy the existing non-null value forward. That’s also what Java does in the CoW path by projecting both columns. With the current PR, every preserved row looks “last updated” by the rewrite commit to Java / PyIceberg / iceberg-rust, which breaks the cross-engine watermark semantics.

Second: `SchemaWithRowID`. It panics if the schema already has `_row_id`, which is legal in v3, and `append(s.Fields(), RowID())` can alias the source schema’s backing array. The deleted `TestProjectionV3SchemaWithRowIDOnly` was the guardrail for the first case, so losing that test hides the issue.

Third: compaction gating. `hasRowLineage` returns true if **any** task has lineage. That means a mixed group can go through the preservation path with null and explicit `_row_id`s in one output file. At read time, synthesized IDs for the null rows can collide with explicit IDs from post-lineage rows — exactly what the uniqueness guarantee is supposed to avoid.

What I’d like before merge:

* fix `_last_updated_sequence_number`, or clearly document that we intentionally diverge from v3 / Java
* make `SchemaWithRowID` idempotent and copy the field slice
* change compaction gating to “all tasks have lineage”, or split mixed groups
* gate `WithRowLineage` on v3 in `Projection()`
* bring back the still-relevant scanner tests: `Select("_row_id")`, schema already has `_row_id`, v2 rejection
* add at least one `ExecuteCompactionGroup` test
* make the `next-row-id` accounting test `Equal(5)`, not `GreaterOrEqual(5)`
* hoist the per-batch bind/convert in `filterRecordBatch` out of the iterator

Once these are covered, happy to take another look and approve.

tanmayrauth:
Thanks for the thorough pass,  addressed all 12 inline comments. Going through them in order:

  **1. `metadata_columns.go` : `SchemaWithRowID` panics on duplicate + aliases the input slice.**
  Renamed to `SchemaWithRowLineage`. `slices.Clone`'d the field slice up front so the returned schema cannot share a backing array with the input, and the function skips appending when a reserved field ID (`_row_id` or  `_last_updated_sequence_number`) is already present. New test `TestProjectionV3SchemaAlreadyHasRowID` locks in the no-panic / no-duplicate-id behavior.

  **2. `table/transaction.go`  : `_last_updated_sequence_number` left null breaks CDC semantics.**
  Agreed, fixed. `rewriteSingleFile` now projects `_last_updated_sequence_number` via `SchemaWithRowLineage`, and `scanTask.DataSequenceNumber` is wired in from the source manifest entry's `FileSequenceNum`  (threaded through a new `fileSeqByPath` map returned from `classifyFilesForFilteredDeletions`). The arrow scanner's `synthesizeRowLineageColumns` keeps any explicit per-row value from the source file and  only synthesizes from the file's `data_sequence_number` for null source rows — so a pure CoW rewrite preserves the original "last logically updated" sequence per row instead of stamping the new commit's  sequence onto every survivor.  `TestCoWRewritePreservesRowID` was updated to assert this directly.
  
  **3. `table/rewrite_data_files.go`  : `hasRowLineage` permits mixed groups.**
  Tightened to `allTasksHaveRowLineage` (renamed function in the other thread too). When even one task in the group lacks `FirstRowID`, the entire group skips the lineage path — avoids the "some explicit  `_row_id`s, some nulls" output you flagged. Splitting mixed groups so each side rewrites with its own semantics is the cleaner long-term fix but requires changing `ExecuteCompactionGroup`'s output shape ,left as follow-up; called it out in the gating comment so we don't lose the thread.

  **4. `table/table.go`  : `WithRowLineage` doc says "ignored on non-v3" but impl projects regardless.**
  `Projection()` now returns an explicit error when `includeRowLineage` is set on a v1/v2 table, and the  `WithRowLineage` doc was rewritten to match. New test `TestProjectionWithRowLineageRequiresV3` covers  it.

  **5. `table/table.go`  : `WithRowLineage` name suggests whole-feature control.**
  Went with the "extend it" branch you suggested: `SchemaWithRowLineage` (the helper `WithRowLineage` calls) now projects both `_row_id` and `_last_updated_sequence_number`. So one switch ↔ both lineage  columns ↔ preservation through rewrite. Kept the name `WithRowLineage` since the semantics now match it.

  **6. `table/scanner.go`  : losing the `Select("_row_id")` escape hatch.**
  Restored. `Projection()` now intercepts row-lineage column names (case-aware) in the user's `selectedFields`, strips them before `curSchema.Select`, and appends the corresponding `RowID()` /  `LastUpdatedSequenceNumber()` fields back onto the projected schema — but only on v3 tables. On v1/v2 the names fall through to `curSchema.Select`, which fails as before with `ErrInvalidSchema`. Restored `TestProjectionV3SelectRowLineageColumns` and `TestProjectionRowLineageRejectedOnV1V2` to lock both behaviors in.

  **7. `table/transaction.go`  : `filterRecordBatch` rebinds per batch + `AlwaysFalse` returns schemaless record.**
  once for the whole rewrite (the complement filter doesn't change between files) and passes the resulting closure into every `rewriteSingleFile` call. The `AlwaysFalse` fast-path now returns `rec.NewSlice(0,  0)` so downstream code that calls `rec.Column(i)` doesn't panic on the empty result.
  
  **8. `table/transaction.go`  : `AlwaysTrue` disables pushdown for the rewrite scan.**
  Left a `TODO(perf)` comment on `rewriteSingleFile` calling out the IO cost of disabling pushdown when preserving lineage, and noted that file-level skipping (against the bound filter's residual on file  stats) could be re-enabled here without breaking `_row_id` synthesis since synthesis only depends on within-file row positions. Punting the actual implementation to a follow-up — agreed it's worth fixing but didn't want to expand this PR into a metrics-eval refactor.
  
  **9. `table/rewrite_data_files.go`  : `hasRowLineage` comment claims a table property but checks per-task.**
  Renamed to `allTasksHaveRowLineage` and rewrote the comment. The function now also returns `false` on an empty task list (defensive), and the call-site comment explains why we gate on "all" rather than "any" (mixed groups → per-file invariant violation, see thread 3).
  
  **10. `table/write_records.go`  : `WithPreserveRowLineage` accepts anything.**
  Added validation in `WriteRecords`: errors when the table format version is below 3, and errors when the input arrow schema doesn't include the `_row_id` column. Docstring rewritten to call this out. Implementation-wise this required a small reshuffle: `cfg` is now built before `checkArrowSchemaCompat` runs so the compat check can be against the projected (with-lineage) schema rather than `tbl.Schema()` — otherwise the compat check itself would reject `_row_id` as unknown.
  
  **11. `table/row_lineage_rewrite_test.go`  : `GreaterOrEqual(5)` is vacuous.**
  Tightened to `assert.Equal(t, int64(5), tbl.Metadata().NextRowID())`. The comment above the assertion already had the exact expected value (3 + 2), so the loose form was just hiding regressions in either  direction.
  
  **12. `table/snapshot_producers.go`  : `ManifestListWriter.V4Writer` doesn't exist.**
  Typo, fixed to `V3Writer`. Same fix applied to the matching comment in the test that referenced V4Writer for context.
  
  ---
  
  **Newly added tests:**
  - `TestProjectionV3SelectRowLineageColumns`  : `Select("_row_id")` works on v3
  - `TestProjectionRowLineageRejectedOnV1V2`  : same select on v1/v2 errors with `ErrInvalidSchema`
  - `TestProjectionWithRowLineageRequiresV3`  : `WithRowLineage` errors on v1/v2 
  - `TestProjectionV3SchemaAlreadyHasRowID`  : idempotency under reserved-id collision
  - `TestExecuteCompactionGroupPreservesRowID`  : end-to-end compaction on a v3 table preserves `_row_id` for every row across two source files
  
 @laskoviymishka  PTAL when you have a moment.

zeroshade:
Rebasing should fix the integration tests

table/transaction.go
```diff
@@ -1488,13 +1501,23 @@ func (t *Transaction) classifyFilesForFilteredDeletions(ctx context.Context, fs
 					localDelete = append(localDelete, df)
 				} else {
 					localRewrite = append(localRewrite, df)
+					// Capture the file's sequence number from the
+					// manifest entry so the rewrite path can synthesize
+					// _last_updated_sequence_number for source rows
+					// that have a null value (or no column) in the file.
+					if fseq := entry.FileSequenceNum(); fseq != nil {
```
laskoviymishka: I think this is reading the wrong manifest field. Spec §Row Lineage says when `_last_updated_sequence_number` is null on read, it gets assigned the manifest entry's `sequence_number` (field id 3, the data sequence number) — not `file_sequence_number` (field id 4, the snapshot that added the file). For back-dated files where they diverge, a Java or iceberg-rust reader would compute a different value for the same row.

I'd switch both this site and the matching one in `scanner.go` to `entry.SequenceNum()`. `SequenceNum()` returns `int64` not `*int64`, so the nil guard collapses — drop the conditional or guard on `>= 0`. While we're here, I'd rename `FileScanTask.DataSequenceNumber` if that's where this plumbs through, so the field name matches what it actually holds.

table/scanner_internal_test.go
```diff
@@ -238,98 +236,71 @@ func TestBuildPartitionEvaluatorWithInvalidSpecID(t *testing.T) {
...
-// columns are absent from the schema).
-func TestProjectionV3PreLineageFile(t *testing.T) {
-	schema := iceberg.NewSchema(
-		1,
-		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64, Required: true},
-		iceberg.NestedField{ID: 2, Name: "payload", Type: iceberg.PrimitiveTypes.String, Required: false},
-	)
+// TestSynthesizeRowLineageColumns verifies that _row_id and _last_updated_sequence_number
+// are filled from task constants when those columns are present and null.
+func TestSynthesizeRowLineageColumns(t *testing.T) {
```
laskoviymishka, line 241: The test only constructs all-null inputs, so it doesn't actually prove the property the feature depends on. If `synthesizeRowLineageColumns` had a bug where it overwrote pre-existing non-null values, this test would still pass — and so would `TestCoWRewritePreservesRowID`, because fresh appends produce all-null source columns too.

The case that matters is the mixed one: some rows have explicit `_row_id` / `_last_updated_sequence_number` values from a prior write, the rest are null. That only arises after multiple rewrites of files that already carry explicit values, which is exactly the scenario the spec's null-coalescing rule exists for.

I'd add a fixture that builds a record with three rows — one explicit non-null, one null, one explicit non-null with a different value — and asserts the non-null ones survive untouched while the null one picks up `firstRowID + position`. Same shape for `_last_updated_sequence_number`.

table/rewrite_data_files.go
```diff
@@ -289,6 +289,20 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
...
+	// Preserve row lineage only when every source file in the group carries
+	// it. A mixed group (some files with FirstRowID, some without — e.g.
+	// legacy files on a v3 table) would otherwise produce one output where
+	// post-lineage rows have explicit _row_id values and pre-lineage rows
+	// have nulls, which violates the per-file uniqueness/coverage
+	// invariant the v3 spec requires. Splitting mixed groups into separate
+	// outputs is a larger refactor and is left as a follow-up; for now we
+	// degrade gracefully (the rewrite still succeeds, but lineage is not
+	// preserved for the surviving rows).
+	preserveLineage := tbl.metadata.Version() >= 3 && allTasksHaveRowLineage(group.Tasks)
```
laskoviymishka, line 302: Mixed-lineage groups silently fall through to the non-lineage path. Compaction succeeds, the output is valid, but `_row_id` is gone — no log, no metric, no flag in `CompactionGroupResult`. For tables migrating to v3, every compaction pass over a group that contains any legacy file will quietly drop lineage on the surviving rows. That's the common case during migration, not the edge case.

I'd at minimum add a `slog.Warn` matching the v3 pos-delete pattern, and ideally a `RowLineagePreserved bool` on `CompactionGroupResult` so callers can observe it.

table/scanner.go
```diff
@@ -261,26 +262,96 @@ func (scan *Scan) Projection() (*iceberg.Schema, error) {
...
+		// Select would fail with "could not find column" on v3 tables
+		// where they are otherwise legal to project. The scanner reads
+		// them from file metadata (or synthesizes them) at scan time;
+		// here we just need to ensure they survive into the projection.
+		//
+		// On v1/v2 tables, Select is left to fail naturally — those
+		// versions don't have row lineage, so requesting these columns
+		// is an error.
+		userFields, lineageFields := splitLineageMetadataFields(scan.selectedFields, scan.caseSensitive)
+		if len(lineageFields) > 0 && curVersion < minFormatVersionRowLineage {
```
laskoviymishka, line 282: The rejection here works only because `Select` happens to fail when `_row_id` isn't in the v1/v2 schema — the user sees `ErrInvalidSchema` rather than a clear "row lineage requires v3". And it's fragile: if a future v2 schema ever carries a column literally named `_row_id` for unrelated reasons, this short-circuits in the wrong direction.

I'd return an explicit error here for the `curVersion < minFormatVersionRowLineage && len(lineageFields) > 0` case, naming the format version requirement. Same logic, but the contract is in the code rather than emergent from `Select`'s error path.

table/scanner.go
```diff
@@ -261,26 +262,96 @@ func (scan *Scan) Projection() (*iceberg.Schema, error) {
...
+		schema, err = curSchema.Select(scan.caseSensitive, userFields...)
 		if err != nil {
 			return nil, err
 		}
+		if len(lineageFields) > 0 {
+			schema = appendMissingLineageFields(schema, lineageFields)
+		}
+	}
+
+	if scan.includeRowLineage {
```
laskoviymishka, line 305: There are now two code paths that select lineage columns into the projection: the `includeRowLineage` flag path (calls `SchemaWithRowLineage`) and the explicit-field-name path (calls `appendMissingLineageFields`). Both run when the user supplied lineage field names *and* set `WithRowLineage`, producing a redundant pass.

More concerning, they have subtly different error behavior — one rejects on v1/v2 with a clear error, the other relies on `Select` to fail. Anyone adding a third lineage column down the road has to update both paths and both test sets.

I'd canonicalize on one path: detect lineage fields in `selectedFields` up front, set `scan.includeRowLineage = true` if any are present, drop them from `selectedFields`, and from there only the `includeRowLineage` branch runs. wdyt?

table/transaction.go
```diff
@@ -1798,3 +1905,53 @@ func (s *StagedTable) Refresh(ctx context.Context) (*Table, error) {
...
+			return rec.NewSlice(0, 0), nil
+		}, nil
+	}
+
+	extSet, substraitFilter, err := substrait.ConvertExpr(schema, bound, caseSensitive)
+	if err != nil {
+		return nil, fmt.Errorf("prepareBatchFilter: convert expression: %w", err)
+	}
+
+	ctx = exprs.WithExtensionIDSet(ctx, exprs.NewExtensionSetDefault(*extSet))
```
laskoviymishka: The closure captures `ctx` at construction time with the extension-ID set baked in, but per-batch allocator and deadline values from the call-site ctx are then lost. If `filterRecords` resolves the allocator from ctx, intermediate arrays end up on the construction-time allocator and `memory.NewCheckedAllocator` accounting goes sideways under test.

I'd take the call-site ctx as a parameter on the returned closure, derive the extension-set context inside, and let allocator/deadline propagate.

While we're in here: the positional-binding assumption (lineage fields at indices N..N+1 in the record batch, filter bound against indices 0..N-1) is correct but undocumented. Worth a one-line comment so a future change to lineage column placement doesn't silently break filter evaluation.

table/transaction.go
```diff
@@ -1530,14 +1565,50 @@ func (t *Transaction) rewriteFilesWithFilter(ctx context.Context, fs io.IO, upda
 	return nil
 }
 
-// rewriteSingleFile reads a single data file, applies the filter, and writes new files with filtered data
-func (t *Transaction) rewriteSingleFile(ctx context.Context, fs io.IO, originalFile iceberg.DataFile, filter iceberg.BooleanExpression, caseSensitive bool, commitUUID uuid.UUID, concurrency int) ([]iceberg.DataFile, error) {
+// rewriteSingleFile reads a single data file, applies the filter, and writes new files with filtered data.
+// fileSeqNum is the source file's file_sequence_number from its manifest entry; required to synthesize
+// _last_updated_sequence_number for rows whose value is null in the source file.
+func (t *Transaction) rewriteSingleFile(ctx context.Context, fs io.IO, originalFile iceberg.DataFile, fileSeqNum *int64, filter iceberg.BooleanExpression, postFilter func(arrow.RecordBatch) (arrow.RecordBatch, error), caseSensitive bool, commitUUID uuid.UUID, concurrency int) ([]iceberg.DataFile, error) {
```
laskoviymishka: Nine positional parameters with several same-typed adjacents (`commitUUID, concurrency`, and now two same-shaped filter closures) — no compile-time order protection if anyone swaps a pair. I'd pull these into a `rewriteSingleFileArgs` struct.

Separate but related: `BindExpr` runs unconditionally in `rewriteSingleFile`, but when `preserveRowLineage == true` the bound filter is unused (scanFilter becomes `AlwaysTrue{}`) and the same expression is bound again inside `prepareBatchFilter`. With `preserveRowLineage == false`, `prepareBatchFilter` runs `BindExpr` + `ConvertExpr` but `postFilter` is never called. One of those two binds is always wasted.

table/transaction.go
```diff
@@ -1798,3 +1905,53 @@ func (s *StagedTable) Refresh(ctx context.Context) (*Table, error) {
...
+// it to substrait once, returning a per-batch filter function that can be
+// reused across every record batch. The setup work (BindExpr, ConvertExpr) is
+// independent of the batch and is the most expensive part of filter-eval, so
+// hoisting it out of the iterator loop is a measurable win on rewrites that
+// produce many batches.
+//
+// The returned function takes ownership of the input batch (it releases it on
+// the AlwaysFalse fast-path) and returns a possibly-new batch the caller is
+// responsible for releasing.
+func prepareBatchFilter(ctx context.Context, filter iceberg.BooleanExpression, schema *iceberg.Schema, caseSensitive bool) (func(arrow.RecordBatch) (arrow.RecordBatch, error), error) {
```
laskoviymishka: The `AlwaysFalse` fast-path does `defer rec.Release()`; the `AlwaysTrue` / bound-nil paths return `rec` unchanged with no Retain. So depending on which fast-path triggers, the caller either does or doesn't own a reference to the returned record. That's two contracts for one return type, and the next caller to add a `Release()` will double-free on the AlwaysFalse path.

I'd have every fast-path do a `rec.Retain()` so the caller always owns exactly one reference on the result, mirroring the bound-filter path's semantics.

metadata_columns.go
```diff
@@ -59,3 +61,38 @@ func LastUpdatedSequenceNumber() NestedField {
...
+
+// SchemaWithRowLineage returns a new schema with the row-lineage metadata columns
+// (_row_id, _last_updated_sequence_number) appended to the given schema's fields.
+// Used when reading source files during a CoW rewrite or compaction so that row
+// identity and per-row update sequence are preserved in the output.
+//
+// Idempotent: if a row-lineage column is already present (by reserved field ID),
+// it is not appended again. The returned schema always allocates a fresh field
+// slice so it cannot alias the input schema's backing array.
+func SchemaWithRowLineage(s *Schema) *Schema {
```
laskoviymishka, line 74: Two things on the new helper.

The comment a few lines up reads as if the second reserved ID is literally 108. Should be "Integer.MAX_VALUE - 107 and Integer.MAX_VALUE - 108" — `2147483540` and `2147483539` respectively.

More substantively, `SchemaWithRowLineage` lives in the top-level `iceberg` package, which makes it part of the v1 public API surface. Java keeps the equivalent logic internal. I'd consider an unexported `schemaWithRowLineage` in `table`, or a method on `*Schema`, so we're not locked into the helper's exact shape if v4 lineage semantics shift. Same for `WithPreserveRowLineage(schema)` — making callers construct the schema feels backwards; could it take no arg and derive from the table internally?

Also worth a nil guard on `s` at the top here — it's exported, a nil receiver will panic on `s.Fields()`.

table/row_lineage_rewrite_test.go
```diff
@@ -0,0 +1,291 @@
...
+	filter := iceberg.EqualTo(iceberg.Reference("id"), int64(20))
+	tbl, err = tbl.Delete(ctx, filter, nil)
+	require.NoError(t, err)
+
+	// next-row-id advances by 3 (original) + 2 (rewritten survivors) = 5,
+	// even though the surviving rows preserve their old IDs. This "wastes"
+	// ID space but doesn't violate uniqueness — actual row IDs come from
+	// the explicit Parquet column, not the global counter.
+	assert.Equal(t, int64(5), tbl.Metadata().NextRowID(),
+		"next-row-id should advance by original (3) + rewritten (2) = 5")
```
laskoviymishka: The comment is arithmetically misleading — the advance is `2` (the new manifest's added rows), and `5 = firstRowID(3) + 2`. The "3" is the starting value, not an addend. Worth rewriting the comment to match the actual computation, and linking to the `snapshot_producers.go` site that mirrors `ManifestListWriter.V3Writer.prepare()` in Java.

Separately, I'd double-check `snapshot.AddedRows` here — Java's `added-rows` summary field is the actual added row count, not `existing + added`. If we're stamping the Java-equivalent of `existing + added` into the snapshot summary, downstream tools reading that field will see inflated numbers.

laskoviymishka (changes requested):
Almost there. The big pieces are in now: `SchemaWithRowLineage` is extracted, `synthesizeRowLineageColumns` has its first unit test, the CoW path carries `_row_id` end-to-end, and `TestExecuteCompactionGroupPreservesRowID` covers the compaction path. Nice cleanup.

I’d still hold merge for two correctness fixes:

* `_last_updated_sequence_number` is synthesized from `file_sequence_number`, but it should use `data_sequence_number`.
* `synthesizeRowLineageColumns` only has an all-null fixture right now, so we still don’t test the null-coalescing behavior this feature depends on.

Both look narrow, but they matter for cross-engine read parity.

A few smaller things are inline too: mixed-group compaction can still silently drop lineage, the `WithRowLineage + selectedFields` v1/v2 path needs tightening, and `prepareBatchFilter` captures `ctx` a bit oddly.

table/row_lineage_rewrite_test.go
tanmayrauth: On snapshot.AddedRows: the existing comment in snapshot_producers.go:920 explicitly says it matches Java's ManifestListWriter.V3Writer accounting (counts existing + added in new manifests). Worth a  separate look since if Java's snapshot summary added-rows field actually means "newly added rows only," that's a pre-existing bug in the v3 writer rather than something this PR introduced. Want me to file  that as a separate issue?

metadata_columns.go
tanmayrauth: On unexporting: SchemaWithRowLineage is already used by callers outside the table package (compaction's ExecuteCompactionGroup builds the projected schema this way before passing it to  WithPreserveRowLineage), so unexporting it now means duplicating the helper or carving out a different API. Worth doing as a follow-up if v4 lineage actually changes the shape, but I'd rather not  preemptively redesign for a hypothetical. Same reasoning on WithPreserveRowLineage(schema) — having the caller construct the schema once keeps scan-side and write-side projections in lockstep, which   matters for the compaction path where both have to match exactly.

table/transaction.go
tanmayrauth: I think the current contract is consistent: input ownership transfers to output, caller releases once. Adding Retain() on the AlwaysTrue path would leak the input (caller would have to release input AND output, but the contract today is "after postFilter(rec), only release the result").

table/transaction.go
tanmayrauth: Bundled the params into rewriteSingleFileArgs. On the wasted bind: hoisted BindExpr into the !preserveRowLineage branch so the per-file bind is no longer wasted on the lineage path. Left the   prepareBatchFilter bind alone — it runs once per rewriteFilesWithFilter call rather than per file, so the cost is amortized across all rewritten files in the operation.

table/transaction.go
tanmayrauth: Changed the closure signature to func(context.Context, arrow.RecordBatch) — extSet is pre-computed once and reused, but the per-call ctx (with allocator/deadline) is attached inside the closure on each  call. Also added a one-line comment on the positional-binding assumption (lineage fields at indices N..N+1, filter bound against 0..N-1).

table/scanner.go
tanmayrauth: Did the smaller piece — skipping appendMissingLineageFields when scan.includeRowLineage is set so the redundant pass goes away. Pushed back on the full canonicalization (auto-setting  includeRowLineage=true whenever any lineage name appears in selectedFields): selecting only _row_id today gives you only _row_id, but with the rewrite it would also pull in _last_updated_sequence_number.  That's a behavior change for the legacy Select(_row_id) escape hatch. The maintainability concern (third column down the road) is mostly handled by splitLineageMetadataFields + SchemaWithRowLineage already centralizing the column list.

table/scanner.go
tanmayrauth: Done, explicit ErrInvalidOperation with the format version named, rather than relying on Select's ErrInvalidSchema. Updated TestProjectionRowLineageRejectedOnV1V2 to match.

table/rewrite_data_files.go
tanmayrauth: Added a slog.Warn matching the v3 pos-delete pattern in ExecuteCompactionGroup when a v3 table sees a mixed group with at least one lineage file. Holding off on the RowLineagePreserved bool on  CompactionGroupResult for now — the log surfaces the silent loss, and adding to the result struct locks in API surface for what should be a transitional migration concern. Happy to add it as a follow-up if it shows up in practice.

table/scanner_internal_test.go
tanmayrauth: Added TestSynthesizeRowLineageColumnsPreservesExplicit with the 3-row mixed fixture (explicit / null / explicit-different) for both _row_id and _last_updated_sequence_number. Asserts the non-null rows  survive untouched and the null row picks up firstRowID + position / task.DataSequenceNumber.

table/transaction.go
tanmayrauth: Good catch, switched both sites to entry.SequenceNum() and added a guard on >= 0 (since SequenceNum() returns -1 for unset). Left FileScanTask.DataSequenceNumber named as-is since the field name  already matches what we're now putting in it (data sequence number); the bug was the source, not the destination.

tanmayrauth:
@laskoviymishka  PTAL when you have a moment.

laskoviymishka (approved):
Thanks for the quick next round. Overall, this looks good to land.

The `SequenceNum()` fix and `TestSynthesizeRowLineageColumnsPreservesExplicit` cover the two main blockers. The `slog.Warn` on mixed-lineage compaction also surfaces the silent-drop case I was worried about.

The three inline pushbacks all look fine to me:

* `postFilter` ownership: fair. Re-reading the cases, the contract really is “one release per batch on every path”.
* Keeping `SchemaWithRowLineage` exported: agreed. `ExecuteCompactionGroup` is already a cross-package caller, and the nil guard plus the v4 follow-up plan is enough.
* `snapshot.AddedRows` accounting: agreed, this is pre-existing `V3Writer` parity, not this PR’s concern.

🚢 👍

PR #1100, opened 2026-05-20
feat(table): add SerializeDV and DVWriter for deletion vector authoring

tanmayrauth:
Adds the serialization half of the deletion vector codec (inverse of DeserializeDV) and a DVWriter that accumulates positions per data file and flushes them as a single Puffin file with one blob per data file.

Returned DataFile entries carry ContentOffset, ContentSizeInBytes, and ReferencedDataFile so they can be passed directly to RowDelta.AddDeletes().

Part 1 of #997

### DIFF
--- a/io/mem.go
+++ b/io/mem.go
@@ -60,6 +60,11 @@ type MemFS struct {
 	files map[string][]byte // keyed by full URI
 }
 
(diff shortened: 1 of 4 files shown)


tanmayrauth:

zeroshade (approved):
LGTM

PR #1101, opened 2026-05-20
feat(table): reject AddDataFiles on v3 when first_row_id is missing (#1000)

### LINKED ISSUE #1000: feat(table): reject AddDataFiles on v3 when first_row_id is missing
Parent: #589

Depends on #998. When users call `Transaction.AddDataFiles` with parquet files written outside iceberg-go on a v3 table, those files have no `first_row_id` and the writer cannot retroactively fabricate one without rescanning every file. Pyiceberg's posture is to error out unless the caller explicitly sets `first_row_id` per file at registration time. Mirror that.

Touches: `table/transaction.go` `AddDataFiles` (or wherever the `filesToDataFiles` validator lives) — branch on `formatVersion >= 3` and require a non-zero `first_row_id` per file. Error message should point users at supplying the row-id range explicitly so they understand the workflow.

Tests: `AddDataFiles` without `first_row_id` on v3 returns a clear error; with `first_row_id` succeeds and metadata `next-row-id` advances correctly. v2 path unchanged. Spec: [Iceberg row lineage](https://iceberg.apache.org/spec/#row-lineage).

tanmayrauth:
Closes #1000                                                                                                                                                                                                   
                                                                                                                                                                                                                 
  When users call `AddDataFiles` (or `ReplaceDataFilesWithDataFiles` / `ReplaceFiles`) with externally-written parquet files on a v3 table, reject the operation if `first_row_id` is not set. This mirrors      
  pyiceberg's behavior — the library cannot fabricate row IDs retroactively, so callers must supply them explicitly via `DataFileBuilder.FirstRowID()`.
                                                                                                                                                                                                                 
  ## Changes                                                                                                                                                                                                     
   
  - Added validation in `validateDataFilesToAdd` to require `first_row_id` on format version >= 3                                                                                                                
  - Error message directs users to `DataFileBuilder.FirstRowID()`
  - v2 tables are unaffected

### DIFF
--- a/table/snapshot_producers_test.go
+++ b/table/snapshot_producers_test.go
@@ -968,3 +968,47 @@ func TestComputeOwnManifests_ParentManifestsIOError(t *testing.T) {
 			"swallows the IO error and returns a programming-bug error would fail this test")
 	require.Nil(t, got, "error path must return nil manifest slice")
 }
+
+func TestAddDataFilesV3RejectsWithoutFirstRowID(t *testing.T) {
+	spec := iceberg.NewPartitionSpec()
+	txn, _ := createTestTransactionWithMemIO(t, spec)
+	txn.meta.formatVersion = 3
+
+	df := newTestDataFile(t, spec, "file://data.parquet", nil)
+
+	err := txn.AddDataFiles(context.Background(), []iceberg.DataFile{df}, nil)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "missing first_row_id")
+	require.ErrorContains(t, err, "required for v3 tables")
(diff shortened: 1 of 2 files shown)


tanmayrauth:

zeroshade (approved):
LGTM and narrowly scoped. I'll wait for @laskoviymishka to take a look before merging as I haven't dug into the pyiceberg behavior on this to make sure we're matching

tanmayrauth:
@laskoviymishka could you please review it?

laskoviymishka (approved):
🛳️ 👍

PR #355, opened 2025-03-21
refactor: Slight refactor of the integration test

Fokko:
This makes it easier to track where the tables come from.

PR #1102, opened 2026-05-20
fix(manifest): stop writing deprecated distinct_counts (field-id 111)

dhananjaykrutika:
The Avro wire schema declared `distinct_counts` on data_file v1 and v2, causing it to be emitted on every manifest entry. The Iceberg spec marks this field as "Deprecated. Do not write." (https://github.com/apache/iceberg/blob/main/format/spec.md?plain=1#L667), so writers should not emit it. The field stands deprecated since 2020 -- https://github.com/apache/iceberg/issues/767#issuecomment-582192665

Strict readers that don't include field-id 111 in their data_file read schema (e.g. PyIceberg 0.10) fail to resolve manifests written by this library with: ResolveError "File/read schema are not aligned for map, got None".


dhananjaykrutika:
The integration test failure comes from PR #355. The change at https://github.com/apache/iceberg-go/pull/1105 fixes this. So this test should pass once #1105 is merged.

twuebi:
cc @laskoviymishka

laskoviymishka (changes requested):
Nice, narrow fix, dropping field `111` from the v1/v2 Avro schemas is the right enforcement layer, and the PyIceberg 0.10 unblock is a very real reason to do it.

I’d hold merge for a couple of cleanup points though. Main one: removing `TestWriteManifestV{1,2}KeepsDistinctCounts` entirely feels a bit overkill :D

We no longer want to **write** `distinct_counts`, but we should still be able to **read** legacy manifests that already have field `111` on the wire. The `dataFile` struct still has the Avro tag, so older iceberg-go / Spark-written manifests should decode fine — but after deleting those tests, nothing pins that behavior.

I’d add a small back-compat test that bypasses `WriteManifest`, builds/uses a raw OCF fixture with the old schema and field `111` present, then reads it through `ReadManifest` and asserts the map comes back.

A few smaller things before merge:

* update `MarshalAvroEntry` / `EncodeDataFile` godoc: `distinct_counts` is dropped on encode for all versions now, not just v3
* rename/broaden `TestWriteManifestV3OmitsDistinctCounts` to cover v1/v2/v3
* either remove the v3-only `prepareEntry` special case, or mirror the nil-clear for v1/v2 too
* add `// Deprecated:` godoc on `DataFileBuilder.DistinctValueCounts`, since it’s now a public setter that won’t encode

Fix itself looks right — I’d just keep the legacy read coverage instead of deleting the whole old test shape.

zeroshade:
@dhananjaykrutika I hope you don't mind that I'm gonna address @laskoviymishka's comments so we can get this across the finish line so that we can create a new RC as this issue is causing us to have an incompatibility with pyiceberg

laskoviymishka (approved):
🚢 🫰

PR #1104, opened 2026-05-20
feat(catalog): add PurgeTable as optional interface for physical file deletion

### LINKED ISSUE #1092: Support purge in DropTable for physical file deletion

I find that there is no `purge` capability built into `iceberg-go`'s drop operation. is that intentional?

### Questions

1. What is the recommended approach for physical file cleanup after `DropTable` today?
2. Is there a plan to add a `purge bool` parameter to `DropTable` (like Java) or a dedicated `PurgeTable()` method?

### Context

It would be great to have a robust physical cleanup mechanism for `DropTable`, similar to how `ExpireSnapshots` in v0.6.0 now handles cleanup via `PostCommit`.

Please let me know whether creating a pr for this would be helpful

mithilgirish:
Fixes #1092

## Summary
Adds `PurgeableTable` as an optional interface so catalogs can support
physical deletion of table files when dropping a table.

Today, `DropTable` only removes the catalog metadata entry — physical
files (Parquet, manifests, metadata JSON) remain on storage. This PR
closes that gap for client-side catalogs without making any breaking
changes to the `Catalog.DropTable` signature.

### Core
- `catalog/catalog.go` — add `PurgeableTable` optional interface
- `catalog/internal/utils.go` — add `PurgeTableFiles` shared helper
  using `WalkDir` + `BulkRemovableIO` fallback for storage-agnostic deletion

### Per-catalog
- **SQL**: load table metadata, drop catalog entry, walk and delete all
  files under the table location via `WalkDir` + `BulkRemovableIO` fallback
- **Glue**: same pattern; falls back to metadata-only drop if table files
  are inaccessible (e.g. S3 permissions)
- **Hive**: same pattern; falls back to metadata-only drop on load failure
- **Hadoop**: delegates to `DropTable` which already removes the directory tree
- **REST**: existing `PurgeTable` (`purgeRequested=true`) satisfies the
  interface automatically — no changes needed

### CLI
- `cmd/iceberg/main.go` — wire `--purge` flag into `drop table` command
  via type assertion against `PurgeableTable`; exits with an error if the
  catalog does not support client-side purge

## Notes
- No breaking changes to `Catalog.DropTable` signature
- Client-side purge is best-effort — if interrupted between catalog drop
  and file deletion, orphaned files may remain. For reliable cleanup,
  prefer `expire-snapshots` + `remove-orphan-files`
- REST catalog delegates purge to the server as before

## Testing
- `catalog/sql/sql_test.go` — verifies physical file deletion on local storage
- `catalog/glue/glue_test.go` — mock test for purge invocation and fallback
- `catalog/hive/hive_test.go` — mock test for purge invocation and fallback
- `cmd/iceberg/drop_test.go` — CLI flag parsing and purge dispatch tests

### DIFF
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -168,7 +168,21 @@ type TransactionalCatalog interface {
 }
 
 // PurgeableTable is an optional interface that catalogs can implement
-// to support physical table deletion.
+// to support physical table deletion (catalog entry + underlying files).
+// Callers should check for this capability via a type assertion:
+//
+//	if purger, ok := cat.(catalog.PurgeableTable); ok {
+//	    err := purger.PurgeTable(ctx, ident)
+//	}
+//
+// For REST catalogs the purge is delegated server-side. For client-side
+// catalogs (SQL, Glue, Hive, Hadoop) the table is first dropped from
+// the catalog, then all files under the table's [table.Metadata.Location]
+// root, plus any referenced files written outside the root (e.g. via
+// write.data.path or write.metadata.path table properties), are deleted.
+// File-deletion errors are propagated to the caller, but because the
+// catalog entry is already removed at that point there is no automatic
+// retry path.
 type PurgeableTable interface {
 	PurgeTable(ctx context.Context, identifier table.Identifier) error
 }
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -167,6 +167,12 @@ type TransactionalCatalog interface {
 	CommitTransaction(ctx context.Context, commits []table.TableCommit) error
 }
 
+// PurgeableTable is an optional interface that catalogs can implement
+// to support physical table deletion.
+type PurgeableTable interface {
+	PurgeTable(ctx context.Context, identifier table.Identifier) error
+}
+
 func ToIdentifier(ident ...string) table.Identifier {
 	if len(ident) == 1 {
 		if ident[0] == "" {
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -179,10 +179,13 @@ type TransactionalCatalog interface {
 // catalogs (SQL, Glue, Hive, Hadoop) the table is first dropped from
 // the catalog, then all files under the table's [table.Metadata.Location]
 // root, plus any referenced files written outside the root (e.g. via
-// write.data.path or write.metadata.path table properties), are deleted.
-// File-deletion errors are propagated to the caller, but because the
-// catalog entry is already removed at that point there is no automatic
-// retry path.
+// write.data.path or write.metadata.path table properties), are physically deleted.
+//
+// File physical deletion is performed on a best-effort basis. If gc.enabled is
+// set to false on the table metadata, physical deletion is skipped entirely.
+// Any file-deletion, directory walk, or listing errors encountered are logged
+// as warnings, and the operation succeeds (returns nil) so that the catalog drop
+// remains the single source of truth.
 type PurgeableTable interface {
 	PurgeTable(ctx context.Context, identifier table.Identifier) error
 }
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -408,11 +408,7 @@ func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) er
 func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) error {
 	tbl, err := c.LoadTable(ctx, identifier)
 	if err != nil {
-		if errors.Is(err, catalog.ErrNoSuchTable) {
-			return err
-		}
-		// If we can't load the table, fall back to dropping metadata only
-		return c.DropTable(ctx, identifier)
+		return err
 	}
 
 	// Drop the table entry from the catalog first
@@ -422,7 +418,9 @@ func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) e
 	}
 
 	// Physically delete all table files on storage
-	_ = internal.PurgeTableFiles(ctx, tbl)
+	if purgeErr := internal.PurgeTableFiles(ctx, tbl); purgeErr != nil {
+		return fmt.Errorf("dropped table but failed to purge files: %w", purgeErr)
+	}
 
 	return nil
 }
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"iter"
+	"log"
 	"maps"
 	"strconv"
 	"strings"
@@ -136,6 +137,8 @@ type glueAPI interface {
 	DeleteDatabase(ctx context.Context, params *glue.DeleteDatabaseInput, optFns ...func(*glue.Options)) (*glue.DeleteDatabaseOutput, error)
 	UpdateDatabase(ctx context.Context, params *glue.UpdateDatabaseInput, optFns ...func(*glue.Options)) (*glue.UpdateDatabaseOutput, error)
 }
+
+var _ catalog.PurgeableTable = (*Catalog)(nil)
 
 type Catalog struct {
 	glueSvc   glueAPI
@@ -417,9 +420,9 @@ func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) e
 		return err
 	}
 
-	// Physically delete all table files on storage
-	if purgeErr := internal.PurgeTableFiles(ctx, tbl); purgeErr != nil {
-		return fmt.Errorf("dropped table but failed to purge files: %w", purgeErr)
+	// Physically delete all table files on storage best-effort
+	if purgeErr := tbl.PurgeFiles(ctx); purgeErr != nil {
+		log.Printf("WARNING: dropped table %s but failed to purge files: %v", identifier, purgeErr)
 	}
 
 	return nil
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -405,6 +405,28 @@ func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) er
 	return nil
 }
 
+func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) error {
+	tbl, err := c.LoadTable(ctx, identifier)
+	if err != nil {
+		if errors.Is(err, catalog.ErrNoSuchTable) {
+			return err
+		}
+		// If we can't load the table, fall back to dropping metadata only
+		return c.DropTable(ctx, identifier)
+	}
+
+	// Drop the table entry from the catalog first
+	err = c.DropTable(ctx, identifier)
+	if err != nil {
+		return err
+	}
+
+	// Physically delete all table files on storage
+	_ = internal.PurgeTableFiles(ctx, tbl)
+
+	return nil
+}
+
 // RenameTable renames an Iceberg table in the Glue catalog.
 func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*table.Table, error) {
 	fromDatabase, fromTable, err := identifierToGlueTable(from)
--- a/catalog/glue/glue_test.go
+++ b/catalog/glue/glue_test.go
@@ -1447,3 +1447,28 @@ func TestIsConcurrentModificationException(t *testing.T) {
 	require.False(t, isConcurrentModificationException(errors.New("network timeout")))
 	require.False(t, isConcurrentModificationException(nil))
 }
+
+func TestGluePurgeTable(t *testing.T) {
+	assert := require.New(t)
+
+	mockGlueSvc := &mockGlueClient{}
+
+	mockGlueSvc.On("GetTable", mock.Anything, &glue.GetTableInput{
+		DatabaseName: aws.String("test_database"),
+		Name:         aws.String("test_table"),
+	}, mock.Anything).Return(&glue.GetTableOutput{
+		Table: &testIcebergGlueTable1,
+	}, nil).Twice()
+
+	mockGlueSvc.On("DeleteTable", mock.Anything, &glue.DeleteTableInput{
+		DatabaseName: aws.String("test_database"),
+		Name:         aws.String("test_table"),
+	}, mock.Anything).Return(&glue.DeleteTableOutput{}, nil).Once()
+
+	glueCatalog := &Catalog{
+		glueSvc: mockGlueSvc,
+	}
+
+	err := glueCatalog.PurgeTable(context.TODO(), TableIdentifier("test_database", "test_table"))
+	assert.NoError(err)
+}
--- a/catalog/glue/glue_test.go
+++ b/catalog/glue/glue_test.go
@@ -1447,28 +1447,3 @@ func TestIsConcurrentModificationException(t *testing.T) {
 	require.False(t, isConcurrentModificationException(errors.New("network timeout")))
 	require.False(t, isConcurrentModificationException(nil))
 }
-
-func TestGluePurgeTable(t *testing.T) {
-	assert := require.New(t)
-
-	mockGlueSvc := &mockGlueClient{}
-
-	mockGlueSvc.On("GetTable", mock.Anything, &glue.GetTableInput{
-		DatabaseName: aws.String("test_database"),
-		Name:         aws.String("test_table"),
-	}, mock.Anything).Return(&glue.GetTableOutput{
-		Table: &testIcebergGlueTable1,
-	}, nil).Twice()
-
-	mockGlueSvc.On("DeleteTable", mock.Anything, &glue.DeleteTableInput{
-		DatabaseName: aws.String("test_database"),
-		Name:         aws.String("test_table"),
-	}, mock.Anything).Return(&glue.DeleteTableOutput{}, nil).Once()
-
-	glueCatalog := &Catalog{
-		glueSvc: mockGlueSvc,
-	}
-
-	err := glueCatalog.PurgeTable(context.TODO(), TableIdentifier("test_database", "test_table"))
-	assert.NoError(err)
-}
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -561,6 +561,10 @@ func (c *Catalog) DropTable(_ context.Context, ident table.Identifier) error {
 	return os.RemoveAll(tablePath)
 }
 
+func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) error {
+	return c.DropTable(ctx, identifier)
+}
+
 func (c *Catalog) RenameTable(_ context.Context, _, _ table.Identifier) (*table.Table, error) {
 	return nil, errors.New("hadoop catalog: rename table is not supported")
 }
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -95,6 +95,8 @@ func validateIdentifier(ident table.Identifier) error {
 // Catalog is a filesystem-based Iceberg catalog that requires no external
 // metastore. All state lives on disk as directories and versioned JSON
 // metadata files. Currently only local filesystem paths are supported.
+var _ catalog.PurgeableTable = (*Catalog)(nil)
+
 type Catalog struct {
 	name      string
 	warehouse string
@@ -562,6 +564,17 @@ func (c *Catalog) DropTable(_ context.Context, ident table.Identifier) error {
 }
 
 func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) error {
+	tbl, err := c.LoadTable(ctx, identifier)
+	if err != nil {
+		return err
+	}
+
+	// For Hadoop catalog, physical files walk must run BEFORE deleting the table directory root
+	if purgeErr := tbl.PurgeFiles(ctx); purgeErr != nil {
+		log.Printf("WARNING: failing to purge some files in Hadoop table %s: %v", identifier, purgeErr)
+	}
+
+	// Delete the table directory root from the local storage
 	return c.DropTable(ctx, identifier)
 }
 
--- a/catalog/internal/utils.go
+++ b/catalog/internal/utils.go
@@ -27,8 +27,10 @@ import (
 	stdfs "io/fs"
 	"maps"
 	"net/url"
+	"os"
 	"path"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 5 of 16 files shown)


catalog/internal/utils.go
```diff
@@ -241,3 +242,37 @@ func UpdateAndStageTable(ctx context.Context, current *table.Table, ident table.
 		),
 	}, nil
 }
+
+// PurgeTableFiles physically deletes all files under the table's warehouse location using WalkDir.
+// It is best-effort — if interrupted, orphaned files may remain.
+func PurgeTableFiles(ctx context.Context, tbl *table.Table) (err error) {
```
laskoviymishka: Walks `Metadata().Location()`, but tables can write outside that root via `write.data.path` / `write.metadata.path` (table/properties.go:27-28). PyIceberg/Java walk manifest entries instead, which catches that case.

At minimum document the limitation on `PurgeableTable`. Better, union in manifest-referenced paths that fall outside `Location()`. wdyt?

catalog/internal/utils.go
```diff
@@ -241,3 +242,37 @@ func UpdateAndStageTable(ctx context.Context, current *table.Table, ident table.
...
+			if err != nil {
+				return nil
+			}
+			if !d.IsDir() {
+				files = append(files, path)
+			}
+
+			return nil
+		})
+	}
```
laskoviymishka: Every deletion error is silently dropped: the `WalkDir` err, the `DeleteFiles` err, every individual `Remove`. Function declares `(err error)` but always returns nil, so callers can't tell a no-op purge from success.

I'd collect with `errors.Join` and return. `BulkRemovableIO` also returns the successfully-deleted list, worth surfacing when `len(deleted) < len(files)`.

catalog/glue/glue.go
```diff
@@ -405,6 +405,28 @@ func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) er
 	return nil
 }
 
+func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) error {
+	tbl, err := c.LoadTable(ctx, identifier)
+	if err != nil {
+		if errors.Is(err, catalog.ErrNoSuchTable) {
+			return err
+		}
```
laskoviymishka: This (same in hive.go, sql.go) silently downgrades to metadata-only on any non-`ErrNoSuchTable` `LoadTable` failure: network blip, S3 403, corrupt metadata JSON. User asked for `--purge`, catalog entry vanishes, files stay, no signal.

I'd return the error. If the user wants metadata-only on load failure they can re-run `drop table` without the flag. At minimum a typed `ErrPurgeIncomplete` the CLI surfaces, not a silent fallback.

catalog/sql/sql_test.go
```diff
@@ -684,6 +684,39 @@ func (s *SqliteCatalogTestSuite) TestDropTableNotExist() {
...
+		s.Require().True(ok, "catalog must implement PurgeableTable")
+
+		s.NoError(purger.PurgeTable(context.Background(), tt.tblID))
+
+		// The table catalog entry should be gone
+		_, err = tt.cat.LoadTable(context.Background(), tt.tblID)
+		s.ErrorIs(err, catalog.ErrNoSuchTable)
+
+		// The physical file should be gone
+		s.NoFileExists(metaFile)
```
laskoviymishka: Only checks the single metadata.json is gone. `CreateTable` against an empty table writes no data files, so the walk + bulk-delete path is never exercised. The test would still pass if `PurgeTableFiles` only removed metadata.json and silently failed on everything else.

I'd extend with an `AppendTable` commit so there are real data + manifest files, then assert the directory is empty.

catalog/glue/glue_test.go
```diff
@@ -1447,3 +1447,28 @@ func TestIsConcurrentModificationException(t *testing.T) {
 	require.False(t, isConcurrentModificationException(errors.New("network timeout")))
 	require.False(t, isConcurrentModificationException(nil))
 }
+
```
laskoviymishka: No `FS` plumbed in, so `PurgeTableFiles` never gets exercised. The test only asserts `LoadTable` + `DropTable` get called, which `DropTable` tests already cover. Same shape in hive_test.go:1207.

Either drop these or thread an in-memory FS so the purge path is real. As written they'd still pass if we deleted `PurgeTableFiles` entirely.

catalog/catalog.go
```diff
@@ -167,6 +167,12 @@ type TransactionalCatalog interface {
 	CommitTransaction(ctx context.Context, commits []table.TableCommit) error
 }
 
+// PurgeableTable is an optional interface that catalogs can implement
```
laskoviymishka, line 170: Godoc could carry the contract callers actually need: purge is best-effort, file-deletion errors may or may not surface, and REST delegates server-side while SQL/Glue/Hive/Hadoop are client-side and may miss files outside the location root. Today a downstream implementor has no signal about what 'physical table deletion' guarantees.

laskoviymishka (changes requested):
Shape looks right overall: `PurgeableTable` as an optional interface, CLI opt-in via type assertion, REST using `purgeRequested=true`, Hadoop delegating to `RemoveAll`. Splitting metadata drop from storage cleanup is the right factoring, and the shared `WalkDir` / `BulkRemovableIO` helper seems like a clean storage-agnostic base.

I’d hold merge for one safety issue though: today `--purge` can look successful while leaving files behind.

Three pieces combine badly:

* `PurgeTableFiles` swallows file-deletion errors and returns `nil`
* catalog wrappers fall back to metadata-only drop on any non-`ErrNoSuchTable` `LoadTable` failure
* the location walk only covers `Metadata().Location()`, so files under `write.data.path` / `write.metadata.path` can be missed

Net effect: user runs `--purge`, sees success, and assumes storage is clean. That’s worse than the current metadata-only behavior, because at least today the limitation is explicit.

The tests don’t catch this yet: the SQL test only checks `metadata.json` is gone, `CreateTable` writes no data files, and the Glue/Hive mocks don’t exercise a filesystem path.

Before merge, I’d want:

* propagate errors from `PurgeTableFiles`
* remove the `LoadTable` fallback-to-`DropTable`, or return something like `ErrPurgeIncomplete`
* extend the SQL test to write real data files and assert the tree is gone
* either make the Glue/Hive tests exercise an in-memory FS, or drop them as purge-path coverage
* document what `PurgeableTable` actually guarantees, including the `write.data.path` / `write.metadata.path` limitation for location-walk implementations

One smaller note: drop-then-purge means a purge failure leaves no easy retry path because the table no longer loads. PyIceberg does the same, so I’m not pushing to invert it, but it’s worth calling out in godoc.

Once those are covered, happy to take another pass.

mithilgirish:
All five concerns are valid and addressed in the updated patch.

* **Error Propagation:** Fixed. Errors from `WalkDir`, `DeleteFiles`, and individual `Remove` calls are now collected via `errors.Join` and returned. `os.IsNotExist` is safely ignored. The declared `(err error)` return now carries actual signal.
* **No Silent Fallbacks:** Fixed. The `LoadTable` fallback in SQL, Glue, and Hive that was quietly downgrading `--purge` on any non-`ErrNoSuchTable` error has been removed. The catalog now returns a wrapped error immediately — the table entry is not dropped until a purge can be attempted.
* **Tracking External Paths:** Fixed. `PurgeTableFiles` now unions the base `Location()` walk with a full crawl of active metadata, manifests, and snapshots. This mirrors the PyIceberg/Java approach and catches files under custom `write.data.path` / `write.metadata.path`.

Holding the push on this last point until we make a call on **Option A vs B**:
* **Option A:** Promote `getReferencedFiles` in `table/orphan_cleanup.go` to a public method on `Table`. No duplication, logic is already correct, but the diff touches `table/`.
* **Option B:** Private helper in `catalog/internal/utils.go` via public metadata APIs. PR stays isolated to `catalog/`, but introduces duplication that needs to stay in sync.

* **Test Coverage:** `TestPurgeTable` now writes real Parquet + a mock `stats.puffin` outside the table root, asserting both are fully wiped. Glue and Hive mock purge tests removed — no filesystem backend meant they weren't covering `PurgeTableFiles` at all.
* **PurgeableTable GoDoc:** Documents the best-effort model, type-assertion pattern, the un-loadable-on-failure warning, and the client-side vs REST (`purgeRequested=true`) distinction.

Let me know your preference on Option A or B, and I will push the updated patch.

laskoviymishka:
I think it's fine to touch table package, so lean to opt A

mithilgirish:
Just pushed the update using Option A. getReferencedFiles is now exported as GetReferencedFiles on the Table struct, keeping the same logic with zero duplication. All review points are fully addressed.

catalog/internal/utils.go
```diff
@@ -241,3 +245,79 @@ func UpdateAndStageTable(ctx context.Context, current *table.Table, ident table.
...
+
+		return "file:///" + cleaned
+	}
+
+	return uri
+}
+
+// PurgeTableFiles physically deletes all files under the table's warehouse location
+// and any referenced files written outside the location root (e.g., via write.data.path
+// or write.metadata.path properties).
```
laskoviymishka: We're deleting data files unconditionally here, but Iceberg's spec gates this on the `gc.enabled` table property — Java's `CatalogUtil.dropTableData` and PyIceberg both refuse to purge data when `gc.enabled=false`, because the same physical files may be referenced by snapshotted clones or branched tables. On a `gc.enabled=false` table this PR will silently corrupt the other reader. I'd read `tbl.Metadata().Properties().Get("gc.enabled", "true")` at the top of `PurgeTableFiles` and short-circuit (or at least skip the data-file portion) when it's false. wdyt?

catalog/internal/utils.go
```diff
@@ -241,3 +245,79 @@ func UpdateAndStageTable(ctx context.Context, current *table.Table, ident table.
 		),
 	}, nil
 }
```
laskoviymishka, line 243: This only strips `file://` — so the LocalFS walk callback (which yields bare OS paths like `/tmp/foo`) and the metadata-side URIs (`file:///tmp/foo`) land in the set as two distinct keys. Today it's masked because the second `Remove` hits `ErrNoSuchFile` and we swallow it, but the union-dedup intent is broken, and for any cloud FS where `WalkDir` emits a different scheme form than the metadata we'll get spurious `BulkRemovableIO` errors. There's already a `normalizeFilePath` helper in `table/orphan_cleanup.go` that does full URL normalization across schemes — I'd reuse that instead of rolling our own here.

catalog/internal/utils.go
```diff
@@ -241,3 +245,79 @@ func UpdateAndStageTable(ctx context.Context, current *table.Table, ident table.
...
+		cleaned = strings.TrimPrefix(cleaned, "/")
+
+		return "file:///" + cleaned
+	}
+
+	return uri
+}
+
+// PurgeTableFiles physically deletes all files under the table's warehouse location
+// and any referenced files written outside the location root (e.g., via write.data.path
```
laskoviymishka: The godoc calls this "best-effort" but `errors.Join` propagates every per-file failure and the catalogs wrap it as "dropped table but failed to purge files". That's not best-effort — that's fail-loud, with no retry path because the catalog row is already gone. Either match Java/PyIceberg (log per-file failures, return nil so the catalog drop stays the source of truth) or keep fail-loud and return a structured `PurgeResult{CatalogDropped bool, FileErrors []error}` so the caller can tell that the catalog state is durable. Right now the doc and the code disagree, which is the worst of both.

catalog/hadoop/hadoop.go
```diff
@@ -561,6 +561,10 @@ func (c *Catalog) DropTable(_ context.Context, ident table.Identifier) error {
 	return os.RemoveAll(tablePath)
 }
 
+func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) error {
```
laskoviymishka, line 566: This delegates to `os.RemoveAll(tablePath)`, which means any data written to `write.data.path` or `write.metadata.path` outside the table root just survives the purge. The interface godoc explicitly claims "files written outside the root… are deleted", so Hadoop is silently violating its own contract. I'd at minimum load the table and run the same union-walk + bulk-delete path the other catalogs use, or call out the caveat in the Hadoop-specific godoc.

table/orphan_cleanup.go
```diff
@@ -225,10 +225,10 @@ func (t Table) executeOrphanCleanup(ctx context.Context, cfg *orphanCleanupConfi
 	return result, nil
 }
 
-// getReferencedFiles collects all files referenced by table metadata: previous metadata
+// GetReferencedFiles collects all files referenced by table metadata: previous metadata
 // files, statistics and partition-statistics paths (Puffin, etc.), and all paths reachable
 // from current snapshots (manifest lists, manifests, data files).
-func (t Table) getReferencedFiles(fs iceio.IO) (map[string]bool, error) {
+func (t Table) GetReferencedFiles(fs iceio.IO) (map[string]bool, error) {
```
laskoviymishka: This is a public method on `Table` whose only caller is `catalog/internal`, takes an `iceio.IO` it'll panic-deref if the table has snapshots and `fs` is nil, and the returned `map[string]bool` has no documented contract about path normalization. I'd consider either folding this into a `Table.PurgeFiles(ctx)` that owns its own FS, or pushing it onto `Metadata` and documenting the nil-fs case. Exporting it as-is locks in an awkward surface for one internal caller.

Also worth a note: this walk doesn't include deletion-vector blob paths. With #1100 just merged, V3 tables with DVs will leak `.dv` files on stores where `ListableIO` isn't available.

catalog/internal/utils.go
```diff
@@ -170,6 +173,7 @@ func getDefaultWarehouseLocation(dbname, tablename string, nsprops, catprops ice
 // ([\w-]{36})      -> UUID (36 characters, including hyphens)
 // (?:\.\w+)?       -> optional codec name
 // \.metadata\.json -> file extension
+// var tableMetadataFileNameRegex = regexp.MustCompile(`^(\d+)-([\w-]{36})(?:\.\w+)?\.metadata\.json`)
```
laskoviymishka: Stray commented-out `tableMetadataFileNameRegex` line that the diff left behind. All three of us stopped on it. Worth a quick clean.

catalog/catalog.go
```diff
@@ -167,6 +167,26 @@ type TransactionalCatalog interface {
...
+//
+// For REST catalogs the purge is delegated server-side. For client-side
+// catalogs (SQL, Glue, Hive, Hadoop) the table is first dropped from
+// the catalog, then all files under the table's [table.Metadata.Location]
+// root, plus any referenced files written outside the root (e.g. via
+// write.data.path or write.metadata.path table properties), are deleted.
+// File-deletion errors are propagated to the caller, but because the
+// catalog entry is already removed at that point there is no automatic
+// retry path.
+type PurgeableTable interface {
```
laskoviymishka, line 189: The catalogs that intentionally satisfy this interface would benefit from a `var _ catalog.PurgeableTable = (*Catalog)(nil)` assertion (REST especially, since it's not exercised in the same test fixture). Cheap compile-time guard against someone renaming the method out from under the interface.

cmd/iceberg/drop_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (m *mockCatalogForDrop) UpdateNamespaceProperties(ctx context.Context, namespace table.Identifier, removals []string, updates iceberg.Properties) (catalog.PropertiesUpdateSummary, error) {
+	return catalog.PropertiesUpdateSummary{}, nil
+}
+
+type mockPurgeableCatalog struct {
+	mockCatalogForDrop
+	purgeCalled bool
+	purgeIdent  table.Identifier
+	purgeErr    error
```
laskoviymishka, line 105: `purgeErr` is declared on the mock but never used in a test case — I'd add a `TestRunDropTablePurgeError` that returns an error here and asserts the wrapping ("dropped table but failed to purge files") so the contract can't silently regress.

catalog/sql/sql_test.go
```diff
@@ -684,6 +686,106 @@ func (s *SqliteCatalogTestSuite) TestDropTableNotExist() {
...
+		s.FileExists(externalStatsPath)
+
+		// Add this external statistics file to the table metadata JSON
+		metaBytes, err := os.ReadFile(metaLoc)
+		s.Require().NoError(err)
+		var metaMap map[string]any
+		s.Require().NoError(json.Unmarshal(metaBytes, &metaMap))
+		metaMap["statistics"] = []any{
+			map[string]any{
+				"snapshot-id":               int64(1),
```
laskoviymishka: Hard-coding `"snapshot-id": int64(1)` here is fragile — if it ever drifts from `tbl.CurrentSnapshot().SnapshotID`, the test will still pass, but for the wrong reason (the external file gets removed by the walk rather than by the referenced-files path, which is what this test is supposed to exercise). I'd read the snapshot id off the loaded table instead.

cmd/iceberg/main.go
```diff
@@ -537,10 +538,21 @@ func runDrop(ctx context.Context, output Output, cat catalog.Catalog, cmd *DropC
...
 	case cmd.Table != nil:
-		err := cat.DropTable(ctx, catalog.ToIdentifier(cmd.Table.Identifier))
+		ident := catalog.ToIdentifier(cmd.Table.Identifier)
+		var err error
+		if cmd.Table.Purge {
+			if purger, ok := cat.(catalog.PurgeableTable); ok {
+				err = purger.PurgeTable(ctx, ident)
+			} else {
+				output.Error(fmt.Errorf("catalog %s does not support purge", cat.CatalogType()))
+				osExit(1)
```
laskoviymishka, line 548: The purge branch uses `osExit(1)` (the swappable test hook), but the namespace branch right next to it still calls bare `os.Exit(1)`. Functionally fine today, but it's the kind of thing that makes future test additions land on a confusing failure mode. I'd make both use `osExit` for consistency.

laskoviymishka (changes requested):
Thanks for the quick turnaround, all landed in shape. 

Almost there. Two things before merge:

The big one is gc.enabled. Iceberg's spec gates data-file deletion on this table property - Java's CatalogUtil.dropTableData and PyIceberg both refuse to purge data files when it's false, because the same physical files may be shared with snapshotted clones or branched tables. As-is this PR will delete those files anyway and silently corrupt the other reader. The fix is small (read the property at the top of PurgeTableFiles), but it has to be there before this lands.

- The other is the union-dedup. normalizeURI only strips file://, so the LocalFS walk (bare OS paths) and the metadata side (file:/// URIs) end up as distinct keys in the set. Today the double-delete is masked by ErrNoSuchFile being swallowed, but the dedup intent is broken and cloud FSs will surface spurious errors as soon as their WalkDir scheme form differs from the metadata. There's already a normalizeFilePath helper in table/orphan_cleanup.go, i'd reuse that.

Plus a few smaller things in inline: the godoc/contract mismatch on "best-effort" vs. propagated errors (pick one), Hadoop's PurgeTable silently ignoring external write.data.path / write.metadata.path, the stray commented-out regex, and the unused purgeErr field on the test mock.

Fix the gc.enabled gate and the dedup, address the contract-vs-godoc choice, and this is good to land. Happy to take another pass once those are in.

mithilgirish:
Thanks for the detailed review! I've pushed a new commit addressing all points:

* `gc.enabled` gate added — skips data-file deletion when false
* URI dedup fixed using existing `normalizeFilePath` helper
* Godoc aligned with Java/PyIceberg best-effort model
* Hadoop now runs full union-walk before `os.RemoveAll`
* Compile-time `var _ catalog.PurgeableTable` guards added to all catalogs
* SQL test snapshot ID reads from `tbl.CurrentSnapshot().SnapshotID`
* `TestRunDropTablePurgeError` added
* CLI unified to `osExit(1)`
* Stray regex removed
Also added `ctx.Err()` checks in traversal loops and safe `os.ErrNotExist` handling.

table/orphan_cleanup.go
```diff
@@ -547,7 +562,23 @@ func deleteFilesParallel(fs iceio.IO, orphanFiles []string, cfg *orphanCleanupCo
 //  3. Path separators and casing may vary across different systems and configurations
 //  4. Without normalization, semantically identical paths would be treated as different,
 //     leading to false positives in orphan detection
-func normalizeFilePath(path string, cfg *orphanCleanupConfig) string {
+//
+// normalizeFilePath normalizes a file path for comparison, handling schemes, authorities, and separators.
+// It also aligns file:// URIs and bare local file system paths so they normalize to the same format.
+func normalizeFilePath(path string) string {
```
laskoviymishka, line 571: This silently eats the authority on RFC 8089 URIs like `file://localhost/foo/bar.parquet`, so `localhost` gets folded into the path. The downstream effect is that `getReferencedFiles` stores keys via this stripper while `identifyOrphanFiles` looks them up via `normalizeFilePathWithConfig` (which does not strip `file:`), so on a LocalFS where `WalkDir` returns `file:///foo/x.parquet` and the referenced set holds `/foo/x.parquet`, the file is classified as an orphan and deleted.

I'd parse with `url.Parse` for the `file:` scheme and take `parsed.Path` directly, then point both call sites at the same function. `file:///foo` and `file://localhost/foo` both land at `/foo` and we don't have to reason about trim ordering.

laskoviymishka (changes requested):
Looks good overall. The catalog wiring and `gc.enabled=false` handling are much closer now.

One potential bug before merge: path normalization. The orphan-file path seems to use two different normalizers on each side of the comparison, so local `file://...` paths can be mismatched and valid files may be treated as orphans. I left the details inline, but I’d unify both sides on one `url.Parse`-based normalizer.

A few smaller things, mostly follow-ups / cleanup:

* `gc.enabled=false` should probably only skip data files; Java still deletes manifests / metadata.
* purge should include deleted data files too; `discardDeleted=true` is right for orphan detection, but not for table purge.
* the purge-error test seems to cover a contract the current catalogs don’t actually expose.
* prefer `slog.WarnContext` over `log.Printf` so callers can route logs.

Everything else I asked for looks addressed. Fix the normalization issue, and I’m happy to take another pass.

mithilgirish:
Pushed the final fixes — unified `file:` normalizers on `url.Parse` with localhost/Windows handling, scoped `gc.enabled=false` to data payloads only, switched to `discardDeleted=false`, added fail-fast on metadata errors, and cleaned up logging/test.

laskoviymishka (approved):
nice work!

mithilgirish:
is there a target date for the next release with this feature? 
we're about to use it in our production application.

laskoviymishka:
@zeroshade -- can we do a patch version release? I'm not sure it worth to do a minor version bump for this, we plan to do 0.7.0 once variant fully in (incl shreding).

PR #1105, opened 2026-05-20
fix(test): TestCreateTable existence check failure

laskoviymishka:
PR #355 renamed the table created in TestCreateTable from "test-table" to "test-create-table" but left the follow-up CheckTableExists and DropTable calls pointing at the old name. CheckTableExists returns false, s.True(exists) fails, the actual table is never dropped, and the orphan in the namespace cascades into failures for every other suite test via ensureNamespace.


nssalian (approved):

nssalian:
Thanks @laskoviymishka. lgtm pending integration tests passing here

dhananjaykrutika (approved):

zeroshade (approved):

PR #1106, opened 2026-05-20
fix(table): tighten strict negative metrics checks

### LINKED ISSUE #1103: Strict metrics NotEqual and NotIn use canContainNulls instead of containsNullsOnly
In #1098, reviewer feedback pointed out that `VisitNotEqual` and `VisitNotIn` still use `canContainNulls(fieldID) || canContainNans(fieldID)`.

These branches have different semantics from the range/equality/IN visitors updated in #1098. Per Tanmay's review, the Java strict metrics evaluator uses `containsNullsOnly` / `containsNaNsOnly` for the negative predicate branches, then falls through to bounds checks. The Go implementation currently uses `canContainNulls` / `canContainNans`, which treats "any nulls/NaNs may exist" as enough to prove all rows match `!=` / `NOT IN`.

That seems too optimistic: a file with some nulls and some values equal to the literal should not be considered guaranteed to match the negative predicate. Only an all-null/all-NaN column can be proven from that metric alone.

Proposed fix:

- update `VisitNotEqual` and `VisitNotIn` to use `containsNullsOnly` / `containsNaNsOnly`
- add regression coverage for mixed null/value files where `canContainNulls` would be too broad
- keep this separate from #1098 because the negative predicate semantics differ from the missing-metric conservative handling there

fallintoplace:
## Summary

- use `containsNullsOnly` / `containsNansOnly` for strict `NotEqual` and `NotIn` metrics evaluation
- remove now-unused strict `canContain*` helpers
- add regression coverage for mixed null/value and mixed NaN/value files

## Why

For strict negative predicates, the evaluator can only prove a whole file matches from null/NaN metrics when the column contains only nulls or only NaNs. Mixed null/value or NaN/value files still need to fall through to bounds checks and remain conservative when bounds do not prove the predicate.

Fixes #1103.

## Testing

- `go test ./table -run TestEvaluators`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/table/evaluators.go
+++ b/table/evaluators.go
@@ -1436,7 +1436,7 @@ func (m *strictMetricsEval) VisitNotEqual(t iceberg.BoundTerm, lit iceberg.Liter
 	field := t.Ref().Field()
 	fieldID := field.ID
 
-	if m.canContainNulls(fieldID) || m.canContainNans(fieldID) {
+	if m.containsNullsOnly(fieldID) || m.containsNansOnly(fieldID) {
 		return rowsMustMatch
 	}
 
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
Thanks for the PR, LGTM!

laskoviymishka (approved):

PR #1108, opened 2026-05-21
feat(table): add DV planning validations to match Java's DeleteFileIndex (#1050)

### LINKED ISSUE #1050: feat(table): match Java's DV planning validations in PlanFiles

tanmayrauth:
Add sequence-number guard and reject multiple DVs per data file during scan planning, matching Java's DeleteFileIndex behavior.

Closes: #1050

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -442,6 +442,7 @@ func buildDVIndex(dvEntries []iceberg.ManifestEntry) (map[string]iceberg.Manifes
 			dvIndex[*ref] = del
 		}
 	}
+
 	return dvIndex, nil
 }
 
@@ -456,6 +457,7 @@ func matchDVToData(dataEntry iceberg.ManifestEntry, dvIndex map[string]iceberg.M
 	if dataEntry.SequenceNum() <= dvEntry.SequenceNum() {
 		return []iceberg.DataFile{dvEntry.DataFile()}
 	}
+
 	return nil
 }
 
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -429,6 +429,36 @@ func partitionsMatch(a, b map[int]any) bool {
 	return true
 }
 
+// buildDVIndex indexes deletion vectors by the data file path they reference.
+// The spec requires at most one DV per data file; a second entry for the same
+// path is rejected with an error.
+func buildDVIndex(dvEntries []iceberg.ManifestEntry) (map[string]iceberg.ManifestEntry, error) {
+	dvIndex := make(map[string]iceberg.ManifestEntry, len(dvEntries))
+	for _, del := range dvEntries {
+		if ref := del.DataFile().ReferencedDataFile(); ref != nil {
+			if _, exists := dvIndex[*ref]; exists {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
I believe that if you rebase it should solve the CI issue.

tanmayrauth:
Let me try rebasing.

zeroshade:
sorry about that, rebase now and that should fix the failing tests

tanmayrauth:
Np! just got it rebased.

table/scanner.go
```diff
@@ -429,6 +429,38 @@ func partitionsMatch(a, b map[int]any) bool {
...
+
+// matchDVToData returns the deletion vector that applies to the given data
+// entry, if any. A DV applies only when the data file's sequence number is
+// less than or equal to the DV's sequence number.
+func matchDVToData(dataEntry iceberg.ManifestEntry, dvIndex map[string]iceberg.ManifestEntry) []iceberg.DataFile {
+	dvEntry, ok := dvIndex[dataEntry.DataFile().FilePath()]
+	if !ok {
+		return nil
+	}
+	if dataEntry.SequenceNum() <= dvEntry.SequenceNum() {
```
laskoviymishka, line 457: One subtle thing for a follow-up, not blocking the merge — when `dataEntry.SequenceNum() > dvEntry.SequenceNum()` we silently return nil where Java throws a `ValidationException` here. And since `SequenceNum()` returns `-1` for a nil pointer (see manifest.go around line 2069), a DV entry whose seq-num pointer is nil compares wrong against any real data seq-num and gets silently skipped. In practice inheritance should populate seq-nums before this path, so the live impact is bounded, but it'd be cleaner to either propagate the spec violation as an error or guard the `-1` sentinel explicitly so the helper's contract isn't load-bearing on upstream behavior. Happy to file a follow-up issue with the details.

laskoviymishka (approved):
Nice work!

I left one small follow up, but this non blocking, mergeable

PR #1110, opened 2026-05-21
fix: reject invalid transform widths

### LINKED ISSUE #1109: ParseTransform accepts invalid bucket and truncate widths
`ParseTransform` currently accepts invalid `bucket[...]` and `truncate[...]` widths.

For `bucket` and `truncate`, it parses the bracket value with `strconv.Atoi` but ignores the error and does not reject zero:

```go
n, _ := strconv.Atoi(matches[1])
```

Because `regexFromBrackets` accepts digits, inputs like `bucket[0]`, `truncate[0]`, and very large values can produce transforms with invalid widths.

That can later panic when the transform is applied. `BucketTransform` takes modulo by `int32(t.NumBuckets)`, and integer/decimal truncate paths take modulo by the width. A zero width therefore becomes a divide-by-zero panic instead of a parse error.

One extra edge case: on 64-bit platforms, a value like `bucket[4294967296]` may fit in `int` but becomes `int32(0)` before the bucket modulo. So validation should reject values larger than `math.MaxInt32`, not only `Atoi` overflow and `<= 0`.

Proposed fix:

- reject `Atoi` errors
- reject widths `<= 0`
- reject widths greater than `math.MaxInt32`
- add parse tests for zero, huge overflow, and int32-overflow widths

fallintoplace:
## Summary

- reject invalid `bucket[...]` and `truncate[...]` widths during `ParseTransform`
- treat `Atoi` errors, zero widths, and values above `math.MaxInt32` as invalid transforms
- add parse regression coverage for zero, huge overflow, and int32-overflow widths

## Why

`ParseTransform` previously ignored the `strconv.Atoi` error and accepted zero widths. Those invalid transforms could later panic when bucket/truncate logic performed modulo by the width. Values larger than `math.MaxInt32` are also unsafe because bucket/truncate implementations narrow widths through int32-sized paths.

Fixes #1109.

## Testing

- `go test . -run TestParseTransform`
- `go test .`
- `git diff --check`

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -47,7 +47,10 @@ func ParseTransform(s string) (Transform, error) {
 			break
 		}
 
-		n, _ := strconv.Atoi(matches[1])
+		n, err := strconv.Atoi(matches[1])
+		if err != nil || n <= 0 || n > math.MaxInt32 {
+			break
+		}
 
 		return BucketTransform{NumBuckets: n}, nil
 	case strings.HasPrefix(s, "truncate"):
@@ -56,7 +59,10 @@ func ParseTransform(s string) (Transform, error) {
 			break
 		}
 
-		n, _ := strconv.Atoi(matches[1])
+		n, err := strconv.Atoi(matches[1])
+		if err != nil || n <= 0 || n > math.MaxInt32 {
+			break
+		}
 
 		return TruncateTransform{Width: n}, nil
 	default:
--- a/transforms_test.go
+++ b/transforms_test.go
@@ -82,6 +82,12 @@ func TestParseTransform(t *testing.T) {
 		{"truncate no val", "truncate[]"},
 		{"bucket neg", "bucket[-1]"},
 		{"truncate neg", "truncate[-1]"},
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
Simple and straightforward. Nice!

laskoviymishka (approved):
Approving. The fix is in the right place, the real bug was `ParseTransform` swallowing the `strconv.Atoi` error, and the new guard covers both that and the zero-width case.

One separate thing to flag, not blocking this PR: the spec does not really put an explicit upper bound on `N` for `bucket[N]` / `truncate[N]`. Java ends up capped at `Integer.MAX_VALUE` because it uses `int`; iceberg-rust uses `u32`, so it accepts a wider range.

This PR makes iceberg-go match Java, which I think is the right call. The only weird edge is that a rust-authored table with `bucket[N]` where `N > MaxInt32` could parse in Rust and now fail in Go. That feels like something to discuss cross-client, probably on the Rust side rather than here.

I’ll file a follow-up issue for that.

PR #1112, opened 2026-05-21
fix(table): normalize timestamp units for partitioned writes

### LINKED ISSUE #1111: Partitioned writes compute timestamp partitions using raw Arrow units
Partitioned writes currently derive partition keys from the input Arrow record before `ToRequestedSchema` normalizes timestamp units to the table schema.

The write path eventually casts Arrow `timestamp[s]` and `timestamp[ms]` values to Iceberg microsecond timestamps before writing data files, but partition routing happens earlier:

```go
partitions, err := p.getPartitions(record)
```

`getRecordPartitions` then calls `getArrowValueAsIcebergLiteral`, which currently casts the raw Arrow timestamp value directly to an Iceberg microsecond timestamp:

```go
case *array.Timestamp:
    return iceberg.NewLiteral(iceberg.Timestamp(arr.Value(row))), nil
```

That ignores `arr.DataType().(*arrow.TimestampType).Unit`. For example, an Arrow `timestamp[ms]` value of `1700000000000` represents milliseconds since epoch, but this path treats it as microseconds while computing day/hour partitions. The data values can later be written correctly after schema projection, while the partition metadata/path was computed with the wrong unit.

This can route rows to the wrong year/month/day/hour partition and break partition pruning for partitioned writes.

Proposed direction:

- keep accepting Arrow `timestamp[s]` and `timestamp[ms]`; existing projection code supports casting them to Iceberg microseconds
- make partition-key extraction convert Arrow timestamp values according to the Arrow unit and the Iceberg source field type
- return microsecond literals for `timestamp` / `timestamptz` sources
- return nanosecond literals for `timestamp_ns` / `timestamptz_ns` sources
- add regression coverage for partitioned fanout writes using non-microsecond timestamp input

fallintoplace:
## Summary

- convert Arrow timestamp partition values according to their Arrow unit before applying Iceberg partition transforms
- use the table source type to choose microsecond vs nanosecond Iceberg timestamp literals
- add partition fanout regression coverage for timestamp seconds, milliseconds, microseconds, and nanoseconds

## Why

Partitioned writes compute partition keys before `ToRequestedSchema` normalizes Arrow timestamp arrays to the table schema. The old partition path cast raw Arrow timestamp values directly to Iceberg microsecond timestamps, so `timestamp[s]` and `timestamp[ms]` inputs could be routed to the wrong day/hour partition even though the data values were later written with normalized units.

Fixes #1111.

## Testing

- `go test ./table -run TestFanoutWriter -count=1`
- `git diff --check`

### DIFF
--- a/table/partitioned_fanout_writer.go
+++ b/table/partitioned_fanout_writer.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"iter"
+	"math"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
@@ -29,6 +30,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/decimal128"
 	"github.com/apache/arrow-go/v18/arrow/extensions"
 	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/internal"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -49,6 +51,12 @@ type partitionInfo struct {
 	rows            []int64
 	partitionValues map[int]any
 	partitionRec    partitionRecord // The actual partition values for generating the path
+}
+
+type partitionFieldInfo struct {
+	sourceField iceberg.PartitionField
+	fieldID     int
+	sourceType  iceberg.Type
 }
 
 // NewPartitionedFanoutWriter creates a new PartitionedFanoutWriter with the specified
@@ -214,28 +222,37 @@ func getRecordPartitions(spec iceberg.PartitionSpec, schema *iceberg.Schema, rec
 	partitionRec := make(partitionRecord, len(partitionFields))
 
 	partitionColumns := make([]arrow.Array, len(partitionFields))
-	partitionFieldsInfo := make([]struct {
-		sourceField *iceberg.PartitionField
-		fieldID     int
-	}, len(partitionFields))
+	partitionFieldsInfo := make([]partitionFieldInfo, len(partitionFields))
 
 	for i := range partitionFields {
 		sourceField := spec.Field(i)
-		colName, _ := schema.FindColumnName(sourceField.SourceID())
-		colIdx := record.Schema().FieldIndices(colName)[0]
-		partitionColumns[i] = record.Column(colIdx)
-		partitionFieldsInfo[i] = struct {
-			sourceField *iceberg.PartitionField
-			fieldID     int
-		}{&sourceField, sourceField.FieldID}
+		colName, ok := schema.FindColumnName(sourceField.SourceID())
+		if !ok {
+			return nil, fmt.Errorf("failed to find source field ID %d in schema", sourceField.SourceID())
+		}
+		colIndices := record.Schema().FieldIndices(colName)
+		if len(colIndices) == 0 {
+			return nil, fmt.Errorf("failed to find source column %q in record schema", colName)
+		}
+		sourceType, ok := schema.FindTypeByID(sourceField.SourceID())
+		if !ok {
+			return nil, fmt.Errorf("failed to find type for source field ID %d in schema", sourceField.SourceID())
+		}
+		partitionColumns[i] = record.Column(colIndices[0])
+		partitionFieldsInfo[i] = partitionFieldInfo{
+			sourceField: sourceField,
+			fieldID:     sourceField.FieldID,
+			sourceType:  sourceType,
+		}
 	}
 
 	for row := range record.NumRows() {
 		for i := range partitionFields {
 			col := partitionColumns[i]
 			if !col.IsNull(int(row)) {
-				sourceField := partitionFieldsInfo[i].sourceField
-				val, err := getArrowValueAsIcebergLiteral(col, int(row))
+				fieldInfo := partitionFieldsInfo[i]
+				sourceField := fieldInfo.sourceField
+				val, err := getArrowValueAsIcebergLiteral(col, int(row), fieldInfo.sourceType)
 				if err != nil {
 					return nil, fmt.Errorf("failed to get arrow values as iceberg literal: %w", err)
 				}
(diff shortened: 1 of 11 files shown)


table/partitioned_fanout_writer.go
```diff
@@ -464,3 +472,66 @@ func getArrowValueAsIcebergLiteral(column arrow.Array, row int) (iceberg.Literal
...
+func arrowTimestampToMicros(value int64, unit arrow.TimeUnit) (int64, error) {
+	switch unit {
+	case arrow.Second:
+		return scaleTimestamp(value, 1_000_000)
+	case arrow.Millisecond:
+		return scaleTimestamp(value, 1_000)
+	case arrow.Microsecond:
+		return value, nil
+	case arrow.Nanosecond:
+		return value / 1_000, nil
```
tanmayrauth: Go's / truncates toward zero, but unit downconversion should floor. For pre-epoch values this rounds the wrong way: ns=-1500 gives -1 here, but the correct μs bin is -2 ([-2000, -1000)). Same class of bug this PR is fixing, wrong partition routing for negative timestamps.

Try
```
case arrow.Nanosecond:
    return math.FloorDiv(value, 1_000), nil
```

Can you please add a regression test with a negative ns value (e.g. one second before epoch with a sub-μs offset) asserting the partition path.

tanmayrauth:
Out of scope: Time64 (line ~381) has the same unit-vs-iceberg.Time bug — follow-up PR.

table/partitioned_fanout_writer.go
```diff
@@ -464,3 +472,75 @@ func getArrowValueAsIcebergLiteral(column arrow.Array, row int) (iceberg.Literal
...
+}
+
+func floorDivInt64(a, b int64) int64 {
+	d := a / b
+	if (a^b) < 0 && d*b != a {
+		d--
+	}
+
+	return d
+}
```
zeroshade: this already exists in the root transforms.go file, we should probably just move the version in transforms.go:579 into an internal/utils.go file and then use that in both places rather than duplicate this function.

table/partitioned_fanout_writer.go
```diff
@@ -464,3 +472,75 @@ func getArrowValueAsIcebergLiteral(column arrow.Array, row int) (iceberg.Literal
...
+	default:
+		return 0, fmt.Errorf("unsupported arrow timestamp unit: %s", unit)
+	}
+}
+
+func scaleTimestamp(value, factor int64) (int64, error) {
+	if (value > 0 && value > math.MaxInt64/factor) ||
+		(value < 0 && value < math.MinInt64/factor) {
+		return 0, fmt.Errorf("arrow timestamp value %d overflows int64 when scaled by %d", value, factor)
+	}
```
zeroshade, line 539: can you add a test that covers this? I don't think it's covered by the current tests

table/partitioned_fanout_writer.go
```diff
@@ -464,3 +472,75 @@ func getArrowValueAsIcebergLiteral(column arrow.Array, row int) (iceberg.Literal
...
+
+	switch sourceType.(type) {
+	case iceberg.TimestampType, iceberg.TimestampTzType:
+		micros, err := arrowTimestampToMicros(value, timestampType.Unit)
+		if err != nil {
+			return nil, err
+		}
+
+		return iceberg.NewLiteral(iceberg.Timestamp(micros)), nil
+	case iceberg.TimestampNsType, iceberg.TimestampTzNsType:
```
zeroshade, line 493: the Tz variants don't seem to get tested, can you add cases that have `TimeZone: "UTC"` so we hit this case?

table/partitioned_fanout_writer.go
```diff
@@ -214,28 +221,33 @@ func getRecordPartitions(spec iceberg.PartitionSpec, schema *iceberg.Schema, rec
...
+		colIndices := record.Schema().FieldIndices(colName)
+		if len(colIndices) == 0 {
+			return nil, fmt.Errorf("failed to find source column %q in record schema", colName)
+		}
+		sourceType, ok := schema.FindTypeByID(sourceField.SourceID())
+		if !ok {
+			return nil, fmt.Errorf("failed to find source field ID %d in schema", sourceField.SourceID())
+		}
+		partitionColumns[i] = record.Column(colIndices[0])
+		partitionFieldsInfo[i] = partitionFieldInfo{&sourceField, sourceField.FieldID, sourceType}
```
zeroshade:
```suggestion
		partitionFieldsInfo[i] = partitionFieldInfo{
		    sourceField: &sourceField, 
		    fieldID:          sourceField.FieldID, 
		    sourceType: sourceType,
		}
```

just so we don't accidentally misorder things

table/partitioned_fanout_writer.go
```diff
@@ -51,6 +52,12 @@ type partitionInfo struct {
 	partitionRec    partitionRecord // The actual partition values for generating the path
 }
 
+type partitionFieldInfo struct {
+	sourceField *iceberg.PartitionField
```
zeroshade: PartitionField is a small struct, why use a pointer here instead of just using it by value?

table/partitioned_fanout_writer.go
```diff
@@ -214,28 +221,33 @@ func getRecordPartitions(spec iceberg.PartitionSpec, schema *iceberg.Schema, rec
...
+		if !ok {
+			return nil, fmt.Errorf("failed to find source field ID %d in schema", sourceField.SourceID())
+		}
+		colIndices := record.Schema().FieldIndices(colName)
+		if len(colIndices) == 0 {
+			return nil, fmt.Errorf("failed to find source column %q in record schema", colName)
+		}
+		sourceType, ok := schema.FindTypeByID(sourceField.SourceID())
+		if !ok {
+			return nil, fmt.Errorf("failed to find source field ID %d in schema", sourceField.SourceID())
```
zeroshade: can we use something like "failed to find type for source field ID" to distinguish this error from the above identical one?

zeroshade (changes requested):

fallintoplace:
@zeroshade @laskoviymishka 

could you please review this PR once again? Feel free to suggest changes or push additional commits.

zeroshade (approved):
LGTM, thanks!!

PR #1113, opened 2026-05-21
feat(table): wire DVWriter into position-delete producer with property gating

tanmayrauth:
Wires DVWriter into the position-delete producer with property gating.

  Adds write.delete.format table property (v2 default: position, v3 default: dv). When set to "dv", positionDeleteRecordsToDataFiles routes through DVWriter instead of the Parquet position-delete writer. The existing v3 warning now only fires when write.delete.format is explicitly set to "position".

  Part 2 of #997

  Out of scope (tracked as follow-ups):
  - In-snapshot DV merging when multiple writes target the same data file.
  - v2 Parquet pos-delete → v3 DV upgrade path within a single snapshot.
  - Full cross-client scan test lands in PR #3.

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1647,19 +1647,18 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string, 
 		}
 	}
 
-	deleteFormat := meta.props.Get(WriteDeleteFormatKey, "")
-	if deleteFormat == "" {
-		if latestMetadata.Version() >= 3 {
-			deleteFormat = WriteDeleteFormatDV
-		} else {
-			deleteFormat = WriteDeleteFormatPosition
-		}
-	}
-
-	if deleteFormat == WriteDeleteFormatDV {
+	// V3+ unpartitioned tables write deletion vectors via Puffin, matching the
+	// Java reference implementation, which hardcodes DV for v3 and does not
+	// expose a property to opt out. DV+partitioned support is deferred to a
+	// follow-up; partitioned v3 writes fall through to the Parquet writer
+	// below and emit the deprecation warning.
+	if latestMetadata.Version() >= 3 && latestMetadata.PartitionSpec().IsUnpartitioned() {
 		return positionDeleteRecordsToDataFilesDV(ctx, rootLocation, args)
 	}
 
+	// V3 and later prefer deletion vectors over Parquet position-delete files;
+	// warn so users migrate. The check is `>= 3` rather than `== 3` so the
+	// warning carries forward to v4+ without churn. See apache/iceberg#12048.
 	if latestMetadata.Version() >= 3 {
 		slog.Warn("writing Parquet position-delete file on a v3 table; prefer deletion vectors",
 			"table_location", latestMetadata.Location())
@@ -1719,6 +1718,7 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
 	return func(yield func(iceberg.DataFile, error) bool) {
 		writer := dv.NewDVWriter(args.fs)
 
+		hasEntries := false
 		for batch, err := range args.itr {
 			if err != nil {
 				yield(nil, err)
@@ -1731,14 +1731,19 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
 
 			for i := range batch.NumRows() {
 				writer.Add(filePaths.Value(int(i)), []int64{positions.Value(int(i))})
+				hasEntries = true
 			}
 		}
 
-		u := uuid.Must(uuid.NewRandom())
-		if args.writeUUID != nil {
-			u = *args.writeUUID
+		if !hasEntries {
+			return
 		}
-		location := rootLocation + "/data/" + fmt.Sprintf("dv-%s.puffin", u)
+
+		if args.writeUUID == nil {
+			u := uuid.Must(uuid.NewRandom())
+			args.writeUUID = &u
+		}
+		location := rootLocation + "/data/" + fmt.Sprintf("00000-0-%s-deletes.puffin", *args.writeUUID)
 
 		dataFiles, err := writer.Flush(ctx, location)
 		if err != nil {
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -40,6 +40,7 @@ import (
 	"github.com/apache/iceberg-go/config"
 	"github.com/apache/iceberg-go/internal"
 	iceio "github.com/apache/iceberg-go/io"
+	"github.com/apache/iceberg-go/table/dv"
 	tblutils "github.com/apache/iceberg-go/table/internal"
 	"github.com/google/uuid"
 	"github.com/pterm/pterm"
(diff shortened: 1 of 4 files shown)


tanmayrauth:
@zeroshade @laskoviymishka Can you please review this part 2?

table/arrow_utils.go
```diff
@@ -1646,10 +1647,19 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string,
...
+		if latestMetadata.Version() >= 3 {
+			deleteFormat = WriteDeleteFormatDV
+		} else {
+			deleteFormat = WriteDeleteFormatPosition
+		}
+	}
+
+	if deleteFormat == WriteDeleteFormatDV {
+		return positionDeleteRecordsToDataFilesDV(ctx, rootLocation, args)
+	}
```
zeroshade: it looks like we're losing track of the partitions here since dv_writer.go:131 doesn't seem to push the partition data into the data files, despite us having the `partitionContextByFile` here. We probably need to push the partition context through the new `positionDeleteRecordsToDataFilesDV` function and all the way into the DVWriter so that we get the correct partition spec ID and partition structs into the dv manifests.

You should be able to create a `TestDVWritePathPartitioned` test for this and assert that every returned DataFile has the proper SpecID and Partition fields.

table/properties.go
```diff
@@ -76,6 +76,11 @@ const (
 	WriteDeleteModeKey     = "write.delete.mode"
 	WriteDeleteModeDefault = WriteModeCopyOnWrite
 
+	WriteDeleteFormatKey = "write.delete.format"
+
+	WriteDeleteFormatPosition = "position"
+	WriteDeleteFormatDV       = "dv"
```
zeroshade: the Java library uses this key for different semantics: `write.delete.format.default` refers to the *file format* and should be `parquet`, `avro` or `orc`. Let's not reuse an existing key and give it different semantics. 

The Java library hardcodes using puffin and DVs for v3 instead of letting users change it with a property. We should probably follow the same behavior instead of introducing the new property.

table/dv_write_path_test.go
```diff
@@ -0,0 +1,123 @@
...
+	mb := newPositionDeleteUnpartitionedMetadata(t, 2)
+
+	emptyItr := func(yield func(arrow.RecordBatch, error) bool) {}
+
+	seq := positionDeleteRecordsToDataFiles(context.Background(), t.TempDir(), mb, nil,
+		recordWritingArgs{
+			sc:  PositionalDeleteArrowSchema,
+			itr: emptyItr,
+			fs:  iceio.LocalFS{},
+		})
```
zeroshade: if `len(w.entries) == 0` in `positionDeleteRecordsToDataFiles` we return early from Flush, so since you're using the empty iterator this isn't actually testing anything. Likewise, the position delete path will produce no files and no errors on an empty case. We need to feed a non-empty iterator and then assert on the FileFormat of the first returned DataFile

table/arrow_utils.go
```diff
@@ -1704,3 +1714,43 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string,
...
+
+			for i := range batch.NumRows() {
+				writer.Add(filePaths.Value(int(i)), []int64{positions.Value(int(i))})
+			}
+		}
+
+		u := uuid.Must(uuid.NewRandom())
+		if args.writeUUID != nil {
+			u = *args.writeUUID
+		}
```
zeroshade: we have this same block (almost, the other version is better) up at line 1668, can we avoid duplicating that? Or at least move the `u := uuid.Must....` inside the if?

table/arrow_utils.go
```diff
@@ -1704,3 +1714,43 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string,
...
+		location := rootLocation + "/data/" + fmt.Sprintf("dv-%s.puffin", u)
```
zeroshade: should we follow the Java lib convention of `00000-0-<UUID>-deletes.puffin` instead?

table/arrow_utils.go
```diff
@@ -1646,10 +1647,19 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string,
 		}
 	}
 
-	// V3 and later prefer deletion vectors over Parquet position-delete files;
-	// warn so users migrate when DV-write support lands. The check is `>= 3`
-	// rather than `== 3` so the warning carries forward to v4+ without churn.
-	// See apache/iceberg#12048.
```
zeroshade, line 1669: we probably shouldn't drop this comment

table/pos_delete_v3_warn_test.go
```diff
@@ -123,20 +123,20 @@ func runPositionDeleteWrite(t *testing.T, mb *MetadataBuilder, partitions map[st
...
-// silently regress this property without it.
+// On v3 with the default (or explicit "dv") format, the DV writer is used
+// instead and no warning is emitted.
 func TestPositionDeleteV3Warning(t *testing.T) {
 	emptyItr := func(yield func(arrow.RecordBatch, error) bool) {}
 
-	t.Run("v3 unpartitioned warns once with table location", func(t *testing.T) {
+	t.Run("v3 with explicit position format warns once with table location", func(t *testing.T) {
+		mb := newPositionDeleteUnpartitionedMetadata(t, 3)
+		mb.props = iceberg.Properties{WriteDeleteFormatKey: WriteDeleteFormatPosition}
```
zeroshade: use `mb.SetProperties` instead of using the internal field here.

table/arrow_utils.go
```diff
@@ -1704,3 +1714,43 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string,
...
+			u = *args.writeUUID
+		}
+		location := rootLocation + "/data/" + fmt.Sprintf("dv-%s.puffin", u)
+
+		dataFiles, err := writer.Flush(ctx, location)
+		if err != nil {
+			yield(nil, err)
+
+			return
+		}
```
zeroshade, line 1753: move this early return above line 1737 and also do a `if len(dataFiles) == 0 { return }` early return. That way we can avoid the uuid generation and location string handling if we're gonna return early.

zeroshade (changes requested):
A few questions and nitpicks:

- Should we consider the partitioned case out of scope for this PR and track it as a follow-up?
- Can we add a test that verifies cancellation via a cancelled context?
- Can we add a test that errors mid-stream (after at least one good batch) and assert that we get the error back?
- Can we add a test that verifies we merge multi-batch deletes which target the same file? i.e. two batches that both target the same file should merge into the same bitmap

table/arrow_utils.go
tanmayrauth: Restructured. The DV producer now drains the iterator (early-return on iterator error), returns early if no positions were added (no UUID, no location, no Flush), then generates the UUID and location and calls Flush, with an early-return on Flush error before any successful yield. UUID and location work is fully skipped on the empty path.

table/arrow_utils.go
tanmayrauth: Restored.

table/arrow_utils.go
tanmayrauth: Switched the DV path to the cleaner form used by the Parquet path (only generate a UUID when args.writeUUID is nil) and moved it below the empty-input early-return so we no longer generate a UUID when  there's nothing to flush.

table/dv_write_path_test.go
tanmayrauth: You're right, the empty iterator made it a no-op assertion. Reworked it: the v2 fallback test now feeds a non-empty batch and asserts that the first returned DataFile's format is Parquet. Also added a v3-partitioned fallback test with the same Parquet assertion. The explicit-property test went away with comment #2.

table/properties.go
tanmayrauth: Agreed,  both the semantic conflict with Java's write.delete.format.default and the "Java doesn't expose a knob" point. Dropped the property and its constants entirely. Routing is now purely version + partitioning driven: v3+ unpartitioned uses DV, everything else uses Parquet position deletes.

table/arrow_utils.go
tanmayrauth: Good catch. Going to defer DV+partitioned support to a follow-up to keep this PR scoped. Updated the routing so v3 partitioned writes fall through to the existing Parquet position-delete writer (which already handles partition fanout correctly) and continue to emit the prefer-DV warning. The DV branch is now gated on v3+ AND unpartitioned, with a comment explaining why. Will open a follow-up to thread the partition context map through DVWriter and add the partitioned test.

tanmayrauth:
@zeroshade PTAL when you have a moment.

tanmayrauth:
@zeroshade  can you please review this PR?

zeroshade (approved):

PR #1118, opened 2026-05-22
perf(table): dedupe shared manifest reads during snapshot expiration

### LINKED ISSUE #1117: perf(table): removeSnapshotsUpdate.PostCommit re-reads shared manifests once per expired snapshot
### Apache Iceberg version
  
  main (and v0.5.1)
  
  ### Description
  
  `table.removeSnapshotsUpdate.PostCommit` (`table/updates.go`) walks every expired
  snapshot's manifest list and opens each referenced manifest file individually:
  
  ```go
  for _, snapId := range u.SnapshotIDs {
      snap := preTable.SnapshotByID(snapId)
      mans, err := snap.Manifests(prefs)
      for _, man := range mans {
          for entry, err := range man.Entries(prefs, false) { ... }
      }   
  }   
  ```

  Iceberg manifests are shared by reference across snapshots — an APPEND commit
  produces a new manifest list pointing at all the existing manifests plus 1–2
  new ones. So a shared manifest gets opened once per expired snapshot that
  references it, instead of once total.

  For a table with 491 incremental-append snapshots, expiring 490 of them
  causes ~sum(1..490) = ~120k manifest-file downloads from object storage where
  ~500 unique reads would suffice. We observed a single-table expire running
  for hours in this state.

  The retained-snapshot pass below has the same shape and could also dedupe
  across retained snapshots.

  Willing to contribute

  - I can contribute a fix for this bug independently

paveon:
Fixes https://github.com/apache/iceberg-go/issues/1117

### What changed
Reworked `removeSnapshotsUpdate.PostCommit` so each unique manifest file is
opened at most once per call, regardless of how many expired or retained
snapshots reference it.

Two passes, both deduped:

1) Build the set of manifest paths reachable from any retained snapshot,
reading only manifest-lists. Cache the resulting `[]ManifestFile` per
snapshot so the retained-side pass below doesn't re-download each list.
2) Walk expired snapshots' manifest lists; for each manifest, skip if it's
in the retained set (its data files are live by definition and the
manifest itself must not be deleted) or if a prior expired snapshot
already enumerated it. Otherwise read its entries once.
3) Subtract live data files via a single walk over each unique retained
manifest. DELETED entries remain tombstones (unchanged from prior
semantics).

### Behavior
Semantically equivalent to the previous implementation — the final
`filesToDelete` set is the same on well-formed metadata. No spec change,
no API change. The only difference is the I/O cost.

### Performance impact
For a 491-snapshot incremental-append table where expiring 490 snapshots
previously triggered ~sum(1..490) ≈ 120,000 manifest-file downloads, the
rewrite reduces that to roughly the count of unique orphaned manifests
(a few hundred in practice). Two-to-three orders of magnitude fewer
object-store reads, in our test.

### DIFF
--- a/table/updates.go
+++ b/table/updates.go
@@ -506,19 +506,21 @@ func (u *removeSnapshotsUpdate) PostCommit(ctx context.Context, preTable *Table,
 		}
 	}
 
-	// Manifest paths kept alive by retained snapshots, plus their
-	// loaded manifest-file slices so the live-data-file pass below
-	// doesn't re-download each manifest list.
+	// Preload retained manifest lists once so the live-data-file pass below can walk them without re-fetching.
+	retainedSnapshots := postTable.Metadata().Snapshots()
 	retainedManifests := make(map[string]struct{})
-	retainedSnapshotManifests := make(map[int64][]iceberg.ManifestFile)
-	for _, snap := range postTable.Metadata().Snapshots() {
+	retainedSnapshotManifests := make([]iceberg.ManifestFile, 0, len(retainedSnapshots))
+	for _, snap := range retainedSnapshots {
 		mans, err := snap.Manifests(prefs)
 		if err != nil {
 			return err
 		}
-		retainedSnapshotManifests[snap.SnapshotID] = mans
 		for _, man := range mans {
-			retainedManifests[man.FilePath()] = struct{}{}
+			manPath := man.FilePath()
+			if _, ok := retainedManifests[manPath]; !ok {
+				retainedManifests[manPath] = struct{}{}
+				retainedSnapshotManifests = append(retainedSnapshotManifests, man)
+			}
 		}
 	}
 
@@ -530,7 +532,7 @@ func (u *removeSnapshotsUpdate) PostCommit(ctx context.Context, preTable *Table,
 	for _, snapId := range u.SnapshotIDs {
 		snap := preTable.SnapshotByID(snapId)
 		if snap == nil {
-			return errors.New("missing snapshot")
+			return fmt.Errorf("missing snapshot %d", snapId)
 		}
 
 		mans, err := snap.Manifests(prefs)
@@ -561,17 +563,9 @@ func (u *removeSnapshotsUpdate) PostCommit(ctx context.Context, preTable *Table,
 	}
 
 	// Keep files still referenced (non-DELETED) by retained manifests,
-	// including data files carried forward as EXISTING by manifest
-	// merges. Each retained manifest is walked at most once.
-	walkedRetained := make(map[string]struct{}, len(retainedManifests))
-	for _, mans := range retainedSnapshotManifests {
-		for _, man := range mans {
-			manPath := man.FilePath()
-			if _, ok := walkedRetained[manPath]; ok {
-				continue
-			}
-			walkedRetained[manPath] = struct{}{}
-
+	// including data files carried forward as EXISTING by manifest merges.
+	if len(filesToDelete) > 0 {
+		for _, man := range retainedSnapshotManifests {
 			for entry, err := range man.Entries(prefs, false) {
 				if err != nil {
 					return err
--- a/table/updates.go
+++ b/table/updates.go
@@ -506,6 +506,27 @@ func (u *removeSnapshotsUpdate) PostCommit(ctx context.Context, preTable *Table,
 		}
 	}
 
+	// Manifest paths kept alive by retained snapshots, plus their
+	// loaded manifest-file slices so the live-data-file pass below
+	// doesn't re-download each manifest list.
+	retainedManifests := make(map[string]struct{})
+	retainedSnapshotManifests := make(map[int64][]iceberg.ManifestFile)
+	for _, snap := range postTable.Metadata().Snapshots() {
+		mans, err := snap.Manifests(prefs)
+		if err != nil {
+			return err
+		}
+		retainedSnapshotManifests[snap.SnapshotID] = mans
+		for _, man := range mans {
+			retainedManifests[man.FilePath()] = struct{}{}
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM! The dedup is clean and the perf win is significant.
  
 Non-blocking nit: would be great to add a test covering a manifest shared between an expired and a retained snapshot, so the dedup semantics are pinned down for future edits to either pass.

table/updates.go
```diff
@@ -506,6 +506,27 @@ func (u *removeSnapshotsUpdate) PostCommit(ctx context.Context, preTable *Table,
 		}
 	}
 
+	// Manifest paths kept alive by retained snapshots, plus their
```
laskoviymishka: small nit: "doesn't re-download each manifest list" reads like the whole function skips manifest-list reads, but the expired loop below still calls `snap.Manifests(prefs)` per expired snapshot. I'd tighten to something like `// Preload retained manifest lists once so the live-data-file pass below can walk them without re-fetching.` to keep the scope honest.

table/updates.go
```diff
@@ -506,6 +506,27 @@ func (u *removeSnapshotsUpdate) PostCommit(ctx context.Context, preTable *Table,
...
+	// loaded manifest-file slices so the live-data-file pass below
+	// doesn't re-download each manifest list.
+	retainedManifests := make(map[string]struct{})
+	retainedSnapshotManifests := make(map[int64][]iceberg.ManifestFile)
```
laskoviymishka: the snapshot-id key here is never read — the only consumer below ranges `for _, mans := range retainedSnapshotManifests`. I'd drop the map and either flatten while we're collecting or keep a `[][]iceberg.ManifestFile` so the walkedRetained pass below isn't paying for a map keyed on a field nobody uses.

while we're here, the inner loop already gives us each `man` once per `(snap, man)` pair — we could just dedupe by `man.FilePath()` during this pass and hand the second pass a flat `[]iceberg.ManifestFile` of unique manifests, which also removes the need for `walkedRetained` below. wdyt?

table/updates.go
```diff
@@ -506,6 +506,27 @@ func (u *removeSnapshotsUpdate) PostCommit(ctx context.Context, preTable *Table,
...
+		retainedSnapshotManifests[snap.SnapshotID] = mans
+		for _, man := range mans {
+			retainedManifests[man.FilePath()] = struct{}{}
+		}
+	}
+
+	// Open each orphaned manifest at most once: skip manifests that
+	// retained snapshots still reference, and dedupe across expired
+	// snapshots that share manifests by reference.
+	visitedManifests := make(map[string]struct{})
```
laskoviymishka, line 530: the dedup invariant this PR is built on isn't covered by any test. `TestRemoveSnapshotsPostCommitDeletesStatisticsFiles` and `TestRemoveSnapshotsPostCommitPreservesStatisticsOfSurvivingSnapshots` both use snapshots with empty manifest lists, so `snap.Manifests()` returns `[]` and neither of the loops you're changing here ever executes.

Adding two cases against a trackingIO-style fs would lock this in: (a) manifest M shared between an expired snapshot and a retained one, asserting M and its data files survive; (b) manifest M shared between two expired snapshots, asserting M+its data files are deleted exactly once and `Entries()` is called once. The second one also pins the perf claim.

table/updates.go
```diff
@@ -529,14 +560,17 @@ func (u *removeSnapshotsUpdate) PostCommit(ctx context.Context, preTable *Table,
...
-	for _, snap := range postTable.Metadata().Snapshots() {
-		mans, err := snap.Manifests(prefs)
-		if err != nil {
-			return err
-		}
-
+	// Keep files still referenced (non-DELETED) by retained manifests,
+	// including data files carried forward as EXISTING by manifest
+	// merges. Each retained manifest is walked at most once.
+	walkedRetained := make(map[string]struct{}, len(retainedManifests))
```
laskoviymishka: minor perf regression in the "nothing to delete" path: when `filesToDelete` is empty (no expired snapshots had unique manifests), this pass still opens every retained manifest and iterates entries to call `delete` on a no-op key. I'd guard the whole retained walk with `if len(filesToDelete) > 0`. Restores the short-circuit and keeps the PR's spirit of "open each manifest at most once, and only when it matters".

laskoviymishka (changes requested):
Nice job!

I walked through both passes, and the final `filesToDelete` result looks the same as before.

One ask before merge: the existing `TestRemoveSnapshotsPostCommit*` cases use snapshots with empty manifest lists, so they don’t really cover the loops being changed here. Could you add a test with:

* one manifest shared by an expired snapshot and a retained snapshot
* one manifest shared by two expired snapshots

That would pin the invariant from the PR description, and the second case also covers the perf claim.

Small cleanup while you’re there: `retainedSnapshotManifests` doesn’t use the map value, so a flat deduped `[]ManifestFile` would work and should also let `walkedRetained` go away.

Left a couple smaller comments inline.

table/updates.go
paveon: Good point, comment changed.

table/updates.go
paveon: Thanks for the suggestion, I simplified it based on your comment and went with the flat `[]iceberg.ManifestFile` slice with unique manifests.

table/updates.go
paveon: Added two new unit tests to verify this behaviour. I also refactored the previously existing tests a little bit to reduce the duplicity of creating metadata strings over and over if that's okay.

table/updates.go
paveon: Good point, added the guard.

laskoviymishka (approved):
Thanks for contributing!

🚢 👍

PR #1120, opened 2026-05-23
feat(table): MoR equality-rewrite - insert/reinsert writer API

### LINKED ISSUE #999: feat(table): preserve row lineage through copy-on-write and merge-on-read

tanmayrauth:
Adds `table.PositionDeltaWriter` for the MoR insert/reinsert split on v3  tables. Mirrors the data-file half of Java's `SparkPositionDeltaWrite`:

  - `Reinsert(batch)` — survivor rows from a position-delta rewrite, with explicit non-null `_row_id` values to preserve identity.
  - `Insert(batch)` — fresh rows; writer appends a null `_row_id` column so  the reader synthesizes a new ID at scan time.

  Output schema is `SchemaWithRowID` (a new helper): `_row_id` written  explicitly, `_last_updated_sequence_number` left absent so it inherits from the manifest entry's `data_sequence_number` — exactly the value the spec requires for rewritten rows.

  Scope: data files only. The position-delete entries that pair with  reinserts (turning an UPDATE into delete-old + reinsert) are the engine driver's responsibility — the writer doc spells this out so a reader doesn't expect Java's full `SparkPositionDeltaWrite` semantics from the name alone.

Part 2 of #999 
Closes: #999

### DIFF
--- a/table/position_delta_writer.go
+++ b/table/position_delta_writer.go
@@ -29,19 +29,28 @@ import (
 	"github.com/google/uuid"
 )
 
-// PositionDeltaWriter implements the position-delta write pattern for MoR
-// updates. It distinguishes between reinserted rows (survivors from a
-// position-delete that preserve their original _row_id) and fresh inserts
-// (new rows that get identity at read time).
+// PositionDeltaWriter writes data files for the position-delta MoR pattern,
+// distinguishing reinserted rows (survivors of a position-delta rewrite that
+// preserve their original _row_id) from fresh inserts (rows that get a new
+// _row_id synthesized at read time).
 //
-// This mirrors Java's SparkPositionDeltaWrite.reinsert(meta, row) vs
-// insert(row) pattern.
+// Scope: this writer produces *data files only*. The position-delete entries
+// that pair with reinserts (and that turn an UPDATE into delete-old + reinsert)
+// are not emitted here — the engine driver is responsible for composing this
+// writer with a position-delete writer and committing both via a RowDelta-style
+// snapshot. This mirrors the data-file half of Java's
+// SparkPositionDeltaWrite.reinsert(meta, row) vs insert(row) split.
+//
+// _last_updated_sequence_number is intentionally not written: the row-id-only
+// file schema lets the reader synthesize it from the manifest entry's
+// data_sequence_number, which after the rewrite is the new snapshot's sequence
+// number — the value the spec requires.
 //
 // Usage:
 //
-//	w := table.NewPositionDeltaWriter(tbl)
-//	w.Reinsert(survivorBatch)  // batch must include _row_id column
-//	w.Insert(freshBatch)       // batch without _row_id (or null values)
+//	w, err := table.NewPositionDeltaWriter(tbl)
+//	w.Reinsert(survivorBatch)  // batch must include _row_id column with non-null values
+//	w.Insert(freshBatch)       // batch without _row_id (writer appends nulls)
 //	dataFiles, err := w.Close(ctx)
 type PositionDeltaWriter struct {
 	tbl       *Table
@@ -71,9 +80,7 @@ func NewPositionDeltaWriter(tbl *Table) (*PositionDeltaWriter, error) {
 
 // Reinsert adds survivor rows that preserve their original _row_id. The batch
 // MUST contain a _row_id column (field name "_row_id") with non-null int64
-// values representing the preserved row identities. _last_updated_sequence_number
-// should be absent or null — it will inherit from the new file's
-// data_sequence_number at read time.
+// values representing the preserved row identities.
 //
 // Per the Iceberg spec, only position-delete rewrites and CoW can preserve
 // lineage. Equality deletes cannot preserve lineage because the engine writes
@@ -101,12 +108,19 @@ func (w *PositionDeltaWriter) Reinsert(batch arrow.RecordBatch) error {
 	return nil
 }
 
-// Insert adds fresh rows that get new identity at read time. The batch should
-// NOT contain a _row_id column, or if it does, all values must be null. These
-// rows represent genuinely new data (not survivors of a rewrite).
+// Insert adds fresh rows that get a new _row_id at read time. The batch should
+// NOT contain a _row_id column; if it does, all values must be null. These rows
+// represent genuinely new data (not survivors of a rewrite).
 func (w *PositionDeltaWriter) Insert(batch arrow.RecordBatch) error {
 	if w.closed {
 		return fmt.Errorf("%w: writer is already closed", ErrInvalidOperation)
+	}
+
+	if indices := batch.Schema().FieldIndices(iceberg.RowIDColumnName); len(indices) > 0 {
+		if batch.Column(indices[0]).NullN() != int(batch.NumRows()) {
+			return fmt.Errorf("%w: Insert batch %s column must be all null (use Reinsert for preserved IDs)",
+				iceberg.ErrInvalidArgument, iceberg.RowIDColumnName)
+		}
 	}
 
 	batch.Retain()
@@ -116,12 +130,12 @@ func (w *PositionDeltaWriter) Insert(batch arrow.RecordBatch) error {
 }
 
 // Close finalizes the writer and returns the data files produced. The returned
-// files contain both reinserted and fresh rows, with the _row_id column
-// written explicitly for reinserted rows (non-null) and left null for fresh
-// inserts.
+// files contain both reinserted and fresh rows, with the _row_id column written
+// explicitly for reinserted rows (non-null) and left null for fresh inserts.
 //
 // The caller is responsible for adding these files to a snapshot (typically via
-// a Transaction's snapshot producer).
+// a Transaction's snapshot producer) along with any position-delete entries
+// that pair with the reinserts.
 func (w *PositionDeltaWriter) Close(ctx context.Context) ([]iceberg.DataFile, error) {
 	if w.closed {
 		return nil, fmt.Errorf("%w: writer is already closed", ErrInvalidOperation)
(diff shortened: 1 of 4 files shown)


tanmayrauth:

table/position_delta_writer.go
```diff
@@ -0,0 +1,241 @@
...
+// iterator. Reinsert batches already have _row_id; insert batches get a null
+// _row_id column appended.
+func (w *PositionDeltaWriter) buildUnifiedIterator() iter.Seq2[arrow.RecordBatch, error] {
+	return func(yield func(arrow.RecordBatch, error) bool) {
+		alloc := memory.NewGoAllocator()
+
+		for _, batch := range w.reinsertBatches {
+			batch.Retain()
+			if !yield(batch, nil) {
+				batch.Release()
```
laskoviymishka: I think there's a double-release on the early-stop path.

`buildUnifiedIterator` calls `batch.Retain()` and `batch.Release()` in the `!yield` branch here (and the analogous `enriched.Release()` at line 206). But `WriteRecords`'s `releasing` wrapper (write_records.go:191) *also* calls `rec.Release()` on its own `!yield` branch — same pointer, two releases. Arrow-go's behavior below zero refcount is undefined, so happy-path tests don't catch it; any IO error mid-write silently corrupts the allocator.

I'd drop the `batch.Release()` (and `enriched.Release()`) from the `!yield` branches here and let `releasing` own it — the contract on `WriteRecords` ("releases each RecordBatch it consumes") already covers that. A regression test that aborts the iterator mid-stream and asserts `CurrentAlloc() == 0` would lock it in.

table/position_delta_writer.go
```diff
@@ -0,0 +1,241 @@
...
+	bldr.AppendNulls(int(nrows))
+	nullCol := bldr.NewArray()
+	defer nullCol.Release()
+
+	fields := append(batch.Schema().Fields(), arrow.Field{
+		Name:     iceberg.RowIDColumnName,
+		Type:     arrow.PrimitiveTypes.Int64,
+		Nullable: true,
+	})
+	cols := append(batch.Columns(), nullCol)
```
laskoviymishka: `batch.Columns()` returns the record's internal `arrs` slice directly. If `len(arrs) < cap(arrs)` (which can happen depending on how the batch was constructed upstream), this `append` writes `nullCol` into `batch.arrs[len]` in place — silently mutating the source batch's column array. Same risk one line up with `batch.Schema().Fields()` if the Arrow version doesn't defensively copy.

Three-index slice it: `cols := append(batch.Columns()[:batch.NumCols():batch.NumCols()], nullCol)`. Or just `make` + `copy` explicitly.

table/position_delta_writer_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	require.Contains(t, rowsByID, int64(1))
+	require.Contains(t, rowsByID, int64(2))
+	require.Contains(t, rowsByID, int64(99))
+
+	assert.Equal(t, int64(0), rowsByID[1], "reinserted id=1 must preserve original _row_id=0")
+	assert.Equal(t, int64(1), rowsByID[2], "reinserted id=2 must preserve original _row_id=1")
+	// The fresh row's _row_id must collide with no preserved survivor's id.
+	assert.NotEqual(t, int64(0), rowsByID[99])
+	assert.NotEqual(t, int64(1), rowsByID[99])
```
laskoviymishka: This test doesn't actually distinguish "we wrote `_row_id` explicitly" from "the reader synthesized it and happened to land on the same values."

If the writer silently dropped `_row_id` (hypothetical bug), the new file's `first_row_id` could easily be 0, and `synthesizeRowLineageColumns` would produce row_ids 0, 1, 2 for the three rows — exactly what the test asserts. The `NotEqual(rowsByID[99], 0/1)` guard only catches it if the file's `first_row_id` happens to overlap reinsert IDs.

Two ways to strengthen, wdyt:

- Pick reinsert `_row_id` values that can't collide with any plausible synthesized range, e.g. `[]int64{1 << 40, (1<<40) + 1}`. Then `Equal` checks become real proof.
- Or read the data file directly and assert the `_row_id` column physically exists with non-null values matching what we passed.

The first is cheaper.

table/position_delta_writer.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	nrows := batch.NumRows()
+	bldr := array.NewInt64Builder(alloc)
+	defer bldr.Release()
+	bldr.AppendNulls(int(nrows))
+	nullCol := bldr.NewArray()
+	defer nullCol.Release()
+
+	fields := append(batch.Schema().Fields(), arrow.Field{
```
laskoviymishka: The appended Arrow field has no `PARQUET:field_id` metadata. In practice the static `arrowSc` we pass to `WriteRecords` (derived from `SchemaToArrowSchema(SchemaWithRowID(...), ...)`) carries the field ID, so the Parquet writer tags the column correctly downstream — but per-batch the field is unannotated, which is brittle: any future code path that reads field IDs from the per-batch schema rather than the static one would silently misroute the column.

Worth attaching `ArrowParquetFieldIDKey → strconv.Itoa(iceberg.RowIDFieldID)` here so the two schemas agree.

table/position_delta_writer.go
```diff
@@ -0,0 +1,241 @@
...
+	reinsertBatches []arrow.RecordBatch
+	// insertBatches hold fresh records without lineage.
+	insertBatches []arrow.RecordBatch
+
+	closed bool
+}
+
+// NewPositionDeltaWriter creates a writer for the position-delta MoR update
+// pattern on the given table. The table must be format version 3 or higher.
+func NewPositionDeltaWriter(tbl *Table) (*PositionDeltaWriter, error) {
```
laskoviymishka: No way for callers to thread `WithTargetFileSize`, `WithMaxWriteWorkers`, or `WithClusteredWrite` through. A large rewrite silently uses table defaults, which may produce a different file count/shape than the caller expects.

Could we take `opts ...WriteRecordOption` and merge them at the `WriteRecords` call site? `WithWriteUUID` and `WithPreserveRowLineage` stay hardcoded; everything else flows from the caller. Thoughts?

table/position_delta_writer.go
```diff
@@ -0,0 +1,241 @@
...
+//	w.Insert(freshBatch)       // batch without _row_id (writer appends nulls)
+//	dataFiles, err := w.Close(ctx)
+type PositionDeltaWriter struct {
+	tbl       *Table
+	writeUUID uuid.UUID
+
+	// reinsertBatches hold records with explicit _row_id values (preserving lineage).
+	reinsertBatches []arrow.RecordBatch
+	// insertBatches hold fresh records without lineage.
+	insertBatches []arrow.RecordBatch
```
laskoviymishka, line 75: Both slices grow unbounded until `Close()` flushes, so a 10GB partition rewrite materializes 10GB of Arrow before a byte hits disk. For the engine-driver use cases this writer targets, that might be fine; for compaction it's not.

Either document the memory model on the type doc explicitly so callers don't get surprised, or take the iterator at construction and stream straight through to `WriteRecords` rather than accumulating. I'd be fine with documenting it for now and revisiting in the position-delete writer PR.

laskoviymishka (changes requested):
The shape matches what we settled on in #999 and reads clean, lines up with #1099 nicely.

One thing I'd want fixed before merge: there's a double-release in buildUnifiedIterator on the early-stop path. The producer Retains and then Releases on `!yield`, but releasing inside `WriteRecords` already releases unconditionally on that branch - happy path is fine but any IO error mid-write quietly takes the refcount below zero. Inline has the fix proposal.

Two other items I'd want pulled into this PR:

- `append(batch.Columns(), nullCol)` aliases the source batch's internal slice if it has spare capacity - three-index slice or explicit make + copy fixes it.
- The `ReinsertPreservesRowID` test passes whether `_row_id` is written explicitly or synthesized on read, because the reinsert IDs (0, 1) overlap any plausible first_row_id of the new file. Bumping the test values to something like `1 << 40` makes the assertion actually prove preservation.

Everything else (PARQUET:field_id on the appended field, WriteRecordOption passthrough, memory-model doc on the accumulator) is polish and can land here or in a follow-up - left inline. Once the double-release and the aliasing are addressed, happy to take another pass.

tanmayrauth:
@laskoviymishka  thanks for the review. I've made the below changes

- Dropped batch.Release() and enriched.Release() from both !yield branches in buildUnifiedIterator; added a doc comment naming the contract that the releasing wrapper in WriteRecords owns the release on both success and abort paths.
- New white-box test TestPositionDeltaWriter_BuildUnifiedIterator_AbortNoDoubleRelease uses CheckedAllocator.AssertSize(t, 0) and exercises the abort branch on the reinsert path, the insert path, and the  phase boundary as subtests.
- Three-index slices on both Fields() and Columns(). The Columns() one is the real risk — Arrow's Record.Columns() returns the internal arrs slice directly; Fields() happens to copy in the current Arrow version but the defensive slice is forward-compatible.
- Appended _row_id field now carries ArrowParquetFieldIDKey → strconv.Itoa(iceberg.RowIDFieldID) so the per-batch and static schemas agree.
- New TestPositionDeltaWriter_AppendNullRowIDColumn_NoSourceAliasing builds a source batch with cap > len, asserts the source schema/columns are untouched, and asserts the field-id metadata is present.
- Bumped preserved IDs in the lineage test to 1 << 40 and (1<<40)+1 and switched to direct Equal assertions, so a hypothetical bug where the writer drops _row_id and the reader synthesizes from first_row_id now fails instead of coinciding.
- NewPositionDeltaWriter(tbl, opts ...WriteRecordOption) threads caller-supplied options (target file size, worker count, clustered write) through to WriteRecords. Writer-owned WithPreserveRowLineage and WithWriteUUID are appended last so they take precedence on conflict; documented on the constructor.
- Added a "Memory model" paragraph on the type doc making the accumulate-then-flush behavior explicit and noting the streaming variant is deferred to the position-delete writer PR.
- Full table test suite passes and go vet is clean.

tanmayrauth:
@laskoviymishka can you please review this when you have a moment?

laskoviymishka (approved):
Nice work!
:shipit:

PR #1126, opened 2026-05-24
fix(table): reject reserved metadata IDs before reassignment

### LINKED ISSUE #1107: Bug(schema): AssignFreshSchemaIDs Java Divergance
### Apache Iceberg version

None

### Please describe the bug 🐞

`AssignFreshSchemaIDs (schema.go:1366)` renumbers reserved metadata column IDs along with everything else. 

When a v3 table's `_row_id` ends up at id=3 in the projection, `arrowAccessor.FieldPartner (table/arrow_utils.go:677)` will look up id=3 in file schemas — Java-written files have _row_id at id=MaxInt32-107, so we'd silently read whichever column the Java file has at id=3 instead.

There are two ways to solve this issue: Either special-casing reserved ids in `AssignFreshSchemaIDs` or having `NewMetadata` reject metadata columns in the user schema (which matches Java). Both are reasonable.

cc @laskoviymishka

officialasishkumar:
Summary

- reject user schemas that contain reserved metadata column IDs before schema ID reassignment
- validate nested struct, list, and map fields for reserved metadata IDs
- keep lineage projection tests focused on appendMissingLineageFields behavior

Fixes #1107

Testing

- go test ./table -run "TestNewMetadataRejectsReservedMetadataColumnIDsBeforeReassignment|TestAppendMissingLineageFieldsAlreadyHasRowID|TestNewMetadata_DuplicateValues" -count=1
- go test ./table -count=1

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2194,6 +2194,10 @@ func NewMetadataWithUUID(sc *iceberg.Schema, partitions *iceberg.PartitionSpec, 
 		}
 	}
 
+	if err := validateNoReservedMetadataColumnIDs(sc); err != nil {
+		return nil, err
+	}
+
 	reassignedIds, err := reassignIDs(sc, partitions, sortOrder)
 	if err != nil {
 		return nil, err
--- a/table/metadata_schema_compatibility.go
+++ b/table/metadata_schema_compatibility.go
@@ -84,6 +84,10 @@ func checkSchemaCompatibility(sc *iceberg.Schema, formatVersion int) error {
 	const defaultValuesMinFormatVersion = 3
 	problems := make([]IncompatibleField, 0)
 
+	if err := validateNoReservedMetadataColumnIDs(sc); err != nil {
+		return err
+	}
+
 	if err := validateUnknownTypes(sc); err != nil {
 		return fmt.Errorf("failed to validate unknown types: %w", err)
 	}
@@ -107,11 +111,6 @@ func checkSchemaCompatibility(sc *iceberg.Schema, formatVersion int) error {
 		colName, found := sc.FindColumnName(field.ID)
 		if !found {
 			panic("invalid schema: field with id " + strconv.Itoa(field.ID) + " not found, this is a bug, please report.")
-		}
-
-		if iceberg.IsMetadataColumn(field.ID) {
-			return fmt.Errorf("%w: field '%s' uses reserved metadata column ID %d",
-				iceberg.ErrInvalidSchema, colName, field.ID)
 		}
 
 		minFormatVersion := minFormatVersionForType(field.Type)
@@ -145,6 +144,54 @@ func checkSchemaCompatibility(sc *iceberg.Schema, formatVersion int) error {
 
 	if len(problems) != 0 {
 		return ErrIncompatibleSchema{fields: problems, formatVersion: formatVersion}
+	}
+
+	return nil
+}
+
+func validateNoReservedMetadataColumnIDs(sc *iceberg.Schema) error {
+	for _, field := range sc.Fields() {
+		if err := validateNoReservedMetadataColumnID(field, field.Name); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func validateNoReservedMetadataColumnID(field iceberg.NestedField, name string) error {
+	if iceberg.IsMetadataColumn(field.ID) {
+		return fmt.Errorf("%w: field '%s' uses reserved metadata column ID %d",
+			iceberg.ErrInvalidSchema, name, field.ID)
+	}
+
+	switch typ := field.Type.(type) {
+	case *iceberg.StructType:
+		if typ == nil {
+			return nil
+		}
+		for _, child := range typ.FieldList {
+			if err := validateNoReservedMetadataColumnID(child, name+"."+child.Name); err != nil {
+				return err
+			}
+		}
+	case *iceberg.ListType:
+		if typ == nil {
+			return nil
+		}
+		if err := validateNoReservedMetadataColumnID(typ.ElementField(), name+".element"); err != nil {
+			return err
+		}
+	case *iceberg.MapType:
+		if typ == nil {
+			return nil
+		}
+		if err := validateNoReservedMetadataColumnID(typ.KeyField(), name+".key"); err != nil {
+			return err
+		}
+		if err := validateNoReservedMetadataColumnID(typ.ValueField(), name+".value"); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/table/metadata_schema_compatibility_test.go
+++ b/table/metadata_schema_compatibility_test.go
@@ -50,3 +50,31 @@ func TestNewMetadata_DuplicateValues(t *testing.T) {
 	assert.ErrorIs(t, err, iceberg.ErrInvalidSchema)
 	assert.Contains(t, err.Error(), "multiple fields for name foo")
 }
+
+func TestNewMetadataRejectsReservedMetadataColumnIDsBeforeReassignment(t *testing.T) {
+	for _, field := range []iceberg.NestedField{
+		iceberg.RowID(),
+		iceberg.LastUpdatedSequenceNumber(),
+	} {
+		t.Run(field.Name, func(t *testing.T) {
+			schema := iceberg.NewSchema(
+				1,
+				iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64, Required: true},
+				field,
+			)
+
+			_, err := NewMetadata(
+				schema,
+				iceberg.UnpartitionedSpec,
+				UnsortedSortOrder,
+				"s3://test-bucket/test_table",
+				iceberg.Properties{PropertyFormatVersion: "3"},
+			)
+
+			require.Error(t, err)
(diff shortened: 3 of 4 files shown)


table/metadata_schema_compatibility.go
```diff
@@ -84,6 +84,10 @@ func checkSchemaCompatibility(sc *iceberg.Schema, formatVersion int) error {
 	const defaultValuesMinFormatVersion = 3
 	problems := make([]IncompatibleField, 0)
 
+	if err := validateNoReservedMetadataColumnIDs(sc); err != nil {
```
laskoviymishka, line 87: I think this call is effectively dead for the `NewMetadata` path — by the time `checkSchemaCompatibility` runs, `reassignIDs` has already remapped every ID to a fresh non-reserved value, so the validator can't find anything. The only meaningful check is the new one in `NewMetadataWithUUID` before reassignment.

If the intent is to also catch callers that reach `checkSchemaCompatibility` without going through reassignment (e.g. `MetadataBuilder.AddSchema` later in a table's life), I'd move this guard into `AddSchema` directly and leave `checkSchemaCompatibility` focused on format-version compatibility. Otherwise we end up with two call sites for the same guard and an unclear invariant about which one is the actual gate. wdyt?

table/metadata_schema_compatibility.go
```diff
@@ -150,6 +149,54 @@ func checkSchemaCompatibility(sc *iceberg.Schema, formatVersion int) error {
...
+	switch typ := field.Type.(type) {
+	case *iceberg.StructType:
+		if typ == nil {
+			return nil
+		}
+		for _, child := range typ.FieldList {
+			if err := validateNoReservedMetadataColumnID(child, name+"."+child.Name); err != nil {
+				return err
+			}
+		}
```
laskoviymishka, line 177: The three `if typ == nil { return nil }` arms (StructType / ListType / MapType) aren't reachable — a successful type assertion on `field.Type.(type)` already discriminates against the nil-interface case, and nothing in the codebase produces a typed-nil `*iceberg.StructType` here. `validateComplexDefault` in the same file does the same switch without these guards. I'd drop them.

table/metadata_schema_compatibility.go
```diff
@@ -150,6 +149,54 @@ func checkSchemaCompatibility(sc *iceberg.Schema, formatVersion int) error {
...
 
+func validateNoReservedMetadataColumnIDs(sc *iceberg.Schema) error {
+	for _, field := range sc.Fields() {
+		if err := validateNoReservedMetadataColumnID(field, field.Name); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
laskoviymishka, line 160: Small consistency thing: the neighboring validators (`validateUnknownTypes`, `validateComplexTypeDefaults`) wrap their error with `fmt.Errorf("failed to validate ...: %w", err)` at the call site, while this one returns bare. Not a regression (the inline check it replaced also returned bare), but worth wrapping at the `checkSchemaCompatibility` call site for symmetry if we keep that call site at all.

table/metadata_schema_compatibility_test.go
```diff
@@ -50,3 +50,31 @@ func TestNewMetadata_DuplicateValues(t *testing.T) {
 	assert.ErrorIs(t, err, iceberg.ErrInvalidSchema)
 	assert.Contains(t, err.Error(), "multiple fields for name foo")
 }
+
```
laskoviymishka, line 53: Two thoughts on the new test:

The PR description says nested struct/list/map fields are validated, but the test only covers top-level reserved IDs. I'd add at least one case with a struct child or a list element carrying `RowIDFieldID` so the recursive walk is actually exercised.

Also worth adding a `NestedField{ID: iceberg.RowIDFieldID, Name: "user_id"}` case — that confirms the error message uses the schema's name, not the canonical constant, which is the point of the `name` argument in `validateNoReservedMetadataColumnID`.

table/scanner_internal_test.go
```diff
@@ -492,45 +492,32 @@ func TestProjectionWithRowLineageRequiresV3(t *testing.T) {
...
 
-// TestProjectionV3SchemaAlreadyHasRowID covers the case where the user schema
-// already declares _row_id (a reserved field id, but legal in v3). The
-// projection helper must be idempotent and not panic on the duplicate ID.
-func TestProjectionV3SchemaAlreadyHasRowID(t *testing.T) {
+// TestAppendMissingLineageFieldsAlreadyHasRowID covers the helper's idempotent
+// behavior when a schema already includes reserved lineage fields. New table
+// metadata rejects user schemas with these IDs, but projections can still
+// contain them after the scanner adds row-lineage columns.
+func TestAppendMissingLineageFieldsAlreadyHasRowID(t *testing.T) {
```
laskoviymishka, line 499: This rename narrows the test from end-to-end (`scan.Projection()` over real v3 metadata) to a unit test of `appendMissingLineageFields` in isolation. The motivation is clear — `NewMetadata` now rejects the old setup — but it leaves the full read-path projection over a schema-with-reserved-IDs uncovered, and existing tables loaded from catalogs can still have those IDs.

I'd keep this unit test and add a sibling that constructs `*Scan` with a hand-built metadata (or via a path that bypasses the new validator) and exercises `Projection()` end-to-end. Otherwise we lose the original regression target.

table/metadata.go
```diff
@@ -2194,6 +2194,10 @@ func NewMetadataWithUUID(sc *iceberg.Schema, partitions *iceberg.PartitionSpec,
 		}
 	}
 
+	if err := validateNoReservedMetadataColumnIDs(sc); err != nil {
+		return nil, err
```
laskoviymishka, line 2198: Worth a one-line comment here explaining *why* the check runs before `reassignIDs` — "user-provided IDs must be validated before reassignment overwrites them with fresh non-reserved IDs". The placement is correct but non-obvious to a future reader.

Also a heads-up: Java's `TableMetadata.newTableMetadata` doesn't reject reserved IDs here; it silently relies on `TypeUtil.assignFreshIds` to remap them. We're imposing a stricter precondition than Java. I think that's the right call (silent reassignment of a user-supplied `_row_id` is a footgun), but worth flagging in case it's intentional vs. accidental divergence.

laskoviymishka (changes requested):
Thanks for closing this gap, silently reassigning a user-supplied `_row_id` was a real footgun, and validating before reassignIDs is the right spot.

Two things I'd tidy before merge:

- Validator runs twice. NewMetadataWithUUID calls it pre-reassignment (load-bearing), and checkSchemaCompatibility calls it again on the already-reassigned schema (dead for this path). Pick one home, or leave a comment explaining which gate covers which caller.
- Test gaps. The new test only covers top-level reserved IDs, the nested struct/list/map walk that's the bulk of the new code is untested. And the `TestProjectionV3SchemaAlreadyHasRowID` rename drops the end-to-end `Scan.Projection()` path, which still matters for tables loaded from catalogs.

Once those are in, happy to take another pass.

table/metadata_schema_compatibility.go
```diff
@@ -150,6 +149,54 @@ func checkSchemaCompatibility(sc *iceberg.Schema, formatVersion int) error {
...
+
+func validateNoReservedMetadataColumnID(field iceberg.NestedField, name string) error {
+	if iceberg.IsMetadataColumn(field.ID) {
```
zeroshade, line 163: This only rejects IDs recognized by `iceberg.IsMetadataColumn`, which currently covers the row-lineage columns but not the rest of Iceberg's reserved metadata IDs. User schemas can still use `_file` (2147483646), `_pos` (2147483645), `_deleted` (2147483644), `_spec_id`, `_partition`, `file_path`, and other IDs in the reserved range before reassignment. Per the spec, table schema field IDs must not exceed 2147483447 (`Integer.MAX_VALUE - 200`). Please reject `field.ID > 2147483447` here, or expand the predicate to cover the full spec-reserved set, and add a regression test for at least one non-row-lineage reserved ID such as 2147483646.

table/metadata_schema_compatibility_test.go
```diff
@@ -50,3 +50,31 @@ func TestNewMetadata_DuplicateValues(t *testing.T) {
...
+func TestNewMetadataRejectsReservedMetadataColumnIDsBeforeReassignment(t *testing.T) {
+	for _, field := range []iceberg.NestedField{
```
zeroshade, line 55: This test only covers the two top-level row-lineage fields. Since the new validator recurses into structs, lists, and maps, please add table-driven cases for a reserved ID on a struct child, list element, map key, and map value. At least one case should use a non-canonical field name so the assertion proves the error reports the actual schema path, not just the metadata column name.

table/scanner_internal_test.go
zeroshade, line 499: Keeping a direct helper test is useful, but this removes the end-to-end `Scan.Projection()` regression coverage for metadata that already contains row-lineage fields. Please add back an end-to-end projection test over v3 metadata containing row-lineage fields so the scanner path remains covered, in addition to this helper-level idempotence check.

zeroshade (changes requested):
Thanks for putting the reserved-ID validation before ID reassignment; that is the right place to catch user schemas before the table builder normalizes IDs. I think the approach still needs a few changes before this is safe to merge: the reserved-ID predicate is incomplete relative to the spec, and the tests should cover the new recursive validation and preserve the end-to-end projection regression coverage.

zeroshade:

PR #1127, opened 2026-05-24
fix(table): cross-check deletion vector cardinality

### LINKED ISSUE #1058: feat(table/dv): cross-validate manifest record_count against blob cardinality property
Parent: #589, follows #1049.

#1049 added cardinality validation in `ReadDV` by reading the puffin blob's `cardinality` property. Java's `BitmapPositionDeleteIndex.deserializeBitmap` does the equivalent check against the manifest entry's `record_count` field (`DeleteFile.recordCount()`), not the puffin property. Both are set to the same value by Java's writer, so iceberg-go and Java agree for spec-conformant tables — but if a third-party writer sets them inconsistently (e.g. stale `record_count` after an incremental merge, fresh blob property), the two implementations would disagree on whether the DV is valid.

`ReadDV` should consult both sources when each is independently available:

- Manifest entry's `dvFile.Count()` when greater than zero — Java's primary source.
- Puffin blob's `cardinality` property — what #1049 already reads.

When both are non-zero and disagree, fail with a clear "manifest record_count N disagrees with puffin cardinality property M" error. When only one is available, use it. When neither, keep the current `slog.Warn` skip path.

Cross-client interop benefit: catches writer bugs that affect manifest/footer agreement before they propagate to scans.

officialasishkumar:
Summary

- compare deletion vector manifest record_count with the Puffin cardinality property when both are available
- fail fast when manifest and Puffin cardinality metadata disagree
- keep bitmap cardinality validation active when metadata sources agree

Fixes #1058

Testing

- go test ./table/dv -run TestReadDVCardinalityValidation -count=1
- go test ./table/dv -count=1

### DIFF
--- a/table/dv/deletion_vector.go
+++ b/table/dv/deletion_vector.go
@@ -132,22 +132,16 @@ func SerializeDV(bitmap *RoaringPositionBitmap) ([]byte, error) {
 // ReadDV reads a deletion vector from a puffin file using the manifest entry metadata.
 // ContentOffset and ContentSizeInBytes must be set on the DataFile (required by v3 spec).
 //
-// When the puffin blob carries a `cardinality` property (spec-mandated for
-// deletion-vector-v1), its value is parsed and used to validate the decoded
-// bitmap — this catches truncated or partially-overwritten blobs whose CRC
-// still validates over the bytes that are present.
+// ReadDV validates cardinality from both independently available sources:
+// the manifest entry's record_count (when greater than zero) and the Puffin
+// blob's `cardinality` property. Java's BitmapPositionDeleteIndex uses the
+// manifest entry value; Java writers also set the Puffin property to the same
+// value, so a disagreement indicates corrupt or non-conformant metadata.
 //
-// Java's BitmapPositionDeleteIndex validates against the manifest entry's
-// `record_count` field rather than the puffin blob property; the two are
-// always set to the same value by Java's writer, so this PR's behavior agrees
-// with Java for spec-conformant tables. A future change should cross-validate
-// both sources when they're independently available (tracked separately).
-//
-// Blobs missing the spec-required cardinality property are accepted with a
-// slog warning rather than rejected — strict enforcement is deferred until
-// the Go DV writer guarantees the property is always present. The per-byte
-// CRC check in DeserializeDV still applies, so missing-property is degraded
-// integrity, not absent integrity.
+// If neither source carries a usable cardinality, ReadDV logs a warning and
+// falls back to CRC-only validation. The per-byte CRC check in DeserializeDV
+// still applies, so missing cardinality is degraded integrity, not absent
+// integrity.
 func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error) {
 	if dvFile.FileFormat() != iceberg.PuffinFile {
 		return nil, fmt.Errorf("expected PUFFIN format for deletion vector, got %s", dvFile.FileFormat())
@@ -179,21 +173,36 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
 		return nil, fmt.Errorf("read DV blob at offset %d: %w", offset, err)
 	}
 
-	cardinality, ok, err := blobCardinality(reader.Blobs(), offset, size)
+	blobCardinality, hasBlobCardinality, err := blobCardinality(reader.Blobs(), offset, size)
 	if err != nil {
 		return nil, fmt.Errorf("DV file %s: %w", dvFile.FilePath(), err)
 	}
-	if !ok {
+	manifestCardinality := dvFile.Count()
+	hasManifestCardinality := manifestCardinality > 0
+
+	var expectedCardinality int64
+	switch {
+	case hasManifestCardinality && hasBlobCardinality:
+		if manifestCardinality != blobCardinality {
+			return nil, fmt.Errorf("manifest record_count %d disagrees with puffin cardinality property %d",
+				manifestCardinality, blobCardinality)
+		}
+		expectedCardinality = manifestCardinality
+	case hasManifestCardinality:
+		expectedCardinality = manifestCardinality
+	case hasBlobCardinality:
+		expectedCardinality = blobCardinality
+	default:
 		// Spec deviation: deletion-vector-v1 MUST carry a cardinality
 		// property. We tolerate the absence to keep reading from third-
 		// party writers that emit non-conformant files; flag so operators
 		// can identify affected tables.
 		slog.Warn("DV blob missing spec-required cardinality property; skipping cardinality validation",
 			"dv_file", dvFile.FilePath(), "offset", offset)
-		cardinality = -1
+		expectedCardinality = -1
 	}
 
-	return DeserializeDV(blobData, cardinality)
+	return DeserializeDV(blobData, expectedCardinality)
 }
 
 // blobCardinality returns the cardinality declared by the puffin blob at the
--- a/table/dv/deletion_vector_test.go
+++ b/table/dv/deletion_vector_test.go
@@ -413,18 +413,29 @@ func TestReadDVCardinalityValidation(t *testing.T) {
 		assert.Equal(t, int64(5), bm.Cardinality())
 	})
 
-	t.Run("mismatched cardinality is rejected", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


table/dv/deletion_vector.go
```diff
@@ -179,21 +173,36 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
...
 	}
 
-	cardinality, ok, err := blobCardinality(reader.Blobs(), offset, size)
+	blobCardinality, hasBlobCardinality, err := blobCardinality(reader.Blobs(), offset, size)
 	if err != nil {
 		return nil, fmt.Errorf("DV file %s: %w", dvFile.FilePath(), err)
 	}
-	if !ok {
+	manifestCardinality := dvFile.Count()
+	hasManifestCardinality := manifestCardinality > 0
```
laskoviymishka, line 181: I think there's a subtle issue here. `record_count` is field 103 in the spec — non-nullable `long`, no absent state — and Java writes `0` for empty DVs (BaseDVFileWriter via `deletes.positions().cardinality()`). Treating `manifestCardinality > 0` as "manifest didn't tell us" means a real disagreement like blob=5 / manifest=0 falls through to "blob-only" and gets accepted silently, which is exactly the case the PR is trying to catch.

Since `RecordCount` is a non-pointer `int64` with no representable absent state, I'd just treat it as always present — `hasManifestCardinality := true` — and let the agree/disagree branches handle zero naturally. wdyt?

table/dv/deletion_vector_test.go
```diff
@@ -413,18 +413,29 @@ func TestReadDVCardinalityValidation(t *testing.T) {
...
+			// Manifest record_count below says 5; claim 99 in the Puffin property.
 			"cardinality": "99",
 		})
 		offset, size := meta.Offset, meta.Length
 		_, err := ReadDV(iceio.LocalFS{}, newDVTestFile(path, 5, &offset, &size))
 		require.Error(t, err)
-		// Pin the specific error path: DeserializeDV's "cardinality
-		// mismatch", not the helper's "invalid cardinality" parse error.
+		assert.Contains(t, err.Error(), "manifest record_count 5 disagrees with puffin cardinality property 99")
+	})
```
laskoviymishka, line 427: Related to this PR but in the sibling subtest just below this one: the existing `"missing cardinality property is tolerated with a warning"` test doesn't reach the `default` branch anymore — `writePuffinWithDVBlob` still embeds `"cardinality": "5"` and the test passes `count=5`, so it hits the agree-path. The warning branch could be deleted and that test would keep passing.

I'd add a sibling helper (or extend `writePuffinWithDVBlobAndProps`) that omits the cardinality property entirely, paired with a manifest `count` that also can't be used, so the test actually pins the `default` warning.

table/dv/deletion_vector.go
```diff
@@ -179,21 +173,36 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
 		return nil, fmt.Errorf("read DV blob at offset %d: %w", offset, err)
 	}
 
-	cardinality, ok, err := blobCardinality(reader.Blobs(), offset, size)
+	blobCardinality, hasBlobCardinality, err := blobCardinality(reader.Blobs(), offset, size)
```
laskoviymishka, line 176: The local `blobCardinality` shadows the package-level function of the same name on this line — readability hazard since the identifier resolves to the int64 result for the rest of the function, including the three uses below (`manifestCardinality != blobCardinality`, the `Errorf` arg, and `expectedCardinality = blobCardinality` in the `case hasBlobCardinality:` arm).

I'd rename the local to `puffinCard` (mirrors the `manifestCardinality` source-prefix naming) and touch up all 4 sites — declaration plus the three uses inside the switch.

table/dv/deletion_vector.go
```diff
@@ -179,21 +173,36 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
...
 	}
-	if !ok {
+	manifestCardinality := dvFile.Count()
+	hasManifestCardinality := manifestCardinality > 0
+
+	var expectedCardinality int64
+	switch {
+	case hasManifestCardinality && hasBlobCardinality:
+		if manifestCardinality != blobCardinality {
+			return nil, fmt.Errorf("manifest record_count %d disagrees with puffin cardinality property %d",
```
laskoviymishka, line 187: Every other error in `ReadDV` carries the DV file path (the `"DV file %s: %w"` wrap above, plus `read DV blob at offset %d`, etc.); this one's the odd one out. Cheap to include and it makes triage from a server log a lot easier:

```go
return nil, fmt.Errorf("DV file %s: manifest record_count %d disagrees with puffin cardinality property %d",
    dvFile.FilePath(), manifestCardinality, blobCardinality)
```

laskoviymishka (changes requested):
The cross-source validation direction looks right, and the disagreement test is exactly the regression pin I’d want here. Almost there, with one blocker before merge:

`hasManifestCardinality := manifestCardinality > 0` mixes up “manifest didn’t tell us” with “manifest said zero.” `record_count` is non-nullable in the spec, and Java writes `0` for empty DVs. So a real `blob=5 / manifest=0` mismatch would silently fall back to “blob-only wins”, which is the opposite of what this PR is trying to catch.

Since `RecordCount` is a non-pointer `int64`, I’d treat it as always present and let zero go through the normal agree / disagree path.

Related: the “missing cardinality property is tolerated with a warning” test doesn’t hit that branch anymore. `writePuffinWithDVBlob` still writes `"cardinality": "5"` and the test passes `count=5`, so it’s just the agree path. Worth adding a sibling helper that omits the property entirely.

Small cleanup while you’re there:

* rename the local `blobCardinality`; it shadows the package-level function
* include the DV file path in the new disagreement error, like the rest of `ReadDV` does

table/dv/deletion_vector.go
zeroshade, line 181: This treats `record_count == 0` as if the manifest cardinality were absent. For an `iceberg.DataFile`, manifest `record_count` is a required non-nullable long, so zero is a valid expected cardinality and still needs to participate in both checks. As written, manifest 0 + Puffin cardinality 5 + decoded bitmap cardinality 5 falls into the blob-only path and is accepted instead of rejected. Please treat the manifest cardinality as present for data files, compare it to the Puffin cardinality whenever the property exists, and validate the decoded bitmap against it, including zero. A regression test for manifest 0 / Puffin 5 / bitmap 5 should error.

table/dv/deletion_vector_test.go
```diff
@@ -413,18 +413,29 @@ func TestReadDVCardinalityValidation(t *testing.T) {
...
+		dir := t.TempDir()
+		path, meta := writePuffinWithDVBlobAndProps(t, dir, dvBlobBytes, map[string]string{
+			"referenced-data-file": "s3://bucket/data/data-001.parquet",
+			// Both metadata sources agree, but the bitmap actually has 5 positions.
+			"cardinality": "99",
+		})
+		offset, size := meta.Offset, meta.Length
+		_, err := ReadDV(iceio.LocalFS{}, newDVTestFile(path, 99, &offset, &size))
+		require.Error(t, err)
 		assert.Contains(t, err.Error(), "cardinality mismatch")
```
zeroshade, line 439: Nearby real line 442 still claims to cover the missing-cardinality-property path, but it calls `writePuffinWithDVBlob`, which writes `cardinality: 5`, so the test exercises the normal cardinality path instead of missing-property tolerance. Please add a fixture/helper that actually omits the cardinality property, and include a zero-cardinality case so the manifest-zero behavior is covered explicitly.

table/dv/deletion_vector.go
zeroshade, line 187: Two polish items while updating this block: line 176 shadows the package-level `blobCardinality` helper, so renaming the local to something like `puffinCardinality` would avoid confusion; and this mismatch error should include the DV file path, consistent with the neighboring `ReadDV` errors, to make corrupt-file diagnostics actionable.

zeroshade (changes requested):
Thanks for moving this toward cross-checking the manifest and Puffin cardinalities; that is the right direction. However, the central manifest check is currently bypassed when record_count is zero even though zero is a valid required manifest cardinality. That leaves a corrupt zero-manifest/nonzero-Puffin case accepted, so I think this needs changes before merge.

zeroshade:

PR #1128, opened 2026-05-24
fix(table): validate scan filters before empty planning returns

### LINKED ISSUE #1119: table: PlanFiles skips rowFilter validation on empty planning paths
`Scan.PlanFiles()` validates malformed `rowFilter` expressions inconsistently.

If planning exits before any evaluator is constructed, invalid filters can be accepted without error.

### Code path

```text
PlanFiles
  -> fetchPartitionSpecFilteredManifests
     -> if scan.Snapshot() == nil { return nil, nil }
     -> manifestEvaluators := newKeyDefaultMapWrapErr(scan.buildManifestEvaluator)
        -> buildManifestEvaluator
           -> partitionFilters.Get(specID)
              -> buildPartitionProjection
                 -> newInclusiveProjection(...).Project(rowFilter)
                    -> BindExpr(...)
  -> if len(manifestList) == 0 { return nil, nil }
  -> collectManifestEntries
     -> newInclusiveMetricsEvaluator(..., scan.rowFilter, ...)
        -> BindExpr(...)
     -> partitionEvaluators := newKeyDefaultMapWrapErr(scan.buildPartitionEvaluator)
        -> buildPartitionEvaluator
           -> ExpressionEvaluator(... partitionFilters.Get(specID) ...)
```

Observed empty-planning cases:

- no current snapshot: `fetchPartitionSpecFilteredManifests()` returns `nil, nil`
- current snapshot with an empty manifest list: `PlanFiles()` returns early on `len(manifestList) == 0`

In both cases, the usual filter-binding paths are never reached.

Relevant code:

- `table/scanner.go:76`
- `table/scanner.go:286`
- `table/scanner.go:300`
- `table/scanner.go:314`
- `table/scanner.go:466`
- `table/scanner.go:500`
- `table/scanner.go:517`
- `table/scanner.go:567`
- `table/scanner.go:675`
- `table/evaluators.go:41`
- `table/evaluators.go:573`
- `table/evaluators.go:676`

### Current behavior

For a malformed scan filter:

- on an empty planning path, `PlanFiles()` can return no tasks and no error
- on a non-empty planning path, validation is reached once evaluator creation runs
- because the lazy evaluator caches use `newKeyDefaultMapWrapErr(...)`, those binding errors may currently surface via `panic` instead of a normal returned error

### Expected behavior

`PlanFiles()` should reject malformed `rowFilter` expressions regardless of whether planning is empty.

In practice, validation likely needs to happen before the early return for empty planning paths.

### Scope

This looks like a validation-timing issue in `PlanFiles()`, not a case-sensitivity issue.

`ReadTasks()` appears narrower here because it binds `scan.rowFilter` eagerly before reading.

officialasishkumar:
Summary

- validate scan row filters before planning can return early for missing snapshots or empty manifest lists
- add coverage for invalid filters on both empty planning paths

Fixes #1119

Testing

- go test ./table -run "TestPlanFilesValidatesRowFilter|TestBuild.*Evaluator" -count=1
- go test ./table -count=1

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -362,6 +362,17 @@ func (scan *Scan) buildPartitionProjection(specID int) (iceberg.BooleanExpressio
 	return buildPartitionProjection(specID, scan.metadata, scan.rowFilter, scan.caseSensitive)
 }
 
+func (scan *Scan) validateRowFilter() error {
+	_, err := newInclusiveMetricsEvaluator(
+		scan.metadata.CurrentSchema(),
+		scan.rowFilter,
+		scan.caseSensitive,
+		scan.options["include_empty_files"] == "true",
+	)
+
+	return err
+}
+
 func buildPartitionProjection(specID int, meta Metadata, rowFilter iceberg.BooleanExpression, caseSensitive bool) (iceberg.BooleanExpression, error) {
 	spec := meta.PartitionSpecByID(specID)
 	if spec == nil {
@@ -657,6 +668,10 @@ func (scan *Scan) PlanFiles(ctx context.Context) ([]FileScanTask, error) {
 		scan.snapshotID = &snapshot.SnapshotID
 		scan.asOfTimestamp = nil
 	}
+	if err := scan.validateRowFilter(); err != nil {
+		return nil, err
+	}
+
 	// Step 1: Retrieve filtered manifests based on snapshot and partition specs.
 	manifestList, err := scan.fetchPartitionSpecFilteredManifests(ctx)
 	if err != nil || len(manifestList) == 0 {
--- a/table/scanner_internal_test.go
+++ b/table/scanner_internal_test.go
@@ -18,6 +18,7 @@
 package table
 
 import (
+	"context"
 	"runtime"
 	"sync"
 	"sync/atomic"
@@ -28,6 +29,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/compute"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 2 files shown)


tanmayrauth (approved):
Thanks for the fix.

 Two test-coverage suggestions:

  - Add a test for the `asOfTimestamp` → empty-manifest path. The ordering (`asOfTimestamp` resolution before `validateRowFilter`) isn't asserted anywhere, so a future refactor could silently regress it.
  - The new tests construct `Scan{}` via struct literals, which bypasses defaults set by `metadata.Scan(...)`. One end-to-end test through `Table.Scan(...).PlanFiles(...)` would lock in the public-API
  contract.

  LGTM on the fix itself.

table/scanner.go
```diff
@@ -657,6 +668,10 @@ func (scan *Scan) PlanFiles(ctx context.Context) ([]FileScanTask, error) {
 		scan.snapshotID = &snapshot.SnapshotID
 		scan.asOfTimestamp = nil
 	}
+	if err := scan.validateRowFilter(); err != nil {
```
laskoviymishka, line 671: I'd think hard about this placement. Java's `SnapshotScan.planFiles` returns `CloseableIterable.empty()` when the snapshot is nil, PyIceberg's `scan_plan_helper` returns `iter([])`, and iceberg-rust does the same — none of them bind the filter on the empty-table path. After this PR we're the only client that hard-errors there.

The fix to the gap is real, but I'd either gate this on `scan.Snapshot() != nil && len(manifestList) > 0` (so we only validate when we'd actually have used the filter), or make the divergence an explicit, documented choice rather than a side effect of validation placement. wdyt?

laskoviymishka (changes requested):
One thing I'd want resolved before merge though: Java, PyIceberg, and iceberg-rust all return an empty iterable when there's no snapshot or no manifests, without binding the filter, so this PR makes iceberg-go the only client that hard-errors on an empty-table scan with an invalid filter. I'd either gate the validation on scan.Snapshot() != nil && len(manifestList) > 0, or make the divergence a conscious, documented choice. The validation primitive itself (newInclusiveMetricsEvaluator with include_empty_files) is also the wrong vehicle — iceberg.BindExpr directly would be clearer and avoid double-binding on the hot path. Left the rest inline.                                                                                                      

Once that's settled, happy to take another pass.

zeroshade:
> Java, PyIceberg, and iceberg-rust all return an empty iterable when there's no snapshot or no manifests, without binding the filter, so this PR makes iceberg-go the only client that hard-errors on an empty-table scan with an invalid filter. 

Given that we should definitely follow and match that behavior rather than decide to hard error

table/scanner.go
```diff
@@ -362,6 +362,17 @@ func (scan *Scan) buildPartitionProjection(specID int) (iceberg.BooleanExpressio
 	return buildPartitionProjection(specID, scan.metadata, scan.rowFilter, scan.caseSensitive)
 }
 
+func (scan *Scan) validateRowFilter() error {
+	_, err := newInclusiveMetricsEvaluator(
+		scan.metadata.CurrentSchema(),
```
zeroshade, line 367: Minor: this constructs a full `newInclusiveMetricsEvaluator` only to check whether the filter binds, then discards it. Since non-empty planning builds the real evaluator later anyway, please prefer validating directly with `RewriteNotExpr` plus `BindExpr(scan.metadata.CurrentSchema(), ...)` instead of allocating an evaluator here.

table/scanner.go
```diff
@@ -657,6 +668,10 @@ func (scan *Scan) PlanFiles(ctx context.Context) ([]FileScanTask, error) {
 		scan.snapshotID = &snapshot.SnapshotID
 		scan.asOfTimestamp = nil
 	}
+	if err := scan.validateRowFilter(); err != nil {
+		return nil, err
```
zeroshade, line 672: Question: this placement intentionally validates filters before the empty-scan paths, which fixes #1119 for iceberg-go. It does diverge from Java/PyIceberg/iceberg-rust behavior described in review, where empty-scan paths can return empty results without binding filters. Please confirm the intended cross-client behavior here: accept the stricter validation in iceberg-go, or gate this validation so it only runs when a snapshot/manifests exist.

zeroshade (changes requested):
Thanks for the fix here. The core behavioral change looks right for the reported iceberg-go issue: invalid row filters should surface an error even when planning would otherwise return an empty scan.

Before this can merge, the PR also needs the branch state resolved. I checked GitHub twice and it currently reports `mergeable=UNKNOWN` / `mergeStateStatus=UNKNOWN`; the earlier stale/conflicting state was `CONFLICTING`/`DIRTY`, so please rebase or otherwise refresh the branch until GitHub reports a clean mergeable state.

I also left two inline notes: one minor implementation simplification, and one question to confirm the intended cross-client behavior divergence.

zeroshade:

PR #1129, opened 2026-05-24
fix(manifest): encode decimal partitions with declared precision

### LINKED ISSUE #1027: Bug(manifest): convertDecimalValue uses string length instead of declared precision for Avro fixed-size encoding
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

**Component**: `manifest.go`
**Affected version**: `main` (v0.5.0)
**Severity**: Data corruption / Avro encode error for decimal partition columns

---

`convertDecimalValue` computes the Avro `fixed[N]` byte size by passing `len(dec.String())` to `internal.DecimalRequiredBytes`, but that function expects the column's declared **precision**. These two quantities are unrelated, causing the wrong number of bytes to be allocated and the Avro encoder to reject the value.

---

## Affected code

`manifest.go`, `convertDecimalValue`:

```go
func convertDecimalValue(v any) any {
    if dec, ok := v.(Decimal); ok {
        fixedSize := internal.DecimalRequiredBytes(len(dec.String()))  // ← BUG
        bytes, err := DecimalLiteral(dec).MarshalBinary()
        ...
        return padOrTruncateBytes(bytes, fixedSize)
    }
    return v
}
```

`internal.DecimalRequiredBytes` signature (`internal/avro_schemas.go`):

```go
// DecimalRequiredBytes returns the required number of bytes to store a
// decimal value with the given precision.
func DecimalRequiredBytes(precision int) int { ... }
```

---

## Root cause

`dec.String()` returns the human-readable decimal string (e.g. `"1.00"`).
Its character length is not the declared precision of the column — it varies
with the specific value being encoded.

| Column type   | Value   | `dec.String()` | `len(...)` | `DecimalRequiredBytes(len)` | Correct `DecimalRequiredBytes(precision)` |
|---------------|---------|----------------|------------|-----------------------------|-------------------------------------------|
| `decimal(10,2)` | `10.00` | `"10.00"` | 5 | 3 bytes | 5 bytes |
| `decimal(10,2)` | `1.00`  | `"1.00"`  | 4 | 2 bytes | 5 bytes ← **wrong** |
| `decimal(18,0)` | `1`     | `"1"`     | 1 | 1 byte  | 8 bytes ← **wrong** |

The `decimal(10,2)` / `10.00` case produces the right size by coincidence
(`len("10.00") == 5 == DecimalRequiredBytes(10)`), masking the bug for that
specific value.

---

## Symptom

Any write of a manifest with a decimal partition column where the value's string representation length differs from the column's precision produces:

```
avro: field data_file.partition.<field>: cannot use []uint8 with Avro type fixed
```

---

## Reproduction

1. Create a table with partition spec `identity(price)` where `price` is `decimal(10, 2)`.
2. Write a `RowDelta` that adds a data file with partition value `1.00`
   (string length 4, but declared precision is 10).
3. Observe the Avro encode error when the manifest is flushed.

---

## Proposed fix

Thread the column's declared precision from the `DecimalType` of the partition field through to `convertDecimalValue`:

```go
// Pass precision from the field's type, not from the value's string length.
func convertDecimalValue(v any, precision int) any {
    if dec, ok := v.(Decimal); ok {
        fixedSize := internal.DecimalRequiredBytes(precision)
        bytes, err := DecimalLiteral(dec).MarshalBinary()
        if err != nil {
            return v
        }

        return padOrTruncateBytes(bytes, fixedSize)
    }

    return v
}
```

The call site in the partition value serialization loop already has access to the `PartitionField` → `DecimalType` → `Precision`.

officialasishkumar:
Summary

- track decimal partition precision separately from decimal scale while building manifest field metadata
- use the declared Avro decimal precision to determine fixed-size encoding width
- add regression coverage for a decimal(10, 2) partition value whose textual length is smaller than the required fixed size

Fixes #1027

Testing

- go test . -run TestManifest/TestWriteManifestDecimalPartitionUsesDeclaredPrecision -count=1
- go test . ./codec -count=1

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -376,7 +376,7 @@ func (m *manifestFile) FetchEntries(fs iceio.IO, discardDeleted bool) (_ []Manif
 	return ReadManifest(m, f, discardDeleted)
 }
 
-func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int) {
+func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int, map[int]int) {
 	root := sc.Root()
 	getField := func(node avro.SchemaNode, name string) *avro.SchemaField {
 		for i := range node.Fields {
@@ -390,7 +390,8 @@ func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int
 
 	result := make(map[string]int)
 	logicalTypes := make(map[int]string)
-	fixedSizes := make(map[int]int)
+	decimalScales := make(map[int]int)
+	decimalPrecisions := make(map[int]int)
 
 	entryField := getField(root, "data_file")
 	partitionField := getField(entryField.Type, "partition")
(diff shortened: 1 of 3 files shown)


manifest.go
```diff
@@ -414,18 +415,19 @@ func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int
...
 			if typ.LogicalType == atype.Decimal {
-				fixedSizes[fid] = typ.Scale
+				decimalScales[fid] = typ.Scale
+				decimalPrecisions[fid] = typ.Precision
 			}
 		}
 	}
 
-	return result, logicalTypes, fixedSizes
+	return result, logicalTypes, decimalScales, decimalPrecisions
```
zeroshade, line 424: can we combine the `decimalScales` and `decimalPrecisions` into a single that maps the fieldID to a combined scale and precision? Compacting that would be better than having two separate maps.

zeroshade (changes requested):

zeroshade:

PR #1130, opened 2026-05-26
fix(substrait): emit decimal literals as 16-byte little-endian

laskoviymishka:
DecimalLiteral.MarshalBinary returns big-endian min-byte; Substrait's expr.NewLiteral needs 16-byte little-endian and was silently returning nil for everything else, panicking downstream. Convert before emit and add whitebox tests covering small/zero/16-byte-negative values.

### DIFF
--- a/table/substrait/substrait.go
+++ b/table/substrait/substrait.go
@@ -19,8 +19,10 @@ package substrait
 
 import (
 	_ "embed"
+	"encoding/binary"
 	"fmt"
 	"strings"
+	"unsafe"
 
 	"github.com/apache/arrow-go/v18/arrow/compute/exprs"
 	"github.com/apache/iceberg-go"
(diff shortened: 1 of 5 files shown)


table/substrait/substrait.go
```diff
@@ -273,13 +273,45 @@ func toDecimalLiteral(typ iceberg.Type, v iceberg.DecimalLiteral) expr.Literal {
 	byts, _ := v.MarshalBinary()
 	result, _ := expr.NewLiteral(&types.Decimal{
 		Scale:     int32(v.Scale),
-		Value:     byts,
+		Value:     decimalBytesBEMinToLE16(byts),
 		Precision: int32(precision),
 	}, false)
```
zeroshade: Instead of doing the `MarshalBinary`, if we're running on a LE machine, then you can just do `unsafe.Slice((*byte)(unsafe.Pointer(&v.Value().Val)), 16)` or something to that end. 

`DecimalLiteral.Value()` returns a `Decimal` and the `.Val` member is an `Decimal128` which is represented by a struct consisting of two uint64s in two's complement. So treating the struct as just the 16 bytes that make it up would be the full 16 byte, little-endian, 2's complement representation. Nice and simple :smile:

For big-endian, we just reverse the bytes, might be simpler than the function you created below.

zeroshade (changes requested):

table/substrait/substrait.go
laskoviymishka: fixed

table/substrait/substrait.go
```diff
@@ -258,6 +260,15 @@ func toByteSliceSubstraitLiteral[T []byte | types.UUID](v T) expr.Literal {
...
 }
 
+// hostIsLittleEndian reports whether decimal128.Num's in-memory layout already
+// matches the Substrait 16-byte little-endian two's-complement encoding. True
+// on x86, arm64, and other LE architectures; false on s390x and other BE.
+var hostIsLittleEndian = func() bool {
+	var probe uint16 = 1
+
+	return *(*byte)(unsafe.Pointer(&probe)) == 1
+}()
```
zeroshade: why not just use a separate file and have `//go:build armbe || arm64be || m68k || mips || mips64 || mips64p32 || ppc || ppc64 || s390 || s390x || shbe || sparc || sparc64` with `const hostIsLE = false` and then a complementary file with `//go:build 386 || amd64 || amd64p32 || alpha || arm || arm64 || loong64 || mipsle || mips64le || mips64p32le || nios2 || ppc64le || riscv || riscv64 || sh || wasm` and `const hostIsLE = true`?

zeroshade (changes requested):

table/substrait/substrait.go
laskoviymishka: want to avoid extra file :)
but yeah, this make sense

laskoviymishka: fixed

zeroshade (approved):

PR #1131, opened 2026-05-26
test(table): add cross-client DV format compliance tests

tanmayrauth:
Validates that Go's DVWriter output conforms to the spec envelope format (length + magic + bitmap + CRC32) and produces Puffin files with correct blob metadata (type, snapshot-id=-1, sequence-number=-1, cardinality, referenced-data-file). Asserts byte equivalence between Go's SerializeDV output and a Java-produced DV fixture for the same input positions, and that multi-blob Puffin files have correct independent offsets.

Part 3 of #997.


table/dv/dv_cross_client_test.go
```diff
@@ -0,0 +1,157 @@
...
+// output conforms to the DV spec envelope format and can be parsed by
+// any spec-compliant reader (Java, pyiceberg, iceberg-rust).
+//
+// Format verified:
+//   - Puffin file structure (magic + blobs + footer)
+//   - Blob type is "deletion-vector-v1"
+//   - snapshot-id = -1, sequence-number = -1
+//   - Blob properties include "cardinality" and "referenced-data-file"
+//   - Inner DV envelope: [length][magic 0xD1D33964][bitmap][CRC32]
+func TestCrossClientDVWriterProducesSpecCompliantBlob(t *testing.T) {
```
laskoviymishka: This is Go-writes-Go-reads — `NewDVWriter` produces the file, `puffin.NewReader` + `DeserializeDV` reads it back, and the cardinality is checked against `df.Count()` which is itself Go-derived. There's no Java/PyIceberg/Rust artifact in the loop, so a Puffin-layer regression in any of those clients wouldn't be caught. `TestDVWriterSingleDataFile` + `verifyDVReadBack` in `dv_writer_test.go` already covers the roundtrip.

What this test could be doing — and what would actually earn the "cross-client" name — is reading a Java-produced single-blob Puffin file from `testdata/` via `ReadDV`, then asserting on cardinality, referenced-data-file, and bitmap contents. The testdata directory has Java-produced DV envelopes (`.bin`) but no Java-produced Puffin file; checking one in here is the unlock.

table/dv/dv_cross_client_test.go
```diff
@@ -0,0 +1,157 @@
...
+		assert.True(t, bm.Contains(uint64(pos)), "position %d should be set", pos)
+	}
+}
+
+// TestCrossClientGoSerializeMatchesJavaFormat verifies that Go's SerializeDV
+// produces byte-identical output to a Java-produced DV fixture for the same
+// input positions. Byte equivalence is a stronger cross-client guarantee than
+// roundtrip-readability: any spec-compliant reader will accept Go's output,
+// and Go's output is indistinguishable from Java's at the wire level.
+func TestCrossClientGoSerializeMatchesJavaFormat(t *testing.T) {
```
laskoviymishka: This is the only test that compares against an external artifact, so it's pulling a lot of weight — but the artifact is the inner DV envelope, not a Puffin file, and the fixture `[1,3,5,7,9]` is exactly the case that's easiest to byte-match: scattered low-cardinality positions, single roaring array container, no run encoding triggered on either side.

That makes the comment overstate the guarantee in two ways. First, it's silent on the Puffin layer (where most cross-client bugs live). Second, byte-equality doesn't generalize — Java's `BitmapPositionDeleteIndex.serialize()` calls `runLengthEncode()` before writing and Go's `RoaringPositionBitmap.Serialize()` doesn't, so for range-style deletes the two outputs diverge bytewise while both remaining spec-compliant.

I'd scope this test to what it actually proves (single-container DV-envelope parity for this fixture) and add coverage at two new axes:

- A Java-produced **Puffin** fixture (not just an inner envelope) read end-to-end through `ReadDV`.
- At least one DV-envelope fixture that crosses the 4096-entry array→bitmap threshold, and ideally one >2^32 to force a second roaring key — the container/key boundaries are where encoding differences actually surface.

table/dv/dv_cross_client_test.go
```diff
@@ -0,0 +1,157 @@
...
+	require.NoError(t, err)
+
+	assert.Equal(t, javaBytes, goBytes,
+		"Go-serialized DV must be byte-identical to Java fixture for the same positions")
+}
+
+// TestCrossClientMultiFileDVPuffin validates that a multi-blob Puffin file
+// (one blob per data file) has correct offsets and each blob is independently
+// readable — matching how Java writes multi-DV Puffin files.
+func TestCrossClientMultiFileDVPuffin(t *testing.T) {
```
laskoviymishka: Same shape as the comment on `TestCrossClientDVWriterProducesSpecCompliantBlob` — Go writes the multi-blob Puffin, Go reads it back, asserts cardinality against `df.Count()`. `TestDVWriterMultipleDataFiles` already covers this. To actually be a cross-client test, this should consume a Java-produced multi-blob Puffin from `testdata/` and check that `ReadDV` recovers the right bitmap for each `DataFile`. The Puffin footer layout for multi-blob files is exactly the place where small encoding choices (offset encoding, blob-meta ordering) cause silent corruption across clients.

laskoviymishka (changes requested):
The structural assertions are fine, envelope layout, blob metadata, multi-blob offsets all check what they should. 

But for a PR titled "cross-client DV format compliance tests," I'd expect the centerpiece to be **golden-file** comparison against **Java**-produced **Puffin** artifacts, and that's missing 😅 . 

Only one of the three tests touches a non-Go artifact, and that artifact is a raw DV envelope (the inner bytes) rather than a Puffin file. 

The Puffin layer — footer offset encoding, magic, compression flag, blob property casing — is where cross-client format bugs actually show up, and nothing here exercises it against a real foreign-produced file.

tanmayrauth:
@laskoviymishka — fair points across the board. Honestly I had some doubts about the scope of "cross-client" before I opened this and should have clarified up front rather than letting you spell it out in  review. Sorry for the round trip. Four things I want to lock down so the rewrite lands in one pass:
  
  1. Disposition of the three current tests. My read: drop TestCrossClientDVWriterProducesSpecCompliantBlob and TestCrossClientMultiFileDVPuffin — they're Go-writes-Go-reads and already covered by TestDVWriterSingleDataFile / TestDVWriterMultipleDataFiles — and replace them with two tests that read Java-produced Puffin fixtures (single-blob and multi-blob) end-to-end through ReadDV. For TestCrossClientGoSerializeMatchesJavaFormat, scope-down + rename rather than delete — keep it narrowly as "Go envelope matches Java for this single-array-container fixture" with a comment that explicitly calls out the RLE divergence. Or would you rather see Test 2 gone too in favour of a single Java-Puffin readback?
  2. Java generator location. I have a one-off GenerateDVFixtures.java + generate.sh that produces the .puffin bytes — it has to live inside an apache/iceberg checkout to compile against the package-private BitmapPositionDeleteIndex. Two options: commit it under dev/dv-fixtures/ for reproducibility, or keep it out and ship just the .puffin bytes + regen instructions in testdata/deletes/README.md. Preference?
  3. Envelope-fixture boundaries. Existing 64maphighvals.bin covers the >2^32 second-roaring-key case. Do you also want a fresh fixture crossing the 4096-entry array→bitmap container threshold, or do 64mapspreadvals.bin / 64map32bitvals.bin already cover mixed-container cases enough for your purposes?
  4. Boundary-crossing assertions. I'm planning read-back-and-assert-positions on the new fixtures rather than byte-equality, since BitmapPositionDeleteIndex.serialize()'s runLengthEncode() step makes byte-match brittle for any range-style fixture. Sound right?

laskoviymishka:
1. yes
2. commit it under dev/dv-fixtures/ for reproducibility -- is good enough here
3. i think it's fine already
4. yes

tanmayrauth:
@laskoviymishka  I've made the below change PTAL when you have a moment.

  - Dropped the two Go-writes-Go-reads tests — TestDVWriterSingleDataFile / TestDVWriterMultipleDataFiles already cover that path.
  - Added TestCrossClientReadJavaSingleBlobDV and TestCrossClientReadJavaMultiBlobDV — both read Java-produced .puffin artifacts end-to-end through ReadDV.
  - Scoped down the byte-equality test to a single-array-container fixture and renamed it accordingly; comment now calls out the Java RLE step as the reason byte-match doesn't generalize.
  - Added a high-key rejection test mirroring Java's testDeserializeUnsupportedRoaringExample so both clients fail-closed on the same spec-illegal input.

laskoviymishka (approved):
🚢 👍

PR #1134, opened 2026-05-27
test(table): end-to-end deletion-vector scan over a real Parquet file

### LINKED ISSUE #1052: test(table): end-to-end deletion-vector scan over a real Parquet data file

laskoviymishka:
Closes #1052.

End-to-end test through the public `ReadTasks` API for a real on-disk
Parquet file. Locks that DV-deleted positions actually disappear from
scan output via the composition through `recordBatchesFromTasksAndDeletes`.

Seven cases:

- no-DV baseline
- interior / boundary / all-rows deletes
- empty bitmap
- two files with asymmetric ranges + DV positions so a bitmap-swap
  aggregates differently
- stale out-of-range positions

Verified by replacing `filterByDeletionVector` with a passthrough and
observing predicted mismatches.

Sits alongside cross-client fixture work in https://github.com/apache/iceberg-go/pull/1041 — this pins Go-side
composition semantics for the scan path, https://github.com/apache/iceberg-go/pull/1041 pins the wire format
against externally-produced fixtures.

### DIFF
--- a/table/dv_scanner_end_to_end_test.go
+++ b/table/dv_scanner_end_to_end_test.go
@@ -159,11 +159,23 @@ func collectDVScanRows(t *testing.T, scan *Scan, tasks []FileScanTask) []int64 {
 // Empirically verified to catch a regression: temporarily replacing the
 // filterByDeletionVector body with a passthrough makes cases 2/3/4/6 fail
 // with surviving-row mismatches. See PR description for details.
+// newDVScanTestEnv allocates a LocalFS-backed scratch dir and a *Table pointed
+// at it. Hoisted out of the subtests so each case states its DV-specific
+// fixture (data file path, bitmap, expected survivors) and nothing else.
+// Returns LocalFS by concrete type so callers can pass it to both the
+// read-side (Table.Scan -> iceio.IO) and write-side (writeIntParquetWithFieldID
+// -> iceio.WriteFileIO) helpers without a re-type-assert.
+func newDVScanTestEnv(t *testing.T) (iceio.LocalFS, string, *Table) {
+	t.Helper()
+	fs := iceio.LocalFS{}
+	tmp := t.TempDir()
+
+	return fs, tmp, buildDVScanTestTable(t, fs, tmp)
+}
+
 func TestDVScanEndToEnd(t *testing.T) {
 	t.Run("no DV — all rows survive", func(t *testing.T) {
-		fs := iceio.LocalFS{}
-		tmp := t.TempDir()
-		tbl := buildDVScanTestTable(t, fs, tmp)
+		fs, tmp, tbl := newDVScanTestEnv(t)
 
 		df := writeIntParquetWithFieldID(t, fs, filepath.Join(tmp, "data-1.parquet"), 0, 10)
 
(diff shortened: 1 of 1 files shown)


table/dv_scanner_end_to_end_test.go
```diff
@@ -0,0 +1,309 @@
...
+		df := writeIntParquetWithFieldID(t, fs, dataPath, 0, 5)
+		puffinPath, offset, length := writeDVPuffinFixture(t, nil, dataPath)
+		dvFile := newDVMockDataFile(puffinPath, dataPath, offset, length)
+
+		got := collectDVScanRows(t, tbl.Scan(), []FileScanTask{{
+			File:                df,
+			DeletionVectorFiles: []iceberg.DataFile{dvFile},
+		}})
+		assert.Equal(t, []int64{0, 1, 2, 3, 4}, got,
+			"empty DV must be a no-op, not skip the file")
```
zeroshade, line 253: Technically this isn't actually exercising `filterByDeletionVector`. Empty bitmaps will short-circuit *before* the filter is appended to the pipeline. Can we clarify in a comment that the no-op is happening via the `IsEmpty` short-circuit?

table/dv_scanner_end_to_end_test.go
```diff
@@ -0,0 +1,309 @@
...
+// nextIdx counter directly), and cross-client byte fixtures (#1041).
+//
+// Empirically verified to catch a regression: temporarily replacing the
+// filterByDeletionVector body with a passthrough makes cases 2/3/4/6 fail
+// with surviving-row mismatches. See PR description for details.
+func TestDVScanEndToEnd(t *testing.T) {
+	t.Run("no DV — all rows survive", func(t *testing.T) {
+		fs := iceio.LocalFS{}
+		tmp := t.TempDir()
+		tbl := buildDVScanTestTable(t, fs, tmp)
```
zeroshade: can we hoist this out of the tests into a helper or something instead rather than duplicating this on every tests? 

Would it make sense to make this a table-based test instead to reduce the duplication or is that not really feasible given the test structure?

table/dv_scanner_end_to_end_test.go
```diff
@@ -0,0 +1,309 @@
...
+				newDVMockDataFile(puffinB, dataPathB, offB, lenB),
+			}},
+		}
+		got := collectDVScanRows(t, tbl.Scan(), tasks)
+		assert.Equal(t, []int64{0, 2, 4, 11, 13, 14}, got,
+			"each file's DV must apply only to that file's rows; a bitmap-swap "+
+				"between files would yield [1,3,4,10,12,14] instead")
+	})
+
+	t.Run("DV with out-of-range position is a no-op for missing rows", func(t *testing.T) {
```
zeroshade, line 292: Can we add one more test case for partial coverage? a bitmap with mixed in-range + out-of-range positions?

e.g. `{1, 100}` on a 5-row file

zeroshade (approved):
Just a few nitpicks that are mostly optional. LGTM

Issue #1135, opened 2026-05-27
feat(table): write deletion vectors for partitioned v3 tables


laskoviymishka:
iceberg-go writes deletion vectors only for v3 unpartitioned tables (#1113); partitioned v3 deletes still fall through to legacy Parquet position-delete files with a deprecation warning. The spec mandates DV on v3 regardless of partitioning. The current path costs storage (Roaring bitmaps vs Parquet position-rows), read latency (one Puffin blob seek per data file vs per-partition file opens), and leaves the deprecation warning firing on every commit.

The split is an iceberg-go implementation choice: `DVWriter.Add` doesn't yet accept partition metadata, so Flush can't propagate `(spec, partition)` to the output DV manifest entries.

Spec: https://iceberg.apache.org/spec/#deletion-vectors
Java reference: `BaseDVFileWriter` in apache/iceberg

Parent: #589.

tanmayrauth:
@laskoviymishka  can I work on this issue?

laskoviymishka:
@tanmayrauth yes! it's a natural follow up of PR you recently landed, i have some ground work done here, but after that - i can hand over this to you

PR #1136, opened 2026-05-27
feat(table/dv): accept partition spec and data on DVWriter.Add

laskoviymishka:
Part of #1135. Matches Java's BaseDVFileWriter.delete signature so DV manifest entries can carry the data file's partition record and spec id, unblocking the follow-up that lifts the unpartitioned-only gate in positionDeleteRecordsToDataFiles.

First Add for a path captures spec+partition; later Adds append positions only, mirroring Java's computeIfAbsent semantics. Both spec and partitionData are defensively copied on capture (NewPartitionSpecID for spec, maps.Clone for partitionData) so caller mutations between Add and Flush can't silently corrupt the entry, mirroring Java's StructLikeUtil.copy(partition).

### DIFF
--- a/table/dv/dv_writer.go
+++ b/table/dv/dv_writer.go
@@ -29,13 +29,17 @@ import (
 	"github.com/apache/iceberg-go/puffin"
 )
 
-// dvEntry holds the per-data-file state a DV manifest entry needs. Each entry
-// is keyed by the referenced data file path; the spec and partition record
-// come from the data file itself and are propagated unchanged onto the output
-// DV manifest entry — matching Java's BaseDVFileWriter.Deletes shape.
+// SpecResolver looks up a PartitionSpec by its id, mirroring
+// Metadata.PartitionSpecByID. Returns nil if the id is not known.
+type SpecResolver func(specID int32) *iceberg.PartitionSpec
+
+// dvEntry holds the per-data-file state a DV manifest entry needs. specID and
+// partitionData come from the data file's own manifest entry; the spec itself
+// is resolved at Flush time via the writer's SpecResolver, matching how the
+// partitioned Parquet writer consults metadata.PartitionSpecByID.
 type dvEntry struct {
 	bitmap        *RoaringPositionBitmap
-	spec          iceberg.PartitionSpec
+	specID        int32
 	partitionData map[int]any
 }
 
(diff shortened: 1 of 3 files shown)


laskoviymishka:
@tanmayrauth  -- can you please take a look on this, it's matching API to BaseDVFileWriter.delete shape, would love your review here

table/dv/dv_writer.go
```diff
@@ -21,51 +21,102 @@ import (
...
+// new spec/partitionData args. This mirrors Java's BaseDVFileWriter, which
+// stores partition metadata via computeIfAbsent on the same key. Callers must
+// not pass conflicting partition values across Adds for the same data file —
+// the writer trusts the first-Add values for the rest of the writer's life.
+func (w *DVWriter) Add(dataFilePath string, positions []int64, spec iceberg.PartitionSpec, partitionData map[int]any) {
+	entry, ok := w.entries[dataFilePath]
 	if !ok {
-		bm = NewRoaringPositionBitmap()
-		w.entries[dataFilePath] = bm
+		// Defensive copies on capture so a caller that mutates or reuses
```
tanmayrauth: Super minor doc nit, feel free to ignore: I tried mutating originalSpec.Fields()[0].SourceIDs[0] after Add and the captured spec did see the change, since PartitionField is value-copied but the inner SourceIDs []int slice still aliases. In practice this almost certainly never matters (nobody mutates SourceIDs), but if you wanted the comment to be airtight, either a slices.Clone on SourceIDs per field  or just softening the wording to something like "shallow copy — partition fields aren't mutated in this codebase" would do it.

tanmayrauth (commented):
Thanks Andrei, this looks good overall.

One thing I wanted to check: I noticed we already have partitionContext{partitionData map[int]any, specID int32} at table/arrow_utils.go:1636, populated in transaction.go:1754 and consumed by the partitioned Parquet writer at pos_delete_partitioned_fanout_writer.go:131 (which resolves the spec via metadata.PartitionSpecByID).

Just curious — did you consider taking (specID int32, partitionData map[int]any) in Add instead of the full PartitionSpec, and resolving via metadata at flush time? It would line up with how the existing partitioned writer already does this. Totally fine either way, just wanted to understand the thinking.
 
One small inline note on the defensive-copy doc, otherwise this reads well to me.

zeroshade (approved):

Issue #1139, opened 2026-05-29
feat(test): add Spark 4.0 to integration test matrix


nssalian:
### Feature Request / Improvement

Iceberg's variant type requires Spark 4.0. The integration test matrix only runs Spark 3.5, so variant cannot be tested against Spark in CI. Add Spark 4.0 to the matrix.

Spark 3.5 must continue to run on every PR.

## Follow-up tests

- [x] Add Spark 4 read tests for iceberg-go variant writes and unknown

Ref: [Iceberg spec, Version 3](https://iceberg.apache.org/spec/#version-3)

PR #1140, opened 2026-05-29
feat(test): add Spark 4.0 integration test infrastructure

nssalian:
First part of #1139 

## Summary

Adds Spark 4.0 alongside Spark 3.5 in the integration test matrix so v3 features can be validated cross-engine in CI. 

## What's in

- `Dockerfile.spark4` (apache/spark:4.0.2 + iceberg-spark-runtime-4.0 + aws-bundle), with companion `spark-defaults.conf` and `entrypoint.sh`
- docker-compose parameterized via `SPARK_IMAGE` and `SPARK_DOCKERFILE`; defaults preserve 3.5 behavior
- `make integration-setup-spark4`
- Workflow matrix entry for 4.0 (`fail-fast: false`)
- 3 CTAS statements in `provision.py` rewritten as `CREATE TABLE` + `INSERT INTO` so the same provisioning runs on both versions

## What's not in

No new tests, no cross-engine v3 feature validation. Follow-up PR.

## Validation

Locally ran on both Spark versions:
- `make integration-setup && make integration-test`
- `make integration-setup-spark4 && make integration-test`

### DIFF
--- a/.github/workflows/go-integration.yml
+++ b/.github/workflows/go-integration.yml
@@ -33,6 +33,12 @@ permissions:
 jobs:
   integration-test:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        spark:
+          - { version: "3.5", setup: "integration-setup" }
+          - { version: "4.0", setup: "integration-setup-spark4" }
     steps:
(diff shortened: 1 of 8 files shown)


nssalian:
@zeroshade @laskoviymishka PTAL when you can

zeroshade (approved):

PR #950, opened 2026-04-28
feat(io/s3): support connect timeout property

mizoz:
Fixes #941.

## Summary

- Remove `s3.connect-timeout` from the unsupported S3 property path.
- Parse `s3.connect-timeout` as a Go duration and apply it to the AWS SDK buildable client's dialer timeout.
- Preserve existing proxy support by sharing the same buildable HTTP client when both proxy and connect timeout are configured.
- Add focused tests for valid and invalid connect timeout values.

## Testing

- `go test ./io/gocloud`
- `git diff --check && go test ./...`

PR #1141, opened 2026-05-30
feat(io/s3): support connect timeout property

### LINKED ISSUE #941: feat: support s3.connect-timeout property
Parent: #471

`io/gocloud/s3.go:43` keeps `io.S3ConnectTimeout` in the `unsupportedS3Props` list and rejects it outright. Java and pyiceberg both honor `s3.connect-timeout`. The fix is to parse the property as a duration and set it on the AWS SDK HTTP client (`aws.Config.HTTPClient`) when constructing the S3 IO, then drop the property from the unsupported list. PR #744 unblocked `s3.signer.uri` the same way and is a good template.

Revanth14:
Fixes #941

## Summary
- Support the `s3.connect-timeout` property for S3 clients.
- Accept PyIceberg-compatible numeric seconds such as `60` and `60.0`,  plus Go duration strings such as `5s`.
- Apply the value to the AWS SDK buildable client's dialer timeout.
- Compose connect timeout with `s3.proxy-uri` by sharing the same  buildable HTTP client.
- Reject zero and negative timeout values with a clear error.

Supersedes #950 and addresses all three items from laskoviymishka's review: float-seconds parsing, zero/negative rejection, and proxy+timeout composition test.

## Testing
- `go test ./io/gocloud/...`

## Release note
S3 now supports the `s3.connect-timeout` property, accepting numeric seconds like `60`/`60.0` and Go duration strings like `5s`.

### DIFF
--- a/io/gocloud/s3.go
+++ b/io/gocloud/s3.go
@@ -21,10 +21,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
-	"slices"
 	"strconv"
 	"time"
 
@@ -40,19 +40,8 @@ import (
 	"gocloud.dev/blob/s3blob"
 )
 
-var unsupportedS3Props = []string{
-	io.S3ConnectTimeout,
-}
-
 // ParseAWSConfig parses S3 properties and returns a configuration.
 func ParseAWSConfig(ctx context.Context, props map[string]string) (*aws.Config, error) {
-	// If any unsupported properties are set, return an error.
-	for k := range props {
-		if slices.Contains(unsupportedS3Props, k) {
-			return nil, fmt.Errorf("unsupported S3 property %q", k)
-		}
-	}
-
 	// Remote S3 request signing is not implemented yet.
 	if v, ok := props[io.S3RemoteSigningEnabled]; ok {
 		if enabled, err := strconv.ParseBool(v); err == nil && enabled {
@@ -61,6 +50,7 @@ func ParseAWSConfig(ctx context.Context, props map[string]string) (*aws.Config, 
 	}
 
 	opts := []func(*config.LoadOptions) error{}
+	var httpClient *awshttp.BuildableClient
 
 	if tok, ok := props["token"]; ok {
 		opts = append(opts, config.WithBearerAuthTokenProvider(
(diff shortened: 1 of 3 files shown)


Revanth14:
@laskoviymishka @tanmayrauth 
This should address the feedback from #950. Could you please review when you have some time?

io/gocloud/s3_test.go
```diff
@@ -80,13 +84,111 @@ func TestParseAWSConfigRemoteSigningEnabled(t *testing.T) {
...
 	_, err := ParseAWSConfig(context.Background(), map[string]string{
-		io.S3ConnectTimeout: "5000",
+		io.S3Region:         "us-east-1",
+		io.S3ConnectTimeout: "not-a-duration",
 	})
-	require.ErrorContains(t, err, "unsupported S3 property")
+	require.ErrorContains(t, err, "invalid s3.connect-timeout")
+}
+
+func TestParseAWSConfigConnectTimeout(t *testing.T) {
```
tanmayrauth, line 97: Nice tests. One small gap: "60.0" proves a decimal point is tolerated but doesn't actually exercise the fractional-seconds path — 60.0 * float64(time.Second) collapses to the same int as 60. Consider  adding a true fractional case so a future change to the float math can't silently break sub-second timeouts:
 {
  name:    "fractional seconds",
  timeout: "1.5",
  want:    1500 * time.Millisecond,
 },

io/gocloud/s3.go
```diff
@@ -103,6 +111,27 @@ func ParseAWSConfig(ctx context.Context, props map[string]string) (*aws.Config,
 	return awscfg, nil
 }
 
+func parseS3ConnectTimeout(timeout string) (time.Duration, error) {
```
tanmayrauth, line 114: Minor: the duration <= 0 check is duplicated in both branches. Easier to read with a single validation at the end:
   
  ```
  func parseS3ConnectTimeout(timeout string) (time.Duration, error) {
       var duration time.Duration
       if seconds, err := strconv.ParseFloat(timeout, 64); err == nil {
           duration = time.Duration(seconds * float64(time.Second))
       } else {
            d, err := time.ParseDuration(timeout)
            if err != nil {
                return 0, fmt.Errorf("invalid s3.connect-timeout %q: must be seconds as a number or a Go duration string", timeout)
           }
            duration = d
        }
       if duration <= 0 {
           return 0, errors.New("s3.connect-timeout must be a positive duration")
        }
        return duration, nil
    }
```

File: io/gocloud/s3.go [hunk 3]
```diff
@@ -83,14 +73,32 @@ func ParseAWSConfig(ctx context.Context, props map[string]string) (*aws.Config,
...
+	}
+
+	if timeout, ok := props[io.S3ConnectTimeout]; ok {
+		duration, err := parseS3ConnectTimeout(timeout)
+		if err != nil {
+			return nil, err
+		}
+
+		if httpClient == nil {
+			httpClient = awshttp.NewBuildableClient()
```
tanmayrauth, line 93: Not a blocker for this PR since the same pattern already exists for s3.proxy-uri, but worth flagging: createS3Bucket (lines 167-176) applies tuned transport options (MaxIdleConnsPerHost: 256, MaxIdleConns: 256, etc.) only when awscfg.HTTPClient == nil. Setting s3.connect-timeout now populates HTTPClient here, so callers fall back to the AWS SDK's awshttp.NewBuildableClient() defaults — notably  MaxIdleConnsPerHost: 10 instead of 256, which can hurt highly parallel scans. Could you open a follow-up issue to apply the same transport tuning inside ParseAWSConfig (or factor a shared  newBuildableClient() helper) so the proxy and connect-timeout paths don't silently regress connection pooling?

io/gocloud/s3_test.go
Revanth14: Good point, thanks. Added a true fractional seconds case with `1.5` expecting `1500 * time.Millisecond`.

io/gocloud/s3.go
Revanth14: Thanks, updated `parseS3ConnectTimeout` to parse into a single `duration` variable and perform the positive-duration validation once at the end.

io/gocloud/s3.go
Revanth14: Thanks for flagging this. Since the same behavior already exists for `s3.proxy-uri`, I opened a follow-up issue to track the shared transport-tuning cleanup: #1148 .

tanmayrauth (approved):
Thanks for the update, LGTM! 
Will wait for Andrei for final comments and merge.

zeroshade (approved):
LGTM!

PR #1143, opened 2026-05-31
perf(manifest): lazy per-group init for dataFile maps and preallocate avroColMapToMap

### LINKED ISSUE #1142: perf(manifest): dataFile.initializeMapData eagerly allocates all column-stats maps on first accessor call
### Description:
When profiling workloads that read manifest entries (compaction, orphan cleanup), map allocations from `dataFile.initializeMapData()` accounted for ~45% of allocated memory in our use case.
The root cause is that it uses a single `sync.Once` to convert all 7 column-stats maps plus the
partition map on the first call to any accessor — even when the caller only needs `Partition()`.

Most code paths outside of table scanning only need partition data, but currently pay the full cost of allocating and populating maps for `ColumnSizes`, `ValueCounts`, `NullValueCounts`, `NaNValueCounts`, `DistinctValueCounts`, `LowerBoundValues`, and `UpperBoundValues`.

Additionally, the underlying `avroColMapToMap` helper does not preallocate the output map, causing repeated rehashing as entries are inserted.

### Proposed fix:
  1. Split the single `sync.Once` into two — one for partition data, one for column stats — so each accessor only triggers the initialization it needs.
  2. Preallocate the map in `avroColMapToMap` using the known slice length.

paveon:
## Summary
  - Split `dataFile.initializeMapData()` into `initPartitionData()` and
    `initColumnStatsData()`, each with its own `sync.Once`, so that 
    partition-only callers (compaction, conflict validation, snapshots)
    no longer allocate column-stats maps they never read.
  - Preallocate the map in `avroColMapToMap` with `make(map[K]V, len(*c))`
    to avoid incremental rehashing.
    
  Fixes [#1142](https://github.com/apache/iceberg-go/issues/1142)
  
  ## Test plan
  - Existing tests cover both init paths — column stats are verified in
    `manifest_test.go` (v1, v2, round-trip) and partition data in
    `table_test.go`, `codec/data_file_test.go`, `dv_writer_test.go`, etc.
  - `go test ./...` passes with no regressions.

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1675,7 +1675,7 @@ func avroColMapToMap[K comparable, V any](c *[]colMap[K, V]) map[K]V {
 		return nil
 	}
 
-	out := make(map[K]V)
+	out := make(map[K]V, len(*c))
 	for _, data := range *c {
 		out[data.Key] = data.Value
 	}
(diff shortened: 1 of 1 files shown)


tanmayrauth (approved):
Looks good to me!

One small nit: the field/method names don't quite line up - `initPartition` field with `initPartitionData()` method, and `initColStats` field with `initColumnStatsData()` method. Could you match them either way (e.g. `initColStats` field + `initColStatsData()` method)? Not a blocker, just reads a bit cleaner.

tanmayrauth (approved):

zeroshade (approved):

PR #759, opened 2026-02-26
feat(table): roll parquet files based on actual compressed size

twuebi:
This change refactors datafile writing to use the actual written file size as iceberg-java & iceberg-rust do instead of the in-memory size.

PR #1150, opened 2026-06-02
refactor(table): clean up unused concurrentDataFileWriter in partitioned write paths 

xixipi-lining:
#### What changed:
This PR removes the  `concurrentDataFileWriter`  parameter from  `PartitionedFanoutWriter`, `PositionDeletePartitionedFanoutWriter` , and  `RollingDataWriter` .
  
#### Why:
The  concurrentDataFileWriter  parameter is currently passed into these partitioned writers but is not used within their implementations. Removing it simplifies the constructor signatures and reduces unnecessary parameter passing.
  
The  concurrentDataFileWriter  struct itself is preserved, as it is still used by unpartitioned delete operations.

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1686,10 +1686,10 @@ func positionDeleteRecordsToDataFiles(ctx context.Context, rootLocation string, 
 	targetFileSize := int64(meta.props.GetInt(WriteTargetFileSizeBytesKey,
 		WriteTargetFileSizeBytesDefault))
 
-	cw := newConcurrentDataFileWriter(func(rootLocation string, fs iceio.WriteFileIO, meta *MetadataBuilder, props iceberg.Properties, opts ...dataFileWriterOption) (dataFileWriter, error) {
-		return newPositionDeleteWriter(rootLocation, fs, meta, props, opts...)
-	}, withSchemaSanitization(false))
 	if latestMetadata.PartitionSpec().IsUnpartitioned() {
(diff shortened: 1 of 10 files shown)


tanmayrauth (changes requested):
Thanks for the cleanup, I confirmed the param became dead in #759 (the rolling-by-actual-compressed-size refactor) when r.concurrentWriter.writeFiles(...) was replaced by direct dataFileWriter.Write()  calls; the param threading was just left behind. 

Two follow-ups before this is complete, though by removing the param from the partitioned writers but leaving the unconditional allocation in callers introduced two new dead allocations:

1. `table/equality_delete_writer.go:161-162` - cw is now only consumed at line 216 in the unpartitioned tail, but still allocated unconditionally above the partitioned branch (which returns at line 190). Can you please move the cw := newConcurrentDataFileWriter(...) block to just before line 193 so it's only constructed when actually used.
2. `table/arrow_utils.go:1689-1691 (in positionDeleteRecordsToDataFiles)` - same shape: cw is only used at line 1716 in the unpartitioned branch, but allocated above the if IsUnpartitioned() check. Can you please move the allocation inside that if block.

`recordsToDataFiles` already does this correctly — it deletes the cw block entirely. The same pattern applies to the two sites above.

 

Otherwise LGTM — happy to approve once those two allocations are gone.

xixipi-lining:
@tanmayrauth Thanks for the detailed review!

I've updated the PR and moved the concurrentDataFileWriter allocations inside their respective unpartitioned blocks for both table/equality_delete_writer.go and table/arrow_utils.go.

Let me know if everything looks good now!

tanmayrauth (approved):
Thanks for the quick update, changes LGTM!
Will wait for Andrei for the final approval.

zeroshade (approved):

PR #1151, opened 2026-06-02
feat(table): write deletion vectors for partitioned v3 tables

### LINKED ISSUE #1135: feat(table): write deletion vectors for partitioned v3 tables

tanmayrauth:
Lifts the IsUnpartitioned() gate in positionDeleteRecordsToDataFiles so all v3+ writes go through the DV path, and threads partitionContextByFilePath into positionDeleteRecordsToDataFilesDV: each row's (specID, partitionData) is looked up per data file and passed to DVWriter.Add (groundwork: #1136), so output DataFile manifest entries carry the source data file's partition record.

A missing partition context is treated as a programming error rather than silently producing a default-partition entry, matching Java's BaseDVFileWriter contract.

The now-unreachable v3 "prefer deletion vectors" slog.Warn is removed.

Closes #1135.


tanmayrauth:
@laskoviymishka @zeroshade  can you please review this PR when you have a moment.

tanmayrauth:
@laskoviymishka  can you please review this PR?

table/arrow_utils.go
```diff
@@ -1753,13 +1750,14 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
...
-				// partitionData) so the partitioned path can pass through
-				// partitionContext directly. This site is still
-				// unpartitioned-only (the gate above routes partitioned
-				// writes elsewhere), so specID=0 + nil partition. PR (2)
-				// drops the gate and threads partitionContext through here.
-				writer.Add(filePaths.Value(int(i)), []int64{positions.Value(int(i))}, 0, nil)
+				filePath := filePaths.Value(int(i))
+				pCtx, ok := partitionContextByFilePath[filePath]
+				if !ok {
+					yield(nil, fmt.Errorf("unexpected missing partition context for path %s", filePath))
```
laskoviymishka, line 1796: this change makes a nil `partitionContextByFilePath` invalid for any v3 row, but `TestDVWritePathCancelledContext` still passes `nil` (it's unchanged in this diff, around line 262 of the test file).

It passes today only because the cancelled iterator yields `(nil, ctx.Err())` before this loop runs — so the lookup never happens. The three other unpartitioned v3 tests in this PR were converted to `unpartitionedContexts(...)`; this one was missed. If anyone later adds a valid row before the cancellation, the failure mode silently flips from "propagates context.Canceled" to "missing partition context." I'd switch it to `unpartitionedContexts()` (or `map[string]partitionContext{}`) to match. Fix that and this lands.

laskoviymishka (changes requested):
Nice work!

One blocking thing before merge: this PR makes a nil partitionContextByFilePath invalid on the v3 path, but TestDVWritePathCancelledContext still passes nil. It's safe today only because the cancelled iterator errors before the row loop runs, so it's a latent footgun, not a live failure. The  three other unpartitioned v3 tests here were converted to unpartitionedContexts(...); this one was missed. 

One-liner to fix.

table/arrow_utils.go
tanmayrauth: Good catch, I've switched TestDVWritePathCancelledContext to pass unpartitionedContexts() instead of nil so it matches the other v3 tests and won't silently flip failure modes if a valid row gets added  before the  cancellation. Test still passes.

laskoviymishka (approved):
:shipit:

PR #1152, opened 2026-06-02
feat(table): read shredded Parquet variants

### LINKED ISSUE #986: feat(table): variant shredded reader
Parent: #589

#929 / PR #932 cover the non-shredded variant path (`struct<metadata: binary, value: binary>`). The shredded path is the next piece: a Parquet variant column appears as `struct<metadata: binary, value: binary, typed_value: STRUCT>` where `typed_value` mirrors the shredded subtree and the reader reconstructs the full variant by preferring `typed_value` per-field and falling back to the residual `value` slice.

The Parquet codec lives in arrow-go's `parquet/variant` (already in `go.mod`); the spec is in [Parquet Variant shredding](https://github.com/apache/parquet-format/blob/master/VariantShredding.md). Add a new file `table/internal/variant_shredded.go` carrying a pure `ReassembleShreddedVariant(metadata, value, typedValue) (variant.Value, error)` that walks the shredded struct per spec, then call it from `table/internal/parquet_files.go` so a shredded variant column reads identically to a non-shredded one. A shredded column should be invisible to the scanner.

Cross-client coverage: a Java-produced shredded variant fixture committed under `table/internal/testdata/` and a golden test asserting iceberg-go reads the same `variant.Value`. Java apache/iceberg PR landing the writer: [apache/iceberg#11500](https://github.com/apache/iceberg/pull/11500) and follow-ups.

matanper:
Closes #986.

## Summary

- Adds `ReassembleShreddedVariant` — a named entry point that reconstructs a `variant.Value` from one row of a shredded Parquet variant column, delegating the spec walk to arrow-go's `*extensions.VariantArray`.
- Adds `WrapShreddedVariantReader`, a `RecordReader` wrapper that replaces shredded variant columns with their non-shredded equivalents on each batch, so the scanner always sees a uniform `struct<metadata, value>` layout regardless of how the Parquet file was written.
- Wires `WrapShreddedVariantReader` into `parquet_files.go`'s `GetRecords` path. For files with no shredded variant columns the wrapper is a no-op (returns the inner reader unchanged).
- Vendors six Java-produced fixture files from apache/parquet-testing (pinned to a specific commit) and adds two tests — one exercising the pure per-row function, one exercising the full scan path — covering primitive shredding, flat objects, nested shredded objects, and partially-shredded objects (typed_value + residual merge).

### Performance note
The requirement that a shredded column be invisible to the scanner means we eagerly reassemble every row on every batch read. This adds per-batch CPU overhead and causes all typed_value sub-columns to be read from Parquet and discarded after reassembly — even when the consumer never accesses the variant value. A lazy approach is not straightforward: every pipeline step that reconstructs a RecordBatch (deletion vector filtering, projection, row lineage) validates that column types exactly match the schema, so a shredded VariantArray cannot pass through a non-shredded schema without panicking. Fixing this properly would require either a wrapper array type that presents as non-shredded but reassembles lazily on .Value(i), or making the entire pipeline shredding-aware.


tanmayrauth:
I think @nssalian  is already working on this PR.

nssalian:
@matanper I was going to put out a proposal before doing this. I'll fix this once I get consensus with the community. Also, please check prior to working on an issue if already assigned.

matanper:
Hi @nssalian first of all thank you for your contributions to this library 🙏 

You are assigned to multiple issues at once without noticeable progress, and since I need this feature the natural thing for me, instead of being blocked, is to implement and contribute back to the community in order to help the project moving forward.

At the end both of us wish for this project to succeed

tanmayrauth:
@matanper  thanks for the PR but this the [general guideline](https://github.com/apache/iceberg-go/blob/main/CONTRIBUTING.md#picking-up-issues) for claiming a issue before woking on the PR.

PR #1154, opened 2026-06-03
Avoid carrying empty overwrite manifests

### LINKED ISSUE #1153: Repeated overwrite commits carry delete-only manifests forward
### Apache Iceberg version

main (development)

## Problem

Repeated overwrite commits can keep carrying old manifests that no longer contribute any live entries to the current snapshot.

In `overwriteFiles.existingManifests`, existing manifest entries are read with deleted entries filtered out:

```go
for entry, err := range of.base.iterManifestEntries(m, true) {
    ...
}
```

If a carried manifest only contains deleted entries, this leaves `notDeleted` empty and `foundDeletedCount == 0`. The current logic then appends the original manifest to `existingFiles`, so the delete-only manifest remains reachable from the new snapshot.

## Why This Matters

For workloads that repeatedly replace the full table contents, every overwrite can create a delete-entry manifest for the previous files. Old delete-only manifests are then carried into later snapshots even though they do not affect current snapshot reads.

Over time, the current snapshot's manifest list grows with stale delete-only manifests and normal orphan cleanup cannot remove those Avro files because they are still reachable.

## Expected Behavior

After deleted entries are filtered out, a manifest with zero remaining live entries should not be carried forward into the next overwrite snapshot.

This should not drop zero-count inherited manifests blindly. The check should be based on actually reading entries with `discardDeleted=true`, not on manifest count metadata, because some valid inherited manifests may have zero or unset counts but still contain live entries.

## Proposed Fix

In the overwrite existing-manifest path, skip an existing manifest if `iterManifestEntries(m, true)` produces no live entries.

Add a regression test with repeated `ReplaceDataFilesWithDataFiles` calls to verify the current snapshot keeps a bounded manifest list instead of accumulating old delete-only manifests.

rraulinio:
Fixes #1153.

Overwrite commits read existing manifest entries with deleted entries filtered out. If no live entries remain, carrying that manifest forward keeps stale deletion history reachable in the current snapshot and causes repeated replace operations to grow the manifest list.

This change skips existing overwrite manifests that have no live entries after `discardDeleted=true` filtering. It does not rely on manifest-level count metadata, so zero-count inherited manifests that still contain live entries are still preserved.

A regression test covers repeated `ReplaceDataFilesWithDataFiles` calls and verifies the current snapshot keeps a bounded manifest list instead of accumulating old delete-only manifests.

### DIFF
--- a/table/snapshot_producers.go
+++ b/table/snapshot_producers.go
@@ -187,13 +187,13 @@ func (of *overwriteFiles) existingManifests() ([]iceberg.ManifestFile, error) {
 			}
 		}
 
-		if foundDeletedCount == 0 {
-			existingFiles = append(existingFiles, m)
-
+		if len(notDeleted) == 0 {
 			continue
 		}
 
-		if len(notDeleted) == 0 {
+		if foundDeletedCount == 0 {
+			existingFiles = append(existingFiles, m)
+
 			continue
 		}
 
--- a/table/table_test.go
+++ b/table/table_test.go
@@ -1275,6 +1275,55 @@ func (t *TableWritingTestSuite) TestReplaceDataFilesWithDataFiles() {
 	t.Equal("2", staged.CurrentSnapshot().Summary.Properties["deleted-data-files"])
 }
 
+func (t *TableWritingTestSuite) TestReplaceDataFilesWithDataFilesDoesNotCarryEmptyManifests() {
+	ident := table.Identifier{"default", "replace_data_files_drops_empty_manifests_v" + strconv.Itoa(t.formatVersion)}
+	tbl := t.createTable(ident, t.formatVersion, *iceberg.UnpartitionedSpec, t.tableSchema)
+
+	writeDataFile := func(name string) iceberg.DataFile {
+		path := fmt.Sprintf("%s/%s/%s.parquet", t.location, ident[1], name)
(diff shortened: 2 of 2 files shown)


tanmayrauth (approved):
LGTM.

  Walked through the four `(foundDeletedCount, len(notDeleted))` cases — only `(0, 0)` changes behavior, which is exactly the all-tombstone manifest from a prior snapshot, and that's right to drop.
  `iterManifestEntries(m, true)` only filters `EntryStatusDELETED` (`manifest.go:835`), so an empty iteration is a reliable signal for "no live data here". The current overwrite's own tombstones are produced separately in `overwriteFiles.deletedEntries`, so this doesn't lose any deletion record. Older snapshots' manifest lists still reference the skipped manifests, so time-travel and expiration paths are unaffected.

  Two small things in the test, both optional:

  - `table/table_test.go:1322`: the trailing `continue` is dead since it's the last statement in the loop body.
  - `table/table_test.go:1295`: `for i := range 5` with `i+1` inside reads a bit oddly; `for i := 1; i <= 5; i++` would let you drop the `+1`.

zeroshade (approved):
LGTM

zeroshade:
@rraulinio let's get the suggestions that @tanmayrauth implemented first, then we'll merge

rraulinio:
@zeroshade Done.

PR #1156, opened 2026-06-03
fix(table): serialize name mapping as JSON in UpdateSchema build updates

hdnpth:
`UpdateSchema.BuildUpdates` was storing the updated name mapping under `schema.name-mapping.default` by calling `NameMapping.String()`. That method is a pretty-printer ((`[\n\t([foo] -> 1)\n]`) and not JSON. The reader on the other side (`MetadataBuilder.NameMapping`) parses the property with `json.Unmarshal` and silently returns `nil` on error, so any mapping written through `UpdateSchema` was effectively lost on the next read


zeroshade (commented):
LGTM in general, but needs to fix the failing CI first

zeroshade (approved):

PR #1157, opened 2026-06-03
feat(table): enforce sort order on write #833

### LINKED ISSUE #833: feat: enforce sort order on write
Parent: #829 (v2 spec completion)
Sort orders are fully modeled in metadata — `SortOrder`, `SortField`, direction, null order, transforms — and tracked through schema/spec evolution. However the write path never consults the table's sort order. Data is written in arrival order regardless of the `default-sort-order` setting.

### Current state
- `SortOrder` / `SortField` types with JSON marshaling, validation, compatibility checks — `table/sorting.go`
- `MetadataBuilder.AddSortOrder`, `SetDefaultSortOrder` — fully wired
- `WriteTask.SortOrderID` field exists (`table/writer.go`) but is never populated or used
- `writerFactory` in `table/rolling_data_writer.go` has access to metadata but doesn't read the sort order
- Arrow Go compute has no sort kernel, but `compute.Take` exists (used for positional deletes in #825, #762) — can reorder batches by permutation index

### Context
Sort order enforcement improves query performance for engines that can exploit sorted data (range scans, merge joins). It's also needed for sort-aware compaction (#832) — when compacting, the output should respect the table's sort order.

tanmayrauth:
Resolve the table's default sort order to Arrow compute SortKeys and apply compute.SortRecordBatch to each batch before flushing to a data file. Skip for unsorted tables (sort_order_id 0) and for delete files, where the table-level order does not apply.

All data write paths route through writerFactory + RollingDataWriter (unpartitioned, clustered, and hash-fanout), so Append, Overwrite, Delete data writes, WriteRecords, and compaction (which calls WriteRecords) all pick up the sort.

Limitations:
- Per-batch, not per-file: each Add() is sorted internally; rows are not merged across batches. Matches PyIceberg.
- Bucket transform falls back to source-column order (the manifest still records sort_order_id verbatim).
- Multi-arg transform sort fields use only the first source column.
- Nested-field sort sources are rejected with a clear error.

Closes #833 (https://github.com/apache/iceberg-go/issues/833).

### DIFF
--- a/table/rolling_data_writer.go
+++ b/table/rolling_data_writer.go
@@ -28,6 +28,7 @@ import (
 	"sync/atomic"
 
 	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/compute"
 	"github.com/apache/iceberg-go"
 	iceio "github.com/apache/iceberg-go/io"
 	tblutils "github.com/apache/iceberg-go/table/internal"
@@ -57,6 +58,7 @@ type writerFactory struct {
 	content          iceberg.ManifestEntryContent
 	equalityFieldIDs []int
 	sortOrderID      int
+	sortKeys         []compute.SortKey
 
 	writers               sync.Map
 	partitionLocProviders sync.Map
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
Looks good, only nitpick is to add documentation pointing out that the sorting is only per-batch and not per-file etc. and the other limitations.

zeroshade:
Might need a rebase and fix some linting issues :smile:

Issue #1096, opened 2026-05-19
feat(catalog): Cloud storage backend support for hadoop


tanmayrauth:
### Feature Request / Improvement

The current implementation only supports local filesystem paths (`file://`). Cloud storage backends (S3, GCS, ABFS) need to be wired through the `io.IO` abstraction instead of direct `os` calls.

C-Loftus:
@tanmayrauth would it be reasonable for me to take this on if no one else is working on it? If so feel free to assign me to this.

For context, I am looking to implement relative path support on disk for a hadoop catalog in my downstream application via a custom `IO` interface and I think allowing the hadoop catalog to use the `IO` abstraction would solve both issues at once. (I know that relative paths may be coming in v4, but mainly looking to get relative paths in some custom code using v3)

tanmayrauth:
Thanks for offering, @C-Loftus. @laskoviymishka / @zeroshade  can help you get assigned this issue. 
  
  Before any code lands I'd want to nail down a few things, because the issue title is broader than what's safe to ship:

  1. Atomic rename / commit semantics. The current implementation relies on os.Rename in three load-bearing spots — version-hint write, namespace metadata write, and CommitTable (catalog/hadoop/hadoop.go ). That rename is the catalog's only mechanism for safe concurrent commits. S3 has no atomic rename (it's copy + delete); GCS and ABFS have different consistency stories than POSIX. The Java HadoopCatalog carries the same limitation and explicitly discourages object-store usage. Wiring io.IO makes the call sites cloud-capable but does not make commits safe — we either need a per-backend conditional-write story (S3 If-Match, GCS generation preconditions, etc.) or an explicit "unsafe for concurrent writers" caveat surfaced on CreateCatalog and in docs. A written plan for this is the part I'd push hardest on.
  2. Interface surface. io.IO today is read/write/delete-oriented. The Hadoop catalog also walks metadata/ to discover vN.metadata.json and stats files for existence checks, so listing + stat capabilities (ListableIO, equivalent for stat) need to be in scope. Worth deciding up front whether you extend io.IO or compose interfaces.
  3. Relative paths are a separate concern. Switching to io.IO is necessary but not sufficient for relative-path support — that's a contract decision about how the IO resolves identifiers, not catalog-side code. I'd suggest keeping the cloud-IO change pure and tracking relative-path support as a follow-up.

[...]

PR #1158, opened 2026-06-03
refactor(catalog/hadoop): Change Hadoop Catalog to Use IO Interface

C-Loftus:
First step of addressing https://github.com/apache/iceberg-go/issues/1096 This just changes the hadoop catalog implementation to support an interface but still requires a local filesystem.

- All `os.` calls have been swapped to using an interface `HadoopCatalogFS`. 
- `HadoopCatalogFS` implements the same exact calls as `os.` 
    - There were a few calls to `os.remove` that didn't handle the `err` value. I made these `_` to explicitly discard the err value.
    - In a follow up PR I can condense the number of methods to reduce interface size. However as a first pass, I wanted to keep the interface and thus the behavior very easy to review.
- The `localfs` struct was extended to support the additional methods. 
    - There were a few places where `mkdir` used `0o777` for a permission. I didn't see any reason to allow for execution by other users so I changed these to `0o755`

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -25,7 +25,6 @@ import (
 	"iter"
 	"log"
 	"net/url"
-	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -98,9 +97,10 @@ func validateIdentifier(ident table.Identifier) error {
 var _ catalog.PurgeableTable = (*Catalog)(nil)
 
 type Catalog struct {
-	name      string
-	warehouse string
-	props     iceberg.Properties
+	name       string
+	warehouse  string
+	filesystem HadoopCatalogFS
+	props      iceberg.Properties
 }
 
 // NewCatalog creates a new Hadoop catalog rooted at the given warehouse path.
@@ -145,7 +145,10 @@ func NewCatalog(name, warehouse string, props iceberg.Properties) (*Catalog, err
 	return &Catalog{
 		name:      name,
 		warehouse: warehouse,
-		props:     props,
+		// for the time being, we default to localfs since there is not yet
+		// support for other filesystems like blob stores
+		filesystem: icebergio.LocalFS{},
+		props:      props,
 	}, nil
 }
 
@@ -182,10 +185,10 @@ func (c *Catalog) defaultTableLocation(ident table.Identifier) string {
 //   - v*.metadata.json (Hadoop catalog format)
 //   - <seq>-<uuid>.metadata.json (Java/PyIceberg format)
 //   - version-hint.text
-func isTableDir(path string) bool {
+func isTableDir(filesystem HadoopCatalogFS, path string) bool {
 	metaDir := filepath.Join(path, "metadata")
 
-	entries, err := os.ReadDir(metaDir)
+	entries, err := filesystem.ReadDir(metaDir)
 	if err != nil {
 		return false
 	}
@@ -207,7 +210,7 @@ func isTableDir(path string) bool {
 }
 
 func (c *Catalog) readVersionHint(ident table.Identifier) int {
-	data, err := os.ReadFile(c.versionHintPath(ident))
+	data, err := c.filesystem.ReadFile(c.versionHintPath(ident))
 	if err != nil {
 		return 0
 	}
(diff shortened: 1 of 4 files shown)


C-Loftus:
In a future follow up PR, the goal I think would be to 

1. Rewrite some of this logic to decrease the number of methods within the `HadoopCatalogFS` interface
    - Once the list of interfaces are stable, it also might be relevant to move these into the main `io.IO` interface list. Unsure.  However, I wanted to keep them in `hadoop` until it was clear what was explicitly needed for the catalog.
2. Add an interface and more direct logic and documentation around the atomicity requirements described in https://github.com/apache/iceberg-go/issues/1096 for a fs to be used for the hadoop catalog after the atomic object store related work is added

catalog/hadoop/io.go
```diff
@@ -0,0 +1,100 @@
...
+// Ensure that the LocalFS implements the extensions
+// thus ensuring that the hadoop catalog can use the LocalFS for its
+// IO operations
+var (
+	_ icebergio.IO = (*icebergio.LocalFS)(nil)
+	_ StatIO       = (*icebergio.LocalFS)(nil)
+	_ RenameIO     = (*icebergio.LocalFS)(nil)
+	_ RemoveAllIO  = (*icebergio.LocalFS)(nil)
+	_ MkdirIO      = (*icebergio.LocalFS)(nil)
+	_ MkdirAllIO   = (*icebergio.LocalFS)(nil)
```
tanmayrauth: LocalFS.WriteFile mode should be 0o644, not 0o755. Your PR description explains the 0o777 → 0o755 change as removing the execute bit, but WriteFile's third arg is a file mode where 0o755 still grants +x to everyone. The hadoop catalog used to write the version-hint file with explicit 0o644 at hadoop.go (pre-PR line 232); after this PR it goes through LocalFS.WriteFile and inherits 0o755, so the version-hint file is now created executable. 0o755 is correct for the directory cases (Mkdir/MkdirAll), just not for WriteFile.

catalog/hadoop/hadoop.go
```diff
@@ -354,13 +357,13 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 	}
 
 	if err := internal.WriteTableMetadata(metadata, icebergio.LocalFS{}, tempPath, compression); err != nil {
```
tanmayrauth: pass `c.filesystem` to  `WriteTableMetadata`, not `icebergio.LocalFS{}`.** `HadoopCatalogFS` embeds  `WriteFileIO`, so `c.filesystem` already satisfies the parameter. Today both  resolve to the same value via the hardcoded constructor, but the whole point  of this PR is that `c.filesystem` can later be swapped — leaving those two call sites pinned to `LocalFS{}` means the eventual cloud-backed catalog would write metadata to local disk while the surrounding rename/cleanup goes to the cloud FS. Worth fixing in this PR so the abstraction is fully wired.

tanmayrauth (commented):
Thanks for kicking this off, the interface decomposition reads well as a first  pass. Two correctness issues that I think should be addressed before merge, plus some smaller cleanup:

  1. **`io/local.go:49` — `LocalFS.WriteFile` mode should be `0o644`, not `0o755`.** Your PR description explains the `0o777 → 0o755` change as removing the execute bit, but `WriteFile`'s third arg is a *file* mode where `0o755` still grants `+x` to everyone. The hadoop catalog used to write the version-hint file with explicit `0o644` at `hadoop.go` (pre-PR line 232); after this PR it  goes through `LocalFS.WriteFile` and inherits `0o755`, so the version-hint file is now created executable. `0o755` is correct for the directory cases (`Mkdir`/`MkdirAll`), just not for `WriteFile`.

  2. **`catalog/hadoop/hadoop.go:359` and `:479` — pass `c.filesystem` to `WriteTableMetadata`, not `icebergio.LocalFS{}`.** `HadoopCatalogFS` embeds `WriteFileIO`, so `c.filesystem` already satisfies the parameter. Today both resolve to the same value via the hardcoded constructor, but the whole point  of this PR is that `c.filesystem` can later be swapped — leaving those two call sites pinned to `LocalFS{}` means the eventual cloud-backed catalog would write metadata to local disk while the surrounding rename/cleanup goes  to the cloud FS. Worth fixing in this PR so the abstraction is fully wired.


  **Smaller things:**

  3. `catalog/hadoop/io.go:43-51` — the per-interface assertions miss `ListableIO`/`ReadFileIO`/`WriteFileIO` and the `IO` one is redundant with  the others. Replace the block with a single `var _ HadoopCatalogFS = (*icebergio.LocalFS)(nil)` so it stays correct automatically when methods are added/removed.

  4. `RenameIO`'s "this should be atomic" comment is load-bearing — the catalog's optimistic-concurrency relies on it at `hadoop.go:365` and `:238`. Worth promoting "should" to "must" and documenting that `StatIO`/`MkdirIO` errors need to satisfy `errors.Is(err, fs.ErrNotExist)` / `fs.ErrExist`, since callers use those predicates already and a non-`os` implementation must wrap errors accordingly.

  5. The new extension interfaces look generic enough to live in `io/`  eventually — fine to defer, but flagging for the follow-up.

C-Loftus:
Thank you very much for the review @tanmayrauth! I appreciate it. I believe I have addressed your feedback now. Happy to iterate on anything else if you wish.

1. Documented which `fs.` errors each method in the `io` file needs to fulfill
2. Switched from a hardcoded `localfs` to the catalog interface.
3. Changed the `writefile` file permissions to your feedback.

tanmayrauth (approved):

zeroshade (approved):

PR #1159, opened 2026-06-03
fix(table): reject time64[ns] in partition literal extraction (#1115)

### LINKED ISSUE #1115: Partitioned writes should handle Arrow Time64 units consistently
In #1112, timestamp partition extraction was updated to respect Arrow timestamp units before applying Iceberg partition transforms.

There is a similar-looking edge in `getArrowValueAsIcebergLiteral` for `*array.Time64`:

```go
case *array.Time64:
    return iceberg.NewLiteral(iceberg.Time(arr.Value(row))), nil
```

Iceberg `time` values are stored as microseconds, but Arrow `Time64` can carry a unit. Today the public Arrow schema conversion path only accepts `time64[us]` and rejects `time64[ns]`, so this is lower priority than the timestamp partition bug. Still, the partition literal helper currently assumes the raw Arrow value is already in Iceberg microseconds.

Proposed follow-up:

- inspect `arr.DataType().(*arrow.Time64Type).Unit`
- keep `time64[us]` unchanged
- either reject `time64[ns]` explicitly in partition extraction or convert it with the same floored downcast semantics used for timestamps
- add focused regression coverage around the helper / partition extraction behavior

This is intentionally separate from #1112 because timestamp writes are already accepted and normalized later, while Time64 has a narrower reachable surface today.

tanmayrauth:
`getArrowValueAsIcebergLiteral` previously assumed every `*array.Time64` already held microseconds and ignored the Arrow `Time64Type.Unit`. This is unreachable today because the schema-conversion path in `arrow_utils.go` only accepts `time64[us]`, but the helper was still silently producing wrong partition keys if `time64[ns]` ever reached it.

Inspect `Time64Type.Unit` and return `iceberg.ErrInvalidSchema` for any unit other than microseconds, mirroring the schema-conversion error. Add a focused unit test covering both the success and rejection paths.

Closes https://github.com/apache/iceberg-go/issues/1115

### DIFF
--- a/table/partitioned_fanout_writer_test.go
+++ b/table/partitioned_fanout_writer_test.go
@@ -597,7 +597,7 @@ func (s *FanoutWriterTestSuite) TestGetArrowValueAsIcebergLiteralTime64() {
 	usArr := usBldr.NewTime64Array()
 	defer usArr.Release()
 
-	lit, err := getArrowValueAsIcebergLiteral(usArr, 0)
+	lit, err := getArrowValueAsIcebergLiteral(usArr, 0, iceberg.PrimitiveTypes.Time)
 	s.Require().NoError(err)
 	s.Equal(iceberg.NewLiteral(iceberg.Time(5_000_000)), lit)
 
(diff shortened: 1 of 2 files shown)


table/partitioned_fanout_writer_test.go
```diff
@@ -588,3 +405,27 @@ func (s *FanoutWriterTestSuite) createComprehensiveTestRecord() arrow.RecordBatc
 
 	return bldr.NewRecordBatch()
 }
+
```
fallintoplace, line 591: The implementation looks good to me aside from type check.

zeroshade (approved):

PR #1162, opened 2026-06-03
fix(table): preserve table props when building partition location provider (#1155)

### LINKED ISSUE #1155: [Bug] write.object-storage.enabled  is completely ignored when writing to partitioned tables
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

When appending data to a partitioned table with the table property  write.object-storage.enabled  set to  true , the object storage layout is not applied. Instead of generating data file paths with entropy hash prefixes (e.g.,  <hash>-<filename>.parquet ), the writer falls back to the default Hadoop-style layout, writing directly to the partition directory (e.g.,  <table_location>/data/<partition_path>/<filename>.parquet ).

This issue does not affect unpartitioned tables.

#### To Reproduce

1. Create a partitioned Iceberg table.
2. Set the table property  write.object-storage.enabled  to  true .
3. Append data to the table using the  iceberg-go  SDK (e.g., via  Transaction.Append ).
4. Observe the final data file paths in the storage. The files are written without the expected hash prefixes.

#### Root Cause Analysis
The bug is caused by property state loss in  table/rolling_data_writer.go .
When  writerFactory  dynamically generates a  LocationProvider  for a specific partition, it drops all the original table properties.

In  func (w *writerFactory) partitionLocProvider(partitionPath string) :

  partitionDataPath := w.rootURL.JoinPath("data", partitionPath).String()

  // BUG: A new empty map is passed, losing all original table properties
  // including 'write.object-storage.enabled'.
  loc, err := LoadLocationProvider(w.rootLocation,
      iceberg.Properties{WriteDataPathKey: partitionDataPath})

Because  LoadLocationProvider  receives a fresh properties map containing only  WriteDataPathKey , it evaluates  ObjectStoreEnabledKey  as  false  and incorrectly falls back to  simpleLocationProvider .

#### Expected behavior
The  writerFactory  should preserve and pass the original table properties (or at least the relevant object-storage properties) down to  LoadLocationProvider  when creating partition location providers, so that     
objectStoreLocationProvider  is correctly instantiated and hashes are added.

tanmayrauth:
writerFactory.partitionLocProvider was passing only WriteDataPathKey to LoadLocationProvider, dropping write.object-storage.enabled and other table props. As a result, partitioned writes always fell back to the simple Hadoop-style layout, ignoring the object-storage entropy prefix.

Store the table props on the factory and copy them when constructing partition props, overlaying only the partition-specific data path.

Fixes: #1155

### DIFF
--- a/table/rolling_data_writer.go
+++ b/table/rolling_data_writer.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"iter"
+	"maps"
 	"net/url"
 	"strings"
 	"sync"
@@ -222,9 +223,7 @@ func (w *writerFactory) partitionLocProvider(partitionPath string) (LocationProv
 
 	partitionDataPath := w.rootURL.JoinPath("data", partitionPath).String()
 	partitionProps := make(iceberg.Properties, len(w.tableProps)+1)
-	for k, v := range w.tableProps {
-		partitionProps[k] = v
-	}
+	maps.Copy(partitionProps, w.tableProps)
 	partitionProps[WriteDataPathKey] = partitionDataPath
 	loc, err := LoadLocationProvider(w.rootLocation, partitionProps)
 	if err != nil {
--- a/table/rolling_data_writer.go
+++ b/table/rolling_data_writer.go
@@ -45,6 +45,7 @@ type writerFactory struct {
 	targetFileSize int64
 
 	locProvider      LocationProvider
+	tableProps       iceberg.Properties
 	fileSchema       *iceberg.Schema
 	arrowSchema      *arrow.Schema
 	writeProps       any
@@ -152,6 +153,7 @@ func newWriterFactory(rootLocation string, args recordWritingArgs, meta *Metadat
 		taskSchema:     taskSchema,
 		targetFileSize: targetFileSize,
 		locProvider:    locProvider,
+		tableProps:     meta.props,
 		fileSchema:     fileSchema,
 		arrowSchema:    arrowSchema,
 		writeProps:     format.GetWriteProperties(meta.props),
@@ -219,8 +221,12 @@ func (w *writerFactory) partitionLocProvider(partitionPath string) (LocationProv
 	}
 
 	partitionDataPath := w.rootURL.JoinPath("data", partitionPath).String()
-	loc, err := LoadLocationProvider(w.rootLocation,
-		iceberg.Properties{WriteDataPathKey: partitionDataPath})
+	partitionProps := make(iceberg.Properties, len(w.tableProps)+1)
+	for k, v := range w.tableProps {
+		partitionProps[k] = v
+	}
+	partitionProps[WriteDataPathKey] = partitionDataPath
+	loc, err := LoadLocationProvider(w.rootLocation, partitionProps)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


xixipi-lining:
Hi @tanmayrauth , thanks for working on this fix!

While this approach resolves the issue, I have an architectural concern regarding writerFactory. Storing tableProps on the factory just to pass them down into LoadLocationProvider for each partition adds extra state to it, which we should try to avoid to keep its responsibilities focused.

If we look at the [Java implementation of LocationProvider](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/io/LocationProvider.java), it actually provides a partition-aware method:
newDataLocation(PartitionSpec spec, StructLike partitionData, String filename)

Long term, evolving our LocationProvider to be partition-aware like Java's would cleanly remove the need for writerFactory to manage tableProps or rebuild providers per partition.

What are your thoughts on taking this direction?

tanmayrauth:
Thanks @xixipi-lining, that's a fair point and I agree on the direction.
  
  On this PR's scope. I'd like to keep this minimal because #1155 is a correctness regression — write.object-storage.enabled is silently ignored on partitioned writes today, so partitioned tables end up with no entropy at all in their data paths. The two extra fields on writerFactory restore the documented behavior without changing any public API. I'd rather land that on its own than gate the bug fix behind an interface change.

  On the architectural direction. Agreed that long-term the partition-aware overload is the right shape. I checked the Java reference and LocationProvider declares it as an abstract method on the interface itself:

  String newDataLocation(PartitionSpec spec, StructLike partitionData, String filename);

  with ObjectStoreLocationProvider folding the partition path into the hash input when includePartitionPaths=true, and dropping the partition entirely when it's false. That's actually a stronger argument than I first realized: the factory-rebuilds-per-partition pattern in Go can't reproduce that layout no matter what props you pass — the hash dirs end up after the partition path instead of before, and the hash input is filename-only. So this PR restores entropy, but it does not yet make Go match Java's spec for partitioned object-storage layout.
  
  That makes the follow-up worth doing properly:

  - add the partition-aware method to LocationProvider
  - have objectStoreLocationProvider produce Java's layout (<storage>/<hash>/<partitionPath>/<filename> with partition folded into the hash) and honor includePartitionPaths=false by dropping the partition
  - remove tableProps and partitionLocProvider from writerFactory once the call sites move over

  That's a public-interface change and a behavior change for anyone already on the current partitioned layout, so it deserves its own PR with a clear note in the changelog. Happy to file the follow-up issue and pick it up after this lands — or, if you'd prefer, I can draft the interface signature first so we can align before either of us writes code.

  WDYT — land this as the regression fix and tackle the interface evolution separately?

xixipi-lining:
@tanmayrauth  Sounds like a solid plan. Let's land this regression fix first.

I completely agree with tackling the LocationProvider evolution in a follow-up issue. Feel free to draft the interface signature there so we can align before diving into the code.

LGTM for this PR. Thanks for driving this!

table/rolling_data_writer.go
```diff
@@ -219,8 +221,12 @@ func (w *writerFactory) partitionLocProvider(partitionPath string) (LocationProv
 	}
 
 	partitionDataPath := w.rootURL.JoinPath("data", partitionPath).String()
-	loc, err := LoadLocationProvider(w.rootLocation,
-		iceberg.Properties{WriteDataPathKey: partitionDataPath})
+	partitionProps := make(iceberg.Properties, len(w.tableProps)+1)
+	for k, v := range w.tableProps {
+		partitionProps[k] = v
+	}
```
zeroshade: `maps.Copy`?

zeroshade (changes requested):

table/rolling_data_writer.go
tanmayrauth: Good call, done, switched to maps.Copy.

zeroshade (approved):

PR #1163, opened 2026-06-04
fix(io/s3): preserve tuned transport options when configuring HTTP client

### LINKED ISSUE #1148: io/s3: preserve tuned transport options when configuring HTTP client
## Issue

When `s3.connect-timeout` or `s3.proxy-uri` is set, `ParseAWSConfig` populates `awscfg.HTTPClient`. The nil guard in `createS3Bucket` then skips applying the tuned transport options:

- `MaxIdleConnsPerHost: 256`
- `MaxIdleConns: 256`
- `MaxConnsPerHost: 256`
- `IdleConnTimeout: 90s`

As a result, these code paths fall back to the AWS SDK defaults, including `MaxIdleConnsPerHost: 10`.

This can hurt highly parallel scans on tables with many files.

## Possible Fixes

- Factor a shared `newTunedBuildableClient()` helper and use it in both `ParseAWSConfig` and `createS3Bucket`.
- Or apply the tuning inside `ParseAWSConfig` directly.

Flagged during review of #1141.

Revanth14:
Fixes #1148.

## Summary
- Factor S3 HTTP client transport tuning into a shared helper.
- Use the tuned buildable client when `s3.proxy-uri` configures an HTTP client.
- Use the tuned buildable client when `s3.connect-timeout` configures an HTTP client.
- Preserve the existing S3 transport tuning values for all configured S3 HTTP clients.

## Testing
- `env GOCACHE=/private/tmp/iceberg-go-gocache go test ./io/gocloud -run 'TestParseAWSConfig' -count=1`
- `env GOCACHE=/private/tmp/iceberg-go-gocache go vet ./io/gocloud/...`
- `git diff --check`

### DIFF
--- a/io/gocloud/s3.go
+++ b/io/gocloud/s3.go
@@ -76,7 +76,7 @@ func ParseAWSConfig(ctx context.Context, props map[string]string) (*aws.Config, 
 			return nil, fmt.Errorf("invalid s3 proxy url %q: %w", proxy, err)
 		}
 
-		httpClient = awshttp.NewBuildableClient().WithTransportOptions(
+		httpClient = newS3BuildableClient().WithTransportOptions(
 			func(t *http.Transport) {
 				t.Proxy = http.ProxyURL(proxyURL)
 			},
@@ -90,7 +90,7 @@ func ParseAWSConfig(ctx context.Context, props map[string]string) (*aws.Config, 
 		}
 
 		if httpClient == nil {
-			httpClient = awshttp.NewBuildableClient()
+			httpClient = newS3BuildableClient()
 		}
 		httpClient = httpClient.WithDialerOptions(func(d *net.Dialer) {
 			d.Timeout = duration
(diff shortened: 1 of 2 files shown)


Revanth14:
@tanmayrauth, this fix is for the separate issue we discussed on pr #1141 
can you please review?

tanmayrauth (approved):
LGTM!

One thing to fix before merge:

 The proxy/timeout paths previously inherited the SDK default `MaxConnsPerHost = 2048`, and this PR brings them down to `s3MaxConnsPerHost = 256` to match the no-proxy default. That's a real cap reduction for anyone running heavy parallel S3 traffic through a proxy. Either call this out explicitly in the PR description as an intentional consistency change, or bump `s3MaxConnsPerHost` back up to 2048 (and keep the other three at 256) so the proxy/timeout paths don't regress on concurrency.

Revanth14:
> LGTM!
> 
> One thing to fix before merge:
> 
> The proxy/timeout paths previously inherited the SDK default `MaxConnsPerHost = 2048`, and this PR brings them down to `s3MaxConnsPerHost = 256` to match the no-proxy default. That's a real cap reduction for anyone running heavy parallel S3 traffic through a proxy. Either call this out explicitly in the PR description as an intentional consistency change, or bump `s3MaxConnsPerHost` back up to 2048 (and keep the other three at 256) so the proxy/timeout paths don't regress on concurrency.

Thanks, fixed it. 
I bumped `MaxConnsPerHost` to `2048` so the proxy/connect-timeout paths keep the previous SDK concurrency cap, while the idle connection tuning remains at `256`/`90s`.

tanmayrauth (approved):

zeroshade (approved):

PR #1164, opened 2026-06-04
fix(table): preserve InitialDefault and WriteDefault in AssignFreshSchemaIDs

hdnpth:
`AssignFreshSchemaIDs` silently drops `InitialDefault` and `WriteDefault` on every field. This corrupts the behaviour`UpdateSchema.AddColumn(..., defaultValue)`, which constructs the new field with defaults and calls `AssignFreshSchemaIDs`.

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -1020,11 +1020,13 @@ func (p *pruneColVisitor) Struct(st StructType, fieldResults []Type) Type {
 			sameType = false
 			// type has changed, create a new field with the projected type
 			selected = append(selected, NestedField{
-				ID:       field.ID,
-				Name:     field.Name,
-				Type:     t,
-				Doc:      field.Doc,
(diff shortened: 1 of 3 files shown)


tanmayrauth (approved):
LGTM!

Few suggestions on test cases:
1. `TestAssignFreshSchemaIDsPreservesDefaults` (schema_test.go:764) only covers top-level  primitive fields. The recursive `setFreshIDs.Struct` path is the one most likely to silently egress. A nested case (default on a field inside a StructType) would be cheap insurance.

 2. The same test could also assert that the IDs were rewritten (e.g. `assert.NotEqual(t, 1, withDefaults.ID)`). Otherwise a future regression that turns `AssignFreshSchemaIDs` into a no-op would still pass.

zeroshade (approved):

PR #1167, opened 2026-06-05
refactor(catalog): unify metadata IO operations by passing Table instances directly

xixipi-lining:
**What changes were proposed in this pull request?**
This PR fixes an authentication issue when updating tables. 

Currently, `internal.UpdateAndStageTable` and `internal.WriteMetadata` only use table properties. This causes IO failures for remote storage (like S3) because catalog-level credentials (e.g., `s3.access-key-id`) are lost.

**Fix:**
- Updated `UpdateAndStageTable` to accept `catprops` and merge it with table properties before initializing the filesystem.
- Applied the same property merge to `internal.WriteMetadata`.
- Note: This aligns the update logic with how `CreateStagedTable` already works. `nsprops` is intentionally omitted here since table locations are already fixed during updates.

**How was this patch tested?**
- `make test` passes successfully. 
- *No specific unit test was added for this fix, as our mock filesystems do not validate remote credentials, making it hard to test credential merging effectively. Relying on alignment with `CreateStagedTable` instead.*

### DIFF
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -334,14 +334,16 @@ func (c *Catalog) CommitTable(ctx context.Context, identifier table.Identifier, 
 	}
 
 	// Create a staging table with the updates applied
-	staged, err := internal.UpdateAndStageTable(ctx, current, identifier, requirements, updates, c)
+	staged, err := internal.UpdateAndStageTable(ctx, c.props, current, identifier, requirements, updates, c)
 	if err != nil {
 		return nil, "", err
 	}
 	if current != nil && staged.Metadata().Equals(current.Metadata()) {
 		return current.Metadata(), current.MetadataLocation(), nil
 	}
-	if err := internal.WriteMetadata(ctx, staged.Metadata(), staged.MetadataLocation(), staged.Properties()); err != nil {
+	ioProps := maps.Clone(c.props)
+	maps.Copy(ioProps, staged.Properties())
+	if err := internal.WriteMetadata(ctx, staged.Metadata(), staged.MetadataLocation(), ioProps); err != nil {
 		return nil, "", err
 	}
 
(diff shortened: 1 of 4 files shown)


tanmayrauth (commented):
LGTM.   Thanks for the PR.

  A couple of notes:

 

```
  ioProps := maps.Clone(<catprops>)
  maps.Copy(ioProps, staged.Properties())
```

  is now duplicated in three catalogs in addition to the equivalent merge inside  UpdateAndStageTable itself. Pushing catprops into WriteMetadata's signature  (symmetric with what you've done for UpdateAndStageTable) would keep the merge in one place and drop six lines from the callers. Up to you whether to fold it in here.

  Latent issue, pre-existing in CreateStagedTable: maps.Clone(nil) returns nil, and maps.Copy(nil_map, src) panics if src is non-empty. catalog/sql/sql.go:200 does props: props with no nil-normalization, so NewCatalog(..., nil) followed by a commit on a table with any property would panic. This PR doesn't introduce the issue — it's already in CreateStagedTable — but it does extend the panic surface to commit/update paths. While catalog/internal/utils.go is open, a two-line if ioProps == nil { ioProps = iceberg.Properties{} } in both UpdateAndStageTable and CreateStagedTable would close it. Equally fine to do as a follow-up.

xixipi-lining:
Thanks for the feedback and for catching the  maps.Copy(nil)  latent issue!

I looked into passing  catprops  to  WriteMetadata  as you suggested, and realized we could actually go one step further. Since  CreateStagedTable  and  UpdateAndStageTable  already embed the fully merged IO properties
inside the created  Table 's filesystem closure, we can just pass the underlying  *table.Table  directly to  WriteMetadata .

This avoids coupling  WriteMetadata  to raw properties or even  StagedTable  explicitly. It completely removed the filesystem boilerplate code in both  CreateTable  and  CommitTable  across all catalogs (Glue, Hive, SQL).
I've also added the nil-map protection in  utils.go .

xixipi-lining:
I've updated the PR title. PTAL!  @tanmayrauth

zeroshade (approved):

PR #1168, opened 2026-06-05
fix: preserve nested field IDs when filling missing columns

xixipi-lining:
Fixes a bug where generating a null array for a completely missing nested column (e.g., during schema evolution) accidentally drops the inner `PARQUET:field_id` metadata due to a hardcoded `false` flag. This correctly propagates `includeFieldIDs` and resolves the `record schema does not match writer's` error when writing to `pqarrow`.


tanmayrauth (approved):
LGTM, thanks for the fix.

zeroshade (approved):

PR #1169, opened 2026-06-06
feat(cli): add branch and tag delete commands

### LINKED ISSUE #1165: cli: add branch delete and tag delete commands

The `iceberg branch create` and `iceberg tag create` commands landed in #1068, but there's no symmetric `delete` yet. Once you've made a stale ref via the  CLI, the only way to clean it up is to drop into a catalog client — which is an awkward gap.

  Proposed shape:

      iceberg branch delete <table-id> <branch-name> [--yes]
      iceberg tag delete    <table-id> <tag-name>    [--yes]

  Both should mirror the existing create commands in `cmd/iceberg/branch_tag.go` — confirmation prompt unless `--yes`, error if the ref doesn't exist, and a structured result via the `Output` interface so `--output json` keeps working.

  The underlying API is already there: `MetadataBuilder.RemoveSnapshotRef` in  `table/metadata.go` and `NewRemoveSnapshotRefUpdate` in `table/updates.go`. A handler can mirror `runBranchCreate` and just swap in the remove call.

  One thing worth flagging: `RemoveSnapshotRef` clears `currentSnapshotID` when the removed ref is `main`, which would leave the table headless. The CLI should reject `branch delete <table> main` before ever calling the API.

  @zeroshade @laskoviymishka  this can be a good-first-issue.

Revanth14:
Fixes #1165.

## Summary
- Add `iceberg branch delete <table-id> <branch-name> [--yes]`.
- Add `iceberg tag delete <table-id> <tag-name> [--yes]`.
- Reject deletion of the `main` branch before any catalog commit, so the table can't be left headless (`RemoveSnapshotRef` clears `currentSnapshotID` for `main`).
- Validate that the requested ref exists and matches the expected ref type (`branch delete` refuses a tag, `tag delete` refuses a branch).
- Commit a `RemoveSnapshotRef` update guarded by `AssertTableUUID` + `AssertRefSnapshotID(ref, observedSnapshot)`. Pinning the observed snapshot is intentional: a concurrent move/delete of the ref between load and commit fails the requirement instead of silently removing a different ref state.
- Add structured delete output (`RefDeletedResult`) for both text and JSON modes, so `--output json` keeps working.

## Implementation notes
- The delete handlers mirror the existing `branch create` / `tag create` paths in `cmd/iceberg/branch_tag.go`, sharing `loadTable`, `confirmAction`, and the `Output` interface.
- The shared `osExit` test hook moved from `upgrade_rollback.go` into a dedicated `cmd/iceberg/exit.go`, and the create handlers were migrated onto it so the whole file uses one exit convention and the error paths are testable.

## Testing
- `env GOCACHE=/private/tmp/iceberg-go-gocache go test ./cmd/iceberg -count=1`
- `env GOCACHE=/private/tmp/iceberg-go-gocache go vet ./cmd/iceberg`
- `gofmt -l cmd/iceberg`
- `git diff --check`

### DIFF
--- a/cmd/iceberg/branch_tag.go
+++ b/cmd/iceberg/branch_tag.go
@@ -33,6 +33,8 @@ func runBranch(ctx context.Context, output Output, cat catalog.Catalog, cmd *Bra
 	switch {
 	case cmd.Create != nil:
 		runBranchCreate(ctx, output, cat, cmd.Create)
+	case cmd.Delete != nil:
+		runBranchDelete(ctx, output, cat, cmd.Delete)
 	}
 }
 
@@ -40,6 +42,8 @@ func runTag(ctx context.Context, output Output, cat catalog.Catalog, cmd *TagCmd
 	switch {
 	case cmd.Create != nil:
 		runTagCreate(ctx, output, cat, cmd.Create)
+	case cmd.Delete != nil:
+		runTagDelete(ctx, output, cat, cmd.Delete)
 	}
 }
 
@@ -47,11 +51,11 @@ func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cm
 	tbl := loadTable(ctx, output, cat, cmd.TableID)
 	meta := tbl.Metadata()
 
-	for name := range meta.Refs() {
-		if name == cmd.BranchName {
-			output.Error(fmt.Errorf("ref %q already exists", cmd.BranchName))
-			os.Exit(1)
-		}
+	if _, found := findSnapshotRef(meta, cmd.BranchName); found {
+		output.Error(fmt.Errorf("ref %q already exists", cmd.BranchName))
+		osExit(1)
+
+		return
 	}
 
 	snapshotID := resolveSnapshotID(output, tbl, cmd.SnapshotID)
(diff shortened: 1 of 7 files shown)


Revanth14:
@tanmayrauth 
Can you please review this.

tanmayrauth (commented):
LGTM, thanks for the PR. 

One thing I wanted to check on the delete path: in `runRefDelete` (branch_tag.go:251) the commit pins `AssertRefSnapshotID(refName, &observed)`, which means the delete also fails if someone just appends to   the branch in the meantime — the head moves and the requirement trips, not only on a move/delete of the ref. The summary frames this as protecting against a concurrent move/delete, so I wasn't sure whether  blocking on a plain concurrent commit is intended too. Totally fine if it is — it's safe-by-default and matches the create path — just want to make sure it's deliberate rather than a side effect.

 Other than that I've got nothing blocking. The `"main"` rejection in `runBranchDelete` is the right call given `RemoveSnapshotRef` will otherwise null out `currentSnapshotID` and leave the table headless, so good that it's caught before the commit.

Revanth14:
@tanmayrauth 
Thanks for reviewing. 
Yes, that behavior is intentional, delete should only remove the exact ref state the user observed. If the ref head advances due to a concurrent append, the requirement fails and the user can re-run against the latest state.

I updated the code comment to make that explicit.

tanmayrauth (approved):

zeroshade (approved):

PR #1170, opened 2026-06-06
feat(cli): support rollback by timestamp

### LINKED ISSUE #1166: cli: support rollback by timestamp

`iceberg rollback` (added in #1071) currently requires `--snapshot-id`. In practice, when you're rolling back during an incident, you usually know the  time — "before the bad job at 03:00 UTC" — not the snapshot ID. Looking it up via `iceberg snapshots` first is an extra step you'd rather skip while paged.

  Both PyIceberg (`rollback_to_timestamp`) and Spark Iceberg (its `rollback_to_timestamp` procedure) support this, so it's mostly catching up on parity.

  Proposed:

      iceberg rollback <table-id> --timestamp 2026-01-15T03:00:00Z [--yes]

  `--snapshot-id` and `--timestamp` would be mutually exclusive, exactly one required.

  The change should stay in `cmd/iceberg/upgrade_rollback.go`. The snapshot list is already chronological — `MetadataBuilder.AddSnapshot` enforces that in `table/metadata.go` — so resolving the timestamp is a linear scan from the end. The resolved ID then feeds into the existing `RollbackToSnapshot` path in `table/transaction.go`, which already handles the ancestry check and the `AssertRefSnapshotID` guard, so no library-level changes are needed.

  If no snapshot satisfies `TimestampMs <= t`, the command should error out clearly rather than rolling back to nothing.

@zeroshade  @laskoviymishka  this can be a good-first-issue

Revanth14:
Fixes #1166.

## Summary
- Add `iceberg rollback <table-id> --timestamp <RFC3339>` to roll back to the snapshot current as of a given time.
- Make `--snapshot-id` and `--timestamp` mutually exclusive, exactly one required(`--snapshot-id` goes from required to optional; existing usage is unchanged).
- Resolve the timestamp via the existing `Table.SnapshotAsOf(timestampMs, true)` and feed the resolved ID into the existing rollback flow, preserving the ancestor check, `AssertRefSnapshotID` guard, and confirmation prompt.
- Validate the selector before catalog init, so invalid usage returns a clear message instead of a catalog connection error.

Timestamp resolution follows the snapshot **log** (main lineage), matching PyIceberg/Spark semantics, so a snapshot never on the main lineage isn't selectable by timestamp (intentional), and no snapshot at/before `t` errors clearly rather than rolling back to nothing.

## Testing
- `env GOCACHE=/private/tmp/iceberg-go-gocache go test ./cmd/iceberg -count=1`
- `env GOCACHE=/private/tmp/iceberg-go-gocache go vet ./cmd/iceberg`
- `gofmt -l cmd/iceberg`
- `git diff --check`

### DIFF
--- a/cmd/iceberg/args_test.go
+++ b/cmd/iceberg/args_test.go
@@ -211,6 +211,28 @@ func TestArgsParsing(t *testing.T) {
 			},
 		},
 		{
+			name: "rollback with snapshot id",
+			args: []string{"rollback", "prod.db.events", "--snapshot-id", "123", "--yes"},
+			check: func(t *testing.T, a Args) {
+				require.NotNil(t, a.Rollback)
+				assert.Equal(t, "prod.db.events", a.Rollback.TableID)
+				require.NotNil(t, a.Rollback.SnapshotID)
+				assert.Equal(t, int64(123), *a.Rollback.SnapshotID)
+				assert.Equal(t, "", a.Rollback.Timestamp)
+				assert.True(t, a.Rollback.Yes)
+			},
+		},
+		{
+			name: "rollback with timestamp",
+			args: []string{"rollback", "prod.db.events", "--timestamp", "2026-01-15T03:00:00Z"},
+			check: func(t *testing.T, a Args) {
+				require.NotNil(t, a.Rollback)
+				assert.Equal(t, "prod.db.events", a.Rollback.TableID)
+				assert.Nil(t, a.Rollback.SnapshotID)
+				assert.Equal(t, "2026-01-15T03:00:00Z", a.Rollback.Timestamp)
+			},
+		},
+		{
 			name:    "describe without identifier errors",
 			args:    []string{"describe"},
 			wantErr: true,
--- a/cmd/iceberg/main.go
+++ b/cmd/iceberg/main.go
@@ -272,6 +272,15 @@ func main() {
 		log.Fatal("unimplemented output type")
 	}
 
+	// Validate the rollback selector before catalog init so invalid flags fail
+	// fast with a clear message instead of a catalog connection error.
+	if args.Rollback != nil {
+		if err := validateRollbackSelector(args.Rollback); err != nil {
+			output.Error(err)
+			os.Exit(1)
+		}
+	}
+
 	cat := initCatalog(ctx, args)
 
 	switch {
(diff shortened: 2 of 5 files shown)


Revanth14:
@tanmayrauth 
When you get a chance, could you please review this?

tanmayrauth (commented):
LGTM, the timestamp resolution reusing SnapshotAsOf on the snapshot log is the right call, and the off-lineage / array-order test cases nail down the semantics nicely.

  One small thing and one nit, neither blocking:

  - validateRollbackSelector runs both in main() before catalog init and again at the top of runRollback (upgrade_rollback.go:90). I think this is intentional so runRollback stays self-contained for the direct unit tests — just confirming that's the reasoning rather than a leftover?

  - the --timestamp help text says "RFC3339" but parseRollbackTimestamp uses time.RFC3339Nano (upgrade_rollback.go:198), so it'll also accept fractional seconds. Strictly more lenient, so no harm,  just  flagging the wording in case you want them to match.

  Nice work, especially the subprocess test for validation-before-init.

Revanth14:
> LGTM, the timestamp resolution reusing SnapshotAsOf on the snapshot log is the right call, and the off-lineage / array-order test cases nail down the semantics nicely.
> 
> One small thing and one nit, neither blocking:
> 
> * validateRollbackSelector runs both in main() before catalog init and again at the top of runRollback (upgrade_rollback.go:90). I think this is intentional so runRollback stays self-contained for the direct unit tests — just confirming that's the reasoning rather than a leftover?
> * the --timestamp help text says "RFC3339" but parseRollbackTimestamp uses time.RFC3339Nano (upgrade_rollback.go:198), so it'll also accept fractional seconds. Strictly more lenient, so no harm,  just  flagging the wording in case you want them to match.
> 
> Nice work, especially the subprocess test for validation-before-init.

Thanks for reviewing @tanmayrauth.

Yes, the double validation is intentional:
- `main()` validates before catalog init so invalid selector usage fails with a clear CLI error instead of a catalog connection error.
- `runRollback` keeps the command handler self-contained/testable and protects direct callers of the handler.

I also updated the `--timestamp` help text to mention that fractional seconds are accepted.

tanmayrauth (approved):

zeroshade (approved):

PR #1174, opened 2026-06-08
feat(catalog/rest): endpoint negotiation

rambleraptor:
The REST Catalog sends out a list of supported endpoints via `/v1/config`. This is important for features like views + server-side planning, where the client has to understand if the catalog supports certain features.

I've created a list of endpoints and a mechanism for iceberg-go to error out if a particular method isn't supported by the server. 

Note: This was created with the help of GenAI, since creating all of the endpoint objects is really duplicative work.

### DIFF
--- a/catalog/rest/endpoints.go
+++ b/catalog/rest/endpoints.go
@@ -18,25 +18,21 @@
 package rest
 
 import (
+	"errors"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"strings"
 )
 
-// ErrEndpointNotSupported indicates the server did not advertise a REST endpoint
-// required by the requested operation. It wraps [ErrRESTError].
-var ErrEndpointNotSupported = fmt.Errorf("%w: endpoint not supported by server", ErrRESTError)
+// ErrEndpointNotSupported means the server did not advertise an endpoint the
+// operation needs.
+var ErrEndpointNotSupported = errors.New("endpoint not supported by server")
 
-// pathPrefix is the part of every endpoint template already encoded in the base
-// URI (API version plus optional catalog prefix); reqPath strips it.
 const pathPrefix = "/v1/{prefix}"
 
-// endpoint identifies a REST catalog operation by HTTP method and path template
-// (e.g. "GET /v1/{prefix}/namespaces"). Servers advertise the endpoints they
-// support in the config response so the client can negotiate capabilities.
-//
-// Each endpoint both gates capability and builds the request path (see
-// [endpoint.reqPath]), making it the single source of truth for its operation.
+// endpoint identifies a REST catalog operation by method and path template. It
+// both gates capability and builds the request path.
 type endpoint struct {
 	method string
 	path   string
@@ -44,9 +40,26 @@ type endpoint struct {
 
 func (e endpoint) String() string { return e.method + " " + e.path }
 
-// reqPath renders the request path relative to the base URI, substituting params
-// in order for the "{...}" placeholders that follow [pathPrefix].
-func (e endpoint) reqPath(params ...string) []string {
+// nparams is the number of "{...}" placeholders reqPath expects.
+func (e endpoint) nparams() int {
+	n := 0
+	for _, s := range strings.Split(strings.TrimPrefix(e.path, pathPrefix+"/"), "/") {
+		if strings.HasPrefix(s, "{") && strings.HasSuffix(s, "}") {
+			n++
+		}
+	}
+
+	return n
+}
+
+// reqPath fills the "{...}" placeholders with params in order, returning the
+// path relative to the base URI. It errors on a param count mismatch.
+func (e endpoint) reqPath(params ...string) ([]string, error) {
+	if n := e.nparams(); len(params) != n {
+		return nil, fmt.Errorf("%w: endpoint %s expects %d path parameter(s), got %d",
+			ErrRESTError, e, n, len(params))
+	}
+
 	segs := strings.Split(strings.TrimPrefix(e.path, pathPrefix+"/"), "/")
 
 	out := make([]string, len(segs))
@@ -59,11 +72,10 @@ func (e endpoint) reqPath(params ...string) []string {
 		}
 	}
 
-	return out
+	return out, nil
 }
 
-// endpointFromString parses an endpoint from its "METHOD PATH" wire form,
-// erroring unless the string is exactly two whitespace-separated tokens.
+// endpointFromString parses an endpoint from its "METHOD PATH" wire form.
 func endpointFromString(s string) (endpoint, error) {
 	fields := strings.Fields(s)
 	if len(fields) != 2 {
@@ -104,32 +116,39 @@ var (
 	endpointRegisterView = endpoint{http.MethodPost, "/v1/{prefix}/namespaces/{namespace}/register-view"}
 )
 
-// fallbackEndpoints is assumed when a server advertises no "endpoints" list. It
-// covers every operation the client can invoke, for compatibility with servers
-// that predate endpoint negotiation.
-var fallbackEndpoints = []endpoint{
-	// namespace and table operations
+// defaultEndpoints is the spec default set, assumed when the server advertises
+// no endpoints. The HEAD existence endpoints are excluded on purpose: their
+// absence is what makes Check*Exists fall back to GET.
+var defaultEndpoints = []endpoint{
 	endpointListNamespaces, endpointLoadNamespace, endpointCreateNamespace,
 	endpointUpdateNamespace, endpointDeleteNamespace,
 	endpointListTables, endpointLoadTable, endpointCreateTable,
 	endpointUpdateTable, endpointDeleteTable, endpointRenameTable,
 	endpointRegisterTable, endpointReportMetrics, endpointCommitTransaction,
+}
 
-	// view operations
+// viewEndpoints is added to the default set only when view-endpoints-supported
+// is set.
+var viewEndpoints = []endpoint{
 	endpointListViews, endpointLoadView, endpointCreateView,
 	endpointUpdateView, endpointDeleteView, endpointRenameView,
+}
 
-	// Existence checks and view registration are included for parity with past
-	// behavior. When an advertised set omits a HEAD endpoint, Check*Exists still
-	// degrades to a GET.
-	endpointNamespaceExists, endpointTableExists, endpointViewExists,
-	endpointRegisterView,
+// allEndpoints lists every endpoint constant, for tests that validate templates.
+var allEndpoints = []endpoint{
+	endpointListNamespaces, endpointLoadNamespace, endpointNamespaceExists,
+	endpointCreateNamespace, endpointUpdateNamespace, endpointDeleteNamespace,
+	endpointCommitTransaction,
+	endpointListTables, endpointLoadTable, endpointTableExists, endpointCreateTable,
+	endpointUpdateTable, endpointDeleteTable, endpointRenameTable, endpointRegisterTable,
+	endpointReportMetrics, endpointTableCredentials,
+	endpointListViews, endpointLoadView, endpointViewExists, endpointCreateView,
+	endpointUpdateView, endpointDeleteView, endpointRenameView, endpointRegisterView,
 }
 
 // endpointSet is the set of endpoints a catalog server supports.
 type endpointSet map[endpoint]struct{}
 
-// newEndpointSet builds a set from a list of endpoints.
 func newEndpointSet(eps []endpoint) endpointSet {
 	s := endpointSet{}
 	for _, e := range eps {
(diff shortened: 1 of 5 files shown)


catalog/rest/endpoints.go
```diff
@@ -0,0 +1,173 @@
...
+
+// reqPath renders the request path relative to the base URI, substituting params
+// in order for the "{...}" placeholders that follow [pathPrefix].
+func (e endpoint) reqPath(params ...string) []string {
+	segs := strings.Split(strings.TrimPrefix(e.path, pathPrefix+"/"), "/")
+
+	out := make([]string, len(segs))
+	p := 0
+	for i, s := range segs {
+		if strings.HasPrefix(s, "{") && strings.HasSuffix(s, "}") {
```
laskoviymishka, line 68: There's no bounds check on `params[p]`, so an under-supplied call panics at runtime rather than failing gracefully. The variadic `...string` signature gives us no compile-time arity enforcement either, so a future endpoint with an extra placeholder — or just a miscounted call site — becomes a production panic on the request hot path.

I'd either return an `error` from `reqPath` and propagate it, or panic with a descriptive message naming the endpoint and the expected vs actual param count. Whichever we pick, an `init()` self-check that renders every endpoint constant with its expected arity would catch miscounts at startup instead of on first call. wdyt?

catalog/rest/rest.go
```diff
@@ -796,7 +802,7 @@ func (r *Catalog) fetchTableCreds(ctx context.Context, ident []string, location
 		return nil, err
 	}
 
-	ret, err := doGet[loadCredentialsResponse](ctx, r.baseURI, []string{"namespaces", ns, "tables", tbl, "credentials"},
+	ret, err := doGet[loadCredentialsResponse](ctx, r.baseURI, endpointTableCredentials.reqPath(ns, tbl),
```
laskoviymishka: This is the one call site that got the `reqPath` migration but not the `check()` gate every other op received.

`endpointTableCredentials` isn't in the spec's default set, so against a legacy server this fires an unnegotiated GET, the 404/405 maps to `ErrNoSuchTable`, and the real cause ("server doesn't advertise credentials") gets masked. I'd add `if err := r.endpoints.check(endpointTableCredentials); err != nil { return nil, err }` at the top, matching the other ops.

catalog/rest/endpoints.go
```diff
@@ -0,0 +1,173 @@
...
+	endpointCreateView   = endpoint{http.MethodPost, "/v1/{prefix}/namespaces/{namespace}/views"}
+	endpointUpdateView   = endpoint{http.MethodPost, "/v1/{prefix}/namespaces/{namespace}/views/{view}"}
+	endpointDeleteView   = endpoint{http.MethodDelete, "/v1/{prefix}/namespaces/{namespace}/views/{view}"}
+	endpointRenameView   = endpoint{http.MethodPost, "/v1/{prefix}/views/rename"}
+	endpointRegisterView = endpoint{http.MethodPost, "/v1/{prefix}/namespaces/{namespace}/register-view"}
+)
+
+// fallbackEndpoints is assumed when a server advertises no "endpoints" list. It
+// covers every operation the client can invoke, for compatibility with servers
+// that predate endpoint negotiation.
```
laskoviymishka: The fallback set doesn't match the spec's mandated default. The REST OpenAPI `CatalogConfig` defines exactly 14 endpoints a client MUST assume when the server sends none, and Java's `DEFAULT_ENDPOINTS` matches that exactly; the view group is a separate set merged only when `view-endpoints-supported=true` (default false).

Two divergences matter here. The HEAD existence endpoints (`endpointNamespaceExists`/`TableExists`/`ViewExists`) shouldn't be in fallback at all — their *absence* is what's supposed to trigger the HEAD→GET degradation, so including them defeats the fallback we built in the Check*Exists path. And the whole view CRUD group is merged unconditionally, which tells the client a legacy non-view server supports every view op.

I'd trim fallback to the 14 spec endpoints, drop the three HEAD entries, and move the view endpoints to a separate set gated by `view-endpoints-supported`.

catalog/rest/rest.go
```diff
@@ -834,8 +840,12 @@ func (r *Catalog) listTablesPage(ctx context.Context, namespace table.Identifier
 	if err := checkValidNamespace(namespace); err != nil {
 		return nil, "", err
 	}
+	// Unsupported listing yields an empty result rather than an error.
+	if r.endpoints != nil && !r.endpoints.contains(endpointListTables) {
```
laskoviymishka: The nil-guard semantics are inconsistent across the file and I think it's worth unifying. `check()` already treats a nil set as "permit everything", but the list and Check*Exists paths re-implement that with an inline `r.endpoints != nil && !contains(...)`. Post-`fetchConfig` the set is never nil, so the guard here is effectively dead — and direct construction of the exported `Catalog` struct (bypassing `fetchConfig`) would hit the nil branch and silently behave differently from a negotiated catalog.

I'd add a small `allowed(e) bool { return s == nil || s.contains(e) }` on `endpointSet` so these sites read `if !r.endpoints.allowed(endpointListTables)`, and let it carry the nil semantics in one place. Same pattern applies to `listNamespacesPage` and `listViewsPage`.

catalog/rest/rest.go
```diff
@@ -834,8 +840,12 @@ func (r *Catalog) listTablesPage(ctx context.Context, namespace table.Identifier
...
+		return nil, "", nil
```
laskoviymishka, line 858: Separate from the nil-guard idiom: silently returning an empty list when the endpoint is unsupported is a real footgun for callers. They can't distinguish "this namespace has no tables" from "the server doesn't support listing tables" — which is dangerous for sync/CDC downstreams that treat empty as authoritative.

This matches Java's behavior (PyIceberg instead raises), so I'm not asking to change it. But I'd at least document it in the `ListTables`/`ListNamespaces`/`ListViews` godoc so the empty-on-unsupported contract is explicit. Same applies to the other two list pagers.

catalog/rest/rest.go
```diff
@@ -1327,7 +1394,22 @@ func (r *Catalog) CheckNamespaceExists(ctx context.Context, namespace table.Iden
 		return false, err
 	}
 
-	err := doHead(ctx, r.baseURI, []string{"namespaces", strings.Join(namespace, namespaceSeparator)},
+	// If the server does not advertise the HEAD existence endpoint, fall back to
+	// a GET (load) for compatibility with servers that predate it.
+	if r.endpoints != nil && !r.endpoints.contains(endpointNamespaceExists) {
```
laskoviymishka: The `r.endpoints != nil &&` here conflicts with the fallback design. Java's condition is just `if (endpoints.contains(V1_TABLE_EXISTS))` with no nil special-case — and that works because HEAD endpoints are *absent* from their default set, so absence naturally routes to GET.

Right now nil means "allow everything" in `check()` but "use HEAD" in this path, which is two meanings for the same state. If we trim the HEAD entries out of fallback (see the fallback-set comment), then `!contains(headEP)` alone handles both the nil-negotiated and legacy cases and the guard can drop the nil check. Same for `CheckTableExists` and `CheckViewExists`.

One related catch: when a server omits *both* the HEAD and the GET endpoint, this fallback calls `LoadNamespaceProperties`, which re-checks its own endpoint and can return `ErrEndpointNotSupported` — which then bubbles up from a bool-returning "does it exist" call, since the branch only catches `ErrNoSuchNamespace`. Worth handling that case explicitly too.

catalog/rest/endpoints.go
```diff
@@ -0,0 +1,173 @@
...
+// resolveEndpoints builds the effective endpoint set from the advertised list.
+func resolveEndpoints(advertised []string) endpointSet {
+	s := endpointSet{}
+	for _, raw := range advertised {
+		if e, err := endpointFromString(raw); err == nil {
+			s[e] = struct{}{}
+		}
+	}
+	// A server that advertises nothing (or only unparseable values) gets the
+	// backward-compatibility fallback set.
```
laskoviymishka: There's a subtle asymmetry in the partial-garbage case. If the advertised list is all unparseable we fall back to the open set; but if even one entry parses, we keep just that one and silently drop the rest — yielding a tiny, highly-restricted set where most ops fail. So a server that adds an annotated entry like `"GET /v1/{prefix}/namespaces (deprecated)"` could break nearly everything while looking like a successful negotiation.

I'd `slog.Warn` per dropped entry at minimum so this is debuggable in prod (right now unparseable entries vanish with no trace) — and consider treating a non-empty advertised list as authoritative rather than silently discarding. Today a meaningful new endpoint Go doesn't yet parse just disappears.

catalog/rest/endpoints.go
```diff
@@ -0,0 +1,173 @@
...
+package rest
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+// ErrEndpointNotSupported indicates the server did not advertise a REST endpoint
+// required by the requested operation. It wraps [ErrRESTError].
```
laskoviymishka: `ErrEndpointNotSupported` wraps `ErrRESTError`, which means any caller doing `errors.Is(err, ErrRESTError)` to retry or log transport failures will also catch capability errors — semantically these are different (Java throws `UnsupportedOperationException`, PyIceberg `NotImplementedError`).

Non-blocking, but I'd lean toward a standalone sentinel that doesn't satisfy `errors.Is(ErrRESTError)`. If we keep the wrap, a doc note telling callers to check `ErrEndpointNotSupported` before `ErrRESTError` would help.

catalog/rest/endpoints_test.go
```diff
@@ -0,0 +1,115 @@
...
+			continue
+		}
+		require.NoError(t, err)
+		assert.Equal(t, tc.want, got)
+	}
+}
+
+func TestEndpointRoundTripString(t *testing.T) {
+	t.Parallel()
+
```
laskoviymishka, line 73: The unit suite never actually exercises negotiation, which is the whole point of the PR. The existing `RestCatalogSuite` fixture returns no `endpoints` field, so `resolveEndpoints` always lands in fallback and `check()` can never return an error — CI would stay green even if `check()` were a no-op.

I'd add a mock-based test where `/v1/config` returns a minimal endpoints list and assert the real behavior: `CreateTable` → `ErrEndpointNotSupported`, `ListNamespaces` still succeeds, and an unsupported list op returns empty rather than erroring. And separately, nothing tests `reqPath` substitution itself (e.g. `LoadTable.reqPath("ns","tbl")` == `["namespaces","ns","tables","tbl"]`, `CommitTransaction.reqPath()` == `["transactions","commit"]`, plus the too-few-params case) — given it's the most failure-prone function and a wrong result silently routes every call to the wrong URL, a table-driven test over 0/1/2/5-param endpoints would be worth adding.

laskoviymishka (changes requested):
Nice change — routing all call sites through `endpoint.reqPath` is much cleaner, the per-op `check()` gating reads well, and the round-trip test over `fallbackEndpoints` is a good touch.

I’d hold before merge. Few things here:

The main one is `reqPath`: there’s no bounds check on `params[p]`, so an under-supplied call panics at runtime. Since the signature is variadic, we also get no compile-time arity check. A future endpoint with one more placeholder, or just a miscounted call site, becomes a hot-path request panic. I’d either return an error or panic with a clear message, and add an `init()` self-check that renders every endpoint constant so miscounts fail at startup.

The other big one is spec parity. `fallbackEndpoints` doesn’t match the REST `CatalogConfig` default set: it includes the `HEAD` existence endpoints, whose absence is what should trigger `HEAD` → `GET` degradation, and it pulls in the full view group unconditionally where Java gates it behind `view-endpoints-supported`. I’d trim fallback to the spec defaults, split views out, and drop the `HEAD` entries.

Also, `fetchTableCreds` got the path migration but not the `check()` gate, so a legacy server now returns a confusing `ErrNoSuchTable` instead of a clear unsupported error.

Things I’d settle before merge:

* bounds-check / arity-guard `reqPath`, plus startup self-check
* gate `fetchTableCreds` behind `endpointTableCredentials`
* trim `fallbackEndpoints` to the spec default set; gate views; drop `HEAD`
* unify nil-guard semantics, probably via `allowed()`
* warn instead of silently dropping unparseable advertised endpoints
* add tests for real negotiation and direct `reqPath` coverage


One small thing while to keep hygiene standarts: mind aligning the PR title with the conventional-commit style the rest of the repo uses? Something like `feat(catalog/rest): endpoint negotiation.`

Once those are fixed, happy to take another pass and approve.

catalog/rest/endpoints.go
rambleraptor: Yep, let's return an error. I really don't like panics.

catalog/rest/endpoints.go
rambleraptor: Yep, sounds good. Thanks for that!

catalog/rest/rest.go
rambleraptor: Added it in a comment. We should match it. I'll actually make a PR in Python to change this.

catalog/rest/endpoints.go
rambleraptor: The standalone error sounds great.

catalog/rest/endpoints_test.go
rambleraptor: Yeah, we should absolutely have that. I really don't love mock tests against the REST spec (how do you know your mocks truly match the spec?), but they're definitely needed in this case.

catalog/rest/rest.go
rambleraptor: Cool, made that fix!

rambleraptor:
@laskoviymishka wow thank you for such a thorough review! I really appreciate it. I made those changes, ptal!

zeroshade:
A few small follow-ups from re-reviewing the negotiation tests and `resolveEndpoints`:

**1. Integration test doesn't actually prove negotiation (re: Copilot's still-open comment on `endpoints_integration_test.go:40`)**

Copilot already flagged that `require.NotEmpty(t, cat.endpoints, ...)` always passes, since `resolveEndpoints` never returns an empty set. Worth extending that point: the per-endpoint loop underneath it doesn't close the gap either, because every endpoint it checks (`endpointListNamespaces`, `endpointCreateNamespace`, `endpointListTables`, `endpointCreateTable`, `endpointLoadTable`, `endpointDeleteTable`, `endpointCommitTransaction`) is part of `defaultEndpoints`. So if the live fixture advertised no `endpoints` field, the client would silently fall back, every `contains` check would still pass, and the test's stated goal — "a format mismatch would surface here ... rather than as a silent fallback" — wouldn't actually hold.

To genuinely prove the set came off the wire rather than from fallback, assert `contains` on at least one endpoint that is **not** in `defaultEndpoints` — e.g. `endpointTableExists` (HEAD), `endpointTableCredentials`, or one of the view endpoints — whichever the fixture advertises.

**2. Duplicated endpoint tables (nit)**

`allEndpoints` (`[]endpoint` in `endpoints.go`) and `allEndpointStrings` (`[]string` in `rest_test.go`) encode the same set twice and can drift independently. Consider deriving the test strings from the constants via `.String()` (e.g. build `allEndpointStrings` from `allEndpoints`) so there's a single source of truth.

**3. Partial-parse semantics in `resolveEndpoints` (nit)**

When the advertised list mixes parseable and unparseable entries, the parseable ones are kept and the rest dropped (now logged via `slog.Warn` — 👍). The edge case worth documenting: a single garbled or annotated entry — e.g. `"GET /v1/{prefix}/namespaces (deprecated)"` — alongside otherwise-valid entries yields a tiny, highly-restricted set where most ops are gated off, while still looking like a successful negotiation. A short doc note on `resolveEndpoints` stating that a non-empty advertised list is treated as authoritative (so parse failures reduce capability) would make that contract explicit.

rambleraptor:
Hey @zeroshade thanks a lot for the review! I made the (improper) decision not to look at the Copilot reviews. I'm in some other OSS projects where they end up adding a lot of noise. I'll be sure to look at those for my future iceberg-go PRs!

Agree on the test. Made all three of the changes. ptal!

zeroshade (approved):
LGTM!

zeroshade:
I'll wait for @laskoviymishka to give the okay before we merge since he has an outstanding "request changes" on this

laskoviymishka (approved):
LGTM now!
Thanks!

PR #1180, opened 2026-06-10
fix(manifest): allow NewManifestReader to read lower-version manifests

twuebi:
currently, a manifest-list with version 2 referencing a manifest with version 1 fails to be read


twuebi:
@laskoviymishka

laskoviymishka (approved):
Nice catch!
🚢 👍

PR #1182, opened 2026-06-11
test(table): validate variant and unknown types against Spark 4

### LINKED ISSUE #1139: feat(test): add Spark 4.0 to integration test matrix

nssalian:
Closes [#1139](https://github.com/apache/iceberg-go/issues/1139) 
Follow-up to [#1140](https://github.com/apache/iceberg-go/pull/1140)

### Summary

- New `recipe.SparkMajorVersion` helper - extracts the running container's pyspark version inline (`python -c`)
- New `requireSpark4()` test-suite helper - skips for Spark 3.5.
- `TestVariantWriteAndScan` now also asserts Spark 4 reads the rows via `to_json` after the existing Go-side scan.
- New `TestUnknownTypeWriteAndScan` - Go writes an `UnknownType` column, Spark 4 confirms via `DESCRIBE` that the column lands as `void` (Spark's `NullType`).

### Test plan

Ran the following locally:

- `make integration-setup && make integration-spark` (Spark 3.5): 10 PASS, 2 SKIP (variant + unknown with the message).
- `make integration-setup-spark4 && make integration-spark` (Spark 4.0): 12 PASS.


table/transaction_test.go
```diff
@@ -73,6 +73,15 @@ func (s *SparkIntegrationTestSuite) SetupTest() {
...
+	s.T().Helper()
+	major, err := recipe.SparkMajorVersion(s.T())
+	s.Require().NoError(err, "spark version extraction failed")
+	if major < 4 {
+		s.T().Skipf("requires Spark 4+ (running Spark %d)", major)
+	}
+}
+
 func (s *SparkIntegrationTestSuite) TestSetProperties() {
 	icebergSchema := iceberg.NewSchema(0,
```
laskoviymishka, line 86: The skip lands after we've created the table, written, committed, and scanned — so on Spark 3.5 this isn't really skipped, it does all the work and then bails.

Two problems: the Go-side assertions above run under a test whose skip message says "requires Spark 4+", which is misleading; and the write side effects are already committed, so a rerun on 3.5 can hit `CreateTable` on an existing `go_variant_events`/`go_unknown_table` and fail with a conflict. I'd move `s.requireSpark4()` to the first line of both this test and `TestUnknownTypeWriteAndScan` (line 144), so nothing executes before the skip decision.

internal/recipe/local_spark.go
```diff
@@ -140,3 +142,33 @@ func ExecuteSpark(t *testing.T, scriptPath string, args ...string) (string, erro
...
 	return string(output), nil
 }
+
+var (
+	sparkMajorOnce sync.Once
+	sparkMajor     int
+	sparkMajorErr  error
+)
+
+// SparkMajorVersion returns the major pyspark version (3 or 4) of the
```
laskoviymishka, line 154: Two things tangled together here, both rooted in caching the result keyed off `t`.

First, `sync.Once` caches the error permanently. If the container is momentarily unavailable on the first call — docker-exec startup timing is flaky — `sparkMajorErr` is set for the whole process and every Spark-4-gated test afterward fails with "extract spark version", masking the real cause. I'd only cache on success: either drop the `Once` and accept one exec per test, or guard with a mutex that leaves the result unset on error so the next caller retries.

Second, the closure captures the *first* caller's `*testing.T` and uses its context/logging for an exec that later tests depend on. If that first test has returned, calling its `t` methods is a data race / "log after test finished" panic. I'd detach detection from `t` entirely — drive the one-time check off `context.Background()` with a raw exec helper, make `SparkMajorVersion()` take no `t`, and have call sites become `recipe.SparkMajorVersion()`. wdyt?

laskoviymishka (changes requested):
Nice coverage, Spark 4 cross-validation for variant/unknown is something we don't get elsewhere. 

Two things to fix before merge (inline): move requireSpark4 to the first line of both tests so we don't commit side effects and then skip on 3.5, and detach the sync.Once helper from *testing.T + stop caching its error. 

Happy to approve once those are in.

table/transaction_test.go
nssalian: Agreed. I moved `requireSpark4()` to the first line of both tests and it should resolve the rerun-conflict on `go_variant_events` / `go_unknown_table`

internal/recipe/local_spark.go
nssalian: Was able to fix this.
switched to a mutex that only writes `sparkMajor` inside the success branches,  so transient exec failures leave it at 0 and the next caller retries.  On the race I couldn't reproduce it (`sync.Once.Do` runs the closure once during the first caller's lifetime, `ExecuteSpark` has no goroutines, so the captured `t` is held but never re-invoked) but the decoupling is still the right cleanup.

laskoviymishka (approved):
<3

PR #1183, opened 2026-06-11
fix(table): apply positional deletes with batch-local Take indices

jeremybarner:
## Problem

`processPositionalDeletes` applies positional (merge-on-read) deletes one Arrow record batch at a time, but `combinePositionalDeletes` builds the surviving-row indices in **global, file-relative** coordinates `[start, end)` and passes them straight into `compute.Take` against the **current batch**, whose valid index range is `[0, NumRows)`.

For the first batch of a data file `start == 0`, so it works. For the second and later batches `start > 0`, every index is `>= NumRows` and the scan fails:

```
index error: <N> out of bounds
```

where `<N>` is a multiple of the parquet read batch size (`read.parquet.batch-size`, default `1<<17 == 131072`). So any data file larger than one batch that also has positional delete files fails to scan at its second batch.

## Fix

Rebase the Take indices to batch-local coordinates (`i - start`). The `deletes` set stays in global coordinates because it is matched against the global position `i`.

```go
for i := start; i < end; i++ {
    if _, ok := deletes[i]; !ok {
        bldr.Append(i - start)
    }
}
```

## Test

Adds `TestProcessPositionalDeletesAcrossBatches`, which feeds two consecutive batches with a delete located in the **second** batch — the case the previous code got wrong. The test fails with `index error: 4 out of bounds` before the fix and passes after.

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -304,13 +304,20 @@ func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_
 
 type set[T comparable] map[T]struct{}
 
+// combinePositionalDeletes builds the surviving-row index list for a single record
+// batch. The deletes set holds file-relative (global) positions, and [start, end) is
+// the global position span the batch covers, so the indices are rebased to batch-local
+// coordinates (i-start): they index into the batch handed to compute.Take, not into the
+// whole file. Without the rebase, the second and later batches of a file would pass
+// indices >= the batch length and compute.Take would fail with "index error: N out of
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM. The fix is right — the survivor indices feed `compute.Take` against the per-batch array, so they have to be batch-local (`i - start`), while the `deletes` membership check correctly stays on the global `i`. 

Nice that the DV path in `filterByDeletionVector` already slices per-batch, so there's no twin bug there.

zeroshade (approved):

PR #1185, opened 2026-06-11
refactor(catalog): Minimize functions in interface for Hadoop catalog

C-Loftus:
Addresses https://github.com/apache/iceberg-go/issues/1096 (as the next step; blob storage support is out of scope for this PR and would be present in a follow up)

- Removes interfaces for `MkdirIO` and `ReadDirIO` to prepare for hadoop cloud storage backends.
- Refactors hadoop catalog to use `MkdirAll` and `WalkDir` in their place.
    - Cleans up the comments for those interfaces to reflect current usage. 
- Adds on new helper function `CheckedMkdirAll`, which before calling `MkdirAll`, walks the directory and ensures that each subdirectory of the full path exists. 
    - Currently this is only used once since only one previous `Mkdir` call checked for errors if the directory already existed. That being said, I think it helps to abstract logic a bit and I'm thinking that it likely may be used in other parts of the Hadoop catalog.

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -793,5 +793,6 @@ func (c *Catalog) CheckedMkdirAll(id table.Identifier) error {
 		// it means that it must already exist
 		return fs.ErrExist
 	}
+
 	return c.filesystem.MkdirAll(path)
 }
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -363,11 +363,7 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 	}
 
 	metaDir := c.metadataDir(ident)
-	if err := c.validSubpaths(ident, false); err != nil {
-		return nil, err
-	}
-
-	if err := c.filesystem.MkdirAll(metaDir); err != nil {
+	if err := c.CheckedMkdirAll(ident, false); err != nil {
 		return nil, fmt.Errorf("hadoop catalog: failed to create metadata directory: %w", err)
 	}
 
@@ -493,10 +489,7 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs 
 
 	// Step 7: Create metadata directory if needed (create-via-commit).
 	metaDir := c.metadataDir(ident)
-	if err := c.validSubpaths(ident, false); err != nil {
-		return nil, "", err
-	}
-	if err := c.filesystem.MkdirAll(metaDir); err != nil {
+	if err := c.CheckedMkdirAll(ident, false); err != nil {
 		return nil, "", fmt.Errorf("hadoop catalog: failed to create metadata directory: %w", err)
 	}
 
@@ -631,16 +624,20 @@ func (c *Catalog) CreateNamespace(_ context.Context, ns table.Identifier, props 
 		return errors.New("hadoop catalog: namespace properties are not supported")
 	}
 
-	path := c.namespaceToPath(ns)
-
 	// Raise an error if the namespace already exists
-	if err := c.validSubpaths(ns, true); err != nil {
-		return err
-	}
+	if err := c.CheckedMkdirAll(ns, true); err != nil {
+		if errors.Is(err, fs.ErrExist) {
+			return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(ns, "."))
+		}
 
-	if err := c.filesystem.MkdirAll(path); err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return fmt.Errorf("%w: parent namespace does not exist for %s",
+				catalog.ErrNoSuchNamespace, strings.Join(ns, "."))
+		}
+
 		return fmt.Errorf("hadoop catalog: failed to create namespace: %w", err)
 	}
+
 	return nil
 }
 
@@ -776,28 +773,23 @@ func (c *Catalog) UpdateNamespaceProperties(_ context.Context, _ table.Identifie
 	return catalog.PropertiesUpdateSummary{}, errors.New("hadoop catalog: UpdateNamespaceProperties not yet implemented")
 }
 
-// Helper function for checking that all subpaths in a path exist and that the
-// final path does not already exist.
-// this means that all parent namespaces must already exist, and the target namespace
-// itself must not already exist; this should be used before calls like MkdirAll
-// which don't raise errors if the target already exists.
-func (c *Catalog) validSubpaths(ns table.Identifier, errorIfExists bool) error {
-	path := c.namespaceToPath(ns)
-	for pathIndex := range ns {
-		subPath := ns[:pathIndex]
+// CheckedMkdirAll is a helper function that checkes
+// all subdirectories of a given identifier path exist before creting the full path.
+// If errIfExists is true, it also checks that the full path does not already exist before creating it.
+func (c *Catalog) CheckedMkdirAll(id table.Identifier, errIfExists bool) error {
+	path := c.namespaceToPath(id)
+	for pathIndex := range id {
+		subPath := id[:pathIndex]
 		parentPath := c.namespaceToPath(subPath)
-		_, err := c.filesystem.Stat(parentPath)
-		if errors.Is(err, fs.ErrNotExist) {
-			return fmt.Errorf("%w: parent namespace does not exist for %s", catalog.ErrNoSuchNamespace, strings.Join(subPath, "."))
-		}
-		if err != nil {
-			return fmt.Errorf("hadoop catalog: failed to stat parent namespace: %w", err)
+		if _, err := c.filesystem.Stat(parentPath); err != nil {
+			return err
 		}
 	}
-	if errorIfExists {
+	if errIfExists {
 		if _, err := c.filesystem.Stat(path); err == nil {
-			return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(ns, "."))
+			return err
 		}
 	}
-	return nil
+	return c.filesystem.MkdirAll(path)
+
 }
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -215,7 +215,6 @@ func isTableDir(filesystem HadoopCatalogFS, path string) bool {
 
 		return nil
 	})
-
 	if err != nil {
 		return false
 	}
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -188,7 +188,7 @@ func (c *Catalog) defaultTableLocation(ident table.Identifier) string {
 func isTableDir(filesystem HadoopCatalogFS, path string) bool {
 	metaDir := filepath.Join(path, "metadata")
 
-	found_metadata := false
+	foundMetadata := false
 	err := filesystem.WalkDir(metaDir, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
@@ -208,7 +208,7 @@ func isTableDir(filesystem HadoopCatalogFS, path string) bool {
 		if versionPattern.MatchString(name) ||
 			uuidMetadataPattern.MatchString(name) ||
 			name == "version-hint.text" {
-			found_metadata = true
+			foundMetadata = true
 
 			return fs.SkipAll
 		}
@@ -216,11 +216,11 @@ func isTableDir(filesystem HadoopCatalogFS, path string) bool {
 		return nil
 	})
 
-	if err != nil && err != fs.SkipAll {
+	if err != nil {
 		return false
 	}
 
-	return found_metadata
+	return foundMetadata
 }
 
 func (c *Catalog) readVersionHint(ident table.Identifier) int {
@@ -566,13 +566,13 @@ func (c *Catalog) ListTables(_ context.Context, ns table.Identifier) iter.Seq2[t
 
 			// Skip anything that is not a table directory.
 			if !isTableDir(c.filesystem, path) {
-				return nil
+				return fs.SkipDir
 			}
 			ident := make(table.Identifier, len(ns)+1)
 			copy(ident, ns)
 			ident[len(ns)] = d.Name()
 			if !yield(ident, nil) {
-				return nil
+				return fs.SkipAll
 			}
 
 			return nil
@@ -627,7 +627,7 @@ func (c *Catalog) CreateNamespace(_ context.Context, ns table.Identifier, props 
 	}
 
 	// Raise an error if the namespace already exists
-	if err := c.CheckedMkdirAll(ns, true); err != nil {
+	if err := c.CheckedMkdirAll(ns); err != nil {
 		if errors.Is(err, fs.ErrExist) {
 			return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(ns, "."))
 		}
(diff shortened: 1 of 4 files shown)


C-Loftus:
Fixing lint issues now

catalog/hadoop/hadoop.go
```diff
@@ -526,29 +549,38 @@ func (c *Catalog) ListTables(_ context.Context, ns table.Identifier) iter.Seq2[t
...
 			ident := make(table.Identifier, len(ns)+1)
 			copy(ident, ns)
-			ident[len(ns)] = e.Name()
+			ident[len(ns)] = d.Name()
 			if !yield(ident, nil) {
-				return
+				return nil
 			}
+
+			return nil
```
zeroshade: `ListTables`'s `WalkDir` callback returns `nil` in every branch, so it never stops descending (unlike `ListNamespaces` below, which uses `fs.SkipDir`). Both reproduced locally:

- **Wrong results:** it recurses into child namespaces and yields nested tables under a flattened identifier — `ns/child/tbl` comes back as `["ns","tbl"]`.
- **Panic on early `break`:** after `!yield` returns false, `return nil` keeps the walk going and calls `yield` again → `range function continued iteration after function for loop body returned false`.
- **Cost:** walks each table's entire `data/`/`metadata/` subtree just to list names — expensive locally, much worse on the cloud backends this PR targets.

Mirror `ListNamespaces`: return `fs.SkipDir` for the non-table dir (L569) and after yielding a table (L578), and `fs.SkipAll` when `!yield` (L575).

catalog/hadoop/hadoop.go
```diff
@@ -723,3 +773,26 @@ func (c *Catalog) LoadNamespaceProperties(_ context.Context, ns table.Identifier
 func (c *Catalog) UpdateNamespaceProperties(_ context.Context, _ table.Identifier, _ []string, _ iceberg.Properties) (catalog.PropertiesUpdateSummary, error) {
 	return catalog.PropertiesUpdateSummary{}, errors.New("hadoop catalog: UpdateNamespaceProperties not yet implemented")
 }
+
+// CheckedMkdirAll is a helper function that checkes
+// all subdirectories of a given identifier path exist before creting the full path.
+// If errIfExists is true, it also checks that the full path does not already exist before creating it.
+func (c *Catalog) CheckedMkdirAll(id table.Identifier, errIfExists bool) error {
```
zeroshade: A few small things on this helper:
- Consider unexporting (`checkedMkdirAll`) — internal helper, no need to widen `*Catalog`'s public API.
- The `errIfExists=false` path has no caller and no test; drop the param or add coverage.
- `Stat`-then-`MkdirAll` isn't atomic like the old single `os.Mkdir`, so the `ErrNamespaceAlreadyExists` guarantee is now best-effort under concurrent `CreateNamespace`. Likely fine, but worth a note.

catalog/hadoop/hadoop.go
```diff
@@ -188,25 +188,39 @@ func (c *Catalog) defaultTableLocation(ident table.Identifier) string {
 func isTableDir(filesystem HadoopCatalogFS, path string) bool {
 	metaDir := filepath.Join(path, "metadata")
 
-	entries, err := filesystem.ReadDir(metaDir)
-	if err != nil {
-		return false
-	}
+	found_metadata := false
```
zeroshade: Non-idiomatic snake_case — rename to `foundMetadata` (matches `foundEntries` added elsewhere in this PR).

catalog/hadoop/hadoop.go
```diff
@@ -188,25 +188,39 @@ func (c *Catalog) defaultTableLocation(ident table.Identifier) string {
...
-			return true
+			found_metadata = true
+
+			return fs.SkipAll
 		}
+
+		return nil
+	})
+
+	if err != nil && err != fs.SkipAll {
```
zeroshade: `LocalFS.WalkDir` delegates to `filepath.WalkDir`, which consumes `fs.SkipAll` and returns `nil`, so `&& err != fs.SkipAll` is unreachable. Simplify to `if err != nil { return false }`.

catalog/hadoop/hadoop.go
```diff
@@ -723,3 +773,26 @@ func (c *Catalog) LoadNamespaceProperties(_ context.Context, ns table.Identifier
 func (c *Catalog) UpdateNamespaceProperties(_ context.Context, _ table.Identifier, _ []string, _ iceberg.Properties) (catalog.PropertiesUpdateSummary, error) {
 	return catalog.PropertiesUpdateSummary{}, errors.New("hadoop catalog: UpdateNamespaceProperties not yet implemented")
 }
+
+// CheckedMkdirAll is a helper function that checkes
```
zeroshade: Typos: "checkes" → "checks", "creting" → "creating".

zeroshade (changes requested):
Direction looks good — narrowing the filesystem interface is the right prep for #1096, and the namespace-creation refactor preserves behavior. One blocker before merge: `ListTables` needs a fix — its `WalkDir` callback never returns `SkipDir`/`SkipAll`, which yields wrong results, panics on an early `break`, and walks whole table subtrees. Details inline; the rest are minor.

catalog/hadoop/hadoop.go
C-Loftus: Thank you very much for catching this. Could you please clarify what you mean by `Both reproduced locally`? Did you run this code with a sample script? At least from what I can see in both my tests and CI on github, this code did not raise any errors in the test suite. Should another test be added to prevent regressions on this?

C-Loftus:
Believe I have addressed all feedback. I am not entirely clear why CI is failing. It says something needs to be formatted but I have all the code formatted in the latest commit and there are no issues locally, nor any changes if I run the formatted again. I am running with go 1.26 so I don't think its due to differences in gofmt. Could someone please advise on the best way to fix this? Is this some caching issues in CI? 

```
> go version
go version go1.26.4 darwin/arm64
> go fmt ./catalog/
> golangci-lint run --fix ./catalog/
0 issues.
> git diff
>

**Comment by C-Loftus** [comment]
Resolved my previous CI mixup, should be good to review again now, thanks!

File: catalog/hadoop/hadoop.go [hunk 6]
```diff
@@ -723,3 +772,26 @@ func (c *Catalog) LoadNamespaceProperties(_ context.Context, ns table.Identifier
...
 	return catalog.PropertiesUpdateSummary{}, errors.New("hadoop catalog: UpdateNamespaceProperties not yet implemented")
 }
+
+// CheckedMkdirAll is a helper function that checks
+// all subdirectories of a given identifier path exist before creating the full path.
+// This function is not atomic and may not return ErrNamespaceAlreadyExists if
+// called concurrently with other calls that change the same identifier
+func (c *Catalog) CheckedMkdirAll(id table.Identifier) error {
+	path := c.namespaceToPath(id)
+	for pathIndex := range id {
```
laskoviymishka: I think there's a real bug in this loop. `pathIndex` starts at 0, so the first iteration uses `id[:0]` — the empty identifier — and stats the warehouse root rather than the first namespace component. For a single-component namespace like `["ns"]` we never verify any actual parent, and we redundantly stat the warehouse on every call.

I'd start at 1: `for pathIndex := 1; pathIndex < len(id); pathIndex++`. `TestCreateNamespaceNestedParentMissing` only passes today because the warehouse always exists in the fixture, so the intended invariant ("all parent namespaces must exist") isn't really being checked. wdyt?

catalog/hadoop/hadoop.go
```diff
@@ -723,3 +772,26 @@ func (c *Catalog) LoadNamespaceProperties(_ context.Context, ns table.Identifier
...
+		subPath := id[:pathIndex]
+		parentPath := c.namespaceToPath(subPath)
+		if _, err := c.filesystem.Stat(parentPath); err != nil {
```
laskoviymishka, line 795: Returning the raw `Stat` error here breaks the `ErrNoSuchNamespace` contract that `CreateNamespace` used to hold. The caller only catches `fs.ErrExist`; everything else — including the `fs.ErrNotExist` from a missing parent — falls through to the generic `fmt.Errorf`, so the error chain no longer contains `catalog.ErrNoSuchNamespace`.

The old `CreateNamespace` had an explicit `if errors.Is(err, fs.ErrNotExist) { ... catalog.ErrNoSuchNamespace ... }` branch that this PR drops. I'd re-add that branch in `CreateNamespace` (alongside the existing `ErrExist` check), which covers both the intermediate-parent and missing-warehouse-root cases in one spot. It'd also help to record which component failed so the message isn't misleading.

catalog/hadoop/hadoop.go
```diff
@@ -723,3 +772,26 @@ func (c *Catalog) LoadNamespaceProperties(_ context.Context, ns table.Identifier
 func (c *Catalog) UpdateNamespaceProperties(_ context.Context, _ table.Identifier, _ []string, _ iceberg.Properties) (catalog.PropertiesUpdateSummary, error) {
 	return catalog.PropertiesUpdateSummary{}, errors.New("hadoop catalog: UpdateNamespaceProperties not yet implemented")
 }
+
+// CheckedMkdirAll is a helper function that checks
+// all subdirectories of a given identifier path exist before creating the full path.
+// This function is not atomic and may not return ErrNamespaceAlreadyExists if
+// called concurrently with other calls that change the same identifier
+func (c *Catalog) CheckedMkdirAll(id table.Identifier) error {
```
laskoviymishka: I'd unexport this to `checkedMkdirAll`. The only caller is `CreateNamespace` in this same package, it bypasses `validateIdentifier`, and exporting it bakes an internal implementation detail into the stable public API right as we're trying to slim the surface.

laskoviymishka (changes requested):
Nice fix, the interface narrows cleanly and the ReadDir→WalkDir conversions preserve the depth-1 semantics. 

Holding on CheckedMkdirAll though: the loop starts at index 0 (stats the warehouse, not the first parent), and it drops the catalog.ErrNoSuchNamespace wrapping CreateNamespace used to guarantee. Details inline.

Once fixed - happy to approve.

catalog/hadoop/hadoop.go
C-Loftus: Hmm sorry I am not following this.

> The old CreateNamespace had an explicit if errors.Is(err, fs.ErrNotExist) { ... catalog.ErrNoSuchNamespace ... } branch that this PR drops. I'd re-add that branch in CreateNamespace (alongside the existing ErrExist check)

I believe I currently have a check for `fs.ErrNotExist` which returns an error with `catalog.ErrNoSuchNamespace` in the result. Could you please clarify what you would like me to change beyond that?

https://github.com/C-Loftus/iceberg-go/blob/3a7d309d7efd5c216991adae0263e90f76719334/catalog/hadoop/hadoop.go#L629-L640

```go
	if err := c.checkedMkdirAll(ns); err != nil {
		if errors.Is(err, fs.ErrExist) {
			return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(ns, "."))
		}

		if errors.Is(err, fs.ErrNotExist) {
			return fmt.Errorf("%w: parent namespace does not exist for %s",
				catalog.ErrNoSuchNamespace, strings.Join(ns, "."))
		}

		return fmt.Errorf("hadoop catalog: failed to create namespace: %w", err)
	}
```

catalog/hadoop/hadoop.go
C-Loftus: Changed

catalog/hadoop/hadoop.go
C-Loftus: Done; sorry this was just a typo on my end. Meant to do that in previous commit

catalog/hadoop/hadoop.go
laskoviymishka: My bad. I misread this. The fs.ErrNotExist branch you linked already catches the missing-parent case and wraps it in catalog.ErrNoSuchNamespace, so the contract is held. Nothing to change here, sorry for the noise

C-Loftus: All good no worries thanks for confirming! 👍

catalog/hadoop/hadoop.go
```diff
@@ -619,7 +649,20 @@ func (c *Catalog) DropNamespace(_ context.Context, ns table.Identifier) error {
 
 	path := c.namespaceToPath(ns)
 
-	entries, err := c.filesystem.ReadDir(path)
+	foundEntries := false
+	err := c.filesystem.WalkDir(path, func(p string, d fs.DirEntry, err error) error {
```
laskoviymishka: I think the ReadDir→WalkDir swap dropped a guard here. If a plain file sits at `warehouse/<name>`, WalkDir calls the callback once with `p == path`, we return nil, `foundEntries` stays false, and we fall through to `c.filesystem.Remove(path)` — so we silently delete the file and return success. The old `ReadDir(path)` returned ENOTDIR on a non-dir and the delete never happened.

I'd Stat the path up front and bail with `ErrNoSuchNamespace` if it isn't a directory, before the walk.

A test that pins it: write a plain file at `warehouse/notadir`, call `DropNamespace(["notadir"])`, assert it errors and the file is still on disk.

catalog/hadoop/hadoop.go
```diff
@@ -526,29 +548,38 @@ func (c *Catalog) ListTables(_ context.Context, ns table.Identifier) iter.Seq2[t
...
 			ident := make(table.Identifier, len(ns)+1)
 			copy(ident, ns)
-			ident[len(ns)] = e.Name()
+			ident[len(ns)] = d.Name()
 			if !yield(ident, nil) {
-				return
+				return fs.SkipAll
 			}
+
+			return nil
```
laskoviymishka: The SkipDir/SkipAll fix here looks right — child namespaces aren't flattened anymore and the early break no longer panics. To answer your question on zeroshade's thread: yes, I'd add a regression test so we can close it confidently.

`TestListTablesDoesNotYieldTablesInChildNamespace`: create `ns/tbl1` and `ns/child_ns/tbl2`, then assert `ListTables(["ns"])` returns exactly `[["ns","tbl1"]]`. That's the exact shape zeroshade reproduced, and it'd catch any future regression to the SkipDir branch.

One thing still here: after yielding a table we `return nil`, so WalkDir descends one level into each table's own `data/`/`metadata/` before SkipDir-ing them. Harmless for correctness, just extra reads — `return fs.SkipDir` after the yield avoids it. wdyt?

laskoviymishka (changes requested):
Thanks, that's a nice refactoring.

One thing I'd wait till merge: the ReadDir→WalkDir swap in DropNamespace lost its directory guard. If a plain file sits at warehouse/<name>, the walk finds no entries and we fall through to os.Remove(path) — silently deleting the file and returning success, where ReadDir used to return ENOTDIR. I'd Stat-and-IsDir-guard before the walk, with a test (file at warehouse/notadir, DropNamespace must error and leave it intact).

And would be nice to close loose ends on zeroshade's thread: the fix is in, it just needs the regression test — ns/tbl1 + ns/child_ns/tbl2, assert ListTables(["ns"]) returns exactly [["ns","tbl1"]].

Fix those two and I'm happy to approve.

catalog/hadoop/hadoop.go
C-Loftus: Done, thanks for the great feedback!

catalog/hadoop/hadoop.go
C-Loftus: Added a test and made the requested `fs.SkipDir` change.

I believe this should reflect it but if not please let me know.
https://github.com/C-Loftus/iceberg-go/blob/fbccce9a2201bccd857f319b46174c574b5e61a8/catalog/hadoop/hadoop.go#L572-L578

catalog/hadoop/hadoop.go
C-Loftus: Added tests in  for this in https://github.com/apache/iceberg-go/pull/1185/commits/fbccce9a2201bccd857f319b46174c574b5e61a8 so resolving this comment

laskoviymishka (approved):
:shipit: , will wait @zeroshade till merge

zeroshade (approved):
LGTM

Issue #1178, opened 2026-06-09
Epic: REST Catalog Scan Planning (server-side scan planning) client support


laskoviymishka:
Bring iceberg-go to parity with Java and Python on REST server-side scan planning. The endpoints were added to the Iceberg REST OpenAPI spec by apache/iceberg#9695 (merged 2024-09-10) and shipped in Java through 2025 (apache/iceberg#13004, #13400, #14480, #14660, #14822, #15184, #15863, #16024, #16197).

Today iceberg-go has zero support: no types, no client methods, no scanner wiring, no capability discovery. `(*Scan).PlanFiles` reads manifests locally regardless of what the catalog can offer.

## What this unlocks

When this lands, Go consumers of an Iceberg REST catalog will be able to:

- Scan tables of any size without downloading manifests locally
- Run stateless from Lambda / Cloud Run / short-lived k8s pods (no manifest cache)
- Enforce catalog-side governance (RLS, masking, time-travel restrictions) on Go clients
- Receive per-scan vended `StorageCredentials` with plan-scoped TTLs
- Do cheap incremental scans for CDC consumers via `start-snapshot-id` / `end-snapshot-id`
- Drive the iceberg-go CLI against production-scale tables (`iceberg files`, `partition-stats`, `clean-orphan-files`, `expire-snapshots`, `compact`)

## Scope of this epic

This is a client-only effort. Non-goals:

- Implementing the server side of the spec (the REST catalog server lives in Java).
- A distributed worker runtime. We expose `FetchScanTasks` so external engines can fan out, but we don't build the engine.
- Reworking the local planning path. Default mode stays `local`; remote planning is strictly opt-in, no breaking change.

## Correctness contract

The acceptance test for the whole effort is the **local-vs-remote parity test**: the same scan (same filter, same table, same snapshot) under `scan-planning-mode=local` and `scan-planning-mode=remote` must produce the same `FileScanTask` set (paths, ranges, deletes, residuals). If it doesn't, the expression JSON codec is wrong and silent result corruption — not a loud error — is the failure mode. That codec is the highest-risk piece of the project.

## Proposed decomposition

[...]

PR #1191, opened 2026-06-12
fix: manifest pruning when partition source columns are dropped

### LINKED ISSUE #1190: fix: manifest pruning when partition source columns are dropped
### Apache Iceberg version

main (development)

## Problem

Manifest-level pruning can read the wrong partition field summary after a partition source column is dropped from the current schema.

Example:

A table is partitioned by two fields in this order:

```text
q_part, p_part
```

Manifest files store partition summaries positionally, so:

```text
slot 0 = q_part summary
slot 1 = p_part summary
```

If the table schema later drops column `q`, iceberg-go currently builds a compacted partition schema that only contains:

```text
slot 0 = p_part
```

But old manifest files still store summaries in the original order:

```text
slot 0 = q_part
slot 1 = p_part
```

So a query like:

```sql
WHERE p = 2
```

can accidentally compare `2` against `q_part`'s summary instead of `p_part`'s summary. If `q_part` has bounds like `10..20`, the manifest evaluator may conclude that `p = 2` cannot match and skip the manifest, even though matching data files exist.

## Expected Behavior

Dropping a source column should not shift the positions of later partition fields during manifest pruning.

Dropped-source partition fields should stay as inert placeholders, so manifest summary positions remain aligned with the partition spec:

```text
slot 0 = dropped q_part placeholder
slot 1 = p_part
```

This matches Java Iceberg's partition type behavior, where dropped partition-source fields are represented with `UnknownType` in the partition type.

## Impact

This can cause incorrect query results because matching manifests may be pruned incorrectly.

## Proposed Fix

When building the partition schema used for manifest evaluation, keep every partition field from the partition spec. If the source column no longer exists in the scan schema, use an `UnknownType` placeholder for that partition field instead of dropping it.

Add a regression test with:

- partition spec: `identity(q), identity(p)`
- current schema after evolution: `q` dropped, `p` still present
- row filter: `p = 2`, projected to the partition filter for `p_part`

Before the fix, the matching manifest can be incorrectly pruned. After the fix, the matching file is returned.

rraulinio:
## Summary

Closes https://github.com/apache/iceberg-go/issues/1190.

Fix manifest-level pruning when a partition source column has been dropped from the current schema.

Manifest partition summaries are positional and correspond to the full partition field list in the manifest's partition spec. Before this change, `newManifestEvaluator` used `PartitionSpec.PartitionType(schema)`, which drops partition fields whose source column is missing from the provided schema. That can shift later partition fields to the wrong position.

For example, if the manifest was written with partition fields `[q_part, p_part]`, and the current schema has dropped `q`, then `p_part` could be evaluated against `q_part`'s summary slot. This can incorrectly prune a manifest and return incomplete results.

This change keeps all partition fields when building the manifest evaluator schema. Dropped-source fields are represented as `UnknownType` placeholders so later fields remain positionally aligned.

This matches Java Iceberg's partition type behavior, where dropped partition-source fields are represented with UnknownType in the partition type.

## Tests

Added `TestManifestEvaluatorWithDroppedPartitionSource`, covering:

- partition spec: `identity(q), identity(p)`
- current schema: `q` dropped, `p` retained
- row filter: `p = 2`, projected to `p_part`
- manifest summaries: `q_part` at slot 0, `p_part` at slot 1

Before the fix, the manifest can be incorrectly pruned. After the fix, it remains eligible.

## Validation

- `go test ./table -run TestManifestEvaluatorWithDroppedPartitionSource -count=1`
- `go test ./table`
- `go test ./...`

### DIFF
--- a/table/evaluators_test.go
+++ b/table/evaluators_test.go
@@ -716,50 +716,83 @@ func TestManifestEvaluator(t *testing.T) {
 }
 
 func TestManifestEvaluatorWithDroppedPartitionSource(t *testing.T) {
-	qMin, err := iceberg.Int64Literal(10).MarshalBinary()
-	require.NoError(t, err)
-	qMax, err := iceberg.Int64Literal(20).MarshalBinary()
-	require.NoError(t, err)
-	pMin, err := iceberg.Int64Literal(2).MarshalBinary()
-	require.NoError(t, err)
-	pMax, err := iceberg.Int64Literal(2).MarshalBinary()
-	require.NoError(t, err)
+	// Manifest partition summaries are positional per the spec's full field
+	// list. When a partition field's source column is dropped from the current
+	// schema, the evaluator must keep that field's slot (as an Unknown-typed
+	// placeholder) rather than compacting it away; otherwise a later, present
+	// field is read against an earlier slot's bounds.
+	bound := func(lo, hi int64) iceberg.FieldSummary {
+		lower, err := iceberg.Int64Literal(lo).MarshalBinary()
+		require.NoError(t, err)
+		upper, err := iceberg.Int64Literal(hi).MarshalBinary()
+		require.NoError(t, err)
 
-	spec := iceberg.NewPartitionSpec(
-		iceberg.PartitionField{SourceIDs: []int{3}, FieldID: 1000, Name: "q_part", Transform: iceberg.IdentityTransform{}},
-		iceberg.PartitionField{SourceIDs: []int{2}, FieldID: 1001, Name: "p_part", Transform: iceberg.IdentityTransform{}},
-	)
+		return iceberg.FieldSummary{ContainsNull: false, LowerBound: &lower, UpperBound: &upper}
+	}
+	manifestWith := func(summaries ...iceberg.FieldSummary) iceberg.ManifestFile {
+		return iceberg.NewManifestFile(1, "", 0, 0, 0).Partitions(summaries).Build()
+	}
+	// evalForCol projects (col == 2) onto the spec and returns the manifest
+	// evaluator. In every case below the dropped field's slot contains 2
+	// exactly when the present field's slot does not, so a compacted read of
+	// the wrong slot flips both the match and the prune assertion.
+	evalForCol := func(spec iceberg.PartitionSpec, schema *iceberg.Schema, col string) func(iceberg.ManifestFile) bool {
+		project := newInclusiveProjection(schema, spec, true)
+		partitionFilter, err := project(iceberg.EqualTo(iceberg.Reference(col), int64(2)))
+		require.NoError(t, err)
+		eval, err := newManifestEvaluator(spec, schema, partitionFilter, true)
+		require.NoError(t, err)
 
-	currentSchema := iceberg.NewSchema(1,
-		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64, Required: true},
-		iceberg.NestedField{ID: 2, Name: "p", Type: iceberg.PrimitiveTypes.Int64, Required: true},
-	)
-	manifest := iceberg.NewManifestFile(1, "", 0, 0, 0).Partitions(
-		[]iceberg.FieldSummary{
-			{
-				ContainsNull: false,
-				LowerBound:   &qMin,
-				UpperBound:   &qMax,
-			},
-			{
-				ContainsNull: false,
-				LowerBound:   &pMin,
-				UpperBound:   &pMax,
-			},
-		}).Build()
+		return func(mf iceberg.ManifestFile) bool {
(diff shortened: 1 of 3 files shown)


rraulinio:
A potentially relevant fix also for https://github.com/apache/iceberg-go/issues/1178.

tanmayrauth (commented):
LGTM. The Unknown-typed placeholder keeps each field at its true index, so p_part's Pos() stays aligned with summary slot 1. Nice fix.

zeroshade:
Thanks for the fix — the positional-alignment bug checks out, and this faithfully mirrors Java's `PartitionSpec.partitionType()` (the `UnknownType` placeholder for dropped sources, introduced in apache/iceberg#11868).

One follow-up worth tracking (non-blocking for this PR): we now have two divergent ways to build a partition struct from a `(spec, schema)` pair:

- `PartitionSpec.PartitionType()` — drops fields whose source column is missing, compacting positions
- the new `manifestPartitionFields()` — keeps every field, placeholding dropped sources with `UnknownType`

Scoping the fix to a local helper here is the right call: changing `PartitionType()` itself would ripple through its other ~10 callers (partition-path generation, codecs, snapshot producers, fanout writers), so keeping the blast radius small is correct.

That said, the divergence is a latent footgun — anyone who later reaches for `PartitionType()` in a context that indexes a positional structure (e.g. a manifest `FieldSummary` slice) will silently reintroduce exactly this bug. Two small suggestions:

1. Add a short doc note on `PartitionType()` stating that it compacts dropped-source fields and therefore must not be used to index positional partition summaries.
2. File a follow-up to converge `PartitionType()` onto Java's `UnknownType` behavior so there's a single source of truth (separate PR, given the blast radius).

For reference, the data-file partition path (`buildPartitionEvaluator` → `GetPartitionRecord`) is unaffected, since `DataFile.Partition()` is keyed by field ID rather than by position.

rraulinio:
@zeroshade Fixed and added https://github.com/apache/iceberg-go/issues/1205.

table/evaluators_test.go
```diff
@@ -715,6 +715,53 @@ func TestManifestEvaluator(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Manifest partition summaries are positional. The manifest was written
+	// with q_part at slot 0 and p_part at slot 1; dropping q from the current
+	// schema must not compact p_part to slot 0 during manifest evaluation.
+	eval, err := newManifestEvaluator(spec, currentSchema, partitionFilter, true)
+	require.NoError(t, err)
+	result, err := eval(manifest)
+	require.NoError(t, err)
+	assert.True(t, result, "p_part must be evaluated against slot 1, not q_part's slot 0")
```
laskoviymishka: The only assertion here is `assert.True`, which a broken fix that returns `rowsMightMatch` unconditionally would also pass — so it doesn't actually prove the evaluator is reading slot 1.

Could we add a negative case? A second manifest where `p_part`'s bounds don't contain 2 (say 3..5), asserting `eval` returns `false`. That pins both that the right slot is read *and* that pruning still works after the fix.

table/evaluators_test.go
```diff
@@ -715,6 +715,53 @@ func TestManifestEvaluator(t *testing.T) {
 	})
 }
 
+func TestManifestEvaluatorWithDroppedPartitionSource(t *testing.T) {
```
laskoviymishka, line 718: This covers a dropped field at slot 0, but the more interesting case is a drop in the *middle* of a 3-field spec, which shifts two subsequent fields rather than one.

I'd add a `[a_part(src=dropped), b_part(src=present), c_part(src=dropped)]` case with the predicate on `b_part`, asserting it's evaluated at slot 1. That's where an off-by-one in the placeholder logic would actually surface — the current single-drop-at-slot-0 case wouldn't catch it.

laskoviymishka (changes requested):
Nice fix, the positional reasoning checks out and the UnknownType placeholder mirrors Java's partitionType() exactly. Scoping it to a local helper is the right call.

Two asks on the test before merge: it only covers a dropped field at slot 0 and only asserts the positive case, so a fix that returned true unconditionally would pass it too. A negative manifest (bounds that shouldn't match → assert false) plus a middle-of-spec drop case [a(dropped), b(present), c(dropped)] with the predicate on b would pin the behavior — that middle case is where an off-by-one would actually show up.

Once the tests are tightened, happy to approve.

table/evaluators_test.go
rraulinio: 👍

table/evaluators_test.go
rraulinio: 👍

rraulinio:
@laskoviymishka Done.

laskoviymishka (approved):
nice work! 👍

PR #1193, opened 2026-06-12
fix: incorrect pruning for NOT STARTS WITH over truncated string partitions

### LINKED ISSUE #1192: fix: incorrect pruning for `NOT STARTS WITH` over truncated string partitions
### Apache Iceberg version

main (development)

## Problem

`iceberg-go` can incorrectly prune files for filters like:

```sql
WHERE name NOT STARTS WITH 'abc'
```

when the table is partitioned by a shorter string truncate transform, for example:

```text
truncate[2](name)
```

In that case, a data file with partition value `"ab"` may contain rows like:

```text
name = "ab"
```

That row **does** satisfy:

```sql
name NOT STARTS WITH 'abc'
```

because `"ab"` does not start with `"abc"`.

However, the current projection logic can truncate the filter literal `"abc"` to `"ab"` and reason from the partition value alone. That can make the planner conclude that partition `"ab"` cannot match the filter, so the file is skipped. This is a false negative and can produce incomplete query results.

## Why This Is Unsafe

For `truncate[W](name)`, the partition value only preserves the first `W` characters of `name`. If the `NOT STARTS WITH` prefix is longer than `W`, the partition value does not contain enough information to prove that every row in the file fails the filter.

Concrete counterexample:

```text
partition transform: truncate[2](name)
partition value:     "ab"
row value:           "ab"
filter:              name NOT STARTS WITH "abc"
```

The row should be returned, so pruning the file is incorrect.

## Expected Behavior

For `NOT STARTS WITH` over a truncated string partition:

- If the filter prefix length is less than or equal to the truncate width, projection can be safe.
- If the filter prefix is longer than the truncate width, the projection should return no partition predicate and avoid pruning based on that transform.

That means the planner should read the candidate file and let row-level filtering decide.

## Proposed Fix

Update the truncate string projection logic so `NOT STARTS WITH` only produces a partition predicate when the filter prefix length is less than or equal to the truncate width.

If the filter prefix is longer than the truncate width, return no projection instead of projecting an unsafe predicate.

Add regression coverage for a case like:

```text
partition: truncate[2](name)
filter:    name NOT STARTS WITH "abc"
value:     "ab"
```

The test should fail before the fix because the file is incorrectly pruned, and pass after the fix because pruning is disabled for that unsafe case.

rraulinio:
## Description

Closes #1192.

This fixes unsafe inclusive projection for `NOT STARTS WITH` predicates over truncated string/binary partition fields.

Previously, `truncate[N]` projection could truncate the filter prefix and produce a `NOT STARTS WITH` predicate on the partition value even when the original prefix was longer than `N`. That is unsafe because the truncated partition value does not contain enough information to prove that all rows fail the original predicate, and can incorrectly prune files that contain matching rows.

The projection now follows the safe behavior used by other Iceberg implementations:

- prefix shorter than truncate width: keep `NOT STARTS WITH`
- prefix exactly equal to truncate width: project to `!=`
- prefix longer than truncate width: return no projection

Regression coverage was added for all three cases.

## Testing

```sh
go test ./table -run TestProjection/TestStringTruncateProjection -count=1
go test ./...
```

### DIFF
--- a/table/evaluators_test.go
+++ b/table/evaluators_test.go
@@ -966,6 +966,9 @@ func (p *ProjectionTestSuite) TestStringTruncateProjection() {
 		{iceberg.NotEqualTo(ref, "aaa"), iceberg.AlwaysTrue{}},
 		{iceberg.IsIn(ref, "aaa", "aab"), iceberg.EqualTo(truncStr, "aa")},
 		{iceberg.NotIn(ref, "aaa", "aab"), iceberg.AlwaysTrue{}},
+		{iceberg.NotStartsWith(ref, "a"), iceberg.NotStartsWith(truncStr, "a")},
+		{iceberg.NotStartsWith(ref, "aa"), iceberg.NotEqualTo(truncStr, "aa")},
+		{iceberg.NotStartsWith(ref, "aaa"), iceberg.AlwaysTrue{}},
 	}
 
 	project := newInclusiveProjection(schema, spec, true)
--- a/table/evaluators_test.go
+++ b/table/evaluators_test.go
@@ -981,6 +981,52 @@ func (p *ProjectionTestSuite) TestStringTruncateProjection() {
 	}
 }
 
+func (p *ProjectionTestSuite) TestBinaryTruncateProjection() {
+	schema := iceberg.NewSchema(0,
+		iceberg.NestedField{ID: 1, Name: "data", Type: iceberg.PrimitiveTypes.Binary},
+	)
+	spec := iceberg.NewPartitionSpec(
+		iceberg.PartitionField{
+			SourceIDs: []int{1}, FieldID: 1000,
+			Transform: iceberg.TruncateTransform{Width: 2}, Name: "data_trunc",
+		},
+	)
+
+	ref, truncBin := iceberg.Reference("data"), iceberg.Reference("data_trunc")
+	notStartsWith := func(t iceberg.UnboundTerm, v string) iceberg.BooleanExpression {
+		return iceberg.LiteralPredicate(iceberg.OpNotStartsWith, t, iceberg.NewLiteral([]byte(v)))
+	}
+	tests := []struct {
+		pred, expected iceberg.BooleanExpression
+	}{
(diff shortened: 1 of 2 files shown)


rraulinio:
A potentially relevant fix also for https://github.com/apache/iceberg-go/issues/1178.

tanmayrauth (commented):
Nice fix. The logic lines up with the Java `Truncate.project` behavior for NOT_STARTS_WITH (prefix < width → NOT_STARTS_WITH with the original literal, == width → !=, > width → no projection), and using `p.Literal()`  instead of the truncated literal is the right call. 

  One small thing: `literalLen` handles `BinaryLiteral`/`FixedLiteral` and the new block sits above the type switch, so binary truncate fields go through the same path — but there's no binary regression test  (only `TestStringTruncateProjection`). Worth adding a binary `NotStartsWith` case so that path is locked in too?

zeroshade:
> One small thing: literalLen handles BinaryLiteral/FixedLiteral and the new block sits above the type switch, so binary truncate fields go through the same path — but there's no binary regression test (only TestStringTruncateProjection). Worth adding a binary NotStartsWith case so that path is locked in too?

I agree with @tanmayrauth here, let's add a test case to cover the binary/FixedLiteral case. 

Also, @kevinjqliu @Fokko I think that pyiceberg might share this same bug, might be worth investigating

laskoviymishka (changes requested):
Agree here with @tanmayrauth , the binary path is entirely untested. literalLen and the new branch both handle BinaryLiteral, and the  byte-length logic looks sound there for the same reason it works for strings, but nothing exercises it. I'd add a TestBinaryTruncateProjection mirroring the string cases across the three length branches so that path is locked in.

kevinjqliu:
thanks for the ping, i added a github issue to track on the iceberg-python side https://github.com/apache/iceberg-python/issues/3493

rraulinio:
@tanmayrauth @zeroshade @laskoviymishka Added.

rraulinio:
@laskoviymishka @zeroshade 😇

laskoviymishka (approved):
🚢 👍

PR #1194, opened 2026-06-12
fix: runtime.staticuint64s+42100384 is not 2-byte aligned on big-endian

strophy:
https://github.com/minio/warp recently added iceberg-go as a dependency. iceberg-go encounters the following compiler error on s390x architecture (big-endian) during packaging for Alpine Linux:
```
# github.com/minio/warp
github.com/apache/iceberg-go/table/internal.newStatAgg[go.shape.bool]: runtime.staticuint64s+42100384 is not 2-byte aligned
github.com/apache/iceberg-go/table/internal.newStatAgg[go.shape.bool]: runtime.staticuint64s+42100384 is not 2-byte aligned
github.com/apache/iceberg-go.getComparator[go.shape.bool]: runtime.staticuint64s+42100384 is not 2-byte aligned
github.com/apache/iceberg-go.getComparator[go.shape.bool]: runtime.staticuint64s+42100384 is not 2-byte aligned
>>> ERROR: warp-s3: build failed
```
The issue here is instantiating the generic bool comparator on a generic zero value, resulting in an alignment bug on big-endian architecture. This fix replaces the generic zero-value comparator initialization with explicit typed comparators, allowing compilation on s390x.

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -60,3 +60,28 @@ jobs:
     # Race detector is opt-in per package/test.
     - name: Run race detector
       run: go test -race -v ./codec/...
+
+  # Guards against big-endian-only failures such as the runtime.staticuint64s
+  # relocation misalignment triggered by boxing scalars in generic code
+  # (golang/go#76744). The fault is a linker error and only s390x's linker
+  # rejects the misaligned relocation, so this must build a main package for
+  # s390x specifically rather than just vet or compile libraries.
+  cross-compile:
+    name: cross-compile linux/s390x
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+    - name: Install Go
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+      with:
+        go-version: '1.26.1'
+        cache: true
+        cache-dependency-path: go.sum
+    - name: Build for big-endian (s390x)
+      env:
+        GOOS: linux
+        GOARCH: s390x
+        CGO_ENABLED: '0'
+      run: go build ./...
--- a/literals.go
+++ b/literals.go
@@ -80,37 +80,43 @@ type NumericLiteral interface {
 	Decrement() Literal
 }
 
-// NewLiteral provides a literal based on the type of T
+// NewLiteral provides a literal based on the type of T.
+//
+// The type switch is performed on &val (a pointer) rather than val itself.
+// Boxing a small scalar such as bool into an interface makes the compiler
+// reference runtime.staticuint64s, which emits a relocation the linker rejects
+// as misaligned on big-endian platforms such as s390x. Switching on a pointer
+// boxes only the pointer, never the scalar, avoiding that reference.
 func NewLiteral[T LiteralType](val T) Literal {
-	switch v := any(val).(type) {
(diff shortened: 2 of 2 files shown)


zeroshade:
Hmm, is there any change we can make to the bool comparator that would let us keep the zero-value initialization?

strophy:
Hi @zeroshade thanks for the lightning-quick review! A quick disclaimer, I'm an Alpine package maintainer, not a specialist in s390x or Golang, so I'm leaning heavily on AI here and careful review (or a better solution) is needed. Apologies in advance if this is against your contibutor guidelines, this is a [genuine bug I'm hitting](https://gitlab.alpinelinux.org/strophy/aports/-/jobs/2387614) and not a drive-by slop PR. :sweat_smile: 

According to Claude, there is no way to keep the zero-value initialization because:

- The failure site is the generic instantiation path, not comparator logic: it blows up at `getComparator`/`newStatAgg` specialization time, before any bool comparison executes.
- The problematic pattern is the zero-value generic conversion path in `iceberg-go/literals.go` and the former generic helper call sites in `iceberg-go/manifest.go` and `iceberg-go/table/internal/utils.go`.
- Changing the body of `BoolLiteral.Comparator` in `iceberg-go/literals.go` does not remove the monomorphization/type-assertion path that triggers the s390x alignment issue.

zeroshade:
Hmm, thanks for the reply. I'll take a little look into this myself shortly and see what I can come up with. I'll also want to see if there's a way we can get a big-endian build into the CI so that we can prevent something like this from recurring.

zeroshade:
Thanks for digging into this, @strophy!

I pushed an alternative fix to this branch (force-pushed over your commit — the previous version is still reachable via the PR's force-push timeline). It resolves the same s390x failure but keeps the generics intact.

**Root cause** (confirmed with a local repro + upstream [golang/go#76744](https://github.com/golang/go/issues/76744)): the crash isn't in the comparator logic — it's the generic instantiation boxing a scalar *zero value* into an interface (`var z T; NewLiteral(z)` with `T = bool`). Boxing a 1-byte value makes the compiler reference `runtime.staticuint64s`, and on big-endian the resulting relocation isn't aligned the way the s390x linker requires.

**Fix:** perform `NewLiteral`'s type switch on `&val` (a pointer) instead of `val`. Pointer values take the compiler's `IsDirectIface` path and skip the `staticuint64s` optimization entirely, so the bad relocation is never emitted. Because the boxing was the only problem, everything else stays as-is:

- `getComparator` and `newStatAgg` keep their original generic bodies — no new exported API, no per-call-site comparators.
- Only `NewLiteral` in `literals.go` changes; `manifest.go`, `parquet_files.go`, and `table/internal/utils.go` are untouched.
- Behavior and allocation counts are unchanged (benchmarked bool/int64/string — identical allocs).

**Verification:** the failure reproduces on any machine with `GOOS=linux GOARCH=s390x go build ./...` (it's a link-time error, so it has to link a main package) — no s390x hardware required. Confirmed fixed, and the full test suite passes.

I also added a `cross-compile` CI job that links for **s390x** so this can't silently regress. Worth noting: s390x specifically is what catches it — I tested ppc64 (also big-endian) and its linker does *not* reject the relocation, so a ppc64 check would stay green even with the bug present.

While I double checked myself using GOOS/GOARCH, can you also test on your end to double-confirm that the new approach works for you?

laskoviymishka (approved):
👍

strophy:
Looks good from my end, iceberg-go compiles with GOARCH=s390x from this PR branch both in isolation and as a dependency of warp. Your fix is much cleaner, thanks for your quick response here!

PR #1195, opened 2026-06-12
fix(exprs): support nanosecond timestamp predicates

fallintoplace:
## Summary

- bind `timestamp_ns` and `timestamptz_ns` references as `TimestampNano` in expression factories
- evaluate bound nanosecond timestamp comparisons through `doCmp`
- cover binding, `IS NULL`, `EQ`, multi-value `IN`, and evaluator comparisons for both nanosecond timestamp types

## Why

V3 nanosecond timestamp columns are valid primitive and literal types, but expression binding and comparison evaluation still missed those type cases. Filters against these columns could panic while binding or fail when evaluating bound comparison predicates.

## Testing

- `go test . -run "TestRefTypes|TestEvaluatorCmpTypes" -count=1`
- `go test ./... -count=1`
- `git diff --check`

### DIFF
--- a/exprs.go
+++ b/exprs.go
@@ -443,6 +443,8 @@ func createBoundRef(field NestedField, acc accessor) BoundReference {
 		return &boundRef[Time]{field: field, acc: acc}
 	case TimestampType, TimestampTzType:
 		return &boundRef[Timestamp]{field: field, acc: acc}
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
LGTM! THanks!

PR #1196, opened 2026-06-12
fix(transforms): support timestamp nanos in year and month

fallintoplace:
## Summary

- add `timestamp_ns` and `timestamptz_ns` support to `YearTransform.Transformer` and `MonthTransform.Transformer`
- make `YearTransform.Apply` and `MonthTransform.Apply` handle `TimestampNsLiteral`
- make transform projection handle `TimestampNsLiteral` through `transformLiteral`
- add regression coverage for Apply, Transformer, and Project paths, including a pre-epoch ns value

## Why

`YearTransform` and `MonthTransform` already report that they can transform nanosecond timestamp types through `canTransformTime`, but their implementation only handled date and microsecond timestamp values. That meant accepted partition transforms could produce invalid/null transform results for timestamp-ns inputs, and projection needed the same literal support once the transformer path succeeded.

## Testing

- `go test . -run "TestYearMonthTransformNanoseconds|TestCanTransform|TestManifestPartitionVals" -count=1`
- `go test ./... -count=1`
- `git diff --check`

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -1031,6 +1031,8 @@ func transformLiteral[T LiteralType](fn func(any) Optional[T], lit Literal) Lite
 		return NewLiteral(fn(Time(l)).Val)
 	case TimestampLiteral:
 		return NewLiteral(fn(Timestamp(l)).Val)
+	case TimestampNsLiteral:
+		return NewLiteral(fn(TimestampNano(l)).Val)
 	case StringLiteral:
 		return NewLiteral(fn(string(l)).Val)
 	case FixedLiteral:
--- a/transforms.go
+++ b/transforms.go
@@ -650,6 +650,17 @@ func (YearTransform) Transformer(src Type) (func(any) Optional[int32], error) {
 				Val:   int32(v.(Timestamp).ToTime().Year() - epochTM.Year()),
 			}
 		}, nil
+	case TimestampNsType, TimestampTzNsType:
+		return func(v any) Optional[int32] {
+			if v == nil {
+				return Optional[int32]{}
+			}
+
+			return Optional[int32]{
+				Valid: true,
+				Val:   int32(v.(TimestampNano).ToTime().Year() - epochTM.Year()),
+			}
+		}, nil
 	}
 
 	return nil, fmt.Errorf("%w: cannot apply year transform for type %s",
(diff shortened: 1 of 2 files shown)


transforms.go
```diff
@@ -668,6 +679,9 @@ func (YearTransform) Apply(value Optional[Literal]) (out Optional[Literal]) {
 	case TimestampLiteral:
 		out.Valid = true
 		out.Val = Int32Literal(Timestamp(v).ToTime().Year() - epochTM.Year())
+	case TimestampNsLiteral:
```
laskoviymishka, line 682: I think this PR makes a latent panic reachable. `transformLiteral` (further down, the `panic("invalid literal type")` arm) has no `TimestampNsLiteral` case.

Now that `Transformer` succeeds for ns timestamps, `projectTimeTransform` obtains a transformer and runs the literal through `truncateNumber`/`setApplyTransform` → `transformLiteral`, which panics. So a range predicate like `ts_col > '2024-01-01'` on a year/month-partitioned ns-timestamp column now panics at runtime in `Transform.Project(...)` — it used to error out before this change.

I'd add the matching case alongside the `TimestampLiteral` one:

```go
case TimestampNsLiteral:
    return NewLiteral(fn(TimestampNano(l)).Val)
```

Worth doing here since this PR is what makes it reachable for Year/Month. wdyt — fix in this PR, or land with an explicit follow-up note?

laskoviymishka (changes requested):
Nice tight fix,
both transforms advertised the ns-timestamp types through canTransformTime but had no matching arms in Transformer/Apply, so this closes a real contract gap and mirrors the Day/Hour pattern cleanly.

One blocking thing before merge: transformLiteral has no TimestampNsLiteral case and falls through to panic("invalid literal type"). That arm was dead for Year/Month before this PR because Transformer errored out for ns timestamps, now that it succeeds, projectTimeTransform reaches the panic. A range predicate on a year/month-partitioned ns-timestamp column will panic at runtime in Transform.Project(...). I'd add the TimestampNsLiteral case alongside TimestampLiteral here, or land this with an explicit follow-up note.

transforms.go
zeroshade: I'd preferably fix here

zeroshade:
Can we also add a pre-epoch (e.g. negative value) ns case for the Year/Month to mirror the existing `TestHourTransformPreEpoch` tests?

transforms.go
fallintoplace: Thank you for your attention. I will be working on a fix in a few hours. @zeroshade

fallintoplace: @laskoviymishka Also welcome back from the hiatus. Could you please check again?

laskoviymishka (approved):
🚢 👍 thanks!

PR #1197, opened 2026-06-12
fix(transforms): support timestamp nanos in bucket

fallintoplace:
Summary
- add `timestamp_ns` and `timestamptz_ns` support to `BucketTransform.Apply` and `BucketTransform.Transformer`
- keep bucket hashing on the physical integer timestamp value
- add regression coverage for Apply, Transformer, and Project paths, including a pre-epoch ns value

Why
`BucketTransform.CanTransform` already accepted nanosecond timestamp types, but `Apply` returned an invalid optional for `TimestampNsLiteral` and `Transformer` could return a closure with a nil hash function. That made `bucket(timestamp_ns)` unsafe for partition value generation and predicate projection.

Testing
- `go test . -run "TestBucketTransformTimestampNanoseconds|TestCanTransform|TestManifestPartitionVals" -count=1`
- `go test ./... -count=1`
- `git diff --check`

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -285,6 +285,8 @@ func (t BucketTransform) Apply(value Optional[Literal]) Optional[Literal] {
 		hash = hashHelperInt[int64](int64(v))
 	case TimestampLiteral:
 		hash = hashHelperInt[int64](int64(v))
+	case TimestampNsLiteral:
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Good catch! Thanks for this!

PR #1198, opened 2026-06-12
fix(transforms): preserve UTF-8 in string truncate

fallintoplace:
Summary
- make `truncate[string]` truncate at UTF-8 rune boundaries instead of byte offsets
- keep `truncate[binary]` byte-based
- add regression coverage for Apply, Transformer, and Project paths

Why
`TruncateTransform.Transformer` handled strings and binary values with the same byte-slicing branch. For non-ASCII strings, a width could split a UTF-8 code point and produce an invalid string partition value.

Testing
- `go test . -run "TestTruncateStringDoesNotSplitUTF8|TestTruncateTransform|TestCanTransform|TestManifestPartitionVals" -count=1`
- `go test ./... -count=1`
- `git diff --check`

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -27,7 +27,6 @@ import (
 	"strconv"
 	"strings"
 	"time"
-	"unicode/utf8"
 	"unsafe"
 
 	"github.com/apache/arrow-go/v18/arrow/decimal128"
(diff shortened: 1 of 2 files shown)


transforms.go
```diff
@@ -477,6 +484,18 @@ func (t TruncateTransform) Transformer(src Type) (func(any) any, error) {
 		ErrInvalidArgument, src)
 }
 
+func truncateString(s string, width int) string {
+	for idx := range s {
+		if width == 0 {
+			return s[:idx]
```
zeroshade, line 532: Isn't this just the same exact thing as we already had? I'm not clear on how this is different at all to the existing code

fallintoplace: Yeah, I can see why it looks similar at first glance.

The difference is that the old code was cutting strings by byte count. So for something like `"éx"` with width `1`, it could return only the first byte of `"é"`, which makes the result invalid UTF-8.

This version still slices the original string, but it only slices at indexes from `range s`, which are valid UTF-8 character boundaries. So width `1` returns `"é"` instead of half of that character.

Binary is still byte-based, so I split that into a separate branch.

zeroshade: Ah, wouldn't it be simpler to just do: 

```go
r := []rune(s)
return string(r[:width])
```

fallintoplace: I checked the Go spec again, and I think the small reason to keep this version is that range over a string already gives us the byte index for each UTF-8 code point boundary:

https://go.dev/ref/spec#For_statements_with_range_clause

So this only walks until it finds the cutoff point and then slices the original string there.

With the []rune version we’d need a bounds check to avoid panicking when width is larger than the number of runes, and it also converts the whole string to a rune slice first.

That said, I agree the []rune version is easier to understand at a glance. I’m happy to either keep this and add a short comment, or switch to the clearer version with the bounds check if you prefer.

zeroshade: Fair enough, can you just add a comment to the function explaining why we need to do the loop like that? We can keep the loop since it's a good point that we don't need to do the copy, let's just add the comment so we don't look at this later and get confused 😄

zeroshade (approved):

PR #1201, opened 2026-06-14
fix(table): Propagate snapshotbyID error in snapshot summaries

dttung2905:
During commit, summary() loads the parent snapshot to inherit its summary properties before calling updateSnapshotSummaries:

https://github.com/apache/iceberg-go/blob/1e4916f7a79462d193ed65d672545ff71fa6f556/table/snapshot_producers.go#L810-L827

If lookup failed, ``previousSnapshot` stayed nil, `previousSummary` was never set, and updateSnapshotSummaries fell back to zero baselines. That produced incorrect incremental snapshot summary properties (total-data-files, total-records, etc.) with no error surfaced to the caller.

The same lookup in `computeOwnManifests` already propagates the error and is covered by `TestComputeOwnManifests_SnapshotByIDError`; the summary path did not.

https://github.com/apache/iceberg-go/blob/1e4916f7a79462d193ed65d672545ff71fa6f556/table/snapshot_producers.go#L833-L842

### DIFF
--- a/table/snapshot_producers.go
+++ b/table/snapshot_producers.go
@@ -809,7 +809,10 @@ func (sp *snapshotProducer) summary(props iceberg.Properties) (Summary, error) {
 
 	var previousSnapshot *Snapshot
 	if sp.parentSnapshotID > 0 {
-		previousSnapshot, _ = sp.txn.meta.SnapshotByID(sp.parentSnapshotID)
+		previousSnapshot, err = sp.txn.meta.SnapshotByID(sp.parentSnapshotID)
+		if err != nil {
+			return Summary{}, fmt.Errorf("summary: lookup parent snapshot %d: %w", sp.parentSnapshotID, err)
+		}
 	}
 
 	var previousSummary iceberg.Properties
--- a/table/snapshot_producers.go
+++ b/table/snapshot_producers.go
@@ -816,7 +816,7 @@ func (sp *snapshotProducer) summary(props iceberg.Properties) (Summary, error) {
 	}
 
 	var previousSummary iceberg.Properties
-	if previousSnapshot != nil {
+	if previousSnapshot != nil && previousSnapshot.Summary != nil {
 		previousSummary = previousSnapshot.Summary.Properties
 	}
 
--- a/table/snapshot_producers_test.go
+++ b/table/snapshot_producers_test.go
@@ -923,6 +923,54 @@ func TestComputeOwnManifests_NewTable(t *testing.T) {
 	require.Nil(t, got, "new table: returned manifests must equal input")
 }
 
+// TestSummary_SnapshotByIDError verifies that summary computation fails when
+// the parent snapshot cannot be found, instead of silently treating totals as
+// starting from zero.
+func TestSummary_SnapshotByIDError(t *testing.T) {
+	spec := iceberg.NewPartitionSpec()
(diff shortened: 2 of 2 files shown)


table/snapshot_producers.go
```diff
@@ -809,7 +809,10 @@ func (sp *snapshotProducer) summary(props iceberg.Properties) (Summary, error) {
...
 	var previousSnapshot *Snapshot
 	if sp.parentSnapshotID > 0 {
-		previousSnapshot, _ = sp.txn.meta.SnapshotByID(sp.parentSnapshotID)
+		previousSnapshot, err = sp.txn.meta.SnapshotByID(sp.parentSnapshotID)
+		if err != nil {
+			return Summary{}, fmt.Errorf("summary: lookup parent snapshot %d: %w", sp.parentSnapshotID, err)
+		}
 	}
 
 	var previousSummary iceberg.Properties
```
laskoviymishka, line 818: I think this fix surfaces a latent crash just below at the `previousSummary = previousSnapshot.Summary.Properties` line.

`Summary` is `*Summary` with `omitempty`, so a V1 table or a parent snapshot written without a summary can legally have it nil. Before this change a failed lookup left `previousSnapshot` nil and the `if previousSnapshot != nil` guard skipped the deref; now a successful lookup guarantees non-nil, so that line panics unconditionally for those inputs. The old silent-zero bug was masking it.

I'd extend the guard the way line ~1066 already does: `if previousSnapshot != nil && previousSnapshot.Summary != nil`. wdyt?

table/snapshot_producers_test.go
```diff
@@ -923,6 +923,54 @@ func TestComputeOwnManifests_NewTable(t *testing.T) {
 	require.Nil(t, got, "new table: returned manifests must equal input")
 }
 
+// TestSummary_SnapshotByIDError verifies that summary computation fails when
+// the parent snapshot cannot be found, instead of silently treating totals as
+// starting from zero.
+func TestSummary_SnapshotByIDError(t *testing.T) {
```
laskoviymishka, line 929: While we're here, neither new test covers the case that actually panics now — parent present but `Summary == nil`. CI stays green while that crash ships.

I'd add a case injecting a parent snapshot with `Summary: nil` so the nil guard above is regression-protected.

laskoviymishka (changes requested):
Thanks for this, one blocking thing from my POV:

The fix exposes a latent nil-deref one line down: previousSummary = previousSnapshot.Summary.Properties dereferences Summary unconditionally, but it's a *Summary with omitempty, so a V1 table or a summary-less parent snapshot can have it nil. The old discarded-error path left previousSnapshot nil and skipped that line; now a successful lookup guarantees non-nil, so it panics on the second commit for that class of input. I'd guard it the same way line ~1066 already does, and add a test case with Summary: nil so it's covered.

table/snapshot_producers.go
dttung2905: Thanks for the comment, I think you brought up a valid concern. I have added extra guard for it

table/snapshot_producers_test.go
dttung2905: added a new test case where parent snapshot with `Summary: nil`

laskoviymishka (approved):
:shipit:

PR #1209, opened 2026-06-15
docs: document ExpireSnapshots in the Maintenance API section

laskoviymishka:
Maintancne docs to cover V2 spec completion.


zeroshade (approved):
Can we also include docs like this in the Go docstrings and package docs? That way the docs would show up on pkg.go.dev?

laskoviymishka:
Good idea, added a doc-strings

PR #1211, opened 2026-06-15
fix(catalog/sql): use catalog name instead of hardcoded "sql"

### LINKED ISSUE #1210: catalog/sql: uses hardcoded "sql" instead of catalog name
### Apache Iceberg version

main (development)

## Description

The SQL Catalog registrar in [`catalog/sql/sql.go`](https://github.com/apache/iceberg-go/blob/0c5552bfd34937e07bcb22f9bf9a3b5c4e33b27c/catalog/sql/sql.go#L95) ignores the catalog name passed to `catalog.Load(ctx, name, props)`, and threads it through `Properties.Get(name, "sql")`. It tries to find the key which is equal to the catalog name. 

This breaks cross-engine interoperability. A catalog loaded as `catalog.Load(ctx, "<catalog_name>", ...)` stores metadata with `catalog_name='sql'`, while Spark, configured with Iceberg Java JDBC catalog use `catalog_name='<catalog_name>'`

## Proposed Solution
Pass `c.name` as the name the to `catalog.Load`, so a `catalog.Load(ctx, "<catalog_name>", …)` shares the same catalog name row space as Spark or any other caller.

badalprasadsingh:
## Problem

Fixes #1210 

The SQL Catalog registrar in `catalog/sql/sql.go` ignores the catalog name passed to `catalog.Load(ctx, name, props)`, and threads it through `Properties.Get(name, "sql")`. 

## Solution
Thread the loader-supplied `name` directly into `NewCatalog`, so it ends up as `c.name` and as the `catalog_name` partition key on every row. A `catalog.Load(ctx, "olake_iceberg", …)` then shares the same row space as Spark loading the same name.

This change aligns with:
- what we have for [REST](https://github.com/apache/iceberg-go/blob/0c5552bfd34937e07bcb22f9bf9a3b5c4e33b27c/catalog/rest/rest.go#L103-L111) and [Hadoop](https://github.com/apache/iceberg-go/blob/0c5552bfd34937e07bcb22f9bf9a3b5c4e33b27c/catalog/hadoop/hadoop.go#L41-L49) as the catalog.
- Iceberg-java's [JdbcCatalog.initialize](https://github.com/apache/iceberg/blob/6ee0a3be308037e280ff17172a5bbd67a2a4131a/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java#L130-L132) stores the caller-supplied name.

## Files Changed

| File | Description |
|------|-------------|
| `catalog/sql/sql.go` | Bug fix |
| `catalog/sql/sql_test.go` | Test for it |

## Testing
- Ingested data into the SQL catalog
- Queried through Spark `3.5.5`

## Changelog Note
Prior versions of the SQL Catalog registrar ignored the `name` argument and always wrote rows with catalog_name=`sql`.

**Upgrade:** if you previously loaded under a name other than `"sql"`, re-key existing rows:
```sql
UPDATE iceberg_tables SET catalog_name='<name>' WHERE catalog_name='sql';
UPDATE iceberg_namespace_properties SET catalog_name='<name>' WHERE catalog_name='sql';
```

Warehouse paths and metadata files are unchanged.

### DIFF
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -92,6 +92,8 @@ func init() {
 			}
 		}()
 
+		// Here, name is the loader-supplied catalog name and is persisted as the
+		// catalog_name partition key on every row in the iceberg_tables and iceberg_namespace_properties.
 		return NewCatalog(name, sqldb, SupportedDialect(dialect), p)
 	}))
 }
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -92,7 +92,7 @@ func init() {
 			}
 		}()
 
-		return NewCatalog(p.Get(name, "sql"), sqldb, SupportedDialect(dialect), p)
+		return NewCatalog(name, sqldb, SupportedDialect(dialect), p)
 	}))
 }
 
--- a/catalog/sql/sql_test.go
+++ b/catalog/sql/sql_test.go
@@ -323,25 +323,42 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
 }
 
 func (s *SqliteCatalogTestSuite) TestCatalogNameMatchesLoaderArg() {
-	const catalogName = "test_catalog"
-	cat, err := catalog.Load(context.Background(), catalogName, iceberg.Properties{
-		"uri":             ":memory:",
+	ctx := context.Background()
+
+	sharedURI := s.catalogUri()
+	props := iceberg.Properties{
+		"uri":             sharedURI,
 		sqlcat.DriverKey:  sqliteshim.ShimName,
 		sqlcat.DialectKey: string(sqlcat.SQLite),
 		"type":            "sql",
 		"warehouse":       "file://" + s.warehouse,
-	})
+	}
+
+	const catalogName = "test_catalog"
+	cat, err := catalog.Load(ctx, catalogName, props)
 	s.Require().NoError(err)
-	sqlCat := cat.(*sqlcat.Catalog)
+	sqlCat, ok := cat.(*sqlcat.Catalog)
+	s.Require().True(ok, "expected *sqlcat.Catalog")
 	s.Equal(catalogName, sqlCat.Name(), "SQL catalog must surface the name passed to catalog.Load")
 
-	ctx := context.Background()
 	ns := table.Identifier{"test"}
 	s.Require().NoError(sqlCat.CreateNamespace(ctx, ns, iceberg.Properties{"created_by": "iceberg-go"}))
 
 	got, err := sqlCat.ListNamespaces(ctx, table.Identifier{})
 	s.Require().NoError(err)
+	s.Len(got, 1, "the loader-named catalog should see exactly the namespace it just created")
 	s.Contains(got, ns, "iceberg-go must list a namespace it just created under its own catalog name")
+
+	otherCat, err := catalog.Load(ctx, "sql", props)
+	s.Require().NoError(err)
+	otherSQLCat, ok := otherCat.(*sqlcat.Catalog)
+	s.Require().True(ok, "expected *sqlcat.Catalog")
+	s.Equal("sql", otherSQLCat.Name())
+
+	otherNS, err := otherSQLCat.ListNamespaces(ctx, table.Identifier{})
+	s.Require().NoError(err)
+	s.NotContains(otherNS, ns,
+		"a catalog loaded under a different name must not see namespaces written under another catalog_name")
 }
 
 func (s *SqliteCatalogTestSuite) TestDropSQLTablesIdempotency() {
--- a/catalog/sql/sql_test.go
+++ b/catalog/sql/sql_test.go
@@ -322,6 +322,28 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
 	s.confirmTablesExist(sqldb)
 }
 
+func (s *SqliteCatalogTestSuite) TestCatalogNameMatchesLoaderArg() {
+	const catalogName = "test_catalog"
+	cat, err := catalog.Load(context.Background(), catalogName, iceberg.Properties{
+		"uri":             ":memory:",
+		sqlcat.DriverKey:  sqliteshim.ShimName,
+		sqlcat.DialectKey: string(sqlcat.SQLite),
+		"type":            "sql",
+		"warehouse":       "file://" + s.warehouse,
+	})
+	s.Require().NoError(err)
+	sqlCat := cat.(*sqlcat.Catalog)
+	s.Equal(catalogName, sqlCat.Name(), "SQL catalog must surface the name passed to catalog.Load")
+
+	ctx := context.Background()
+	ns := table.Identifier{"test"}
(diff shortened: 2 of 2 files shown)


badalprasadsingh:
@laskoviymishka, please take a look when you are free

catalog/sql/sql.go
```diff
@@ -92,7 +92,7 @@ func init() {
 			}
 		}()
 
-		return NewCatalog(p.Get(name, "sql"), sqldb, SupportedDialect(dialect), p)
+		return NewCatalog(name, sqldb, SupportedDialect(dialect), p)
```
laskoviymishka, line 97: This is the right fix — `p.Get(name, "sql")` was treating `name` as a lookup key (never set), so it always fell back to "sql". `name` flows straight into `c.name` as the `catalog_name` partition key on every query, which matches REST and Hadoop here.

The one thing this changes silently: existing deployments that loaded under a name other than "sql" had everything written under `catalog_name='sql'`, and now those rows won't be found. I'd document the upgrade in the CHANGELOG with the re-key `UPDATE iceberg_tables SET catalog_name='<name>' WHERE catalog_name='sql'` (and the same for `iceberg_namespace_properties`), and note warehouse paths are unchanged. A one-line comment here explaining `name` is the catalog_name partition key would also save the next person from re-deriving the `p.Get` confusion. wdyt?

catalog/sql/sql_test.go
```diff
@@ -322,6 +322,28 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
...
+	sqlCat := cat.(*sqlcat.Catalog)
+	s.Equal(catalogName, sqlCat.Name(), "SQL catalog must surface the name passed to catalog.Load")
+
+	ctx := context.Background()
+	ns := table.Identifier{"test"}
+	s.Require().NoError(sqlCat.CreateNamespace(ctx, ns, iceberg.Properties{"created_by": "iceberg-go"}))
+
+	got, err := sqlCat.ListNamespaces(ctx, table.Identifier{})
+	s.Require().NoError(err)
+	s.Contains(got, ns, "iceberg-go must list a namespace it just created under its own catalog name")
```
laskoviymishka, line 350: This is the main thing I'd want reworked. As written, this asserts a single-instance round-trip — create under "test_catalog", list under "test_catalog" — which passes under the *old* code too, since one catalog is always self-consistent regardless of what name it stored.

The regression is cross-instance isolation. I'd open a second catalog with name "sql" against the same `:memory:` DB and assert it does *not* see the namespace created under "test_catalog". That's the assertion that actually fails before the fix and passes after.

catalog/sql/sql_test.go
```diff
@@ -322,6 +322,28 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
...
+	const catalogName = "test_catalog"
+	cat, err := catalog.Load(context.Background(), catalogName, iceberg.Properties{
+		"uri":             ":memory:",
+		sqlcat.DriverKey:  sqliteshim.ShimName,
+		sqlcat.DialectKey: string(sqlcat.SQLite),
+		"type":            "sql",
+		"warehouse":       "file://" + s.warehouse,
+	})
+	s.Require().NoError(err)
+	sqlCat := cat.(*sqlcat.Catalog)
```
laskoviymishka: Minor: the bare assertion panics instead of failing cleanly if `Load` ever returns a different type. The two-value form reads better in a test:

```go
sqlCat, ok := cat.(*sqlcat.Catalog)
s.Require().True(ok, "expected *sqlcat.Catalog")
```

Matches the pattern in TestPurgeTable. Not blocking.

catalog/sql/sql_test.go
```diff
@@ -322,6 +322,28 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
...
+	s.Equal(catalogName, sqlCat.Name(), "SQL catalog must surface the name passed to catalog.Load")
+
+	ctx := context.Background()
+	ns := table.Identifier{"test"}
+	s.Require().NoError(sqlCat.CreateNamespace(ctx, ns, iceberg.Properties{"created_by": "iceberg-go"}))
+
+	got, err := sqlCat.ListNamespaces(ctx, table.Identifier{})
```
laskoviymishka, line 347: `s.Contains` alone won't notice a phantom or duplicated namespace. An `s.Len(got, 1)` before the Contains (like TestListNamespaces) makes it strict. Also `context.Background()` is called twice in this test — the surrounding multi-call tests declare `ctx` once and reuse it. Both small, fold them in if you're touching this anyway.

laskoviymishka (changes requested):
Nice catch! the fix is correct. p.Get(name, "sql") treated the catalog name as a map key (never set), so it always fell back to "sql" and the loader name never reached c.name. Passing name directly matches Java JdbcCatalog, PyIceberg, and our own REST/Hadoop registrars.

Two things I'd settle before merging. The migration break: existing deployments under a name other than "sql" have all their rows keyed under catalog_name='sql', and after this they go invisible with no error — I'd want a CHANGELOG note flagging it as breaking plus the re-key UPDATE. And the test: it's a single-instance round-trip that passes under the old code too — to catch the actual regression, load a second catalog named "sql" against the same DB and assert it doesn't see the namespace created under "test_catalog".

Both addressed and I'm happy to approve.

catalog/sql/sql.go
badalprasadsingh: Yes, you are right.

1. Added the migration UPDATEs in the PR body's **Changelog Note** section since the repo doesn't keep a `CHANGELOG.md`.
2. Added a two-line comment at the call site noting that `name` is the `catalog_name` partition key persisted in `iceberg_tables` / `iceberg_namespace_properties`.

catalog/sql/sql_test.go
badalprasadsingh: Done. 

With one small substitution: I used a file-backed sqlite DB via `s.catalogUri()` instead of `:memory:`, because two separate `:memory:` opens give you isolated `DBs` and the cross-instance assertion would silently pass for the wrong reason.

Now, second catalog loaded as `"sql"` against the same DB, asserts `NotContains` on the namespace created under `"test_catalog"`. Confirmed the assertion fails on the pre-fix `p.Get(name, "sql")` code and passes after.

catalog/sql/sql_test.go
badalprasadsingh: Done.

catalog/sql/sql_test.go
badalprasadsingh: Done. Folded both in.

badalprasadsingh:
@laskoviymishka, both addressed:

1. **Changelog**: Since, iceberg-go doesn't maintain a `CHANGELOG.md`, I've added a **Changelog Note** section to the PR body with the breaking change callout and the re-key `UPDATE`s for both `iceberg_tables` and `iceberg_namespace_properties`.
2. **Test**: Done - second catalog loaded as `sql` against the same DB asserts `NotContains` on the namespace created under `"test_catalog"`. Verified it locally: that the new assertion fails on the pre-fix code and passes after.

laskoviymishka (approved):
Thanks! 

Change log notes - is good enough here!

🚢 👍

PR #1212, opened 2026-06-15
feat(exprs): expression parsing for server-side planning

rambleraptor:
I'm hoping to build out server-side planning support in Go. There's a bunch of prerequisites we need to do before we can implement.

One of them is parsing Iceberg Expressions, which will be sent over the wire. There's work being done to formalize the Expression spec, so it's as good a time as any to have this support in iceberg-go.

Apologies, this is longer than I'd like because the expression spec is lengthy.

### DIFF
--- a/expr_json.go
+++ b/expr_json.go
@@ -67,6 +67,22 @@ var jsonToOp = func() map[string]Operation {
 	return m
 }()
 
+// Predicate operations grouped by the shape they carry on the wire. Spelled out
+// rather than keyed off iota ranges, so reordering the Operation constants can't
+// silently reclassify a predicate.
+var (
+	unaryOps = map[Operation]bool{
+		OpIsNull: true, OpNotNull: true, OpIsNan: true, OpNotNan: true,
+	}
+	literalOps = map[Operation]bool{
+		OpLT: true, OpLTEQ: true, OpGT: true, OpGTEQ: true,
+		OpEQ: true, OpNEQ: true, OpStartsWith: true, OpNotStartsWith: true,
+	}
+	setOps = map[Operation]bool{
+		OpIn: true, OpNotIn: true,
+	}
+)
+
 // exprNode is the wire form of an expression node. omitempty leaves only the
 // keys relevant to a given node; field order matches Java's output.
 type exprNode struct {
@@ -91,10 +107,21 @@ type transformNode struct {
 // a task's residual filter).
 //
 // With a schema, literals take the referenced field's type (e.g. "2022-08-14"
-// on a date column becomes a DateLiteral). Without one they fall back to the
-// base JSON kind: Int64Literal, Float64Literal, StringLiteral, or BoolLiteral.
+// on a date column becomes a DateLiteral) and field names are resolved
+// case-sensitively, since REST field names are authoritative. Without a schema
+// literals fall back to the base JSON kind: Int64Literal, Float64Literal,
+// StringLiteral, or BoolLiteral. The result is unbound.
+//
+// Transform terms (e.g. {"type":"transform","transform":"bucket[16]","term":"id"})
+// parse into an UnboundTransform. The term resolves its result type against the
+// schema; binding a full predicate over a transform term to the typed bound
+// machinery is not yet supported.
 func ParseExpr(data []byte, schema *Schema) (BooleanExpression, error) {
-	return decodeExpr(json.RawMessage(data), schema)
+	return parseExpr(json.RawMessage(data), schema, true)
+}
+
+func parseExpr(raw json.RawMessage, schema *Schema, caseSensitive bool) (BooleanExpression, error) {
+	return decodeExpr(raw, schema, caseSensitive)
 }
 
 // MarshalJSON emits the REST form, so an expression can be used as a "filter"
@@ -256,6 +283,19 @@ func (b *BoundTransform) MarshalJSON() ([]byte, error) {
 	})
 }
 
+func (u *UnboundTransform) MarshalJSON() ([]byte, error) {
+	ref, ok := u.term.(Reference)
+	if !ok {
+		return nil, fmt.Errorf("%w: cannot serialize transform over a non-reference term", ErrInvalidArgument)
+	}
+
+	return json.Marshal(transformNode{
+		Type:      exprKeyTransform,
+		Transform: u.transform.String(),
+		Term:      string(ref),
+	})
+}
+
 // Each literal writes its REST form (Java's SingleValueParser).
 func (l BoolLiteral) MarshalJSON() ([]byte, error)   { return json.Marshal(bool(l)) }
 func (l Int32Literal) MarshalJSON() ([]byte, error)  { return json.Marshal(int32(l)) }
@@ -350,17 +390,22 @@ func (v *literalValue) UnmarshalJSON(b []byte) error {
 // without it for timestamp.
 func marshalTimestamp(tm time.Time, typ Type) ([]byte, error) {
 	switch typ.(type) {
-	case TimestampTzType, TimestampTzNsType:
+	case TimestampTzType:
 		// "-07:00" prints +00:00 for a UTC instant, matching Java.
 		return json.Marshal(tm.Format("2006-01-02T15:04:05.999999-07:00"))
-	case TimestampType, TimestampNsType:
+	case TimestampTzNsType:
+		// Nanosecond precision needs nine fractional digits.
+		return json.Marshal(tm.Format("2006-01-02T15:04:05.999999999-07:00"))
+	case TimestampType:
 		return json.Marshal(tm.Format("2006-01-02T15:04:05.999999"))
+	case TimestampNsType:
+		return json.Marshal(tm.Format("2006-01-02T15:04:05.999999999"))
 	default:
 		return nil, fmt.Errorf("%w: serializing a timestamp filter needs the field type; bind the expression to a schema first", ErrInvalidArgument)
 	}
 }
 
-func decodeExpr(raw json.RawMessage, schema *Schema) (BooleanExpression, error) {
+func decodeExpr(raw json.RawMessage, schema *Schema, caseSensitive bool) (BooleanExpression, error) {
 	// A bare boolean is AlwaysTrue/AlwaysFalse; anything else is a node object.
 	tok, err := firstToken(raw)
 	if err != nil {
@@ -384,6 +429,11 @@ func decodeExpr(raw json.RawMessage, schema *Schema) (BooleanExpression, error) 
 
 	// {"type":"literal","value":<bool>} is an alternate spelling of a constant.
 	if node.Type == "literal" {
+		// A missing or null value would unmarshal to false and silently become
+		// AlwaysFalse, turning a corrupt payload into "filter everything out".
+		if len(node.Value) == 0 || bytes.Equal(node.Value, []byte("null")) {
+			return nil, fmt.Errorf("%w: literal expression is missing a boolean value", ErrInvalidArgument)
+		}
 		var bv bool
 		if err := json.Unmarshal(node.Value, &bv); err != nil {
 			return nil, fmt.Errorf("%w: cannot parse literal expression: %s", ErrInvalidArgument, err)
@@ -402,18 +452,18 @@ func decodeExpr(raw json.RawMessage, schema *Schema) (BooleanExpression, error) 
 
 	switch op {
 	case OpNot:
-		child, err := decodeExpr(node.Child, schema)
+		child, err := decodeExpr(node.Child, schema, caseSensitive)
 		if err != nil {
 			return nil, err
 		}
 
 		return NewNot(child), nil
 	case OpAnd, OpOr:
-		left, err := decodeExpr(node.Left, schema)
+		left, err := decodeExpr(node.Left, schema, caseSensitive)
 		if err != nil {
 			return nil, err
 		}
-		right, err := decodeExpr(node.Right, schema)
+		right, err := decodeExpr(node.Right, schema, caseSensitive)
 		if err != nil {
 			return nil, err
 		}
@@ -424,7 +474,7 @@ func decodeExpr(raw json.RawMessage, schema *Schema) (BooleanExpression, error) 
 		return NewOr(left, right), nil
 	}
 
-	return decodePredicate(op, node, schema)
+	return decodePredicate(op, node, schema, caseSensitive)
 }
 
 // firstToken returns the first JSON token of raw, used to classify a node
@@ -435,7 +485,7 @@ func firstToken(raw json.RawMessage) (json.Token, error) {
 	return dec.Token()
 }
 
-func decodePredicate(op Operation, node exprNode, schema *Schema) (BooleanExpression, error) {
+func decodePredicate(op Operation, node exprNode, schema *Schema, caseSensitive bool) (BooleanExpression, error) {
 	term, err := decodeTerm(node.Term)
 	if err != nil {
 		return nil, err
@@ -444,7 +494,7 @@ func decodePredicate(op Operation, node exprNode, schema *Schema) (BooleanExpres
 	// Resolve the field type once; nil means schema-less.
 	var typ Type
 	if schema != nil {
-		bound, err := term.Bind(schema, false)
+		bound, err := term.Bind(schema, caseSensitive)
 		if err != nil {
 			return nil, err
 		}
@@ -452,13 +502,13 @@ func decodePredicate(op Operation, node exprNode, schema *Schema) (BooleanExpres
 	}
 
 	switch {
-	case op >= OpIsNull && op <= OpNotNan:
+	case unaryOps[op]:
 		if len(node.Value) != 0 || node.Values != nil {
 			return nil, fmt.Errorf("%w: unary predicate %s must not have a value", ErrInvalidArgument, node.Type)
 		}
 
 		return UnaryPredicate(op, term), nil
-	case op >= OpLT && op <= OpNotStartsWith:
+	case literalOps[op]:
 		if len(node.Value) == 0 {
 			return nil, fmt.Errorf("%w: predicate %s is missing a value", ErrInvalidArgument, node.Type)
 		}
@@ -471,7 +521,7 @@ func decodePredicate(op Operation, node exprNode, schema *Schema) (BooleanExpres
 		}
 
 		return LiteralPredicate(op, term, lit), nil
-	case op >= OpIn && op <= OpNotIn:
+	case setOps[op]:
 		if node.Values == nil {
 			return nil, fmt.Errorf("%w: predicate %s is missing values", ErrInvalidArgument, node.Type)
 		}
@@ -517,7 +567,13 @@ func decodeTerm(raw json.RawMessage) (UnboundTerm, error) {
 		// {"type":"reference","term":"name"}
 		return Reference(t.Term), nil
 	case exprKeyTransform:
-		return nil, fmt.Errorf("%w: transform terms are not supported when parsing expressions", ErrNotImplemented)
+		// {"type":"transform","transform":"bucket[16]","term":"id"}
+		tf, err := ParseTransform(t.Transform)
+		if err != nil {
+			return nil, fmt.Errorf("%w: cannot parse transform term: %s", ErrInvalidArgument, err)
+		}
+
+		return NewUnboundTransform(tf, Reference(t.Term)), nil
 	default:
 		return nil, fmt.Errorf("%w: cannot parse term with type %q", ErrInvalidArgument, t.Type)
 	}
@@ -552,7 +608,11 @@ func convertValue(base Literal, typ Type) (Literal, error) {
 		if err != nil {
 			return nil, fmt.Errorf("%w: invalid hex for %s value: %s", ErrInvalidArgument, typ, err)
 		}
-		if _, ok := typ.(FixedType); ok {
+		if ft, ok := typ.(FixedType); ok {
+			if len(decoded) != ft.Len() {
+				return nil, fmt.Errorf("%w: fixed[%d] value has %d bytes", ErrInvalidArgument, ft.Len(), len(decoded))
+			}
+
 			return FixedLiteral(decoded), nil
 		}
 
--- a/expr_json.go
+++ b/expr_json.go
@@ -30,8 +30,8 @@ import (
 	"github.com/google/uuid"
 )
 
-// JSON serialization for boolean expressions, used for the "filter" field of a
-// REST scan-planning request. Mirrors Java's ExpressionParser.
+// JSON for boolean expressions, used as the "filter" of a REST scan-planning
+// request. Mirrors Java's ExpressionParser.
 
 // exprKeyTransform is the "type" value identifying a transform term.
 const exprKeyTransform = "transform"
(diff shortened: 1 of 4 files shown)


expr_json.go
```diff
@@ -0,0 +1,502 @@
...
+		return json.Marshal(string(l))
+	case DateLiteral:
+		return json.Marshal(Date(l).ToTime().Format("2006-01-02"))
+	case TimeLiteral:
+		// "9"s trim trailing fractional zeros (and the point when zero), as Java does.
+		return json.Marshal(time.UnixMicro(int64(l)).UTC().Format("15:04:05.999999"))
+	case TimestampLiteral:
+		return json.Marshal(Timestamp(l).ToTime().Format("2006-01-02T15:04:05.999999"))
+	case TimestampNsLiteral:
+		return json.Marshal(TimestampNano(l).ToTime().Format("2006-01-02T15:04:05.999999999"))
```
dillitz: I believe `timestamptz` columns serialize incorrectly here. `encodeLiteral` switches on the Go literal type, and both `timestamp` and `timestamptz` values are carried by `TimestampLiteral` (its `Type()` always reports `timestamp`, and `To(TimestampTzType)` returns the literal unchanged), so the timestamp branch formats with no zone offset.

Java appends a `+00:00` offset for `shouldAdjustToUTC` types: the [`SingleValueParser` TIMESTAMP case](https://github.com/apache/iceberg/blob/85ffa1984e115e80ba1571f3eb017fcf0ba39031/core/src/main/java/org/apache/iceberg/SingleValueParser.java#L297-L305) routes to [`microsToIsoTimestamptz`](https://github.com/apache/iceberg/blob/85ffa1984e115e80ba1571f3eb017fcf0ba39031/api/src/main/java/org/apache/iceberg/util/DateTimeUtil.java#L148-L150), which formats with the offset. So for a `timestamptz` filter value:

- Go emits `"2022-08-14T10:00:00"` where Java emits `"2022-08-14T10:00:00+00:00"`, so the bytes don't match.
- It doesn't round-trip its own output either: parsing that offset-less string back with a tz schema fails, since `To(TimestampTzType)` goes through `time.RFC3339` and needs the offset.
- A server parsing it with `isoTimestamptzToMicros` (`OffsetDateTime.parse`) needs the offset too, so it would reject the filter.

expr_json.go
```diff
@@ -0,0 +1,502 @@
...
+
+// MarshalJSON emits the REST JSON form, so an expression can be used directly
+// as a request's "filter" field. Tag such fields omitempty to drop a nil one.
+func (e AlwaysTrue) MarshalJSON() ([]byte, error)  { return encodeExpr(e) }
+func (e AlwaysFalse) MarshalJSON() ([]byte, error) { return encodeExpr(e) }
+func (e NotExpr) MarshalJSON() ([]byte, error)     { return encodeExpr(e) }
+func (e AndExpr) MarshalJSON() ([]byte, error)     { return encodeExpr(e) }
+func (e OrExpr) MarshalJSON() ([]byte, error)      { return encodeExpr(e) }
+
+func (p *unboundUnaryPredicate) MarshalJSON() ([]byte, error)   { return encodeExpr(p) }
```
dillitz: As far as I understand, it could be problematic that `MarshalJSON` is defined only on the unbound predicate types, which makes `json.Marshal` of a *bound* expression fall through to default struct marshaling and silently returns `{}` with a nil error (the bound types' fields are unexported). The scan path serializes the unbound filter, so this isn't hit today, but `json.Marshal` quietly "succeeding" with `{}` is a sharp edge if anyone later marshals a bound expression. Worth having the bound predicate types either delegate to `encodeExpr` or return an error, rather than producing `{}`.

expr_json.go
```diff
@@ -0,0 +1,502 @@
...
+	case BoolLiteral:
+		return json.Marshal(bool(l))
+	case Int32Literal:
+		return json.Marshal(int32(l))
+	case Int64Literal:
+		return json.Marshal(int64(l))
+	case Float32Literal:
+		return json.Marshal(float32(l))
+	case Float64Literal:
+		return json.Marshal(float64(l))
```
dillitz: I think this errors on non-finite values (`json: unsupported value: NaN` / `+Inf`), so a float predicate with a `NaN`/`Inf` bound can't be serialized at all. The parse side also rejects the bare `NaN`/`Infinity` tokens Java emits. These are constructible via `NewLiteral`, so it's reachable, though uncommon (NaN checks usually go through `is-nan`/`not-nan`). Might be worth deciding the policy explicitly: return a wrapped error, or special-case them to match Java. Worth noting Java's bare `NaN`/`Infinity` tokens aren't valid standard JSON anyway.

expr_json.go
```diff
@@ -0,0 +1,502 @@
...
+	var t transformNode
+	if err := json.Unmarshal(b, &t); err != nil {
+		return nil, fmt.Errorf("%w: cannot parse term: %s", ErrInvalidArgument, err)
+	}
+	switch t.Type {
+	case "reference":
+		// {"type":"reference","term":"name"}
+		return Reference(t.Term), nil
+	case exprKeyTransform:
+		return nil, fmt.Errorf("%w: transform terms are not supported when parsing expressions", ErrNotImplemented)
```
dillitz: `encodeTerm` serializes a `*BoundTransform` to `{"type":"transform",...}`, but `decodeTerm` rejects that same shape with `ErrNotImplemented`. A transform term is a legal `Term` per the spec ([`Term` = `Reference | TransformTerm`](https://github.com/apache/iceberg/blob/85ffa1984e115e80ba1571f3eb017fcf0ba39031/open-api/rest-catalog-open-api.yaml#L2651-L2674)), and Java reads it back ([`ExpressionParser.term()`](https://github.com/apache/iceberg/blob/85ffa1984e115e80ba1571f3eb017fcf0ba39031/core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java#L404-L417)), so a legal residual filter whose term is a partition transform would fail to parse here. The NotImplemented error seems fine if we don't want to support this yet.

expr_json.go
```diff
@@ -0,0 +1,502 @@
...
+		lit, err := decodeValue(node.Value, typ)
+		if err != nil {
+			return nil, err
+		}
+
+		return LiteralPredicate(op, term, lit), nil
+	case op >= OpIn && op <= OpNotIn:
+		if node.Values == nil {
+			return nil, fmt.Errorf("%w: predicate %s is missing values", ErrInvalidArgument, node.Type)
+		}
```
dillitz: The unary branch rejects a stray `value`/`values`, but the literal-op branch doesn't reject a stray `values` and the set branch doesn't reject a stray `value`, so `{"type":"eq","term":"a","value":1,"values":[1,2]}` parses to `eq(a,1)` instead of erroring. Java rejects all three combinations ([`predicateFromJson`](https://github.com/apache/iceberg/blob/85ffa1984e115e80ba1571f3eb017fcf0ba39031/core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java#L331-L375)).

dillitz (commented):
Thank you for working on this! I did a pass and found a few things; the `timestamptz` handling appears like the most critical finding.

expr_json.go
rambleraptor: Great catch! Fixed this.

expr_json.go
rambleraptor: Yeah, this is already complicated enough (as you've seen with your great review!). Let's have that be a follow-up.

expr_json.go
```diff
@@ -0,0 +1,542 @@
...
+	OpNotStartsWith: "not-starts-with",
+	OpIn:            "in",
+	OpNotIn:         "not-in",
+	OpNot:           "not",
+	OpAnd:           "and",
+	OpOr:            "or",
+}
+
+var jsonToOp = func() map[string]Operation {
+	m := make(map[string]Operation, len(opToJSON))
```
zeroshade, line 62: I'm surprised there isn't a `maps.Invert` function in the stdlib lol

expr_json.go
```diff
@@ -0,0 +1,542 @@
...
+// With a schema, literals take the referenced field's type (e.g. "2022-08-14"
+// on a date column becomes a DateLiteral). Without one they fall back to the
+// base JSON kind: Int64Literal, Float64Literal, StringLiteral, or BoolLiteral.
+func ParseExpr(data []byte, schema *Schema) (BooleanExpression, error) {
+	return decodeExpr(json.RawMessage(data), schema)
+}
+
+func encodeExpr(e BooleanExpression) (json.RawMessage, error) {
+	switch v := e.(type) {
+	case AlwaysTrue:
```
zeroshade: Why not just have AlwaysTrue's MarshalJSON hard-coded as `return []byte("true")`?

expr_json.go
```diff
@@ -0,0 +1,542 @@
...
+		return json.RawMessage("true"), nil
+	case AlwaysFalse:
+		return json.RawMessage("false"), nil
```
zeroshade: Same comment as above

expr_json.go
```diff
@@ -0,0 +1,542 @@
...
+	case NotExpr:
+		child, err := encodeExpr(v.child)
```
zeroshade: Couldn't NotExpr's MarshalJSON just call MarshalJSON on its child and then do the return?

Ultimately, why do we even need this giant encodeExpr function when we're just using a large switch anyways. Each individual bit can just live in the appropriate MarshalJSON function for that expr type instead of using this large switch

expr_json.go
```diff
@@ -0,0 +1,542 @@
...
+		typ = bt.Type()
+	}
+
+	node := exprNode{Type: js}
+	if node.Term, err = encodeTerm(term); err != nil {
+		return nil, err
+	}
+
+	switch {
+	case op >= OpIsNull && op <= OpNotNan:
```
zeroshade: Don't we have types for these that we can just implement MarshalJSON on instead of needing to explicitly switch on this?

expr_json.go
```diff
@@ -0,0 +1,542 @@
...
+		return nil, fmt.Errorf("%w: cannot serialize expression with operation %s", ErrInvalidArgument, op)
+	}
+
+	return json.Marshal(node)
+}
+
+func encodeTerm(term Term) (json.RawMessage, error) {
+	switch t := term.(type) {
+	case Reference:
+		return json.Marshal(string(t))
```
zeroshade: Same thing here, this logic can just live in the MarshalJSON methods for each of these types

expr_json.go
```diff
@@ -0,0 +1,542 @@
...
+		return nil, fmt.Errorf("%w: cannot serialize term of type %T", ErrInvalidArgument, term)
+	}
+}
+
+// encodeLiteral writes a non-null literal in the JSON form for its Iceberg type
+// (see Java's SingleValueParser). typ is the resolved field type, used to tell a
+// timestamptz literal from a plain timestamp; it may be nil for an unbound term.
+func encodeLiteral(lit Literal, typ Type) (json.RawMessage, error) {
+	switch l := lit.(type) {
+	case BoolLiteral:
```
zeroshade: Again, just add MarshalJSON methods to the literals instead of this

expr_json.go
```diff
@@ -0,0 +1,542 @@
...
+		return p.lits.Members(), nil
+	case BoundSetPredicate:
+		return p.Literals().Members(), nil
+	default:
+		return nil, fmt.Errorf("%w: expected a set predicate, got %T", ErrInvalidArgument, e)
+	}
+}
+
+func decodeExpr(raw json.RawMessage, schema *Schema) (BooleanExpression, error) {
+	b := bytes.TrimSpace(raw)
```
zeroshade: Use json.NewDecoder and get the tokens instead of having to deal with spaces and trying to parse the JSON yourself.

zeroshade (changes requested):
You should really encapsulate the logic in the explicit MarshalJSON and UnmarshalJSON functions on each type rather than using the large switch in a huge encode/decodeExpr function. You can use a similar intermediate helper like we do for the types in types.go to handle discriminators

expr_json.go
rambleraptor: I might be able to find the right person to complain to lol. The go stdlib has always had some interesting omissions.

rambleraptor:
Yeah, great call about moving things towards MarshalJSON. One of the big reasons I had a giant function was to keep type information intact. We still need something like that (you need type information, especially for Timestamp types as @dillitz pointed out), but everything is much simpler + more Go-native now.

expr_json.go
```diff
@@ -0,0 +1,594 @@
...
+	return nil
+}
+
+// marshalTimestamp emits the REST form: with the UTC offset for timestamptz,
+// without it for timestamp.
+func marshalTimestamp(tm time.Time, typ Type) ([]byte, error) {
+	switch typ.(type) {
+	case TimestampTzType, TimestampTzNsType:
+		// "-07:00" prints +00:00 for a UTC instant, matching Java.
+		return json.Marshal(tm.Format("2006-01-02T15:04:05.999999-07:00"))
```
laskoviymishka, line 395: This breaks round-trip for any sub-second timestamp, and it's the one I'd most want fixed before merge.

We marshal a fractional form here, but the parse side (`StringLiteral.To(TimestampTzType)` uses `time.RFC3339`, `To(TimestampType)` uses `"2006-01-02T15:04:05"` — both with no fractional component) can't read it back. A server emitting a valid `TimestampTzTypeValue` like `"2022-08-14T10:00:00.123456+00:00"` parses to an error. The current test only exercises a whole-second value, so this passes CI — I'd add a sub-second case to make the break visible.

The parse-side fix lives in `literals.go` (switch those formats to `RFC3339Nano` / a fractional-aware layout — Java's `isoTimestamptzToMicros` uses `ISO_DATE_TIME`, which accepts arbitrary fractional precision); I'd be careful not to change `StringLiteral.To` semantics for existing callers without checking them. Separately, the `TimestampNsType`/`TimestampTzNsType` branches here cap at six 9s and silently truncate nanoseconds — the OpenAPI `TimestampNanoTypeValue` example is `"2007-12-03T10:15:30.123456789"`, so the ns branches need nine 9s. wdyt — fix in `literals.go` this PR, or pull the timestamp work into a follow-up gated behind a documented limitation?

expr_json.go
```diff
@@ -0,0 +1,594 @@
...
+
+	var node exprNode
+	if err := json.Unmarshal(raw, &node); err != nil {
+		return nil, fmt.Errorf("%w: cannot parse expression: %s", ErrInvalidArgument, err)
+	}
+
+	// {"type":"literal","value":<bool>} is an alternate spelling of a constant.
+	if node.Type == "literal" {
+		var bv bool
+		if err := json.Unmarshal(node.Value, &bv); err != nil {
```
laskoviymishka, line 438: `{"type":"literal","value":null}` silently misparses as `AlwaysFalse{}`.

`json.Unmarshal([]byte("null"), &bv)` succeeds and leaves `bv` at its zero value `false`, so a null (or absent) value falls through to `AlwaysFalse{}` with no error. For a planning filter that means a corrupt payload quietly becomes "filter everything out" rather than a rejected request.

I'd check for JSON null (and empty `node.Value`) explicitly before the bool unmarshal and error out. A `{"type":"literal","value":null}` case in `TestUnmarshalExpressionErrors` would lock it.

expr_json.go
```diff
@@ -0,0 +1,594 @@
...
+	var t transformNode
+	if err := json.Unmarshal(raw, &t); err != nil {
+		return nil, fmt.Errorf("%w: cannot parse term: %s", ErrInvalidArgument, err)
+	}
+	switch t.Type {
+	case "reference":
+		// {"type":"reference","term":"name"}
+		return Reference(t.Term), nil
+	case exprKeyTransform:
+		return nil, fmt.Errorf("%w: transform terms are not supported when parsing expressions", ErrNotImplemented)
```
laskoviymishka: We can marshal transform terms (`BoundTransform.MarshalJSON`) but not parse them — `decodeTerm` returns `ErrNotImplemented` for `"transform"`. Java's `term()` parses these via `Transforms.fromString`, so a Java server returning a residual like `{"type":"lt-eq","term":{"type":"transform","transform":"bucket[100]","term":"id"},"value":50}` makes the whole scan fail on our side.

For the stated server-side-planning goal I think shipping this as a documented known-gap is acceptable for now, but not as a silent one. I'd add a godoc note on `ParseExpr` calling out that transform terms in residuals aren't yet parsed, and open a tracking issue. If you'd rather close the gap here, a Go equivalent of `Transforms.fromString` plus a `BoundTransform`/unbound transform-term construction is the path. Which way do you want to take it?

expr_json.go
```diff
@@ -0,0 +1,594 @@
...
+func decodePredicate(op Operation, node exprNode, schema *Schema) (BooleanExpression, error) {
+	term, err := decodeTerm(node.Term)
+	if err != nil {
+		return nil, err
+	}
+
+	// Resolve the field type once; nil means schema-less.
+	var typ Type
+	if schema != nil {
+		bound, err := term.Bind(schema, false)
```
laskoviymishka: `caseSensitive` is hardcoded `false` here, so name resolution is silently case-insensitive even though REST field names are authoritative. I'd thread a `caseSensitive` param through — an internal `parseExpr(data, schema, caseSensitive)` with the public `ParseExpr` passing `true`, documented as "schema is used for type resolution with case-sensitive name matching, result is unbound." If we deliberately want lenient matching, that's defensible too — but it should be explicit rather than a buried `false`.

expr_json.go
```diff
@@ -0,0 +1,594 @@
...
+	case BinaryType, FixedType:
+		s, ok := base.(StringLiteral)
+		if !ok {
+			return nil, fmt.Errorf("%w: expected a string for %s value", ErrInvalidArgument, typ)
+		}
+		decoded, err := hex.DecodeString(string(s))
+		if err != nil {
+			return nil, fmt.Errorf("%w: invalid hex for %s value: %s", ErrInvalidArgument, typ, err)
+		}
+		if _, ok := typ.(FixedType); ok {
```
laskoviymishka: The decoded length isn't validated against `FixedType.Len()`, so a structurally invalid `FixedLiteral` flows downstream unchecked. Note `StringLiteral.To(FixedType)` does reject a length mismatch, so this path is the odd one out.

```go
if ft, ok := typ.(FixedType); ok {
	if len(decoded) != ft.Len() {
		return nil, fmt.Errorf("%w: fixed[%d] value has %d bytes", ErrInvalidArgument, ft.Len(), len(decoded))
	}
	return FixedLiteral(decoded), nil
}
```

expr_json.go
```diff
@@ -0,0 +1,594 @@
...
+	if schema != nil {
+		bound, err := term.Bind(schema, false)
+		if err != nil {
+			return nil, err
+		}
+		typ = bound.Type()
+	}
+
+	switch {
+	case op >= OpIsNull && op <= OpNotNan:
```
laskoviymishka: These range checks lean on `iota` ordering, and the "do not change the order" warning lives at the definition site in `exprs.go`, not here. A new op inserted mid-range would silently get misclassified as unary/literal/set. I'd make the classification explicit — either spell out the ops as `switch` cases, or build `unaryOps`/`literalOps`/`setOps` sets keyed off `opToJSON` — so the categorization can't drift out from under us.

laskoviymishka (changes requested):
This is a clean, well-structured foundation for the server-side planning work — the `literalValue{Literal, typ}` wrapper threading the field type through timestamp serialization, the `opToJSON`/`jsonToOp` inverse maps, and the bound/unbound `MarshalJSON` parity are all how I'd build this, and the exact-bytes assertions are exactly the wire-parity guard I want on a file like this.

The one I care most about: sub-second timestamps don't round-trip. We marshal a fractional form, but the parse path (`StringLiteral.To` with `RFC3339` / a no-fractional layout) can't read fractional seconds back, so a real payload like `"2022-08-14T10:00:00.123456+00:00"` fails to parse. The whole-second test value is what's keeping CI green. The fix lives mostly in `literals.go`, and the ns-timestamp branches separately truncate to microseconds and need the full nine digits.

Second: `{"type":"literal","value":null}` silently parses to `AlwaysFalse{}`, because `json.Unmarshal("null", &bool)` succeeds and leaves `false`. For a planning filter that turns a corrupt payload into "filter everything out" rather than an error — I'd reject it explicitly.

- sub-second timestamp round-trip (marshal + parse), plus nanosecond precision for the ns types
- `{"type":"literal","value":null}` should error, not become `AlwaysFalse`
- thread `caseSensitive` through `ParseExpr` instead of hardcoding `false`
- validate `FixedType` length on parse
- make the unary/literal/set op classification explicit rather than relying on iota ranges
- decide and document the transform-term parse gap — I'm fine treating it as a known-gap with a godoc note and a tracking issue, but not as a silent `ErrNotImplemented`

Left the smaller idiom and test-coverage notes out of this pass since it's an unblocking PR; happy to take another look once the above are addressed.

expr_json.go
rambleraptor: Did the second. Thanks!

expr_json.go
rambleraptor: If you're up for the review, let's just fix it while we're here. I made those changes.

expr_json.go
rambleraptor: Let's just fix it while we're here. If you're up for the review, I'd rather try to get it right here, rather than have it linger for an undetermined amount of time.

rambleraptor:
@laskoviymishka wow, thank you so much for the deep review! I added the changes you made. It does make this a *long* PR, so let me know if you would prefer if I split up some of those follow-ups after all.

laskoviymishka:
@rambleraptor i think it's fine :) this is foundational work for big feature.

laskoviymishka (approved):
I think we are good to merge here.

Great work @rambleraptor!

A few things to keep in mind for the next PRs around this scope:

1. Test the full round trip. Marshal-only and parse-only tests are not enough. For every wire type, do `marshal → parse → equal` with values that actually exercise the format: fractional timestamps, mixed-case hex, multi-element sets, negative and large integers.

2. Use test values that can fail meaningfully. Whole-second timestamps, `01 02 03`, or simple ordered sets often pass by accident. Pick values where encoding details matter.

3. Reject `null` when it would turn into a meaningful zero value. `json.Unmarshal("null", &bool)` succeeds and gives `false`. On REST input, that can silently change semantics. Treat absent/null explicitly instead of letting it become a default.

4. Don’t classify enums by `iota` range. Checks like `op >= OpX && op <= OpY` break when enum order changes. Use an explicit switch or set.

5. Keep marshal and parse symmetric. If we can emit something we cannot read back, either implement both sides or document the gap and link a follow-up. Don’t leave it as a silent `ErrNotImplemented`.

6. Thread policy flags through the call stack. Things like `caseSensitive` should not become hardcoded literals deep in helpers. Pass them through and make the public default explicit.

7. Validate decoded values against the declared type. Fixed-length hex, decimal precision/scale, and similar cases should be checked before constructing the literal.

8. Verify parity claims against the actual Java/PyIceberg code. For expressions, check formats, field names, ordering, set handling, and hex casing instead of relying on “mirrors Java” in the PR description.

9. Keep the code idiomatic to the package. `slices.SortFunc`, consistent receivers, proper `%w` wrapping, and similar details are small, but they add friction in review when they drift.

laskoviymishka:
Will wait for @zeroshade  input here.

rambleraptor:
I appreciate the follow-up comments! I'll definitely keep those in mind. Appreciate you going through a PR with this size. I'll also do my best to avoid these kinds of large PRs in the future.

zeroshade (approved):
LGTM! Nice job!

PR #1214, opened 2026-06-16
feat(table): public helper to build datafile from in-heap parquet metadata

### LINKED ISSUE #1186: feat(table): public helper to build datafile from in-heap parquet metadata

External writers that create Parquet files in-process already hold `*metadata.FileMetaData` after `(*file.Writer).Close()`. After uploading the bytes to object storage, they need a fully populated iceberg.DataFile with per-column statistics to commit via `Transaction.AddDataFiles` or `RowDelta`.

Today the only path is `Transaction.AddFiles(filePaths)`, which re-opens each file from storage to read the footer — the exact metadata the caller already has in memory. This adds unnecessary I/O on every commit.

**Proposed Solution**

A new exported function at package table, which would validate the inputs, call the already existing internals to extract statistics, set EqualityFieldIDs when Content == EntryContentEqDeletes and return the DataFile.

**Contribution**

I’d be happy to work on this enhancement. Please feel free to assign this issue to me and I can prepare a PR implementing it.

badalprasadsingh:
## Feature

Adds an exported helper in the table package to construct an iceberg.DataFile directly from in-memory metadata (*metadata.FileMetaData).

Fixes #1186 

## Testing
- Created in-memory parquets (both data and deletes) using `arrow-go` and added into the iceberg table using an `overwrite` operation 
- Ingested into an unpartitioned iceberg table in Glue Catalog and queried using Trino


zeroshade (changes requested):
I've managed to keep all Parquet specific stuff limited to the internal package and exposed via generic FileFormat interfaces, can we move this and expose something generic instead?

badalprasadsingh:
Hey @zeroshade, 

1. Removed the `parquet/metadata` import.
2. The exported fields are now format-agnostic — `iceberg.FileFormat` selects the implementation and Metadata `any` carries the format-specific in-memory metadata object.
3. Tried to keep the dispatch generic via the internal `FileFormat` something similar to what we have in `writer.go`.
4. Kept `Metadata` as any.

Please take a look whenever you are free.

table/data_file_meta.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to build stats plan: %w", err)
+	}
+
+	pathMapping, err := format.PathToIDMapping(args.Schema)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build path mapping: %w", err)
+	}
+
+	stats := format.DataFileStatsFromMeta(
```
laskoviymishka: This is a public function that can panic instead of returning an error, and the `stats == nil` guard doesn't catch the real risk.

`DataFileStatsFromMeta` does a hard `meta.(*metadata.FileMetaData)` assertion — pass any other concrete type with `Format: ParquetFile` and it panics, never reaching the nil check. The column-mapping path inside it and the multi-value `PartitionValue()` path inside `ToDataFile` panic too. Internally this is fine because `filesToDataFiles` wraps `fileToDataFile` in `recover()`, but here there's no net, so a bad arg crashes the caller's goroutine.

I'd at minimum comma-ok the metadata type up front and return something like `fmt.Errorf("unsupported metadata type for %s: %T", args.Format, args.Metadata)`, and wrap the body in a deferred `recover()` that converts the deeper panics into error returns. wdyt?

table/data_file_meta.go
```diff
@@ -0,0 +1,178 @@
...
+	if len(args.EqualityFieldIDs) > 0 {
+		stats.EqualityFieldIDs = append(stats.EqualityFieldIDs, args.EqualityFieldIDs...)
+	}
+
+	partitionValues := args.PartitionValues
+	if partitionValues == nil {
+		partitionValues = make(map[int]any)
+	}
+
+	df := stats.ToDataFile(tblutils.DataFileOpts{
```
laskoviymishka: `FirstRowID` never gets set, which means the result is dead on arrival for v3 tables.

`Transaction.AddDataFiles` enforces that every `DataFile` carries a `FirstRowID` on format-version-3 tables, and this helper is documented as the entry point into exactly that. So a v3 caller following the godoc gets an immediate rejection at commit.

I'd add a `FirstRowID *int64` to `DataFileArgs` and plumb it through to the builder, with a doc note that it's required for v3. Thoughts?

table/data_file_meta.go
```diff
@@ -0,0 +1,178 @@
...
+		Schema:          args.Schema,
+		Spec:            args.Spec,
+		Path:            args.FilePath,
+		Format:          args.Format,
+		Content:         args.Content,
+		FileSize:        args.FileSize,
+		PartitionValues: partitionValues,
+		SortOrderID:     args.SortOrderID,
```
laskoviymishka: Positional deletes need a null sort order id, and we pass `args.SortOrderID` unconditionally here.

The spec requires pos-delete files to set sort order id to null, and our own writer enforces this at `writer.go:134` ("position delete file claims sort order id %d; the spec requires a null sort order id"). This helper happily accepts a nonzero `SortOrderID` for `EntryContentPosDeletes` and produces a file the writer would have rejected.

I'd return an error when `Content == EntryContentPosDeletes && SortOrderID != UnsortedSortOrderID` so the two paths agree.

table/data_file_meta.go
```diff
@@ -0,0 +1,178 @@
...
+	// Schema is the Iceberg schema the file's contents conform to. The
+	// file's columns MUST carry per-leaf field-ID metadata whose values
+	// match this schema's field IDs (for Parquet this is the
+	// "PARQUET:field_id" key on each leaf).
+	Schema *iceberg.Schema
+
+	// Spec is the partition spec the resulting DataFile is bound to —
+	// typically tbl.Spec(). For an unpartitioned table pass
+	// *iceberg.UnpartitionedSpec.
+	Spec iceberg.PartitionSpec
```
laskoviymishka, line 49: There's no validation that partition values line up with the spec, and the failure modes here are bad.

A zero-value `PartitionSpec{}` equals the unpartitioned spec, so a caller who forgets `Spec` on a partitioned table silently gets a `DataFile` with no partition info — which breaks partition pruning on other clients and only surfaces as a `validateDataFilePartitionData` failure at commit. Worse, if they do supply `PartitionValues` but the file spans more than one value for a field, `ToDataFile → PartitionValue()` panics (see the recover comment above).

I'd validate that when `Spec` is partitioned, `PartitionValues` has an entry per field, and return an error rather than deferring to a commit-time failure or a crash. Also a small doc nit: the comment says pass `*iceberg.UnpartitionedSpec` but the field is a value `iceberg.PartitionSpec` — the zero value is already equivalent, so I'd just say that.

table/data_file_meta.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, errors.New("file metadata is required")
+	}
+	if args.FilePath == "" {
+		return nil, errors.New("file path is required")
+	}
+	if args.FileSize <= 0 {
+		return nil, fmt.Errorf("file size must be greater than 0, got %d", args.FileSize)
+	}
+
+	switch args.Content {
```
laskoviymishka, line 205: We validate `Content` shape but never validate the field IDs against the schema, which is the check the real writers do.

For eq-deletes the spec says every id in `equality_ids` must be present in the delete file, and `equality_delete_writer.go:44` validates exactly that at write time; here we accept any ints. For pos-deletes the spec mandates the reserved field IDs (`2147483546`/`2147483545`) and we never check the schema matches `iceberg.PositionalDeleteSchema`. Either way an invalid file gets built and is only caught at `RowDelta.Commit`, after the bytes are already uploaded.

I'd loop the `EqualityFieldIDs` through `args.Schema.FindFieldByID` for the eq case, and verify schema compatibility with `PositionalDeleteSchema` for the pos case. wdyt?

table/data_file_meta.go
```diff
@@ -0,0 +1,178 @@
...
+		statsPlan,
+		pathMapping,
+		tblutils.VariantFieldIDsFromSchema(args.Schema),
+	)
+	if stats == nil {
+		return nil, errors.New("failed to build data file stats from metadata")
+	}
+
+	if len(args.EqualityFieldIDs) > 0 {
+		stats.EqualityFieldIDs = append(stats.EqualityFieldIDs, args.EqualityFieldIDs...)
```
laskoviymishka: `stats.EqualityFieldIDs` is always nil coming out of `DataFileStatsFromMeta`, so the `append` is just a misleading way to write an assignment — it implies accumulation that isn't happening, and if `DataFileStatsFromMeta` ever does populate it we'd silently duplicate. The internal write path uses a direct assignment; I'd match it:

```go
stats.EqualityFieldIDs = args.EqualityFieldIDs
```

table/data_file_meta.go
```diff
@@ -0,0 +1,178 @@
...
+	switch args.Content {
+	case iceberg.EntryContentData,
+		iceberg.EntryContentEqDeletes,
+		iceberg.EntryContentPosDeletes:
+	default:
+		return nil, fmt.Errorf("invalid content: %v", args.Content)
+	}
+
+	if args.Content == iceberg.EntryContentEqDeletes && len(args.EqualityFieldIDs) == 0 {
+		return nil, errors.New("equalityFieldIDs is required for equality-delete files")
```
laskoviymishka: Small one: these messages say `equalityFieldIDs` but the exported field is `EqualityFieldIDs`, and the rest of the codebase uses the exported casing (e.g. `row_delta.go:134` "equality delete file must have non-empty EqualityFieldIDs"). Lowercase doesn't grep back to the symbol — I'd capitalize both.

table/data_file_meta_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, df)
+
+	assert.Equal(t, iceberg.EntryContentData, df.ContentType())
+	assert.Equal(t, iceberg.ParquetFile, df.FileFormat())
+	assert.Equal(t, filePath, df.FilePath())
+	assert.Equal(t, int64(len(pqBytes)), df.FileSizeBytes())
+	assert.EqualValues(t, 3, df.Count())
+
+	assert.Len(t, df.ValueCounts(), 2, "one value-count entry per column")
```
laskoviymishka: These assertions only check that the bounds maps have two entries — they'd pass even if every bound were zeroed or the min/max were swapped, so the stats path isn't really being exercised.

For `id` 1-3 and `name` alpha/gamma the bounds are computable, so I'd decode at least one field with `iceberg.LiteralFromBytes` and assert the actual lower/upper values. While we're here, all three tests are happy-path and near-identical skeletons — I'd fold them into one table-driven `TestDataFileFromMetadata` with subtests and add the error cases (nil schema, nil/wrong-type metadata, empty path, non-positive size, invalid content, eq-delete with empty `EqualityFieldIDs`). Several of those paths panic today rather than error, so the table is how a fix here actually gets caught.

laskoviymishka (changes requested):
Thanks for putting this together, this is a genuinely handy addition, and the overall shape is great. Routing through the same `computeStatsPlan → PathToIDMapping → DataFileStatsFromMeta → ToDataFile` pipeline the internal writer uses is exactly the right instinct: stats encoding stays identical to the normal write path, and not touching the filesystem is the correct shape for an `AddDataFiles`/`NewRowDelta` entry point. The Trino round-trip in the description is reassuring too.

I left a handful of inline comments. Most are small, but since this is a brand-new exported surface I'd love to get the contract nailed down before it lands — these are hard to walk back once people start building on them.

The two I'd most want to resolve first:

- `DataFileFromMetadata` can panic instead of returning an error. The `if stats == nil` guard doesn't catch the real risk — a wrong `Metadata` concrete type hits a hard type assertion deep in `DataFileStatsFromMeta` and panics, as do the partition-value and column-mapping paths in `ToDataFile`. The internal writer survives this only because `filesToDataFiles` wraps it in `recover()`; this helper has no net, so a bad arg crashes the caller's goroutine. Since this is public, I think callers will reasonably expect an error there.
- v3 tables: `AddDataFiles` rejects any `DataFile` without `FirstRowID` set on format-version-3 tables, and this helper never sets it — so the documented happy path doesn't work on v3 yet.

The rest are smaller things I'd love to settle in this PR:

- convert the internal panics (wrong metadata type, multi-value partition, missing column mapping) into errors via an up-front comma-ok assertion plus a deferred recover
- add `FirstRowID` to `DataFileArgs` and plumb it through for v3
- reject a nonzero `SortOrderID` for positional deletes, the way the writer already does
- validate field IDs against the schema — `EqualityFieldIDs` present in the schema for eq-deletes, schema compatible with `PositionalDeleteSchema` for pos-deletes
- validate partition values against the spec instead of letting a forgotten `Spec`/`PartitionValues` fail at commit or panic
- reconsider `Metadata any` — at a public boundary it type-erases the one dependency the compiler and godoc should be checking; a small interface or a concrete `*metadata.FileMetaData` field would make the requirement explicit
- the tests are all happy-path and assert only bound-map lengths — I'd decode and check at least one real min/max, fold the three into a table-driven test, and add the error cases

None of this is far off, and the core is solid — happy to take another pass as soon as these are in. Thanks again!

table/data_file_meta.go
zeroshade: I agree here, we need to have a corresponding change in the `parquet_files.go` to update the assertion so that we can error rather than panic on the wrong metadata type.

and definitely add a deferred recover to catch the deeper panics. We currently let the internal packages panic because the internal writer is always somewhere below a recover and it gives us a fast abort path. But if we're going to create a public interface like this, we need to prevent panics at the public API boundary.

table/data_file_meta.go
```diff
@@ -0,0 +1,178 @@
...
+	// Format identifies the on-disk file format of the file being
+	// described and selects the internal FileFormat implementation used
+	// to extract per-column statistics from Metadata.
+	Format iceberg.FileFormat
+
+	// Metadata is the format-specific, in-memory file metadata object
+	// produced by the external writer (for example
+	// *parquet/metadata.FileMetaData when Format is
+	// [iceberg.ParquetFile]).
+	Metadata any
```
zeroshade, line 61: I wonder if there's a small interface we can define so we can add at least *some* kind of constraint (not a blocking nit)

zeroshade (changes requested):
@laskoviymishka caught most of the issues I see already, just one more nit: please add a test that exercises a partitioned spec since that's the most panic-prone path and is currently entirely uncovered. While this helper takes Spec and PartitionValues, neither is validated so a mismatch will either panic in `PartitionValue`/`ToDataFile` or will fail later at the commit `validateDataFilePartitionData`.

table/data_file_meta.go
badalprasadsingh: Done.

1. Added a deferred `recover()` that converts the deeper panics (column-mapping, multi-value `PartitionValue()` into error returns.
2. Added an upfront comma-ok assertion for unsupported metadata type.

Left the `parquet_files.go` assertion as it is - happy to change it in a separate PR.

table/data_file_meta.go
badalprasadsingh: Done.

Added `FirstRowID *int64` to `DataFileArgs`. Added a test as well for the same.

table/data_file_meta.go
badalprasadsingh: Done.

table/data_file_meta.go
badalprasadsingh: Done.

1. Added the validation logic in `validatePartitionValues` for the same: when the `spec` is _partitioned_, every `spec` field must have a value and no stray `field IDs` are allowed, otherwise we error instead of deferring to commit/panic.
2. Fixed the doc to just say the zero-value `PartitionSpec` is equivalent to `unpartitioned`.

table/data_file_meta.go
badalprasadsingh: Done.

1. **eq-delete** IDs are checked against `Schema.FindFieldByID`
2. **pos-delete** schemas are verified for the reserved field IDs/types against `PositionalDeleteSchema`.

table/data_file_meta.go
badalprasadsingh: Done.

table/data_file_meta.go
badalprasadsingh: Done.

table/data_file_meta_test.go
badalprasadsingh: Done.

Added a single table-driven `TestDataFileFromMetadata` with subtests that decode real lower/upper bounds via `LiteralFromBytes`, plus a errors table covering nil schema, nil/wrong-type metadata, empty path, non-positive size, invalid content, and the eq/pos-delete validation paths.

table/data_file_meta.go
badalprasadsingh: Kept `Metadata any` for now with the explicit _comma-ok_ check at the boundary.

Happy to introducing a small marker interface in a follow-up PR.

badalprasadsingh:
Hey @laskoviymishka , @zeroshade thank you so much for the review. I have addressed all of the above mentioned comments.

The two major _blockers_:
1. **Panic -> errors**: Added an up-front _comma-ok_ assertion on `Metadata`, plus a deferred `recover()` in `DataFileFromMetadata`.
2. **V3 tables**: Added `FirstRowID *int64` to `DataFileArgs` and plumbed it through `ToDataFile`.

The other smaller comments, all in:
1. Positional deletes now reject a nonzero `SortOrderID`, matching the writer.
2. Field-ID validation: `EqualityFieldIDs` must exist in the schema for _eq-deletes_, and the schema must be compatible with `PositionalDeleteSchema` ( _reserved IDs_ present and correctly typed ) for _pos-deletes_.
3. Partition values are validated against the _spec_ up front ( missing / extra field IDs ) instead of failing at commit or panicking.
4. Tests are now _table-driven_, decode and assert real _min/max_ bounds (not just map lengths), and cover the error cases — including the recover path and the validation failures above.
5. Added `validatePartitionValues` so a _Spec/PartitionValues_ mismatch fails at the boundary instead of at commit, and the one mismatch that depends on file stats (multi-value partition) is now caught by a _deferred_ `recover()` and returned as an error. <br></br> Added partitioned-spec tests covering the happy path, missing value, unknown field id, and the multi-value panic path. cc: @zeroshade 

On `Metadata`: I have kept it as `any` but made the requirement explicit and fail-fast via the comma-ok assertion. Reason: the entry point routes through the internal FileFormat seam (`DataFileStatsFromMeta(meta Metadata, …)`), which is intentionally format-agnostic.



Ready for another _pass_ whenever you have a moment — thanks again!

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	df = stats.ToDataFile(tblutils.DataFileOpts{
+		Schema:          args.Schema,
+		Spec:            args.Spec,
+		Path:            args.FilePath,
+		Format:          args.Format,
+		Content:         args.Content,
+		FileSize:        args.FileSize,
+		PartitionValues: partitionValues,
```
laskoviymishka: I think there's a real problem here, and the current tests don't catch it because they only exercise files where the partition column is constant. `ToDataFile` (utils.go ~211-237) calls `d.PartitionValue(field, opts.Schema)` and prefers the *stats-inferred* value over the `PartitionValues` we pass in — and that inference panics with "more than one value for partition field" whenever the stats show more than one distinct value for a partition column.

So an identity-partitioned file spanning `id=1..10` panics, gets caught by the blanket recover, and comes back as an opaque error — even though the caller supplied the correct partition value. That's exactly the #1186 case: an external writer registering an already-written partitioned file.

I'd skip the stats-based `d.PartitionValue` for any field the caller supplied a non-nil value for (or thread a suppress-inference flag through `DataFileOpts`). The existing `fileToDataFile` path supplies no caller values, so it stays unaffected. A test with a non-constant partitioned column would lock this down. wdyt?

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+func validateContentSchema(args DataFileArgs) error {
+	switch args.Content {
+	case iceberg.EntryContentEqDeletes:
+		for _, id := range args.EqualityFieldIDs {
+			if _, ok := args.Schema.FindFieldByID(id); !ok {
+				return fmt.Errorf(
+					"EqualityFieldIDs contains field id %d which is not present in the supplied schema", id)
+			}
+		}
+	case iceberg.EntryContentPosDeletes:
```
laskoviymishka, line 382: There's no way to set `referenced_data_file` here, and no path to the V3 `content_offset` / `content_size_in_bytes` (spec notes 4-5 require these for deletion vectors). So a positional-delete file built through this helper lands without the data-file reference — legal-but-suboptimal for V2 (engines lose the data-file scoping), and not enough if anyone reaches for V3 DV semantics.

The PR mentions this was exercised with pos-deletes via Trino — do the entries it produced actually carry `referenced_data_file`, or is Trino just tolerating its absence? I'd either add a `ReferencedDataFile` field plumbed to the builder, or scope this PR to data + equality deletes and split pos-deletes into a follow-up. At minimum the DV scope limit deserves a line in the doc comment. Shipping a half-wired pos-delete path on a public API is the part that's hard to undo.

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+	df = stats.ToDataFile(tblutils.DataFileOpts{
+		Schema:          args.Schema,
+		Spec:            args.Spec,
+		Path:            args.FilePath,
+		Format:          args.Format,
+		Content:         args.Content,
+		FileSize:        args.FileSize,
+		PartitionValues: partitionValues,
+		SortOrderID:     args.SortOrderID,
+		FirstRowID:      args.FirstRowID,
```
laskoviymishka: Two spec problems here, both from §First Row ID Inheritance, and the field doc ("Set it for v3 data files") points callers the wrong way.

For delete files, `first_row_id` must always be null — and nothing rejects a non-nil `FirstRowID` on eq/pos-delete content; it's plumbed straight through. For *newly written* data files it must also be null: "its first_row_id field is set to null because it is not assigned until the snapshot is successfully committed" — row IDs come from inheritance at commit. Since this helper produces ADDED-status entries for `Transaction.AddDataFiles`, the spec-correct value for the normal case is nil, and hand-supplying one risks drifting from the table's `next-row-id` counter under concurrent commits.

I'd gate `bldr.FirstRowID(...)` on `opts.Content == EntryContentData` over in `table/internal/utils.go` (the `if opts.FirstRowID != nil` block), reject a non-nil `FirstRowID` on delete content here, and reword the doc to: "nil for newly-written files — row IDs are assigned by inheritance at commit; set only when re-registering an existing file whose first_row_id was previously assigned; always nil for delete files." Worth asking whether exposing `FirstRowID` at all is right for an append-oriented helper.

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+	// Format identifies the on-disk file format of the file being
+	// described and selects the internal FileFormat implementation used
+	// to extract per-column statistics from Metadata.
+	Format iceberg.FileFormat
+
+	// Metadata is the format-specific, in-memory file metadata object
+	// produced by the external writer (for example
+	// *parquet/metadata.FileMetaData when Format is
+	// [iceberg.ParquetFile]).
+	Metadata any
```
laskoviymishka, line 61: `Format` and `Metadata` are two fields that have to agree, and the only thing enforcing that is the runtime type-assert below — a caller passing `ParquetFile` plus an Avro metadata object compiles fine and fails at call time. Since Parquet is the only supported format today, `any` isn't buying us real generality yet.

I'd lean toward typing it now (`Metadata *metadata.FileMetaData`, and drop or default `Format`) while there are no callers to break. If we want `any` for forward-compat, that's defensible, but then I'd document the Format↔Metadata invariant much more loudly right here. wdyt?

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+// (delete files); it is NOT committed automatically. The file bytes
+// described by args.Metadata must already have been uploaded to
+// args.FilePath.
+func DataFileFromMetadata(args DataFileArgs) (df iceberg.DataFile, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			df = nil
+			switch e := r.(type) {
+			case error:
+				err = fmt.Errorf("error while building an iceberg datafile: %w", e)
```
laskoviymishka: Wrapping the entire function in `recover` means a genuine bug in this package's own code (a nil-map write, an index-out-of-range) gets laundered into a generic "error while building" rather than surfacing as the panic it is. It's also currently masking the two real, reachable input failures (the partition-override panic, and the missing data-schema validation in `validateContentSchema`).

Once those are validated up front, I'd either drop the blanket recover or scope it tightly to the `DataFileStatsFromMeta` + `ToDataFile` calls, with a comment naming which call panics and why. Two small things while you're in here: `df = nil` is redundant (the named return is already nil), and the message wording ("error while building an iceberg datafile") diverges from the package's "error encountered during …" pattern in `arrow_utils.go` / `equality_delete_writer.go`.

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+		FileSize:        args.FileSize,
+		PartitionValues: partitionValues,
+		SortOrderID:     args.SortOrderID,
+		FirstRowID:      args.FirstRowID,
+	})
+
+	return df, nil
+}
+
+func validateContentSchema(args DataFileArgs) error {
```
laskoviymishka, line 371: The eq-delete and pos-delete branches validate their schema up front, but there's no `case iceberg.EntryContentData`, so a data file with an empty or mismatched schema slips through to `computeStatsPlan` + `DataFileStatsFromMeta` and panics there ("column chunk %q not found in column mapping" in parquet_files.go) — again laundered through the recover into an opaque error.

A schema/Parquet mismatch is a likely caller mistake for this new API, and it's currently untested. I'd add a superset check here (Parquet field IDs ⊇ Iceberg schema field IDs) so it returns a clean error. If the asymmetry is intentional, an explicit `case iceberg.EntryContentData: // no schema constraints` with a one-line comment would at least make it read as deliberate.

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+	switch args.Content {
+	case iceberg.EntryContentEqDeletes:
+		for _, id := range args.EqualityFieldIDs {
+			if _, ok := args.Schema.FindFieldByID(id); !ok {
```
laskoviymishka: This checks the ID exists but not that it resolves to a leaf primitive. You can't equality-delete on a struct/list/map field, and Java's writers enforce primitivity here. A caller passing the ID of a nested struct field would pass this check and produce an entry other clients reject. I'd tighten this to also assert the resolved field's type is primitive.

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+	return nil
+}
+
+func validatePartitionValues(spec iceberg.PartitionSpec, values map[int]any) error {
+	if spec.Equals(*iceberg.UnpartitionedSpec) {
+		return nil
+	}
+
+	expected := make(map[int]string, spec.NumFields())
+	for _, field := range spec.Fields() {
```
laskoviymishka, line 409: Separate from the inference panic: once caller-supplied values actually get used, the type of each value starts to matter — an identity transform over an int column wants `int32`, a bucket transform wants the `int32` bucket index, and a void-transform field must be nil. This loop validates presence and rejects extras but doesn't check any of that.

Is type-checking the partition tuple something we want to own here, or is that the caller's responsibility? If the latter, a line in the `PartitionValues` doc (including the void→nil rule) would help. wdyt?

table/data_file_meta.go
```diff
@@ -0,0 +1,270 @@
...
+	if args.Metadata == nil {
+		return nil, errors.New("file metadata is required")
+	}
+	pqMeta, ok := args.Metadata.(*metadata.FileMetaData)
+	if !ok {
+		return nil, fmt.Errorf(
+			"unsupported metadata type: expected *metadata.FileMetaData, got %T",
+			args.Metadata)
+	}
+	if pqMeta == nil {
```
laskoviymishka, line 195: Both nil-checks are correct and both are reachable: the `args.Metadata == nil` check above only catches a literally-unset field, while a typed-nil `(*metadata.FileMetaData)(nil)` slips past it and gets caught here. Nice that the test already locks this in. I'd add a one-line comment explaining why both exist, otherwise someone will "simplify" one away later. One small thing: both paths return the same "file metadata is required" — a distinct message for the typed-nil case ("metadata pointer is nil") would tell a confused caller which mistake they made.

table/data_file_meta_test.go
```diff
@@ -0,0 +1,477 @@
...
+					iceberg.NestedField{ID: 2147483545, Name: "pos", Type: iceberg.PrimitiveTypes.String, Required: true},
+				)
+			},
+			wantSub: "must have type",
+		},
+		{
+			name: "invalid metrics mode property fails stats plan",
+			mutate: func(a *table.DataFileArgs) {
+				a.Properties = iceberg.Properties{
+					table.DefaultWriteMetricsModeKey: "truncate(0)",
```
laskoviymishka, line 782: A few gaps worth closing while the API is new. The only metrics-mode coverage is this error path (`truncate(0)`) — there's no positive case for `truncate(N)` truncating bounds or `none` suppressing stats, and that's exactly where cross-client bound parity lives (Java/PyIceberg default to `truncate(16)`). The happy-path cases also never assert `NullValueCounts` or `ColumnSizes`, and null counts drive pruning.

`TestDataFileFromMetadata_FirstRowID` only checks the set case — the v1/v2 nil case (`df.FirstRowID()` should be nil) is worth pinning. There's no multi-row-group fixture, where split offsets and aggregated counts can break silently, and no schema-mismatch case to cover the panic path above. If it's not too much, an end-to-end case that hands the result to `txn.AddDataFiles` / `NewRowDelta` would be the strongest signal the output is actually consumable. (Minor: the substring checks could use `require.ErrorContains`, the repo idiom.)

table/data_file_meta_test.go
```diff
@@ -0,0 +1,477 @@
...
+			mutate: func(a *table.DataFileArgs) {
+				a.Spec = iceberg.NewPartitionSpec(iceberg.PartitionField{
+					SourceIDs: []int{1},
+					FieldID:   1000,
+					Name:      "id",
+					Transform: iceberg.IdentityTransform{},
+				})
+				a.PartitionValues = map[int]any{1000: int32(1)}
+			},
+			wantSub: "more than one value for partition field",
```
laskoviymishka: This case is green only because of the blanket recover — it's pinning the recovered-panic string as if it were the contract, and it's the same underlying panic as the partition-override issue. Once that's fixed and partition handling is validated up front, this should flip to asserting on a clean validation error, or become a success case where the caller-supplied value is used without inference.

laskoviymishka (changes requested):
This is a solid addition overall. The stats path reuses `DataFileStatsFromMeta`, the bounds are checked against real values, and the partition round-trip is covered. The direction for #1186 makes sense.

A few issues here are more than cleanup.

The main one is partition handling. `ToDataFile` prefers the partition inferred from stats over `PartitionValues`, and that inference panics if the partition column is not constant in the file. So a file with `id=1..10` fails even when the caller supplied the correct partition value. The current tests only use constant partition columns, so they miss this. Caller-provided values should skip inference for those fields.

`first_row_id` also needs another look. Per spec, it must be null for delete files and newly written data files; commit inheritance assigns it later. Since this helper creates `ADDED` entries, the current doc points callers in the wrong direction, and nothing restricts it by content type.

The positional-delete API is also incomplete: there’s no way to set `referenced_data_file` or the v3 DV fields. Since this is a new exported surface, I’d either add that support now or explicitly leave pos-deletes out of scope.

Before merge I’d settle:

* caller partition values override stats inference
* `first_row_id` is rejected for deletes and documented correctly for new data files
* pos-delete support is completed or scoped out
* validate Parquet field IDs against the schema, or document why not
* replace `Metadata any` with a typed value, or clearly document the `Format`/`Metadata` contract
* narrow the blanket `recover` after validating known panic cases
* require equality field IDs to point to primitive leaves
* add coverage for metrics mode, multiple row groups, null counts, nil `FirstRowID`, and schema mismatch

Once the partition override is fixed, happy to take another pass.

table/data_file_meta.go
badalprasadsingh: Done, you are right.

1. Now, we prioritize caller-supplied values whenever the field has any non-nil caller value. We fallback to getting it from column statistics in-case the user doesn't supply it. 
2. Added a test for the same.

table/data_file_meta.go
badalprasadsingh: Well,

1. I have limited the scope of this PR to _not_ include `V3 deletion vectors` in the scope of this public function right now. Though, position delete files are accepted for `V2` spec.
2. `referenced_data_file` cannot be set for all position deletes. Someone writing with a [DeleteGranuliarity](https://iceberg.apache.org/javadoc/1.5.2/org/apache/iceberg/deletes/DeleteGranularity.html) as Partition might have `null` as the value for this field, because he/she might not have a single referenced data file for the position deletes. A case which we are missing in [validateNoNewDeletesForRewrittenFiles](https://github.com/apache/iceberg-go/blob/54df1f058f2b813b99f2b42148bcc375b5a45c72/table/conflict_validation.go#L623-L634) function, where the compaction skips the position delete files with reference data file value as `null` (But they can be parition-scoped). <br>
Happy to raise an issue for the same.

table/data_file_meta.go
badalprasadsingh: Well,

In Iceberg-Go, I find externally written parquet files it cannot cannot 'fabricate row IDs for files it did not write', check [here](https://github.com/apache/iceberg-go/blob/54df1f058f2b813b99f2b42148bcc375b5a45c72/table/transaction.go#L589-L598). 

Thus, we have to expose the field to the user to supply the value. Later on, we can deprecate it, once iceberg-go's add path assigns `first_row_id` automatically via inheritance. 

Added the other validations which was asked.

table/data_file_meta.go
badalprasadsingh: Done.

Chose to keep `any` for forward-compat and documented the invariant loudly on `Format`, `Metadata`, and the function doc.

table/data_file_meta.go
badalprasadsingh: Done.

Recover is now scoped to just the stats-extraction + build closure with a comment explaining the two panic sources; all up-front input failures return clean errors before it. Others `nit` level fixes done as well.

table/data_file_meta.go
badalprasadsingh: Done.

Made the asymmetry deliberate with an explicit `EntryContentData` case + comment, and added `TestDataFileFromMetadata_SchemaMismatchYieldsError` proving a mismatch yields a clean wrapped error.

table/data_file_meta.go
badalprasadsingh: Done.

Added the primitive-leaf assertion with a dedicated error and a struct-field test case.

table/data_file_meta.go
badalprasadsingh: Done.

Documented this as caller responsibility (including the void→nil rule) on the PartitionValues field, and added a void-transform test confirming nil is accepted and recorded.

table/data_file_meta.go
badalprasadsingh: Done. Added the explanatory comment.

table/data_file_meta_test.go
badalprasadsingh: Done.

Added the tests for:
- Metrics modes: `TestDataFileFromMetadata_MetricsModes` (none, counts, truncate(2)).
- Null counts + column sizes: `TestDataFileFromMetadata_NullCountsAndColumnSizes`.
nil FirstRowID: `TestDataFileFromMetadata_FirstRowIDNil`.
- Multiple row groups: `TestDataFileFromMetadata_MultiRowGroup` (split offsets + aggregated bounds/count).
- Schema mismatch: `TestDataFileFromMetadata_SchemaMismatchYieldsError`.
- End-to-end: `TestDataFileFromMetadata_EndToEndAddDataFiles`

table/data_file_meta_test.go
badalprasadsingh: Done.

table/internal/utils.go
```diff
@@ -264,16 +268,12 @@ func (d *DataFileStatistics) ToDataFile(opts DataFileOpts) iceberg.DataFile {
...
-				val := d.PartitionValue(field, opts.Schema)
-				if val != nil {
-					fieldIDToPartitionData[field.FieldID] = val
-				} else {
-					fieldIDToPartitionData[field.FieldID] = partitionVal
-				}
+				fieldIDToPartitionData[field.FieldID] = partitionVal
 			} else {
-				fieldIDToPartitionData[field.FieldID] = nil
+				fieldIDToPartitionData[field.FieldID] = d.PartitionValue(field, opts.Schema)
```
laskoviymishka, line 277: This inverts the precedence for every existing caller of `ToDataFile`, not just the new helper. The nil branch now calls `d.PartitionValue`, which `must(...)`-panics on a multi-valued column, and `ParquetFileWriter.Close` reaches this via `w.partition` with no guarantee every spec field is present. A misrouted file whose rows span more than one partition now panics with no recover at that call site — the old else-branch set nil unconditionally, which was wrong-but-safe.

I'd gate the infer branch on `field.Transform.PreservesOrder()` so non-order-preserving transforms stay absent and the panic is reachable only from `DataFileFromMetadata`, where the recover catches it. There's also no `utils_test.go` case exercising `ToDataFile` with a partitioned spec and a non-nil `PartitionValues[fieldID]` — worth adding one that asserts the caller's value is recorded.

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil, fmt.Errorf("failed to build path mapping: %w", err)
+	}
+
+	partitionValues := args.PartitionValues
+	if partitionValues == nil {
+		partitionValues = make(map[int]any)
+	}
+
+	var df iceberg.DataFile
+	// DataFileStatsFromMeta and ToDataFile are the only calls wrapped in recover because they may panic on invalid input
```
laskoviymishka, line 278: The comment says both calls can panic, but `DataFileStatsFromMeta` returns nil on failure (you check it right below) — only `ToDataFile` panics. Bundling both under one recover with the same "error encountered during extracting stats and building the data file" prefix means a genuine programmer error (nil deref, index OOB) in the builder gets swallowed into a wrapped string, indistinguishable from a recoverable schema mismatch.

I'd move `DataFileStatsFromMeta` out of the closure and, at minimum, capture `runtime/debug.Stack()` in the recover so a real bug isn't invisible. wdyt?

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+	// Content == [iceberg.EntryContentEqDeletes] and MUST be empty
+	// for every other Content value. Every listed ID must be present in Schema.
+	EqualityFieldIDs []int
+
+	// FirstRowID sets the data file's first_row_id (row lineage, v3+).
+	//
+	// Leave it nil for newly-written files — row IDs are assigned by inheritance at commit;
+	// set only when re-registering an existing file whose first_row_id was previously assigned;
+	// always nil for delete files.
+	FirstRowID *int64
```
laskoviymishka, line 134: `first_row_id` (field 142) and `referenced_data_file` (field 143) are both v3-only — they're absent from the v1/v2 data-file Avro schemas. Nothing here checks the table's format version, so a caller who sets `FirstRowID` (or `ReferencedDataFile`) against a v2 table gets it silently dropped by the Avro encoder: no error, no warning, and they believe row lineage or the delete correlation was recorded.

I'd either thread a `FormatVersion` into `DataFileArgs` and error when these are set below v3, or document plainly that both are silently ignored on pre-v3 tables. For a public API, erroring feels better. wdyt?

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+	Content iceberg.ManifestEntryContent
+
+	// PartitionValues maps spec field ID → partition value. Every spec field
+	// must be present; leave nil/empty for unpartitioned tables, where any
+	// supplied value is rejected.
+	//
+	// Values must already be the transform results in the expected Go types.
+	// A nil value is treated as "not supplied" and is inferred from the file's
+	// column statistics; void-transform fields therefore should be nil (the
+	// void transform always yields null), though this is not enforced.
```
laskoviymishka: This reads as contradictory: "Every spec field must be present" up top, then "void-transform fields therefore should be nil ... though this is not enforced" here. A caller will reasonably read the second half as "I can omit void fields" and then hit `missing partition value for field id %d` at runtime, since `validatePartitionValues` requires the key to be present.

I'd replace "though this is not enforced" with something explicit — void-transform fields must still appear in the map with a nil value; omitting the key entirely is an error.

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+
+	if args.Content == iceberg.EntryContentEqDeletes && len(args.EqualityFieldIDs) == 0 {
+		return nil, errors.New("EqualityFieldIDs is required for equality-delete files")
+	}
+
+	if args.Content != iceberg.EntryContentEqDeletes && len(args.EqualityFieldIDs) > 0 {
+		return nil, fmt.Errorf("EqualityFieldIDs must be empty for content %v", args.Content)
+	}
+
+	if args.Content == iceberg.EntryContentPosDeletes && args.SortOrderID != UnsortedSortOrderID {
```
laskoviymishka: The spec says pos-delete files "should" set sort_order_id to null and readers "must ignore" it — that's a should for writers, not a must, and the Java writer doesn't error here. A file legitimately pre-sorted by (file_path, pos) can carry a real sort order. It's also asymmetric with eq-deletes, which accept any `SortOrderID` a few lines up.

I'd relax this to a no-op — or at most validate the ID references a real sort order — rather than reject outright. wdyt?

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+	// FileFormat implementation used to extract per-column statistics from
+	// Metadata. It must match the concrete type of Metadata: the only supported
+	// combination today is [iceberg.ParquetFile] with a *metadata.FileMetaData.
+	Format iceberg.FileFormat
+
+	// Metadata is the format-specific, in-memory file metadata object produced
+	// by the external writer.
+	//
+	// Today only Format == [iceberg.ParquetFile] with *metadata.FileMetaData is supported
+	Metadata any
```
laskoviymishka, line 61: With `Metadata any`, passing `Format == iceberg.OrcFile` and ORC metadata returns "unsupported metadata type: expected *metadata.FileMetaData" — but the real problem is that ORC isn't supported, not that the metadata type is wrong. Since Parquet is the only supported combo today, I'd either make this `*metadata.FileMetaData` directly and drop the type assertion, or split the error so "unsupported format" reads distinctly from "wrong metadata type".

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+		// from the schema's field-id mapping and reports
+		// them as an error through the scoped recover.
+	case iceberg.EntryContentEqDeletes:
+		for _, id := range args.EqualityFieldIDs {
+			f, ok := args.Schema.FindFieldByID(id)
+			if !ok {
+				return fmt.Errorf(
+					"EqualityFieldIDs contains field id %d which is not present in the supplied schema", id)
+			}
+			if _, ok := f.Type.(iceberg.PrimitiveType); !ok {
```
laskoviymishka: This rejects any equality field whose type isn't a `PrimitiveType`, but the existing writer path (`equality_delete_writer.go`'s `validateEqualityFieldIDs`) only rejects floating-point types. Two entry points now disagree on what a valid equality field is. I'd align with the existing writer — reject floats, allow the rest — so callers don't get different answers depending on which door they came through.

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil, err
+	}
+
+	if err := validatePartitionValues(args.Spec, args.PartitionValues); err != nil {
+		return nil, err
+	}
+
+	format := tblutils.GetFileFormat(args.Format)
+	if format == nil {
+		return nil, fmt.Errorf("unsupported file format: %s", args.Format)
```
laskoviymishka: `iceberg.FileFormat` is a named string type, and perfsprint (enabled in `.golangci.yml`) flags `%s` on those — this will likely trip the linter in CI. `errors.New("unsupported file format: " + string(args.Format))` sidesteps it.

table/data_file_meta_test.go
```diff
@@ -0,0 +1,949 @@
...
+
+	result, err := tx.Commit(t.Context())
+	require.NoError(t, err)
+
+	snap := result.CurrentSnapshot()
+	require.NotNil(t, snap)
+	assert.Equal(t, table.OpAppend, snap.Summary.Operation)
+	assert.Equal(t, "1", snap.Summary.Properties["added-data-files"])
+	assert.Equal(t, "2", snap.Summary.Properties["added-records"])
+}
```
laskoviymishka, line 1228: The two end-to-end tests assert snapshot summary counts and the pos-delete ContentType, but never read the committed manifest entries back to check that LowerBoundValues/UpperBoundValues/NullValueCounts/ColumnSizes survived the round-trip. A regression that dropped column stats somewhere between here and commit wouldn't be caught. I'd iterate the manifest entries after Commit and assert at least one field's bounds and null count made it through.

table/data_file_meta_test.go
```diff
@@ -0,0 +1,949 @@
...
+		Metadata: pqMeta,
+		FilePath: "s3://bucket/data/mismatch.parquet",
+		FileSize: int64(len(pqBytes)),
+		Content:  iceberg.EntryContentData,
+	})
+	require.Error(t, err)
+	require.Nil(t, df)
+	require.ErrorContains(t, err, "error encountered during extracting stats and building the data file")
+	require.ErrorContains(t, err, "not found in column mapping")
+}
```
laskoviymishka, line 991: This covers the strict-subset mismatch (a column dropped), but not the case the doc explicitly warns about: a same-count schema with a renamed column. Since field-id metadata isn't consulted, a rename doesn't error — it silently produces a DataFile with missing bounds, which is the more dangerous path. I'd add a negative test showing a renamed column returns no error but yields empty bounds for that field, so the silent-drop contract is pinned down.

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+
+		return nil
+	}
+
+	expected := make(map[int]string, spec.NumFields())
+	for _, field := range spec.Fields() {
+		expected[field.FieldID] = field.Name
+		if _, ok := values[field.FieldID]; !ok {
+			return fmt.Errorf("missing partition value for field id %d (%s) in spec id %d",
```
laskoviymishka: Requiring every field ID to be present means a caller of a 10-field spec who knows 2 values and wants inference for the rest has to enumerate all 10 with explicit nils. I'd consider treating an absent key as equivalent to nil (infer) — it reads more naturally, and it would also dissolve the void-transform confusion in the `PartitionValues` doc above. Not blocking, just a contract call worth making deliberately. Thoughts?

laskoviymishka (changes requested):
Really nice to see a clean public path from in-heap Parquet metadata to a DataFile without a filesystem round-trip — the helper is well-factored and the test matrix (error cases, metrics modes, multi-row-group aggregation, end-to-end AddDataFiles and RowDelta) is genuinely thorough.

The one that worries me most isn't in the new helper — it's the partition-value precedence change in `ToDataFile`. Inverting the logic so the nil branch calls `d.PartitionValue` changes behavior for every existing caller, and `ParquetFileWriter.Close` reaches that branch with `w.partition`. A misrouted file whose rows span more than one partition now panics through `must(...)` with no recover at that call site, where the old code was wrong-but-safe. I'd gate the infer branch behind `PreservesOrder()` so the panic stays reachable only from `DataFileFromMetadata`.

The other thing I'd settle before this goes in is the v3-only fields on the public struct. `FirstRowID` and `ReferencedDataFile` are both silently dropped by the Avro encoder on a v1/v2 table — a caller who sets them gets no error and believes they were recorded. For a public API I'd rather error when they're set below v3 than no-op quietly.

- gate the `ToDataFile` nil-infer branch on `PreservesOrder()` and add a `ToDataFile` regression test with a partitioned spec + non-nil caller value
- error (or clearly document) `FirstRowID`/`ReferencedDataFile` on pre-v3 tables
- split the recover so a real builder panic isn't swallowed into the same string as a schema mismatch
- reconcile the contradictory `PartitionValues` doc so void-transform fields have an unambiguous contract
- relax the positional-delete sort-order rejection to match the spec's "should"

The rest are smaller — a lint fix, a validation-consistency mismatch with the existing eq-delete writer, and a couple of test gaps — all noted inline.

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+	EqualityFieldIDs []int
+
+	// FirstRowID sets the data file's first_row_id (row lineage, v3+).
+	//
+	// Leave it nil for newly-written files — row IDs are assigned by inheritance at commit;
+	// set only when re-registering an existing file whose first_row_id was previously assigned;
+	// always nil for delete files.
+	FirstRowID *int64
+
+	// ReferencedDataFile, when non-nil, records the single data file that a
```
zeroshade, line 136: [MAJOR] This says newly-written files should leave FirstRowID nil, but v3 AddDataFiles requires data files to carry first_row_id, so callers following this doc can still build a DataFile that only fails at commit time. Please add table-version/target-operation context to DataFileArgs validation/docs so v3 AddDataFiles requires FirstRowID, and add a v3 AddDataFiles test.

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+	//
+	// For partition-scoped position deletes whose file_path column references more
+	// than one data file (cardinality > 1) — such a file has no single referenced data file.
+	ReferencedDataFile *string
+}
+
+// DataFileFromMetadata builds a fully populated [iceberg.DataFile] from a
+// file's in-memory, format-specific metadata object, extracting per-column
+// statistics (column_sizes, value_counts, null_value_counts, lower_bounds,
+// upper_bounds), split offsets, and the manifest-entry shape required for
```
zeroshade, line 150: [MAJOR] ReferencedDataFile is accepted based only on being non-empty, without checking that the position-delete file actually scopes to that data file. Please require field 2147483546 (file_path) to have lower and upper bounds that decode to the supplied reference when ReferencedDataFile is set; otherwise return an error.

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+	// greater than zero.
+	FileSize int64
+
+	// Content selects which manifest-entry kind to produce — one of
+	// [iceberg.EntryContentData], [iceberg.EntryContentEqDeletes] or
+	// [iceberg.EntryContentPosDeletes].
+	Content iceberg.ManifestEntryContent
+
+	// PartitionValues maps spec field ID → partition value. Every spec field
+	// must be present; leave nil/empty for unpartitioned tables, where any
```
zeroshade: [MINOR] The PartitionValues docs still read as contradictory: every spec field must be present, but nil means infer. Please clarify that present-with-nil is the inference signal, or change validation to treat absent keys as inference too.

table/data_file_meta.go
```diff
@@ -0,0 +1,349 @@
...
+	switch args.Content {
+	case iceberg.EntryContentData,
+		iceberg.EntryContentEqDeletes,
+		iceberg.EntryContentPosDeletes:
+	default:
+		return nil, fmt.Errorf("invalid content: %v", args.Content)
+	}
+
+	if args.Content == iceberg.EntryContentEqDeletes && len(args.EqualityFieldIDs) == 0 {
+		return nil, errors.New("EqualityFieldIDs is required for equality-delete files")
```
zeroshade: [NIT] Error strings use exported field-name casing inconsistently (for example EqualityFieldIDs/FirstRowID/ReferencedDataFile). Consider normalizing these messages if you touch this area again.

zeroshade (changes requested):
Requesting changes for remaining DataFileFromMetadata correctness gaps.

Re-review note: prior feedback around panic-to-error conversion and partition coverage is largely addressed; the items below still block this helper from being safe for the advertised commit paths.

- v3 AddDataFiles can still fail at commit time because FirstRowID is documented as optional and there is no v3 AddDataFiles coverage.
- ReferencedDataFile is accepted without proving it matches the position-delete file_path stats.
- PartitionValues docs still conflict around required keys versus nil-driven inference.
- Error strings use exported field-name casing inconsistently.

table/data_file_meta.go
badalprasadsingh: Well, `DataFileStatsFromMeta` can _panic_ as well on a schema/file mismatch (panics with "not found in column mapping").

Thus, not moving it out as it would break the recovery.

table/data_file_meta.go
badalprasadsingh: Well, I fear, `referenced_data_file` is **not** a v3-only spec. Its _optional_ for v2 spec. Please check [here](https://github.com/apache/iceberg/blob/2d3f5c24e0697b361fe95362d8ee13b915e98048/format/spec.md?plain=1#L737).

Thus, kept `first_row_id` for v3 or more than v3 spec only. Letting the caller set the value for `referenced_data_file` in v2 spec.

table/data_file_meta.go
badalprasadsingh: Done.

1. Now, ignoring the `key` or supplying the key with `nil` value behaves the same - not supplied by the user.
2. Updated the spec as well.

table/data_file_meta.go
badalprasadsingh: Done. 

Relaxed it to a no-op.

table/data_file_meta.go
badalprasadsingh: Done.

Added a `tblutils.GetFileFormat(args.Format)` upfront to avoid any file format other than parquet, as a separate error than "wrong metadata type".

table/data_file_meta.go
badalprasadsingh: Done.

Now it uses the same `validateEqualityFieldIDs` validation here. Thus, keeping the same answers for both call paths.

table/data_file_meta.go
badalprasadsingh: Done.

table/data_file_meta_test.go
badalprasadsingh: Done.

table/data_file_meta_test.go
badalprasadsingh: Done.

Added the test `TestDataFileFromMetadata_RenamedColumnErrors` for the same.

table/data_file_meta.go
badalprasadsingh: Done.

From now on, passing just the key with `nil` value and skipping the key entirely - behaves the same.

table/data_file_meta.go
badalprasadsingh: Done.

1. `FirstRowID` is now REQUIRED only for `v3` spec.
2. Explicitly pointed out in the spec when should the caller supply its value and when it can pass nil and let it get supplied from inheritance according to the user's use cases.
3. Added the test `TestDataFileFromMetadata_V3AddDataFilesRequiresFirstRowID` for the same.

table/data_file_meta.go
badalprasadsingh: Done. 

1. Added a validation to check if the lower_bounds or upper_bounds match the `reference_data_file` value.
2. Made sure the `MetricsModeColumnConfPrefix+".file_path"` is explicitly set to `full`.

table/data_file_meta.go
badalprasadsingh: Done.

No longer, the caller needs to supply all the fields. For both the cases, not passing the key or passing the key with nil value requires inference from stats.

table/data_file_meta.go
badalprasadsingh: Done.

table/internal/utils.go
badalprasadsingh: Done.

Added inference on `field.Transform.PreservesOrder()`. So a caller value always wins, non-order-preserving transforms stay nil, and the `must(...)` panic is only reachable from `DataFileFromMetadata`, where the recover converts it to an error. Added the regression tests for it in `utils_test.go`.

badalprasadsingh:
Thank you @laskoviymishka & @zeroshade so much for the re-reivew. I have addressed the comments, please do take a look and tell me if anything else is required from my side.

Just a quick heads up: `referenced_data_file` is not a v3-only spec. Its _optional_ for v2 spec. Please check [here](https://github.com/apache/iceberg/blob/2d3f5c24e0697b361fe95362d8ee13b915e98048/format/spec.md?plain=1#L737). Thus, keeping it exposed for users to set it for position deletes in v2 spec (file-scoped position delete files). Happy to know if any further fix is required for the same in iceberg-go.

table/data_file_meta.go
```diff
@@ -0,0 +1,417 @@
...
+		if statsProps == nil {
+			statsProps = iceberg.Properties{}
+		}
+		statsProps[MetricsModeColumnConfPrefix+".file_path"] = string(tblutils.MetricModeFull)
+	}
+
+	statsPlan, err := computeStatsPlan(args.Schema, statsProps)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build stats plan: %w", err)
+	}
```
laskoviymishka, line 265: Non-blocking, but cheap to foreclose: this override key is hardcoded to `"file_path"` while the metrics plan resolves the same column via `schema.FindColumnName(fieldID)`. The spec only pins the reserved id and type for the pos-delete path column, not its name — so a schema that names that field anything else silently misses the full-metrics override, falls back to `truncate(16)`, and then `verifyReferencedDataFile` fails for any path over 16 bytes (or passes spuriously for short ones). Resolving via `args.Schema.FindColumnName(referencedDataFilePathFieldID)` sidesteps it. Fine as a follow-up if you'd rather not touch it here.

table/data_file_meta.go
```diff
@@ -0,0 +1,417 @@
...
+			if r := recover(); r != nil {
+				err = fmt.Errorf(
+					"error encountered during extracting stats and building the data file: %v\n%s",
+					r, debug.Stack())
+			}
+		}()
+
+		stats := format.DataFileStatsFromMeta(
+			args.Metadata,
+			statsPlan,
```
laskoviymishka, line 294: Non-blocking nit: `debug.Stack()` bakes a raw goroutine stack into an error string that reaches public callers — every other recover site in the package (`arrow_utils.go`, `equality_delete_writer.go`) wraps a clean message without the trace, so I'd drop it for consistency. Separately, `r` here can itself be an error (the `PartitionValue` panic is a `fmt.Errorf` value), so a type-assert + `%w` would preserve `errors.Is/As` for callers. Take it or leave it.

table/data_file_meta_test.go
```diff
@@ -0,0 +1,1244 @@
...
+	require.NoError(t, err)
+
+	assert.EqualValues(t, 2, df.NullValueCounts()[2], "name column has 2 nulls")
+	assert.EqualValues(t, 0, df.NullValueCounts()[1], "id column has no nulls")
+
+	require.NotEmpty(t, df.ColumnSizes(), "column sizes must be populated")
+	assert.Contains(t, df.ColumnSizes(), 1)
+	assert.Contains(t, df.ColumnSizes(), 2)
+}
+
```
laskoviymishka, line 939: Optional coverage: the new `PreservesOrder()` infer branch also fires for truncate (it preserves order), but there's no truncate case here. A sibling to this test with a truncate-partitioned column — shared prefix infers, mixed prefix errors cleanly rather than panicking through recover — would pin that path. Not blocking.

laskoviymishka (approved):
This has converged, approving :shipit: .

Every hard blocker from the last rounds now has a fix and a test behind it: panic to error, v3 `FirstRowID` plumbing and below-v3 rejection, caller partition values overriding stats inference, the `PreservesOrder()` gate, v3 `AddDataFiles` coverage, and `verifyReferencedDataFile` proving the file-scoped delete matches its `file_path` bounds.

The end-to-end matrix is thorough for a new public surface: `AddDataFiles`, `RowDelta`, replace/rewrite, metrics modes, and multi-row-group coverage.

And you’re right on `referenced_data_file`, it’s optional in v2 for file-scoped position deletes, not v3-only. Good catch, keeping it exposed for v2 is correct.

On `Metadata any`, I still slightly prefer a typed constructor, but your reasoning holds. This goes through the `internalFileFormat` seam, which is intentionally format-agnostic, and the comma-ok assertion keeps the failure mode clear. Not going to relitigate it, this is a reasonable call and reversible if we add ORC/Avro later.

Left three non-blocking nits inline: the `file_path` override key resolving by literal vs schema name, one wording issue, and a truncate-inference test. None of them block the merge, take or leave them, or fold into a follow-up.

Very nice long substantial PR <3.

badalprasadsingh:
Thank you so much, @laskoviymishka! 🙏

@zeroshade, please let me know if there's anything else you'd like me to address for this feature.

<br>

**EDIT**: Folded all three nits in this PR rather than deferring.

table/data_file_meta.go
badalprasadsingh: Done.

We now resolve the metrics-override key via `args.Schema.FindColumnName(referencedDataFilePathFieldID)`, instead of the hard-coded `file_path`. Really nice observation.

table/data_file_meta.go
badalprasadsingh: Done.

Dropped `debug.Stack()`. This now matches the shape like the other mentioned paths.

table/data_file_meta_test.go
badalprasadsingh: Done.

Added the test `TestDataFileFromMetadata_TruncatePartitionInferred` with two subsets:
- _shared prefix_ : `"alphabet"`, `"alpine"` → both truncate to `"alp"`, this infers `"alp"` as the partition value
- _mixed prefix_ : `"alpha"`, `"beta"` → `"alp"` vs `"bet"`, returns the recovered `"cannot infer partition value"` error

table/data_file_meta.go
```diff
@@ -0,0 +1,421 @@
...
+		return nil, errors.New("unsupported file format: " + string(args.Format))
+	}
+
+	// Both nil checks are intentional:
+	//   - args.Metadata == nil (above) catches an unset interface, while
+	//   - pqMeta == nil below catches a typed-nil (*metadata.FileMetaData)(nil)
+	pqMeta, ok := args.Metadata.(*metadata.FileMetaData)
+	if !ok {
+		return nil, fmt.Errorf(
+			"unsupported metadata type for format %s: expected *metadata.FileMetaData, got %T",
```
zeroshade, line 192: nit: this error formats `iceberg.FileFormat` with `%s`, whereas the sibling error a few lines up uses `string(args.Format)`. Worth normalizing if the perfsprint linter flags it — not blocking.

zeroshade (approved):
LGTM — the panic→error conversion, partition-value precedence, v3 FirstRowID handling, and metrics/test coverage from the earlier rounds all look addressed. One tiny lint nit inline.

badalprasadsingh:
Thank you so much @zeroshade! Addressed the inline comment.

badalprasadsingh:
Thanks @laskoviymishka!

Issue #1216, opened 2026-06-16
feat(table): partition-match predicate builder for dynamic overwrite (phase 1)


laskoviymishka:
Phase 1 of #1215.

## What

Add an internal, transform-aware helper that, given the current partition spec and a deduplicated set of partition tuples (as produced by `DataFile.Partition()`, keyed by partition field ID), returns a `BooleanExpression` matching exactly those partitions and nothing else.

## Shape

- Input: the partition spec plus a set of partition tuples (field-ID → already-transformed value).
- Output: an OR of per-partition predicates; each per-partition predicate is an AND across the spec's fields. A null partition value maps to `IsNull`; a present value maps to an equality built through the field's `Transform`, not an identity-only switch on Go types.
- Deduplicate tuples canonically so repeated partitions collapse into one predicate.
- Empty input maps to `AlwaysFalse`.

## Why standalone

This is the only genuinely new logic in the feature and the part most likely to need iteration (transform handling, null semantics). Landing it alone keeps the Phase 2 wiring trivial and the predicate independently testable.

## Open design point

The mechanism for non-identity transforms (e.g. matching files in `bucket(x) = 3` or `day(ts) = N`) needs to be settled here so Phase 2 can consume it directly — either a source-column row predicate where the transform is invertible, or partition-level matching against the manifest tuple. Resolving this is part of the phase.

## Tests

- Identity transform, single and multiple fields.
- Non-identity transform (bucket / day) handled correctly.
- Null partition value → `IsNull`.
- Duplicate tuples dedup to a single predicate.
- Empty set → `AlwaysFalse`.

## Parity

Mirror partition matching in `BaseReplacePartitions` / PyIceberg `_build_partition_predicate`.

Issue #1215, opened 2026-06-16
feat(table): Support Dynamic Partition Overwrite (ReplacePartitions)


laskoviymishka:
## Goal

Add `Transaction.DynamicPartitionOverwrite` — the iceberg-go equivalent of Java's `ReplacePartitions` (`BaseReplacePartitions`) and PyIceberg's `Transaction.dynamic_partition_overwrite`. Given an Arrow table, it detects the partitions present in the incoming data, atomically deletes the existing data in exactly those partitions, and appends the new data, leaving untouched partitions alone.

## Background

A first attempt was made in #482, but it predates the partitioned-write and copy-on-write-overwrite machinery that has since landed, so most of that PR is now redundant:

- Partitioned writes are native — `recordsToDataFiles` routes to the partitioned fanout / clustered writers.
- Overwrite-by-filter already exists — `Transaction.Overwrite`, `performCopyOnWriteDeletion`, `mergeOverwrite`.
- Transform-aware predicate projection exists — `Transform.Project` — which removes the identity-transform-only limitation of the original attempt.

What remains is small and composable: derive a partition-matching predicate from the written data files, then drive the existing overwrite path. Landing it in reviewable slices rather than one drop.

## Scope (decomposable across PRs)

- **Phase 1 — Partition-match predicate.** A transform-aware helper that turns a set of touched partition tuples plus the partition spec into a `BooleanExpression` selecting exactly those partitions. Standalone and unit-tested, no transaction wiring.
- **Phase 2 — `DynamicPartitionOverwrite` API.** Compose the partitioned write, touched-partition collection via `DataFile.Partition()`, the Phase 1 predicate, and the existing copy-on-write overwrite. Resolve the deletion mechanism for non-identity transforms. Keep the unpartitioned and empty-table guards.
- **Phase 3 — Happy-path & interop tests.** Multi-partition, null partition, copy-on-write vs merge-on-read, and a Spark round-trip for cross-engine parity.
- **Phase 4 (optional) — Docs & example/CLI exposure.**

## Parity references

- Java: `org.apache.iceberg.BaseReplacePartitions`
- PyIceberg: `Transaction.dynamic_partition_overwrite`

Credit to @dttung2905 for the original implementation in #482.

## Tracking

- [ ] Phase 1 — Partition-match predicate: #1216
- [ ] Phase 2 — `DynamicPartitionOverwrite` API
- [ ] Phase 3 — Happy-path & interop tests
[...]

PR #1217, opened 2026-06-16
feat(table): add partition-match predicate for dynamic overwrite

laskoviymishka:
Phase 1 of #1215: BuildPartitionMatchPredicate (in table/internal) builds the partition-matching filter the upcoming DynamicPartitionOverwrite will use to delete existing data in the partitions touched by incoming data before appending. Identity-only for now; non-identity transforms are rejected and deferred. Refs #1216.

### DIFF
--- a/table/internal/partition_predicate.go
+++ b/table/internal/partition_predicate.go
@@ -109,11 +109,6 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 				continue
 			}
 
-			lit, err := literalForPartitionValue(val)
-			if err != nil {
-				return nil, fmt.Errorf("partition field %q: %w", fr.name, err)
-			}
-
 			if isNaN(val) {
 				// x == NaN is never true (IEEE 754), so a NaN partition value must
 				// match via IsNaN. This intentionally diverges from PyIceberg's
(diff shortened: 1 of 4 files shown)


laskoviymishka:
@nssalian / @dttung2905 PTAL

table/internal/partition_predicate.go
```diff
@@ -0,0 +1,215 @@
...
+			// entry rather than an absent key; !ok is a defensive guard.
+			val, ok := part[fr.id]
+			if !ok || val == nil {
+				clause = iceberg.NewAnd(clause, iceberg.IsNull(ref))
+				sigParts = append(sigParts, strconv.Itoa(fr.id)+":null")
+
+				continue
+			}
+
+			lit, err := literalForPartitionValue(val)
```
zeroshade: Should you move this after the isNan then?

table/internal/partition_predicate.go
```diff
@@ -0,0 +1,215 @@
...
+	if lit, ok := v.(iceberg.Literal); ok {
+		return lit, nil
+	}
+
+	// The remaining cases cover the raw Go values manifest decoding emits
+	// (int32/int64/float/string/[]byte and the Date/Time/Timestamp aliases).
+	// The int and iceberg.Decimal cases are accepted defensively for
+	// hand-constructed inputs; the decoder itself yields int32/int64 and, for
+	// decimals, a DecimalLiteral caught by the passthrough above.
+	switch val := v.(type) {
```
zeroshade, line 183: I thought we already had a function in literals.go which takes an any and does this?

zeroshade (approved):
Just some nitpicks

PR #1218, opened 2026-06-17
fix(manifest): expose day-transform partition values as Date

### LINKED ISSUE #1200: Expose day-transform partition values consistently as Date
Follow up to #915, relates to #1199

Users can see different Go value types for the same logical `day(...)` partition value depending on how a manifest encoded it in Avro.

On current `main` (`1e4916f`), this is reproducible:

- `day(...)` encoded as Avro `int` with `logicalType: date` reads as `iceberg.Date(19000)`
- the same `day(...)` value encoded as legacy plain Avro `int` reads as `int32(19000)`

Both encodings represent days since `1970-01-01`, so iceberg-go should not leak that Avro encoding difference to users. When reading manifests, `day(...)` partition values should consistently be exposed as Iceberg `Date`.

This should only apply to fields whose partition spec transform is `day(...)`; other plain integer partition values should remain integers.

tanmayrauth:
A day(...) partition value read as iceberg.Date when encoded as an Avro int with the date logical type, but as a plain int32 when encoded as a legacy Avro int, leaking the Avro encoding to users. Use the manifest's partition spec to normalize day-transform fields to iceberg.Date for both encodings; other integer partition values are unaffected.

Closes #1200

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -429,6 +429,12 @@ func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int
 // encoding difference. Only day transforms are affected; other plain integer
 // partition values (including hour/month/year transforms) are left untouched.
 //
+// The Iceberg spec gives the day transform result type "date" and requires
+// readers to also accept the plain-int encoding, interpreting each integer as
+// the number of days since 1970-01-01 (see the transform table and note [1] in
+// apache/iceberg#16446). Accepting both encodings is therefore spec-mandated,
+// not merely cross-engine convention.
+//
 // Absent or malformed partition-spec metadata is ignored: the reader simply
 // falls back to the logical types declared in the Avro schema.
 func applyDayTransformDates(specJSON []byte, fieldIDToType map[int]string) {
@@ -1894,6 +1900,12 @@ func (d *dataFile) convertAvroValueToIcebergType(v any, fieldID int) any {
 				return Date(val)
 			}
 
+			// Unreachable with twmb/avro: an int+date logical type decodes
+			// to time.Time and a plain int to int32, so v is always one of
+			// the two cases above. Returning v rather than panicking keeps a
+			// future decoder change from crashing here, but callers that type
+			// assert iceberg.Date would then fail; do not add a guard that
+			// silently coerces other types, which reintroduces #1200.
 			return v
 		case atype.TimeMillis:
 			if val, ok := v.(time.Duration); ok {
--- a/manifest.go
+++ b/manifest.go
@@ -422,6 +422,32 @@ func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int
 	return result, logicalTypes, fixedSizes
 }
 
+// applyDayTransformDates marks every day(...) partition field described by the
+// manifest's partition-spec metadata as a "date" logical type. This normalizes
+// day-transform partition values to iceberg.Date even when they were written as
+// a plain Avro int with no logical type, so callers never observe the Avro
+// encoding difference. Only day transforms are affected; other plain integer
+// partition values (including hour/month/year transforms) are left untouched.
+//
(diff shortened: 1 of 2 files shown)


manifest.go
```diff
@@ -1858,8 +1890,11 @@ func (d *dataFile) convertAvroValueToIcebergType(v any, fieldID int) any {
 			if val, ok := v.(time.Time); ok {
 				return Date(val.Truncate(24*time.Hour).Unix() / int64((time.Hour * 24).Seconds()))
 			}
+			if val, ok := v.(int32); ok {
+				return Date(val)
+			}
 
-			return Date(v.(int32))
+			return v
```
laskoviymishka, line 1909: The `int32` guard is the right addition. My one concern is the `return v` it falls through to: this slot is for `atype.Date`, so a caller will do `partition[fieldID].(iceberg.Date)` and panic far from here if `v` is ever neither `time.Time` nor `int32`.

In practice this is unreachable today — `twmb/avro` decodes an int+date logical type to `time.Time` and a plain int to `int32`, so there's no third type to land here. But the old `Date(v.(int32))` panicked loudly at the conversion site on anything unexpected, and `return v` trades that for a silent wrong-typed value if a decoder change ever fires it.

I'd either keep the loud-failure behavior (a panic or a typed error) or add a one-line comment stating why this is unreachable, so nobody later "fixes" it by adding a guard that reintroduces #1200. wdyt?

manifest.go
```diff
@@ -422,6 +422,32 @@ func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int
 	return result, logicalTypes, fixedSizes
 }
 
+// applyDayTransformDates marks every day(...) partition field described by the
```
laskoviymishka, line 425: The godoc explains the why well, but it frames the two-encoding acceptance as de-facto cross-engine behavior. As of apache/iceberg#16446 it's no longer just convention — the spec's transform table now lists `day` with result type `date`, plus note [1]: "Readers must also accept `int` values for the `day` transform, interpreting each integer as a `date` represented by the number of days since 1970-01-01." That's exactly what this function implements.

I'd cite that note here (and in the test's rationale comment, which currently says "other Iceberg implementations and engines accept both encodings") so the normalization reads as spec-mandated rather than a compatibility nicety — it also tells the next reader this is required behavior, not something safe to drop.

laskoviymishka (changes requested):
This looks like the right fix. With apache/iceberg#16446, the spec now says the day transform result type is `date`, and note [1] requires readers to also accept the plain-int encoding as days since `1970-01-01`.

So this PR is basically the iceberg-go reader side of that clarification from https://github.com/apache/iceberg/pull/16446. It also lines up with Java, PyIceberg, and iceberg-rust. The two-encoding round-trip test is the right coverage.

One small ask: cite that spec note in the `applyDayTransformDates` godoc and the test rationale, so the day→date normalization reads as required by the spec, not just copied from other engines.

The one thing I’d still settle before merge is the `return v` fallthrough in `convertAvroValueToIcebergType`. It’s effectively dead with current `twmb/avro`, but as written it can silently return a raw value into a slot callers type-assert as `iceberg.Date`.

Once that’s fixed, happy to approve.

tanmayrauth:
Thanks for the thorough review! 

I've pushed a commit addressing all three points. The applyDayTransformDates godoc and the test rationale now cite apache/iceberg#16446 (the day transform's date result type and note [1] requiring readers to accept the plain-int encoding), so the normalization reads as  spec-mandated rather than convention. For the atype.Date fallthrough, I kept return v but added a comment explaining why it's unreachable with twmb/avro and warning against adding a coercing guard that would reintroduce #1200. Happy to switch that to a loud failure (panic/typed error) instead if you'd prefer,   let me know!

laskoviymishka (approved):
Thanks!

PR #915, opened 2026-04-17
fix(manifest): encode day-transform partition fields with Avro date logical type

artbcf:
`DayTransform.Apply()` returns an `int32` value representing days since the epoch. For manifest partition data, Iceberg implementations commonly write `day(...)` partition fields as Avro `int` with the `date` logical type, while readers should be able to accept both plain Avro `int` and Avro `date`.

Previously, `DayTransform.ResultType()` returned `Int32`, which caused `PartitionSpec.PartitionType()` to model day-transform partition fields as plain `Int32`. As a result, Iceberg Go wrote day-partition fields as plain Avro `int`, instead of Avro `date`, diverging from the writer behavior of other Iceberg implementations.

This PR changes `DayTransform.ResultType()` to return `DateType`. That lets the existing partition type and Avro schema conversion paths naturally emit Avro `date` for day-transform partition fields.

Tests added:

- Verify that `day(ts)` partition fields are encoded with the Avro `date` logical type.
- Verify that non-day `Int32` partition fields still use plain Avro `int` encoding.

PR #1219, opened 2026-06-17
test(manifest): read day-transform partition value as avro int or date

### LINKED ISSUE #1199: test day transform write as avro date, and can read partition value encoded as either avro int or avro date
LGTM 

lets merge this and add the tests as a follow up. 
- test day transform writes as avro date
- test can read partition value encoded as either avro int or avro date 

we also dont necessarily need `TestNonDayInt32PartitionAvroIntEncoding`, bucket transform encoding is likely tested elsewhere

_Originally posted by @kevinjqliu in https://github.com/apache/iceberg-go/pull/915#pullrequestreview-4472530481_

tanmayrauth:
Verify a day-transform partition value is readable whether it was written as a plain Avro int or an Avro int with the date logical type, matching what other Iceberg engines produce. Follow-up to #915.

Closes #1199

### DIFF
--- a/manifest_day_partition_test.go
+++ b/manifest_day_partition_test.go
@@ -1,0 +1,156 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package iceberg
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/apache/iceberg-go/internal"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/twmb/avro"
+	"github.com/twmb/avro/atype"
+	"github.com/twmb/avro/ocf"
+)
+
+// dayPartitionFieldID is the partition field id used by the day-transform tests.
+const dayPartitionFieldID = 1000
+
+// writeDayPartitionManifest encodes a single-entry v2 manifest whose only
+// partition field ("ts_day", produced by a day transform) is serialized using
+// partFieldNode. By varying partFieldNode between a plain Avro int and an Avro
+// int with the "date" logical type, the same day value can be written in either
+// of the two encodings that Iceberg engines produce in the wild.
+func writeDayPartitionManifest(t *testing.T, partFieldNode avro.SchemaNode, dayValue int32) ([]byte, ManifestFile) {
+	t.Helper()
+
+	spec := NewPartitionSpecID(0, PartitionField{
+		FieldID:   dayPartitionFieldID,
+		SourceIDs: []int{1},
+		Name:      "ts_day",
+		Transform: DayTransform{},
+	})
+
+	builder, err := NewDataFileBuilder(
+		spec,
+		EntryContentData,
+		"s3://bucket/namespace/table/data/00000-0-day.parquet",
+		ParquetFile,
+		map[int]any{dayPartitionFieldID: dayValue},
+		map[int]string{},
+		map[int]int{},
+		100,
+		1024,
+	)
+	require.NoError(t, err)
+
+	snapshotID := int64(42)
+	seqNum := int64(1)
+	entry := NewManifestEntry(EntryStatusADDED, &snapshotID, &seqNum, &seqNum, builder.Build())
+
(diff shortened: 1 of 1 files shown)


tanmayrauth:
@laskoviymishka, could you please take a look when you have a moment?

manifest_day_partition_test.go
```diff
@@ -0,0 +1,156 @@
...
+	require.NoError(t, err)
+
+	var buf bytes.Buffer
+	wr, err := ocf.NewWriter(&buf, entrySchema,
+		// WithSchema writes the explicit JSON schema, preserving the field-id
+		// props the manifest reader relies on to map partition values.
+		ocf.WithSchema(entrySchema.String()),
+		ocf.WithMetadata(map[string][]byte{
+			"format-version": []byte("2"),
+			"content":        []byte("data"),
```
laskoviymishka, line 90: I think this test doesn't exercise what it's documenting. The OCF metadata here only carries `format-version` and `content`, no `partition-spec`. `applyDayTransformDates` (manifest.go:711) returns immediately when `partition-spec` is absent, so for the plain-int case the day-transform field is never forced to `date` — the reader just hands back the raw `int32`. The assertion below passes because normalization was skipped, not because it worked; this test would still pass if `applyDayTransformDates` were deleted entirely.

That's also the scenario #1199 actually cares about: real manifests from older iceberg-go writers carry `partition-spec` but encode the field as plain int. To reach the path under test we need to write it here:

```go
ocf.WithMetadata(map[string][]byte{
    "format-version":    []byte("2"),
    "content":           []byte("data"),
    "partition-spec":    specFieldsJSON, // marshaled PartitionField slice
    "partition-spec-id": []byte("0"),
})
```

This is exactly the change that landed on `main` in #1218 (commit `41131be`).

manifest_day_partition_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	cases := []struct {
+		name     string
+		partNode avro.SchemaNode
+		want     any
+	}{
+		{
+			name:     "avro int (no logical type)",
+			partNode: internal.NullableNode(internal.IntNode),
+			want:     dayValue,
```
laskoviymishka, line 128: The `want` field asserts different Go types for the two encodings — `int32` for plain-int, `Date` for date-logical — but the normalization layer exists specifically to erase that difference (apache/iceberg#16446). Callers pattern-matching on `Partition()` values (predicate eval, projection, CDC) shouldn't have to handle both `int32` and `Date` for the same logical field, and a test that accepts both teaches the opposite.

It also slips past the assertion: `assert.IsType` is non-fatal, and `assert.EqualValues(t, dayValue, got)` coerces `Date(19000)` and `int32(19000)` to the same number, so a regression to raw `int32` goes unreported even on the case that's supposed to catch it.

I'd drop `want` and assert uniformly for both cases — `assert.IsType(t, Date(0), got)` plus `assert.Equal(t, Date(dayValue), got)` for typed equality. That's what's on `main` today.

manifest_day_partition_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return buf.Bytes(), file
+}
+
+// TestDayPartitionReadsAvroIntAndDate verifies that iceberg-go can read a
+// day-transform partition value regardless of whether it was written as a plain
+// Avro int or as an Avro int carrying the "date" logical type. Other Iceberg
+// implementations (Java, PyIceberg, iceberg-rust) and engines (Trino, Spark)
+// accept both encodings, and so must iceberg-go; a manifest normalized to the
+// date encoding by one writer must remain readable here. See apache/iceberg#16414.
```
laskoviymishka, line 111: Small cross-reference inconsistency: the docstring cites #16414 (the issue) while the production code at manifest.go:436 cites #16446 (the clarification PR that mandates the normalization). Worth pointing both at #16446 so the test and the code it guards line up. Minor — fold it into the rebase.

laskoviymishka (changes requested):
Thanks for picking up the #1199 follow-up, the docstring is genuinely good, it lays out the int-vs-date-logical encoding difference clearly and points at the cross-engine motivation.

I have to flag the bigger thing first though: this diff has been superseded. The corrected version of this exact test already landed on `main` as commit `41131be` (PR #1218), and GitHub already marks this PR as conflicting :D. The version on `main` writes `partition-spec` metadata and asserts `Date` uniformly for both encodings; the version in this diff does neither. If this merged as-is it would overwrite the good test with a weaker one, so I'd close this in favor of `main` (or hard-rebase and keep the #1218 shape).

For the record on why the diff's shape is the weaker one: it omits `partition-spec` from the OCF metadata, which makes `applyDayTransformDates` return early, so the plain-int normalization path — the one #1199 is actually about — never runs. The table then asserts `int32` for plain-int and `Date` for date-logical, and `assert.EqualValues` coerces both to the same number, so the test passes by bypassing normalization rather than exercising it. It'd still pass if `applyDayTransformDates` were deleted.

The fix is already written — it's what's on `main`: write `partition-spec` + `partition-spec-id`, drop the per-case `want`, and assert `Date(0)`/`Date(dayValue)` for both. Once this is rebased onto that, happy to take another pass.

tanmayrauth:
Thanks for the thorough review, @laskoviymishka  you're right on all counts. I rebased onto main and the commit emptied out entirely: the corrected test landed in #1218 (41131be) already writes partition-spec/partition-spec-id, drops the per-case want in favor of asserting Date uniformly.
  
  Closing this in favor of main as you suggested.

PR #1221, opened 2026-06-17
fix(dv): deletion vectors orphaned by data-file rewrite

twuebi:
Compaction replaces a data file carrying a deletion vector with new compacted files, dropping the old file's manifest entry, but leaves the DV's manifest entry pointing at a data file the table no longer references.

## Cause

DV removal was keyed by path, but a DV entry is 1:1 with its referenced data file and one Puffin can hold DVs for several files. Compaction never collected DVs (they ride in `FileScanTask.DeletionVectorFiles`, not `DeleteFiles`), and a path-keyed removal would have dropped sibling DVs in the same Puffin too.

## Fix

Key DV removal by referenced data file end to end. `CollectSafeDeletionVectors` gathers a group's DVs; `CompactionGroupResult.SafeDeletionVectors` carries them; both commit paths expunge them, dropping the rewritten file's DV while sibling DVs in the same Puffin survive. Use `RewriteFiles.ApplyResult`, not `Apply`, which can't carry them.

## Tests

`TestRewriteDataFilesRemovesDeletionVectors` (no orphan after compaction) and `TestRewriteDataFilesPreservesSiblingDeletionVector` (shared Puffin: rewritten DV expunged, sibling survives).

### DIFF
--- a/table/dv_rewrite_test.go
+++ b/table/dv_rewrite_test.go
@@ -35,7 +35,9 @@ import (
 
 // TestRewriteDataFilesRemovesDeletionVectors guarantees that compacting a data
 // file carrying a deletion vector expunges that DV in the same rewrite snapshot,
-// leaving no manifest entry referencing the removed data file.
+// leaving no manifest entry referencing the removed data file. Both commit
+// paths of [Transaction.RewriteDataFiles] are covered: the single atomic
+// snapshot and the per-group partial-progress path.
 //
 // Background: a DV manifest entry is 1:1 with its referenced data file but rides
 // in FileScanTask.DeletionVectorFiles, separate from the pos-delete files in
@@ -46,87 +48,64 @@ import (
 // referenced data file is gone — so the assertion walks the raw delete manifests
 // of the post-compaction snapshot.
 func TestRewriteDataFilesRemovesDeletionVectors(t *testing.T) {
-	ctx := context.Background()
-	mem := memory.DefaultAllocator
-	fs := iceio.LocalFS{}
-
-	tbl := newV3RowLineageTestTable(t)
-
-	arrowSchema := arrow.NewSchema([]arrow.Field{
-		{Name: "id", Type: arrow.PrimitiveTypes.Int64, Nullable: false},
-		{Name: "data", Type: arrow.BinaryTypes.String, Nullable: true},
-	}, nil)
-
-	// Two appends produce two data files so the planner forms a compaction group.
-	for _, rows := range []string{
-		`[{"id": 1, "data": "a"}, {"id": 2, "data": "b"}]`,
-		`[{"id": 3, "data": "c"}, {"id": 4, "data": "d"}]`,
-	} {
-		tab, err := array.TableFromJSON(mem, arrowSchema, []string{rows})
-		require.NoError(t, err)
-		tbl, err = tbl.Append(ctx, array.NewTableReader(tab, -1), nil)
-		tab.Release()
-		require.NoError(t, err)
-	}
-	assertRowCount(t, tbl, 4)
-
-	tasks, err := tbl.Scan().PlanFiles(ctx)
-	require.NoError(t, err)
-	require.Len(t, tasks, 2)
-	dvTarget := tasks[0].File.FilePath()
-
-	// Author a real DV (puffin) deleting position 0 (id=1) of the first data
-	// file, then commit it as a delete manifest entry.
-	specByID := func(id int32) *iceberg.PartitionSpec {
-		if id == 0 {
-			return iceberg.UnpartitionedSpec
+	for _, partialProgress := range []bool{false, true} {
+		name := "atomic"
+		if partialProgress {
+			name = "partial-progress"
 		}
+		t.Run(name, func(t *testing.T) {
+			ctx := context.Background()
+			tbl, dvTarget := seedV3TableWithDV(t)
 
-		return nil
+			groups, rewritten := planDVCompaction(t, tbl, dvTarget)
+
+			rtx := tbl.NewTransaction()
+			_, err := rtx.RewriteDataFiles(ctx, groups,
+				table.RewriteDataFilesOptions{PartialProgress: partialProgress})
+			require.NoError(t, err)
+			tbl, err = rtx.Commit(ctx)
+			require.NoError(t, err)
+
+			assertRowCount(t, tbl, 3) // DV applied during the rewrite; id=1 stays gone
+			assert.Empty(t, deleteEntriesReferencing(t, tbl, rewritten),
+				"compaction must remove deletion vectors for rewritten data files")
+		})
 	}
-	w := dv.NewDVWriter(fs, specByID)
-	w.Add(dvTarget, []int64{0}, 0, nil)
-	dvFiles, err := w.Flush(ctx, filepath.Join(filepath.Dir(dvTarget), "dv-0001.puffin"))
-	require.NoError(t, err)
-	require.Len(t, dvFiles, 1)
-	require.Equal(t, dvTarget, *dvFiles[0].ReferencedDataFile())
+}
+
+// TestRewriteFilesApplyResultRemovesDeletionVectors drives the distributed
+// coordinator path — a worker runs [ExecuteCompactionGroup], then a leader
+// stages the results with [Transaction.NewRewrite] + [RewriteFiles.ApplyResult]
+// + Commit. It locks the ApplyResult→ReplaceFiles coupling that carries
+// SafeDeletionVectors and expunges them by referenced data file.
+func TestRewriteFilesApplyResultRemovesDeletionVectors(t *testing.T) {
+	ctx := context.Background()
+	tbl, dvTarget := seedV3TableWithDV(t)
+
+	groups, rewritten := planDVCompaction(t, tbl, dvTarget)
+
+	var results []table.CompactionGroupResult
+	dvCount := 0
+	for _, g := range groups {
+		gr, err := table.ExecuteCompactionGroup(ctx, tbl, g)
+		require.NoError(t, err)
+		results = append(results, gr)
+		dvCount += len(gr.SafeDeletionVectors)
+	}
+	require.Positive(t, dvCount, "worker must collect the DV for removal")
 
 	tx := tbl.NewTransaction()
-	require.NoError(t, tx.NewRowDelta(nil).AddDeletes(dvFiles...).Commit(ctx))
-	tbl, err = tx.Commit(ctx)
-	require.NoError(t, err)
-	assertRowCount(t, tbl, 3) // id=1 deleted by the DV
-
-	tasks, err = tbl.Scan().PlanFiles(ctx)
-	require.NoError(t, err)
-
-	cfg := defaultTestCompactionCfg
-	cfg.DeleteFileThreshold = 1 // force compaction of the DV-bearing file
-	plan, err := cfg.PlanCompaction(tasks)
-	require.NoError(t, err)
-	require.NotEmpty(t, plan.Groups)
-
-	rewritten := map[string]struct{}{}
-	for _, g := range plan.Groups {
-		for _, tk := range g.Tasks {
-			rewritten[tk.File.FilePath()] = struct{}{}
-		}
+	rewrite := tx.NewRewrite(nil)
+	for _, gr := range results {
+		rewrite.ApplyResult(gr)
 	}
-	require.Contains(t, rewritten, dvTarget,
-		"the DV's referenced data file must be in the rewrite set for this test to exercise the bug")
-
-	rtx := tbl.NewTransaction()
-	_, err = rtx.RewriteDataFiles(ctx, toTaskGroups(plan.Groups), table.RewriteDataFilesOptions{})
-	require.NoError(t, err)
-	tbl, err = rtx.Commit(ctx)
+	require.NoError(t, rewrite.Commit(ctx))
+	tbl, err := tx.Commit(ctx)
 	require.NoError(t, err)
 
-	assertRowCount(t, tbl, 3) // DV applied during the rewrite; id=1 stays gone
-
-	orphans := deleteEntriesReferencing(t, tbl, rewritten)
-	assert.Empty(t, orphans,
-		"compaction must remove deletion vectors for rewritten data files; "+
-			"these DV entries still reference files removed by the rewrite: %v", orphans)
+	assertRowCount(t, tbl, 3)
+	assert.Empty(t, deleteEntriesReferencing(t, tbl, rewritten),
+		"coordinator commit must remove deletion vectors for rewritten data files")
 }
 
 // TestRewriteDataFilesPreservesSiblingDeletionVector pins the granularity of DV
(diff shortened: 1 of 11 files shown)


table/snapshot_producers.go
```diff
@@ -588,6 +590,34 @@ func (sp *snapshotProducer) removeDeleteFile(df iceberg.DataFile) *snapshotProdu
 	return sp
 }
 
+func (sp *snapshotProducer) removeDeletionVector(df iceberg.DataFile) *snapshotProducer {
+	sp.deletedDVsByRef[*df.ReferencedDataFile()] = df
```
laskoviymishka: This dereferences `ReferencedDataFile()` with nothing guarding nil — it works today only because every caller pre-filters through `isDeletionVectorFile`. That precondition isn't documented or enforced here, and the analogous `removeDeleteFile` doesn't deref a pointer at all, so there's no precedent for the unsafe contract.

I'd add a guard at the top so a future caller wired in without the pre-check fails closed rather than panicking:

```go
ref := df.ReferencedDataFile()
if ref == nil {
	return sp
}
sp.deletedDVsByRef[*ref] = df
```

Fix this and the snapshot side is good to land.

table/snapshot_producers.go
```diff
@@ -588,6 +590,34 @@ func (sp *snapshotProducer) removeDeleteFile(df iceberg.DataFile) *snapshotProdu
...
+	if _, ok := sp.deletedDeleteFiles[df.FilePath()]; ok {
+		return true
+	}
+	if isDeletionVectorFile(df) {
+		_, ok := sp.deletedDVsByRef[*df.ReferencedDataFile()]
+
+		return ok
+	}
+
+	return false
```
laskoviymishka, line 617: We now have two near-identical predicates in the `table` package: this one, and `isDeletionVector` in `scanner.go:182`, which checks only `ReferencedDataFile() != nil`. Same concept, different semantics, confusingly similar names.

The divergence isn't harmless — Java's `ContentFileUtil.isDV` keys purely on `format() == PUFFIN`, and the spec permits file-scoped Parquet pos-delete files that set `referenced_data_file`. The `scanner.go` version would misclassify one of those as a DV and route it to `DeletionVectorFiles` instead of `DeleteFiles`. This new one is closer to correct.

I'd consolidate to a single predicate keyed on the Puffin format check and use it in both places. wdyt?

table/rewrite_data_files.go
```diff
@@ -524,3 +537,25 @@ func CollectSafePositionDeletes(tasks []FileScanTask) []iceberg.DataFile {
 
 	return safe
 }
+
+// CollectSafeDeletionVectors returns the deletion vectors attached to the
+// tasks. A DV is bound 1:1 to the data file it references, and that file is
+// always in the rewrite set (it is the task's own file), so every DV here is
+// safe to expunge. Deduplicated by referenced data file.
+func CollectSafeDeletionVectors(tasks []FileScanTask) []iceberg.DataFile {
```
laskoviymishka, line 548: The "always in the rewrite set" claim is correct, but only because scan planning matches a DV to exactly its own task's file — and that invariant isn't stated. `CollectSafePositionDeletes` carries a full caller-contract warning right above it; this one just asserts the conclusion.

The gap bites a caller that hand-constructs tasks (a test helper, a non-standard coordinator): if a `FileScanTask` has a `DeletionVectorFiles[i]` whose `ReferencedDataFile()` isn't `task.File.FilePath()`, we'd collect a DV pointing at a live data file and mark it safe to expunge — silently resurrecting deleted rows in that other file.

I'd mirror the `CollectSafePositionDeletes` contract note here, spelling out that safety relies on the scan-planning invariant that each task owns the DVs it carries.

table/rewrite_files.go
```diff
@@ -189,7 +189,12 @@ func (r *RewriteFiles) Apply(deletes, adds, safeDeletes []iceberg.DataFile) *Rew
 //	}
 //	if err := rewrite.Commit(ctx); err != nil { ... }
 func (r *RewriteFiles) ApplyResult(gr CompactionGroupResult) *RewriteFiles {
-	return r.Apply(gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes)
+	r.Apply(gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes)
+	for _, dv := range gr.SafeDeletionVectors {
```
laskoviymishka: This works, but the correctness is invisible: a DV routed through `DeleteFile` lands in `deleteFilesToRemove` (because its `ContentType` is `EntryContentPosDeletes`), and `ReplaceFiles` then re-identifies it as a DV via `isDeletionVectorFile` to split it back out. That coupling breaks non-obviously if someone teaches `DeleteFile` to reject Puffin files or refactors the splitting in `ReplaceFiles`. A one-line comment here saying why `DeleteFile(dv)` is the right channel would save the next reader the trace.

The other half: `Apply` is still public and undeprecated, and a coordinator calling `rewriteFiles.Apply(...)` directly still orphans DVs exactly as before — its signature can't carry `SafeDeletionVectors`. The doc now says coordinators "must use ApplyResult" but nothing enforces it. I'd add `// Deprecated: use ApplyResult, which carries SafeDeletionVectors.` to `Apply` so the compiler nudges existing callers over.

table/rewrite_data_files.go
```diff
@@ -430,7 +441,8 @@ func (t *Transaction) rewriteDataFilesPartial(ctx context.Context, groups []Comp
 			continue
 		}
 
-		if err := t.ReplaceFiles(ctx, gr.OldDataFiles, gr.NewDataFiles, gr.SafePosDeletes,
+		deletesToRemove := append(slices.Clone(gr.SafePosDeletes), gr.SafeDeletionVectors...)
```
laskoviymishka, line 444: Both tests run with the default `PartialProgress: false`, so this branch — concatenating `SafeDeletionVectors` into the `ReplaceFiles` input — never runs in CI, and it's the path operators are most likely to hit on large jobs (a checkpoint per group).

I'd parameterize one of the existing DV tests over `PartialProgress: true`, or add a sibling case, asserting the same orphan-absence. While we're at it, a test that drives the coordinator pattern directly — `tx.NewRewrite(nil)` → `ApplyResult(gr)` for a group with DVs → `Commit` → assert no orphaned DV entry — would lock the `ApplyResult`/`ReplaceFiles` coupling from the comment on `rewrite_files.go`. Both paths the PR targets would then be covered.

table/transaction.go
```diff
@@ -932,6 +949,9 @@ func (t *Transaction) ReplaceFiles(ctx context.Context, dataFilesToDelete, dataF
 	if len(markedDeleteForRemoval) != len(setDeleteFilesToRemove) {
 		return errors.New("cannot remove delete files that do not belong to the table")
 	}
+	if len(markedDVsForRemoval) != len(dvRefsToRemove) {
```
laskoviymishka, line 957: This count check is fine for a well-formed table, but it leans on the same unstated assumption as the sibling data/pos-delete checks: that no `referenced_data_file` appears twice. If a broken table had two entries for the same ref, both would append to `markedDVsForRemoval` while the `dvRefsToRemove` map dedups, so `!=` wouldn't catch the over-count and one entry would survive. Not worth a guard, but a one-line comment noting the distinct-ref invariant this relies on would help.

laskoviymishka (changes requested):
This is the right model. Tracking DV removal by `referenced_data_file` instead of Puffin path is what shared-Puffin tables need, and `PreservesSiblingDeletionVector` pins the exact case a path-keyed impl would break.

One blocker before merge: `removeDeletionVector` dereferences `*df.ReferencedDataFile()` with no nil guard. It’s only safe today because current callers pre-filter through `isDeletionVectorFile`, but that contract isn’t enforced at the call site. A simple nil-ref guard at the top would close the panic risk.

I’d also prefer to cover the two target commit paths here rather than follow up: both current tests run `PartialProgress: false`, so `rewriteDataFilesPartial` and the distributed coordinator path (`NewRewrite` + `ApplyResult` + `Commit`) are still untested.

Smaller inline comments: consolidate the two DV predicates, add the same caller-contract note to `CollectSafeDeletionVectors`, and maybe deprecate public `Apply` toward `ApplyResult`.

Once the nil guard and path coverage are in, happy to approve.

twuebi:
@laskoviymishka addressed your points

twuebi:
got another fix lined up where _row_id/_last_updated_sequence_number were synthesized after delete/DV/filter were applied, that lead to survivors getting renumbered `0..N` instead of keeping first_row_id + their original file position. 

Its tests depend on some of the machinery introduced here, so waiting to merge this one before submitting the next PR.

table/scanner.go
```diff
@@ -179,8 +179,13 @@ func openManifest(io io.IO, manifest iceberg.ManifestFile,
...
 }
 
+// isDeletionVector reports whether df is a deletion vector, keyed on the Puffin
+// format to mirror Java's ContentFileUtil.isDV. Format is authoritative: a
+// spec-legal Parquet position-delete file may also set referenced_data_file, so
+// keying on that would misclassify it. Callers that key removal by
+// referenced_data_file guard nil at each site, since a malformed DV may omit it.
 func isDeletionVector(df iceberg.DataFile) bool {
-	return df.ReferencedDataFile() != nil
+	return df.FileFormat() == iceberg.PuffinFile
```
laskoviymishka: I'd add a content-type guard here: `return df.FileFormat() == iceberg.PuffinFile && df.ContentType() == iceberg.EntryContentPosDeletes`.

The Java reference (`ContentFileUtil.isDV`) only checks `format() == PUFFIN`, but it's called with a `DeleteFile`, which is already restricted to pos/eq-delete content. Our `isDeletionVector` takes an `iceberg.DataFile`, which can carry any content type — so a Puffin file with some other content type (a stats blob written as a delete-manifest entry by a buggy external writer) would pass this and get routed as a DV during a scan. Our own `DataFileBuilder` enforces `PuffinFile → EntryContentPosDeletes`, but external writers aren't bound by that, so the guard is what makes the classifier safe on the read side. wdyt?

table/transaction.go
```diff
@@ -881,10 +881,23 @@ func (t *Transaction) ReplaceFiles(ctx context.Context, dataFilesToDelete, dataF
 	}
 
 	setDeleteFilesToRemove := make(map[string]struct{}, len(deleteFilesToRemove))
+	dvRefsToRemove := make(map[string]struct{}, len(deleteFilesToRemove))
 	for i, df := range deleteFilesToRemove {
 		if df == nil {
 			return fmt.Errorf("nil delete file at index %d for ReplaceFiles", i)
 		}
+		if isDeletionVector(df) {
```
laskoviymishka, line 893: The DV branch `continue`s before the `path == ""` guard runs, so a DV with a non-nil ref but an empty `FilePath()` slips past all validation and lands in the snapshot as a delete-manifest entry with an empty path string — malformed metadata that confuses future readers.

The in-process flow never hits this since `CollectSafeDeletionVectors` returns live scan-planning objects with real paths, but `ReplaceFiles` is public and a serialization round-trip through `ApplyResult` could drop the path field. I'd add the same non-empty check inside this block before the `continue`:

```go
if df.FilePath() == "" {
    return errors.New("deletion vector paths must be non-empty for ReplaceFiles")
}
```

table/transaction.go
```diff
@@ -932,6 +952,12 @@ func (t *Transaction) ReplaceFiles(ctx context.Context, dataFilesToDelete, dataF
 	if len(markedDeleteForRemoval) != len(setDeleteFilesToRemove) {
 		return errors.New("cannot remove delete files that do not belong to the table")
 	}
+	// Relies on the spec invariant of at most one DV per referenced data file:
```
laskoviymishka: I think the comment has the failure mode backwards, and the check is unsound in that case. `markedDVsForRemoval` is built by scanning `s.dataFiles`, while `dvRefsToRemove` is a deduped map. If the table has the same DV ref in two distinct manifest entries (a failed OCC retry that rewrote one manifest but left the old one un-GC'd, or a corrupt table), the scan yields two entries for one ref, so `len(markedDVsForRemoval)=2 > len(dvRefsToRemove)=1` — the check *fails* and rejects a valid request. Duplicates don't "slip past" the check; they trip it.

I'd dedupe `markedDVsForRemoval` by ref before comparing (or count distinct refs found), and reword the comment to match. Otherwise compaction can spuriously error on tables with duplicate DV manifest entries.

table/snapshot_producers.go
```diff
@@ -588,6 +590,33 @@ func (sp *snapshotProducer) removeDeleteFile(df iceberg.DataFile) *snapshotProdu
 	return sp
 }
 
+func (sp *snapshotProducer) removeDeletionVector(df iceberg.DataFile) *snapshotProducer {
+	ref := df.ReferencedDataFile()
+	if ref == nil {
```
laskoviymishka, line 595: `removeDeletionVector` silently drops the request when `ref == nil` and returns `sp` unchanged. `ReplaceFiles` guards this upstream so normal flow never reaches here with a nil ref — but any future direct caller would silently lose the removal and orphan the exact DV this PR is fixing, with no error signal.

I'd either return an error here or add a doc line stating callers must guarantee a non-nil ref.

table/rewrite_data_files.go
```diff
@@ -524,3 +537,31 @@ func CollectSafePositionDeletes(tasks []FileScanTask) []iceberg.DataFile {
...
+// tasks, deduplicated by referenced data file.
+//
+// Safety rests on the scan-planning invariant that a task carries only the DV
+// referencing its own data file (matchDVToData keys on task.File's path), so
+// every DV returned references a file in the rewrite set. A caller that
+// hand-builds a [FileScanTask] with a DV whose ReferencedDataFile is some other
+// live data file would have that DV marked safe to expunge here, silently
+// resurrecting its deleted rows — populate DeletionVectorFiles only from scan
+// planning.
+func CollectSafeDeletionVectors(tasks []FileScanTask) []iceberg.DataFile {
```
laskoviymishka, line 548: Minor, but `map[string]bool` as a sentinel set is a lint target — `map[string]struct{}` is the idiomatic form. `CollectSafePositionDeletes` right above has the same pattern; while we're here I'd switch both. Not blocking.

table/dv_rewrite_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	var results []table.CompactionGroupResult
+	dvCount := 0
+	for _, g := range groups {
+		gr, err := table.ExecuteCompactionGroup(ctx, tbl, g)
+		require.NoError(t, err)
+		results = append(results, gr)
+		dvCount += len(gr.SafeDeletionVectors)
+	}
+	require.Positive(t, dvCount, "worker must collect the DV for removal")
```
laskoviymishka, line 95: The tests cover the happy paths well, but none of the error paths `ReplaceFiles` now introduces are exercised — the missing-ref error, the distinct-refs error, and the "DV not in table" count check. Given those branches are where the soundness questions live, I'd add at least one test that drives a DV-not-in-table removal and one for the duplicate-ref case once that check is settled. wdyt?

laskoviymishka (changes requested):
Nice work tracking this down, the diagnosis is right. DVs ride in `FileScanTask.DeletionVectorFiles` separate from the pos-delete files, so a rewrite that collected only the latter orphaned them, and now both commit paths expunge by referenced data file. The `deleteFileRemoved` helper is the right abstraction, and the sibling-DV test (two blobs in one shared Puffin, only one data file rewritten) pins exactly the case I'd worry about — keying removal by path instead of ref would silently resurrect the sibling's deleted rows.

I'd hold this before merging though. The biggest one is `isDeletionVector`: it now keys purely on Puffin format with no content-type guard. Java's `ContentFileUtil.isDV` gets away with that because it's handed a `DeleteFile` that's already restricted to delete content, but our version takes an `iceberg.DataFile` that can carry anything, so a non-pos-delete Puffin file from an external writer would be misclassified as a DV during a scan. I'd add `&& df.ContentType() == iceberg.EntryContentPosDeletes`.

The other concern is the new validation in `ReplaceFiles`. The DV branch `continue`s before the empty-path guard, so a zero-path DV slips through into the snapshot as malformed metadata; and the count-equality check can spuriously fail on a table with duplicate DV manifest entries — the comment defending it actually has the logic backwards (duplicates trip the check, they don't slip past it). Both are public-API surfaces, so I'd want them tight even though the in-process flow doesn't hit them today.

submit major things — the ones I'd want settled in this PR before merge:

- content-type guard in `isDeletionVector`
- empty-path guard for the DV branch in `ReplaceFiles`
- dedupe `markedDVsForRemoval` by ref (and fix the comment) so the count check can't spuriously reject valid removals
- a test or two for the new error paths

The `map[string]bool` set and the nil-ref `removeDeletionVector` behavior are minor — happy to take those as nits. Once the above are addressed, happy to take another pass and approve.

laskoviymishka (approved):
<3

PR #1227, opened 2026-06-18
fix(transforms): preserve "+00:00" UTC offset for TimestampTz / TimestampTzNs in IdentityTransform.ToHumanStr

### LINKED ISSUE #1226: transforms: IdentityTransform.ToHumanStr omits UTC offset for TimestampTz / TimestampTzNs values
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

In Iceberg-Go, `iceberg.IdentityTransform.ToHumanStr` formats `TimestampTzType` and `TimestampTzNsType` partition values without the `+00:00` UTC offset.

This creates a different partition directory than what is expected.

expected: `created_ts_tz=2024-01-15T10%3A30%3A00%2B00%3A00`
created: `created_ts_tz=2024-01-15T10%3A30%3A00`

### Proposed Solution

Implement the function to dispatch on Type and appends `+00:00` for `TimestampTzType` / `TimestampTzNsType`. 

Something similar to what Iceberg Java's [**DateTimeUtil**](https://github.com/apache/iceberg/blob/8924046f7ac5326e454d1cd48a6b30a6d16ba196/api/src/main/java/org/apache/iceberg/util/DateTimeUtil.java)  implements.

badalprasadsingh:
## Problem

In Iceberg-Go, `iceberg.IdentityTransform.ToHumanStr` formats both timezone-bearing (`TimestampTzType`, `TimestampTzNsType`) and zone-naive (`TimestampType`, `TimestampNsType`) timestamps with the same **offset-less** ISO string.

Fixes: #1226 

## Solution

Implement the function to dispatch on Type and appends `+00:00` for `TimestampTzType` / `TimestampTzNsType`. Something similar to what Iceberg Java's [**DateTimeUtil**](https://github.com/apache/iceberg/blob/8924046f7ac5326e454d1cd48a6b30a6d16ba196/api/src/main/java/org/apache/iceberg/util/DateTimeUtil.java)  implements.

## Files changed

| File | Description |
|---|---|
| `partitions.go` | Update PartitionToPath to use a type-aware human-string |
| `transforms.go` | Add type-aware timestamp formatting for IdentityTransform |
| `transforms_test.go` | Regression Test |

## Testing
- Ingested data into a partitioned iceberg table in Glue Catalog

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -552,7 +552,12 @@ func (ps *PartitionSpec) PartitionToPath(data StructLike, sc *Schema) string {
 		sb.WriteByte('=')
 
 		// Only escape the value (which changes per row)
-		valueStr := ps.fields[i].Transform.ToHumanStr(data.Get(i))
+		var valueStr string
+		if t, ok := ps.fields[i].Transform.(typedHumanStringer); ok {
+			valueStr = t.ToHumanStrType(partType.FieldList[i].Type, data.Get(i))
+		} else {
+			valueStr = ps.fields[i].Transform.ToHumanStr(data.Get(i))
+		}
 		sb.WriteString(url.QueryEscape(valueStr))
 	}
 
--- a/transforms.go
+++ b/transforms.go
@@ -103,6 +103,10 @@ type Transform interface {
 	ToHumanStr(any) string
 }
 
+type typedHumanStringer interface {
+	ToHumanStrType(typ Type, val any) string
+}
+
 // IdentityTransform uses the identity function, performing no transformation
 // but instead partitioning on the value itself.
 type IdentityTransform struct{}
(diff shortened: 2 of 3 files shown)


badalprasadsingh:
@laskoviymishka, please take a look when you are free.

tanmayrauth (commented):
Thanks for the PR, it looks good.

  One thing I want to confirm: `ToHumanStr(any)` still emits the offset-less string for timestamptz, since both timestamp and timestamptz reach it as the same Go `Timestamp` value, only the new `ToHumanStrType` carries the offset. That's the right fix for the partition-path bug, but `ToHumanStr` is the documented `Transform` method, so anyone calling it directly on a timestamptz won't get the offset. Is the partition-path-only scope intentional? A short comment on `ToHumanStrType` explaining why the offset lives only there would save a future reader some confusion.
  
  Minor: the new `case TimestampNano` in `ToHumanStr` (transforms.go:158) isn't reachable from the partition path anymore — since IdentityTransform now implements typedHumanStringer, PartitionToPath always routes through ToHumanStrTyp, so it's currently untested. Fine to keep as completeness, just noting it.

badalprasadsingh:
Hey @tanmayrauth, first of all, thanks for review.

1. Yes, its intentional. Though the principle is _type-aware-render_, not partition-path. Currently `PartitionToPath` is the only such caller in iceberg-go.<br>
I understand any caller calling `ToHumanStr` won't get the offset, but that is **by design**: at that boundary the source `Type` has been erased and `Timestamp` / `TimestampTz` values are indistinguishable (as, both arrive as `iceberg.Timestamp` int64 micros), so the offset cannot be recovered there. <br>
In Iceberg-Java, we follow the same approach, check [here](https://github.com/apache/iceberg/blob/54ea9a04d287f21bca16c5a1d40bb20e332bd864/api/src/main/java/org/apache/iceberg/transforms/Transform.java#L148-L202). <br>
With, the value-only `toHumanString(T value)` already [depricated](https://github.com/apache/iceberg/blob/54ea9a04d287f21bca16c5a1d40bb20e332bd864/api/src/main/java/org/apache/iceberg/transforms/Transform.java#L148-L166) in Iceberg-Java, and staled for removal in `2.0.0`. Happy to do an iceberg-go counterpart (deprecating `Transform.ToHumanStr(any)` in favor of the type-aware form) in a follow-up PR if you'd like - but that's a public Transform interface **change** so probably deserves its own discussion.<br>
Added comments on `ToHumanStrType` & `typedHumanStringer`.
<br></br>
2. Yes, we should keep it because `ToHumanStr(any)` is still part of the public Transform interface for callers without source-type context. An external caller doing `IdentityTransform.ToHumanStr(<someTimestampNano>)` is still served by this arm. <br>
Pinned it with a row in the existing `TestToHumanString`.

zeroshade (approved):
Your explanation makes sense, this looks good to me. 

> With, the value-only toHumanString(T value) already depricated in Iceberg-Java, and staled for removal in 2.0.0. Happy to do an iceberg-go counterpart (deprecating Transform.ToHumanStr(any) in favor of the type-aware form) in a follow-up PR if you'd like - but that's a public Transform interface change so probably deserves its own discussion.

Please do! That would be very helpful as a follow up PR.

badalprasadsingh:
Thanks @zeroshade! I'll open a follow-up PR to discuss the depricating `Transform.ToHumanStr(any)` in favor of the type-aware API

PR #1229, opened 2026-06-18
fix(table): handle string file_path layouts in position deletes

fallintoplace:
Summary

- collect position-delete `file_path` values from string, large-string, and dictionary Arrow chunks
- return invalid-schema errors for unsupported or null `file_path` layouts instead of panicking
- pass the caller compute context through the `file_path` unique path
- release per-path filter masks immediately after filtering
- keep the existing Arrow filter path for position values

Testing

- `go test ./table -run 'TestGroupPosDeletesByFilePath|TestProcessPositionalDeletesAcrossBatches|TestRowDeltaIntegrationPosDeleteRoundTrip' -count=1`
- `git diff --check`
- `go test ./table -count=1`
- `go test ./... -count=1`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -30,6 +30,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/compute"
 	"github.com/apache/arrow-go/v18/arrow/compute/exprs"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/arrow/scalar"
 	"github.com/apache/iceberg-go"
 	iceinternal "github.com/apache/iceberg-go/internal"
 	iceio "github.com/apache/iceberg-go/io"
@@ -251,6 +252,153 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
 	return *a.ContentOffset() == *b.ContentOffset()
 }
 
+func filePathValueType(dt arrow.DataType) arrow.DataType {
+	if dictType, ok := dt.(*arrow.DictionaryType); ok {
+		return dictType.ValueType
+	}
+
+	return dt
+}
+
+func filePathValues(values arrow.Array) (arrow.TypedArray[string], error) {
+	switch arr := values.(type) {
+	case *array.Dictionary:
+		wrapped, err := array.NewDictWrapper[string](arr)
+		if err != nil {
+			return nil, fmt.Errorf("%w: file_path column is not string: %w",
+				iceberg.ErrInvalidSchema, err)
+		}
+
+		return wrapped, nil
+	case arrow.TypedArray[string]:
+		return arr, nil
+	default:
+		return nil, fmt.Errorf("%w: unsupported file_path column type %s in position delete file",
+			iceberg.ErrInvalidSchema, values.DataType())
+	}
+}
+
+func validateFilePathValues(values arrow.TypedArray[string], arr arrow.Array) error {
+	var dict arrow.Array
+	var indices *array.Dictionary
+	if dictionary, ok := arr.(*array.Dictionary); ok {
+		indices = dictionary
+		dict = dictionary.Dictionary()
+	}
+
+	for i := 0; i < values.Len(); i++ {
+		if values.IsNull(i) {
+			return fmt.Errorf("%w: null file_path in position delete file",
+				iceberg.ErrInvalidSchema)
+		}
+		if dict != nil && dict.IsNull(indices.GetValueIndex(i)) {
+			return fmt.Errorf("%w: null file_path dictionary value in position delete file",
+				iceberg.ErrInvalidSchema)
+		}
+	}
+
+	return nil
+}
+
+func distinctPosDeleteFilePaths(ctx context.Context, filePathCol *arrow.Chunked) (arrow.Array, error) {
+	if filePathCol.NullN() > 0 {
+		return nil, fmt.Errorf("%w: null file_path in position delete file", iceberg.ErrInvalidSchema)
+	}
+	if filePathValueType(filePathCol.DataType()).ID() == arrow.STRING_VIEW {
+		return nil, fmt.Errorf("%w: unsupported file_path column type %s in position delete file",
+			iceberg.ErrInvalidSchema, filePathCol.DataType())
+	}
+
+	filePaths := compute.NewDatum(filePathCol)
+	unique, err := compute.Unique(ctx, filePaths)
+	filePaths.Release()
+	if err != nil {
+		return nil, err
+	}
+
+	result, ok := unique.(*compute.ArrayDatum)
+	if !ok {
+		unique.Release()
+
+		return nil, fmt.Errorf("%w: unique file_path result is %s",
+			iceberg.ErrInvalidSchema, unique.Kind())
+	}
+
+	arr := result.MakeArray()
+	unique.Release()
+
+	return arr, nil
+}
+
+func groupPosDeletesByFilePath(ctx context.Context, filePathCol, posCol *arrow.Chunked) (map[string]*arrow.Chunked, error) {
+	uniquePaths, err := distinctPosDeleteFilePaths(ctx, filePathCol)
+	if err != nil {
+		return nil, err
+	}
+	defer uniquePaths.Release()
+
+	paths, err := filePathValues(uniquePaths)
+	if err != nil {
+		return nil, err
+	}
+	if err := validateFilePathValues(paths, uniquePaths); err != nil {
+		return nil, err
+	}
+
+	results := make(map[string]*arrow.Chunked, uniquePaths.Len())
+	for i := 0; i < uniquePaths.Len(); i++ {
+		sc, err := scalar.GetScalar(uniquePaths, i)
+		if err != nil {
+			releasePosDeletes(results)
+
+			return nil, err
+		}
+		scDatum := compute.NewDatum(sc)
+		if releasable, ok := sc.(scalar.Releasable); ok {
+			releasable.Release()
+		}
+		mask, err := compute.CallFunction(ctx, "equal", nil,
+			compute.NewDatumWithoutOwning(filePathCol), scDatum)
+		scDatum.Release()
+		if err != nil {
+			releasePosDeletes(results)
+
+			return nil, err
+		}
+
+		filtered, err := compute.Filter(ctx, compute.NewDatumWithoutOwning(posCol),
+			mask, *compute.DefaultFilterOptions())
+		mask.Release()
+		if err != nil {
+			releasePosDeletes(results)
+
+			return nil, err
+		}
+
+		filteredChunked, ok := filtered.(*compute.ChunkedDatum)
+		if !ok {
+			filtered.Release()
+			releasePosDeletes(results)
+
+			return nil, fmt.Errorf("%w: filtered position delete result is %s",
+				iceberg.ErrInvalidSchema, filtered.Kind())
+		}
+		filteredChunked.Value.Retain()
+		results[paths.Value(i)] = filteredChunked.Value
+		filtered.Release()
+	}
+
+	return results, nil
+}
+
+func releasePosDeletes(deletes map[string]*arrow.Chunked) {
+	for _, chunk := range deletes {
+		if chunk != nil {
+			chunk.Release()
+		}
+	}
+}
+
 func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_ map[string]*arrow.Chunked, err error) {
 	src, err := internal.GetFile(ctx, fs, dataFile, true)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tanmayrauth (approved):
LGTM, please check the lint issue.

table/arrow_scanner.go
```diff
@@ -251,6 +252,158 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
+	}
+
+	return scalar.NewStringScalar(v)
+}
+
+type stringArrayValues interface {
+	Len() int
+	IsNull(int) bool
+	Value(int) string
+}
```
zeroshade: Just use `array.StringLike` (https://github.com/apache/arrow-go/blob/main/arrow/array/string.go#L32)

table/arrow_scanner.go
```diff
@@ -251,6 +252,158 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
+		if values.IsNull(i) {
+			return fmt.Errorf("%w: null file_path in position delete chunk %d row %d",
+				iceberg.ErrInvalidSchema, chunkIdx, i)
+		}
+
+		addDistinctFilePath(seen, paths, values.Value(i))
+	}
+
+	return nil
+}
```
zeroshade: Could you just use the [`compute.UniqueArray`](https://pkg.go.dev/github.com/apache/arrow-go/v18/arrow/compute#UniqueArray) function instead?

table/arrow_scanner.go
```diff
@@ -251,6 +252,158 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
+		}
+
+		dictIdx := arr.GetValueIndex(i)
+		if values.IsNull(dictIdx) {
+			return fmt.Errorf("%w: null file_path dictionary value in position delete chunk %d row %d",
+				iceberg.ErrInvalidSchema, chunkIdx, i)
+		}
+
+		addDistinctFilePath(seen, paths, values.Value(dictIdx))
+	}
```
zeroshade: the same UniqueArray function works on dictionaries too! 😄

table/arrow_scanner.go
```diff
@@ -251,6 +252,158 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
+
+	return nil
+}
+
+func distinctPosDeleteFilePaths(filePathCol *arrow.Chunked) ([]string, error) {
```
zeroshade: yea, this entire function can be replaced with just calling [`compute.Unique(compute.NewDatumWithoutOwning(filePathCol))`](https://pkg.go.dev/github.com/apache/arrow-go/v18/arrow/compute#Unique) and using the resulting array which would contain the unique paths

table/arrow_scanner.go
```diff
@@ -277,29 +430,8 @@ func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_
 
 	filePathCol := tbl.Column(tbl.Schema().FieldIndices("file_path")[0]).Data()
 	posCol := tbl.Column(tbl.Schema().FieldIndices("pos")[0]).Data()
-	dict := filePathCol.Chunk(0).(*array.Dictionary).Dictionary().(*array.String)
```
zeroshade, line 280: I think that `compute.Unique` will always output a dictionary, which means that you could simply just do something like this:

```go
result, err := compute.Unique(compute.NewDatumWithoutOwning(filePathCol))
if err != nil {
    return nil, err
}
defer result.Release()

dictChunks := result.(compute.ArrayLikeDatum).Chunks()
for _, c := range dictChunks {
    arr, err := array.NewDictWrapper[string](c.(*array.Dictionary))
    if err != nil {
        return nil, fmt.Errorf("file path col not string: %w", err)
    }
    for i := range arr.Len() {
        v := arr.Value(i)
        ....
    }
}
```

that would probably simplify this a lot while still achieving your goal.

zeroshade (changes requested):

table/arrow_scanner.go
```diff
@@ -251,48 +252,127 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
-	tbl, err = array.UnifyTableDicts(compute.GetAllocator(ctx), tbl)
+	return paths, nil
+}
+
+func distinctPosDeleteFilePaths(ctx context.Context, filePathCol *arrow.Chunked) ([]string, error) {
+	if filePathCol.NullN() > 0 {
+		return nil, fmt.Errorf("%w: null file_path in position delete file", iceberg.ErrInvalidSchema)
+	}
+
+	unique, err := compute.Unique(compute.WithAllocator(ctx, memory.DefaultAllocator),
```
laskoviymishka: I'd pass `ctx` straight through here rather than forcing `memory.DefaultAllocator`.

There's no leak today — `unique` is released via `defer unique.Release()` and the array via `releaseChunks` — but hardcoding the allocator means `Unique`'s allocations bypass whatever the caller wired in: a `CheckedAllocator` in tests, or a bounded/tracked allocator in production. It's defensive code that quietly defeats the thing it looks like it's protecting.

This pairs with the two regression tests below — once they use a checked allocator, this override has to go or the `AssertSize` can't see `Unique`'s allocations. wdyt?

table/arrow_scanner.go
```diff
@@ -251,48 +252,127 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
 		mask, err := compute.CallFunction(ctx, "equal", nil,
-			compute.NewDatumWithoutOwning(filePathCol), compute.NewDatum(v))
+			compute.NewDatumWithoutOwning(filePathCol), scDatum)
+		scDatum.Release()
 		if err != nil {
+			releasePosDeletes(results)
+
 			return nil, err
 		}
 		defer mask.Release()
```
laskoviymishka: `defer` here fires at function return, not per loop iteration, so we hold one full boolean mask (one bit per row of `filePathCol`) live for every distinct path at once. For a delete file referencing thousands of data files that's a real spike.

I'd release it right after `Filter` instead, and on the error paths:

```go
filtered, err := compute.Filter(ctx, compute.NewDatumWithoutOwning(posCol), mask, *compute.DefaultFilterOptions())
mask.Release()
if err != nil {
    releasePosDeletes(results)
    return nil, err
}
```

Pre-existing, but the refactor is the natural moment to fix it.

table/arrow_scanner.go
```diff
@@ -251,48 +252,127 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
+	defer unique.Release()
+
+	var chunks []arrow.Array
+	var releaseChunks func()
+	switch result := unique.(type) {
+	case *compute.ArrayDatum:
+		arr := result.MakeArray()
+		chunks = []arrow.Array{arr}
+		releaseChunks = arr.Release
+	case *compute.ChunkedDatum:
```
laskoviymishka: I think this branch is unreachable — `compute.Unique` registers with `OutputChunked = false`, so `WrapResults` hands back the single datum un-wrapped and the result is always `*compute.ArrayDatum`.

The asymmetry bugs me a little too: the array branch sets `releaseChunks = arr.Release` while this one is a no-op. I'd collapse to an `*compute.ArrayDatum` assert plus the `default` error case and drop this. wdyt?

table/arrow_scanner.go
```diff
@@ -251,48 +252,127 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
+	case array.StringLike:
+		for i := 0; i < arr.Len(); i++ {
+			if arr.IsNull(i) {
+				return nil, fmt.Errorf("%w: null file_path in position delete file",
+					iceberg.ErrInvalidSchema)
+			}
+
+			paths = append(paths, arr.Value(i))
+		}
+	case *array.Dictionary:
```
laskoviymishka, line 265: The `NullN()>0` guard, the per-row null checks, and this `NewDictWrapper[string]` error branch all carry `ErrInvalidSchema` paths that nothing exercises — they'd still pass if we deleted them entirely.

I'd add a case feeding a `file_path` chunk with a null and asserting `ErrInvalidSchema`, and optionally a dictionary with a non-string value type to hit the `NewDictWrapper` failure. The reject test already gives you the pattern to extend.

table/arrow_scanner_posdelete_regression_test.go
```diff
@@ -21,12 +21,139 @@ import (
...
+					chunkB.Release()
+					dict.Release()
+					idxA.Release()
+					idxB.Release()
+				}
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			mem := memory.DefaultAllocator
```
laskoviymishka, line 118: This is a "does it crash" test rather than a "does it leak" test — with `memory.DefaultAllocator` and no `AssertSize`, a leak anywhere in `groupPosDeletesByFilePath` would pass silently.

`TestProcessPositionalDeletesAcrossBatches` just below in this file already does it the way I'd want: `memory.NewCheckedAllocator(memory.DefaultAllocator)` with `defer mem.AssertSize(t, 0)`. I'd mirror that here (needs the allocator-override fix above so the checked allocator actually sees `Unique`'s allocations).

table/arrow_scanner_posdelete_regression_test.go
```diff
@@ -21,12 +21,139 @@ import (
...
+			posCol.Release()
+			posA.Release()
+			posB.Release()
+			releaseFilePathCol()
+		})
+	}
+}
+
+func TestGroupPosDeletesByFilePathRejectsUnsupportedFilePathLayout(t *testing.T) {
+	mem := memory.DefaultAllocator
```
laskoviymishka: Same here — checked allocator + `defer mem.AssertSize(t, 0)`. The error path allocates and rejects, so it's exactly where you'd want to confirm nothing's left dangling.

table/row_delta_test.go
```diff
@@ -490,10 +508,13 @@ func TestRowDeltaIntegrationPosDeleteRoundTrip(t *testing.T) {
...
 	posDelArrowSc := table.PositionalDeleteArrowSchema
 	posDelPath := location + "/data/pos-del-001.parquet"
-	writeParquetFile(t, posDelPath, posDelArrowSc, `[
+	writeParquetFileWithProperties(t, posDelPath, posDelArrowSc, `[
 		{"file_path": "`+dataPath+`", "pos": 1},
 		{"file_path": "`+dataPath+`", "pos": 3}
-	]`)
+	]`, 1, parquet.NewWriterProperties(
+		parquet.WithStats(true),
+		parquet.WithDictionaryDefault(false),
```
laskoviymishka, line 517: Nice touch forcing plain-string encoding with `WithDictionaryDefault(false)` — that's exactly the layout the old code panicked on.

One gap: the assertion downstream is `assertRowCount(t, tbl, 3)`, so if the filter deleted the wrong two rows the test would still pass. Since the survivors should be "alpha", "gamma", "epsilon", I'd assert the actual values, not just the count — that's what proves we dropped positions 1 and 3 and not some others.

table/row_delta_test.go
```diff
@@ -434,9 +447,14 @@ func writeParquetFile(t testing.TB, path string, sc *arrow.Schema, jsonData stri
...
-		parquet.NewWriterProperties(parquet.WithStats(true)),
-		pqarrow.DefaultWriterProps()))
+	if rowGroupSize <= 0 {
+		rowGroupSize = rec.NumRows()
+	}
+	if writerProps == nil {
+		writerProps = parquet.NewWriterProperties(parquet.WithStats(true))
+	}
+
+	require.NoError(t, pqarrow.WriteTable(tbl, fw, rowGroupSize, writerProps, pqarrow.DefaultWriterProps()))
```
laskoviymishka, line 458: `fw` from `fs.Create(path)` just above is never closed — copied from `writeParquetFile`, so pre-existing, but worth a `defer fw.Close()` while we're in here. Tempdir so it's low-stakes, but it leaks FDs and has bitten Windows CI before.

laskoviymishka (changes requested):
The fix is right — the old `readDeletes` assumed `file_path` was dictionary-encoded and panicked on the plain-string / large-string layouts that Trino and dict-disabled writers produce. Routing through `compute.Unique` plus a type switch is a clean shape, and the `ErrInvalidSchema` rejections line up with the spec.

I'd hold this before merging though. The main thread: `distinctPosDeleteFilePaths` hardcodes `memory.DefaultAllocator` into `compute.Unique`, which defeats any allocator a caller injects — and the two new tests run on the default allocator with no `AssertSize`, so they prove "doesn't crash" but not "doesn't leak." Drop the override and the tests can assert zero leaks the way `TestProcessPositionalDeletesAcrossBatches` already does. I'd also release `mask` right after `Filter` rather than deferring to function return. Details inline — once those are settled, happy to take another pass and approve.

table/arrow_scanner.go
fallintoplace: Thanks, pushed `37c45a8`.

One note: I left the happy-path layout matrix on the default allocator. When I ran that through a checked allocator, Arrow's string/dictionary `Unique` path still showed live allocations after the returned objects were released. The reject cases use a checked allocator now, so the cleanup/error paths are still covered.

laskoviymishka (approved):
LGTM now!

Will wait @zeroshade to merge.

laskoviymishka (approved):
LGTM now!

Will wait @zeroshade to merge.

zeroshade (approved):
**Non-blocking nits:**
1. Success-path tests use `DefaultAllocator`, and compute uses the ctx allocator (also Default), so the happy-path Retain/Release balance is not leak-checked. Wrapping the success case in `compute.WithAllocator(ctx, checked)` + `AssertSize(0)` would harden it.
2. `filePathScalar` only special-cases `LARGE_STRING`; a `StringView` column (accepted by the `StringLike` branch) would get a plain `StringScalar` for the `equal` comparison. Practically never hit for `file_path`, but an explicit arm or reject would be tidier.

fallintoplace:
@zeroshade I can work on the nitpicking comments right now. Hold on a bit before I push this.

zeroshade:
Quick question @fallintoplace: is there a reason not to adopt the approach i mentioned in https://github.com/apache/iceberg-go/pull/1229#discussion_r3444338656 which should simplify things further?

table/arrow_scanner.go
fallintoplace: I don’t think we can assume the Unique result is always dictionary-encoded. My understanding is that unique returns a single ArrayDatum because OutputChunked=false, but the array’s concrete type follows the input layout: plain string / large string inputs can produce a string-like array, while dictionary input produces a dictionary array. Since the main bug here is plain-string file_path layouts, the dict-only wrapper path would miss or panic on the case this PR is fixing.

I’m happy to keep the simplification around assuming ArrayDatum, but I think we still need the array-level type switch for array.StringLike vs *array.Dictionary.

zeroshade: Ah, I was misreading the uniqueFinalizeDictionary and missing that it only returns the dictionary if the input was a dictionary. (ironic since i'm the one who wrote the whole thing....)

We should definitely simplify using `ArrayDatum` and then do the type switch so that if it's a dictionary, create the NewDictWrapper so it looks something like:

```go
var values array.StringLike
switch v := arr.(type) {
case *array.Dictionary:
    values = array.NewDictWrapper[string](v)
default:
    values = arr.(array.StringLike)
}

// do stuff....
```

which would still simplify the code a lot and avoid needing the whole scalar handling you're doing

PR #1234, opened 2026-06-19
fix(table): return error instead of panicking on unsupported geo CRS

### LINKED ISSUE #1232: Return an error instead of panicking on unsupported CRS in icebergCRSToGeoArrowMetadata
Follow-up to #1138.

`icebergCRSToGeoArrowMetadata` panics on an unsupported (projjson) CRS, while the symmetric read function `geoArrowCRSToIcebergCRS` returns a clean error for the same case. It only works today because `iceberg.Visit`'s deferred recover catches the panic — call the function outside a visitor and it blows up.

Proposal: give `icebergCRSToGeoArrowMetadata` a `(geoarrow.Metadata, error)` signature, and have `VisitGeometry`/`VisitGeography` panic-with-error to thread it through the visitor recover (the convention the other visitor methods already use). That removes the read/write asymmetry and makes the function safe to call on its own.

This is the only one of the #1138 follow-ups that touches a signature, so it's cleaner as its own PR.

laskoviymishka:
icebergCRSToGeoArrowMetadata panicked on a projjson CRS while the read-side geoArrowCRSToIcebergCRS returned a clean error; it only worked because iceberg.Visit's recover caught the panic. Give it a (geoarrow.Metadata, error) signature; VisitGeometry/VisitGeography re-panic to thread the error through the visitor recover (the file's existing idiom). Public API behavior unchanged. Adds direct + end-to-end Geometry/Geography projjson coverage.

Fixes #1232.

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -673,7 +673,13 @@ func (c convertToArrow) VisitVariant() arrow.Field {
 }
 
 func (c convertToArrow) VisitGeometry(g iceberg.GeometryType) arrow.Field {
-	meta := icebergCRSToGeoArrowMetadata(g.CRS())
+	meta, err := icebergCRSToGeoArrowMetadata(g.CRS())
+	if err != nil {
+		// Panic to thread the error through iceberg.Visit's recover, matching the
+		// convention used by the other visitor methods.
+		panic(err)
+	}
+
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Lgtm

PR #1237, opened 2026-06-19
fix(v3): _row_id renumbering when rows are dropped in a scan

twuebi:
## Problem

Scanning or compacting a v3 table renumbers surviving rows' `_row_id` whenever rows are dropped — by a deletion vector, positional/equality deletes, or a row filter. The spec requires `_row_id = first_row_id + the row's original position in its file`, so survivors must keep their ids, instead they came out as a dense `0..N` sequence.

## Cause

Row-lineage synthesis runs last in the scan pipeline, after the row-dropping steps, and counts positions over the already-filtered batch. Independently, Parquet row-group pruning skips whole groups before the batch reaches the pipeline, leaving survivor positions non-contiguous.

## Fix

Synthesize `_row_id` / `_last_updated_sequence_number` first, before any row-dropping step, counting over the full unfiltered batch; later filters drop rows already carrying the correct id. Skip row-group pruning while synthesizing lineage so positions stay contiguous — the per-row filter still enforces the predicate.

## Tests

`TestRewriteDataFilesPreservesRowIDThroughDeletionVector` (atomic + partial-progress) and `TestScanRowLineagePreservedAcrossPrunedRowGroups` (a filter prunes a leading row group; survivors keep their ids).

### DIFF
--- a/metadata_columns.go
+++ b/metadata_columns.go
@@ -63,22 +63,17 @@ func IsMetadataColumn(fieldID int) bool {
 	return fieldID == RowIDFieldID || fieldID == LastUpdatedSequenceNumberFieldID
 }
 
-// SchemaWithRowLineage returns a new schema with both row-lineage metadata columns
-// (_row_id, _last_updated_sequence_number) appended to the given schema's fields.
-// Used when reading source files during a CoW rewrite or compaction so that row
-// identity and per-row update sequence are preserved in the output.
+// SchemaWithRowLineage appends both row-lineage metadata columns (_row_id,
+// _last_updated_sequence_number) so a CoW rewrite or compaction preserves them.
 func SchemaWithRowLineage(s *Schema) *Schema {
 	return SchemaWithRowLineageColumns(s, true, true)
 }
 
-// SchemaWithRowLineageColumns returns a new schema with the requested row-lineage
-// metadata columns appended: _row_id when rowID is set, _last_updated_sequence_number
-// when lastUpdatedSeq is set. Callers that synthesize these columns conditionally must
-// request exactly the ones they will materialize so the read schema stays aligned with
-// the produced batch.
+// SchemaWithRowLineageColumns appends the requested row-lineage columns: _row_id
+// when rowID, _last_updated_sequence_number when lastUpdatedSeq. Request exactly
+// the columns you will materialize so the read schema matches the produced batch.
 //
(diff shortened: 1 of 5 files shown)


twuebi:
@laskoviymishka got another fix lined up for mor positional deletes, depends on some things from here

zeroshade (approved):
Thanks for this fix — the pipeline reordering is spec-correct (`_row_id = first_row_id + _pos` computed over the row's original file position), memory-safe, and well covered by the new regression tests. Approving. A few non-blocking observations:

**1. Performance trade-off: row-group pruning is disabled whenever `_row_id` is projected.**
Because `needsOriginalPositions = synthesizeRowID`, any scan that projects `_row_id` now reads every row group of each selected file, even under a selective row filter. This is mitigated by file/manifest-level pruning still applying in `PlanFiles` (whole files can still be skipped) — only intra-file row-group skipping is given up, and it's necessary to keep survivor positions contiguous. A possible future optimization is to track each row group's base position and feed it into the offset accounting, which would let pruning stay enabled while preserving `first_row_id + pos`; that needs the reader to expose row-group boundaries, so leaving it out of scope here seems reasonable. Might be worth a short code comment calling out the perf implication for row-lineage scans.

**2. Pre-existing positional-delete / DV counters rely on the same position-contiguity assumption.**
`processPositionalDeletes` and `filterByDeletionVector` advance a running `nextIdx` per batch (arrow_scanner.go ~330-336 and ~365-373), i.e. they assume the reader emits absolute, contiguous file positions. With a row filter active + row-group pruning enabled + no row-lineage projection, that counter can drift from true file positions. This is independent of this PR — you don't regress it, and you actually fix it for the lineage-projected path since pruning is disabled there. Flagging it only because it's the same class of bug; presumably the positional-deletes follow-up you mentioned covers it. Worth confirming that follow-up also addresses the non-lineage scan path.

**3. Minor: `SchemaWithRowID` now overlaps `SchemaWithRowLineageColumns(s, true, false)`.**
With the new generalized helper, `SchemaWithRowID(s)` is equivalent to `SchemaWithRowLineageColumns(s, true, false)`. The former could delegate to the latter to keep a single code path. Purely cosmetic.

None of these block merge.

twuebi:
already prepared a followup for 2. can fold 3 into that, will see what we can do about 1. next week

PR #1238, opened 2026-06-19
feat(table): variant shredded reader + bump arrow-go version

### LINKED ISSUE #986: feat(table): variant shredded reader

nssalian:
Closes [#986](https://github.com/apache/iceberg-go/issues/986).

At scan time, reassembles shredded variant columns (metadata + residual value + typed_value) back into the plain unshredded layout, so a shredded variant reads the same as a non-shredded one. arrow-go's VariantArray does the per-row work. The code change is only in the projection visitor's Variant method.

Note: This PR also bumps arrow-go to 4a540f5 (https://github.com/apache/arrow-go/commit/4a540f5144843a25db8cffe828dd7deeb34260e1) (v18.6.1-0.20260623163317-4a540f514484) to pick up `UnshredVariant` (https://github.com/apache/arrow-go/pull/859), which does the per-row reassembly.

## Fixture based Testing

The fixtures come from [apache/parquet-testing](https://github.com/apache/parquet-testing/tree/master/shredded_variant) (written by Iceberg Java's TestVariantReaders). I first ran all 137 upstream cases through arrow-go to see where it stands (128 pass, 4 reject, 2 lenient, 3 spec-invalid), then committed 17 of them. These represent one per distinct reader behaviour instead of all 137:

- 9 success cases checked against the Java golden value (array, empty array, scalar, object with a field that falls back to the residual value, object with no value column, nested array, null variant, unshredded data under a shredded schema, and a multi-row column with nulls). Compared as JSON because reassembly is valid but not byte-identical for nested objects.
- 5 cases the reader should reject (value/typed_value conflicts, non-object-with-shredded-fields, and a spec-invalid file).
- 3 cases where arrow-go reads something Java would reject (uint32 and fixed_size_binary typed_value, plus one spec-invalid file it reads). These pin the current behaviour so we notice if arrow-go changes it.

## Additional Testing

- compute.Take / FilterRecordBatch leave a shredded column alone (they run before projection).
- rows surviving a positional delete still reassemble correctly.
- a leak check on the reassembly path.
- a Spark 4 integration test that writes a shredded table and reads it back (checks the file is actually shredded first).


nssalian:
@laskoviymishka @zeroshade PTAL when you get a chance.

table/arrow_utils.go
```diff
@@ -1190,8 +1178,33 @@ func (a *arrowProjectionVisitor) Primitive(_ iceberg.PrimitiveType, arr arrow.Ar
 	return arr
 }
 
+// Variant reassembles a shredded variant back into the unshredded layout so the
+// projected schema matches; without it the scanner fails to assemble the table.
```
zeroshade, line 1195: I'm surprised that this is the case, this feels like a bug in arrow-go/parquet rather than something we should have to do here. Can you take a look at arrow-go/parquet and validate why it isn't already reassembling the shredded column into a variant array properly and whether or not this is something we should fix on that side instead?

nssalian: I looked at it for a bit. It's not really a bug: the reader builds the variant array's storage straight from the file's physical schema, so a shredded file comes back as a shredded VariantArray (with Shredded()/UntypedValues() for typed-column access), and Value(i) reassembles each row on demand. What's missing is a way to collapse a whole shredded array into the plain layout, which is what the scan needs since it can't mix shredded and non-shredded arrays in one column. 

I prototyped that as a small helper in arrow-go and used it here. It passes the tests locally. I'll put it up as an arrow-go PR (will tag you for a review) then update this PR to use it and drop the manual reassembly here.

zeroshade: the utility is merged, you can update the arrow-go dep here and change this to use it (I'll plan on doing a new release of arrow-go in the next couple weeks)

nssalian:
The arrow-go PR has been merged, I'll bring the fix in here in an upcoming commit

table/arrow_utils.go
```diff
@@ -1191,8 +1191,22 @@ func (a *arrowProjectionVisitor) Primitive(_ iceberg.PrimitiveType, arr arrow.Ar
...
 func (a *arrowProjectionVisitor) Variant(_ iceberg.VariantType, arr arrow.Array) arrow.Array {
```
zeroshade, line 1196: Correctness here quietly depends on `VariantType` NOT implementing `arrow.NestedType`: in `Struct` (line ~1078) the `defer arr.Release()` only fires for nested types, so the owned +1 from `UnshredVariant` is balanced instead by `castIfNeeded`s `Retain()`/deferred `Release()`. If arrow-go ever made `VariantType` a `NestedType`, line 1078 would fire and double-release. The two checked-allocator leak tests defend this, but a one-line comment noting the invariant (relies on castIfNeeded retain/release; VariantType is intentionally not NestedType) would make it explicit for the next reader.

zeroshade (approved):
Reviewed the full change and verified it locally against the PR branch: `go build ./...`, `go test -race` (all variant/shredded tests), `golangci-lint run` (0 issues), and `go mod tidy` (clean) all pass.

The production change is small and well-contained — only `(*arrowProjectionVisitor).Variant` reassembles a shredded `*extensions.VariantArray` via `extensions.UnshredVariant`, leaving non-shredded/non-variant arrays untouched. Nice work on the fixture-based testing: ASF-licensed goldens vendored from `apache/parquet-testing` at a pinned commit, reproducible via `gen_fixtures.sh`, with each case mapped to the reader behaviour it pins (success / reject / leniency), and JSON comparison for nested objects. Coverage of the `compute.Take`/`FilterRecordBatch`-before-projection ordering and the positional-delete path is a good touch.

I spent most of the review on memory ownership, since `UnshredVariant` returns a newly-allocated, caller-owned array. The two `CheckedAllocator` + `AssertSize(t, 0)` tests (`TestShreddedVariantReassemblyNoLeak` and `TestShreddedVariantSurvivesPositionalDelete`) run the real projection path and pass clean under `-race`, so the owned reference is balanced correctly by `castIfNeeded`'s retain/release. The `panic` on `UnshredVariant` error is fine here: the `Variant` method has no error return, and `VisitSchemaWithPartner` recovers panics into errors, so a malformed shredded file becomes a scan error rather than a crash — consistent with the existing `castIfNeeded`/`Struct` convention in this file.

LGTM. A couple of non-blocking notes inline and below.

Non-blocking:
- The testdata README references `go-notes/986-shredded-conformance-and-patches.md`, which doesn't appear to be committed in this PR. Either commit the survey note or drop the reference so readers aren't pointed at a missing file.
- Worth confirming the Spark 4 shredded integration test actually executed in a Spark-4 CI lane rather than silently skipping on Spark 3.5.

table/arrow_utils.go
nssalian: Clarified this with a comment.

nssalian:
Thanks for the detailed review @zeroshade.

Changes:
- Removed the reference to the conformance. I had done local tests for all the cases and captured that in a note. Doesn't need to be checked in here.
- Checked the Github Actions run log for Spark4 to see that all tests are passing. I kept them distinct so they are individually visible that they are passing.

From the run
```
    --- PASS: TestSparkIntegration/TestShreddedVariantSparkWriteGoRead (20.65s)
        --- PASS: TestSparkIntegration/TestShreddedVariantSparkWriteGoRead/nested_object_reassembles (0.00s)
        --- PASS: TestSparkIntegration/TestShreddedVariantSparkWriteGoRead/partial_shred_merges_residual (0.00s)
        --- PASS: TestSparkIntegration/TestShreddedVariantSparkWriteGoRead/top-level_scalar (0.00s)
        --- PASS: TestSparkIntegration/TestShreddedVariantSparkWriteGoRead/top-level_array (0.00s)
        --- PASS: TestSparkIntegration/TestShreddedVariantSparkWriteGoRead/json_null_is_present (0.00s)
        --- PASS: TestSparkIntegration/TestShreddedVariantSparkWriteGoRead/sql_null_is_physical_null (0.00s)
        --- PASS: TestSparkIntegration/TestShreddedVariantSparkWriteGoRead/spark_reads_merged_variant_back (9.91s)
    --- PASS: TestSparkIntegration/TestUnknownTypeWriteAndScan (7.57s)
    --- PASS: TestSparkIntegration/TestUpdateSpec (7.75s)
    --- PASS: TestSparkIntegration/TestVariantWriteAndScan (9.81s)
```

PR #1239, opened 2026-06-20
fix(table): harden GeoArrow CRS read parsing

### LINKED ISSUE #1231: Harden CRS read path in geoArrowCRSToIcebergCRS (form-based matching, reject projjson strings)
Follow-up to #1138.

The Arrow→Iceberg CRS parsing in `geoArrowCRSToIcebergCRS` leans on a `len(crs) <= 32` length check in the string-CRS `default` branch, which accepts arbitrary short strings and rejects long-but-valid ones. A few things worth tightening up:

- Replace the length heuristic with form-based matching — pass `authority:code`-shaped strings through regardless of length (the `authorityCodeCRS` regex already used on the write side works for this), and reject other forms with a specific error.
- Reject projjson-typed *string* CRS on read. Right now a `CRSTypePROJJSON` string slips through the same `default` branch and is accepted silently, even though the write path rejects projjson.
- Give the WKT2:2019 case a consistent "not supported" message whether or not `crs_type` is set. Today a bare WKT2 string falls into the length branch and errors with "crs length too long", which is misleading.
- The JSON-object branch returns the same `errors.New("unsupported CRS")` from six different failure points. Putting the offending member into the message (or wrapping a sentinel) would make a parse failure tell you what was actually wrong.

None of this is on the common path, so it's a robustness + diagnostics cleanup rather than a correctness fix — good first issue territory.

Revanth14:
closes #1231 

## Summary
- Reject unsupported GeoArrow string CRS encodings for `projjson` and WKT2:2019.
- Keep non-empty string CRS values readable to preserve existing data and read/write round-tripping.
- Preserve canonical `OGC:CRS84` / `EPSG:4326` handling even when `crs_type:"srid"` is present.
- Improve JSON-object CRS diagnostics for invalid `id`, `id.authority`, and `id.code` members.

## Testing
- go test ./table -run 'Test(IcebergGeoTypesToArrowSchema|GeoArrowCRSToIcebergCRS|IcebergCRSToGeoArrowMetadata)'
- go test ./table

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1875,11 +1875,14 @@ func checkCRSJSON(rawCrs json.RawMessage) bool {
 // package does not support yet.
 var errWKT2CRSNotSupported = errors.New("CRS type wkt2:2019 not supported")
 
+// errPROJJSONStringCRSNotSupported is returned for projjson-typed string CRS
+// values, which this package does not support yet.
+var errPROJJSONStringCRSNotSupported = errors.New("CRS type projjson not supported for string CRS")
+
 // isWKT2CRSString reports whether crs looks like a WKT2 (ISO 19162) CRS
 // definition. It is a best-effort heuristic used only to surface the clearer
-// errWKT2CRSNotSupported message: a WKT2 string fails the authority:code check
-// regardless, so a missing keyword only changes the error text, not the
-// outcome. The list covers the WKT2:2019 CRS keywords and their long forms.
+// errWKT2CRSNotSupported message. The list covers the WKT2:2019 CRS keywords
+// and their long forms.
 func isWKT2CRSString(crs string) bool {
 	upper := strings.ToUpper(strings.TrimSpace(crs))
 
@@ -1920,12 +1923,13 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 		if err := json.Unmarshal(meta.CRS, &crs); err != nil {
 			return "", fmt.Errorf("invalid geoarrow CRS metadata: %w", err)
 		}
+		if crs == "" {
+			return "", errors.New("unsupported CRS: empty string CRS")
+		}
 
 		switch meta.CRSType {
-		case geoarrow.CRSTypeSRID:
-			return "srid:" + crs, nil
 		case geoarrow.CRSTypePROJJSON:
-			return "", errors.New("CRS type projjson not supported for string CRS")
+			return "", errPROJJSONStringCRSNotSupported
 		case geoarrow.CRSTypeWKT22019:
 			return "", errWKT2CRSNotSupported
 		}
@@ -1936,11 +1940,15 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 		if isWKT2CRSString(crs) {
 			return "", errWKT2CRSNotSupported
 		}
-		if authorityCodeCRS.MatchString(crs) {
-			return crs, nil
+		// SRID is resolved after the canonical and WKT2 checks so a
+		// contradictory crs_type=srid on a value like "EPSG:4326" still
+		// canonicalizes rather than producing "srid:EPSG:4326". Real numeric
+		// SRIDs match none of the checks above and fall through to here.
+		if meta.CRSType == geoarrow.CRSTypeSRID {
+			return "srid:" + crs, nil
 		}
 
-		return "", errors.New("unsupported CRS string: expected authority:code form")
+		return crs, nil
 	case checkCRSJSON(meta.CRS):
 		var crs map[string]json.RawMessage
 
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1871,6 +1871,43 @@ func checkCRSJSON(rawCrs json.RawMessage) bool {
 	return len(b) > 0 && b[0] == '{'
 }
 
+// errWKT2CRSNotSupported is returned for WKT2:2019 CRS definitions, which this
+// package does not support yet.
+var errWKT2CRSNotSupported = errors.New("CRS type wkt2:2019 not supported")
+
+// isWKT2CRSString reports whether crs looks like a WKT2 (ISO 19162) CRS
+// definition. It is a best-effort heuristic used only to surface the clearer
+// errWKT2CRSNotSupported message: a WKT2 string fails the authority:code check
+// regardless, so a missing keyword only changes the error text, not the
+// outcome. The list covers the WKT2:2019 CRS keywords and their long forms.
+func isWKT2CRSString(crs string) bool {
+	upper := strings.ToUpper(strings.TrimSpace(crs))
+
+	for _, prefix := range []string{
+		"BOUNDCRS[",
+		"COMPOUNDCRS[",
+		"DERIVEDPROJCRS[",
+		"ENGCRS[",
+		"ENGINEERINGCRS[",
+		"GEODCRS[",
+		"GEODETICCRS[",
+		"GEOGCRS[",
+		"GEOGRAPHICCRS[",
+		"PARAMETRICCRS[",
+		"PROJCRS[",
+		"PROJECTEDCRS[",
+		"TIMECRS[",
+		"VERTCRS[",
+		"VERTICALCRS[",
+	} {
+		if strings.HasPrefix(upper, prefix) {
+			return true
+		}
+	}
+
+	return false
+}
+
 func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 	if len(meta.CRS) == 0 {
 		return "srid:0", nil
@@ -1884,22 +1921,26 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 			return "", fmt.Errorf("invalid geoarrow CRS metadata: %w", err)
 		}
 
-		if strings.EqualFold(crs, "OGC:CRS84") || strings.EqualFold(crs, "EPSG:4326") {
-			return "OGC:CRS84", nil
-		}
-
 		switch meta.CRSType {
 		case geoarrow.CRSTypeSRID:
 			return "srid:" + crs, nil
+		case geoarrow.CRSTypePROJJSON:
+			return "", errors.New("CRS type projjson not supported for string CRS")
 		case geoarrow.CRSTypeWKT22019:
-			return "", errors.New("CRS type wkt2:2019 not supported")
-		default:
-			if len(crs) <= 32 {
-				return crs, nil
-			}
-
-			return "", errors.New("crs length too long")
+			return "", errWKT2CRSNotSupported
 		}
+
+		if strings.EqualFold(crs, "OGC:CRS84") || strings.EqualFold(crs, "EPSG:4326") {
+			return "OGC:CRS84", nil
+		}
+		if isWKT2CRSString(crs) {
+			return "", errWKT2CRSNotSupported
+		}
+		if authorityCodeCRS.MatchString(crs) {
+			return crs, nil
+		}
+
+		return "", errors.New("unsupported CRS string: expected authority:code form")
 	case checkCRSJSON(meta.CRS):
 		var crs map[string]json.RawMessage
 
@@ -1909,34 +1950,41 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 
 		idRaw, ok := crs["id"]
 		if !ok || len(idRaw) == 0 {
-			return "", errors.New("unsupported CRS")
+			return "", errors.New("unsupported CRS: missing id")
 		}
 
 		var id map[string]json.RawMessage
 		if err := json.Unmarshal(idRaw, &id); err != nil {
-			return "", errors.New("unsupported CRS")
+			return "", fmt.Errorf("unsupported CRS: invalid id: %w", err)
 		}
 
 		var authority string
-		if err := json.Unmarshal(id["authority"], &authority); err != nil || authority == "" {
-			return "", errors.New("unsupported CRS")
+		authorityRaw, ok := id["authority"]
+		if !ok || len(authorityRaw) == 0 {
+			return "", errors.New("unsupported CRS: missing id.authority")
+		}
+		if err := json.Unmarshal(authorityRaw, &authority); err != nil {
+			return "", fmt.Errorf("unsupported CRS: invalid id.authority: %w", err)
+		}
+		if authority == "" {
+			return "", errors.New("unsupported CRS: empty id.authority")
 		}
 
 		codeRaw := id["code"]
 		if len(codeRaw) == 0 {
-			return "", errors.New("unsupported CRS")
+			return "", errors.New("unsupported CRS: missing id.code")
 		}
 
 		var code string
 		if err := json.Unmarshal(codeRaw, &code); err != nil {
 			var codeNum json.Number
 			if err := json.Unmarshal(codeRaw, &codeNum); err != nil {
-				return "", errors.New("unsupported CRS")
+				return "", fmt.Errorf("unsupported CRS: invalid id.code: %w", err)
 			}
 			code = codeNum.String()
 		}
 		if code == "" {
-			return "", errors.New("unsupported CRS")
+			return "", errors.New("unsupported CRS: empty id.code")
 		}
 
 		authorityCode := authority + ":" + code
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1889,6 +1889,9 @@ func isWKT2CRSString(crs string) bool {
 	for _, prefix := range []string{
 		"BOUNDCRS[",
 		"COMPOUNDCRS[",
+		"COORDINATEMETADATA[",
+		"COORDINATEOPERATION[",
+		"CRS[",
 		"DERIVEDPROJCRS[",
 		"ENGCRS[",
 		"ENGINEERINGCRS[",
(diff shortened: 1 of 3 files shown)


tanmayrauth (commented):
Nice hardening — the per-member JSON id errors are a real readability win, and the projjson-string rejection is good to have.

  One thing I want to make sure we've thought through: the new authority:code regex at arrow_utils.go:1939 is now the only accepted shape for a string CRS on read, but the write side doesn't enforce it. icebergCRSToGeoArrowMetadata  (arrow_utils.go:2050) serializes any string, and GeometryTypeOf/GeographyTypeOf (types.go:847, :907)  accept any non-empty CRS with no shape check. So something like geometry("my-custom-crs") writes fine today but fails to read back after this change — which the old `len(crs) <= 32` branch used to allow. That makes it a forward-compat break for tables already written with a free-form CRS string. Is the plan to also tighten the constructor/write path so those values can't be produced, or should read stay lenient for already-written data? Right now read and write disagree on what's valid.

Revanth14:
> Nice hardening — the per-member JSON id errors are a real readability win, and the projjson-string rejection is good to have.
> 
> One thing I want to make sure we've thought through: the new authority:code regex at arrow_utils.go:1939 is now the only accepted shape for a string CRS on read, but the write side doesn't enforce it. icebergCRSToGeoArrowMetadata (arrow_utils.go:2050) serializes any string, and GeometryTypeOf/GeographyTypeOf (types.go:847, :907) accept any non-empty CRS with no shape check. So something like geometry("my-custom-crs") writes fine today but fails to read back after this change — which the old `len(crs) <= 32` branch used to allow. That makes it a forward-compat break for tables already written with a free-form CRS string. Is the plan to also tighten the constructor/write path so those values can't be produced, or should read stay lenient for already-written data? Right now read and write disagree on what's valid.

@tanmayrauth  thanks for the review.

Good catch. This does create a round-trip break: `geometry("my-custom-crs")` can be written today, but the new read path rejects it.

I think read should stay lenient for non-empty string CRS values, since tightening write would not help existing data and would still leave constructors more permissive than the writer. We can still reject `projjson` string CRS and WKT2, since those are unsupported.

Will wait for @laskoviymishka thoughts before updating. If that direction sounds right, I’ll add a round-trip regression test for `GeometryTypeOf("my-custom-crs")`.

table/arrow_utils.go
```diff
@@ -1884,22 +1921,26 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
...
-			return "", errors.New("CRS type wkt2:2019 not supported")
-		default:
-			if len(crs) <= 32 {
-				return crs, nil
-			}
+			return "", errWKT2CRSNotSupported
+		}
 
-			return "", errors.New("crs length too long")
+		if strings.EqualFold(crs, "OGC:CRS84") || strings.EqualFold(crs, "EPSG:4326") {
```
laskoviymishka: Moving this canonicalization below the `CRSType` switch quietly changes behavior for typed inputs.

Before this PR the OGC:CRS84/EPSG:4326 check ran first, so `{"crs":"EPSG:4326","crs_type":"srid"}` read back as `OGC:CRS84`. Now it hits the `CRSTypeSRID` case first and returns `srid:EPSG:4326` — an SRID-typed authority string that won't round-trip as a real SRID. Same shape for `{"crs":"OGC:CRS84","crs_type":"srid"}` -> `srid:OGC:CRS84`.

I'd put the canonical check back above the switch so it wins regardless of the annotated type. If we deliberately want the `CRSType` annotation to take precedence here, that's defensible too — but then it needs a test row asserting `srid:EPSG:4326` on purpose, since nothing covers the SRID + canonical-value combination today. wdyt?

table/arrow_utils.go
```diff
@@ -1884,22 +1921,26 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
...
 		}
 
-		if strings.EqualFold(crs, "OGC:CRS84") || strings.EqualFold(crs, "EPSG:4326") {
-			return "OGC:CRS84", nil
-		}
-
 		switch meta.CRSType {
 		case geoarrow.CRSTypeSRID:
 			return "srid:" + crs, nil
+		case geoarrow.CRSTypePROJJSON:
```
laskoviymishka, line 1938: This branch flips a previously-tolerated input into a hard read error — the same behavior the deleted `geometry_epsg_4326_incorrect_type` case used to exercise.

`{"crs":"EPSG:4326","crs_type":"projjson"}` is spec-invalid (projjson means the crs was written as PROJJSON, not as a string), but the old code accepted it best-effort via the CRS84/EPSG check and returned `OGC:CRS84`. A client that erroneously wrote that string now fails on read.

I lean toward erroring being the more correct behavior — but it's a read-tolerance regression that isn't called out in the description, and the schema-roundtrip case that pinned the old direction got deleted rather than converted. Either keep the CRS84/EPSG canonicalization ahead of this branch so canonical values still read, or keep the rejection and note it in the PR description so it's clearly intentional. Which way do you want it?

laskoviymishka (changes requested):
Nice hardening pass, the new `TestGeoArrowCRSToIcebergCRS` table and the per-member `id.authority`/`id.code` errors are a clear step up from the old single "unsupported CRS" string, and dropping the `len <= 32` heuristic for explicit checks reads much better.

One thing I'd settle before merge: relocating the OGC:CRS84/EPSG:4326 canonicalization to *after* the `CRSType` switch silently changes how typed-but-canonical inputs read. `{"crs":"EPSG:4326","crs_type":"srid"}` now comes back as `srid:EPSG:4326`, and the same value with `crs_type:"projjson"` now hard-errors where it used to be tolerated. Both were loosely covered by the `geometry_epsg_4326_incorrect_type` case this PR deletes, so the change is real but nothing pins the intended outcome anymore. Details are inline: let's make the call explicit and lock it with a test either way.

table/arrow_utils.go
Revanth14: Good catch. Agreed, and addressed.

I moved SRID resolution after the canonical check, so the `OGC:CRS84` / `EPSG:4326` collapse wins even with `crs_type:"srid"`:

- `{"crs":"EPSG:4326","crs_type":"srid"}` → `OGC:CRS84`
- `{"crs":"OGC:CRS84","crs_type":"srid"}` → `OGC:CRS84`
- `{"crs":"3857","crs_type":"srid"}` → `srid:3857`

I added test coverage for both the canonical+SRID case and the numeric SRID case, and left a comment explaining why SRID is resolved last.

One intentional exception: `projjson`- and `wkt2`-typed strings still reject before canonicalization, since those encodings are unsupported by this path.

table/arrow_utils.go
Revanth14: I kept the rejection here intentionally.

`crs_type:"projjson"` says the CRS payload should be PROJJSON, so a string CRS with that annotation is invalid for this path. I agree it is a read-tolerance change from the old best-effort canonicalization, but this is one of the unsupported encodings the PR is meant to reject.

I kept that behavior covered with the `projjson` string CRS test, and I’ll call it out in the PR description so the compatibility change is explicit.

table/arrow_utils.go
```diff
@@ -1883,23 +1923,32 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
...
-		if strings.EqualFold(crs, "OGC:CRS84") || strings.EqualFold(crs, "EPSG:4326") {
-			return "OGC:CRS84", nil
+		if crs == "" {
+			return "", errors.New("unsupported CRS: empty string CRS")
 		}
 
 		switch meta.CRSType {
-		case geoarrow.CRSTypeSRID:
-			return "srid:" + crs, nil
+		case geoarrow.CRSTypePROJJSON:
```
laskoviymishka, line 1938: I think there's a subtle inconsistency here. The SRID arm was deliberately moved below the canonical check (with the comment a few lines down) so that `EPSG:4326 + crs_type=srid` still canonicalizes to `OGC:CRS84` rather than becoming `srid:EPSG:4326`. But projjson and wkt2:2019 fire *before* the canonical check, so `{"crs":"EPSG:4326","crs_type":"projjson"}` and `{"crs":"EPSG:4326","crs_type":"wkt2:2019"}` both error — that's exactly the case the deleted `geometry_epsg_4326_incorrect_type` test used to cover.

A canonical authority code is unambiguous regardless of a bogus `crs_type`, so I'd apply the same rule you already applied to SRID: run the `OGC:CRS84`/`EPSG:4326` check first, then let projjson/wkt2:2019 reject only non-canonical strings. That makes all four arms consistent. wdyt?

table/arrow_utils.go
```diff
@@ -1883,23 +1923,32 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
...
-			return "", errors.New("CRS type wkt2:2019 not supported")
-		default:
-			if len(crs) <= 32 {
-				return crs, nil
-			}
+			return "", errWKT2CRSNotSupported
+		}
 
-			return "", errors.New("crs length too long")
+		if strings.EqualFold(crs, "OGC:CRS84") || strings.EqualFold(crs, "EPSG:4326") {
```
laskoviymishka: This is the move I'd make to fix the ordering above: lift this canonical block to the very top of the string case, right after the empty-string guard and before `switch meta.CRSType`. Once it short-circuits first, the projjson and wkt2:2019 arms can stay as-is and the SRID special-casing comment below becomes unnecessary, since canonical will already have returned.

table/arrow_utils.go
```diff
@@ -1871,6 +1871,46 @@ func checkCRSJSON(rawCrs json.RawMessage) bool {
...
+
+	for _, prefix := range []string{
+		"BOUNDCRS[",
+		"COMPOUNDCRS[",
+		"DERIVEDPROJCRS[",
+		"ENGCRS[",
+		"ENGINEERINGCRS[",
+		"GEODCRS[",
+		"GEODETICCRS[",
+		"GEOGCRS[",
```
laskoviymishka, line 1900: The keyword set is missing a few valid WKT2:2019 top-level forms — `COORDINATEMETADATA[`, `COORDINATEOPERATION[`, and the abstract `CRS[` wrapper — so a string wrapped in those passes the heuristic and gets stored as an opaque CRS instead of hitting the clearer error. I'd add them.

While we're here: this is a best-effort prefix match, so it'll also fire on a non-WKT2 authority string that happens to start with one of these tokens — worth a one-line comment that it's intentionally heuristic and only used to pick the friendlier error.

table/arrow_utils.go
```diff
@@ -1883,23 +1923,32 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
...
+		case geoarrow.CRSTypePROJJSON:
+			return "", errPROJJSONStringCRSNotSupported
 		case geoarrow.CRSTypeWKT22019:
-			return "", errors.New("CRS type wkt2:2019 not supported")
-		default:
-			if len(crs) <= 32 {
-				return crs, nil
-			}
+			return "", errWKT2CRSNotSupported
+		}
```
laskoviymishka, line 1945: This switch handles two of the four `CRSType` constants with no `default`, which the exhaustive linter will flag, and means a future fifth constant silently falls through to `return crs, nil`. I'd add an explicit `default:` with a short comment that SRID/AuthorityCode/unset are intentionally handled by the checks below.

table/arrow_utils.go
```diff
@@ -1883,23 +1923,32 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
...
 		}
+		// SRID is resolved after the canonical and WKT2 checks so a
+		// contradictory crs_type=srid on a value like "EPSG:4326" still
+		// canonicalizes rather than producing "srid:EPSG:4326". Real numeric
+		// SRIDs match none of the checks above and fall through to here.
+		if meta.CRSType == geoarrow.CRSTypeSRID {
+			return "srid:" + crs, nil
+		}
+
+		return crs, nil
```
laskoviymishka, line 1956: This final `return crs, nil` is the catch-all passthrough for AuthorityCode and unset types, but that policy is implicit. One line — something like "unrecognized type or bare string: accept as opaque CRS per spec (any unique identifier is valid)" — would save the next reader the trace-through.

table/arrow_utils_internal_test.go
```diff
@@ -491,3 +492,128 @@ func TestIcebergCRSToGeoArrowMetadata(t *testing.T) {
...
+		assert.ErrorContains(t, err, "unsupported CRS: empty string CRS")
+	})
+
+	t.Run("rejects projjson string CRS", func(t *testing.T) {
+		_, err := geoArrowCRSToIcebergCRS(geoarrow.Metadata{
+			CRS:     stringCRS("EPSG:4326"),
+			CRSType: geoarrow.CRSTypePROJJSON,
+		})
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "CRS type projjson not supported for string CRS")
```
laskoviymishka: These two are package-level sentinels (`errPROJJSONStringCRSNotSupported` here and `errWKT2CRSNotSupported` just below), and this is the internal test, so it can reach them — I'd assert with `assert.ErrorIs(t, err, errPROJJSONStringCRSNotSupported)` rather than matching the message text. Keeps the test from breaking on wording changes and actually pins the sentinel. (The external `arrow_utils_test.go` can't see the unexported vars, so `ErrorContains` is correct there.)

table/arrow_utils_internal_test.go
```diff
@@ -491,3 +492,128 @@ func TestIcebergCRSToGeoArrowMetadata(t *testing.T) {
 		assert.Empty(t, meta.CRSType)
 	})
 }
+
+func TestGeoArrowCRSToIcebergCRS(t *testing.T) {
+	stringCRS := func(s string) json.RawMessage {
```
laskoviymishka, line 497: `stringCRS` closes over the outer `t`, so a `require.NoError` failure inside a subtest gets reported against the parent. `json.Marshal` of a string can't fail, so I'd just drop the error check (or take a `testing.TB` and call `tb.Helper()` if you want to keep it defensive).

table/arrow_utils_test.go
```diff
@@ -608,11 +615,6 @@ func TestIcebergGeoTypesToArrowSchema(t *testing.T) {
 			ice:              defaultGeometry,
 			geoarrowMetaJSON: `{"crs":"epsg:4326"}`,
 		},
-		{
-			name:             "geometry_epsg_4326_incorrect_type",
```
laskoviymishka, line 612: Once the canonical-first reordering lands, this case should come back with its original canonicalizing assertion (`OGC:CRS84`) — it's the regression guard for the exact ordering bug. I'd also add the sibling `{"crs":"EPSG:4326","crs_type":"wkt2:2019"}` case, since that path is currently untested and is where the expected behavior is easiest to get wrong.

laskoviymishka (changes requested):
Nice hardening pass, dropping the 32-char cap is the right call, and the new `TestGeoArrowCRSToIcebergCRS` table plus the per-member JSON error messages (`unsupported CRS: missing id.authority` and friends) are a clear improvement over the old single "unsupported CRS".

The one thing I'd want sorted before this lands is the new ordering, which reads as inconsistent in a way that's a bug rather than a deliberate call. SRID was deliberately moved *below* the canonical `OGC:CRS84`/`EPSG:4326` check (there's a comment explaining that `EPSG:4326 + crs_type=srid` should still canonicalize), but projjson and wkt2:2019 fire *above* it. So `{"crs":"EPSG:4326","crs_type":"projjson"}` and `{"crs":"EPSG:4326","crs_type":"wkt2:2019"}` now error — which is why the `geometry_epsg_4326_incorrect_type` test had to go. A canonical authority code is unambiguous regardless of a bogus annotation, so I'd lift the canonical check above the whole switch and let projjson/wkt2:2019 reject only non-canonical strings. No conforming producer emits that combination, so this is a correctness/consistency fix rather than an interop break.

Second thing: the WKT2 heuristic is missing a few real WKT2:2019 keywords (`COORDINATEMETADATA[`, `COORDINATEOPERATION[`, `CRS[`), and since it's a prefix match a short comment that it's intentionally best-effort would help.

Things worth settling in this PR before merge:
- move the canonical check above the `crs_type` switch so canonical wins for all annotations, not just SRID
- restore `geometry_epsg_4326_incorrect_type` to its canonicalizing assertion and add the `crs_type=wkt2:2019` sibling case
- add the missing WKT2:2019 keywords + a passthrough-policy comment
- add a `default:` arm to the `crs_type` switch (exhaustive linter, and future-proofing)
- switch the internal-test sentinel assertions to `ErrorIs`

table/arrow_utils.go
Revanth14: Agreed, that inconsistency is fixed in the latest revision.

The canonical `OGC:CRS84` / `EPSG:4326` check now runs before the whole `crs_type` switch, so canonical strings win for `srid`, `projjson`, and `wkt2:2019` annotations alike. Non-canonical `projjson` / WKT2 strings still reject as unsupported.

I also restored the `geometry_epsg_4326_incorrect_type` canonicalizing case and added the `crs_type:"wkt2:2019"` sibling.

table/arrow_utils.go
Revanth14: Done, The canonical block is now immediately after the empty-string guard and before the `crs_type` switch, so it short-circuits for all annotations. I also removed the SRID special-casing comment since it’s no longer needed.

table/arrow_utils.go
Revanth14: Done. I added `COORDINATEMETADATA[`, `COORDINATEOPERATION[`, and `CRS[` to the WKT2 prefix list, and added coverage for those forms in `TestGeoArrowCRSToIcebergCRS`.

The helper comment now also calls out that this is intentionally best-effort and only used to return the clearer WKT2 unsupported error.

table/arrow_utils.go
Revanth14: Done. I added an explicit `default:` arm with a comment that unset, `AuthorityCode`, `SRID`, and future `CRSType` values intentionally continue to the string-shape checks below.

table/arrow_utils.go
Revanth14: Done. I added a short comment above the passthrough return explaining that bare or unrecognized string CRS values are treated as opaque identifiers per the Iceberg type contract and preserved for read/write compatibility.

table/arrow_utils_internal_test.go
Revanth14: Done. The internal tests now use `assert.ErrorIs` for both package-level sentinels: `errPROJJSONStringCRSNotSupported` and `errWKT2CRSNotSupported`. The external tests still use `ErrorContains` where the unexported sentinels are not visible.

table/arrow_utils_internal_test.go
Revanth14: I removed the defensive `require.NoError` from `stringCRS` and now ignore the marshal error with a short `nolint` comment, since marshaling a string cannot fail.

table/arrow_utils_test.go
Revanth14: Done. I restored `geometry_epsg_4326_incorrect_type` with the canonical `OGC:CRS84` expectation and added the sibling `{"crs":"EPSG:4326","crs_type":"wkt2:2019"}` case to pin the same canonical-first behavior.

laskoviymishka (approved):
:shipit:

PR #1241, opened 2026-06-21
fix(table): abort parquet data file writes after batch write failure

fallintoplace:
Summary

This changes the Parquet `WriteDataFile` error path to abort the open writer when a batch write fails, rather than calling `Close` and trying to finalize the file. If abort cleanup also fails, the returned error now includes both the original write error and the abort error.

The regression test uses a fake file writer to verify the write-failure path calls `Abort`, does not call `Close`, preserves the raw write error when abort succeeds, and joins both errors when abort also fails.

Testing

- `git diff --check`
- `go test ./table/internal -run '^TestWriteDataFileBatchesAbortsOnWriteError$' -count=1`
- `go test ./table/internal -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m ./table/internal/...`
- `go test ./... -count=1`

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -312,9 +312,15 @@ func (p parquetFormat) WriteDataFile(ctx context.Context, fs iceio.WriteFileIO, 
 		return nil, err
 	}
 
+	return writeDataFileBatches(w, batches)
+}
+
+func writeDataFileBatches(w FileWriter, batches []arrow.RecordBatch) (iceberg.DataFile, error) {
 	for _, batch := range batches {
 		if err := w.Write(batch); err != nil {
-			w.Close()
+			if abortErr := w.Abort(); abortErr != nil {
+				return nil, errors.Join(err, abortErr)
+			}
 
 			return nil, err
 		}
--- a/table/internal/parquet_files_internal_test.go
+++ b/table/internal/parquet_files_internal_test.go
@@ -1,0 +1,90 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package internal
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/iceberg-go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type failingWriteFileWriter struct {
+	writeErr error
+	abortErr error
+
+	abortCalled bool
+	closeCalled bool
+}
+
+func (w *failingWriteFileWriter) Write(arrow.RecordBatch) error {
+	return w.writeErr
+}
+
+func (*failingWriteFileWriter) BytesWritten() int64 {
+	return 0
+}
+
+func (w *failingWriteFileWriter) Close() (iceberg.DataFile, error) {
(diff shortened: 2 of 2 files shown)


table/internal/parquet_files_internal_test.go
```diff
@@ -0,0 +1,73 @@
...
+	return w.writeErr
+}
+
+func (*failingWriteFileWriter) BytesWritten() int64 {
+	return 0
+}
+
+func (w *failingWriteFileWriter) Close() (iceberg.DataFile, error) {
+	w.closeCalled = true
+	panic("Close should not be called after write failure")
```
laskoviymishka: This is the one I'd cover before merge. The panic here crashes the whole test binary if `Close` is ever called — the rest of the package's tests don't run, and you get a stack trace instead of a readable failure. The `closeCalled` set just above it is unreachable for assertion purposes, and the `assert.False(t, writer.closeCalled)` at the call site never runs if the panic fires.

Since we already track `closeCalled`, I'd drop the panic and let `assert.False(closeCalled)` do the work — it's the idiomatic way to say "this must not be called."

table/internal/parquet_files.go
```diff
@@ -312,11 +312,13 @@ func (p parquetFormat) WriteDataFile(ctx context.Context, fs iceio.WriteFileIO,
...
+	return writeDataFileBatches(w, batches)
+}
+
+func writeDataFileBatches(w FileWriter, batches []arrow.RecordBatch) (iceberg.DataFile, error) {
 	for _, batch := range batches {
 		if err := w.Write(batch); err != nil {
-			w.Close()
-
-			return nil, err
+			return nil, errors.Join(err, w.Abort())
```
laskoviymishka: Minor, take or leave. On the common path — write fails, abort succeeds — `errors.Join(err, nil)` doesn't return `err`, it wraps it in a `*joinError` with a different concrete type (and a trailing newline in `.Error()`). `errors.Is` still finds the write error so anything matching on it is fine, but a `==` sentinel check or type assertion on the returned error silently stops matching, and that's the expected case here, not an edge one.

I'd keep the raw error when abort succeeds:

```go
if abortErr := w.Abort(); abortErr != nil {
    return nil, errors.Join(err, abortErr)
}
return nil, err
```

table/internal/parquet_files_internal_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+func TestWriteDataFileBatchesAbortsOnWriteError(t *testing.T) {
+	writeErr := errors.New("write failed")
+	abortErr := errors.New("abort failed")
+	writer := &failingWriteFileWriter{
+		writeErr: writeErr,
+		abortErr: abortErr,
+	}
+
+	df, err := writeDataFileBatches(writer, []arrow.RecordBatch{nil})
```
laskoviymishka: Nit. This only covers the both-errors case. The common path — write fails, abort succeeds (`abortErr == nil`) — is the one that exercises the type-wrapping I flagged in `parquet_files.go`, and it's untested. I'd make this table-driven with two subcases (abort succeeds / abort fails) and assert in the abort-succeeds case that the returned error is the raw `writeErr`, not a join.

While you're here: `[]arrow.RecordBatch{nil}` works only because the fake's `Write` ignores its arg — a one-line comment saying so would save the next reader from thinking this validates nil-batch handling.

laskoviymishka (changes requested):
Nice, this is the right fix. Routing the error path through `Abort()` instead of `Close()` avoids writing a footer on a half-finished file and sidesteps a likely panic in `pqarrow.FileWriter.Close()` on partial state, and the whitebox helper plus fake is a clean way to pin the control-flow choice.

One thing I'd like to cover before merge: the fake panics on `Close`, which crashes the whole package test binary if `Close` ever gets called - you get a stack trace instead of a readable failure and the rest of the package's tests don't run. We already track `closeCalled`, so `assert.False` can carry that assertion cleanly. Details inline.

The rest are nits, take or leave. On the common path `errors.Join(err, nil)` wraps the write error in a `*joinError` instead of returning it raw (`errors.Is` still works), and the test only exercises the both-fail case. I'd add the abort-succeeds path. Both inline.

One for a follow-up, not this PR: `Abort()` closes the file handle but doesn't delete it, so for blob backends an aborted write leaves an orphaned footer-less object in the store. It's pre-existing (same call in `rolling_data_writer.go`) and invisible to readers since the snapshot never commits, but it's worth a tracking issue to mirror Java's `BaseTaskWriter.abort()` and actually remove the file.

laskoviymishka (approved):
Thanks for that!

:shipit:

PR #1242, opened 2026-06-21
fix(table): equality delete null key encoding

fallintoplace:
## Summary

This makes the equality-delete fast-path encoders write the same null tag as the generic encoder before reading raw Arrow values. Without that guard, nullable numeric/date/time/timestamp keys could be encoded as non-null zero or epoch-like values.

The tests cover the raw fast-path encoders directly and scan-level equality deletes where a null delete key must remove only null rows, not non-null zero or epoch rows.

## Testing

- `go test ./table -run 'TestEqualityDeleteNullable|TestMakeColEncoderMatchesGenericForNullFastPathTypes' -count=1`
- `git diff --check`
- `go test ./table -count=1`
- `go test ./... -count=1`


table/equality_delete_reader.go
```diff
@@ -361,76 +371,120 @@ func makeColEncoder(arr arrow.Array) colEncoder {
 		vals := a.Int8Values()
 
 		return func(buf *bytes.Buffer, row int) {
+			if writeNullTagIfNull(buf, a, row) {
```
laskoviymishka: This is the main thing I'd like to settle before the rest. This two-line guard is now copy-pasted into all 11 numeric closures, and there's nothing structural stopping the next fast-path someone adds from forgetting it — which is exactly how this bug got in.

I'd lift it to a wrapper applied at the `makeColEncoder` return site:

```go
func withNullGuard(arr arrow.Array, enc colEncoder) colEncoder {
	return func(buf *bytes.Buffer, row int) {
		if writeNullTagIfNull(buf, arr, row) {
			return
		}
		enc(buf, row)
	}
}
```

Then each branch becomes `return withNullGuard(a, func(buf, row) { buf.WriteByte(1); ... })` — one guard to review instead of eleven, and the string/binary/boolean arms could fold into it too so we stop carrying three null-guard idioms in one function. If the extra indirection has measurable hot-path cost I'd keep the inline form but leave a comment marking it load-bearing. wdyt?

table/equality_delete_reader_test.go
```diff
@@ -280,3 +280,226 @@ func TestEqualityDeleteMultiColumnKey(t *testing.T) {
...
+	}{
+		{
+			name:    "int",
+			keyType: iceberg.PrimitiveTypes.Int32,
+			dataJSON: `[
+				{"row_id": 1, "key": null},
+				{"row_id": 2, "key": 0},
+				{"row_id": 3, "key": 7}
+			]`,
+			deleteJSON: `[{"key": null}]`,
```
laskoviymishka: Every case here uses `deleteJSON: [{"key": null}]`, so we only ever prove a null delete key matches null rows — the easy half. The half with the data-loss teeth is the converse: a non-null delete key, especially `0`, must *not* delete a null-key row.

Pre-fix, a null row encoded as `0x01` + a zero payload, byte-identical to a real key of `0`. So `deleteJSON: [{"key": 0}]` against this exact data (row 1 null, row 2 = 0) would have deleted *both* rows, and should now delete only row 2. That case catches the bug from the dangerous direction and guards against a future regression where the null tag collides with a zero value. Could we add a subcase (per type, or a representative few) with `deleteJSON: [{"key": 0}]` and the null row expected to survive?

table/equality_delete_reader_internal_test.go
```diff
@@ -0,0 +1,190 @@
...
+		{name: "float64", arr: buildNullableFloat64Array(mem)},
+		{name: "date32", arr: buildNullableDate32Array(mem)},
+		{name: "date64", arr: buildNullableDate64Array(mem)},
+		{name: "time32", arr: buildNullableTime32Array(mem)},
+		{name: "time64", arr: buildNullableTime64Array(mem)},
+		{name: "timestamp", arr: buildNullableTimestampArray(mem)},
+	}
+
+	for _, tt := range tests {
+		defer tt.arr.Release()
```
laskoviymishka: `defer tt.arr.Release()` inside the range loop stacks all 11 releases on the parent function's defer list rather than running per-iteration — they all fire when the test returns, after `mem.AssertSize`. It's correct today because the subtests are synchronous, but it reads as if it's scoped per-iteration when it isn't, and the day someone adds `t.Parallel()` in the `t.Run` closure the parent can return while a subtest still holds a live ref. I'd move it inside the `t.Run` closure so the lifetime matches the scope.

table/equality_delete_reader_internal_test.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+		})
+	}
+}
+
+func buildNullableInt8Array(mem memory.Allocator) arrow.Array {
+	builder := array.NewInt8Builder(mem)
+	defer builder.Release()
+
+	builder.Append(7)
```
laskoviymishka: Every builder produces the same shape — `[non-null, null, non-null]` with the null fixed at index 1. That leaves a few edges unexercised: null at index 0, null at `Len-1`, an all-null array, and (the one I'd actually want) an all-valid array with `NullN == 0`. The all-valid case is the only thing that proves `writeNullTagIfNull` is a true no-op when there's no null bitmap. A 4-element variant with nulls at the ends plus one all-valid subcase would round this out.

table/equality_delete_reader_test.go
```diff
@@ -280,3 +280,226 @@ func TestEqualityDeleteMultiColumnKey(t *testing.T) {
...
+			require.NoError(t, rd.Commit(t.Context()))
+			tbl, err = tx2.Commit(t.Context())
+			require.NoError(t, err)
+
+			assert.Equal(t, []int64{2, 3}, collectRowIDs(t, tbl))
+		})
+	}
+}
+
+func TestEqualityDeleteNullableCompositeFastPathKey(t *testing.T) {
```
laskoviymishka: This composite case only covers `{id: 7, deleted_at: null}` — one non-null field, one null. The concat case I'd want is the all-null key `{id: null, deleted_at: null}` producing `[0x00][0x00]`, which is what proves key concatenation handles a fully-null composite. Worth a subcase where both fields are null and only the matching all-null row is deleted.

table/equality_delete_reader_test.go
```diff
@@ -280,3 +280,226 @@ func TestEqualityDeleteMultiColumnKey(t *testing.T) {
...
+	require.NoError(t, err)
+
+	var ids []int64
+	for rec, err := range itr {
+		require.NoError(t, err)
+
+		indices := rec.Schema().FieldIndices("row_id")
+		require.NotEmpty(t, indices)
+
+		col := rec.Column(indices[0]).(*array.Int64)
```
laskoviymishka: Bare type assertion will panic with an unhelpful message if the column ever comes back as something other than `Int64`. Safe given the hardcoded schema, but `require.IsType` or the two-value form keeps the failure readable. Minor companion just below: the loop uses `for i := 0; i < col.Len(); i++` — `for i := range col.Len()` is available on Go 1.25. Both very minor.

laskoviymishka (changes requested):
Nice catch, and the fix itself is correct — `writeNullTagIfNull` plus the guard at the top of each fast-path closure makes the numeric/temporal encoders consult the null bitmap before reading `vals[row]`, so a null key now encodes as the `0x00` tag instead of `0x01` + a zero/epoch payload. That lines up with `encodeArrowValue` and with Java's `Comparators.nullsFirst()` both-null-equal semantics, and since the same encoder builds both the delete-key set and the data-row keys, the fix closes the mismatch on both sides.

The main thing I'd like to settle before going deeper is structural, not the fix itself. We've now got the same two-line guard copy-pasted into all 11 numeric closures, and three different null-guard idioms living in one function — the inline `a.IsNull` on string/binary/boolean, the new `writeNullTagIfNull` on the numerics, and the default delegation. Nothing stops the next fast-path branch someone adds from silently forgetting the guard, which is exactly how this bug got in. I left an inline suggestion to lift it to a `withNullGuard` wrapper applied at the `makeColEncoder` return site — one guard to review instead of eleven. I'd like to agree on that direction first, because it reshapes where the guard lives and a couple of the test notes below get easier to place once we know the final structure.

Once we've settled the structure, the one coverage gap I'd still want regardless is the converse case: every test here drives a null delete key against null rows, but none prove that a non-null/zero delete key leaves a null row alone — and that's the direction with the data-loss teeth. The rest are small test cleanups (per-iteration `defer`, a couple of edge shapes, an all-null composite key, a type-assertion nit) that we can fold in cheaply once the structure's locked.

I'll hold off approving until we've talked through the wrapper — happy to take another pass right after.

table/equality_delete_reader_test.go
```diff
@@ -280,3 +280,270 @@ func TestEqualityDeleteMultiColumnKey(t *testing.T) {
...
+		{
+			name:    "long",
+			keyType: iceberg.PrimitiveTypes.Int64,
+			dataJSON: `[
+				{"row_id": 1, "key": null},
+				{"row_id": 2, "key": 0},
+				{"row_id": 3, "key": 7}
+			]`,
+			nullDeleteJSON: `[{"key": null}]`,
+			zeroDeleteJSON: `[{"key": 0}]`,
```
laskoviymishka, line 312: I'd drop the `float` and `double` sub-cases here.

The spec says float and double columns can't be used as equality delete columns, so these set `EqualityFieldIDs` on a key type that Java/PyIceberg/Rust would refuse to write — the round-trip we're asserting can't actually occur, and it quietly implies we support float-keyed deletes. The encoder-level test (`TestMakeColEncoderMatchesGenericForNullFastPathTypes`) is the right place to exercise float/double, since it checks the encoding layer in isolation without claiming they're valid keys — happy to keep them there.

Separately, there's no validation rejecting float/double equality field IDs at write/commit time — pre-existing, worth a follow-up, not this PR.

laskoviymishka (changes requested):
Almost there, one blocker to merge: the scan-level test uses `float`/`double` as equality-delete key columns, which the spec disallows in `equality_ids`. Details in the inline note.

Everything else is in good shape — the core fix is symmetric across the delete-key and data-row paths, the string/binary/bool/fixed cases were already guarded, and the null-matches-null semantics line up with Java's nulls-first comparators. Drop the float/double scan cases and this is good to land.

laskoviymishka (approved):
nice one!

fallintoplace:
I will fix the merge conflict later today.

zeroshade (approved):
LGTM. The earlier blocking compile error in `equality_delete_reader_internal_test.go` is fixed - the imports are back in the import block and `buildTimestampArray` is properly closed (`go test ./table` passes). The `withNullGuard` / `writeNullTagIfNull` change correctly restores null-tag symmetry so null equality-delete keys no longer collide with zero/epoch values: fast-path raw reads are guarded, and all-valid arrays skip the wrapper via `NullN() == 0`. The encoder-symmetry internal tests plus the scan-level null-vs-zero/epoch tests cover the high-risk cases.

Optional, non-blocking: the scan-level tests exercise int/long/date/timestamp but not every type the internal encoder-symmetry tests cover (e.g. float/time/date64) - a couple more scan-level cases would round things out, though the internal tests already pin the encoder behavior. Nice fix.

PR #1244, opened 2026-06-22
fix(catalog/sql): migrate legacy V0 iceberg_tables schema and tolerate null iceberg_type

### LINKED ISSUE #1188: fix(catalog): catalog/sql cannot read iceberg tables previously created by jdbc catalog at default schema-version (V0)
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Iceberg-Go's SQL catalog assumes that iceberg table always has the `V1` schema column `iceberg_type` (cf. `JdbcUtil.RECORD_TYPE`).

When the same postgres database has already been initialised by jdbc catalog at its default schema-version (V0), the `iceberg_type` column does not exist. Every read query bun emits via the `sqlIcebergTable` model fails immediately.
 
**Proposed Solution**

On catalog init, run `ALTER TABLE iceberg_tables ADD COLUMN iceberg_type VARCHAR(5)` if the column is missing, and change every iceberg-go SQL query that filters on iceberg_type to accept the legacy NULL form.

**Contribution**

I’d be happy to work on this issue. Please feel free to assign this issue to me and I can prepare a PR implementing it.

badalprasadsingh:
## Description

Resolves the case where the same database has already been initialised by the JDBC catalog at its default V0 schema-version, which omits the `iceberg_type` column added in V1. Iceberg-Go currently assumes V1 unconditionally, so every query fails immediately with `no such column: iceberg_type`.

Fixes: #1188 

## Solution
- **migrateVOSchema** step after `CreateSQLTables` If `iceberg_type` is missing, it issues `ALTER TABLE iceberg_tables ADD COLUMN iceberg_type VARCHAR(5)` (nullable, so legacy rows keep `NULL`), along with `icebergTypeColumnExists` check
- Every table-scoped query that filtered on `iceberg_type = 'TABLE'` now uses `(iceberg_type = ? OR iceberg_type IS NULL)` so legacy rows continue to load, list, commit, drop, and rename. View queries are deliberately untouched.

## Testing
- Tested against `JDBC` catalog with `postgres` as db


badalprasadsingh:
@laskoviymishka, please take a look, whenever you are free.

catalog/sql/sql.go
```diff
@@ -247,7 +247,76 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+	if probeErr != nil {
+		log.Printf("WARNING: skipping V0 schema migration; iceberg_type column probe failed: %v", probeErr)
+
+		return nil
+	}
+	if hasCol {
+		return nil
+	}
+
+	if _, err := c.db.ExecContext(ctx, "ALTER TABLE iceberg_tables ADD COLUMN iceberg_type VARCHAR(5)"); err != nil {
```
laskoviymishka: This ALTER is hardcoded, but `ADD COLUMN` isn't valid on every dialect the probe handles. SQL Server wants `ALTER TABLE iceberg_tables ADD iceberg_type VARCHAR(5)` (no `COLUMN`) and Oracle wants `ADD (iceberg_type VARCHAR2(5))`. Since the probe already has `mssql` and `oracle` cases, a V0 catalog on either engine reaches this line, the ALTER fails, the re-probe still finds the column missing, and `migrateV0Schema` returns a hard error — so the catalog can't be opened at all. I'd switch on the dialect here the same way the probe does and emit dialect-specific DDL.

Separately, `VARCHAR(5)` diverges from the unbounded `VARCHAR` bun emits for fresh Go-created tables, so migrated and fresh DBs end up with different column types — and it fits `"TABLE"` exactly, which is tight. I'd drop the length bound to match what bun generates.

catalog/sql/sql.go
```diff
@@ -247,7 +247,76 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
-	return c.CreateSQLTables(context.Background())
+	ctx := context.Background()
+	if err := c.CreateSQLTables(ctx); err != nil {
+		return err
+	}
+
+	return c.migrateV0Schema(ctx)
+}
+
+func (c *Catalog) migrateV0Schema(ctx context.Context) error {
```
laskoviymishka: This is the piece I'd most want to settle before merge. The migration runs on every open where `init_catalog_tables` is set (the default), so we issue one-way DDL against what may be a catalog shared with Java or PyIceberg, with no opt-out.

Java made this an explicit choice — `JdbcCatalog` only migrates when `jdbc.schema-version=V1` is set, otherwise it warns and stays on V0. I'd mirror that: take a `jdbc.schema-version` property (same key as Java, for interop), run the ALTER only when it's `V1`, and otherwise lean on the NULL-tolerant read path this PR already adds. There's also no exported constant for that knob the way `init_catalog_tables` has one. wdyt?

catalog/sql/sql.go
```diff
@@ -247,7 +247,76 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+		return fmt.Errorf("migration of V0 schema failed: %w", err)
+	}
+
+	return nil
+}
+
+func (c *Catalog) icebergTypeColumnExists(ctx context.Context) (bool, error) {
+	var query string
+	switch dialectName := c.db.Dialect().Name().String(); dialectName {
+	case "pg":
```
laskoviymishka: The Postgres probe doesn't scope to the current schema, and the `mssql` case below has the same gap. `information_schema.columns` spans every schema the user can see, so if another schema has an `iceberg_tables` that already carries `iceberg_type`, this returns true, we skip the migration, and the first real query against our table fails. Multi-schema Postgres is common, and the JDBC catalog supports a configurable schema. I'd add `AND table_schema = current_schema()` here and `AND table_schema = SCHEMA_NAME()` to the MSSQL case — MySQL already gets this right with `table_schema = DATABASE()`.

catalog/sql/sql.go
```diff
@@ -247,7 +247,76 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+	if err := c.CreateSQLTables(ctx); err != nil {
+		return err
+	}
+
+	return c.migrateV0Schema(ctx)
+}
+
+func (c *Catalog) migrateV0Schema(ctx context.Context) error {
+	hasCol, probeErr := c.icebergTypeColumnExists(ctx)
+	if probeErr != nil {
```
laskoviymishka: Swallowing the probe error and returning nil means `NewCatalog` hands back a catalog that looks fine, then the first table op fails with a cryptic `no such column: iceberg_type` far from the real cause. The `default` case in the probe gets silently no-op'd the same way. I'd return the wrapped error here (`fmt.Errorf("V0 schema migration: column probe failed: %w", probeErr)`) so we fail fast at construction; if the goal is to let read-only users who can't probe still connect, I'd make that an explicit option rather than the silent default.

(Minor, while here: `log.Printf` writes to the default logger and bypasses whatever logging the caller configured — consistent with `PurgeTable` today, but worth keeping in mind.)

catalog/sql/sql.go
```diff
@@ -247,7 +247,76 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+
+		return fmt.Errorf("migration of V0 schema failed: %w", err)
+	}
+
+	return nil
+}
+
+func (c *Catalog) icebergTypeColumnExists(ctx context.Context) (bool, error) {
+	var query string
+	switch dialectName := c.db.Dialect().Name().String(); dialectName {
```
laskoviymishka: Small robustness thing: switching on `c.db.Dialect().Name().String()` compares against bun's internal strings and leaks one back out in the `default` error (e.g. `pg`, not the public `postgres`). bun exports the typed `dialect.Name` constants (`dialect.PG`, `dialect.SQLite`, …) — switching on those instead is sturdier against name drift and lets the `exhaustive` linter actually check this switch.

catalog/sql/sql.go
```diff
@@ -349,11 +418,11 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
 				CatalogName:              c.name,
 				TableNamespace:           strings.Join(ns, "."),
 				TableName:                tblName,
-				IcebergType:              TableType,
+				IcebergType:              sql.NullString{String: TableType, Valid: true},
```
laskoviymishka, line 579: Heads up — because this UPDATE is driven off the model struct, bun puts every non-PK field in the SET clause, so we write `iceberg_type = 'TABLE'` back onto the row on every commit. Java's V1 commit only sets `metadata_location`/`previous_metadata_location` and leaves `iceberg_type` alone (it's WHERE-only there). It's harmless — NULL and `'TABLE'` read identically through the new predicate — but it's an intentional-looking divergence the test asserts as correct, so either a one-line comment saying we opportunistically heal V0 rows on commit, or explicit `Set(...)` calls to match Java.

catalog/sql/sql.go
```diff
@@ -247,7 +247,76 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+	default:
+		return false, fmt.Errorf("unsupported dialect for V0 migration: %s", dialectName)
+	}
+	var dummy int
+	if err := c.db.QueryRowContext(ctx, query).Scan(&dummy); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return false, nil
+		}
+
+		return false, err
```
laskoviymishka, line 440: Bare `err` here — the rest of the file wraps with `fmt.Errorf("...: %w", err)`. Worth wrapping so the caller (and the warning log) gets some context about what failed.

catalog/sql/sql.go
```diff
@@ -247,7 +247,76 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+		query = `SELECT 1 FROM pragma_table_info('iceberg_tables')
+		         WHERE name = 'iceberg_type' LIMIT 1`
+	case "oracle":
+		query = `SELECT 1 FROM user_tab_columns
+ 			    WHERE UPPER(table_name) = 'ICEBERG_TABLES'
+  			  	 AND UPPER(column_name) = 'ICEBERG_TYPE'`
+	default:
+		return false, fmt.Errorf("unsupported dialect for V0 migration: %s", dialectName)
+	}
+	var dummy int
```
laskoviymishka: `dummy` undersells it — the scanned value is a real existence sentinel. `var exists int` reads better, or skip the scan target entirely by probing with `COUNT(*)` into an int and checking `> 0`.

catalog/sql/sql.go
```diff
@@ -247,7 +247,76 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+		           AND column_name = 'iceberg_type' LIMIT 1`
+	case "mssql":
+		query = `SELECT TOP 1 1 FROM information_schema.columns
+		         WHERE table_name = 'iceberg_tables'
+		           AND column_name = 'iceberg_type'`
+	case "sqlite":
+		query = `SELECT 1 FROM pragma_table_info('iceberg_tables')
+		         WHERE name = 'iceberg_type' LIMIT 1`
+	case "oracle":
+		query = `SELECT 1 FROM user_tab_columns
```
laskoviymishka: Nit: this query's indentation mixes tabs and spaces where the other cases use tabs. It's inside a raw string so gofmt won't catch it and the SQL runs fine, but it'll look off in future diffs — worth realigning to match the others.

catalog/sql/sql_test.go
```diff
@@ -322,6 +322,182 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
 	s.confirmTablesExist(sqldb)
 }
 
+func (s *SqliteCatalogTestSuite) TestV0SchemaFromIcebergIsMigrated() {
+	sqldb := s.getDB()
```
laskoviymishka, line 350: The coverage here is all SQLite, which leaves the riskiest paths untested: the `mssql`/`oracle` probe + ALTER (where the `ADD COLUMN` syntax bug lives), the probe-error → skip-migration branch, and the concurrent-ALTER re-probe recovery. The first in particular means CI won't catch a V0 catalog being unopenable on those engines. Even a focused test that injects a failing probe would cover the swallow path.

catalog/sql/sql_test.go
```diff
@@ -322,6 +322,182 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
...
+	var got sql.NullString
+	err = sqldb.QueryRow(
+		`SELECT iceberg_type FROM iceberg_tables WHERE catalog_name = ? AND table_namespace = ? AND table_name = ?`,
+		catName, nsName, tblName,
+	).Scan(&got)
+	s.Require().NoError(err)
+	s.False(got.Valid, "pre-V0 row should keep iceberg_type IS NULL after migration")
+
+	_ = s.loadCatalogForTableCreation()
+}
```
laskoviymishka, line 447: This second `loadCatalogForTableCreation()` looks like it's meant to check the migration is idempotent on re-open, but with no assertion or comment after it, it isn't really pinning anything down. A line of comment plus a re-check (column still present, row still NULL) would make the intent stick.

catalog/sql/sql_test.go
```diff
@@ -322,6 +322,182 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
...
+	err = sqldb.QueryRow(
+		`SELECT COUNT(*) FROM iceberg_tables WHERE catalog_name = ? AND table_namespace = ? AND table_name = ?`,
+		catName, nsName, tblName,
+	).Scan(&remaining)
+	s.Require().NoError(err)
+	s.Equal(0, remaining, "DropTable must delete rows with iceberg_type IS NULL")
+}
+
+func (s *SqliteCatalogTestSuite) TestV0SchemaLoadCommitRenameAcceptNullIcebergType() {
+	sqldb := s.getDB()
```
laskoviymishka, line 566: This exercises a row created at V1 and then manually NULLed — a fine proxy, but not the same as a genuine V0 row where the column was absent until the migration ran; a one-line comment noting the setup depends on migration having already happened would help. Couple of smaller things in here too: the closure hardcodes `"default"` as the catalog name (`cat.Name()` would be less fragile), and `EqualValues(1, n)` could be `Equal(int64(1), n)` for a stricter check.

laskoviymishka (changes requested):
The compatibility approach here is sound — tolerating `iceberg_type IS NULL` on the read paths is exactly what a V0 catalog needs, and widening the field to `sql.NullString` is the right mechanism. The `LoadTable` filter even closes a latent bug where a view row could come back for a table looked up by PK, so the shape is good.

What I'd want settled before this merges is the migration contract. We run `ALTER TABLE` on every catalog open whenever `init_catalog_tables` is on (the default), which means one-way DDL against a catalog that may be shared with Java or PyIceberg clients, with no way to turn it off. Java made this deliberate: `JdbcCatalog` only migrates when `jdbc.schema-version=V1` is set and otherwise stays on V0 with a warning. I'd rather we adopt the same contract than diverge on something this load-bearing — gate the DDL on a `jdbc.schema-version` property and let the NULL-tolerant read path carry the default case.

There's also a hard correctness bug independent of that: the `ADD COLUMN` ALTER is hardcoded with syntax MSSQL and Oracle both reject, and the probe routes those dialects straight into it, so a V0 catalog on either engine can't be opened at all. That one's a clear blocker regardless of how the migration-contract question lands.

Smaller things worth fixing in this pass:
- the Postgres and MSSQL probes don't scope to the current schema, so a same-named table elsewhere can mask the migration
- a probe error is swallowed and hands back a usable-looking catalog that then fails cryptically on the first query — I'd fail fast at open instead
- the suite only exercises SQLite, so neither the MSSQL/Oracle ALTER nor the error paths are covered
- worth a glance: `TestLoadTableInvalidMetadata` now passes for a different reason than before — the new `LoadTable` filter finds its NULL-type row, and the `ErrNoSuchTable` comes from the null `metadata_location` check rather than the row being absent

Details and the line-level nits are inline. Once the migration contract and the MSSQL/Oracle ALTER are sorted I'm happy to take another pass.

badalprasadsingh:
Hey @laskoviymishka, 

1. It now adopts the _Java contract_: the `ALTER` now runs only when `jdbc.schema-version=V1` is set, otherwise we _warn_ and stay on `V0` behind the NULL-tolerant read path.
2. For actual testing of the `MSSQL`/`Oracle` `ALTER`/`probe`, we would be needing those engines in the CI. For now, can we keep it out of scope, and open a new PR for it?
3. Other than that, all the smaller comments addressed: `ADD COLUMN` is replaced with dialect-specific `DDL`, the `PG/MSSQL` probes are scoped to the current schema, and probe errors now fail fast at open. 

Please do have a look whenever you are free.

catalog/sql/sql.go
```diff
@@ -61,8 +62,37 @@ const (
...
+// schemaVer tracks which physical layout of the iceberg_tables table the catalog is operating against.
+//
+// schemaV1 has the iceberg_type column. schemaV0 is the legacy JDBC default that lacks the column entirely.
+// Queries against a V0 catalog must never reference iceberg_type
+//
+// The zero value is schemaV1
+type schemaVer int
+
+const (
+	schemaV1 schemaVer = iota
```
laskoviymishka: I'd flip this so `schemaV0` is the zero value.

Right now any `Catalog` constructed without going through `NewCatalog` — anything that builds `&Catalog{...}` directly — starts in V1 and emits `iceberg_type` predicates against whatever schema it's pointed at. That's backwards from the safe default: a zero-value catalog should never reference a column that might not exist, and Java initializes `schemaVersion` to `V0` explicitly for the same reason.

Swapping the iota order so `schemaV0 = iota` makes the zero value column-free-safe, and V1 is only reached after a probe. The `// The zero value is schemaV1` comment would flip too. wdyt?

catalog/sql/sql.go
```diff
@@ -211,6 +246,10 @@ func NewCatalog(name string, db *sql.DB, dialect SupportedDialect, props iceberg
 		return cat, cat.ensureTablesExist()
 	}
 
+	if err := cat.detectSchemaVersion(context.Background()); err != nil {
```
laskoviymishka: I think the migration is only reachable on one of the two open paths, and I'd want it to always run.

`migrateV0Schema` is called only from `ensureTablesExist`, which `NewCatalog` reaches only when `init_catalog_tables` is true. On the false branch we fall through to `detectSchemaVersion`, which never migrates. So a user who sets `init_catalog_tables=false` (to manage DDL themselves) and `jdbc.schema-version=V1` gets no migration, no ALTER, and no diagnostic — the opt-in is silently consumed. That diverges from Java's `JdbcCatalog.initialize()`, where `updateSchemaIfRequired()` runs unconditionally regardless of `initializeCatalogTables`.

I'd collapse `detectSchemaVersion` and `migrateV0Schema` into one function that always detects and, when opted in, migrates — then gate only `CreateSQLTables` behind `init_catalog_tables`. That also removes the redundancy where both functions probe and both assign `schemaVersion`. wdyt?

catalog/sql/sql.go
```diff
@@ -247,12 +286,141 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+func (c *Catalog) detectSchemaVersion(ctx context.Context) error {
+	hasCol, err := c.icebergTypeColumnExists(ctx)
+	if err != nil {
+		return fmt.Errorf("detecting catalog schema version: %w", err)
+	}
+
+	if hasCol {
+		c.schemaVersion = schemaV1
+	} else {
+		c.schemaVersion = schemaV0
```
laskoviymishka: There's a subtle edge here when the table doesn't exist yet on the init=false path.

If `iceberg_tables` hasn't been created, the probe returns `(false, nil)` and we land on `schemaV0`. If the caller then runs `CreateSQLTables()` (it's public), they get a V1 schema with `iceberg_type`, but the catalog stays in V0 mode for its whole lifetime — `ExcludeColumn` on every query, `iceberg_type` never written to new rows.

I'd distinguish "column absent" from "table absent": if the table doesn't exist, treat it as a fresh V1 catalog rather than legacy V0. Or, if we'd rather keep it simple, document that `init_catalog_tables=false` requires the table to already exist. Either's fine, but the current behavior is a silent trap.

catalog/sql/sql.go
```diff
@@ -247,12 +286,141 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("V0 schema migration: column probe failed: %w", err)
+	}
+	if hasCol {
+		c.schemaVersion = schemaV1
+
+		return nil
+	}
+
+	if !strings.EqualFold(c.props.Get(SchemaVersionKey, ""), SchemaVersionV1) {
```
laskoviymishka, line 328: I'd make this an exact comparison rather than `EqualFold`.

`jdbc.schema-version=v1` (lowercase) would trip the one-way, irreversible ALTER here, and Java parses this case-sensitively via `SchemaVersion.valueOf`, so `v1` wouldn't migrate there. That's a portability gap on a destructive operation. I'd use `!= SchemaVersionV1`. If we genuinely want case-insensitivity, let's document it and add a test that `v1` triggers migration — right now nothing covers that path.

catalog/sql/sql.go
```diff
@@ -247,12 +286,141 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+		log.Printf("WARNING: iceberg_tables is on the legacy V0 schema (no iceberg_type column); "+
```
laskoviymishka: This warning fires on every `NewCatalog` against a V0 db without the opt-in, straight to the global logger.

For apps that reconstruct a catalog per operation or per connection, that's one line per call indefinitely, and callers can't redirect or suppress it. A `sync.Once` to emit it once, or surfacing it through structured logging, would be friendlier. It's consistent with the existing `log.Printf("WARNING: dropped table…")`, so not new ground — but worth a doc note either way.

catalog/sql/sql.go
```diff
@@ -247,12 +286,141 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+
+	c.schemaVersion = schemaV1
+
+	return nil
+}
+
+func addIcebergTypeColumnDDL(name dialect.Name) (string, error) {
+	switch name {
+	case dialect.PG, dialect.SQLite:
+		return "ALTER TABLE iceberg_tables ADD COLUMN iceberg_type VARCHAR", nil
```
laskoviymishka, line 370: Small interop note: Java's `V1_UPDATE_CATALOG_SQL` adds this as `VARCHAR(5)`.

Values are only ever `TABLE`/`VIEW`, so there's no corruption risk — this only shows up in `DESCRIBE`/`information_schema`. I'd either match Java's `VARCHAR(5)` or pick one width uniformly across dialects; the current mix (bare `VARCHAR` for PG/SQLite, `VARCHAR(255)` elsewhere) is a little arbitrary.

catalog/sql/sql.go
```diff
@@ -247,12 +286,141 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+	case dialect.MSSQL:
+		return `SELECT TOP 1 1 FROM information_schema.columns
+		         WHERE table_schema = SCHEMA_NAME()
+		           AND table_name = 'iceberg_tables'
+		           AND column_name = 'iceberg_type'`, nil
+	case dialect.SQLite:
+		return `SELECT 1 FROM pragma_table_info('iceberg_tables')
+		         WHERE name = 'iceberg_type' LIMIT 1`, nil
+	case dialect.Oracle:
+		return `SELECT 1 FROM user_tab_columns
```
laskoviymishka, line 403: Every other dialect's probe caps at one row (`LIMIT 1` / `TOP 1`); the Oracle branch doesn't.

I'd add `AND ROWNUM = 1` here for consistency — harmless on a pure existence check, but it's the odd one out otherwise.

catalog/sql/sql.go
```diff
@@ -247,12 +286,141 @@ func (c *Catalog) DropSQLTables(ctx context.Context) error {
...
+}
+
+func (c *Catalog) icebergTypeColumnExists(ctx context.Context) (bool, error) {
+	query, err := icebergTypeColumnExistsQuery(c.db.Dialect().Name())
+	if err != nil {
+		return false, err
+	}
+
+	var exists int
+	if err := c.db.QueryRowContext(ctx, query).Scan(&exists); err != nil {
```
laskoviymishka: This probe runs directly on the pool, unlike every other read which goes through `withReadTx`.

Harmless on SQLite with `MaxOpenConns(1)`, but on Postgres under serializable it's an inconsistency that's harder to audit. I'd wrap it in a read tx to match, or drop a comment noting the deliberate exception. wdyt?

catalog/sql/sql.go
```diff
@@ -1013,6 +1225,10 @@ func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) err
 
 // CheckViewExists returns true if a view exists in the catalog.
 func (c *Catalog) CheckViewExists(ctx context.Context, identifier table.Identifier) (bool, error) {
+	if c.isV0() {
+		return false, errViewsUnsupportedOnV0
```
laskoviymishka: I'd have `CheckViewExists` return `(false, nil)` on V0 rather than an error.

By signature it's a pure predicate, and the correct answer for a V0 catalog is "no, there are no views" — not a failure. Returning an error breaks symmetry with `CheckTableExists` (which returns `(false, nil)`), and any direct caller now has to special-case it. I'd keep the error gate on the mutation/load paths (`CreateView`/`DropView`/`LoadView`) where refusing is the right call, and let `ListViews` return empty. `CreateView`'s own guard fires first, so it's unaffected either way.

catalog/sql/sql_test.go
```diff
@@ -322,6 +336,358 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
...
+
+	cat := s.getCatalogSqlite()
+	ctx := context.Background()
+
+	// This is a proxy for a genuine V0 row: we create the row at V1 (so the
```
laskoviymishka: This test's own comment flags the gap: it's a proxy, not the real path.

It creates the row at V1 and manually NULLs it, so it exercises the NULL-tolerant WHERE predicate on a schema that physically has `iceberg_type`. It never exercises the `ExcludeColumn` branch against a genuinely column-less V0 schema for load/commit/rename — which is the code path this PR is actually adding. `TestV0SchemaFullLifecycleWithoutOptIn` doesn't close it either, since its commit isn't a real metadata commit.

I'd add one test that starts from a real column-less V0 table with a pre-written metadata file, migrates to V1 (column present, row NULL), and drives `CommitTable`/`RenameTable` end-to-end through that post-migration NULL row. That's the scenario a user hits, and nothing here would fail if the NULL-tolerant predicate regressed. wdyt?

catalog/sql/sql_test.go
```diff
@@ -322,6 +336,358 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
...
+	_, err = sqldb.Exec(`CREATE TABLE "iceberg_namespace_properties" (
+		"catalog_name" VARCHAR NOT NULL,
+		"namespace" VARCHAR NOT NULL,
+		"property_key" VARCHAR NOT NULL,
+		"property_value" VARCHAR,
+		PRIMARY KEY ("catalog_name", "namespace", "property_key"))`)
+	s.Require().NoError(err)
+}
+
+func (s *SqliteCatalogTestSuite) loadCatalogV0NoOptIn() *sqlcat.Catalog {
```
laskoviymishka, line 528: `loadCatalogV0NoOptIn` doesn't set `init_catalog_tables`, so it defaults to true — the false branch of `NewCatalog` is never exercised end-to-end.

Combined with the migration-suppression issue above, this is the path I'd most want covered: a helper that sets `"init_catalog_tables":"false"` against an existing column-less schema and confirms V0 is detected (and, once the suppression is fixed, that `jdbc.schema-version=V1` still migrates). The internal unit tests call `detectSchemaVersion` directly, which skips the branch.

catalog/sql/sql_test.go
```diff
@@ -322,6 +336,358 @@ func (s *SqliteCatalogTestSuite) TestCreationAllTablesExist() {
...
+	_, err = cat.CheckViewExists(ctx, viewID)
+	s.ErrorContains(err, "V0", "CheckViewExists must be rejected on a V0 catalog")
+
+	_, err = cat.LoadView(ctx, viewID)
+	s.ErrorContains(err, "V0", "LoadView must be rejected on a V0 catalog")
+
+	err = cat.DropView(ctx, viewID)
+	s.ErrorContains(err, "V0", "DropView must be rejected on a V0 catalog")
+
+	for _, listErr := range cat.ListViews(ctx, []string{nsName}) {
```
laskoviymishka: This loop passes vacuously if `ListViews` ever yields zero items.

If the V0 guard regresses and `ListViews` returns empty instead of erroring, the body never runs and the test still goes green. I'd collect into a slice, `require.Len(..., 1)`, then assert `ErrorContains` on the first element so a missing error actually fails.

laskoviymishka (changes requested):
Nice work on this — the NULL-tolerant predicate is the right way to round-trip rows written by clients that never set `iceberg_type`, and the compat-mode-plus-warning design is a sensible way to avoid forcing a destructive ALTER on people. The test coverage is genuinely thorough.

The one I care most about: the migration is only reachable when `init_catalog_tables=true`, because `migrateV0Schema` is only called from `ensureTablesExist`. On the `init_catalog_tables=false` branch we run `detectSchemaVersion`, which never migrates — so someone who manages their DDL externally and sets `jdbc.schema-version=V1` gets no migration, no ALTER, and no diagnostic. The opt-in is silently consumed. Java's `JdbcCatalog.initialize()` runs `updateSchemaIfRequired()` unconditionally for exactly this reason. I'd collapse the two functions into one that always detects and optionally migrates, and gate only `CreateSQLTables` behind `init_catalog_tables`.

Things I'd want to settle in this PR before merge:
- migration always runs regardless of `init_catalog_tables` (the opt-in path above)
- `schemaV0` as the zero value, not `schemaV1`, so a default-constructed catalog is column-free-safe
- exact comparison instead of `EqualFold`, so a lowercase `v1` can't trip the one-way ALTER
- `CheckViewExists` returning `(false, nil)` on V0 instead of an error
- one test that drives commit/rename through a genuinely migrated column-less V0 row, not the current proxy

The rest are smaller and noted inline. Once those are addressed, happy to take another pass and approve.

catalog/sql/sql.go
zeroshade: [major] With `init_catalog_tables=false`, this path only detects the existing schema and never calls `migrateV0Schema`, so `jdbc.schema-version=V1` does not actually migrate an existing V0 catalog. Please run one detect-and-maybe-migrate on open, and gate only `CreateSQLTables` behind `init_catalog_tables`.

catalog/sql/sql.go
zeroshade: [major] A missing `iceberg_tables` table is collapsed into `schemaV0` because the column probe returns false for both legacy V0 and table-absent catalogs. If a caller opens with `init_catalog_tables=false` and later calls `CreateSQLTables`, the physical schema is V1 while this `Catalog` stays in V0 mode and writes omit `iceberg_type`. Please probe table existence separately and either treat table-absent as fresh V1 or fail/document this path.

catalog/sql/sql.go
zeroshade, line 328: [major] `strings.EqualFold(..., "V1")` lets lowercase `v1` trigger the irreversible ALTER, which diverges from Java's case-sensitive handling. Please compare exactly to `SchemaVersionV1`, or explicitly document and test the intended case-insensitive behavior.

catalog/sql/sql.go
zeroshade: [minor] `CheckViewExists` should return `(false, nil)` on V0 rather than an unsupported-view error, mirroring `CheckTableExists` existence-check semantics. Consider making `ListViews` return empty on V0 as well.

catalog/sql/sql.go
zeroshade: [minor] The zero value of `schemaVer` is currently V1, which is the less safe default if construction or probing is skipped. Please make `schemaV0` the iota zero value and set V1 only after a successful probe or migration.

catalog/sql/sql.go
zeroshade: [minor] The schema probe bypasses `withReadTx`, unlike the other read paths in this catalog. Please wrap this in `withReadTx` as well, or document why the schema-version probe intentionally runs outside the transaction helper.

catalog/sql/sql.go
zeroshade, line 403: [nit] The Oracle column probe should also limit to one row, e.g. add `AND ROWNUM = 1`, to match the other dialect probes' bounded existence checks.

catalog/sql/sql_test.go
zeroshade: [nit] This `ListViews` assertion can pass vacuously if the iterator yields no values. Please assert that the iterator produced the expected error (or collect and assert one error) so the V0 behavior is actually exercised.

zeroshade (changes requested):
Thanks for moving the migration behind jdbc.schema-version=V1; the unconditional-ALTER concern is mostly addressed, but the implementation still has blocking schema-state issues.

Notes:
- The remaining blockers are around open/init flows, table-absence detection, and case-sensitive migration opt-in semantics.
- View behavior on V0 and transaction consistency need small follow-ups.
- Tests are currently SQLite-only, so dialect-specific probes/DDL need broader coverage or stronger unit assertions.

catalog/sql/sql.go
badalprasadsingh: Done.

catalog/sql/sql.go
badalprasadsingh: Done.

Covered end-to-end in `TestV0InitDisabledDetectsV0AndMigratesWithOptIn`.

catalog/sql/sql.go
badalprasadsingh: Done.

Added the unit test for it as well in `TestInitSchemaVersionTableAbsentIsV1`.

catalog/sql/sql.go
badalprasadsingh: Well

I checked, Iceberg-java makes it case-insensititve in case update Schema is required. Please check [here](https://github.com/apache/iceberg/blob/04a35e412d2f37bb03e0cc68d56111be5d8dbb5b/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java#L250). Happy to know in case we want to case-sensitive.

catalog/sql/sql.go
badalprasadsingh: Done.

catalog/sql/sql.go
badalprasadsingh: Well,

The point is the SQL catalogs which are already created in the past by Iceberg-Go, would still have the current mix of `VARCHAR` as do not set `VARCHAR(5)` during initial `V1` schema catalog creation. Happy to open a new PR for this change across everywhere to match Java's `VARCHAR(5)`.

catalog/sql/sql.go
badalprasadsingh: Done.

catalog/sql/sql.go
badalprasadsingh: Done.

You are right, this would make it consistent with the rest of the catalog's read paths now.

catalog/sql/sql.go
badalprasadsingh: Done.

catalog/sql/sql_test.go
badalprasadsingh: Done.

You are right, added the test for it.

catalog/sql/sql_test.go
badalprasadsingh: Done.

catalog/sql/sql.go
badalprasadsingh: Done.

`TestV0InitDisabledDetectsV0AndMigratesWithOptIn` covers it.

catalog/sql/sql.go
badalprasadsingh: Done.

Now, for the case, we consider it as `V1`.

catalog/sql/sql.go
badalprasadsingh: Well

In, Iceberg-java, its case-insensititve. Please check [here](https://github.com/apache/iceberg/blob/04a35e412d2f37bb03e0cc68d56111be5d8dbb5b/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java#L250). Happy to know in case we want to case-sensitive.

catalog/sql/sql.go
badalprasadsingh: Done.

catalog/sql/sql.go
badalprasadsingh: Done.

catalog/sql/sql.go
badalprasadsingh: Done.

catalog/sql/sql.go
badalprasadsingh: Done.

badalprasadsingh:
Hey @laskoviymishka & @zeroshade, thank you again for the re-review!

I've addressed all of the comments mentioned above. Please let me know if there's anything else that needs to be changed.

Just a quick heads up: [Iceberg-Java](https://github.com/apache/iceberg/blob/04a35e412d2f37bb03e0cc68d56111be5d8dbb5b/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java#L250), makes this check case-insensitive, so I've updated the Go implementation to match that behavior for parity. Happy to change it if we're thinking differently.

Thanks.

zeroshade (approved):
LGTM — the schema-state and init-flow blockers from the earlier review are addressed (mode-gated migration, dialect-specific DDL/probes, table-absence handling). One residual to keep in mind: the MSSQL/Oracle/PG paths are only asserted via generated SQL strings in `catalog/sql/migrate_internal_test.go`, not against live engines — fine for this PR, but those remain the riskiest paths to watch.

laskoviymishka (approved):
:shipit:

PR #1245, opened 2026-06-22
fix(table): clean up partial positional deletes on error

fallintoplace:
Summary
- return partially loaded positional-delete chunks when a concurrent delete-file read fails
- keep successfully read chunks on the result channel even if another worker cancels the group
- add checked-allocator regression coverage for partial failure cleanup

Why
If one positional-delete file is loaded and a later delete-file read fails, the scanner callers already try to release any partial result. readAllDeleteFiles was returning nil on error, so those loaded Arrow chunks could be lost instead of released.

Testing
- go test ./table -run 'TestReadAllDeleteFilesReturnsPartialDeletesOnError|TestReleasePerFilePosDeletes' -count=1\n- go test ./table -count=1\n- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.8.0 run --timeout=10m ./table/...

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -108,12 +108,12 @@ func readAllDeleteFiles(ctx context.Context, fs iceio.IO, tasks []FileScanTask, 
 				if deletes == nil {
 					return nil
 				}
-				select {
-				case perFileChan <- deletes:
-					return nil
-				case <-gctx.Done():
-					return gctx.Err()
-				}
+				// Safe even after gctx cancellation: the channel buffer equals
+				// g's concurrency limit, so every in-flight worker can send
+				// without blocking while the for-range below drains to close.
+				perFileChan <- deletes
+
+				return nil
 			})
 		}
 		_ = g.Wait()
@@ -126,7 +126,7 @@ func readAllDeleteFiles(ctx context.Context, fs iceio.IO, tasks []FileScanTask, 
 	}
 
 	if err := g.Wait(); err != nil {
-		return nil, err
+		return deletesPerFile, err
 	}
 
 	return deletesPerFile, nil
@@ -1136,6 +1136,8 @@ func (as *arrowScan) GetRecords(ctx context.Context, tasks []FileScanTask) (*arr
 	// no intermediate position set.
 	dvBitmaps, err := readAllDeletionVectors(ctx, as.fs, tasks, as.concurrency)
 	if err != nil {
+		releasePerFilePosDeletes(deletesPerFile)
+
 		return nil, nil, err
 	}
 
--- a/table/arrow_scanner_test.go
+++ b/table/arrow_scanner_test.go
@@ -18,17 +18,79 @@
 package table
 
 import (
+	"errors"
+	iofs "io/fs"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/compute"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/pqarrow"
+	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+type failAfterGoodCloseFS struct {
+	*iceio.MemFS
+
+	goodPath string
+	badPath  string
+
+	goodClosed     chan struct{}
+	goodClosedOnce sync.Once
+}
+
+func (f *failAfterGoodCloseFS) Open(name string) (iceio.File, error) {
+	if name == f.goodPath {
+		file, err := f.MemFS.Open(name)
+		if err != nil {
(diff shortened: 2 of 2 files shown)


table/arrow_scanner.go
```diff
@@ -108,12 +108,9 @@ func readAllDeleteFiles(ctx context.Context, fs iceio.IO, tasks []FileScanTask,
 				if deletes == nil {
 					return nil
 				}
-				select {
-				case perFileChan <- deletes:
-					return nil
-				case <-gctx.Done():
-					return gctx.Err()
-				}
+				perFileChan <- deletes
```
laskoviymishka, line 114: Dropping the select is the right simplification, but this send is now safe only because the channel capacity equals `g`'s limit — every concurrent worker can send without blocking, even after `gctx` is cancelled, and the for-range below drains until close. That invariant is load-bearing now, not an optimization, and nothing enforces it: if the buffer and `g.SetLimit` ever drift apart this deadlocks. I'd drop a short comment right here so the next person doesn't "fix" one without re-deriving the other.

```go
// Unconditional send is safe: the channel buffer equals g's concurrency
// limit, so every worker can send without blocking even after gctx is
// cancelled, and the for-range below drains until perFileChan is closed.
perFileChan <- deletes
```

table/arrow_scanner_test.go
```diff
@@ -18,18 +18,70 @@
...
 )
 
+type failAfterGoodCloseFS struct {
+	*iceio.MemFS
+
+	goodPath string
+	badPath  string
+
+	goodClosed      chan struct{}
+	closeGoodClosed sync.Once
```
laskoviymishka: `closeGoodClosed` parses as an adjective phrase rather than the `sync.Once` guarding `close(goodClosed)`. I'd rename it `goodClosedOnce` to mirror the channel name — reads as "the Once for goodClosed" at the call site.

table/arrow_scanner_test.go
```diff
@@ -142,6 +223,41 @@ func chunkedPosDelete(t *testing.T, mem memory.Allocator, positions []int64) *ar
...
+		goodClosed: make(chan struct{}),
+	}
+	tasks := []FileScanTask{{
+		DeleteFiles: []iceberg.DataFile{
+			newPosDeleteFile(t, goodDeletePath, 2, 128),
+			newPosDeleteFile(t, badDeletePath, 1, 128),
+		},
+	}}
+
+	deletesPerFile, err := readAllDeleteFiles(ctx, testFS, tasks, 2)
```
laskoviymishka, line 263: With concurrency=2 both files run in the same wave, so the good worker finishes and sends before the bad one fails — this lands in the success-then-return-partial path but never the cancel-while-between-send window the description calls out, which means it'd pass against the old code too and doesn't isolate that race. The `return deletesPerFile, err` half is what this actually pins down (without it `deletesPerFile` comes back nil and `require.NotNil` fails), so the leak side is covered — I'd just add a line noting concurrency=2 is enough to observe the partial result and you're deliberately not asserting the cancellation race.

While we're here, `require.NotEmpty` passes on any non-empty chunk — the good file has 2 rows (pos 1, 3), so `require.Len(t, deletesPerFile[dataPath], 1)` (and ideally checking the two positions) would catch a regression that accumulates only a partial chunk.

table/arrow_scanner_test.go
```diff
@@ -18,18 +18,70 @@
...
+		return &closeSignalFile{
+			File: file,
+			onClose: func() {
+				f.closeGoodClosed.Do(func() { close(f.goodClosed) })
+			},
+		}, nil
+	}
+
+	if name == f.badPath {
+		<-f.goodClosed
```
laskoviymishka: This blocking receive is the whole point of the helper and it's fine as-is, but it has no escape hatch: if the good file's `Close` ever stops firing (a panic before the defer, or a future refactor that drops it) the test hangs with no diagnostic. A `select` with a `case <-time.After(...)` → `t.Fatal` would turn that into an explicit failure instead of a silent timeout. Minor — only if you feel like hardening the helper.

laskoviymishka (changes requested):
Almost there — the fix itself is correct, with one small thing I'd want before merge and one that's your call.

I traced the unconditional `perFileChan <- deletes` and it can't block: the channel buffer equals `g`'s concurrency limit and the drain loop runs the full lifetime, so every in-flight worker can always send, even after `gctx` is cancelled. And `return deletesPerFile, err` is what finally lets the callers release the partial chunks — the `releasePerFilePosDeletes` guards at both call sites were vacuous before this. The `failAfterGoodCloseFS` ordering harness is a clean, deterministic way to land in the partial-failure state, and `AssertSize(0)` proves the leak is gone.

The blocking ask is small: that send now depends entirely on `buffer == concurrency`, and nothing comments or enforces it — a refactor that changes one and not the other deadlocks. One comment on the send line covers it (left inline).

Separately, and this one's your call since it's pre-existing and outside this diff: inside `GetRecords`, the `readAllDeletionVectors` error path returns without `releasePerFilePosDeletes(deletesPerFile)` — it sits between the `readAllDeleteFiles` release you just fixed and the equality-delete release. Before this PR `deletesPerFile` was empty there so it didn't matter; now it can be fully populated, so the same leak you're closing reopens one step downstream. Since you're already auditing this release contract and adding a comment right above that site, it'd be natural to fold in, but a follow-up is just as reasonable.

Add the invariant comment and decide on the DV-path release, and this is good to land — happy to take another pass.

laskoviymishka (approved):
Nice!

PR #1249, opened 2026-06-22
Truncate projection counts UTF-8 string prefixes in bytes instead of characters

### LINKED ISSUE #1248: Truncate projection counts UTF-8 string prefixes in bytes instead of characters
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

# Description

When projecting `NOT STARTS WITH` predicates through a `truncate[N]` string partition transform, iceberg-go checks whether the filter prefix fits within the truncate width.

That check currently uses `len(StringLiteral)`, which counts bytes. But the string truncate transform itself truncates by Unicode characters/code points, not bytes. For ASCII strings those are the same, but for UTF-8 strings they can differ.

Example:

- partition transform: `truncate[2](name)`
- filter: `name NOT STARTS WITH "éé"`
- `"éé"` is 2 characters, but 4 bytes

The prefix fits within `truncate[2]` in terms of Iceberg string truncate semantics, but the current byte-length check treats it as too long. That can make the projection choose the wrong fallback behavior for multibyte UTF-8 prefixes.

The fix is to count string literals with `utf8.RuneCountInString` while keeping binary/fixed literals byte-counted. This matches the existing truncate implementation, which already truncates strings on UTF-8 code point boundaries.

This is separate from, but adjacent to, the `NOT STARTS WITH` truncate projection bug in #1192/#1193. It was found while reviewing that same area.

# Expected behavior

String truncate projection should compare string prefix length using the same unit as the string truncate transform: Unicode characters/code points.

Binary and fixed truncate projection should continue using byte length.

# Impact

This can affect pruning/projection correctness for tables partitioned by string `truncate[N]` when filters use multibyte UTF-8 prefixes.

# Proposed fix

Use `utf8.RuneCountInString(string(l))` for `StringLiteral` in the truncate projection length helper, and add regression coverage with prefixes such as `"é"`, `"éé"`, and `"ééa"`.

rraulinio:
## Summary

Closes https://github.com/apache/iceberg-go/issues/1248.

Fixes UTF-8 handling in truncate projection length checks for string `NOT STARTS WITH` predicates.

String `truncate[N]` already truncates by Unicode characters/code points, but the projection helper used `len(StringLiteral)`, which counts bytes. For multibyte UTF-8 prefixes like `"éé"`, that made the projection compare a byte length against a character-based truncate width.

This changes string literal length checks to use `utf8.RuneCountInString`, while keeping binary/fixed literals byte-counted. It also adds regression coverage for Unicode prefixes in string truncate projection.

### DIFF
--- a/table/evaluators_test.go
+++ b/table/evaluators_test.go
@@ -1049,6 +1049,9 @@ func (p *ProjectionTestSuite) TestStringTruncateProjection() {
 		{iceberg.NotStartsWith(ref, "a"), iceberg.NotStartsWith(truncStr, "a")},
 		{iceberg.NotStartsWith(ref, "aa"), iceberg.NotEqualTo(truncStr, "aa")},
 		{iceberg.NotStartsWith(ref, "aaa"), iceberg.AlwaysTrue{}},
+		{iceberg.NotStartsWith(ref, "é"), iceberg.NotStartsWith(truncStr, "é")},
+		{iceberg.NotStartsWith(ref, "éé"), iceberg.NotEqualTo(truncStr, "éé")},
+		{iceberg.NotStartsWith(ref, "ééa"), iceberg.AlwaysTrue{}},
 	}
 
 	project := newInclusiveProjection(schema, spec, true)
--- a/transforms.go
+++ b/transforms.go
@@ -27,6 +27,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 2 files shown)


zeroshade (approved):

tanmayrauth (approved):
LGTM. Thanks for the fix.

PR #1250, opened 2026-06-22
fix(v3): re-enable row-group pruning with correct row lineage tracking

twuebi:
Follow-up to #1246 and #1237 

Re-enable row-group pruning while correctly tracking row lineage & add adequate tests.


table/arrow_scanner.go
```diff
@@ -358,38 +355,59 @@ func processPositionalDeletes(ctx context.Context, deletes set[int64]) recProces
...
-		sliced := array.NewSlice(full, currentIdx, nextIdx).(*array.Boolean)
-		defer sliced.Release()
+		bldr := array.NewBooleanBuilder(mem)
+		defer bldr.Release()
+		bldr.Reserve(int(nrows))
+		for range nrows {
+			pos := cursor.next()
+			// Test keep bit at absolute position pos: byte pos/8, bit pos%8,
+			// LSB-first (the layout the fast path's array.NewBoolean reads).
+			bldr.Append(keepBits[pos>>3]&(1<<(uint(pos)&7)) != 0)
```
laskoviymishka, line 405: This is the one thing I'd want settled before merge.

`keepBits` is sized `⌈rowCount/8⌉` from the manifest's `File.Count()`, but `pos` here comes from `cursor.next()` walking `RowGroupSpan`s drawn from Parquet row-group metadata. If those two ever disagree — a writer bug, a hand-built file, corruption — `pos` can reach or exceed `rowCount` and `keepBits[pos>>3]` is an out-of-range index that panics. The fast path is immune because `array.NewSlice` validates; this slow path has no guard.

It also contradicts the `next()` contract a few lines up ("reading past the total row count degrades... rather than panicking") — that only holds for callers that don't index a bounded slice, and this one does. I'd add the guard and keep-by-default:

```go
if pos >= rowCount {
    bldr.Append(true)
    continue
}
```

If we'd rather treat an overshoot as a true invariant violation, the alternative is to panic here with a message that names the position overflow — but a silent OOB is the worst of both. wdyt?

laskoviymishka (changes requested):
Almost there, the cursor design is the right shape and the regression coverage is solid. One blocking thing before merge:

In `filterByDeletionVector`'s slow path we index `keepBits[pos>>3]` with a `pos` from `cursor.next()` (sourced from Parquet row-group metadata), while `keepBits` is sized from the manifest's `File.Count()`. There's no bounds check, so if the two ever disagree — a writer bug, a hand-built file, corruption — that's an out-of-range panic, not a wrong-but-bounded result. And it contradicts the `next()` doc that promises overshoot "degrades... rather than panicking": that only holds for callers that don't index a bounded buffer, and this one does. I'd add a `pos >= rowCount` keep-by-default guard, or panic with a message that names the overflow — a silent OOB is the worst option.

The rest landed well:
- `rowPositionSource`/`rowPositionCursor` walking surviving spans in file order keeps `_row_id = first_row_id + position` correct across pruned-group gaps
- the one-source-many-cursors split reads cleanly, and the DV-supersedes-positional-delete rule matches the v3 spec
- `row_lineage_prune_delete_test.go` exercises the real pruning/delete interaction it's meant to guard
- scan-time only — no file-format or cross-client interop change

A couple of non-blocking nits I'll leave to you: the `positions` slice in `synthesizeRowLineageColumns` is allocated/filled every batch even on the all-explicit-`_row_id` path where it's never read, and the `Survivors` doc says "left empty" where the code sets it to nil. Neither blocks.

Fix the slow-path bound (and reconcile the `next()` degrade contract) and this is good to land.

table/arrow_scanner.go
twuebi: added the guard, thanks!

laskoviymishka (approved):

PR #1252, opened 2026-06-22
feat(transform): deprecate Transform.ToHumanStr(any) in favor of type-aware ToHumanStrType

### LINKED ISSUE #1251: Deprecate Transform.ToHumanStr(any) in favor of type-aware ToHumanStrType(Type, any)
## Feature Request / Improvement

This is a follow-up on the [discussion](https://github.com/apache/iceberg-go/pull/1227#issuecomment-4746198642) at PR #1227.

`iceberg.Transform.ToHumanStr(any) string` is value-only. Because Iceberg's `TimestampType` and `TimestampTzType` both surface as `iceberg.Timestamp` in Go, the source type cannot be recovered at the call site, so `IdentityTransform` cannot render `TimestampTz` partition values with the `+00:00` suffix.

A previous PR ( fixing #1226 ) worked around this by introducing an unexported `typedHumanStringer` extension interface, type-asserting it inside `PartitionToPath`, and falling back to `ToHumanStr` otherwise. That works but leaves a private _shim_ in place and keeps the public API **misleading**.

## Proposal

Mirror `Iceberg-Java`'s deprecation path (`Transform.toHumanString(T)` → `Transform.toHumanString(Type, T)`).

## Compatibility

This is a **breaking change** for external implementers of the public `Transform` interface — they will need to add `ToHumanStrType`. All in-tree implementations are updated.

badalprasadsingh:
## Description

Closes #1251 

In iceberg-go, the public Transform.ToHumanStr(any) string cannot distinguish TimestampType from TimestampTzType (both are iceberg.Timestamp in Go), so formatting that depends on source type cannot be recovered at the call site.

Following up on the [discussion](https://github.com/apache/iceberg-go/pull/1227#issuecomment-4746198642) at PR #1227.

## Changelog Note
This is a **breaking change** for external implementers of the public `Transform` interface; all in-tree implementations are updated. The deprecated `ToHumanStr` remains for the deprecation window.

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -552,12 +552,7 @@ func (ps *PartitionSpec) PartitionToPath(data StructLike, sc *Schema) string {
 		sb.WriteByte('=')
 
 		// Only escape the value (which changes per row)
-		var valueStr string
-		if t, ok := ps.fields[i].Transform.(typedHumanStringer); ok {
-			valueStr = t.ToHumanStrType(partType.FieldList[i].Type, data.Get(i))
-		} else {
-			valueStr = ps.fields[i].Transform.ToHumanStr(data.Get(i))
-		}
+		valueStr := ps.fields[i].Transform.ToHumanStrType(partType.FieldList[i].Type, data.Get(i))
 		sb.WriteString(url.QueryEscape(valueStr))
 	}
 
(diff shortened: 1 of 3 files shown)


laskoviymishka (approved):
🚢 👍

PR #1255, opened 2026-06-22
fix(table): read non-dictionary file_path columns in position deletes

laskoviymishka:
readDeletes assumed a position delete file's file_path column was always dictionary-encoded with a string value type and cast it straight to *array.Dictionary / *array.String. Files that store file_path as a plain or large string column (PyIceberg, arrow-rs, or any writer with dictionary encoding turned off) made that cast panic on the scan path.

Parse the column by layout instead. Plain string, large string, and dictionary variants of both are supported; any other type is rejected with ErrInvalidSchema. Distinct file paths are gathered in Go rather than through compute.Unique, whose kernel concatenated chunked input into a scratch array that was never released.

Build each per-path position list with an explicit mask release after the filter, a checked assertion on the filter result, and chunk cleanup on every error return.

Tests cover the four string layouts, distinct per-chunk dictionaries, null and unsupported-layout rejection, and the RowDelta round-trip now writes file_path without dictionary encoding. The unit tests run under a checked allocator to pin reference counting.

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -30,6 +30,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/compute"
 	"github.com/apache/arrow-go/v18/arrow/compute/exprs"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/arrow/scalar"
 	"github.com/apache/iceberg-go"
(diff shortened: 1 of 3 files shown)


zeroshade:
Isn't this a duplicate of #1229 ?

laskoviymishka:
yep, it was stashed for some time before i pushed it :D, i drafted this quite some time ago, https://github.com/apache/iceberg-go/pull/1229 is better here

PR #1256, opened 2026-06-22
fix(table): reject read-only file io in write paths

fallintoplace:
Summary

- add a shared WriteFileIO requirement helper for table write paths
- replace unchecked transaction WriteFileIO assertions with normal errors
- keep ErrWriteIORequired compatible with the older ErrNotImplemented match from WriteRecords
- run the write-IO guard before automatic name-mapping mutations
- reuse the copy-on-write deletion WriteFileIO in Overwrite instead of resolving the FS twice
- add regression coverage for append, overwrite, copy-on-write delete, merge-on-read delete, row delta, AddDataFiles, AddFiles, replacement add-only paths, WriteEqualityDeletes, and WriteRecords

Why

Custom FileIO implementations can satisfy io.IO without supporting writes. A few table write paths still used direct WriteFileIO assertions, so a read-only implementation could panic instead of returning an error.

Testing

- go test ./table -run 'TestWritePathsReturnErrorForReadOnlyIO|TestWriteRecords/TestFSNotWritable|TestDoCommit_NonWriteFileIOReturnsError' -count=1
- go test ./table -count=1
- git diff --check

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -54,12 +54,25 @@ import (
 // issue #830).
 var ErrCommitFailed = errors.New("commit failed, refresh and try again")
 
-// ErrWriteIORequired is returned by doCommit when the table's file system
-// does not implement io.WriteFileIO. Manifest-list rebuild on retry requires
-// write access; failing fast here is preferable to silently skipping the
-// rebuild and reintroducing the stale-parent data-loss bug. Callers that
-// need to detect this condition should use errors.Is(err, ErrWriteIORequired).
-var ErrWriteIORequired = errors.New("commit: file system does not implement WriteFileIO")
+// ErrWriteIORequired is returned by write paths when the table's file system
+// does not implement io.WriteFileIO. Commit retries also fail fast on this
+// condition because manifest-list rebuilds need write access; skipping that
+// rebuild can reintroduce stale-parent data loss. Callers should use
+// errors.Is(err, ErrWriteIORequired) to detect the precise condition, or
+// errors.Is(err, iceberg.ErrNotImplemented) for compatibility with older
+// WriteRecords behavior.
+var ErrWriteIORequired = fmt.Errorf("%w: file system does not implement WriteFileIO", iceberg.ErrNotImplemented)
+
+// requireWriteFileIO should run immediately after resolving the table FS and
+// before mutating transaction state such as automatic name mapping.
+func requireWriteFileIO(fs icebergio.IO) (icebergio.WriteFileIO, error) {
+	wfs, ok := fs.(icebergio.WriteFileIO)
+	if !ok {
+		return nil, ErrWriteIORequired
+	}
+
+	return wfs, nil
+}
 
 // ErrSnapshotNotFound is returned (wrapped) by metadata lookups and by
 // computeOwnManifests when a snapshot ID does not exist in the table's
@@ -392,9 +405,9 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
 	// Every real commit-path FS implements WriteFileIO. Failing here is
 	// preferable to silently skipping the manifest-list rebuild inside the
 	// retry loop — a skip reintroduces the original stale-parent data loss.
-	wfs, ok := fs.(icebergio.WriteFileIO)
-	if !ok {
-		return nil, fmt.Errorf("%w: manifest list rebuild requires write access", ErrWriteIORequired)
+	wfs, err := requireWriteFileIO(fs)
+	if err != nil {
+		return nil, fmt.Errorf("%w: manifest list rebuild requires write access", err)
 	}
 
 	var (
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -154,9 +154,9 @@ func (t *Transaction) addValidator(v conflictValidatorFunc) {
 	t.validators = append(t.validators, v)
 }
 
-func (t *Transaction) appendSnapshotProducer(afs io.IO, props iceberg.Properties) *snapshotProducer {
+func (t *Transaction) appendSnapshotProducer(wfs io.WriteFileIO, props iceberg.Properties) *snapshotProducer {
 	manifestMerge := t.meta.props.GetBool(ManifestMergeEnabledKey, ManifestMergeEnabledDefault)
-	updateSnapshot := t.updateSnapshot(afs, props, OpAppend)
+	updateSnapshot := t.updateSnapshot(wfs, props, OpAppend)
 	if manifestMerge {
 		return updateSnapshot.mergeAppend()
 	}
@@ -164,10 +164,10 @@ func (t *Transaction) appendSnapshotProducer(afs io.IO, props iceberg.Properties
 	return updateSnapshot.fastAppend()
 }
 
-func (t *Transaction) updateSnapshot(fs io.IO, props iceberg.Properties, operation Operation) snapshotUpdate {
+func (t *Transaction) updateSnapshot(wfs io.WriteFileIO, props iceberg.Properties, operation Operation) snapshotUpdate {
 	return snapshotUpdate{
 		txn:           t,
-		io:            fs.(io.WriteFileIO),
+		io:            wfs,
 		snapshotProps: props,
 		operation:     operation,
 	}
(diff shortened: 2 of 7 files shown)


tanmayrauth (commented):
LGTM, the type change on updateSnapshot/appendSnapshotProducer is the nice part, since it makes the write-IO requirement compiler-enforced at every call site rather than relying on each path remembering to assert.

  One question: WriteRecords used to wrap iceberg.ErrNotImplemented and now returns ErrWriteIORequired instead (write_records.go:181). Nothing in-tree depends on the old sentinel, but WriteRecords is exported — is dropping the ErrNotImplemented match intentional? Fine by me if so, just want to make sure it's a deliberate consolidation rather than an accidental compat break.

  One nit on coverage: the new test exercises append/overwrite/CoW-delete/row-delta, but performMergeOnReadDeletion has its own requireWriteFileIO check that nothing hits — the test's Delete uses default CoW mode, so the MoR branch is never taken. Might be worth one case that forces merge-on-read just so that path is pinned too.

zeroshade (approved):

zeroshade:
Merge pending answers to the question and nit from @tanmayrauth's comment

table/write_records.go
```diff
@@ -178,9 +177,9 @@ func WriteRecords(ctx context.Context, tbl *Table,
 		return internal.SingleErrorIter[iceberg.DataFile](err)
 	}
 
-	writeFS, ok := fs.(iceio.WriteFileIO)
-	if !ok {
-		return internal.SingleErrorIter[iceberg.DataFile](fmt.Errorf("%w: filesystem does not support writing", iceberg.ErrNotImplemented))
+	writeFS, err := requireWriteFileIO(fs)
```
laskoviymishka, line 180: This drops the `iceberg.ErrNotImplemented` wrapping. `WriteRecords` is exported, so anyone matching on `errors.Is(err, iceberg.ErrNotImplemented)` silently stops matching after this — a quiet breaking change.

I'd have `ErrWriteIORequired` wrap `ErrNotImplemented` so both `errors.Is` checks keep passing, e.g. define it as `fmt.Errorf("%w: file system does not implement WriteFileIO", ErrNotImplemented)`, or wrap here: `fmt.Errorf("%w: %w", iceberg.ErrNotImplemented, ErrWriteIORequired)`. `ErrWriteIORequired` is the more accurate sentinel, so I lean toward the former and keeping the relationship. wdyt?

table/transaction.go
```diff
@@ -1224,10 +1251,14 @@ func (t *Transaction) Overwrite(ctx context.Context, rdr array.RecordReader, sna
 	if err != nil {
 		return err
 	}
+	wfs, err := requireWriteFileIO(fs)
```
laskoviymishka: `Overwrite` already calls `performCopyOnWriteDeletion` above this, which now guards immediately after its own `fsF`. So on a read-only FS we return from there long before reaching this second guard — it's dead for the read-only case, and we've done two `fsF(ctx)` round-trips.

I'd have `performCopyOnWriteDeletion` return the `WriteFileIO` it already resolved and thread it down here, so there's one fetch and one guard. If there's a reason the second fetch has to be independent (credential refresh, etc.) let's drop a comment, since otherwise it reads as redundant.

table/transaction.go
```diff
@@ -510,8 +514,13 @@ func (t *Transaction) ReplaceDataFiles(ctx context.Context, filesToDelete, files
 		}
 	}
 
+	wfs, err := requireWriteFileIO(fs)
```
laskoviymishka: If `ensureNameMapping`/`SetProperties(DefaultNameMappingKey)` runs above this guard in `ReplaceDataFiles` (and the same shape in `ReplaceDataFilesWithDataFiles`, `ReplaceFiles`, `AddDataFiles`, `AddFiles`), then on a read-only FS we've already mutated the transaction's in-memory name-mapping before bailing out — the txn is left half-applied.

`performCopyOnWriteDeletion` got this right by guarding immediately after `fsF`, before any name-mapping write. I'd move `requireWriteFileIO` up to right after `t.tbl.fsF(ctx)` in each of these so the gate runs before any state mutation. Worth a one-line comment on `requireWriteFileIO` noting the canonical placement so a new call site doesn't reintroduce this.

table/table.go
```diff
@@ -54,12 +54,19 @@ import (
...
-// ErrWriteIORequired is returned by doCommit when the table's file system
-// does not implement io.WriteFileIO. Manifest-list rebuild on retry requires
-// write access; failing fast here is preferable to silently skipping the
-// rebuild and reintroducing the stale-parent data-loss bug. Callers that
-// need to detect this condition should use errors.Is(err, ErrWriteIORequired).
-var ErrWriteIORequired = errors.New("commit: file system does not implement WriteFileIO")
+// ErrWriteIORequired is returned by write paths when the table's file system
+// does not implement io.WriteFileIO. Callers that need to detect this
+// condition should use errors.Is(err, ErrWriteIORequired).
+var ErrWriteIORequired = errors.New("file system does not implement WriteFileIO")
```
laskoviymishka: Generalizing this to "write paths" makes sense, but it drops the manifest-list rebuild rationale entirely — and this doc is the only place a consumer reading `go doc`/IDE hover learns *why* `doCommit` fails fast instead of skipping the rebuild and reintroducing the stale-parent data-loss bug (#830). The inline comment at the `doCommit` site has it, but the exported symbol is what people actually read.

I'd keep a sentence of that rationale here while generalizing the rest. Separately, the sentinel string also changed (`"commit: file system..."` → `"file system..."`); `ErrWriteIORequired` is new as of the prior PR so the window is small, but worth a line in the PR description for anyone matching on the message.

table/write_file_io_test.go
```diff
@@ -0,0 +1,147 @@
...
+			},
+		},
+		{
+			name: "row delta",
+			run: func(ctx context.Context, tbl *Table) error {
+				return tbl.NewTransaction().NewRowDelta(nil).
+					AddRows(newWriteIORequiredDataFile(t)).
+					Commit(ctx)
+			},
+		},
```
laskoviymishka, line 142: This covers append, overwrite, CoW delete, MoR delete, and row delta — but the PR patches ~11 write paths. `AddDataFiles`, `AddFiles`, `WriteEqualityDeletes`, `ReplaceDataFiles`, `ReplaceDataFilesWithDataFiles`, and `ReplaceFiles` all gained guards and none are exercised here.

`AddDataFiles` and `ReplaceDataFilesWithDataFiles` are cheap to add since `newWriteIORequiredDataFile` already exists. I'd add at least those plus `WriteEqualityDeletes`. Also worth noting: `rewriteSingleFile` and `writePositionDeletesForFiles` only run when the table has existing data partially matching the filter, so the empty-table cases never reach them — those are exactly the spots that would have panicked on the old unsafe assertion, so a case seeded with one data file would be the most valuable addition.

table/write_file_io_test.go
```diff
@@ -0,0 +1,147 @@
...
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			tbl := newReadOnlyWriteTable(t, tc.props)
+			var err error
+			require.NotPanics(t, func() {
```
laskoviymishka, line 189: The `NotPanics` wrapper is load-bearing here — it's guarding against the pre-fix `fs.(io.WriteFileIO)` panic — but that's not obvious to a future reader. A one-line comment would make it clear this isn't just ceremony, e.g. `// NotPanics guards against the pre-fix unsafe type assertion on a read-only FS.`

table/table.go
```diff
@@ -54,12 +54,19 @@ import (
...
+
+func requireWriteFileIO(fs icebergio.IO) (icebergio.WriteFileIO, error) {
```
laskoviymishka, line 68: Minor, while we're here: `requireWriteFileIO` lives in `table.go` but is now used across every write-path file. Not wrong, but a small `write_helpers.go` (or a comment here noting the canonical "call immediately after `fsF`" placement, tied to the guard-ordering point) would make the shared contract easier to find. Non-blocking.

laskoviymishka (changes requested):
Solid, focused fix — centralizing the capability check in `requireWriteFileIO` and threading `io.WriteFileIO` up through `updateSnapshot`/`appendSnapshotProducer` cleanly removes the unsafe `fs.(io.WriteFileIO)` assertions, and `write_file_io_test.go` proves the gate fires for the main write paths.

I'd hold this before merging though. The blocking one: `WriteRecords` previously returned `iceberg.ErrNotImplemented` and now returns a bare `ErrWriteIORequired` with no `errors.Is` relationship between them — since `WriteRecords` is exported, any caller matching on `errors.Is(err, iceberg.ErrNotImplemented)` silently stops matching. I'd make `ErrWriteIORequired` wrap `ErrNotImplemented` so both checks keep passing.

The other thing I'd settle here is the guard placement in the `Replace*`/`Add*` paths — if `ensureNameMapping` runs before the new guard, we mutate the transaction's name-mapping before bailing on a read-only FS, leaving it half-applied. `performCopyOnWriteDeletion` gets the ordering right (guard immediately after `fsF`); I'd match that everywhere.

A few smaller things I'd want, nut can be deffered to follow up:
- `Overwrite` adds a second `requireWriteFileIO` after `performCopyOnWriteDeletion` already gated — dead for the read-only case and a double `fsF` fetch; thread the `WriteFileIO` down instead.
- The exported `ErrWriteIORequired` doc dropped the manifest-list rebuild / stale-parent data-loss rationale (#830) — keep a sentence of it on the exported symbol.
- The new test covers 5 of ~11 patched paths; add at least `AddDataFiles`, `WriteEqualityDeletes`, and a data-file-seeded case so `rewriteSingleFile`/`writePositionDeletesForFiles` actually get exercised.

laskoviymishka (approved):
very nice now!

Issue #1247, opened 2026-06-22
Abort should remove incomplete Parquet data files


fallintoplace:
Follow-up from #1241.

That PR changes the failed batch-write path to call `Abort()` instead of `Close()`, which is the right control flow for avoiding Parquet footer finalization after a write error. One piece it intentionally leaves alone: `ParquetFileWriter.Abort()` currently only closes the underlying file handle.

For local files this is mostly just an orphan cleanup concern. For object/blob backends, though, it can leave a footer-less object in storage. Readers should not see it because the snapshot is never committed, but the object can still sit around until orphan cleanup catches it.

Expected behavior:

- `Abort()` closes the open writer/file handle.
- `Abort()` removes the incomplete file path from the `FileIO`.
- If close and remove both fail, both errors are returned, for example with `errors.Join`.

This also applies to `RollingDataWriter`'s deferred abort path, since it calls the same writer-level `Abort()` when a stream exits with an open file.

Suggested tests:

- `Abort()` removes the created file.
- remove errors are returned.
- `RollingDataWriter` abort removes the incomplete current file, if practical.

tanmayrauth:
@zeroshade  @laskoviymishka  I think we can close it.

PR #1257, opened 2026-06-22
fix(table): remove parquet files on abort

fallintoplace:
Summary
- remove the incomplete Parquet file when ParquetFileWriter.Abort is called
- return both close and remove failures from Abort with errors.Join
- cover direct abort cleanup and the rolling writer deferred-abort path

Why
#1241 switched failed batch writes to Abort instead of Close, but Abort still only closed the handle. On object stores that could leave a footer-less object behind even though the snapshot never commits it.

Testing
- go test ./table/internal -run 'TestParquetFileWriterAbort|TestWriteDataFileBatchesAbortsOnWriteError' -count=1\n- go test ./table -run 'TestRollingDataWriter/TestAbortWithZeroRowsWritten|TestRollingDataWriter/TestStreamErrorPathUsesAbort' -count=1\n- go test ./table/internal -count=1\n- go test ./table -count=1\n- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.8.0 run --timeout=10m ./table/...\n- git diff --check\n\nCloses #1247

### DIFF
--- a/io/gocloud/blob.go
+++ b/io/gocloud/blob.go
@@ -133,7 +133,15 @@ func (bfs *blobFileIO) Remove(name string) error {
 		return &fs.PathError{Op: "remove", Path: name, Err: err}
 	}
 
-	return bfs.Delete(bfs.ctx, name)
+	if err := bfs.Delete(bfs.ctx, name); err != nil {
+		if gcerrors.Code(err) == gcerrors.NotFound {
+			err = fs.ErrNotExist
+		}
+
+		return &fs.PathError{Op: "remove", Path: name, Err: err}
+	}
+
+	return nil
 }
 
 func (bfs *blobFileIO) Create(name string) (icebergio.FileWriter, error) {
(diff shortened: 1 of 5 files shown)


tanmayrauth (commented):
The close-then-remove ordering via errors.Join is the right call here — finalizing the handle before deleting handles the object-store/Windows case cleanly, and the tests cover both the direct and deferred-abort paths nicely.

  One thing I'm curious about: in writeDataFileBatches  (table/internal/parquet_files.go:321) the Abort error gets joined into the returned write error. On an object store where the failed write never persisted anything,  blobFileIO.Remove → gocloud Delete can come back with a not-found error, which would then tack  itself onto the real write failure as noise. In practice gocloud's Writer.Close() tends to materialize even an empty object so Remove finds it — but a custom lazy WriteFileIO might not. Do you think Abort should swallow a NotExist from Remove, or is surfacing it intentional?

zeroshade (changes requested):
@tanmayrauth brings up a good question (and an interaction I didn't realize myself). 

We should definitely consider that situation and swallow the NotExist in that scenario with the abort.

zeroshade (approved):

PR #1258, opened 2026-06-22
fix(table): stop manifest iteration spin

fallintoplace:
Summary
- stop reading from AllManifests' error channel after it closes cleanly
- keep the early-exit drain from blocking once that channel has been disabled
- add a regression that drains manifests from several snapshots after the clean close path has happened

Why
A closed error channel is always ready. When it closed without an error, the select loop could keep taking the nil error case instead of draining ready manifest results.

Testing
- go test ./table -run '^TestAllManifestsCompletesAfterErrorChannelCloses$' -count=1
- go test ./table -count=1
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.8.0 run --timeout=10m ./table/...
- git diff --check
- go test ./... -count=1

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -303,14 +303,21 @@ func (t Table) AllManifests(ctx context.Context) iter.Seq2[iceberg.ManifestFile,
 			go func() {
 				for range results {
 				}
-				for range errch {
+				if errch != nil {
+					for range errch {
+					}
 				}
 			}()
 		}()
 
 		for {
 			select {
-			case err := <-errch:
+			case err, ok := <-errch:
+				if !ok {
+					errch = nil
+
+					continue
+				}
 				if err != nil {
 					yield(nil, err)
 
--- a/table/all_manifests_internal_test.go
+++ b/table/all_manifests_internal_test.go
@@ -1,0 +1,128 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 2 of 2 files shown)


tanmayrauth (commented):
Fix looks right to me. Nil-ing `errch` once it closes so the closed channel stops winning the select is the correct move, and the `if errch != nil` guard on the drain is needed — otherwise it'd `for range nil` and leak the drain goroutine. Error ordering is safe too: `errch <- err` always precedes `close(errch)`, so the `!ok` branch can't swallow a pending error.

  One note on the test, not a blocker: I ran `TestAllManifestsCompletesAfter  ErrorChannelCloses` against the pre-fix table.go and it passes in ~0.01s (also under GOMAXPROCS=1). That's expected — `errch` only closes after g.Wait(), by which point every manifest is already buffered in `results`, so the old code is a tail-end CPU micro-spin rather than a hang, and the random select still drains to completion fast. So this reads more as defensive cleanup than a fix for an observable hang, and the test exercises the close path but won't fail if someone reverts the channel handling. Totally fine to keep as-is — might just be worth a one-line comment so it isn't mistaken for a hang regression.

table/table.go
```diff
@@ -303,14 +303,21 @@ func (t Table) AllManifests(ctx context.Context) iter.Seq2[iceberg.ManifestFile,
...
 
 		for {
 			select {
-			case err := <-errch:
+			case err, ok := <-errch:
+				if !ok {
+					errch = nil
+
+					continue
+				}
```
zeroshade, line 320: this is only safe because it's in the select right? otherwise it would block indefinitely?

zeroshade (approved):

table/table.go
fallintoplace: Setting `errch = nil` disables that case in the select. Outside a select it would block forever, which is why the drain path checks `errch != nil`.

laskoviymishka (approved):

PR #1259, opened 2026-06-22
fix(io): validate cloud object URI buckets

fallintoplace:
Summary

- parse cloud object locations without URL-unescaping, so raw `%` and spaces in keys keep working
- reject fully-qualified object paths whose authority differs from the opened bucket or container
- apply the same authority check through S3/GCS/OSS and ADLS
- use the parsed object location for WalkDir path reconstruction, keep the root trailing slash, and reject malformed authority-only roots instead of walking the whole bucket
- add regression coverage for raw key characters, cross-authority rejection, ADLS authority checks, WalkDir roots, and malformed roots

Why

The current gocloud FileIO is bucket/container-bound. Silently folding `s3://other-bucket/...` into the opened bucket can read, write, or delete the wrong object, so this now fails loudly when a manifest path points at a different authority. If iceberg-go should support cross-authority metadata and data paths, that should be handled with explicit per-authority routing in a follow-up instead of pretending the current bucket-bound FileIO can safely serve both.

Testing

- go test ./io/gocloud -run 'TestDefaultKeyExtractor|TestBlobFileIORejectsWrongBucketObjectPaths|TestBlobFileIOWalkDirRejectsWrongBucket|TestBlobFileIOWalkDirRejectsWrongAzureAuthority|TestAdlsKeyExtractor' -count=1
- go test ./io/gocloud -count=1
- go test ./io -count=1
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m ./io/gocloud ./io
- git diff --check

### DIFF
--- a/io/config.go
+++ b/io/config.go
@@ -17,6 +17,14 @@
 
 package io
 
+// Constants for generic object store configuration options.
+const (
+	// ObjectStoreStrictAuthorityValidation rejects fully-qualified object paths
+	// whose URI authority differs from the bucket or container used to create
+	// the FileIO.
+	ObjectStoreStrictAuthorityValidation = "object-store.strict-authority-validation"
+)
+
 // Constants for S3 configuration options
 const (
 	S3Region          = "s3.region"
--- a/io/config.go
+++ b/io/config.go
@@ -17,15 +17,6 @@
 
 package io
 
-// Constants for generic object store configuration options.
-const (
-	// ObjectStoreStrictAuthorityValidation rejects fully-qualified object paths
-	// whose URI authority differs from the bucket or container used to create
-	// the FileIO. This is honored only by the Go gocloud-backed FileIO and is
-	// ignored by other Iceberg implementations.
-	ObjectStoreStrictAuthorityValidation = "object-store.strict-authority-validation"
-)
-
 // Constants for S3 configuration options
 const (
 	S3Region          = "s3.region"
--- a/io/config.go
+++ b/io/config.go
@@ -21,7 +21,8 @@ package io
 const (
 	// ObjectStoreStrictAuthorityValidation rejects fully-qualified object paths
 	// whose URI authority differs from the bucket or container used to create
-	// the FileIO.
+	// the FileIO. This is honored only by the Go gocloud-backed FileIO and is
+	// ignored by other Iceberg implementations.
 	ObjectStoreStrictAuthorityValidation = "object-store.strict-authority-validation"
 )
 
--- a/io/gocloud/azure.go
+++ b/io/gocloud/azure.go
@@ -230,6 +230,11 @@ func adlsObjectLocationExtractor(parsedURL *url.URL, opts ...keyExtractorOption)
 		}
 
 		uriPath := matches[3]
+		if uriPath != "" && !strings.HasPrefix(uriPath, "/") {
+			return objectLocation{}, fmt.Errorf("URI authority %q must be followed by an object path: %s",
+				authority, location)
+		}
+
 		key := strings.TrimPrefix(uriPath, "/")
 
 		parsed := objectLocation{
--- a/io/gocloud/azure.go
+++ b/io/gocloud/azure.go
@@ -220,7 +220,8 @@ func adlsObjectLocationExtractor(parsedURL *url.URL) objectLocationExtractor {
 
 		authority := matches[2]
 		if authority != expectedAuthority {
-			return objectLocation{}, fmt.Errorf("URI authority %q does not match configured authority %q",
+			return objectLocation{}, fmt.Errorf("%w: URI authority %q does not match configured authority %q",
+				ErrUnsupportedObjectAuthority,
 				authority, expectedAuthority)
 		}
 
@@ -233,6 +234,7 @@ func adlsObjectLocationExtractor(parsedURL *url.URL) objectLocationExtractor {
 		key := strings.TrimPrefix(uriPath, "/")
 
 		parsed := objectLocation{
+			scheme:       matches[1],
 			authority:    authority,
 			key:          key,
 			uriPrefix:    matches[1] + "://" + authority + "/",
@@ -240,7 +242,7 @@ func adlsObjectLocationExtractor(parsedURL *url.URL) objectLocationExtractor {
 		}
 
 		if key == "" {
-			return parsed, fmt.Errorf("%w: %s", errEmptyObjectKey, location)
+			return parsed, fmt.Errorf("%w: %s", ErrEmptyObjectKey, location)
 		}
 
 		return parsed, nil
--- a/io/gocloud/azure.go
+++ b/io/gocloud/azure.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log/slog"
 	"net/url"
 	"regexp"
 	"strings"
@@ -210,12 +209,7 @@ func adlsAuthority(parsed *url.URL) string {
 	return parsed.User.String() + "@" + parsed.Host
 }
 
-func adlsObjectLocationExtractor(parsedURL *url.URL, opts ...keyExtractorOption) objectLocationExtractor {
-	var cfg keyExtractorConfig
-	for _, opt := range opts {
-		opt(&cfg)
-	}
-
+func adlsObjectLocationExtractor(parsedURL *url.URL) objectLocationExtractor {
 	expectedAuthority := adlsAuthority(parsedURL)
 
 	return func(location string) (objectLocation, error) {
@@ -225,14 +219,9 @@ func adlsObjectLocationExtractor(parsedURL *url.URL, opts ...keyExtractorOption)
 		}
 
 		authority := matches[2]
-		if cfg.strictAuthorityValidation && authority != expectedAuthority {
+		if authority != expectedAuthority {
 			return objectLocation{}, fmt.Errorf("URI authority %q does not match configured authority %q",
 				authority, expectedAuthority)
-		}
-		if !cfg.strictAuthorityValidation && authority != expectedAuthority {
-			slog.Warn("using cross-authority ADLS URI with configured container",
-				"authority", authority,
-				"configured_authority", expectedAuthority)
 		}
 
 		uriPath := matches[3]
@@ -260,6 +249,6 @@ func adlsObjectLocationExtractor(parsedURL *url.URL, opts ...keyExtractorOption)
 }
 
 // adlsKeyExtractor creates a key extractor for Azure schemes using the adlsURIPattern pattern.
-func adlsKeyExtractor(parsedURL *url.URL, opts ...keyExtractorOption) KeyExtractor {
-	return keyExtractorFromObjectLocation(adlsObjectLocationExtractor(parsedURL, opts...))
+func adlsKeyExtractor(parsedURL *url.URL) KeyExtractor {
+	return keyExtractorFromObjectLocation(adlsObjectLocationExtractor(parsedURL))
 }
--- a/io/gocloud/azure.go
+++ b/io/gocloud/azure.go
@@ -203,10 +203,10 @@ func createAzureBucket(ctx context.Context, parsed *url.URL, props map[string]st
 
 func adlsAuthority(parsed *url.URL) string {
 	if parsed.User == nil {
-		return parsed.Host
+		return parsed.Hostname()
 	}
 
-	return parsed.User.String() + "@" + parsed.Host
+	return parsed.User.Username() + "@" + parsed.Hostname()
 }
 
 func adlsObjectLocationExtractor(parsedURL *url.URL) objectLocationExtractor {
@@ -235,7 +235,6 @@ func adlsObjectLocationExtractor(parsedURL *url.URL) objectLocationExtractor {
 		parsed := objectLocation{
 			authority:    authority,
 			key:          key,
-			uriKey:       key,
 			uriPrefix:    matches[1] + "://" + authority + "/",
 			hasAuthority: true,
 		}
@@ -246,9 +245,4 @@ func adlsObjectLocationExtractor(parsedURL *url.URL) objectLocationExtractor {
 
 		return parsed, nil
 	}
-}
-
-// adlsKeyExtractor creates a key extractor for Azure schemes using the adlsURIPattern pattern.
-func adlsKeyExtractor(parsedURL *url.URL) KeyExtractor {
-	return keyExtractorFromObjectLocation(adlsObjectLocationExtractor(parsedURL))
 }
--- a/io/gocloud/azure.go
+++ b/io/gocloud/azure.go
@@ -235,6 +235,7 @@ func adlsObjectLocationExtractor(parsedURL *url.URL, opts ...keyExtractorOption)
 		parsed := objectLocation{
 			authority:    authority,
 			key:          key,
+			uriKey:       key,
 			uriPrefix:    matches[1] + "://" + authority + "/",
 			hasAuthority: true,
 		}
--- a/io/gocloud/azure.go
+++ b/io/gocloud/azure.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"net/url"
 	"regexp"
 	"strings"
@@ -227,6 +228,11 @@ func adlsObjectLocationExtractor(parsedURL *url.URL, opts ...keyExtractorOption)
 		if cfg.strictAuthorityValidation && authority != expectedAuthority {
 			return objectLocation{}, fmt.Errorf("URI authority %q does not match configured authority %q",
 				authority, expectedAuthority)
+		}
+		if !cfg.strictAuthorityValidation && authority != expectedAuthority {
+			slog.Warn("using cross-authority ADLS URI with configured container",
+				"authority", authority,
+				"configured_authority", expectedAuthority)
 		}
 
 		uriPath := matches[3]
--- a/io/gocloud/azure.go
+++ b/io/gocloud/azure.go
@@ -201,21 +201,53 @@ func createAzureBucket(ctx context.Context, parsed *url.URL, props map[string]st
 	return azureblob.OpenBucket(ctx, client, nil)
 }
 
-// adlsKeyExtractor creates a key extractor for Azure schemes using the adlsURIPattern pattern
-func adlsKeyExtractor() KeyExtractor {
-	return func(location string) (string, error) {
+func adlsAuthority(parsed *url.URL) string {
+	if parsed.User == nil {
+		return parsed.Host
+	}
+
+	return parsed.User.String() + "@" + parsed.Host
+}
+
+func adlsObjectLocationExtractor(parsedURL *url.URL, opts ...keyExtractorOption) objectLocationExtractor {
+	var cfg keyExtractorConfig
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+
+	expectedAuthority := adlsAuthority(parsedURL)
+
+	return func(location string) (objectLocation, error) {
 		matches := adlsURIPattern.FindStringSubmatch(location)
 		if len(matches) < 4 {
-			return "", fmt.Errorf("invalid ADLS location: %s", location)
+			return objectLocation{}, fmt.Errorf("invalid ADLS location: %s", location)
+		}
+
+		authority := matches[2]
+		if cfg.strictAuthorityValidation && authority != expectedAuthority {
+			return objectLocation{}, fmt.Errorf("URI authority %q does not match configured authority %q",
+				authority, expectedAuthority)
 		}
 
 		uriPath := matches[3]
 		key := strings.TrimPrefix(uriPath, "/")
 
-		if key == "" {
-			return "", fmt.Errorf("URI path is empty: %s", location)
+		parsed := objectLocation{
+			authority:    authority,
+			key:          key,
+			uriPrefix:    matches[1] + "://" + authority + "/",
+			hasAuthority: true,
 		}
 
-		return key, nil
+		if key == "" {
+			return parsed, fmt.Errorf("%w: %s", errEmptyObjectKey, location)
+		}
+
+		return parsed, nil
 	}
+}
+
+// adlsKeyExtractor creates a key extractor for Azure schemes using the adlsURIPattern pattern.
+func adlsKeyExtractor(parsedURL *url.URL, opts ...keyExtractorOption) KeyExtractor {
+	return keyExtractorFromObjectLocation(adlsObjectLocationExtractor(parsedURL, opts...))
 }
--- a/io/gocloud/azure_test.go
+++ b/io/gocloud/azure_test.go
@@ -207,14 +207,14 @@ func TestAdlsKeyExtractor(t *testing.T) {
 			name:        "URI with no path",
 			input:       "abfs://container@account.dfs.core.windows.net",
 			expectedErr: "object key is empty",
-			wantErrIs:   errEmptyObjectKey,
+			wantErrIs:   ErrEmptyObjectKey,
 			shouldError: true,
 		},
 		{
 			name:        "URI with empty path",
 			input:       "abfs://container@account.dfs.core.windows.net/",
 			expectedErr: "object key is empty",
-			wantErrIs:   errEmptyObjectKey,
+			wantErrIs:   ErrEmptyObjectKey,
 			shouldError: true,
 		},
 		{
@@ -227,6 +227,7 @@ func TestAdlsKeyExtractor(t *testing.T) {
 			name:        "URI with different container",
 			input:       "abfs://other@account.dfs.core.windows.net/path/to/file.parquet",
 			expectedErr: "does not match configured authority",
+			wantErrIs:   ErrUnsupportedObjectAuthority,
 			shouldError: true,
 		},
 		{
--- a/io/gocloud/azure_test.go
+++ b/io/gocloud/azure_test.go
@@ -178,6 +178,7 @@ func TestAdlsKeyExtractor(t *testing.T) {
 		input       string
 		expectedKey string
 		expectedErr string
+		wantErrIs   error
 		shouldError bool
 	}{
 		{
@@ -205,11 +206,15 @@ func TestAdlsKeyExtractor(t *testing.T) {
 		{
 			name:        "URI with no path",
 			input:       "abfs://container@account.dfs.core.windows.net",
+			expectedErr: "object key is empty",
+			wantErrIs:   errEmptyObjectKey,
 			shouldError: true,
 		},
 		{
 			name:        "URI with empty path",
 			input:       "abfs://container@account.dfs.core.windows.net/",
+			expectedErr: "object key is empty",
+			wantErrIs:   errEmptyObjectKey,
 			shouldError: true,
 		},
 		{
@@ -221,11 +226,13 @@ func TestAdlsKeyExtractor(t *testing.T) {
 		{
 			name:        "URI with different container",
 			input:       "abfs://other@account.dfs.core.windows.net/path/to/file.parquet",
+			expectedErr: "does not match configured authority",
 			shouldError: true,
 		},
 		{
 			name:        "invalid ADLS location - invalid scheme",
 			input:       "s3://bucket/path/to/file.parquet",
+			expectedErr: "invalid ADLS location",
 			shouldError: true,
 		},
 	}
@@ -239,16 +246,19 @@ func TestAdlsKeyExtractor(t *testing.T) {
 			parsed, err := url.Parse(root)
 			require.NoError(t, err)
 
-			extractor := adlsKeyExtractor(parsed)
+			extractor := keyExtractorFromObjectLocation(adlsObjectLocationExtractor(parsed))
 			key, err := extractor(test.input)
 
 			if test.shouldError {
-				assert.Error(t, err, "Expected error for input: %s", test.input)
+				require.Error(t, err, "Expected error for input: %s", test.input)
 				if test.expectedErr != "" {
 					assert.ErrorContains(t, err, test.expectedErr, "Expected specific error for input: %s", test.input)
 				}
+				if test.wantErrIs != nil {
+					assert.ErrorIs(t, err, test.wantErrIs)
+				}
 			} else {
-				assert.NoError(t, err, "Unexpected error for input: %s", test.input)
+				require.NoError(t, err, "Unexpected error for input: %s", test.input)
 				assert.Equal(t, test.expectedKey, key, "Key mismatch for input: %s", test.input)
 			}
 		})
--- a/io/gocloud/azure_test.go
+++ b/io/gocloud/azure_test.go
@@ -23,6 +23,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCreateAzureBucketDefaultCredentialCalled(t *testing.T) {
@@ -171,12 +172,17 @@ func TestNewAdlsLocationUriParsing(t *testing.T) {
 }
 
 func TestAdlsKeyExtractor(t *testing.T) {
-	extractor := adlsKeyExtractor()
+	parsed, err := url.Parse("abfs://container@account.dfs.core.windows.net/")
+	require.NoError(t, err)
+
+	extractor := adlsKeyExtractor(parsed)
+	strictExtractor := adlsKeyExtractor(parsed, withStrictAuthorityValidation())
 
 	tests := []struct {
 		name        string
 		input       string
 		expectedKey string
+		strict      bool
 		shouldError bool
 	}{
 		{
@@ -210,6 +216,12 @@ func TestAdlsKeyExtractor(t *testing.T) {
 			shouldError: true,
 		},
 		{
+			name:        "strict URI with different container",
+			input:       "abfs://other@account.dfs.core.windows.net/path/to/file.parquet",
+			strict:      true,
+			shouldError: true,
+		},
+		{
 			name:        "invalid ADLS location - invalid scheme",
 			input:       "s3://bucket/path/to/file.parquet",
 			shouldError: true,
@@ -218,7 +230,11 @@ func TestAdlsKeyExtractor(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			key, err := extractor(test.input)
+			extract := extractor
+			if test.strict {
+				extract = strictExtractor
+			}
+			key, err := extract(test.input)
 
 			if test.shouldError {
 				assert.Error(t, err, "Expected error for input: %s", test.input)
--- a/io/gocloud/azure_test.go
+++ b/io/gocloud/azure_test.go
@@ -182,6 +182,7 @@ func TestAdlsKeyExtractor(t *testing.T) {
 		name        string
 		input       string
 		expectedKey string
+		expectedErr string
 		strict      bool
 		shouldError bool
 	}{
@@ -216,6 +217,12 @@ func TestAdlsKeyExtractor(t *testing.T) {
 			shouldError: true,
 		},
 		{
+			name:        "URI with query but no path",
+			input:       "abfs://container@account.dfs.core.windows.net?prefix=data",
+			expectedErr: "must be followed by an object path",
+			shouldError: true,
+		},
+		{
 			name:        "strict URI with different container",
 			input:       "abfs://other@account.dfs.core.windows.net/path/to/file.parquet",
 			strict:      true,
@@ -238,6 +245,9 @@ func TestAdlsKeyExtractor(t *testing.T) {
 
 			if test.shouldError {
 				assert.Error(t, err, "Expected error for input: %s", test.input)
+				if test.expectedErr != "" {
+					assert.ErrorContains(t, err, test.expectedErr, "Expected specific error for input: %s", test.input)
+				}
 			} else {
 				assert.NoError(t, err, "Unexpected error for input: %s", test.input)
 				assert.Equal(t, test.expectedKey, key, "Key mismatch for input: %s", test.input)
--- a/io/gocloud/azure_test.go
+++ b/io/gocloud/azure_test.go
@@ -172,18 +172,12 @@ func TestNewAdlsLocationUriParsing(t *testing.T) {
 }
 
 func TestAdlsKeyExtractor(t *testing.T) {
-	parsed, err := url.Parse("abfs://container@account.dfs.core.windows.net/")
-	require.NoError(t, err)
-
-	extractor := adlsKeyExtractor(parsed)
-	strictExtractor := adlsKeyExtractor(parsed, withStrictAuthorityValidation())
-
 	tests := []struct {
 		name        string
+		root        string
 		input       string
 		expectedKey string
 		expectedErr string
-		strict      bool
 		shouldError bool
 	}{
 		{
@@ -198,11 +192,13 @@ func TestAdlsKeyExtractor(t *testing.T) {
 		},
 		{
 			name:        "wasb valid URI",
+			root:        "wasb://container@account.blob.core.windows.net/",
 			input:       "wasb://container@account.blob.core.windows.net/path/to/file.parquet",
 			expectedKey: "path/to/file.parquet",
 		},
 		{
 			name:        "wasbs valid URI",
+			root:        "wasbs://container@account.blob.core.windows.net/",
 			input:       "wasbs://container@account.blob.core.windows.net/path/to/file.parquet",
 			expectedKey: "path/to/file.parquet",
 		},
@@ -223,9 +219,8 @@ func TestAdlsKeyExtractor(t *testing.T) {
 			shouldError: true,
 		},
 		{
-			name:        "strict URI with different container",
+			name:        "URI with different container",
 			input:       "abfs://other@account.dfs.core.windows.net/path/to/file.parquet",
-			strict:      true,
 			shouldError: true,
 		},
 		{
@@ -237,11 +232,15 @@ func TestAdlsKeyExtractor(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			extract := extractor
-			if test.strict {
-				extract = strictExtractor
+			root := test.root
+			if root == "" {
+				root = "abfs://container@account.dfs.core.windows.net/"
 			}
-			key, err := extract(test.input)
+			parsed, err := url.Parse(root)
+			require.NoError(t, err)
+
+			extractor := adlsKeyExtractor(parsed)
+			key, err := extractor(test.input)
 
 			if test.shouldError {
 				assert.Error(t, err, "Expected error for input: %s", test.input)
--- a/io/gocloud/blob.go
+++ b/io/gocloud/blob.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
-	"net/url"
 	"path/filepath"
 	"strings"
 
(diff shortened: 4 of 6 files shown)


io/gocloud/blob.go
```diff
@@ -72,25 +72,61 @@ func (f *blobOpenFile) Stat() (fs.FileInfo, error) { return f, nil }
 // KeyExtractor extracts the object key from an input path
 type KeyExtractor func(path string) (string, error)
 
+var errEmptyObjectKey = errors.New("object key is empty")
```
zeroshade: should this be public so that consumers can use `errors.Is`?

io/gocloud/blob.go
```diff
@@ -72,25 +72,61 @@ func (f *blobOpenFile) Stat() (fs.FileInfo, error) { return f, nil }
...
+				}
+			}
+		} else {
+			_, after, found := strings.Cut(location, "://")
+			if found {
+				location = after
+			}
+
+			key = strings.TrimPrefix(location, bucketName+"/")
+		}
```
zeroshade, line 161: what's the case where the `url.Parse` succeeds but the condition at line 87 fails? can't we still just use `parsed` to figure this out instead of needing to do `strings.Cut`?

io/gocloud/blob.go
```diff
@@ -72,25 +72,61 @@ func (f *blobOpenFile) Stat() (fs.FileInfo, error) { return f, nil }
...
+			return "", fmt.Errorf("%w: %s", errEmptyObjectKey, location)
 		}
 
 		return key, nil
 	}
 }
 
+func validatesURIAuthority(scheme string) bool {
+	switch scheme {
+	case "", "s3", "s3a", "s3n", "oss", "gs":
```
zeroshade: isn't this missing azure authorities and others?

zeroshade (changes requested):

tanmayrauth (commented):
Nice fix — the wrong-bucket rejection is clearly an improvement, and the regression tests cover it well. One thing I want to confirm before this lands:

  Switching the extractor to `url.Parse` (blob.go:81) means every location is now parsed as a URL up front, including the string-trim fallback branch. The old code never parsed, so it tolerated object keys with a literal `%`. Now anything that isn't a valid URL escape gets rejected outright — and because the parse  happens before the scheme gate, it hits relative keys and non-cloud schemes too, not just s3/gs:

  - `s3://bucket/data/100%off/f.parquet`     -> old: key "data/100%off/f.parquet";  new: error `invalid URL escape "%of"`
  - `data/100%off/file.parquet`              -> same, even with no scheme
  - `abfs://c@acct/data/50%done.pq`          -> same

  Literal `%` in a key is legal in S3/GCS, so this looks like a read-path tightening that would reject paths previously written/readable. Is that intended scope, or should the fallback branch keep handling non-parseable keys the old way?

  Smaller note: `EscapedPath()` at blob.go:95 re-encodes raw characters, so `s3://bucket/data/city=New York/f.parquet` becomes `...city=New%20York/...`. Symmetric within iceberg-go since write and read share the extractor, but worth calling out as a deliberate behavior change.

io/gocloud/blob.go
```diff
@@ -69,22 +68,70 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
-			location = after
+		parsed, err := splitObjectLocation(location)
+		if err != nil {
+			return "", err
 		}
 
-		key := strings.TrimPrefix(location, bucketName+"/")
+		key := parsed.key
+		if parsed.hasAuthority {
+			if parsed.authority != bucketName {
```
laskoviymishka, line 154: My one real hesitation about merging this: the hard authority check runs on the default read path, and I think it breaks a supported Iceberg layout.

A table can set `write.data.path = s3://data-bucket/` while its location and metadata live in `s3://metadata-bucket/...`. Manifests then carry full URIs like `s3://data-bucket/data/00000.parquet`. As I read it, the scan opens those through the single table IO keyed to the metadata bucket — so `parsed.authority` is `data-bucket`, `bucketName` is `metadata-bucket`, and we now reject a file that Java `S3FileIO` and PyIceberg read without complaint (neither restricts which bucket a path resolves to).

I'd keep the validation but make it not the default on reads — either construct a per-authority IO when a data-file URI's authority differs from the table's, or gate strict equality behind an opt-in property. Before I lean too hard on this though: can you confirm the scan path really does reuse one IO across differing data-path authorities? If it already loads an FS per data path, this concern mostly evaporates and I'd drop it to a nit. wdyt?

io/gocloud/blob_test.go
```diff
@@ -229,13 +333,67 @@ func TestBlobFileIOWalkDir(t *testing.T) {
...
+
+func TestBlobFileIOWalkDirRejectsWrongAzureAuthority(t *testing.T) {
+	ctx := context.Background()
+
+	bucket := memblob.OpenBucket(nil)
+	defer bucket.Close()
+
+	bfs := &blobFileIO{
+		Bucket:       bucket,
+		keyExtractor: defaultKeyExtractor("container@account.dfs.core.windows.net"),
```
laskoviymishka: This test doesn't actually exercise the Azure path. Production Azure/ADLS schemes are registered with `adlsKeyExtractor` (register.go), which extracts the path but never compares the authority — so `WriteFile("abfs://other-container@account.../data/file.parquet")` still silently lands in the opened container with the old bug intact.

The test only rejects because it wires up `defaultKeyExtractor` here, so it's proving the S3 extractor's behavior under an Azure-looking string rather than the real Azure code path. As written it implies a cross-container protection that doesn't exist.

I'd either add a configured authority + mismatch check to `adlsKeyExtractor` and point this test at it, or drop the test and note in the PR that ADLS authority validation is out of scope for now.

io/gocloud/blob.go
```diff
@@ -193,20 +240,30 @@ func createBlobFS(ctx context.Context, bucket *blob.Bucket, keyExtractor KeyExtr
...
+			return &fs.PathError{Op: "walkdir", Path: root, Err: err}
+		}
 
-	walkPath := strings.TrimPrefix(parsed.Path, "/")
-	if walkPath == "" {
 		walkPath = "."
 	}
 
-	parsed.Path = ""
+	location, err := splitObjectLocation(root)
```
laskoviymishka: We parse `root` twice here — once via `preprocess` (which runs `keyExtractor`) and again directly through `splitObjectLocation` — then use the second parse's `uriPrefix` to reconstruct the callback paths.

That's fine for `defaultKeyExtractor`, but for any extractor whose authority boundary differs from the naive up-to-`/?#` cut — `adlsKeyExtractor` today, or any custom one — the prefix used for reconstruction can diverge from what the extractor actually accepted. The two parsers disagreeing is exactly the kind of bug that stays invisible until someone plugs in a non-default extractor.

I'd parse once and share the `objectLocation` for both the key extraction and the prefix reconstruction, so there's a single source of truth for the authority boundary.

io/gocloud/blob.go
```diff
@@ -193,20 +240,30 @@ func createBlobFS(ctx context.Context, bucket *blob.Bucket, keyExtractor KeyExtr
...
+	if err != nil {
+		return &fs.PathError{Op: "walkdir", Path: root, Err: err}
+	}
 
 	return fs.WalkDir(bfs.Bucket, walkPath, func(path string, d fs.DirEntry, err error) error {
-		return fn(parsed.JoinPath(path).String(), d, err)
+		if location.hasAuthority {
+			if path == "." {
+				path = strings.TrimSuffix(location.uriPrefix, "/")
```
laskoviymishka: The root-entry path changed shape here. The old code returned `parsed.JoinPath(".")` → `s3://bucket/` for the root entry; this returns `strings.TrimSuffix(uriPrefix, "/")` → `s3://bucket` (no trailing slash), while every child is `uriPrefix + path` → `s3://bucket/...`. So the root entry no longer matches the form callers passed in as the walk root.

Current callers all filter with `!d.IsDir()`, so the root `"."` entry is skipped and nothing breaks today — but it's a silent behavior change waiting for the first caller that compares a walked entry against the stored root. I'd use `uriPrefix` consistently for the `"."` case so the root keeps its trailing slash.

io/gocloud/blob.go
```diff
@@ -69,22 +68,70 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
+	if authorityEnd == -1 {
+		authorityEnd = len(rest)
+	}
+
+	authority := rest[:authorityEnd]
+	if authority == "" {
+		return objectLocation{}, fmt.Errorf("URI authority is empty: %s", location)
+	}
+
+	key := ""
```
laskoviymishka, line 108: A malformed authority-only URI quietly degrades into a full bucket walk, which surprised me. For `s3://my-bucket?foo/bar`, `IndexAny` stops at `?`, the authority is `my-bucket`, and since `rest[authorityEnd]` is `?` not `/`, `key` stays empty. That bubbles up as `ErrEmptyObjectKey`, which `WalkDir` then treats as a bucket root and promotes to `walkPath = "."` — walking the entire bucket from a malformed input.

Iceberg locations shouldn't carry a query or fragment, so this is an edge, but silently turning a bad URI into a full walk feels risky. I'd either reject a non-`/` continuation after the authority explicitly, or at least keep that case from reaching the walk-everything fallback.

io/gocloud/blob.go
```diff
@@ -193,20 +240,30 @@ func createBlobFS(ctx context.Context, bucket *blob.Bucket, keyExtractor KeyExtr
 }
 
 func (bfs *blobFileIO) WalkDir(root string, fn fs.WalkDirFunc) error {
-	parsed, err := url.Parse(root)
+	walkPath, err := bfs.preprocess(root)
 	if err != nil {
-		return fmt.Errorf("invalid URL %s: %w", root, err)
-	}
+		if !errors.Is(err, ErrEmptyObjectKey) {
+			return &fs.PathError{Op: "walkdir", Path: root, Err: err}
```
laskoviymishka: Small one: the rest of this file uses space-separated op names (`"write file"`, `"new writer"`), so I'd make this `"walk dir"` to match.

While we're here — the authority-mismatch error is a plain `fmt.Errorf` with no sentinel, so callers can't `errors.Is` it the way they can `ErrEmptyObjectKey`. If we expect anyone to branch on it, an exported `ErrWrongAuthority` wrapped with `%w` would be more consistent. Not blocking.

io/gocloud/blob_test.go
```diff
@@ -84,6 +147,45 @@ func TestDefaultKeyExtractor(t *testing.T) {
...
+			name:   "gcs",
+			path:   "gs://other-bucket/data/file.parquet",
+			oldKey: "other-bucket/data/file.parquet",
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			err := bfs.WriteFile(tt.path, []byte("content"))
+			require.ErrorContains(t, err, "does not match configured authority")
+
+			exists, err := bucket.Exists(ctx, tt.oldKey)
```
laskoviymishka: This `oldKey` existence check is vacuously true — nothing ever seeds `oldKey`, so `Exists` returns false even if the write had been misrouted somewhere else, and the `require.ErrorContains` above already guarantees no write happened.

If we want this to actually prove "not misrouted to the old key", seed `oldKey` with a known sentinel first and assert it's unchanged after the failed write. Otherwise I'd just drop the `Exists` block — the `ErrorContains` carries the assertion on its own.

io/gocloud/blob_test.go
```diff
@@ -57,25 +85,60 @@ func TestDefaultKeyExtractor(t *testing.T) {
...
+			}
+			extractor := defaultKeyExtractor(bucketName)
 			key, err := extractor(test.input)
 
-			if test.shouldError {
-				assert.Error(t, err, "Expected error for input: %s", test.input)
+			if test.wantErrContains != "" {
+				assert.ErrorContains(t, err, test.wantErrContains, "Expected error for input: %s", test.input)
+				if test.wantErrIs != nil {
+					assert.True(t, errors.Is(err, test.wantErrIs), "expected errors.Is(%v)", test.wantErrIs)
```
laskoviymishka: `assert.ErrorIs(t, err, test.wantErrIs)` reads better here and prints the actual error chain on failure, versus `assert.True` over a bare bool.

Separately, the gating assertions in this table are `assert`, not `require` — on a nil-when-error-expected case the `ErrorContains` fails but execution continues into this `errors.Is` check (and on the success path `NoError` continues into `Equal`), so one root cause produces two failures. I'd make the gate a `require`.

io/gocloud/blob.go
```diff
@@ -69,22 +68,70 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
 
-// KeyExtractor extracts the object key from an input path
+// KeyExtractor extracts the object key from an input path.
 type KeyExtractor func(path string) (string, error)
 
-// defaultKeyExtractor extracts the object key by removing the scheme and bucket name from the URI
-// e.g., s3://bucket/path/file -> path/file
+// ErrEmptyObjectKey is returned when a location names a bucket or authority
+// without an object key.
+var ErrEmptyObjectKey = errors.New("object key is empty")
```
laskoviymishka, line 77: `ErrEmptyObjectKey` is exported, but its only consumer is the internal `errors.Is` in `WalkDir` — bucket-root handling is otherwise transparent to callers. Exporting it is a permanent API commitment, so I'd either unexport it to `errEmptyObjectKey`, or document it explicitly as part of the `KeyExtractor` author contract ("custom extractors must wrap this for bucket-root inputs") — which also ties into the `adlsKeyExtractor` gap.

laskoviymishka (changes requested):
The authority validation here closes a genuinely nasty gap — before this, a manifest path like `s3://other-bucket/data/file.parquet` got silently rewritten into the configured bucket and read/written/deleted from the wrong place, and `TestBlobFileIORejectsWrongBucketObjectPaths` proves that's now blocked for S3/GCS/OSS.

I'd hold this before merging, though, because as written the strict check may break legitimately cross-bucket tables on the default read path. Iceberg supports `write.data.path` pointing at a different bucket than the table location, so tables written by Java/PyIceberg/Spark embed full `s3://data-bucket/...` URIs in manifests while metadata lives in `s3://metadata-bucket/...`. If the scan reuses one IO keyed to the metadata bucket, this now rejects files that Java `S3FileIO` and PyIceberg read without restriction. I'd keep the validation but make strict equality opt-in (or load a per-authority IO), not the default. I left that scan-path assumption as a question inline — if it turns out we already load an FS per data path, it drops to a nit.

The other thing is Azure: the description reads as if cross-container writes are covered, but production Azure goes through `adlsKeyExtractor`, which has no authority check — the Azure test only passes because it wires up `defaultKeyExtractor`. So that protection isn't actually there yet.

- Confirm or repair the cross-bucket `write.data.path` read path — opt-in strict mode or a per-authority IO rather than a hard default reject.
- Either add authority validation to `adlsKeyExtractor` and retarget the Azure test at it, or drop the test and state Azure is out of scope.
- Parse `root` once in `WalkDir` and share the `objectLocation` for both key extraction and prefix reconstruction.
- Keep the trailing slash on the root `"."` entry so the walked root matches its stored form.
- Decide whether the malformed-authority-URI → full-bucket-walk fallback is intended; I'd reject it instead.

Once those are sorted, happy to take another pass.

io/gocloud/blob.go
fallintoplace: @laskoviymishka Can you please check again when you have time? Thank you for your attention.

io/gocloud/blob.go
```diff
@@ -188,25 +284,52 @@ func (bfs *blobFileIO) NewWriter(ctx context.Context, path string, overwrite boo
 		nil
 }
 
-func createBlobFS(ctx context.Context, bucket *blob.Bucket, keyExtractor KeyExtractor) icebergio.IO {
-	return &blobFileIO{Bucket: bucket, keyExtractor: keyExtractor, ctx: ctx}
+func createBlobFS(ctx context.Context, bucket *blob.Bucket, keyExtractor KeyExtractor, extractObject ...objectLocationExtractor) icebergio.IO {
```
laskoviymishka: I'd make `extractObject` a required parameter rather than variadic.

The variadic reads as composable but isn't — only `[0]` is ever used, a second arg is silently dropped, and `createBlobFS(ctx, bucket, ext, nil)` is legal and quietly installs a nil extractor that sends every call down the fallback path below. The only callers relying on the optionality are a few old 3-arg test helpers; I'd rather update those four call sites than keep a fallback that papers over a real semantic difference.

io/gocloud/blob.go
```diff
@@ -188,25 +284,52 @@ func (bfs *blobFileIO) NewWriter(ctx context.Context, path string, overwrite boo
...
+	if bfs.extractObject != nil {
+		return bfs.extractObject(root)
+	}
+
+	location, err := splitObjectLocation(root)
+	if err != nil {
+		return objectLocation{}, err
+	}
+
+	key, err := bfs.preprocess(root)
```
laskoviymishka: This fallback computes the key two incompatible ways and they disagree on cross-authority input.

For `s3://other-bucket/path`, `splitObjectLocation` leaves `location.uriPrefix = "s3://other-bucket/"` with `key = "path"`, but `bfs.preprocess(root)` (non-strict `defaultKeyExtractor`) then overwrites `location.key` with `legacyAuthorityKey` = `"other-bucket/path"` while `uriPrefix` still points at `other-bucket`. WalkDir reconstructs `s3://other-bucket/other-bucket/path` — authority doubled. If we make `extractObject` required (see the `createBlobFS` comment) this branch goes away entirely; if it stays, the key needs to come from `splitObjectLocation` too so prefix and key agree.

io/gocloud/blob.go
```diff
@@ -188,25 +284,52 @@ func (bfs *blobFileIO) NewWriter(ctx context.Context, path string, overwrite boo
...
 
-	parsed.Path = ""
-
 	return fs.WalkDir(bfs.Bucket, walkPath, func(path string, d fs.DirEntry, err error) error {
-		return fn(parsed.JoinPath(path).String(), d, err)
+		if location.hasAuthority {
+			if path == "." {
+				path = location.uriPrefix
+			} else {
+				path = location.uriPrefix + path
```
laskoviymishka: The same doubled-authority bug surfaces here on the non-strict path even with `extractObject` set: `defaultObjectLocationExtractor` folds the authority into `parsed.key` via `legacyAuthorityKey` but leaves `uriPrefix` as the original `s3://other-bucket/`, so `walkPath = "other-bucket/data"`, the walk yields `other-bucket/data/file.parquet`, and this line prepends `s3://other-bucket/` to give `s3://other-bucket/other-bucket/data/file.parquet`.

A table case would lock this down: write a file at `other-bucket/data/file.parquet`, walk `s3://other-bucket/data` non-strict against a FileIO bound to `my-bucket`, and assert no returned path contains a duplicated authority. Every current WalkDir test uses a matching authority or strict mode, so this path is unguarded.

io/gocloud/blob.go
```diff
@@ -69,33 +68,130 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
+		if parsed.hasAuthority {
+			if parsed.authority != bucketName {
+				if cfg.strictAuthorityValidation {
+					return objectLocation{}, fmt.Errorf("URI authority %q does not match configured authority %q",
+						parsed.authority, bucketName)
+				}
 
-		if key == "" {
-			return "", fmt.Errorf("URI path is empty: %s", location)
+				parsed.key = legacyAuthorityKey(parsed)
```
laskoviymishka: In non-strict mode a cross-authority URI silently becomes a key inside the configured bucket, which I think is riskier than the old behavior.

`s3://data-bucket/snap-1.avro` against a FileIO bound to `meta-bucket` writes an object at key `data-bucket/snap-1.avro` inside `meta-bucket` and reports success, where the SDK previously returned a clear not-found. I'd either make strict the default with an explicit `object-store.allow-cross-authority` opt-out, or at minimum emit a structured warning here so a misconfigured FileIO leaves a trace instead of silently misplacing data. Which way would you prefer — strict-by-default, or keep the legacy fold but make it loud?

io/config.go
```diff
@@ -17,6 +17,14 @@
 
 package io
 
+// Constants for generic object store configuration options.
+const (
+	// ObjectStoreStrictAuthorityValidation rejects fully-qualified object paths
+	// whose URI authority differs from the bucket or container used to create
+	// the FileIO.
+	ObjectStoreStrictAuthorityValidation = "object-store.strict-authority-validation"
```
laskoviymishka: This sits at the top level next to `s3.*`/`gcs.*`/`adls.*`, all of which are mirrored in Java and PyIceberg — but `object-store.strict-authority-validation` is Go-only. Java's `S3FileIO`/`ADLSFileIO` and PyIceberg's fsspec create per-authority FileIO instances and have no cross-authority knob, so a user who sets this in table properties expecting portable behavior gets it honored only by the gocloud backend and silently ignored elsewhere.

I'd either move it into the gocloud package unexported and wire it through the functional option you already have, or, if it stays exported, add a doc comment that it's honored only by the gocloud FileIO and ignored by Java/PyIceberg/Rust. wdyt?

laskoviymishka (changes requested):
Thanks for pushing this, replacing the hard wrong-bucket rejection with an opt-in strict mode is the right thing, and it settles the worry from the last round about valid cross-authority layouts getting bounced. The `objectLocation` refactor reads well, and the `errEmptyObjectKey` sentinel makes the WalkDir empty-root handling a lot cleaner than the old `url.Parse` path.

For the next step, the part I'd like us to work through is the non-strict (default) path, since that's where most users land. Two related things show up there: a cross-authority URI silently becomes a key inside the configured bucket instead of erroring, and WalkDir reconstructs paths with the authority doubled (`s3://other-bucket/other-bucket/...`). The old code surfaced both as a clean not-found from the SDK, so it's more a small regression on the default path than a new gap — and since Iceberg tables routinely split data and metadata across buckets, it's worth getting right. None of the new tests exercise it yet.

A couple of smaller things to settle alongside that:
- whether `ObjectStoreStrictAuthorityValidation` should be a top-level exported property at all, given there's no Java/PyIceberg/Rust counterpart — a user setting it expecting portable behavior would be silently ignored everywhere but the gocloud backend
- dropping the variadic `extractObject` for a required parameter, which also removes the nil-extractor fallback the doubled-authority path leans on

I left the specifics inline. None of it is far off — once the default-path behavior and the property question are sorted (plus a cross-authority WalkDir test to lock it in), I think we're close. Happy to take another pass; could land the round after, might even be ready then.

io/gocloud/blob.go
fallintoplace: Thank you for your attention, I will have a look tomorrow evening.

io/gocloud/azure.go
```diff
@@ -201,21 +201,53 @@ func createAzureBucket(ctx context.Context, parsed *url.URL, props map[string]st
...
+		}
+
+		authority := matches[2]
+		if cfg.strictAuthorityValidation && authority != expectedAuthority {
+			return objectLocation{}, fmt.Errorf("URI authority %q does not match configured authority %q",
+				authority, expectedAuthority)
 		}
 
 		uriPath := matches[3]
 		key := strings.TrimPrefix(uriPath, "/")
```
zeroshade, line 234: `adlsURIPattern` is `^(abfss?|wasbs?)://([^/?#]+)(.*)?$`, so group 3 captures everything after the authority *including* a leading `?` or `#`. For `abfs://container@account.dfs.core.windows.net?prefix=data`, `uriPath` is `?prefix=data` and `key` becomes `?prefix=data` - non-empty, so it's accepted as the object key. `splitObjectLocation` now explicitly rejects this shape for s3/gs (`must be followed by an object path`), and since `objectLocation()` also feeds `WalkDir`, the two paths disagree on the same malformed input. Suggest rejecting a non-empty `uriPath` that doesn't begin with `/` here too, to keep ADLS consistent with the generic parser (a matching `abfs://...?prefix=data` test would lock it in).

zeroshade (changes requested):
The opt-in strict authority validation and the `splitObjectLocation` hardening read well. One consistency gap that I don't think the earlier rounds covered (separate from the cross-authority / doubled-authority threads): the ADLS extractor doesn't reject malformed authority-only URIs the way the new generic parser now does. Details inline.

io/gocloud/register.go
```diff
@@ -30,6 +31,14 @@ func init() {
 	registerAzureSchemes()
 }
 
+func keyExtractorOptions(props map[string]string) []keyExtractorOption {
+	if enabled, err := strconv.ParseBool(props[icebergio.ObjectStoreStrictAuthorityValidation]); err == nil && enabled {
```
zeroshade: Strict-mode opt-in silently swallows `ParseBool` errors — a typo like `"ture"`/`"yes"` disables validation invisibly. Since the prop only accepts the two bool spellings, consider returning the parse error to the factory instead of ignoring it.

io/gocloud/blob.go
```diff
@@ -69,33 +68,138 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
+func defaultKeyExtractor(bucketName string, opts ...keyExtractorOption) KeyExtractor {
+	return keyExtractorFromObjectLocation(defaultObjectLocationExtractor(bucketName, opts...))
+}
+
 type blobFileIO struct {
 	*blob.Bucket
 
-	keyExtractor KeyExtractor
-	ctx          context.Context
+	keyExtractor  KeyExtractor
```
zeroshade: `blobFileIO` stores both `keyExtractor` and `extractObject`, but `keyExtractor` is always derived from `extractObject` via `keyExtractorFromObjectLocation` in every path. Storing both invites drift — consider keeping only `extractObject` and deriving the extractor where needed (and dropping it from the `createBlobFS` signature).

io/gocloud/blob.go
```diff
@@ -198,25 +302,64 @@ func (bfs *blobFileIO) NewWriter(ctx context.Context, path string, overwrite boo
...
+	}
+
+	uriKey := walked
+	if location.key != location.uriKey {
+		storageRoot := strings.TrimSuffix(location.key, "/")
+		if walked == location.key || walked == storageRoot {
+			uriKey = location.uriKey
+		} else {
+			prefix := storageRoot + "/"
+			if suffix, ok := strings.CutPrefix(walked, prefix); ok {
```
zeroshade: If `walked` matches none of `location.key` / `storageRoot` / `storageRoot+"/"`, this `ok=false` path leaves `uriKey = walked`, producing `uriPrefix + walked` with the bucket doubled (e.g. `s3://other-bucket/other-bucket/foo`). Unreachable today since `WalkDir` only emits descendants, but a silent fallback — consider an explicit error or an invariant comment.

zeroshade (changes requested):
Thanks @fallintoplace — a few low-severity items from review (none blocking):

laskoviymishka (changes requested):
The `url.Parse` → string-splitting fix is a clear win. Raw `%` and spaces now round-trip correctly, and the tests cover it. I’d happily take that as a standalone PR.

The cross-authority part feels like a separate architectural decision, and probably the reason this has taken a few rounds.

A gocloud `FileIO` is tied to one bucket when it’s opened, while Iceberg metadata can point to files in another bucket — separate data/metadata buckets, `write.data.path`, object-storage layout, etc. Rewriting the key can’t actually reach that other bucket, so we either read from the wrong place or fail in strict mode. Java’s `ResolvingFileIO` and PyIceberg solve this by routing each path to the right backend instead.

So I think we should first decide: do we want per-authority routing in gocloud, or do we treat cross-authority paths as unsupported and reject them?

The one behavior I would avoid is the current default: silently folding a foreign path into the opened bucket.

My preference:

* land the `%` fix separately
* reject cross-authority paths by default
* keep folding as opt-in only if there’s a real use case
* handle per-authority routing in a follow-up

Would like your steer on routing vs reject before digging into another round.

fallintoplace:
My preference would be:

- reject cross-authority paths by default in the current gocloud FileIO
- avoid exposing the strict flag as a portable top-level Iceberg property
- only keep the legacy fold behind an explicit opt-in if someone has a real compatibility case for it
- handle proper per-authority routing in a follow-up

Does that direction sound right to you? @laskoviymishka @zeroshade

laskoviymishka:
Sound good to me

fallintoplace:
@laskoviymishka Can you please check again this PR?

io/gocloud/blob.go
```diff
@@ -69,33 +68,103 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
+func defaultObjectLocationExtractor(bucketName string) objectLocationExtractor {
+	return func(location string) (objectLocation, error) {
+		parsed, err := splitObjectLocation(location)
+		if err != nil {
+			return objectLocation{}, err
+		}
+
+		if parsed.hasAuthority {
+			if parsed.authority != bucketName {
+				return objectLocation{}, fmt.Errorf("URI authority %q does not match configured authority %q",
```
laskoviymishka: This is the crux of the PR and I'd want to settle it before merge.

Rejecting any authority that differs from the open bucket assumes one FileIO only ever serves one authority. But the spec defines `file_path` as a full URI that can point anywhere — the Path Resolution example even mixes an `hdfs://` data file under an `s3://` table. Java's `ResolvingFileIO` routes by scheme, not authority (one `S3FileIO` serves every `s3://` bucket), and PyIceberg/iceberg-rust do the same. In iceberg-go a single `blobFileIO` bound to one bucket handles every Open across manifest-list/manifest/data-file reads, so a table with `write.data.path` on a second bucket, or manifests pointing at another bucket post-migration, produces `file_path`s this now hard-rejects.

I take your point in the description that a bucket-bound IO can't safely serve two authorities and that cross-authority routing is its own follow-up — and I agree writing to the wrong bucket should be caught. But a blanket read+write rejection breaks spec-legal tables today, whereas the property you actually want ("don't silently write to the wrong bucket") is narrower. Would erroring only on a scheme mismatch (which genuinely can't be routed), or making the authority check a write-open-time opt-in, get the safety without breaking cross-bucket reads? wdyt?

io/gocloud/blob.go
```diff
@@ -198,25 +272,70 @@ func (bfs *blobFileIO) NewWriter(ctx context.Context, path string, overwrite boo
...
+		return location.uriPrefix, nil
+	}
+
+	return location.uriPrefix + uriKey, nil
 }
 
 func (bfs *blobFileIO) WalkDir(root string, fn fs.WalkDirFunc) error {
-	parsed, err := url.Parse(root)
+	location, err := bfs.objectLocation(root)
+	walkPath := location.key
```
laskoviymishka: `walkPath := location.key` reads the result before we've checked `err`.

It's safe with today's extractors — a zero-value `location` gives `walkPath == ""` and the authority-mismatch path returns early — but the invariant is implicit. Any future extractor that returns a partially-populated `objectLocation` alongside a non-`errEmptyObjectKey` error would set `walkPath` from it and never correct it. I'd move the assignment below the error-kind check so it only runs where `location` is actually valid.

io/gocloud/blob.go
```diff
@@ -198,25 +272,70 @@ func (bfs *blobFileIO) NewWriter(ctx context.Context, path string, overwrite boo
...
+	return bfs.extractObject(root)
+}
+
+func walkedURIPath(location objectLocation, walked string) (string, error) {
+	if walked == "." {
+		return location.uriPrefix, nil
+	}
+
+	uriKey := walked
+	if location.key != location.uriKey {
```
laskoviymishka: Every extractor sets `uriKey = key`, so `location.key != location.uriKey` is never true in production and this whole block — the storageRoot trimming, the `CutPrefix`, the "outside storage root" error — is unreachable.

`TestWalkedURIPathRejectsPathOutsideStorageRoot` hand-builds an `objectLocation` with `key != uriKey`, a state no extractor emits, so it's testing dead code rather than any reachable behavior.

I'd drop `uriKey` and the branch and collapse this to the `.` shortcut plus `return location.uriPrefix + walked`, then add `uriKey` back when a concrete extractor actually needs the divergence. If you'd rather keep it, it needs an end-to-end WalkDir test that reaches the branch through the extractor path, plus a comment on the invariant. wdyt?

io/gocloud/azure.go
```diff
@@ -201,21 +201,54 @@ func createAzureBucket(ctx context.Context, parsed *url.URL, props map[string]st
...
 
-// adlsKeyExtractor creates a key extractor for Azure schemes using the adlsURIPattern pattern
-func adlsKeyExtractor() KeyExtractor {
-	return func(location string) (string, error) {
+func adlsAuthority(parsed *url.URL) string {
+	if parsed.User == nil {
+		return parsed.Host
+	}
+
+	return parsed.User.String() + "@" + parsed.Host
```
laskoviymishka: `User.String()` percent-encodes reserved userinfo characters, but the regex captures the container raw (`[^/?#]+`), so for any container with an encodable char these won't string-match and you get a spurious authority mismatch.

`getContainerNameFromURI` right above (azure.go:75) already uses `.User.Username()` for exactly this reason — I'd use `parsed.User.Username()` + `parsed.Hostname()` here to stay symmetric with the raw capture.

io/gocloud/blob.go
```diff
@@ -69,33 +68,103 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
+	}
+
+	key := ""
+	if authorityEnd < len(rest) {
+		if rest[authorityEnd] != '/' {
+			return objectLocation{}, fmt.Errorf("URI authority %q must be followed by an object path: %s",
+				authority, location)
+		}
+
+		key = strings.TrimPrefix(rest[authorityEnd:], "/")
```
laskoviymishka, line 117: Worth a comment here on why we keep the path raw — two things fall out of it that aren't obvious.

No `url.PathUnescape`, so `s3://bucket/data/100%off/file.parquet` yields the literal-`%` key `data/100%off/file.parquet`. That's correct for gocloud (keys are opaque), but a table written by Java — which URL-decodes via `URI.getPath()` — addresses `city%3DNew%20York` where we address the raw `city=New York`, so the same logical file resolves to different object keys across engines.

And because `IndexAny` stops at the first `/`, a trailing `?param=value#fragment` stays in the key verbatim (the test pins `expectedKey: "...parquet?param=value#fragment"`). I'm not sure that round-trips at the SDK layer — some backends treat `?` as a query separator — so I'd confirm the object actually reads back, and document the raw-key contract either way.

io/gocloud/blob.go
```diff
@@ -198,25 +272,70 @@ func (bfs *blobFileIO) NewWriter(ctx context.Context, path string, overwrite boo
...
-	walkPath := strings.TrimPrefix(parsed.Path, "/")
-	if walkPath == "" {
 		walkPath = "."
 	}
 
-	parsed.Path = ""
-
 	return fs.WalkDir(bfs.Bucket, walkPath, func(path string, d fs.DirEntry, err error) error {
-		return fn(parsed.JoinPath(path).String(), d, err)
+		if location.hasAuthority {
```
laskoviymishka, line 327: For a relative (no-scheme) root, `hasAuthority` is false so we skip the rewrite and hand the raw walked key straight to the callback. The old code always ran `parsed.JoinPath(path).String()`, so this is a behavioral change for relative inputs.

Probably fine, but nothing pins it — I'd add a no-scheme WalkDir case documenting that callbacks receive bare keys in that mode, so the contract is explicit.

io/gocloud/azure_test.go
```diff
@@ -209,6 +212,17 @@ func TestAdlsKeyExtractor(t *testing.T) {
...
+		{
+			name:        "URI with query but no path",
+			input:       "abfs://container@account.dfs.core.windows.net?prefix=data",
+			expectedErr: "must be followed by an object path",
+			shouldError: true,
+		},
+		{
+			name:        "URI with different container",
+			input:       "abfs://other@account.dfs.core.windows.net/path/to/file.parquet",
+			shouldError: true,
```
laskoviymishka, line 231: This case only asserts that some error came back, not which one. Since the point is the cross-container rejection, I'd assert `does not match configured authority` here via `expectedErr`, the way the query case above does — otherwise a wrong-but-still-erroring path passes silently.

Same for the empty-path cases: worth asserting `object key is empty`/`errEmptyObjectKey` to match the tightened `TestDefaultKeyExtractor`.

io/gocloud/blob_test.go
```diff
@@ -281,11 +489,7 @@ func TestBlobFileIOWalkDirAzureURI(t *testing.T) {
 
 	require.NoError(t, bucket.WriteAll(ctx, "path/to/file.parquet", []byte("data"), nil))
```
laskoviymishka: `TestBlobFileIOWalkDir` now covers raw `%` and space in the s3 walk, but the ADLS walk only exercises `path/to/file.parquet`.

Since the walked-path rewriting is the whole point of this PR, I'd add the same raw-special-char fixtures here (e.g. `path/100%off/file.parquet`, `path/city=New York/file.parquet`) so the ADLS rewrite path gets the same coverage as s3.

io/gocloud/azure.go
```diff
@@ -201,21 +201,54 @@ func createAzureBucket(ctx context.Context, parsed *url.URL, props map[string]st
...
+			return parsed, fmt.Errorf("%w: %s", errEmptyObjectKey, location)
 		}
 
-		return key, nil
+		return parsed, nil
 	}
 }
+
+// adlsKeyExtractor creates a key extractor for Azure schemes using the adlsURIPattern pattern.
+func adlsKeyExtractor(parsedURL *url.URL) KeyExtractor {
```
laskoviymishka: `adlsKeyExtractor` has no production caller now — `registerAzureSchemes` calls `adlsObjectLocationExtractor` directly, and only `TestAdlsKeyExtractor` uses this wrapper. I'd drop it and have the test call `adlsObjectLocationExtractor` (through `keyExtractorFromObjectLocation` if it wants the string form), rather than keep a wrapper alive purely for a test.

io/gocloud/blob.go
```diff
@@ -198,25 +272,70 @@ func (bfs *blobFileIO) NewWriter(ctx context.Context, path string, overwrite boo
...
 
-func createBlobFS(ctx context.Context, bucket *blob.Bucket, keyExtractor KeyExtractor) icebergio.IO {
-	return &blobFileIO{Bucket: bucket, keyExtractor: keyExtractor, ctx: ctx}
+func createBlobFS(ctx context.Context, bucket *blob.Bucket, extractObject objectLocationExtractor) icebergio.IO {
+	return &blobFileIO{Bucket: bucket, extractObject: extractObject, ctx: ctx}
+}
+
+func (bfs *blobFileIO) objectLocation(root string) (objectLocation, error) {
+	if bfs.extractObject == nil {
+		return objectLocation{}, errors.New("blob file IO missing object location extractor")
```
laskoviymishka, line 295: This guards a construction-time invariant at call time — `extractObject` is always set by `createBlobFS` and the test helpers. It's harmless, but if the goal is to make "no extractor" impossible, requiring it at construction says it more directly than a runtime check on every `objectLocation` call. Take it or leave it.

laskoviymishka (changes requested):
Nice cleanup overall — folding the scheme/bucket parsing into `splitObjectLocation` + the `objectLocation` struct reads much better than the old inline `strings.Cut` dance, and the tightened `TestDefaultKeyExtractor` matrix (raw `%`, spaces, empty authority, wrong bucket) is a real step up. I'd hold it before merging though, mostly on one design question plus a couple of correctness things.

The design question is the cross-authority rejection, and I don't think it's a slam-dunk either way. Hard-rejecting any authority that differs from the open bucket assumes a FileIO only ever serves one authority — but the spec defines `file_path` as a full URI that can point anywhere (the Path Resolution example even mixes `hdfs://` data under an `s3://` table), and Java `ResolvingFileIO` / PyIceberg / iceberg-rust all route by scheme, not bucket. So a `write.data.path` on a second bucket, or post-migration manifests pointing at another bucket, become unreadable under this change. I take your point that a bucket-bound IO genuinely can't serve two authorities and that routing is a separate follow-up — that's fair. But the safety property you're after ("don't silently write to the wrong bucket") is narrower than a blanket read+write reject. I'd like us to land on one of: reject only on scheme mismatch, or make the authority check a write-open-time opt-in — before this goes in. Left the detail inline.

On correctness: `walkPath := location.key` is read before the `err` check (safe today, fragile the moment an extractor returns a partial location with an error), and the `walkedURIPath` `key != uriKey` branch is dead — no extractor ever makes those diverge, and the test for it hand-builds a state production can't reach. Both are quick.

- settle the cross-authority behavior above (the one I actually care about)
- fix the read-before-err-check in WalkDir
- drop the dead `uriKey`/`walkedURIPath` branch or give it a reachable test
- use `User.Username()`, not `User.String()`, in `adlsAuthority` (percent-encoding vs the raw regex capture)
- document the raw-key contract (no unescaping; trailing `?`/`#` kept verbatim) and confirm the query/fragment case round-trips
- tighten the ADLS tests: assert the specific error strings, and cover raw special chars in the ADLS walk

Once the cross-authority question is settled and those are addressed, happy to take another pass and approve.

io/gocloud/blob.go
```diff
@@ -69,33 +68,102 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
+
+func defaultObjectLocationExtractor(bucketName string) objectLocationExtractor {
+	return func(location string) (objectLocation, error) {
+		parsed, err := splitObjectLocation(location)
+		if err != nil {
+			return objectLocation{}, err
 		}
 
-		key := strings.TrimPrefix(location, bucketName+"/")
+		if parsed.hasAuthority {
```
zeroshade, line 149: [MAJOR] `splitObjectLocation` records only authority-derived state while `defaultObjectLocationExtractor` validates only `authority == bucketName`, so an S3-backed FileIO can accept `gs://same-bucket/path` (and the reverse) as long as the authority matches. Carry the URI scheme in `objectLocation`, validate it against the backend's allowed schemes from registration, and add s3+gs wrong-scheme/same-authority tests.

io/gocloud/azure.go
```diff
@@ -201,21 +201,48 @@ func createAzureBucket(ctx context.Context, parsed *url.URL, props map[string]st
...
+
+	return func(location string) (objectLocation, error) {
 		matches := adlsURIPattern.FindStringSubmatch(location)
 		if len(matches) < 4 {
-			return "", fmt.Errorf("invalid ADLS location: %s", location)
+			return objectLocation{}, fmt.Errorf("invalid ADLS location: %s", location)
+		}
+
+		authority := matches[2]
+		if authority != expectedAuthority {
```
zeroshade, line 222: [MAJOR] This hard authority check also applies to the generic blob path at `io/gocloud/blob.go:139-143`, so read/write/delete/walk reject object URIs whose bucket/container differs from the opened root. Iceberg tables may legally reference data in another bucket/container. Please make an explicit design choice: route per authority, return a documented unsupported-mode error, or narrow the rejection to write/delete only.

io/gocloud/blob.go
```diff
@@ -69,33 +68,102 @@ func (f *blobOpenFile) Sys() any                   { return f.b }
...
 func (f *blobOpenFile) Stat() (fs.FileInfo, error) { return f, nil }
 
-// KeyExtractor extracts the object key from an input path
+// KeyExtractor extracts the object key from an input path.
 type KeyExtractor func(path string) (string, error)
 
-// defaultKeyExtractor extracts the object key by removing the scheme and bucket name from the URI
-// e.g., s3://bucket/path/file -> path/file
-func defaultKeyExtractor(bucketName string) KeyExtractor {
+var errEmptyObjectKey = errors.New("object key is empty")
```
zeroshade: [MINOR] `errEmptyObjectKey` is now surfaced through public IO errors (including `WalkDir` handling around `io/gocloud/blob.go:301-304`) but remains unexported, so consumers cannot use `errors.Is` against it. Export it or document that callers should not depend on this sentinel.

io/gocloud/blob_test.go
```diff
@@ -57,42 +86,142 @@ func TestDefaultKeyExtractor(t *testing.T) {
 			input:       "gs://my-bucket/path/to/file.parquet",
 			expectedKey: "path/to/file.parquet",
 		},
+		{
+			name:        "azure URI with path",
+			bucketName:  "container@account.dfs.core.windows.net",
+			input:       "abfs://container@account.dfs.core.windows.net/path/to/file.parquet",
+			expectedKey: "path/to/file.parquet",
+		},
+		{
```
zeroshade, line 96: [MINOR] The tests exercise different-bucket rejection here, but they still do not cover the wrong-scheme/same-authority bypass. Please add cases for an s3 FileIO receiving `gs://same-bucket/...` and a gs FileIO receiving `s3://same-bucket/...`.

zeroshade (changes requested):
Re-review: prior zeroshade feedback is largely addressed, but the remaining object-location handling issues below are still blocking.

- [MAJOR] URI scheme is parsed but not validated, allowing wrong-scheme same-authority paths.
- [MAJOR] Cross-authority object URIs are hard-rejected for all operations, which can break spec-legal tables with data outside the opened bucket/container.
- [MINOR] errEmptyObjectKey is exposed through public IO errors but remains unexported/undocumented.
- [MINOR] tests cover different-bucket rejection but not wrong-scheme same-authority bypass.

zeroshade (approved):
Re-reviewed after your latest push — all four of my earlier points are addressed: the scheme is now carried and validated against per-backend scheme sets, cross-authority rejection uses the documented ErrUnsupportedObjectAuthority sentinel, ErrEmptyObjectKey is exported, and the wrong-scheme/same-authority cases are tested. Thanks for the thorough follow-up. LGTM.

laskoviymishka (approved):
:shipit:

Issue #1262, opened 2026-06-23
Windows support: enable `windows-latest` in CI


abnobdoss:
## Summary

Once the CI `os` matrix actually picks runners (#1261), `windows-latest` fails across several packages. Opening this so Windows can be tracked and enabled. I don't have enough experience with iceberg-go's internals to be confident about the right fixes, so I'm reporting what breaks rather than proposing solutions.

## What breaks

- `io`: scheme inference rejects drive paths like `C:\dir`, and paths with `%` in them fail to parse.
- `table`: writes fail with "is a directory" because the data and metadata dirs never nest.
- `table` orphan cleanup: `file://C:/...` gets turned into UNC-looking `\\C:\...` with `%5C` left behind.
- `catalog/hadoop`: `NewCatalog` rejects drive-letter warehouses, and file URIs come out with backslashes.
- `catalog/sql`: DSNs like `file://C:\...` are rejected, and SQLite files can't be deleted while they're still open ("sharing violation").
- `catalog/rest`: `TestSigv4ConcurrentSigners` fails on windows go1.26.1 only. It passes on windows go1.25.8 and everywhere else, and isn't related to paths, so it may belong in its own issue.

## What I noticed

Most of these look like a Windows path being treated as a URL. `url.Parse("C:\\dir")` reads the drive letter `C:` as a scheme, which seems to be behind the `io`, `table`, and normalization failures. The `sql` "sharing violation" looks like the Windows rule that you can't remove a file while it's still open. The `rest` failure looks unrelated (that package isn't path-aware) and more like a go1.26 networking change or a fragile test. I'm not confident enough in the codebase to say these are the real root causes, so treat them as starting points rather than conclusions.

## Success criteria

`windows-latest` is enabled in the CI matrix and runs green with no errors.

deepak-2605:
url.Parse misreads Windows drive letters as URL schemes

Write path (table/locations.go): simpleLocationProvider/objectStoreLocationProvider wrap a *url.URL built from url.Parse(rootLocation). When rootLocation is a raw Windows path (no file:// prefix — e.g. what t.TempDir() returns, or any local/Hadoop-catalog warehouse given as a bare path), url.Parse reads the drive letter as a scheme:
`u, _ := url.Parse(`C:\Users\me\warehouse`)
// Scheme="c"  Opaque="\Users\me\warehouse"  Path=""`
[...]

PR #1264, opened 2026-06-23
ci: run lint-and-test on the real os matrix

### LINKED ISSUE #1261: CI: the OS matrix only ever runs on Linux
## Summary

`go-ci.yml` advertises a multi-OS matrix, but every job runs on Linux, so Windows and macOS are never actually tested.

## Current behavior

- `runs-on:` is hardcoded to `ubuntu-latest` while `name:` uses `${{ matrix.os }}`, so the `os` axis has no effect on where jobs run.
- The matrix just runs duplicate Linux jobs; macOS and Windows builds never happen.

## Expected behavior

- Each `os` value runs on its own runner, so macOS (and later Windows) is actually tested.

## Proposed fix

- Set `runs-on: ${{ matrix.os }}`. Enable macOS now. Windows needs path fixes first (separate issue), so leave it commented out for the moment.
- Add `.gitattributes` with `*.go text eol=lf` so gofmt doesn't fail on CRLF checkouts once non-Linux runners are in play. It's a no-op on Linux/macOS.

abnobdoss:
Closes #1261

## What
Set `runs-on: ${{ matrix.os }}` so the matrix actually selects runners, and turn on macOS alongside Linux. `windows-latest` stays commented out until the path issues in #1262 are sorted. Also add `*.go text eol=lf` to `.gitattributes` so gofmt passes when Go files are checked out on a non-Linux runner.

## Why
`runs-on:` was hardcoded to `ubuntu-latest` while `name:` used `${{ matrix.os }}`, so the `os` axis had no effect and every job ran on Linux. macOS was listed but never actually tested. Making the matrix real gets macOS covered now. Windows needs filesystem-path fixes first (#1262), so it stays off for the moment. The `.gitattributes` rule keeps the gofmt linter from failing if a runner checks out Go sources with CRLF.

## Validation
Tested in a fork: the macOS cells ran lint, tests, and the race job green.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,3 +21,7 @@
 # shell scripts with many flags. Unfortunately, directories themselves
 # won't recursively ignore, so we need the top level directories
 # as well as their files.
+
+# Force LF for Go sources so gofmt passes when checked out on Windows
+# (Windows runners would otherwise convert LF to CRLF and fail the gofmt linter).
+*.go text eol=lf
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -33,7 +33,7 @@ permissions:
 jobs:
   lint-and-test:
     name: ${{ matrix.os }} go${{ matrix.go }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     # Cache dirs are pinned to dot-prefixed paths under the workspace so they are
     # identical across Linux/Windows/macOS (no per-OS path branching) and are
     # excluded from `go`/golangci-lint `./...` expansion (leading-dot dirs are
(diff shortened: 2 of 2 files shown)


.gitattributes
```diff
@@ -20,4 +20,8 @@
 # uses an industry standard as opposed to possibly hard to read
 # shell scripts with many flags. Unfortunately, directories themselves
 # won't recursively ignore, so we need the top level directories
-# as well as their files.
\ No newline at end of file
+# as well as their files.
+
+# Force LF for Go sources so gofmt passes when checked out on Windows
+# (Windows runners would otherwise convert LF to CRLF and fail the gofmt linter).
+*.go text eol=lf
```
zeroshade: add a newline to the end of the file?

zeroshade (approved):
Just needs a rebase and a nipick

abnobdoss:
Thank you for the review! I've rebased and added the EOF newline.

Issue #479, opened 2025-07-08
"table X.Y is not an iceberg table" on a valid iceberg table


vbekiaris:
### Apache Iceberg version

None

### Please describe the bug 🐞

We are appending data to a table in AWS Glue Catalog using PyIceberg & PyArrow
then try to load the table from `iceberg-go` but `LoadTable` fails
with "table X.Y is not an iceberg table". The table parameters map
contains key "table_type" as required by [Glue catalog implementation](https://github.com/apache/iceberg-go/blob/237da8fd2034d772cd5969a372d2e411e3dfc66d/catalog/glue/glue.go#L697-L699)
but its value is not all-upper "ICEBERG", rather it's lower case.

On the writer side (PyIceberg/PyArrow) we use the [AWS Glue Iceberg REST endpoint](https://docs.aws.amazon.com/glue/latest/dg/connect-glu-iceberg-rest.html).

Both [PyIceberg](https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/glue.py#L376) and [java](https://github.com/apache/iceberg/blob/061ae58986db3495ff3af6f1932a96dd086e5fbd/aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java#L329) implementations
perform the same check ignoring the case of the value.
It seems that the condition check should be implemented in case-insensitive
way for `iceberg-go` too, wdyt?

PR #537, opened 2025-08-19
feat(glue): add optimistic locking and skip-archive support to lue catalog

xixipi-lining:
### Summary
This PR implements optimistic locking for table commits in the Glue catalog to prevent concurrent modification conflicts, and adds support for skipping table version archiving to improve performance for high-frequency workloads.

### Changes
- **Optimistic Locking**: Use Glue table `VersionId` to implement optimistic locking in `CommitTable` method
- **Skip Archive**: Add `glue.skip-archive` configuration option (default: `true`) to skip archiving old table versions, consistent with Java and Python implementations
- **Code Refactoring**: Extract `convertGlueToIceberg` method for better code reusability
- **Enhanced Testing**: Add comprehensive integration tests for optimistic locking scenarios

### Key Features
- ✅ Prevents concurrent table modification conflicts
- ✅ Improves performance for streaming workloads by skipping unnecessary archiving
- ✅ Maintains backward compatibility with Java and Python Iceberg implementations
- ✅ Handles both table creation and update scenarios

### Testing
- All existing unit tests pass
- Added new integration test `TestCommitTableOptimisticLockingIntegration` to verify concurrent commit behavior
- Enhanced test data with complete database name fields

PR #480, opened 2025-07-08
fix(catalog/glue): case insensitive type match

vbekiaris:
table_type and database_type key values are now
compared against "ICEBERG" string in case-insensitive way
(similar to how PyIceberg and Java implementations work).

Fixes #479

PR #1269, opened 2026-06-23
fix(catalog/glue): case-insensitive type match in convertGlueToIceberg

### LINKED ISSUE #1267: Glue catalog: case-sensitive table_type check in convertGlueToIceberg regresses #479
### Apache Iceberg version

main (development)

### Summary

`Catalog.convertGlueToIceberg` in `catalog/glue/glue.go` compares the Glue `table_type` parameter against the string `"ICEBERG"` with a strict `!=`. This regresses the case-insensitive behaviour added in #480, and reproduces the original failure mode from #479 for any Glue table whose `table_type` parameter is the lowercase `"iceberg"` value emitted by PyIceberg and the AWS Glue Iceberg REST endpoint.

This was already flagged by @tommika in a comment on #479 but no follow-up landed; filing a fresh issue so it doesn't fall off the radar.

### Where

`catalog/glue/glue.go`, line 703 on `main`:

```go
if glueTable.Parameters[tableParamTableType] != glueTypeIceberg {
    return nil, fmt.Errorf("table %s.%s is not an iceberg table", database, tableName)
}
```

The other two call sites in the same file already use `strings.EqualFold` (lines 676 and 770), so this is the only remaining case-sensitive check.

### How it regressed

- #480 (commit `ad83ccd`) made the two original call sites case-insensitive and closed #479.
- #537 (commit `1aa2a52`, "optimistic locking and skip-archive support") later added a new call site, `convertGlueToIceberg`, using a strict `!=` comparison.

### Reproduction

`LoadTable` first calls `getTable` (case-insensitive after #480), then `convertGlueToIceberg`. On a Glue table written by PyIceberg or via the Glue Iceberg REST endpoint, both of which write `table_type=iceberg` (lowercase), the first check passes but the second rejects the table with `table X.Y is not an iceberg table`. Same symptom as #479.

### Expected behaviour

`convertGlueToIceberg` should match the case-insensitive semantics of the other two call sites in this file, and of the PyIceberg and Java Glue implementations.

### Proposed fix

Replace the `!=` with `!strings.EqualFold(...)` to align with the other two sites. I have a one-line patch with a parallel unit test (`TestGlueConvertGlueToIcebergCaseInsensitive`, mirroring the existing `TestGlueGetTableCaseInsensitive`); PR to follow.

prafulla-kiran:
Fixes #1267 (related: #479, regressed by #537).

### Summary

`convertGlueToIceberg` is the only remaining case-sensitive `table_type` check in `catalog/glue/glue.go` — the other two sites (`getTable`, `filterTableListByType`) already use `strings.EqualFold` since #480. This brings the third site in line so `LoadTable` accepts Glue tables written by PyIceberg or via the AWS Glue Iceberg REST endpoint (both emit lowercase `iceberg`), matching the case-insensitive semantics of the PyIceberg and Java Glue implementations.

### Why

`LoadTable` → `getTable` now passes case-insensitively but the subsequent `convertGlueToIceberg` rejects the same table, reproducing the symptom from #479 (`table X.Y is not an iceberg table` on a valid Iceberg table). Originally flagged by @tommika in a comment on #479.

### Change

- One-line fix at `catalog/glue/glue.go:703`: `!=` → `!strings.EqualFold(...)`.
- New unit test `TestGlueConvertGlueToIcebergCaseInsensitive` mirroring the existing `TestGlueGetTableCaseInsensitive`, covering `ICEBERG`, `iceberg`, `IcEbErG`, and a non-Iceberg negative case. Verified by reverting the fix and confirming the new test fails.

### Test plan

- [x] `go test ./catalog/glue/` passes
- [x] `golangci-lint run ./catalog/glue/...` clean
- [x] New test fails without the fix, passes with it

### DIFF
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -700,7 +700,8 @@ func (c *Catalog) convertGlueToIceberg(ctx context.Context, glueTable *types.Tab
 		return nil, fmt.Errorf("missing parameters for table %s", tableName)
 	}
 
-	if glueTable.Parameters[tableParamTableType] != glueTypeIceberg {
+	// Keep this case-insensitive check consistent with getTable and filterTableListByType.
+	if !strings.EqualFold(glueTable.Parameters[tableParamTableType], glueTypeIceberg) {
 		return nil, fmt.Errorf("table %s.%s is not an iceberg table", database, tableName)
 	}
 
--- a/catalog/glue/glue_test.go
+++ b/catalog/glue/glue_test.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"testing"
@@ -250,6 +251,60 @@ func TestGlueGetTableCaseInsensitive(t *testing.T) {
 			} else {
 				assert.NoError(err)
 				assert.Equal("s3://test-bucket/test_table/metadata/abc123-123.metadata.json", tbl.Parameters[tableParamMetadataLocation])
+			}
+		})
+	}
+}
+
+func TestGlueConvertGlueToIcebergCaseInsensitive(t *testing.T) {
+	// Minimal valid v2 table metadata so NewFromLocation can succeed on the
+	// iceberg cases. Mirrors table/testdata/TableMetadataV2ValidMinimal.json.
+	const metadataJSON = `{"current-schema-id":0,"current-snapshot-id":-1,"default-sort-order-id":0,"default-spec-id":0,"format-version":2,"last-column-id":3,"last-partition-id":999,"last-sequence-number":0,"last-updated-ms":1602638573590,"location":"s3://bucket/test/location","metadata-log":[],"partition-specs":[{"fields":[],"spec-id":0}],"properties":{},"refs":{},"schemas":[{"fields":[{"id":1,"name":"x","required":true,"type":"long"},{"doc":"comment","id":2,"name":"y","required":true,"type":"long"},{"id":3,"name":"z","required":true,"type":"long"}],"schema-id":0,"type":"struct"}],"snapshot-log":[],"snapshots":[],"sort-orders":[{"fields":[],"order-id":0}],"statistics":[],"table-uuid":"9c12d441-03fe-4693-9a96-a0705ddf69c1"}`
+
+	metaPath := filepath.Join(t.TempDir(), "metadata.json")
+	require.NoError(t, os.WriteFile(metaPath, []byte(metadataJSON), 0o644))
+	metadataLocation := "file://" + metaPath
+
+	testCases := []struct {
+		name      string
+		tableType string
+		omitType  bool
+		shouldErr bool
+	}{
+		{"uppercase", "ICEBERG", false, false},
+		{"lowercase", "iceberg", false, false},
+		{"mixed case", "IcEbErG", false, false},
+		{"non-iceberg", "HIVE", false, true},
+		{"empty", "", false, true},
+		{"absent", "", true, true},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert := require.New(t)
+
+			params := map[string]string{
+				tableParamMetadataLocation: metadataLocation,
+			}
+			if !tc.omitType {
+				params[tableParamTableType] = tc.tableType
+			}
+			glueTable := &types.Table{
+				Name:         aws.String("test_table"),
+				DatabaseName: aws.String("test_database"),
+				Parameters:   params,
+			}
+
+			glueCatalog := &Catalog{}
+
+			tbl, err := glueCatalog.convertGlueToIceberg(context.TODO(), glueTable)
+			if tc.shouldErr {
+				assert.Error(err)
+				assert.Contains(err.Error(), "is not an iceberg table")
+			} else {
+				assert.NoError(err)
+				assert.NotNil(tbl)
+				assert.Equal([]string{"test_database", "test_table"}, tbl.Identifier())
 			}
 		})
 	}


tanmayrauth (approved):
LGTM, thanks for the fix.

catalog/glue/glue_test.go
```diff
@@ -255,6 +255,39 @@ func TestGlueGetTableCaseInsensitive(t *testing.T) {
...
 
+func TestGlueConvertGlueToIcebergCaseInsensitive(t *testing.T) {
+	assert := require.New(t)
+
+	testCases := []struct {
+		name      string
+		tableType string
+		wantErr   string
+	}{
+		{"uppercase", "ICEBERG", "missing metadata location"},
```
laskoviymishka: The three iceberg cases here don't actually pin the type check — they assert on "missing metadata location", which is a downstream error from the `metadata_location` lookup, not from the type guard. If someone reordered those two checks, or broke the type check in a way that still fell through to the metadata lookup, this test keeps passing. The whole point of the test is to prove lowercase `table_type` is now accepted, and right now it doesn't prove that.

I'd follow the `TestGlueGetTableCaseInsensitive` pattern instead: add `tableParamMetadataLocation` to the Parameters map for the iceberg cases, switch to a `shouldErr bool`, and on the success path assert `NoError` plus something concrete on the returned `*table.Table`. At minimum, assert the error does *not* contain "is not an iceberg table". While we're here it'd be cheap to add `""` and absent-key cases too, both expecting "is not an iceberg table". wdyt?

catalog/glue/glue_test.go
```diff
@@ -255,6 +255,39 @@ func TestGlueGetTableCaseInsensitive(t *testing.T) {
 	}
 }
 
+func TestGlueConvertGlueToIcebergCaseInsensitive(t *testing.T) {
+	assert := require.New(t)
```
laskoviymishka: `assert` is bound to the outer `t` but used inside the `t.Run` subtests, so a failure calls `FailNow()` on the parent — it kills the remaining subtests and attributes the failure to the parent rather than the named case. I'd move `assert := require.New(t)` inside the closure so it binds to the subtest's `t`.

This is copied from `TestGlueGetTableCaseInsensitive`, which has the same issue, so it's a pre-existing pattern rather than something this PR introduced — but since we're touching the success-path assertions anyway, worth fixing here rather than carrying it forward.

catalog/glue/glue.go
```diff
@@ -700,7 +700,7 @@ func (c *Catalog) convertGlueToIceberg(ctx context.Context, glueTable *types.Tab
 		return nil, fmt.Errorf("missing parameters for table %s", tableName)
 	}
 
-	if glueTable.Parameters[tableParamTableType] != glueTypeIceberg {
+	if !strings.EqualFold(glueTable.Parameters[tableParamTableType], glueTypeIceberg) {
```
laskoviymishka, line 704: This is now a second gate — every caller funnels through `getTable`, which already does the same `EqualFold` check, so the only path that relies on this one directly is `CommitTable`. That's fine as defensive depth, but the duplication is exactly what let #537 regress: a new caller site got the old comparison and nobody caught it. A one-line comment noting this must stay consistent with the `getTable` check would help the next person. Not blocking.

laskoviymishka (changes requested):
Nice catch on the root cause — this was the one `table_type` check #480 missed and #537 brought back, and lining it up with `EqualFold` matches both Java and PyIceberg. The write side always emits canonical uppercase, so there's no round-trip risk in relaxing the read.

I'd hold on the test before merging though. The three iceberg sub-cases assert on a downstream "missing metadata location" error rather than on the type check passing, so the test stays green even if the type guard breaks — it doesn't actually pin the behavior the PR is about. Following the `TestGlueGetTableCaseInsensitive` shape (populate `metadata_location`, assert the success path) fixes that.

One adjacent thing for a follow-up, not this PR: the `EXTERNAL_TABLE` check in `getTable` is still case-sensitive and PyIceberg doesn't enforce it at all, so a table registered via the Glue native Iceberg API could still get rejected at exactly the point this PR is trying to unblock. Worth its own issue.

Fix the test assertions and this is good to land — happy to take another pass once that's in.

prafulla-kiran:
Thanks for the close read, @laskoviymishka — both points are fair.

**Test reshaped.** Force-pushed an amend that pivots the iceberg sub-cases onto the success path, mirroring `TestGlueGetTableCaseInsensitive`. The test now writes a minimal valid v2 metadata document to `t.TempDir()`, sets `metadata_location` to a `file://` URI, and asserts `NoError` + a non-nil `*table.Table` with the right identifier for `ICEBERG` / `iceberg` / `IcEbErG`. The `HIVE` case keeps the `is not an iceberg table` substring check. Confirmed the test fails on the lowercase sub-case if the `!=` is restored, so it now pins the type-check behavior directly rather than leaning on a downstream error message.

**`EXTERNAL_TABLE` follow-up.** Agreed that's a separate concern and worth its own issue — I'll keep this PR scoped to the `convertGlueToIceberg` regression and file a follow-up for the `TableType` check after this lands. (PyIceberg's behavior aligns with what you suggested: it doesn't enforce `EXTERNAL_TABLE` at all, so Glue native Iceberg REST-registered tables are accepted there.)

Re-requesting review.

prafulla-kiran:
Apologies — I missed your two other inline notes on the first pass (only addressed the test-shape critique). Just force-pushed a second amend covering all three:

- **`glue.go:703`** — added the one-line consistency comment.
- **`glue_test.go:260`** — moved `assert := require.New(t)` inside the `t.Run` closure so failures bind to the subtest's `t` instead of the parent. Verified by reverting the production change: both `lowercase` and `mixed_case` now fail with correct per-subtest attribution, where previously the parent `FailNow()` killed `mixed_case` before it ran.
- **`glue_test.go`** — added `empty` (`table_type=""`) and `absent` (key not in Parameters map) sub-cases, both expecting `is not an iceberg table`. Needed an `omitType bool` on the case struct to build the Parameters map conditionally for the absent case.

Full `./catalog/glue/` suite green, lint clean. Sorry for the back-and-forth.

laskoviymishka (approved):

laskoviymishka:
@prafulla-kiran welcome to contributors gang!

PR #1279, opened 2026-06-23
fix(table): recover panics in RollingDataWriter.stream

laskoviymishka:
stream runs in its own goroutine (spawned by newRollingDataWriter) and has no caller able to recover a panic. Inside, closeWriter -> FileWriter.Close -> stats.ToDataFile panics when NewDataFileBuilder fails, so that panic escaped the goroutine and crashed the whole process instead of failing the write.

Add a recover at the top of stream that converts a panic into an error delivered on errorCh, mirroring the recover already present in equalityDeleteRecordsToDataFiles. The defer is declared after the close(r.errorCh) defer so it runs first (LIFO) and sends while errorCh is still open; the existing Abort defer still cleans up the half-written file.

### DIFF
--- a/table/rolling_data_writer.go
+++ b/table/rolling_data_writer.go
@@ -322,7 +322,23 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
 	defer close(r.errorCh)
 
 	var currentWriter tblutils.FileWriter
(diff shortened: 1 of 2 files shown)


table/rolling_data_writer.go
zeroshade, line 345: can we just combine these two defers rather than having so many individual ones?

zeroshade (approved):

PR #1268, opened 2026-06-23
feat(catalog/hadoop): Allow opt-in for arbitrary schemes for Hadoop Catalog

C-Loftus:
Next step for #1096 

This PR allows clients to opt in to use the hadoop catalog with arbitrary schemes. However, they have to explicitly set `allow-unsafe-commits`. I followed the guidance on this as specified here: https://github.com/apache/iceberg-go/issues/1096#issuecomment-4656700687 Note that this allows any arbitrary scheme. It is unclear if this should be expand to just add support for blob stores?

Note, I was going to add an integration test for minio with hadoop but as far as I can tell, I can't get the integration test runner to work for me. Unclear if I am doing something wrong or if we maybe need to make a change to this to use relative paths or something analogous? (I see the Dockerfile is there)

Note that this is the current integration test file without any changes to the runner.
```
> make integration-hadoop
go test -tags=integration -v -run="^TestHadoopIntegration" ./catalog/hadoop/...
=== RUN   TestHadoopIntegration
2026/06/23 12:18:27 github.com/testcontainers/testcontainers-go - Connected to docker: 
  Server Version: 29.1.3
  API Version: 1.52
  Operating System: Docker Desktop
  Total Memory: 7836 MB
  Labels:
    com.docker.desktop.address=unix:///Users/cloftus/Library/Containers/com.docker.docker/Data/docker-cli.sock
  Testcontainers for Go Version: v0.42.0
  Resolved Docker Host: unix:///var/run/docker.sock
  Resolved Docker Socket Path: /var/run/docker.sock
  Test SessionID: b94c401c54d8169c5101c563021f3b088636ea1765586a79ac332f17f6fff9fc
  Test ProcessID: 7e1b2a5e-9f59-4ff3-a4bf-7b3afef90979
#1 [internal] load local bake definitions
#1 reading from stdin 642B done
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 2B done
#2 DONE 0.0s
    hadoop_integration_test.go:58: 
                Error Trace:    /Users/cloftus/github/iceberg-go/catalog/hadoop/hadoop_integration_test.go:58
                                                        /Users/cloftus/go/pkg/mod/github.com/stretchr/testify@v1.11.1/suite/suite.go:211
                                                        /Users/cloftus/github/iceberg-go/catalog/hadoop/hadoop_integration_test.go:315
                Error:          Received unexpected error:
[...]

PR #1280, opened 2026-06-23
fix(catalog/hadoop): Load arbitrary filesystems dynamically for hadoop catalog

C-Loftus:
Following up on https://github.com/apache/iceberg-go/pull/1268 as next step for https://github.com/apache/iceberg-go/issues/1096 

The previous PR mistakenly specified `filesystem: icebergio.LocalFS{},` so the hadoop catalog will continue to only use the local filesystem which is incorrect. Note that for the time being, only `LocalFS` is wired up to support the interface for hadoop. Support for blob stores will come in a future PR.

- I moved the path normalization into an if statement to only do path cleanups if the uri is local. Otherwise for schemes like `s3://` it will use it directly as the client provides it. 
- I added a type check to see if the filesystem implements the `HadoopCatalogFS` interface. This is dynamic since we presumably want to support external users creating their own implementations if needed and loading them via `io.LoadFS`. We could of course also change the `NewCatalog` function to directly take in a filesystem which fulfils the interface but I assumed we didn't want to edit the constructor in that way.


catalog/hadoop/hadoop.go
```diff
@@ -135,35 +135,45 @@ func NewCatalog(name, warehouse string, props iceberg.Properties) (*Catalog, err
...
+			return nil, fmt.Errorf("hadoop catalog: failed to resolve absolute warehouse path: %w", err)
+		}
 
-	// Normalize to absolute path so the synthetic "location" property
-	// always produces a valid file:// URI.
-	absWarehouse, err := filepath.Abs(warehouse)
+		warehouse = absWarehouse
+	}
+
+	filesystem, err := io.LoadFS(context.Background(), nil, warehouse)
```
laskoviymishka: We're passing `nil` here, but the catalog's `props` are in scope and the cloud factories (`createS3Bucket`, `createGCSBucket`, `createAzureBucket`) read region, access key, secret, endpoint, etc. straight off the props map. With `nil` every lookup returns zero-value, so the FS falls back entirely to ambient SDK credentials and silently ignores anything the user set via `s3.access-key-id` / `s3.region` / `s3.endpoint`.

It's also inconsistent with the table path — `io.LoadFSFunc(c.props, metaPath)` a few lines down passes the real props, so the catalog FS and table FS end up configured differently.

`iceberg.Properties` is `map[string]string`, so this is just `io.LoadFS(context.Background(), props, warehouse)`. wdyt?

catalog/hadoop/hadoop.go
```diff
@@ -135,35 +135,45 @@ func NewCatalog(name, warehouse string, props iceberg.Properties) (*Catalog, err
...
 	if err != nil {
-		return nil, fmt.Errorf("hadoop catalog: failed to resolve absolute warehouse path: %w", err)
+		return nil, fmt.Errorf("hadoop catalog: failed to load filesystem: %w", err)
 	}
 
-	warehouse = absWarehouse
+	hadoopFs, ok := filesystem.(HadoopCatalogFS)
```
laskoviymishka, line 170: I think this assertion can never succeed for a remote store today, which means the PR's headline capability isn't actually reachable.

`HadoopCatalogFS` requires `Stat`, `Rename`, `RemoveAll`, `MkdirAll` on top of the read/write/list set. `blobFileIO` — what all the S3/GCS/Azure backends return — only implements `Open`, `Remove`, `Create`, `WriteFile`, `WalkDir`, `ReadFile`, `DeleteFiles`. So for any non-local scheme `LoadFS` succeeds, `ok` is false, and `NewCatalog` errors out even with `allow-unsafe-commits=true`. A user who correctly imports the gocloud backend sees "does not implement necessary functions" and reasonably reads it as a library bug.

The hard part is `Rename` — it's fundamentally non-atomic on object stores, which is exactly why Java warns against S3-backed `HadoopCatalog`. So I'd either extend `blobFileIO` with the four methods (with `Rename` as copy+delete) before merging, or scope this PR as a preparatory step and say so in the description. Leaving the assertion with nothing that can satisfy it is the dead end I'd want to avoid. Thoughts?

catalog/hadoop/hadoop.go
```diff
@@ -135,35 +135,45 @@ func NewCatalog(name, warehouse string, props iceberg.Properties) (*Catalog, err
 
 	if !isLocal && !allowUnsafeCommits {
 		return nil, fmt.Errorf("hadoop catalog: when using warehouse scheme %q, `allow-unsafe-commits` must be set to true", u.Scheme)
-	}
+	} else if isLocal {
```
laskoviymishka: There's a subtle issue with the third branch that this control flow hides. When `isLocal` is false and `allowUnsafeCommits` is true, we fall through with `warehouse` left as the raw URI, e.g. `s3://bucket/wh`.

Every path-building method — `namespaceToPath`, `tableToPath`, `metadataDir`, `metadataFilePath`, `versionHintPath`, `defaultTableLocation` — then calls `filepath.Join(c.warehouse, ...)`, and `filepath.Join` runs `filepath.Clean`, which collapses `s3://bucket` to `s3:/bucket`. So the `location` we write into table metadata comes out as `s3:/bucket/wh/...` — a single slash after the scheme — which Java, PyIceberg, and iceberg-rust all reject when they parse it as a URI.

I'd make the join scheme-aware: `filepath.Join` for local, and `path.Join` (or `strings.TrimRight(base, "/") + "/" + ...`) preserving the `scheme://authority` prefix for remote. Probably worth a single `joinPath` helper that branches on `isLocal`.

Also, while we're here — the `else if` after an unconditional `return` isn't idiomatic; dropping the `else` and using a plain `if isLocal {` would make the non-local fallthrough visible rather than implied.

catalog/hadoop/hadoop.go
laskoviymishka, line 170: Related, but worth its own note since it lives outside the diff: `LoadNamespaceProperties` builds the location property as `(&url.URL{Scheme: "file", Path: path}).String()` with a hardcoded `file://` scheme. Once a non-local warehouse is supported, an S3 namespace returns `file://s3:/bucket/wh/ns` — a `file://` URL wrapping the corrupted S3 path. I'd thread the actual warehouse scheme through there (or reuse the `namespaceToPath` result) as part of the same scheme-awareness fix.

catalog/hadoop/hadoop.go
```diff
@@ -135,35 +135,45 @@ func NewCatalog(name, warehouse string, props iceberg.Properties) (*Catalog, err
...
-	warehouse = absWarehouse
+	hadoopFs, ok := filesystem.(HadoopCatalogFS)
+	if !ok {
+		return nil, fmt.Errorf("hadoop catalog: filesystem %T does not implement necessary functions to be used as a Hadoop catalog", filesystem)
+	}
 
 	return &Catalog{
 		name:      name,
 		warehouse: warehouse,
 		// for the time being, we default to localfs since there is not yet
```
laskoviymishka: This comment now contradicts the code right below it — we're no longer defaulting to localfs, the FS is loaded dynamically from the registry by scheme. I'd drop it or replace with something like "filesystem is resolved dynamically from the IO registry based on the warehouse scheme."

catalog/hadoop/hadoop.go
laskoviymishka: `context.Background()` is the only option here since `NewCatalog` has no `ctx` param, which is fine and out of scope to change. But the S3 factory does use the context for `config.LoadDefaultConfig` and the bucket handle, so it'd be good to leave a `// TODO: propagate caller context once NewCatalog accepts one` here as a marker for the follow-up.

catalog/hadoop/hadoop.go
laskoviymishka, line 170: Minor, but the message is a little wordy given the `hadoop catalog:` prefix — naming the interface directly is more actionable: `"hadoop catalog: %T does not implement HadoopCatalogFS"`.

laskoviymishka (commented):
Thanks for picking this up — the diagnosis is right. #1268 left `filesystem` pinned to `LocalFS{}`, which made the `allow-unsafe-commits` opt-in for remote schemes pointless, and switching to a dynamic `io.LoadFS` lookup is the correct fix. The local-warehouse path looks good and existing users are unaffected.

I'd hold this before merging though, because as written it doesn't actually deliver the S3 capability in the title. `io.LoadFS` is called with `nil` props, so every cloud factory loses the credentials and region the user configured on the catalog — a silent correctness failure that happens to work under ambient IAM and breaks with static credentials. And no registered cloud FS implements `HadoopCatalogFS`, so the type assertion fails for every non-local scheme; an S3 warehouse with `allow-unsafe-commits=true` gets "does not implement necessary functions" and no way forward.

There's also a third issue just outside the diff: once a non-local warehouse is stored as a raw URI, every `filepath.Join(c.warehouse, ...)` collapses `s3://` to `s3:/`, so the `location` we'd write into metadata is unreadable by Java/PyIceberg/Rust. `LoadNamespaceProperties` has the matching problem with its hardcoded `file://` scheme.

- pass `c.props` (not `nil`) to `io.LoadFS`
- make the remote path real — either extend `blobFileIO` to satisfy `HadoopCatalogFS` (with copy+delete `Rename`), or explicitly scope this PR as a prep step in the description
- make path-building scheme-aware so remote URIs survive the join (and fix the `file://` literal in `LoadNamespaceProperties`)
- add tests: happy path with a stub `HadoopCatalogFS` + `allow-unsafe-commits=true`, the assertion-failure path, and confirm `TestNewCatalogRequiresOptInForRemoteSchemes` still holds under the new control flow

C-Loftus:
Thanks for your review @laskoviymishka ! I will make those changes. 

> make the remote path real — either extend blobFileIO to satisfy HadoopCatalogFS (with copy+delete Rename), or explicitly scope this PR as a prep step in the description

To clarify I mainly wanted to get in a fix for the localfs issue. As such this PR is just the next step for https://github.com/apache/iceberg-go/issues/1096 and not the full implementation. I have edited the PR description to make that more explicit.  I will save that logic for a follow up so it does not block iterations.

laskoviymishka:
@C-Loftus this would work

catalog/hadoop/hadoop.go
```diff
@@ -137,33 +139,45 @@ func NewCatalog(name, warehouse string, props iceberg.Properties) (*Catalog, err
...
+		// Normalize to absolute path so the synthetic "location" property
+		// always produces a valid file:// URI.
+		absWarehouse, err := filepath.Abs(warehouse)
+		if err != nil {
+			return nil, fmt.Errorf("hadoop catalog: failed to resolve absolute warehouse path: %w", err)
+		}
+
+		warehouse = absWarehouse
+	}
+	// TODO: propagate caller context once NewCatalog accepts one
```
C-Loftus, line 164: Added the requested TODO comment

catalog/hadoop/hadoop_test.go
```diff
@@ -75,6 +76,25 @@ func (f *barrierRenameNoReplaceFS) RenameNoReplace(oldpath, newpath string) erro
 	return f.LocalFS.RenameNoReplace(oldpath, newpath)
 }
 
+// a mock hadoop catalog filesystem to ensure that we
+// can load arbitrary new filesystem implementations
+// as long as they full the HadoopCatalogFS interface
+type stubHadoopCatalogFS struct {
+	icebergio.LocalFS
+}
```
C-Loftus, line 84: Requested stub helper for testing against another mock implementation

catalog/hadoop/hadoop.go
```diff
@@ -172,24 +186,50 @@ func (c *Catalog) CatalogType() catalog.Type {
 	return catalog.Hadoop
 }
 
+// joinPath is a helper that allows paths to be joined as both local filesystem
+// paths or as remote URIs needed for filesystems like blob stores.
+func joinPath(isLocal bool, base string, parts ...string) string {
```
C-Loftus, line 191: Condensed all joinPath logic into this function here. This is private to the package and tested.

C-Loftus (commented):
@laskoviymishka at your convenience this should be ready to review. I have added a few minor comments as a Github review to call out where I addressed your feedback.

laskoviymishka (approved):
🚢 👍

PR #1282, opened 2026-06-23
fix: keep ExpireSnapshots retention defaults at int64 on 32-bit builds

### LINKED ISSUE #1278: fix: keep ExpireSnapshots retention defaults at int64 on 32-bit builds
Follow-up to #877.

`ExpireSnapshots` reads the retention defaults with `int64(props.GetInt(...))`, but `Properties.GetInt` parses then narrows to platform-width `int`, and `MaxSnapshotAgeMsDefault` / `MaxRefAgeMsDefault` are `math.MaxInt`. On a 32-bit target `math.MaxInt` is `MaxInt32` (~25 days in ms), and the value is already truncated before the `int64()` widening — so the "keep everything" sentinel collapses and refs or snapshots older than ~25 days get expired even when no retention was configured.

Add a `Properties.GetInt64(key string, defVal int64) int64` helper (parse with bit size 64, return the full `int64`), use it for the two age properties, and retype the `*Default` constants to `int64(math.MaxInt64)`.

nanookclaw:
`Properties.GetInt` parses with `strconv.ParseInt(v, 10, 64)` then narrows to platform-width `int`. On 32-bit targets `math.MaxInt` is `MaxInt32` (~25 days in ms), so the "keep everything" sentinel for `MaxSnapshotAgeMsDefault` and `MaxRefAgeMsDefault` collapses before the `int64()` widening in `ExpireSnapshots` — snapshots or refs older than ~25 days get expired even when no retention was configured.

**Fix:** Add `Properties.GetInt64(key string, defVal int64) int64` that returns the full `int64` without platform truncation, retype the two age defaults to `int64(math.MaxInt64)`, and update `ExpireSnapshots` to use `GetInt64` for those properties.

`MinSnapshotsToKeepDefault` remains `math.MaxInt` with `GetInt` because its downstream type is `int`, not `int64`.

Closes #1278

### DIFF
--- a/table/properties.go
+++ b/table/properties.go
@@ -95,10 +95,10 @@ const (
 	MinSnapshotsToKeepDefault = math.MaxInt
 
 	MaxSnapshotAgeMsKey     = "max-snapshot-age-ms"
-	MaxSnapshotAgeMsDefault = math.MaxInt
+	MaxSnapshotAgeMsDefault = int64(math.MaxInt64)
 
 	MaxRefAgeMsKey     = "max-ref-age-ms"
-	MaxRefAgeMsDefault = math.MaxInt
+	MaxRefAgeMsDefault = int64(math.MaxInt64)
 
 	// CommitNumRetriesKey is the number of commit retry attempts before
 	// giving up on ErrCommitFailed from the catalog.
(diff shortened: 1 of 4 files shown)


nanookclaw:
Fixed the CI formatting failure in `types_test.go` and pushed `1500ace`.

Local verification after the fix:

- `go test ./ -run 'TestGetInt64|TestExpireSnapshots' -count=1`
- `git diff --check`

The previous Go matrix failures were all the same `gofmt` issue; the new workflow run is queued now.

laskoviymishka (approved):
👍

PR #1283, opened 2026-06-23
feat(cli): RewriteManifests Maintenance Task

rambleraptor:
Maintenance tasks are awesome! This introduces support for RewriteManifests and the associated CLI command.

The goal of RewriteManifests is to "rewrite" manifests by combining them together into larger manifests.

Note: I had to introduce a new parameter to the SnapshotProducer impl called `rebuildFromInheritedOnly`. I tried to match the others. Let me know if there's a better way to express this.

### DIFF
--- a/cmd/iceberg/rewrite_manifests.go
+++ b/cmd/iceberg/rewrite_manifests.go
@@ -1,0 +1,65 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/apache/iceberg-go/catalog"
+	"github.com/apache/iceberg-go/table"
+)
+
+func runRewriteManifests(ctx context.Context, output Output, cat catalog.Catalog, cmd *RewriteManifestsCmd) {
+	tbl := loadTable(ctx, output, cat, cmd.TableID)
+
+	var opts []table.RewriteManifestsOpt
+	if cmd.TargetManifestSize > 0 {
+		opts = append(opts, table.WithManifestTargetSize(cmd.TargetManifestSize))
+	}
+	if cmd.SpecID >= 0 {
+		opts = append(opts, table.WithRewriteSpecID(cmd.SpecID))
+	}
+
+	tx := tbl.NewTransaction()
+	res, err := tx.RewriteManifests(ctx, opts...)
+	if err != nil {
+		output.Error(fmt.Errorf("rewrite manifests failed: %w", err))
+		osExit(1)
+	}
+
+	// No-op — nothing was written or staged, so there is no commit to make. The
+	// reason distinguishes an empty table from an already-optimal layout.
+	if res.IsNoOp() {
+		if res.NoOpReason == table.NoOpNoSnapshot {
+			output.Text("Table has no current snapshot; nothing to rewrite.")
+		} else {
+			output.Text("Manifests already optimal; nothing to rewrite.")
+		}
+
+		return
+	}
+
+	if _, err := tx.Commit(ctx); err != nil {
+		output.Error(fmt.Errorf("commit failed: %w", err))
+		osExit(1)
+	}
+
+	output.Text(fmt.Sprintf("Rewrote %d manifests into %d.",
+		len(res.RewrittenManifests), len(res.AddedManifests)))
+}
--- a/table/rewrite_manifests.go
+++ b/table/rewrite_manifests.go
@@ -1,0 +1,423 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"slices"
+	"strconv"
+
+	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
+)
+
+// Snapshot summary keys for a manifest rewrite.
+const (
+	manifestsCreatedKey  = "manifests-created"  // new manifests the merge wrote
+	manifestsReplacedKey = "manifests-replaced" // input manifests the merge consumed
+	manifestsKeptKey     = "manifests-kept"     // manifests left untouched, delete manifests included
+	entriesProcessedKey  = "entries-processed"  // data-file entries, not rows
+)
+
+// NoOpReason explains why a rewrite changed nothing. It lets callers tell a
+// table with no current snapshot apart from one whose manifests are already
+// optimal, which would otherwise both surface as an empty result.
+type NoOpReason int
+
+const (
+	// NoOpNone means the rewrite produced changes (not a no-op).
+	NoOpNone NoOpReason = iota
+	// NoOpNoSnapshot means the table had no current snapshot to rewrite.
+	NoOpNoSnapshot
+	// NoOpAlreadyOptimal means the eligible manifests were already optimal.
+	NoOpAlreadyOptimal
+)
+
+func (r NoOpReason) String() string {
+	switch r {
+	case NoOpNoSnapshot:
+		return "no current snapshot"
+	case NoOpAlreadyOptimal:
+		return "manifests already optimal"
+	default:
+		return "none"
+	}
+}
+
+// RewriteManifestsResult reports the manifests changed by a rewrite.
+type RewriteManifestsResult struct {
+	// RewrittenManifests are the old manifests that were replaced.
+	RewrittenManifests []iceberg.ManifestFile
+	// AddedManifests are the new manifests written in their place.
+	AddedManifests []iceberg.ManifestFile
+	// NoOpReason is set when the rewrite changed nothing, distinguishing a
+	// missing snapshot from an already-optimal layout. NoOpNone otherwise.
+	NoOpReason NoOpReason
+}
+
+// IsNoOp reports whether the rewrite changed nothing. Callers should skip the
+// commit in that case rather than staging an empty REPLACE snapshot. NoOpReason
+// is the single source of truth.
+func (r *RewriteManifestsResult) IsNoOp() bool {
+	return r.NoOpReason != NoOpNone
+}
+
+type rewriteManifestsCfg struct {
+	targetSizeBytes int64
+	specID          *int
+	predicate       func(iceberg.ManifestFile) bool
+}
+
+// RewriteManifestsOpt configures [Transaction.RewriteManifests].
+type RewriteManifestsOpt func(*rewriteManifestsCfg)
+
+// WithManifestTargetSize overrides the target manifest size in bytes. A
+// non-positive size is ignored, leaving the default from the
+// commit.manifest.target-size-bytes property.
+func WithManifestTargetSize(size int64) RewriteManifestsOpt {
+	return func(c *rewriteManifestsCfg) {
+		if size > 0 {
+			c.targetSizeBytes = size
+		}
+	}
+}
+
+// WithRewriteSpecID restricts the rewrite to manifests of one partition spec.
+func WithRewriteSpecID(id int) RewriteManifestsOpt {
+	return func(c *rewriteManifestsCfg) { c.specID = &id }
+}
+
+// WithRewriteManifestPredicate only rewrites manifests for which pred is true.
+// Manifests that don't match are left untouched.
+func WithRewriteManifestPredicate(pred func(iceberg.ManifestFile) bool) RewriteManifestsOpt {
+	return func(c *rewriteManifestsCfg) { c.predicate = pred }
+}
+
+// rewriteManifests is a producer that merges small data manifests into
+// fewer, target-sized ones, committed as a metadata-only REPLACE snapshot.
+type rewriteManifests struct {
+	base *snapshotProducer
+	cfg  rewriteManifestsCfg
+
+	rewritten []iceberg.ManifestFile
+	added     []iceberg.ManifestFile
+
+	// superseded accumulates merged manifest files written by a rebuild on a
+	// prior OCC attempt that a later attempt replaced. doCommit removes them
+	// after a successful commit so retries don't leak orphaned manifests.
+	superseded []string
+
+	// result is the value returned to the caller. record() rewrites its
+	// fields on every pass, including OCC retries, so the pointer the caller
+	// holds reflects the manifests actually committed, not attempt 0's.
+	result *RewriteManifestsResult
+}
+
+func newRewriteManifestsProducer(txn *Transaction, fs iceio.WriteFileIO, props iceberg.Properties, cfg rewriteManifestsCfg) *snapshotProducer {
+	prod := createSnapshotProducer(OpReplace, txn, fs, nil, props)
+	prod.producerImpl = &rewriteManifests{base: prod, cfg: cfg, result: &RewriteManifestsResult{}}
+
+	return prod
+}
+
+// supersededManifests returns the merged manifests orphaned across OCC retries,
+// safe to delete once the commit resolves. When the commit never landed
+// (committed is false) the final attempt's manifests are orphaned too — a
+// snapshot that never committed references nothing — so they are appended.
+func (r *rewriteManifests) supersededManifests(committed bool) []string {
+	if committed {
+		return r.superseded
+	}
+
+	paths := slices.Clone(r.superseded)
+	for _, m := range r.added {
+		paths = append(paths, m.FilePath())
+	}
+
+	return paths
+}
+
+func (r *rewriteManifests) existingManifests(parent *Snapshot) ([]iceberg.ManifestFile, error) {
+	if parent == nil {
+		return nil, nil
+	}
+
+	return parent.Manifests(r.base.io)
+}
+
+func (r *rewriteManifests) deletedEntries(context.Context, *Snapshot) ([]iceberg.ManifestEntry, error) {
+	return nil, nil
+}
+
+func (r *rewriteManifests) processManifests(manifests []iceberg.ManifestFile) ([]iceberg.ManifestFile, error) {
+	var toRewrite, kept []iceberg.ManifestFile
+	for _, m := range manifests {
+		if r.eligible(m) {
+			toRewrite = append(toRewrite, m)
+		} else {
+			kept = append(kept, m)
+		}
+	}
+
+	mgr := manifestMergeManager{
+		targetSizeBytes: r.cfg.targetSizeBytes,
+		minCountToMerge: 1,    // force a merge regardless of count
+		mergeEnabled:    true, // explicit op ignores commit.manifest-merge.enabled
+		snap:            r.base,
+	}
+	merged, err := mgr.mergeManifests(toRewrite)
+	if err != nil {
+		return nil, err
+	}
+
+	// mergeManifests already wrote the new .avro files. If this pass fails
+	// before record() takes ownership, nothing will reference them, so delete
+	// them here rather than leak orphans that no commit/retry cleanup can reach.
+	owned := false
+	defer func() {
+		if !owned {
+			r.deleteWritten(toRewrite, merged)
+		}
+	}()
+
+	// Record on every pass, not just the first. An OCC retry re-runs this
+	// against the fresh parent and writes a different merged set; the last
+	// pass is the one that commits, so its counts are the ones that must win.
+	// record also runs the file-count guard, so an error here leaves the
+	// merged files to the defer above for cleanup.
+	if err := r.record(toRewrite, merged, kept); err != nil {
+		return nil, err
+	}
+	owned = true
+
+	return slices.Concat(merged, kept), nil
+}
+
+// eligible reports whether m is a data manifest selected for rewrite.
+func (r *rewriteManifests) eligible(m iceberg.ManifestFile) bool {
+	if m.ManifestContent() != iceberg.ManifestContentData {
+		return false
+	}
+	if r.cfg.specID != nil && int(m.PartitionSpecID()) != *r.cfg.specID {
+		return false
+	}
+	if r.cfg.predicate != nil && !r.cfg.predicate(m) {
+		return false
+	}
+
+	return true
+}
+
+// deleteWritten removes the manifests mergeManifests newly wrote for this pass.
+// Single-manifest bins pass through with their input path and are left alone;
+// only the merged files (in the output but not the input) are deleted.
+func (r *rewriteManifests) deleteWritten(input, merged []iceberg.ManifestFile) {
+	inPaths := make(map[string]struct{}, len(input))
+	for _, m := range input {
+		inPaths[m.FilePath()] = struct{}{}
+	}
+	for _, m := range merged {
+		if _, ok := inPaths[m.FilePath()]; ok {
+			continue
+		}
+		if err := r.base.io.Remove(m.FilePath()); err != nil {
+			log.Printf("Warning: failed to delete orphaned merged manifest %s: %v", m.FilePath(), err)
+		}
+	}
+}
+
+func (r *rewriteManifests) record(toRewrite, merged, kept []iceberg.ManifestFile) error {
+	// A prior pass (a superseded OCC attempt) wrote its own merged manifests
+	// that the catalog never referenced. Carry their paths into the
+	// producer's superseded set so commit() can delete them on success;
+	// otherwise every retry leaks a full set of merged .avro files.
+	for _, m := range r.added {
+		r.superseded = append(r.superseded, m.FilePath())
+	}
+	r.added, r.rewritten = nil, nil
+
+	inPaths := make(map[string]struct{}, len(toRewrite))
+	for _, m := range toRewrite {
+		inPaths[m.FilePath()] = struct{}{}
+	}
+	outPaths := make(map[string]struct{}, len(merged))
+	for _, m := range merged {
+		outPaths[m.FilePath()] = struct{}{}
+	}
+
+	// Bins of a single manifest pass through unchanged; only the manifests
+	// that actually appear on one side and not the other are added/replaced.
+	// This drives the returned RewriteManifestsResult.
+	for _, m := range merged {
+		if _, ok := inPaths[m.FilePath()]; !ok {
+			r.added = append(r.added, m)
+		}
+	}
+	for _, m := range toRewrite {
+		if _, ok := outPaths[m.FilePath()]; !ok {
+			r.rewritten = append(r.rewritten, m)
+		}
+	}
+
+	// Guard against a merge dropping or duplicating data files: the rewritten
+	// manifests must hold the same active files as the ones written in their
+	// place. Pass-through bins appear on both sides and cancel, so comparing the
+	// rewritten/added diff is equivalent to comparing the full input/output and
+	// avoids re-walking the pass-throughs. The processed count doubles as the
+	// entries-processed summary value.
+	processed, err := manifestActiveFiles(r.base.io, r.rewritten)
+	if err != nil {
+		return err
+	}
+	written, err := manifestActiveFiles(r.base.io, r.added)
+	if err != nil {
+		return err
+	}
+	if processed != written {
+		return fmt.Errorf("rewrite manifests changed active file count: %d before, %d after", processed, written)
+	}
+
+	// Publish into the shared result the caller holds so a retry's counts win.
+	r.result.RewrittenManifests = r.rewritten
+	r.result.AddedManifests = r.added
+
+	// The summary follows Java's Appendix-F accounting, which counts only the
+	// manifests a merge actually wrote and consumed — not single-manifest bins
+	// that pass through untouched — so a Go-written snapshot reads the same as a
+	// Java one.
+	r.base.snapshotProps[manifestsCreatedKey] = strconv.Itoa(len(r.added))
+	r.base.snapshotProps[manifestsReplacedKey] = strconv.Itoa(len(r.rewritten))
+	r.base.snapshotProps[manifestsKeptKey] = strconv.Itoa(len(kept))
+	r.base.snapshotProps[entriesProcessedKey] = strconv.FormatInt(processed, 10)
+
+	return nil
+}
+
+func (r *rewriteManifests) validate(*conflictContext) error { return nil }
+func (r *rewriteManifests) needsValidation() bool           { return false }
+
+// manifestActiveFiles sums the active (added + existing) data files across
+// manifests. It uses the manifest header counts when present and falls back to
+// reading and counting the live entries when a count is absent — V1 manifests
+// don't populate the count fields, so without the fallback the count guard
+// would silently not run on exactly the tables that most need it.
+func manifestActiveFiles(fs iceio.IO, manifests []iceberg.ManifestFile) (int64, error) {
+	var total int64
+	for _, m := range manifests {
+		added, existing := m.AddedDataFiles(), m.ExistingDataFiles()
+		if added < 0 || existing < 0 {
+			// discardDeleted drops DELETED entries, leaving added + existing.
+			for _, err := range m.Entries(fs, true) {
+				if err != nil {
+					return 0, fmt.Errorf("count active files in manifest %s: %w", m.FilePath(), err)
+				}
+				total++
+			}
+
+			continue
+		}
+		total += int64(added) + int64(existing)
+	}
+
+	return total, nil
+}
+
+// RewriteManifests merges small data manifests in the current snapshot into
+// fewer, target-sized ones and stages the result as a REPLACE snapshot. It
+// rewrites metadata only; no data files are read or written. Delete manifests
+// are left untouched.
+//
+// Manifests are clustered by size only (bin-packed toward the target size);
+// clustering by partition or sort key, which Java exposes via clusterBy, is a
+// future extension.
+//
+// On a V3 table each rewrite advances next-row-id by the eligible manifests'
+// row count even though no rows are written, so running it on a cadence
+// steadily consumes row-ID space.
+//
+// A no-op result (IsNoOp) means there was nothing to do — either the table has
+// no current snapshot (NoOpNoSnapshot) or the eligible manifests are already
+// optimal (NoOpAlreadyOptimal). A no-op writes no manifest list and stages
+// nothing on the transaction, so a following Commit is a true no-op; callers
+// can skip it either way.
+func (t *Transaction) RewriteManifests(ctx context.Context, opts ...RewriteManifestsOpt) (*RewriteManifestsResult, error) {
+	if t.meta.currentSnapshot() == nil {
+		return &RewriteManifestsResult{NoOpReason: NoOpNoSnapshot}, nil
+	}
+
+	cfg := rewriteManifestsCfg{
+		targetSizeBytes: int64(t.meta.props.GetInt(ManifestTargetSizeBytesKey, ManifestTargetSizeBytesDefault)),
+	}
+	for _, o := range opts {
+		o(&cfg)
+	}
+
+	if cfg.specID != nil {
+		if _, err := t.meta.GetSpecByID(*cfg.specID); err != nil {
+			return nil, fmt.Errorf("cannot rewrite manifests: unknown partition spec id %d: %w", *cfg.specID, err)
+		}
+	}
+
+	fs, err := t.tbl.fsF(ctx)
+	if err != nil {
+		return nil, err
+	}
+	wfs, ok := fs.(iceio.WriteFileIO)
+	if !ok {
+		return nil, ErrWriteIORequired
+	}
+
+	prod := newRewriteManifestsProducer(t, wfs, iceberg.Properties{}, cfg)
+	impl, ok := prod.producerImpl.(*rewriteManifests)
+	if !ok {
+		return nil, fmt.Errorf("internal error: unexpected producer type %T", prod.producerImpl)
+	}
+
+	parent, err := prod.parentSnapshot()
+	if err != nil {
+		return nil, err
+	}
+
+	// Plan the merge first (this writes merged manifests only when a real merge
+	// happens) so a no-op is detected before commit writes the manifest list or
+	// apply stages an empty REPLACE. A single-manifest bin passes through and
+	// writes nothing, so an empty diff means nothing was written.
+	newManifests, addedContent, err := prod.buildManifests(ctx, parent)
+	if err != nil {
+		return nil, err
+	}
+	if len(impl.result.AddedManifests) == 0 && len(impl.result.RewrittenManifests) == 0 {
+		impl.result.NoOpReason = NoOpAlreadyOptimal
+
+		return impl.result, nil
+	}
+
+	updates, reqs, err := prod.commitManifests(newManifests, addedContent)
+	if err != nil {
+		return nil, err
+	}
+	if err := t.apply(updates, reqs); err != nil {
+		return nil, err
+	}
+
+	// impl.result aliases the struct record() updates on every pass, so it
+	// reflects the winning attempt after the caller commits the transaction.
+	return impl.result, nil
+}
--- a/table/rewrite_manifests_test.go
+++ b/table/rewrite_manifests_test.go
@@ -1,0 +1,881 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table_test
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet/pqarrow"
+	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
+	"github.com/apache/iceberg-go/table"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// activeFiles sums the added + existing data files across manifests. It mirrors
+// the production accounting in manifestActiveFiles: a manifest that reports a
+// negative (unknown) count fails the test rather than being silently folded in,
+// so a count we can't trust never masquerades as a valid total.
+func activeFiles(t *testing.T, manifests []iceberg.ManifestFile) int {
+	t.Helper()
+	var n int
+	for _, m := range manifests {
+		added, existing := m.AddedDataFiles(), m.ExistingDataFiles()
+		require.False(t, added < 0 || existing < 0,
+			"manifest %s reports an unknown data-file count", m.FilePath())
+		n += int(added) + int(existing)
+	}
+
+	return n
+}
+
+// rewriteArrowSchema is the single-column schema used by the rewrite tests.
+var rewriteArrowSchema = arrow.NewSchema([]arrow.Field{
+	{Name: "id", Type: arrow.PrimitiveTypes.Int32, Nullable: false},
+}, nil)
+
+// writeOneRowParquet writes a one-row Parquet file at path.
+func writeOneRowParquet(t *testing.T, fs iceio.WriteFileIO, path string) {
+	t.Helper()
+
+	bldr := array.NewInt32Builder(memory.DefaultAllocator)
+	defer bldr.Release()
+
+	bldr.AppendValues([]int32{1}, nil)
+	col := bldr.NewArray()
+	defer col.Release()
+
+	rec := array.NewRecordBatch(rewriteArrowSchema, []arrow.Array{col}, 1)
+	defer rec.Release()
+
+	arrTbl := array.NewTableFromRecords(rewriteArrowSchema, []arrow.RecordBatch{rec})
+	defer arrTbl.Release()
+
+	fo, err := fs.Create(path)
+	require.NoError(t, err)
+	require.NoError(t, pqarrow.WriteTable(arrTbl, fo, arrTbl.NumRows(),
+		nil, pqarrow.DefaultWriterProps()))
+}
+
+func rewriteSchema() *iceberg.Schema {
+	return iceberg.NewSchema(1,
+		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int32, Required: true},
+	)
+}
+
+// appendSeparateManifests performs n single-row appends, each in its own
+// commit, leaving n separate data manifests behind (merge must be disabled on
+// the table). It returns the table after the final commit.
+func appendSeparateManifests(t *testing.T, ctx context.Context, tbl *table.Table, fs iceio.WriteFileIO, dir, prefix string, n int) *table.Table {
+	t.Helper()
+	var err error
+	for i := range n {
+		filePath := fmt.Sprintf("%s/%s-%d.parquet", dir, prefix, i)
+		writeOneRowParquet(t, fs, filePath)
+
+		txn := tbl.NewTransaction()
+		require.NoError(t, txn.AddFiles(ctx, []string{filePath}, nil, false))
+		tbl, err = txn.Commit(ctx)
+		require.NoError(t, err)
+	}
+
+	return tbl
+}
+
+// TestRewriteManifests merges several small data manifests into one without
+// changing the data files they reference.
+func TestRewriteManifests(t *testing.T) {
+	ctx := context.Background()
+	dir := t.TempDir()
+	fs := iceio.LocalFS{}
+
+	// Merge disabled, so each append leaves its own manifest behind.
+	meta, err := table.NewMetadata(rewriteSchema(), iceberg.UnpartitionedSpec,
+		table.UnsortedSortOrder, dir, iceberg.Properties{
+			table.ManifestMergeEnabledKey: "false",
+		})
+	require.NoError(t, err)
+
+	cat := &mergeCatalog{meta: meta}
+	ident := table.Identifier{"default", "test_rewrite_manifests"}
+	tbl := table.New(ident, meta, dir+"/metadata/00000.json",
+		func(_ context.Context) (iceio.IO, error) { return fs, nil },
+		cat,
+	)
+
+	const numCommits = 3
+	tbl = appendSeparateManifests(t, ctx, tbl, fs, dir, "data", numCommits)
+
+	before, err := tbl.CurrentSnapshot().Manifests(fs)
+	require.NoError(t, err)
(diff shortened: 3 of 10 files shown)


table/rewrite_manifests.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return nil, err
+	}
+	wfs, ok := fs.(iceio.WriteFileIO)
+	if !ok {
+		return nil, ErrWriteIORequired
+	}
+
+	prod := newRewriteManifestsProducer(t, wfs, iceberg.Properties{}, cfg)
+	updates, reqs, err := prod.commit(ctx)
```
tanmayrauth: This always stages a REPLACE snapshot even when nothing was eligible (merge returns the inputs unchanged, added/rewritten empty) — only the CLI guards against committing the empty snapshot. A library caller doing RewriteManifests + Commit would get an empty REPLACE snapshot. Worth detecting the no-op here, or at least documenting that callers must check the result before committing?

table/rewrite_manifests.go
```diff
@@ -0,0 +1,261 @@
...
+	for _, m := range toRewrite {
+		if _, ok := outPaths[m.FilePath()]; !ok {
+			r.rewritten = append(r.rewritten, m)
+		}
+	}
+
+	r.base.snapshotProps[manifestsCreatedKey] = strconv.Itoa(len(r.added))
+	r.base.snapshotProps[manifestsReplacedKey] = strconv.Itoa(len(r.rewritten))
+	r.base.snapshotProps[manifestsKeptKey] = strconv.Itoa(len(merged) - len(r.added) + len(kept))
+	r.base.snapshotProps[entriesProcessedKey] = strconv.FormatInt(manifestActiveFiles(r.rewritten), 10)
```
tanmayrauth: `manifestActiveFiles` returns -1 on an unknown count, so this can write "entries-processed": "-1" into the summary. Might be cleaner to omit the key when the count is unknown.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil, err
+	}
+
+	if err := validateRewriteFileCounts(toRewrite, merged); err != nil {
+		return nil, err
+	}
+
+	// Capture results and summary once, from the first (synchronous) pass so
+	// they reach the attempt-0 summary; retries reuse that summary.
+	r.once.Do(func() { r.record(toRewrite, merged, kept) })
```
tanmayrauth: The `once.Do` freezes `rewritten`/`added` and the summary props at attempt 0, but on an OCC retry `rebuildFn` re-runs `processManifests` against the fresh parent and writes a different set of merged manifests. So under contention the returned `RewriteManifestsResult` points at attempt-0 manifest files that got superseded/orphaned, and the committed `manifests-created`/`manifests-replaced` counts may not match what was actually written. Is reusing attempt-0 results here intentional, or should the result/summary reflect the winning attempt?

table/rewrite_manifests_test.go
```diff
@@ -0,0 +1,155 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
tanmayrauth, line 1: Could we add coverage for the spec-id filter, the predicate option, delete manifests being left untouched, and the already-optimal no-op path? The OCC-retry rebuild in particular (tied to the once.Do question above) is currently untested.

tanmayrauth (commented):
Nicely done — clean integration with the producer/OCC path and the rebuildFromInheritedOnly hook reads well. A couple of questions and small nits inline, nothing blocking.

table/rewrite_manifests.go
laskoviymishka: I think there's a correctness bug here once an OCC retry fires.

On retry, `commit()` has cleared `ownManifests`, so the rebuild re-runs `processManifests` against the fresh parent and `mergeManifests` writes a brand-new set of merged manifests. But `r.once` already fired on attempt 0, so `record()` is skipped — which means `r.added`/`r.rewritten` and the four `snapshotProps` summary keys still describe the attempt-0 manifests, not the ones actually committed. So both the `RewriteManifestsResult` we hand back and the committed snapshot summary are frozen at values that don't match what's on disk.

I'd drop the `sync.Once` and make `record()` re-entrant — reset `r.added`/`r.rewritten` at the top and call it on every pass. The last pass is the one that commits, so its counts are the ones that should win. wdyt?

table/rewrite_manifests.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+
+	mgr := manifestMergeManager{
+		targetSizeBytes: r.cfg.targetSizeBytes,
+		minCountToMerge: 1,    // force a merge regardless of count
+		mergeEnabled:    true, // explicit op ignores commit.manifest-merge.enabled
+		snap:            r.base,
+	}
+	merged, err := mgr.mergeManifests(toRewrite)
```
laskoviymishka, line 185: Related to the retry path: each `processManifests` call writes fresh merged `.avro` files via `createManifest`, but only the final attempt's manifests end up referenced by the committed snapshot. The earlier attempts' files are never referenced and never cleaned up, so every OCC retry leaks a full set of merged manifests permanently.

The manifest-list orphan handling we already have only covers manifest-list files, not the inner manifest avro. This exists in a milder form in `mergeAppendFiles`, but it's structurally worse here because a rewrite re-merges everything on every retry rather than appending.

Java tracks created manifests and runs `cleanUncommitted` for exactly this. I'd either track the paths we write so they can be cleaned on a superseded attempt, or skip the re-merge when the input already equals the fresh parent. Either works — I'd just want one of them before merge.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,261 @@
...
+
+	r.base.snapshotProps[manifestsCreatedKey] = strconv.Itoa(len(r.added))
+	r.base.snapshotProps[manifestsReplacedKey] = strconv.Itoa(len(r.rewritten))
+	r.base.snapshotProps[manifestsKeptKey] = strconv.Itoa(len(merged) - len(r.added) + len(kept))
+	r.base.snapshotProps[entriesProcessedKey] = strconv.FormatInt(manifestActiveFiles(r.rewritten), 10)
+}
+
+// rebuildFromInheritedOnly tells commit() to drop ownManifests so OCC retries
+// re-merge the fresh parent's manifests rather than the stale rewrite.
+func (r *rewriteManifests) rebuildFromInheritedOnly() bool { return true }
```
laskoviymishka: On your note in the PR description — yeah, I'd express this differently. This reads as a producer-lifecycle concern (how `commit()` should handle a retry) rather than a producer-contract concern (what the producer does), so putting it on the `producerImpl` interface means every implementor has to stub `return false` for something that isn't really about them.

I'd make it a plain field on `snapshotProducer` instead and have `newRewriteManifestsProducer` set it:

```go
type snapshotProducer struct {
    // ...
    rebuildMode bool // re-derive manifests from fresh parent on OCC retry
}
```

Then `commit()` checks `sp.rebuildMode` directly and the three `return false` stubs on `fastAppendFiles`/`overwriteFiles`/`errorOnDeletedEntries` all go away. Keeps the interface about producer behavior and the lifecycle knob where it's actually consumed. wdyt?

cmd/iceberg/rewrite_manifests.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if tbl.CurrentSnapshot() == nil {
+		output.Text("No current snapshot; nothing to rewrite.")
+
+		return
+	}
+
+	var opts []table.RewriteManifestsOpt
+	if cmd.TargetManifestSize > 0 {
+		opts = append(opts, table.WithManifestTargetSize(int(cmd.TargetManifestSize)))
```
laskoviymishka: `cmd.TargetManifestSize` is `int64` but `WithManifestTargetSize` takes `int`, so this truncates on 32-bit builds. Java's target size is a `long` and `ManifestFile.Length()` is `int64`, so I'd make the option `int64` to match — that also drops the cast here. It's a public API, so easier to get right now than to widen later.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+	for _, m := range toRewrite {
+		if _, ok := outPaths[m.FilePath()]; !ok {
+			r.rewritten = append(r.rewritten, m)
+		}
+	}
+
+	r.base.snapshotProps[manifestsCreatedKey] = strconv.Itoa(len(r.added))
+	r.base.snapshotProps[manifestsReplacedKey] = strconv.Itoa(len(r.rewritten))
```
laskoviymishka, line 306: The summary counts diverge from Java's Appendix-F accounting. Java counts `manifests-replaced` as all predicate-matching manifests and `entries-processed` as all live entries across them — including the single-manifest bins that pass through unchanged — whereas here both are computed off `r.rewritten`, which excludes the pass-throughs.

The pass-through optimization itself is correct and I wouldn't touch it; this is purely about what the summary reports. I'd compute these against the eligible set instead: `manifestsReplacedKey = len(toRewrite)`, `manifestsCreatedKey = len(merged)`, and `entriesProcessedKey = manifestActiveFiles(toRewrite)`. That way the summary reflects everything we considered, pass-throughs included, and reads the same as a Java-written snapshot.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,261 @@
...
+	prod := newRewriteManifestsProducer(t, wfs, iceberg.Properties{}, cfg)
+	updates, reqs, err := prod.commit(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if err := t.apply(updates, reqs); err != nil {
+		return nil, err
+	}
+
+	impl := prod.producerImpl.(*rewriteManifests)
```
laskoviymishka: Bare type assertion will panic if the producer impl is ever anything other than `*rewriteManifests`. The function already returns an error, so I'd use comma-ok:

```go
impl, ok := prod.producerImpl.(*rewriteManifests)
if !ok {
    return nil, fmt.Errorf("internal error: unexpected producer type %T", prod.producerImpl)
}
```

cmd/iceberg/rewrite_manifests.go
```diff
@@ -0,0 +1,67 @@
...
+	"os"
+
+	"github.com/apache/iceberg-go/catalog"
+	"github.com/apache/iceberg-go/table"
+)
+
+func runRewriteManifests(ctx context.Context, output Output, cat catalog.Catalog, cmd *RewriteManifestsCmd) {
+	tbl := loadTable(ctx, output, cat, cmd.TableID)
+
+	if tbl.CurrentSnapshot() == nil {
```
laskoviymishka: We've got two different answers for the empty-snapshot case: the CLI returns exit-0 with a "nothing to rewrite" message here, but the library returns a hard error from `RewriteManifests`. That's inconsistent and means the check lives in two places.

I'd pick one. Simplest is to drop this CLI pre-check and have the API return `&RewriteManifestsResult{}, nil` for the no-snapshot case so it flows through the same empty-result path as "nothing eligible." Alternatively, export a sentinel (`ErrNoCurrentSnapshot`, like the existing `ErrSnapshotNotFound`) and have the CLI translate it. Either is fine, but it should be one source of truth.

cmd/iceberg/rewrite_manifests.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if cmd.SpecID >= 0 {
+		opts = append(opts, table.WithRewriteSpecID(cmd.SpecID))
+	}
+
+	tx := tbl.NewTransaction()
+	res, err := tx.RewriteManifests(ctx, opts...)
+	if err != nil {
+		output.Error(fmt.Errorf("rewrite manifests failed: %w", err))
+		os.Exit(1)
```
laskoviymishka: The repo wires `os.Exit` through the `osExit` seam in `exit.go` (`var osExit = os.Exit`) so exit paths stay testable. I'd use `osExit(1)` here and at the commit-failure site below to match the rest of the CLI.

table/rewrite_manifests_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// activeFiles sums the added + existing data files across manifests.
+func activeFiles(t *testing.T, manifests []iceberg.ManifestFile) int {
+	t.Helper()
+	var n int
+	for _, m := range manifests {
+		n += int(m.AddedDataFiles()) + int(m.ExistingDataFiles())
```
laskoviymishka: This helper doesn't mirror the production `manifestActiveFiles`, which returns -1 when a manifest reports an unknown count. Here a -1 widens into a large positive int and silently corrupts the sum, so a test could pass on a count that production would have flagged as unvalidatable. I'd mirror production and `require.False` on a negative count instead of folding it in.

table/rewrite_manifests_test.go
```diff
@@ -0,0 +1,155 @@
...
+	snap := tbl.CurrentSnapshot()
+	require.NotNil(t, snap)
+	assert.Equal(t, table.OpReplace, snap.Summary.Operation)
+	assert.Equal(t, "1", snap.Summary.Properties["manifests-created"])
+	assert.Equal(t, strconv.Itoa(numCommits), snap.Summary.Properties["manifests-replaced"])
+
+	after, err := snap.Manifests(fs)
+	require.NoError(t, err)
+	assert.Len(t, after, 1, "manifests should be merged into one")
+	assert.Equal(t, wantFiles, activeFiles(t, after), "rewrite must preserve the data file count")
```
laskoviymishka, line 162: The happy path is covered well, but the OCC-retry path isn't tested at all — and that's exactly where the `record()`/`sync.Once` staleness bug lives, so right now nothing would catch it. I'd add a case that forces a retry (the `blockingTrackingIO` + `occ_scenario_test.go` setup shows the pattern) and asserts that both the `RewriteManifestsResult` and the committed summary counts match the committed manifests, not attempt 0.

While we're in here, a couple of cheaper gaps worth closing: `WithRewriteSpecID`, `WithRewriteManifestPredicate`, and `WithManifestTargetSize` binning are all unexercised, and this test doesn't prove delete-manifest isolation — adding a delete manifest and asserting it survives untouched would lock that in.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,261 @@
...
+
+	return slices.Concat(merged, kept), nil
+}
+
+// eligible reports whether m is a data manifest selected for rewrite.
+func (r *rewriteManifests) eligible(m iceberg.ManifestFile) bool {
+	if m.ManifestContent() != iceberg.ManifestContentData {
+		return false
+	}
+	if r.cfg.specID != nil && int(m.PartitionSpecID()) != *r.cfg.specID {
```
laskoviymishka, line 218: `WithRewriteSpecID` with a spec ID that doesn't exist in `t.meta.specs` silently matches nothing and rewrites zero manifests, which is hard to distinguish from "already optimal." I'd validate the id against the known specs up front and error on an unknown one. Minor, but it turns a silent no-op into a clear signal.

laskoviymishka (changes requested):
This is a clean, well-scoped first cut — the producer reuses `manifestMergeManager`, the metadata-only REPLACE framing is right, delete manifests are correctly left alone, and the single-manifest pass-through is a legitimate optimization over Java's always-rewrite (correct per spec, don't change it). Happy path is delivered and tested.

I'd hold this before merging though. The `sync.Once` guarding `record()` makes both the returned `RewriteManifestsResult` and the committed snapshot summary go stale the moment an OCC retry fires: `commit()` clears `ownManifests`, the rebuild re-runs `processManifests` against the fresh parent and writes new merged manifests, but `record()` is skipped, so the result and the `manifests-*`/`entries-processed` counts still describe the attempt-0 manifests that never got committed. Dropping the `Once` and making `record()` re-entrant fixes it.

Traveling with that: every retry also leaks its merged `.avro` files with no cleanup (Java's `cleanUncommitted` covers this), and the summary counts diverge from Java's Appendix-F semantics even on the happy path — they should reflect the full eligible set, pass-throughs included.

On your `rebuildFromInheritedOnly` question — I wouldn't make it a 5th interface method. It's a lifecycle concern, not a producer contract, and the `return false` stubs everywhere are the tell. A plain `rebuildMode bool` field on `snapshotProducer` set by `newRewriteManifestsProducer` keeps the interface about behavior; details inline.

- drop `sync.Once`, make `record()` re-entrant so result + summary match the committed manifests
- clean up (or avoid) the orphaned merged manifests on retry
- fix the summary count semantics to match Java's eligible-set accounting
- move `rebuildFromInheritedOnly` off the interface
- widen `WithManifestTargetSize` to `int64`
- add a test for the OCC-retry path (where the headline bug lives)

table/rewrite_manifests.go
rambleraptor: Sounds great!

table/rewrite_manifests.go
rambleraptor: Did the first, thanks!

rambleraptor:
@laskoviymishka made those changes. thanks!

table/rewrite_manifests.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	return total
+}
+
+// validateRewriteFileCounts guards against dropping or duplicating data files:
+// the rewritten manifests must hold the same active files as their inputs.
+func validateRewriteFileCounts(before, after []iceberg.ManifestFile) error {
+	in, out := manifestActiveFiles(before), manifestActiveFiles(after)
+	if in < 0 || out < 0 {
```
laskoviymishka: This guard is doing important work — it's the main thing protecting against a merge accidentally dropping or duplicating data files. The one thing I'd love to tighten: it returns nil the moment a manifest reports an unknown count, and V1 tables don't populate the count fields at all, so on those tables the check quietly doesn't run — which is a shame, because they're exactly the ones where I'd most want it.

Could we fall back to reading the live entries and counting them directly when the header counts are missing? If that feels like too much for this PR, even a loud log plus a godoc note that the guard is a no-op on V1 would keep us honest about what it does and doesn't cover. wdyt?

table/rewrite_manifests.go
```diff
@@ -0,0 +1,297 @@
...
+// rewrites metadata only; no data files are read or written. Delete manifests
+// are left untouched.
+//
+// An empty result (no AddedManifests and no RewrittenManifests) means there
+// was nothing to do — either the table has no current snapshot or the
+// manifests are already optimal. Callers should treat that as a no-op and
+// skip the commit rather than staging an empty REPLACE snapshot.
+func (t *Transaction) RewriteManifests(ctx context.Context, opts ...RewriteManifestsOpt) (*RewriteManifestsResult, error) {
+	if t.meta.currentSnapshot() == nil {
+		return &RewriteManifestsResult{}, nil
```
laskoviymishka: Two small things tangled together here, both about the no-op shape — flagging them now mostly because they're public API and harder to change later.

The no-snapshot case and the already-optimal case both return the same empty `RewriteManifestsResult{}`, so a caller can't tell them apart — the CLI ends up printing "Manifests already optimal" even for an empty table. A sentinel error for no-snapshot (or a little `IsNoOp`/reason on the result) would let callers branch cleanly.

And further down, `RewriteManifests` calls `prod.commit` + `t.apply` unconditionally, so an empty rewrite still stages a REPLACE. The godoc kindly tells callers to skip the commit, but nothing enforces it, so a `RewriteManifests` + `Commit` lands an empty snapshot. Could we make the no-op real — skip `t.apply` when the result is empty — so `Commit` is a true no-op no matter what the caller does?

table/snapshot_producers.go
```diff
@@ -524,6 +524,19 @@ type snapshotProducer struct {
 	deletedFiles       map[string]iceberg.DataFile
 	deletedDeleteFiles map[string]iceberg.DataFile
 	snapshotProps      iceberg.Properties
+
+	// rebuildMode re-derives manifests from the fresh parent on an OCC
+	// retry rather than carrying forward this producer's manifests. It is a
+	// lifecycle knob consumed by commit(), not part of the producer
+	// contract; newRewriteManifestsProducer sets it.
+	rebuildMode bool
```
laskoviymishka: Picking up the question from your PR description about whether there's a cleaner way to express this — I think your earlier instinct toward `rebuildFromInheritedOnly() bool` on `producerImpl` was the right one. As a field on the shared `snapshotProducer`, every other producer ends up carrying `rebuildMode` and `supersededManifests` zero-valued, and `commit()` grows an `if sp.rebuildMode` branch that (as the comment itself notes) isn't really part of the producer contract.

Putting it back behind an interface method, and scoping `supersededManifests` to the `rewriteManifests` type — surfaced to `doCommit` through the `addSnapshotUpdate` you're already threading — would keep the base struct generic. Worth settling before more producers grow on top of it. wdyt?

table/table.go
```diff
@@ -530,6 +530,15 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
 		return nil, err
 	}
 
+	// The commit succeeded. Inner data manifests written by superseded retry
+	// attempts (a rewrite re-merges everything on each retry) are now orphaned
+	// objects; the committed snapshot references only the final attempt's.
+	for _, u := range updates {
+		if su, ok := u.(*addSnapshotUpdate); ok && su.supersededManifests != nil {
+			orphanedManifests = append(orphanedManifests, *su.supersededManifests...)
```
laskoviymishka: This handles the happy path well. The only gap I see is the error path: if `CommitTable` returns something other than `ErrCommitFailed` (a 5xx after a few retries, say), cleanup is skipped, and the manifests written by the earlier attempts — which nothing ever referenced — become orphans no one cleans up.

Definitely fine as a follow-up rather than a blocker. The nice thing is those pre-failure superseded manifests are always safe to delete, unlike the winning attempt's, so cleaning them unconditionally on the error path (or just documenting the leak as a known edge) would close it.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	// Publish into the shared result the caller holds so a retry's counts win.
+	r.result.RewrittenManifests = r.rewritten
+	r.result.AddedManifests = r.added
+
+	// The summary follows Java's Appendix-F accounting over the full eligible
+	// set, pass-throughs included, so it reads the same as a Java-written
+	// snapshot — distinct from the added/rewritten diff above.
+	r.base.snapshotProps[manifestsCreatedKey] = strconv.Itoa(len(merged))
```
laskoviymishka: The comment mentions matching Java's Appendix-F accounting — I think it's close but drifts a little once single-manifest pass-throughs are in play. Java counts created/replaced over the manifests actually written and consumed by a merge, and entries-processed over entries physically read, whereas here a bin of one manifest passes through untouched but still lands in `len(merged)` and `len(toRewrite)`.

Concretely, with 10 eligible manifests where 5 are already above target and 5 merge into 1, this reports created=6/replaced=10 while Java reports created=1/replaced=5. The good news is you already compute the real added/rewritten diff for the result — driving the summary off `len(r.added)`/`len(r.rewritten)` (and counting entries over `r.rewritten`) would make a Go-written snapshot read the same as a Java one. wdyt?

table/rewrite_manifests_test.go
```diff
@@ -0,0 +1,525 @@
...
+	const freshCount = staleCount + 1
+	_ = appendSeparateManifests(t, ctx, tbl, track, dir, "fresh", 1)
+	freshSnap := cat.current.CurrentSnapshot()
+	require.NotNil(t, freshSnap)
+	freshManifests, err := freshSnap.Manifests(track)
+	require.NoError(t, err)
+	require.Len(t, freshManifests, freshCount, "catalog should now hold four manifests")
+
+	// Force exactly one conflict so Writer A retries against the fresh parent.
+	cat.conflictsLeft = 1
```
laskoviymishka, line 829: Really glad this test asserts the orphan cleanup at all — that's the part I'd most worry about. One extension I'd suggest: it only forces a single conflict, and `supersededManifests` is built to accumulate across multiple retries, so the per-generation `record()` accumulation only gets exercised once.

A `conflictsLeft=2` case that checks every superseded generation gets cleaned up would lock that down, and pairing it with an error-path case (inject a `Create` failure on a later retry, per the table.go note) would cover the leak corner too.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,297 @@
...
+
+	cfg := rewriteManifestsCfg{
+		targetSizeBytes: int64(t.meta.props.GetInt(ManifestTargetSizeBytesKey, ManifestTargetSizeBytesDefault)),
+	}
+	for _, o := range opts {
+		o(&cfg)
+	}
+
+	if cfg.specID != nil {
+		if _, err := t.meta.GetSpecByID(*cfg.specID); err != nil {
```
laskoviymishka, line 373: Tiny one: `GetSpecByID`'s error gets dropped and replaced with a fresh string, so the `ErrPartitionSpecNotFound` sentinel is lost and callers can't `errors.Is` it — which is why the test has to string-match on "unknown partition spec id 999".

Wrapping it — `fmt.Errorf("cannot rewrite manifests: unknown partition spec id %d: %w", *cfg.specID, err)` — and switching the test to `errors.Is` would make the failure programmatically distinguishable.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil // counts unknown; can't validate
+	}
+	if in != out {
+		return fmt.Errorf("rewrite manifests changed active file count: %d before, %d after", in, out)
+	}
+
+	return nil
+}
+
+// RewriteManifests merges small data manifests in the current snapshot into
```
laskoviymishka, line 342: One small doc note: Java's `RewriteManifests` exposes `clusterBy(...)` to group by partition/sort key, and this bin-packs by size only. That's a totally reasonable subset for a first cut — I'd just call it out in this godoc (size-only today, partition/sort clustering a future extension) so the contract stays honest until a `WithClusterBy` lands.

laskoviymishka (changes requested):
Thanks for taking this on — manifest rewriting is one of the more delicate maintenance ops and this is a strong first cut. The producer reuses `manifestMergeManager` cleanly, the REPLACE summary keys line up with Java's `SnapshotSummary` constants, and the OCC handling is genuinely well thought through: re-deriving the merge against the fresh parent on retry and tracking superseded manifests for cleanup is the hard part, and it's mostly here, with a test that forces a conflict and checks the winning attempt's counts win. The coverage is broad too — predicate, delete-manifest passthrough, already-optimal, no-snapshot.

I left a handful of inline notes. Most are small, but a few I'd love to settle in this PR since the rest of the maintenance stack will build on the producer contract:

- the count guard (`validateRewriteFileCounts`) quietly no-ops when a manifest reports an unknown count, which is the normal state for V1 tables — so the one real data-safety check sits idle on exactly the tables that most need it. This is the one I'd most want to nail down.
- the no-op contract collapses "no snapshot" and "already optimal" into the same empty result, and `RewriteManifests` stages a REPLACE unconditionally — both are public-API shapes worth getting right now.
- to your own question in the description: I think `rebuildMode` reads more cleanly as the `rebuildFromInheritedOnly()` interface method you'd sketched, with `supersededManifests` scoped to the rewrite type.
- the summary counts drift from Java once single-manifest pass-throughs appear, and the OCC test only forces one conflict.

None of this is far off — the core mechanism is sound, and once the count guard and the two API shapes are sorted I'd be glad to take another pass. Nice work.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return nil, err
+	}
+	wfs, ok := fs.(iceio.WriteFileIO)
+	if !ok {
+		return nil, ErrWriteIORequired
+	}
+
+	prod := newRewriteManifestsProducer(t, wfs, iceberg.Properties{}, cfg)
+	updates, reqs, err := prod.commit(ctx)
```
zeroshade: [Major] `prod.commit(ctx)` writes the `snap-*.avro` manifest-list file while building the staged snapshot, and the `t.apply(updates, reqs)` just below stages the REPLACE update - both run *before* the empty-result check further down. So for a no-op rewrite that still has a current snapshot (already optimal, a valid spec-id matching no manifests, or a predicate selecting nothing), this writes an unreferenced manifest list and stages an empty REPLACE on the transaction. That contradicts this method's own doc comment, which tells callers to skip the commit rather than stage an empty REPLACE snapshot: by the time the CLI sees the empty result and skips `tx.Commit`, the orphan object and the staged update already exist.

Suggest detecting the no-op before `prod.commit`/`t.apply` - run manifest processing first and return the empty `RewriteManifestsResult` when there are no added/rewritten manifests, or have the producer signal a no-op before it writes the manifest list. A test asserting the no-op leaves no new manifest-list object and stages nothing would lock it in.

table/table.go
```diff
@@ -530,6 +530,15 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
 		return nil, err
 	}
 
+	// The commit succeeded. Inner data manifests written by superseded retry
+	// attempts (a rewrite re-merges everything on each retry) are now orphaned
+	// objects; the committed snapshot references only the final attempt's.
+	for _, u := range updates {
```
zeroshade, line 592: [Minor] This success-path cleanup (appending superseded retry manifests to `orphanedManifests`) only runs after the commit succeeds. When retries are exhausted and `CommitTable` returns `ErrCommitFailed`, `doCommit` returns at the `return nil, err` just above (~line 530) before reaching this loop, so the inner data manifests written by superseded retry attempts are never added to cleanup and leak as orphaned objects. Consider appending `su.supersededManifests` on the safe failure path too (e.g. exhausted `ErrCommitFailed`).

zeroshade (changes requested):
The data-manifest selection, merge, and OCC retry/rebuild design are solid, and the test coverage (merge, no-snapshot, already-optimal, unknown spec, predicate, delete-manifest preservation, successful retry) is good. Requesting changes on one structural issue in the no-op path plus a related failure-path cleanup gap; both inline.

table/rewrite_manifests.go
rambleraptor: Yep, sounds good. Made the fix.

table/table.go
rambleraptor: Yeah, I ended up leaving this as a follow-up if that's alright.

table/rewrite_manifests.go
rambleraptor: I ended up doing the full approach. Hope that works.

rambleraptor:
@zeroshade @laskoviymishka thanks for the review! I went ahead and made the changes you requested.

table/snapshot_producers.go
```diff
@@ -60,6 +60,15 @@ type producerImpl interface {
 	// unconditionally a no-op; commit() skips validator registration
 	// entirely when this returns false, so validate will never run.
 	needsValidation() bool
+	// rebuildFromInheritedOnly reports whether OCC retries should re-derive
+	// this producer's manifests from the fresh parent rather than carrying
+	// forward the manifests built on the first attempt. A rewrite re-expresses
+	// inherited manifests, so it rebuilds; appends carry forward. Default false.
+	rebuildFromInheritedOnly() bool
```
laskoviymishka: On the `rebuildFromInheritedOnly` question — yeah, I'd pull these off the shared interface rather than widen it. Both `rebuildFromInheritedOnly()` and `supersededManifests()` are only meaningful for the rewrite producer, but every impl now has to carry them: `fastAppendFiles`, `overwriteFiles`, and `mergeAppendFiles` all stub out to `false`/`nil`.

We already use the optional-capability + type-assertion pattern elsewhere. I'd define two unexported interfaces — something like `inheritedOnlyRebuilder` and `supersededAccumulator` — and type-assert on `producerImpl` at the two existing `commitManifests` call sites (the `ownManifests = nil` clear and the `addSnap.supersededManifests` wire-up). The rewrite producer opts in by implementing them, and the other three impls lose the stubs entirely.

That also lets `supersededManifests` drop the `*[]string` return, which is a little unusual for Go — the accumulator interface can just hand back the paths. wdyt?

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+// Manifests are clustered by size only (bin-packed toward the target size);
+// clustering by partition or sort key, which Java exposes via clusterBy, is a
+// future extension.
+//
+// A no-op result (IsNoOp) means there was nothing to do — either the table has
+// no current snapshot (NoOpNoSnapshot) or the eligible manifests are already
+// optimal (NoOpAlreadyOptimal). A no-op writes no manifest list and stages
+// nothing on the transaction, so a following Commit is a true no-op; callers
+// can skip it either way.
+func (t *Transaction) RewriteManifests(ctx context.Context, opts ...RewriteManifestsOpt) (*RewriteManifestsResult, error) {
```
laskoviymishka, line 360: Not a bug, but worth a line in the godoc here: on a V3 table each rewrite reserves row-ID space equal to the total row count of the eligible manifests, even though no new data is written. The merged manifest goes out with `FirstRowIDValue == nil`, so the manifest-list writer treats it as new and advances `next-row-id` by `ExistingRowsCount + AddedRowsCount`.

That matches Java's `ManifestListWriter.V3Writer` and the existing overcount comment in `snapshot_producers.go`, so it's correct — it's just surprising that a metadata-only rewrite burns ID space, and anyone running this on a cadence should know that going in.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+
+func (r *rewriteManifests) record(toRewrite, merged, kept []iceberg.ManifestFile) {
+	// A prior pass (a superseded OCC attempt) wrote its own merged manifests
+	// that the catalog never referenced. Carry their paths into the
+	// producer's superseded set so commit() can delete them on success;
+	// otherwise every retry leaks a full set of merged .avro files.
+	for _, m := range r.added {
+		r.superseded = append(r.superseded, m.FilePath())
+	}
+	r.added, r.rewritten = nil, nil
```
laskoviymishka, line 254: I think there's a small leak on the total-retry-exhaustion path. `record()` only moves the *previous* attempt's `r.added` into `superseded`, and `doCommit` reads the accumulator after the loop — so on the final failing attempt the merged `.avro` we wrote is never cleaned up. `TestRewriteManifestsCleansSupersededOnExhaustedRetries` even calls this out as "the one allowed leak", but there's no signal in the production code.

Either drop a comment explaining it, or on exhaustion append the last `r.added` to `superseded` too — that's safe, since a snapshot that never committed references nothing. I'd lean toward the latter. wdyt?

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+	r.result.AddedManifests = r.added
+
+	// The summary follows Java's Appendix-F accounting, which counts only the
+	// manifests a merge actually wrote and consumed — not single-manifest bins
+	// that pass through untouched. Driving it off the added/rewritten diff (and
+	// counting entries over the rewritten set) makes a Go-written snapshot read
+	// the same as a Java one.
+	r.base.snapshotProps[manifestsCreatedKey] = strconv.Itoa(len(r.added))
+	r.base.snapshotProps[manifestsReplacedKey] = strconv.Itoa(len(r.rewritten))
+	r.base.snapshotProps[manifestsKeptKey] = strconv.Itoa(len(kept))
```
laskoviymishka, line 307: Small interop gap here in the summary props: Java's `BaseRewriteManifests` hard-codes `setPartitionSummaryLimit(0)`, but we fall back to the table property (`write.summary.partition-limit`, default 0). On a table that sets that property > 0, a Go-written REPLACE would emit per-partition summaries Java never writes for a rewrite. I'd force the limit to 0 in this producer's summary path so the two engines' REPLACE snapshots stay identical.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+
+	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
+)
+
+// Snapshot summary keys for a manifest rewrite.
+const (
+	manifestsCreatedKey  = "manifests-created"
+	manifestsReplacedKey = "manifests-replaced"
+	manifestsKeptKey     = "manifests-kept"
```
laskoviymishka: Two of these counts are correct but non-obvious, and I'd document them on the const block so nobody "fixes" them later.

`manifests-kept` includes delete manifests, not just unmerged data manifests — that actually matches Java (`BaseRewriteManifests` routes `containsDeletes` manifests into `keptManifests`, and `KEPT_MANIFESTS_COUNT` includes them), so it's fine for interop. But it's worth a one-liner here, plus a test asserting the count on a table with a delete manifest present, to lock the behavior in.

`entries-processed` counts data-file entries, not rows — also worth spelling out, since the name reads ambiguously.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+	// manifests a merge actually wrote and consumed — not single-manifest bins
+	// that pass through untouched. Driving it off the added/rewritten diff (and
+	// counting entries over the rewritten set) makes a Go-written snapshot read
+	// the same as a Java one.
+	r.base.snapshotProps[manifestsCreatedKey] = strconv.Itoa(len(r.added))
+	r.base.snapshotProps[manifestsReplacedKey] = strconv.Itoa(len(r.rewritten))
+	r.base.snapshotProps[manifestsKeptKey] = strconv.Itoa(len(kept))
+	// Best-effort: the rewritten set was just read by validateRewriteFileCounts,
+	// so a count here should succeed; drop the key rather than fail if it can't.
+	if entries, err := manifestActiveFiles(r.base.io, r.rewritten); err == nil {
```
laskoviymishka: `record()` recomputes the active-file count over `r.rewritten` here, but `validateRewriteFileCounts` already walked the same manifests a few lines up. On a V1 table (no cached `AddedDataFiles`/`ExistingDataFiles`) that's a full second pass over every entry, on every retry attempt. Could we thread the count out of `validateRewriteFileCounts` and reuse it for `entries-processed` instead of recomputing?

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+)
+
+// NoOpReason explains why a rewrite changed nothing. It lets callers tell a
+// table with no current snapshot apart from one whose manifests are already
+// optimal, which would otherwise both surface as an empty result.
+type NoOpReason string
+
+const (
+	// NoOpNone means the rewrite produced changes (not a no-op).
+	NoOpNone NoOpReason = ""
```
laskoviymishka: `NoOpReason` is exported with `NoOpNone = ""`, so the zero value is indistinguishable from "unset" — and since this is public API it's awkward to change later. I'd either make it an `int`/`iota` type or give `NoOpNone` a real value like `"none"`.

While we're here, `IsNoOp()` keys off the slice lengths rather than the reason. If `NoOpReason` were the single source of truth for both, the two couldn't drift out of sync — worth considering.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+// RewriteManifestsOpt configures [Transaction.RewriteManifests].
+type RewriteManifestsOpt func(*rewriteManifestsCfg)
+
+// WithManifestTargetSize overrides the target manifest size in bytes.
+// The default comes from the commit.manifest.target-size-bytes property.
+func WithManifestTargetSize(size int64) RewriteManifestsOpt {
+	return func(c *rewriteManifestsCfg) {
+		if size > 0 {
```
laskoviymishka, line 96: `WithManifestTargetSize` silently drops a size `<= 0`. The CLI guards against that, but a library caller passing 0 just gets the default with no signal. I'd at least document the "≤0 is ignored" contract in the godoc — or return an error if that reads cleaner.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+		minCountToMerge: 1,    // force a merge regardless of count
+		mergeEnabled:    true, // explicit op ignores commit.manifest-merge.enabled
+		snap:            r.base,
+	}
+	merged, err := mgr.mergeManifests(toRewrite)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := validateRewriteFileCounts(r.base.io, toRewrite, merged); err != nil {
```
laskoviymishka: There's no coverage for `mergeManifests`/`createManifest` or `validateRewriteFileCounts` returning an error. The merged `.avro` files are written before `validateRewriteFileCounts` runs, so an error here orphans them — and the cleanup behavior on that path is both unspecified and untested. I'd add a test that forces the count mismatch and asserts what happens to the staged files.

table/rewrite_manifests_test.go
```diff
@@ -0,0 +1,833 @@
...
+		tbl, err = txn.Commit(ctx)
+		require.NoError(t, err)
+	}
+
+	return tbl
+}
+
+// TestRewriteManifests merges several small data manifests into one without
+// changing the data files they reference.
+func TestRewriteManifests(t *testing.T) {
```
laskoviymishka, line 113: Every test here uses the default target size, which is far larger than the tiny test manifests — so every rewrite produces exactly one output manifest, and the 3-inputs→2-outputs case (`manifests-created = 2`) is never actually exercised. Could we add a case with `WithManifestTargetSize` small enough to force more than one output bin? That's the path most likely to regress silently.

table/snapshot_producers.go
```diff
@@ -973,6 +1004,10 @@ func (sp *snapshotProducer) commit(ctx context.Context) (_ []Update, _ []Require
 	snapshotID := sp.snapshotID
 	commitUUID := sp.commitUuid
 	capturedSnapshot := snapshot // copy the value so the closure is self-contained
+	// Capture the live snapshotProps map. processManifests updates it on each
+	// retry (a rewrite re-merges the fresh parent), so the rebuilt summary
+	// below can reflect the winning attempt's operation counts.
+	liveProps := sp.snapshotProps
```
laskoviymishka: `liveProps` is a live alias into the map, not a copy — which is intentional, so `maps.Copy(deltaSummary.Properties, liveProps)` picks up the counts `record()` wrote on the current attempt. It's correct, but fragile to a future reader; a one-line comment on the capture noting it's deliberately live would save someone a debugging session.

laskoviymishka (changes requested):
Really nice to see `RewriteManifests` land on both the CLI and the library side, and the test setup deserves a callout — exercising the real retry machinery (`doCommit`, `rebuildSnapshotUpdates`, the actual `.avro` writes) instead of mocking it is exactly right, and it's what gives me confidence in the OCC-retry paths. I checked the V3 row-lineage interaction and the manifests-kept-includes-deletes behavior against Java and both hold up.

On the `rebuildFromInheritedOnly` question you raised — yeah, I think there's a cleaner way, and it's the one thing I'd settle before merge. Right now both new methods live on the shared `producerImpl` interface even though they only mean anything for the rewrite path, so `fastAppendFiles`/`overwriteFiles`/`mergeAppendFiles` all carry `false`/`nil` stubs. We already use the optional-capability + type-assertion pattern elsewhere; a couple of unexported interfaces asserted at the two `commitManifests` call sites keeps the rewrite-specific logic contained. Details inline.

Everything else is polish. A few things I'd like to settle before merge:

* Extract the rewrite-only methods into optional interfaces (this is the answer to your question)
* Decide on `NoOpReason`'s zero value while it's still easy to change — it's public API
* Document the V3 next-row-id space reservation and the deliberately-live `liveProps` capture
* Add the missing multi-output-bin test and settle cleanup on the merge-error path

None of these are big. Once they're addressed, happy to take another pass and approve.

table/table.go
```diff
@@ -526,6 +526,20 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
...
+	// re-merges everything on each retry) are orphaned objects: on success the
+	// committed snapshot references only the final attempt's, and on a failed
+	// commit nothing references any of them. They are always safe to delete —
+	// the winning attempt's manifests are never added to this set — so collect
+	// them on both the success and the safe-failure (exhausted ErrCommitFailed)
+	// paths. The defer skips cleanup only on the unsafe non-ErrCommitFailed
+	// path, which already returned above with cleanupOrphans = false.
+	for _, u := range updates {
+		if su, ok := u.(*addSnapshotUpdate); ok && su.supersededManifests != nil {
+			orphanedManifests = append(orphanedManifests, *su.supersededManifests...)
```
zeroshade: This only collects `supersededManifests`, but `record()` only moves the previous attempt's `r.added` into that slice (table/rewrite_manifests.go:200-203). When retries are exhausted with `ErrCommitFailed`, the final failed attempt's merged manifest was written but is never moved into `supersededManifests`, so this cleanup misses it; the new test currently codifies that final leak as allowed. Please also clean the current attempt's uncommitted rewrite outputs on the safe exhausted-failure path.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	mgr := manifestMergeManager{
+		targetSizeBytes: r.cfg.targetSizeBytes,
+		minCountToMerge: 1,    // force a merge regardless of count
+		mergeEnabled:    true, // explicit op ignores commit.manifest-merge.enabled
+		snap:            r.base,
+	}
+	merged, err := mgr.mergeManifests(toRewrite)
+	if err != nil {
```
zeroshade, line 186: `mergeManifests` can write merged manifests before `validateRewriteFileCounts` runs. If validation (or any later path before `record()`) returns an error, those newly created paths have not been registered in `added`/`superseded`, so they are orphaned. Please register the created manifest paths immediately and then hand cleanup ownership to the commit/retry mechanism.

table/snapshot_producers.go
```diff
@@ -60,6 +60,15 @@ type producerImpl interface {
 	// unconditionally a no-op; commit() skips validator registration
 	// entirely when this returns false, so validate will never run.
 	needsValidation() bool
+	// rebuildFromInheritedOnly reports whether OCC retries should re-derive
+	// this producer's manifests from the fresh parent rather than carrying
+	// forward the manifests built on the first attempt. A rewrite re-expresses
```
zeroshade: These hooks are rewrite-only behavior, but adding them to `producerImpl` forces every producer and test fake to implement no-op methods. Please keep the shared interface focused on common producer behavior and use unexported optional capability interfaces at the call sites that need rewrite-specific retry/cleanup behavior.

table/rewrite_manifests.go
zeroshade: `NoOpNone` is the empty-string zero value while `IsNoOp()` derives no-op state from slice lengths, so the explicit reason can drift from the result state and the zero value is indistinguishable from unset. Please make this an iota-style enum with a concrete `NoOpNone`, and make `NoOpReason` the single source of truth for whether the result is a no-op.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,368 @@
...
+
+// RewriteManifestsOpt configures [Transaction.RewriteManifests].
+type RewriteManifestsOpt func(*rewriteManifestsCfg)
+
+// WithManifestTargetSize overrides the target manifest size in bytes.
+// The default comes from the commit.manifest.target-size-bytes property.
+func WithManifestTargetSize(size int64) RewriteManifestsOpt {
+	return func(c *rewriteManifestsCfg) {
+		if size > 0 {
+			c.targetSizeBytes = size
```
zeroshade, line 97: `WithManifestTargetSize` silently ignores zero or negative sizes. Please either document that non-positive values mean 'use the table property' or return an error/validate earlier so invalid caller input is not silently accepted.

zeroshade (changes requested):
Re-review: the prior no-op/empty-REPLACE concern is addressed, but cleanup-on-failure/error paths still block this change.

- Final exhausted ErrCommitFailed still leaves the last attempt's merged manifest unreferenced and uncleaned.
- Errors after mergeManifests but before record() can orphan newly written manifests.
- Please keep rewrite-specific retry hooks out of the shared producer interface, and tighten the NoOp/target-size API behavior.

table/snapshot_producers.go
rambleraptor: I took @laskoviymishka from earlier and followed their pattern. This should match what you're talking about as well. Thanks for the suggestion!

table/snapshot_producers.go
rambleraptor: Sounds great. @zeroshade had a similar suggestion. Made that change and removed the interface.

table/rewrite_manifests.go
rambleraptor: Great, I'll add the last r.added to superseded and call that good!

table/rewrite_manifests.go
rambleraptor: Ack, sounds good.

table/rewrite_manifests.go
rambleraptor: Yep, that's a great call.

table/rewrite_manifests.go
rambleraptor: Yeah, the reason should be the source of truth. I don't like when array lengths end up having magical meanings either.

table/rewrite_manifests.go
rambleraptor: I documented the issue, since returning the error will end up changing the interface.

table/rewrite_manifests.go
rambleraptor: Yep, great call. I ended up putting this in a doc, since I didn't want to change the interface to include an error.

table/rewrite_manifests.go
rambleraptor: Added that test. I also specified the cleanup behavior.

table/rewrite_manifests_test.go
rambleraptor: Added the test.

table/snapshot_producers.go
rambleraptor: Cool, added.

table/rewrite_manifests.go
rambleraptor: Yeah, that's a valid concern. I end up biasing towards "if Java does it, it's fine + doesn't need explanation", but I added the comment.

table/rewrite_manifests.go
rambleraptor: @laskoviymishka mentioned the same thing. Fixed.

table/table.go
rambleraptor: Yep, we should handle that. (I really hate the all of the edge cases we have to deal with for retries...)

Made that change. @laskoviymishka mentioned that too, so we're in the clear.

table/rewrite_manifests.go
rambleraptor: This sounds like what @laskoviymishka mentioned in https://github.com/apache/iceberg-go/pull/1283#discussion_r3515619582. Does that match your understanding?

I added a test called `TestRewriteManifestsCleansOrphanOnValidationError` and then had a defer that calls deleteWritten. Will that suffice? It's slightly different than what you described, but I think it'll handle the situation.

rambleraptor:
@laskoviymishka @zeroshade thanks so much for your diligent reviews. Sorry this has taken so long. The retry logic is... nuanced to say the least. I'm glad we're getting it right.

table/rewrite_manifests.go
```diff
@@ -0,0 +1,424 @@
...
+
+	mgr := manifestMergeManager{
+		targetSizeBytes: r.cfg.targetSizeBytes,
+		minCountToMerge: 1,    // force a merge regardless of count
+		mergeEnabled:    true, // explicit op ignores commit.manifest-merge.enabled
+		snap:            r.base,
+	}
+	merged, err := mgr.mergeManifests(toRewrite)
+	if err != nil {
+		return nil, err
```
zeroshade, line 187: If `mergeManifests` errors after one or more bins have already written merged `.avro` files (bins are written concurrently in `mergeGroup`), this early return leaks them — the `deleteWritten` defer is only installed after a successful return a few lines below. This is the remaining orphan-manifest cleanup gap from the earlier review.

zeroshade (changes requested):
Most of the earlier feedback is addressed — one cleanup path still leaks (inline).

table/rewrite_manifests.go
rambleraptor: Ugh, yeah, that's a great point. Thanks for pointing that out. I added a test for that as well.

(There was also a pretty big rebase)

table/snapshot_producers.go
```diff
@@ -444,12 +454,32 @@ func (m *manifestMergeManager) mergeGroup(firstManifest iceberg.ManifestFile, sp
 	}
 
 	if err := g.Wait(); err != nil {
+		// Delete bins that finished before this one failed.
+		m.removeOrphans(manifests, slices.Concat(binResults...))
```
zeroshade, line 458: Minor / non-blocking, and pre-existing — flagging for a possible follow-up rather than this PR.

This correctly cleans up bins that finished before a sibling failed. The one case it can't reach is a bin that fails *inside* `createManifest` after `newManifestWriter` has already opened its output file: that path never makes it into `binResults`, so its partial `.avro` is left behind. Since `createManifest` is shared with `mergeAppendFiles` and predates this change, I'd treat it as a separate follow-up, not a blocker here.

zeroshade (approved):
Re-review — approving.

The one remaining blocker from my last pass, the concurrent-bin merge leak (`mergeManifests` returning after sibling bins already wrote their `.avro` files), is fixed correctly: `mergeGroup` cleans up the bins that finished before the failing one via `removeOrphans` before returning the error, and `mergeManifests` drops earlier groups' outputs when a later group fails. `TestRewriteManifestsCleansOrphansOnMergeFailure` covers it directly, and the exhausted-retry path is sound too — `doCommit` accumulates orphans (including the final failed attempt via `supersededManifests(false)`) before the error return and deletes them in the defer, while a non-`ErrCommitFailed` error correctly skips cleanup so a snapshot the catalog may have accepted isn't removed.

The two threads still showing as unresolved are both stale after the rebase — the bare type assertion is now the comma-ok form, and `validateRewriteFileCounts` is gone with the file-count guard folded into `record()` — so I've resolved them with notes. I also confirmed the rebase dropped `rebuildFromInheritedOnly` from the shared `producerImpl` interface in favor of the `supersededSource` capability interface, which was the earlier ask.

Left one minor, pre-existing note inline (not blocking).

Holding the actual merge until @laskoviymishka gets a final look, since they drove most of the detailed review here.

table/rewrite_manifests.go
zeroshade: Resolving — addressed by the rebase. This is now the comma-ok form with an error return (`impl, ok := prod.producerImpl.(*rewriteManifests)`) at `table/rewrite_manifests.go:388-391`, so there's no bare-assertion panic path anymore.

table/rewrite_manifests.go
zeroshade: Resolving — the rebase removed `validateRewriteFileCounts` entirely, so the double walk this flagged is gone. The active-file guard is now a single pass inside `record()` (`table/rewrite_manifests.go:285-295`).

laskoviymishka (approved):
LGTM! This one quite huge

PR #1284, opened 2026-06-23
test: pin geo type JSON wire format

### LINKED ISSUE #1271: test: pin geo type JSON wire format with a cross-implementation fixture
Follow-up to #984.

`TestGeographyType` and the geo schema round-trips assert with `assert.JSONEq`, which is key-order- and whitespace-agnostic — so nothing pins the byte-exact type-string emission that other Iceberg clients have to parse. The only geo fixtures we ship are the Parquet files under `table/testdata/geo/` from `apache/parquet-testing`, and those exercise the Parquet read path, not the Iceberg geo type-string JSON.

Add a fixture test that loads a Java/PyIceberg-emitted schema JSON with a `geography` field carrying a non-default CRS + algorithm (e.g. `geography(srid:4269, karney)`) and a `geometry` field with a custom CRS, unmarshal/re-marshal it, and compare the emitted type string byte-for-byte — not `JSONEq` — against the checked-in fixture. Document the producing implementation and version on the fixture, the way `puffin/dv_golden_test.go` does for the DV blob.

Locking the wire format before more geo PRs build on it is newcomer-sized: no domain depth, clear pass/fail.

Revanth14:
## Summary

Adds a cross-implementation schema JSON fixture to pin the Iceberg geo type-string wire format.

The fixture includes:

- `geometry(srid:3857)`
- `geography(srid:4269, karney)`

The new test loads the fixture, unmarshals and re-marshals it through Go, then compares the raw emitted JSON `type` string literals byte-for-byte instead of using `assert.JSONEq`.

## Why

Existing geo tests validated JSON equivalence, but did not pin the exact type strings that other Iceberg clients parse. This locks the current wire format before more geo-related work builds on it.

Fixes #1271.

## Testing

```sh
go test ./...

### CODE REVIEW

File: schema_test.go [hunk 1]
```diff
@@ -19,6 +19,8 @@ package iceberg_test
```
tanmayrauth, line 19: The fixture covers the two-arg forms (geometry(srid), geography(srid, algo)) but not the single-arg geography branches — geography(crs) with the default algorithm, and geography(OGC:CRS84, algo) with the default CRS. Those are exactly the cases where Go's Type() drops the default, so they're the most valuable to pin cross-impl. Worth adding fields for them while you're capturing Java output? Totally optional.

tanmayrauth (commented):
LGTM,clean test-only addition, and the layering is nice: the map comparison pins round-trip stability while the hardcoded literals pin the absolute wire format.

schema_test.go
Revanth14: Thanks, good call. I added both single-arg geography cases to the Java fixture:

- `geography(srid:4269)` for custom CRS + default algorithm
- `geography(OGC:CRS84, karney)` for default CRS + custom algorithm

Also added byte-exact assertions for both emitted type strings.

schema_test.go
```diff
@@ -28,6 +30,26 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// geoSchemaJavaFixtureName is a schema JSON fixture emitted by Apache Iceberg
+// Java SchemaParser.toJson on apache/iceberg main at
+// df6ea0820030d217d5f645469419cdbb3f54a41f.
```
laskoviymishka: This commit isn't the one that defines geo serialization — `df6ea082` is "Data: Add metrics reporter to generic scan builder (#16664)", which touches `IcebergGenerics.java` and has nothing to do with `SchemaParser` or geo types. It looks like it was just the tip of main when the fixture was generated.

The geo format was actually added in `e1e0a740` (#12346, "Add geometry and geography types support") and last touched in `8de302bf`. I'd cite one of those instead so a reader can actually verify the fixture against the source — e.g. "geo type serialization is stable since e1e0a740 (#12346)". The values themselves are right; it's only the citation that's misleading.

schema_test.go
```diff
@@ -28,6 +30,26 @@ import (
...
+//	    Types.NestedField.optional(2, "geom", Types.GeometryType.of("srid:3857")),
+//	    Types.NestedField.optional(3, "geog",
+//	        Types.GeographyType.of("srid:4269", EdgeAlgorithm.KARNEY)),
+//	    Types.NestedField.optional(4, "geog_default_algo",
+//	        Types.GeographyType.of("srid:4269")),
+//	    Types.NestedField.optional(5, "geog_default_crs",
+//	        Types.GeographyType.of("OGC:CRS84", EdgeAlgorithm.KARNEY))))
+//
+// Java and Go write object keys in different orders, so the fixture test pins
+// the raw JSON string literals for field types instead of whole-object bytes.
```
laskoviymishka, line 55: This pins the Java/Go form, but I'd be careful calling it cross-implementation. PyIceberg emits *quoted* CRS — `geometry('srid:3857')` — and its `GEOMETRY_REGEX` requires the quotes, so it won't parse the unquoted form this fixture pins, and conversely Go's regex would capture `'srid:3857'` (quotes and all) from PyIceberg's output. That's a real bidirectional divergence.

The filename already scopes this to Java, which is good. I'd just add a sentence here noting PyIceberg differs so nobody reads this as the canonical cross-client format. Might also be worth a tracking issue for standardizing the CRS quoting across clients — wdyt?

schema_test.go
```diff
@@ -1177,6 +1199,47 @@ func TestSchemaWithGeometryGeographyTypes(t *testing.T) {
...
+
+	var schema iceberg.Schema
+	require.NoError(t, json.Unmarshal(fixtureBytes, &schema))
+
+	freshBytes, err := json.Marshal(&schema)
+	require.NoError(t, err)
+
+	fixtureTypes := rawTopLevelFieldTypes(t, fixtureBytes)
+	freshTypes := rawTopLevelFieldTypes(t, freshBytes)
+	require.Equal(t, fixtureTypes, freshTypes)
```
laskoviymishka: This `require.Equal` is doing almost nothing once the four literal asserts below it are in place — those already pin every `freshTypes` value against the exact fixture string. In the realistic failure mode (Go drops the space and emits `geography(srid:4269,karney)`) the literal asserts catch it directly and point at the field, while the map compare just dumps a diff.

I'd either drop it and keep the literal asserts, or add a `require.Equal(t, 5, len(freshTypes))` so it earns its place by catching a silently-dropped field on round-trip (which the literal asserts won't). The latter is probably the more useful version.

schema_test.go
```diff
@@ -1177,6 +1199,47 @@ func TestSchemaWithGeometryGeographyTypes(t *testing.T) {
...
+			Type json.RawMessage `json:"type"`
+		} `json:"fields"`
+	}
+	require.NoError(t, json.Unmarshal(data, &schema))
+	require.NotEmpty(t, schema.Fields)
+
+	types := make(map[string]string, len(schema.Fields))
+	for _, field := range schema.Fields {
+		require.NotEmpty(t, field.Name)
+		require.NotEmptyf(t, field.Type, "field %q is missing type", field.Name)
```
laskoviymishka, line 1253: `field.Type` is a `json.RawMessage` (`[]byte`), so a JSON `null` token is four non-empty bytes and sails through `NotEmptyf` while being semantically empty. If you want this guard to mean what it reads as, `require.NotEqualf(t, "null", string(field.Type), ...)` alongside the length check would actually catch a null type.

testdata/geo-schema-java-main.json
```diff
@@ -0,0 +1 @@
+{"type":"struct","schema-id":17,"fields":[{"id":1,"name":"id","required":true,"type":"long"},{"id":2,"name":"geom","required":false,"type":"geometry(srid:3857)"},{"id":3,"name":"geog","required":false,"type":"geography(srid:4269, karney)"},{"id":4,"name":"geog_default_algo","required":false,"type":"geography(srid:4269)"},{"id":5,"name":"geog_default_crs","required":false,"type":"geography(OGC:CRS84, karney)"}]}
```
laskoviymishka: All four geo cases here carry params, but the most common form — bare `geometry`/`geography` with the default CRS omitted (Java's `GeometryType.crs84().toString()` returns just `"geometry"`) — isn't pinned anywhere. A regression turning bare `GeometryType{}.Type()` into `"geometry()"` would slip right through.

I'd add two more fields (a plain `geometry` and a plain `geography`) plus their asserts so the default path is locked down too.

laskoviymishka (changes requested):
Nice addition — this locks down the exact `geometry(...)`/`geography(...)` type-string literals that Java's `SchemaParser.toJson` emits, which is exactly the kind of thing that silently drifts when someone tweaks `String()` formatting. The four parameterized cases you picked are right.

Two things make the fixture hard to trust as a reference, though, and since the whole point is to be the thing others build on, I'd fix them before merge. The cited Java commit `df6ea082...` isn't the geo-serialization commit — it's a metrics-reporter PR (#16664), so anyone tracing the provenance hits a dead end. And the comment frames this as a cross-implementation format, but PyIceberg emits *quoted* CRS (`geometry('srid:3857')`) and won't parse the unquoted Java/Go form — so "cross-implementation" overstates what's pinned here.

On the test itself: the top-level `require.Equal` is mostly carried by the four literal asserts below it, and the most common wire form — bare `geometry`/`geography` with default CRS — isn't covered. Both worth tightening.

Fix the commit citation and the cross-impl note, close the default-case gap, and this is good to land.

schema_test.go
Revanth14: Good catch, thanks. Updated the fixture comment to cite the geo-related Iceberg commits instead of the unrelated main tip: `e1e0a740` for the original geo type support and `8de302bf` as the latest geo type-string touchpoint when the fixture was captured.

schema_test.go
Revanth14: Agreed. I added a note that this fixture is scoped to Java's type-string form and that PyIceberg currently emits quoted CRS values, so this shouldn't be read as the canonical cross-client format.

The CRS quoting divergence is worth tracking separately. I'll open an issue for standardizing it across clients and link it here. Kept this PR focused on pinning the Java/Go fixture behavior.

schema_test.go
Revanth14: Done. I replaced the broad map equality with explicit expected type strings for each field, plus `require.Len` checks on both fixture and fresh maps so a silently dropped or extra field still fails clearly.

schema_test.go
Revanth14: Done. Kept the existing non-empty guard and added an explicit `"null"` check so a raw JSON null type fails as missing instead of passing as non-empty bytes.

testdata/geo-schema-java-main.json
Revanth14: Done. Added plain `geom_default` and `geog_default` fields to the Java fixture and pinned their exact emitted type strings as `"geometry"` and `"geography"` in the test.

Revanth14:
> Nice addition — this locks down the exact `geometry(...)`/`geography(...)` type-string literals that Java's `SchemaParser.toJson` emits, which is exactly the kind of thing that silently drifts when someone tweaks `String()` formatting. The four parameterized cases you picked are right.
> 
> Two things make the fixture hard to trust as a reference, though, and since the whole point is to be the thing others build on, I'd fix them before merge. The cited Java commit `df6ea082...` isn't the geo-serialization commit — it's a metrics-reporter PR (#16664), so anyone tracing the provenance hits a dead end. And the comment frames this as a cross-implementation format, but PyIceberg emits _quoted_ CRS (`geometry('srid:3857')`) and won't parse the unquoted Java/Go form — so "cross-implementation" overstates what's pinned here.
> 
> On the test itself: the top-level `require.Equal` is mostly carried by the four literal asserts below it, and the most common wire form — bare `geometry`/`geography` with default CRS — isn't covered. Both worth tightening.
> 
> Fix the commit citation and the cross-impl note, close the default-case gap, and this is good to land.

Thanks for the detailed pass. I pushed the follow-up with the provenance fixed to the geo-related Iceberg commits, added the PyIceberg quoted-CRS note so the fixture is clearly scoped to Java's form, replaced the broad map equality with explicit per-field byte-string assertions plus length checks, and added bare `geometry` / `geography` default cases to the Java fixture.

laskoviymishka (approved):

PR #1285, opened 2026-06-23
test: cover Glue and Hive purge table

### LINKED ISSUE #1272: test: cover Glue and Hive PurgeTable file deletion
Follow-up to #1104.

`glue.PurgeTable` and `hive.PurgeTable` both call `DropTable` then `tbl.PurgeFiles`, swallowing a `PurgeFiles` error with a warning log — and neither path has a test. `TestGlueDropTable` / `TestHiveDropTable` exercise `DropTable` alone and never plumb a `FileIO`, so `PurgeFiles` is never invoked under the mocks. `catalog/sql`'s `TestPurgeTable` / `TestPurgeTableGCDisabled` and `catalog/rest`'s `TestPurgeTable204` already show the shape.

Add `TestGluePurgeTable` and `TestHivePurgeTable` that load a table backed by an in-memory or local `FileIO` with real data files, call `PurgeTable`, and assert both that the drop fires and that the files are physically gone. Include a case where `PurgeFiles` fails, to lock in the best-effort swallow-and-log behavior.

Revanth14:
## Summary

Adds unit coverage for `PurgeTable` in the Glue and Hive catalogs.

The new tests verify that:

- `PurgeTable` loads the table and drops the catalog entry
- local table metadata/data files are physically deleted
- `PurgeFiles` failures are swallowed after the drop
- the warning path is logged when file purge fails

## Why

`glue.PurgeTable` and `hive.PurgeTable` already call `DropTable` followed by best-effort `tbl.PurgeFiles`, but only `DropTable` had test coverage. These tests lock in the intended purge behavior and the existing swallow-and-log error handling.

Fixes #1272.

## Testing

```sh
go test ./catalog/glue ./catalog/hive
go test ./...

### DIFF
--- a/catalog/glue/glue_test.go
+++ b/catalog/glue/glue_test.go
@@ -488,7 +488,20 @@ func TestGluePurgeTableSwallowsPurgeFilesError(t *testing.T) {
 	assert := require.New(t)
 	ctx := context.Background()
 	const scheme = "gluepurgefail"
-	failingFS := failRemoveIO{MemFS: iceio.NewMemFS(), err: errGluePurgeRemove}
+	dropCalled := false
+	removeBeforeDrop := false
+	removeCalls := 0
+	failingFS := failRemoveIO{
+		MemFS: iceio.NewMemFS(),
+		err:   errGluePurgeRemove,
+		onRemove: func() {
+			removeCalls++
+			if !dropCalled {
+				removeBeforeDrop = true
+			}
+		},
+	}
+	iceio.Unregister(scheme)
 	iceio.Register(scheme, func(context.Context, *url.URL, map[string]string) (iceio.IO, error) {
 		return failingFS, nil
 	})
@@ -497,8 +510,51 @@ func TestGluePurgeTableSwallowsPurgeFilesError(t *testing.T) {
 	tableLocation := scheme + "://bucket/test_database/test_table"
 	metadataLocation := tableLocation + "/metadata/v1.metadata.json"
 	dataFile := tableLocation + "/data/file.parquet"
-	writeGluePurgeTableMetadata(t, failingFS, tableLocation, metadataLocation)
-	require.NoError(t, failingFS.WriteFile(dataFile, []byte("data")))
+	writeGluePurgeTableMetadata(t, failingFS, tableLocation, metadataLocation, nil)
+	assert.NoError(failingFS.WriteFile(dataFile, []byte("data")))
+	glueTable := gluePurgeTable(metadataLocation)
+
+	mockGlueSvc := &mockGlueClient{}
+	mockGlueSvc.On("GetTable", mock.Anything, &glue.GetTableInput{
+		DatabaseName: aws.String("test_database"),
+		Name:         aws.String("test_table"),
+	}, mock.Anything).Return(&glue.GetTableOutput{Table: &glueTable}, nil).Twice()
+	mockGlueSvc.On("DeleteTable", mock.Anything, &glue.DeleteTableInput{
+		DatabaseName: aws.String("test_database"),
+		Name:         aws.String("test_table"),
+	}, mock.Anything).Run(func(mock.Arguments) {
+		dropCalled = true
+	}).Return(&glue.DeleteTableOutput{}, nil).Once()
+
+	var logs bytes.Buffer
+	originalLogOutput := log.Writer()
+	// This test redirects the process-global logger, so keep it non-parallel.
+	log.SetOutput(&logs)
+	t.Cleanup(func() { log.SetOutput(originalLogOutput) })
+
+	glueCatalog := &Catalog{glueSvc: mockGlueSvc}
+
+	assert.NoError(glueCatalog.PurgeTable(ctx, TableIdentifier("test_database", "test_table")))
+	assert.True(dropCalled)
+	assert.Positive(removeCalls)
+	assert.False(removeBeforeDrop, "PurgeTable should drop the catalog entry before removing files")
+	assert.Contains(logs.String(), "WARNING: dropped table")
+	assert.Contains(logs.String(), "test_database")
+	assert.Contains(logs.String(), "test_table")
+	assert.Contains(logs.String(), errGluePurgeRemove.Error())
+	file, err := failingFS.Open(dataFile)
+	assert.NoError(err, "data file should remain when FileIO remove fails")
+	assert.NotNil(file)
+	assert.NoError(file.Close())
+	mockGlueSvc.AssertExpectations(t)
+}
+
+func TestGluePurgeTableWithGCDisabled(t *testing.T) {
+	assert := require.New(t)
+	ctx := context.Background()
+	tableLocation := filepath.Join(t.TempDir(), "warehouse", "test_database", "test_table")
+	metadataLocation, dataFile := writeGluePurgeTableFilesWithProperties(
+		t, tableLocation, iceberg.Properties{"gc.enabled": "false"})
 	glueTable := gluePurgeTable(metadataLocation)
 
 	mockGlueSvc := &mockGlueClient{}
@@ -511,28 +567,25 @@ func TestGluePurgeTableSwallowsPurgeFilesError(t *testing.T) {
 		Name:         aws.String("test_table"),
 	}, mock.Anything).Return(&glue.DeleteTableOutput{}, nil).Once()
 
-	var logs bytes.Buffer
-	originalLogOutput := log.Writer()
-	log.SetOutput(&logs)
-	t.Cleanup(func() { log.SetOutput(originalLogOutput) })
-
 	glueCatalog := &Catalog{glueSvc: mockGlueSvc}
 
 	assert.NoError(glueCatalog.PurgeTable(ctx, TableIdentifier("test_database", "test_table")))
-	assert.Contains(logs.String(), "failed to purge files")
-	assert.Contains(logs.String(), errGluePurgeRemove.Error())
-	file, err := failingFS.Open(dataFile)
-	assert.NoError(err, "data file should remain when FileIO remove fails")
-	assert.NoError(file.Close())
+	assert.NoFileExists(metadataLocation)
+	assert.FileExists(dataFile)
 	mockGlueSvc.AssertExpectations(t)
 }
 
 type failRemoveIO struct {
 	*iceio.MemFS
-	err error
+	err      error
+	onRemove func()
 }
 
 func (f failRemoveIO) Remove(string) error {
+	if f.onRemove != nil {
+		f.onRemove()
+	}
+
 	return f.err
 }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: catalog/hive/hive_test.go [hunk 1]
```diff
@@ -523,6 +530,114 @@ func TestHiveDropTable(t *testing.T) {
...
+	mockClient.AssertExpectations(t)
+}
+
+func TestHivePurgeTableSwallowsPurgeFilesError(t *testing.T) {
+	assert := require.New(t)
+	ctx := context.Background()
+	const scheme = "hivepurgefail"
+	failingFS := failRemoveIO{MemFS: iceio.NewMemFS(), err: errHivePurgeRemove}
+	iceio.Register(scheme, func(context.Context, *url.URL, map[string]string) (iceio.IO, error) {
+		return failingFS, nil
```
tanmayrauth, line 571: Since MemFS isn't BulkRemovableIO, these swallow tests pin the per-file Remove path but not the bulk DeleteFiles branch in PurgeFiles. PurgeTable's swallow-and-log is the same either way so I don't think it matters for *this* PR's goal — just flagging that the bulk path stays uncovered if that's something you'd want to lock in too (probably better as a PurgeFiles test than here). Up to you.

tanmayrauth (commented):
LGTM, just a small suggestion commented.

catalog/hive/hive_test.go
Revanth14: Thanks, agreed. These tests are focused on `PurgeTable` dropping the catalog entry and swallowing/logging `PurgeFiles` failures, so the per-file remove path is enough for this PR.

The `BulkRemovableIO` branch would be better covered directly in `PurgeFiles` tests if we want to lock that behavior in separately.

catalog/glue/glue_test.go
```diff
@@ -451,6 +459,116 @@ func TestGlueDropTable(t *testing.T) {
...
+	}, mock.Anything).Return(&glue.DeleteTableOutput{}, nil).Once()
+
+	var logs bytes.Buffer
+	originalLogOutput := log.Writer()
+	log.SetOutput(&logs)
+	t.Cleanup(func() { log.SetOutput(originalLogOutput) })
+
+	glueCatalog := &Catalog{glueSvc: mockGlueSvc}
+
+	assert.NoError(glueCatalog.PurgeTable(ctx, TableIdentifier("test_database", "test_table")))
```
laskoviymishka, line 529: These tests don't prove the contract they're named for. We assert the purge error is swallowed, but nothing asserts `DeleteTable` actually ran *before* the failing `Remove` — if someone reordered `PurgeTable` to purge first and drop second, this test still passes and we'd miss the regression.

I'd record call order on the mock (testify `InOrder`, or a counter the `Remove` stub reads) and assert the table was dropped before any `Remove`. Same gap in the Hive twin.

catalog/glue/glue_test.go
```diff
@@ -451,6 +459,116 @@ func TestGlueDropTable(t *testing.T) {
...
+	assert.Contains(logs.String(), "failed to purge files")
```
laskoviymishka: The substring here is loose enough to false-pass. The actual line is `WARNING: dropped table %s but failed to purge files: %v`, and since `log.SetOutput` redirects a process-global buffer, any other line containing "failed to purge files" lands in the same `logs`.

I'd anchor it to the specific call — assert `Contains` on `WARNING: dropped table` (or the full `fmt.Sprintf` with the table identifier) so it's tied to this purge and not just any warning that happens to share the phrase.

catalog/glue/glue_test.go
```diff
@@ -451,6 +459,116 @@ func TestGlueDropTable(t *testing.T) {
...
+
+	return metadataLocation, dataFile
+}
+
+func writeGluePurgeTableMetadata(t *testing.T, fs iceio.WriteFileIO, tableLocation, metadataLocation string) {
+	t.Helper()
+	schema := iceberg.NewSchema(1, iceberg.NestedField{
+		ID: 1, Name: "foo", Type: iceberg.PrimitiveTypes.String, Required: true,
+	})
+	meta, err := table.NewMetadata(schema, iceberg.UnpartitionedSpec, table.UnsortedSortOrder, tableLocation, nil)
```
laskoviymishka: The `gc.enabled=false` branch has no coverage in either suite. `PurgeFiles` skips data-file removal when `gc.enabled=false` but always removes metadata — and that's exactly the branch a purge test should pin down, since it's the one place we deliberately keep data.

The default is `true`, so the happy path only exercises the `true` side. I'd add a case that builds metadata with `iceberg.Properties{"gc.enabled": "false"}` and asserts the metadata file is gone but the data file survives.

catalog/glue/glue_test.go
```diff
@@ -451,6 +459,116 @@ func TestGlueDropTable(t *testing.T) {
...
+		Name:         aws.String("test_table"),
+	}, mock.Anything).Return(&glue.GetTableOutput{Table: &glueTable}, nil).Twice()
+	mockGlueSvc.On("DeleteTable", mock.Anything, &glue.DeleteTableInput{
+		DatabaseName: aws.String("test_database"),
+		Name:         aws.String("test_table"),
+	}, mock.Anything).Return(&glue.DeleteTableOutput{}, nil).Once()
+
+	var logs bytes.Buffer
+	originalLogOutput := log.Writer()
+	log.SetOutput(&logs)
```
laskoviymishka: `log.SetOutput` mutates process-global state with nothing guarding against parallel execution — a future `t.Parallel()` or `go test -count=2` would race this `bytes.Buffer`. The existing `table_test.go` does the same thing with `defer` rather than `t.Cleanup`, so the PR also mixes the two conventions for one pattern.

The clean fix is to assert only that `PurgeTable` returns nil (the real contract) and drop the log inspection, or inject a `*slog.Logger` you can scope per-test. If we keep the global redirect, I'd at least drop a `// do not add t.Parallel()` note on it.

catalog/glue/glue_test.go
```diff
@@ -451,6 +459,116 @@ func TestGlueDropTable(t *testing.T) {
...
+	assert.NoFileExists(dataFile)
+	mockGlueSvc.AssertExpectations(t)
+}
+
+func TestGluePurgeTableSwallowsPurgeFilesError(t *testing.T) {
+	assert := require.New(t)
+	ctx := context.Background()
+	const scheme = "gluepurgefail"
+	failingFS := failRemoveIO{MemFS: iceio.NewMemFS(), err: errGluePurgeRemove}
+	iceio.Register(scheme, func(context.Context, *url.URL, map[string]string) (iceio.IO, error) {
```
laskoviymishka, line 503: `iceio.Register` panics if the scheme is already registered, and there's no guard for a re-run where cleanup didn't fire. If anything between here and `t.Cleanup` panics, the scheme stays registered and the next run panics on this line.

Minor, but `defer iceio.Unregister(scheme)` immediately after `Register` (or a register-if-absent) closes it. Same in the Hive test.

catalog/glue/glue_test.go
```diff
@@ -451,6 +459,116 @@ func TestGlueDropTable(t *testing.T) {
...
+		return failingFS, nil
+	})
+	t.Cleanup(func() { iceio.Unregister(scheme) })
+
+	tableLocation := scheme + "://bucket/test_database/test_table"
+	metadataLocation := tableLocation + "/metadata/v1.metadata.json"
+	dataFile := tableLocation + "/data/file.parquet"
+	writeGluePurgeTableMetadata(t, failingFS, tableLocation, metadataLocation)
+	require.NoError(t, failingFS.WriteFile(dataFile, []byte("data")))
```
laskoviymishka: Small consistency thing: the test body aliases `assert := require.New(t)` and uses `assert.NoError(...)` everywhere, then reaches back into package-level `require.NoError(t, ...)` right here. I'd use `assert.NoError(failingFS.WriteFile(...))` to match. (The helpers without the alias are fine as-is.)

catalog/glue/glue_test.go
```diff
@@ -451,6 +459,116 @@ func TestGlueDropTable(t *testing.T) {
...
+	originalLogOutput := log.Writer()
+	log.SetOutput(&logs)
+	t.Cleanup(func() { log.SetOutput(originalLogOutput) })
+
+	glueCatalog := &Catalog{glueSvc: mockGlueSvc}
+
+	assert.NoError(glueCatalog.PurgeTable(ctx, TableIdentifier("test_database", "test_table")))
+	assert.Contains(logs.String(), "failed to purge files")
+	assert.Contains(logs.String(), errGluePurgeRemove.Error())
+	file, err := failingFS.Open(dataFile)
```
laskoviymishka, line 533: `file.Close()` on the next line runs without a nil guard — it's only safe because the `assert` alias is `require.New` (FailNow on the preceding line). If that alias ever became plain `testify/assert`, this panics on a nil `file`. A `require.NotNil(t, file)` after `Open` makes the safety explicit rather than load-bearing on the alias.

catalog/hive/hive_test.go
```diff
@@ -523,6 +530,114 @@ func TestHiveDropTable(t *testing.T) {
...
+func (f failRemoveIO) Remove(string) error {
+	return f.err
+}
+
+func hivePurgeTable(metadataLocation, tableLocation string) *hive_metastore.Table {
+	return &hive_metastore.Table{
+		TableName: "test_table",
+		DbName:    "test_database",
+		TableType: TableTypeExternalTable,
+		Parameters: map[string]string{
```
laskoviymishka, line 644: This fixture omits `EXTERNAL=TRUE` and the `storage_handler` parameter. A real Iceberg table in Hive — whether written by Java's `HiveOperationsBase` or this client's `catalog/hive/schema.go` — always carries both; with only `table_type=ICEBERG` and the metadata location, the fixture models a table that could only exist from an incomplete writer.

Not blocking the purge logic under test, but adding the two keys makes it the round-trip shape we actually produce. wdyt?

laskoviymishka (changes requested):
Nice — this fills a real gap, and the `.Twice()` on `GetTable` plus the swallow-vs-propagate split is exactly the right shape for these two paths. I'd hold it for a few changes before merge though.

The main one: the `SwallowsPurgeFilesError` tests don't yet prove the invariant they're named for. We assert the error is swallowed, but nothing asserts the table was dropped *before* the failing `Remove`. If someone reordered `PurgeTable` to purge-then-drop, both tests still pass — and that ordering is the whole point of the contract. I'd record call order on the mock and assert drop-before-remove. The log check is also matching a substring (`failed to purge files`) loose enough that an unrelated line in the shared global buffer could false-pass; I'd tie it to `WARNING: dropped table` or the full formatted line.

Second, the `gc.enabled=false` branch has zero coverage. That's the one path where `PurgeFiles` deliberately keeps data files while still removing metadata — exactly what a purge test should pin down. I'd add a case here that sets `gc.enabled=false` and asserts metadata gone / data preserved.

- guard `iceio.Register` against a missed-cleanup re-run (it panics on duplicate scheme)
- the global `log.SetOutput` mutation is unprotected against `t.Parallel()` / `-count=2` — either drop the log inspection for a plain return-nil assert, or note the constraint
- the `require`/`assert`-alias mix and the unguarded `file.Close()` in the Swallows bodies
- Hive fixture is missing `EXTERNAL=TRUE` + `storage_handler` (non-blocking)

The duplicated `failRemoveIO` across both packages is unavoidable without a shared test helper — fine to leave, maybe a follow-up.

Once the ordering assertion and the `gc.enabled=false` case are in, happy to take another pass and approve.

catalog/glue/glue_test.go
Revanth14: Addressed in both Glue and Hive. The failing `Remove` path now records whether the catalog drop had already run, and the tests assert we drop before any file removal attempt. This should fail if `PurgeTable` is reordered to purge first and drop second.

catalog/glue/glue_test.go
Revanth14: Updated this to anchor on the real warning path instead of the loose substring. The assertions now check `WARNING: dropped table`, the table identifier pieces, and the injected purge error, so the log match is tied to this specific `PurgeTable` call.

The test is kept non-parallel while redirecting the process-global logger.

catalog/glue/glue_test.go
Revanth14: Added coverage for this in both Glue and Hive. The new `gc.enabled=false` cases build metadata with `iceberg.Properties{"gc.enabled": "false"}` and assert `PurgeTable` removes the metadata file while preserving the data file.

catalog/glue/glue_test.go
Revanth14: Good point. I dropped the log inspection entirely so the tests no longer mutate the process-global logger. The swallow tests now focus on the actual contract: `PurgeTable` returns nil, the catalog entry is dropped before any file removal attempt, and the data file remains when removal fails.

catalog/glue/glue_test.go
Revanth14: Addressed in both Glue and Hive. The tests now call `iceio.Unregister(scheme)` before `Register` to guard against stale registrations from a prior failed run, and still register cleanup with `t.Cleanup` for the normal path.

catalog/glue/glue_test.go
Revanth14: Addressed. Switched the in-test `WriteFile` assertions to use the local `assert := require.New(t)` helper in both Glue and Hive.

catalog/glue/glue_test.go
Revanth14: Addressed in both Glue and Hive. Added an explicit `assert.NotNil(file)` after `Open` before calling `Close`, so the nil safety is no longer implicit in the `require.New(t)` alias.

catalog/hive/hive_test.go
Revanth14: Agreed. Added both `EXTERNAL=TRUE` and `storage_handler` to the Hive purge fixture so it matches the table shape produced by the Hive catalog.

laskoviymishka (approved):
👍

PR #1286, opened 2026-06-23
refactor: require ListableIO for orphan cleanup walk

### LINKED ISSUE #1276: refactor: remove the dead reflection fallback in orphan-cleanup walkDirectory
Follow-up to #916.

`walkDirectory` first tries `fsys.(iceio.ListableIO)` and returns immediately if it matches, then falls through to a `switch` with a `LocalFS` case and a blob `default` that calls `getBucketName`, which reaches into the FS with `reflect` to pull a `Bucket` field. Every concrete IO now implements `WalkDir` / `ListableIO` — `LocalFS`, `blobFileIO` (all gocloud constructors return it), and `MemFS` — so the `ListableIO` branch always wins and the switch plus `getBucketName` can never run. The thread asked to remove this once #917 landed; #917 is merged.

Delete the post-`ListableIO` fallback and `getBucketName` from `walkDirectory`, return an error if a non-`ListableIO` ever shows up, and drop the now-unused `reflect` import.

Revanth14:
Fixes #1276.

## Summary

Removes the dead fallback path in orphan-cleanup directory walking.

`walkDirectory` now requires the table `FileIO` to implement `ListableIO` and returns a clear error if it does not. This also removes the old reflection-based `getBucketName` helper and the unused `reflect` import.

## Why

All concrete IO implementations used here now implement `WalkDir` / `ListableIO`, so the old post-`ListableIO` fallback was unreachable. Removing it gets rid of brittle reflection against blob internals.

## Testing

```sh
go test ./table
go test ./...

### DIFF
--- a/table/orphan_cleanup.go
+++ b/table/orphan_cleanup.go
@@ -166,6 +166,11 @@ type OrphanCleanupResult struct {
 	TotalSizeBytes int64
 }
 
+// DeleteOrphanFiles identifies files under a table location that are no longer
+// referenced by table metadata and deletes them unless dry-run is enabled.
+//
+// The table filesystem must implement iceio.ListableIO so orphan cleanup can
+// fully enumerate candidate files before deciding what is safe to delete.
 func (t Table) DeleteOrphanFiles(ctx context.Context, opts ...OrphanCleanupOption) (OrphanCleanupResult, error) {
 	cfg := &orphanCleanupConfig{
 		location:           "",             // empty means use table's data location
@@ -422,7 +427,7 @@ func walkDirectory(fsys iceio.IO, root string, fn func(path string, info stdfs.F
 		})
 	}
 
-	return fmt.Errorf("filesystem %T does not implement io.ListableIO", fsys)
+	return fmt.Errorf("filesystem %T does not implement iceio.ListableIO", fsys)
 }
 
 func isFileOrphan(file string, referencedFiles map[string]bool, normalizedReferencedFiles map[string]string, cfg *orphanCleanupConfig) (bool, error) {
--- a/table/orphan_cleanup.go
+++ b/table/orphan_cleanup.go
@@ -26,7 +26,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"reflect"
 	"runtime"
 	"slices"
 	"strings"
@@ -404,7 +403,6 @@ func (t Table) getReferencedFiles(ctx context.Context, fs iceio.IO, maxConcurren
 }
 
 func walkDirectory(fsys iceio.IO, root string, fn func(path string, info stdfs.FileInfo) error) error {
-	// Prefer ListableIO when available.
 	if listable, ok := fsys.(iceio.ListableIO); ok {
 		return listable.WalkDir(root, func(path string, d stdfs.DirEntry, err error) error {
 			if err != nil {
@@ -424,64 +422,7 @@ func walkDirectory(fsys iceio.IO, root string, fn func(path string, info stdfs.F
 		})
 	}
 
-	// Fallback to original implementation for IO types that don't
-	// implement ListableIO yet.
-	switch v := fsys.(type) {
-	case iceio.LocalFS:
-		cleanRoot := strings.TrimPrefix(root, "file://")
-		if cleanRoot == "" {
-			cleanRoot = "."
-		}
-
-		return filepath.WalkDir(cleanRoot, makeFileWalkFunc(fn, func(path string) string {
-			return path
-		}))
-
-	default:
-		// For blob storage: direct field access since we know the structure.
-		bucket := getBucketName(v)
-
-		parsed, err := url.Parse(root)
-		if err != nil {
-			return fmt.Errorf("invalid URL %s: %w", root, err)
-		}
-
-		walkPath := strings.TrimPrefix(parsed.Path, "/")
-		if walkPath == "" {
-			walkPath = "."
-		}
-
-		return stdfs.WalkDir(bucket, walkPath, makeFileWalkFunc(fn, func(path string) string {
-			return parsed.Scheme + "://" + parsed.Host + "/" + path
-		}))
-	}
-}
-
-// getBucketName gets the Bucket field from blob storage - absolute minimal approach.
-func getBucketName(fsys iceio.IO) stdfs.FS {
-	v := reflect.ValueOf(fsys).Elem()
-
-	return v.FieldByName("Bucket").Interface().(stdfs.FS)
-}
-
-// makeFileWalkFunc creates a WalkDirFunc that processes only files with path transformation.
-func makeFileWalkFunc(fn func(path string, info stdfs.FileInfo) error, pathTransform func(string) string) stdfs.WalkDirFunc {
-	return func(path string, d stdfs.DirEntry, err error) error {
-		if err != nil {
-			return err
-		}
-
-		if d.IsDir() {
-			return nil
-		}
-
-		info, err := d.Info()
-		if err != nil {
-			return err
-		}
-
-		return fn(pathTransform(path), info)
-	}
+	return fmt.Errorf("filesystem %T does not implement io.ListableIO", fsys)
 }
 
 func isFileOrphan(file string, referencedFiles map[string]bool, normalizedReferencedFiles map[string]string, cfg *orphanCleanupConfig) (bool, error) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: table/orphan_cleanup.go [hunk 1]
```diff
@@ -424,64 +422,7 @@ func walkDirectory(fsys iceio.IO, root string, fn func(path string, info stdfs.F
...
-		}
-
-		info, err := d.Info()
-		if err != nil {
-			return err
-		}
-
-		return fn(pathTransform(path), info)
-	}
+	return fmt.Errorf("filesystem %T does not implement io.ListableIO", fsys)
```
tanmayrauth: Minor consistency thought, not blocking: PurgeFiles at line 780 silently skips the walk when the FS isn't listable, while this now returns a hard error. I think erroring is the right call for orphan cleanup (vs. best-effort purge)

tanmayrauth (commented):
LGTM, One small thought below.

table/orphan_cleanup.go
Revanth14: Thanks, agreed. `PurgeFiles` is best-effort so skipping a non-listable FS makes sense there, but orphan cleanup needs to know whether it fully walked the location before deciding what is orphaned. Keeping this as a hard error feels safer.

laskoviymishka: The type here is `iceio.ListableIO`, but `io` is the stdlib alias also in scope, so this error points readers at the wrong package — they'll go searching stdlib `io` and come up empty.

I'd name the local alias: `fmt.Errorf("filesystem %T does not implement iceio.ListableIO", fsys)`. The test assertion at `orphan_cleanup_test.go:632` hard-codes the current string, so it needs to move with this.

laskoviymishka: `DeleteOrphanFiles` and `FSysF` are public, and this turns "IO that implements only `iceio.IO`" into a hard runtime failure on orphan cleanup. For all the built-in schemes that's fine — they all implement ListableIO — but a downstream caller injecting a custom IO would hit this with no signal until runtime.

I'd add a line of godoc on `DeleteOrphanFiles` noting that the table's IO must implement `iceio.ListableIO`. Worth aligning with `PurgeFiles` too while we're here — it does its own inline ListableIO check and silently skips the walk, so the two paths handle the same condition in opposite ways. Not blocking, but the asymmetry is a future trap.

table/orphan_cleanup.go
```diff
@@ -404,7 +403,6 @@ func (t Table) getReferencedFiles(ctx context.Context, fs iceio.IO, maxConcurren
 }
 
 func walkDirectory(fsys iceio.IO, root string, fn func(path string, info stdfs.FileInfo) error) error {
-	// Prefer ListableIO when available.
 	if listable, ok := fsys.(iceio.ListableIO); ok {
```
laskoviymishka, line 411: `blobFileIO` (S3/GCS/ADLS) is now the load-bearing ListableIO type for the whole gocloud path, but it has no compile-time assertion the way `LocalFS` does — if its `WalkDir` signature drifts, the break surfaces at runtime in orphan cleanup rather than at compile time.

I'd add `var _ icebergio.ListableIO = (*blobFileIO)(nil)` over in `io/gocloud/blob.go` to lock it down. wdyt?

table/orphan_cleanup_test.go
```diff
@@ -621,6 +622,16 @@ func TestDeleteFilesFallsBackToExistingBehavior(t *testing.T) {
...
 }
 
+func TestWalkDirectoryRequiresListableIO(t *testing.T) {
+	err := walkDirectory(&mockPlainIO{}, "s3://bucket/data", func(string, stdfs.FileInfo) error {
+		require.Fail(t, "walk callback should not run for non-listable IO")
+
+		return nil
+	})
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "does not implement io.ListableIO")
```
laskoviymishka: This is the only unit test touching `walkDirectory`, and it only covers the error branch. With the fallback gone, nothing at the unit level exercises the success path — if the ListableIO callback regresses (path transform, file-vs-dir filtering, callback ordering) the integration tests would catch it but they're build-tag gated and don't run in normal CI.

Could we add a case that drives a `MemFS` (or a small ListableIO mock) with a few seeded paths through `walkDirectory` and asserts the callback receives the full expected set? Otherwise the test would still pass if `walkDirectory`'s body were deleted down to the error return. wdyt?

laskoviymishka (changes requested):
This is a clean refactor and the core claim holds up — LocalFS, MemFS, and blobFileIO behind every gocloud scheme all implement WalkDir, so the fallback was genuinely dead and dropping the reflection-based getBucketName is a nice cleanup.

I'd hold it before merging on two things. The error string says `io.ListableIO`, but the real type is `iceio.ListableIO` and stdlib `io` is also in scope — anyone hitting that error will go looking in the wrong package. And with the fallback gone, the only unit test on `walkDirectory` covers the error branch; the success path has no coverage, so a regression in the ListableIO callback would slip through normal CI. I left an inline suggesting a MemFS-driven case.

One design point worth a thought, not blocking: `DeleteOrphanFiles` and `FSysF` are public, and this makes a custom IO that implements only `iceio.IO` fail hard at runtime. A line of godoc on `DeleteOrphanFiles` would close that gap, and it'd be worth aligning with `PurgeFiles`, which handles the same condition by silently skipping the walk.

Once the error string and the test gap are addressed, happy to take another pass and approve.

table/orphan_cleanup.go
Revanth14: Done, updated this to `iceio.ListableIO` and moved the test assertion with it.

Revanth14: Added godoc on `DeleteOrphanFiles` noting that the table filesystem must implement `iceio.ListableIO`.

I left `PurgeFiles` behavior unchanged here: purge is best-effort after catalog drop, while orphan cleanup needs complete location enumeration before it can safely decide what is orphaned, so the hard failure is intentional.

table/orphan_cleanup.go
Revanth14: Good call. Added `var _ icebergio.ListableIO = (*blobFileIO)(nil)` in `io/gocloud/blob.go`, so a `WalkDir` signature drift breaks at build time instead of surfacing in orphan cleanup at runtime.

table/orphan_cleanup_test.go
Revanth14: Added `TestWalkDirectoryUsesListableIO`, which drives a seeded `ListableIO` mock through `walkDirectory` and asserts the callback receives exactly the expected files: directories filtered out, paths and sizes intact. It fails if the body is reduced to the error return.

I went with a small mock rather than `MemFS` to keep the test focused on the `ListableIO` contract, but happy to switch to `MemFS` if you’d prefer.

Revanth14:
> This is a clean refactor and the core claim holds up — LocalFS, MemFS, and blobFileIO behind every gocloud scheme all implement WalkDir, so the fallback was genuinely dead and dropping the reflection-based getBucketName is a nice cleanup.
> 
> I'd hold it before merging on two things. The error string says `io.ListableIO`, but the real type is `iceio.ListableIO` and stdlib `io` is also in scope — anyone hitting that error will go looking in the wrong package. And with the fallback gone, the only unit test on `walkDirectory` covers the error branch; the success path has no coverage, so a regression in the ListableIO callback would slip through normal CI. I left an inline suggesting a MemFS-driven case.
> 
> One design point worth a thought, not blocking: `DeleteOrphanFiles` and `FSysF` are public, and this makes a custom IO that implements only `iceio.IO` fail hard at runtime. A line of godoc on `DeleteOrphanFiles` would close that gap, and it'd be worth aligning with `PurgeFiles`, which handles the same condition by silently skipping the walk.
> 
> Once the error string and the test gap are addressed, happy to take another pass and approve.

Thanks for the review. Addressed the two blocking items:

- Updated the error string and assertion from `io.ListableIO` to `iceio.ListableIO`.
- Added `TestWalkDirectoryUsesListableIO` to cover the success path through `walkDirectory`.

Also added the `DeleteOrphanFiles` godoc note and the `blobFileIO` compile-time `ListableIO` assertion from the follow-up comments. I left `PurgeFiles` behavior unchanged because purge is best-effort after catalog drop, while orphan cleanup needs complete location enumeration before deciding what is safe to delete.

laskoviymishka (approved):
:shipit:

PR #1290, opened 2026-06-23
fix(table): release empty scan batches

fallintoplace:
Summary

Release zero-row record batches before `createIterator` skips them. The iterator owns batches after pulling them from the sequenced channel, so an empty batch still needs the same cleanup as any other batch that is not yielded.

The test uses a checked allocator and a zero-row slice backed by a real Arrow buffer, so it catches the leak directly instead of relying on a full scan setup.

Testing

- `go test ./table -run '^TestCreateIteratorReleasesEmptyRecordBatches$' -count=1`
- `go test ./table -count=1`
- `git diff --check`


tanmayrauth (commented):
LGTM. Confirmed the test actually catches the leak — reverting just the rec.Release() at arrow_scanner.go:1039 makes TestCreateIteratorReleasesEmptyRecordBatches fail with the checked allocator reporting the leaked buffer, and the fix closes it. Nice that it puts empty batches on the same cleanup footing as the rowLimit paths right above. Thanks for the targeted test.

tanmayrauth (approved):

laskoviymishka (approved):
:shipit:

laskoviymishka:
@fallintoplace good to merge, but you need to fix merge conficts.

fallintoplace:
@laskoviymishka @zeroshade @tanmayrauth Can you please check again when you have time? Thank you for the support.

PR #1291, opened 2026-06-23
fix(catalog): avoid mutating load properties

fallintoplace:
Summary

Clone caller-provided properties in `catalog.Load` before filling defaults from the configured catalog. The merged copy is still passed to the catalog registrar, but the caller-owned map is left alone.

This avoids surprising side effects when callers reuse a config map across loads, or share one between goroutines.

Testing

- `go test ./catalog -run '^TestCatalogLoadDoesNotMutateProps$' -count=1`
- `go test ./catalog -count=1`
- `git diff --check`


laskoviymishka (approved):
:shipit:

PR #1292, opened 2026-06-23
fix(catalog/hadoop): load discovered metadata files

fallintoplace:
Summary

Make the Hadoop catalog load the actual metadata file it discovers instead of reconstructing vN.metadata.json after finding a version. This keeps gzip metadata and UUID-style metadata names consistent across LoadTable, CheckTableExists, and ListTables.

This also stops a lone version-hint.text from making a directory look like a table when there is no loadable metadata file.

Follow-up changes from review:

- propagate metadata directory read errors instead of flattening them into ErrNoSuchTable
- keep CheckTableExists on the fast metadata-file existence path
- reuse the metadata version found by LoadTable during CommitTable
- reject UUID metadata sequence 00000 and use 00001 in the UUID fixture
- pin metadata tie-break ordering with focused tests

Testing

- go test ./catalog/hadoop -run 'TestMetadataFileBetterThan|TestHadoopCatalogTestSuite/(TestLoadTablePropagatesMetadataReadError|TestCheckTableExistsPropagatesMetadataReadError|TestLoadTableGzipMetadata|TestLoadTableUUIDMetadata|TestVersionHintWithoutMetadataIsNotLoadableTable|TestIsTableDirFalseVersionHintOnly|TestFindVersionGzipOnlyWithHint|TestFindVersionMixedGzipAndPlain|TestMetadataFileFromNameRejectsZeroUUIDVersion)' -count=1
- git diff --check

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -60,8 +60,70 @@ var versionPattern = regexp.MustCompile(`^v([0-9]+)(?:\.gz)?\.metadata\.json$`)
 // 00000-<uuid>.gz.metadata.json. The sequence is a 5-digit zero-padded
 // number and the UUID is in canonical 8-4-4-4-12 hex format.
 var uuidMetadataPattern = regexp.MustCompile(
-	`^[0-9]{5}-[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(?:\.gz)?\.metadata\.json$`,
+	`^([0-9]{5})-[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(?:\.gz)?\.metadata\.json$`,
 )
+
+type metadataFile struct {
+	location   string
+	version    int
+	hadoopName bool
+	compressed bool
+}
+
+func metadataFileFromName(path, name string) (metadataFile, bool) {
+	if matches := versionPattern.FindStringSubmatch(name); len(matches) == 2 {
+		version, err := strconv.Atoi(matches[1])
+		if err != nil || version <= 0 {
+			return metadataFile{}, false
+		}
+
+		return metadataFile{
+			location:   path,
+			version:    version,
+			hadoopName: true,
+			compressed: strings.Contains(name, ".gz.metadata.json"),
+		}, true
+	}
+
+	if matches := uuidMetadataPattern.FindStringSubmatch(name); len(matches) == 2 {
+		version, err := strconv.Atoi(matches[1])
+		if err != nil || version < 0 {
+			return metadataFile{}, false
+		}
+
+		return metadataFile{
+			location:   path,
+			version:    version,
+			compressed: strings.Contains(name, ".gz.metadata.json"),
+		}, true
+	}
+
+	return metadataFile{}, false
+}
+
+func (m metadataFile) betterThan(current metadataFile) bool {
+	switch {
+	case current.location == "":
+		return true
+	case m.version != current.version:
+		return m.version > current.version
+	case m.hadoopName != current.hadoopName:
+		// If both naming styles exist for one version, prefer the Hadoop vN
+		// file because this catalog writes and conflict-checks that sequence.
+		return m.hadoopName
+	case m.compressed != current.compressed:
+		return !m.compressed
+	default:
+		// This keeps scans deterministic under map iteration; within one
+		// metadata directory it reduces to a stable filename tie-break.
+		return m.location > current.location
+	}
+}
+
+func shouldSuppressPermissionError(err error) bool {
+	return errors.Is(err, fs.ErrPermission) ||
+		(err != nil && strings.Contains(err.Error(), "AuthorizationPermissionMismatch"))
+}
 
 // validateIdentifier checks that an identifier is non-empty and that each
 // component is safe for use as a path segment. It rejects nil/empty
@@ -237,12 +299,11 @@ func (c *Catalog) defaultTableLocation(ident table.Identifier) string {
 }
 
 // isTableDir reports whether path is a table directory by checking for
-// metadata files in its metadata/ subdirectory. It recognizes:
+// loadable metadata files in its metadata/ subdirectory. It recognizes:
 //   - v*.metadata.json (Hadoop catalog format)
 //   - <seq>-<uuid>.metadata.json (Java/PyIceberg format)
-//   - version-hint.text
-func isTableDir(filesystem HadoopCatalogFS, isLocal bool, path string) bool {
-	metaDir := joinPath(isLocal, path, "metadata")
+func isTableDir(filesystem HadoopCatalogFS, isLocal bool, tablePath string) (bool, error) {
+	metaDir := joinPath(isLocal, tablePath, "metadata")
 
 	foundMetadata := false
 	err := filesystem.WalkDir(metaDir, func(path string, d fs.DirEntry, err error) error {
@@ -260,10 +321,7 @@ func isTableDir(filesystem HadoopCatalogFS, isLocal bool, path string) bool {
 			return fs.SkipDir
 		}
 
-		name := d.Name()
-		if versionPattern.MatchString(name) ||
-			uuidMetadataPattern.MatchString(name) ||
-			name == "version-hint.text" {
+		if _, ok := metadataFileFromName(path, d.Name()); ok {
 			foundMetadata = true
 
 			return fs.SkipAll
@@ -272,10 +330,25 @@ func isTableDir(filesystem HadoopCatalogFS, isLocal bool, path string) bool {
 		return nil
 	})
 	if err != nil {
-		return false
+		if errors.Is(err, fs.ErrNotExist) {
+			return false, nil
+		}
+
+		return false, err
 	}
 
-	return foundMetadata
+	return foundMetadata, nil
+}
+
+func isTableDirForListing(filesystem HadoopCatalogFS, isLocal bool, tablePath string) (bool, error) {
+	isTable, err := isTableDir(filesystem, isLocal, tablePath)
+	if err != nil && shouldSuppressPermissionError(err) {
+		log.Printf("hadoop catalog: unable to inspect metadata directory %s: %v", joinPath(isLocal, tablePath, "metadata"), err)
+
+		return false, nil
+	}
+
+	return isTable, err
 }
 
 func (c *Catalog) readVersionHint(ident table.Identifier) int {
@@ -310,43 +383,11 @@ func (c *Catalog) writeVersionHint(ident table.Identifier, version int) {
 	}
 }
 
-// metadataVersionExists checks whether a metadata file for the given version
-// exists in either plain or gzip-compressed form.
-func (c *Catalog) metadataVersionExists(ident table.Identifier, version int) bool {
+func (c *Catalog) scanMetadataFiles(ident table.Identifier) (map[int]metadataFile, metadataFile, error) {
 	dir := c.metadataDir(ident)
-	plain := joinPath(c.isLocal, dir, fmt.Sprintf("v%d.metadata.json", version))
+	byVersion := map[int]metadataFile{}
+	var latest metadataFile
 
-	if _, err := c.filesystem.Stat(plain); err == nil {
-		return true
-	}
-
-	gz := joinPath(c.isLocal, dir, fmt.Sprintf("v%d.gz.metadata.json", version))
-
-	_, err := c.filesystem.Stat(gz)
-
-	return err == nil
-}
-
-func (c *Catalog) scanForward(ident table.Identifier, start int) int {
-	ver := start
-	for c.metadataVersionExists(ident, ver+1) {
-		ver++
-	}
-
-	return ver
-}
-
-func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
-	hint := c.readVersionHint(ident)
-	if hint > 0 && c.metadataVersionExists(ident, hint) {
-		return c.scanForward(ident, hint), nil
-	}
-
-	dir := c.metadataDir(ident)
-
-	maxVer := 0
-	// Walk just one directory level to find metadata files,
-	// ignoring any subdirectories that may exist
 	err := c.filesystem.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
@@ -358,28 +399,84 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
 			return fs.SkipDir
 		}
 
-		name := d.Name()
-		matches := versionPattern.FindStringSubmatch(name)
-		if len(matches) == 2 {
-			v, _ := strconv.Atoi(matches[1])
-			if v > maxVer {
-				maxVer = v
-			}
+		file, ok := metadataFileFromName(path, d.Name())
+		if !ok {
+			return nil
+		}
+
+		if file.betterThan(byVersion[file.version]) {
+			byVersion[file.version] = file
+		}
+		if file.betterThan(latest) {
+			latest = file
 		}
 
 		return nil
 	})
+
+	return byVersion, latest, err
+}
+
+func (c *Catalog) metadataVersionExists(ident table.Identifier, version int) (bool, error) {
+	dir := c.metadataDir(ident)
+	found := false
+
+	// This intentionally scans: UUID metadata filenames contain an
+	// unpredictable UUID, so Stat checks cannot cover every same-version race.
+	err := c.filesystem.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if path == dir {
+			return nil
+		}
+		if d.IsDir() {
+			return fs.SkipDir
+		}
+
+		file, ok := metadataFileFromName(path, d.Name())
+		if !ok || file.version != version {
+			return nil
+		}
+
+		found = true
+
+		return fs.SkipAll
+	})
 	if err != nil {
-		return 0, fmt.Errorf("hadoop catalog: cannot read metadata directory for %s: %w",
+		return false, err
+	}
+
+	return found, nil
+}
+
+func (c *Catalog) findMetadataLocation(ident table.Identifier) (string, int, error) {
+	_, latest, err := c.scanMetadataFiles(ident)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return "", 0, fmt.Errorf("hadoop catalog: cannot read metadata directory for %s: %w: %w",
+				strings.Join(ident, "."), catalog.ErrNoSuchTable, err)
+		}
+
+		return "", 0, fmt.Errorf("hadoop catalog: cannot read metadata directory for %s: %w",
+			strings.Join(ident, "."), err)
+	}
+
+	if latest.location == "" {
+		return "", 0, fmt.Errorf("hadoop catalog: no metadata files found for table %s: %w",
 			strings.Join(ident, "."), catalog.ErrNoSuchTable)
 	}
 
-	if maxVer == 0 {
-		return 0, fmt.Errorf("hadoop catalog: no metadata files found for table %s: %w",
-			strings.Join(ident, "."), catalog.ErrNoSuchTable)
+	return latest.location, latest.version, nil
+}
+
+func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
+	_, version, err := c.findMetadataLocation(ident)
+	if err != nil {
+		return 0, err
 	}
 
-	return c.scanForward(ident, maxVer), nil
+	return version, nil
 }
 
 func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *iceberg.Schema, opts ...catalog.CreateTableOpt) (*table.Table, error) {
@@ -409,7 +506,11 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 		return nil, errors.New("hadoop catalog: custom table locations are not supported")
 	}
 
-	if isTableDir(c.filesystem, c.isLocal, loc) {
+	exists, err := isTableDir(c.filesystem, c.isLocal, loc)
+	if err != nil {
+		return nil, fmt.Errorf("hadoop catalog: failed to inspect table directory: %w", err)
+	}
+	if exists {
 		return nil, fmt.Errorf("%w: %s", catalog.ErrTableAlreadyExists, strings.Join(ident, "."))
 	}
 
@@ -457,19 +558,28 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 	return tbl, nil
 }
 
-func (c *Catalog) LoadTable(ctx context.Context, ident table.Identifier) (*table.Table, error) {
+func (c *Catalog) loadTable(ctx context.Context, ident table.Identifier) (*table.Table, int, error) {
 	if err := validateTableIdentifier(ident); err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 
-	ver, err := c.findVersion(ident)
+	metaPath, version, err := c.findMetadataLocation(ident)
 	if err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 
-	metaPath := c.metadataFilePath(ident, ver)
+	tbl, err := table.NewFromLocation(ctx, ident, metaPath, io.LoadFSFunc(c.props, metaPath), c)
+	if err != nil {
+		return nil, 0, err
+	}
 
-	return table.NewFromLocation(ctx, ident, metaPath, io.LoadFSFunc(c.props, metaPath), c)
+	return tbl, version, nil
+}
+
+func (c *Catalog) LoadTable(ctx context.Context, ident table.Identifier) (*table.Table, error) {
+	tbl, _, err := c.loadTable(ctx, ident)
+
+	return tbl, err
 }
 
 func (c *Catalog) CheckTableExists(_ context.Context, ident table.Identifier) (bool, error) {
@@ -477,7 +587,12 @@ func (c *Catalog) CheckTableExists(_ context.Context, ident table.Identifier) (b
 		return false, nil
 	}
 
-	return isTableDir(c.filesystem, c.isLocal, c.tableToPath(ident)), nil
+	exists, err := isTableDir(c.filesystem, c.isLocal, c.tableToPath(ident))
+	if err != nil {
+		return false, err
+	}
+
+	return exists, nil
 }
 
 func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs []table.Requirement, updates []table.Update) (table.Metadata, string, error) {
@@ -486,7 +601,7 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs 
 	}
 
 	// Step 1: Load current table (nil for create-via-commit).
-	current, err := c.LoadTable(ctx, ident)
+	current, currentVersion, err := c.loadTable(ctx, ident)
 	if err != nil && !errors.Is(err, catalog.ErrNoSuchTable) {
 		return nil, "", err
 	}
@@ -530,15 +645,6 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs 
 	}
 
 	// Step 6: Determine next version number.
-	var currentVersion int
-
-	if current != nil {
-		currentVersion, err = c.findVersion(ident)
-		if err != nil {
-			return nil, "", err
-		}
-	}
-
 	newVersion := currentVersion + 1
 
 	// Step 7: Create metadata directory if needed (create-via-commit).
@@ -558,6 +664,19 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs 
 		return nil, "", fmt.Errorf("hadoop catalog: failed to write table metadata: %w", err)
 	}
 
+	exists, err := c.metadataVersionExists(ident, newVersion)
+	if err != nil {
+		_ = c.filesystem.Remove(tempPath)
+
+		return nil, "", fmt.Errorf("hadoop catalog: failed to inspect metadata directory for version %d: %w",
+			newVersion, err)
+	}
+	if exists {
+		_ = c.filesystem.Remove(tempPath)
+
+		return nil, "", fmt.Errorf("hadoop catalog: version %d already exists for table %s",
+			newVersion, strings.Join(ident, "."))
+	}
 	if err := c.commitMetadataFile(ident, tempPath, newMetaPath, table.ErrCommitFailed); err != nil {
 		return nil, "", err
 	}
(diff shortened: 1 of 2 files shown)


catalog/hadoop/hadoop.go
```diff
@@ -318,28 +332,65 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
...
+
+		current = next
+	}
+}
+
+func (c *Catalog) findMetadataLocation(ident table.Identifier) (string, int, error) {
+	files, latest, err := c.scanMetadataFiles(ident)
 	if err != nil {
-		return 0, fmt.Errorf("hadoop catalog: cannot read metadata directory for %s: %w",
+		return "", 0, fmt.Errorf("hadoop catalog: cannot read metadata directory for %s: %w",
```
laskoviymishka, line 461: I think there's a real bug here: any error out of `scanMetadataFiles` (permission denied, transient FS failure) gets wrapped as `catalog.ErrNoSuchTable`, and the original error is dropped — it's not threaded through `%w`.

Downstream, `CheckTableExists` does `errors.Is(err, catalog.ErrNoSuchTable)` and returns `(false, nil)`, so a metadata dir we simply couldn't read looks exactly like a table that doesn't exist. That's the kind of thing that leads to a `CreateTable` stomping live data.

I'd only map to `ErrNoSuchTable` when `errors.Is(err, fs.ErrNotExist)`, and otherwise return the original error wrapped. wdyt?

catalog/hadoop/hadoop.go
```diff
@@ -439,7 +488,12 @@ func (c *Catalog) CheckTableExists(_ context.Context, ident table.Identifier) (b
 		return false, nil
 	}
 
-	return isTableDir(c.filesystem, c.tableToPath(ident)), nil
+	_, _, err := c.findMetadataLocation(ident)
+	if errors.Is(err, catalog.ErrNoSuchTable) {
+		return false, nil
+	}
+
+	return err == nil, err
```
laskoviymishka: `return err == nil, err` is doing two jobs in one line and obscures the contract. Combined with the error-swallowing in `findMetadataLocation` above, a non-`ErrNoSuchTable` failure flows through here as `(false, <err>)` — which is the right shape only once that function stops masking real errors.

I'd make it explicit: `if err != nil { return false, err }` then `return true, nil`. Easier to see that "not found" and "couldn't tell" are different outcomes.

catalog/hadoop/hadoop.go
```diff
@@ -439,7 +488,12 @@ func (c *Catalog) CheckTableExists(_ context.Context, ident table.Identifier) (b
 		return false, nil
 	}
 
-	return isTableDir(c.filesystem, c.tableToPath(ident)), nil
+	_, _, err := c.findMetadataLocation(ident)
```
laskoviymishka: This swaps the old `isTableDir` short-circuit (first match → `fs.SkipAll`) for a full `scanMetadataFiles` walk on every existence check. Java's `HadoopCatalog` uses a filtered `listStatus` that bails on the first hit; a table with a few hundred versions now pays an O(metadata-file-count) walk per `CheckTableExists`, and this is called before `CreateTable` and to guard commits.

`isTableDir` now correctly excludes hint-only dirs (same `metadataFileFromName`), so it's still a valid fast path. I'd keep `CheckTableExists` on `isTableDir` and reserve the full scan for callers that actually need the location. Thoughts?

catalog/hadoop/hadoop.go
```diff
@@ -495,7 +549,7 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
 	var currentVersion int
 
 	if current != nil {
-		currentVersion, err = c.findVersion(ident)
+		_, currentVersion, err = c.findMetadataLocation(ident)
```
laskoviymishka: `CommitTable` already called `LoadTable` (→ `findMetadataLocation` → full walk) just above, and now walks the same directory a second time here just to recover the version. Two independent full scans of the same metadata dir per commit.

Since `findMetadataLocation` already returns the version alongside the path, I'd have `LoadTable` hand both back (or read the version off the already-loaded metadata) and reuse it here. wdyt?

catalog/hadoop/hadoop.go
```diff
@@ -59,9 +59,59 @@ var versionPattern = regexp.MustCompile(`^v([0-9]+)(?:\.gz)?\.metadata\.json$`)
...
+			location:   path,
+			version:    version,
+			compressed: strings.Contains(name, ".gz.metadata.json"),
+		}, true
+	}
+
+	return metadataFile{}, false
+}
+
+func (m metadataFile) betterThan(current metadataFile) bool {
```
laskoviymishka, line 104: This tie-breaker is the most novel piece of the PR and it's the resolution policy the whole discovery rewrite leans on, but nothing tests it. I'd add a focused unit test on `betterThan`: empty current loses; higher version wins; same version → hadoop name beats uuid; same version+style → uncompressed beats gz; and the final lexicographic fallback.

The hadoop-over-uuid preference at equal version is also an undocumented policy with no Java precedent (Java never mixes the two naming schemes). A one-line comment on *why* we prefer the hadoop name would save the next reader a real head-scratch. Once there's a test pinning these, I'm comfortable with the logic.

catalog/hadoop/hadoop.go
```diff
@@ -59,9 +59,59 @@ var versionPattern = regexp.MustCompile(`^v([0-9]+)(?:\.gz)?\.metadata\.json$`)
...
+		return metadataFile{
+			location:   path,
+			version:    version,
+			hadoopName: true,
+			compressed: strings.Contains(name, ".gz.metadata.json"),
+		}, true
+	}
+
+	if matches := uuidMetadataPattern.FindStringSubmatch(name); len(matches) == 2 {
+		version, _ := strconv.Atoi(matches[1])
```
laskoviymishka: The hadoop branch rejects `version <= 0`, but this UUID branch doesn't, so a `00000-<uuid>.metadata.json` parses to `version: 0` and is accepted. In Java/PyIceberg the first commit writes `00001-` (`%05d` on `newVersion`, which starts at 1), and version 0 is the "table doesn't exist" sentinel — so a real client never produces `00000-`.

I'd mirror the guard in this branch:

```go
if version <= 0 {
    return metadataFile{}, false
}
```

That also lets the UUID test fixture move onto `00001-`, which it should be anyway.

catalog/hadoop/hadoop_test.go
```diff
@@ -230,6 +232,35 @@ func (s *HadoopCatalogTestSuite) createMetadataFile(ident table.Identifier, vers
...
+	s.Require().NoError(os.Remove(plainPath))
+
+	return gzPath
+}
+
+func (s *HadoopCatalogTestSuite) replaceMetadataWithUUIDName(ident table.Identifier, version int) string {
+	plainPath := s.cat.metadataFilePath(ident, version)
+	uuidPath := filepath.Join(
+		s.cat.metadataDir(ident),
+		"00000-a1b2c3d4-e5f6-7890-abcd-ef1234567890.metadata.json",
```
laskoviymishka: `replaceMetadataWithUUIDName` hardcodes the `00000-` prefix, which parses to version 0. That means the UUID file lands at `files[0]`, so the hint forward-scan in `findMetadataLocation` (`files[hint]`, hint=1) never finds it — the test passes via the `latest` fallback, not the path it looks like it's exercising.

I'd switch the fixture to `00001-<uuid>` so it actually drives the hint lookup, and it matches what real Java/PyIceberg clients write. Pairs with adding the `version <= 0` guard on the UUID branch.

catalog/hadoop/hadoop.go
```diff
@@ -318,28 +332,65 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
...
+	if hint > 0 {
+		if hinted, ok := files[hint]; ok {
+			latest = scanForwardMetadata(files, hinted)
+		}
+	}
+
+	return latest.location, latest.version, nil
+}
+
+func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
```
laskoviymishka, line 473: `findVersion` is now just a thin wrapper that throws away the location, and the only callers left are tests — `LoadTable` and `CommitTable` both go straight to `findMetadataLocation`. So `TestFindVersionScanForward` and friends are testing an intermediary that production no longer uses.

I'd point those tests at `findMetadataLocation` directly and drop `findVersion`, so the discovery tests cover the real call path. Not blocking, but it's dead production code today.

laskoviymishka (changes requested):
Nice fix — this is aimed at the right thing and the core change is correct. Loading the discovered metadata path instead of reconstructing `vN.metadata.json` is the right move, the gzip/UUID-name consistency across `LoadTable`, `CheckTableExists`, and `ListTables` is a real improvement, and dropping `version-hint.text` as a standalone table marker actually brings us closer to Java (`HadoopCatalog.TABLE_FILTER` only matches `.metadata.json`). The existing discovery tests all reproduce the same results under the new logic.

Almost there — one blocking thing before merge: `findMetadataLocation` flattens every error from `scanMetadataFiles` into `ErrNoSuchTable` without threading the original through `%w`. So a permission-denied (or any transient FS failure) on the metadata dir makes `CheckTableExists` return `(false, nil)` — a table we couldn't read now looks identical to one that doesn't exist, which is how you end up clobbering live data on a subsequent create. I'd only map to `ErrNoSuchTable` when `errors.Is(err, fs.ErrNotExist)` and return the original error otherwise, and make `CheckTableExists` propagate it explicitly.

A few things I'd want to settle alongside it:

- `betterThan` is the most novel part of this PR and has no test coverage — I'd pin its tie-break rules (version, then hadoop-over-uuid, then uncompressed-over-gz, then lexicographic) with a focused unit test, plus a one-line comment on why hadoop names win at equal version.
- `CheckTableExists` and `CommitTable` both traded an O(1) hint fast-path for full directory walks — `CommitTable` does two per commit now. Worth keeping `isTableDir` for existence checks and reusing the version `LoadTable` already discovered.
- The UUID branch accepts version 0 (`00000-`) while the hadoop branch guards `<= 0`; the UUID test fixture uses `00000-`, which makes the hint forward-scan path silently vacuous. Move it to `00001-` and mirror the guard.

Fix the error-swallowing and add the `betterThan` coverage and this is good to land — happy to take another pass once those are in.

catalog/hadoop/hadoop.go
```diff
@@ -55,13 +55,68 @@ var _ catalog.Catalog = (*Catalog)(nil)
...
+			location:   path,
+			version:    version,
+			hadoopName: true,
+			compressed: strings.Contains(name, ".gz.metadata.json"),
+		}, true
+	}
+
+	if matches := uuidMetadataPattern.FindStringSubmatch(name); len(matches) == 2 {
+		version, _ := strconv.Atoi(matches[1])
+		if version <= 0 {
```
laskoviymishka: Really nice that this PR teaches the catalog to read UUID-named metadata — that's exactly the interop gap worth closing. One thing I think we need to handle before merge: this guard rejects `00000-<uuid>.metadata.json`, which is the first metadata file other engines write for a brand-new table.

Both reference implementations start the sequence at zero: Java's `BaseMetastoreTableOperations` starts `version` at -1, so the first commit is `(-1)+1 = 0` -> `String.format("%05d-%s%s", 0, uuid, ext)` = `00000-<uuid>.metadata.json`; PyIceberg's `new_table_metadata_file_location(new_version=0)` produces the same; and Java's own `HadoopCatalog.isTableDir` filters on `endsWith(".metadata.json")`, so it's happy with `00000-*`. With `version <= 0` here, a table whose only metadata file is `00000-<uuid>.metadata.json` ends up invisible to this catalog — `LoadTable` returns `ErrNoSuchTable`, `ListTables` skips it, `CheckTableExists` returns false — which is the cross-engine round-trip this PR is trying to enable.

I think the two naming schemes just want different guards: keep `<= 0` on the Hadoop `vN` branch (Java starts `vN` at 1, so `v0` really is invalid there), and use `< 0` on this UUID branch. A small fixture with `00000-<uuid>.metadata.json` asserting `isTableDir`/`LoadTable` find it would lock it in.

(Tiny related thing: the doc-comment edit just above, from `00000` to `00001`, points the wrong way for the same reason — `00000` was correct — so I'd revert that hunk alongside this.)

catalog/hadoop/hadoop.go
```diff
@@ -318,28 +341,70 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
...
+	if latest.location == "" {
+		return "", 0, fmt.Errorf("hadoop catalog: no metadata files found for table %s: %w",
 			strings.Join(ident, "."), catalog.ErrNoSuchTable)
 	}
 
-	return c.scanForward(ident, maxVer), nil
+	hint := c.readVersionHint(ident)
+	if hint > 0 {
+		if hinted, ok := files[hint]; ok {
+			latest = scanForwardMetadata(files, hinted)
```
laskoviymishka: I think there's a subtle case here that can quietly re-open the very bug this PR is closing — flagging it because three of us landed on it independently.

`latest` already holds the global best across every discovered file, and then when the hint resolves we overwrite it with `scanForwardMetadata`, which walks contiguous versions and stops at the first gap. So with files `{v1, v2, v5}` and `hint=1`, the walk goes 1 -> 2 -> (no v3) -> stops at v2, and we load v2 while v5 is sitting right there.

The old code only trusted the hint when that exact version existed and otherwise fell back to the max, so it didn't trip on gaps. A one-line guard keeps the fast path while staying safe:

```go
if hint > 0 {
    if hinted, ok := files[hint]; ok {
        if walked := scanForwardMetadata(files, hinted); walked.betterThan(latest) {
            latest = walked
        }
    }
}
```

A `{v1,v2,v5}+hint=1` test would be a nice regression guard here — it's currently uncovered.

catalog/hadoop/hadoop.go
```diff
@@ -55,13 +55,68 @@ var _ catalog.Catalog = (*Catalog)(nil)
...
+type metadataFile struct {
+	location   string
+	version    int
+	hadoopName bool
+	compressed bool
+}
+
+func metadataFileFromName(path, name string) (metadataFile, bool) {
+	if matches := versionPattern.FindStringSubmatch(name); len(matches) == 2 {
+		version, _ := strconv.Atoi(matches[1])
```
laskoviymishka: Small one: the version capture is unbounded (`[0-9]+`), so something like `v99999999999999999999.metadata.json` overflows `Atoi` to `(0, ErrRange)`. We drop the error and then `version <= 0` rejects it — so the file does get skipped, but for the wrong reason, and a legitimately huge version would vanish the same way. I'd check the error explicitly here (and on the UUID branch, with `< 0` per the note above):

```go
version, err := strconv.Atoi(matches[1])
if err != nil || version <= 0 {
    return metadataFile{}, false
}
```

catalog/hadoop/hadoop.go
```diff
@@ -580,7 +657,11 @@ func (c *Catalog) ListTables(_ context.Context, ns table.Identifier) iter.Seq2[t
 			}
 
 			// Skip anything that is not a table directory.
-			if !isTableDir(c.filesystem, path) {
+			isTable, err := isTableDir(c.filesystem, path)
```
laskoviymishka: Returning the `isTableDir` error instead of swallowing it is a good instinct. The one thing I'd weigh: returning it straight out of the `WalkDir` callback means a single unreadable metadata subdir now aborts the whole `ListTables` (and `ListNamespaces` does the same), where the old code skipped the bad entry and kept going.

Java's `HadoopCatalog.isTableDir` has a `shouldSuppressPermissionError` guard for exactly this — it log-warns and treats the dir as not-a-table so listing survives mixed-permission warehouses. I'd lean toward matching that (suppress permission errors, skip the dir), but I'm happy either way as long as it's a conscious choice — which behavior do you want here?

catalog/hadoop/hadoop.go
```diff
@@ -318,28 +341,70 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
...
+	}
+}
+
+func (c *Catalog) findMetadataLocation(ident table.Identifier) (string, int, error) {
+	files, latest, err := c.scanMetadataFiles(ident)
 	if err != nil {
-		return 0, fmt.Errorf("hadoop catalog: cannot read metadata directory for %s: %w",
-			strings.Join(ident, "."), catalog.ErrNoSuchTable)
+		if errors.Is(err, fs.ErrNotExist) {
+			return "", 0, fmt.Errorf("%w: cannot read metadata directory for %s: %w",
```
laskoviymishka: Two small things on this branch. The message leads with the sentinel (`"table does not exist: cannot read metadata directory for ns.tbl: file does not exist"`), while everything else in this file — including the sibling branch right below — leads with the `hadoop catalog:` prefix. Keeping the prefix and moving the sentinel into the wrap chain reads more consistently:

```go
return "", 0, fmt.Errorf("hadoop catalog: cannot read metadata directory for %s: %w: %w",
    strings.Join(ident, "."), catalog.ErrNoSuchTable, err)
```

The slightly bigger one: this maps `fs.ErrNotExist` on the metadata dir to `ErrNoSuchTable`, while `isTableDir` maps the same condition to `(false, nil)`. Same physical state, two interpretations across the two callpaths — might be worth picking one as canonical so they agree.

catalog/hadoop/hadoop.go
```diff
@@ -55,13 +55,68 @@ var _ catalog.Catalog = (*Catalog)(nil)
...
+func (m metadataFile) betterThan(current metadataFile) bool {
+	switch {
+	case current.location == "":
+		return true
+	case m.version != current.version:
+		return m.version > current.version
+	case m.hadoopName != current.hadoopName:
+		// If both naming styles exist for one version, prefer the Hadoop vN
+		// file because this catalog writes and conflict-checks that sequence.
+		return m.hadoopName
```
laskoviymishka, line 113: This tie-break reads fine for the common case; I just want to flag the assumption baked in. Preferring `vN` over `00001-<uuid>` for the same version treats the UUID `<seq>` and the Hadoop version as the same namespace — but Java derives its version purely from `v<N>.metadata.json` + version-hint and never parses UUID names; those come from object-store/REST writers where the sequence is just a monotonic counter. So for a single writer this is great, but for a genuinely mixed-writer table the two files at "version 1" aren't guaranteed to be the same snapshot. I don't think this needs to change here — a comment noting it's a best-effort heuristic rather than spec would help the next reader. wdyt?

catalog/hadoop/hadoop.go
```diff
@@ -318,28 +341,70 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
...
+			latest = file
 		}
 
 		return nil
 	})
+
+	return byVersion, latest, err
+}
+
+func scanForwardMetadata(files map[int]metadataFile, start metadataFile) metadataFile {
```
laskoviymishka: Nice clean helper. One behavior worth a comment: because it walks `files` purely by version, it'll happily chain from a UUID file at vN to a Hadoop `vN+1` (and back) — reasonable for a table that switched writers, but novel relative to Java (which only walks the `vN` sequence) and currently untested. A one-line note that the walk is naming-agnostic, plus a small mixed-style test, would pin the intent.

catalog/hadoop/hadoop_test.go
```diff
@@ -1222,6 +1382,17 @@ func (s *HadoopCatalogTestSuite) TestLoadTableNotExists() {
 	s.ErrorIs(err, catalog.ErrNoSuchTable)
 }
 
+func (s *HadoopCatalogTestSuite) TestLoadTablePropagatesMetadataReadError() {
+	ctx := context.Background()
+	walkErr := errors.New("metadata walk failed")
+	s.cat.filesystem = failingWalkFS{HadoopCatalogFS: s.cat.filesystem, err: walkErr}
```
laskoviymishka: This and `TestCheckTableExistsPropagatesMetadataReadError` swap `s.cat.filesystem` in place and don't restore it. If `SetupTest` rebuilds `s.cat` fresh per test, we're totally fine — I just couldn't see it in the diff. If it reuses the instance, a later test could inherit `failingWalkFS` and flake depending on order; a deferred restore would make it bulletproof. Could you confirm `SetupTest` reconstructs `s.cat`?

catalog/hadoop/hadoop_test.go
```diff
@@ -1240,6 +1411,67 @@ func (s *HadoopCatalogTestSuite) TestLoadTableStaleHint() {
...
+	metaDir := s.cat.metadataDir(ident)
+	s.Require().NoError(os.MkdirAll(metaDir, 0o755))
+	s.cat.writeVersionHint(ident, 1)
+
+	exists, err := s.cat.CheckTableExists(ctx, ident)
+	s.Require().NoError(err)
+	s.False(exists)
+
+	var listed []table.Identifier
+	for ident, err := range s.cat.ListTables(ctx, []string{"ns"}) {
```
laskoviymishka: Tiny nit: the range variable `ident` shadows the outer `ident` for this test. No bug under Go 1.22+ scoping, but vet/staticcheck flags it and it reads a little confusingly — renaming the range var to `tblIdent` would clear it up.

catalog/hadoop/hadoop_test.go
```diff
@@ -54,6 +58,125 @@ func TestHadoopCatalogTestSuite(t *testing.T) {
...
 }
 
+func (s *HadoopCatalogTestSuite) requireIsTableDir(path string) bool {
+	isTable, err := isTableDir(s.cat.filesystem, path)
+	s.Require().NoError(err)
+
+	return isTable
+}
+
+func TestMetadataFileBetterThan(t *testing.T) {
```
laskoviymishka, line 141: Solid table test. Two small things: it uses bare `t.Fatalf` while the rest of the suite is on testify — `require.Equal(t, tt.want, got)` would match the house style and give a nicer diff on failure. And since `scanMetadataFiles` relies on `betterThan` being a strict weak ordering under nondeterministic map iteration, a couple of property rows (`x.betterThan(x) == false`, plus the case where both compression and naming differ at the same version) would lock that contract down.

catalog/hadoop/hadoop.go
```diff
@@ -55,13 +55,68 @@ var _ catalog.Catalog = (*Catalog)(nil)
...
+	case m.version != current.version:
+		return m.version > current.version
+	case m.hadoopName != current.hadoopName:
+		// If both naming styles exist for one version, prefer the Hadoop vN
+		// file because this catalog writes and conflict-checks that sequence.
+		return m.hadoopName
+	case m.compressed != current.compressed:
+		return !m.compressed
+	default:
+		return m.location > current.location
```
laskoviymishka, line 119: Just a readability nit: the lexicographic fallback is perfectly fine for determinism, but "higher path wins" isn't a meaningful Iceberg ordering, so a future reader may pause on it. A one-line comment that it's purely a stable tie-breaker under map-iteration order (and that within a single scan all locations share a parent dir, so it reduces to filename order) would explain itself.

catalog/hadoop/hadoop.go
```diff
@@ -197,11 +252,10 @@ func (c *Catalog) defaultTableLocation(ident table.Identifier) string {
 }
 
 // isTableDir reports whether path is a table directory by checking for
-// metadata files in its metadata/ subdirectory. It recognizes:
+// loadable metadata files in its metadata/ subdirectory. It recognizes:
 //   - v*.metadata.json (Hadoop catalog format)
 //   - <seq>-<uuid>.metadata.json (Java/PyIceberg format)
-//   - version-hint.text
-func isTableDir(filesystem HadoopCatalogFS, path string) bool {
+func isTableDir(filesystem HadoopCatalogFS, path string) (bool, error) {
```
laskoviymishka: Good change — folding detection into `metadataFileFromName` is cleaner. Worth calling out that a directory with only `version-hint.text` and no metadata file is no longer treated as a table. I checked, and this matches Java (its `isTableDir` never counted `version-hint.text`), so it's correct parity rather than a regression — but since it's a user-visible change to `CheckTableExists`/`ListTables`, a line in the PR description noting it's intentional would help.

laskoviymishka (changes requested):
Thanks for taking this on — and for the thorough follow-up round (the error propagation, the version reuse in `CommitTable`, the tie-break tests). The core change is the right one: the old path rebuilt `v{N}.metadata.json` from a bare version number, which silently broke for gzip- and UUID-named metadata, and loading the actual discovered file fixes that cleanly. The `loadTable`/`LoadTable` split is tidy, and I like that the new gzip/UUID tests assert on `MetadataLocation()` — exactly the surface that was wrong before.

I left inline notes; most are small, but there are a few I'd love to settle before this lands:

- The UUID branch rejects `00000-<uuid>.metadata.json` (`version <= 0`), but that's the first metadata file Java and PyIceberg write for a brand-new table, so a table first-committed by another engine ends up invisible here. Splitting the guard — keep `<= 0` for `vN`, use `< 0` for the UUID branch — and reverting the `00000`->`00001` comment edit fixes it.
- The hint-walk in `findMetadataLocation` can adopt a lower version than the global best when there's a gap in the sequence (`{v1,v2,v5}` + `hint=1` loads v2, skipping v5). Only adopting the walked result when it's `betterThan(latest)` keeps it safe.
- A discarded `strconv.Atoi` error means an oversized version is rejected for the wrong reason — worth checking the error explicitly.

Smaller stuff: listing now fails hard if a single metadata subdir is unreadable (Java suppresses permission errors there — worth deciding which behavior you want); an error-message prefix inconsistency; and a few documentation/test nits inline.

A couple of things I dug into that are totally fine: reusing `currentVersion` in `CommitTable` doesn't widen the commit race — the Stat-then-Rename check is still the authoritative guard — and dropping the hint-only-directory case is correct Java parity (worth a line in the description, since it's user-visible).

Really close. Once the `00000` handling and the hint-walk guard are in, I'd be glad to take another pass.

catalog/hadoop/hadoop.go
```diff
@@ -492,15 +598,6 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
...
-	var currentVersion int
-
-	if current != nil {
-		currentVersion, err = c.findVersion(ident)
-		if err != nil {
-			return nil, "", err
-		}
-	}
-
 	newVersion := currentVersion + 1
```
zeroshade, line 650: [Major] Reusing `currentVersion` here is the right call. The gap is downstream: the pre-rename conflict check a few lines below (`Stat(newMetaPath)`, where `newMetaPath` is `v{newVersion}.metadata.json`) only probes the Hadoop-style filename. Now that this PR teaches the catalog to *read* UUID- and gzip-named metadata, a concurrent writer (Java/PyIceberg) that lands `%05d-<uuid>.metadata.json` or `v{newVersion}.gz.metadata.json` for the same version won't be detected, and we can end up writing a second `v{newVersion}.metadata.json` for the same logical version.

Suggest rescanning the metadata dir (via `scanMetadataFiles`/`metadataFileFromName`) for any recognized file whose parsed version == `newVersion` immediately before the rename, failing if one exists - plus a regression test that stages a UUID/gzip next-version file between load and commit. (HadoopCatalog isn't a substitute for a locking catalog under concurrent multi-writer use, but this keeps the write/conflict path consistent with the new read support.)

zeroshade (changes requested):
Thanks for the thorough round - the stale-hint/version-gap fix, the error-propagation fix (no more `(false, nil)` when the metadata dir is unreadable), the `00000-<uuid>` handling, the `strconv.Atoi` overflow checks, and the `betterThan` tie-break tests all look good. One remaining item before merge; details inline.

catalog/hadoop/hadoop.go
```diff
@@ -318,28 +392,70 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
...
+	if latest.location == "" {
+		return "", 0, fmt.Errorf("hadoop catalog: no metadata files found for table %s: %w",
 			strings.Join(ident, "."), catalog.ErrNoSuchTable)
 	}
 
-	return c.scanForward(ident, maxVer), nil
+	hint := c.readVersionHint(ident)
+	if hint > 0 {
+		if hinted, ok := files[hint]; ok {
+			if walked := scanForwardMetadata(files, hinted); walked.betterThan(latest) {
```
zeroshade: This hint-forward walk reads as dead code: `latest` is already the global best via `betterThan`, so `walked.betterThan(latest)` can never be true. Consider removing `scanForwardMetadata` + this check, or documenting why it stays.

catalog/hadoop/hadoop.go
```diff
@@ -270,43 +343,44 @@ func (c *Catalog) writeVersionHint(ident table.Identifier, version int) {
...
+	return byVersion, latest, err
 }
 
-func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
-	hint := c.readVersionHint(ident)
-	if hint > 0 && c.metadataVersionExists(ident, hint) {
-		return c.scanForward(ident, hint), nil
-	}
-
+func (c *Catalog) metadataVersionExists(ident table.Identifier, version int) (bool, error) {
```
zeroshade, line 420: `metadataVersionExists` now does a full directory walk on every `CommitTable` (hot path); with Hadoop naming `v10` sorts before `v2`, so it usually scans most entries before short-circuiting. Consider targeted `Stat`s for the plain/gz/UUID names, or document the tradeoff if the walk is intentional (UUID-race coverage).

catalog/hadoop/hadoop.go
```diff
@@ -419,27 +539,44 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
...
 	if err := validateTableIdentifier(ident); err != nil {
 		return false, nil
 	}
 
-	return isTableDir(c.filesystem, c.tableToPath(ident)), nil
+	exists, err := isTableDir(c.filesystem, c.tableToPath(ident))
+	if err != nil {
+		return false, err
+	}
+	if !exists {
```
zeroshade: Two small things: (1) `CheckTableExists` now propagates `isTableDir` errors where it previously returned `(false, nil)` — asymmetric vs. the tolerant listing path; confirm intended or route via `isTableDirForListing`. (2) `if !exists { return false, nil }; return true, nil` collapses to `return exists, nil`.

zeroshade (changes requested):
Thanks @fallintoplace — correct overall; a few low-severity notes:

laskoviymishka (approved):
Need a rebase, but overall looks good to me now!

catalog/hadoop/hadoop.go
```diff
@@ -457,27 +558,41 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
...
+	return tbl, err
 }
 
 func (c *Catalog) CheckTableExists(_ context.Context, ident table.Identifier) (bool, error) {
 	if err := validateTableIdentifier(ident); err != nil {
 		return false, nil
 	}
 
-	return isTableDir(c.filesystem, c.isLocal, c.tableToPath(ident)), nil
+	exists, err := isTableDir(c.filesystem, c.isLocal, c.tableToPath(ident))
```
zeroshade, line 592: Optional (non-blocking): `CheckTableExists` now propagates `isTableDir` errors, while `ListTables`/`ListNamespaces` intentionally swallow permission errors via `isTableDirForListing`. That strict-vs-tolerant split is defensible, but a one-line comment here explaining the distinction would save the next reader a double-take.

zeroshade (approved):
Re-reviewed. The UUID/gzip next-version conflict detection (`metadataVersionExists` + `scanMetadataFiles`) resolves my earlier blocking concern, discovery now canonicalizes the resolved path, and coverage is solid. Agreed with the prior approval — LGTM. One optional doc-only note inline.

PR #1293, opened 2026-06-24
fix: drop redundant hadoop namespace stat

### LINKED ISSUE #1273: fix: drop the redundant Stat before WalkDir in hadoop DropNamespace
Follow-up to #963.

#963 deliberately dropped the standalone existence check and let `ReadDir` map `fs.ErrNotExist` to `ErrNoSuchNamespace`. The #1185 interface-minimization refactor re-introduced a `Stat` before `WalkDir` in `DropNamespace` — but `WalkDir` already propagates `fs.ErrNotExist` and the handler already converts it, so the leading `Stat` is redundant and only widens the check-to-use window where a concurrent caller can delete or recreate the directory between the two calls.

Drop the standalone `Stat` and rely on `WalkDir`'s error handling; if the not-a-directory case needs preserving, fold it into the `WalkDir` callback on the root entry rather than a separate pre-check.

Revanth14:
Fixes #1273.

## Summary

Drops the redundant pre-`Stat` call from Hadoop catalog `DropNamespace` and relies on `WalkDir` to surface missing namespace paths. The root `WalkDir` entry still preserves the existing behavior for a file at the namespace path by returning `ErrNoSuchNamespace`.

## Testing

- `go test ./catalog/hadoop`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -670,21 +670,18 @@ func (c *Catalog) DropNamespace(_ context.Context, ns table.Identifier) error {
 
 	path := c.namespaceToPath(ns)
 
-	info, err := c.filesystem.Stat(path)
-	if errors.Is(err, fs.ErrNotExist) || (err == nil && !info.IsDir()) {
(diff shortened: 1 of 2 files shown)


laskoviymishka (approved):
Nice work here!

PR #1296, opened 2026-06-24
feat(compaction): force compaction on file-scoped delete dead space

twuebi:
The candidate planner only counted positional and equality delete files toward delete pressure, so a right-sized data file accumulating file-scoped deletes (a deletion vector, or a path-scoped positional delete) was never rewritten — its dead rows persisted as read amplification indefinitely.

Add a DeleteRatioThreshold (default 0.3, mirroring Java's delete-ratio-threshold). A file is forced into compaction when its file-scoped deletes shadow at least that fraction of its rows. Following Java's BinPackRewriteFilePlanner.tooHighDeleteRatio, the deleted-row count sums the record counts of file-scoped deletes only — deletion vectors and path-scoped positional deletes (those carrying a referenced data file), identified by isFileScoped (= referenced data file present, matching ContentFileUtil.isFileScoped) — capped at the data file's record count. Equality deletes and partition-scoped positional deletes are not file-scoped and stay under the existing count rule, to which deletion vectors are now also added.

Unlike Java, a threshold of 0 disables the rule so a zero-valued Config stays valid; the default remains 0.3.

### DIFF
--- a/table/compaction/compaction.go
+++ b/table/compaction/compaction.go
@@ -19,9 +19,10 @@
 package compaction
 
 import (
+	"bytes"
 	"fmt"
 
-	iceberg "github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go"
 	"github.com/apache/iceberg-go/internal"
 	"github.com/apache/iceberg-go/table"
 )
@@ -53,10 +54,11 @@ type Config struct {
 	DeleteFileThreshold int
 
 	// DeleteRatioThreshold forces a data file into compaction when the fraction
-	// of its rows shadowed by a deletion vector reaches this value, regardless
-	// of file size. Deletion vectors are 1:1 with their data file and carry an
+	// of its rows shadowed by file-scoped deletes (deletion vectors and
+	// path-scoped positional deletes) reaches this value, regardless of file
+	// size. File-scoped deletes apply to exactly one data file and carry an
 	// exact deleted-row count, so size-based selection alone never reclaims a
-	// right-sized file's DV dead space; this rule does. Range [0,1]; 0 disables
+	// right-sized file's dead space; this rule does. Range [0,1]; 0 disables
 	// it. Default: DefaultDeleteRatioThreshold.
 	DeleteRatioThreshold float64
 
@@ -96,8 +98,9 @@ const (
 	DefaultPackingLookback uint = 128
 
 	// DefaultDeleteRatioThreshold mirrors Java's delete-ratio default: a file
-	// whose deletion vector shadows at least 30% of its rows is rewritten.
-	DefaultDeleteRatioThreshold = 0.3
+	// whose file-scoped deletes (deletion vectors and path-scoped positional
+	// deletes) shadow at least 30% of its rows is rewritten.
+	DefaultDeleteRatioThreshold float64 = 0.3
 )
 
 // DefaultConfig returns a Config with production defaults.
@@ -138,7 +141,7 @@ func (cfg Config) Validate() error {
 		return fmt.Errorf("delete file threshold must be >= 1, got %d", cfg.DeleteFileThreshold)
 	}
 	if cfg.DeleteRatioThreshold < 0 || cfg.DeleteRatioThreshold > 1 {
-		return fmt.Errorf("delete ratio threshold must be in [0,1], got %v", cfg.DeleteRatioThreshold)
+		return fmt.Errorf("delete ratio threshold must be in [0,1], got %g", cfg.DeleteRatioThreshold)
 	}
 
 	return nil
@@ -276,6 +279,13 @@ func (cfg Config) PlanCompaction(tasks []table.FileScanTask) (Plan, error) {
 // isCandidate returns true if a file should be considered for compaction.
 func (cfg Config) isCandidate(t table.FileScanTask) bool {
 	size := t.File.FileSizeBytes()
+
+	// Deletion vectors are counted here alongside positional and equality
+	// delete files, matching Java's unified task.deletes() count. A data file
+	// has at most one DV, so at the minimum DeleteFileThreshold of 1 a single
+	// DV trips this rule before the ratio rule below runs; at the default of 5
+	// it does not. The ratio rule still owns right-sized DV-backed files at any
+	// threshold above 1.
 	deleteCount := len(t.DeleteFiles) + len(t.EqualityDeleteFiles) + len(t.DeletionVectorFiles)
 
 	// Too many delete files — always compact regardless of size.
@@ -286,16 +296,13 @@ func (cfg Config) isCandidate(t table.FileScanTask) bool {
 	// Enough rows shadowed by file-scoped deletes — compact to reclaim the dead
 	// space. Mirrors Java BinPackRewriteFilePlanner.tooHighDeleteRatio: sum the
 	// record counts of file-scoped deletes (deletion vectors and path-scoped
-	// positional deletes — those carrying a referenced data file), cap at the
-	// data file's record count, and compare the ratio. Equality deletes and
-	// partition-scoped positional deletes are not file-scoped (their counts
-	// cannot be attributed to one data file) and are left to the count rule.
+	// positional deletes — see isFileScoped), cap at the data file's record
+	// count, and compare the ratio. Equality deletes and partition-scoped
+	// positional deletes are not file-scoped (their counts cannot be attributed
+	// to one data file) and are left to the count rule.
 	if cfg.DeleteRatioThreshold > 0 {
 		if records := t.File.Count(); records > 0 {
-			deleted := fileScopedDeletedRows(t)
-			if deleted > records {
-				deleted = records
-			}
+			deleted := min(fileScopedDeletedRows(t), records)
 			if float64(deleted)/float64(records) >= cfg.DeleteRatioThreshold {
 				return true
 			}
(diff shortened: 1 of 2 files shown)


table/compaction/compaction.go
```diff
@@ -279,3 +315,31 @@ func (cfg Config) isCandidate(t table.FileScanTask) bool {
...
+	}
+
+	return deleted
+}
+
+// isFileScoped reports whether a delete file applies to exactly one data file.
+// Mirrors Java ContentFileUtil.isFileScoped: deletion vectors and path-scoped
+// positional deletes carry a referenced data file; partition-scoped positional
+// deletes and equality deletes do not.
+func isFileScoped(d iceberg.DataFile) bool {
```
laskoviymishka, line 353: This only catches DVs in practice, not the path-scoped positional deletes the PR is really aiming at.

`referenced_data_file` is optional in V2 — the Java/PyIceberg/iceberg-go pos-delete writers don't set it, they signal file-scope through equal `file_path` lower/upper bounds. Java's `ContentFileUtil.referencedDataFile` does a two-step check: return the explicit ref if present, else fall back to `lower_bounds[file_path] == upper_bounds[file_path]`. We only do the first step, so any V2 table accumulating Parquet pos-delete rows on right-sized files never gets compacted by this rule — which is the dominant case it's meant to handle. (In our own scan planning, `matchDeletesToData` never populates `referenced_data_file` on `DeleteFiles`, so the path-scoped test only passes because the fixture sets it.)

I'd add the bounds fallback, and while we're here drop out eq-deletes up front to match Java (it returns null for them regardless):

```go
func isFileScoped(d iceberg.DataFile) bool {
	if d.ContentType() == iceberg.EntryContentEqDeletes {
		return false
	}
	if d.ReferencedDataFile() != nil {
		return true
	}
	const filePathFieldID = 2147483546
	lb := d.LowerBoundValues()[filePathFieldID]
	ub := d.UpperBoundValues()[filePathFieldID]
	return len(lb) > 0 && bytes.Equal(lb, ub)
}
```

table/compaction/compaction.go
```diff
@@ -259,13 +276,32 @@ func (cfg Config) PlanCompaction(tasks []table.FileScanTask) (Plan, error) {
 // isCandidate returns true if a file should be considered for compaction.
 func (cfg Config) isCandidate(t table.FileScanTask) bool {
 	size := t.File.FileSizeBytes()
-	deleteCount := len(t.DeleteFiles) + len(t.EqualityDeleteFiles)
+	deleteCount := len(t.DeleteFiles) + len(t.EqualityDeleteFiles) + len(t.DeletionVectorFiles)
```
laskoviymishka, line 289: I'm of two minds on folding DVs into `deleteCount`. It matches Java's unified `task.deletes()` list, so there's a parity argument for it.

But the count rule is for "lots of separate delete files piling up," and a file has at most one DV — so at `DeleteFileThreshold=1` (the minimum valid value) a single DV trips the count rule before the ratio rule ever runs, which makes the ratio rule dead code for DV-backed files at that threshold and inflates `group.DeleteFileCount` by 1 per DV-backed file for downstream consumers. Was the intent to have both signals fire on DVs, or should DVs stay out of the count and be owned solely by the ratio rule? wdyt?

table/compaction/compaction.go
```diff
@@ -51,6 +52,14 @@ type Config struct {
 	// Default: 5.
 	DeleteFileThreshold int
 
+	// DeleteRatioThreshold forces a data file into compaction when the fraction
+	// of its rows shadowed by a deletion vector reaches this value, regardless
+	// of file size. Deletion vectors are 1:1 with their data file and carry an
+	// exact deleted-row count, so size-based selection alone never reclaims a
+	// right-sized file's DV dead space; this rule does. Range [0,1]; 0 disables
+	// it. Default: DefaultDeleteRatioThreshold.
+	DeleteRatioThreshold float64
```
laskoviymishka, line 63: The doc says "deletion vector" throughout, but the rule (correctly) also counts path-scoped positional deletes via `fileScopedDeletedRows`. I'd broaden the wording to "file-scoped deletes (deletion vectors and path-scoped positional deletes)" so the scope isn't misleading — same applies to the `DefaultDeleteRatioThreshold` const comment below.

table/compaction/compaction.go
```diff
@@ -123,6 +137,9 @@ func (cfg Config) Validate() error {
 	if cfg.DeleteFileThreshold < 1 {
 		return fmt.Errorf("delete file threshold must be >= 1, got %d", cfg.DeleteFileThreshold)
 	}
+	if cfg.DeleteRatioThreshold < 0 || cfg.DeleteRatioThreshold > 1 {
+		return fmt.Errorf("delete ratio threshold must be in [0,1], got %v", cfg.DeleteRatioThreshold)
```
laskoviymishka: `%g` reads better than `%v` for a float64 here, and it's closer to the sibling `%d` int errors above in being explicit about the type.

table/compaction/compaction.go
```diff
@@ -21,6 +21,7 @@ package compaction
 import (
 	"fmt"
 
+	iceberg "github.com/apache/iceberg-go"
```
laskoviymishka: The alias is an identity alias (the package is already named `iceberg`), and every other file in this package imports it unaliased — I'd drop it for consistency.

table/compaction/compaction.go
```diff
@@ -85,19 +94,24 @@ const (
 
 	// DefaultPackingLookback is the default packing lookback.
 	DefaultPackingLookback uint = 128
+
+	// DefaultDeleteRatioThreshold mirrors Java's delete-ratio default: a file
+	// whose deletion vector shadows at least 30% of its rows is rewritten.
+	DefaultDeleteRatioThreshold = 0.3
```
laskoviymishka: The siblings are explicitly typed (`DefaultMinInputFiles uint = 5`, `DefaultPackingLookback uint = 128`). I'd make this `float64` too — `DefaultDeleteRatioThreshold float64 = 0.3` — otherwise it's an untyped constant in a block of typed ones.

table/compaction/compaction_test.go
```diff
@@ -261,6 +308,109 @@ func TestPlanCompaction_DeleteFilesForcesCompaction(t *testing.T) {
...
+// size-based selection alone would skip it — isolating the deletion-vector
+// ratio as the only thing that can force compaction.
+func dvRatioConfig() compaction.Config {
+	return compaction.Config{
+		TargetFileSizeBytes:  2600 * 1024 * 1024,
+		MinFileSizeBytes:     384 * 1024 * 1024,
+		MaxFileSizeBytes:     3000 * 1024 * 1024,
+		MinInputFiles:        2,
+		DeleteFileThreshold:  5,
+		DeleteRatioThreshold: 0.3,
```
laskoviymishka: `dvRatioConfig` hardcodes `0.3` while `PackingLookback` just above uses the exported const. If the default ever moves, this helper silently keeps testing the old value — I'd use `compaction.DefaultDeleteRatioThreshold` here too.

laskoviymishka (changes requested):
Nice addition — the ratio rule is the right tool for reclaiming file-scoped delete dead space that size-based selection structurally can't see, and the test config that pins files right-sized so only the ratio can force compaction is a clean way to isolate the signal.

I'd hold it before merging, though, because the trigger condition misses the case the PR is really motivated by. `isFileScoped` is `ReferencedDataFile() != nil`, but `referenced_data_file` is optional in V2 and the Java/PyIceberg/iceberg-go positional-delete writers don't set it — they signal file-scope through equal `file_path` lower/upper bounds instead. Java's `ContentFileUtil.referencedDataFile` falls back to those bounds when the explicit field is absent; we don't. As written, only Puffin DVs hit the new rule, and a table accumulating ordinary V2 Parquet pos-deletes on right-sized files never gets compacted. The path-scoped test passes only because the fixture sets `referenced` — in real scan planning `matchDeletesToData` never does. I'd add the bounds fallback (and drop eq-deletes up front, matching Java) — details inline.

Separately, I'd like to settle whether deletion vectors belong in the count-based `deleteCount` now that the ratio rule owns them. It's defensible as Java parity, but at `DeleteFileThreshold=1` a single DV trips the count rule before the ratio rule runs, making the ratio rule dead code for DV-backed files and inflating `group.DeleteFileCount` downstream. Not blocking on a particular answer — just want the intent to be explicit.

- `isFileScoped` falls back to equal `file_path` lower/upper bounds when `referenced_data_file` is absent, and returns false for equality deletes.
- a decision on whether DVs should stay in `deleteCount` (or be owned solely by the ratio rule), with the `DeleteFileThreshold=1` interaction documented or tested.
- at-threshold (exactly 30%) and negative-ratio validation tests, so a `>` vs `>=` or sign slip can't pass green.
- the godoc/const wording broadened from "deletion vector" to "file-scoped deletes (deletion vectors and path-scoped positional deletes)".

table/compaction/compaction.go
twuebi: done, added the eq delete short circuit and equal file path fallback & added a test exercising the fallback

table/compaction/compaction.go
twuebi: went for parity with java, which as far as i understand unifies deletes, so a single dv may trip the count rule, but at default 5 it wouldn't. happy to go the other way if you prefer

table/compaction/compaction.go
twuebi: broadened to "file-scoped deletes"

table/compaction/compaction.go
twuebi: fixed

table/compaction/compaction.go
twuebi: dropped

table/compaction/compaction.go
twuebi: fixed

table/compaction/compaction_test.go
twuebi: fixed

laskoviymishka (approved):
<3

PR #1300, opened 2026-06-24
fix(catalog/hadoop): publish metadata without replacement

fallintoplace:
Summary

- add a no-replace rename path for Hadoop metadata publishes
- use it when creating `v1.metadata.json` and when committing the next table metadata version
- return `ErrTableAlreadyExists` for create races and `table.ErrCommitFailed` for commit races
- add deterministic concurrent create/commit coverage where two writers publish the same metadata version

Why

Hadoop commits wrote metadata to a temp file, checked whether the final path existed, then used `Rename` to publish it. On local filesystems `os.Rename` replaces the destination, so another writer could win between the existence check and rename and then be overwritten by the later commit.

The metadata file publish now fails if the final path already exists. `version-hint.text` stays best-effort after the metadata file is committed.

Testing

- go test ./catalog/hadoop -run 'TestHadoopCatalogTestSuite/(TestCreateTableConcurrentMetadataPublishConflict|TestCommitTableConcurrentMetadataPublishConflict|TestCommitTableConflictDetection)' -count=1
- go test ./io -run '^TestLocalFSRenameNoReplaceDoesNotOverwrite$' -count=1
- go test ./io -count=1
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m ./io
- git diff --check

### DIFF
--- a/io/local.go
+++ b/io/local.go
@@ -80,3 +80,18 @@ func (LocalFS) Stat(name string) (fs.FileInfo, error) {
 func (LocalFS) Rename(oldpath, newpath string) error {
 	return os.Rename(strings.TrimPrefix(oldpath, "file://"), strings.TrimPrefix(newpath, "file://"))
 }
+
+func (LocalFS) RenameNoReplace(oldpath, newpath string) error {
+	oldpath = strings.TrimPrefix(oldpath, "file://")
+	newpath = strings.TrimPrefix(newpath, "file://")
+
+	if err := os.Link(oldpath, newpath); err != nil {
+		return err
+	}
+
+	// Publishing already succeeded once the hard link exists. Removing the
+	// source temp file is best-effort cleanup.
+	_ = os.Remove(oldpath)
+
+	return nil
+}
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -400,10 +400,8 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 		return nil, fmt.Errorf("hadoop catalog: failed to write table metadata: %w", err)
 	}
 
-	if err := c.filesystem.Rename(tempPath, metaPath); err != nil {
-		_ = c.filesystem.Remove(tempPath)
-
-		return nil, fmt.Errorf("hadoop catalog: failed to commit metadata file: %w", err)
+	if err := c.commitMetadataFile(ident, tempPath, metaPath, catalog.ErrTableAlreadyExists); err != nil {
+		return nil, err
 	}
 
 	c.writeVersionHint(ident, version)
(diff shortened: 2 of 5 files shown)


io/local.go
```diff
@@ -80,3 +80,16 @@ func (LocalFS) Stat(name string) (fs.FileInfo, error) {
 func (LocalFS) Rename(oldpath, newpath string) error {
 	return os.Rename(strings.TrimPrefix(oldpath, "file://"), strings.TrimPrefix(newpath, "file://"))
 }
+
+func (LocalFS) RenameNoReplace(oldpath, newpath string) error {
+	oldpath = strings.TrimPrefix(oldpath, "file://")
+	newpath = strings.TrimPrefix(newpath, "file://")
+
+	if err := os.Link(oldpath, newpath); err != nil {
```
zeroshade, line 88: `os.Link` fails with `EXDEV` across filesystems. Here temp and final share the metadata dir (same FS), so it is fine in practice, but a one-line doc note on the method would help since this is a new exported `io` primitive. On Linux, `unix.Renameat2(..., RENAME_NOREPLACE)` is the truly-atomic kernel primitive; the link-based approach is the conventional portable fallback. Reasonable as-is.

io/local.go
```diff
@@ -80,3 +80,16 @@ func (LocalFS) Stat(name string) (fs.FileInfo, error) {
...
+		return err
+	}
+
+	_ = os.Remove(oldpath)
```
zeroshade, line 94: Dropped `os.Remove` error: the publish already succeeded via `os.Link`, so this only risks leaking a stray temp file if removal fails (correctness is unaffected). A short comment noting it is intentional best-effort cleanup would prevent a future "unchecked error" flag.

zeroshade (approved):
Approve. Correct fix for a real TOCTOU race, well tested.

**Bug:** `CreateTable`/`CommitTable` did `Stat`-then-`Rename`, and `os.Rename` replaces the destination on local FS, so two writers racing the same version could both "succeed" with the later overwriting the winner.

**Fix:** `RenameNoReplace` publishes via `os.Link` (atomic create-or-fail) and maps `fs.ErrExist` to `ErrTableAlreadyExists` (create) / `table.ErrCommitFailed` (commit). The atomic gate is the `os.Link` step itself, so conflict detection does not depend on the link+remove pair being atomic.

**Verified locally:** `errors.Is(os.Link-EEXIST, fs.ErrExist)` is true (Go maps `syscall.EEXIST` to `ErrExist`, and `*os.LinkError` implements `Unwrap`), so the conflict mapping is sound. Concurrency tests pass and deterministically assert exactly 1 success + 1 conflict; `./io` and `./catalog/hadoop` tests pass; golangci-lint clean. Adding `RenameNoReplaceIO` to `HadoopCatalogFS` is safe since the catalog only supports `LocalFS` (compile-time assertion).

Two non-blocking nits inline.

laskoviymishka (approved):
:shipit:

PR #1301, opened 2026-06-24
fix(table): write added manifests with file spec

fallintoplace:
Summary

- group newly added manifest entries by the added file's partition spec ID
- write each added-file manifest with that matching partition spec
- use the added file's own spec when collecting snapshot summary partition metrics
- add a regression with added position deletes from old and current specs after spec evolution

Why

Position delete files can inherit their partition data and spec ID from the data file they reference. After partition spec evolution, a row-level delete can add a delete file whose spec ID is not the table's current spec ID.

The added-file manifest path wrote every new data/delete file using the current spec. That could put an added delete file into a manifest whose `partition_spec_id` did not match the delete file's own spec. Deleted entries already group by spec ID before writing manifests; this applies the same rule to added files.

Testing

- go test ./table -run '^TestAddedDeleteManifestsUseDeleteFileSpecID$' -count=1
- go test ./table -count=1
- git diff --check

### DIFF
--- a/table/snapshot_producers.go
+++ b/table/snapshot_producers.go
@@ -754,46 +754,46 @@ func (sp *snapshotProducer) manifests(ctx context.Context) (_ []iceberg.Manifest
 
 func (sp *snapshotProducer) manifestProducer(content iceberg.ManifestContent, files []iceberg.DataFile, output *[]iceberg.ManifestFile) func() (err error) {
 	return func() (err error) {
-		out, path, err := sp.newManifestOutput()
-		if err != nil {
-			return err
-		}
-		defer internal.CheckedClose(out, &err)
-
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approve. Correctly fixes added delete/data files being written into a manifest stamped with the table's current spec instead of the file's own spec after partition-spec evolution. Grouping added files by `df.SpecID()` mirrors the existing deleted-entry grouping, and `summary()` now uses each file's own spec.

Verified: race-free (each producer writes a distinct output pointer; filenames via atomic counter), `sp.spec()` falls back safely on unknown IDs, regression test is load-bearing (old-spec delete would land in a current-spec manifest and fail the assertion). Full ./table suite passes; golangci-lint clean.

PR #1302, opened 2026-06-24
fix(io): keep MemFS WalkDir within root

fallintoplace:
Summary

- make MemFS.WalkDir match roots on path boundaries instead of raw string prefixes
- normalize trailing slashes on the walk root
- add regression coverage for sibling prefixes like `table2` and `table_backup`

Why

MemFS stored full keys and filtered them with `strings.HasPrefix(key, root)`. Walking `mem://bucket/table` could therefore include `mem://bucket/table2/...` or `mem://bucket/table_backup/...`, which is surprising for tests and in-memory cleanup flows.

Testing

- go test ./io -run 'TestMemIO_WalkDir' -count=1
- go test ./io -count=1
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m ./io
- git diff --check

### DIFF
--- a/io/mem.go
+++ b/io/mem.go
@@ -104,8 +104,9 @@ func (m *MemFS) WalkDir(root string, fn fs.WalkDirFunc) error {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 
+	root = strings.TrimRight(root, "/")
 	for key, data := range m.files {
-		if !strings.HasPrefix(key, root) {
+		if !memPathInRoot(key, root) {
 			continue
 		}
 		info := &memFileInfo{name: filepath.Base(key), size: int64(len(data))}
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approve. Fixes `MemFS.WalkDir` matching raw string prefixes, so walking `.../table` no longer leaks `table2` / `table_backup` siblings. `memPathInRoot` matches on path boundaries (empty root, exact, or `root+"/"`) with trailing-slash normalization, preserving the prior empty-root and exact-key behavior. WalkDir is the only MemFS method that filters by prefix, so the fix is complete. Test is load-bearing and covers both `table` and `table/` roots. io tests pass; golangci-lint clean.

PR #1303, opened 2026-06-24
fix(table): reject empty equality delete keys

fallintoplace:
Summary

- reject equality delete files with no equality field IDs while loading scan deletes
- add reader-level regression coverage for malformed delete metadata before any delete file is read

Why

Equality delete keys are built from the file's equality field IDs. If that list is empty, the reader can build an empty key, and every data row also encodes to that same empty key. A malformed equality delete file could therefore filter every row from an applicable scan.

Writers already reject empty equality field IDs; this adds the same guard to the reader path for externally-written or malformed delete files.

Testing

- go test ./table -run '^TestReadAllEqualityDeleteFilesRejectsEmptyEqualityFieldIDs$' -count=1
- go test ./table -count=1
- git diff --check

### DIFF
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -65,6 +65,10 @@ func readAllEqualityDeleteFiles(ctx context.Context, fs iceio.IO, schema *iceber
 				continue
 			}
 
+			if len(d.EqualityFieldIDs()) == 0 {
+				return nil, fmt.Errorf("%w: equality delete file %s", ErrEmptyEqualityFieldIDs, d.FilePath())
+			}
+
 			hasAny = true
 			if _, ok := uniqueDeletes[d.FilePath()]; !ok {
 				uniqueDeletes[d.FilePath()] = deleteFileInfo{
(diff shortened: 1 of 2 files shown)


table/equality_delete_reader.go
```diff
@@ -65,6 +65,10 @@ func readAllEqualityDeleteFiles(ctx context.Context, fs iceio.IO, schema *iceber
 				continue
 			}
 
+			if len(d.EqualityFieldIDs()) == 0 {
+				return nil, fmt.Errorf("equality delete file %s has no equality field IDs", d.FilePath())
```
zeroshade: Non-blocking: the writer side exposes `table.ErrEmptyEqualityFieldIDs` (equality_delete_writer.go) and its tests assert it via `errors.Is`. Consider wrapping that sentinel here so callers can match programmatically: `fmt.Errorf("%w: equality delete file %s", ErrEmptyEqualityFieldIDs, d.FilePath())` (or `iceberg.ErrInvalidSchema`, which the writer uses for its related field-ID validation). The reader's sibling error just below ("field ID not found") is also a plain error, so this is consistency polish rather than a defect.

zeroshade (approved):
Approve. Solid defense-in-depth fix. An equality delete file with no equality field IDs builds an empty key, and every data row also encodes to that empty key, so a malformed file would filter out every row of an applicable scan. The guard rejects it at load time, before any delete bytes are read, and mirrors the existing writer-side check. Placement is the correct single chokepoint (`readAllEqualityDeleteFiles`); the regression test is load-bearing; `./table` tests pass and golangci-lint is clean.

One non-blocking consistency nit inline.

PR #1304, opened 2026-06-24
fix(catalog/rest): close SigV4 signing body clone

fallintoplace:
Summary

- close the request body clone used to hash SigV4 payloads
- return close errors from the signing path instead of sending a request after cleanup fails
- add regression coverage for both successful close and close-error paths

Why

The REST session transport uses request.GetBody() to hash signed request payloads without consuming the original request body. That cloned reader is owned by the signing path, but it was left open after io.Copy finished. Under repeated signed POST or commit requests, that can leak any resources held by custom body clones.

Testing

- go test ./catalog/rest -run 'TestSigv4(ClosesClonedRequestBody|ReturnsClonedRequestBodyCloseError|ContentSha256Header)' -count=1
- go test ./catalog/rest -count=1
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m ./catalog/rest/...
- git diff --check

### DIFF
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -250,6 +250,14 @@ func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 
 			h := s.newHash()
 			if _, err = io.Copy(h, rdr); err != nil {
+				if closeErr := rdr.Close(); closeErr != nil {
+					err = errors.Join(err, closeErr)
+				}
+
+				return nil, err
+			}
+
+			if err = rdr.Close(); err != nil {
 				return nil, err
 			}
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approve. Fixes a resource leak in the SigV4 signing path: the `r.GetBody()` clone used to hash the payload was never closed after `io.Copy`, leaking resources held by custom body clones across repeated signed requests. The fix closes the clone, joining the close error on the copy-error path and returning it on success.

Verified: `rdr` is the GetBody() clone (not `r.Body`), so closing it is safe and does not affect the request that gets signed and sent; no use-after-close or double-close; tests cover both close-success and close-error paths (request not sent on close error); golangci-lint clean.

Non-blocking: on the success path the payload hash is already computed before Close, so a close error on the throwaway hash-clone now aborts an otherwise-valid request. Standard bodies use io.NopCloser (never errors), so this only affects custom resource-backed GetBody clones, where failing fast is defensible. Surfacing it is the conservative choice; fine to keep.

PR #1305, opened 2026-06-24
fix(table): return error when rolling writer is closed

fallintoplace:
Summary

- return ErrWriterClosed when RollingDataWriter.Add observes a closed error channel
- remove stopped rolling writers from the partition writer cache
- add regression coverage for the closed-channel Add path and stale writer cache cleanup

Why

RollingDataWriter.Add was reading errorCh with a single-value receive. Once the stream goroutine closed that channel, Add could receive the zero error value, release the caller's record, and return nil. That makes a stopped writer look like it accepted a batch.

The stream now also removes itself from the factory cache when it exits, so later partition writes do not keep reusing a dead writer.

Testing

- go test ./table -run 'TestRollingDataWriter/Test(AddReturnsErrorAfterStreamStopsCleanly|StreamErrorRemovesWriterFromFactory|StreamErrorPathUsesAbort)' -count=1
- go test ./table -count=1
- go test ./table/... -count=1
- git diff --check

### DIFF
--- a/table/rolling_data_writer.go
+++ b/table/rolling_data_writer.go
@@ -19,6 +19,7 @@ package table
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"iter"
 	"maps"
@@ -34,6 +35,10 @@ import (
 	tblutils "github.com/apache/iceberg-go/table/internal"
 	"github.com/google/uuid"
 )
+
+// ErrWriterClosed is returned when records are added to a writer after its
+// streaming goroutine has already stopped.
+var ErrWriterClosed = errors.New("writer is closed")
 
 // writerFactory manages the creation and lifecycle of RollingDataWriter instances
 // for different partitions, providing shared configuration and coordination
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approve. Fixes a silent data-loss bug: `Add` used a single-value receive on `errorCh`, so once `stream` exited and closed the channel, `Add` received the zero (nil) error, released the record, and returned nil — a stopped writer falsely reporting success. The two-value receive now returns `ErrWriterClosed` on a closed channel, and `stream` evicts the dead writer from the factory cache via CompareAndDelete.

Verified (incl. -race): errorCh is buffered cap 1 with a non-blocking sendError, so a real error is received (ok=true) before the closed-channel ok=false path — the distinction is correct and there's no goroutine leak. CompareAndDelete (not Delete) correctly avoids clobbering a newer writer under the same key. Defer order is abort -> evict -> close errorCh -> wg.Done. The two fixes compose: a briefly-stale cached writer now fails loudly instead of dropping records. Tests are load-bearing and pass under the race detector; golangci-lint clean.

PR #1306, opened 2026-06-24
fix(manifest): commit writer state after encode

fallintoplace:
Summary

- encode manifest entries before updating ManifestWriter counters and partition summaries
- encode built-in data files through an Avro-field clone so partition conversion does not mutate the caller's DataFile
- add regression coverage for an encode failure leaving manifest writer state unchanged

Why

ManifestWriter.addEntry updated file counts, row counts, partition summaries, and data file partition data before prepareEntry and writer.Encode could fail. If Encode rejected an entry, the writer looked as if the entry had been written even though it was not in the manifest.

This keeps those mutations on the success side of Encode.

Testing

- go test . -run '^TestManifests/TestManifestWriterDoesNotCommitStateOnEncodeError$' -count=1
- go test . -count=1
- go test ./table -count=1
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m . ./table/...
- git diff --check

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1332,54 +1332,52 @@ func (w *ManifestWriter) addEntry(entry *manifestEntry) error {
 		return errors.New("cannot add entry to closed manifest writer")
 	}
 
-	switch entry.Status() {
+	status := entry.Status()
+	switch status {
+	case EntryStatusADDED, EntryStatusEXISTING, EntryStatusDELETED:
+	default:
+		return fmt.Errorf("unknown entry status: %v", status)
+	}
+	count := entry.DataFile().Count()
+	partition := entry.Data.Partition()
+
+	entryToEncode := *entry
+	if dataFile, ok := entry.DataFile().(*dataFile); ok {
+		encodeDataFile := cloneDataFileAvroFields(dataFile)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approve. Fixes ManifestWriter.addEntry committing writer state (file/row counters, partition summaries, minSeqNum) and mutating the caller's DataFile before prepareEntry/Encode could fail — so a rejected entry left the writer looking as if it had been written. State mutations now happen only after a successful Encode, and the encoded form is built on a clone so the caller's DataFile is untouched.

Verified the refactor is encoding-equivalent: cloneDataFileAvroFields copies only avro-tagged fields into a fresh dataFile; avroEncodePartitionData is the same value-conversion + name-keying as the removed inline logic; prepareEntry never reads the field-name/logical-type maps, so dropping the old setFieldNameToIDMap call in addEntry doesn't change encoding (that setter is still used on the decode path). Regression test is load-bearing (forces an Encode failure and asserts zero counters, empty partitions, minSeqNum -1, and unchanged caller PartitionData). Root + ./table lint clean.

PR #1307, opened 2026-06-24
refactor(table): remove orphan cleanup walk fallback

### LINKED ISSUE #1276: refactor: remove the dead reflection fallback in orphan-cleanup walkDirectory

Herrtian:
﻿## Summary
- Remove the unreachable LocalFS/blob fallback from orphan cleanup `walkDirectory`
- Return an explicit error if a future IO implementation does not implement `ListableIO`
- Add a regression test for that defensive error

Fixes #1276.

## Tests
- `go test ./table -run TestWalkDirectoryRequiresListableIO -count=1`

I also tried the broader orphan-cleanup subset on Windows; it currently fails in existing path-normalization tests that expect S3 URL paths with `/` while Windows produces escaped backslashes (`%5C`).

### DIFF
--- a/table/orphan_cleanup.go
+++ b/table/orphan_cleanup.go
@@ -26,7 +26,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"reflect"
 	"runtime"
 	"slices"
 	"strings"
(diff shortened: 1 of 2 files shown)


Herrtian:
Closing this draft.

PR #1308, opened 2026-06-24
fix: reject invalid parameterized transforms

fallintoplace:
Summary

- parse bucket and truncate transforms with exact syntax checks
- reject bogus transform names like bucketx[5] and truncatefoo[10]
- add negative parser coverage for invalid bucket/truncate prefixes

Why

ParseTransform used prefix checks for bucket and truncate before applying a generic bracket parser. That let strings with extra transform-name text parse as valid parameterized transforms.

The parser now only accepts the Iceberg transform forms bucket[N] and truncate[N].

Testing

- go test . -run '^TestParseTransform$' -count=1
- go test . -count=1
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m .
- git diff --check

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approve. Fixes ParseTransform accepting bogus parameterized transforms: the old code checked only a 'bucket'/'truncate' prefix and then applied a generic `^\w+\[(\d+)\]$` regex, so `bucketx[5]`, `bucket_extra[5]`, `truncatefoo[10]` parsed as valid. Dedicated anchored regexes (`^bucket\[(\d+)\]$`, `^truncate\[(\d+)\]$`) now require the exact spec forms.

Verified: ToLower and the n>0 && n<=MaxInt32 bounds are preserved; invalid-but-matching inputs (bucket[0], int32/atoi overflow) still fall through to ErrInvalidTransform; the shared regexFromBrackets remains used by type-name parsing in types.go (no dead code). New negative tests are load-bearing. Tests pass; lint clean.

PR #1311, opened 2026-06-24
fix(table): track actual metadata removals

fallintoplace:
Summary

- record only partition spec and schema IDs that were actually removed
- avoid emitting remove updates when no requested IDs match existing metadata
- make partition spec removal capacity safe when callers pass more IDs than existing specs

Why

RemovePartitionSpecs and RemoveSchemas preallocated the removed ID slice with length instead of capacity. That made any non-empty request look like a successful removal, and the emitted update used the requested IDs rather than the IDs that were actually removed.

Testing

- go test ./table -run 'Test(AddRemovePartitionSpec|RemovePartitionSpecsNoMatchDoesNotUpdate|RemoveSchemas|RemoveSchemasNoMatchDoesNotUpdate)$' -count=1
- go test ./table -count=1
- git diff --check

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1194,8 +1194,8 @@ func (b *MetadataBuilder) RemovePartitionSpecs(ints []int) error {
 		return fmt.Errorf("%w: can't remove default partition spec with id %d", iceberg.ErrInvalidArgument, b.defaultSpecID)
 	}
 
-	newSpecs := make([]iceberg.PartitionSpec, 0, len(b.specs)-len(ints))
-	removed := make([]int, len(ints))
+	newSpecs := make([]iceberg.PartitionSpec, 0, len(b.specs))
+	removed := make([]int, 0, len(ints))
 	for _, spec := range b.specs {
 		if slices.Contains(ints, spec.ID()) {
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM,  one tiny optional thing, not blocking: RemoveSchemas early-returns on empty ints but RemovePartitionSpecs doesn't — harmless now since empty input just yields no update, but might be worth matching them for symmetry while you're here.

zeroshade (approved):
LGTM. +1 to @tanmayrauth's suggestion: let's add the matching `if len(ints) == 0 { return nil }` early return to `RemovePartitionSpecs` so it stays symmetric with `RemoveSchemas`. It's harmless today (empty input just yields no update), but worth clearing up the asymmetry while we're here. The core fix (length -> capacity on the `removed` slice, and emitting `removed` instead of `ints`) is correct, and the new no-match tests covering the spurious-update and negative-capacity cases are a nice touch.

fallintoplace:
@tanmayrauth @zeroshade Thank you for your support. Let me resolve the nit.

laskoviymishka (approved):
:shipit:

PR #1312, opened 2026-06-24
fix(types): validate fixed and decimal type bounds

fallintoplace:
## Summary

Validate fixed and decimal type bounds at construction time and while parsing JSON type strings.

- Fixed lengths must be at least 1.
- Decimal precision must be 1..38 and scale must be within 0..precision.
- Integer conversion errors, including overflow, are returned instead of being ignored.
- Fixed and decimal type strings are anchored so suffixed names are rejected.
- Literal type paths avoid panics for empty fixed literals and out-of-range decimal scales.

JSON parsing intentionally applies the same validation, so malformed persisted schemas fail early instead of reaching Arrow, Parquet, or reader paths with impossible types.

## Testing

- `go test . -run "Test(TypeUnmarshalRejectsInvalidFixedAndDecimal|FixedType|DecimalType)$" -count=1`
- `go test ./table -run "^TestArrowToIceberg$" -count=1`
- `go test . -count=1`
- `go test ./table -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m . ./table/...`
- `git diff --check`


literals.go
```diff
@@ -1246,12 +1246,16 @@ func (DecimalLiteral) Comparator() Comparator[Decimal] {
...
-func (d DecimalLiteral) Type() Type     { return DecimalTypeOf(9, d.Scale) }
+// Type returns a DecimalType built from the literal's scale and a placeholder
+// precision that is at least 9. The precision is NOT the originating column's
+// declared precision; DecimalLiteral does not carry precision. Callers that
+// need the real column precision must consult the bound field's type rather
+// than lit.Type(). See https://github.com/apache/iceberg-go/issues/1028.
+func (d DecimalLiteral) Type() Type {
+	precision := max(9, d.Scale)
+
+	return DecimalTypeOf(precision, d.Scale)
```
tanmayrauth: Good catch using `max(9, d.Scale)` here — without it the new precision/scale validation would panic for literals with scale > 9.
  
  The same reasoning seems to apply to `FixedLiteral.Type()` at line 1122, which still does `FixedTypeOf(len(f))`. A zero-length fixed literal (e.g. the zero value, or `UnmarshalBinary` with empty data) now panics there instead of returning `FixedType{0}` as it used to. It's an unusual input so maybe not worth much, but since you guarded the decimal side, was leaving the fixed side intentional? A `len(f) == 0` guard (or just documenting it) would keep the two paths consistent.

tanmayrauth (commented):
Nice tightening, anchoring the regexes is a real fix, and the validation bounds look right. One question about symmetry with the fixed-literal path.

literals.go
fallintoplace: Thank you for the comment. I have pushed the fix.

types.go
```diff
@@ -183,12 +185,16 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
 		default:
 			switch {
 			case strings.HasPrefix(typename, "fixed"):
-				matches := regexFromBrackets.FindStringSubmatch(typename)
+				matches := fixedRegex.FindStringSubmatch(typename)
 				if len(matches) != 2 {
 					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
 				}
 
 				n, _ := strconv.Atoi(matches[1])
```
zeroshade: [Major] The `strconv.Atoi` error is dropped here, and `validateFixedLength` only rejects `n <= 0`. On overflow `Atoi` returns `math.MaxInt` together with an `ErrRange`, so `fixed[99999999999999999999]` is accepted as `fixed[math.MaxInt]` instead of being rejected - a validation bypass in the exact path this PR hardens. (The decimal branch below is incidentally safe because its `> 38` upper bound catches the saturated value; fixed has no upper bound.) Suggest checking the error:

```go
n, err := strconv.Atoi(matches[1])
if err != nil {
    return fmt.Errorf("%w: %s: %w", ErrInvalidTypeString, typename, err)
}
```

literals.go
```diff
@@ -1246,12 +1246,16 @@ func (DecimalLiteral) Comparator() Comparator[Decimal] {
...
-// real column precision must consult the bound field's type rather than
-// lit.Type(). See https://github.com/apache/iceberg-go/issues/1028.
-func (d DecimalLiteral) Type() Type     { return DecimalTypeOf(9, d.Scale) }
+// Type returns a DecimalType built from the literal's scale and a placeholder
+// precision that is at least 9. The precision is NOT the originating column's
+// declared precision; DecimalLiteral does not carry precision. Callers that
+// need the real column precision must consult the bound field's type rather
+// than lit.Type(). See https://github.com/apache/iceberg-go/issues/1028.
+func (d DecimalLiteral) Type() Type {
+	precision := max(9, d.Scale)
```
zeroshade: [Minor] `max(9, d.Scale)` guards the lower precision bound but not the upper one: if `d.Scale > 38`, `precision` becomes `> 38` and the now-validating `DecimalTypeOf` panics - whereas the old code returned a placeholder without panicking. `FixedLiteral.Type()` is fully guarded by `max(1, len(f))`, so this is an asymmetry. If `Scale` can exceed 38, consider clamping (e.g. `precision := min(maxDecimalPrecision, max(9, d.Scale))`) or confirm scale is always <= 38.

literals.go
```diff
@@ -1119,7 +1119,7 @@ func (b *BinaryLiteral) UnmarshalBinary(data []byte) error {
 type FixedLiteral []byte
 
 func (FixedLiteral) Comparator() Comparator[[]byte] { return bytes.Compare }
-func (f FixedLiteral) Type() Type                   { return FixedTypeOf(len(f)) }
+func (f FixedLiteral) Type() Type                   { return FixedTypeOf(max(1, len(f))) }
```
zeroshade: nit: unlike `DecimalLiteral.Type()` just below (which documents its placeholder precision), this `max(1, len(f))` has no comment explaining why an empty `FixedLiteral` reports `fixed[1]` (to avoid the new `FixedTypeOf` panic on length 0). A one-line comment would help the next reader.

zeroshade (changes requested):
Nice fix - the bounds match the Iceberg spec (decimal precision 1-38, scale 0..precision; fixed length >= 1), and validating both the exported constructors and the JSON parse path (plus anchoring the `fixed[...]`/`decimal(...)` regexes) is the right call. Requesting changes on one validation gap and one new panic edge; the third is just a doc nit. Details inline.

types.go
```diff
@@ -200,16 +205,10 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
...
-				if err != nil {
-					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
-				}
-				scale, err := strconv.Atoi(matches[2])
-				if err != nil {
-					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
-				}
-				if err := validateDecimalPrecisionScale(prec, scale); err != nil {
-					return fmt.Errorf("%w: %w", ErrInvalidTypeString, err)
+				prec, _ := strconv.Atoi(matches[1])
```
laskoviymishka, line 208: Same one zeroshade flagged — agreed it needs fixing. Dropping the `Atoi` errors here is only safe by accident: on `decimal(99999999999999999999, 0)` the parse saturates to `math.MaxInt`, and the `prec > 38` bound happens to catch it, so the message comes out as "precision must be between 1 and 38: 9223372036854775807" instead of "value out of range".

The tell is that the `fixed` path two blocks up (line 196) *does* propagate the `Atoi` error — this path used to as well. If `maxDecimalPrecision` ever moves, the accidental safety disappears and it becomes a real bypass.

I'd restore both err checks to match the fixed path, and add a `decimal(99999999999999999999, 0)` case asserting "value out of range" — the fixed overflow case is already tested, decimal isn't.

types.go
```diff
@@ -571,17 +570,17 @@ func (m *MapType) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-func validateFixedLength(length int) error {
-	if length < 0 {
-		return fmt.Errorf("fixed length must be non-negative, got %d", length)
+func validateFixedLength(n int) error {
+	if n <= 0 {
```
laskoviymishka, line 574: This now runs on the JSON parse path, and that's where I get nervous. The spec says `fixed(L)` with no stated lower bound on L, and Java's `Types.FixedType.ofLength` does no validation at all — so a metadata JSON containing `fixed[0]`, written by Java or older iceberg-go, used to unmarshal fine and now hard-errors.

Enforcing `>= 1` inside `FixedTypeOf` is fine — that's our constructor, our contract. But rejecting it when *reading* someone else's table is a compatibility decision. I'd keep the parse path permissive (`n < 0`) and enforce the stricter bound only in the constructor. If the strict read is intentional, could we say so in the PR description? wdyt?

types.go
```diff
@@ -200,16 +205,10 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
...
-				}
-				scale, err := strconv.Atoi(matches[2])
-				if err != nil {
-					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
-				}
-				if err := validateDecimalPrecisionScale(prec, scale); err != nil {
-					return fmt.Errorf("%w: %w", ErrInvalidTypeString, err)
+				prec, _ := strconv.Atoi(matches[1])
+				scale, _ := strconv.Atoi(matches[2])
+				if err := validateDecimalParams(prec, scale); err != nil {
```
laskoviymishka, line 210: Same class as the `fixed[0]` note above, decimal side. `validateDecimalParams` on the parse path now rejects `decimal(0, 0)` and any `scale > precision`, but Java's `DecimalType.of` only checks `precision <= 38` — it happily constructs `decimal(2, 5)`. So a Java-written table with `scale > precision` stops unmarshalling here.

If we go permissive-on-read, this and the fixed case move together: strict in the constructors, lenient in `UnmarshalJSON`.

literals.go
```diff
@@ -1180,10 +1180,13 @@ func (g GeoLiteral) MarshalBinary() ([]byte, error) {
...
 
 func (FixedLiteral) Comparator() Comparator[[]byte] { return bytes.Compare }
-func (f FixedLiteral) Type() Type                   { return FixedType{len: len(f)} }
-func (f FixedLiteral) Value() []byte                { return []byte(f) }
-func (f FixedLiteral) Any() any                     { return f.Value() }
-func (f FixedLiteral) String() string               { return string(f) }
+
+// Empty fixed literals can arise from zero values or empty unmarshaled data.
+// Report fixed[1] as a placeholder so Type() stays non-panicking.
+func (f FixedLiteral) Type() Type     { return FixedTypeOf(max(1, len(f))) }
```
laskoviymishka, line 1186: Trading the panic for a placeholder makes sense, but `fixed[1]` is a claim about the data that isn't true — a zero-length `FixedLiteral` reports length 1 while the underlying bytes are empty. Anything using `lit.Type()` for a length decision (Parquet schema, equality, scan planning) gets the wrong answer, and `lit.To(FixedType{len: 1})` fails since `0 != 1`.

Since `Type()` is meant to report what the literal *is*, not construct a type for registration, I'd return `FixedType{len: len(f)}` via the struct literal directly and skip the `FixedTypeOf` floor. If the floor is deliberate to dodge the constructor panic, the comment should say the length is a non-representative placeholder.

literals.go
```diff
@@ -1307,13 +1310,18 @@ func (DecimalLiteral) Comparator() Comparator[Decimal] {
...
+// Type returns a DecimalType built from the literal's scale and a placeholder
+// precision that is at least 9. The precision is NOT the originating column's
+// declared precision; DecimalLiteral does not carry precision. Callers that
+// need the real column precision must consult the bound field's type rather
+// than lit.Type(). See https://github.com/apache/iceberg-go/issues/1028.
 func (d DecimalLiteral) Type() Type {
-	return DecimalType{precision: 9, scale: d.Scale}
+	// Clamp the placeholder scale into Iceberg's valid schema range so odd
+	// literal values still report a non-panicking placeholder type.
+	scale := min(maxDecimalPrecision, max(0, d.Scale))
```
laskoviymishka, line 1321: Same shape as the fixed case. Clamping `scale` to dodge the `DecimalTypeOf` panic means a literal with `Scale: 39` reports `decimal(38, 38)` — a caller doing `lit.Type().(DecimalType).Scale()` expecting 39 silently gets 38. The old code was wrong about precision but at least honest about scale.

Not a blocker, but I'd either return the struct literal directly (`DecimalType{precision: max(9, d.Scale), scale: d.Scale}`) or document that `Type()` is a clamped, non-representative placeholder. wdyt?

table/substrait/substrait.go
```diff
@@ -263,9 +263,9 @@ func toByteSliceSubstraitLiteral[T []byte | types.UUID](v T) expr.Literal {
 // toDecimalLiteral converts an iceberg.DecimalLiteral into a Substrait decimal
 // literal. Precision is taken from the bound field's iceberg.DecimalType (typ),
 // NOT from v.Type(). DecimalLiteral does not carry the originating column's
-// precision, so v.Type() always reports a hardcoded precision of 9 (see
+// precision, so v.Type() reports a placeholder precision (see
 // https://github.com/apache/iceberg-go/issues/1028). If typ is not a
-// DecimalType, this falls back to v.Type()'s (hardcoded) precision.
+// DecimalType, this falls back to v.Type()'s placeholder precision.
```
laskoviymishka, line 268: This comment is now factually off — the fallback at line 270 is `precision := int32(9)`, which never calls `v.Type()`. The old "hardcoded precision of 9" wording was accurate. I'd make it something like "falls back to a hardcoded placeholder precision of 9 (independent of `v.Type()`)".

types_test.go
```diff
@@ -73,14 +73,74 @@ func TestTypesBasic(t *testing.T) {
 	}
 }
 
+func TestTypeUnmarshalRejectsInvalidFixedAndDecimal(t *testing.T) {
```
laskoviymishka, line 76: Small cleanups while we're here: this table overlaps `TestDecimalTypeInvalidParse` (both cover zero precision, precision too large, negative scale) — `scale > precision` is the only genuinely new coverage, so I'd fold it into the existing test rather than duplicate. Also, unlike the existing test, this one has no `name` field, so `t.Run` uses the raw type string ("decimal(2, 5)"), whose parens/commas need escaping for `go test -run`. A `name` field would fix that.

laskoviymishka (changes requested):
Agreed with where tanmayrauth and zeroshade already landed — the bounds match the spec, anchoring the `fixed[...]`/`decimal(...)` regexes is a real fix, and validating both the exported constructors and the JSON parse path is the right instinct. I went through this independently and I hit zeroshade's two blockers too, so the inline on the swallowed `strconv.Atoi` error is me corroborating, not raising anything new.

I'd hold this before merging though, mostly on one question I don't think has come up yet: is tightening validation on the JSON *parse* path an intentional read-compatibility break? `validateFixedLength` now rejects `fixed[0]` and `validateDecimalParams` rejects `decimal(0,0)` and any `scale > precision`, all at unmarshal time. Java's `FixedType.ofLength` has no length check, and `DecimalType.of` only guards `precision <= 38` — so a table written by Java or older iceberg-go with any of those shapes now fails to load in iceberg-go instead of reading. Enforcing these in the `FixedTypeOf`/`DecimalTypeOf` constructors is clearly right; doing it on the read path is a different call. I'd lean toward keeping the parse path permissive and enforcing only in the constructors, but if the break is deliberate let's call it out in the description.

Separately, the two `Type()` placeholder changes (`FixedLiteral` → `fixed[1]`, `DecimalLiteral` clamping scale) swap a panic for a value that's quietly wrong about the data — worth a look but not blockers on their own.

- restore the `strconv.Atoi` error checks in the decimal parse path to match the fixed path, and add the `decimal(99999999999999999999, 0)` overflow test (the fixed overflow case is already covered)
- settle the parse-vs-constructor strictness question for `fixed[0]` / `decimal(0,0)` / `scale > precision`
- fix the `substrait.go` comment — it says the fallback uses `v.Type()`'s placeholder precision, but line 270 hardcodes `int32(9)` and never calls `v.Type()`

The `Type()` placeholder honesty and the test-table consolidation are things I'd like but can live with as follow-ups. Once the parse-path question and the `Atoi` fix land, happy to take another pass and approve.

types.go
zeroshade, line 574: Non-blocking: tightening this to reject `n <= 0` means `fixed[0]` / `FixedTypeOf(0)` now error/panic where they were previously accepted. That's correct per the Iceberg spec (fixed length must be ≥ 1), but it's technically a behavior change for any existing caller or persisted data using `fixed[0]` — worth calling out in the PR description / changelog.

types.go
zeroshade, line 208: Minor: the decimal branch now discards the `strconv.Atoi` errors (`prec, _ :=` / `scale, _ :=`), relying on `validateDecimalParams` to reject the resulting 0 on overflow. The `fixed` branch just above still checks its `Atoi` error and surfaces `value out of range`. Consider handling the error here too so overflow input like `decimal(99999999999999999999, 0)` gets a consistent message.

zeroshade (commented):
A couple of non-blocking nits on the validation changes (inline).

PR #1313, opened 2026-06-24
fix(table): validate sort field source IDs

fallintoplace:
Summary

- reject sort fields that omit both source-id and source-ids
- reject zero or negative sort source IDs for both JSON metadata and NewSortOrder
- add regression coverage for missing, empty, zero, negative, and mixed invalid source IDs

Why

SortField.UnmarshalJSON treated a missing source-id as the Go zero value and accepted it as source ID 0. That allowed invalid sort metadata to enter the table model and fail later during compatibility checks or other sort-order use.

Testing

- go test ./table -run 'Test(NewSortOrderRejectsInvalidSourceIDs|SortFieldMultiArgSourceIDs|UnmarshalSortOrderDefaults|UnmarshalInvalidSort)' -count=1
- go test ./table -count=1
- git diff --check

### DIFF
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -330,21 +330,28 @@ func (s *SortOrder) CheckCompatibility(schema *iceberg.Schema) error {
 	}
 
 	for _, field := range s.fields {
-		f, ok := schema.FindFieldByID(field.SourceID())
-		if !ok {
-			return fmt.Errorf("sort field with source id %d not found in schema", field.SourceID())
-		}
-
-		if _, ok := f.Type.(iceberg.PrimitiveType); !ok {
-			return fmt.Errorf("cannot sort by non-primitive source field: %s", f.Type.Type())
-		}
-
 		if field.Transform == nil {
 			return fmt.Errorf("%w: sort field with source id %d has no transform", ErrInvalidTransform, field.SourceID())
 		}
 
-		if !field.Transform.CanTransform(f.Type) {
-			return fmt.Errorf("invalid source type %s for transform %s", f.Type.Type(), field.Transform)
+		var firstField iceberg.NestedField
+		for idx, sourceID := range field.SourceIDs {
+			f, ok := schema.FindFieldByID(sourceID)
+			if !ok {
+				return fmt.Errorf("sort field with source id %d not found in schema", sourceID)
+			}
+
+			if _, ok := f.Type.(iceberg.PrimitiveType); !ok {
+				return fmt.Errorf("cannot sort by non-primitive source field: %s", f.Type.Type())
+			}
+
+			if idx == 0 {
+				firstField = f
+			}
+		}
+
+		if !field.Transform.CanTransform(firstField.Type) {
+			return fmt.Errorf("invalid source type %s for transform %s", firstField.Type.Type(), field.Transform)
 		}
 	}
 
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -44,6 +44,7 @@ const (
 
 var (
 	ErrInvalidSortOrderID   = errors.New("invalid sort order ID")
+	ErrInvalidSortSourceID  = errors.New("invalid sort source ID")
 	ErrInvalidTransform     = errors.New("invalid transform, must be a valid transform string or a transform object")
 	ErrInvalidSortDirection = errors.New("invalid sort direction, must be 'asc' or 'desc'")
 	ErrInvalidNullOrder     = errors.New("invalid null order, must be 'nulls-first' or 'nulls-last'")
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):

table/sorting.go
```diff
@@ -271,6 +303,9 @@ func NewSortOrder(orderID int, fields []SortField) (SortOrder, error) {
 		fields = []SortField{}
 	}
 	for idx, field := range fields {
+		if err := validateSortSourceIDs(field.SourceIDs); err != nil {
```
zeroshade: This validates that every entry in `SourceIDs` is present and positive - good. The remaining gap is *existence in the schema*: `SortOrder.CheckCompatibility` only looks up `field.SourceID()`, i.e. the first source id (sorting.go:298), so a multi-arg sort field such as `SourceIDs: [1, 999]` still passes when `999` isn't in the schema. Worth iterating all of `field.SourceIDs` there (looking each up via `FindFieldByID` and checking it's primitive), with a regression case pairing a valid id and a nonexistent one.

zeroshade (changes requested):
Thanks - validating source-id shape (presence + positivity) at unmarshal and in `NewSortOrder` is a solid addition, and the table tests are thorough. Requesting changes on one related gap with multi-arg `source-ids` - inline.

table/sorting.go
```diff
@@ -180,6 +190,28 @@ func (s *SortField) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+func validateSortSourceID(id int) error {
+	if id <= 0 {
```
laskoviymishka: I'd drop the `id <= 0` guard from the parse path — it's stricter than the spec. The spec doesn't reserve source-id 0 for sort fields (only order-id 0, for unsorted), and Java's `SortOrderParser`, iceberg-rust, and PyIceberg all accept it with no range check.

Since this fires in `UnmarshalJSON`, a table another client wrote with source-id 0 becomes unreadable here — we'd fail to load the metadata at all. If we want to reject nonsensical *writes*, the place for it is `CheckCompatibility` or `MarshalJSON`, not the reader. wdyt?

table/sorting.go
```diff
@@ -133,10 +134,13 @@ func (s *SortField) UnmarshalJSON(b []byte) error {
...
-		if _, ok := raw["source-ids"]; ok {
-			return errors.New("sort field cannot contain both source-id and source-ids")
-		}
+	_, hasSourceID := raw["source-id"]
+	_, hasSourceIDs := raw["source-ids"]
+	if hasSourceID && hasSourceIDs {
+		return errors.New("sort field cannot contain both source-id and source-ids")
+	}
+	if !hasSourceID && !hasSourceIDs {
+		return fmt.Errorf("%w: sort field must contain source-id or source-ids", ErrInvalidSortSourceID)
```
laskoviymishka: V3 marks both `source-id` and `source-ids` as `optional` in the sort field JSON, so a reader hard-failing when both are absent isn't spec-compliant — the "single-arg writes source-id, multi-arg writes source-ids" rule is a writer obligation, not a reader parse-rejection rule.

Same interop risk as the zero-check: a V3 writer that omits both (a future extension, or a quirk in another client) makes the whole table unreadable here. I'd let `SourceIDs` stay empty through the parse and let `CheckCompatibility` reject it downstream. Keep the `hasSourceID && hasSourceIDs` rejection though — that one's genuinely malformed.

table/sorting.go
laskoviymishka: This changes the error a caller sees for `SortField{SourceIDs: nil, Transform: nil}` from `ErrInvalidTransform` to `ErrInvalidSortSourceID`, since the source-id check now runs first. `NewSortOrder` is public with named sentinels, so anyone doing `errors.Is(err, ErrInvalidTransform)` on programmatic construction quietly breaks — worth documenting the new validation order in the godoc, or reordering so transform stays first.

While here: `fmt.Errorf("%w: sort field at index %d", err, idx)` wraps an already-wrapped `err`, so the message reads "invalid sort source ID: source-ids must not be empty: sort field at index 0" with the index trailing. The transform line right below it puts the sentinel first in a self-contained message — I'd flatten this to match.

laskoviymishka (changes requested):
Nice tight fix, the `ErrInvalidSortSourceID` sentinel, the flat both/neither branching in `UnmarshalJSON`, and the table-driven `NewSortOrder` rejection test all read well.

There's really one thing standing between this and merge: two of the new validations are stricter than the spec and fire in the worst place — `UnmarshalJSON`, where they block loading the table's metadata entirely. The `id <= 0` guard rejects source-id 0, which the spec doesn't reserve for sort fields and which Java, iceberg-rust, and PyIceberg all accept — so a table another client wrote with source-id 0 becomes unreadable here. And the `!hasSourceID && !hasSourceIDs` hard-fail rejects metadata where both keys are absent, but V3 marks both `optional`; that's a writer obligation, not a reader rejection rule.

The fix keeps your fail-fast intent without the interop risk: have `UnmarshalJSON` reject only the genuinely-malformed "both keys present" case, and move the range/presence checks to `CheckCompatibility` (or the write path). That's the whole blocker.

One more to decide before it lands: `NewSortOrder` now returns `ErrInvalidSortSourceID` where it used to return `ErrInvalidTransform` for a nil/nil field, since the source-id check runs first — a behavior change on a public sentinel. Document the new order (or reorder the checks) and I'm happy to approve on the next pass.

laskoviymishka (approved):
👍

zeroshade (approved):

PR #1315, opened 2026-06-24
fix: validate schema field IDs during unmarshal

fallintoplace:
Summary

- validate duplicate field IDs while unmarshalling schemas from JSON
- add coverage for top-level and nested duplicate IDs returning ErrInvalidSchema

Why

Schema constructors already reject duplicate positive field IDs, but Schema.UnmarshalJSON assigned decoded fields without running the same check. Malformed schema JSON could enter the system and later corrupt ID-based lookups, projections, name mapping, or writes.

Testing

- go test . -run '^TestUnmarshalSchemaRejectsDuplicateFieldIDs$' -count=1\n- go test . -count=1\n- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m .\n- git diff --check

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -278,6 +278,10 @@ func (s *Schema) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
+	if err := checkDuplicateFieldIDs(nil, aux.Fields); err != nil {
+		return err
+	}
+
 	s.init()
 
 	s.fields = aux.Fields
--- a/schema_test.go
+++ b/schema_test.go
@@ -545,7 +545,7 @@ func TestUnmarshalSchemaRejectsDuplicateFieldIDs(t *testing.T) {
 	}
 }
 
-func TestUnmarshalSchemaRejectsMissingOrZeroCollectionFieldIDs(t *testing.T) {
+func TestUnmarshalSchemaRejectsMissingCollectionFieldIDs(t *testing.T) {
 	tests := []struct {
 		name     string
 		schema   string
@@ -573,28 +573,6 @@ func TestUnmarshalSchemaRejectsMissingOrZeroCollectionFieldIDs(t *testing.T) {
 			contains: "field is missing required 'element-id' key in JSON",
 		},
 		{
-			name: "zero list element id",
-			schema: `{
-				"type": "struct",
-				"fields": [
-					{
-						"id": 1,
-						"name": "items",
-						"type": {
-							"type": "list",
-							"element-id": 0,
-							"element": "int",
-							"element-required": false
-						},
-						"required": false
-					}
-				],
-				"schema-id": 1,
-				"identifier-field-ids": []
-			}`,
-			contains: "field 'element-id' must not be 0",
-		},
-		{
 			name: "missing map key id",
 			schema: `{
 				"type": "struct",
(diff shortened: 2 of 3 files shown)


schema_test.go
```diff
@@ -442,6 +442,61 @@ func TestUnmarshalSchema(t *testing.T) {
 	assert.True(t, tableSchemaSimple.Equals(&schema))
 }
 
+func TestUnmarshalSchemaRejectsDuplicateFieldIDs(t *testing.T) {
```
tanmayrauth, line 445: Nice — the top-level and nested-struct cases are the important ones. Since checkDuplicateFieldIDs also walks list element IDs and map key/value IDs (schema.go:118-130), would you consider adding a list or map dup-ID case too? Those branches are only exercised indirectly right now, and they're the easiest to silently break later.

tanmayrauth (commented):
LGTM, one nit on test coverage.

schema.go
```diff
@@ -278,6 +278,10 @@ func (s *Schema) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
+	if err := checkDuplicateFieldIDs(nil, aux.Fields); err != nil {
```
zeroshade, line 281: This catches duplicate *positive* IDs, but missing list/map IDs still slip through. `checkDuplicateFieldIDs` only records `f.ID > 0` (schema.go:101), and an omitted list `element-id` or map `key-id`/`value-id` decodes to `0` - so a schema that omits one of those still unmarshals cleanly (a lone `0` isn't a duplicate). If the goal is to validate field IDs on unmarshal, it'd be worth rejecting missing/zero element/key/value IDs too (e.g. decode them as `*int` and error when absent), while still allowing the negative placeholders used during construction. Is the broader validation in scope here, or is this intentionally limited to duplicates?

zeroshade (changes requested):
Thanks for adding this. The duplicate-ID detection is correct, and I like the recursion into struct/list/map plus the negative-placeholder carve-out. Requesting changes on one gap relative to the PR title (`validate schema field IDs`) - noted inline.

types.go
```diff
@@ -444,15 +444,22 @@ func (l *ListType) String() string { return fmt.Sprintf("list<%s>", l.Element) }
...
 		return err
 	}
 
-	l.ElementID = aux.ID
+	if aux.ID == nil {
+		return fmt.Errorf("%w: field is missing required 'element-id' key in JSON", ErrInvalidSchema)
+	}
+	if *aux.ID == 0 {
+		return fmt.Errorf("%w: field 'element-id' must not be 0", ErrInvalidSchema)
+	}
```
laskoviymishka: The nil check here is right — it mirrors Java's `JsonUtil.getInt` existence check, so keep it. The `== 0` rejection is the blocker, though, and I'd drop it.

The spec only requires field IDs to be unique in the table schema — it doesn't reserve 0 or require positivity, and Java, PyIceberg, and iceberg-rust all parse 0 without complaint. So this makes iceberg-go permanently unable to read otherwise-valid metadata any of them wrote with `element-id: 0`. It also breaks our own round-trip: `ListType.ElementID` is a plain `int`, so `json.Marshal(&iceberg.ListType{})` emits `element-id: 0` and this same method then refuses it.

```suggestion
	if aux.ID == nil {
		return fmt.Errorf("%w: field is missing required 'element-id' key in JSON", ErrInvalidSchema)
	}
```

Reject 0 vs. accept it as the spec allows — wdyt?

types.go
```diff
@@ -524,18 +531,31 @@ func (m *MapType) String() string {
...
+	if *aux.KeyID == 0 {
+		return fmt.Errorf("%w: field 'key-id' must not be 0", ErrInvalidSchema)
+	}
+	if aux.ValueID == nil {
+		return fmt.Errorf("%w: field is missing required 'value-id' key in JSON", ErrInvalidSchema)
+	}
+	if *aux.ValueID == 0 {
+		return fmt.Errorf("%w: field 'value-id' must not be 0", ErrInvalidSchema)
+	}
```
laskoviymishka: Same blocker on the map side — keep both nil checks, drop the two `== 0` checks. Java's `mapFromJson` parses zero key-id/value-id fine, so rejecting them breaks reading Java/Py/Rust-written maps the same way. The suggestion also spaces the two guards out, which keeps nlreturn happy.

```suggestion
	if aux.KeyID == nil {
		return fmt.Errorf("%w: field is missing required 'key-id' key in JSON", ErrInvalidSchema)
	}

	if aux.ValueID == nil {
		return fmt.Errorf("%w: field is missing required 'value-id' key in JSON", ErrInvalidSchema)
	}
```

schema_test.go
```diff
@@ -442,6 +442,217 @@ func TestUnmarshalSchema(t *testing.T) {
...
+						"required": false
+					}
+				],
+				"schema-id": 1,
+				"identifier-field-ids": []
+			}`,
+			contains: "field is missing required 'element-id' key in JSON",
+		},
+		{
+			name: "zero list element id",
```
laskoviymishka, line 640: Once the `== 0` checks come out, this row and "zero map value id" assert the wrong outcome — they'd need to flip to `require.NoError`. While here: the table never exercises zero map key-id or missing map value-id (both are real branches), and there's no positive case proving a valid non-zero list/map schema still unmarshals after the `*int` change — `tableSchemaNested` goes through the constructor, not JSON decode, so a regression there wouldn't be caught. I'd add a round-trip case asserting `require.NoError`.

laskoviymishka (changes requested):
Almost there — the missing-key (nil) checks and the `checkDuplicateFieldIDs` reuse across the constructor and the decode path are exactly the right shape. One blocker before this lands.

The `== 0` rejection on element-id/key-id/value-id is stricter than the spec asks for. The spec only requires these IDs to be unique in the table schema — it doesn't reserve 0 or require positivity, and Java, PyIceberg, and iceberg-rust all parse 0 without complaint. So this makes iceberg-go permanently unable to read otherwise-valid metadata those clients wrote, and it breaks our own round-trip: `MarshalJSON(&ListType{})` emits `element-id: 0`, which this same `UnmarshalJSON` now refuses. Worth noting too that these methods fire on every list/map decode — partition fields, sort orders, REST responses — so the effect is a global tightening of list/map decoding, not just schema validation.

Before merge:
- Drop the three `*aux.ID == 0` / `*aux.KeyID == 0` / `*aux.ValueID == 0` checks; keep the nil checks and the duplicate-ID check.
- Add a blank line before each consecutive guard block — nlreturn will fail CI otherwise.
- Flip the two "zero …" test rows to expect success, add the missing zero-map-key-id / missing-map-value-id cases, and add one positive round-trip case asserting a valid list/map schema still unmarshals.

The nil checks and duplicate-ID validation are good as-is. Happy to take another pass once the zero-checks come out.

laskoviymishka (approved):

zeroshade (approved):

PR #1317, opened 2026-06-25
fix(literals): incorrect comparison result for (false, false) in BoolLiteral.Comparator

### LINKED ISSUE #1316: Bug: BoolLiteral Comparator returns incorrect result for equal false values
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

`BoolLiteral.Comparator()` incorrectly returns `-1` when comparing two equal false values. Check [here](https://github.com/apache/iceberg-go/blob/8a8556b0e0011b867116bb661e5c2d8a22e85103/literals.go#L377-L389).

When `v1` is `false`, the comparator returns `-1` unconditionally, ignoring the value of `v2`. As a result, `false == false` is reported as `false < false`.

badalprasadsingh:
## Description 

Fix a **bug** in `BoolLiteral.Comparator()` where comparing two `false` values returned `-1` instead of `0`.

The previous implementation returned `-1` whenever `v1` was `false`, regardless of the value of `v2`, causing the comparator to report: `false < false`.

Fixes #1316 

## Testing

Added a regression test covering all boolean comparison combinations:
- `(false, false)` → `0`
- `(true, true)` → `0`
- `(false, true)` → `-1`
- `(true, false)` → `1`

### DIFF
--- a/literals.go
+++ b/literals.go
@@ -376,11 +376,10 @@ type BoolLiteral bool
 
 func (BoolLiteral) Comparator() Comparator[bool] {
 	return func(v1, v2 bool) int {
+		if v1 == v2 {
+			return 0
+		}
 		if v1 {
-			if v2 {
-				return 0
-			}
-
 			return 1
 		}
 
--- a/literals_test.go
+++ b/literals_test.go
@@ -736,6 +736,27 @@ func TestInvalidBoolLiteralConversions(t *testing.T) {
 	})
 }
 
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
LGTM. The old closure fell through `if v1` unconditionally, so `(false, false)` returned -1; the `if v1 == v2 { return 0 }` form is correct for all four pairs, and since `bool` isn't `constraints.Ordered` the hand-written comparator is the right call. Test covers every combination. Thanks!

PR #1320, opened 2026-06-26
feat(cmd/config): add aws-profile support for Glue catalog

rasta-rocket:
## Problem

When using the Glue catalog, the CLI always resolves AWS credentials through the SDK default chain.
Users working with multiple AWS accounts or roles had to fall back to environment variables (`AWS_PROFILE`) or profile-switching wrappers instead of expressing the profile in their iceberg-go config (1 catalog = 1 account = 1 profile).

## Fix

Add an `aws-profile` field to `CatalogConfig` in `config/config.go`.
When the Glue catalog is initialised in `initCatalog`, the profile is passed to `awsconfig.LoadDefaultConfig` via `WithSharedConfigProfile`.
A matching `--aws-profile` CLI flag is also wired up and takes precedence over the config file value (following the same merge pattern used by the other catalog flags).

## How to test

Add an entry to `~/.iceberg-go.yaml`:

```yaml
catalog:
  default:
    type: glue
    aws-profile: my-aws-profile
```

Run any Glue command and confirm it authenticates with the named profile:
```
iceberg list
```

Alternatively, supply it on the command line:
```
iceberg --catalog glue --aws-profile my-aws-profile list
```

Unit test:
```
go test ./config/...
```

### DIFF
--- a/cmd/iceberg/main.go
+++ b/cmd/iceberg/main.go
@@ -202,6 +202,7 @@ type Args struct {
 	Warehouse   string `arg:"--warehouse" help:"warehouse to use"`
 	Scope       string `arg:"--scope" default:"catalog" help:"OAuth scope"`
 	Config      string `arg:"--config" help:"path to configuration file"`
+	AwsProfile  string `arg:"--aws-profile" help:"AWS profile to use (Glue catalog)"`
 
 	RestOptions *config.RestOptions `arg:"-"`
 }
@@ -379,7 +380,12 @@ func initCatalog(ctx context.Context, args Args) catalog.Catalog {
 			log.Fatal(err)
 		}
 	case catalog.Glue:
-		awscfg, err := awsconfig.LoadDefaultConfig(ctx)
+		var awsLoadOpts []func(*awsconfig.LoadOptions) error
+		if args.AwsProfile != "" {
+			awsLoadOpts = append(awsLoadOpts, awsconfig.WithSharedConfigProfile(args.AwsProfile))
+		}
+
+		awscfg, err := awsconfig.LoadDefaultConfig(ctx, awsLoadOpts...)
 		if err != nil {
 			log.Fatal(err)
 		}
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -47,6 +47,7 @@ type CatalogConfig struct {
 	Output      string       `yaml:"output"`
 	Credential  string       `yaml:"credential"`
 	Warehouse   string       `yaml:"warehouse"`
+	AwsProfile  string       `yaml:"aws-profile,omitempty"`
```
zeroshade: nit: `AwsProfile` carries `omitempty` while the sibling string fields (`Output`, `Credential`, `Warehouse`) don't. Harmless on unmarshal, but worth making consistent one way or the other.

cmd/iceberg/main.go
```diff
@@ -810,6 +816,10 @@ func mergeConf(fileConf *config.CatalogConfig, args *Args, explicitFlags map[str
 		args.Warehouse = fileConf.Warehouse
 	}
 
+	if !explicitFlags["aws-profile"] && len(fileConf.AwsProfile) > 0 {
```
zeroshade, line 819: nit: this `mergeConf` branch and the `--aws-profile` flag itself aren't covered by `args_test.go`. A case asserting (a) `--aws-profile` parses onto `Args`, and (b) the file-config value is applied only when the flag is absent, would lock in the precedence.

zeroshade (approved):
LGTM. The profile plumbs through correctly - `WithSharedConfigProfile` into `LoadDefaultConfig`, then on to the Glue client and S3 ops - and `mergeConf` follows the same explicit-flag-wins precedence as the other fields. A couple of nits inline. One follow-up thought: the programmatic Glue path (`toAwsConfig` via `iceberg.Properties`) has no `glue.profile` key, so this is CLI-only for now - might be worth a follow-up issue for parity.

laskoviymishka (approved):
:shipit:

PR #735, opened 2026-02-17
feat(table): Add support for row lineage in v3

dttung2905:
This should fully support read path and partially support write path
Unsupported write path:
 - Rewrite/compaction: When overwrite or rewrite copies existing rows into new data files, existing non-null `_row_id` and `_last_updated_sequence_number` are not copied into the new files. Row lineage is preserved for appends and for metadata/manifest list; it is not yet preserved when rewriting data files.
 - Explicit null columns on append: New data files do not write `_row_id/_last_updated_sequence_number` as null columns (they are omitted); that is allowed by the [spec](https://iceberg.apache.org/spec/#reserved-field-ids) and is not planned in this PR.

>A data file with only new rows for the table may omit the _last_updated_sequence_number and _row_id. If the columns are missing, readers should treat both columns as if they exist and are set to null for all rows.

PR #1321, opened 2026-06-26
feat: let WriteManifest set and return v3 first_row_id

### LINKED ISSUE #1274: feat: let WriteManifest set and return v3 first_row_id
Follow-up to #735.

`WriteManifest` still carries its pre-#735 signature and returns only `(ManifestFile, error)` — it forwards no writer or file options, so `ToManifestFile` never sets `FirstRowIDValue` and the returned `ManifestFile.FirstRowID()` is always nil. The real v3 path (`NewManifestListWriterV3` plus the inheritance logic) handles `first_row_id`, but it's reached through `WriteManifestList`, bypassing `WriteManifest` entirely. `WriteManifest` has no production callers — it's a public helper for library consumers, who currently can neither set nor observe `first_row_id` through it, and there's no `WriteManifestV3`.

Either add a `WriteManifestV3` helper that takes a `firstRowID` and returns the assigned value alongside the `ManifestFile`, or add a `WithManifestFileFirstRowID` option and forward variadic opts through `WriteManifest`. Keep `WriteManifest` back-compatible.

This one touches public API, so it's worth a quick maintainer nod on the shape before someone runs with it.

the-onewho-knocks:
**Summary**

* add `WithManifestFileFirstRowID`, a `ManifestFileOption` that sets `first_row_id` on a v3+ data manifest
* forward variadic `ManifestFileOption` args through `WriteManifest` to `ToManifestFile`

**Why**

`WriteManifest` predates the v3 work in #735 and forwarded no options to `ToManifestFile`, so `ManifestFile.FirstRowID()` was always `nil` for manifests written through this public helper, even though `ManifestWriter`/`ToManifestFile` already supported setting it via `opts`. Library consumers had no way to set or observe `first_row_id` through `WriteManifest`.

`WriteManifest`'s signature change is purely additive (`opts` is the last, variadic parameter), so existing call sites are unaffected.

**Testing**

* `go test ./... -run '^TestWriteManifestWithFirstRowIDOption$' -count=1`
* `go test ./... -count=1`
* `gofmt -l .`
* `golangci-lint run`

Fixes #1274

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1560,10 +1560,8 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 						if m.nextRowID != nil {
 							firstRowID := *m.nextRowID
 							wrapped.FirstRowIDValue = &firstRowID
+							*m.nextRowID += wrapped.ExistingRowsCount + wrapped.AddedRowsCount
 						}
-					}
-					if m.nextRowID != nil {
-						*m.nextRowID += wrapped.ExistingRowsCount + wrapped.AddedRowsCount
 					}
 				}
 			}
--- a/manifest.go
+++ b/manifest.go
@@ -1274,6 +1274,13 @@ func WithManifestFileContent(content ManifestContent) ManifestFileOption {
 	}
 }
 
+// WithManifestFileFirstRowID sets the first_row_id on a v3+ data manifest.
+func WithManifestFileFirstRowID(firstRowID int64) ManifestFileOption {
+	return func(mf *manifestFile) {
+		mf.FirstRowIDValue = &firstRowID
+	}
+}
+
 func (w *ManifestWriter) ToManifestFile(location string, length int64, opts ...ManifestFileOption) (ManifestFile, error) {
 	if err := w.Close(); err != nil {
 		return nil, err
@@ -1628,6 +1635,7 @@ func WriteManifest(
 	schema *Schema,
 	snapshotID int64,
 	entries []ManifestEntry,
+	opts ...ManifestFileOption,
 ) (mf ManifestFile, err error) {
 	cnt := &internal.CountingWriter{W: out}
 
@@ -1648,7 +1656,7 @@ func WriteManifest(
 		return nil, err
 	}
 
-	return w.ToManifestFile(filename, cnt.Count)
+	return w.ToManifestFile(filename, cnt.Count, opts...)
 }
 
 // ManifestEntryStatus defines constants for the entry status of
--- a/manifest.go
+++ b/manifest.go
@@ -1274,16 +1274,6 @@ func WithManifestFileContent(content ManifestContent) ManifestFileOption {
 	}
 }
 
-// WithManifestFileFirstRowID sets the first_row_id on a v3+ data manifest.
-// Silently no-ops on v1/v2 manifests since the field does not exist in those formats.
-func WithManifestFileFirstRowID(firstRowID int64) ManifestFileOption {
-	return func(mf *manifestFile) {
-		if mf.version >= 3 {
-			mf.FirstRowIDValue = &firstRowID
-		}
-	}
-}
-
 func (w *ManifestWriter) ToManifestFile(location string, length int64, opts ...ManifestFileOption) (ManifestFile, error) {
 	if err := w.Close(); err != nil {
 		return nil, err
@@ -1314,6 +1304,10 @@ func (w *ManifestWriter) ToManifestFile(location string, length int64, opts ...M
 	}
 	for _, apply := range opts {
 		apply(&mf)
+	}
+
+	if mf.Content == ManifestContentDeletes && mf.FirstRowIDValue != nil {
+		return nil, errors.New("first_row_id must not be set on delete manifests")
 	}
 
 	return &mf, nil
@@ -1561,10 +1555,14 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 			wrapped := *(file.(*manifestFile))
 			if m.version == 3 {
 				// Ref: https://github.com/apache/iceberg/blob/ea2071568dc66148b483a82eefedcd2992b435f7/core/src/main/java/org/apache/iceberg/ManifestListWriter.java#L157-L168
-				if wrapped.Content == ManifestContentData && wrapped.FirstRowIDValue == nil {
+				if wrapped.Content == ManifestContentData {
+					if wrapped.FirstRowIDValue == nil {
+						if m.nextRowID != nil {
+							firstRowID := *m.nextRowID
+							wrapped.FirstRowIDValue = &firstRowID
+						}
+					}
 					if m.nextRowID != nil {
-						firstRowID := *m.nextRowID
-						wrapped.FirstRowIDValue = &firstRowID
 						*m.nextRowID += wrapped.ExistingRowsCount + wrapped.AddedRowsCount
 					}
 				}
@@ -1630,10 +1628,6 @@ func WriteManifestList(version int, out io.Writer, snapshotID int64, parentSnaps
 	return writer.AddManifests(files)
 }
 
-// WriteManifest writes a manifest file (a list of manifest entries) as Avro.
-// opts may include ManifestFileOption values such as WithManifestFileContent or
-// WithManifestFileFirstRowID (v3+ only) to set descriptor fields on the returned
-// ManifestFile.
 func WriteManifest(
 	filename string,
 	out io.Writer,
@@ -1642,7 +1636,6 @@ func WriteManifest(
 	schema *Schema,
 	snapshotID int64,
 	entries []ManifestEntry,
-	opts ...ManifestFileOption,
 ) (mf ManifestFile, err error) {
 	cnt := &internal.CountingWriter{W: out}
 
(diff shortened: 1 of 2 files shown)


manifest.go
```diff
@@ -1274,6 +1274,13 @@ func WithManifestFileContent(content ManifestContent) ManifestFileOption {
 	}
 }
 
+// WithManifestFileFirstRowID sets the first_row_id on a v3+ data manifest.
+func WithManifestFileFirstRowID(firstRowID int64) ManifestFileOption {
+	return func(mf *manifestFile) {
+		mf.FirstRowIDValue = &firstRowID
```
zeroshade: nit: `mf.version` is already set before options are applied, so this could self-gate to v3+ where `first_row_id` actually exists:

```go
if mf.version >= 3 {
    mf.FirstRowIDValue = &firstRowID
}
```

As written, calling this against a v1/v2 writer returns a `ManifestFile` whose `FirstRowID()` is non-nil but gets silently dropped on serialize. `WithManifestFileContent` has the same no-guard pattern so I'm fine either way, but the v3 exclusivity is a bit stronger here.

manifest_test.go
```diff
@@ -1158,6 +1158,66 @@ func (m *ManifestTestSuite) TestV3DataManifestFirstRowIDInheritanceSkipsDeletedE
 	m.EqualValues(1000+liveCount, *read[2].DataFile().FirstRowID())
 }
 
+func (m *ManifestTestSuite) TestWriteManifestWithFirstRowIDOption() {
```
zeroshade: nit: consider adding (a) a v1/v2 + `WithManifestFileFirstRowID` case to pin the intended behavior (no-op vs non-nil `FirstRowID()`), and (b) a delete-manifest case, since the spec restricts `first_row_id` to data manifests. The read side already gates inheritance on `formatVersion >= 3 && content == data`, so this just locks down the write side.

manifest.go
```diff
@@ -1628,6 +1635,7 @@ func WriteManifest(
 	schema *Schema,
 	snapshotID int64,
 	entries []ManifestEntry,
+	opts ...ManifestFileOption,
```
zeroshade: nit (pre-existing): `WriteManifest` is exported but has no doc comment. Since you're already touching the signature, a one-liner noting that `opts` set v3-specific descriptor fields (e.g. `WithManifestFileFirstRowID`) would be a welcome add.

zeroshade (approved):
LGTM. The variadic `opts` plumbing through `WriteManifest` is additive and backward-compatible, and the v3 round-trip test (write with the option, read back, assert inheritance) is exactly right. A few small nits inline around v3-only semantics and coverage; none blocking.

manifest.go
the-onewho-knocks: added the mf.version >= 3 guard in b6f3bbb, with a doc comment noting it's a no-op on v1/v2 agreed it's worth being explicit here even though WithManifestFileContent doesn't guard the same way.

manifest_test.go
the-onewho-knocks: Added both in b6f3bbb:

a v1 manifest + WithManifestFileFirstRowID case confirming it's a no-op (FirstRowID() is nil)
a v3 delete-manifest case confirming the field can be set on the manifest itself (the guard only checks version), but entries don't inherit first_row_id on read since that's separately gated on content == data in the reader

manifest.go
the-onewho-knocks: Added a doc comment on WriteManifest in b6f3bbb, calling out that opts can include v3-specific options like WithManifestFileFirstRowID.

manifest.go
```diff
@@ -1274,6 +1274,16 @@ func WithManifestFileContent(content ManifestContent) ManifestFileOption {
 	}
 }
```
laskoviymishka, line 1276: Assuming the option survives the design question above — the doc says "v3+ data manifest" but the closure enforces neither the version nor the content. On v1/v2 we set `FirstRowIDValue` in memory and then `toV1()` silently drops it at encode time, so what you read back diverges from what's persisted, with no error. And nothing stops `WithManifestFileContent(ManifestContentDeletes)` + this producing a delete manifest with a non-null `first_row_id`, which the spec says is always null (Java's list writer asserts `content() != DATA || firstRowId() != null` rather than allowing it).

`ManifestFileOption` is `func(mf *manifestFile)`, so it can't return an error or see the version at apply time — the in-closure guard (`if mf.version < 3 { return }`) is really the only place for the version check. The delete case is trickier because option order matters, so I'd validate that one in `ToManifestFile` after all opts are applied rather than in the closure.

manifest.go
```diff
@@ -1274,6 +1274,16 @@ func WithManifestFileContent(content ManifestContent) ManifestFileOption {
...
 
+// WithManifestFileFirstRowID sets the first_row_id on a v3+ data manifest.
+// Silently no-ops on v1/v2 manifests since the field does not exist in those formats.
+func WithManifestFileFirstRowID(firstRowID int64) ManifestFileOption {
```
laskoviymishka: Independent of the API-shape question — this stores the address of the by-value parameter, so applying the same option instance to two manifests shares one `*int64`. Single-call is fine, but reuse aliases the storage, and any downstream mutation through the pointer would leak across manifests.

Copy into the closure so each application gets its own allocation, the way `AddManifests` does (`firstRowID := *m.nextRowID` before taking its address):

```go
return func(mf *manifestFile) {
	v := firstRowID
	mf.FirstRowIDValue = &v
}
```

manifest_test.go
```diff
@@ -1158,6 +1158,110 @@ func (m *ManifestTestSuite) TestV3DataManifestFirstRowIDInheritanceSkipsDeletedE
...
+	m.Require().NoError(err)
+	m.Require().Len(read, 2)
+	m.Require().NotNil(read[0].DataFile().FirstRowID())
+	m.EqualValues(firstRowID, *read[0].DataFile().FirstRowID())
+	m.Require().NotNil(read[1].DataFile().FirstRowID())
+	m.EqualValues(firstRowID+count, *read[1].DataFile().FirstRowID())
+
+	// Test 2: WriteManifest without option leaves FirstRowID nil (backward compat).
+	var bufNoID bytes.Buffer
+	mf2, err := WriteManifest("/manifest.avro", &bufNoID, 3, partitionSpec, testSchema, entrySnapshotID, entries)
```
laskoviymishka: Once the design and the guards settle, the coverage will want to follow. Right now the two cases cover the happy path and backward-compat, but the no-op behaviors the doc promises are exactly what's untested — v1/v2 dropping the value, and a delete manifest rejecting it. I'd also add an `AddManifests` case with one pre-set and one nil-ID manifest asserting non-overlapping ranges, since that's the scenario most likely to bite.

Smaller things while you're here: these read more naturally as two focused subtests than comment-delimited blocks in one method (if Test 1's `Require` aborts, Test 2 never runs), and the block mixes `m.Equal` and `m.EqualValues` on the same `int64` — neighboring tests use `EqualValues` throughout.

laskoviymishka (changes requested):
Nice, this closes the gap from #1274: `WriteManifest` forwarding opts means `FirstRowID()` finally comes back non-nil for v3, and the read-back test confirms inheritance fires. The functional-option shape matches `WithManifestFileContent` and the variadic widening is ABI-safe.

Before I get into the line-level stuff, though, there's a design question I'd like to settle first, because the answer changes everything below it: should `first_row_id` be settable here at all?

The concern is that pre-seeding it collides with the commit-time allocation in `AddManifests`, which only assigns (and advances its counter) when the value is nil. Pre-set one manifest, hand it to a v3 list writer alongside nil-ID siblings, and the siblings get overlapping row-ID ranges — a uniqueness violation, which is a hard v3 requirement. Java keeps this assignment in the list writer for exactly that reason; its public `ManifestFiles.write()` overloads all pass null. The #1274 use case (round-trip reconstruction, reading the value back) is legitimate — but a free-form setter on the public writer lets a caller silently corrupt the row-ID namespace.

So the question I'd want answered before anything else: do we keep this as a settable option — with a hard doc contract that it's reconstruction-only and must not be set on manifests committed through a new list writer — or make it a dedicated `WriteManifestV3(...) (ManifestFile, int64, error)` that takes and returns the value and makes the constraint impossible to misuse? I lean toward the latter, but I'd like your read first.

I've left inline notes on the concrete issues — version/delete guards, a pointer-aliasing bug, test gaps — but they're all downstream of this decision, so let's settle the API shape before chasing them. wdyt?

manifest.go
the-onewho-knocks: agreed I'll go with WriteManifestV3(...) (ManifestFile, int64, error). A dedicated v3 entrypoint makes the reconstruction only use case explicit at the call site, rather than relying on a doc contract for a general option that looks identical to WithManifestFileContent but carries this footgun. Neither shape changes where the actual safety boundary lives that's still AddManifests's nil-check but this makes accidental misuse much less likely.
Will drop WithManifestFileFirstRowID and rework as described below

the-onewho-knocks: moving to WriteManifestV3 resolves the v1/v2 silent-drop issue structurally, since the function only exists for v3 callers, so there's no toV1() divergence to worry about. For the delete-manifest case, I'll validate in the function itself after all options are applied (inside what's currently ToManifestFile), returning an error if Content == ManifestContentDeletes and a first_row_id was requested, rather than trying to enforce it from inside a single option closure.

manifest.go
the-onewho-knocks: Since WriteManifestV3 takes firstRowID directly as a parameter rather than going through a reusable ManifestFileOption closure, this particular aliasing pattern goes away but I'll still make sure the value is copied into its own allocation at the point it's stored on the manifestFile, same as AddManifests does, rather than taking the address of a shared variable.

manifest_test.go
the-onewho-knocks: Will rework the tests once WriteManifestV3 lands splitting into focused subtests (so an early Require failure doesn't silently skip later cases), adding the delete-manifest rejection case, and an AddManifests case with one pre-set and one nil-ID manifest to assert non-overlapping ranges. Will also switch to EqualValues consistently for the int64 comparisons. Thanks for flagging the test-isolation issue too clearly worth fixing regardless of the API shape.

laskoviymishka (approved):

manifest.go
```diff
@@ -1306,6 +1306,10 @@ func (w *ManifestWriter) ToManifestFile(location string, length int64, opts ...M
 		apply(&mf)
 	}
 
+	if mf.Content == ManifestContentDeletes && mf.FirstRowIDValue != nil {
```
zeroshade, line 1309: This guard is unreachable from the public API: no `ManifestFileOption` sets `FirstRowIDValue` (only `WithManifestFileContent` exists) and `ToManifestFile` inits it to nil. The PR description mentions a `WithManifestFileFirstRowID` option + variadic `WriteManifest`, but neither is in the diff — only this check landed. Either add the missing option/variadic, or drop the dead guard for now.

manifest.go
```diff
@@ -1651,6 +1657,52 @@ func WriteManifest(
...
+	}
+	defer internal.CheckedClose(w, &err)
+
+	for _, entry := range entries {
+		if err := w.addEntry(entry.(*manifestEntry)); err != nil {
+			return nil, 0, err
+		}
+	}
+
+	if err := w.Close(); err != nil {
```
zeroshade, line 1689: `w.Close()` ends up called three times on success (here, again inside `ToManifestFile`, and via the deferred `CheckedClose`). Idempotent so safe, but this explicit close is redundant — `ToManifestFile` already closes to finalize `cnt.Count`. Consider dropping it to mirror `WriteManifest`.

manifest.go
```diff
@@ -1651,6 +1657,52 @@ func WriteManifest(
...
+	}
+
+	mf, err = w.ToManifestFile(filename, cnt.Count)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	raw := mf.(*manifestFile)
+	v := firstRowID
+	raw.FirstRowIDValue = &v
```
zeroshade, line 1700: `FirstRowIDValue` is set post-finalize with no precondition on entry content/status, so passing delete entries yields `Content=ManifestContentData` + a `FirstRowID` anyway — bypassing the guard above. Consider asserting `Content == ManifestContentData` (or rejecting non-data entries) before assigning.

manifest_test.go
```diff
@@ -1158,6 +1159,129 @@ func (m *ManifestTestSuite) TestV3DataManifestFirstRowIDInheritanceSkipsDeletedE
 	m.EqualValues(1000+liveCount, *read[2].DataFile().FirstRowID())
 }
 
+func (m *ManifestTestSuite) TestWriteManifestV3() {
```
zeroshade, line 1162: Coverage gap: only `EntryStatusADDED` entries are exercised. Add a mixed ADDED/EXISTING/DELETED case asserting `nextFirstRowID == firstRowID + AddedRowsCount + ExistingRowsCount` (and that DELETED is excluded) to lock the documented contract.

zeroshade (changes requested):
Thanks @the-onewho-knocks — low-severity notes; the first is the important one (diff vs. description mismatch):

PR #1323, opened 2026-06-26
fix(literals): handle non-int64 big-int correctly when casting DecimalLiteral to Int32

### LINKED ISSUE #1322: DecimalLiteral.To(Int32Type) incorrectly determines overflow direction for values outside int64 range
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

In Iceberg-Go, `DecimalLiteral.To(Int32Type)` in `literals.go` converts a decimal to `Int32` like this:
```go
case Int32Type:
    v := d.Val.BigInt().Int64()
    if v > math.MaxInt32 {
        return Int32AboveMaxLiteral(), nil
    } else if v < math.MinInt32 {
        return Int32BelowMinLiteral(), nil
    }
    return Int32Literal(int32(v)), nil
```

A `(*big.Int).Int64()` does not preserve the original value when the value does not fit in the `int64` range. 

Consequence: the comparison against `MaxInt32` / `MinInt32`:
- pick the _wrong_ sentinel direction (e.g. unscaled `2^63` wraps to `math.MinInt64` and is reported as `Int32BelowMin` instead of `Int32AboveMax`), or
- pick no sentinel at all, returning a spurious in-range `Int32Literal` (e.g. unscaled `-(2^64-1)` wraps to `+1`).

badalprasadsingh:
## Description

Fixes #1322, the incorrect overflow direction detection in `DecimalLiteral.To(Int32Type)` for decimal values whose unscaled `big.Int` exceeds the `int64` range.

## Testing
Added regression tests covering:
1. Positive values outside the `int64` range (`n4`).
2. Negative values that previously produced an incorrect in-range `Int32Literal` (`n5`).
3. Negative values outside the `int64` range that previously returned the wrong overflow sentinel (`n6`).

### DIFF
--- a/literals.go
+++ b/literals.go
@@ -1268,14 +1268,22 @@ func (d DecimalLiteral) To(t Type) (Literal, error) {
 		return nil, fmt.Errorf("%w: could not convert %v to %s",
 			ErrBadCast, d, t)
 	case Int32Type:
-		v := d.Val.BigInt().Int64()
-		if v > math.MaxInt32 {
+		v := d.Val.BigInt()
+		if !v.IsInt64() {
+			if v.Sign() > 0 {
+				return Int32AboveMaxLiteral(), nil
+			} else if v.Sign() < 0 {
+				return Int32BelowMinLiteral(), nil
+			}
+		}
+		i := v.Int64()
+		if i > math.MaxInt32 {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Thanks!

PR #1324, opened 2026-06-26
feat(rest): implement scan planning client methods

Revanth14:
Part of #1178 

## Summary

This PR implements low-level REST client support for server-side scan planning.

It adds scan-planning endpoint constants, capability checks, request/header plumbing, and the four low-level client methods:

- `PlanTableScan`
- `FetchPlanningResult`
- `CancelPlanning`
- `FetchScanTasks`

This is intentionally scoped to the client-method layer only. Higher-level scan orchestration is left for follow-up work.

### Endpoint Negotiation

Added constants for the scan-planning REST operations:

- `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan`
- `GET /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{plan-id}`
- `DELETE /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{plan-id}`
- `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks`

These are included in `allEndpoints` for template validation, but deliberately not in `defaultEndpoints`. Catalogs must explicitly advertise scan-planning support.

### Capability Checks

Implemented:

- `SupportsPlanTableScan`
- `SupportsFullRemoteScanPlanning`
- `SupportsRemoteScanPlanning`

`SupportsFullRemoteScanPlanning` requires all four endpoints because an initial plan may return async or fanout handles that require polling and task fetch support.

### Request/Header Handling

Added per-request support for:

- `Idempotency-Key`
- `X-Iceberg-Access-Delegation`

`PlanTableScan` always sends an idempotency key and uses either a per-request access-delegation value or the session default.

`FetchPlanningResult` uses a per-request access-delegation value when provided, otherwise the session default.

`FetchScanTasks` sends an idempotency key and suppresses access delegation.

`CancelPlanning` suppresses access delegation and intentionally defers cancel-specific idempotency/access-delegation API shape until an options type is added.

### Shared REST Plumbing

Generalized internal REST helpers so requests can:

- set per-request headers
- suppress selected session-default headers
- allow HTTP 204 No Content

`sessionTransport.RoundTrip` now treats any header already present on a request as overriding the session default. This is needed so scan-planning calls can override or suppress the default `X-Iceberg-Access-Delegation: vended-credentials` header.

### Response Validation

Preserved `PlanTableScanResponse.UnmarshalJSON` validation.

Added validation for `FetchPlanningResultResponse`:

- failed responses must include an error payload
- unknown statuses are rejected
- completed, submitted, and cancelled are accepted for polling

A failed `PlanTableScan` response still decodes as `(resp, nil)` when the failed arm contains an error payload; callers must branch on `resp.Status`.

## Out Of Scope

This PR does not implement:

- `WaitForPlan`
- `Catalog.PlanFiles`
- scanner delegation
- expression JSON public wrappers
- file/delete task decoding
- residual decoding
- plan-scoped FileIO wiring
- retry orchestration

## Tests

Added focused tests for endpoint rendering, capability checks, endpoint gating, request headers, default/override access delegation behavior, idempotency key behavior, response validation, `ErrPlanExpired` mapping, and request-helper body closing.

Validated with:

```bash
go test ./catalog/rest
go test ./...

### DIFF
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -72,6 +72,12 @@ const (
 	bearerPrefix        = "Bearer"
 	keyPrefix           = "prefix"
 
+	// headerIcebergAccessDelegation requests server-side storage-credential
+	// vending. It is a session default (defaultAccessDelegation) and a
+	// per-request header on the scan-planning endpoints that vend credentials.
+	headerIcebergAccessDelegation = "X-Iceberg-Access-Delegation"
+	defaultAccessDelegation       = "vended-credentials"
+
 	keyNamespaceSeparator     = "namespace-separator"
 	defaultNamespaceSeparator = "%1F"
 
@@ -223,9 +229,11 @@ const emptyStringHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991
 
 func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 	for k, v := range s.defaultHeaders {
-		// Skip Content-Type if it's already set in the request
-		// to avoid duplicate headers (e.g., when using PostForm)
-		if http.CanonicalHeaderKey(k) == "Content-Type" && r.Header.Get("Content-Type") != "" {
+		// Per-request headers override session defaults. The check is on key
+		// *presence*, not value, so suppressRequestHeaders can opt out of a
+		// default by setting a present-but-empty entry. Keep this in sync with
+		// suppressRequestHeaders.
+		if _, ok := r.Header[http.CanonicalHeaderKey(k)]; ok {
 			continue
 		}
 		for _, hdr := range v {
@@ -287,7 +295,48 @@ func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 	return s.RoundTripper.RoundTrip(r)
 }
 
-func do[T any](ctx context.Context, method string, baseURI *url.URL, path []string, cl *http.Client, override map[int]error, allowNoContent bool) (ret T, err error) {
+// reqConfig holds the optional per-request knobs shared by do and doPost:
+// headers to set, session-default headers to suppress, and whether an HTTP 204
+// is a valid empty result.
+type reqConfig struct {
+	headers         map[string]string
+	suppressHeaders []string
+	allowNoContent  bool
+}
+
+type reqOption func(*reqConfig)
+
+// withHeaders sets per-request headers. They take precedence over session
+// defaults because sessionTransport.RoundTrip skips any default whose key is
+// already present on the request.
+func withHeaders(headers map[string]string) reqOption {
+	return func(c *reqConfig) { c.headers = headers }
+}
+
+// withSuppressedHeaders prevents the named session-default headers from being
+// sent on this request. See suppressRequestHeaders for the mechanism.
+func withSuppressedHeaders(names ...string) reqOption {
+	return func(c *reqConfig) { c.suppressHeaders = names }
+}
+
+// allowNoContent treats an HTTP 204 No Content response as a valid empty result
+// rather than trying to decode a body.
+func allowNoContent() reqOption {
+	return func(c *reqConfig) { c.allowNoContent = true }
+}
+
+func newReqConfig(opts []reqOption) reqConfig {
+	var cfg reqConfig
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+
+	return cfg
+}
+
+func do[T any](ctx context.Context, method string, baseURI *url.URL, path []string, cl *http.Client, override map[int]error, opts ...reqOption) (ret T, err error) {
+	cfg := newReqConfig(opts)
+
 	var (
 		req *http.Request
 		rsp *http.Response
@@ -297,6 +346,8 @@ func do[T any](ctx context.Context, method string, baseURI *url.URL, path []stri
 	if req, err = http.NewRequestWithContext(ctx, method, uri, nil); err != nil {
 		return ret, err
 	}
+	setRequestHeaders(req, cfg.headers)
+	suppressRequestHeaders(req, cfg.suppressHeaders)
 
 	if rsp, err = cl.Do(req); err != nil {
 		return ret, err
@@ -306,7 +357,7 @@ func do[T any](ctx context.Context, method string, baseURI *url.URL, path []stri
 		_ = rsp.Body.Close()
 	}()
 
-	if allowNoContent && rsp.StatusCode == http.StatusNoContent {
+	if cfg.allowNoContent && rsp.StatusCode == http.StatusNoContent {
 		return ret, err
 	}
 
@@ -325,25 +376,23 @@ func do[T any](ctx context.Context, method string, baseURI *url.URL, path []stri
 	return ret, err
 }
 
-func doGet[T any](ctx context.Context, baseURI *url.URL, path []string, cl *http.Client, override map[int]error) (ret T, err error) {
-	return do[T](ctx, http.MethodGet, baseURI, path, cl, override, false)
+func doGet[T any](ctx context.Context, baseURI *url.URL, path []string, cl *http.Client, override map[int]error, opts ...reqOption) (ret T, err error) {
+	return do[T](ctx, http.MethodGet, baseURI, path, cl, override, opts...)
 }
 
-func doDelete[T any](ctx context.Context, baseURI *url.URL, path []string, cl *http.Client, override map[int]error) (ret T, err error) {
-	return do[T](ctx, http.MethodDelete, baseURI, path, cl, override, true)
+func doDelete[T any](ctx context.Context, baseURI *url.URL, path []string, cl *http.Client, override map[int]error, opts ...reqOption) (ret T, err error) {
+	return do[T](ctx, http.MethodDelete, baseURI, path, cl, override, append([]reqOption{allowNoContent()}, opts...)...)
 }
 
 func doHead(ctx context.Context, baseURI *url.URL, path []string, cl *http.Client, override map[int]error) error {
-	_, err := do[struct{}](ctx, http.MethodHead, baseURI, path, cl, override, true)
+	_, err := do[struct{}](ctx, http.MethodHead, baseURI, path, cl, override, allowNoContent())
 
 	return err
 }
 
-func doPost[Payload, Result any](ctx context.Context, baseURI *url.URL, path []string, payload Payload, cl *http.Client, override map[int]error) (ret Result, err error) {
-	return doPostAllowNoContent[Payload, Result](ctx, baseURI, path, payload, cl, override, false)
-}
+func doPost[Payload, Result any](ctx context.Context, baseURI *url.URL, path []string, payload Payload, cl *http.Client, override map[int]error, opts ...reqOption) (ret Result, err error) {
+	cfg := newReqConfig(opts)
 
-func doPostAllowNoContent[Payload, Result any](ctx context.Context, baseURI *url.URL, path []string, payload Payload, cl *http.Client, override map[int]error, allowNoContent bool) (ret Result, err error) {
 	var (
 		req  *http.Request
 		rsp  *http.Response
@@ -360,6 +409,8 @@ func doPostAllowNoContent[Payload, Result any](ctx context.Context, baseURI *url
 	if err != nil {
 		return ret, err
 	}
+	setRequestHeaders(req, cfg.headers)
+	suppressRequestHeaders(req, cfg.suppressHeaders)
 
 	rsp, err = cl.Do(req)
 	if err != nil {
@@ -370,7 +421,7 @@ func doPostAllowNoContent[Payload, Result any](ctx context.Context, baseURI *url
 		_ = rsp.Body.Close()
 	}()
 
-	if allowNoContent && rsp.StatusCode == http.StatusNoContent {
+	if cfg.allowNoContent && rsp.StatusCode == http.StatusNoContent {
 		return ret, err
 	}
 
@@ -387,6 +438,24 @@ func doPostAllowNoContent[Payload, Result any](ctx context.Context, baseURI *url
 	}
 
 	return ret, err
+}
+
+func setRequestHeaders(req *http.Request, headers map[string]string) {
+	for k, v := range headers {
+		req.Header.Set(k, v)
+	}
+}
+
+// suppressRequestHeaders prevents the named session-default headers from being
+// sent on this request. It writes a present-but-empty header entry rather than
+// calling Header.Del: sessionTransport.RoundTrip decides whether to apply a
+// default by testing key *presence* (not value), so a present-but-empty key
+// both blocks the default and emits nothing, whereas deleting the key would let
+// the default reappear. Keep RoundTrip's presence check in sync with this.
+func suppressRequestHeaders(req *http.Request, headers []string) {
+	for _, h := range headers {
+		req.Header[http.CanonicalHeaderKey(h)] = nil
+	}
 }
 
 func handleNon200(rsp *http.Response, override map[int]error) error {
@@ -695,7 +764,7 @@ func (r *Catalog) createSession(ctx context.Context, opts *options) (*http.Clien
 	session.defaultHeaders.Set("X-Client-Version", icebergRestSpecVersion)
 	session.defaultHeaders.Set("Content-Type", "application/json")
 	session.defaultHeaders.Set("User-Agent", "GoIceberg/"+iceberg.Version())
-	session.defaultHeaders.Set("X-Iceberg-Access-Delegation", "vended-credentials")
+	session.defaultHeaders.Set(headerIcebergAccessDelegation, defaultAccessDelegation)
 
 	for k, v := range opts.headers {
 		session.defaultHeaders.Set(k, v)
@@ -1156,7 +1225,7 @@ func (r *Catalog) CommitTransaction(ctx context.Context, commits []table.TableCo
 		return err
 	}
 
-	_, err = doPostAllowNoContent[payload, struct{}](
+	_, err = doPost[payload, struct{}](
 		ctx, r.baseURI, path,
 		payload{TableChanges: changes}, r.cl,
 		map[int]error{
@@ -1167,7 +1236,7 @@ func (r *Catalog) CommitTransaction(ctx context.Context, commits []table.TableCo
 			http.StatusServiceUnavailable:  ErrCommitStateUnknown,
 			http.StatusGatewayTimeout:      ErrCommitStateUnknown,
 		},
-		true,
+		allowNoContent(),
 	)
 
 	return err
@@ -1359,8 +1428,8 @@ func (r *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
 		return nil, err
 	}
 
-	_, err = doPostAllowNoContent[payload, any](ctx, r.baseURI, path, payload{Source: src, Destination: dst}, r.cl,
-		map[int]error{http.StatusNotFound: catalog.ErrNoSuchTable}, true)
+	_, err = doPost[payload, any](ctx, r.baseURI, path, payload{Source: src, Destination: dst}, r.cl,
+		map[int]error{http.StatusNotFound: catalog.ErrNoSuchTable}, allowNoContent())
 	if err != nil {
 		return nil, err
 	}
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -228,12 +228,17 @@ type sessionTransport struct {
 const emptyStringHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 
 func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
+	// A session default is applied unless the request already carries that
+	// header (a per-request override of any default, not just Content-Type
+	// wins) or explicitly opted out of it via withSuppressedHeaders (carried on
+	// the context as an explicit set, never inferred from header values).
+	suppressed := suppressedHeadersFrom(r.Context())
 	for k, v := range s.defaultHeaders {
-		// Per-request headers override session defaults. The check is on key
-		// *presence*, not value, so suppressRequestHeaders can opt out of a
-		// default by setting a present-but-empty entry. Keep this in sync with
-		// suppressRequestHeaders.
-		if _, ok := r.Header[http.CanonicalHeaderKey(k)]; ok {
+		ck := http.CanonicalHeaderKey(k)
+		if _, ok := r.Header[ck]; ok {
+			continue
+		}
+		if _, ok := suppressed[ck]; ok {
 			continue
 		}
 		for _, hdr := range v {
@@ -295,6 +300,11 @@ func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 	return s.RoundTripper.RoundTrip(r)
 }
 
+// suppressHeadersKey carries the set of session-default header keys (canonical
+// form) that a request opted out of, so sessionTransport.RoundTrip can consult
+// an explicit set rather than inferring suppression from header values.
+type suppressHeadersKey struct{}
+
 // reqConfig holds the optional per-request knobs shared by do and doPost:
 // headers to set, session-default headers to suppress, and whether an HTTP 204
 // is a valid empty result.
@@ -306,17 +316,24 @@ type reqConfig struct {
 
 type reqOption func(*reqConfig)
 
-// withHeaders sets per-request headers. They take precedence over session
-// defaults because sessionTransport.RoundTrip skips any default whose key is
-// already present on the request.
+// withHeaders sets per-request headers, merging with any already set so options
+// compose. A per-request header takes precedence over a session default because
+// sessionTransport.RoundTrip skips any default whose key is already present.
 func withHeaders(headers map[string]string) reqOption {
-	return func(c *reqConfig) { c.headers = headers }
+	return func(c *reqConfig) {
+		if c.headers == nil {
+			c.headers = make(map[string]string, len(headers))
+		}
+		maps.Copy(c.headers, headers)
+	}
 }
 
 // withSuppressedHeaders prevents the named session-default headers from being
-// sent on this request. See suppressRequestHeaders for the mechanism.
+// sent on this request. Names accumulate so options compose. Suppression is
+// carried through the request context (see RoundTrip), not encoded as a header
+// value, so req.Header only ever means "headers to send".
 func withSuppressedHeaders(names ...string) reqOption {
-	return func(c *reqConfig) { c.suppressHeaders = names }
+	return func(c *reqConfig) { c.suppressHeaders = append(c.suppressHeaders, names...) }
 }
 
 // allowNoContent treats an HTTP 204 No Content response as a valid empty result
@@ -334,6 +351,27 @@ func newReqConfig(opts []reqOption) reqConfig {
 	return cfg
 }
 
+// withSuppressedHeadersCtx returns ctx carrying the canonicalized suppression
+// set, or ctx unchanged when nothing is suppressed.
+func withSuppressedHeadersCtx(ctx context.Context, names []string) context.Context {
+	if len(names) == 0 {
+		return ctx
+	}
+
+	set := make(map[string]struct{}, len(names))
+	for _, n := range names {
+		set[http.CanonicalHeaderKey(n)] = struct{}{}
+	}
+
+	return context.WithValue(ctx, suppressHeadersKey{}, set)
+}
+
+func suppressedHeadersFrom(ctx context.Context) map[string]struct{} {
+	set, _ := ctx.Value(suppressHeadersKey{}).(map[string]struct{})
+
+	return set
+}
+
 func do[T any](ctx context.Context, method string, baseURI *url.URL, path []string, cl *http.Client, override map[int]error, opts ...reqOption) (ret T, err error) {
 	cfg := newReqConfig(opts)
 
@@ -343,11 +381,11 @@ func do[T any](ctx context.Context, method string, baseURI *url.URL, path []stri
 	)
 
 	uri := baseURI.JoinPath(path...).String()
+	ctx = withSuppressedHeadersCtx(ctx, cfg.suppressHeaders)
 	if req, err = http.NewRequestWithContext(ctx, method, uri, nil); err != nil {
 		return ret, err
 	}
 	setRequestHeaders(req, cfg.headers)
-	suppressRequestHeaders(req, cfg.suppressHeaders)
 
 	if rsp, err = cl.Do(req); err != nil {
 		return ret, err
@@ -405,12 +443,12 @@ func doPost[Payload, Result any](ctx context.Context, baseURI *url.URL, path []s
 		return ret, err
 	}
 
+	ctx = withSuppressedHeadersCtx(ctx, cfg.suppressHeaders)
 	req, err = http.NewRequestWithContext(ctx, http.MethodPost, uri, bytes.NewReader(data))
 	if err != nil {
 		return ret, err
 	}
 	setRequestHeaders(req, cfg.headers)
-	suppressRequestHeaders(req, cfg.suppressHeaders)
 
 	rsp, err = cl.Do(req)
 	if err != nil {
@@ -443,18 +481,6 @@ func doPost[Payload, Result any](ctx context.Context, baseURI *url.URL, path []s
 func setRequestHeaders(req *http.Request, headers map[string]string) {
 	for k, v := range headers {
 		req.Header.Set(k, v)
-	}
-}
-
-// suppressRequestHeaders prevents the named session-default headers from being
-// sent on this request. It writes a present-but-empty header entry rather than
-// calling Header.Del: sessionTransport.RoundTrip decides whether to apply a
-// default by testing key *presence* (not value), so a present-but-empty key
-// both blocks the default and emits nothing, whereas deleting the key would let
-// the default reappear. Keep RoundTrip's presence check in sync with this.
-func suppressRequestHeaders(req *http.Request, headers []string) {
-	for _, h := range headers {
-		req.Header[http.CanonicalHeaderKey(h)] = nil
 	}
 }
 
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -246,6 +246,12 @@ func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 		}
 	}
 
+	// Ordering is load-bearing: authManager runs after the default-header loop
+	// and Set-overwrites, so the managed auth header always wins even though the
+	// loop above lets an arbitrary per-request header (withHeaders) override a
+	// session default of the same key. A caller cannot suppress or spoof the
+	// Authorization header by supplying its own. Do not reorder this before the
+	// default-header loop.
 	if s.authManager != nil && r.Context().Value(skipOAuth) == nil {
 		k, v, err := s.authManager.AuthHeader()
 		if err != nil {
@@ -306,12 +312,14 @@ func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 type suppressHeadersKey struct{}
 
 // reqConfig holds the optional per-request knobs shared by do and doPost:
-// headers to set, session-default headers to suppress, and whether an HTTP 204
-// is a valid empty result.
+// headers to set, session-default headers to suppress, a body error.type ->
+// error override, and whether an HTTP 204 is a valid empty result.
 type reqConfig struct {
-	headers         map[string]string
-	suppressHeaders []string
-	allowNoContent  bool
+	headers           map[string]string
+	suppressHeaders   []string
+	errorTypeOverride map[string]error
+	allowNoContent    bool
+	requireBody       bool
 }
 
 type reqOption func(*reqConfig)
@@ -336,10 +344,33 @@ func withSuppressedHeaders(names ...string) reqOption {
 	return func(c *reqConfig) { c.suppressHeaders = append(c.suppressHeaders, names...) }
 }
 
+// withErrorTypeOverride maps a non-200 response to a specific error by the REST
+// error.type in the response body. It takes precedence over the status-code
+// override so a shared status (e.g. a 404 that can be a missing plan, table, or
+// namespace) can split into distinct sentinels. Entries accumulate so options
+// compose; an unset or unmatched error.type falls through to the status-code
+// override.
+func withErrorTypeOverride(byType map[string]error) reqOption {
+	return func(c *reqConfig) {
+		if c.errorTypeOverride == nil {
+			c.errorTypeOverride = make(map[string]error, len(byType))
+		}
+		maps.Copy(c.errorTypeOverride, byType)
+	}
+}
+
 // allowNoContent treats an HTTP 204 No Content response as a valid empty result
 // rather than trying to decode a body.
 func allowNoContent() reqOption {
 	return func(c *reqConfig) { c.allowNoContent = true }
+}
+
+// requireBody rejects an HTTP 200 with an empty body rather than decoding it to
+// the zero result. Use it when a 200 must carry a payload, so a truncated or
+// proxy-stripped response cannot masquerade as a successful empty result (e.g.
+// fetchScanTasks returning "zero tasks" and silently dropping work).
+func requireBody() reqOption {
+	return func(c *reqConfig) { c.requireBody = true }
 }
 
 func newReqConfig(opts []reqOption) reqConfig {
@@ -400,7 +431,7 @@ func do[T any](ctx context.Context, method string, baseURI *url.URL, path []stri
 	}
 
 	if rsp.StatusCode != http.StatusOK {
-		return ret, handleNon200(rsp, override)
+		return ret, handleNon200(rsp, override, cfg.errorTypeOverride)
 	}
 
 	if method == http.MethodHead || method == http.MethodDelete {
@@ -464,10 +495,14 @@ func doPost[Payload, Result any](ctx context.Context, baseURI *url.URL, path []s
 	}
 
 	if rsp.StatusCode != http.StatusOK {
-		return ret, handleNon200(rsp, override)
+		return ret, handleNon200(rsp, override, cfg.errorTypeOverride)
 	}
 
 	if rsp.ContentLength == 0 {
+		if cfg.requireBody {
+			return ret, fmt.Errorf("%w: empty response body on %s %s", ErrRESTError, req.Method, req.URL.Path)
+		}
+
 		return ret, err
 	}
 
@@ -484,7 +519,7 @@ func setRequestHeaders(req *http.Request, headers map[string]string) {
 	}
 }
 
-func handleNon200(rsp *http.Response, override map[int]error) error {
+func handleNon200(rsp *http.Response, override map[int]error, typeOverride map[string]error) error {
 	var e errorResponse
 
 	// Only try to decode if there's a body (HEAD requests don't have one)
@@ -509,9 +544,23 @@ func handleNon200(rsp *http.Response, override map[int]error) error {
 		}
 	}
 
+	// The status-code override maps a status to a sentinel; a body error.type
+	// override refines it further (e.g. splitting a 404 into missing-plan vs.
+	// table vs. namespace). The type refinement only applies to a status the
+	// caller already mapped, because these error.type values are defined for
+	// specific statuses (404) in the REST spec: a 503 body carrying, say,
+	// NoSuchPlanIdException must still resolve to ErrServiceUnavailable rather
+	// than a 404 sentinel.
 	if override != nil {
-		if err, ok := override[rsp.StatusCode]; ok {
-			e.wrapping = err
+		if statusErr, ok := override[rsp.StatusCode]; ok {
+			if typeOverride != nil && e.Type != "" {
+				if typeErr, ok := typeOverride[e.Type]; ok {
+					e.wrapping = typeErr
+
+					return e
+				}
+			}
+			e.wrapping = statusErr
 
 			return e
 		}
--- a/catalog/rest/rest_internal_test.go
+++ b/catalog/rest/rest_internal_test.go
@@ -914,22 +914,22 @@ func TestResponseBodyLeak(t *testing.T) {
 		baseURI, err := url.Parse(srv.URL)
 		require.NoError(t, err)
 
-		_, err = do[struct{}](context.Background(), http.MethodGet, baseURI, []string{"test"}, client, nil, false)
+		_, err = do[struct{}](context.Background(), http.MethodGet, baseURI, []string{"test"}, client, nil)
 		require.Error(t, err)
 
 		assert.True(t, tracker.body.closed,
 			"response body should be closed on non-200 status")
 	})
 
-	t.Run("doPostAllowNoContent", func(t *testing.T) {
+	t.Run("doPost", func(t *testing.T) {
 		tracker := &trackingTransport{transport: http.DefaultTransport}
 		client := &http.Client{Transport: tracker}
 
 		baseURI, err := url.Parse(srv.URL)
 		require.NoError(t, err)
 
-		_, err = doPostAllowNoContent[map[string]string, struct{}](
-			context.Background(), baseURI, []string{"test"}, map[string]string{"key": "value"}, client, nil, false)
+		_, err = doPost[map[string]string, struct{}](
+			context.Background(), baseURI, []string{"test"}, map[string]string{"key": "value"}, client, nil, allowNoContent())
 		require.Error(t, err)
 
 		assert.True(t, tracker.body.closed,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by Revanth14** [comment]
@laskoviymishka Please take a look when you got time.

File: catalog/rest/scan_planning.go [hunk 1]
```diff
@@ -80,29 +91,97 @@ func (r *Catalog) PlanFiles(ctx context.Context, req table.ScanPlanningRequest)
...
+	return doGet[FetchPlanningResultResponse](
+		ctx, r.baseURI, path, r.cl,
+		map[int]error{http.StatusNotFound: ErrPlanExpired}, withHeaders(headers))
 }
 
 // CancelPlanning cancels a server-side plan. Callers should cancel on context
-// cancellation using a detached context with a short timeout.
+// cancellation using a detached context with a short timeout. The spec supports
+// idempotency and access-delegation headers on cancel; this low-level method
+// deliberately defers those until a cancel options type is added, and suppresses
```
zeroshade, line 277: nit: per the REST spec, `cancelPlanning` does accept an `Idempotency-Key` (the `idempotency-key` parameter on the DELETE in rest-catalog-open-api.yaml), so deferring it here is reasonable - just flagging so the follow-up that introduces the cancel options type doesn't lose it. The best-effort cancel + 204 handling and access-delegation suppression here look correct.

zeroshade (commented):
A couple of non-blocking documentation nits from the scan-planning client methods (comment-only, not requesting changes; the low-level surface looks spec-compliant otherwise):

**Stale file header - `catalog/rest/scan_planning.go` lines 17-24:** the top-of-file comment still says the method bodies are intentionally unimplemented stubs (returning `ErrNotImplemented`) and that endpoint capability discovery lands separately in the Phase 0 PR. This PR implements `PlanTableScan` / `FetchPlanningResult` / `CancelPlanning` / `FetchScanTasks` plus the `Supports*` capability checks, so that header is now stale and worth updating or removing. (Noting it here rather than inline because those lines aren't part of the diff.)

Revanth14:
> A couple of non-blocking documentation nits from the scan-planning client methods (comment-only, not requesting changes; the low-level surface looks spec-compliant otherwise):
> 
> **Stale file header - `catalog/rest/scan_planning.go` lines 17-24:** the top-of-file comment still says the method bodies are intentionally unimplemented stubs (returning `ErrNotImplemented`) and that endpoint capability discovery lands separately in the Phase 0 PR. This PR implements `PlanTableScan` / `FetchPlanningResult` / `CancelPlanning` / `FetchScanTasks` plus the `Supports*` capability checks, so that header is now stale and worth updating or removing. (Noting it here rather than inline because those lines aren't part of the diff.)

Updated the stale scan_planning.go file header to reflect the implemented low-level client methods and remaining follow-up scope.

catalog/rest/scan_planning.go
```diff
@@ -113,6 +189,48 @@ func (r *Catalog) WaitForPlan(ctx context.Context, ident table.Identifier, planI
...
+	}
+
+	return headers, nil
+}
+
+func idempotencyHeaderValue(idempotencyKey *string) (string, error) {
+	if idempotencyKey == nil {
+		// Plan and task POSTs always send an idempotency key. There is no
+		// transport retry here, so nil means a fresh key for this call.
+		return uuid.NewString(), nil
```
laskoviymishka: `uuid.NewString()` gives us a UUIDv4, but the spec pins this header to "UUIDv7 in string form (RFC 9562)" and Java generates it via `UUIDUtil.generateUuidV7()`. Servers that key the idempotency window off the embedded timestamp will treat our v4 keys as undefined, so cross-client dedup against a Java-based server is broken.

`google/uuid` v1.6.0 is already in go.mod and exposes `uuid.NewV7()` — I'd swap to that and propagate the error it returns.

While we're here: Java only sends idempotency headers when the server advertises a key lifetime (`config.idempotencyKeyLifetime()`), whereas we send one on every plan and task POST. The header's optional, so this isn't wrong, but strict servers could reject the unsolicited header. Worth a follow-up to gate it on config, even if not in this PR.

catalog/rest/scan_planning.go
```diff
@@ -80,29 +88,97 @@ func (r *Catalog) PlanFiles(ctx context.Context, req table.ScanPlanningRequest)
...
+	}
+
+	headers, err := scanPlanningHeaders(nil, opts.AccessDelegation, false)
+	if err != nil {
+		return FetchPlanningResultResponse{}, err
+	}
+
+	return doGet[FetchPlanningResultResponse](
+		ctx, r.baseURI, path, r.cl,
+		map[int]error{http.StatusNotFound: ErrPlanExpired}, withHeaders(headers))
```
laskoviymishka: This maps every 404 to `ErrPlanExpired`, but the spec's `fetchPlanningResult` 404 can be `NoSuchPlanIdException`, `NoSuchTableException`, or `NoSuchNamespaceException`. A dropped table comes back as "plan expired," which is the one distinction `WaitForPlan` actually needs — "plan expired, retry with a new plan" vs. "table's gone, abort."

I'd settle this before the polling layer lands on top of it, since fixing it later means reworking the retry contract. The clean version is inspecting the body's `error.type` before mapping — either a pre-decode pass or a typed override hook in `handleNon200` that can split on body content. wdyt?

catalog/rest/scan_planning.go
```diff
@@ -80,29 +88,97 @@ func (r *Catalog) PlanFiles(ctx context.Context, req table.ScanPlanningRequest)
...
+	}
+
+	headers, err := scanPlanningHeaders(req.IdempotencyKey, nil, true)
+	if err != nil {
+		return FetchScanTasksResponse{}, err
+	}
+
+	return doPost[FetchScanTasksRequest, FetchScanTasksResponse](
+		ctx, r.baseURI, path, req, r.cl,
+		map[int]error{http.StatusNotFound: catalog.ErrNoSuchTable},
```
laskoviymishka: Same shape as the `FetchPlanningResult` 404, the other direction: here every 404 becomes `catalog.ErrNoSuchTable`, but the spec's `fetchScanTasks` 404 also covers `NoSuchPlanTaskException`. A caller fanning out over plan-task handles will read `ErrNoSuchTable` and conclude the table vanished when really a task handle expired. If we go the `error.type` inspection route for the `FetchPlanningResult` 404, the same hook covers this — and we'd want an `ErrNoSuchPlanTask` sentinel to map onto.

catalog/rest/scan_planning.go
```diff
@@ -80,29 +88,97 @@ func (r *Catalog) PlanFiles(ctx context.Context, req table.ScanPlanningRequest)
 // --- Low-level client methods -----------------------------------------------
 
 // PlanTableScan submits a scan plan. The result is either completed inline,
-// submitted (returns a plan-id to poll), or failed.
+// submitted (returns a plan-id to poll), or failed. A failed planning response
+// decodes as (resp, nil); callers must branch on resp.Status.
 func (r *Catalog) PlanTableScan(ctx context.Context, ident table.Identifier, req PlanTableScanRequest) (PlanTableScanResponse, error) {
```
laskoviymishka, line 195: I think this contract is going to bite callers. Returning `(resp, nil)` for a `failed` plan means anyone using the `if err != nil` idiom treats a failure as success, then reads `resp.PlanID` (nil for failed) and panics. The doc comment is honest about it, but the comment is the only thing standing between this and a nil-deref downstream.

I'd lean toward returning an error on `failed` — something like `fmt.Errorf("%w: plan failed: %s", ErrRESTError, resp.Error.Message)`, or a typed error wrapping the `PlanningError` so callers can still get at the structured detail. The `completed`/`submitted` split stays in the response.

Either way I'd settle it here rather than after `WaitForPlan` and the scanner are written against the current shape — and the same note needs to go on `FetchPlanningResult`, which has the identical `(resp, nil)` contract and is exactly what the poller will loop on. Thoughts?

catalog/rest/scan_planning.go
```diff
@@ -54,21 +59,24 @@ var ErrPlanExpired = fmt.Errorf("%w: scan plan expired", ErrRESTError)
...
+		r.endpoints.contains(endpointFetchScanTasks)
 }
 
 // --- table.ScanPlanner implementation ---------------------------------------
 
 // SupportsRemoteScanPlanning reports whether this catalog can complete a remote
 // plan end-to-end; backed by the split capability checks above.
 func (r *Catalog) SupportsRemoteScanPlanning() bool {
-	return false
+	return r.SupportsFullRemoteScanPlanning()
```
laskoviymishka: This requires all four endpoints, where Java's minimal check only needs the submit endpoint for the sync-only case (`RESTSessionCatalog.java:641`). A server that advertises only `plan` — completed-inline, no polling — passes Java but gets rejected here and falls back to client-side planning.

The doc comment ("complete a plan end-to-end") is consistent with requiring all four, so this isn't a contradiction so much as a semantics call: is `SupportsRemoteScanPlanning` "can I do the full async dance" or "can the server plan at all"? I'd lean toward the latter to match Java — return `SupportsPlanTableScan()`, and let callers needing fanout check `SupportsFullRemoteScanPlanning()` explicitly. wdyt?

catalog/rest/rest.go
```diff
@@ -389,6 +440,24 @@ func doPostAllowNoContent[Payload, Result any](ctx context.Context, baseURI *url
...
+
+// suppressRequestHeaders prevents the named session-default headers from being
+// sent on this request. It writes a present-but-empty header entry rather than
+// calling Header.Del: sessionTransport.RoundTrip decides whether to apply a
+// default by testing key *presence* (not value), so a present-but-empty key
+// both blocks the default and emits nothing, whereas deleting the key would let
+// the default reappear. Keep RoundTrip's presence check in sync with this.
+func suppressRequestHeaders(req *http.Request, headers []string) {
+	for _, h := range headers {
+		req.Header[http.CanonicalHeaderKey(h)] = nil
```
laskoviymishka: This leans on net/http not writing a header whose value slice is nil — correct today, but it's an internal wire invariant, not a documented contract, and the comment calls it "present-but-empty" when it's actually nil. If a runtime change ever altered that, the suppression breaks silently and `X-Iceberg-Access-Delegation` reappears on cancel and task-fetch.

Two ways to firm this up. The lighter one is `req.Header.Set(h, "")` plus a comment — present, empty string, still skipped by the presence check, no reliance on nil-slice behavior. The sturdier one is moving suppression into `reqConfig` as an explicit suppress-set that `RoundTrip` consults directly, since `RoundTrip` already owns session-default logic — then there's no two-sided "keep in sync" invariant at all. I'd lean toward the latter given a follow-up PR will be poking at this. Thoughts?

catalog/rest/rest.go
```diff
@@ -287,7 +295,48 @@ func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+// withHeaders sets per-request headers. They take precedence over session
+// defaults because sessionTransport.RoundTrip skips any default whose key is
+// already present on the request.
+func withHeaders(headers map[string]string) reqOption {
+	return func(c *reqConfig) { c.headers = headers }
+}
+
+// withSuppressedHeaders prevents the named session-default headers from being
+// sent on this request. See suppressRequestHeaders for the mechanism.
+func withSuppressedHeaders(names ...string) reqOption {
```
laskoviymishka, line 343: Both options assign rather than merge, so two `withSuppressedHeaders(...)` on one call silently drops all but the last. Not a bug today since each is passed once, but it's a trap for the next caller who reasonably assumes options compose.

I'd make `withSuppressedHeaders` append (`c.suppressHeaders = append(c.suppressHeaders, names...)`) and merge maps in `withHeaders`.

catalog/rest/rest.go
```diff
@@ -223,9 +229,11 @@ const emptyStringHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991
...
 func (s *sessionTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 	for k, v := range s.defaultHeaders {
-		// Skip Content-Type if it's already set in the request
-		// to avoid duplicate headers (e.g., when using PostForm)
-		if http.CanonicalHeaderKey(k) == "Content-Type" && r.Header.Get("Content-Type") != "" {
+		// Per-request headers override session defaults. The check is on key
+		// *presence*, not value, so suppressRequestHeaders can opt out of a
+		// default by setting a present-but-empty entry. Keep this in sync with
+		// suppressRequestHeaders.
+		if _, ok := r.Header[http.CanonicalHeaderKey(k)]; ok {
```
laskoviymishka: This broadens the old behavior: it used to skip only Content-Type, now it skips any session default whose key is already present. That's almost certainly what we want, but the oauth2 Content-Type correctness that the old special-case guaranteed is now a side-effect of the general rule. I'd just note in the comment that the rule applies to any per-request header sharing a key with a session default, so the next person doesn't re-add a Content-Type special case.

catalog/rest/scan_planning_test.go
```diff
@@ -94,3 +183,298 @@ func TestPlanTableScanResponseRejectsUnknownStatus(t *testing.T) {
...
+		IdempotencyKey: &idempotencyKey,
+		PlanTask:       "task-1",
+	})
+	require.NoError(t, err)
+	assert.Equal(t, []string{"child-task"}, resp.PlanTasks)
+	assert.Len(t, resp.FileScanTasks, 1)
+	assert.Len(t, resp.DeleteFiles, 1)
+}
+
+func TestScanPlanningEndpointGating(t *testing.T) {
```
laskoviymishka, line 708: This runs all four endpoint checks in one body, so a failing `require` on the first masks the other three. `TestScanPlanningCapabilities` right above it uses `t.Run` subtests — I'd match that here so each endpoint reports independently.

catalog/rest/scan_planning_test.go
```diff
@@ -94,3 +183,298 @@ func TestPlanTableScanResponseRejectsUnknownStatus(t *testing.T) {
 	err := json.Unmarshal([]byte(`{"status":"bogus"}`), &resp)
 	require.ErrorIs(t, err, ErrRESTError)
 }
+
+func TestFetchPlanningResultResponseRejectsMalformedResponses(t *testing.T) {
```
laskoviymishka, line 227: The `cancelled` arm in `FetchPlanningResultResponse.UnmarshalJSON` is accepted silently with no test, so a refactor routing `cancelled` into the `default` error case would slip through — and `cancelled` is specifically valid for poll results, unlike `PlanTableScanResponse`. A subtest unmarshaling `{"status":"cancelled"}` asserting NoError + `PlanStatusCancelled` would pin it.

Same gap on the round-trip side: `TestFetchPlanningResultRequest` only covers `completed`, while `TestPlanTableScanRequest` has a `failed` arm. A `failed`-status HTTP case here would exercise the validation through the real client path, not just direct `json.Unmarshal`.

catalog/rest/rest_internal_test.go
```diff
@@ -914,22 +914,22 @@ func TestResponseBodyLeak(t *testing.T) {
 		baseURI, err := url.Parse(srv.URL)
 		require.NoError(t, err)
 
-		_, err = do[struct{}](context.Background(), http.MethodGet, baseURI, []string{"test"}, client, nil, false)
+		_, err = do[struct{}](context.Background(), http.MethodGet, baseURI, []string{"test"}, client, nil)
```
laskoviymishka, line 1022: The generalized presence-check override in `RoundTrip` is only tested transitively through the scan-planning end-to-end tests. Since it's the most consequential behavior change in the PR, I'd add a direct unit test here: build a `sessionTransport` with a default header, send a request that already sets that header, call `RoundTrip`, and assert the default wasn't appended. That pins the invariant independently of the scan-planning callers.

catalog/rest/rest.go
zeroshade: Consumer side of the suppression trick: this has to test key *presence* (not value) so that a present-but-`nil` entry both blocks the default and emits nothing. That presence-not-value invariant is what I'd like to retire (see the review summary) — with an explicit suppress set carried on the request context, this stays a plain "explicit per-request override wins" check and the nil-entry coupling disappears.

catalog/rest/rest.go
zeroshade: Producer side of the same trick: writing a `nil` entry to signal "suppress" overloads `req.Header` with two meanings. Prefer threading `suppressHeaders` through the request context so `RoundTrip` consults an explicit set, rather than inferring intent from a nil header value.

zeroshade (changes requested):
Thanks @Revanth14 — this is solid, and the functional-options refactor (`reqOption`) is a genuinely nice cleanup over the old boolean params. One change request before merge, scoped to **how header suppression is implemented**.

**The concern.** Per-request *override* of session defaults is good. The issue is *suppression*: `suppressRequestHeaders` writes a present-but-`nil` entry into `req.Header`, and `RoundTrip` then decides whether to apply a default by testing key **presence, not value**. That:
- couples two distant sites with an invariant ("RoundTrip must check presence, never value") that a future refactor can silently break, and
- overloads `req.Header` to mean two different things — "override with this value" *and* "suppress this default".

**Preferred (Option B): carry the suppression set explicitly instead of encoding it as a nil header.**

Keep real per-request overrides exactly as they are (a real value via `setRequestHeaders`, which `RoundTrip` then lets win). For *suppression*, thread the set through the request **context** rather than `req.Header[k] = nil`:

```go
// in do/doPost, when cfg.suppressHeaders is non-empty:
ctx = context.WithValue(ctx, suppressHeadersKey{}, cfg.suppressHeaders)

// RoundTrip:
suppressed := suppressedFrom(r.Context()) // set of canonical keys
for k, v := range s.defaultHeaders {
    ck := http.CanonicalHeaderKey(k)
    if _, ok := r.Header[ck]; ok { continue } // explicit per-request override wins
    if suppressed[ck] { continue }            // explicit suppression
    for _, h := range v { r.Header.Add(k, h) }
}
```

**Why it's preferred**
1. **Single source of truth** — suppression intent lives in one explicit place; the "keep these two checks in sync" comment goes away.
2. **No overloaded semantics** — `req.Header` only ever means "headers to send"; the presence-vs-value subtlety disappears.
3. **Refactor-safe** — removing the fragile `if _, ok := r.Header[...]` presence-not-value invariant means a future RoundTrip change can't silently re-enable a suppressed default.
4. **Behavior-identical** — callers see no difference; this is purely how suppression is plumbed.

Everything else (capability gating, idempotency-key handling, access-delegation defaulting, the httptest coverage) looks great. Happy to discuss if you'd prefer to keep the current approach — but I'd like the suppression decoupled from header presence before this merges.

laskoviymishka (changes requested):
The client-method layer is in good shape, and the scope discipline is exactly right — four methods wired, capability checks split three ways, header plumbing through functional options, and `WaitForPlan`/scanner delegation cleanly deferred. The `reqConfig`/`reqOption` refactor reads well and the `UnmarshalJSON` union validation is a nice touch.

I'd hold this before merging though. The idempotency key is generated as a UUIDv4 (`uuid.NewString()`), but the spec pins the header to UUIDv7 and Java generates it via `UUIDUtil.generateUuidV7()`. Servers that key their dedup window off the embedded timestamp will treat our keys as undefined, so cross-client interop is broken. `google/uuid` already gives us `NewV7()`, so it's a one-line swap plus error propagation.

The other thing I'd settle in this PR is the 404 mapping on `FetchPlanningResult`. Every 404 collapses to `ErrPlanExpired`, but the spec's 404 there can also be a dropped table or namespace — and "plan expired, retry" vs. "table gone, abort" is exactly the distinction `WaitForPlan` will need. Fixing it after the poller is written means reworking the retry contract, so I'd rather get it right here. `FetchScanTasks` has the mirror version, collapsing `NoSuchPlanTaskException` into `ErrNoSuchTable`.

Things I'd like to settle in this PR before the follow-ups build on it:

- generate the idempotency key as UUIDv7 (`uuid.NewV7()`) and propagate the error
- split the `FetchPlanningResult`/`FetchScanTasks` 404s on the body's `error.type` so plan-expired, plan-task-not-found, and table-not-found stay distinct
- decide the `(resp, nil)`-on-failed-plan contract now — I'd lean toward returning an error so `if err != nil` callers don't nil-deref `resp.PlanID`
- decide whether `SupportsRemoteScanPlanning` should require all four endpoints or just `plan` (Java uses the single endpoint for the sync-only case)
- firm up the header suppression so it doesn't rely on net/http's nil-slice wire behavior, and make `withHeaders`/`withSuppressedHeaders` compose rather than last-writer-wins

A few smaller test asks too: subtests in `TestScanPlanningEndpointGating`, a `cancelled` and a `failed` round-trip case for `FetchPlanningResult`, and a direct `RoundTrip` unit test for the new presence-check override.

catalog/rest/rest.go
Revanth14: Thanks, addressed this by removing the nil-header suppression path entirely.

Suppression is now carried explicitly on the request context via a canonicalized suppression set. `RoundTrip` still treats a real per-request header as an override, but suppression is checked separately from `req.Header`, so `req.Header` only represents headers that should actually be sent.

Also added direct coverage for:
- default header applied when absent
- per-request override avoiding duplicate defaults
- context suppression dropping the default
- `withHeaders` / `withSuppressedHeaders` composition

catalog/rest/rest.go
Revanth14: Yep, addressed this with the same refactor as the consumer-side thread.

`suppressRequestHeaders` was removed, and suppression is no longer represented by writing a nil entry into `req.Header`. `do` / `doPost` now attach a canonicalized suppression set to the request context, and `RoundTrip` consults that explicit set when deciding whether to apply session defaults.

So the producer side no longer overloads `req.Header`; real request headers remain only the headers we intend to send.

catalog/rest/scan_planning.go
Revanth14: Addressed the UUID version issue.

Generated idempotency keys now use `uuid.NewV7()` and propagate generation errors instead of using `uuid.NewString()` / UUIDv4. I also tightened caller-provided keys to require UUIDv7, and updated the request docs/tests accordingly.

I left the “only send when the server advertises key lifetime” behavior as a follow-up, as suggested. This PR still intentionally sends an idempotency key on plan/tasks POSTs; the config-gated behavior can land with the catalog config plumbing.

catalog/rest/scan_planning.go
Revanth14: I’m deferring the granular `error.type` split out of this PR.

For the low-level client methods, `FetchPlanningResult` keeps the existing status-based 404 mapping to `ErrPlanExpired`, and the comment now calls out that finer plan/table/namespace classification should land with the polling layer that consumes the retry-vs-abort contract.

I’ll handle that in the `WaitForPlan` follow-up, where we can pin the exact `error.type` strings against the spec/OpenAPI and test the retry behavior directly.

catalog/rest/scan_planning.go
Revanth14: I’m deferring the granular `error.type` split here as well.

For this PR, `FetchScanTasks` keeps the low-level status-based mapping. The plan-task-vs-table distinction should land together with the polling/fanout layer that consumes it, so we can add the right `ErrNoSuchPlanTask` contract and test retry/abort behavior against the exact spec/OpenAPI `error.type` values.

For now this PR stays focused on wiring the low-level client method and preserving the existing REST status override style.

catalog/rest/scan_planning.go
Revanth14: Agreed, addressed by making terminal failed plans return errors instead of `(resp, nil)`.

`PlanTableScan` now returns a zero response plus `*PlanFailedError` when the server returns `status: failed`; the structured server error is available through `errors.As`. `FetchPlanningResult` follows the same pattern for failed results, and cancelled poll results now return `ErrPlanCancelled` so terminal non-success states do not slip past the `if err != nil` idiom.

The completed/submitted distinction still stays in the response as before.

catalog/rest/scan_planning.go
Revanth14: I kept `SupportsRemoteScanPlanning` as the end-to-end/full-endpoint predicate.

The reason is that this method is consumed by `table.Scan` auto mode as the gate for whether to route the scan to remote planning. Once auto mode chooses remote, it does not get a second chance to fall back to local if the server returns `submitted` or plan-task fanout and the follow-up endpoints are missing.

So `SupportsPlanTableScan()` remains the narrow “plan endpoint exists / sync-inline may work” check, while `SupportsRemoteScanPlanning()` keeps the table-side contract of “can complete remote planning end to end.” That avoids making auto mode fail on plan-only catalogs that could otherwise plan locally.

For the Java-style “can the server plan at all” signal, `SupportsPlanTableScan()` is the exposed predicate; auto mode intentionally routes on the stricter end-to-end check.

catalog/rest/rest.go
Revanth14: Went with the sturdier version here.

The nil-slice suppression path was removed entirely. Suppression is now carried as an explicit canonicalized set on the request context, and `RoundTrip` consults that set directly when applying session defaults. That removes the dependency on net/http’s nil-slice behavior and also removes the two-sided “keep this in sync” invariant.

I added direct coverage for the three default-header paths: apply default when absent, let a real per-request header override without duplication, and suppress a default via context.

catalog/rest/rest.go
Revanth14: Addressed.

`withHeaders` now merges into the existing header map instead of replacing it, and `withSuppressedHeaders` appends to the existing suppression list instead of replacing it. I also added a direct `TestReqOptionsCompose` test that exercises multiple `withHeaders` and multiple `withSuppressedHeaders` options on the same request config.

catalog/rest/rest.go
Revanth14: Addressed in the `RoundTrip` comment.

It now explicitly states that a session default is skipped when the request already carries the same header key, and that this applies to any per-request override, not just `Content-Type`. The old Content-Type special case is now covered by the general rule, and suppression is handled separately through the context-carried suppression set.

catalog/rest/scan_planning_test.go
Revanth14: Addressed.

`TestScanPlanningEndpointGating` now uses `t.Run` subtests for plan, fetch-result, cancel, and fetch-tasks, so each endpoint-gating assertion reports independently instead of the first `require` masking the rest.

catalog/rest/scan_planning_test.go
Revanth14: Addressed.

Added direct coverage that `FetchPlanningResultResponse.UnmarshalJSON` accepts `{"status":"cancelled"}` and sets `PlanStatusCancelled`.

Also added real client-path coverage for terminal poll results: `FetchPlanningResult` now has round-trip cases for both `cancelled` and `failed`. `cancelled` returns `ErrPlanCancelled`, and `failed` returns `*PlanFailedError`, so those terminal states are exercised through the HTTP client path rather than only direct JSON unmarshalling.

catalog/rest/rest_internal_test.go
Revanth14: Addressed.

Added direct `sessionTransport.RoundTrip` coverage in `TestRoundTripDefaultHeaderHandling`. It now verifies:
- a session default is applied when the request lacks the header
- a real per-request header overrides the session default without appending/duplicating it
- context-carried suppression drops the session default

So the generalized default-header behavior is pinned independently of the scan-planning request tests.

Revanth14:
Thanks for the detailed review @laskoviymishka @zeroshade 
Made the required changes, please take a look and let me know if any changes needed.

catalog/rest/scan_planning.go
```diff
@@ -79,30 +125,125 @@ func (r *Catalog) PlanFiles(ctx context.Context, req table.ScanPlanningRequest)
...
+	}
+
+	headers, err := scanPlanningHeaders(nil, opts.AccessDelegation, false)
+	if err != nil {
+		return FetchPlanningResultResponse{}, err
+	}
+
+	resp, err := doGet[FetchPlanningResultResponse](
+		ctx, r.baseURI, path, r.cl,
+		map[int]error{http.StatusNotFound: ErrPlanExpired}, withHeaders(headers))
```
zeroshade: This still makes every 404 from `FetchPlanningResult` indistinguishable from an expired plan, which conflicts with the documented follow-up contract at lines 41-45. A dropped table or namespace will now surface as `ErrPlanExpired` instead of `catalog.ErrNoSuchTable` / `catalog.ErrNoSuchNamespace`, so the polling layer cannot make the required distinction. Please decode the REST error model's `error.type` on 404 and map `NoSuchPlanIdException` to `ErrPlanExpired`, `NoSuchTableException` to `catalog.ErrNoSuchTable`, and `NoSuchNamespaceException` to `catalog.ErrNoSuchNamespace`, with httptest coverage for each.

catalog/rest/scan_planning.go
```diff
@@ -79,30 +125,125 @@ func (r *Catalog) PlanFiles(ctx context.Context, req table.ScanPlanningRequest)
...
+	}
+
+	headers, err := scanPlanningHeaders(req.IdempotencyKey, nil, true)
+	if err != nil {
+		return FetchScanTasksResponse{}, err
+	}
+
+	return doPost[FetchScanTasksRequest, FetchScanTasksResponse](
+		ctx, r.baseURI, path, req, r.cl,
+		map[int]error{http.StatusNotFound: catalog.ErrNoSuchTable},
```
zeroshade: This 404 override hides `NoSuchPlanTaskException` as `catalog.ErrNoSuchTable`. For the follow-up task-fetch layer to distinguish a missing table from an expired/unknown plan-task handle, please add a distinct `ErrNoSuchPlanTask` sentinel and map the 404 by REST `error.type` instead of status alone, with matching tests.

catalog/rest/scan_planning_test.go
```diff
@@ -94,3 +186,385 @@ func TestPlanTableScanResponseRejectsUnknownStatus(t *testing.T) {
...
+
+	cat := newScanPlanningTestCatalog(t, []endpoint{endpointFetchPlanResult}, func(mux *http.ServeMux) {
+		mux.HandleFunc("/v1/namespaces/db/tables/tbl/plan/expired-plan", func(w http.ResponseWriter, req *http.Request) {
+			require.Equal(t, http.MethodGet, req.Method)
+			w.WriteHeader(http.StatusNotFound)
+		})
+	})
+
+	_, err := cat.FetchPlanningResult(context.Background(), table.Identifier{"db", "tbl"}, "expired-plan", FetchPlanningResultOptions{})
+	require.ErrorIs(t, err, ErrPlanExpired)
```
zeroshade: This test only covers an empty-body 404 and therefore pins the lossy `ErrPlanExpired` behavior. Please switch these 404 cases to typed REST error bodies for plan-id/table/namespace and add corresponding `FetchScanTasks` 404 coverage, including the `NoSuchPlanTaskException` case.

zeroshade (changes requested):
Re-review: the prior header-suppression-via-context feedback is addressed, but the 404 error-type contract for the follow-up polling layer is still blocking.

- FetchPlanningResult currently collapses all 404s to ErrPlanExpired; it should decode REST error.type and distinguish NoSuchPlanIdException, NoSuchTableException, and NoSuchNamespaceException.
- FetchScanTasks currently collapses 404s to ErrNoSuchTable; it should expose a distinct NoSuchPlanTask sentinel for NoSuchPlanTaskException.
- The tests pin the lossy empty-body 404 behavior and do not cover FetchScanTasks 404s; add typed REST error-body coverage for the expected cases.

catalog/rest/scan_planning.go
laskoviymishka: This is the 404 split from last round, still here — and zeroshade re-flagged it too, so I'd treat it as the merge blocker for this pass.

The problem with mapping every 404 to `ErrPlanExpired` (and `PlanTableScan`/`FetchScanTasks` mapping theirs to `ErrNoSuchTable`) is that `error.type` never gets consulted, so a table-gone 404 is indistinguishable from an expired plan-id. `WaitForPlan` is the consumer, and once it's real it needs `errors.Is(err, ErrPlanExpired)` to mean "retry" and a table-gone to mean "abort" — with the collapse baked in here, the poll loop spins forever where Java and PyIceberg terminate. The comment defers this to the polling layer, but this method is the only place with the `error.type`, so the split has to happen here, not downstream.

zeroshade already laid out the full type enumeration on the other thread (`NoSuchPlanIdException` vs `NoSuchTableException`/`NoSuchNamespaceException`, plus the distinct `NoSuchPlanTask` sentinel for `/tasks`) — I'd follow that rather than restate it. Concretely, read `errorResponse.Type` in `handleNon200` and key the override on it: `NoSuchTableException`/`NoSuchNamespaceException` → `catalog.ErrNoSuchTable`, and keep `ErrPlanExpired` for `NoSuchPlanIdException`. Once that's in for all three call sites I'm happy to approve.

catalog/rest/scan_planning.go
```diff
@@ -180,8 +378,8 @@ type CompletedPlanningResult struct {
 // wire type and the seam stay in agreement.
 type PlanTableScanRequest struct {
 	// IdempotencyKey is sent as the Idempotency-Key header, not in the JSON body.
-	// If set, it must be a UUID string; nil lets the implementation choose a
-	// safe retry key.
+	// If set, it must be a UUIDv7 string (RFC 9562); nil lets the implementation
+	// generate a safe UUIDv7 retry key.
```
laskoviymishka: Calling this a "retry key" is a bit of a footgun given the implementation generates a fresh UUIDv7 on every invocation — so a caller who passes nil and retries on a transient error sends a different key each attempt, which silently defeats the server's dedup. And since the generated key isn't returned, they can't reuse it either.

I'd reword to something like "a new key is generated per call; pass an explicit key to get idempotency across retries." wdyt about also returning the generated key so callers can reuse it?

catalog/rest/rest.go
```diff
@@ -222,10 +228,17 @@ type sessionTransport struct {
...
+	// header (a per-request override of any default, not just Content-Type
+	// wins) or explicitly opted out of it via withSuppressedHeaders (carried on
+	// the context as an explicit set, never inferred from header values).
+	suppressed := suppressedHeadersFrom(r.Context())
 	for k, v := range s.defaultHeaders {
-		// Skip Content-Type if it's already set in the request
-		// to avoid duplicate headers (e.g., when using PostForm)
-		if http.CanonicalHeaderKey(k) == "Content-Type" && r.Header.Get("Content-Type") != "" {
+		ck := http.CanonicalHeaderKey(k)
+		if _, ok := r.Header[ck]; ok {
```
laskoviymishka, line 238: This is the right general semantics for the suppression rework, but widening the skip from Content-Type to all default headers quietly changes how Authorization behaves — a future `withHeaders` carrying an Authorization value would now suppress the session default. It's safe today only because `authManager` runs after this loop, which is a pretty fragile ordering dependency to leave implicit.

I'd add a one-line comment noting that ordering is load-bearing, plus a small regression test asserting `authManager.AuthHeader()` still wins over a `withHeaders`-supplied Authorization.

catalog/rest/scan_planning_test.go
```diff
@@ -94,3 +186,385 @@ func TestPlanTableScanResponseRejectsUnknownStatus(t *testing.T) {
...
+				_, err := w.Write([]byte(`{"status":"failed","error":{"message":"boom","type":"ServerError","code":500}}`))
+				require.NoError(t, err)
+			})
+		})
+
+		_, err := cat.FetchPlanningResult(context.Background(), table.Identifier{"db", "tbl"}, "plan-123", FetchPlanningResultOptions{})
+		require.ErrorIs(t, err, ErrPlanFailed)
+		var pfe *PlanFailedError
+		require.ErrorAs(t, err, &pfe)
+		assert.Equal(t, "boom", pfe.Detail.Message)
```
laskoviymishka, line 656: `pfe.Detail` is dereferenced here without a nil guard — if `Detail` were ever nil this panics and takes down the test process rather than failing cleanly. The `TestPlanTableScanRequest` sibling guards it with `require.NotNil(t, pfe.Detail)`; worth mirroring that here.

laskoviymishka (changes requested):
This is close now — every substantive thing from round 1 has landed. UUIDv7 generation plus version validation, the `PlanFailedError`/`ErrPlanCancelled` contract so a failed or cancelled plan can't be mistaken for an empty scan, `SupportsRemoteScanPlanning` gating on all four endpoints, the context-based access-delegation suppression with composing options, and the tests I asked for (cancelled/failed round-trips, the invalid-key rejection, endpoint gating). Thanks for working through all of it.

The one thing still open is the 404 error-type split I raised last round — and zeroshade re-flagged it on the 3rd, so I'd treat it as the blocker for this pass. `FetchPlanningResult` still collapses every 404 to `ErrPlanExpired`, and `PlanTableScan`/`FetchScanTasks` collapse theirs to `ErrNoSuchTable`, all on status code alone with `error.type` never consulted. The comment defers this to the polling layer, but this is the only place with the type info, and a poll loop treating `ErrPlanExpired` as retry-forever will spin indefinitely when the table's actually gone — Java and PyIceberg abort there. I left the concrete fix inline and pointed at zeroshade's type enumeration rather than restating it.

Two small new things while I was in here: the `IdempotencyKey` doc still calls it a "retry key" even though a fresh key is generated per call (so nil-then-retry silently defeats server dedup), and the RoundTrip change now lets a per-request header override any session default including a future Authorization — safe today only because authManager runs afterward, which is worth a comment and a regression test. Plus one test nit on an unguarded `pfe.Detail` deref.

Fix the 404 split and I'm happy to take another pass and approve.

catalog/rest/scan_planning.go
zeroshade: Mapping every 404 to `ErrPlanExpired` loses `NoSuchTableException`/`NoSuchNamespaceException` — a poller built on top can't distinguish retry-with-a-new-plan from table/namespace-gone-abort. Please split on the REST response `error.type` rather than collapsing all 404s here.

catalog/rest/scan_planning.go
zeroshade: Same issue here — 404 → `ErrNoSuchTable` hides `NoSuchPlanTaskException`; a task-handle expiry gets reported as a missing table, so callers can't tell retry-able from non-retry-able 404s.

zeroshade (changes requested):
The main outstanding item is typed error mapping for 404s (inline).

Revanth14:
Working on the requested changes, will update this PR once all changes are made.

catalog/rest/scan_planning.go
```diff
@@ -15,60 +15,106 @@
...
 
 // SupportsRemoteScanPlanning reports whether this catalog can complete a remote
-// plan end-to-end; backed by the split capability checks above.
+// plan end-to-end, including a `submitted` reply that must be polled. It requires
+// all four endpoints so table.Scan's auto mode only routes to the server when it
+// can finish without a fallback. Callers wanting the narrow "can plan inline"
+// signal should use SupportsPlanTableScan.
 func (r *Catalog) SupportsRemoteScanPlanning() bool {
-	return false
+	return r.SupportsFullRemoteScanPlanning()
```
zeroshade: Non-blocking: `SupportsRemoteScanPlanning` can now return true via `SupportsFullRemoteScanPlanning()`, while `PlanFiles` still returns `ErrNotImplemented` in this phase. If any auto-mode scanner routes on this capability before the follow-up phase lands, it would hit `ErrNotImplemented` rather than falling back to local planning. Since the phased split is intentional, it might be safer to keep this returning false (or add a guard + tracking issue) until `PlanFiles` is wired end-to-end.

zeroshade (commented):
Left one non-blocking note on the capability signal vs the still-stubbed `PlanFiles`.

catalog/rest/scan_planning.go
Revanth14: Addressed. `FetchPlanningResult` now inspects the REST response’s `error.type` and maps:

- `NoSuchPlanIdException` → `ErrPlanExpired`
- `NoSuchTableException` → `catalog.ErrNoSuchTable`
- `NoSuchNamespaceException` → `catalog.ErrNoSuchNamespace`

Bare or unrecognized 404s remain `ErrRESTError` rather than being guessed. Added `httptest` coverage for all mappings and fallback cases.

catalog/rest/scan_planning.go
Revanth14: Addressed. Added the distinct `ErrNoSuchPlanTask` sentinel, and `FetchScanTasks` now maps 404 responses using the REST `error.type`:

- `NoSuchPlanTaskException` → `ErrNoSuchPlanTask`
- `NoSuchTableException` → `catalog.ErrNoSuchTable`
- `NoSuchNamespaceException` → `catalog.ErrNoSuchNamespace`

Bare or unrecognized 404s remain `ErrRESTError`. Added matching `httptest` coverage for every mapping and fallback case.

catalog/rest/scan_planning_test.go
Revanth14: Addressed. Replaced the lossy empty-body-only coverage with typed REST error-body cases for:

- `FetchPlanningResult`: plan ID, table, and namespace not found
- `FetchScanTasks`: plan task, table, and namespace not found

Also retained bare and unrecognized 404 cases to verify they fall back to `ErrRESTError` without being misclassified.

Revanth14:
> Re-review: the prior header-suppression-via-context feedback is addressed, but the 404 error-type contract for the follow-up polling layer is still blocking.
> 
> * FetchPlanningResult currently collapses all 404s to ErrPlanExpired; it should decode REST error.type and distinguish NoSuchPlanIdException, NoSuchTableException, and NoSuchNamespaceException.
> * FetchScanTasks currently collapses 404s to ErrNoSuchTable; it should expose a distinct NoSuchPlanTask sentinel for NoSuchPlanTaskException.
> * The tests pin the lossy empty-body 404 behavior and do not cover FetchScanTasks 404s; add typed REST error-body coverage for the expected cases.

Thanks for the re-review. The blocking 404 error-type contract is now addressed:

- `FetchPlanningResult` distinguishes `NoSuchPlanIdException`, `NoSuchTableException`, and `NoSuchNamespaceException`.
- `FetchScanTasks` now exposes `ErrNoSuchPlanTask` for `NoSuchPlanTaskException` and separately maps missing tables and namespaces.
- Bare or unrecognized 404s conservatively remain `ErrRESTError`.
- Added typed REST error-body coverage for every expected mapping on both endpoints, plus bare and unrecognized fallback cases.

All related inline conversations have been resolved.

catalog/rest/scan_planning.go
Revanth14: Addressed this. `handleNon200` now supports mapping by REST `error.type`, and the typed override is applied at all three scan-planning call sites:

- `PlanTableScan`: distinguishes missing table and namespace.
- `FetchPlanningResult`: distinguishes expired plan ID, missing table, and missing namespace.
- `FetchScanTasks`: distinguishes missing plan task, table, and namespace using the new `ErrNoSuchPlanTask` sentinel.

Bare or unknown 404s remain `ErrRESTError`, so they are not incorrectly classified as retryable. Added coverage for every typed mapping and fallback case.

catalog/rest/scan_planning.go
Revanth14: Addressed the documentation footgun for both request types. It now states that `nil` generates a fresh UUIDv7 per call and that callers must provide an explicit key to preserve idempotency across retries.

I kept the generated key internal for this PR because returning it would require expanding the public method contract or mutating caller-owned request state. Callers that need retry dedup can generate one UUIDv7 before the first attempt and reuse it through `IdempotencyKey`; we can revisit a returned-key API separately if needed.

catalog/rest/rest.go
Revanth14: Addressed. Added a load-bearing ordering comment documenting that `authManager` must run after the default-header loop and overwrite any caller-supplied `Authorization` value.

Also added a regression test confirming that `authManager.AuthHeader()` wins over a per-request `Authorization` header.

catalog/rest/scan_planning_test.go
Revanth14: Addressed. Added `require.NotNil(t, pfe.Detail)` before dereferencing `pfe.Detail.Message`, matching the sibling test and ensuring a missing detail fails cleanly instead of panicking.

Revanth14:
> This is close now — every substantive thing from round 1 has landed. UUIDv7 generation plus version validation, the `PlanFailedError`/`ErrPlanCancelled` contract so a failed or cancelled plan can't be mistaken for an empty scan, `SupportsRemoteScanPlanning` gating on all four endpoints, the context-based access-delegation suppression with composing options, and the tests I asked for (cancelled/failed round-trips, the invalid-key rejection, endpoint gating). Thanks for working through all of it.
> 
> The one thing still open is the 404 error-type split I raised last round — and zeroshade re-flagged it on the 3rd, so I'd treat it as the blocker for this pass. `FetchPlanningResult` still collapses every 404 to `ErrPlanExpired`, and `PlanTableScan`/`FetchScanTasks` collapse theirs to `ErrNoSuchTable`, all on status code alone with `error.type` never consulted. The comment defers this to the polling layer, but this is the only place with the type info, and a poll loop treating `ErrPlanExpired` as retry-forever will spin indefinitely when the table's actually gone — Java and PyIceberg abort there. I left the concrete fix inline and pointed at zeroshade's type enumeration rather than restating it.
> 
> Two small new things while I was in here: the `IdempotencyKey` doc still calls it a "retry key" even though a fresh key is generated per call (so nil-then-retry silently defeats server dedup), and the RoundTrip change now lets a per-request header override any session default including a future Authorization — safe today only because authManager runs afterward, which is worth a comment and a regression test. Plus one test nit on an unguarded `pfe.Detail` deref.
> 
> Fix the 404 split and I'm happy to take another pass and approve.

Thanks for the detailed re-review. All remaining items are now addressed:

- Added REST `error.type` mapping for all three scan-planning call sites, distinguishing plan ID, plan task, table, and namespace 404s.
- Added `ErrNoSuchPlanTask`; bare and unknown 404s conservatively remain `ErrRESTError`.
- Added typed error-body tests for every expected mapping and fallback case.
- Reworded the idempotency-key documentation to clarify that `nil` generates a fresh key per call and explicit keys must be reused across retries.
- Documented the load-bearing `authManager` ordering and added a regression test confirming managed authorization wins.
- Added the missing `pfe.Detail` nil guard.
- Kept `SupportsRemoteScanPlanning` false while `PlanFiles` remains unimplemented.

catalog/rest/scan_planning.go
Revanth14: Addressed. `FetchPlanningResult` now splits 404 responses by REST `error.type`:

- `NoSuchPlanIdException` → `ErrPlanExpired`
- `NoSuchTableException` → `catalog.ErrNoSuchTable`
- `NoSuchNamespaceException` → `catalog.ErrNoSuchNamespace`

Bare or unrecognized 404s remain `ErrRESTError`, so the polling layer only retries when the server explicitly reports a missing plan ID. Added coverage for every mapping and fallback case.

catalog/rest/scan_planning.go
Revanth14: Addressed. Added the distinct `ErrNoSuchPlanTask` sentinel, and `FetchScanTasks` now splits 404s by REST `error.type`:

- `NoSuchPlanTaskException` → `ErrNoSuchPlanTask`
- `NoSuchTableException` → `catalog.ErrNoSuchTable`
- `NoSuchNamespaceException` → `catalog.ErrNoSuchNamespace`

Bare or unrecognized 404s remain `ErrRESTError`. Added tests for every mapping and fallback case.

catalog/rest/scan_planning.go
Revanth14: Addressed. `SupportsRemoteScanPlanning` now remains `false` while `PlanFiles` is unimplemented, ensuring auto mode falls back to local planning instead of reaching `ErrNotImplemented`.

`SupportsFullRemoteScanPlanning` remains available as the endpoint-capability check and will back `SupportsRemoteScanPlanning` once `PlanFiles` is wired end-to-end. Added coverage confirming the auto-mode signal stays disabled even when all four endpoints are advertised.

Revanth14:
Thanks @zeroshade and @laskoviymishka for the detailed reviews.

All requested changes have been addressed, including the typed 404 mappings, `ErrNoSuchPlanTask`, idempotency documentation, authorization-ordering regression coverage, capability gating, and test hardening.

Verification completed successfully:

- `go test -count=1 ./...`
- `go test -race -count=1 ./catalog/rest`
- `go vet ./catalog/rest .`

Please take a look when you have time.

laskoviymishka (approved):
Thanks @Revanth14, this is there now. 

The 404 error-type split I held on last round has landed: `handleNon200` takes a type-keyed override that refines the status-code mapping, so `FetchPlanningResult` distinguishes a forgotten plan-id (`ErrPlanExpired`) from a gone table/namespace, and `FetchScanTasks` gets its own `ErrNoSuchPlanTask` sentinel — with direct `handleNon200` coverage for the type-wins-over-status and unmatched-type-falls-through cases. The refinement being scoped to an already-mapped status is exactly right; a 503 carrying a plan exception still resolves as a 503.

The two smaller things are in too: the idempotency-key doc now says a fresh key is generated per call, and the RoundTrip Authorization ordering has its load-bearing comment and a regression test. Nice work across all three rounds. 

Overall LGTM, nothing of this blocking. Great job! :shipit:

zeroshade (approved):

PR #1326, opened 2026-06-27
fix(hadoop): use codec-aware metadata filenames

fallintoplace:
Summary

- write Hadoop metadata files with suffixes that match the configured metadata compression codec
- load the actual discovered metadata file path instead of reconstructing an uncompressed-looking path
- teach Hadoop metadata discovery to recognize plain, gzip, and zstd metadata files consistently

Why

Hadoop currently always writes metadata to vN.metadata.json even when write.metadata.compression-codec requests gzip or zstd. The table metadata reader infers decompression from the filename suffix, so compressed bytes behind a plain .metadata.json path can fail to reload.

Using codec-aware filenames on write and carrying the discovered metadata path through LoadTable keeps create, commit, and reload consistent for plain, gzip, and zstd metadata.

Testing

- go test ./catalog/hadoop -run 'TestHadoopCatalogTestSuite/(TestMetadataFilePathForCompression|TestVersionPatternMatches|TestIsTableDirWithZstdMetadata|TestFindVersionGzipOnlyWithHint|TestFindVersionMixedGzipAndPlain|TestFindVersionZstdOnlyWithHint|TestFindVersionMixedZstdAndPlain|TestCreateTableGzipMetadata|TestCreateTableZstdMetadata|TestCommitTableGzipMetadata|TestCommitTableZstdMetadata)$' -count=1
- go test ./... -run '^$' -count=1
- git diff --check

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -290,6 +290,20 @@ func (c *Catalog) metadataFilePath(ident table.Identifier, version int) string {
 	return joinPath(c.isLocal, c.metadataDir(ident), fmt.Sprintf("v%d.metadata.json", version))
 }
 
+func (c *Catalog) metadataFilePathForCompression(ident table.Identifier, version int, compression string) (string, error) {
+	var suffix string
+	switch compression {
+	case table.MetadataCompressionCodecNone:
+		suffix = ".metadata.json"
+	case table.MetadataCompressionCodecGzip:
+		suffix = ".gz.metadata.json"
+	default:
+		return "", fmt.Errorf("unsupported write metadata compression codec: %s", compression)
+	}
+
+	return joinPath(c.isLocal, c.metadataDir(ident), fmt.Sprintf("v%d%s", version, suffix)), nil
+}
+
 func (c *Catalog) versionHintPath(ident table.Identifier) string {
 	return joinPath(c.isLocal, c.metadataDir(ident), "version-hint.text")
 }
@@ -525,15 +539,19 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 	}
 
 	version := 1
-	metaPath := c.metadataFilePath(ident, version)
-	tempPath := joinPath(c.isLocal, metaDir, uuid.New().String()+".metadata.json")
-
 	compression := table.MetadataCompressionDefault
 	if cfg.Properties != nil {
 		if v, ok := cfg.Properties[table.MetadataCompressionKey]; ok {
 			compression = v
 		}
 	}
+
+	metaPath, err := c.metadataFilePathForCompression(ident, version, compression)
+	if err != nil {
+		return nil, fmt.Errorf("hadoop catalog: failed to determine metadata file path: %w", err)
+	}
+
+	tempPath := joinPath(c.isLocal, metaDir, uuid.New().String()+".metadata.json")
 
 	if err := internal.WriteTableMetadata(metadata, c.filesystem, tempPath, compression); err != nil {
 		_ = c.filesystem.Remove(tempPath)
@@ -655,10 +673,13 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs 
 		return nil, "", fmt.Errorf("hadoop catalog: failed to create metadata directory: %w", err)
 	}
 
-	newMetaPath := c.metadataFilePath(ident, newVersion)
-	tempPath := joinPath(c.isLocal, metaDir, uuid.New().String()+".metadata.json")
-
 	compression := updated.Properties().Get(table.MetadataCompressionKey, table.MetadataCompressionDefault)
+	newMetaPath, err := c.metadataFilePathForCompression(ident, newVersion, compression)
+	if err != nil {
+		return nil, "", fmt.Errorf("hadoop catalog: failed to determine metadata file path: %w", err)
+	}
+
+	tempPath := joinPath(c.isLocal, metaDir, uuid.New().String()+".metadata.json")
 
 	if err := internal.WriteTableMetadata(updated, c.filesystem, tempPath, compression); err != nil {
 		_ = c.filesystem.Remove(tempPath)
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -300,6 +300,10 @@ func (c *Catalog) metadataFilePathForCompression(ident table.Identifier, version
 	return joinPath(c.isLocal, c.metadataDir(ident), fmt.Sprintf("v%d%s", version, suffix)), nil
 }
 
+func (c *Catalog) metadataVersionClaimPath(ident table.Identifier, version int) string {
+	return joinPath(c.isLocal, c.metadataDir(ident), fmt.Sprintf("v%d.metadata.json.claim", version))
+}
+
 func (c *Catalog) versionHintPath(ident table.Identifier) string {
 	return joinPath(c.isLocal, c.metadataDir(ident), "version-hint.text")
 }
@@ -460,6 +464,16 @@ func (c *Catalog) metadataVersionExists(ident table.Identifier, version int) (bo
 	return found, nil
 }
 
+func (c *Catalog) metadataVersionLocation(ident table.Identifier, version int) (string, bool) {
+	files, _, err := c.scanMetadataFiles(ident)
+	if err != nil {
+		return "", false
+	}
+
+	file, ok := files[version]
+	return file.location, ok
+}
+
 func (c *Catalog) findMetadataLocation(ident table.Identifier) (string, int, error) {
 	_, latest, err := c.scanMetadataFiles(ident)
 	if err != nil {
@@ -555,7 +569,7 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 		return nil, fmt.Errorf("hadoop catalog: failed to write table metadata: %w", err)
 	}
 
-	if err := c.commitMetadataFile(ident, tempPath, metaPath, catalog.ErrTableAlreadyExists); err != nil {
+	if err := c.commitMetadataFile(ident, version, tempPath, metaPath, catalog.ErrTableAlreadyExists); err != nil {
 		return nil, err
 	}
 
@@ -683,20 +697,7 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs 
 		return nil, "", fmt.Errorf("hadoop catalog: failed to write table metadata: %w", err)
 	}
 
-	exists, err := c.metadataVersionExists(ident, newVersion)
-	if err != nil {
-		_ = c.filesystem.Remove(tempPath)
-
-		return nil, "", fmt.Errorf("hadoop catalog: failed to inspect metadata directory for version %d: %w",
-			newVersion, err)
-	}
-	if exists {
-		_ = c.filesystem.Remove(tempPath)
-
-		return nil, "", fmt.Errorf("hadoop catalog: version %d already exists for table %s",
-			newVersion, strings.Join(ident, "."))
-	}
-	if err := c.commitMetadataFile(ident, tempPath, newMetaPath, table.ErrCommitFailed); err != nil {
+	if err := c.commitMetadataFile(ident, newVersion, tempPath, newMetaPath, table.ErrCommitFailed); err != nil {
 		return nil, "", err
 	}
 
@@ -706,10 +707,32 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs 
 	return updated, newMetaPath, nil
 }
 
-func (c *Catalog) commitMetadataFile(ident table.Identifier, tempPath, metaPath string, conflictErr error) error {
-	if err := c.filesystem.RenameNoReplace(tempPath, metaPath); err != nil {
+func (c *Catalog) commitMetadataFile(ident table.Identifier, version int, tempPath, metaPath string, conflictErr error) error {
+	claimPath := c.metadataVersionClaimPath(ident, version)
+	if err := c.filesystem.RenameNoReplace(tempPath, claimPath); err != nil {
 		_ = c.filesystem.Remove(tempPath)
 
+		if errors.Is(err, fs.ErrExist) {
+			return fmt.Errorf("%w: metadata version already claimed for table %s: %s",
+				conflictErr, strings.Join(ident, "."), claimPath)
+		}
+
+		return fmt.Errorf("hadoop catalog: failed to claim metadata version: %w", err)
+	}
+
+	removeClaim := true
+	defer func() {
+		if removeClaim {
+			_ = c.filesystem.Remove(claimPath)
+		}
+	}()
+
+	if existingPath, exists := c.metadataVersionLocation(ident, version); exists {
+		return fmt.Errorf("%w: metadata file already exists for table %s: %s",
+			conflictErr, strings.Join(ident, "."), existingPath)
+	}
+
+	if err := c.filesystem.RenameNoReplace(claimPath, metaPath); err != nil {
 		if errors.Is(err, fs.ErrExist) {
 			return fmt.Errorf("%w: metadata file already exists for table %s: %s",
 				conflictErr, strings.Join(ident, "."), metaPath)
@@ -717,6 +740,8 @@ func (c *Catalog) commitMetadataFile(ident table.Identifier, tempPath, metaPath 
 
 		return fmt.Errorf("hadoop catalog: failed to commit metadata file: %w", err)
 	}
+
+	removeClaim = false
 
 	return nil
 }
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -471,6 +471,7 @@ func (c *Catalog) metadataVersionLocation(ident table.Identifier, version int) (
 	}
 
 	file, ok := files[version]
+
 	return file.location, ok
 }
 
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -30,6 +30,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/apache/iceberg-go"
 	"github.com/apache/iceberg-go/catalog"
@@ -62,6 +63,8 @@ var versionPattern = regexp.MustCompile(`^v([0-9]+)(?:\.gz)?\.metadata\.json$`)
 var uuidMetadataPattern = regexp.MustCompile(
 	`^([0-9]{5})-[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(?:\.gz)?\.metadata\.json$`,
 )
+
+const metadataClaimStaleAfter = time.Minute
 
 type metadataFile struct {
 	location   string
@@ -710,15 +713,45 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs 
 
 func (c *Catalog) commitMetadataFile(ident table.Identifier, version int, tempPath, metaPath string, conflictErr error) error {
 	claimPath := c.metadataVersionClaimPath(ident, version)
-	if err := c.filesystem.RenameNoReplace(tempPath, claimPath); err != nil {
-		_ = c.filesystem.Remove(tempPath)
+	for {
+		if err := c.filesystem.RenameNoReplace(tempPath, claimPath); err != nil {
+			if !errors.Is(err, fs.ErrExist) {
+				_ = c.filesystem.Remove(tempPath)
 
-		if errors.Is(err, fs.ErrExist) {
-			return fmt.Errorf("%w: metadata version already claimed for table %s: %s",
-				conflictErr, strings.Join(ident, "."), claimPath)
+				return fmt.Errorf("hadoop catalog: failed to claim metadata version: %w", err)
+			}
+
+			existingPath, exists := c.metadataVersionLocation(ident, version)
+			if exists {
+				_ = c.filesystem.Remove(tempPath)
+
+				return fmt.Errorf("%w: metadata file already exists for table %s: %s",
+					conflictErr, strings.Join(ident, "."), existingPath)
+			}
+
+			claimInfo, err := c.filesystem.Stat(claimPath)
+			if err != nil {
+				_ = c.filesystem.Remove(tempPath)
+
+				return fmt.Errorf("hadoop catalog: failed to inspect stale metadata claim %s: %w", claimPath, err)
+			}
+			if time.Since(claimInfo.ModTime()) < metadataClaimStaleAfter {
+				_ = c.filesystem.Remove(tempPath)
+
+				return fmt.Errorf("%w: metadata version already claimed for table %s: %s",
+					conflictErr, strings.Join(ident, "."), claimPath)
+			}
+
+			if err := c.filesystem.Remove(claimPath); err != nil && !errors.Is(err, fs.ErrNotExist) {
+				_ = c.filesystem.Remove(tempPath)
+
+				return fmt.Errorf("hadoop catalog: failed to clear stale metadata claim %s: %w", claimPath, err)
+			}
+
+			continue
 		}
 
-		return fmt.Errorf("hadoop catalog: failed to claim metadata version: %w", err)
+		break
 	}
 
 	removeClaim := true
@@ -737,6 +770,16 @@ func (c *Catalog) commitMetadataFile(ident table.Identifier, version int, tempPa
 		if errors.Is(err, fs.ErrExist) {
 			return fmt.Errorf("%w: metadata file already exists for table %s: %s",
 				conflictErr, strings.Join(ident, "."), metaPath)
+		}
+		if errors.Is(err, fs.ErrNotExist) {
+			existingPath, exists := c.metadataVersionLocation(ident, version)
+			if exists {
+				return fmt.Errorf("%w: metadata file already exists for table %s: %s",
+					conflictErr, strings.Join(ident, "."), existingPath)
+			}
+
+			return fmt.Errorf("%w: metadata claim for table %s and version %d was removed before publish",
+				conflictErr, strings.Join(ident, "."), version)
 		}
 
 		return fmt.Errorf("hadoop catalog: failed to commit metadata file: %w", err)
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -286,10 +286,6 @@ func (c *Catalog) metadataDir(ident table.Identifier) string {
 	return joinPath(c.isLocal, c.tableToPath(ident), "metadata")
 }
 
-func (c *Catalog) metadataFilePath(ident table.Identifier, version int) string {
-	return joinPath(c.isLocal, c.metadataDir(ident), fmt.Sprintf("v%d.metadata.json", version))
-}
-
 func (c *Catalog) metadataFilePathForCompression(ident table.Identifier, version int, compression string) (string, error) {
 	var suffix string
 	switch compression {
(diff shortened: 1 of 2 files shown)


catalog/hadoop/hadoop.go
tanmayrauth: After this change `metadataFilePath` no longer has a production caller, CreateTable/CommitTable/LoadTable all go through metadataFilePathForCompression /  findMetadataLocation now. It's still referenced from the tests so it's not dead, but might be worth either keeping it explicitly as a test-only helper or pointing those tests at the new helper to avoid drift.

tanmayrauth (approved):
Looks good — write suffix now matches the read-path codec inference in table.go, and the create→load / commit→load round-trip tests cover both gzip and zstd. One tiny nit.

catalog/hadoop/hadoop.go
fallintoplace: This seems to be a valid comment, I will amend the PR. Have a nice weekend.

catalog/hadoop/hadoop.go
```diff
@@ -184,8 +186,20 @@ func (c *Catalog) metadataDir(ident table.Identifier) string {
...
-	return filepath.Join(c.metadataDir(ident), fmt.Sprintf("v%d.metadata.json", version))
+func (c *Catalog) metadataFilePathForCompression(ident table.Identifier, version int, compression string) (string, error) {
+	var suffix string
+	switch compression {
+	case table.MetadataCompressionCodecNone:
+		suffix = ".metadata.json"
+	case table.MetadataCompressionCodecGzip:
+		suffix = ".gz.metadata.json"
+	case table.MetadataCompressionCodecZstd:
+		suffix = ".zstd.metadata.json"
```
laskoviymishka: I think the zstd branch here is a cross-client trap. `zstd` isn't a value Java's `TableMetadataParser.Codec` understands — its enum is only `NONE("")` and `GZIP(".gz")`, and `HadoopTableOperations.getMetadataFile` only ever probes `Codec.values()`, so a `vN.zstd.metadata.json` is never even looked for. PyIceberg falls back to `NOOP_COMPRESSOR` for any non-`.gz` suffix and reads the compressed bytes as garbage, and iceberg-rust rejects zstd outright in `parse_metadata_file_compression`.

So a Hadoop-catalog table we write with zstd is invisible-or-corrupt to every other Iceberg client, and the Hadoop catalog is precisely the shared-filesystem case where that interop matters most.

I'd gate zstd at this layer for the Hadoop catalog — return the same "unsupported codec" error you already produce for `brotli`, so the failure is loud at write time rather than a silent unreadable file. If the intent is genuinely a go-to-go-only extension, that needs a prominent warning in the PR description and CHANGES.md and a clear note in the code that this is non-portable. wdyt?

catalog/hadoop/hadoop.go
```diff
@@ -509,10 +550,13 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
 		return nil, "", fmt.Errorf("hadoop catalog: failed to create metadata directory: %w", err)
 	}
 
-	newMetaPath := c.metadataFilePath(ident, newVersion)
-	tempPath := filepath.Join(metaDir, uuid.New().String()+".metadata.json")
-
 	compression := updated.Properties().Get(table.MetadataCompressionKey, table.MetadataCompressionDefault)
+	newMetaPath, err := c.metadataFilePathForCompression(ident, newVersion, compression)
```
laskoviymishka, line 661: The conflict guard below only stats the codec-specific path for `newVersion`, so cross-codec races slip through. If writer A plans `v2.gz.metadata.json` and writer B plans `v2.metadata.json` concurrently, neither `Stat` sees the other and both commit — two different files both claiming version 2, breaking the monotonic chain.

Java's `getMetadataFile` iterates every codec variant, so its conflict detection is codec-agnostic by construction.

I'd make the guard probe all variants: call `metadataVersionLocation(ident, newVersion)` and fail the commit if it finds any file for that version, regardless of suffix. That also covers the migration case where someone switches codecs between commits.

catalog/hadoop/hadoop.go
```diff
@@ -330,16 +355,30 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
...
 			strings.Join(ident, "."), catalog.ErrNoSuchTable)
 	}
 
 	if maxVer == 0 {
-		return 0, fmt.Errorf("hadoop catalog: no metadata files found for table %s: %w",
+		return "", 0, fmt.Errorf("hadoop catalog: no metadata files found for table %s: %w",
+			strings.Join(ident, "."), catalog.ErrNoSuchTable)
+	}
+
+	metaPath, ok := c.metadataVersionLocation(ident, maxVer)
```
laskoviymishka: The `WalkDir` above already saw the exact filename for `maxVer` (it matched `versionPattern` to compute the version), but we throw the name away and only keep the integer — then re-probe here with the fixed plain→gz→zstd order.

That re-probe is where stale data wins: if both `v5.metadata.json` (old) and `v5.gz.metadata.json` (new) exist, plain is probed first and silently shadows the compressed current version, so the catalog serves stale uncompressed metadata even though the authoritative version is the gzip one. A compression migration would be quietly defeated.

I'd capture the matched filename in the `WalkDir` callback (it's the `path` param) when `v > maxVer`, and thread that straight into `scanForward` so we never re-probe a file we just saw. That removes the shadowing and drops the extra `Stat`s per lookup on the slow path.

catalog/hadoop/hadoop_test.go
```diff
@@ -392,6 +416,33 @@ func (s *HadoopCatalogTestSuite) TestFindVersionMixedGzipAndPlain() {
...
+	ident := []string{"ns", "tbl"}
+	dir := s.cat.metadataDir(ident)
+	s.Require().NoError(os.MkdirAll(dir, 0o755))
+	s.Require().NoError(os.WriteFile(filepath.Join(dir, "v1.zstd.metadata.json"), nil, 0o644))
+	s.Require().NoError(os.WriteFile(filepath.Join(dir, "v2.zstd.metadata.json"), nil, 0o644))
+	s.cat.writeVersionHint(ident, 1)
+
+	ver, err := s.cat.findVersion(ident)
+	s.Require().NoError(err)
+	s.Equal(2, ver)
```
laskoviymishka: `TestFindVersionZstdOnlyWithHint` and `TestFindVersionMixedZstdAndPlain` only assert the version number, never the returned path — so the part of the fix that actually matters (that `findMetadataLocation` returns the codec-suffixed path, not a reconstructed plain one) isn't directly exercised here. The empty `nil` files are fine since these call only `Stat`, but the path is the thing the LoadTable fix hinges on.

I'd add a direct `findMetadataLocation` assertion that the returned path ends in `.zstd.metadata.json` for the zstd-only case and `.gz.metadata.json` for a gzip-only one. That's also the natural place to add the shadowing case from the comment on `findMetadataLocation` — write both `v3.metadata.json` and `v3.gz.metadata.json` and assert which path wins, so the resolution is pinned by a test.

catalog/hadoop/hadoop_test.go
```diff
@@ -129,12 +129,24 @@ func (s *HadoopCatalogTestSuite) TestMetadataDir() {
...
 
-	path = s.cat.metadataFilePath([]string{"ns", "tbl"}, 42)
-	s.Equal(filepath.Join(s.warehouse, "ns", "tbl", "metadata", "v42.metadata.json"), path)
+	path, err = s.cat.metadataFilePathForCompression(ident, 1, table.MetadataCompressionCodecGzip)
+	s.Require().NoError(err)
+	s.Equal(filepath.Join(s.warehouse, "ns", "tbl", "metadata", "v1.gz.metadata.json"), path)
+
+	path, err = s.cat.metadataFilePathForCompression(ident, 1, table.MetadataCompressionCodecZstd)
+	s.Require().NoError(err)
+	s.Equal(filepath.Join(s.warehouse, "ns", "tbl", "metadata", "v1.zstd.metadata.json"), path)
```
laskoviymishka: The old `TestMetadataFilePath` covered version 1 and version 42; the rewrite only keeps version 1 across the three codecs. The multi-digit case is the normal path for any table past v9, and `fmt.Sprintf("v%d%s", ...)` is exactly where a width/format slip would hide.

I'd carry a `version=42` assertion over for at least one codec while we're here.

catalog/hadoop/hadoop.go
```diff
@@ -51,15 +51,17 @@ func init() {
...
 // uuidMetadataPattern matches UUID-style metadata filenames produced by
-// Java/PyIceberg catalogs: 00000-<uuid>.metadata.json or
-// 00000-<uuid>.gz.metadata.json. The sequence is a 5-digit zero-padded
+// Java/PyIceberg catalogs: 00000-<uuid>.metadata.json,
+// 00000-<uuid>.gz.metadata.json, or 00000-<uuid>.zstd.metadata.json.
+// The sequence is a 5-digit zero-padded
 // number and the UUID is in canonical 8-4-4-4-12 hex format.
 var uuidMetadataPattern = regexp.MustCompile(
-	`^[0-9]{5}-[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(?:\.gz)?\.metadata\.json$`,
+	`^[0-9]{5}-[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(?:\.(?:gz|zstd))?\.metadata\.json$`,
```
laskoviymishka: Two small things here. The `// Java/PyIceberg catalogs` comment above is now inaccurate for the zstd alternative — neither Java nor PyIceberg ever emits `00000-<uuid>.zstd.metadata.json`, so I'd note the zstd part is an iceberg-go-only extension rather than implying parity.

And recognizing UUID-style zstd here (used by `isTableDir`/`ListTables`) while `findMetadataLocation`'s `WalkDir` still keys off `versionPattern` only creates a confusing split: a user could create a table in Python, see it in `ListTables`, then get `ErrNoSuchTable` on `LoadTable`. That load gap is pre-existing, but I'd at least not widen the recognition surface without a `// recognized but not loadable` note. The new zstd UUID branch also has no test backing it.

catalog/hadoop/hadoop.go
```diff
@@ -270,36 +284,47 @@ func (c *Catalog) writeVersionHint(ident table.Identifier, version int) {
...
 	dir := c.metadataDir(ident)
-	plain := filepath.Join(dir, fmt.Sprintf("v%d.metadata.json", version))
-
-	if _, err := c.filesystem.Stat(plain); err == nil {
-		return true
+	for _, path := range []string{
+		filepath.Join(dir, fmt.Sprintf("v%d.metadata.json", version)),
+		filepath.Join(dir, fmt.Sprintf("v%d.gz.metadata.json", version)),
+		filepath.Join(dir, fmt.Sprintf("v%d.zstd.metadata.json", version)),
+	} {
```
laskoviymishka: Worth a one-line comment that any non-`ErrNotExist` `Stat` error (permission denied, transient FS error) is treated as "not found" here. It's the same behavior as the old `metadataVersionExists`, so not a regression, but on a slow filesystem a transient error could make a real version look absent — calling it out keeps the next reader from assuming it's only an existence check.

catalog/hadoop/hadoop_test.go
```diff
@@ -990,6 +1041,50 @@ func (s *HadoopCatalogTestSuite) TestCreateTableAndLoad() {
 	s.Equal(len(created.Schema().Fields()), len(loaded.Schema().Fields()))
 }
 
+func (s *HadoopCatalogTestSuite) TestCreateTableGzipMetadata() {
```
laskoviymishka, line 1485: These four new integration tests (`TestCreateTableGzipMetadata`, `TestCreateTableZstdMetadata`, `TestCommitTableGzipMetadata`, `TestCommitTableZstdMetadata`) repeat the same `os.Mkdir` + props map + `CreateTable` + `FileExists` boilerplate, where the existing suite leans on a `createTestTable` helper.

I'd fold the four into a `createTestTableWithCompression` helper or table-driven subtests to match the rest of the suite, and while you're adding zstd coverage, a `TestIsTableDirTrueUUIDMetadataZstd` to mirror the existing UUID+gzip case would catch a `ztsd`-style typo in `uuidMetadataPattern`. Low priority next to the blockers above.

zeroshade:
This LGTM on review, but it currently has merge conflicts with `main` (the metadata-discovery code moved). Could you rebase onto the latest `main`? Once it's clean I'll get it merged. Thanks @fallintoplace!

laskoviymishka (changes requested):
Nice catch on the underlying bug — `CreateTable`/`CommitTable` were writing to `vN.metadata.json` regardless of the codec, so a gzip file behind a plain filename would fail to parse on the next `LoadTable`, and routing the discovered path from `findMetadataLocation` straight into `NewFromLocation` is the right fix. The gzip side of this is solid.

I'd hold it before merging though. The zstd half goes beyond what the ecosystem supports: Java's `Codec` enum is only `NONE`/`GZIP` and never probes for `.zstd`, PyIceberg falls back to `NOOP_COMPRESSOR` and reads the bytes as garbage, and iceberg-rust rejects zstd outright. So a Hadoop-catalog table we write with zstd is unreadable by every other client — and the Hadoop catalog is exactly the shared-filesystem case where that matters. I'd gate zstd at this layer (same error you already raise for `brotli`) unless the intent is an explicit go-only extension, in which case it needs a loud warning in the PR description and CHANGES.md.

The other blocker is independent of zstd: on commit we only stat the new codec's own filename, and on load we re-probe with a fixed plain→gz→zstd order even though `WalkDir` already saw the real filename. Both let a stale plain `vN` shadow a newer compressed `vN`, which silently defeats a compression migration and can produce two different files at the same version under a concurrent cross-codec write.

- gate (or loudly document) zstd for the Hadoop catalog
- make the commit conflict check probe all codec variants for the new version
- thread the filename `WalkDir` already found into `scanForward` instead of re-probing, and add a test pinning which file wins when `vN.metadata.json` and `vN.gz.metadata.json` both exist
- assert the returned path (not just the version) in the gzip/zstd `findVersion` tests

catalog/hadoop/hadoop.go
```diff
@@ -364,22 +386,31 @@ func (c *Catalog) findVersion(ident table.Identifier) (int, error) {
 			v, _ := strconv.Atoi(matches[1])
 			if v > maxVer {
 				maxVer = v
+				maxPath = path
```
laskoviymishka: `maxPath` is set to whichever variant `WalkDir` visits first, so when both `v3.metadata.json` and `v3.gz.metadata.json` exist for a version the loaded file depends on listing order. On local FS `.gz` sorts ahead of `.metadata` so it's stable and the test passes, but on a remote/NFS listing it's arbitrary — and it disagrees with the hint path, which goes through `metadataVersionLocation` and always prefers plain. Same table can then resolve to two different metadata files depending on which discovery branch runs.

I'd drop `maxPath` and call `metadataVersionLocation(ident, maxVer)` after the walk so both discovery paths land on the same file. wdyt?

laskoviymishka (changes requested):
The fix itself is right — the gzip round-trip tests exercise the previously-broken write-then-reload path with real compressed bytes and assert UUID equality, so I'm confident the reported bug (compressed bytes behind a plain filename) is closed. The zstd hard-reject on write is a nice improvement over silently writing something that fails on read.

One thing I'd fix before this lands: `findMetadataLocation` returns `maxPath` straight from the `WalkDir` callback, so when both `v{n}.metadata.json` and `v{n}.gz.metadata.json` exist for a version, the file we load depends on listing order. On local FS the `.gz` variant happens to sort first so the test passes, but on a remote/NFS listing it's arbitrary — and it disagrees with the hint path, which resolves through `metadataVersionLocation` and deterministically prefers plain. Same table, two discovery branches, potentially two different files. Canonicalizing on `metadataVersionLocation(ident, maxVer)` after the walk closes that.

Left the detail inline. Fix that and I think this is good to land.

catalog/hadoop/hadoop.go
```diff
@@ -547,17 +584,27 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
...
+	tempPath := joinPath(c.isLocal, metaDir, uuid.New().String()+".metadata.json")
 
 	if err := internal.WriteTableMetadata(updated, c.filesystem, tempPath, compression); err != nil {
 		_ = c.filesystem.Remove(tempPath)
 
 		return nil, "", fmt.Errorf("hadoop catalog: failed to write table metadata: %w", err)
 	}
 
+	if existingPath, exists := c.metadataVersionLocation(ident, newVersion); exists {
+		_ = c.filesystem.Remove(tempPath)
```
zeroshade: [MAJOR] The cross-codec commit conflict check is non-atomic. Two concurrent writers can both compute the same newVersion, both pass metadataVersionLocation(ident, newVersion) before either publishes, then one RenameNoReplace can publish vN.metadata.json while the other publishes vN.gz.metadata.json; both succeed, leaving two files claiming the same version. Fix by taking an atomic per-version claim independent of codec suffix before publishing (for example a RenameNoReplace sentinel for vN), and add a concurrent mixed-codec commit test where both writers pass preflight before either final rename.

catalog/hadoop/hadoop.go
```diff
@@ -424,16 +459,20 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
...
-	tempPath := joinPath(c.isLocal, metaDir, uuid.New().String()+".metadata.json")
-
 	compression := table.MetadataCompressionDefault
 	if cfg.Properties != nil {
 		if v, ok := cfg.Properties[table.MetadataCompressionKey]; ok {
 			compression = v
 		}
 	}
 
+	metaPath, err := c.metadataFilePathForCompression(ident, version, compression)
```
zeroshade, line 533: [MAJOR] CreateTable has the same duplicate-version race with different compression. Two concurrent creators can both pass isTableDir, then publish v1.metadata.json and v1.gz.metadata.json successfully. Reuse a codec-independent per-version/table-creation claim before publishing metadata, and add a concurrent create test with default and gzip writers.

zeroshade (changes requested):
Remaining blockers are concurrency races in the new cross-codec Hadoop metadata paths. laskoviymishka's zstd-rejection and path-canonicalization threads appear addressed; items below are the remaining blockers (concurrency).

- [MAJOR] CommitTable cross-codec version conflict check is non-atomic; add a codec-independent per-version claim and mixed-codec concurrent commit test.
- [MAJOR] CreateTable has the same default/gzip duplicate-v1 race; reuse the claim and add a concurrent create test.

catalog/hadoop/hadoop.go
```diff
@@ -568,10 +612,32 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
 	return updated, newMetaPath, nil
 }
 
-func (c *Catalog) commitMetadataFile(ident table.Identifier, tempPath, metaPath string, conflictErr error) error {
-	if err := c.filesystem.RenameNoReplace(tempPath, metaPath); err != nil {
+func (c *Catalog) commitMetadataFile(ident table.Identifier, version int, tempPath, metaPath string, conflictErr error) error {
+	claimPath := c.metadataVersionClaimPath(ident, version)
+	if err := c.filesystem.RenameNoReplace(tempPath, claimPath); err != nil {
```
zeroshade: [MAJOR] A stale `.claim` file can permanently block this metadata version. The claim is created here with `RenameNoReplace(tempPath, claimPath)`, then is only removed by the deferred cleanup when `removeClaim` remains true or intentionally retained after the final rename. If the process dies after taking the claim but before publishing the final metadata file, `vN.metadata.json.claim` remains with no final `vN.*.metadata.json`, so every future attempt for that version hits `fs.ErrExist` and returns a conflict even though `metadataVersionLocation` finds no actual metadata file. Please make stale claims recoverable: when `claimPath` already exists, check whether any final metadata variant exists via `metadataVersionLocation`; if none exists, treat the claim as abandoned by reclaiming/removing it, or encode enough information to distinguish in-progress vs abandoned. Add a regression test that leaves a `vN.metadata.json.claim` with no `vN.*.metadata.json` and verifies recovery.

zeroshade (changes requested):
Thanks for the update — the two earlier concurrency concerns are addressed now: the codec-independent vN.metadata.json.claim path closes the mixed-codec version race, and the new concurrent mixed-codec create/commit tests cover the real scenarios. Nice work.

I found one new follow-up around crash recovery for retained claim files that should be addressed before merge.

zeroshade:
Following up after the latest commit: it only addresses the nlreturn lint, so my earlier request-changes still stands. The stale-claim recovery gap is unaddressed — if the process dies between taking the vN.metadata.json.claim and the final publish, that claim permanently blocks the version because the fs.ErrExist branch never checks whether a final metadata file actually exists. Please add the recovery check (on fs.ErrExist, call metadataVersionLocation and reclaim or remove an abandoned claim when no final metadata exists) plus a regression test that leaves a claim without final metadata.

catalog/hadoop/hadoop.go
```diff
@@ -450,6 +467,17 @@ func (c *Catalog) metadataVersionExists(ident table.Identifier, version int) (bo
 	return found, nil
 }
 
+func (c *Catalog) metadataVersionLocation(ident table.Identifier, version int) (string, bool) {
+	files, _, err := c.scanMetadataFiles(ident)
+	if err != nil {
+		return "", false
```
laskoviymishka: This is the path canonicalization from last round and it reads well — but the helper it's built on drops its error: any `scanMetadataFiles` failure returns `("", false)`, which is indistinguishable from "this version doesn't exist."

That's load-bearing at the post-claim call to this in `commitMetadataFile` — it's the only cross-codec conflict guard left. If `WalkDir` hits a transient error there, we get `false`, fall through to `RenameNoReplace` onto the plain filename (which succeeds because that path doesn't exist), and end up with `v2.metadata.json` and `v2.gz.metadata.json` both at the same version. `betterThan` then prefers the plain file and the compressed commit is silently gone.

I'd give it an error return and fail the commit on it rather than folding it into absence:

```go
func (c *Catalog) metadataVersionLocation(ident table.Identifier, version int) (string, bool, error) {
	files, _, err := c.scanMetadataFiles(ident)
	if err != nil {
		return "", false, err
	}
	file, ok := files[version]
	return file.location, ok, nil
}
```

Then propagate it at the call site as an infra error (not wrapped in `conflictErr`), so a scan failure aborts the commit instead of publishing over a live version.

catalog/hadoop/hadoop.go
```diff
@@ -689,18 +711,82 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
...
+			existingPath, exists := c.metadataVersionLocation(ident, version)
+			if exists {
+				_ = c.filesystem.Remove(tempPath)
+
+				return fmt.Errorf("%w: metadata file already exists for table %s: %s",
+					conflictErr, strings.Join(ident, "."), existingPath)
+			}
+
+			claimInfo, err := c.filesystem.Stat(claimPath)
+			if err != nil {
```
laskoviymishka, line 709: Small one that folds into the crash-recovery thread zeroshade opened: if a competitor cleans the claim and publishes between the failing rename above and this `Stat`, `Stat` returns `fs.ErrNotExist` and we return a generic "failed to inspect stale metadata claim" that isn't wrapped in `conflictErr` — so a caller distinguishing retryable-conflict from abort-infra spuriously aborts on a race that actually resolved fine.

I'd special-case `fs.ErrNotExist` here and just `continue` — the top of the loop re-checks `metadataVersionLocation` and will surface a proper conflict if it was published.

catalog/hadoop/hadoop.go
```diff
@@ -655,31 +687,21 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
...
-			newVersion, err)
-	}
-	if exists {
-		_ = c.filesystem.Remove(tempPath)
-
-		return nil, "", fmt.Errorf("hadoop catalog: version %d already exists for table %s",
-			newVersion, strings.Join(ident, "."))
-	}
-	if err := c.commitMetadataFile(ident, tempPath, newMetaPath, table.ErrCommitFailed); err != nil {
+	if err := c.commitMetadataFile(ident, newVersion, tempPath, newMetaPath, table.ErrCommitFailed); err != nil {
```
laskoviymishka, line 674: With this pre-check gone, `metadataVersionExists` loses its last caller — the claim protocol supersedes it. I'd drop it while we're here so it doesn't drift out of sync with the codec-aware path.

catalog/hadoop/hadoop_test.go
```diff
@@ -564,18 +613,29 @@ func (s *HadoopCatalogTestSuite) TestIsTableDirWithGzipMetadata() {
 	s.True(s.requireIsTableDir(tableDir))
 }
 
+func (s *HadoopCatalogTestSuite) TestIsTableDirIgnoresZstdMetadata() {
+	tableDir := filepath.Join(s.warehouse, "ns", "tbl")
+	metaDir := filepath.Join(tableDir, "metadata")
+	s.Require().NoError(os.MkdirAll(metaDir, 0o755))
+	s.Require().NoError(os.WriteFile(filepath.Join(metaDir, "v1.zstd.metadata.json"), nil, 0o644))
+
+	s.False(isTableDir(s.cat.filesystem, s.cat.isLocal, tableDir))
```
laskoviymishka: `isTableDir` returns an error alongside the bool, and passing the call straight into `s.False` slides it into the assertion's message args instead of checking it. The sibling tests go through `requireIsTableDir`, which surfaces it — I'd use that here too so a real scan failure can't masquerade as a clean false.

catalog/hadoop/hadoop_test.go
```diff
@@ -1488,7 +1605,144 @@ func (s *HadoopCatalogTestSuite) TestCreateTableConcurrentMetadataPublishConflic
...
+	s.Require().NoError(err)
+	gzipPath, err := s.cat.metadataFilePathForCompression(ident, 1, table.MetadataCompressionCodecGzip)
+	s.Require().NoError(err)
+
+	existing := 0
+	for _, path := range []string{plainPath, gzipPath} {
+		if _, err := os.Stat(path); err == nil {
+			existing++
+		} else {
+			s.True(os.IsNotExist(err))
```
laskoviymishka: `os.IsNotExist` doesn't unwrap, so this misses a wrapped not-exist. `s.ErrorIs(err, fs.ErrNotExist)` is the unwrapping form — same pattern in the other new mixed-codec test.

catalog/hadoop/hadoop_test.go
```diff
@@ -1488,7 +1605,144 @@ func (s *HadoopCatalogTestSuite) TestCreateTableConcurrentMetadataPublishConflic
...
+		err     error
+	}
+
+	results := make(chan createResult, 2)
+	var wg sync.WaitGroup
+	for _, props := range []iceberg.Properties{
+		nil,
+		{table.MetadataCompressionKey: table.MetadataCompressionCodecGzip},
+	} {
+		props := props
```
laskoviymishka: go.mod is on 1.25, so the per-iteration loop var is automatic — this copy (and the one in the commit-side test) is redundant.

laskoviymishka (changes requested):
Everything I flagged last time is in now. `zstd` is a hard reject on write in `metadataFilePathForCompression`, and discovery canonicalizes through `metadataVersionLocation`, so the hint and the walk agree on which file wins. That closes both earlier threads.

One new thing before this lands, in the same helper: `metadataVersionLocation` swallows the `scanMetadataFiles` error and returns `("", false)`, which looks exactly like “version does not exist.” That helper is the last cross-codec guard after the claim rename in `commitMetadataFile`, so a transient list error there could let plain `vN` publish next to compressed `vN` and silently drop a commit.

I’d give the helper an error return and fail the commit on scan errors. Left the detail inline.

The claim-protocol and stale-recovery threads are zeroshade’s. I only left a small note where `Stat` on a vanished claim gets classified as infra rather than conflict, since that feeds into the crash-recovery follow-up.

Fix the swallowed error and I think this is good to land.

catalog/hadoop/hadoop.go
```diff
@@ -63,6 +64,8 @@ var uuidMetadataPattern = regexp.MustCompile(
 	`^([0-9]{5})-[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(?:\.gz)?\.metadata\.json$`,
 )
 
+const metadataClaimStaleAfter = time.Minute
```
zeroshade, line 70: nit: the stale-claim recovery timeout is hardcoded to 1 minute; a very slow live committer could lose its claim. It fails safe (errors rather than publishing duplicate metadata), but a comment noting the tradeoff — or making it tunable — would help operators.

zeroshade (approved):
LGTM — codec-aware filenames, zstd rejection, the codec-independent version-claim protocol, and scan-error propagation all look good. One operational nit inline.

laskoviymishka (approved):

PR #1328, opened 2026-06-27
fix(table): dedupe transaction requirements by semantics

fallintoplace:
Summary

- dedupe transaction requirements by their full semantic payload instead of GetType() alone
- preserve distinct requirements of the same type, including multiple AssertRefSnapshotID checks for different refs
- add focused transaction regressions for distinct same-type requirements and repeated identical requirements

Why

Transaction.apply currently dedupes requirements by GetType() only. That drops semantically distinct requirements that share a type, especially AssertRefSnapshotID checks for different refs. Operations such as expiring snapshots can build multiple ref assertions, and collapsing them weakens optimistic concurrency control by letting commits proceed without validating every referenced branch or tag.

Using the full requirement payload as the dedupe key keeps equivalent duplicates collapsed while preserving distinct assertions.

Testing

- go test ./table -run 'TestTransactionApply(KeepsDistinctRequirementsOfSameType|DedupesEquivalentRequirementsWithinAndAcrossCalls)$' -count=1
- go test ./table -count=1
- go test ./... -run '^$' -count=1
- git diff --check

### DIFF
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -114,12 +114,23 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 
 	existing := map[string]struct{}{}
 	for _, r := range t.reqs {
-		existing[r.GetType()] = struct{}{}
+		key, err := requirementSemanticKey(r)
+		if err != nil {
+			return err
+		}
+
+		existing[key] = struct{}{}
 	}
 
 	for _, r := range reqs {
-		if _, ok := existing[r.GetType()]; !ok {
+		key, err := requirementSemanticKey(r)
+		if err != nil {
+			return err
+		}
+
+		if _, ok := existing[key]; !ok {
 			t.reqs = append(t.reqs, r)
+			existing[key] = struct{}{}
 		}
 	}
 
@@ -140,6 +151,15 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 	}
 
 	return nil
+}
+
+func requirementSemanticKey(r Requirement) (string, error) {
+	data, err := json.Marshal(r)
+	if err != nil {
+		return "", fmt.Errorf("marshal requirement %q: %w", r.GetType(), err)
+	}
+
+	return string(data), nil
 }
 
 // addValidator appends a conflict validator under t.mx. Producers
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -153,6 +153,8 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 	return nil
 }
 
+// requirementSemanticKey assumes Requirement JSON marshaling is canonical and
+// deterministic for every requirement type that participates in dedupe.
 func requirementSemanticKey(r Requirement) (string, error) {
 	data, err := json.Marshal(r)
 	if err != nil {
--- a/table/transaction_internal_test.go
+++ b/table/transaction_internal_test.go
@@ -38,8 +38,8 @@ func TestTransactionApplyKeepsDistinctRequirementsOfSameType(t *testing.T) {
 	require.NoError(t, err)
 
 	require.Len(t, txn.reqs, 2)
-	requireContainsRequirement(t, txn.reqs, AssertRefSnapshotID(MainBranch, &mainSnapshotID))
-	requireContainsRequirement(t, txn.reqs, AssertRefSnapshotID("feature", &featureSnapshotID))
+	requireContainsRefSnapshotRequirement(t, txn.reqs, MainBranch, &mainSnapshotID)
+	requireContainsRefSnapshotRequirement(t, txn.reqs, "feature", &featureSnapshotID)
 }
 
 func TestTransactionApplyDedupesEquivalentRequirementsWithinAndAcrossCalls(t *testing.T) {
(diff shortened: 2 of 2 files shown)


table/transaction.go
```diff
@@ -142,6 +153,15 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 	return nil
 }
 
+func requirementSemanticKey(r Requirement) (string, error) {
+	data, err := json.Marshal(r)
```
tanmayrauth, line 159: The dedupe key relies on encoding/json being deterministic. That holds for all current requirement types,  fields serialize in declaration order and none contain maps — but a future requirement that adds a map field would silently break dedupe (Go sorts map keys, so it'd still be stable... actually that's fine — the real risk is any field whose marshaling isn't canonical). Could you add a one-line doc comment on requirementSemanticKey noting it assumes canonical/deterministic marshaling, so the constraint is visible to whoever adds the next Requirement type?

table/transaction_internal_test.go
```diff
@@ -0,0 +1,111 @@
...
+		TimestampMs:      now + 1,
+	}))
+
+	require.NoError(t, txn.meta.SetSnapshotRef(MainBranch, 10, BranchRef))
+	require.NoError(t, txn.meta.SetSnapshotRef("feature", 20, BranchRef))
+
+	return txn
+}
+
+func requireContainsRequirement(t *testing.T, requirements []Requirement, expected Requirement) {
```
tanmayrauth: requireContainsRequirement asserts membership via requirementSemanticKey — the same function the production change uses. If that function had a bug, both sides would agree and the test would still pass.
  Comparing on the concrete fields (ref + snapshot id) instead would make the assertion independent of the code under test.

tanmayrauth (commented):
Thanks for the PR, confirmed this is a real concurrency fix, not cleanup, the old GetType() key dropped every ref assertion after the first across apply calls. Two small non-blocking things inline.

table/transaction_internal_test.go
fallintoplace: Thank you, I push the fix here. [e35a1e1](https://github.com/apache/iceberg-go/pull/1328/commits/e35a1e131cebd3e35e9b73a5dbe27c66db8aa055)

table/transaction.go
```diff
@@ -114,12 +114,23 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
...
 	}
 
 	for _, r := range reqs {
-		if _, ok := existing[r.GetType()]; !ok {
+		key, err := requirementSemanticKey(r)
+		if err != nil {
+			return err
+		}
+
+		if _, ok := existing[key]; !ok {
```
laskoviymishka, line 131: Now that the key is the full payload, I'd like to think through one case with you. If `apply` runs more than once on the same `Transaction` for the same ref at different snapshot IDs — say an append on an empty table emits `AssertRefSnapshotID("main", nil)`, then a later append emits `AssertRefSnapshotID("main", &firstID)` — those serialize to different JSON (`"snapshot-id":null` vs `:<id>`), so both survive dedupe and both go to the catalog. The base state only matches one of them, so I'd expect the other to 409. The old type-only key collapsed them into one.

Is that path reachable in practice, or does `apply` only ever see one assertion per ref per transaction? If it's reachable, the design question is what the key should really be: full payload preserves the genuinely-distinct refs you're fixing, but it also preserves same-ref/different-id pairs that we probably don't want both of. Java threads this with a per-ref `changedRefs` set — first assertion per ref wins. One option here is to key `assertRefSnapshotID` on type + ref and fall back to payload for the rest. How are you thinking about it?

table/transaction.go
```diff
@@ -142,6 +153,17 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 	return nil
 }
 
+// requirementSemanticKey assumes Requirement JSON marshaling is canonical and
```
laskoviymishka, line 156: More of a design question than a blocker: this comment names a load-bearing invariant — every `Requirement` marshals to canonical, deterministic JSON — that nothing enforces. It holds today because every requirement is flat scalars in declaration order, but `Requirement` is exported, and a future field that's a map (non-deterministic order) or one tagged `json:"-"` (two distinct requirements collapsing to one key) would break the key silently.

Two ways to think about it. Keep `json.Marshal` but guard the invariant with a test that round-trips each concrete requirement type twice and asserts equal keys — cheap, and it turns the comment into something enforced. Or give `Requirement` a typed `SemanticKey()` (e.g. `"assert-ref-snapshot-id:main:10"`) so the equivalence contract is explicit rather than a side effect of marshaling — more code, but it's the more honest abstraction and it'd sidestep the same-ref question above too. Which direction feels right to you?

zeroshade (approved):
LGTM — semantic (JSON-marshaled) dedupe key keeps distinct same-type requirements while still collapsing true duplicates. Reviewed + CI green. Thanks @fallintoplace!

laskoviymishka (changes requested):
This is the right fix for a real bug — the old type-string key collapsed every `AssertRefSnapshotID` into one entry, so `ExpireSnapshots` and friends only ever asserted a single ref. Canonicalizing on the payload is the correct direction, it lines up with Java's per-ref behavior, and the multi-ref test demonstrates it cleanly.

I'd hold this before merging though, on one open question. Once we key on the full JSON, two operations in the same `Transaction` that touch the same ref at different snapshot IDs (e.g. an append on an empty table, then a second append) produce two distinct `AssertRefSnapshotID` for that ref — `"snapshot-id":null` vs a concrete id — and both get sent to the catalog. I'd expect the catalog to 409 that, where the old type-only key silently dropped the second. I left the full walkthrough inline; I can't tell from the diff whether that path is actually reachable, so I'd like your read before this lands. If it is reachable, keying `assertRefSnapshotID` on type + ref (Java's `changedRefs` pattern) and falling back to JSON for the rest fixes it.

Smaller things:

- the `requirementSemanticKey` comment documents a determinism invariant that nothing enforces — either guard it with a per-type round-trip test, or switch to a typed `SemanticKey()`
- the new `existing[key]` insert is an intentional within-call dedupe change that deserves a one-line comment
- reuse the pointer helper already in the package test scope instead of adding `transactionTestPtr`/`transactionTestInt64PtrEqual`
- a same-type/different-value case (and a nil-vs-id snapshot case) would prove the fix generalizes past `assertRefSnapshotID`

Sort out the same-ref question and tighten the helper comment and this is good to land. Happy to take another pass once those are addressed.

fallintoplace:
@laskoviymishka should I still have a look later if this is already merged?

laskoviymishka:
@fallintoplace nit's up to you tbh

laskoviymishka:
The design question is i think worth to park in separate refactor issue, i will file this later

PR #1331, opened 2026-06-27
fix(table): preserve version hint URIs in orphan cleanup

fallintoplace:
Summary

build the orphan-cleanup version-hint path with a URI-aware join for URL locations
keep the existing filepath join fallback for plain local filesystem paths
add regression coverage for URI and local-path version-hint construction, plus the referenced-file set

Why

`Table.getReferencedFiles` adds `metadata/version-hint.text` to the referenced set so orphan cleanup will not delete it. That path was being built with `filepath.Join(metadata.Location(), ...)`, which is fine for plain local paths but corrupts object-store URIs like `s3://bucket/table` into `s3:/bucket/table/...`.

Once that malformed path hit orphan cleanup's normalizer, it no longer looked like a URL and could diverge from the real `s3://.../version-hint.text` candidate path. That made the real version-hint file eligible to be treated as an orphan.

Testing

go test ./table -run 'Test(VersionHintLocation|GetReferencedFiles_IncludesStatisticsFiles|.*Orphan.*)$' -count=1
go test ./table -count=1
git diff --check

### DIFF
--- a/table/orphan_cleanup.go
+++ b/table/orphan_cleanup.go
@@ -304,7 +304,7 @@ func (t Table) getReferencedFiles(ctx context.Context, fs iceio.IO, maxConcurren
 
 	// Add version hint file (for Hadoop-style tables)
 	// Following Java's ReachableFileUtil.versionHintLocation() logic:
-	versionHintPath := filepath.Join(metadata.Location(), "metadata", "version-hint.text")
+	versionHintPath := versionHintLocation(metadata.Location())
 	referenced[normalizeFilePath(versionHintPath)] = false
 
 	for sf := range metadata.Statistics() {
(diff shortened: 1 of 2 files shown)


table/orphan_cleanup.go
```diff
@@ -597,6 +597,16 @@ func normalizeFilePathWithConfig(path string, cfg *orphanCleanupConfig) string {
 	return normalizeNonURLPath(path)
 }
 
+func versionHintLocation(tableLocation string) string {
+	if strings.Contains(tableLocation, "://") || strings.HasPrefix(tableLocation, "file:") {
```
tanmayrauth, line 601: Nice catch adding `|| strings.HasPrefix(tableLocation, "file:")` — since `file:///x` and `file://host/x` already contain `://`, this branch is really only here to cover the single-slash `file:/x` opaque form. Just confirming that's the intent.

table/orphan_cleanup_test.go
```diff
@@ -157,6 +158,36 @@ func TestNormalizeNonURLPath(t *testing.T) {
 	}
 }
 
+func TestVersionHintLocation(t *testing.T) {
```
tanmayrauth, line 161: The three cases here all either contain `://` (s3, file:///) or fall through to filepath.Join (local), so none of them actually exercise the `file:`-prefix-only branch above. A `file:/tmp/table` →  `file:/tmp/table/metadata/version-hint.text` case would lock in the one input that branch uniquely guards.

tanmayrauth (commented):
Clean fix and the write→read round-trip holds for s3/gs/file URIs and local paths alike. Just one small test-coverage nit. Details inline.

zeroshade (approved):
LGTM — `versionHintLocation` preserves URI schemes via `url.JoinPath`, falling back to `filepath.Join` for local paths. Reviewed + CI green. Thanks @fallintoplace!

PR #1332, opened 2026-06-27
fix(partitions): align paths with compacted partition fields

fallintoplace:
Summary

make PartitionToPath walk the same active compacted partition-field sequence as PartitionType
reuse that active-field computation in both code paths to keep them aligned
add a regression test for a spec whose leading partition source column has been dropped from the current schema

Why

`PartitionSpec.PartitionType` explicitly compacts away partition fields whose source columns are no longer present in the current schema. But `PartitionSpec.PartitionToPath` still indexed into `ps.fields` positionally while reading values from that compacted partition record.

When an earlier partition source column had been dropped, the compacted record shifted left while `PartitionToPath` kept using names and transforms from the full spec. That could label values under the wrong partition field names or apply the wrong transform when generating partition paths.

Testing

go test . -run 'Test(PartitionType|PartitionSpecToPath|PartitionSpecToPathWithDroppedLeadingSourceColumn)$' -count=1
go test . -count=1
go test ./table -count=1
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m . ./table/...
git diff --check

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -482,20 +482,20 @@ func (ps *PartitionSpec) LastAssignedFieldID() int {
 	return id
 }
 
-type partitionFieldType struct {
+type activePartitionField struct {
 	field      PartitionField
 	resultType Type
 }
 
-func (ps *PartitionSpec) activePartitionFields(schema *Schema) []partitionFieldType {
-	fields := make([]partitionFieldType, 0, len(ps.fields))
+func (ps *PartitionSpec) activePartitionFields(schema *Schema) []activePartitionField {
+	fields := make([]activePartitionField, 0, len(ps.fields))
 	for _, field := range ps.fields {
 		sourceType, ok := schema.FindTypeByID(field.SourceID())
 		if !ok {
 			continue
 		}
 
-		fields = append(fields, partitionFieldType{
+		fields = append(fields, activePartitionField{
 			field:      field,
 			resultType: field.Transform.ResultType(sourceType),
 		})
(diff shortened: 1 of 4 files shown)


partitions.go
```diff
@@ -482,6 +482,28 @@ func (ps *PartitionSpec) LastAssignedFieldID() int {
 	return id
 }
 
+type partitionFieldType struct {
```
tanmayrauth: nit: this is really a (field, resolvedType) pair rather than a "type" — activePartitionField might read a touch clearer.

tanmayrauth (commented):
Nice, well-scoped fix,  traced it through the snapshot-summary caller and the compacted record lines up with the new active-field indexing. LGTM. One out-of-scope question and a naming nit inline.

getRecordPartitions in table/partitioned_fanout_writer.go has the same shape this PR is fixing,  it sizes its loop by spec.PartitionType(schema).FieldList (compacted) but indexes spec.Field(i) (full spec). It shouldn't trigger today since the active write spec shouldn't have a dropped source column, but is that guaranteed? Might be worth aligning it on activePartitionFields too (or dropping a comment) so the two paths don't drift. Happy to leave it for a follow-up since it's outside this PR's scope.

zeroshade (approved):
LGTM — partition paths/records now indexed by active (compacted) fields, keeping them aligned when leading sources are dropped. Reviewed + CI green. Thanks @fallintoplace!

PR #1333, opened 2026-06-27
fix(table): return scanner memoization errors

fallintoplace:
Summary

replace the panic-wrapping memoized factory helper with an error-carrying cache
propagate cached factory errors through scanner planning, filtered overwrite classification, and conflict validation
add regression coverage for cached factory failures and an invalid-spec manifest path that used to panic

Why

The scanner code memoized several per-spec builders using `newKeyDefaultMapWrapErr`, which converted any factory error into a panic. That meant ordinary failures like unknown partition spec ids or projection/evaluator build errors could crash the process instead of being returned to the caller.

The shared helper is also reused in transaction-side manifest classification and conflict validation, so the same panic behavior leaked there too.

Testing

go test ./table -run 'Test(KeyDefaultMapWrapErrCachesError|FetchPartitionSpecFilteredManifests_InvalidSpecIDDoesNotPanic|BuildPartitionProjectionWithInvalidSpecID|BuildManifestEvaluatorWithInvalidSpecID|BuildPartitionEvaluatorWithInvalidSpecID|FetchPartitionSpecFilteredManifests_PropagatesEvalError)$' -count=1
go test ./table -count=1
git diff --check

### DIFF
--- a/table/conflict_validation.go
+++ b/table/conflict_validation.go
@@ -378,7 +378,7 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
 
 			mEval, err := manifestEvals.Get(int(mf.PartitionSpecID()))
 			if err != nil {
-				return fmt.Errorf("building manifest evaluator for spec %d: %w", mf.PartitionSpecID(), err)
+				return fmt.Errorf("failed to build manifest evaluator for spec %d: %w", mf.PartitionSpecID(), err)
 			}
 			keep, err := mEval(mf)
 			if err != nil {
@@ -390,7 +390,7 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
 
 			pEval, err := partitionEvals.Get(int(mf.PartitionSpecID()))
 			if err != nil {
-				return fmt.Errorf("building partition evaluator for spec %d: %w", mf.PartitionSpecID(), err)
+				return fmt.Errorf("failed to build partition evaluator for spec %d: %w", mf.PartitionSpecID(), err)
 			}
 			for e, err := range mf.Entries(ctx.fs, false) {
 				if err != nil {
--- a/table/conflict_validation.go
+++ b/table/conflict_validation.go
@@ -376,7 +376,10 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
 				continue
 			}
 
-			mEval := manifestEvals.Get(int(mf.PartitionSpecID()))
+			mEval, err := manifestEvals.Get(int(mf.PartitionSpecID()))
+			if err != nil {
+				return fmt.Errorf("building manifest evaluator for spec %d: %w", mf.PartitionSpecID(), err)
+			}
 			keep, err := mEval(mf)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


table/transaction.go
```diff
@@ -1542,7 +1542,10 @@ func (t *Transaction) classifyFilesForFilteredDeletions(ctx context.Context, fs
 	for _, manifest := range manifests {
 		manifest := manifest // capture loop variable
 		g.Go(func() error {
-			manifestEval := manifestEvaluators.Get(int(manifest.PartitionSpecID()))
+			manifestEval, err := manifestEvaluators.Get(int(manifest.PartitionSpecID()))
+			if err != nil {
+				return fmt.Errorf("failed to build manifest evaluator for spec %d: %w", manifest.PartitionSpecID(), err)
+			}
 			if manifestEval != nil {
```
tanmayrauth: The `if manifestEval != nil` guard is dead now that the `err != nil` return runs first, `buildManifestEvaluator` only ever returns a nil func alongside a non-nil error, so past line 1548 this is always true. Harmless, but could be dropped to make the intent clearer.

table/conflict_validation.go
```diff
@@ -376,7 +376,10 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
 				continue
 			}
 
-			mEval := manifestEvals.Get(int(mf.PartitionSpecID()))
+			mEval, err := manifestEvals.Get(int(mf.PartitionSpecID()))
+			if err != nil {
+				return fmt.Errorf("building manifest evaluator for spec %d: %w", mf.PartitionSpecID(), err)
```
tanmayrauth: Tiny consistency nit: here the wrap is "building manifest evaluator for spec %d" (and "building partition evaluator..." on 393), but the equivalent wraps in scanner.go:626 and transaction.go:1547 say "failed to build ...". Worth aligning the prefix across the three so the messages read uniformly.

table/scanner.go
```diff
@@ -73,17 +85,33 @@ func newKeyDefaultMap[K comparable, V any](factory func(K) V) *keyDefaultMap[K,
...
+func (k *keyDefaultMapErr[K, V]) Get(key K) (V, error) {
+	k.mx.RLock()
+	if v, ok := k.data[key]; ok {
+		k.mx.RUnlock()
+
+		return v.value, v.err
+	}
+
+	k.mx.RUnlock()
+	k.mx.Lock()
```
tanmayrauth, line 90: Just confirming the intent: caching the error means a failing factory is never retried for that spec id (the new TestKeyDefaultMapWrapErrCachesError pins this). That's fine for these deterministic errors, unknown spec id, projection build failure but wanted to flag it explicitly since it's a behavior the helper now guarantees, not just an implementation detail.

tanmayrauth (commented):
Clean fix, turning these memoization panics into propagated errors is the right call, and reusing the existing double-checked-locking pattern keeps the concurrency behavior identical. Just two nits and one confirmation, all inline.

laskoviymishka (approved):

fallintoplace:
@tanmayrauth @laskoviymishka Thank you both for the reviews!

PR #1334, opened 2026-06-27
fix(manifest): preserve 64-bit list metadata ids

fallintoplace:
Summary

format manifest-list header snapshot ids, sequence numbers, parent snapshot ids, and first-row ids directly from int64
remove the intermediate `int` conversion in V1/V2/V3 manifest list writers
add regression coverage that reads the OCF header metadata back for large 64-bit values

Why

The manifest-list writers were serializing several 64-bit metadata fields by converting them to `int` and then calling `strconv.Itoa`. That is conceptually wrong for Iceberg's 64-bit ids and unsafe on 32-bit builds.

Using `strconv.FormatInt(..., 10)` keeps the full value intact for snapshot ids, sequence numbers, parent snapshot ids, and v3 first-row ids.

Testing

go test . -run 'TestManifests/TestManifestListWriterMetadataPreservesInt64Values$' -count=1
go test . -count=1
go test ./table -count=1
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m . ./table/...
git diff --check

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1427,12 +1427,12 @@ func NewManifestListWriterV1(out io.Writer, snapshotID int64, parentSnapshot *in
 
 	parentSnapshotStr := "null"
 	if parentSnapshot != nil {
-		parentSnapshotStr = strconv.Itoa(int(*parentSnapshot))
+		parentSnapshotStr = strconv.FormatInt(*parentSnapshot, 10)
 	}
 
 	return m, m.init(map[string][]byte{
 		"format-version":     []byte(strconv.Itoa(m.version)),
-		"snapshot-id":        []byte(strconv.Itoa(int(snapshotID))),
+		"snapshot-id":        []byte(strconv.FormatInt(snapshotID, 10)),
 		"parent-snapshot-id": []byte(parentSnapshotStr),
 	})
 }
@@ -1447,13 +1447,13 @@ func NewManifestListWriterV2(out io.Writer, snapshotID, sequenceNumber int64, pa
 
 	parentSnapshotStr := "null"
 	if parentSnapshot != nil {
-		parentSnapshotStr = strconv.Itoa(int(*parentSnapshot))
+		parentSnapshotStr = strconv.FormatInt(*parentSnapshot, 10)
 	}
 
 	return m, m.init(map[string][]byte{
 		"format-version":     []byte(strconv.Itoa(m.version)),
-		"snapshot-id":        []byte(strconv.Itoa(int(snapshotID))),
-		"sequence-number":    []byte(strconv.Itoa(int(sequenceNumber))),
+		"snapshot-id":        []byte(strconv.FormatInt(snapshotID, 10)),
+		"sequence-number":    []byte(strconv.FormatInt(sequenceNumber, 10)),
 		"parent-snapshot-id": []byte(parentSnapshotStr),
 	})
 }
@@ -1468,14 +1468,14 @@ func NewManifestListWriterV3(out io.Writer, snapshotId, sequenceNumber, firstRow
 	}
 	parentSnapshotStr := "null"
 	if parentSnapshot != nil {
-		parentSnapshotStr = strconv.Itoa(int(*parentSnapshot))
+		parentSnapshotStr = strconv.FormatInt(*parentSnapshot, 10)
 	}
 
 	return m, m.init(map[string][]byte{
 		"format-version":     []byte(strconv.Itoa(m.version)),
-		"snapshot-id":        []byte(strconv.Itoa(int(snapshotId))),
-		"sequence-number":    []byte(strconv.Itoa(int(sequenceNumber))),
-		"first-row-id":       []byte(strconv.Itoa(int(firstRowID))),
+		"snapshot-id":        []byte(strconv.FormatInt(snapshotId, 10)),
+		"sequence-number":    []byte(strconv.FormatInt(sequenceNumber, 10)),
+		"first-row-id":       []byte(strconv.FormatInt(firstRowID, 10)),
 		"parent-snapshot-id": []byte(parentSnapshotStr),
 	})
 }
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -2023,6 +2023,73 @@ func (m *ManifestTestSuite) TestV3ManifestListWriterPersistsPerManifestFirstRowI
 	m.EqualValues(5022, *writer.NextRowID())
 }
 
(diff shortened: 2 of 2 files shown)


manifest.go
```diff
@@ -1468,14 +1468,14 @@ func NewManifestListWriterV3(out io.Writer, snapshotId, sequenceNumber, firstRow
 	}
 	parentSnapshotStr := "null"
 	if parentSnapshot != nil {
-		parentSnapshotStr = strconv.Itoa(int(*parentSnapshot))
+		parentSnapshotStr = strconv.FormatInt(*parentSnapshot, 10)
 	}
 
 	return m, m.init(map[string][]byte{
 		"format-version":     []byte(strconv.Itoa(m.version)),
```
tanmayrauth, line 1475: Nice that you caught first-row-id here too, not just the fields shared with v1/v2, FormatInt is the right call for all of these since they're all 64-bit per the spec.

manifest_test.go
```diff
@@ -2023,6 +2023,73 @@ func (m *ManifestTestSuite) TestV3ManifestListWriterPersistsPerManifestFirstRowI
 	m.EqualValues(5022, *writer.NextRowID())
 }
 
+func (m *ManifestTestSuite) TestManifestListWriterMetadataPreservesInt64Values() {
```
tanmayrauth, line 2027: 1<<40 already proves the >2^32 case well. Might be worth also asserting a value up near math.MaxInt64 (a realistic snapshot id like 9223372036854775807), since that's the boundary a future regression to a
  32-bit path would corrupt first.

tanmayrauth (commented):
LGTM, correct 64-bit fix across all three writer versions. One optional test idea inline.

manifest_test.go
```diff
@@ -2023,6 +2024,101 @@ func (m *ManifestTestSuite) TestV3ManifestListWriterPersistsPerManifestFirstRowI
...
+
+			reader, err := ocf.NewReader(&buf)
+			m.Require().NoError(err)
+			defer func() {
+				m.Require().NoError(reader.Close())
+			}()
+
+			meta := reader.Metadata()
+			for key, want := range tt.expected {
+				m.Equal(want, string(meta[key]))
```
laskoviymishka, line 2088: I think this test is a no-op on our actual CI. On a 64-bit host `int` is 64-bit, so `strconv.Itoa(int(x))` and `strconv.FormatInt(x, 10)` produce identical bytes for every value here — including the near-MaxInt64 subtest below. It'll pass against the pre-fix code too, and there's no GOARCH=386/wasm step in the matrix to exercise the truncating path, so it doesn't actually fence the bug on any supported target.

Could we make it fail without the fix? A `// +build 386`-tagged variant that asserts the truncated vs. full output differ would document the gap, or — simpler — a platform-independent invariant: round-trip the value through `int64 → string → ParseInt` and assert equality, which only holds when the writer used the 64-bit formatter. wdyt?

Separately, while we're here: the iteration over `tt.expected` is map-ordered, so a failure cites a different key each run. Worth passing the key as msgAndArgs — `m.Equal(want, string(meta[key]), "OCF metadata key %q", key)` — so failures are reproducible.

manifest_test.go
```diff
@@ -2023,6 +2024,101 @@ func (m *ManifestTestSuite) TestV3ManifestListWriterPersistsPerManifestFirstRowI
...
+		expected map[string]string
+	}{
+		{
+			name: "v1",
+			build: func(w io.Writer) (*ManifestListWriter, error) {
+				return NewManifestListWriterV1(w, (1<<40)+1, &parentSnapshot)
+			},
+			expected: map[string]string{
+				"format-version":     "1",
+				"snapshot-id":        "1099511627777",
```
laskoviymishka, line 2042: These string literals need mental arithmetic to tie back to the `(1<<40)+N` inputs. The `v3_near_max_int64` subtest gets this right by computing the expected value (`strconv.FormatInt(...)`) instead of hardcoding it. Could we match that here so the table is self-documenting and can't silently drift from the inputs?

manifest_test.go
```diff
@@ -2023,6 +2024,101 @@ func (m *ManifestTestSuite) TestV3ManifestListWriterPersistsPerManifestFirstRowI
...
+	for _, tt := range tests {
+		m.Run(tt.name, func() {
+			var buf bytes.Buffer
+			writer, err := tt.build(&buf)
+			m.Require().NoError(err)
+			m.Require().NoError(writer.Close())
+
+			reader, err := ocf.NewReader(&buf)
+			m.Require().NoError(err)
+			defer func() {
```
laskoviymishka, line 2082: A deferred `Require` (which does `FailNow` → `Goexit`) can surface a `Close()` error as a confusing panic rather than a clean failure in some testify versions. The rest of this file closes readers inline with `m.NoError(reader.Close())` after the assertions. I'd either move the close inline or drop to a plain `defer reader.Close()` if the close error isn't meaningful here — applies to both subtests.

laskoviymishka (approved):
Nice catch — `strconv.Itoa(int(x))` on an int64 silently truncates to 32 bits on 32-bit targets, and these are all `long` header fields in the spec, so this is a genuine cross-client correctness bug. `FormatInt(x, 10)` is the right fix and matches what Java/Python/Rust emit byte-for-byte. The "null" parent sentinel and `format-version` staying on `Itoa` are both correct. Happy to approve this.

One thing worth knowing about the regression test, even if you'd rather not touch it here: on a 64-bit host `int` is 64-bit, so the old and new code emit identical bytes for every value in the table — the test passes against pre-fix code too, and there's no 32-bit step in the CI matrix to exercise the truncating path. So it documents intent more than it actually fences the regression. I left an inline note on making it fail without the fix (a platform-independent round-trip invariant, or a GOARCH=386-tagged case) — your call.

A few smaller, optional things inline too — the map-ordered assertion makes failures non-reproducible, the hardcoded `"1099511627777"`-style literals would read better computed from the inputs like the near-max subtest does, and the deferred `Require(Close())` is worth swapping for an inline close to match the rest of the file. All at your discretion.

Two sibling truncation sites I noticed but that are out of scope here — `cmd/iceberg/output.go` `LastUpdatedMillis` and `table/snapshots.go` snapshot totals are the same bug class. Worth a follow-up issue.

zeroshade (approved):
LGTM — manifest-list writers now use `strconv.FormatInt` to preserve full int64 IDs (snapshot/sequence/parent/first-row). Reviewed for correctness + style; CI green. Thanks @fallintoplace!

PR #1337, opened 2026-06-27
fix(table): validate pos delete position chunks

fallintoplace:
Summary

- replace unchecked positional-delete `pos` casts with a shared validated collector
- return `ErrInvalidSchema` when a delete file exposes a non-`int64` `pos` chunk
- add regression coverage for malformed `pos` chunk types

Why

The Arrow scanner assumed every positional-delete `pos` chunk was `*array.Int64` and cast it directly in both scan paths. If a malformed delete file carried the wrong Arrow type, the scanner panicked instead of returning a schema error.

Testing

- `go test ./table -run 'Test(GroupPosDeletesByFilePathSupportsStringLayouts|GroupPosDeletesByFilePathRejectsUnsupportedFilePathLayout|CollectPosDeletePositionsRejectsUnsupportedPosType|ProcessPositionalDeletesAcrossBatches)$' -count=1`
- `go test ./table -count=1`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -391,6 +391,32 @@ func groupPosDeletesByFilePath(ctx context.Context, filePathCol, posCol *arrow.C
 	return results, nil
 }
 
+func collectPosDeletePositions(positionalDeletes positionDeletes) (set[int64], error) {
+	deletes := set[int64]{}
+	for _, chunk := range positionalDeletes {
+		if chunk == nil {
+			return nil, fmt.Errorf("%w: nil pos column chunk in position delete file",
+				iceberg.ErrInvalidSchema)
+		}
+		if chunk.DataType().ID() != arrow.INT64 {
+			return nil, fmt.Errorf("%w: unsupported pos column type %s in position delete file",
+				iceberg.ErrInvalidSchema, chunk.DataType())
+		}
+		for _, arr := range chunk.Chunks() {
+			posArr, ok := arr.(*array.Int64)
+			if !ok {
+				return nil, fmt.Errorf("%w: unsupported pos chunk array type %T in position delete file",
+					iceberg.ErrInvalidSchema, arr)
+			}
+			for _, v := range posArr.Int64Values() {
+				deletes[v] = struct{}{}
+			}
+		}
+	}
+
+	return deletes, nil
+}
+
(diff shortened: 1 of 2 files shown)


table/arrow_scanner.go
```diff
@@ -391,6 +391,32 @@ func groupPosDeletesByFilePath(ctx context.Context, filePathCol, posCol *arrow.C
 	return results, nil
 }
 
+func collectPosDeletePositions(positionalDeletes positionDeletes) (set[int64], error) {
+	deletes := set[int64]{}
+	for _, chunk := range positionalDeletes {
+		if chunk == nil {
```
tanmayrauth, line 397: Is the nil-chunk guard meant purely as defensiveness? `groupPosDeletesByFilePath` only ever stores retained, non-nil chunks into the map, so this branch shouldn't be reachable from the scan path — just want to make sure I'm not missing a caller that can pass a nil chunk in.

table/arrow_scanner.go
```diff
@@ -391,6 +391,32 @@ func groupPosDeletesByFilePath(ctx context.Context, filePathCol, posCol *arrow.C
...
+			return nil, fmt.Errorf("%w: nil pos column chunk in position delete file",
+				iceberg.ErrInvalidSchema)
+		}
+		if chunk.DataType().ID() != arrow.INT64 {
+			return nil, fmt.Errorf("%w: unsupported pos column type %s in position delete file",
+				iceberg.ErrInvalidSchema, chunk.DataType())
+		}
+		for _, arr := range chunk.Chunks() {
```
tanmayrauth, line 405: The inner `arr.(*array.Int64)` assertion is unreachable here — once the outer `chunk.DataType().ID() == arrow.INT64` check at line 401 passes, every chunk of that column is already `*array.Int64`, so the
  `!ok` branch and its separate "unsupported pos chunk array type" message can't fire. Not a problem to leave as defensive code, but you could drop the `, ok` check (or fold the two error messages into one) to
  keep it tight.

tanmayrauth (approved):
LGTM, nice catch turning the blind pos cast into a clean schema error. Just one nit and a question, details inline.

zeroshade (approved):
LGTM — pos-delete collection extracted into a helper with nil/INT64 validation, replacing duplicated panicking assertions. Reviewed + CI green. Thanks @fallintoplace!

PR #1338, opened 2026-06-27
fix(glue): revalidate source table during rename

fallintoplace:
Summary

- re-fetch the source Glue table after creating the rename destination and before deleting the source
- roll back the destination table and fail the rename if the source table's VersionId or metadata location changed during the rename window
- add regression coverage for source-table drift during rename alongside the existing rollback path

Why

Glue RenameTable currently snapshots the source table, creates the destination table from that snapshot, and then deletes the source table. If another writer commits between the initial GetTable and DeleteTable, the destination can point at stale metadata while the source entry holding the newer commit is deleted.

Glue exposes VersionId optimistic locking on UpdateTable, but not on DeleteTable, so this change hardens rename by revalidating the source table immediately before delete and failing closed if it drifted.

Testing

- `go test ./catalog/glue -run 'TestGlueRenameTable|TestGlueRenameTable_DeleteTableFailureRollback|TestGlueRenameTable_SourceTableChangedRollback|TestIsConcurrentModificationException' -count=1`
- `go test ./catalog/glue -count=1`
- `git diff --check`

### DIFF
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -461,6 +461,9 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
 		return nil, fmt.Errorf("failed to create the table %s.%s: %w", fromDatabase, fromTable, err)
 	}
 
+	// Revalidate the source table immediately before delete to narrow the
+	// rename race window. Glue does not offer a conditional DeleteTable, so
+	// this cannot eliminate the race if a commit lands after this read.
 	currentFromGlueTable, err := c.getTable(ctx, fromDatabase, fromTable)
 	if err != nil {
 		c.rollbackRenamedTable(ctx, toDatabase, toTable)
@@ -495,7 +498,7 @@ func (c *Catalog) rollbackRenamedTable(ctx context.Context, database, tableName 
 		Name:         aws.String(tableName),
 	})
 	if rollbackErr != nil {
-		fmt.Printf("failed to rollback the new table %s.%s: %v", database, tableName, rollbackErr)
+		log.Printf("failed to rollback the new table %s.%s: %v", database, tableName, rollbackErr)
 	}
 }
 
@@ -867,6 +870,9 @@ func glueTableChangedDuringRename(original, current *types.Table) bool {
 	originalVersionID := aws.ToString(original.VersionId)
 	currentVersionID := aws.ToString(current.VersionId)
 	if originalVersionID != "" || currentVersionID != "" {
+		// Glue returns the same VersionId across consecutive GetTable calls
+		// unless the table was updated, making this a stronger drift check
+		// than comparing metadata_location when both versions are present.
 		return originalVersionID != currentVersionID
 	}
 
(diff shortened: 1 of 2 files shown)


catalog/glue/glue.go
```diff
@@ -461,29 +461,44 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
 		return nil, fmt.Errorf("failed to create the table %s.%s: %w", fromDatabase, fromTable, err)
 	}
 
+	currentFromGlueTable, err := c.getTable(ctx, fromDatabase, fromTable)
```
tanmayrauth, line 467: This closes most of the race, but there's still a TOCTOU gap between this revalidation GetTable and the DeleteTable below — a writer that commits in that sub-window still loses its commit. The summary already notes Glue has no conditional DeleteTable, so I think this is the best available mitigation (and create-then-delete is the right order). Could you add a line to the comment making clear it narrows rather than eliminates the race? Just so a future reader doesn't assume it's airtight.

catalog/glue/glue.go
```diff
@@ -843,3 +858,25 @@ func constructDatabaseInput(database string, props iceberg.Properties) *types.Da
...
 }
+
+func glueTableChangedDuringRename(original, current *types.Table) bool {
+	if original == nil || current == nil {
+		return true
+	}
+
+	originalVersionID := aws.ToString(original.VersionId)
+	currentVersionID := aws.ToString(current.VersionId)
+	if originalVersionID != "" || currentVersionID != "" {
```
tanmayrauth, line 872: The version-first / metadata-location-fallback logic looks right to me, VersionId bumps on every Glue UpdateTable, so it's strictly stronger than comparing metadata_location, and only falling back when both
  versions are empty is sensible. This does rely on two back-to-back GetTable calls returning the same VersionId absent a write (true for Glue), so no spurious rename failures — worth a one-line comment to
  capture that assumption.

catalog/glue/glue.go
```diff
@@ -461,29 +461,44 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
...
 }
 
+func (c *Catalog) rollbackRenamedTable(ctx context.Context, database, tableName string) {
+	_, rollbackErr := c.glueSvc.DeleteTable(ctx, &glue.DeleteTableInput{
+		CatalogId:    c.catalogId,
+		DatabaseName: aws.String(database),
+		Name:         aws.String(tableName),
+	})
+	if rollbackErr != nil {
+		fmt.Printf("failed to rollback the new table %s.%s: %v", database, tableName, rollbackErr)
```
tanmayrauth: nit: the Printf on rollback failure is pre-existing code you've just relocated, so not a blocker but since it's moving anyway, worth routing through a logger and adding the missing trailing newline.

tanmayrauth (commented):
Solid fix and the test coverage is good. One question on the residual window and a couple of small notes

zeroshade (approved):
LGTM — revalidates the source table immediately before the Glue delete to narrow the rename race, with a clean rollback helper. Reviewed + CI green. Thanks @fallintoplace!

Issue #1236, opened 2026-06-19
Proposal: Metrics Reporting framework for iceberg-go


tanmayrauth:
## Summary

Today Java ships Iceberg's **Metrics Reporting API**, PyIceberg has none of it, and neither does Go. This proposal builds the **full framework** in iceberg-go, modeled on Java's long-standing implementation: a pluggable `Reporter` that receives a `ScanReport` after scan planning and a `CommitReport` after a commit, the `Counter`/`Timer`/`Context` primitives that feed them, the scan- and commit-path instrumentation that populates them, and the built-in `Logging`, `InMemory`, and `REST` reporters.

**OpenTelemetry is the final, optional piece** — one opt-in adapter layered on top, not the core abstraction. The framework is the spec-compliant contract; any backend (including OTel) plugs in behind it.

## Upstream alignment

**Framework (PRs 1–6) — modeled on Java's existing implementation.** Java has shipped the full Metrics Reporting framework for years: `MetricsReporter`, `ScanReport`/`CommitReport`, `ScanMetrics`/`CommitMetrics`, `Counter`/`Timer`/`MetricsContext`, the `Logging`/`InMemory`/`REST` reporters, and the scan- and commit-path instrumentation that drives them. The bulk of this proposal ports that proven design to Go, adapting mechanics to Go idioms (e.g. a registry instead of reflection for reporter selection). There is no recent "discussion" to cite for this part — it predates the current dev process, so the reference is the Java codebase itself.

[...]

PR #1343, opened 2026-06-29
feat(metrics): add Reporter contract and built-in reporters (#1236)

tanmayrauth:
First PR of the metrics reporting framework. 
Adds the metrics package with the Reporter interface, the sealed Report marker, and the built-in Nop, Logging, InMemory, and Combine reporters. Reporting is opt-in and must never block or fail the observed scan/commit.

Concrete report types and instrumentation follow in later PRs.

Related: #1236

### DIFF
--- a/metrics/reporter.go
+++ b/metrics/reporter.go
@@ -17,27 +17,33 @@
 
 // Package metrics implements Iceberg's Metrics Reporting API for iceberg-go.
 //
-// A [Reporter] is a pluggable sink that receives a [Report] — a ScanReport
-// after scan planning or a CommitReport after a commit — describing what the
-// client did (files and manifests considered, scanned and skipped; bytes read;
-// commit attempts and durations). Reporting gives operators a standard way to
-// aggregate these otherwise-invisible client-side metrics across many clients.
+// A [Reporter] is a pluggable sink that receives a [MetricsReport] — a
+// ScanReport after scan planning or a CommitReport after a commit — describing
+// what the client did (files and manifests considered, scanned and skipped;
+// bytes read; commit attempts and durations). Reporting gives operators a
+// standard way to aggregate these otherwise-invisible client-side metrics
+// across many clients.
 //
 // Reporting is strictly opt-in: with no reporter configured the instrumented
 // code paths do no work. Reporters must never block or fail the scan/commit
 // they observe — see [Reporter] for the contract.
 //
-// This package provides the contract and the built-in reporters ([Nop],
-// [LoggingReporter], [InMemoryReporter], and [Combine]). The concrete report
-// types and the scan/commit instrumentation are layered on top in later work.
+// This package provides the contract and the built-in reporters
+// ([NopReporter], [LoggingReporter], [InMemoryReporter], and [Combine]). The
+// concrete report types and the scan/commit instrumentation are layered on top
+// in later work.
 package metrics
 
 import "context"
 
-// Report is the sealed marker interface implemented by the concrete report
-// types (ScanReport and CommitReport). It is sealed to this package so the set
-// of report types stays under the framework's control.
-type Report interface {
+// MetricsReport is the sealed marker interface implemented by the concrete
+// report types (ScanReport and CommitReport). It is sealed to this package so
+// the set of report types stays under the framework's control.
+//
+// The name mirrors the MetricsReport type in the Java and Python Iceberg
+// implementations, easing ports across the ecosystem and disambiguating the
+// type from the [Reporter.Report] method that consumes it.
+type MetricsReport interface {
 	isMetricsReport()
 }
 
@@ -50,5 +56,5 @@ type Report interface {
 // (logged and swallowed), never propagated back to the caller. Report must be
 // safe for concurrent use by multiple goroutines.
 type Reporter interface {
-	Report(ctx context.Context, report Report)
+	Report(ctx context.Context, report MetricsReport)
 }
--- a/metrics/reporter.go
+++ b/metrics/reporter.go
@@ -36,16 +36,20 @@ package metrics
 
 import "context"
 
-// MetricsReport is the sealed marker interface implemented by the concrete
-// report types (ScanReport and CommitReport). It is sealed to this package so
-// the set of report types stays under the framework's control.
+// MetricsReport is the marker interface implemented by the concrete report
+// types (ScanReport and CommitReport). It is intentionally empty, mirroring the
+// open MetricsReport interface in the Java and Python Iceberg implementations,
+// so downstream operators can implement it for their own report wrappers.
+//
+// The set of report types is controlled by this package for now, but that is a
+// convention documented here rather than a structural guarantee — the interface
+// is deliberately not sealed, leaving room for third-party reports without a
+// future breaking change.
 //
 // The name mirrors the MetricsReport type in the Java and Python Iceberg
 // implementations, easing ports across the ecosystem and disambiguating the
 // type from the [Reporter.Report] method that consumes it.
-type MetricsReport interface {
-	isMetricsReport()
-}
+type MetricsReport any
 
 // Reporter is a pluggable sink for metrics reports.
 //
--- a/metrics/reporter.go
+++ b/metrics/reporter.go
@@ -29,9 +29,9 @@
 // they observe — see [Reporter] for the contract.
 //
 // This package provides the contract and the built-in reporters
-// ([NopReporter], [LoggingReporter], [InMemoryReporter], and [Combine]). The
-// concrete report types and the scan/commit instrumentation are layered on top
-// in later work.
+// ([NopReporter], [LoggingReporter], [InMemoryReporter], [Combine], and
+// [CombineWithLogger]). The concrete report types and the scan/commit
+// instrumentation are layered on top in later work.
 package metrics
 
 import "context"
@@ -42,9 +42,16 @@ import "context"
 // so downstream operators can implement it for their own report wrappers.
 //
 // The set of report types is controlled by this package for now, but that is a
-// convention documented here rather than a structural guarantee — the interface
-// is deliberately not sealed, leaving room for third-party reports without a
-// future breaking change.
+// convention rather than a structural guarantee: the interface is deliberately
+// not sealed. That openness is the irreversible choice — sealing it later would
+// be the breaking change — so it is worth being precise about what it does and
+// does not buy. Custom report types are honored only by the in-process reporters
+// ([LoggingReporter], [InMemoryReporter], and the [Combine] fan-out). The REST
+// reporter (once it lands) can serialize only the discriminators it knows
+// (scan-report and commit-report), so a third-party MetricsReport carries no
+// discriminator and is silently dropped by that reporter — a spec-compliant
+// catalog would reject it outright. Do not build a custom report expecting it to
+// reach a catalog endpoint.
 //
 // The name mirrors the MetricsReport type in the Java and Python Iceberg
 // implementations, easing ports across the ecosystem and disambiguating the
--- a/metrics/reporter.go
+++ b/metrics/reporter.go
@@ -1,0 +1,54 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package metrics implements Iceberg's Metrics Reporting API for iceberg-go.
+//
+// A [Reporter] is a pluggable sink that receives a [Report] — a ScanReport
+// after scan planning or a CommitReport after a commit — describing what the
+// client did (files and manifests considered, scanned and skipped; bytes read;
+// commit attempts and durations). Reporting gives operators a standard way to
+// aggregate these otherwise-invisible client-side metrics across many clients.
+//
+// Reporting is strictly opt-in: with no reporter configured the instrumented
+// code paths do no work. Reporters must never block or fail the scan/commit
+// they observe — see [Reporter] for the contract.
+//
+// This package provides the contract and the built-in reporters ([Nop],
+// [LoggingReporter], [InMemoryReporter], and [Combine]). The concrete report
+// types and the scan/commit instrumentation are layered on top in later work.
+package metrics
+
+import "context"
+
+// Report is the sealed marker interface implemented by the concrete report
+// types (ScanReport and CommitReport). It is sealed to this package so the set
+// of report types stays under the framework's control.
+type Report interface {
+	isMetricsReport()
+}
+
+// Reporter is a pluggable sink for metrics reports.
+//
+// Report is invoked inline at the scan/commit completion point, so
+// implementations must return promptly and must never block or fail the
+// operation being observed: network-backed reporters should dispatch the
+// actual send on a background worker, and any error must be handled internally
+// (logged and swallowed), never propagated back to the caller. Report must be
+// safe for concurrent use by multiple goroutines.
+type Reporter interface {
+	Report(ctx context.Context, report Report)
+}
--- a/metrics/reporters.go
+++ b/metrics/reporters.go
@@ -21,8 +21,26 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"reflect"
+	"runtime/debug"
 	"sync"
 )
+
+// isNilReport reports whether report carries no value. A plain report == nil
+// catches only an untyped nil interface; once the concrete report types land, a
+// typed nil pointer (e.g. (*ScanReport)(nil)) is a non-nil interface value that
+// would slip past that check, so the reporters below would log or store a nil
+// report — and nil-deref if the concrete type grows an [slog.LogValuer] on a nil
+// receiver. This also treats such typed nils as "no report" so the guards honor
+// the documented "nil reports are ignored" behavior.
+func isNilReport(report MetricsReport) bool {
+	if report == nil {
+		return true
+	}
+	v := reflect.ValueOf(report)
+
+	return v.Kind() == reflect.Ptr && v.IsNil()
+}
 
 // NopReporter is a [Reporter] that discards every report. It is the default
 // when no reporter is configured, so that instrumentation is free unless a user
@@ -51,7 +69,7 @@ func NewLoggingReporter(logger *slog.Logger) *LoggingReporter {
 
 // Report logs report at info level.
 func (r *LoggingReporter) Report(ctx context.Context, report MetricsReport) {
-	if report == nil {
+	if isNilReport(report) {
 		return
 	}
 	logger := r.logger
@@ -72,7 +90,7 @@ var _ Reporter = (*InMemoryReporter)(nil)
 
 // Report appends report to the retained set.
 func (r *InMemoryReporter) Report(_ context.Context, report MetricsReport) {
-	if report == nil {
+	if isNilReport(report) {
 		return
 	}
 	r.mu.Lock()
@@ -149,15 +167,16 @@ func (c *compositeReporter) Report(ctx context.Context, report MetricsReport) {
 			defer func() {
 				if v := recover(); v != nil {
 					// Swallow per the Reporter contract, but surface the
-					// failure (with the offending reporter type) so a broken
-					// reporter is traceable rather than showing up only as
-					// mysteriously-missing metrics.
+					// failure (with the offending reporter type and the stack
+					// at the panic) so a broken reporter is traceable rather
+					// than showing up only as mysteriously-missing metrics.
 					logger := c.logger
 					if logger == nil {
 						logger = slog.Default()
 					}
 					logger.WarnContext(ctx, "iceberg metrics reporter panicked; recovered",
-						"reporter", fmt.Sprintf("%T", r), "panic", v)
+						"reporter", fmt.Sprintf("%T", r), "panic", v,
+						slog.String("stack", string(debug.Stack())))
 				}
 			}()
 			r.Report(ctx, report)
--- a/metrics/reporters.go
+++ b/metrics/reporters.go
@@ -19,6 +19,7 @@ package metrics
 
 import (
 	"context"
+	"fmt"
 	"log/slog"
 	"sync"
 )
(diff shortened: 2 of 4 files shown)


metrics/reporter.go
```diff
@@ -0,0 +1,54 @@
...
+// [LoggingReporter], [InMemoryReporter], and [Combine]). The concrete report
+// types and the scan/commit instrumentation are layered on top in later work.
+package metrics
+
+import "context"
+
+// Report is the sealed marker interface implemented by the concrete report
+// types (ScanReport and CommitReport). It is sealed to this package so the set
+// of report types stays under the framework's control.
+type Report interface {
```
zeroshade: **Naming — the one thing worth settling before this public API ships.**

The marker interface is `Report` while the contract method is `Reporter.Report(ctx, report Report)`, so signatures read as `Report(context.Context, Report)` — a method named like its own parameter type. Java/PyIceberg call the marker **`MetricsReport`**, with `ScanReport`/`CommitReport` implementing it and `MetricsReporter.report(MetricsReport)`.

Suggest renaming this interface `Report` → `MetricsReport` (keep `Reporter.Report(...)` as the verb). That (a) disambiguates the type from the method, and (b) aligns the Go API with the rest of the Iceberg ecosystem, easing ports from Java. Since `ScanReport`/`CommitReport` arrive in later phases, settling the marker name now avoids a churny rename across those PRs.

metrics/reporters.go
```diff
@@ -0,0 +1,131 @@
...
+import (
+	"context"
+	"log/slog"
+	"sync"
+)
+
+// Nop is a [Reporter] that discards every report. It is the default when no
+// reporter is configured, so that instrumentation is free unless a user opts
+// in. The zero value is ready to use.
+type Nop struct{}
```
zeroshade: Tiny consistency nit: the other built-ins are `LoggingReporter` / `InMemoryReporter`, but this one is `Nop`. Consider `NopReporter` so the set reads uniformly — or keep `Nop` intentionally as the short form; just flagging the asymmetry. Non-blocking.

metrics/reporters.go
```diff
@@ -0,0 +1,131 @@
...
+		if r != nil {
+			nonNil = append(nonNil, r)
+		}
+	}
+
+	switch len(nonNil) {
+	case 0:
+		return Nop{}
+	case 1:
+		return nonNil[0]
```
zeroshade: A few minor notes here (all non-blocking):

- **Isolation asymmetry:** `Combine(r)` returns `r` directly, so a *single* reporter is not panic-isolated, whereas the composite path below (L124-129) recovers per reporter. Fine as an optimization, but worth a doc line that the panic safety-net only applies with 2+ reporters (or always wrap for uniform behavior).
- **Silent recover (L127):** `defer func() { _ = recover() }()` swallows a panicking reporter entirely. Per the `Reporter` contract that's intentional, but a fully-silent recover makes a broken reporter invisible — consider logging the recovered value at debug level.
- **FYI (not a change):** this is Phase 1 scaffolding — no concrete `Report` types or scan/commit instrumentation yet, so nothing produces reports in-tree until the later phases of #1236. Noting for anyone reading the package in isolation.

zeroshade (changes requested):
Really nice PR, @tanmayrauth — the package/symbol docs, `var _ Reporter` assertions, mutex-guarded `InMemoryReporter`, and the test suite (incl. concurrency + panic isolation) are all excellent.

Requesting changes only to **settle the public naming before it ships** (inline on `Report`) — renaming a published interface later is the painful kind of change, and the concrete `ScanReport`/`CommitReport` types land in follow-up phases. The other two inline notes are minor and non-blocking; take or leave them.

metrics/reporters.go
tanmayrauth: Renamed to NopReporter so the built-in set reads uniformly (NopReporter / LoggingReporter / InMemoryReporter), and updated the Combine zero-reporter case and docs to match.

metrics/reporter.go
tanmayrauth: Renamed the marker interface to MetricsReport and kept Reporter.Report(ctx, MetricsReport) as the verb, so signatures now read Report(context.Context, MetricsReport) with no type/method collision.  This also lines the Go API up with Java/PyIceberg's MetricsReport, which should ease the ScanReport/CommitReport ports in the later phases.

metrics/reporters.go
```diff
@@ -0,0 +1,141 @@
...
+		return NopReporter{}
+	case 1:
+		return nonNil[0]
+	default:
+		return compositeReporter(nonNil)
+	}
+}
+
+type compositeReporter []Reporter
+
```
laskoviymishka, line 163: `_ = recover()` drops the panic completely — no log, no type, no signal. The doc framing ("a misbehaving reporter never affects the observed operation") is right, but the contract is "don't propagate", not "stay silent". A reporter that nil-derefs in prod just shows up as mysteriously-missing metrics with nothing to grep for.

The rest of the codebase re-surfaces recovers (schema.go, visitors.go do `if r := recover(); r != nil { ... }`). I'd match that — at minimum `slog.Warn` the reporter type (`%T`) and the recovered value before continuing. To wire a logger in cleanly, `compositeReporter` could become a struct with a `*slog.Logger` field that `Combine` sets, rather than the bare slice.

metrics/reporters.go
```diff
@@ -0,0 +1,141 @@
...
+// if called directly, without the safety net.
+func Combine(reporters ...Reporter) Reporter {
+	nonNil := make([]Reporter, 0, len(reporters))
+	for _, r := range reporters {
+		if r != nil {
+			nonNil = append(nonNil, r)
+		}
+	}
+
+	switch len(nonNil) {
```
laskoviymishka: Typed-nil slips through here and defeats the isolation guarantee. `var lr *LoggingReporter = nil; Combine(lr)` passes the `r != nil` filter above (the interface is non-nil even though the concrete pointer is nil), hits this `case 1`, and gets returned unwrapped — so a later `.Report()` dereferences `r.logger` and panics with no `compositeReporter` recover to catch it. That's the exact failure mode the doc promises to prevent.

I'd always wrap in `compositeReporter`, even for one reporter — it drops the fast path but kills the whole bug class, and the cost is one range over a one-element slice. It also closes the related gap that `Combine(oneReporter)` doesn't currently get the isolation the doc advertises. If you'd rather keep the passthrough, the filter needs a reflect-based typed-nil check instead. wdyt?

metrics/reporters_test.go
```diff
@@ -0,0 +1,143 @@
...
+	require.Len(t, r.Reports(), 2)
+
+	r.Reset()
+	require.Empty(t, r.Reports())
+}
+
+func TestLoggingReporterNilLoggerAndReport(t *testing.T) {
+	r := NewLoggingReporter(nil) // must fall back to slog.Default
+	require.NotNil(t, r)
+	assert.NotPanics(t, func() {
```
laskoviymishka: This asserts only NotPanics — it'd still pass if the `InfoContext` call were deleted, logged at the wrong level, or dropped the report value. For the reporter we're calling the "convenient default for dev/debugging", a silently-broken log call is exactly what nobody would notice.

table_test.go has the prior art: swap in a `slog.NewTextHandler` over a `bytes.Buffer`, then assert the output contains "iceberg metrics report" and the report value, at INFO level.

metrics/reporters_test.go
```diff
@@ -0,0 +1,143 @@
...
+		after := &countingReporter{}
+		r := Combine(panickingReporter{}, after)
+		assert.NotPanics(t, func() {
+			r.Report(context.Background(), testReport{})
+		})
+		assert.Equal(t, 1, after.calls(), "reporters after a panic still run")
+	})
+}
+
+func TestInMemoryReporterConcurrent(t *testing.T) {
```
laskoviymishka, line 198: The whole point of this test is race detection, but `test-race` in the Makefile only covers `./codec/...`, not `./metrics/...` — so on a normal run this just checks that 100 goroutines finish, which they always will. I'd extend `test-race` to include `./metrics/...` in this PR, since the test exists for exactly that.

While here, the production pattern is concurrent `Report()` plus a foreground `Reports()`/`Reset()`. A goroutine looping on `Reports()` while the writers run would actually exercise the read-under-write path under `-race`.

tanmayrauth:
@zeroshade  can you PTAL when you get chance?

laskoviymishka (changes requested):
Nice foundation. The package docs are clear on the opt-in, non-blocking contract, and the tests are thoughtful — the white-box marker trick to build reports before the concrete types land, the copy-isolation check, the panic-isolation case. Good place to start the framework from.

I'd hold it for two things before merge: a correctness bug in `Combine`, and one contract decision on `Report` that's much cheaper to settle now than after this goes public. Both blockers and the two test gaps are inline.

The bug: a typed-nil reporter — `var lr *LoggingReporter = nil; Combine(lr)` — slips past the `r != nil` filter, takes the single-reporter fast path, and gets returned unwrapped, so the eventual `.Report()` nil-derefs with no recover to catch it. That's exactly the failure the contract promises to prevent. Always wrapping in `compositeReporter` kills the bug class and also closes the gap that `Combine(oneReporter)` doesn't currently get the isolation we advertise. Relatedly, the `compositeReporter` recover does `_ = recover()` and drops the panic entirely — all three reviewers flagged this independently. Non-blocking should mean "don't propagate", not "stay silent": a prod nil-deref would just look like missing metrics with nothing to grep. The rest of the codebase re-surfaces recovers (schema.go, visitors.go), and I'd match that.

The contract decision is the sealed `Report` marker. Java's `MetricsReport` is intentionally open — empty public interface, no marker — and it already has a third concrete type (`CacheMetricsReport`) beyond Scan/Commit, fired outside the scan/commit cycle. Sealing locks this package into defining every future report type itself and blocks downstream operators from implementing `Report` for their own wrappers. Since `Report` goes public the moment this merges, I'd rather decide it now: drop `isMetricsReport()`, keep the empty marker mirroring Java, and document the "controlled for now" intent in godoc instead of enforcing it structurally. On the same note, `Report(ctx, report Report)` reads awkwardly with the type and method sharing a name — renaming the type to `MetricsReport` (Java's name) clears that up.

Before merge:

- Fix the typed-nil hole in `Combine` (always-wrap is simplest).
- Make the `compositeReporter` panic observable instead of swallowed.
- Decide the sealed-marker question (I'd open it up) and the `Report`-type/method naming.
- Make the two tests actually assert what they're named for.

Smaller stuff I'd happily take as nits: `compositeReporter` is missing the `var _ Reporter` assertion the other types have; `NewLoggingReporter` snapshots `slog.Default()` at construction, so a later `slog.SetDefault()` (which this repo does in tests) is ignored; and `Reset()` isn't a barrier against in-flight `Report()` calls. The `initialize`/`Close` lifecycle Java's reporter has is fine to defer — worth capturing in #1236 before the REST integration lands, though.

Once the bug and the contract call are settled, happy to take another pass and approve.

metrics/reporters_test.go
tanmayrauth: Both done. Extended test-race in the Makefile to include ./metrics/..., and the test now runs a foreground goroutine looping on Reports() while the 100 writers run, so -race actually exercises the read-under-write path instead of  just confirming the goroutines finish. Verified locally with go test -race ./metrics/....

metrics/reporters_test.go
tanmayrauth: Splitted into two subtests: the nil-logger fallback keeps the NotPanics check, and a new subtest wires a slog.NewTextHandler over a bytes.Buffer and asserts the output contains level=INFO, iceberg metrics report, and the  report value — and that a nil report produces no output. It'll now fail if the log call is deleted, mis-leveled, or drops the report.

metrics/reporters.go
tanmayrauth: Agreed, went with always-wrap. Dropped the case 1 fast path entirely — Combine now returns NopReporter{} for zero non-nil reporters and a *compositeReporter otherwise, so even a lone reporter gets the isolation we advertise and   (*LoggingReporter)(nil) can no longer escape unwrapped to nil-deref later. The one extra range over a single-element slice is well worth closing the bug class. Added a test that drives a typed-nil pointer through Combine and  asserts it doesn't panic.

metrics/reporters.go
tanmayrauth: compositeReporter is now a struct carrying an optional *slog.Logger, and the recover logs at Warn with the offending reporter's %T plus the recovered value before continuing — matching the re-surface pattern in  schema.go/visitors.go. Still swallowed for propagation, just no longer silent.

tanmayrauth:
Thanks for the thorough pass @laskoviymishka . 

Addressed all four blockers: Combine always wraps now (typed-nil hole closed), panics re-surface at warn with the reporter %T, dropped the sealed marker + renamed to MetricsReport, and both tests   assert what they're named for (log output + read-under-write under -race, with test-race now covering ./metrics/...). Also took the easy nits — var _ Reporter assertion and lazy slog.Default resolution. Left Reset()-barrier and  the initialize/Close lifecycle for the #1236 follow-up. Ready for another look.

tanmayrauth:
@laskoviymishka  can you please review it when you get chance?

metrics/reporter.go
```diff
@@ -0,0 +1,64 @@
...
+//
+// The set of report types is controlled by this package for now, but that is a
+// convention documented here rather than a structural guarantee — the interface
+// is deliberately not sealed, leaving room for third-party reports without a
+// future breaking change.
+//
+// The name mirrors the MetricsReport type in the Java and Python Iceberg
+// implementations, easing ports across the ecosystem and disambiguating the
+// type from the [Reporter.Report] method that consumes it.
+type MetricsReport any
```
laskoviymishka, line 59: I'd make this a distinct named interface rather than an alias for `any` — `type MetricsReport interface{}`.

Right now `reporter.Report(ctx, 42)` compiles, and when the REST reporter lands it'll type-switch over an unrestricted interface with nothing to verify arm coverage. Java's is a real (empty) interface enforced by `instanceof`.

Zero cost to tighten now, and a more disruptive break once concrete types ship downstream. wdyt?

metrics/reporters.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	return &compositeReporter{reporters: nonNil}
+}
+
+// compositeReporter fans a report out to several reporters, isolating each from
+// the others' panics.
+type compositeReporter struct {
+	reporters []Reporter
+	logger    *slog.Logger // nil means resolve slog.Default at call time
```
laskoviymishka, line 159: This field is never set — `Combine` is the only constructor and it builds `&compositeReporter{reporters: nonNil}`, so `logger` is permanently nil and the nil-fallback in the recover only works because of that. It reads like configurable state that isn't.

I'd either drop the field and call `slog.Default()` directly in the recover, or add a `CombineWithLogger(logger, reporters...)` that actually wires it. Dropping it is simpler — unless you want the injectable logger for the panic-log test below, in which case they're worth deciding together.

metrics/reporters_test.go
```diff
@@ -0,0 +1,188 @@
...
+	})
+
+	t.Run("fans out to all", func(t *testing.T) {
+		a, b := &countingReporter{}, &countingReporter{}
+		Combine(a, b).Report(context.Background(), testReport{})
+		assert.Equal(t, 1, a.calls())
+		assert.Equal(t, 1, b.calls())
+	})
+
+	t.Run("isolates a panicking reporter", func(t *testing.T) {
```
laskoviymishka, line 169: This proves the panic is isolated and the next reporter runs, but it doesn't cover the warn-log path — nothing asserts `WarnContext` fired, or that the output carries the reporter type and the panic value. If a future change drops the `WarnContext` call, this stays green.

Same trick as the LoggingReporter test would do it: build the composite with a buffer-backed logger and assert the output contains "panicked", the type name, and "boom". That's currently blocked by the un-injectable logger field above, so worth deciding the two together.

laskoviymishka (changes requested):
Very nice — this is in good shape. The panic isolation in `Combine`, typed-nil handling, concurrent `Reports()` test, and godocs all look solid. This should be the last round.

Two things before merge:

* `compositeReporter.logger` is never set. Since `Combine` is the only constructor, it stays nil forever, and the recovery fallback only works because of that. I’d either drop the field and use `slog.Default()` directly, or add a `CombineWithLogger` path and test that the warning is actually emitted.
* `MetricsReport` is declared as `interface{}` rather than an alias to `any`, so future type switches won’t get compiler help. Cheap to fix now, harder once concrete report types build on it.

Fix those two and I’m happy to approve.

tanmayrauth:
I've made the changes @laskoviymishka, but two things: type MetricsReport any is already a defined type, not an alias, an alias would need = any, so it's already distinct from any for type-identity purposes. And our own .golangci.yml has a gofmt  rewrite rule interface{} -> any, so interface{} gets formatted straight back to any on commit.

The deeper point — neither spelling restricts what satisfies it: Report(ctx, 42) compiles either way, since any value satisfies an empty interface, and type switches get no compiler-enforced arm coverage from an open interface  regardless of spelling. If we want that guarantee when the REST reporter lands, the real lever is sealing the interface with an unexported marker method (isMetricsReport()), which the concrete types implement. That's a deliberate call though — it bars third-party report types, which the current godoc intentionally allows. Happy to seal it if you'd prefer closed; otherwise I think any is the honest representation of an open marker.

zeroshade (approved):
Re-reviewed after the follow-ups. Public naming is settled (`MetricsReport` / `NopReporter`), `Combine` now always wraps in a composite with panic isolation + warn-level logging of recovered panics, and the concurrency test is wired into `make test-race`. My earlier change request is addressed — LGTM.

tanmayrauth:
@laskoviymishka can you please review it?

metrics/reporter.go
```diff
@@ -0,0 +1,64 @@
...
+import "context"
+
+// MetricsReport is the marker interface implemented by the concrete report
+// types (ScanReport and CommitReport). It is intentionally empty, mirroring the
+// open MetricsReport interface in the Java and Python Iceberg implementations,
+// so downstream operators can implement it for their own report wrappers.
+//
+// The set of report types is controlled by this package for now, but that is a
+// convention documented here rather than a structural guarantee — the interface
+// is deliberately not sealed, leaving room for third-party reports without a
```
laskoviymishka: The "without a future breaking change" framing reads backwards — keeping the marker open is the irreversible call; a future attempt to *seal* it is what would break.

And the extensibility it advertises won't reach the REST path: once the REST reporter lands it can only serialize the discriminators it knows (`scan-report`/`commit-report`), so a third-party `MetricsReport` has no discriminator and a spec-compliant server rejects it. I'd soften this to say the marker is open for the local reporters but that custom report types are silently dropped by the REST reporter — open extensibility for `Logging`/`InMemory`, not for REST. Not relitigating the open decision (still think it's right); I just don't want the doc to over-promise. wdyt?

metrics/reporters.go
```diff
@@ -0,0 +1,166 @@
...
+// NewLoggingReporter returns a [LoggingReporter] that logs to logger. If logger
+// is nil, [slog.Default] is resolved at each Report call, so a later
+// [slog.SetDefault] is honored rather than snapshotted at construction.
+func NewLoggingReporter(logger *slog.Logger) *LoggingReporter {
+	return &LoggingReporter{logger: logger}
+}
+
+// Report logs report at info level.
+func (r *LoggingReporter) Report(ctx context.Context, report MetricsReport) {
+	if report == nil {
```
laskoviymishka: One concrete edge the open `any` introduces: this only catches an untyped nil. Once `ScanReport`/`CommitReport` arrive as pointer types, a `(*ScanReport)(nil)` is a non-nil interface and slips past — `LoggingReporter` logs a typed-nil (and nil-derefs if the concrete type ever grows a `slog.LogValuer` on a nil receiver), and `InMemoryReporter` below silently stores it, which is the opposite of the "nil reports are ignored" the guard advertises.

Latent today, but cheap to close before those types land — either a doc line that only untyped nil is skipped, or a reflect `Kind()==Ptr && IsNil()` check.

metrics/reporter.go
```diff
@@ -0,0 +1,64 @@
...
+// bytes read; commit attempts and durations). Reporting gives operators a
+// standard way to aggregate these otherwise-invisible client-side metrics
+// across many clients.
+//
+// Reporting is strictly opt-in: with no reporter configured the instrumented
+// code paths do no work. Reporters must never block or fail the scan/commit
+// they observe — see [Reporter] for the contract.
+//
+// This package provides the contract and the built-in reporters
+// ([NopReporter], [LoggingReporter], [InMemoryReporter], and [Combine]). The
```
laskoviymishka: The overview lists `Combine` but not `CombineWithLogger`, which landed in the same round — worth adding it here so the package doc matches the exported surface.

metrics/reporters.go
```diff
@@ -0,0 +1,166 @@
...
+				if v := recover(); v != nil {
+					// Swallow per the Reporter contract, but surface the
+					// failure (with the offending reporter type) so a broken
+					// reporter is traceable rather than showing up only as
+					// mysteriously-missing metrics.
+					logger := c.logger
+					if logger == nil {
+						logger = slog.Default()
+					}
+					logger.WarnContext(ctx, "iceberg metrics reporter panicked; recovered",
```
laskoviymishka, line 177: Now that we warn on a recovered reporter panic, logging the stack too would make a prod nil-deref actually locatable — right now an operator gets the value and the type but no call site, which is the one thing they'd need to find it.

`slog.String("stack", string(debug.Stack()))` here (`debug` = `runtime/debug`) covers it. Take or leave.

laskoviymishka (changes requested):
Thanks for the follow-ups. This addresses my earlier rounds: `CombineWithLogger` wires the logger cleanly, recovered panics now warn with the reporter type, `MetricsReport` is a proper named marker, and `test-race` actually exercises `./metrics/...`.

One thing before approval: the marker godoc still frames openness the wrong way.

This isn’t reopening open vs sealed. I still think open is the right choice for a client library. The issue is narrower: the doc says openness leaves room for third-party reports without a future breaking change, but openness is the irreversible choice. Sealing it later would be the breaking change.

It also won’t work the same across reporters. Local reporters can accept custom `MetricsReport` types, but once the REST reporter lands, those reports won’t have a valid scan-report or commit-report discriminator. They’ll be dropped locally or rejected by the server. So it’s open for local reporters, but effectively closed for REST.

I’d reword the godoc to make that boundary clear, so nobody builds a custom report expecting it to reach a catalog endpoint.

The rest is non-blocking: the typed-nil `report == nil` edge, package docs missing `CombineWithLogger`, and logging `debug.Stack()` on panic. `io.Closer` can still wait for #1236, but should be settled before REST integration.

Reword the marker godoc and I’m happy to merge.

metrics/reporter.go
tanmayrauth: Yeah, the framing was backwards. Reworded the doc now says keeping the marker open is the irreversible choice (sealing it later is what would break), and it's explicit that custom report types are honored only by the in-process reporters (LoggingReporter, InMemoryReporter, Combine). The REST reporter,  once it lands, can serialize only the scan-report/commit-report discriminators, so a third-party MetricsReport has no discriminator and is silently dropped there — a spec-compliant catalog rejects it. Added the "don't build a custom report expecting it to reach a catalog endpoint" caveat. Not relitigating open   vs sealed — agreed it's right for a client lib.

metrics/reporters.go
tanmayrauth: Closed it rather than just documenting. Added an isNilReport helper (reflect Kind()==Ptr && IsNil()) now used by both LoggingReporter and InMemoryReporter, so a (*ScanReport)(nil) is treated as "no report" instead of being logged/stored (and it avoids the nil-receiver LogValuer deref you flagged). Added  TestReportersIgnoreTypedNil covering both reporters.

metrics/reporter.go
tanmayrauth: the overview now lists NopReporter, LoggingReporter, InMemoryReporter, Combine, and CombineWithLogger, matching the exported surface.

metrics/reporters.go
tanmayrauth: The warn log now includes slog.String("stack", string(debug.Stack())) alongside the reporter type and panic value, so an operator gets the call site. Added a stack= assertion to the recovered-panic test.

laskoviymishka (approved):
🦵 ⚽ 🥅 👍

PR #1345, opened 2026-06-29
fix(rest): require absolute auth URL

fallintoplace:
Summary

- follow-up to #1336
- require rest.authorization-url to include a scheme and host after url.Parse succeeds
- extract authorization URL parsing into a testable helper
- extend regression coverage for malformed local props, missing scheme/host, valid local config, and server overrides

Why

#1336 made malformed rest.authorization-url values fail instead of being silently ignored. url.Parse still accepts common typos like example.com/auth, localhost:8080, and the empty string, though. Those values made authUri non-nil, so credential-based auth skipped the default OAuth endpoint and pointed at a bad authorization URL instead of rejecting the config.

Testing

- go test ./catalog/rest -run 'Test(LoadRegisteredCatalogRejectsInvalidAuthURL|LoadRegisteredCatalogAcceptsValidAuthURL|NewCatalogRejectsInvalidAuthURLFromConfig|NewCatalogAcceptsValidAuthURLFromConfig)$' -count=1
- go test ./catalog/rest -count=1
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m ./catalog/rest/...
- git diff --check

### DIFF
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -449,6 +449,14 @@ func handleNon200(rsp *http.Response, override map[int]error) error {
 }
 
 func fromProps(props iceberg.Properties, o *options) error {
+	if v, ok := props[keyAuthUrl]; ok {
+		u, err := parseAuthURL(v)
+		if err != nil {
+			return err
+		}
+		o.authUri = u
+	}
+
 	for k, v := range props {
 		switch k {
 		case keyWarehouseLocation:
@@ -462,11 +470,6 @@ func fromProps(props iceberg.Properties, o *options) error {
 		case keyRestSigV4Service:
 			o.sigv4Service = v
 		case keyAuthUrl:
-			u, err := url.Parse(v)
-			if err != nil {
-				return fmt.Errorf("invalid %s %q: %w", keyAuthUrl, v, err)
-			}
-			o.authUri = u
 		case keyOauthCredential:
 			o.credential = v
 		case keyScope:
@@ -498,6 +501,18 @@ func fromProps(props iceberg.Properties, o *options) error {
 	}
 
 	return nil
+}
+
+func parseAuthURL(raw string) (*url.URL, error) {
+	u, err := url.Parse(raw)
+	if err != nil {
+		return nil, fmt.Errorf("invalid %s %q: %w", keyAuthUrl, raw, err)
+	}
+	if u.Scheme == "" || u.Host == "" {
+		return nil, fmt.Errorf("invalid %s %q: missing scheme or host", keyAuthUrl, raw)
+	}
+
+	return u, nil
 }
 
 func toProps(o *options) iceberg.Properties {
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -449,14 +449,6 @@ func handleNon200(rsp *http.Response, override map[int]error) error {
 }
 
 func fromProps(props iceberg.Properties, o *options) error {
-	if v, ok := props[keyAuthUrl]; ok {
-		u, err := parseAuthURL(v)
-		if err != nil {
-			return err
-		}
-		o.authUri = u
-	}
-
 	for k, v := range props {
 		switch k {
 		case keyWarehouseLocation:
@@ -470,6 +462,11 @@ func fromProps(props iceberg.Properties, o *options) error {
 		case keyRestSigV4Service:
 			o.sigv4Service = v
 		case keyAuthUrl:
+			u, err := parseAuthURL(v)
+			if err != nil {
+				return err
+			}
+			o.authUri = u
 		case keyOauthCredential:
 			o.credential = v
 		case keyScope:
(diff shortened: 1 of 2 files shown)


catalog/rest/rest.go
```diff
@@ -462,11 +470,6 @@ func fromProps(props iceberg.Properties, o *options) error {
 		case keyRestSigV4Service:
 			o.sigv4Service = v
 		case keyAuthUrl:
```
laskoviymishka, line 464: This empty `case keyAuthUrl:` reads like dead code, but it's load-bearing — it's the only thing keeping the auth URL from falling into `default` and getting pushed into `additionalProps`. Someone skimming will see every other key with a one-line assignment and "clean this up" by deleting it, which silently regresses the URL into `additionalProps`.

I'd just fold `parseAuthURL` back into the case here instead of hoisting it above the loop:

```go
case keyAuthUrl:
    u, err := parseAuthURL(v)
    if err != nil {
        return err
    }
    o.authUri = u
```

That keeps the testable-helper win, drops the pre-loop lookup, and the case is self-evident again. wdyt?

catalog/rest/rest_internal_test.go
```diff
@@ -54,35 +54,101 @@ import (
...
 	t.Parallel()
 
-	cat, err := catalog.Load(context.Background(), "rest", iceberg.Properties{
-		"uri":                    "http://example.com",
-		"rest.authorization-url": "http://[::1",
-	})
-	require.Error(t, err)
-	assert.Nil(t, cat)
-	assert.ErrorContains(t, err, "invalid rest.authorization-url")
+	tests := []struct {
```
laskoviymishka, line 57: Every case in both tables asserts an error — there's no positive case proving a well-formed `rest.authorization-url` (say `https://auth.example.com/oauth/token`) still parses and lands on `o.authUri` as the token endpoint.

The thing we actually shipped is "reject broken, keep valid", and right now a regression that over-tightens the check — rejecting good URLs too — wouldn't be caught by anything here. I'd add a success case asserting `Load`/`NewCatalog` succeeds and the URL ends up where it should.

catalog/rest/rest_internal_test.go
```diff
@@ -54,35 +54,101 @@ import (
...
+			authURL: "http://[::1",
+			wantErr: "invalid rest.authorization-url",
+		},
+		{
+			name:    "missing scheme",
+			authURL: "example.com/auth",
+			wantErr: "missing scheme or host",
+		},
+		{
+			name:    "missing host",
```
laskoviymishka: `url.Parse("localhost:8080")` parses `localhost` as the scheme and `8080` as `Opaque` — so `Host` is empty and the assertion passes, but for a different reason than "missing host" implies. There's no case that actually exercises a valid scheme with a genuinely empty host.

I'd rename this to something like `"scheme+opaque, no host"` and add a real `"http://"` (or `"http:///path"`) case to cover the true empty-host-with-valid-scheme path.

catalog/rest/rest.go
```diff
@@ -500,6 +503,18 @@ func fromProps(props iceberg.Properties, o *options) error {
 	return nil
 }
 
+func parseAuthURL(raw string) (*url.URL, error) {
```
laskoviymishka, line 503: Two small things here, neither blocking. The `"missing scheme or host"` message doesn't say which one is actually missing — a touch more precision would help whoever's debugging a bad config.

And this only covers the config-string path; `WithAuthURI` (options.go) still lets a caller set a relative `*url.URL` directly, and `setupOAuthManager` calls `.String()` on it unconditionally. Scoping this PR to the config path is reasonable, but the invariant "if `authUri != nil` it's absolute" isn't enforced catalog-wide. Fine to leave for a follow-up — flagging so it's a deliberate choice.

laskoviymishka (changes requested):
The fix itself is right and nicely scoped — rejecting `Scheme == "" || Host == ""` closes a real bug where a non-nil-but-useless `*url.URL` silently became the OAuth token endpoint, and pulling the parse into `parseAuthURL` makes it testable. The table-driven conversion is a nice breadth bump too.

Just a couple of small hygiene things before merge, nothing structural. The empty `case keyAuthUrl:` left in the switch reads like dead code but is actually load-bearing — folding the helper call back into the case (instead of hoisting it above the loop) kills the smell and drops the redundant lookup. And the test table only asserts errors, so I'd add one positive case proving a valid absolute URL still lands on `o.authUri` and an over-tightening regression gets caught.

One nit on the description: the "validate up front so invalid config fails deterministically" line doesn't match a real bug — `o` is discarded on error at every call site and the old inline `url.Parse` already returned immediately, so there's no map-iteration nondeterminism being fixed. I'd reword to just "extract validation into a testable helper" so nobody hunts for a bug that isn't there.

Clean these up and it's good to land.

laskoviymishka (approved):
:shipit:

PR #1347, opened 2026-06-30
fix(table): reject float equality delete fields

### LINKED ISSUE #1288: Validate equality-delete field IDs reject float and double
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Iceberg equality delete field IDs should not include float or double columns. Today iceberg-go appears to accept equality delete files/row deltas with float or double equality IDs instead of rejecting them when the delete file is written or committed.

That can make callers or tests appear to support a delete shape other engines would reject.

Expected:
- equality delete write/commit paths reject float/double equality field IDs
- the error identifies the field and type
- writer/row-delta coverage pins this behavior

Context: this came up while fixing #1242. Encoder-level float/double coverage is still useful, but scan-level equality-delete round trips should not use float/double keys.

cc: @laskoviymishka

fallintoplace:
Summary

- reject float and double columns in equality delete field IDs
- share equality-ID schema validation between WriteEqualityDeletes and RowDelta commit
- add writer and RowDelta regression coverage for both float and double fields
- cover the partitioned equality-delete writer path with float rejection and decimal acceptance

Why

Iceberg equality delete keys cannot be float or double. Before this change, WriteEqualityDeletes and manually-built equality delete DataFiles could accept those field IDs, which made iceberg-go appear to support delete shapes other engines reject.

Closes #1288.

Testing

- go test ./table -run 'Test(WriteEqualityDeleteFilesRejectsFloatAndDoubleFieldIDs|WriteEqualityDeleteFilesPartitionedTableRejectsFloatFieldID|WriteEqualityDeleteFilesPartitionedTableAcceptsDecimalFieldID|WriteEqualityDeleteFilesPartitionedTable|WriteEqualityDeleteFilesRejectsInvalidFieldID|WriteEqualityDeleteFilesRejectsEmptyFieldIDs|RowDeltaRejectsEqDeleteWithFloatAndDoubleFieldIDs|RowDeltaRejectsEqDeleteWithInvalidFieldID|RowDeltaRejectsEqDeleteWithoutFieldIDs|RowDeltaCommitWithEqualityDeletes)$' -count=1
- go test ./table -count=1
- go test ./table/... -count=1
- git diff --check

### DIFF
--- a/table/equality_delete_writer.go
+++ b/table/equality_delete_writer.go
@@ -55,12 +55,15 @@ func validateEqualityFieldIDs(tableSchema *iceberg.Schema, fieldIDs []int) ([]st
 			return nil, fmt.Errorf("%w: field ID %d not found in table schema", iceberg.ErrInvalidSchema, id)
 		}
 
+		// Keep both lookups: FindColumnName returns the dotted path Schema.Select
+		// needs for nested fields, while FindFieldByID exposes the field type.
 		field, ok := tableSchema.FindFieldByID(id)
 		if !ok {
+			// Defensive only: FindColumnName already found this field ID.
 			return nil, fmt.Errorf("%w: field ID %d not found in table schema", iceberg.ErrInvalidSchema, id)
 		}
 		if isFloatingPointType(field.Type) {
-			return nil, fmt.Errorf("%w: equality field ID %d (%s) has unsupported type %s: float and double cannot be used in equality delete field IDs",
+			return nil, fmt.Errorf("%w: equality field ID %d (%s) has unsupported floating-point type %s: floating-point columns cannot be used as equality delete keys",
 				iceberg.ErrInvalidSchema, id, name, field.Type)
 		}
 
--- a/table/equality_delete_writer.go
+++ b/table/equality_delete_writer.go
@@ -35,6 +35,15 @@ var ErrEmptyEqualityFieldIDs = errors.New("equality field IDs must not be empty"
 // equalityDeleteSchema projects a table schema to only the fields specified
 // by the given field IDs, using Schema.Select for proper nested field handling.
 func equalityDeleteSchema(tableSchema *iceberg.Schema, fieldIDs []int) (*iceberg.Schema, error) {
+	names, err := validateEqualityFieldIDs(tableSchema, fieldIDs)
+	if err != nil {
+		return nil, err
+	}
+
+	return tableSchema.Select(true, names...)
+}
+
+func validateEqualityFieldIDs(tableSchema *iceberg.Schema, fieldIDs []int) ([]string, error) {
 	if len(fieldIDs) == 0 {
 		return nil, ErrEmptyEqualityFieldIDs
 	}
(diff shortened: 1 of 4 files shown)


table/equality_delete_writer.go
```diff
@@ -46,10 +55,28 @@ func equalityDeleteSchema(tableSchema *iceberg.Schema, fieldIDs []int) (*iceberg
 			return nil, fmt.Errorf("%w: field ID %d not found in table schema", iceberg.ErrInvalidSchema, id)
 		}
 
+		field, ok := tableSchema.FindFieldByID(id)
```
laskoviymishka, line 60: This lookup pair looks redundant but isn't — worth a comment so nobody "simplifies" it.

`FindColumnName(id)` above returns the dotted path (`address.zipcode`) that `Schema.Select` needs to project nested fields; `FindFieldByID(id).Name` only gives the bare short name (`zipcode`), so it can't replace it. We keep `FindColumnName` for the projection name and `FindFieldByID` for the `Type`. Dropping the former would silently break nested-field equality deletes — a one-liner here (and noting the `!ok` branch is unreachable after `FindColumnName` already succeeded) prevents that.

table/equality_delete_writer.go
```diff
@@ -46,10 +55,28 @@ func equalityDeleteSchema(tableSchema *iceberg.Schema, fieldIDs []int) (*iceberg
...
+		if !ok {
+			return nil, fmt.Errorf("%w: field ID %d not found in table schema", iceberg.ErrInvalidSchema, id)
+		}
+		if isFloatingPointType(field.Type) {
+			return nil, fmt.Errorf("%w: equality field ID %d (%s) has unsupported type %s: float and double cannot be used in equality delete field IDs",
```
laskoviymishka: For a float field this reads "...has unsupported type float: float and double cannot be used..." — the "double" in the static suffix is noise when only float triggered. I'd tighten to something like "...has unsupported floating-point type %s; floating-point columns cannot be used as equality delete keys". The test asserts on the `float`/`double` substring via the `field.Type` interpolation, so that still holds.

table/equality_delete_writer_test.go
```diff
@@ -226,6 +226,53 @@ func TestWriteEqualityDeleteFilesRejectsInvalidFieldID(t *testing.T) {
...
+
+	records := func(yield func(arrow.RecordBatch, error) bool) {}
+
+	tests := []struct {
+		name      string
+		fieldID   int
+		fieldName string
+		fieldType string
+	}{
+		{name: "float", fieldID: 2, fieldName: "score", fieldType: "float"},
```
laskoviymishka, line 260: The most valuable test gap to close: nothing exercises `equalityDeleteWriteSchema` (the partitioned path) with a float key end-to-end — this unpartitioned table only hits `equalityDeleteSchema`, and the refactor changed `equalityDeleteWriteSchema`'s shape the most. A partitioned-table case rejecting a float key would cover it. While the scaffold's right here, a decimal field asserted as *accepted* is a cheap negative control that pins `isFloatingPointType`'s `default: false` so nobody later conflates "non-integer numeric" with "floating point."

table/row_delta.go
```diff
@@ -134,11 +134,8 @@ func (rd *RowDelta) Commit(ctx context.Context) error {
 					f.FilePath())
 			}
 
-			for _, id := range eqIDs {
-				if _, ok := schema.FindFieldByID(id); !ok {
-					return fmt.Errorf("equality field ID %d not found in table schema: %s",
-						id, f.FilePath())
-				}
+			if _, err := validateEqualityFieldIDs(schema, eqIDs); err != nil {
+				return fmt.Errorf("invalid equality delete file %s: %w", f.FilePath(), err)
```
laskoviymishka, line 138: Good catch wrapping with `%w` — the old "not found" error was unwrapped, so `errors.Is(err, ErrInvalidSchema)` now succeeds and the "not found in table schema" substring is preserved. Just flagging that this commit-path wording ("invalid equality delete file %s") differs from the writer call sites, and the two test files re-derive the expected substrings independently — if the float message changes per the other comment, both need updating in lockstep.

laskoviymishka (changes requested):
Nice, focused fix — routing all three enforcement points through a single `validateEqualityFieldIDs` is exactly right, and that shared validator is what prevents the drift that let #1288 happen (`RowDelta.Commit` had its own inlined existence check that never grew a type guard). Float/double rejection is correct, decimals are correctly left alone, and the `%w` wrap to `ErrInvalidSchema` is a strict improvement.

Almost there — one blocking thing: the `FindColumnName` + `FindFieldByID` pair in `validateEqualityFieldIDs` needs a comment. It looks like a redundant double lookup, but `FindColumnName` returns the dotted path `Schema.Select` needs for nested fields while `FindFieldByID(id).Name` only gives the bare name — dropping the former would silently break nested-field equality deletes. A one-liner explaining dotted-vs-bare closes it.

Rest is nits inline. One follow-up worth a tracked issue rather than this PR: the spec also restricts equality-delete keys nested in maps/lists and non-primitive struct/list/map columns (iceberg-rust enforces this at write time), so closing #1288 isn't quite "equality-delete validation is now spec-complete." Add the comment and this is good to land.

laskoviymishka (approved):
👍

PR #1348, opened 2026-06-30
fix(table): clarify v3 next-row-id overflow guard, harden tests

laskoviymishka:
Follow-up to #1054. cleanup the next-row-id overflow guard in validateAndUpdateRowLineage to check overflow before computing the wrapped sum (added-rows > MaxInt64 - next-row-id) and reword the error to say what actually happened ("adding N rows to next-row-id M overflows int64") rather than the misleading "regressed". Document the invariants the guard relies on: added-rows is non-nil/non-negative and next-row-id is non-negative at that point.

Tests: drop the redundant positive-advance case (covered by TestAddSnapshotV3UpdatesNextRowID), make the zero-added-rows test falsifiable by seeding a non-zero cursor first, and add the MaxInt64 accept boundary alongside the overflow reject.

### DIFF
--- a/table/metadata_builder_internal_test.go
+++ b/table/metadata_builder_internal_test.go
@@ -20,6 +20,7 @@ package table
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"slices"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


table/metadata_builder_internal_test.go
```diff
@@ -1702,13 +1703,61 @@ func TestAddSnapshotV3AcceptsFirstRowIDEqualToNextRowID(t *testing.T) {
...
+	// the maximum so a positive added-rows tips it over. This is the only input
+	// that reaches the guard: negative added-rows is rejected earlier by
+	// Snapshot.ValidateRowLineage.
+	builder := builderWithoutChanges(3)
+	nearMax := int64(math.MaxInt64 - 10)
+	builder.nextRowID = &nearMax
+
+	schemaID := 0
+	firstRowID := nearMax
+	addedRows := int64(20)
```
zeroshade, line 1760: Nit (optional): this rejects at `MaxInt64 - 10 + 20` = `MaxInt64 + 10`. To pin the exact boundary against a future off-by-one, consider `addedRows := int64(11)` (sum == `MaxInt64 + 1`) for the reject case, plus a `MaxInt64 - 1` accept case.

zeroshade (approved):
LGTM. Moving the overflow check ahead of the addition is the correct fix, and the added zero-added-rows and accept-at-`math.MaxInt64` boundary cases are good. One optional test-tightening nit inline.

PR #1350, opened 2026-07-01
feat(catalog/hadoop): Add support for hadoop catalog fs interface methods to blob filesystem

### LINKED ISSUE #1096: feat(catalog): Cloud storage backend support for hadoop

C-Loftus:
Last PR needed for #1096 ; closes #1096

- Move io interfaces needed for HadoopCatalogFS into the `io.go` file since they are now presumed to be relatively stable.
- Add implementations for any methods not previously fulfilled for the blob storage struct in order to have it fulfill HadoopCatalogFS
- Add an integration test for using the hadoop catalog with minio locally


C-Loftus:
Marking this as draft until I have confirmation all the integration tests pass in the github workflows CI. Then may go back and tidy a few things up and will ping maintainers for review

catalog/hadoop/io.go
```diff
@@ -29,59 +28,15 @@ type HadoopCatalogFS interface {
...
-type MkdirAllIO interface {
-	icebergio.IO
-
-	// MkdirAll should not raise an error if the directory already exists,
-	// and should create any parent directories in the path if they
-	// do not already exist.
-	MkdirAll(path string) error
-}
+// BlobFileIO can be used to implement a Hadoop catalog with a blob storage bucket.
+var _ HadoopCatalogFS = (*gocloud.BlobFileIO)(nil)
```
C-Loftus, line 42: I have made `BlobFileIO` a public export now so that we can have a compile time interface assertion in the `hadoop` package.

io/gocloud/blob.go
```diff
@@ -256,10 +319,178 @@ func (bfs *blobFileIO) DeleteFiles(ctx context.Context, paths []string) ([]strin
 	return deleted, errs
 }
 
+// MkdirAll mimics creating a directory by creating a zero-length object for each component of the path
+func (bfs *BlobFileIO) MkdirAll(path string) error {
```
C-Loftus, line 459: There was a fair bit of discussion in #1096 regarding how to implement MkdirAll in a way that makes sense for blob stores that don't have a notion of a directory. Worth reviewing this function a bit closer to ensure that even if it passes all tests, it performs as desired.

C-Loftus (commented):
At your convenience this should be ready to review @zeroshade , thanks!

catalog/hadoop/hadoop_minio_integration_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	var tables []table.Identifier
+	for found, err := range s.cat.ListTables(s.ctx, ns) {
+		s.Require().NoError(err)
+		tables = append(tables, found)
+	}
+	s.Empty(tables)
+}
+
+func TestHadoopMinIOIntegration(t *testing.T) {
```
tanmayrauth: The repo's `integration-hadoop` make target runs `-run="^TestHadoopIntegration"`, and this name (TestHadoopMinIOIntegration) doesn't match that anchored pattern — it diverges at "Hadoop*M*inIO". So this whole suite is skipped in CI and the green checks aren't exercising any of the new blob layout. Could we rename this to TestHadoopIntegrationMinIO so the existing target picks it up (or otherwise wire it into the integration run)? Everything else here really wants this suite executing.

io/gocloud/blob.go
```diff
@@ -256,10 +319,178 @@ func (bfs *blobFileIO) DeleteFiles(ctx context.Context, paths []string) ([]strin
...
+
+	if err := bfs.Delete(bfs.ctx, oldKey); err != nil && gcerrors.Code(err) != gcerrors.NotFound {
+		return blobErrToFsErr("Rename", oldpath, err)
+	}
+
+	return nil
+}
+
+// RenameNoReplace renames a file or directory from oldpath to newpath, returning an error if newpath already exists.
+func (bfs *BlobFileIO) RenameNoReplace(oldpath, newpath string) error {
```
tanmayrauth, line 594: Just noting for the record since this is the commit primitive: this is Stat-then-Copy+Delete, so two concurrent writers can both pass the Stat check and both Copy — the losing commit is silently dropped rather than rejected. That's correctly gated behind allow-unsafe-commits with a fail-closed default, so no change needed here. One precision worth capturing in the follow-up tracking issue:  the hazard is concurrent lost-update, not single-writer crash corruption (S3 CopyObject is atomic at the destination). The real fix is a conditional-write primitive (S3 If-None-Match:*, GCS if-generation-match:0). Is that follow-up issue tracked anywhere yet?

io/gocloud/blob.go
```diff
@@ -256,10 +319,178 @@ func (bfs *blobFileIO) DeleteFiles(ctx context.Context, paths []string) ([]strin
...
+			sys:  obj,
+		}, nil
+	} else if listErr != io.EOF {
+		return nil, blobErrToFsErr("Stat", path, listErr)
+	}
+
+	return nil, &fs.PathError{Op: "Stat", Path: path, Err: fs.ErrNotExist}
+}
+
+// Rename renames a file or directory from oldpath to newpath, replacing newpath if it already exists.
```
tanmayrauth: Doc says "renames a file or directory", but the body is a single Copy(newKey, oldKey) — with the marker layout, a directory's children (and its own marker) wouldn't move. Current callers only rename single files so it's fine in practice; can we tighten the comment to single-object rename to avoid surprising a future caller?

io/gocloud/blob.go
```diff
@@ -256,10 +319,178 @@ func (bfs *blobFileIO) DeleteFiles(ctx context.Context, paths []string) ([]strin
...
+	key, err := bfs.preprocess(path)
+	if err != nil {
+		return nil, &fs.PathError{Op: "Stat", Path: path, Err: err}
+	}
+
+	attrs, err := bfs.Attributes(bfs.ctx, key)
+	// if there is no error and we have attributes, we can return a FileInfo for object
+	if err == nil {
+		return blobFileInfo{
+			name:    filepath.Base(key),
```
tanmayrauth: filepath.Base here vs path.Base in directoryName for the dir case — object keys are always  '/'-separated, so filepath.Base won't split them on Windows. Suggest path.Base for consistency.

io/gocloud/blob.go
```diff
@@ -102,13 +104,65 @@ type blobFileIO struct {
...
+func (bfs *BlobFileIO) preprocess(path string) (string, error) {
 	return bfs.keyExtractor(path)
 }
 
-func (bfs *blobFileIO) Open(path string) (icebergio.File, error) {
+// blobErrToFsErr converts a gocloud blob error to an error from the fs package; this is
+// necessary for comply with certain iceberg-go/io interface functions that expect fs.ErrNotExist
+// for missing files, rather than the gocloud error. If there is no corresponding fs error, or the
+// mapped error wouldn't be needed, the original error is returned.
+func blobErrToFsErr(op, name string, err error) error {
```
tanmayrauth, line 222: Small doc mismatch: the comment says the original error is returned when there's no fs mapping, but this always wraps in *fs.PathError. Behavior's fine (errors.Is still works through the wrap) — just the comment.

tanmayrauth (commented):
Nice work, clean tail of #1096, and the read path degrading gracefully under a bad
  version-hint is what makes the blob layout tolerable. One blocker: the new minio suite won't be picked up by the repo's integration make target, so none of the new behavior runs in CI. Beyond that, a design question on the MkdirAll markers and a few nits. Details inline, plus one cross-engine note below that isn't tied to a changed line.

io/gocloud/blob.go
C-Loftus: Makes sense

>  Is that follow-up issue tracked anywhere yet?

My understanding is that #1096 is currently scoped to baseline hadoop catalog blob store support and that there is not an existing issue for conditional writes. I think it probably makes sense to make that once #1096  is closed.

C-Loftus:
@tanmayrauth thank you very much for your feedback. I believe I have addressed it and at your convenience, it should be ready to review again. (Accidentally changed the test suite struct name instead of the test itself at first, but should be all set now!)

io/gocloud/blob.go
```diff
@@ -256,10 +317,178 @@ func (bfs *blobFileIO) DeleteFiles(ctx context.Context, paths []string) ([]strin
...
+	if err := bfs.Delete(bfs.ctx, oldKey); err != nil && gcerrors.Code(err) != gcerrors.NotFound {
+		return blobErrToFsErr("Rename", oldpath, err)
+	}
+
+	return nil
+}
+
+// RenameNoReplace renames a file or directory from oldpath to newpath, returning an error if newpath already exists.
+func (bfs *BlobFileIO) RenameNoReplace(oldpath, newpath string) error {
+	if _, err := bfs.Stat(newpath); err == nil {
```
laskoviymishka, line 595: This is the one thing I'd want settled before merge. RenameNoReplace here is a check-then-act — Stat(newpath), then fall through to Copy+Delete — with no atomicity between the two steps, but RenameNoReplaceIO's godoc (io/io.go:190) promises the move happens "atomically" and the hadoop commit path leans on that. Two concurrent commits writing the same v1.metadata.json both pass the Stat("not found") check, both Copy, and the second silently overwrites the first — the losing writer's snapshot chain is gone. And commitMetadataFile's fs.ErrExist branch never fires, because nothing on the blob side ever returns it.

The `allow-unsafe-commits` gate is the right instinct and does mean nobody hits this without opting in, but the interface doc still over-promises for blob backends. I'd either soften the RenameIO/RenameNoReplaceIO godoc to state blob implementations aren't atomic and make the `allow-unsafe-commits` error spell out the last-writer-wins risk, or enforce the no-replace server-side with a conditional put (If-None-Match: * on S3/GCS). wdyt?

laskoviymishka (changes requested):
Nice step for #1096. The five fs-interface methods port cleanly, the directory-marker behavior is consistent with S3/HDFS, and the MinIO integration tests are useful.

One thing before merge: `RenameNoReplace` is still a `Stat` → `Copy` check-then-act, while the `RenameNoReplaceIO` godoc promises atomic behavior and the Hadoop commit path relies on that. With concurrent commits, both writers can pass the existence check and the second copy can overwrite the first. `commitMetadataFile` will never see `fs.ErrExist` because the blob path never returns it.

The `allow-unsafe-commits` gate is enough to land this without implementing true atomic rename. I’d just soften the `RenameIO` / `RenameNoReplaceIO` docs for blob backends and make the gate error explicitly call out the last-writer-wins risk.

A conditional put with `If-None-Match` would be the stronger fix, but that can be a follow-up.

Left the details inline. Once the contract wording is fixed, happy to approve.

io/io.go
```diff
@@ -167,6 +167,54 @@ type ReadDirFile interface {
...
+// method for renaming (moving) files or directories; this must be
+// atomic and can be used for committing metadata updates
+type RenameIO interface {
+	IO
+
+	Rename(oldpath, newpath string) error
+}
+
+// RenameNoReplaceIO is an extension of IO interface that includes the
+// RenameNoReplace method for atomically moving files only when the destination
```
zeroshade: [MAJOR] The RenameIO/RenameNoReplaceIO docs here promise atomic commit/no-replace semantics, but BlobFileIO.RenameNoReplace at io/gocloud/blob.go:441 is check-then-copy-then-delete (Stat(newpath)->Copy->Delete). Concurrent commits can both pass the Stat and silently last-writer-wins, so Hadoop commitMetadataFile will not reliably get fs.ErrExist for conflict detection. Please either implement provider-level conditional create/copy for no-replace, or soften these IO docs and the Hadoop allow-unsafe-commits error to state blob-backed commits are non-atomic and can lose concurrent updates.

io/gocloud/blob.go
```diff
@@ -256,10 +317,178 @@ func (bfs *blobFileIO) DeleteFiles(ctx context.Context, paths []string) ([]strin
...
+	if err != nil {
+		return nil, blobErrToFsErr("ReadFile", path, err)
+	}
+
+	return data, nil
+}
+
+// Stat interprets the input path as a directory or file and returns the corresponding FileInfo.
+// If the path does not exist, it returns fs.ErrNotExist
+func (bfs *BlobFileIO) Stat(path string) (fs.FileInfo, error) {
```
zeroshade, line 498: [MINOR] This Stat implementation checks Attributes(key) before interpreting directory markers, so Stat(".../a/") on a marker created by MkdirAll returns IsDir()==false instead of matching local-FS directory behavior. Please TrimRight(key, "/") for stat/dir purposes, or detect keys ending in "/" and return directory info for marker objects.

io/gocloud/blob.go
zeroshade, line 595: [NIT] The comment says RenameNoReplace renames "a file or directory", but it delegates to single-object Rename and is not recursive. Please reword this to "one file/object" or document the non-recursive behavior.

zeroshade (changes requested):
Requesting changes for the remaining atomicity issue.
- laskoviymishka's RenameNoReplace atomicity change-request is not yet addressed.
- tanmayrauth's naming/path items appear addressed.

io/gocloud/blob.go
C-Loftus: Addressed this in https://github.com/apache/iceberg-go/pull/1350/commits/58c4d690d17e332d8b8185fcb5e4a1d50c312d9f Let me know if that seems ok, thanks!

C-Loftus:
Thanks for reviews all, @laskoviymishka / @zeroshade this should be ready for review again at your convenience

io/gocloud/blob.go
```diff
@@ -368,10 +429,194 @@ func (bfs *blobFileIO) DeleteFiles(ctx context.Context, paths []string) ([]strin
...
+		return &fs.PathError{Op: "mkdir", Path: path, Err: err}
+	}
+
+	key = strings.Trim(key, "/")
+	if key == "" || key == "." {
+		return nil
+	}
+
+	parts := strings.Split(key, "/")
+	for idx := range parts {
```
laskoviymishka, line 471: This writes a marker for every component including the leaf, so an empty namespace ends up with a lone `warehouse/ns/` object. After the tables are dropped that marker is all that's left, and `DropNamespace`'s emptiness walk sees it as a file entry named `ns` and returns `ErrNamespaceNotEmpty` — so an empty blob namespace can never be dropped.

I'd either skip writing the leaf marker here, or have the drop walk skip the namespace's own marker. A CreateNamespace→CreateTable→DropTable→DropNamespace test would have caught it and nothing currently exercises that path. wdyt?

laskoviymishka (changes requested):
One blocker before merge. `MkdirAll` writes a zero-byte marker for the leaf namespace (`warehouse/ns/`), and after the tables are dropped that marker is the only object left under the prefix. `DropNamespace`'s emptiness walk reads it as a file entry named `ns` and returns `ErrNamespaceNotEmpty` — so an empty blob namespace can never be dropped. Details inline; a CreateNamespace→CreateTable→DropTable→DropNamespace test would have caught it. Fix that and I'm happy to approve.

C-Loftus:
@laskoviymishka I believe I should have addressed your feedback. Thanks!

io/gocloud/blob.go
```diff
@@ -368,10 +455,194 @@ func (bfs *blobFileIO) DeleteFiles(ctx context.Context, paths []string) ([]strin
...
+	}
+
+	if err := bfs.Delete(bfs.ctx, oldKey); err != nil && gcerrors.Code(err) != gcerrors.NotFound {
+		return blobErrToFsErr("Rename", oldpath, err)
+	}
+
+	return nil
+}
+
+// RenameNoReplace renames a file or directory from oldpath to newpath, returning an error if newpath already exists.
```
zeroshade: nit: this still says it renames a file or directory, but the implementation delegates to single-object `Rename`. Suggest wording like 'one file/object (non-recursive)' to avoid implying a recursive directory rename.

catalog/hadoop/hadoop_minio_integration_test.go
```diff
@@ -0,0 +1,197 @@
...
+	ident := table.Identifier{"drop_table_then_ns", "tbl"}
+
+	s.Require().NoError(s.cat.CreateNamespace(s.ctx, ns, nil))
+	_, err := s.cat.CreateTable(s.ctx, ident, s.testSchema())
+	s.Require().NoError(err)
+
+	s.Require().NoError(s.cat.DropTable(s.ctx, ident))
+
+	// Ensure that the namespace can be dropped after dropping the table
+	// and that driectory marker logic is handled properly, not preventing
```
zeroshade: nit: typo — 'driectory' should be 'directory'.

zeroshade (approved):
LGTM — the interface docs, non-atomic-commit messaging, and the namespace directory-marker fix all look addressed. Two doc/comment nits inline.

C-Loftus:
Thanks for catching that on the comments! Should be fixed now

laskoviymishka (approved):
👍

PR #1351, opened 2026-07-01
feat(table): variant shredded writer

### LINKED ISSUE #987: feat(table): variant shredded writer
Parent: #589

Once the shredded reader (#986) lands, the writer needs to decide what to shred per row and emit the shredded `STRUCT` alongside `metadata` / residual `value`. Java's posture: no shredding unless explicitly configured. Mirror that with a new table property `write.variant.shredding-paths` carrying a list of `$.path.expressions` — same property name as Java, parsed by the same property layer that already handles `write.metadata.compression-codec` etc.

Add `table/internal/variant_shredder.go` with `ShredVariant(value variant.Value, schema ShreddingSchema) (typed any, residual []byte, err error)`, then hook it into the parquet writer in `table/internal/parquet_files.go` so shredded columns appear in the parquet schema and are populated per row. Statistics on shredded typed columns piggyback the existing column-stats path — typed columns get min/max for free, which is the substrate the follow-up stats issue builds on.

Cross-client coverage: write a shredded variant via iceberg-go, read it via Java/pyiceberg, assert equal to source. Round-trip via the iceberg-go reader is also required. Spec: [Parquet Variant shredding](https://github.com/apache/parquet-format/blob/master/VariantShredding.md).

nssalian:
Closes #987 

## Summary

Adds opt-in shredding of top-level variant columns on the write path. Frequently-occurring
variant fields are written as typed Parquet `typed_value` columns (with the `value` blob kept
as fallback) instead of fully-encoded binary, so engines can read and filter them directly.
arrow-go performs the physical shredding; this PR adds the inference (which fields and types to
shred) and the per-file writer wiring.

Disabled by default - no change to existing writes.

Note: The change to the inference algorithm in java is still under works so I've keep the parity currently in this PR. In the future, if the inference algorithm is modified or adjusted, will follow up here to make sure  iceberg-go is following the same analyzer contract.

## Changes

- `AnalyzeVariantShredding` infers the inner Arrow type from a sample of variant values,
  mirroring Iceberg Java's `VariantShreddingAnalyzer`: most-common-type selection with explicit
  tie-break, integer/decimal widening, a 10% field-frequency floor, and 300-field / depth /
  intermediate-field bounds. Inference is deterministic (same values produce the same schema).
- The `RollingDataWriter` buffers the first N converted batches per file, infers the shredding
  schema, opens the file with it, replays the buffer, and re-infers on roll. The logical
  `FileSchema` is unchanged; only the physical Arrow schema carries the shredded layout.
- Shredded decimals require the spec's INT32/INT64/FLBA-by-precision physical types
  (VariantShredding.md); arrow-go emits those only with `StoreDecimalAsInteger`, enabled
  alongside shredding so default writes are unaffected.

## Properties (both default to current behavior)

| Key | Default |
|---|---|
| `write.parquet.shred-variants` | `false` |
| `write.parquet.variant-inference-buffer-size` | `100` |

## Testing

- Unit: analyzer inference (widening, frequency floor, field cap, nested/array, decimal/temporal/
  uuid leaves), the shred-transform `(value, typed_value)` truth table, and round-trips through
  `UnshredVariant`.
- Write path: real writer to Parquet and back, including no-leak under a checked allocator across
  roll/re-bootstrap, row conservation, partitioned fanout, scalar physical-type assertions, a
  precision > 18 decimal, and a regular (non-variant) decimal column whose manifest bounds stay
  correct under the file-global `StoreDecimalAsInteger`.
- Cross-engine: Go writes a shredded file and Spark 4 reads it back and reassembles the values
  (integration-tagged). Locally ran both integration tests.


nssalian:
~~Based on the variant sync discussion today, I'm taking another look at this and checking the better strategy with some performance testing. Will update when that is ready.~~

No changes made to the current algorithm, this PR continues to be in parity with java

table/rolling_data_writer.go
```diff
@@ -402,37 +441,155 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
...
 		if err != nil {
 			r.sendError(err)
 
 			return
 		}
 
-		if currentWriter.BytesWritten() >= r.factory.targetFileSize {
-			if err := closeWriter(); err != nil {
+		if bootstrapping {
+			buf = append(buf, converted)
```
zeroshade: [MAJOR] `shredBufferRows` is documented as rows buffered per file, but this appends the entire converted batch before checking the row limit. A single large upstream batch (for example 1M rows with a small configured limit) will still be fully retained until bootstrap flushes, causing the memory spike the limit is meant to prevent and violating the property contract. Please either slice batches so only the first `shredBufferRows` rows are retained for bootstrap and write the remainder after the file opens, or document this as a batch-granular limit with an additional byte/batch bound. A regression test with one large batch and a small `write.parquet.variant-inference-buffer-size` would catch this.

table/internal/parquet_files.go
```diff
@@ -305,6 +311,12 @@ func (parquetFormat) GetWriteProperties(props iceberg.Properties) any {
 		writerProps = append(writerProps, parquet.WithBloomFilterEnabledFor(colName, enabled))
 	}
 
+	// Shredded decimals need INT32/INT64/FLBA-by-precision (VariantShredding.md);
+	// arrow-go emits those only with StoreDecimalAsInteger. Gated to keep default writes unchanged.
+	if props.GetBool(ParquetShredVariantsKey, ParquetShredVariantsDefault) {
```
zeroshade: [MAJOR] This enables `StoreDecimalAsInteger` solely from the table property, not from the per-file inference result. As a result, `write.parquet.shred-variants=true` changes the physical encoding of ordinary decimal columns in unrelated files even when no shredded decimal `typed_value` is present. Please select decimal-as-integer per file after inference determines a shredded decimal is needed, or explicitly document the table-level encoding side effect and add property-on/no-variant-column coverage.

table/internal/variant_shredding.go
```diff
@@ -0,0 +1,421 @@
...
+
+			return names[i] < names[j]
+		})
+		for _, name := range names[maxShreddedFields:] {
+			delete(node.objectChildren, name)
+		}
+	}
+
+	for _, child := range node.objectChildren {
+		pruneInfrequent(child, totalRows)
```
zeroshade, line 303: [MINOR] The `maxIntermediateFields` and `maxShreddedFields` caps are applied per object node (see the child creation/pruning above), not as a global inferred-schema budget. Deep or broad samples can still infer a very large recursive schema. Please add a global field/node budget, or rename/document these as per-object caps and add a nested high-cardinality test.

table/rolling_data_writer.go
```diff
@@ -402,37 +441,155 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
...
+			va, ok := b.Column(col).(*extensions.VariantArray)
+			if !ok {
+				continue
+			}
+			for i := 0; i < va.Len() && len(sample) < limit; i++ {
+				if va.Storage().IsNull(i) {
+					continue
+				}
+				v, err := va.Value(i)
+				if err != nil {
```
zeroshade, line 640: [QUESTION] Should `va.Value(i)` errors be surfaced instead of silently skipped? Ignoring malformed values can infer shredding from a biased subset of rows. Returning an error, or at least logging/tracking skipped rows, would make this easier to diagnose.

table/rolling_data_writer.go
```diff
@@ -402,37 +441,155 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
...
 }
 
+// inferShreddingFromBatches infers the inner type per top-level variant column,
+// sampling up to limit non-null values from the buffer. Keyed by column index.
+func inferShreddingFromBatches(buf []arrow.RecordBatch, limit int) map[int]arrow.DataType {
+	if len(buf) == 0 {
+		return nil
+	}
+
+	inferred := make(map[int]arrow.DataType)
```
zeroshade, line 620: [NIT] Since this map is keyed by column index, future changes beyond top-level use may be easier to follow if the key carries the field ID or name instead.

zeroshade (changes requested):
Reviewing early since the PR is substantial and the direction is sound — these are the blocking items to resolve before it's review-ready.

- Bound the variant inference bootstrap buffer by rows/bytes or document the current batch-granular behavior and test it.
- Avoid changing ordinary decimal column encoding globally when variant shredding is enabled.
- Clarify/enforce shredding schema size limits and consider surfacing malformed variant values during inference.

table/rolling_data_writer.go
nssalian: Thanks for catching this. I bounded the bootstrap buffer to `shredBufferRows`: an oversized batch is sliced, only the head is buffered, the tail writes steady-state. Wrote a test `TestShreddedVariantWriteLargeBootstrapBatch` (3000-row batch, 100-row buffer, 1-byte target) that asserts the batch splits across files - whole-batch buffering would produce one - plus row conservation under the checked allocator.

table/internal/parquet_files.go
nssalian: Fixed this too. `StoreDecimalAsInteger` is now decided per file in `NewFileWriter` from the inferred schema, not the table property, so unrelated decimal columns keep FLBA. The props slice is `slices.Clone`d before the append so concurrent partition writers don't race (`TestShreddedVariantPartitionedDecimalRace` fails under `-race` without it). Covered: INT32/INT64/FLBA by precision, and wrote a test `TestShreddedVariantPartitionedGoWriteSparkRead` that reads a Go-written partitioned shredded decimal back via Spark.

table/internal/variant_shredding.go
nssalian: Per-node on purpose, matching `VariantShreddingAnalyzer` in Java Iceberg (per-node cap in `pruneInfrequentFields`, frequency measured against root row count at every depth, same 0.10 / 300 / 50 / 1000 constants). A global budget would diverge from Java. Fixed the doc comment and added  TestAnalyzeNestedHighCardinality + TestAnalyzeDepthCap.

table/rolling_data_writer.go
nssalian: Now this counts and logs undecodable sample values instead of dropping them silently; kept non-fatal since a malformed value is already hard-errored at shred time. `TestInferShreddingSkipsUndecodable` covers it in testing.

table/rolling_data_writer.go
nssalian: Keyed by column index on purpose: `ShreddedArrowSchema` applies each type to `fields[idx]` by position. Clarified the doc; happy to switch to field-id if you prefer.

nssalian:
Merged latest main. The one conflict was in `parquet_files.go`: main added the validating `getWriteProperties` helper and this branch added per-file `StoreDecimalAsInteger` gating for shredded decimals. Folded them together - getWriteProperties now takes the arrow schema and does the clone + conditional append inside, so the validation and 
the shredded-decimal encoding live in one place.

zeroshade (approved):

PR #1352, opened 2026-07-02
Preserve reserved Glue and Hive table parameters

fallintoplace:
## Summary

Glue and Hive table construction copied user table properties after setting catalog-owned parameters. That allowed properties such as `table_type`, `metadata_location`, `EXTERNAL`, and Hive `storage_handler` to overwrite the values required for Iceberg catalog entries.

This changes the parameter construction order so user properties are copied first and catalog-required parameters are assigned last. Glue also keeps `previous_metadata_location` catalog-owned by deriving it from the previous Glue table during updates and omitting user-provided values during create/register. Hive table creation/register does the same for `previous_metadata_location`, and Hive views now preserve their required reserved parameters as well.

## Testing

- `go test ./catalog/glue ./catalog/hive`
- `go test ./...`

### DIFF
--- a/catalog/hive/schema.go
+++ b/catalog/hive/schema.go
@@ -102,6 +102,8 @@ func constructHiveViewTable(dbName, viewName, location, metadataLocation string,
 			parameters[k] = v
 		}
 	}
+	delete(parameters, PreviousMetadataLocationKey)
+
 	parameters[TableTypeKey] = TableTypeIcebergView
 	parameters[MetadataLocationKey] = metadataLocation
 	parameters[ExternalKey] = "TRUE"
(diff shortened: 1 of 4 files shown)


catalog/hive/schema.go
```diff
@@ -97,14 +97,14 @@ const (
...
-	parameters[MetadataLocationKey] = metadataLocation
-	parameters[ExternalKey] = "TRUE"
 	for k, v := range props {
 		if v != "" {
 			parameters[k] = v
 		}
 	}
+	parameters[TableTypeKey] = TableTypeIcebergView
+	parameters[MetadataLocationKey] = metadataLocation
+	parameters[ExternalKey] = "TRUE"
```
zeroshade, line 109: Consistency (optional): the table path below deletes `PreviousMetadataLocationKey` from user props before setting reserved keys, but this view path does not. If a user passes `previous_metadata_location` for a view it will be persisted as-is — intentional? If not, mirror the table path here.

zeroshade (approved):
LGTM. Merging user props first and then stamping the reserved catalog keys (so `table_type` / `metadata_location` / `EXTERNAL` always win) is correct, and dropping `previous_metadata_location` from user input on the table path is a good catch. One optional consistency note inline.

PR #1353, opened 2026-07-02
Clone metadata builder properties

fallintoplace:
## Summary

`MetadataBuilder.SetProperties` kept the caller-provided properties map when the builder had no existing properties. That meant later changes to the caller map could mutate builder state after `SetProperties` returned.

This clones incoming properties before storing them in the builder and before recording the pending set-properties update. The builder aggregate properties and the recorded update now stay independent from caller-owned maps and from later builder property merges.

## Testing

- `go test ./table`
- `go test ./...`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -780,11 +780,12 @@ func (b *MetadataBuilder) SetProperties(props iceberg.Properties) error {
 		}
 	}
 
-	b.updates = append(b.updates, NewSetPropertiesUpdate(props))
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Cloning the properties map with `maps.Clone` fixes the aliasing, and the regression test confirms caller mutation no longer leaks into the builder or the recorded update. `maps.Clone` is sufficient here since `Properties` is `map[string]string`.

PR #1354, opened 2026-07-02
fix(rest): validate table and view identifiers

fallintoplace:
## Summary

REST table and view operations accepted one-part identifiers such as `["table"]`. The path helper treated that as an empty namespace plus table name, which could produce malformed REST paths or JSON identifiers.

This adds a shared `catalog.ValidateTableIdentifier` helper that requires at least one namespace level and a table/view name, then wires REST table and view operations through it. Rename and transaction commits now validate identifiers before constructing REST identifier payloads too.

## Testing

- `go test ./catalog ./catalog/rest`
- `go test ./...`

### DIFF
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -214,6 +214,16 @@ func NamespaceFromIdent(ident table.Identifier) table.Identifier {
 	return ident[:len(ident)-1]
 }
 
+// ValidateTableIdentifier checks that an identifier contains at least one namespace level and a table name.
+func ValidateTableIdentifier(ident table.Identifier) error {
+	if len(ident) < 2 {
+		return fmt.Errorf("%w: missing namespace or invalid identifier %v",
+			ErrNoSuchTable, strings.Join(ident, "."))
+	}
+
+	return nil
+}
+
 type CreateTableOpt func(*CreateTableCfg)
 
 func WithLocation(location string) CreateTableOpt {
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -34,6 +34,7 @@ import (
 	"iter"
 	"maps"
 	"strings"
+	"unicode"
 
 	"github.com/apache/iceberg-go"
 	iceinternal "github.com/apache/iceberg-go/internal"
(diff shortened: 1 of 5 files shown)


catalog/catalog.go
```diff
@@ -214,6 +214,16 @@ func NamespaceFromIdent(ident table.Identifier) table.Identifier {
 	return ident[:len(ident)-1]
 }
 
+// ValidateTableIdentifier checks that an identifier contains at least one namespace level and a table name.
+func ValidateTableIdentifier(ident table.Identifier) error {
+	if len(ident) < 2 {
```
zeroshade, line 219: This only rejects identifiers with fewer than two elements, so path-altering components still pass validation: `table.Identifier{"ns", ""}`, `{"ns", "."}`, `{"ns", ".."}`, and `{"ns", "a/b"}`. In the REST client these flow through `splitIdentForPath` into `url.URL.JoinPath`, where `/`, `.`, and `..` are interpreted as path structure/cleaning, so the REST path-injection issue is only partially fixed. Please reject empty components, `.`, `..`, `/`, and control characters in every component; preferably also path-encode the final table/view name like Java's `RESTUtil.encodeString` / pyiceberg's `safe=""`, and add tests showing no malformed REST path is generated.

catalog/rest/rest.go
```diff
@@ -1156,8 +1155,8 @@ func (r *Catalog) CommitTransaction(ctx context.Context, commits []table.TableCo
 
 	changes := make([]tableChange, len(commits))
 	for i, c := range commits {
-		if len(c.Identifier) == 0 {
-			return catalog.ErrMissingIdentifier
+		if err := catalog.ValidateTableIdentifier(c.Identifier); err != nil {
+			return fmt.Errorf("%w: %w", catalog.ErrMissingIdentifier, err)
```
zeroshade: Wrapping every malformed identifier as `ErrMissingIdentifier` conflates an absent identifier with a present-but-invalid one. Please keep `ErrMissingIdentifier` only for `len(c.Identifier) == 0`, and otherwise return the original `ValidateTableIdentifier` error.

catalog/rest/rest_test.go
```diff
@@ -2688,13 +2688,41 @@ func (r *RestCatalogSuite) TestCatalogWithCustomTransport() {
...
+
+	cat, err := rest.NewCatalog(context.Background(), "rest", r.srv.URL, rest.WithCustomTransport(&transport))
+	r.Require().NoError(err)
+	r.Require().Len(transport.calls, 1)
+
+	_, err = cat.LoadTable(context.Background(), table.Identifier{"table"})
+	r.ErrorIs(err, catalog.ErrNoSuchTable)
+
+	_, err = cat.LoadView(context.Background(), table.Identifier{"view"})
+	r.ErrorIs(err, catalog.ErrNoSuchTable)
```
zeroshade: Question: invalid view identifiers now surface `ErrNoSuchTable` rather than `ErrNoSuchView` via the shared table-named helper. That may be acceptable, but it is a little surprising for view APIs.

zeroshade (changes requested):
Thanks for tightening identifier validation; a path-component gap remains and one error mapping should be preserved before this is safe to merge.

- Major: validate every identifier component and/or path-encode table/view names so malformed REST paths cannot be generated.
- Minor: keep ErrMissingIdentifier distinct from present-but-invalid identifiers in CommitTransaction.
- Question: view APIs currently surface ErrNoSuchTable for invalid view identifiers via the shared helper; that may be acceptable but is surprising.

zeroshade (approved):
Re-reviewed — all three of my earlier concerns are addressed: component validation now rejects empty, dot, dot-dot, slash, and control characters; CommitTransaction distinguishes absent from invalid identifiers; and view identifiers validate via ValidateViewIdentifier and surface ErrNoSuchView. One optional follow-up: a test that a valid table/view name needing URL-encoding produces the expected encoded path. Not blocking. LGTM.

PR #1355, opened 2026-07-02
fix(manifest): encode decimal partitions with declared width

fallintoplace:
## Summary

Decimal partition values now use the fixed byte width declared by partition-schema precision instead of a width inferred from the formatted value.

- keep decimal read scale separate from fixed encoding width;
- sign-extend negative values to the declared width;
- reject oversized values unless discarded bytes are redundant sign extension;
- apply the declared width in data-file statistics conversion and `MarshalAvroEntry`;
- keep builder-created partition values directly typed; scale reconstruction is only required after Avro decoding.

## Testing

- table-driven positive, zero, negative, boundary, padding, and overflow vectors
- manifest write/read decimal round trips
- `DataFileStatistics.ToDataFile` → manifest round trip
- `MarshalAvroEntry` → decoder round trip
- `go test . ./table/internal`

### DIFF
--- a/data_file_codec.go
+++ b/data_file_codec.go
@@ -143,7 +143,6 @@ func unmarshalAvroDataFileEntry(data []byte, spec PartitionSpec, schema *Schema,
 	df.specID = int32(spec.ID())
 	df.fieldNameToID = maps.nameToID
 	df.fieldIDToLogicalType = maps.idToType
-	df.fieldIDToFixedSize = maps.idToFixedSize
 	df.fieldIDToDecimalScale = maps.idToDecimalScale
 
 	return df, nil
@@ -295,15 +294,10 @@ func manifestEntrySchemaFor(spec PartitionSpec, schema *Schema, version int) (*a
 	if err != nil {
 		return nil, dataFileFieldMaps{}, err
 	}
-	n2i, i2t, i2s, i2sc := getFieldIDMap(fullSchema)
+	fieldMaps := getFieldIDMap(fullSchema)
 	entry := &dataFileSchemaEntry{
 		schema: fullSchema,
-		maps: dataFileFieldMaps{
-			nameToID:         n2i,
-			idToType:         i2t,
-			idToFixedSize:    i2s,
-			idToDecimalScale: i2sc,
-		},
+		maps:   fieldMaps,
 	}
 	dataFileSchemaCache.Add(key, entry)
 
--- a/data_file_codec.go
+++ b/data_file_codec.go
@@ -107,7 +107,11 @@ func (d *dataFile) MarshalAvroEntry(spec PartitionSpec, schema *Schema, version 
 		return nil, err
 	}
 	clone := cloneDataFileAvroFields(d)
-	clone.PartitionData = avroEncodePartitionData(d.Partition(), maps.nameToID, maps.idToType)
+	partitionData, err := avroEncodePartitionData(d.Partition(), maps.nameToID, maps.idToType, maps.idToFixedSize)
+	if err != nil {
+		return nil, err
+	}
+	clone.PartitionData = partitionData
 
 	return s.Encode(newEncodeEntry(version, clone))
 }
@@ -140,6 +144,7 @@ func unmarshalAvroDataFileEntry(data []byte, spec PartitionSpec, schema *Schema,
 	df.fieldNameToID = maps.nameToID
 	df.fieldIDToLogicalType = maps.idToType
 	df.fieldIDToFixedSize = maps.idToFixedSize
+	df.fieldIDToDecimalScale = maps.idToDecimalScale
 
 	return df, nil
 }
@@ -205,8 +210,11 @@ func cloneDataFileAvroFields(src *dataFile) *dataFile {
 // iceberg-typed values like Date or Decimal) into the name-keyed
 // avro-friendly map the manifest-entry schema expects. Idempotent:
 // values already in primitive form pass through unchanged.
-func avroEncodePartitionData(idKeyed map[int]any, nameToID map[string]int, logicalTypes map[int]string) map[string]any {
-	converted := avroPartitionData(idKeyed, logicalTypes)
+func avroEncodePartitionData(idKeyed map[int]any, nameToID map[string]int, logicalTypes map[int]string, fixedSizes map[int]int) (map[string]any, error) {
+	converted, err := avroPartitionData(idKeyed, logicalTypes, fixedSizes)
+	if err != nil {
+		return nil, err
+	}
 	out := make(map[string]any, len(converted))
 	for name, id := range nameToID {
 		if v, ok := converted[id]; ok {
@@ -214,13 +222,14 @@ func avroEncodePartitionData(idKeyed map[int]any, nameToID map[string]int, logic
 		}
 	}
 
-	return out
+	return out, nil
 }
 
 type dataFileFieldMaps struct {
-	nameToID      map[string]int
-	idToType      map[int]string
-	idToFixedSize map[int]int
+	nameToID         map[string]int
+	idToType         map[int]string
+	idToFixedSize    map[int]int
+	idToDecimalScale map[int]int
 }
 
 // dataFileSchemaCacheKey identifies a cached avro schema by the
@@ -286,13 +295,14 @@ func manifestEntrySchemaFor(spec PartitionSpec, schema *Schema, version int) (*a
 	if err != nil {
 		return nil, dataFileFieldMaps{}, err
 	}
-	n2i, i2t, i2s := getFieldIDMap(fullSchema)
+	n2i, i2t, i2s, i2sc := getFieldIDMap(fullSchema)
 	entry := &dataFileSchemaEntry{
 		schema: fullSchema,
 		maps: dataFileFieldMaps{
-			nameToID:      n2i,
-			idToType:      i2t,
-			idToFixedSize: i2s,
+			nameToID:         n2i,
+			idToType:         i2t,
+			idToFixedSize:    i2s,
+			idToDecimalScale: i2sc,
 		},
 	}
 	dataFileSchemaCache.Add(key, entry)
--- a/manifest.go
+++ b/manifest.go
@@ -1943,7 +1943,7 @@ func encodeDecimalBytes(dec DecimalLiteral, fixedSize int) ([]byte, error) {
 		return nil, err
 	}
 
-	return padOrTruncateBytes(bytes, fixedSize), nil
+	return fitDecimalBytes(bytes, fixedSize)
 }
 
 func convertUUIDValue(v any) any {
@@ -1954,13 +1954,41 @@ func convertUUIDValue(v any) any {
 	return v
 }
 
-func padOrTruncateBytes(bytes []byte, size int) []byte {
-	if len(bytes) >= size {
-		return bytes[len(bytes)-size:]
+func fitDecimalBytes(bytes []byte, size int) ([]byte, error) {
+	if len(bytes) == 0 {
+		return make([]byte, size), nil
 	}
-	padded := slices.Grow(bytes, size-len(bytes))
 
-	return append(make([]byte, size-len(bytes)), padded...)
+	if len(bytes) == size {
+		return bytes, nil
+	}
+
+	if len(bytes) < size {
+		signByte := byte(0x00)
+		if bytes[0]&0x80 != 0 {
+			signByte = 0xff
+		}
+		padded := make([]byte, size)
+		for i := range padded[:size-len(bytes)] {
+			padded[i] = signByte
+		}
+		copy(padded[size-len(bytes):], bytes)
+
+		return padded, nil
+	}
+
+	retained := bytes[len(bytes)-size:]
+	signByte := byte(0x00)
+	if retained[0]&0x80 != 0 {
+		signByte = 0xff
+	}
+	for _, b := range bytes[:len(bytes)-size] {
+		if b != signByte {
+			return nil, fmt.Errorf("decimal value does not fit fixed size %d", size)
+		}
+	}
+
+	return retained, nil
 }
 
 type dataFile struct {
--- a/manifest.go
+++ b/manifest.go
@@ -432,7 +432,7 @@ func (m *manifestFile) FetchEntries(fs iceio.IO, discardDeleted bool) (_ []Manif
 	return ReadManifest(m, f, discardDeleted)
 }
 
-func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int, map[int]int) {
+func getFieldIDMap(sc *avro.Schema) dataFileFieldMaps {
 	root := sc.Root()
 	getField := func(node avro.SchemaNode, name string) *avro.SchemaField {
 		for i := range node.Fields {
@@ -477,7 +477,12 @@ func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int
 		}
 	}
 
-	return result, logicalTypes, fixedSizes, decimalScales
+	return dataFileFieldMaps{
+		nameToID:         result,
+		idToType:         logicalTypes,
+		idToFixedSize:    fixedSizes,
+		idToDecimalScale: decimalScales,
+	}
 }
 
 // applyDayTransformDates marks every day(...) partition field described by the
@@ -515,7 +520,6 @@ func applyDayTransformDates(specJSON []byte, fieldIDToType map[int]string) {
 type hasFieldToIDMap interface {
 	setFieldNameToIDMap(map[string]int)
 	setFieldIDToLogicalTypeMap(map[int]string)
-	setFieldIDToFixedSizeMap(map[int]int)
 	setFieldIDToDecimalScaleMap(map[int]int)
 }
 
@@ -679,7 +683,6 @@ type ManifestReader struct {
 	content        ManifestContent
 	fieldNameToID  map[string]int
 	fieldIDToType  map[int]string
-	fieldIDToSize  map[int]int
 	fieldIDToScale map[int]int
 
 	// The rest are lazily populated, on demand. Most readers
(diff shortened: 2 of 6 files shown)


manifest.go
```diff
@@ -1848,18 +1864,28 @@ func convertTimestampMicrosValue(v any) any {
...
+		return nil, fmt.Errorf("missing decimal fixed size")
 	}
 
-	return v
+	bytes, err := dec.MarshalBinary()
+	if err != nil {
+		return nil, err
+	}
+
+	return padOrTruncateBytes(bytes, fixedSize), nil
```
zeroshade: [BLOCKER] This call still routes through `padOrTruncateBytes`, whose implementation at lines 1899-1905 left-pads short decimal byte slices with `0x00` unconditionally. That is only correct for non-negative values; a negative unscaled value such as -1 for precision 10 (5 bytes) becomes `00 00 00 00 ff` and decodes as +255. Fix with sign-aware fixed-width encoding: when `len(bytes) < fixedSize`, prepend `0xff` if `bytes[0]&0x80 != 0`, otherwise `0x00` (two's-complement sign extension), and add negative/zero/max-precision-38 vectors.

zeroshade: [MAJOR] `padOrTruncateBytes` also silently truncates oversized decimals to the low-order bytes (see real lines 1899-1901), which can change the value instead of reporting that it does not fit the declared fixed size. Since this path already returns an error, make `len(bytes) > fixedSize` verify the dropped prefix is redundant sign extension and otherwise return an overflow/precision error.

manifest_test.go
```diff
@@ -2044,6 +2046,25 @@ func TestReadManifestDecodesNilLogicalPartitionValueFromNullableUnion(t *testing
 	}
 }
 
+func TestAvroEncodePartitionDataUsesDeclaredDecimalFixedSize(t *testing.T) {
```
zeroshade, line 2177: [MINOR] This test only covers one positive decimal, so it misses the negative sign-extension bug and oversized-value truncation. Please make this table-driven with positive, zero, negative, max-width, and too-large vectors, and include a `ManifestWriter`/`ReadManifest` round-trip.

zeroshade (changes requested):
Decimal partition encoding still corrupts negative/oversized fixed decimals and needs broader coverage.

- [BLOCKER] Fixed-width decimal padding must sign-extend negatives instead of always prepending 0x00.
- [MAJOR] Oversized decimals must be rejected unless the dropped prefix is redundant sign extension.
- [MINOR] Add table-driven decimal vectors and a manifest round-trip to cover this behavior.

manifest.go
```diff
@@ -2006,7 +2061,7 @@ func (d *dataFile) convertAvroValueToIcebergType(v any, fieldID int) any {
 			return TimestampNano(v.(int64))
 		case atype.Decimal:
 			if r, ok := v.(*big.Rat); ok {
-				scale := d.fieldIDToFixedSize[fieldID]
+				scale := d.fieldIDToDecimalScale[fieldID]
```
laskoviymishka, line 2122: Now that the read path reads `d.fieldIDToDecimalScale` here, `d.fieldIDToFixedSize` has no reader left anywhere — the write path encodes off `ManifestWriter.partFieldIDToSize` (schema-derived), not the per-`dataFile` field. That makes the `fieldIDToFixedSize` argument to `NewDataFileBuilder` silently a no-op: a caller passing 0 or a wrong width gets neither an error nor a different result.

I'd drop `fieldIDToFixedSize` from the struct and the `NewDataFileBuilder` signature; if it has to stay for the `hasFieldToIDMap` interface, at least comment that it's unread on the write path so we're not handing callers a knob that does nothing.

table/internal/utils.go
```diff
@@ -290,7 +291,7 @@ func (d *DataFileStatistics) ToDataFile(opts DataFileOpts) iceberg.DataFile {
 					fieldIDToLogicalType[field.FieldID] = atype.TimestampMicros
 				case iceberg.DecimalType:
 					fieldIDToLogicalType[field.FieldID] = atype.Decimal
-					fieldIDToFixedSize[field.FieldID] = rt.Scale()
+					fieldIDToFixedSize[field.FieldID] = iceinternal.DecimalRequiredBytes(rt.Precision())
```
laskoviymishka: This is the same class of bug the manifest fix addresses, but on the path Arrow writers actually use to build `DataFile`s for commit — and it has no test. A regression here would corrupt every decimal partition value written by an Arrow writer without a single test failing.

I'd add one that runs a `DataFileStatistics` with a decimal partition column through `ToDataFile` → `WriteManifest` → read back and asserts the decoded value round-trips.

manifest.go
```diff
@@ -1916,6 +1970,7 @@ type dataFile struct {
 	fieldIDToLogicalType   map[int]string
 	fieldIDToPartitionData map[int]any
 	fieldIDToFixedSize     map[int]int
+	fieldIDToDecimalScale  map[int]int
```
laskoviymishka, line 2031: `fieldIDToDecimalScale` only ever gets populated by the `ReadEntry` setter path — `NewDataFileBuilder` doesn't take a scale map. So a `dataFile` built via the constructor and read back with `Partition()` without going through `ManifestReader` reads `scale=0`, and `(123, scale=2)` silently decodes as `123`.

Mirror image of the dead `fieldIDToFixedSize` above — I'd either add a `fieldIDToDecimalScale` param to `NewDataFileBuilder` or derive scale from the spec+schema in the constructor. wdyt?

manifest.go
```diff
@@ -1870,13 +1896,41 @@ func convertUUIDValue(v any) any {
 	return v
 }
 
-func padOrTruncateBytes(bytes []byte, size int) []byte {
-	if len(bytes) >= size {
-		return bytes[len(bytes)-size:]
+func fitDecimalBytes(bytes []byte, size int) ([]byte, error) {
+	if len(bytes) == 0 {
```
laskoviymishka: The `len(bytes) == 0` branch is unreachable — `DecimalLiteral.MarshalBinary` always emits at least one byte (zero gives `[0x00]`), so no caller can deliver an empty slice. Meanwhile there's no `size <= 0` guard: called with `size == 0` and non-empty bytes, it falls through to the trim branch and `retained[0]` panics on the empty slice.

`encodeDecimalBytes` catches `fixedSize <= 0` today so it's not reachable in production, but since this is unexported and test-callable I'd swap the dead empty-bytes check for a `size <= 0` guard that returns an error.

manifest_test.go
```diff
@@ -2044,6 +2047,202 @@ func TestReadManifestDecodesNilLogicalPartitionValueFromNullableUnion(t *testing
...
+				t.Fatal(err)
+			}
+			if !bytes.Equal(got, tt.want) {
+				t.Fatalf("fitDecimalBytes() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestAvroEncodePartitionDataUsesDeclaredDecimalFixedSize(t *testing.T) {
```
laskoviymishka, line 2177: The vectors here are solid, but they only hit precision 10 and 38 — I'd add the `requiredLength` tier boundaries too, since those are exactly where an off-by-one in `DecimalRequiredBytes` would show up: precision 1 (size 1, value 5 → `[0x05]`), precision 9 (size 4), and a negative at precision 18 (size 9, sign-extended).

While we're here, `TestFitDecimalBytesAllowsOnlyRedundantSignExtension` only exercises the trim path — a direct padding case like `{0x01}` size 3 → `{0x00,0x00,0x01}` plus a negative `{0xff}` → `{0xff,0xff,0xff}` would cover the pad branch directly rather than only through the encode test.

manifest.go
```diff
@@ -376,7 +376,7 @@ func (m *manifestFile) FetchEntries(fs iceio.IO, discardDeleted bool) (_ []Manif
 	return ReadManifest(m, f, discardDeleted)
 }
 
-func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int) {
+func getFieldIDMap(sc *avro.Schema) (map[string]int, map[int]string, map[int]int, map[int]int) {
```
laskoviymishka: `getFieldIDMap` now returns four positional maps, two of them `map[int]int` — at the call sites the size map and the scale map are indistinguishable by type, and one site already discards the fourth with `_`. I'd return the existing `dataFileFieldMaps` struct (or named returns) so a future caller can't swap size and scale by accident. wdyt?

manifest.go
```diff
@@ -1848,18 +1864,28 @@ func convertTimestampMicrosValue(v any) any {
...
 }
 
-func convertDecimalValue(v any) any {
-	if dec, ok := v.(Decimal); ok {
-		fixedSize := internal.DecimalRequiredBytes(len(dec.String()))
-		bytes, err := DecimalLiteral(dec).MarshalBinary()
-		if err != nil {
-			return v
-		}
+func convertDecimalValue(v any, fixedSize int) (any, error) {
```
laskoviymishka, line 1926: `convertDecimalValue` now handles both `Decimal` (from `NewDataFileBuilder`) and `DecimalLiteral` (from a round-trip through `initPartitionData`) — that dual case is a bit of a symptom of the partition value model not converging on one type. Not blocking, but a one-line comment on when each shows up would save the next reader the archaeology. wdyt?

manifest.go
```diff
@@ -1870,13 +1896,41 @@ func convertUUIDValue(v any) any {
...
 
-	return append(make([]byte, size-len(bytes)), padded...)
+	retained := bytes[len(bytes)-size:]
+	signByte := byte(0x00)
+	if retained[0]&0x80 != 0 {
+		signByte = 0xff
+	}
+	for _, b := range bytes[:len(bytes)-size] {
+		if b != signByte {
+			return nil, fmt.Errorf("decimal value does not fit fixed size %d", size)
```
laskoviymishka: Small thing: this reports only `size`, so at the top level you get `decimal value does not fit fixed size 5` with no hint whether the value was 6 bytes or 16. Including `len(bytes)` — `decimal value of %d bytes does not fit fixed size %d` — makes it debuggable at a glance.

manifest_test.go
```diff
@@ -2044,6 +2047,202 @@ func TestReadManifestDecodesNilLogicalPartitionValueFromNullableUnion(t *testing
...
+			value:     Decimal{Val: decimal128.FromI64(-1), Scale: 2},
+			precision: 10,
+			want:      []byte{0xff, 0xff, 0xff, 0xff, 0xff},
+		},
+		{
+			name:      "keeps max precision decimal",
+			value:     DecimalLiteral{Val: decimal128.FromBigInt(maxPrecision38), Scale: 0},
+			precision: 38,
+			want: []byte{
+				0x4b, 0x3b, 0x4c, 0xa8, 0x5a, 0x86, 0xc4, 0x7a,
```
laskoviymishka, line 2230: Could we drop a one-liner on what these bytes are — e.g. `// two's-complement big-endian of 10^38 - 1`? A bare 16-byte literal is the kind of thing that's impossible to verify or update later without recomputing it by hand.

laskoviymishka (changes requested):
Nice work on the revision — the pieces zeroshade asked for all look like they landed: negatives sign-extend in `fitDecimalBytes`, oversized values are rejected unless the dropped prefix is redundant sign extension, and there's a table-driven vector set plus a negative round-trip through `WriteManifest`/`ReadManifest`. The core fix is right, too — splitting write-width from read-scale in `getFieldIDMap` is exactly the bug, and the encoded bytes line up with Java and PyIceberg.

I'd hold before merging though, mostly on one thing I kept running into from different directions. Now that the read path reads `fieldIDToDecimalScale`, the old `fieldIDToFixedSize` state has no reader left — the write path encodes off the schema-derived `partFieldIDToSize`, not the per-`dataFile` field — but `NewDataFileBuilder` still accepts `fieldIDToFixedSize` as a parameter. So it's a public argument that silently does nothing, and its mirror (`fieldIDToDecimalScale`) has no constructor entry point at all, which means a builder-built file read back outside `ManifestReader` decodes at `scale=0`. I'd like the constructor surface to tell the truth about which of these actually affects the bytes.

The other one is the `DataFileStatistics.ToDataFile` change — same fix, right change, but it's the path Arrow writers use to produce data files for commit and it has no test, so a future regression there would corrupt decimal partitions with nothing failing.

- Make `NewDataFileBuilder` honest about `fieldIDToFixedSize` (drop it, or document it's unread on the write path) and give it a way to set scale.
- Add a test for the decimal path through `ToDataFile` → `WriteManifest` → read back.
- Extend the vectors to the `requiredLength` tier boundaries (precision 1/9/18), and cover `fitDecimalBytes`'s padding branch directly.

Everything else is small — a dead `len(bytes)==0` guard, error messages that could carry the byte length, the four positional maps out of `getFieldIDMap`. Fix the API surface and the `ToDataFile` test and I'm happy to take another pass and approve.

manifest.go
```diff
@@ -2260,6 +2320,7 @@ func NewDataFileBuilder(
 	recordCount int64,
 	fileSize int64,
 ) (*DataFileBuilder, error) {
+	_ = fieldIDToFixedSize
```
laskoviymishka, line 2387: I said document-or-drop last round and you documented it — that's fair by what I asked. Seeing `_ = fieldIDToFixedSize` actually land though, I've come around to preferring drop.

Both reviewers on this pass read it as a live parameter and assumed it mattered, and the round-trip test still passes a real `internal.DecimalRequiredBytes(10)` into it that goes nowhere. We're pre-1.0, so dropping it is cheaper than carrying a dead arg everyone has to be told to ignore.

Not blocking — your call. Just flagging that's where I'd land now.

manifest.go
```diff
@@ -2313,7 +2374,6 @@ func NewDataFileBuilder(
 			fieldIDToPartitionData: fieldIDToPartitionData,
 			fieldNameToID:          fieldNameToID,
 			fieldIDToLogicalType:   fieldIDToLogicalType,
```
laskoviymishka, line 2440: This is the scale half of what I asked last round and it's still open — a dataFile built here has no way to populate `fieldIDToDecimalScale`; only `ManifestReader.ReadEntry` sets it via `setFieldIDToDecimalScaleMap`. So a builder-built file decoded outside a manifest read hits `convertAvroValueToIcebergType` with scale=0 and mis-scales the decimal.

What I can't tell is whether that path is actually reachable — is anything meant to decode a builder-built dataFile without going through `ManifestReader`? If not, this is moot and I'll drop it; if it is, we'd want a scale setter on the builder. wdyt?

data_file_codec.go
```diff
@@ -107,7 +107,11 @@ func (d *dataFile) MarshalAvroEntry(spec PartitionSpec, schema *Schema, version
 		return nil, err
 	}
 	clone := cloneDataFileAvroFields(d)
-	clone.PartitionData = avroEncodePartitionData(d.Partition(), maps.nameToID, maps.idToType)
+	partitionData, err := avroEncodePartitionData(d.Partition(), maps.nameToID, maps.idToType, maps.idToFixedSize)
```
laskoviymishka, line 110: There are two write paths that encode decimal partitions — `ManifestWriter.addEntry` and this one — and the new tests only cover the first. The old bug lived in both and this fix touches both, so I'd want a round-trip here too: `MarshalAvroEntry` → `unmarshalAvroDataFileEntry` on a decimal partition. Otherwise a regression on this path slips through. Could you add one?

laskoviymishka (changes requested):
Really nice revision. Everything I flagged last time is covered now.

`ToDataFile` has a round-trip test, the padding branch in `fitDecimalBytes` is covered directly, including the precision 1/9/18/38 boundaries, and `getFieldIDMap` now returns a `dataFileFieldMaps` struct instead of four positional maps. The encoding still matches Java and PyIceberg.

I have two narrow things left before approving.

First, the scale part of the builder question is still open. A `dataFile` created through `NewDataFileBuilder` has no way to populate `fieldIDToDecimalScale`; only `ManifestReader` does that. So if that file is decoded anywhere outside a manifest read, decimals would be interpreted with scale 0.

I’m not sure that path is actually reachable. If builder-created files are never decoded without going through `ManifestReader`, then this is moot and I’m happy to drop it. I’d just like to confirm that rather than assume.

Second, `MarshalAvroEntry` is another path that writes decimal partitions, and there still isn’t a decimal test for it. Since this change affects that path too, I’d like to see a round-trip test there.

On `fieldIDToFixedSize`, I asked for either documenting or removing it last time, and you documented it, which is fair. But now that `_ = fieldIDToFixedSize` is in the code, I think removing it would be cleaner. Both reviewers read it as live state, and the test passes a real value that is then ignored. Since we’re pre-1.0, this seems cheap to clean up. Not blocking, just my preference.

Everything else @zeroshade and I raised is in:

* negative values sign-extend correctly
* oversized values are rejected unless the removed prefix is redundant
* the table-driven vectors are there
* write width and read scale are separated cleanly
* the `ToDataFile` test and boundary cases landed

The smaller items from last round, like the `len(bytes) == 0` guard and including the byte length in errors, are still optional.

Once the scale question is settled and there’s a `MarshalAvroEntry` decimal test, I’m happy to approve.

zeroshade (approved):
Re-reviewed the current revision — my earlier blocker is resolved. `fitDecimalBytes` now sign-extends negative decimals (0xff when the high bit is set) and rejects oversized values unless the dropped prefix bytes are redundant sign extension, with the added round-trip/precision coverage. Thanks @fallintoplace! (Leaving @laskoviymishka's review to them.)

zeroshade:
@laskoviymishka when you get a chance, could you take another look here? Your last note said everything you'd flagged was covered, and I re-reviewed the current revision — `fitDecimalBytes` now sign-extends negative decimals and rejects oversized values unless the dropped prefix is redundant sign extension, with the added round-trip/precision coverage. I've approved on my side; if you're happy with it, converting your review would let it merge. Thanks!

manifest.go
```diff
@@ -2324,6 +2384,7 @@ func NewDataFileBuilder(
 	recordCount int64,
 	fileSize int64,
 ) (*DataFileBuilder, error) {
+	_ = fieldIDToFixedSize
```
zeroshade, line 2387: Non-blocking: `NewDataFileBuilder` still retains and ignores `fieldIDToFixedSize` (`_ = fieldIDToFixedSize`). Consider dropping the dead parameter now that the write path derives width from the manifest schema.

zeroshade (commented):
Re-reviewed at head — the blocking decimal issues are resolved (negative sign-extension, error-on-oversized, both write paths + the `table/internal` Arrow path, and builder scale via the Avro decode). One non-blocking nit below. Note: #1509 targets the same bug; this PR supersedes it. CI green.

laskoviymishka (approved):

PR #1356, opened 2026-07-02
fix(view): join view metadata location

fallintoplace:
### Summary
- build view metadata paths with `url.JoinPath` instead of raw string concatenation
- add a regression test for trailing slash view locations

### Testing
- `go test ./view`
- `go test ./catalog/sql ./catalog/hive ./catalog/rest`
- `go test ./...`

### DIFF
--- a/view/view.go
+++ b/view/view.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"slices"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Code LGTM — replacing the raw string concat with `net/url.JoinPath` matches the table metadata path handling, and the trailing-slash regression test covers the reported bug. Heads-up: the `integration-test (3.5)` check is currently red on this PR; worth confirming that is an unrelated flake before this goes in.

PR #1357, opened 2026-07-02
fix(metadata): replace updates on unmarshal

fallintoplace:
### Summary
- decode table and view updates into fresh slices before assigning to the receiver
- decode table and view requirements the same way, including the matching view requirement path found during inspection
- add receiver-reuse regressions that also confirm failed decodes do not partially mutate prior results

### Testing
- `go test ./table ./view`
- `go test ./...`

### DIFF
--- a/table/updates.go
+++ b/table/updates.go
@@ -79,6 +79,10 @@ func (u *Updates) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
+	var updates Updates
+	if len(rawUpdates) > 0 {
+		updates = make(Updates, 0, len(rawUpdates))
+	}
 	for _, raw := range rawUpdates {
 		var base baseUpdate
 		if err := json.Unmarshal(raw, &base); err != nil {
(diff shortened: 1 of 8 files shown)


table/updates.go
```diff
@@ -79,6 +79,10 @@ func (u *Updates) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
+	var updates Updates
```
zeroshade, line 82: Optional nit: when the incoming array is empty this assigns a nil slice, so an incoming `[]` round-trips to `null` on re-marshal (same pattern in table/requirements.go and the view/updates.go + view/requirements.go equivalents). If you want `[]` to stay `[]`, allocate the temp with make(Updates, 0, len(rawUpdates)) even when len==0. Not blocking.

zeroshade (approved):
Nice catch on the receiver-reuse bug — decoding into a fresh slice and assigning only after a successful decode is the right fix, and the regression tests cover it. Approving. One optional nit inline.

table/updates.go
fallintoplace: Thank you @zeroshade.

Let me fix this in a few hours. Have a nice weekend.

PR #1358, opened 2026-07-02
fix(scan): use snapshot schema for pruning

fallintoplace:
### Summary
- resolve the effective scan schema from the selected snapshot before local planning
- pass that schema into partition projection, manifest pruning, partition evaluation, metrics pruning, and read-time row filter binding
- keep transaction conflict validation on the current metadata schema by passing it explicitly
- add schema-evolution regressions for time-travel manifest and metrics pruning

### Testing
- `go test ./table -run TestTimeTravel`
- `go test ./table`
- `go test ./...`

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -287,24 +287,11 @@ func (scan *Scan) Snapshot() *Snapshot {
 }
 
 func (scan *Scan) Projection() (*iceberg.Schema, error) {
-	curSchema := scan.metadata.CurrentSchema()
-	curVersion := scan.metadata.Version()
-	if scan.snapshotID != nil {
-		snap := scan.metadata.SnapshotByID(*scan.snapshotID)
-		if snap == nil {
-			return nil, fmt.Errorf("%w: snapshot not found: %d", ErrInvalidOperation, *scan.snapshotID)
-		}
-
-		if snap.SchemaID != nil {
-			for _, schema := range scan.metadata.Schemas() {
-				if schema.ID == *snap.SchemaID {
-					curSchema = schema
-
-					break
-				}
-			}
-		}
+	curSchema, err := scan.effectiveSchema()
+	if err != nil {
+		return nil, err
 	}
+	curVersion := scan.metadata.Version()
 
 	if scan.includeRowLineage && curVersion < minFormatVersionRowLineage {
 		return nil, fmt.Errorf("%w: row lineage requires format version %d, table is v%d",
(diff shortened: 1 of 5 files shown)


table/scanner.go
```diff
@@ -353,6 +340,39 @@ func (scan *Scan) Projection() (*iceberg.Schema, error) {
...
+		return curSchema, nil
+	}
+
+	for _, schema := range scan.metadata.Schemas() {
+		if schema.ID == *snap.SchemaID {
+			return schema, nil
+		}
+	}
+
+	return curSchema, nil
```
zeroshade: When `snap.SchemaID` is non-nil but no matching schema exists, this falls back to `CurrentSchema()`. That differs from the legacy no-schema-id fallback and can silently bind predicates/pruning against the wrong field IDs, incorrectly dropping candidate manifests/files on malformed or partially-loaded metadata. Please only fall back when `snap.SchemaID == nil`; if `snap.SchemaID != nil` and lookup fails, return an `ErrInvalidMetadata`/`ErrInvalidOperation`-wrapped error naming the snapshot ID and missing schema ID. Also add a test for a non-nil-but-unknown snapshot schema-id.

table/scanner.go
```diff
@@ -353,6 +340,39 @@ func (scan *Scan) Projection() (*iceberg.Schema, error) {
...
+			if entry.TimestampMs <= *scan.asOfTimestamp {
+				snap = scan.metadata.SnapshotByID(entry.SnapshotID)
+
+				break
+			}
+		}
+	}
+
+	if snap == nil || snap.SchemaID == nil {
+		return curSchema, nil
```
zeroshade, line 369: For a default scan with no explicit snapshot/as-of, this returns `CurrentSchema()` rather than resolving `CurrentSnapshot().SchemaID`. Is that intentional after schema-only metadata updates? If so, please add a short comment documenting why default scans differ from `WithSnapshotID(currentSnapshotID)`.

zeroshade (changes requested):
Thanks for tightening schema handling for snapshot scans; I found one correctness issue that should be fixed before merge.

- effectiveSchema should not silently fall back to the current schema when a snapshot declares an unknown schema-id.
- Please also document whether default scans intentionally use CurrentSchema rather than CurrentSnapshot().SchemaID after schema-only metadata updates.

table/scanner.go
```diff
@@ -403,24 +423,34 @@ func appendMissingLineageFields(s *iceberg.Schema, lineageFields []iceberg.Neste
...
 	if spec == nil {
 		return nil, fmt.Errorf("%w: id %d", ErrPartitionSpecNotFound, specID)
 	}
-	project := newInclusiveProjection(meta.CurrentSchema(), *spec, caseSensitive)
+	project := newInclusiveProjection(schema, *spec, caseSensitive)
 
 	return project(rowFilter)
 }
 
 func (scan *Scan) buildManifestEvaluator(specID int) (func(iceberg.ManifestFile) (bool, error), error) {
```
tanmayrauth: After this refactor buildManifestEvaluator and buildPartitionEvaluator (scanner.go:468) have no production callers left — PlanFiles goes through fetchPartitionSpecFilteredManifestsWithSchema / collectManifestEntriesWithSchema,  which build evaluators inline via partitionFiltersForSchema(schema). These two methods (and the scan.partitionFilters field at :243 they read) are reachable only from the internal tests now. Worth removing them, unless they're kept deliberately.

table/scanner_internal_test.go
```diff
@@ -392,6 +393,133 @@ func TestBuildPartitionEvaluatorWithInvalidSpecID(t *testing.T) {
...
+	upper := int32Bound(20)
+	manifest := iceberg.NewManifestFile(2, "mem://default/table/metadata/manifest.avro", 100, int32(spec.ID()), snapshotID).
+		Partitions([]iceberg.FieldSummary{{
+			ContainsNull: false,
+			LowerBound:   &lower,
+			UpperBound:   &upper,
+		}}).
+		Build()
+
+	eval, err := scan.buildManifestEvaluator(spec.ID())
```
tanmayrauth: This drives scan.buildManifestEvaluator directly — the dead method above — rather than the production manifest-pruning path in fetchPartitionSpecFilteredManifestsWithSchema. I checked the gap: forcing that production path to use CurrentSchema() (the exact bug this PR fixes) leaves this test and the entire ./table suite green. Routing this through PlanFiles (like the metrics test uses collectManifestEntries) would make it a real regression guard.

tanmayrauth (commented):
Agreed with @zeroshade's points. The unknown-schema-id fall-through to  CurrentSchema does look like it could silently mis-prune, and erroring there would line up with Java's SnapshotUtil.schemaFor. On the default-scan branch — I think using the current schema for live scans and the snapshot schema only for time-travel is right and matches Java/PyIceberg, so maybe just the doc comment he suggested is enough there.

Found a couple of small follow-ons while reading through, both from the same refactor, left them inline.

zeroshade (approved):
Re-reviewed — both points addressed: pruning no longer silently falls back to CurrentSchema when a snapshot references an unknown schema id (it now returns ErrInvalidMetadata naming the snapshot and schema id), and the default-scan behavior is documented. The new regression test covers the unknown-schema-id case. LGTM.

PR #1359, opened 2026-07-03
feat(table): union schema evolution

### LINKED ISSUE #1187: feat(table): add UpdateSchema.UnionByNameWith for name-based schema union
#### Description

`table.UpdateSchema`, right now, does not provide a way to evolve a table's schema to the union of its current schema and an incoming schema matched by field name.

Iceberg-java's `org.apache.iceberg.UpdateSchema` exposes this as `unionByNameWith(Schema)`.

**Proposed Solution**
```go
func (u *UpdateSchema) UnionByNameWith(newSchema *iceberg.Schema) *UpdateSchema
```

Case-sensitivity should follow the existing caseSensitive flag already on UpdateSchema.

**Contribution**

I’d be happy to work on this enhancement. Please feel free to assign this issue to me and I can prepare a PR implementing it.

badalprasadsingh:
### Description

In Iceberg-go, `table.UpdateSchema`, right now, does not provide a way to evolve a table's schema to the union of its current schema and an incoming schema matched by field name.

Iceberg-java's `org.apache.iceberg.UpdateSchema` exposes this as `unionByNameWith(Schema)`. This PR solves the same for Iceberg-Go.

Fixes #1187


badalprasadsingh:
Hey @laskoviymishka , please take a look whenever you get time.

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+		}
+		if existingType.Equals(newType) {
+			return true
+		}
+		_, err := iceberg.PromoteType(newPrimitive, existingType)
+
+		return err == nil
+	}
+
+	_, ok := newType.(iceberg.PrimitiveType)
```
laskoviymishka: I think there's a real bug in this fallthrough. Once we're past the primitive branch, we return `true` (ignorable) whenever `newType` isn't primitive — so `list<int>`→`map<string,int>`, `struct`→`list`, `map`→`struct` all read as no-ops regardless of whether the kinds actually match.

The problem is what happens next: `unionField` then switches on the *new* type and recurses its children against the *existing* field. For `list`→`map` we look up `path.value` in a list, miss, and `unionAddColumn` grafts the map-value subtree onto the list element. The field IDs won't line up with the physical columns, so Java/PyIceberg/rust can't read the result — and for other inputs it surfaces later as a misleading "cannot add field to non-struct type" at the wrong path.

I'd gate this on same-kind: only return `true` when `newType` is the same complex kind as `existingType`, otherwise `false` so `unionUpdateColumn` emits a real (and rejectable) type change. That also transitively fixes the map-key guard below — right now `existing.Type.(*iceberg.MapType)` silently skips the key-immutability check whenever the existing field isn't already a map, so a `list`→`map` never validates the key at all. Worth an explicit error when `newField.Type` is a map but `existing.Type` isn't. Could we add cross-kind cases (`list↔map`, `struct↔list`, `struct↔map`) to lock it in? wdyt?

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+func isIgnorableTypeUpdate(existingType, newType iceberg.Type) bool {
+	if _, ok := existingType.(iceberg.PrimitiveType); ok {
+		newPrimitive, ok := newType.(iceberg.PrimitiveType)
+		if !ok {
+			return false
+		}
+		if existingType.Equals(newType) {
+			return true
+		}
+		_, err := iceberg.PromoteType(newPrimitive, existingType)
```
laskoviymishka: This is the one spot in the file that calls `PromoteType(new, existing)` — every other call, including `updateColumn`, is `PromoteType(existing, new)`. It happens to work here only because the reversal flips the widening test in the direction we want, so it's a latent trap: a future maintainer "fixing" the order to match the convention introduces a silent correctness bug.

I'd either invert the test to match the rest of the file, or drop a comment right here spelling out that the argument order is intentionally reversed and why.

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+
+		return err == nil
```
laskoviymishka: Treating anything `PromoteType` accepts as "ignorable" pulls in promotions Java's `isPromotionAllowed` explicitly rejects. `PromoteType` allows string↔binary, fixed[16]→uuid, and decimal scale *widening* (`scale <= scale`), whereas Java permits only int→long, float→double, and decimal with exact scale equality.

Concretely: existing `string` + incoming `binary` returns ignorable and silently keeps `string`; incoming `uuid` on existing `fixed[16]` stages a `uuid` field over `fixed[16]` physical data. Both write schemas Java/PyIceberg can't resolve. I'd add a narrow `isPromotionAllowed`-style helper mirroring Java's exact switch rather than delegating the whole spec-allowed set to `PromoteType`. wdyt?

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+	}
+
+	if newField.WriteDefault != nil && !reflect.DeepEqual(newField.WriteDefault, existing.WriteDefault) {
+		u.setWriteDefault(existing, newField.WriteDefault)
+	}
+
+	return nil
+}
+
+func (u *UpdateSchema) setWriteDefault(existing iceberg.NestedField, writeDefault any) {
```
laskoviymishka: This writes straight into `u.updates` and skips everything `updateColumn` guards. Two things worry me.

There's no `u.isDeleted(existing.ID)` check — a transaction that deletes a field and then unions a schema with a differing write-default inserts a stale entry for a deleted field. It's harmless today only because delete wins in `applyChanges`, but that's an ordering coincidence, not an invariant.

And taking `writeDefault any` instead of an `iceberg.Literal` means an incompatible default (an `int64` for an `int32` field) is accepted with no type check and can blow up downstream at encode time. I'd route this through the same `ColumnUpdate`/`updateColumn` path the rest of the defaults use, or at minimum add the `isDeleted` guard plus a type-compat check.

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+
+		if _, ok := parentField.Type.(*iceberg.StructType); !ok {
+			return fmt.Errorf("cannot add field to non-struct type: %s", parentFullPath)
+		}
+
+		parentID = parentField.ID
+	}
+
+	name := path[len(path)-1]
+	for _, add := range u.adds[parentID] {
```
laskoviymishka, line 752: `addColumn` also scans `u.updates[parentID]` for a name collision here, and `unionAddColumn` drops that scan. Since `findField` only sees the current schema, not pending renames: if a prior op renamed `old`→`new` and the incoming schema has `new`, `findField("new")` misses, we fall in here, and append a second `new` to `u.adds` — a duplicate name in the applied schema. I'd mirror the `u.updates` scan before the `u.adds` check.

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+	existing, ok := u.findField(strings.Join(path, "."))
+	if !ok {
+		return u.unionAddColumn(path, newField)
+	}
+
+	if err := u.unionUpdateColumn(path, existing, newField); err != nil {
+		return err
+	}
+
+	switch t := newField.Type.(type) {
```
laskoviymishka, line 687: This switch omits `PrimitiveType` and `VariantType`. The implicit no-op is correct (neither has nested fields), but the exhaustive linter CI runs will flag a type switch with no `default`, and `applyChanges.Variant` already handles variant explicitly elsewhere. I'd add a `default:` with a one-line comment noting there's nothing to recurse into, so CI stays green.

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
 	return u
 }
 
+func (u *UpdateSchema) UnionByNameWith(newSchema *iceberg.Schema) *UpdateSchema {
```
laskoviymishka, line 655: Every other exported `UpdateSchema` method carries a godoc comment, and this one has enough load-bearing semantics — forced-optional on new fields, doc/write-default updates, case-sensitivity — that a short doc block would help callers a lot. Worth adding before this goes in.

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+			return fmt.Errorf("field already exists in adds: %s", fullName)
+		}
+	}
+
+	field := iceberg.NestedField{
+		Name:           name,
+		Type:           newField.Type,
+		Required:       false,
+		Doc:            newField.Doc,
+		InitialDefault: newField.InitialDefault,
```
laskoviymishka, line 778: `addColumn` guards initial defaults with a `defaultValue.Type().Equals(t)` check; here we copy `InitialDefault` verbatim with no type-compatibility check. Not blocking, but it'd be nice to keep the two add paths consistent so a mistyped default can't slip through the union route.

laskoviymishka (changes requested):
Nice work here — the ops-queue integration is clean, the `unionField`/`unionAddColumn`/`unionUpdateColumn` recursion tracks the Java visitor structure well, and the happy-path coverage is thorough. This gets us real parity with `unionByNameWith` for the common cases.

I'd hold this PR until we sort out one blocker. `isIgnorableTypeUpdate` returns `true` for any pair of non-primitive types, so a cross-kind change like `list<int>`→`map<string,int>` or `struct`→`list` passes through as a no-op. The recursion then walks the *new* type's structure against the *old* field's children and silently grafts mismatched subfields on — field IDs stop matching the physical columns, so Java/PyIceberg/rust can't read the result, and for other inputs it throws a misleading "cannot add field to non-struct type" at the wrong path. Java sidesteps this entirely because the visitor casts the partner type and throws on a kind mismatch. Gating `isIgnorableTypeUpdate` on same-kind fixes both the corruption and the missing map-key check.

- Same-kind guard in `isIgnorableTypeUpdate` (the blocker above), with cross-kind tests to lock it in.
- The primitive-promotion path leans on `PromoteType`, which accepts string↔binary, fixed→uuid, and decimal scale-widening that Java's `isPromotionAllowed` rejects — I'd mirror Java's exact allowed set.
- `PromoteType` is called `(new, existing)` here, reversed from everywhere else in the file. It works by luck of the flipped test; either invert it to match or comment why it's intentional.
- `setWriteDefault` bypasses `updateColumn` — no `isDeleted` guard and a raw `any` default with no type check. Route it through the normal path or add the guards.
- `unionAddColumn` drops the `u.updates` name-collision scan that `addColumn` has, so a pending rename can produce a duplicate name.
- The type switch in `unionField` needs a `default:` or the exhaustive linter will fail CI.

On tests: alongside the cross-kind cases, I'd add two sequential `UnionByNameWith` calls and a union chained with other ops (e.g. `AddColumn(...).UnionByNameWith(...)` where the added column is also in the incoming schema) — the second-call and dedup behaviors are currently correct-or-not-by-accident and untested. Also the `assert.NotPanics` wrapper in `TestUnionByNameRejectsMapKeyChange` passes regardless of whether `err` is nil as long as nothing panics; `require.Error` directly is what we want there.

table/update_schema.go
zeroshade: Two issues in `isIgnorableTypeUpdate`: (1) the non-primitive branch effectively `return !ok` (around line 796) treats any non-primitive→non-primitive pair as ignorable, so cross-kind changes (list→map, struct→list) become no-ops and then recurse into the wrong child layout — risking corrupted field IDs / invalid schemas. (2) `PromoteType(newPrimitive, existingType)` has its arguments reversed vs. the `updateColumn` convention used elsewhere; even if it works today it's fragile and leans on the broader `PromoteType` allowed set rather than the narrower promotion rules.

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+	if u.updates[parentID] == nil {
+		u.updates[parentID] = make(map[int]iceberg.NestedField)
+	}
+
+	updatedField, ok := u.updates[parentID][existing.ID]
+	if !ok {
+		updatedField = existing
+	}
+	updatedField.WriteDefault = writeDefault
+	u.updates[parentID][existing.ID] = updatedField
```
zeroshade, line 857: `setWriteDefault` writes directly into `u.updates`, bypassing `updateColumn`'s deleted-field guard and default validation, and accepts a raw `any` — so an incompatible write default can enter schema metadata unchecked.

table/update_schema.go
```diff
@@ -616,6 +617,185 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+		Doc:            newField.Doc,
+		InitialDefault: newField.InitialDefault,
+		WriteDefault:   newField.WriteDefault,
+	}
+
+	sch, err := iceberg.AssignFreshSchemaIDs(iceberg.NewSchema(0, field), u.assignNewColumnID)
+	if err != nil {
+		return fmt.Errorf("failed to assign field id: %w", err)
+	}
+	u.adds[parentID] = append(u.adds[parentID], sch.Field(0))
```
zeroshade, line 786: `unionAddColumn` scans `u.adds` for name collisions but not `u.renames` (which `addColumn` does). A chained rename followed by a union can add a duplicate name, because `findField` only sees the original schema.

zeroshade (changes requested):
The union-evolution logic can currently produce invalid schemas — a few correctness issues inline. (laskoviymishka's earlier CHANGES_REQUESTED is also still outstanding.)

table/update_schema.go
badalprasadsingh: Done.

1. Complex-existing now returns true only if `newType` is the same complex kind
2. Cross-kind fails to `unionUpdateColumn` which errors with `cannot change column type: <path>: <from> -> <to>`
3. Map-key guard in `unionField` is also tightened
4. Locked in with `TestUnionByNameRejectsCrossKindChanges`

table/update_schema.go
badalprasadsingh: Done.

Kept the reversal, and, added a godoc block for the same.

table/update_schema.go
badalprasadsingh: Done.

1. Added the `isPromotionAllowed` helper. `isIgnorableTypeUpdate` now delegates the primitive branch to it, and `unionUpdateColumn` gates any emitted type-change on it
2. Added the test `TestUnionByNameRejectsInvalidPromotions` : `string<->binary`, `fixed[16]->uuid`, and `decimal` scale widening now produce `cannot change column type` errors

table/update_schema.go
badalprasadsingh: Done.

1. Added the `isDeleted` guard and complex-typed-column guard.
2. Couldn't do the Full type-vs-type check as, `NestedField.WriteDefault` is `any`, and there is no exposed runtime `any + Type -> Literal` validator (`iceberg.NewLiteral` is generic / compile-time only). <br>
Routing through `updateColumn` as suggested is not possible here, because that path requires an `iceberg.Literal` which we don't have. <br>
Thus, added a clear `TODO` for it. Happy to open a PR for it!

table/update_schema.go
badalprasadsingh: Done.

A pending rename `old->new` followed by a union whose incoming schema contains `new` now errors as `field already exists: new` instead of double-adding.

table/update_schema.go
badalprasadsingh: Done.

table/update_schema.go
badalprasadsingh: Done.

table/update_schema.go
badalprasadsingh: Done.

Added a structural-guard as same as `addColumn`: complex-typed fields with a non-nil `InitialDefault` or `WriteDefault` are rejected via `default values are not supported for <type>`. Full value-vs-type verification for primitive is limited by the same missing-Literal-constructor issue, as I have documented in the `TODO` above `setWriteDefault`.

table/update_schema.go
badalprasadsingh: Done.

1. The non-primitive branch now returns true only when `newType` is the same complex kind as `existingType`. Cross-kind pairs return false and fall through to `unionUpdateColumn`, which errors with `cannot change column type`
2. The primitive branch no longer calls `iceberg.PromoteType`; it calls `isPromotionAllowed` helper which is strictly narrower than the global `PromoteType` allow-set
3. For the intentionally-reversed argument, added a godoc block
4. Regression-locked by `TestUnionByNameRejectsCrossKindChanges` and `TestUnionByNameRejectsNonJavaPromotions`

table/update_schema.go
badalprasadsingh: Done.

1. This now returns `error` propagated by `unionUpdateColumn`.
2. Rejects writes against a deleted field.
3. Rejects non-primitive columns.
4. For your last concern, here is a limitation, `NestedField.WriteDefault` is `any`, and `iceberg.NewLiteral` is generic / compile-time only - there is no runtime `any` to `Literal` converter to check compatibility. <br>
 Here, `updateColumn` has existing default validation which only fires when the caller supplies an `iceberg.Literal`, which is unavailable here because we've received an already-unwrapped `any` from the incoming schema. <br>
 Have added a `TODO` for it. Happy to open a PR for it!

table/update_schema.go
badalprasadsingh: Done.

badalprasadsingh:
@laskoviymishka and @zeroshade , thank you so much for the reviews! Please take a look whenever you get time.

zeroshade (changes requested):
Re-reviewing the current revision — this is still outstanding from my earlier review:

**Medium — union-by-name default handling bypasses typed validation** (`table/update_schema.go:779`)

`UnionByNameWith` copies incoming `InitialDefault`/`WriteDefault` values directly from `NestedField` as raw `any`, and `setWriteDefault` does the same for existing fields. Unlike `AddColumn`, this never validates that primitive defaults match the field type, so an invalid union schema can be accepted and later panic or produce invalid metadata on write paths.

**Suggested fix:** add runtime validation/conversion for raw defaults against the target Iceberg type before staging added columns or write-default updates, and reject mismatches during `Apply`.

badalprasadsingh:
Done. Removed the `TODO` that I had added for it. Couldn't reuse `AddColumn's Literal.Type().Equals` check directly since union defaults are raw `any`. 

Added a `validateDefaultValue` helper that runs each raw default through the same `defaultToScalar` conversion used on the read/write paths, so anything that would panic downstream is now rejected during Apply instead. Added it into `unionAddColumn` and `setWriteDefault`. Added respective tests for it as well.

zeroshade (changes requested):
Re-reviewing — the default handling has changed since my last review, but two issues remain:

**Medium — write default validated against pre-promotion type** (`table/update_schema.go:847`)
`setWriteDefault` validates the incoming default against `existing.Type` even when the same union op just promoted the column. A default valid for the promoted type but not the old one (e.g. a wider `decimal(18,2)` default during `decimal(9,2) -> decimal(18,2)`) is wrongly rejected.
*Fix:* validate against the pending updated field type (pass `newField.Type` after the promotion is staged).

**Medium — valid complex-type defaults rejected** (`table/update_schema.go:767`)
New struct/list/map fields with non-null `InitialDefault`/`WriteDefault` are rejected, even though the metadata-compatibility code accepts complex defaults as JSON object/array values (and the new API comment says defaults are preserved).
*Fix:* replace the blanket non-primitive rejection with validation matching the repo's metadata-compatibility rules for complex defaults.

table/update_schema.go
```diff
@@ -631,6 +632,368 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+//     change errors.
+//   - A Doc is updated only when the incoming Doc is non-empty and differs (an
+//     empty Doc never clears an existing one).
+//   - A WriteDefault is updated when the incoming value differs; an existing
+//     column's InitialDefault is never modified.
+//   - Map keys are immutable and cross-kind changes (list->map, struct->list,
+//     etc.) are rejected rather than silently grafted onto the existing column.
+//
+// The change is queued and applied with other pending updates on Apply/Commit,
+// so fields added by an earlier op in the same UpdateSchema are visible here.
```
laskoviymishka: This sentence isn't quite true, and I think it'll trip someone up. `findField` only consults `u.schema`, not `u.adds` — so if an earlier op added a struct and a later `UnionByNameWith` tries to extend it, the lookup misses and we error with `parent field not found`, or worse `field already exists in adds: <container>`, which names the container rather than the child and reads like a bug.

Teaching `findField` about `u.adds` is more than this PR should take on. For now I'd drop this sentence and note that schemas should be pre-merged before a single call — and maybe add a test pinning the current limitation, so the boundary is captured rather than implied by that comment in the test file. wdyt?

table/update_schema.go
```diff
@@ -631,6 +632,368 @@ func (u *UpdateSchema) SetIdentifierField(paths [][]string) *UpdateSchema {
...
+		}
+		if err := validateNestedDefaults(newField.Type); err != nil {
+			return err
+		}
+	} else {
+		if err := validateDefaultValue(newField.Type, newField.InitialDefault); err != nil {
+			return fmt.Errorf("invalid initial-default for %s: %w", fullName, err)
+		}
+		if err := validateDefaultValue(newField.Type, newField.WriteDefault); err != nil {
+			return fmt.Errorf("invalid write-default for %s: %w", fullName, err)
```
laskoviymishka: agreed with zeroshade that the raw-`any` default gap on `setWriteDefault` is still the thing to close here. Re-reading this pass I noticed it's not only the write path: this copies `InitialDefault` (and `WriteDefault` just below) straight off `newField` with no check against `newField.Type`, so a mistyped `any` — say an int64 for an `Int32Type` — lands in the metadata silently, the same failure mode as the `setWriteDefault` case the TODO flags but not covered by that TODO.

`InitialDefault` is the worse one, since it can't change once set — a wrong value is permanently wrong. `AddColumn` routes defaults through a typed `iceberg.Literal` and validates `Type().Equals(t)`; I'd want the union add path to go through that same typed check rather than copying the raw `any`. wdyt?

laskoviymishka (changes requested):
This is really close now, almost everything from last round landed, and it landed cleanly.

Everything I asked for in the first pass is in:
- the same-kind guard in `isIgnorableTypeUpdate`
- the local `isPromotionAllowed` (int→long, float→double, decimal same-scale wider-precision, exact match) replacing the permissive `PromoteType`
- the pending-rename collision scan in `unionAddColumn`
- the exhaustive-switch `default` case in `unionField`
- the cross-kind / sequential-call / dedup tests, and the `require.Error` fix on the map-key test

The reversed-arg `isPromotionAllowed(newPrimitive, existingPrim)` call in `isIgnorableTypeUpdate` reads correctly too, and the comment explaining the ordering is exactly what I was after — thanks for that.

One blocking thing is still open from last round: the raw-`any` default validation. zeroshade re-flagged it on the 16th and I'm with him — `setWriteDefault` still copies an untyped `any` with no check against the field's primitive type, and the TODO honestly captures why. Re-reading this pass I noticed it's not just the write path: `unionAddColumn` copies `InitialDefault` straight off the incoming field too, and the TODO doesn't cover that one. `InitialDefault` is the one I'd worry about most, since it can't change once set. I'd route both through the same typed-`Literal` check `AddColumn` uses before this merges. Details inline.

One thing I hadn't caught last round: the doc comment claims fields added by an earlier op in the same `UpdateSchema` are visible to a later `UnionByNameWith`, but `findField` only looks at `u.schema`, so that isn't actually true for extending a struct added by a prior op. Not asking you to make `findField` adds-aware here — just correct the doc and maybe pin the limitation with a test. Also inline.

Fix the default validation and tighten that doc and I'm happy to approve — we're basically there.

table/update_schema.go
badalprasadsingh: Yes you are right. 

Done. Corrected the Doc.

table/update_schema.go
badalprasadsingh: Done.

Added `InitialDefault` validation on the add path as well, rejecting native defaults with mismatched types (e.g. `int64` for `Int32Type`) and covered it with a **test**.

badalprasadsingh:
> **Medium — write default validated against pre-promotion type** (`table/update_schema.go:847`) `setWriteDefault` validates the incoming default against `existing.Type` even when the same union op just promoted the column. A default valid for the promoted type but not the old one (e.g. a wider `decimal(18,2)` default during `decimal(9,2) -> decimal(18,2)`) is wrongly rejected. _Fix:_ validate against the pending updated field type (pass `newField.Type` after the promotion is staged).

Fixed. `unionUpdateColumn` now validates the write-default against the promoted type when the same op stages a type change (falling back to the existing type otherwise) and passes it into `setWriteDefault`. 

Added the test `TestUnionByNamePromoteDecimalKeepsWiderWriteDefault` for the same.

> **Medium — valid complex-type defaults rejected** (`table/update_schema.go:767`) New struct/list/map fields with non-null `InitialDefault`/`WriteDefault` are rejected, even though the metadata-compatibility code accepts complex defaults as JSON object/array values (and the new API comment says defaults are preserved). _Fix:_ replace the blanket non-primitive rejection with validation matching the repo's metadata-compatibility rules for complex defaults.

Done. It now validates `InitialDefault` on the add path as well. Added test `TestUnionByNameRejectsMistypedNativeDefault` for the same.

badalprasadsingh:
Hey @zeroshade & @laskoviymishka, thanks for your patience, and sorry for the delay. I've addressed all the review comments now. Whenever you have some time, I'd really appreciate another look.

Also, while addressing the comments, I noticed: setting an _initial-default_ or _write-default_ on a column of type `timestamptz` or `timestamptz_ns` is always rejected, even for valid values. 

EDIT: This happens because the default is validated with `defaultValue.Type().Equals(...)`, and the only literal that carries a timestamp value is `TimestampLiteral/TimestampNsLiteral`, whose `Type()` is the `non-tz` variant — so the check can never pass for a `tz` column.`UnionByNameWith` works same as `AddColumn` right now, so it affects both paths. Happy to address this issue for both of them separately in a new PR. 

Again, thanks a lot!

laskoviymishka (approved):
Looks good now!

table/update_schema.go
zeroshade: Re-checked at current head (`8065414`) — I don't think this is fully resolved for the **rename** case, so flagging before it's marked done.

`unionAddColumn` does now scan `u.updates[parentID]` for a name collision, but the scan is gated behind `findField(fullName)`:

```go
// guard against colliding with a pending rename
if field, ok := u.findField(fullName); ok {   // false when the new name is a pending rename target
    if !u.isDeleted(field.ID) {
        for _, upd := range u.updates[parentID] {
            if upd.Name == name {
                return fmt.Errorf("field already exists: %s", fullName)
            }
        }
    }
}
```

`findField` only consults the current schema, so for a *pending* rename the new name isn't there yet and the whole block is skipped:

```go
schema = { a }
UpdateSchema().RenameColumn([]string{"a"}, "b").UnionByNameWith(schema{ b })
```

`unionField(["b"])` → `findField("b")` is false → `unionAddColumn(["b"])` → the guarded scan is skipped → a second `b` is added. Net pending state: rename `a→b` **and** add `b`.

Compare `updateColumn`'s own rename guard, which scans `u.updates[parentID]` unconditionally (not gated on `findField`):

```go
for _, upd := range u.updates[parentID] {
    if upd.Name == update.Name.Val && upd.ID != field.ID {
        return fmt.Errorf("cannot rename field to renamed field: %s", newFullName)
    }
}
```

Suggested fix: move the `u.updates[parentID]` name-collision scan out of the `if findField(...)` block so it runs unconditionally (the same latent gap exists in `addColumn`), and add a `RenameColumn(a→b)` + `UnionByNameWith(schema{b})` regression test — the current test only exercises a prior `AddColumn`, not a prior `RenameColumn`.

zeroshade (approved):

PR #1361, opened 2026-07-03
fix(table): bound manifest merge concurrency

fallintoplace:
## Summary

Bound manifest merge bin processing with an errgroup concurrency limit instead of starting one goroutine per packed bin.

The limit defaults to runtime.GOMAXPROCS(0) and can be tuned with the table property commit.manifest-merge.max-concurrency. Non-positive or invalid values fall back to the default.

## Why

Large commits can produce many manifest merge bins. Launching all bin merges at once can spike goroutines, file handles, CPU, and memory during commit even though the work is naturally parallel only up to a reasonable worker limit.

## Testing

- go test ./table -run 'TestManifestMerge(MaxConcurrency|GroupLimits)' -count=1
- go test ./table -count=1
- go test ./... -skip '^TestCreateAzureBucketManagedIdentityCredentialCalled$'
- git diff --check

### DIFF
--- a/table/properties.go
+++ b/table/properties.go
@@ -70,6 +70,9 @@ const (
 	ManifestMinMergeCountKey     = "commit.manifest.min-count-to-merge"
 	ManifestMinMergeCountDefault = 100
 
+	ManifestMergeMaxConcurrencyKey     = "commit.manifest-merge.max-concurrency"
+	ManifestMergeMaxConcurrencyDefault = 0
+
 	WritePartitionSummaryLimitKey     = "write.summary.partition-limit"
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM. Bounding the manifest-merge fan-out with `errgroup.Group.SetLimit` is the right approach, and normalizing non-positive config to `runtime.GOMAXPROCS(0)` avoids the `SetLimit(0)` deadlock. The worker-cap behavior test covers it. Optional/non-blocking: table-driven cases for negative or invalid string property values.

PR #1362, opened 2026-07-03
fix(sql): return error for unsupported dialect

fallintoplace:
## Summary
- Changed SQL catalog dialect resolution to return typed errors instead of panicking for unsupported dialects.
- Updated registry usage to rely on clean error flow from \'sql.NewCatalog\'.
- Added test for direct catalog creation via NewCatalog with unsupported dialect.

## Testing
- go test ./catalog/sql -run 'TestInvalidDialect|TestNewCatalogInvalidDialect' -count=1
- go test ./catalog/sql -count=1

### DIFF
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -86,12 +86,6 @@ func init() {
 			return nil, err
 		}
 
-		defer func() {
-			if r := recover(); r != nil {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Replacing the unsupported-dialect panic with a proper error return is the right call, the supported dialect set is preserved, and both the loader and direct-constructor paths are covered by tests.

PR #1363, opened 2026-07-03
fix(transforms): guard bucket transformer unsupported source types

fallintoplace:
## Summary
- Fixes `BucketTransform.Transformer` for unsupported source types by returning an invalid optional instead of calling a nil hash function.
- Keeps the existing method signature so public callers are unaffected while avoiding latent panics.
- Adds regression coverage for unsupported source types through `TestBucketTransformUnsupportedTypeReturnsInvalidOptional`.

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -425,6 +425,10 @@ func (t BucketTransform) Project(name string, pred BoundPredicate) (UnboundPredi
 		return projectTransformPredicate(t, name, pred)
 	}
 
+	if !t.CanTransform(pred.Term().Type()) {
+		return nil, nil
+	}
+
 	transformer := t.Transformer(pred.Term().Type())
 	switch p := pred.(type) {
 	case BoundUnaryPredicate:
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. The guard matches the spec source-type set for bucket transforms and Project no longer builds unsupported direct predicates. Optional: a broader table-driven matrix over every unsupported source type would be nice, but not blocking.

zeroshade:
Thanks @fallintoplace — this is approved, but it now conflicts with main after recent merges. Could you rebase onto the latest main and resolve the conflicts so we can get it in?

zeroshade:
Thanks @fallintoplace — this LGTM and already has an approval, but it now conflicts with `main` after the recent merges. Could you rebase onto the latest `main` and resolve the conflicts so we can get it merged? 🙏

fallintoplace:
@zeroshade I am not sure if GitHub is behind or outdated. Let me have a look.
image

PR #1364, opened 2026-07-03
fix(dv): reject negative deletion-vector positions in DVWriter.Add

fallintoplace:
## Summary
- Validate DVWriter.Add positions are non-negative.
- Change DVWriter.Add to return an error on negative positions.
- Plumb error handling where DVWriter.Add is called.
- Add regression coverage for negative position rejection in dv_writer tests.

## Testing
- go test ./table/dv -run 'TestDVWriter'
- go test ./table -run 'TestScanPruningWithDeletionVector|TestReadTaskDeletionVectorSupersedesPositionalDeletes|TestRewriteDataFilesPreservesSiblingDeletionVector'
- go test ./table ./table/dv

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1867,7 +1867,11 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
 
 					return
 				}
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
LGTM. Rejecting negative positions at the DVWriter.Add entry point before mutating writer state is correct, the production caller now propagates the error, and the negative/zero cases are well covered.

PR #1365, opened 2026-07-03
fix(transforms): validate BucketTransform NumBuckets

fallintoplace:
## Summary
- Make `BucketTransform` validate `NumBuckets` before modulo and path/transform application.
- Guard `Apply`, `Transformer`, and `Project` so invalid values (`NumBuckets <= 0`) return safe, non-panicking outcomes.
- Return invalid `Optional` from `Apply`/`Transformer` and `ErrInvalidArgument` from `Project` when bucket count is invalid.
- Add regression tests covering invalid bucket counts for `Apply`, `Transformer`, and `Project`.
- Harden unsupported source-type handling in `Transformer` to return invalid values instead of panicking.

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -244,6 +244,14 @@ func (t BucketTransform) MarshalText() ([]byte, error) {
 
 func (t BucketTransform) String() string { return fmt.Sprintf("bucket[%d]", t.NumBuckets) }
 
+func (t BucketTransform) validateNumBuckets() error {
+	if t.NumBuckets <= 0 {
+		return fmt.Errorf("%w: bucket transform requires numBuckets > 0", ErrInvalidArgument)
+	}
+
+	return nil
+}
+
 func (BucketTransform) CanTransform(t Type) bool {
 	switch t.(type) {
 	case Int32Type,
@@ -292,6 +300,9 @@ func (t BucketTransform) Apply(value Optional[Literal]) Optional[Literal] {
 	if !value.Valid {
 		return Optional[Literal]{}
 	}
+	if err := t.validateNumBuckets(); err != nil {
+		return Optional[Literal]{}
+	}
 
 	var hash uint32
 	switch v := value.Val.(type) {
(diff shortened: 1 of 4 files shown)


transforms.go
```diff
@@ -244,6 +244,14 @@ func (t BucketTransform) MarshalText() ([]byte, error) {
 
 func (t BucketTransform) String() string { return fmt.Sprintf("bucket[%d]", t.NumBuckets) }
 
+func (t BucketTransform) validateNumBuckets() error {
```
zeroshade, line 251: [MAJOR] This helper is used for Apply/Transformer/Project, but BucketTransform.MarshalText (~line 241) and NewPartitionSpecID in partitions.go (~line 353) still allow a direct BucketTransform{NumBuckets <= 0} to be stored and serialized as invalid "bucket[0]"/"bucket[-1]". Only ParseTransform / JSON-unmarshal validate. Please validate in BucketTransform.MarshalText (return ErrInvalidArgument for NumBuckets<=0) and/or in the error-returning spec construction paths (NewPartitionSpecOpts/addSpecFieldInternal), or add a validating NewBucketTransform constructor. Add tests for negative direct structs and for json.Marshal/spec construction rejecting NumBuckets<=0.

zeroshade (changes requested):
Thanks for adding the divide-by-zero guard in Apply/Transformer/Project; that is good. One construction/serialization path still needs to be closed before this is safe.

fallintoplace:
Implemented the remaining validation gap in this PR.

What I added on top of bbc3571:
- BucketTransform.MarshalText now validates NumBuckets and returns ErrInvalidArgument for <=0.
- Added transform validation in partition construction (addSpecFieldInternal) so NewPartitionSpecOpts paths reject invalid bucket transforms with the same invalid-argument root cause.
- Added regression tests:
  - TestBucketTransform_MarshalTextRejectsInvalidBuckets
  - TestPartitionSpecRejectsInvalidBucketTransform
  - TestPartitionSpec_MarshalTextRejectsInvalidBucketTransform

Validation run:
- go test ./... -count=1 ✅

zeroshade (approved):
Re-reviewed — my earlier point is addressed: MarshalText and the error-returning spec construction now validate NumBuckets, and json.Marshal / NewPartitionSpecOpts reject non-positive counts, with tests. LGTM.

PR #1366, opened 2026-07-03
fix(literals): return BinaryLiteral for FixedLiteral→binary cast

fallintoplace:
## Summary

- FixedLiteral.To(BinaryType) now returns a copied BinaryLiteral value instead of returning FixedLiteral directly.
- Added regression coverage in TestFixedLiteral to verify:
  - cast result type is BinaryLiteral
  - target type is binary
  - conversion returns an independent byte slice from the source fixed bytes

This avoids type mismatch issues during predicate binding/serialization paths when casting fixed literals to binary.

### DIFF
--- a/literals.go
+++ b/literals.go
@@ -27,6 +27,7 @@ import (
 	"math"
 	"math/big"
 	"reflect"
+	"slices"
 	"strconv"
 	"time"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Fixed-to-binary now correctly returns a BinaryLiteral with cloned bytes, and the regression test verifies the type, byte values, and no aliasing.

PR #1367, opened 2026-07-03
fix(sort): validate sort source IDs during unmarshal

fallintoplace:
## Summary

- Require exactly one of `source-id` or `source-ids` in `SortField.UnmarshalJSON`.
- Reject malformed sort fields at unmarshal time when source IDs are missing, empty, zero, or negative.
- Tightened source ID validation (`> 0`) to prevent invalid source ID 0 from being treated as a valid value.
- Updated sort tests to assert rejection and add coverage for these invalid paths.

## Testing

- `go test ./table -run 'TestSortField|TestUnmarshalSortOrder|TestNewSortOrder|TestSortOrder' -count=1`
- `go test ./table -count=1`
- `go test ./... -count=1`

### DIFF
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -139,6 +139,9 @@ func (s *SortField) UnmarshalJSON(b []byte) error {
 	if hasSourceID && hasSourceIDs {
 		return errors.New("sort field cannot contain both source-id and source-ids")
 	}
+	if !hasSourceID && !hasSourceIDs {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Validation is syntactic (presence and positivity of source ids) at unmarshal time while schema-membership stays in CheckCompatibility, so valid positive-id sort orders still round-trip. Good coverage on the constructor and unmarshal paths.

PR #1368, opened 2026-07-04
fix(table): separate rolling data writer close/abort paths

fallintoplace:
## Summary
- Split rolling-data writer lifecycle into graceful close and abort paths.
- Graceful close now only closes writer input (no context cancellation), while abort cancels context and closes input.
- fanout writers now use a decoupled context for coordination so `errgroup.Wait()` does not cancel post-success draining.
- writerFactory.closeAll uses graceful close so normal completion can flush queued records.
- Unpartitioned write error paths now use abort path; success path now uses `closeAndWait` for consistent lifecycle behavior.
- Added regression tests for normal close flushing, context-cancel fanout regression, and abort cancellation behavior.

## Testing
- `go test ./table -count=1`

### DIFF
--- a/table/partitioned_fanout_writer.go
+++ b/table/partitioned_fanout_writer.go
@@ -80,16 +80,17 @@ func (p *partitionedFanoutWriter) Write(ctx context.Context, workers int) iter.S
 	inputRecordsCh := make(chan arrow.RecordBatch, workers)
 	outputDataFilesCh := make(chan iceberg.DataFile, workers)
 
-	fanoutWorkers, ctx := errgroup.WithContext(ctx)
-	startRecordFeeder(ctx, p.itr, fanoutWorkers, inputRecordsCh)
+	fanoutWorkers, fanoutCtx := errgroup.WithContext(ctx)
+	writerCtx, writerCancel := context.WithCancel(ctx)
+	startRecordFeeder(fanoutCtx, p.itr, fanoutWorkers, inputRecordsCh)
 
 	for range workers {
 		fanoutWorkers.Go(func() error {
-			return p.fanout(ctx, inputRecordsCh, outputDataFilesCh)
+			return p.fanout(fanoutCtx, writerCtx, inputRecordsCh, outputDataFilesCh)
 		})
 	}
 
-	return p.yieldDataFiles(fanoutWorkers, outputDataFilesCh)
+	return p.yieldDataFiles(fanoutWorkers, outputDataFilesCh, writerCancel)
 }
 
 func startRecordFeeder(ctx context.Context, itr iter.Seq2[arrow.RecordBatch, error], fanoutWorkers *errgroup.Group, inputRecordsCh chan<- arrow.RecordBatch) {
@@ -115,7 +116,7 @@ func startRecordFeeder(ctx context.Context, itr iter.Seq2[arrow.RecordBatch, err
 	})
 }
 
-func (p *partitionedFanoutWriter) fanout(ctx context.Context, inputRecordsCh <-chan arrow.RecordBatch, dataFilesChannel chan<- iceberg.DataFile) error {
+func (p *partitionedFanoutWriter) fanout(ctx context.Context, writerCtx context.Context, inputRecordsCh <-chan arrow.RecordBatch, dataFilesChannel chan<- iceberg.DataFile) error {
 	for {
 		select {
 		case <-ctx.Done():
@@ -126,7 +127,7 @@ func (p *partitionedFanoutWriter) fanout(ctx context.Context, inputRecordsCh <-c
 				return nil
 			}
 
-			if err := p.processRecord(ctx, record, dataFilesChannel); err != nil {
+			if err := p.processRecord(ctx, writerCtx, record, dataFilesChannel); err != nil {
 				return err
 			}
 		}
@@ -136,7 +137,7 @@ func (p *partitionedFanoutWriter) fanout(ctx context.Context, inputRecordsCh <-c
 // processRecord partitions a single record batch and writes sub-batches to
 // the appropriate rolling data writers. The record is released when this
 // function returns, bounding Arrow memory to one batch per fanout worker.
-func (p *partitionedFanoutWriter) processRecord(ctx context.Context, record arrow.RecordBatch, dataFilesChannel chan<- iceberg.DataFile) error {
+func (p *partitionedFanoutWriter) processRecord(ctx context.Context, writerCtx context.Context, record arrow.RecordBatch, dataFilesChannel chan<- iceberg.DataFile) error {
 	defer record.Release()
 
 	partitions, err := p.getPartitions(record)
@@ -157,7 +158,7 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, record arro
 		}
 
 		partitionPath := p.partitionPath(val.partitionRec)
-		rollingDataWriter, err := p.writerFactory.getOrCreateRollingDataWriter(ctx, partitionPath, val.partitionValues, dataFilesChannel)
+		rollingDataWriter, err := p.writerFactory.getOrCreateRollingDataWriter(writerCtx, partitionPath, val.partitionValues, dataFilesChannel)
 		if err != nil {
 			partitionRecord.Release()
 
@@ -174,18 +175,37 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, record arro
 	return nil
 }
 
-func (p *partitionedFanoutWriter) yieldDataFiles(fanoutWorkers *errgroup.Group, outputDataFilesCh chan iceberg.DataFile) iter.Seq2[iceberg.DataFile, error] {
-	return yieldDataFiles(p.writerFactory, fanoutWorkers, outputDataFilesCh)
+func (p *partitionedFanoutWriter) yieldDataFiles(fanoutWorkers *errgroup.Group, outputDataFilesCh chan iceberg.DataFile, writerCancel context.CancelFunc) iter.Seq2[iceberg.DataFile, error] {
+	return yieldDataFiles(
+		p.writerFactory,
+		fanoutWorkers,
+		outputDataFilesCh,
+		p.writerFactory.closeAll,
+		p.writerFactory.abortAll,
+		writerCancel,
+	)
 }
 
-func yieldDataFiles(writerFactory *writerFactory, fanoutWorkers *errgroup.Group, outputDataFilesCh chan iceberg.DataFile) iter.Seq2[iceberg.DataFile, error] {
+func yieldDataFiles(
+	writerFactory *writerFactory,
+	fanoutWorkers *errgroup.Group,
+	outputDataFilesCh chan iceberg.DataFile,
+	closeAll func() error,
+	abortAll func(),
+	writerCancel context.CancelFunc,
+) iter.Seq2[iceberg.DataFile, error] {
 	// Use a channel to safely communicate the error from the goroutine
 	// to avoid a data race between writing err in the goroutine and reading it in the iterator.
 	errCh := make(chan error, 1)
 	go func() {
 		defer close(outputDataFilesCh)
+		defer writerCancel()
 		err := fanoutWorkers.Wait()
-		err = errors.Join(err, writerFactory.closeAll())
+		if err != nil {
+			abortAll()
+		} else {
+			err = errors.Join(err, closeAll())
+		}
 		errCh <- err
 		close(errCh)
 	}()
--- a/table/partitioned_fanout_writer_test.go
+++ b/table/partitioned_fanout_writer_test.go
@@ -175,6 +175,7 @@ func (s *FanoutWriterTestSuite) TestCloseAllFlushesAfterFanoutSuccessContextCanc
 		for dataFile, iterErr := range dataFiles {
 			if iterErr != nil {
 				resultCh <- result{err: iterErr}
+
 				return
 			}
 			total += dataFile.Count()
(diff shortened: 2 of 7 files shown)


table/rolling_data_writer.go
```diff
@@ -440,13 +441,22 @@ func (r *RollingDataWriter) sendError(err error) {
...
+}
+
+// abort cancels in-flight work and then closes input so the stream can exit.
+func (r *RollingDataWriter) abort() {
 	r.cancel()
-	close(r.recordCh)
+	r.closeInput()
 }
 
 func (r *RollingDataWriter) closeAndWait() error {
```
zeroshade, line 468: [BLOCKER] This graceful path avoids calling `cancel`, but it still does not cover the fanout case: in `partitioned_fanout_writer.go` the rolling writers are created from the `errgroup.WithContext` context around line 160, and `fanoutWorkers.Wait()` cancels that context before `closeAll()` runs around line 188. Records drained here can still run through `stream()`/`openFileWriter`/`ToRequestedSchema`/`SortRecordBatch` with a canceled `r.ctx` and fail with `context.Canceled`. Fix by decoupling the fanout coordination context from the per-writer drain context: after a successful `Wait()`, drain with a context not canceled by `errgroup.Wait()`; on `Wait()` error or consumer abort, use the abort path that cancels writers before closing input.

table/rolling_data_writer.go
```diff
@@ -440,13 +441,22 @@ func (r *RollingDataWriter) sendError(err error) {
...
-	r.close()
+	r.closeInput()
 	r.factory.writers.Delete(r.partitionKey)
```
zeroshade, line 470: [MINOR] Normal `closeAndWait` no longer calls the writer `CancelFunc` at all, so the `context.WithCancel` child lives until its parent is canceled. Call `cancel` after the stream finishes draining, e.g. after `wg.Wait()`.

table/rolling_data_writer_test.go
```diff
@@ -187,6 +187,67 @@ func (s *RollingDataWriterTestSuite) TestRollsMultipleFiles() {
...
+		{Name: "id", Type: arrow.PrimitiveTypes.Int32, Nullable: true},
+		{Name: "name", Type: arrow.BinaryTypes.String, Nullable: true},
+	}, nil)
+
+	loc := filepath.ToSlash(s.T().TempDir())
+	factory, _ := s.createWriterFactory(loc, arrSchema, 1024*1024)
+
+	outputCh := make(chan iceberg.DataFile, 10)
+	writerCtx, cancel := context.WithCancel(s.ctx)
+	defer cancel()
```
zeroshade, line 248: [MINOR] This regression test derives the writer context from a live context and calls `closeAll()` before canceling the parent, so it does not exercise the fanout-after-`errgroup`-cancel scenario. Add a test where records are queued, the fanout workers return successfully and cancel the `errgroup` context via `Wait()`, and `closeAll()` must still flush all rows without `context.Canceled`.

zeroshade (changes requested):
Thanks for splitting graceful close from abort; that is the right API direction. However, this does not fully fix the underlying bug from PR #1225: the fanout rolling writers are still derived from the errgroup context, and errgroup cancels that context when Wait() returns before closeAll() drains those writers. As a result, graceful drain can still observe context.Canceled. Both #1225 and this PR need the same deeper fix so successful fanout completion drains with a context that is not canceled by errgroup.Wait().

zeroshade (approved):
Re-reviewed — the main concern is resolved: the fanout coordination context and the per-writer drain context are now decoupled (fanoutCtx vs writerCtx), so graceful drain after the errgroup Wait no longer observes a canceled context, and there is a fanout regression test. This also addresses the underlying issue from #1225. One optional, non-blocking follow-up: the unpartitionedWrite success path still closes recordCh directly rather than going through closeAndWait, so it never calls the writer CancelFunc — worth routing it through closeAndWait for consistency. Approving.

zeroshade:
Thanks @fallintoplace — this is approved and it resolves the #1225 fanout graceful-drain issue, but it now conflicts with main after recent merges. Could you rebase onto the latest main? While you are at it, the earlier optional follow-up still stands: route the unpartitionedWrite success path through closeAndWait so it calls the writer CancelFunc for consistency.

laskoviymishka (approved):

table/partitioned_fanout_writer_test.go
```diff
@@ -92,6 +93,101 @@ func (s *FanoutWriterTestSuite) createCustomTestRecord(arrSchema *arrow.Schema,
...
+	payload := strings.Repeat("p", payloadSize)
+	for i := range rows {
+		bldr.Field(0).(*array.Int64Builder).Append(idOffset + int64(i))
+		bldr.Field(1).(*array.StringBuilder).Append(payload)
+	}
+
+	return bldr.NewRecordBatch()
+}
+
+func (s *FanoutWriterTestSuite) TestCloseAllFlushesAfterFanoutSuccessContextCancel() {
```
laskoviymishka, line 109: This test doesn't exercise the race it's guarding against. The name says "ContextCancel" but nothing here cancels a context — it submits 10000 rows through a single writer with a 256MB target and asserts they all land. On local I/O the stream drains well before `fanoutWorkers.Wait()` returns, so the old errgroup-ctx cancellation never catches an in-flight write. This passes on the pre-fix code too, which means it isn't defending the fix.

To actually pin the bug I'd drop to the `RollingDataWriter` level: inject a `FileWriter` (or a write-path hook) that blocks mid-write, cancel while the stream is parked in it, then assert `closeAll()` flushes every record and `abortAll()` discards them. That's the invariant this PR changes, and it's doable entirely in `rolling_data_writer_test.go`. wdyt?

table/partitioned_fanout_writer.go
```diff
@@ -115,7 +116,7 @@ func startRecordFeeder(ctx context.Context, itr iter.Seq2[arrow.RecordBatch, err
 	})
 }
 
-func (p *partitionedFanoutWriter) fanout(ctx context.Context, inputRecordsCh <-chan arrow.RecordBatch, dataFilesChannel chan<- iceberg.DataFile) error {
+func (p *partitionedFanoutWriter) fanout(ctx context.Context, writerCtx context.Context, inputRecordsCh <-chan arrow.RecordBatch, dataFilesChannel chan<- iceberg.DataFile) error {
```
laskoviymishka, line 119: Two adjacent `context.Context` params through `fanout`/`processRecord`/`processBatch` makes me nervous — Go has no named args at the call site, so a future reorder swaps them silently and the compiler won't blink. And a swap here reintroduces exactly the race this PR is fixing. Every test call already reads `processBatch(ctx, ctx, ...)`, which shows how easy the confusion is.

Since `writerCtx` has the same lifetime as the `Write()` call, I'd attach it to the `partitionedFanoutWriter` / `positionDeletePartitionedFanoutWriter` struct during `Write()` and drop it from these signatures entirely. wdyt?

table/rolling_data_writer.go
```diff
@@ -450,26 +451,64 @@ func (r *RollingDataWriter) sendError(err error) {
...
+		if closeErr := writer.closeAndWait(); closeErr != nil && err == nil {
+			err = closeErr
+		}
+	}
+
+	return err
+}
+
+func (w *writerFactory) abortAll() {
+	defer w.stopCount()
```
laskoviymishka, line 504: Both `closeAll()` and `abortAll()` defer `w.stopCount()`, and the new tests also `defer factory.closeAll()` on top of the normal path — so `stopCount()` can run more than once per factory. It's harmless today because `iter.Pull`'s stop is idempotent, but the ownership is implicit and would turn into a real bug the moment `stopCount` wraps anything non-idempotent. Might be worth pulling the `stopCount()` responsibility to a single owner.

table/rolling_data_writer.go
```diff
@@ -450,26 +451,64 @@ func (r *RollingDataWriter) sendError(err error) {
...
 
+	r.cancel()
+
 	return nil
 }
 
+func (r *RollingDataWriter) abortAndWait() {
+	r.abort()
+	r.factory.writers.Delete(r.partitionKey)
+	r.wg.Wait()
```
laskoviymishka, line 487: `abortAndWait()` never drains `r.errorCh` after `Wait()`, so if `stream()` sent an error it's silently dropped. No leak since `errorCh` is buffered at 1, but it makes a failed write harder to debug. Even just draining and logging it would help.

table/partitioned_fanout_writer.go
```diff
@@ -174,18 +175,37 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, record arro
...
+		fanoutWorkers,
+		outputDataFilesCh,
+		p.writerFactory.closeAll,
+		p.writerFactory.abortAll,
+		writerCancel,
+	)
 }
 
-func yieldDataFiles(writerFactory *writerFactory, fanoutWorkers *errgroup.Group, outputDataFilesCh chan iceberg.DataFile) iter.Seq2[iceberg.DataFile, error] {
+func yieldDataFiles(
```
laskoviymishka, line 189: `writerFactory` is dead now that `closeAll`/`abortAll` come in as explicit callbacks — the body doesn't touch it anymore. Both call sites already pass the two closures, so I'd just drop the param.

table/rolling_data_writer.go
```diff
@@ -450,26 +451,64 @@ func (r *RollingDataWriter) sendError(err error) {
...
 	r.factory.writers.Delete(r.partitionKey)
 	r.wg.Wait()
 
 	if err := <-r.errorCh; err != nil {
+		r.cancel()
+
 		return fmt.Errorf("error in rolling data writer: %w", err)
 	}
 
+	r.cancel()
```
laskoviymishka, line 479: Both branches call `r.cancel()` right before returning. `CancelFunc` is idempotent so it's safe, but a single `defer r.cancel()` at the top of `closeAndWait()` reads cleaner and won't get dropped by a future early return.

table/rolling_data_writer.go
```diff
@@ -479,12 +518,5 @@ func (w *writerFactory) closeAll() error {
...
 
-	var err error
-	for _, writer := range writers {
-		if closeErr := writer.closeAndWait(); closeErr != nil && err == nil {
-			err = closeErr
-		}
-	}
-
-	return err
+	return
```
laskoviymishka, line 521: nlreturn is enabled in `.golangci.yml`, and a naked `return` on a named result tends to trip it — worth confirming this doesn't fail CI. Either way I'd return the local explicitly (`return writers`) or drop the named return; it reads better than the bare form here.

table/partitioned_fanout_writer.go
```diff
@@ -174,18 +175,37 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, record arro
...
+	outputDataFilesCh chan iceberg.DataFile,
+	closeAll func() error,
+	abortAll func(),
+	writerCancel context.CancelFunc,
+) iter.Seq2[iceberg.DataFile, error] {
 	// Use a channel to safely communicate the error from the goroutine
 	// to avoid a data race between writing err in the goroutine and reading it in the iterator.
 	errCh := make(chan error, 1)
 	go func() {
 		defer close(outputDataFilesCh)
```
laskoviymishka, line 201: The LIFO ordering here is load-bearing — `writerCancel()` has to run before `close(outputDataFilesCh)`, and reversing these two lines silently reverses the semantics with nothing to catch it. A one-line comment on why the order matters would save the next reader.

table/rolling_data_writer_test.go
```diff
@@ -234,6 +234,67 @@ func (s *RollingDataWriterTestSuite) TestNewWriterFactoryReturnsErrorForInvalidF
...
+
+	writer, err := factory.getOrCreateRollingDataWriter(writerCtx, "", nil, outputCh)
+	s.Require().NoError(err)
+	record := s.buildRecord(arrSchema, 5)
+	s.Require().NoError(writer.Add(record))
+	record.Release()
+
+	writer.abortAndWait()
+	s.Require().ErrorIs(writer.ctx.Err(), context.Canceled)
+}
```
laskoviymishka, line 296: This only checks the context flipped to cancelled — it reaches into the unexported `ctx` field and never verifies the incomplete Parquet file was actually cleaned up. Walking the temp dir afterward (like `TestStreamErrorPathUsesAbort` does) would prove abort went through `Abort()` rather than `Close()`, which is the behavior that matters. Asserting on a public-ish signal instead of `writer.ctx` would also loosen the white-box coupling.

laskoviymishka (changes requested):
Almost there — the fix itself is right. Splitting `fanoutCtx` from `writerCtx` and deferring `writerCancel()` until after `Wait()` is the correct shape, and dispatching to `closeAll()` vs `abortAll()` on the `Wait()` result reads cleanly. `closeAndWait()` deferring the cancel past the drain and the matching `arrow_utils.go` unpartitioned update are both good.

One blocking thing before merge: the regression test doesn't exercise the race. `TestCloseAllFlushesAfterFanoutSuccessContextCancel` cancels no context, and with a 256MB target on local I/O the stream drains before anything would be cancelled — so it passes on the pre-fix code too and doesn't defend the change. I'd move this down to the `RollingDataWriter` level with a blocking `FileWriter` so we can cancel mid-write deterministically and assert `closeAll()` flushes while `abortAll()` discards.

Everything else is minor and can ride along in this PR:

- the two adjacent `context.Context` params on `fanout`/`processRecord`/`processBatch` — I'd thread `writerCtx` through the struct instead
- dead `writerFactory` param on `yieldDataFiles`
- `closeAndWait()` double `r.cancel()` → single `defer`
- `abortAll()`/`closeAll()` dual ownership of `stopCount()`
- `abortAndWait()` dropping the `errorCh` error
- naked `return` in `writerList()` (likely trips nlreturn)

Separately, I noticed a few pre-existing gaps this PR surfaces but doesn't need to fix: `abortAll()` doesn't delete already-rolled data files (Java's `abort()` iterates completed files and deletes them), `stream()` doesn't check `ctx.Done()` at the top of its loop so an unpartitioned/no-sort table can still emit files after abort, and `clustered_writer.go` closes its channel directly on the error path instead of aborting. Those feel like good follow-up issues rather than blockers here.

Fix the test so it actually pins the race and this is good to land — happy to take another pass once it's in.

PR #1369, opened 2026-07-04
fix(rest): accept oauth2-server-uri for the OAuth token endpoint

### LINKED ISSUE #1346: rest: align OAuth endpoint config with oauth2-server-uri
Follow-up from #1336 and #1345.

While looking at REST OAuth URL handling, I noticed iceberg-go uses `rest.authorization-url` for the OAuth token endpoint:

- `catalog/rest/rest.go` has `keyAuthUrl = "rest.authorization-url"`
- `toProps` also serializes the retained value back under that key

The broader Iceberg clients/docs appear to use `oauth2-server-uri` for the same concept:

- Java: `OAuth2Properties.OAUTH2_SERVER_URI = "oauth2-server-uri"`
- Java docs list `oauth2-server-uri` in REST OAuth2 catalog properties
- PyIceberg reads `OAUTH2_SERVER_URI = "oauth2-server-uri"`
- iceberg-rust reads `oauth2-server-uri` in the REST catalog config

There is also an iceberg-go comment in `WithOAuthTLSConfig` that already refers to `oauth2-server-uri`, so the public surface is a little mixed.

Why this matters: a catalog config that works in Java/PyIceberg/Rust, for example:

```properties
credential=client:secret
oauth2-server-uri=https://auth.example.com/oauth/tokens
```

does not configure the OAuth endpoint in iceberg-go today. The property is retained as an additional property, but `setupOAuthManager` never uses it, so the client falls back to `{catalog}/v1/oauth/tokens` unless `rest.authorization-url` is also set.

Possible direction:

- accept `oauth2-server-uri` as the portable/preferred key
- keep `rest.authorization-url` as a compatibility alias if needed
- define precedence if both are set
- update comments/tests around `WithAuthURI` and retained props

I wanted to raise this separately from the URL validation fixes because this feels more like an API compatibility decision than a validation bug.

tanmayrauth:
Read the portable oauth2-server-uri property used by Java, PyIceberg and iceberg-rust as the OAuth2 token endpoint, so a catalog config that works in those clients also configures iceberg-go. The existing rest.authorization-url key is kept as a compatibility alias, and oauth2-server-uri takes precedence when both are set. toProps now serializes the endpoint under the portable key.

Closes #1346

### DIFF
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -593,11 +593,12 @@ func toProps(o *options) iceberg.Properties {
 
 	setIf(keyPrefix, o.prefix)
 	if o.authUri != nil {
-		// Advertise the endpoint under the portable oauth2-server-uri key while
-		// also retaining the legacy rest.authorization-url alias, so existing
-		// consumers that read the properties back keep working.
+		// Advertise the endpoint only under the portable oauth2-server-uri key.
+		// We canonicalize to this key everywhere else (fromProps precedence,
+		// resolveAuthURLAlias), so emitting the legacy rest.authorization-url
+		// alias too would leave the endpoint under two names in r.props and any
+		// table config cloned from it - two sources of truth that can drift.
 		setIf(keyOAuth2ServerURI, o.authUri.String())
-		setIf(keyAuthUrl, o.authUri.String())
 	}
 
 	return props
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -476,15 +476,15 @@ func fromProps(props iceberg.Properties, o *options) error {
 		case keyRestSigV4Service:
 			o.sigv4Service = v
 		case keyAuthUrl:
-			u, err := parseAuthURL(v)
+			u, err := parseAuthURL(keyAuthUrl, v)
 			if err != nil {
 				return err
 			}
 			authURL = u
 		case keyOAuth2ServerURI:
-			u, err := url.Parse(v)
+			u, err := parseAuthURL(keyOAuth2ServerURI, v)
 			if err != nil {
-				return fmt.Errorf("invalid %s %q: %w", keyOAuth2ServerURI, v, err)
+				return err
 			}
 			oauth2ServerURI = u
 		case keyOauthCredential:
@@ -529,16 +529,16 @@ func fromProps(props iceberg.Properties, o *options) error {
 	return nil
 }
 
-func parseAuthURL(raw string) (*url.URL, error) {
+func parseAuthURL(key, raw string) (*url.URL, error) {
 	u, err := url.Parse(raw)
 	if err != nil {
-		return nil, fmt.Errorf("invalid %s %q: %w", keyAuthUrl, raw, err)
+		return nil, fmt.Errorf("invalid %s %q: %w", key, raw, err)
 	}
 	if u.Scheme == "" {
-		return nil, fmt.Errorf("invalid %s %q: missing scheme", keyAuthUrl, raw)
+		return nil, fmt.Errorf("invalid %s %q: missing scheme", key, raw)
 	}
 	if u.Host == "" {
-		return nil, fmt.Errorf("invalid %s %q: missing host", keyAuthUrl, raw)
+		return nil, fmt.Errorf("invalid %s %q: missing host", key, raw)
 	}
 
 	return u, nil
@@ -577,8 +577,11 @@ func toProps(o *options) iceberg.Properties {
 
 	setIf(keyPrefix, o.prefix)
 	if o.authUri != nil {
-		// Serialize under the portable oauth2-server-uri key.
+		// Advertise the endpoint under the portable oauth2-server-uri key while
+		// also retaining the legacy rest.authorization-url alias, so existing
+		// consumers that read the properties back keep working.
 		setIf(keyOAuth2ServerURI, o.authUri.String())
+		setIf(keyAuthUrl, o.authUri.String())
 	}
 
 	return props
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -529,6 +529,22 @@ func fromProps(props iceberg.Properties, o *options) error {
 	return nil
 }
 
+// resolveAuthURLAlias collapses the oauth2-server-uri / rest.authorization-url
+// alias within a single configuration layer into the canonical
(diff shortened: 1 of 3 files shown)


catalog/rest/rest.go
```diff
@@ -466,11 +472,17 @@ func fromProps(props iceberg.Properties, o *options) error {
...
 			o.sigv4Service = v
 		case keyAuthUrl:
-			u, err := parseAuthURL(v)
+			u, err := url.Parse(v)
+			if err != nil {
+				return fmt.Errorf("invalid %s %q: %w", keyAuthUrl, v, err)
+			}
+			authURL = u
+		case keyOAuth2ServerURI:
+			u, err := url.Parse(v)
```
fallintoplace: Do we have better options than url.Parse(v)?

catalog/rest/rest.go
```diff
@@ -501,19 +517,28 @@ func fromProps(props iceberg.Properties, o *options) error {
 		}
 	}
 
+	// oauth2-server-uri is the portable, preferred key and takes precedence over
+	// the rest.authorization-url alias when both are set.
```
zeroshade, line 521: This global alias precedence loses REST config layer precedence after fetchConfig merges defaults, client props, and server overrides. If the client supplied WithAuthURI(A), toProps now adds oauth2-server-uri=A, and a server /config override of rest.authorization-url=B leaves both keys in the merged map. This switch then chooses A, so the server override is ignored. Please resolve the oauth2-server-uri/rest.authorization-url alias within each source layer before merging, or otherwise preserve defaults → client → overrides precedence, and add a regression test for WithAuthURI(A) plus a server rest.authorization-url=B override.

zeroshade (changes requested):
Thanks for adding oauth2-server-uri support. I found one precedence issue that should be fixed before merge:

- The merged configuration can now contain a client-provided oauth2-server-uri and a server override rest.authorization-url. Because fromProps globally prefers oauth2-server-uri over rest.authorization-url after the maps have already been merged, the client value wins and the server override is ignored. REST config overrides should apply after defaults and client properties. Please preserve layer precedence and only resolve the oauth2-server-uri/rest.authorization-url alias within a single source layer, with a regression test for WithAuthURI(A) plus /config override rest.authorization-url=B.

catalog/rest/rest.go
tanmayrauth: Good catch, thanks — you're right that the global alias resolution broke layer precedence.
  
Instead of resolving oauth2-server-uri/rest.authorization-url globally after the merge, I now collapse the alias within each source layer before merging in fetchConfig. A new resolveAuthURLAlias helper
  canonicalizes each layer to the single oauth2-server-uri key (oauth2-server-uri still wins over the alias within a layer), so the defaults → client → overrides map merge preserves precedence naturally. A server
  rest.authorization-url override now correctly wins over a client-provided oauth2-server-uri.

  Added TestFetchConfigAuthURLOverridePrecedence covering exactly the WithAuthURI(A) + /config override rest.authorization-url=B case — asserts B wins. fromProps's existing precedence switch is unchanged since it's still correct for the single-layer newCatalogFromProps path.

tanmayrauth:

catalog/rest/rest.go
```diff
@@ -552,6 +593,10 @@ func toProps(o *options) iceberg.Properties {
 
 	setIf(keyPrefix, o.prefix)
 	if o.authUri != nil {
+		// Advertise the endpoint under the portable oauth2-server-uri key while
+		// also retaining the legacy rest.authorization-url alias, so existing
+		// consumers that read the properties back keep working.
+		setIf(keyOAuth2ServerURI, o.authUri.String())
```
laskoviymishka, line 601: This is the one design call I'd want to settle before merge. Emitting both keys means `r.props` always carries the endpoint under two names, and that duplicate then gets cloned into table config downstream — two sources of truth that can drift.

I get the intent is backward-compat for Go consumers reading props back, but since we canonicalize to `oauth2-server-uri` everywhere else, I'd lean toward emitting only the portable key here and dropping the legacy one. If we do want to keep both, can we at least note in the comment that it's a Go-consumer concession, not something other clients read? wdyt?

laskoviymishka (changes requested):
The alias handling reads well, and the precedence issue zeroshade flagged looks properly addressed with the per-layer collapse.

One design call I'd want to settle before merge: toProps emitting both keys leaves r.props carrying the endpoint under two names, which then gets cloned into table config downstream - two sources of truth that can drift. Left an inline note on it. Since we canonicalize to oauth2-server-uri everywhere else, I'd lean toward emitting only the portable key.

catalog/rest/rest.go
tanmayrauth: Good call, settled on dropping the legacy key. toProps now emits only oauth2-server-uri, so r.props (and the table config cloned from it) carries the endpoint under a single canonical name, matching resolveAuthURLAlias/fromProps. rest.authorization-url is still accepted as an input alias for backward compat; we just don't advertise it back anymore. I will update the PR.

tanmayrauth: @laskoviymishka  I've updated the PR XD

laskoviymishka (approved):
Look good to me, but need to fix build :)

tanmayrauth:
> Look good to me, but need to fix build :)

Oh yes, let me fix it.

zeroshade (changes requested):
The alias/precedence logic looks correct, but CI is currently red on all four test-matrix jobs (plus Audit and Verify) due to one stale test assertion.

In `catalog/rest/rest_internal_test.go`, the `TestNewCatalogRejectsInvalidAuthURLFromConfig` "malformed URL in defaults" case still asserts `wantErr: invalid rest.authorization-url` (around line 160). After `resolveAuthURLAlias` renames the key before `fromProps` parses it, that error now surfaces as `invalid oauth2-server-uri`, so the assertion fails. Please update the expected substring to `invalid oauth2-server-uri` — the behavior itself is correct. (That line falls outside this PR's diff hunks, so I couldn't anchor it inline.)

tanmayrauth:
@zeroshade thanks,  I've already pushed the fix.

zeroshade (approved):
Re-reviewed after the latest commit — the stale assertion is fixed (`wantErr: "invalid oauth2-server-uri"` at `rest_internal_test.go:192`) and CI is now green across all matrix jobs. LGTM — thanks!

PR #1370, opened 2026-07-04
fix(table): dedupe repeated ref assertions by ref name

### LINKED ISSUE #1344: table: clarify transaction requirement dedupe for repeated ref assertions
Follow-up from #1328.

I think the same-ref case @laskoviymishka called out is reachable through normal transaction use, and is worth deciding explicitly rather than leaving it as a side effect of the JSON payload key.

The shape I traced is:

1. Start a transaction on a table with no `main` ref yet.
2. First append builds `AssertRefSnapshotID("main", nil)` and then applies updates that create `main -> snapshot-1` in the transaction builder.
3. A second append in the same transaction builds `AssertRefSnapshotID("main", &snapshot-1)` and applies another snapshot.
4. `Commit` sends both requirements to the catalog.

Those two requirements are contradictory against the pre-transaction base metadata: `main` must not exist, and `main` must already point at `snapshot-1`. A validating catalog should reject that commit, even though the transaction itself is a reasonable sequence of operations.

The fix I think we should consider is to special-case `assert-ref-snapshot-id` dedupe to key by requirement type + ref name, while keeping the current semantic payload key for other requirement types. That would preserve the important #1328 behavior where different refs survive dedupe, but avoid carrying multiple base-state assertions for the same ref as the transaction mutates its own builder state.

A test for the intended behavior could cover:

- two appends in one new-table transaction keep only the first `main == nil` assertion
- two appends on an existing table keep only the original base-head assertion
- assertions for different refs still both survive
- identical non-ref requirements still dedupe as before

tanmayrauth:
Key assert-ref-snapshot-id dedupe by requirement type and ref name, ignoring the asserted snapshot id. As a transaction mutates its own builder ref state across operations, the same ref would otherwise carry multiple, mutually contradictory base-state assertions that a validating catalog rejects. Other requirement types keep the canonical JSON payload key, so assertions for different refs still survive dedupe.

Closes #1344

### DIFF
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -153,9 +153,25 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 	return nil
 }
 
-// requirementSemanticKey assumes Requirement JSON marshaling is canonical and
-// deterministic for every requirement type that participates in dedupe.
+// requirementSemanticKey computes the dedupe key for a requirement.
+//
+// For most requirement types the key is the canonical JSON marshaling, which
+// keeps semantically distinct requirements (e.g. assertions on different refs)
+// from collapsing into one another.
+//
+// assert-ref-snapshot-id is special-cased to key by requirement type + ref name
+// only, deliberately ignoring the asserted snapshot id. Within a single
+// transaction the builder mutates its own ref state across operations (e.g. the
+// first append asserts main == nil, a later append asserts main == snapshot-1),
+// which would otherwise produce multiple, mutually contradictory base-state
+// assertions for the same ref against the pre-transaction metadata. Keying by
+// ref name keeps only the first assertion for each ref while still letting
+// assertions for different refs survive dedupe.
 func requirementSemanticKey(r Requirement) (string, error) {
+	if ref, ok := r.(*assertRefSnapshotID); ok {
+		return fmt.Sprintf("%s\x00%s", reqAssertRefSnapshotID, ref.Ref), nil
+	}
+
 	data, err := json.Marshal(r)
 	if err != nil {
 		return "", fmt.Errorf("marshal requirement %q: %w", r.GetType(), err)
--- a/table/transaction_internal_test.go
+++ b/table/transaction_internal_test.go
@@ -60,6 +60,116 @@ func TestTransactionApplyDedupesEquivalentRequirementsWithinAndAcrossCalls(t *te
 	requireContainsRefSnapshotRequirement(t, txn.reqs, MainBranch, &mainSnapshotID)
 }
 
+// TestTransactionApplyDedupesSameRefAssertionsNewTable covers two appends in a
+// single new-table transaction: the first asserts main must not exist, and the
+// second (after the builder has created main) asserts main == the new snapshot.
+// Only the first main == nil assertion, which reflects the pre-transaction base
+// state, must be retained.
+func TestTransactionApplyDedupesSameRefAssertionsNewTable(t *testing.T) {
+	txn, _ := createTestTransactionWithMemIO(t, *iceberg.UnpartitionedSpec)
+
+	// First append: main does not exist yet in the pre-transaction metadata.
+	err := txn.apply(nil, []Requirement{AssertRefSnapshotID(MainBranch, nil)})
+	require.NoError(t, err)
+	require.Len(t, txn.reqs, 1)
+	requireContainsRefSnapshotRequirement(t, txn.reqs, MainBranch, nil)
+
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
Re-reviewed — deduping ref-snapshot requirements by ref name with first-assertion-wins is the correct catalog semantics, and the tests cover same-ref new/existing and distinct-ref cases. LGTM.

tanmayrauth:
@laskoviymishka can you please review it?

table/transaction.go
```diff
@@ -153,9 +153,25 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
...
+// assert-ref-snapshot-id is special-cased to key by requirement type + ref name
+// only, deliberately ignoring the asserted snapshot id. Within a single
+// transaction the builder mutates its own ref state across operations (e.g. the
+// first append asserts main == nil, a later append asserts main == snapshot-1),
+// which would otherwise produce multiple, mutually contradictory base-state
+// assertions for the same ref against the pre-transaction metadata. Keying by
+// ref name keeps only the first assertion for each ref while still letting
+// assertions for different refs survive dedupe.
 func requirementSemanticKey(r Requirement) (string, error) {
+	if ref, ok := r.(*assertRefSnapshotID); ok {
```
laskoviymishka, line 171: The whole first-wins-is-base-state property only holds because `commit()` captures the builder's ref state before `apply()` applies that op's updates — so the first assertion we see for a ref is the pre-transaction one. That's true today, but it's an implicit ordering contract, not something the types enforce.

I'd add a line here (and near the `commit()` call site in `snapshot_producers.go`) spelling out that dedupe keeps the first assertion *because* it's the base-state one, and that this breaks silently — wrong precondition to the catalog, no error — if a future refactor ever advances the builder before `commit()` runs. Cheap insurance. wdyt?

table/transaction.go
```diff
@@ -153,9 +153,25 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
...
+		return fmt.Sprintf("%s\x00%s", reqAssertRefSnapshotID, ref.Ref), nil
```
laskoviymishka, line 172: Tiny one: this can just be `reqAssertRefSnapshotID + "\x00" + ref.Ref` — perfsprint won't love a `Sprintf` that's really string concatenation, and it's a hair faster.

While we're here, a half-line noting `\x00` is a safe separator because Iceberg ref names can't contain NUL would save the next person a double-take.

table/transaction_internal_test.go
```diff
@@ -60,6 +60,116 @@ func TestTransactionApplyDedupesEquivalentRequirementsWithinAndAcrossCalls(t *te
...
+		ManifestList:   "mem://default/table-location/metadata/manifest-10.avro",
+		Summary:        &Summary{Operation: OpAppend},
+		TimestampMs:    time.Now().UnixMilli(),
+	}))
+	require.NoError(t, txn.meta.SetSnapshotRef(MainBranch, 10, BranchRef))
+
+	// Second append asserts main == 10; it must dedupe against the first
+	// assertion for main rather than adding a contradictory base-state check.
+	newHead := int64(10)
+	err = txn.apply(nil, []Requirement{AssertRefSnapshotID(MainBranch, &newHead)})
```
laskoviymishka, line 90: Optional: every one of these uses two separate `apply()` calls, so the cross-call dedupe path is well covered but the within-a-single-`apply()` case — same ref twice in one reqs slice, different snapshot ids — isn't. The loop handles it today, but a future "build a set then append" refactor would pass all these tests while quietly breaking the within-call case. A quick variant of this test with both assertions in one `apply()` would close that.

laskoviymishka (approved):
Nice, this is the right fix — keeping the first `assert-ref-snapshot-id` per ref is exactly the base-state semantics the catalog wants, and the distinct-ref test proves we don't over-collapse. Agree with zeroshade, this is good to merge.

The one thing I'd genuinely like before it goes in is a comment nailing down *why* first-wins is correct: it works because `commit()` captures the ref's pre-transaction state before `apply()` advances the builder. That's an implicit ordering contract right now, and if it ever inverts this breaks with no error. A couple of lines here and at the `commit()` site is cheap.

One thing for a follow-up, not this PR: the same intra-transaction mutation problem exists for the single-valued requirements (`assert-current-schema-id`, `assert-last-assigned-field-id`, and friends). They have no discriminant like a ref name, so JSON-key dedupe happens to keep both if the asserted values differ, and a spec-compliant catalog would 409 on the second. Worth an issue to extend this same keying to those types by requirement type alone — happy to file it if useful.

Rest are small nits inline. Nice work.

PR #1371, opened 2026-07-04
feat(table): emit geo bounds in manifest DataFile

### LINKED ISSUE #992: feat(table): emit Parquet GeoStatistics on writes, parse on reads
Parent: #989

Depends on #991. Parquet's GeoStatistics column-chunk metadata carries a 4D bounding box (min/max for X, Y, Z, M) and the set of geometry types present. Spec: [Parquet Geospatial types](https://github.com/apache/parquet-format/blob/master/Geospatial.md).

Plumb the writer in `table/internal/parquet_files.go` to emit GeoStatistics when a geo column is flagged; the parquet-go library exposes the API once the column type carries the geo logical type. The reader extracts GeoStatistics into the existing internal column-stats struct so downstream pruning sees it as if it were any other typed bound.

Tests assert GeoStatistics present on round-tripped parquet plus a golden fixture from a Java-produced geo parquet for cross-client coverage.

tanmayrauth:
Compute geometry/geography bounding boxes from WKB during writes and emit WKB single-point lower/upper bounds into the manifest DataFile, which is what Iceberg pruning consumes. arrow-go cannot emit Parquet column-chunk GeospatialStatistics, so bounds are computed in iceberg-go.

Also fixes a panic when writing geo columns through the stats path, and lets geo bounds round-trip via LiteralFromBytes.

Closes: #992

### DIFF
--- a/literals.go
+++ b/literals.go
@@ -170,7 +170,9 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
 		err := v.UnmarshalBinary(data)
 
 		return v, err
-	case BinaryType:
+	case BinaryType, GeometryType, GeographyType:
+		// Geometry and Geography bounds are stored as WKB byte payloads
+		// (single-point bounds), handled as opaque binary on read.
 		var v BinaryLiteral
 		err := v.UnmarshalBinary(data)
 
--- a/literals.go
+++ b/literals.go
@@ -170,9 +170,23 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
 		err := v.UnmarshalBinary(data)
 
 		return v, err
-	case BinaryType, GeometryType, GeographyType:
-		// Geometry and Geography bounds are stored as WKB byte payloads
-		// (single-point bounds), handled as opaque binary on read.
+	case BinaryType:
+		var v BinaryLiteral
+		err := v.UnmarshalBinary(data)
+
+		return v, err
+	case GeometryType, GeographyType:
+		// Geometry/Geography single-value bounds use the Iceberg geospatial
+		// serialization (spec Appendix D): little-endian float64 coordinates in
+		// X, Y[, Z][, M] order, i.e. 16, 24, or 32 bytes — not WKB. iceberg-go
+		// has no geo literal for predicate evaluation, so the validated
+		// coordinate bytes are returned as an opaque BinaryLiteral.
+		switch len(data) {
+		case 16, 24, 32:
+		default:
+			return nil, fmt.Errorf("%w: geometry/geography bound must be 16, 24, or 32 bytes, got %d",
+				ErrInvalidBinSerialization, len(data))
+		}
 		var v BinaryLiteral
 		err := v.UnmarshalBinary(data)
 
--- a/literals.go
+++ b/literals.go
@@ -178,21 +178,19 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
 	case GeometryType, GeographyType:
 		// Geometry/Geography single-value bounds use the Iceberg geospatial
 		// serialization (spec Appendix D): little-endian float64 coordinates in
-		// X, Y[, Z][, M] order, i.e. 16, 24, or 32 bytes — not WKB. iceberg-go
-		// has no geo literal for predicate evaluation, so the validated
-		// coordinate bytes are returned as an opaque BinaryLiteral. Callers doing
-		// geo pruning must not treat this as a plain binary bound: check the
-		// original type first, as the bytes are coordinates, not comparable binary.
+		// X, Y[, Z][, M] order, i.e. 16, 24, or 32 bytes — not WKB. The
+		// coordinates have no total order, so they are returned as a GeoLiteral
+		// (not a plain BinaryLiteral): it reports the real geo type and is not
+		// comparable, so any attempt to order it errors instead of silently
+		// running bytes.Compare over the coordinate bytes.
 		switch len(data) {
-		case 16, 24, 32: // valid coordinate lengths (XY / XYZ|XYM / XYZM), fall through
+		case 16, 24, 32: // valid coordinate lengths (XY / XYZ|XYM / XYZM)
 		default:
 			return nil, fmt.Errorf("%w: geometry/geography bound must be 16, 24, or 32 bytes, got %d",
 				ErrInvalidBinSerialization, len(data))
 		}
-		var v BinaryLiteral
-		err := v.UnmarshalBinary(data)
 
-		return v, err
+		return GeoLiteral{val: data, typ: typ}, nil
 	case FixedType:
 		if len(data) != t.Len() {
 			// looks like some writers will write a prefix of the fixed length
@@ -1131,6 +1129,51 @@ func (b *BinaryLiteral) UnmarshalBinary(data []byte) error {
 	*b = BinaryLiteral(data)
 
 	return nil
+}
+
+// GeoLiteral is a non-null geometry or geography single-value bound. Iceberg
+// serializes geospatial bounds as the concatenation of little-endian float64
+// coordinates in X, Y[, Z][, M] order (spec Appendix D), not as WKB. Those
+// coordinates have no total order, so - unlike every other literal - GeoLiteral
+// is deliberately not a TypedLiteral: it exposes no Comparator and refuses to be
+// cast to an orderable type. That keeps ordering-based pruning from silently
+// running bytes.Compare over coordinate bytes (which would yield wrong answers)
+// and makes it error instead. Type() reports the concrete GeometryType or
+// GeographyType, so callers that key off the literal's own type - pruning, To,
+// logging - see the truth rather than plain binary.
+type GeoLiteral struct {
+	val []byte
+	typ Type
+}
+
+func (g GeoLiteral) Type() Type    { return g.typ }
+func (g GeoLiteral) Value() []byte { return g.val }
+func (g GeoLiteral) Any() any      { return g.val }
+func (g GeoLiteral) String() string {
+	return fmt.Sprintf("%s(%x)", g.typ, g.val)
+}
+
+func (g GeoLiteral) To(typ Type) (Literal, error) {
+	// Only an identity cast is meaningful. Casting to any orderable type is
+	// rejected so a geo bound can never be smuggled into a comparison.
+	if g.typ.Equals(typ) {
+		return g, nil
+	}
+
+	return nil, fmt.Errorf("%w: GeoLiteral to %s", ErrBadCast, typ)
+}
+
+func (g GeoLiteral) Equals(other Literal) bool {
+	rhs, ok := other.(GeoLiteral)
+	if !ok {
+		return false
+	}
+
+	return g.typ.Equals(rhs.typ) && bytes.Equal(g.val, rhs.val)
+}
+
+func (g GeoLiteral) MarshalBinary() ([]byte, error) {
+	return g.val, nil
 }
 
 type FixedLiteral []byte
--- a/literals.go
+++ b/literals.go
@@ -180,9 +180,11 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
 		// serialization (spec Appendix D): little-endian float64 coordinates in
 		// X, Y[, Z][, M] order, i.e. 16, 24, or 32 bytes — not WKB. iceberg-go
 		// has no geo literal for predicate evaluation, so the validated
-		// coordinate bytes are returned as an opaque BinaryLiteral.
+		// coordinate bytes are returned as an opaque BinaryLiteral. Callers doing
+		// geo pruning must not treat this as a plain binary bound: check the
+		// original type first, as the bytes are coordinates, not comparable binary.
 		switch len(data) {
-		case 16, 24, 32:
+		case 16, 24, 32: // valid coordinate lengths (XY / XYZ|XYM / XYZM), fall through
 		default:
 			return nil, fmt.Errorf("%w: geometry/geography bound must be 16, 24, or 32 bytes, got %d",
 				ErrInvalidBinSerialization, len(data))
--- a/table/internal/geo_codec.go
+++ b/table/internal/geo_codec.go
@@ -53,6 +53,17 @@ type geoBoundsAccumulator struct {
 	max [geoNumDims]float64
 	has [geoNumDims]bool
 
+	// geoms counts the leaf geometries that contributed coordinates; zGeoms and
+	// mGeoms count how many of those carried a Z / M dimension. The has[] flags
+	// are sticky across rows, so a column mixing e.g. XYZ and XYM geometries
+	// would set both has[geoDimZ] and has[geoDimM] without any single row having
+	// both. An optional dimension is only safe to emit when every geometry
+	// carried it (count == geoms); otherwise the bound would imply rows that lack
+	// the dimension have a value in range. See layout.
+	geoms  int
+	zGeoms int
+	mGeoms int
+
 	// isGeography marks a column whose edges are geodesics on a sphere, for
 	// which raw vertex min/max is not a safe bounding box (see Bounds).
 	isGeography bool
@@ -94,6 +105,13 @@ func (a *geoBoundsAccumulator) extend(g geom.T) {
 
 	layout := g.Layout()
 	zIdx, mIdx := layout.ZIndex(), layout.MIndex()
+	a.geoms++
+	if zIdx >= 0 {
+		a.zGeoms++
+	}
+	if mIdx >= 0 {
+		a.mGeoms++
+	}
 	for base := 0; base+stride <= len(flat); base += stride {
 		a.update(geoDimX, flat[base])
 		a.update(geoDimY, flat[base+1])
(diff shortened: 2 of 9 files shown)


table/internal/geo_codec.go
```diff
@@ -0,0 +1,215 @@
...
+
+// Bounds encodes the lower and upper bound points as WKB. Both are nil when no
+// bounding box could be produced (e.g. an all-null column).
+func (a *geoBoundsAccumulator) Bounds() (lower, upper []byte, err error) {
+	layout, ok := a.layout()
+	if !ok {
+		return nil, nil, nil
+	}
+
+	lower, err = wkb.Marshal(geom.NewPointFlat(layout, geoPointCoords(a.min, layout)), wkb.NDR)
```
zeroshade: This serializes the manifest lower bound as a WKB point, and line 166 does the same for the upper bound. Iceberg v3 geo single-value bounds are not WKB; they are concatenated little-endian float64 coordinates (XY=16 bytes, XYZ=24, XYM/XYZM=32 with XYM Z=NaN). Please encode/decode these coordinate bytes instead, update `LiteralFromBytes` so geo bounds are parsed as the spec format rather than opaque WKB, and assert exact byte lengths/decoded coordinates in the tests.

table/internal/parquet_files.go
```diff
@@ -384,14 +419,57 @@ func (p parquetFormat) NewFileWriter(ctx context.Context, fs iceio.WriteFileIO,
...
+		if !ok {
+			continue
+		}
+
+		acc := w.geoAccs[gc.fieldID]
+		for i := range storage.Len() {
+			if storage.IsNull(i) {
+				continue
+			}
+			if err := acc.AddWKB(storage.Value(i)); err != nil {
```
zeroshade, line 474: The accumulator is fed raw WKB bytes without knowing whether the Iceberg field is geometry or geography, so geography bounds are computed by simple vertex min/max. That misses geodesic edge interpolation and antimeridian wrapping; for geography crossing the antimeridian the correct longitude interval can require `xmin > xmax`. Please make accumulation type-aware and compute safe geography bounds, or omit geography bounds when they cannot be computed safely, with antimeridian/geodesic coverage.

zeroshade (changes requested):
Thanks for adding geo bounds support. I think this needs changes before merge:

- Geometry/geography manifest lower_bounds/upper_bounds are being serialized as WKB point payloads, but Iceberg v3 geo single-value bounds use concatenated little-endian float64 coordinates (XY=16 bytes, XYZ=24, XYM/XYZM=32 with XYM Z=NaN). This will put WKB headers in manifests and break readers expecting coordinate bytes.
- Geography bounds are accumulated as raw vertex min/max without type awareness, edge interpolation, or antimeridian wraparound handling. For geography, the correct longitude interval may wrap with xmin > xmax, so these bounds can be incorrect unless computed geodesically/type-aware or omitted when unsafe.

tanmayrauth:
@zeroshade  Thanks for both good catches. Switched manifest bounds to the spec coordinate encoding and made geography omit bounds for now.  


On table/internal/geo_codec.go (the encodeGeoBound/Bounds area):
  Fixed — bounds are now the Iceberg geospatial single-value serialization (little-endian float64 X, Y[, Z][, M]: 16/24/32 bytes, XYM writes Z=NaN), not WKB. `LiteralFromBytes` for geometry/geography now validates the 16/24/32  length instead of treating the payload as opaque WKB, and the tests assert  exact byte lengths and the decoded coordinates (including the XYM NaN slot).

  On table/internal/parquet_files.go (the AddWKB accumulation loop):
  Made accumulation type-aware: the accumulator now knows geometry vs geography  (from the Iceberg field type), and geography bounds are omitted entirely. Since geography edges are geodesics, vertex min/max isn't a safe box — latitude can bulge past the endpoints and longitude may need to wrap the antimeridian — so rather than emit possibly-wrong bounds that silently prune valid rows, I leave geography unbounded until we add geodesic/antimeridian-aware computation. Geometry (planar edges) still gets exact bounds.

tanmayrauth:

table/internal/geo_codec.go
```diff
@@ -0,0 +1,235 @@
...
+// layout selects the point layout for the bound points from the dimensions that
+// saw finite values. The bool is false when the box has no X or Y dimension and
+// therefore should not be emitted.
+func (a *geoBoundsAccumulator) layout() (geom.Layout, bool) {
+	if !a.has[geoDimX] || !a.has[geoDimY] {
+		return geom.NoLayout, false
+	}
+
+	switch {
+	case a.has[geoDimZ] && a.has[geoDimM]:
```
laskoviymishka: I think there's a real bug here for a column that mixes dimensionalities.

`has[]` is set the first time *any* geometry contributes a given dimension, so a column with some `POINT Z` rows and some `POINT M` rows sets both `has[geoDimZ]` and `has[geoDimM]` and promotes to XYZM — but the Z range came only from the XYZ subset and the M range only from the XYM subset. The emitted bounds then claim every row has a valid Z *and* a valid M in those ranges, which no single row does. A pruner filtering on M could drop rows that only carry Z (and vice versa), so this is wrong-answer pruning once the read side exists.

I'd track "seen Z and M together in one geometry" vs "seen either at all," and when they never co-occur, omit the ambiguous dimension rather than promoting to XYZM. wdyt on that as the rule — omit-on-ambiguity?

table/internal/parquet_files.go
```diff
@@ -436,6 +524,44 @@ func (w *ParquetFileWriter) Abort() error {
 	return errors.Join(closeErr, removeErr)
 }
 
+// applyGeoBounds injects the WKB single-point bounds accumulated during the
+// write into the file statistics, so they flow through ToDataFile into the
+// manifest entry like any other typed bound.
+func (w *ParquetFileWriter) applyGeoBounds(stats *DataFileStatistics) error {
+	for fieldID, acc := range w.geoAccs {
+		// Honor the column's metrics mode: counts/none modes do not record bounds.
+		switch w.info.StatsCols[fieldID].Mode.Typ {
```
laskoviymishka: `w.info.StatsCols[fieldID]` on a missing key returns a zero-value `StatisticsCollector` whose `Mode.Typ` is `""`, which matches neither `MetricModeNone` nor `MetricModeCounts` — so we fall through and write bounds for a column the caller never registered. That silently overrides an intent to skip.

I'd gate on presence first, and match the positive-check idiom the rest of this file already uses so the exhaustive linter stays happy:

```go
sc, ok := w.info.StatsCols[fieldID]
if !ok || sc.Mode.Typ == MetricModeNone || sc.Mode.Typ == MetricModeCounts {
    continue
}
```

table/internal/geo_codec.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (g *geoStatsAgg) Max() iceberg.Literal {
+	if g.upper == nil {
+		return nil
+	}
+
+	return iceberg.BinaryLiteral(g.upper)
+}
+
+func (g *geoStatsAgg) Update(interface{ HasMinMax() bool }) {}
```
laskoviymishka, line 265: This no-op is safe only because of injection ordering — nothing currently iterates `ColAggs` and calls `Update` on a geo agg. But that's an implicit invariant, and the first call site that does iterate will silently drop updates on geo fields.

At minimum I'd document why it's a no-op. Better would be to not pretend to be a live `StatsAgg` at all — a separate precomputed-bounds carrier (exposing just `MinAsBytes`/`MaxAsBytes`) drained on its own in `ToDataFile` would make the "these are precomputed, don't Update them" contract explicit instead of load-bearing on ordering. Either's fine, but I'd make it explicit.

table/internal/geo_codec.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+		if mIdx >= 0 {
+			a.update(geoDimM, flat[base+mIdx])
+		}
+	}
+}
+
+func (a *geoBoundsAccumulator) update(dim int, v float64) {
+	if math.IsNaN(v) {
+		return
```
laskoviymishka, line 129: `nlreturn` wants a blank line before this early `return`, and there's a second one in `extend()` after the `GeometryCollection` recursion. golangci-lint will fail the build on both — worth a quick pass so CI goes green.

table/internal/parquet_files.go
```diff
@@ -384,14 +425,57 @@ func (p parquetFormat) NewFileWriter(ctx context.Context, fs iceio.WriteFileIO,
...
+		ext, ok := batch.Column(gc.colIdx).(array.ExtensionArray)
+		if !ok {
+			continue
+		}
+		storage, ok := ext.Storage().(wkbStorage)
+		if !ok {
+			continue
+		}
+
+		acc := w.geoAccs[gc.fieldID]
```
laskoviymishka: `acc` is safe by construction today, but a missing key here nil-panics on `acc.AddWKB`. Cheap to make it a two-return + `continue` so a future refactor that changes how `geoAccs` is populated can't turn this into a crash.

table/internal/parquet_files_test.go
```diff
@@ -1348,3 +1349,87 @@ func TestShreddedVariantReadRoundTrip(t *testing.T) {
 		assert.EqualValues(t, 2, obj.NumElements(), "row %d should have a + city", i)
 	}
 }
+
+// TestWriteDataFileGeoBounds writes geometry and geography columns through the
+// full ParquetFileWriter path and asserts that the resulting DataFile carries
+// geometry single-point bounds (Iceberg geospatial single-value serialization)
+// in the manifest entry, while geography bounds are omitted as unsafe.
+func TestWriteDataFileGeoBounds(t *testing.T) {
```
laskoviymishka, line 1362: Two gates go untested right now. `TestWriteDataFileGeoBounds` runs `MetricModeFull` end-to-end, so it happens to cover the old panic path — but its stated intent is bounds correctness, and a future maintainer won't read it as the regression guard for the `createStatsAgg` panic. A one-line comment anchoring it to that issue (or a dedicated tiny test) would keep the guard from silently rotting.

Separately, I'd add sub-cases for `MetricModeNone` and `MetricModeCounts` asserting the geo field ID does *not* appear in lower/upper — that's exactly the fallthrough I flagged in `applyGeoBounds`, and a test there would have caught it.

literals.go
```diff
@@ -174,6 +174,22 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
...
 
+		return v, err
+	case GeometryType, GeographyType:
+		// Geometry/Geography single-value bounds use the Iceberg geospatial
+		// serialization (spec Appendix D): little-endian float64 coordinates in
+		// X, Y[, Z][, M] order, i.e. 16, 24, or 32 bytes — not WKB. iceberg-go
+		// has no geo literal for predicate evaluation, so the validated
+		// coordinate bytes are returned as an opaque BinaryLiteral.
+		switch len(data) {
+		case 16, 24, 32:
```
laskoviymishka: Two small things. The empty `case 16, 24, 32:` reads like a forgotten body — a `// valid sizes, fall through` comment (or restructuring to a positive length check like `FixedType` above) makes the intent obvious.

And the geo case returns a plain `BinaryLiteral`, indistinguishable from a real binary bound. No bug today since nothing does geo pruning, but a future geo predicate evaluator calling `LiteralFromBytes` would get raw coordinate bytes with no signal they're geo. A short comment warning callers to check the original `typ` (or a thin wrapper type later) would save someone a debugging session.

table/internal/parquet_files.go
```diff
@@ -343,6 +344,34 @@ type ParquetFileWriter struct {
...
+	colIdx  int
+	fieldID int
+}
+
+// collectGeoColumns finds top-level WKB-encoded geo columns in the Arrow schema
+// and pairs each with its Iceberg field ID. Geo bounds for columns nested inside
+// structs/lists/maps are not yet computed.
+func collectGeoColumns(sc *arrow.Schema, colMapping map[string]int) []geoColumn {
+	var result []geoColumn
+	for i, f := range sc.Fields() {
```
laskoviymishka, line 364: `collectGeoColumns` only walks top-level fields, so a geo column nested in a struct/list/map gets no bounds — silently, and diverging from Java/PyIceberg. That's a fine scope cut for this PR, but I'd file a tracking issue and drop a `// TODO(#issue): nested geo columns` here so it's discoverable rather than invisible.

laskoviymishka (changes requested):
Thanks for turning this around — this is a solid revision. The switch to little-endian float64 coordinate bounds (16/24/32 bytes) and omitting geography bounds both address the encoding and antimeridian concerns from the last round, and I checked the XYM Z=NaN layout against Java's `GeospatialBound.toByteBuffer` — it's bit-for-bit correct. The `isGeoType` guard is a clean fix for the `createStatsAgg` panic too.

I'd hold this before merging though, on one correctness issue in particular.

`layout()` promotes to XYZM whenever both `has[geoDimZ]` and `has[geoDimM]` are set, but those flags are sticky across rows. A column mixing `POINT Z` and `POINT M` geometries ends up with XYZM bounds where the Z range comes only from the XYZ rows and the M range only from the XYM rows — the bounds then imply every row has a valid Z and M, which no row does. Once a reader wires up geo pruning, that's wrong-answer pruning. I think the right rule is to omit the ambiguous dimension when Z and M never co-occur in a single geometry rather than promoting, but I'd like to settle that here.

The second one: `applyGeoBounds` reads `w.info.StatsCols[fieldID].Mode.Typ`, and a missing key returns a zero-value mode of `""` that falls through the `None`/`Counts` switch, so we write bounds even for columns the caller asked to skip. A presence check fixes it.

- omit-vs-promote decision for mixed-dimensionality geometries in `layout()`
- the `StatsCols` missing-key fallthrough in `applyGeoBounds`
- the `nlreturn` blank lines in `update()`/`extend()` (CI will fail otherwise)
- test sub-cases for `MetricModeNone`/`MetricModeCounts` gating, and a comment anchoring the existing test to the panic regression
- documenting (or restructuring away) the `geoStatsAgg.Update` no-op

The rest are minor — nested geo columns, the opaque geo literal, a nil-guard. Once the mixed-dimensionality and `StatsCols` items are addressed, happy to take another pass and approve.

table/internal/geo_codec.go
tanmayrauth: Good catch — this is a real wrong-answer bug. I went slightly broader than the Z/M co-occurrence flag, because tracking only "seen together" still leaves an identical hole: an XYZM + XYZ mix would keep M and emit bounds claiming the XYZ row has an M in range.

table/internal/parquet_files.go
tanmayrauth: Added MetricModeNone and MetricModeCounts sub-cases to TestWriteDataFileGeoBounds asserting the geo field ID is absent from lower/upper — that's exactly the fallthrough you flagged.

table/internal/geo_codec.go
tanmayrauth: Documented the contract explicitly rather than restructuring: geoStatsAgg carries bounds precomputed from raw WKB, is injected into ColAggs only after DataFileStatsFromMeta has finished folding Parquet stats, and is drained via MinAsBytes/MaxAsBytes in ToDataFile — so nothing calls Update, and a call would  discard precomputed bounds. I kept it on the StatsAgg interface since a separate carrier means a parallel path through DataFileStatistics/ToDataFile, which felt heavier than the risk warranted. Happy to switch to the explicit carrier if you'd still prefer it.

table/internal/geo_codec.go
tanmayrauth: golangci-lint run is green on the current tip (0 issues), including the early returns in update()/extend() and the new returns in layout() — looks like this was already resolved. Let me know if CI still flags something and I'll chase the exact line.

table/internal/parquet_files.go
tanmayrauth: Done — made it a two-return + continue so a future change to how geoAccs is populated can't turn a missing key into a nil-panic on acc.AddWKB.

table/internal/parquet_files_test.go
tanmayrauth: Added a comment stating the test guards the createStatsAgg panic via the isGeoType gate (geo types have no generic stats aggregator), and split the full-mode path into its own named sub-case so the intent reads clearly. The MetricModeNone/Counts sub-cases above cover the applyGeoBounds gating.

literals.go
tanmayrauth: Both addressed: annotated case 16, 24, 32: as an intentional valid-sizes fall-through, and added a warning that the returned BinaryLiteral holds raw little-endian coordinates — a future geo predicate evaluator must check the original type rather than treating it as a comparable binary bound.

table/internal/parquet_files.go
tanmayrauth: Added // TODO(#992): compute geo bounds for geo columns nested in structs/lists/maps and noted the divergence from Java/PyIceberg. I anchored it to #992 (this work's parent) for now — happy to file a dedicated follow-up issue and repoint the TODO if you'd prefer it tracked separately.

zeroshade (approved):
LGTM from my side — the geo-bounds encoding, geography-unbounded handling, and mixed-dimensionality concerns from the earlier round look addressed. Holding the merge until @laskoviymishka's requested changes are resolved.

tanmayrauth:
@zeroshade  I think we can merge this. I've incorporated all the requested changes.

laskoviymishka:
@tanmayrauth give me couple more moments, almost done

literals.go
```diff
@@ -174,6 +174,24 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
 		var v BinaryLiteral
 		err := v.UnmarshalBinary(data)
 
+		return v, err
+	case GeometryType, GeographyType:
```
laskoviymishka, line 178: The validation and byte layout here are right, but the returned `BinaryLiteral` reports `.Type() == Binary`, not Geometry/Geography. We're safe today only because `evaluators.go` dispatches on `field.Type` rather than `lit.Type()` — the moment anything keys off the literal's own type (pruning, `.To()`, logging), it sees binary and routes through `bytes.Compare`, which isn't a valid ordering over LE coords and gives silently-wrong answers instead of erroring.

I'd rather not leave that as a doc comment asking every caller to check the source type. Could we add a thin `GeoLiteral` wrapper (or parametrize `BinaryLiteral`'s `Type()`) so `.Type()` is honest, and gate the ordering-comparison paths to reject geo terms explicitly? wdyt?

laskoviymishka (changes requested):
@tanmayrauth  - sorry, one small gap and i good to go :)

literals.go
tanmayrauth: Good catch, agreed. I'll add a dedicated geo literal so .Type() reports Geometry/Geography honestly, and make its ordering paths (Comparator, To) error instead of silently running bytes.Compare, LE coord bytes have no valid ordering anyway. I'll also gate the ordering visitors in evaluators.go so a geo term  returns rowsMightMatch rather than routing through the comparator. Pushing that now with a test. Thanks!

tanmayrauth:
@laskoviymishka I've pushed the changes.

laskoviymishka (approved):

PR #1374, opened 2026-07-04
fix(table): reject nil table in PositionDeltaWriter

fallintoplace:
## Problem
`NewPositionDeltaWriter` currently panics when called with a nil `*Table` because it dereferences `tbl.metadata` immediately.

For a public constructor, that turns a bad call site into a runtime crash instead of an actionable error.

## Fix
- Add a nil guard in `NewPositionDeltaWriter`:
  - returns `fmt.Errorf("%w: table is nil", iceberg.ErrInvalidArgument)` when `tbl` is nil
- Keep the existing version check behavior unchanged for non-nil tables.
- Add regression test `TestPositionDeltaWriter_NilTable` asserting:
  - an error is returned
  - error is wrapped by `iceberg.ErrInvalidArgument`
  - returned writer is nil

## Validation
- `go test ./table -run TestPositionDeltaWriter_NilTable -count=1`
- `go test ./table -run "TestPositionDeltaWriter_" -count=1`

## Scope
No user-visible data-path behavior change; this is a targeted API-safety fix for invalid constructor input.

### DIFF
--- a/table/position_delta_writer.go
+++ b/table/position_delta_writer.go
@@ -86,6 +86,10 @@ type PositionDeltaWriter struct {
 // those two options; everything else (target file size, worker count,
 // clustered write) flows through.
 func NewPositionDeltaWriter(tbl *Table, opts ...WriteRecordOption) (*PositionDeltaWriter, error) {
+	if tbl == nil {
+		return nil, fmt.Errorf("%w: table is nil", iceberg.ErrInvalidArgument)
+	}
+
 	if tbl.metadata.Version() < 3 {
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM,  clean nil guard, correctly placed before the metadata deref, and the regression test covers all three assertions. Nothing blocking from me.

zeroshade (approved):
LGTM. The nil-table guard at the constructor returns ErrInvalidArgument before the metadata dereference, and the regression test covers it.

PR #1375, opened 2026-07-04
fix(table): return errors for invalid writer file schemas

fallintoplace:
## Summary
- Change writer factory options to return errors so `withFactoryFileSchema` can propagate Arrow schema conversion failures instead of panicking.
- Update `withFactoryFileSchema` to return a construction error when `SchemaToArrowSchema` fails.
- Update `newWriterFactory` to stop option iteration and return the first option error.
- Add regression test: `TestNewWriterFactoryReturnsErrorForInvalidFileSchema` verifies invalid/unsupported file schemas fail with error and no panic.

## Verification
- `go test ./table -run TestRollingDataWriter/TestNewWriterFactoryReturnsErrorForInvalidFileSchema -count=1`
- `go test ./table -run TestRollingDataWriter -count=1`
- `go test ./table -count=1`

### DIFF
--- a/table/rolling_data_writer.go
+++ b/table/rolling_data_writer.go
@@ -73,28 +73,34 @@ type writerFactory struct {
 	mu                    sync.Mutex
 }
 
-type writerFactoryOption func(*writerFactory)
+type writerFactoryOption func(*writerFactory) error
 
 func withContentType(content iceberg.ManifestEntryContent) writerFactoryOption {
-	return func(w *writerFactory) {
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM, clean, well-scoped fix. Nice one.

zeroshade (approved):
LGTM. Invalid writer file schemas now surface as returned/iterated errors instead of a factory-time panic, and the option-application error path is covered.

PR #1376, opened 2026-07-04
fix(partitions): copy partition fields in NewPartitionSpecID

fallintoplace:
## Summary

Fixes mutable aliasing when creating a partition spec.

`NewPartitionSpecID` now copies the variadic field arguments and each field's `SourceIDs` slice before initialization, so callers can safely reuse/mutate their input slice after constructing a spec.

Also added a regression test:
- `TestNewPartitionSpecIDCopiesFields` in `partitions_test.go`

## Validation
- `go test ./... -count=1`

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -351,7 +351,14 @@ func NewPartitionSpec(fields ...PartitionField) PartitionSpec {
 //
 // The fields are not verified against a schema, use NewPartitionSpecOpts if you have to ensure compatibility.
 func NewPartitionSpecID(id int, fields ...PartitionField) PartitionSpec {
-	ret := PartitionSpec{id: id, fields: fields}
+	fieldCopies := make([]PartitionField, len(fields))
+	for i, field := range fields {
+		fieldCopies[i] = field
+		if field.SourceIDs != nil {
+			fieldCopies[i].SourceIDs = slices.Clone(field.SourceIDs)
+		}
+	}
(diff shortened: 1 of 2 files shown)


partitions.go
```diff
@@ -351,7 +351,14 @@ func NewPartitionSpec(fields ...PartitionField) PartitionSpec {
 //
 // The fields are not verified against a schema, use NewPartitionSpecOpts if you have to ensure compatibility.
 func NewPartitionSpecID(id int, fields ...PartitionField) PartitionSpec {
-	ret := PartitionSpec{id: id, fields: fields}
+	fieldCopies := make([]PartitionField, len(fields))
+	for i, field := range fields {
+		fieldCopies[i] = field
+		if field.SourceIDs != nil {
```
tanmayrauth, line 357: this nil guard is redundant — `slices.Clone(nil)` already returns nil,  so you can drop the `if` and just always `fieldCopies[i].SourceIDs = slices.Clone(field.SourceIDs)`. Behavior is identical, one less branch.

tanmayrauth (commented):
LGTM, correct fix and the regression test pins exactly the right thing. One nit inline.

zeroshade (approved):
LGTM. Copying the variadic partition fields and cloning each SourceIDs slice before initialize fixes the input-aliasing bug, and the mutation test covers it. The separate bucket-count validation is handled in #1365.

PR #1377, opened 2026-07-04
fix(gocloud): parse gcs.usejsonapi as boolean

fallintoplace:
## Summary

`ParseGCSConfig` used to enable JSON reads on `gcs.usejsonapi` whenever the key existed, so values like `"false"` still switched it on.

This change now parses the property as a boolean using `strconv.ParseBool` and only enables JSON reads when it resolves to `true`.

Also included:
- Unit tests covering `true`, `false`, absent, and invalid values (`not-a-bool`).
- Documentation update in `website/src/configuration.md`.
- Removed outdated comment in `io/config.go` that implied truthy-by-presence behavior.

## Validation

- `go test ./io/gocloud -run TestParseGCSConfigUseJSONAPI -count=1`

### DIFF
--- a/io/config.go
+++ b/io/config.go
@@ -40,7 +40,7 @@ const (
 	GCSKeyPath    = "gcs.keypath"
 	GCSJSONKey    = "gcs.jsonkey"
 	GCSCredType   = "gcs.credtype"
-	GCSUseJSONAPI = "gcs.usejsonapi" // set to anything to enable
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
LGTM. Parsing gcs.usejsonapi via strconv.ParseBool preserves the disabled-by-default behavior and matches how other boolean properties are handled; tests cover unset/true/false/invalid.

PR #1378, opened 2026-07-04
fix(table): close parquet input file on reader init failure

fallintoplace:
## What changed
- Fixes a resource leak in table/internal/parquet_files.go where parquetFormat.Open returned without closing the opened file when constructing the parquet/parquet-arrow reader failed.
- Added regression tests covering both failure paths: failing file->parquet reader initialization and failing file->arrow reader initialization.

## Validation
- go test ./table/internal -run TestParquetOpenClosesFileOnNew -count=1
- go test ./table/internal -count=1

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -74,7 +74,7 @@ const (
 
 type parquetFormat struct{}
 
-func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (result FileReader, err error) {
+func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (_ FileReader, err error) {
 	inputfile, err := fs.Open(path)
 	if err != nil {
 		return nil, err
@@ -82,7 +82,9 @@ func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (result
 
 	defer func() {
 		if err != nil {
-			_ = inputfile.Close()
+			if closeErr := inputfile.Close(); closeErr != nil {
+				err = errors.Join(err, closeErr)
+			}
 		}
 	}()
 
@@ -97,9 +99,7 @@ func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (result
 		return nil, err
 	}
 
-	result = wrapPqArrowReader{arrRdr}
-
-	return result, nil
+	return wrapPqArrowReader{arrRdr}, nil
 }
 
 func (parquetFormat) PathToIDMapping(sc *iceberg.Schema) (map[string]int, error) {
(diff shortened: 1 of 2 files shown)


table/internal/parquet_files.go
```diff
@@ -74,12 +74,18 @@ const (
...
-func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (FileReader, error) {
+func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (result FileReader, err error) {
 	inputfile, err := fs.Open(path)
 	if err != nil {
 		return nil, err
 	}
 
+	defer func() {
+		if err != nil {
+			_ = inputfile.Close()
```
zeroshade: This cleanup path closes the input file on reader initialization errors, but it drops any Close error. Elsewhere in this file cleanup failures are aggregated with errors.Join; please preserve that behavior here too, e.g. `if closeErr := inputfile.Close(); closeErr != nil { err = errors.Join(err, closeErr) }`.

zeroshade (changes requested):
Thanks for fixing parquetFormat.Open; that is the right cleanup pattern for the file handle opened there. I do need to request changes because the same open -> parquet reader init -> arrow reader init sequence still exists in ParquetFileSource.GetReader and can leak the opened file on initialization errors. That path is used by scan/delete reader construction, so it should get the same error-path cleanup before this lands.

Specifically, table/internal/parquet_files.go real lines 933-958 still opens pf, then returns directly if file.NewParquetReader or pqarrow.NewFileReader fails, without closing pf. Please add the same cleanup pattern there (for example, named err plus a defer after Open, or explicit Close with errors.Join on each error return).

There is also one cleanup-error propagation issue inline: the new parquetFormat.Open defer closes the file on error, but drops any Close error instead of aggregating it as this file does elsewhere.

table/internal/parquet_files.go
```diff
@@ -74,12 +74,18 @@ const (
 
 type parquetFormat struct{}
 
-func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (FileReader, error) {
+func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (result FileReader, err error) {
```
tanmayrauth: The defer only reads `err`, never `result`, so the named `result` isn't buying anything — you could drop it to `(_ FileReader, err error)` and keep `return wrapPqArrowReader{arrRdr}, nil` at the bottom, which still satisfies nlreturn without the `result = ...; return result, nil` two-step.

tanmayrauth (commented):
The fix looks right — NewParquetReader doesn't close on failure so there's no double-close, and on success Close() still flows through to the input file. One thing worth folding in: GetReader at parquet_files.go:933-961 has the identical leak — pf (opened at :934) isn't closed if file.NewParquetReader (:941) or pqarrow.NewFileReader (:957) fails. Same pattern, two functions down. Plus a couple of nits inline.

table/internal/parquet_files.go
```diff
@@ -74,12 +74,20 @@ const (
...
```
zeroshade: nit: the named return `result` isn't needed here — the defer only inspects `err`. `(_ FileReader, err error)` plus `return wrapPqArrowReader{...}, nil` would be a touch cleaner. Not blocking.

zeroshade (approved):
LGTM — the input file is now closed with `errors.Join` on both the `Open` and `GetReader` init-failure paths, and the tests cover all four failure branches. One style nit inline.

laskoviymishka (approved):
:shipit:

PR #1379, opened 2026-07-04
fix(cli): resolve partition spec fields by schema name

fallintoplace:
## Summary
- Resolve CLI --partition-spec entries against the parsed table schema instead of positional field IDs.
- Update table create path to pass schema into partition spec parser.
- Add regression tests for correct field-id mapping and unknown field rejection.

## Testing
- go test ./cmd/iceberg -run TestParsePartitionSpec -count=1
- go test ./cmd/iceberg -count=1

### DIFF
--- a/cmd/iceberg/utils.go
+++ b/cmd/iceberg/utils.go
@@ -48,31 +48,33 @@ func parseProperties(propStr string) (iceberg.Properties, error) {
 	return props, nil
 }
 
-func parsePartitionSpec(specStr string) (*iceberg.PartitionSpec, error) {
+func parsePartitionSpec(specStr string, schema *iceberg.Schema) (*iceberg.PartitionSpec, error) {
+	if schema == nil {
+		return nil, fmt.Errorf("schema is required for partition spec parsing")
+	}
+
 	if specStr == "" {
 		return iceberg.UnpartitionedSpec, nil
 	}
 
 	fields := strings.Split(specStr, ",")
-	var partitionFields []iceberg.PartitionField
+	opts := make([]iceberg.PartitionOption, 0)
 
-	for i, field := range fields {
+	for _, field := range fields {
 		field = strings.TrimSpace(field)
 		if field == "" {
 			continue
 		}
 
-		partitionFields = append(partitionFields, iceberg.PartitionField{
-			SourceIDs: []int{i + 1},
-			FieldID:   i + iceberg.PartitionDataIDStart,
-			Name:      field,
-			Transform: iceberg.IdentityTransform{},
-		})
+		opts = append(opts, iceberg.AddPartitionFieldByName(field, field, iceberg.IdentityTransform{}, schema, nil))
 	}
-	if len(partitionFields) == 0 {
+	if len(opts) == 0 {
 		return iceberg.UnpartitionedSpec, nil
 	}
-	spec := iceberg.NewPartitionSpec(partitionFields...)
+	spec, err := iceberg.NewPartitionSpecOpts(opts...)
+	if err != nil {
+		return nil, err
+	}
 
 	return &spec, nil
 }
--- a/cmd/iceberg/utils.go
+++ b/cmd/iceberg/utils.go
@@ -18,6 +18,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -50,7 +51,7 @@ func parseProperties(propStr string) (iceberg.Properties, error) {
 
 func parsePartitionSpec(specStr string, schema *iceberg.Schema) (*iceberg.PartitionSpec, error) {
 	if schema == nil {
-		return nil, fmt.Errorf("schema is required for partition spec parsing")
+		return nil, errors.New("schema is required for partition spec parsing")
 	}
 
 	if specStr == "" {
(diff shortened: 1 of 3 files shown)


cmd/iceberg/utils.go
```diff
@@ -48,31 +49,33 @@ func parseProperties(propStr string) (iceberg.Properties, error) {
...
 
 	fields := strings.Split(specStr, ",")
-	var partitionFields []iceberg.PartitionField
+	opts := make([]iceberg.PartitionOption, 0)
 
-	for i, field := range fields {
+	for _, field := range fields {
 		field = strings.TrimSpace(field)
 		if field == "" {
 			continue
```
zeroshade, line 68: Passing `nil` here leaves each generated `PartitionField.FieldID` at the internal unassigned value (`0`). `NewPartitionSpecOpts` validates/initializes the spec but does not assign partition field IDs, and REST create-table serializes this spec directly in the request payload. Please assign IDs while parsing (for example `iceberg.PartitionDataIDStart + len(opts)`) and add assertions that parsed fields receive `1000`, `1001`, ... plus a REST create-payload regression test.

zeroshade (changes requested):
Thanks for improving partition-spec parsing against the schema. I found one issue that needs to be fixed before this is safe for REST catalogs.

- `parsePartitionSpec` now builds partition fields via `AddPartitionFieldByName` with a nil field ID. That leaves the resulting `PartitionField.FieldID` at the internal unassigned value, and `NewPartitionSpecOpts` only initializes the spec rather than assigning IDs. Non-REST catalogs may normalize this later, but REST create-table sends `cfg.PartitionSpec` directly in the request payload, so `create table --partition-spec ... --catalog rest` can send field IDs as `0` instead of `1000+`. Please assign stable partition field IDs during parsing and add coverage for parsed field IDs plus the REST create payload.

cmd/iceberg/utils.go
```diff
@@ -48,31 +49,33 @@ func parseProperties(propStr string) (iceberg.Properties, error) {
...
 		}
 
-		partitionFields = append(partitionFields, iceberg.PartitionField{
-			SourceIDs: []int{i + 1},
-			FieldID:   i + iceberg.PartitionDataIDStart,
-			Name:      field,
-			Transform: iceberg.IdentityTransform{},
-		})
+		opts = append(opts, iceberg.AddPartitionFieldByName(field, field, iceberg.IdentityTransform{}, schema, nil))
```
tanmayrauth: `iceberg.AddPartitionFieldByName(field, field, iceberg.IdentityTransform{}, schema, nil)` — the nil field ID is the problem. `NewPartitionSpecOpts` only runs `initialize()`, never `assignPartitionFieldIds`, so every field keeps the
  unassigned value (0). rest.go serializes `cfg.PartitionSpec` directly into the create-table payload, so the server gets `field-id: 0` for all of them — which is also a duplicate and invalid.
  
  Assign a stable ID here instead, e.g.:
  
      id := iceberg.PartitionDataIDStart + len(opts)
      opts = append(opts, iceberg.AddPartitionFieldByName(field, field, iceberg.IdentityTransform{}, schema, &id))
      
  Since empty entries are `continue`d before the append, `len(opts)` gives dense 1000, 1001, … even for input like `a,,b` (the old `i + PartitionDataIDStart` would've left a gap there — small bonus).

cmd/iceberg/utils_test.go
```diff
@@ -81,40 +81,90 @@ func TestParseProperties(t *testing.T) {
...
 				t.Errorf("parsePartitionSpec() returned nil for valid input")
 			}
+			if len(tt.wantSourceIDs) == 0 {
+				require.Equal(t, iceberg.UnpartitionedSpec, got)
+
+				return
+			}
+			require.Equal(t, len(tt.wantSourceIDs), got.NumFields())
+			for i, wantIDs := range tt.wantSourceIDs {
+				require.Equal(t, wantIDs, got.Field(i).SourceIDs)
```
tanmayrauth, line 176: `require.Equal(t, wantIDs, got.Field(i).SourceIDs)` — the tests only assert SourceIDs, never FieldID, which is why they pass with field-id: 0 still going out. Could you add a FieldID assertion (1000, 1001, …) and a REST create-payload regression test that marshals the spec and checks `field-id` isn't 0? That locks in the fix above.

tanmayrauth (commented):
gree with the field-id concern, confirmed the spec goes out with field-id: 0 (and it's a duplicate across fields). One test-gap note too. Details inline.

cmd/iceberg/utils_test.go
```diff
@@ -242,3 +305,123 @@ func TestParseSortOrder(t *testing.T) {
...
+
+	field0, ok := fields[0].(map[string]any)
+	require.True(t, ok)
+	field1, ok := fields[1].(map[string]any)
+	require.True(t, ok)
+	require.EqualValues(t, 1000, field0["field-id"])
+	require.EqualValues(t, 1001, field1["field-id"])
+}
+
+func tableMetadataJSON(tableUUID string) string {
```
zeroshade, line 385: nit: this helper builds a schema that doesn't match the one exercised by the test above — harmless since only the partition spec matters for the assertion, but a matching schema would read less confusingly.

zeroshade (approved):
LGTM — partition field IDs are now assigned from `PartitionDataIDStart` and the REST create-payload test confirms `field-id` is serialized correctly. One cosmetic test nit inline.

laskoviymishka (approved):
🚢

PR #1380, opened 2026-07-04
fix(requirements): validate missing last partition spec id

fallintoplace:
## Summary

Fix a panic in requirement validation when metadata lacks last partition spec ID (for example v1 or malformed metadata).

The previous implementation directly dereferenced `meta.LastPartitionSpecID()` and could panic. This change now checks for nil and returns a clean requirement failure message:

- `requirement failed: last assigned partition id is missing`

When the field is present, behavior is unchanged and the previous comparison path still validates for mismatches.

## Testing

- `go test ./table -run TestAssertLastAssignedPartitionIDValidate -count=1`

### DIFF
--- a/table/requirements.go
+++ b/table/requirements.go
@@ -264,8 +264,13 @@ func (a *assertLastAssignedPartitionId) Validate(meta Metadata) error {
 		return errors.New("Requirement failed: current table metadata does not exist")
 	}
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. The nil LastPartitionSpecID case now returns a requirement failure instead of panicking, consistent with the other requirement validators, and the tests cover missing/match/mismatch.

PR #1381, opened 2026-07-04
fix(sorting): validate sort source IDs

fallintoplace:
## Summary

This change validates sort order source IDs in `NewSortOrder`:

- sort fields with missing/empty `SourceIDs` are rejected
- any non-positive source ID is rejected

This prevents malformed sort metadata from slipping through (for example zero-value sort fields becoming source-id 0).

## Testing

- `go test ./table -count=1`

### DIFF
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -184,8 +184,8 @@ func (s *SortField) UnmarshalJSON(b []byte) error {
 }
 
 func validateSortSourceID(id int) error {
-	if id < 0 {
-		return fmt.Errorf("source ID must be non-negative: %d", id)
+	if id <= 0 {
+		return fmt.Errorf("source ID must be positive: %d", id)
 	}
 
 	return nil
--- a/table/sorting_test.go
+++ b/table/sorting_test.go
@@ -88,12 +88,20 @@ func TestNewSortOrderRejectsInvalidSourceIDs(t *testing.T) {
 			sourceIDs: []int{},
 		},
 		{
+			name:      "zero",
+			sourceIDs: []int{0},
+		},
+		{
 			name:      "negative",
 			sourceIDs: []int{-1},
 		},
 		{
 			name:      "multi arg with negative",
 			sourceIDs: []int{1, -1},
+		},
+		{
+			name:      "multi arg with zero",
+			sourceIDs: []int{1, 0},
 		},
 	} {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


table/sorting.go
```diff
@@ -184,8 +184,8 @@ func (s *SortField) UnmarshalJSON(b []byte) error {
 }
 
 func validateSortSourceID(id int) error {
-	if id < 0 {
-		return fmt.Errorf("source ID must be non-negative: %d", id)
+	if id <= 0 {
```
zeroshade, line 187: [MAJOR] This check now rejects zero only when validation is invoked, but the JSON unmarshal path still bypasses it: SortField.UnmarshalJSON stores the decoded source IDs without presence/positivity validation, and SortOrder.UnmarshalJSON calls newSortOrder(..., false). A missing source-id can still decode leniently and later marshal as source-id: 0, so the round-trip safety issue remains. Please adopt #1367's unmarshal-time presence/positivity checks, or explicitly document that unmarshal is intentionally lenient and validation is deferred. Anchored on the nearest changed line; the affected code is in SortField.UnmarshalJSON and SortOrder.UnmarshalJSON.

table/sorting_test.go
```diff
@@ -185,7 +168,13 @@ func TestSortOrderCheckCompatibilityRejectsInvalidSourceIDs(t *testing.T) {
 		{
 			name:                "negative",
 			jsonData:            `{"order-id": 1, "fields": [{"source-id": -1, "transform": "identity", "direction": "asc", "null-order": "nulls-first"}]}`,
-			wantErr:             "source ID must be non-negative: -1",
+			wantErr:             "source ID must be positive: -1",
+			wantInvalidSourceID: true,
+		},
+		{
+			name:                "zero",
+			jsonData:            `{"order-id": 1, "fields": [{"source-id": 0, "transform": "identity", "direction": "asc", "null-order": "nulls-first"}]}`,
```
zeroshade, line 176: [MINOR] This test still encodes deferred validation: malformed sort source IDs unmarshal successfully and are only rejected later by CheckCompatibility. That directly contradicts #1367 and the stated goal to validate source IDs during unmarshal. Please replace these with rejection tests like #1367, or document that lenient unmarshal is intentional. Anchored on the nearest changed line in the current diff.

zeroshade (changes requested):
CONFLICT: PR #1367 (same author, "fix(sort): validate sort source IDs during unmarshal") is already approved and validates at the unmarshal boundary; #1381 keeps lenient unmarshal and only validates in NewSortOrder/CheckCompatibility, and its tests assert the opposite behavior — the two cannot both land.

Please consolidate into #1367 by adopting unmarshal-time validation, or close #1381, unless the maintainers explicitly want lenient unmarshal plus deferred validation. As written, the behavior and tests disagree with the already-approved direction and weaken round-trip safety.

PR #1382, opened 2026-07-04
fix(table): protect identifier from external mutation

fallintoplace:
## Summary

`Table` stores the caller-provided identifier slice directly, and `Identifier()` returned that same slice. Because `Identifier` is `[]string`, callers could mutate table identity after construction and affect catalog operations.

This patch makes `Table` defensive against aliasing:

- `New(...)` now clones the input identifier before storing it.
- `Identifier()` now returns a cloned identifier.
- Added regression coverage to ensure external identifier mutations cannot affect internal table state.

## Testing

- `go test ./table -run TestTableIdentifierIsolation -count=1`
- `go test ./table -count=1`

### DIFF
--- a/table/commit_test.go
+++ b/table/commit_test.go
@@ -110,6 +110,7 @@ func TestTableCommitDoesNotMarkTransactionCommitted(t *testing.T) {
 func TestTableCommitReturnsCopies(t *testing.T) {
 	tbl := newCommitTestTable(t)
 	tx := tbl.NewTransaction()
+	original := tbl.Identifier()
 
 	require.NoError(t, tx.SetProperties(map[string]string{"key": "value"}))
 
@@ -123,7 +124,10 @@ func TestTableCommitReturnsCopies(t *testing.T) {
 	if len(tc1.Requirements) > 0 {
 		tc1.Requirements = tc1.Requirements[:0]
 	}
+	tc1.Identifier[0] = "different_db"
 	assert.NotEmpty(t, tc2.Requirements)
+	assert.Equal(t, original, tbl.Identifier())
+	assert.Equal(t, original, tc2.Identifier)
 }
 
 func newV1CommitTestTable(t *testing.T) *table.Table {
--- a/table/commit_test.go
+++ b/table/commit_test.go
@@ -49,11 +49,14 @@ func newCommitTestTable(t *testing.T) *table.Table {
 func TestTableCommitFromEmptyTransaction(t *testing.T) {
 	tbl := newCommitTestTable(t)
 	tx := tbl.NewTransaction()
+	original := tbl.Identifier()
 
 	tc, err := tx.TableCommit()
 	require.NoError(t, err)
 
 	assert.Equal(t, table.Identifier{"db", "test_table"}, tc.Identifier)
+	tc.Identifier[0] = "different_db"
+	assert.Equal(t, original, tbl.Identifier())
 	assert.NotNil(t, tc.Requirements, "Requirements must be non-nil empty slice for JSON serialization")
 	assert.NotNil(t, tc.Updates, "Updates must be non-nil empty slice for JSON serialization")
 	assert.Empty(t, tc.Requirements)
--- a/table/scan_planning_test.go
+++ b/table/scan_planning_test.go
@@ -116,6 +116,22 @@ func TestScanPlanningPassesIdentifierCopy(t *testing.T) {
 	assert.Equal(t, Identifier{"db", "scan-copy-test"}, scan.identifier)
 }
 
+func TestTransactionScanCopiesIdentifier(t *testing.T) {
+	txn, _ := createTestTransactionWithMemIO(t, *iceberg.UnpartitionedSpec)
+	txn.tbl.planner = &fakeScanPlanner{
+		result:   ScanPlanningResult{Tasks: []FileScanTask{{}}},
+		supports: true,
+	}
+
+	scan, err := txn.Scan(WithScanPlanningMode(ScanPlanningAuto))
+	require.NoError(t, err)
+	_, err = scan.PlanFiles(context.Background())
+	require.NoError(t, err)
+
+	scan.identifier[0] = "corrupt"
+	assert.Equal(t, Identifier{"db", "tbl"}, txn.tbl.identifier)
+}
+
 func TestScanPlanningUnknownModeErrors(t *testing.T) {
 	t.Parallel()
 
@@ -136,8 +152,7 @@ type fakeScanPlanner struct {
 func (f *fakeScanPlanner) SupportsRemoteScanPlanning() bool { return f.supports }
 
 func (f *fakeScanPlanner) PlanFiles(_ context.Context, req ScanPlanningRequest) (ScanPlanningResult, error) {
-	// Capture a defensive copy to verify call-site identifiers are not shared.
-	f.receivedIdentifier = append([]string(nil), req.Identifier...)
+	f.receivedIdentifier = req.Identifier
 
 	return f.result, f.err
 }
--- a/table/scan_planning_test.go
+++ b/table/scan_planning_test.go
@@ -136,6 +136,7 @@ type fakeScanPlanner struct {
 func (f *fakeScanPlanner) SupportsRemoteScanPlanning() bool { return f.supports }
 
 func (f *fakeScanPlanner) PlanFiles(_ context.Context, req ScanPlanningRequest) (ScanPlanningResult, error) {
+	// Capture a defensive copy to verify call-site identifiers are not shared.
 	f.receivedIdentifier = append([]string(nil), req.Identifier...)
 
 	return f.result, f.err
--- a/table/scan_planning_test.go
+++ b/table/scan_planning_test.go
@@ -94,6 +94,28 @@ func TestScanPlanningAutoUsesCapablePlanner(t *testing.T) {
 	assert.Len(t, tasks, 1)
 }
 
+func TestScanPlanningPassesIdentifierCopy(t *testing.T) {
+	t.Parallel()
+
+	scan := &Scan{
+		planner: &fakeScanPlanner{
+			result:   ScanPlanningResult{Tasks: []FileScanTask{{}}},
+			supports: true,
+		},
+		planningMode:   ScanPlanningAuto,
+		identifier:     Identifier{"db", "scan-copy-test"},
+		selectedFields: []string{"*"},
+	}
+
+	tasks, err := scan.PlanFiles(context.Background())
+	require.NoError(t, err)
+	assert.Len(t, tasks, 1)
+
+	planReq := scan.planner.(*fakeScanPlanner)
+	planReq.receivedIdentifier[0] = "corrupt"
+	assert.Equal(t, Identifier{"db", "scan-copy-test"}, scan.identifier)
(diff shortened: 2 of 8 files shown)


table/table.go
```diff
@@ -103,7 +103,9 @@ func (t Table) Equals(other Table) bool {
 		t.metadata.Equals(other.metadata)
 }
 
-func (t Table) Identifier() Identifier                       { return t.identifier }
+func (t Table) Identifier() Identifier {
+	return slices.Clone(t.identifier)
```
zeroshade: This getter clone is correct, but the same identifier slice still crosses the external CatalogIO boundary. In the current head, Refresh and commit/retry paths still call LoadTable/CommitTable with t.identifier directly (table/table.go lines 152, 480, and 536), which lets a catalog implementation retain or mutate the table's backing array. Please pass slices.Clone(t.identifier) at each catalog call and add a mutating fake-catalog regression test proving tbl.Identifier() stays stable across refresh/commit.

zeroshade (changes requested):
Thanks for adding defensive copies in New and Identifier(); those paths look right. The protection is still incomplete because there are remaining paths that expose the table identifier backing array to callers or external catalog implementations. In table/commit.go, both TableCommit() return paths still set Identifier to t.tbl.identifier directly, so mutating the returned TableCommit.Identifier can mutate the table/transaction identifier. Please clone in both return paths and add a regression test that mutates the returned identifier and verifies tbl.Identifier() remains unchanged. Also, the catalog boundary still needs the same treatment as noted inline.

tanmayrauth (commented):
The clones in `New()` and `Identifier()` look right. Agreeing with @zeroshade that the same backing array still escapes elsewhere — `TableCommit()` (commit.go:63 and :77) returns `t.tbl.identifier` directly, and `Refresh`/commit-retry (table.go:152, :480, :536) hand `t.identifier` straight to the external `CatalogIO`. One more not yet mentioned: the Scan path aliases too — `table.go:856` sets `Scan.identifier` from `t.identifier` uncloned, and it then crosses the `ScanPlanner` boundary at `scanner.go:823`, same class as the catalog case.

  Only flagging that all of these are pre-existing lines outside this PR's two-file diff, so they can't be reviewed inline here. Happy either way — fold them into this PR so the fix is complete, or land this and track the remaining boundaries in a quick follow-up. Whichever, a regression test that mutates a returned `TableCommit.Identifier` (and one that mutates through a fake catalog/planner) would lock it down.

table/table.go
```diff
@@ -554,7 +556,7 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
 
 	deleteOldMetadata(fs, t.metadata, newMeta)
 
-	return New(t.identifier, newMeta, newLoc, t.fsF, t.cat), nil
+	return New(slices.Clone(t.identifier), newMeta, newLoc, t.fsF, t.cat), nil
```
zeroshade: nit: `New()` already clones the identifier on store, so `slices.Clone(t.identifier)` here is a double-clone on the commit path. Harmless for 2–3 element identifiers — fine to leave, just noting it.

zeroshade (approved):
LGTM — every identifier boundary (getter, Refresh, doCommit, Scan, remote plan, TableCommit) now clones, and the tests demonstrate isolation. One harmless redundancy inline.

table/scan_planning_test.go
```diff
@@ -107,11 +129,16 @@ type fakeScanPlanner struct {
...
+	receivedIdentifier Identifier
 }
 
 func (f *fakeScanPlanner) SupportsRemoteScanPlanning() bool { return f.supports }
 
-func (f *fakeScanPlanner) PlanFiles(context.Context, ScanPlanningRequest) (ScanPlanningResult, error) {
+func (f *fakeScanPlanner) PlanFiles(_ context.Context, req ScanPlanningRequest) (ScanPlanningResult, error) {
+	// Capture a defensive copy to verify call-site identifiers are not shared.
+	f.receivedIdentifier = append([]string(nil), req.Identifier...)
+
```
laskoviymishka, line 156: This test can't fail. The fake planner copies `req.Identifier` into its own backing array before storing it (`append([]string(nil), ...)`), so mutating `planReq.receivedIdentifier[0]` only touches that copy — it can never reach `scan.identifier` whether or not `planFilesRemote` clones. Remove the production clone at `scanner.go` and this stays green.

I'd store the slice as-is instead — `f.receivedIdentifier = req.Identifier` — so the captured slice aliases whatever the call site passed. Then mutating it corrupts `scan.identifier` iff `planFilesRemote` didn't clone, which is the property we actually want to assert.

table/table.go
```diff
@@ -531,7 +533,7 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
 			return nil, context.Cause(retryCtx)
 		}
 
-		newMeta, newLoc, err = t.cat.CommitTable(retryCtx, t.identifier, reqs, updates)
+		newMeta, newLoc, err = t.cat.CommitTable(retryCtx, slices.Clone(t.identifier), reqs, updates)
```
laskoviymishka, line 568: The retry branch here is the one that's still open from last round. I listed `table.go:480` alongside `Refresh` and `CommitTable` — those two got the clone, but the `LoadTable` at the top of the retry loop (line 479, just above this) still hands `t.identifier` straight to the catalog. Same `CatalogIO` boundary, same fix: `t.cat.LoadTable(retryCtx, slices.Clone(t.identifier))`.

Worth a test too — the `identifierCapturingCatalog` scaffolding extends naturally to force a retry (first `CommitTable` fails, capture the identifier on the second `LoadTable`, mutate it, assert `t.Identifier()` is unchanged). If you go that route, `loadIdentifier` will want to accumulate into a slice since it's last-write-wins today.

table/table.go
```diff
@@ -851,7 +853,7 @@ func WithRowLineage() ScanOption {
 
 func (t Table) Scan(opts ...ScanOption) *Scan {
 	s := &Scan{
-		identifier:       t.identifier,
+		identifier:       slices.Clone(t.identifier),
```
laskoviymishka, line 904: Same class as this Scan path — `Table.Scan()` clones here now, but the parallel `Transaction.Scan()` at `transaction.go:2177` still does `identifier: t.tbl.identifier` raw. A `ScanPlanner` that retains the slice past `PlanFiles` and mutates it would corrupt the owning table's identity through that path. `slices.Clone` there too (`slices` is already imported in `transaction.go`).

`TestScanPlanningPassesIdentifierCopy` builds the `Scan` by hand, so it doesn't cover `Transaction.Scan()` — once that test can actually fail (see the other comment), pointing one variant through `Transaction.Scan()` would close this.

laskoviymishka (changes requested):
Thanks for turning this around — the boundaries I listed last round are in: both `TableCommit()` return paths clone, `Refresh` and the `CommitTable` call in `doCommit` clone, and `Scan` plus the remote-plan request clone. The regression tests through the fake catalog and fake planner are exactly what I was hoping to see.

Almost there. A few small things before I'm comfortable calling the identifier fully protected:

The `doCommit` retry branch is the one that's still open from last round. I listed `table.go:480` alongside `Refresh` and `CommitTable` — those two got the clone, but the `LoadTable` at the top of the retry loop still hands `t.identifier` straight to the catalog. Same boundary, same one-line fix.

`Transaction.Scan()` is a sibling of the Scan path we just fixed — `Table.Scan()` clones now, but `transaction.go:2177` still aliases `t.tbl.identifier` into the returned scan. Worth closing the same way.

And `TestScanPlanningPassesIdentifierCopy` can't actually fail: the fake planner copies the identifier into its own backing array before storing it, so the mutation never reaches `scan.identifier` — remove the production clone and the test stays green. Storing the slice as-is fixes that.

On the clone-on-both-ends redundancy @zeroshade flagged — agreed, `New()` already clones at construction so the per-read clone in `Identifier()` is belt-and-suspenders. Harmless, not blocking; your call whether to trim it.

- clone in the `doCommit` retry-branch `LoadTable` (table.go:479)
- clone in `Transaction.Scan()` (transaction.go:2177)
- make `TestScanPlanningPassesIdentifierCopy` able to fail (store `req.Identifier` as-is)

Close the retry-branch one and the `Transaction.Scan` sibling the same way as the rest, get that scan test failing on a missing clone, and I'm happy to approve. Really nice cleanup on the boundaries that are already in.

table/table.go
laskoviymishka, line 568: The straight-commit path here is covered nicely by `TestDoCommitPassesIdentifierCopy` — it mutates `cat.commitIdentifier` post-commit and confirms `tbl.Identifier()` holds. The one case we don't exercise is an actual retry: the fake catalog always succeeds, so the loop never re-enters and the clone on the retry attempt goes unchecked.

Not blocking — but if the fake rejected once before succeeding, we'd close that gap too. wdyt?

table/table.go
```diff
@@ -103,7 +103,9 @@ func (t Table) Equals(other Table) bool {
 		t.metadata.Equals(other.metadata)
 }
 
-func (t Table) Identifier() Identifier                       { return t.identifier }
+func (t Table) Identifier() Identifier {
```
laskoviymishka: Tiny thing — the siblings right below (`Metadata()`, `MetadataLocation()`, `FS()`) are all aligned one-liners, so the block body stands out. I'd keep it on one line:

```go
func (t Table) Identifier() Identifier { return slices.Clone(t.identifier) }
```

table/commit_test.go
```diff
@@ -123,7 +124,10 @@ func TestTableCommitReturnsCopies(t *testing.T) {
 	if len(tc1.Requirements) > 0 {
 		tc1.Requirements = tc1.Requirements[:0]
 	}
+	tc1.Identifier[0] = "different_db"
 	assert.NotEmpty(t, tc2.Requirements)
+	assert.Equal(t, original, tbl.Identifier())
+	assert.Equal(t, original, tc2.Identifier)
```
laskoviymishka, line 133: Nice — this locks down the non-empty-updates return. The zero-updates branch of `TableCommit()` got the same `slices.Clone` but isn't exercised here. Totally optional, but a second case that commits with no updates and mutates the returned `Identifier` would cover it symmetrically.

laskoviymishka (approved):
Thanks for turning this around again — all three things from last round are in, and they hold up. The `doCommit` retry-branch `LoadTable` clones now, `Transaction.Scan()` clones `t.tbl.identifier` the same way `Table.Scan()` does, and `TestScanPlanningPassesIdentifierCopy` can actually fail now that the fake planner stores `req.Identifier` as-is. That's the identifier protected across every boundary, and the regression tests are real guards rather than vacuous ones.

I left a couple of small, non-blocking notes inline — a narrow retry-path coverage gap on the `CommitTable` clone (the straight-commit path is guarded, an actual retry isn't exercised) and a one-line formatting nit on `Identifier()`. Both are take-it-or-leave-it.

One heads-up, explicitly not for this PR: `view.View` has the identical aliasing pattern — `View.Identifier()` returns the raw slice and `view.New()` stores `ident` uncloned, the same class of bug we just fixed here for `table.Table`. I'd file a quick follow-up so the sister type stays consistent rather than growing this PR.

This is essentially there for me. Address the inline nits if you like, and I'm happy to approve.

PR #1383, opened 2026-07-04
fix(types): safe compare defaults in NestedField.Equals

fallintoplace:
## Summary

`NestedField.Equals` compared `InitialDefault` and `WriteDefault` with `==`, which can panic when defaults contain non-comparable values (maps/slices/etc).

This changes equality to use `reflect.DeepEqual` for defaults so schema/field comparisons stay safe and correct with nested defaults.

Also adds a regression test that verifies:
- non-comparable defaults no longer panic in `Equals`, and
- equivalent deep structures compare equal while different defaults compare false.

## Testing

- `go test ./... -run 'TestNestedFieldEqualsWithNonComparableDefaults|TestUnknownTypeInNestedStructs|TestTypesBasic' -count=1`

### DIFF
--- a/types.go
+++ b/types.go
@@ -20,6 +20,7 @@ package iceberg
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"regexp"
 	"slices"
 	"strconv"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Switching to reflect.DeepEqual makes NestedField.Equals panic-safe for non-comparable defaults while preserving raw-value equality; the non-comparable-defaults test covers it.

PR #1384, opened 2026-07-04
fix(view): validate SQL representations in NewVersion

fallintoplace:
## Summary

- Add strict validation in `view.NewVersion` for each representation:
  - `Type` must be exactly `"sql"`
  - `Sql` must be non-empty
  - `Dialect` must be non-empty
  - duplicate dialects rejected case-insensitively
- Add regression test coverage for malformed/duplicate representation combinations in `view/metadata_builder_test.go`.
- Keep `NewVersionFromSQL` behavior aligned through delegated constructor validation.

## Testing
- `go test ./view -count=1`

### DIFF
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -173,6 +173,18 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
 		return nil, errors.New("invalid view version: must have at least one representation")
 	}
 
+	seenDialects := make(map[string]struct{})
+	for _, repr := range representations {
+		normalizedDialect, err := validateRepresentation(repr)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %s", ErrInvalidViewMetadata, err)
+		}
+		if _, ok := seenDialects[normalizedDialect]; ok {
+			return nil, fmt.Errorf("%w: Invalid view version: Cannot add multiple queries for dialect %s", ErrInvalidViewMetadata, normalizedDialect)
+		}
+		seenDialects[normalizedDialect] = struct{}{}
+	}
+
 	version := &Version{
 		VersionID:        id,
 		SchemaID:         schemaID,
@@ -187,6 +199,23 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
 	}
 
 	return version, nil
+}
+
+func validateRepresentation(repr Representation) (string, error) {
+	if repr.Type != "sql" {
+		return "", errors.New("invalid view representation: type should be \"sql\"")
+	}
+
+	if strings.TrimSpace(repr.Sql) == "" {
+		return "", errors.New("invalid view representation: sql is required")
+	}
+
+	dialect := strings.TrimSpace(repr.Dialect)
+	if dialect == "" {
+		return "", errors.New("invalid view representation: dialect is required")
+	}
+
+	return strings.ToLower(dialect), nil
 }
 
 // NewVersionFromSQL creates a new Version with a single representation
@@ -446,10 +475,14 @@ func (m *metadata) checkDialectsUnique() error {
 	for _, version := range m.VersionList {
 		seenDialects := make(map[string]bool)
 		for _, repr := range version.Representations {
-			dialect := strings.ToLower(repr.Dialect)
+			dialect, err := validateRepresentation(repr)
+			if err != nil {
+				return fmt.Errorf("%w: version %d: %s", ErrInvalidViewMetadata, version.VersionID, err)
+			}
+
 			if seenDialects[dialect] {
 				return fmt.Errorf("%w: version %d has duplicate dialect %s",
-					ErrInvalidViewMetadata, version.VersionID, repr.Dialect)
+					ErrInvalidViewMetadata, version.VersionID, dialect)
 			}
 			seenDialects[dialect] = true
 		}
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -252,10 +252,14 @@ func (v *Version) Clone() *Version {
 // sqlDialects returns a set of strings representing the SQL dialects supported in this version.
 // Dialects are deduplicated by trimmed, lowercase comparison.
 func (v *Version) sqlDialects() internal.Set[string] {
-	return internal.ToSet(internal.MapSlice(
-		v.Representations,
-		func(r Representation) string { return strings.ToLower(strings.TrimSpace(r.Dialect)) },
-	))
+	dialects := make([]string, 0, len(v.Representations))
+	for _, repr := range v.Representations {
+		if repr.Type == "sql" {
+			dialects = append(dialects, strings.ToLower(strings.TrimSpace(repr.Dialect)))
+		}
+	}
+
+	return internal.ToSet(dialects)
 }
 
 type VersionLogEntry struct {
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -180,7 +180,7 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
 			return nil, fmt.Errorf("%w: %s", ErrInvalidViewMetadata, err)
 		}
 		if _, ok := seenDialects[normalizedDialect]; ok {
-			return nil, fmt.Errorf("%w: Invalid view version: Cannot add multiple queries for dialect %s", ErrInvalidViewMetadata, normalizedDialect)
+			return nil, fmt.Errorf("%w: Invalid view version: Cannot add multiple queries for dialect %s", ErrInvalidViewMetadata, repr.Dialect)
 		}
 		seenDialects[normalizedDialect] = struct{}{}
 	}
@@ -201,6 +201,8 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
 	return version, nil
 }
 
+// validateRepresentation validates a SQL representation and returns its
+// normalized dialect for duplicate detection.
 func validateRepresentation(repr Representation) (string, error) {
 	if repr.Type != "sql" {
 		return "", errors.New("invalid view representation: type should be \"sql\"")
@@ -248,11 +250,11 @@ func (v *Version) Clone() *Version {
 }
 
 // sqlDialects returns a set of strings representing the SQL dialects supported in this version.
-// Dialects are deduplicated by lowercase comparison
+// Dialects are deduplicated by trimmed, lowercase comparison.
 func (v *Version) sqlDialects() internal.Set[string] {
 	return internal.ToSet(internal.MapSlice(
 		v.Representations,
-		func(r Representation) string { return strings.ToLower(r.Dialect) },
+		func(r Representation) string { return strings.ToLower(strings.TrimSpace(r.Dialect)) },
 	))
 }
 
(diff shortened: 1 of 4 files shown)


view/metadata.go
```diff
@@ -173,6 +173,28 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
 		return nil, errors.New("invalid view version: must have at least one representation")
 	}
 
+	seenDialects := make(map[string]struct{})
+	for _, repr := range representations {
+		if repr.Type != "sql" {
```
zeroshade: The NewVersion checks are a good addition, but this leaves the public builder path able to bypass them. Callers can still construct a *Version directly and pass it to MetadataBuilder.AddVersion; addVersion only checks empty representations/duplicate dialects and metadata.validate only calls checkDialectsUnique, so non-sql types or blank/whitespace SQL/dialect values can still be built. Please extract shared representation validation and call it from NewVersion, MetadataBuilder.addVersion, and metadata.validate.

view/metadata.go
```diff
@@ -173,6 +173,28 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
...
+			return nil, errors.New("invalid view representation: type should be \"sql\"")
+		}
+
```
zeroshade: These new representation validation errors should wrap ErrInvalidViewMetadata, like the duplicate-dialect error below and the rest of metadata validation. As written, errors.Is(err, ErrInvalidViewMetadata) is inconsistent for invalid representation cases. Please wrap the type/sql/dialect errors with ErrInvalidViewMetadata.

zeroshade (changes requested):
Thanks for adding the NewVersion representation validation; that is the right direction. I am requesting changes because the validation currently only covers that constructor path, while the public builder still accepts manually constructed versions and can build invalid metadata. Please share the representation validation across the constructor, builder, and metadata validation paths so the invariant is enforced consistently.

view/metadata.go
```diff
@@ -173,6 +173,28 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
...
+		if strings.TrimSpace(repr.Sql) == "" {
+			return nil, errors.New("invalid view representation: sql is required")
+		}
+
+		dialect := strings.TrimSpace(repr.Dialect)
+		if dialect == "" {
+			return nil, errors.New("invalid view representation: dialect is required")
+		}
+
+		normalizedDialect := strings.ToLower(dialect)
```
tanmayrauth: One thing to settle when you extract the helper: this dedup normalizes with TrimSpace + ToLower, but addVersion (metadata_builder.go:184) and checkDialectsUnique (metadata.go:449) only ToLower without the trim — so " spark " vs "spark" collide here but stay distinct in the other two paths. Picking trim+lower in the shared helper fixes that, and also lets you unify the message (checkDialectsUnique says "version %d has duplicate dialect %s" while these say  "Cannot add multiple queries for dialect %s").

view/metadata.go
```diff
@@ -173,6 +173,28 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
 		return nil, errors.New("invalid view version: must have at least one representation")
 	}
 
+	seenDialects := make(map[string]struct{})
+	for _, repr := range representations {
+		if repr.Type != "sql" {
+			return nil, errors.New("invalid view representation: type should be \"sql\"")
```
tanmayrauth: +1 to wrapping these with ErrInvalidViewMetadata — this line plus :183 (sql) and :188 (dialect) use bare errors.New, so errors.Is(err, ErrInvalidViewMetadata) is false for them while the dup-dialect error just below at :193 wraps  it. Worth pulling all of this into one validateRepresentations(reprs) helper and calling it from NewVersion, addVersion (metadata_builder.go:179-189), and metadata.validate — right now the builder only checks empty/dup-dialect and validate() only calls checkDialectsUnique, so a hand-built *Version or any metadata coming through ParseMetadataBytes/UnmarshalJSON can still carry a non-sql type or blank sql/dialect. The parse path is the one I'd most want  covered since that's untrusted input from disk.

tanmayrauth (commented):
Seconding @zeroshade — the checks only guard NewVersion; the builder and the parse path can still build invalid metadata. Two extra details on the shared helper inline so it lands cleanly.

zeroshade (approved):
LGTM — the shared `validateRepresentation` helper unifies validation across `NewVersion`, the builder, and `checkDialectsUnique`, with consistent trim/lowercase and `ErrInvalidViewMetadata` wrapping, and the JSON-deserialization path is covered. Minor pre-existing note (not introduced here): `validateRepresentation` is unexported, so external callers constructing a `Representation` directly can't validate without going through `NewVersion`.

view/metadata.go
```diff
@@ -446,10 +475,14 @@ func (m *metadata) checkDialectsUnique() error {
 	for _, version := range m.VersionList {
 		seenDialects := make(map[string]bool)
 		for _, repr := range version.Representations {
-			dialect := strings.ToLower(repr.Dialect)
+			dialect, err := validateRepresentation(repr)
```
laskoviymishka, line 488: This is the one thing I'd want to tighten before merge. Wiring `validateRepresentation` into `checkDialectsUnique` means the read path (`UnmarshalJSON` → `validate`) now rejects any stored metadata with a non-`sql` type or empty sql/dialect — but Java tolerates unknown representation types via `UnknownViewRepresentation` and only runs dialect dedup over `SQLViewRepresentation`. So a view that Spark/Java writes today with a future representation type loads fine there and fails to parse here.

I'd keep the strict content validation where it belongs — on the write paths (`NewVersion`, `addVersion`) — and on the read path only dedup, guarding the content checks behind `if repr.Type == "sql"` so we skip unknown types the way Java does. Otherwise upgrading the library becomes a breaking change for anyone holding a stored view we'd now refuse to open. wdyt?

view/metadata.go
```diff
@@ -446,10 +475,14 @@ func (m *metadata) checkDialectsUnique() error {
...
+			if err != nil {
+				return fmt.Errorf("%w: version %d: %s", ErrInvalidViewMetadata, version.VersionID, err)
+			}
+
 			if seenDialects[dialect] {
 				return fmt.Errorf("%w: version %d has duplicate dialect %s",
-					ErrInvalidViewMetadata, version.VersionID, repr.Dialect)
+					ErrInvalidViewMetadata, version.VersionID, dialect)
```
laskoviymishka: Small one, but worth a note: this now prints the normalized (trimmed + lowercased) dialect in the duplicate error, where before it was `repr.Dialect` as-written. The `TestDuplicateDialects` change (`"SPARK"` → `" SPARK "`, asserting `"spark"`) quietly bakes that in, so the behavior change isn't obvious from the test.

I'd pass `repr.Dialect` to the message and keep the normalized value only as the dedup key — closer to what someone reading the error would expect to see.

view/metadata.go
```diff
@@ -189,6 +201,23 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
...
+	if strings.TrimSpace(repr.Sql) == "" {
+		return "", errors.New("invalid view representation: sql is required")
+	}
+
+	dialect := strings.TrimSpace(repr.Dialect)
+	if dialect == "" {
+		return "", errors.New("invalid view representation: dialect is required")
+	}
+
+	return strings.ToLower(dialect), nil
```
laskoviymishka, line 220: While we're here — now that this canonicalizes with trim-then-lowercase, `sqlDialects()` just below is out of step: it still does `strings.ToLower(r.Dialect)` without the `TrimSpace`. A stored `" spark "` normalizes to `"spark"` through this helper but stays padded in `sqlDialects()`, and that key feeds the drop-dialect guard in `checkIfDialectIsDropped` — so the two disagree on whether a dialect is present and we could throw a false "dropping dialects" error.

I'd give `sqlDialects()` the same trim before lowercasing so there's one canonical form.

view/metadata.go
```diff
@@ -189,6 +201,23 @@ func NewVersion(id int64, schemaID int, representations []Representation, defaul
 	return version, nil
 }
 
+func validateRepresentation(repr Representation) (string, error) {
```
laskoviymishka, line 206: Minor: `validateRepresentation` also normalizes and returns the dialect, so `validate` is doing double duty. It works, but a future caller could easily call it for the check, discard the return, and keep using `repr.Dialect` — reintroducing the exact case/whitespace inconsistency this is fixing.

Not blocking — either a rename toward `normalizeRepresentation`, or a one-line doc on the return, would make the contract obvious.

view/metadata_builder_test.go
```diff
@@ -97,6 +97,84 @@ func TestBuild_NullAndMissingFields(t *testing.T) {
 	assert.ErrorContains(t, err, "cannot set uuid to null")
 }
 
+func TestNewVersion_RepresentationValidation(t *testing.T) {
```
laskoviymishka, line 100: These cases are all sequential `_, err =` assignments sharing one `err`, so if any assertion `t.Fatal`s the rest silently skip — and the happy path is mixed in with the error cases. `TestInvalidRepresentationInJSON` right below already does the table + `t.Run` version of this; I'd match that here.

One gap worth adding while restructuring: every invalid case uses a single bad representation. A `[{valid}, {type: "hive"}]` case — first repr good, second bad — would cover the loop actually reaching the second element, which is the one that'd regress silently if the iteration order ever changed.

laskoviymishka (changes requested):
Nice follow-up. The shared `validateRepresentation` helper resolves the earlier thread cleanly.

Coming in after the approval, I’d only tighten one thing before merge: `checkDialectsUnique` now makes JSON reads strict too, so stored metadata with an unknown representation type fails to load. Java tolerates unknown types and only dedups SQL representations. I’d keep strict validation on write, but make the read path dedup-only and only validate SQL fields when `repr.Type == "sql"`.

Two small fixes:

* keep the original dialect in duplicate errors, use the normalized value only as the key
* apply the same trim + lowercase logic in `sqlDialects()`

Once the read path is relaxed, looks good to me.

view/metadata.go
```diff
@@ -219,11 +250,11 @@ func (v *Version) Clone() *Version {
 }
 
 // sqlDialects returns a set of strings representing the SQL dialects supported in this version.
-// Dialects are deduplicated by lowercase comparison
+// Dialects are deduplicated by trimmed, lowercase comparison.
 func (v *Version) sqlDialects() internal.Set[string] {
 	return internal.ToSet(internal.MapSlice(
 		v.Representations,
-		func(r Representation) string { return strings.ToLower(r.Dialect) },
+		func(r Representation) string { return strings.ToLower(strings.TrimSpace(r.Dialect)) },
```
zeroshade: `sqlDialects()` maps over *all* representations without filtering to `Type == "sql"`. Since unknown/non-sql representation types are now permitted on reads, such a representation contributes its (or an empty) dialect into the set used by the dialect-drop / uniqueness check. That can make `MetadataBuilderFromBase(...).SetCurrentVersion(...).Build()` fail for metadata parsed from JSON that contains a non-sql representation, even though `checkDialectsUnique` skips them. Please filter to `Type == "sql"` before collecting dialects, and add a round-trip regression test with a non-sql representation present.

zeroshade (changes requested):
The SQL representation validation is a useful addition. One gap in the dialect-uniqueness handling (inline).

laskoviymishka (approved):
LGTM to me, just fix @zeroshade 's sqlDialects filter

zeroshade (approved):

PR #1385, opened 2026-07-04
fix(table): clone scan options map in WithOptions

fallintoplace:
## Summary

- Fix `table.WithOptions` to copy the caller's map with `maps.Clone` so scan behavior no longer changes if the caller mutates options after scan creation.
- Add regression test `TestScanWithOptionsAliasing` in `table/table_test.go` to verify scan behavior is isolated from later external map mutations.

## Testing
- `go test ./table -run TestTable -count=1`

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"iter"
 	"log"
+	"maps"
 	"math/rand/v2"
 	"runtime"
 	"slices"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Cloning the options map in WithOptions with maps.Clone prevents caller-mutation aliasing (same pattern as #1353), and the regression test covers it.

PR #1387, opened 2026-07-04
fix: validate format-version in metadata constructors

fallintoplace:
## Why

Table and view metadata creation accepted invalid `format-version` values silently. A bad property like `format-version: banana` currently fell back to the default version and also mutated the caller's props map, which can hide user mistakes and corrupt input ownership.

## What changed

- table metadata (`table/metadata.go`):
  - In `NewMetadataWithUUID`, `format-version` is now parsed with strict `Atoi` validation.
  - Invalid values now return `iceberg.ErrInvalidFormatVersion` instead of silently defaulting.
  - `format-version` is removed from a **cloned** props map before passing into metadata builder, so caller-provided maps are not mutated.

- view metadata (`view/metadata.go`):
  - Mirrored the same strict validation in view metadata creation.
  - Invalid values now return `iceberg.ErrInvalidFormatVersion`.
  - Uses a cloned props map to avoid external mutation.

## Testing

- `go test ./table -run 'TestNewMetadataRejectInvalidFormatVersion|TestNewMetadataWithExplicitV1Format|TestNewMetadataV2Format' -count=1`
- `go test ./view -run 'TestNewMetadataRejectInvalidFormatVersion|TestNewMetadata|TestUnmarshalViewMetadata' -count=1`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2205,16 +2205,20 @@ func NewMetadataWithUUID(sc *iceberg.Schema, partitions *iceberg.PartitionSpec, 
 	if tableUuid == uuid.Nil {
 		tableUuid = uuid.New()
 	}
+	var inputProps iceberg.Properties
 	var err error
 	formatVersion := DefaultFormatVersion
 	if props != nil {
-		verStr, ok := props[PropertyFormatVersion]
+		inputProps = maps.Clone(props)
+		verStr, ok := inputProps[PropertyFormatVersion]
 		if ok {
 			if formatVersion, err = strconv.Atoi(verStr); err != nil {
-				formatVersion = DefaultFormatVersion
+				return nil, fmt.Errorf("%w: %s", iceberg.ErrInvalidFormatVersion, verStr)
 			}
-			delete(props, PropertyFormatVersion)
+			delete(inputProps, PropertyFormatVersion)
 		}
+	} else {
+		inputProps = props
 	}
 
 	reassignedIds, err := reassignIDs(sc, partitions, sortOrder)
@@ -2259,7 +2263,7 @@ func NewMetadataWithUUID(sc *iceberg.Schema, partitions *iceberg.PartitionSpec, 
 		return nil, err
 	}
 
-	if err = builder.SetProperties(props); err != nil {
+	if err = builder.SetProperties(inputProps); err != nil {
 		return nil, err
 	}
 
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2217,8 +2217,6 @@ func NewMetadataWithUUID(sc *iceberg.Schema, partitions *iceberg.PartitionSpec, 
 			}
 			delete(inputProps, PropertyFormatVersion)
 		}
-	} else {
-		inputProps = props
 	}
 
 	reassignedIds, err := reassignIDs(sc, partitions, sortOrder)
--- a/table/metadata_internal_test.go
+++ b/table/metadata_internal_test.go
@@ -780,6 +780,22 @@ func TestNewMetadataWithExplicitV1Format(t *testing.T) {
 	assert.Truef(t, expected.Equals(actual), "expected: %s\ngot: %s", expected, actual)
 }
 
+func TestNewMetadataRejectInvalidFormatVersion(t *testing.T) {
+	props := iceberg.Properties{
+		"format-version": "banana",
+		"foo":            "bar",
+	}
+	schema := iceberg.NewSchema(1, iceberg.NestedField{ID: 1, Name: "x", Type: iceberg.PrimitiveTypes.Int64})
+
+	metadata, err := NewMetadata(schema, iceberg.UnpartitionedSpec, UnsortedSortOrder, "s3://bucket/path", props)
+	require.Error(t, err)
+	require.ErrorIs(t, err, iceberg.ErrInvalidFormatVersion)
+	assert.Nil(t, metadata)
+	assert.Equal(t, "banana", props["format-version"])
+	assert.Equal(t, "bar", props["foo"])
+	assert.Len(t, props, 2)
+}
+
 func TestNewMetadataV2Format(t *testing.T) {
 	schema := iceberg.NewSchemaWithIdentifiers(10,
 		[]int{22},
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -522,16 +522,20 @@ func NewMetadataWithUUID(version *Version, sc *iceberg.Schema, location string, 
 	formatVersion := DefaultViewFormatVersion
 	if props != nil {
 		inputProps = maps.Clone(props)
-		verStr, ok := inputProps["format-version"]
+		verStr, ok := inputProps[table.PropertyFormatVersion]
 		if ok {
 			var err error
 			if formatVersion, err = strconv.Atoi(verStr); err != nil {
 				return nil, fmt.Errorf("%w: %s", iceberg.ErrInvalidFormatVersion, verStr)
 			}
-			delete(inputProps, "format-version")
+			delete(inputProps, table.PropertyFormatVersion)
 		}
 	} else {
 		inputProps = props
+	}
+
+	if formatVersion != DefaultViewFormatVersion {
+		return nil, fmt.Errorf("%w: %d", iceberg.ErrInvalidFormatVersion, formatVersion)
 	}
 
 	// We assume that this constructor is used for building metadata for a new view.
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -530,11 +530,9 @@ func NewMetadataWithUUID(version *Version, sc *iceberg.Schema, location string, 
 			}
 			delete(inputProps, table.PropertyFormatVersion)
 		}
-	} else {
-		inputProps = props
 	}
 
-	if formatVersion != DefaultViewFormatVersion {
+	if formatVersion > SupportedViewFormatVersion {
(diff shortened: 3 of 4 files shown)


view/metadata.go
```diff
@@ -518,16 +518,20 @@ func NewMetadataWithUUID(version *Version, sc *iceberg.Schema, location string,
...
 	formatVersion := DefaultViewFormatVersion
 	if props != nil {
-		verStr, ok := props["format-version"]
+		inputProps = maps.Clone(props)
+		verStr, ok := inputProps["format-version"]
 		if ok {
 			var err error
 			if formatVersion, err = strconv.Atoi(verStr); err != nil {
-				formatVersion = DefaultViewFormatVersion
+				return nil, fmt.Errorf("%w: %s", iceberg.ErrInvalidFormatVersion, verStr)
```
tanmayrauth, line 529: This wraps ErrInvalidFormatVersion for a non-numeric value, but a numeric-but-invalid view version doesn't get the same treatment: format-version "2" / "0" / "-1" flow through to the builder's SetFormatVersion  (view/metadata_builder.go:322/315), which returns plain strings like "unsupported format version 2" and "downgrading format version from 1 to 0 is not allowed" — neither satisfies errors.Is(err, ErrInvalidFormatVersion). The table path wraps the sentinel for those cases (via NewMetadataBuilder), so the two constructors behave differently. Since the goal here is a uniform ErrInvalidFormatVersion on bad versions, would you want to wrap the sentinel in the view
  builder too (or is that intentionally out of scope for this PR)?

view/metadata.go
```diff
@@ -518,16 +518,20 @@ func NewMetadataWithUUID(version *Version, sc *iceberg.Schema, location string,
 		viewUUID = uuid.New()
 	}
 
+	var inputProps iceberg.Properties
 	formatVersion := DefaultViewFormatVersion
 	if props != nil {
-		verStr, ok := props["format-version"]
+		inputProps = maps.Clone(props)
+		verStr, ok := inputProps["format-version"]
```
tanmayrauth: Nit: the table side keys off the PropertyFormatVersion constant while the view side repeats the "format-version" literal (also at lines 531, 257, 298). Not introduced by this PR, but a package-level const would avoid drift if the  key ever changes.

tanmayrauth (commented):
Core fix looks right, strict validation plus the clone-before-delete to stop mutating the caller's map, and the tests lock that down nicely. One consistency question on the view path and a tiny nit, both inline.

view/metadata.go
```diff
@@ -518,16 +518,24 @@ func NewMetadataWithUUID(version *Version, sc *iceberg.Schema, location string,
...
+				return nil, fmt.Errorf("%w: %s", iceberg.ErrInvalidFormatVersion, verStr)
 			}
-			delete(props, "format-version")
+			delete(inputProps, table.PropertyFormatVersion)
 		}
+	} else {
+		inputProps = props
+	}
+
+	if formatVersion != DefaultViewFormatVersion {
```
laskoviymishka: I'd make this a range check rather than an equality check — `if formatVersion > SupportedViewFormatVersion`. Right now it happens to be equivalent since both constants are 1, but once view V2 lands and they bump, this rejects a valid `"1"` and every other still-supported version below the max. The table path deliberately accepts anything in `[1, supported]` and lets the builder do range validation; I'd mirror that here so the two constructors don't drift. wdyt?

table/metadata.go
```diff
@@ -2205,16 +2205,20 @@ func NewMetadataWithUUID(sc *iceberg.Schema, partitions *iceberg.PartitionSpec,
...
+		verStr, ok := inputProps[PropertyFormatVersion]
 		if ok {
 			if formatVersion, err = strconv.Atoi(verStr); err != nil {
-				formatVersion = DefaultFormatVersion
+				return nil, fmt.Errorf("%w: %s", iceberg.ErrInvalidFormatVersion, verStr)
 			}
-			delete(props, PropertyFormatVersion)
+			delete(inputProps, PropertyFormatVersion)
 		}
+	} else {
```
laskoviymishka: This `else` is a no-op — `inputProps` is already the nil zero value and we only get here when `props == nil`, so it assigns nil to nil. I'd drop the whole branch and just do `inputProps := maps.Clone(props)` up top (`maps.Clone(nil)` returns nil), gating the key lookup on `props != nil`. Same dead `else` over in `view/metadata.go` around line 534.

table/metadata.go
```diff
@@ -2205,16 +2205,20 @@ func NewMetadataWithUUID(sc *iceberg.Schema, partitions *iceberg.PartitionSpec,
...
 	var err error
 	formatVersion := DefaultFormatVersion
 	if props != nil {
-		verStr, ok := props[PropertyFormatVersion]
+		inputProps = maps.Clone(props)
+		verStr, ok := inputProps[PropertyFormatVersion]
 		if ok {
 			if formatVersion, err = strconv.Atoi(verStr); err != nil {
-				formatVersion = DefaultFormatVersion
+				return nil, fmt.Errorf("%w: %s", iceberg.ErrInvalidFormatVersion, verStr)
```
laskoviymishka, line 2216: This rejects non-numeric strings but a numeric-but-out-of-range value like `"0"` or `"99"` still sails past here and only gets caught later by `NewMetadataBuilder`. Both wrap the same sentinel so `errors.Is` is fine, but the error origin and message differ depending on the failure mode. Is that intentional? If we want the constructor to be the single validation point I'd add a range check right after the `Atoi`.

view/metadata.go
```diff
@@ -518,16 +518,24 @@ func NewMetadataWithUUID(version *Version, sc *iceberg.Schema, location string,
...
 	formatVersion := DefaultViewFormatVersion
 	if props != nil {
-		verStr, ok := props["format-version"]
+		inputProps = maps.Clone(props)
+		verStr, ok := inputProps[table.PropertyFormatVersion]
 		if ok {
 			var err error
 			if formatVersion, err = strconv.Atoi(verStr); err != nil {
-				formatVersion = DefaultViewFormatVersion
+				return nil, fmt.Errorf("%w: %s", iceberg.ErrInvalidFormatVersion, verStr)
```
laskoviymishka, line 529: Heads up that the view package already has its own `ErrInvalidViewMetadataFormatVersion`, used by `validateViewMetadata` on the deserialization path. Wrapping `iceberg.ErrInvalidFormatVersion` here means the same package now has two sentinels for the same kind of failure, so anyone doing `errors.Is(err, view.ErrInvalidViewMetadataFormatVersion)` will miss constructor errors. I'd lean toward reusing the existing view sentinel here for consistency — thoughts?

table/metadata_internal_test.go
```diff
@@ -780,6 +780,22 @@ func TestNewMetadataWithExplicitV1Format(t *testing.T) {
 	assert.Truef(t, expected.Equals(actual), "expected: %s\ngot: %s", expected, actual)
 }
 
+func TestNewMetadataRejectInvalidFormatVersion(t *testing.T) {
+	props := iceberg.Properties{
+		"format-version": "banana",
```
laskoviymishka, line 785: This only exercises the non-numeric case. The numeric out-of-range path (`"0"`, `"99"`) goes through a different branch — the builder rather than this constructor — so it's worth a case here too, especially if we tighten the constructor validation. The view test is already table-driven; might be worth matching that shape so both add the extra case in one spot.

view/metadata_test.go
```diff
@@ -88,6 +89,46 @@ func TestNewMetadata(t *testing.T) {
 	assert.Equal(t, []VersionLogEntry{{TimestampMS: 1000, VersionID: 1}}, md.VersionLog())
 }
 
+func TestNewMetadataRejectInvalidFormatVersion(t *testing.T) {
+	tests := []struct {
+		name         string
+		formatVer    string
+		versionValue string
```
laskoviymishka: `versionValue` is always equal to `formatVer` in both cases — I'd drop it and just assert against `tc.formatVer`. Separately, this file mixes the new `table.PropertyFormatVersion` constant with the raw `"format-version"` literal the other ~30 cases use; I'd pick one (raw literal avoids the `table` import entirely) so the file isn't half-and-half.

laskoviymishka (changes requested):
Nice, focused fix — clamping format-version at construction instead of silently defaulting is the right call, and cloning the props map so we stop mutating the caller's input is a good catch while you were in here. The sentinel wrapping means `errors.Is` works for callers.

Almost there. One thing I'd want fixed before merge: the view guard uses `formatVersion != DefaultViewFormatVersion` instead of a `> SupportedViewFormatVersion` range check. It's harmless today because both constants are 1, but it bakes in a forward-incompat — once view V2 ships and both bump, a caller passing `"1"` gets `ErrInvalidFormatVersion` even though V1 is still valid. That's the opposite of the table path, which accepts anything in `[1, supported]` and defers range validation to the builder. One-line change to mirror it.

The rest is minor and can ride along or land as follow-ups: a dead `else` branch in both files, the table path only rejecting non-numeric versions (numeric out-of-range still defers to the builder, so the error origin differs from the view path), a second format-version sentinel now living in the view package alongside `ErrInvalidViewMetadataFormatVersion`, and a couple of test gaps (no numeric out-of-range case on the table side, redundant `versionValue` field / mixed constant-vs-literal style on the view side).

Fix the view guard and this is good to land — happy to take another pass once it's in.

laskoviymishka (approved):
:shipit:

PR #1388, opened 2026-07-04
fix(table): surface transaction initialization errors

fallintoplace:
## What changed
- Added NewTransactionOnBranchE with explicit metadata builder init error return.
- Preserved old NewTransaction/NewTransactionOnBranch API by storing init error on Transaction.
- Propagated initialization errors through transaction methods (including commit, scan, staged table, append/write/delete paths) and RowDelta commit.
- Added regression tests for broken metadata current-schema handling.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -131,20 +131,53 @@ func (t Table) LocationProvider() (LocationProvider, error) {
 }
 
 func (t Table) NewTransaction() *Transaction {
-	return t.NewTransactionOnBranch(MainBranch)
+	txn, err := t.NewTransactionOnBranchWithError(MainBranch)
+	if err != nil {
+		return t.newBrokenTransaction(MainBranch, err)
+	}
+
+	return txn
 }
 
 // NewTransactionOnBranch creates a new transaction that commits to the named
 // branch. Use [NewTransaction] to commit to the default "main" branch.
 func (t Table) NewTransactionOnBranch(branch string) *Transaction {
-	meta, _ := MetadataBuilderFromBase(t.metadata, t.metadataLocation)
+	txn, err := t.NewTransactionOnBranchWithError(branch)
+	if err != nil {
+		return t.newBrokenTransaction(branch, err)
+	}
+
+	return txn
+}
+
+func (t Table) newBrokenTransaction(branch string, err error) *Transaction {
+	return &Transaction{
+		tbl:     &t,
+		initErr: err,
+		branch:  branch,
+		reqs:    []Requirement{},
+	}
+}
+
+// NewTransactionOnBranchWithError creates a new transaction and returns any metadata
+// initialization error that prevents builder construction.
+//
+// This preserves the old non-failing constructor contract while allowing
+// callers to receive the precise initialization error instead of hitting
+// panic/undefined behavior later.
+func (t Table) NewTransactionOnBranchWithError(branch string) (*Transaction, error) {
+	meta, err := MetadataBuilderFromBase(t.metadata, t.metadataLocation)
+	if err != nil {
+		return nil, err
+	}
 
 	return &Transaction{
-		tbl:    &t,
-		meta:   meta,
-		branch: branch,
-		reqs:   []Requirement{},
-	}
+		tbl:     &t,
+		meta:    meta,
+		branch:  branch,
+		reqs:    []Requirement{},
+		initErr: nil,
+	}, nil
 }
 
 func (t *Table) Refresh(ctx context.Context) error {
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -75,9 +75,10 @@ func (s snapshotUpdate) mergeAppend() *snapshotProducer {
 }
 
 type Transaction struct {
-	tbl    *Table
-	meta   *MetadataBuilder
-	branch string
+	tbl     *Table
+	meta    *MetadataBuilder
+	branch  string
+	initErr error
 
 	reqs []Requirement
 
@@ -94,9 +95,24 @@ type Transaction struct {
 	committed bool
 }
 
+func (t *Transaction) ensureInitialized() error {
+	if t.initErr != nil {
+		return t.initErr
+	}
+	if t.meta == nil {
+		return fmt.Errorf("%w: transaction metadata builder is not initialized", ErrInvalidMetadata)
+	}
+
+	return nil
+}
+
 func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 	t.mx.Lock()
 	defer t.mx.Unlock()
+
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
 
 	if t.committed {
 		return errors.New("transaction has already been committed")
@@ -223,6 +239,9 @@ func (t *Transaction) updateSnapshot(wfs io.WriteFileIO, props iceberg.Propertie
 }
 
 func (t *Transaction) SetProperties(props iceberg.Properties) error {
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
 	if len(props) > 0 {
 		return t.apply([]Update{NewSetPropertiesUpdate(props)}, nil)
 	}
@@ -233,10 +252,17 @@ func (t *Transaction) SetProperties(props iceberg.Properties) error {
 // UpgradeFormatVersion upgrades the table to the given format version. Downgrading
 // is not allowed. If the table is already at the given version, this is a no-op.
 func (t *Transaction) UpgradeFormatVersion(version int) error {
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
+
 	return t.apply([]Update{NewUpgradeFormatVersionUpdate(version)}, nil)
 }
 
 func (t *Transaction) RollbackToSnapshot(snapshotID int64) error {
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
 	cs := t.meta.currentSnapshot()
 	if cs == nil {
 		return errors.New("cannot rollback: table has no current snapshot")
@@ -346,6 +372,9 @@ func WithPostCommit(postCommit bool) ExpireSnapshotsOpt {
 //
 // The "iceberg expire-snapshots" CLI command wraps the same operation.
 func (t *Transaction) ExpireSnapshots(opts ...ExpireSnapshotsOpt) error {
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
 	var (
 		cfg         = expireSnapshotsCfg{postCommit: true}
 		updates     []Update
@@ -449,6 +478,9 @@ func (t *Transaction) ExpireSnapshots(opts ...ExpireSnapshotsOpt) error {
 }
 
 func (t *Transaction) AppendTable(ctx context.Context, tbl arrow.Table, batchSize int64, snapshotProps iceberg.Properties) error {
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
 	rdr := array.NewTableReader(tbl, batchSize)
 	defer rdr.Release()
 
@@ -456,6 +488,9 @@ func (t *Transaction) AppendTable(ctx context.Context, tbl arrow.Table, batchSiz
 }
 
 func (t *Transaction) Append(ctx context.Context, rdr array.RecordReader, snapshotProps iceberg.Properties) error {
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
 	fs, err := t.tbl.fsF(ctx)
 	if err != nil {
 		return err
@@ -496,6 +531,9 @@ func (t *Transaction) Append(ctx context.Context, rdr array.RecordReader, snapsh
 //
 // For now, we'll keep using an overwrite operation.
 func (t *Transaction) ReplaceDataFiles(ctx context.Context, filesToDelete, filesToAdd []string, snapshotProps iceberg.Properties) error {
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
 	if len(filesToDelete) == 0 {
 		if len(filesToAdd) > 0 {
 			return t.AddFiles(ctx, filesToAdd, snapshotProps, false)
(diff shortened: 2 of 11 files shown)


tanmayrauth (approved):
LGTM, thanks for the PR.

table/transaction.go
```diff
@@ -206,10 +225,17 @@ func (t *Transaction) SetProperties(props iceberg.Properties) error {
...
 func (t *Transaction) UpgradeFormatVersion(version int) error {
+	if err := t.ensureInitialized(); err != nil {
+		return err
+	}
+
 	return t.apply([]Update{NewUpgradeFormatVersionUpdate(version)}, nil)
 }
 
 func (t *Transaction) RollbackToSnapshot(snapshotID int64) error {
+	if err := t.ensureInitialized(); err != nil {
```
laskoviymishka, line 266: `UpdateSchema` just below here (around line 252) never got this guard, and it's the one method where the omission still panics instead of deferring — `NewUpdateSchema` derefs `txn.meta.LastColumnID()` at construction before anything checks `meta`. So on a broken txn, `txn.UpdateSchema()` panics, which is precisely the case this PR is trying to eliminate.

Adding `ensureInitialized()` at the top of `Transaction.UpdateSchema` fixes it (or pass the last-column-id by value the way `UpdateSpec` already does). Either way I'd want a test that calls `UpdateSchema` on a broken txn.

table/transaction.go
```diff
@@ -93,10 +94,25 @@ type Transaction struct {
...
+	}
+
+	return nil
+}
+
 func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 	t.mx.Lock()
 	defer t.mx.Unlock()
 
+	if err := t.ensureInitialized(); err != nil {
```
laskoviymishka, line 116: `apply()` funnels most mutations through this one guard, which is nice — but every entry point that doesn't route through `apply` needs its own copy, and it's easy to miss one. `UpdateSpec` surfaces the error late (at `Commit` rather than at the call), and `TableCommit`, `WriteEqualityDeletes`, and `RewriteFiles` aren't guarded at all.

Rather than hand-copying the guard into ~20 methods with no compile-time enforcement, could we expose a `txnMeta() (*MetadataBuilder, error)` accessor and have methods pull `meta` through it? Then a new method physically can't touch `meta` without going through the check. wdyt?

table/table.go
```diff
@@ -131,20 +131,43 @@ func (t Table) LocationProvider() (LocationProvider, error) {
...
+// NewTransactionOnBranchE creates a new transaction and returns any metadata
+// initialization error that prevents builder construction.
+//
+// This preserves the old non-failing constructor contract while allowing
+// callers to receive the precise initialization error instead of hitting
+// panic/undefined behavior later.
+func (t Table) NewTransactionOnBranchE(branch string) (*Transaction, error) {
+	meta, err := MetadataBuilderFromBase(t.metadata, t.metadataLocation)
+	if err != nil {
+		return &Transaction{
```
laskoviymishka: Returning a non-nil `*Transaction` alongside a non-nil error is a surprising contract in Go — callers who check `err` first will drop the txn, and callers who don't will keep a poisoned object.

I'd return `nil, err` here and let the legacy `NewTransaction` / `NewTransactionOnBranch` wrappers build the poisoned-but-usable txn themselves. That keeps the `(T, error)` signature idiomatic and confines the swallow to the back-compat path.

table/table.go
```diff
@@ -131,20 +131,43 @@ func (t Table) LocationProvider() (LocationProvider, error) {
...
+	return txn
+}
+
+// NewTransactionOnBranchE creates a new transaction and returns any metadata
+// initialization error that prevents builder construction.
+//
+// This preserves the old non-failing constructor contract while allowing
+// callers to receive the precise initialization error instead of hitting
+// panic/undefined behavior later.
+func (t Table) NewTransactionOnBranchE(branch string) (*Transaction, error) {
```
laskoviymishka: The `...E` suffix isn't a convention we use anywhere else in the codebase, so the correct constructor ends up hidden behind an unfamiliar name while `NewTransaction`/`NewTransactionOnBranch` quietly hand back a broken object.

I'd either spell it out (`NewTransactionOnBranchWithError`) or, better, make the `(T, error)` form the canonical constructor and deprecate the swallowing ones. Not blocking, but worth settling before this lands as public API.

table/transaction.go
```diff
@@ -93,10 +94,25 @@ type Transaction struct {
 	committed bool
 }
 
+func (t *Transaction) ensureInitialized() error {
+	if t.initErr != nil {
+		return t.initErr
+	}
+	if t.meta == nil {
```
laskoviymishka, line 105: The `t.meta == nil` branch looks unreachable in practice — `initErr` is set whenever `meta` is nil, so the first check always returns first.

If you're keeping it as a defensive net, I'd wrap it as `fmt.Errorf("%w: ...", ErrInvalidMetadata)` so it stays consistent with the `initErr` path — a bare `errors.New` here wouldn't match `errors.Is(err, ErrInvalidMetadata)` if the branch ever were reached.

table/transaction_internal_test.go
```diff
@@ -60,6 +62,59 @@ func TestTransactionApplyDedupesEquivalentRequirementsWithinAndAcrossCalls(t *te
...
+	baseMeta, err := NewMetadata(simpleSchema(), iceberg.UnpartitionedSpec, UnsortedSortOrder, "table-location", nil)
+	require.NoError(t, err, "new metadata")
+
+	txn := New(Identifier{"db", "broken"}, brokenMetadata{
+		Metadata: baseMeta,
+	}, "metadata.json", func(context.Context) (iceio.IO, error) {
+		return nil, nil
+	}, nil).NewTransaction()
+	err = txn.SetProperties(iceberg.Properties{"k": "v"})
+	require.ErrorContains(t, err, "current schema is missing")
```
laskoviymishka: This is the exact test that would've caught the `UpdateSchema` panic if it exercised it — I'd add a `txn.UpdateSchema(...)` case here on the broken txn.

Two smaller things: this func drives both `SetProperties` and `RowDelta.Commit`, so I'd split it into `t.Run` blocks; and `brokenMetadata` overriding `CurrentSchema()` to nil is a bit synthetic. A round-trip through `ParseMetadataBytes` with a missing current-schema-id would exercise the real deserialization failure this PR cares about.

laskoviymishka (changes requested):
Almost there — the `initErr` + `ensureInitialized()` approach is the right way to turn a broken transaction from a later panic into an error, and I like that the test drives a nil `CurrentSchema()` all the way to a `Commit` error.

One blocking thing before merge: `UpdateSchema` still panics on a broken transaction. It never got the guard, and unlike the other methods the missing guard here doesn't defer an error — `NewUpdateSchema` derefs `txn.meta` at construction, so `txn.UpdateSchema()` panics, which is the exact case this PR is meant to fix. Guarding `Transaction.UpdateSchema` (or passing last-column-id by value like `UpdateSpec` does) closes it.

The related thing I'd think about is that this guard is hand-copied into ~20 methods with no compile-time enforcement, which is how `UpdateSchema`, `UpdateSpec`, `TableCommit`, and `WriteEqualityDeletes` got missed. A single `txnMeta() (*MetadataBuilder, error)` accessor would make it structurally impossible to reach `meta` without the check — worth considering rather than auditing the list by hand.

- guard `UpdateSchema` (plus `UpdateSpec` / `TableCommit` / `WriteEqualityDeletes`), with a test that calls `UpdateSchema` on a broken txn
- `NewTransactionOnBranchE` should return `nil, err` rather than a non-nil txn alongside a non-nil error
- settle the `...E` naming before it becomes public API

table/update_spec.go
```diff
@@ -54,6 +55,26 @@ type transformKey struct {
...
+		transformToField:      make(map[transformKey]iceberg.PartitionField),
+		transformToAddedField: make(map[transformKey]iceberg.PartitionField),
+		renames:               make(map[string]string),
+		addedTimeFields:       make(map[int]iceberg.PartitionField),
+		caseSensitive:         caseSensitive,
+		adds:                  make([]iceberg.PartitionField, 0),
+		deletes:               make(map[int]bool),
+	}
+
+	if t != nil {
```
laskoviymishka: the guard here only fires when `t` is non-nil — if `t` (or `t.tbl`) is nil, `us.err` stays nil, we skip the `return us`, and fall straight through to `t.tbl.Metadata()` on line 80 and panic. `NewUpdateSchema` has the same shape a few lines down at `txn.meta.LastColumnID()`. it's the exact bug this PR is closing everywhere else.

to be fair, a nil `*Transaction` isn't constructable through the public API — these are always called off a live transaction — so this is defense-in-depth, not a live crash. but since we're already paying for the guard I'd set an explicit error in the nil branch instead of skipping it:

```go
if t == nil {
    us.err = errors.New("transaction is nil")
    return us
}
us.err = t.ensureInitialized()
```

wdyt?

table/table.go
```diff
@@ -131,22 +131,55 @@ func (t Table) LocationProvider() (LocationProvider, error) {
...
 	return &Transaction{
-		tbl:    &t,
-		meta:   meta,
-		branch: branch,
-		reqs:   []Requirement{},
+		tbl:     &t,
+		initErr: err,
+		branch:  branch,
+		reqs:    []Requirement{},
 	}
```
laskoviymishka, line 159: not a blocker for me, but this one still feels off: there's no other `...E`-suffixed constructor anywhere in the codebase, and the Go idiom would usually make the error-returning form the primary one (or add a `Must`-style wrapper going the other way) rather than an `E` variant. the closest analogues we already have are the swallow-error `NewTransaction` / `NewTransactionOnBranch` right above — so whatever we pick here sets the precedent for how that pair evolves too.

since it's public surface we can't rename later without a break, I'd rather not decide it unilaterally. @zeroshade — do you have a preference on the naming convention here? happy to go with whatever you'd bless.

table/transaction.go
```diff
@@ -93,10 +94,25 @@ type Transaction struct {
 	committed bool
 }
 
+func (t *Transaction) ensureInitialized() error {
```
laskoviymishka, line 98: not for this PR, but flagging it while we're here: this leaves ~20 entry points each responsible for remembering to call `ensureInitialized()`, with nothing enforcing it — the round-1 miss on `UpdateSchema`/`UpdateSpec` was exactly this failure mode.

longer term I'd rather funnel meta access through a single accessor (something like `txnMeta() (*MetadataBuilder, error)`) so a new method physically can't forget the guard. non-blocking and probably its own follow-up — just don't want us to lose the thread.

laskoviymishka (changes requested):
almost there — everything I asked for last round is in. the `ensureInitialized()` guard is on `UpdateSchema`, `UpdateSpec`, `TableCommit`, and `WriteEqualityDeletes` now, `NewTransactionOnBranchE` returns `(nil, err)` on failure, and the `require.NotPanics` test pins the exact case that used to panic. nice tight pass.

one thing I'd fix before merge: the guard in `NewUpdateSpec`/`NewUpdateSchema` only runs when the transaction is non-nil — a nil one skips it and falls through to a deref, which is the same bug class we're fixing. not reachable from the public API today, so it's defense-in-depth, but cheap to close properly while we're in here (suggestion inline).

the `...E` constructor name isn't a blocker for me, but it still feels off and it's public surface we can't rename later — want here @zeroshade in for a second opinion rather than holding on it.

the guard-by-convention point from last round still stands (nothing stops a new method from forgetting `ensureInitialized()`), but I think that's a follow-up rather than something to hold this on.

fix the nil-txn fall-through and I'm happy to approve — the naming can settle in parallel.

table/update_spec.go
```diff
@@ -54,6 +55,29 @@ type transformKey struct {
...
+		transformToAddedField: make(map[transformKey]iceberg.PartitionField),
+		renames:               make(map[string]string),
+		addedTimeFields:       make(map[int]iceberg.PartitionField),
+		caseSensitive:         caseSensitive,
+		adds:                  make([]iceberg.PartitionField, 0),
+		deletes:               make(map[int]bool),
+	}
+
+	if t == nil {
+		us.err = errors.New("transaction is nil")
```
laskoviymishka: The one thing I'd tidy from the round-2 fix: every other init error wraps `ErrInvalidMetadata` (via `ensureInitialized` and `MetadataBuilderFromBase`), but the nil-txn case here — and the mirror in `NewUpdateSchema` — uses a bare `errors.New`, so `errors.Is(err, ErrInvalidMetadata)` comes back false only for this path.

I'd wrap it: `fmt.Errorf("%w: transaction is nil", ErrInvalidMetadata)`. If you do, the two nil-txn tests should switch from `assert.EqualError` to `require.ErrorIs(err, ErrInvalidMetadata)` so they aren't pinned to the exact string. Non-blocking, but it closes the last inconsistency in the guard set.

table/transaction.go
```diff
@@ -94,10 +95,25 @@ type Transaction struct {
...
+	if t.initErr != nil {
```
laskoviymishka, line 102: Defense-in-depth thought, not blocking: this reads `t.initErr` on the receiver with no `t == nil` check, so it'd panic rather than error on a nil `*Transaction`. Can't happen today — every constructor returns a non-nil txn, so `rd.txn.ensureInitialized()` and friends are always called on a live pointer. But since the whole point of the PR is turning panics into errors, a one-line `if t == nil { return fmt.Errorf("%w: transaction is nil", ErrInvalidMetadata) }` at the top would make the guard robust if a future caller ever hands it a nil. wdyt?

table/transaction.go
```diff
@@ -2156,6 +2224,10 @@ func (t *Transaction) Commit(ctx context.Context) (*Table, error) {
 	t.mx.Lock()
 	defer t.mx.Unlock()
 
+	if err := t.ensureInitialized(); err != nil {
```
laskoviymishka, line 2230: The suite covers `TableCommit()` but not `Commit(ctx)`, which is the terminal path most likely to be hit on a broken txn. I'd add one subtest asserting `errors.Is(err, ErrInvalidMetadata)` on `txn.Commit(ctx)` alongside the existing ones.

laskoviymishka (approved):
LGTM — the nil-txn fall-through I flagged last round is fixed (`NewUpdateSpec`/`NewUpdateSchema` check `if t == nil` first now, and the two new tests pin it), and the `WithError` rename reads well.

Two small non-blockers I left inline: the nil-txn guards return a bare `errors.New` instead of wrapping `ErrInvalidMetadata` like the rest, so `errors.Is` misses that one case, and `Commit(ctx)` isn't in the broken-txn test sweep. The guard-by-convention point (~20 hand-copied guards) still stands as the follow-up we agreed on, not something for this PR.

On naming: `WithError` is fine by me — still worth @zeroshade's take before we call it settled public API, but not blocking. Approving.

PR #1389, opened 2026-07-04
fix(hadoop): return invalid identifier errors from CheckTableExists

fallintoplace:
## What changed
- `Hadoop` catalog `CheckTableExists` now returns the identifier validation error instead of swallowing malformed identifiers as `(false, nil)`.
- Preserves missing-table behavior for valid identifiers while surfacing malformed identifier inputs as errors.
- Updated `catalog/hadoop` tests to assert short and invalid identifiers return `catalog.ErrNoSuchTable`.

## Testing
- `go test ./catalog/hadoop -count=1`

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -133,33 +133,33 @@ func shouldSuppressPermissionError(err error) bool {
 // Note: this is POSIX-best-effort validation — it does not catch NUL bytes
 // or Windows reserved names (NUL, CON, COM1, etc.).
 func validateIdentifier(ident table.Identifier) error {
-	return validateIdentifierParts(ident, catalog.ErrNoSuchNamespace, "namespace identifier", "identifier component")
+	return validateIdentifierParts(ident, "namespace identifier", "identifier component")
 }
 
 func validateTableIdentifier(ident table.Identifier) error {
 	if len(ident) < 2 {
-		return fmt.Errorf("%w: table identifier must have at least a namespace and table name", catalog.ErrNoSuchTable)
+		return fmt.Errorf("%w: table identifier must have at least a namespace and table name", catalog.ErrInvalidIdentifier)
 	}
 
-	return validateIdentifierParts(ident, catalog.ErrNoSuchTable, "table identifier", "table identifier component")
+	return validateIdentifierParts(ident, "table identifier", "table identifier component")
 }
 
-func validateIdentifierParts(ident table.Identifier, errType error, identifierName, componentName string) error {
+func validateIdentifierParts(ident table.Identifier, identifierName, componentName string) error {
 	if len(ident) == 0 {
-		return fmt.Errorf("%w: %s must not be empty", errType, identifierName)
+		return fmt.Errorf("%w: %s must not be empty", catalog.ErrInvalidIdentifier, identifierName)
 	}
 
 	for _, part := range ident {
 		if part == "" {
-			return fmt.Errorf("%w: %s must not be empty", errType, componentName)
+			return fmt.Errorf("%w: %s must not be empty", catalog.ErrInvalidIdentifier, componentName)
 		}
 
 		if part == "." || part == ".." {
-			return fmt.Errorf("%w: invalid %s %q", errType, componentName, part)
+			return fmt.Errorf("%w: invalid %s %q", catalog.ErrInvalidIdentifier, componentName, part)
 		}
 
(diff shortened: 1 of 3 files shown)


tanmayrauth (commented):
LGTM, this lines hadoop up with glue/hive/rest/sql, which all return the identifier error from CheckTableExists rather than swallowing it.

catalog/hadoop/hadoop.go
```diff
@@ -584,7 +584,7 @@ func (c *Catalog) LoadTable(ctx context.Context, ident table.Identifier) (*table
 
 func (c *Catalog) CheckTableExists(_ context.Context, ident table.Identifier) (bool, error) {
 	if err := validateTableIdentifier(ident); err != nil {
-		return false, nil
+		return false, err
```
laskoviymishka, line 587: Returning the error instead of a silent `false` is the right move, but `validateTableIdentifier` wraps `catalog.ErrNoSuchTable`, so `errors.Is(err, ErrNoSuchTable)` now returns true for a malformed identifier like `["tbl"]`. That sentinel means "catalog queried, table absent" — a caller doing `if errors.Is(err, ErrNoSuchTable) { createTable(...) }` would treat garbage input as "not found, go ahead and create".

I'd introduce a distinct `catalog.ErrInvalidIdentifier` and wrap that from the validation helpers instead — which also lines Hadoop up with Glue/Hive/REST, none of which wrap `ErrNoSuchTable` for invalid identifiers. Worth checking `CheckNamespaceExists`/`validateIdentifierParts` too, since they have the symmetric issue with `ErrNoSuchNamespace`. wdyt?

catalog/hadoop/hadoop_test.go
```diff
@@ -1900,7 +1901,8 @@ func (s *HadoopCatalogTestSuite) TestCheckTableExistsInvalidIdentifierFalse() {
 	for _, tt := range tests {
 		s.Run(tt.name, func() {
 			exists, err := s.cat.CheckTableExists(context.Background(), tt.ident)
-			s.Require().NoError(err)
+			s.Require().Error(err)
+			s.ErrorIs(err, catalog.ErrNoSuchTable)
 			s.False(exists)
```
laskoviymishka, line 1924: This asserts the exact sentinel I flagged in `CheckTableExists`, so it encodes the defect into the suite — if we switch to `ErrInvalidIdentifier`, both this and the `ShortIdentifier` case have to be rewritten. I'd hold the test change until the sentinel is settled, then assert on the new one. It'd also be worth one case proving an invalid identifier is distinguishable from a genuine not-found, which isn't possible while both wrap the same error.

catalog/hadoop/hadoop_test.go
```diff
@@ -1876,7 +1876,8 @@ func (s *HadoopCatalogTestSuite) TestCheckTableExistsPropagatesMetadataReadError
 
 func (s *HadoopCatalogTestSuite) TestCheckTableExistsShortIdentifier() {
 	exists, err := s.cat.CheckTableExists(context.Background(), []string{"tbl"})
-	s.Require().NoError(err)
+	s.Require().Error(err)
+	s.ErrorIs(err, catalog.ErrNoSuchTable)
```
laskoviymishka: Minor: `s.ErrorIs` at assert level won't halt the test, so a non-nil-but-wrong-type error gives you two failure lines. `s.Require().ErrorIs(err, ...)` folds the `Require().Error` and the type check into one and stops cleanly on mismatch.

laskoviymishka (changes requested):
Almost there — the direction is right. Returning `(false, nil)` for a malformed identifier was a genuine bug, since it reported "doesn't exist" for input we never actually checked against the catalog.

One blocking thing before merge: `validateTableIdentifier` wraps `catalog.ErrNoSuchTable`, so this now makes `errors.Is(err, ErrNoSuchTable)` true for a malformed identifier. That sentinel means "queried, table absent", not "identifier is invalid" — a caller doing `if errors.Is(err, ErrNoSuchTable) { createTable(...) }` would create a table on garbage input. It's also inconsistent with Glue/Hive/REST, which return plain errors here. I'd add a distinct `catalog.ErrInvalidIdentifier`, wrap it from the validation helpers (worth doing the `CheckNamespaceExists` side at the same time), and assert on that in the tests — otherwise the current test changes bake in the wrong sentinel and get rewritten anyway.

Once the sentinel is sorted, happy to take another pass and approve.

laskoviymishka (approved):
:shipit:

PR #1390, opened 2026-07-04
fix(table): clone selected fields in scan options

fallintoplace:
## Summary
- fix(table): clone selected fields inside WithSelectedFields
- add regression test that mutating caller slice after option creation does not affect scan.selectedFields

## Testing
- go test ./table -run TestWithSelectedFields -count=1
- go test ./table -count=1

### DIFF
--- a/cmd/iceberg/output.go
+++ b/cmd/iceberg/output.go
@@ -59,6 +59,14 @@ type Output interface {
 
 type textOutput struct{}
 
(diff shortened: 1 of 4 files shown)


tanmayrauth (approved):
LGTM, thanks for the PR.

laskoviymishka (approved):

PR #1392, opened 2026-07-04
docs(table): clarify mixed-group lineage-drop behavior

fallintoplace:
## Summary
- Clarify the current mixed-group row-lineage fallback in compaction on v3 tables.
- Document that mixed groups can appear during a v1/v2->v3 migration because `first_row_id` is assigned lazily at the next v3 manifest-list write.
- Keep the existing conservative behavior: when a mixed group is compacted, `_row_id` is dropped for the merged output.
- Restore the warning text and helper docs so they describe the current behavior accurately.

## Testing
- `go test ./table -run 'TestExecuteCompactionGroupPreservesRowIDAcrossV2Upgrade|TestExecuteCompactionGroupPreservesRowID|TestCoWRewritePreservesRowID|TestCoWRewriteRowIDNextRowIDAccounting' -count=1`

### DIFF
--- a/table/rewrite_data_files.go
+++ b/table/rewrite_data_files.go
@@ -329,7 +329,7 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 			}
 
 			safePosDeletes := mergeTaskFiles(preserve.SafePosDeletes, legacy.SafePosDeletes)
-			safeDeletionVectors := mergeTaskFiles(preserve.SafeDeletionVectors, legacy.SafeDeletionVectors)
+			safeDeletionVectors := mergeTaskFilesByReferencedDataFile(preserve.SafeDeletionVectors, legacy.SafeDeletionVectors)
 
 			return CompactionGroupResult{
 				PartitionKey:        group.PartitionKey,
@@ -343,7 +343,7 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 		}
 	}
 
-	// Preserve row lineage only when every source file in the group carries
+	// Preserve row lineage only when every source file in the group carries row-lineage metadata.
 	preserveLineage := tbl.metadata.Version() >= 3 && allTasksHaveRowLineage(group.Tasks)
 
 	return executeCompactionTaskGroup(ctx, tbl, scanOpts, group, cfg.targetFileSize, preserveLineage)
@@ -458,6 +458,42 @@ func mergeTaskFiles(a, b []iceberg.DataFile) []iceberg.DataFile {
 	}
 
 	return out
+}
+
+func mergeTaskFilesByReferencedDataFile(a, b []iceberg.DataFile) []iceberg.DataFile {
+	if len(a) == 0 {
+		return slices.Clone(b)
+	}
+	if len(b) == 0 {
+		return slices.Clone(a)
+	}
+
+	seen := make(map[string]struct{}, len(a)+len(b))
+	for _, df := range a {
+		ref := dvReferencedDataFile(df)
+		seen[ref] = struct{}{}
+	}
+
+	out := make([]iceberg.DataFile, 0, len(a)+len(b))
+	out = append(out, a...)
+	for _, df := range b {
+		ref := dvReferencedDataFile(df)
+		if _, ok := seen[ref]; ok {
+			continue
+		}
+		seen[ref] = struct{}{}
+		out = append(out, df)
+	}
+
+	return out
+}
+
+func dvReferencedDataFile(df iceberg.DataFile) string {
+	if ref := df.ReferencedDataFile(); ref != nil && *ref != "" {
+		return *ref
+	}
+
+	return df.FilePath()
 }
 
 // allTasksHaveRowLineage returns true iff every task in the group has a
--- a/table/rewrite_data_files.go
+++ b/table/rewrite_data_files.go
@@ -300,38 +300,60 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 		scanOpts = append(scanOpts, WitMaxConcurrency(cfg.scanConcurrency))
 	}
 
-	// Preserve row lineage only when every source file in the group carries
-	// it. A mixed group (some files with FirstRowID, some without — e.g.
-	// legacy files on a v3 table) would otherwise produce one output where
-	// post-lineage rows have explicit _row_id values and pre-lineage rows
-	// have nulls, which violates the per-file uniqueness/coverage
-	// invariant the v3 spec requires. Splitting mixed groups into separate
-	// outputs is a larger refactor and is left as a follow-up; for now we
-	// degrade gracefully (the rewrite still succeeds, but lineage is not
-	// preserved for the surviving rows).
-	preserveLineage := tbl.metadata.Version() >= 3 && allTasksHaveRowLineage(group.Tasks)
-	if preserveLineage {
-		scanOpts = append(scanOpts, WithRowLineage())
-	} else if tbl.metadata.Version() >= 3 {
-		// Mixed group on a v3 table — at least one source file lacks
-		// FirstRowID so we drop lineage on the surviving rows. This is the
-		// common case during a v1/v2→v3 migration; surface it so operators
-		// can detect silent lineage loss instead of having to diff
-		// metadata before/after.
-		var lineageFiles, legacyFiles int
-		for _, t := range group.Tasks {
-			if t.FirstRowID != nil {
-				lineageFiles++
-			} else {
-				legacyFiles++
-			}
-		}
-		if lineageFiles > 0 {
-			slog.Warn("compaction group has mixed row lineage; dropping _row_id on output",
+	if tbl.metadata.Version() >= 3 {
+		lineageTasks, legacyTasks := splitTasksByLineage(group.Tasks)
+		if len(lineageTasks) > 0 && len(legacyTasks) > 0 {
+			slog.Info("compaction group has mixed row-lineage tasks; splitting to preserve _row_id for lineage-capable files",
 				"partition_key", group.PartitionKey,
-				"lineage_files", lineageFiles,
-				"legacy_files", legacyFiles)
+				"lineage_tasks", len(lineageTasks),
+				"legacy_tasks", len(legacyTasks))
+
+			lineageGroup := CompactionTaskGroup{
+				PartitionKey:   group.PartitionKey,
+				Tasks:          lineageTasks,
+				TotalSizeBytes: sumTaskBytes(lineageTasks),
+			}
+			legacyGroup := CompactionTaskGroup{
+				PartitionKey:   group.PartitionKey,
+				Tasks:          legacyTasks,
+				TotalSizeBytes: sumTaskBytes(legacyTasks),
+			}
+
+			preserve, err := executeCompactionTaskGroup(ctx, tbl, scanOpts, lineageGroup, cfg.targetFileSize, true)
+			if err != nil {
+				return CompactionGroupResult{}, err
+			}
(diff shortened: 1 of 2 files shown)


table/rewrite_data_files.go
```diff
@@ -300,38 +300,61 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
...
+			if err != nil {
+				return CompactionGroupResult{}, err
+			}
+			legacy, err := executeCompactionTaskGroup(ctx, tbl, scanOpts, legacyGroup, cfg.targetFileSize, false)
+			if err != nil {
+				return CompactionGroupResult{}, err
+			}
+
+			safePosDeletes := mergeTaskFiles(preserve.SafePosDeletes, legacy.SafePosDeletes)
+			safeDeletionVectors := mergeTaskFiles(preserve.SafeDeletionVectors, legacy.SafeDeletionVectors)
```
tanmayrauth: `mergeTaskFiles` dedups by `df.FilePath()`, which is the Puffin path — but a single Puffin file can hold DV blobs for multiple data files, so DVs have to be deduped by `ReferencedDataFile()`, not path. That's the invariant the rest of the codebase already relies on (see `readAllDeletionVectors` keying by referenced data file, and `deleteFileRemoved`/`removeDeletionVector` in snapshot_producers.go using `deletedDVsByRef`).
  
  Since the split sends lineage-capable and legacy data files into different subgroups, a shared Puffin holding one DV blob for a lineage file and one for a legacy file produces two entries with the same `FilePath()` but different `ReferencedDataFile()`. This merge collapses them to one, so the dropped DV never reaches `ApplyResult` — its data file gets rewritten while its DV is left orphaned (stale delete referencing a removed file). This bites on tables  whose DVs were written by Java/Spark, which bundle blobs. 
  
  Pos-deletes are fine to dedup by path, so I'd suggest a separate DV merge keyed by referenced data file (mirroring `CollectSafeDeletionVectors`) rather than reusing the path-based helper for both. Could you also add a mixed-group case where a single DV Puffin references one lineage and one legacy data file, and assert both DVs are expunged?

table/rewrite_data_files.go
```diff
@@ -300,38 +300,61 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
...
+				NewDataFiles:        append(preserve.NewDataFiles, legacy.NewDataFiles...),
+				SafePosDeletes:      safePosDeletes,
+				SafeDeletionVectors: safeDeletionVectors,
+				BytesBefore:         preserve.BytesBefore + legacy.BytesBefore,
+				BytesAfter:          preserve.BytesAfter + legacy.BytesAfter,
+			}, nil
+		}
+	}
+
 	// Preserve row lineage only when every source file in the group carries
```
tanmayrauth, line 303: This comment lost its tail when the old degradation block was removed — it now ends mid-sentence at "carries". Worth completing or trimming it.

tanmayrauth (commented):
Nice PR, splitting mixed groups is the right fix, and the scanOpts clone is a good catch. One real correctness issue on how the deletion-vector outputs get merged, plus a small comment nit. Details inline.

twuebi:
Thanks for picking up the follow-up from the code comment! I don't think the mixed state is reachable through spec-compliant metadata, though.

The v3 spec ([Row Lineage](https://iceberg.apache.org/spec/#row-lineage)) requires that every data file on a committed v3 snapshot has an effective `first_row_id`:

> Any null (unassigned) `first_row_id` must be assigned via inheritance, even if the data file is existing. This ensures that row IDs are assigned to existing data files in upgraded tables in the first commit after upgrading to v3.

This repo implements that on both the write side (the v3 manifest-list writer assigns `first_row_id` to any data manifest lacking one) and the read side (inheritance during manifest reading), so `PlanFiles` always yields a non-nil `FirstRowID` on v3 — which is also why the new tests have to hand-mutate `tasks[1].FirstRowID = nil` to create the scenario. A mixed group would take a spec-violating writer, and for that the current conservative gate + warning feels like the safer behavior. Splitting also has a real cost: one group becomes two output streams, so we can end up with two undersized files instead of one target-size file.

So my suggestion would be to instead update the stale "left as a follow-up" comment to note the mixed case is unreachable for spec-compliant metadata. Happy to be corrected if there's a real commit sequence that produces a mixed group!

fallintoplace:
I did find some ways nil lineage can leak into compaction:

- malformed/non-compliant v3 manifest lists could produce mixed tasks
- future/custom remote scan planning could return tasks without FirstRowID
- distributed/custom task serialization could drop FirstRowID
- a stale plan made immediately after a metadata-only v2→v3 upgrade can have all-nil tasks, then a concurrent first v3 commit can assign row IDs to those same source files before the rewrite commits

twuebi:
Thanks for bringing these up, going over them one by one, though, I still don't see a case where the split path helps:

- **Malformed / non-compliant manifest lists**, if the metadata is already spec-violating, preserving lineage derived from one half of a broken manifest list seems riskier than the current conservative warn-and-drop. Garbage in shouldn't become half-trusted lineage out.
- **Custom remote scan planning / custom task serialization**, those are callers constructing invalid `FileScanTask`s. Not sure if the library should grow a compaction-splitting path to compensate for that. The in-repo codec does round-trip `FirstRowID` (`codec/file_scan_task.go`), so task serialization within this repo isn't affected.
- **Stale plan across a v2→v3 upgrade race**, don't *all* tasks have nil `FirstRowID` in this scenario, so the group is homogeneous legacy and never enters the new split path?

So the mixed case remains reachable only via broken metadata or caller bugs, and for both the existing gate + warning feels like the right response, especially given the split's cost of turning one compaction group into two output streams.

fallintoplace:
I’ll rework this to drop the split behavior and instead update the comment around the existing conservative gate. I’m also considering whether mixed v3 tasks should fail closed rather than warn-and-drop, since that state seems to indicate invalid input rather than a normal migration case.

table/rewrite_data_files.go
```diff
@@ -301,23 +301,17 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
...
-	// post-lineage rows have explicit _row_id values and pre-lineage rows
-	// have nulls, which violates the per-file uniqueness/coverage
-	// invariant the v3 spec requires. Splitting mixed groups into separate
-	// outputs is a larger refactor and is left as a follow-up; for now we
-	// degrade gracefully (the rewrite still succeeds, but lineage is not
-	// preserved for the surviving rows).
+	// it. A mixed group (some files with FirstRowID, some without) would
+	// otherwise produce one output where some rows keep explicit _row_id
+	// values and others do not, which violates the per-file uniqueness and
+	// coverage invariants the v3 spec requires. Spec-compliant v3 planning
```
laskoviymishka: I think this framing is spec-incorrect for the migration case, and it's the one that landed. Pre-upgrade snapshots are spec-valid with null `first_row_id` — the spec assigns row IDs lazily "in the first commit after upgrading to v3," at manifest-list write time (see `manifest.go` ~1558). A file carried forward from v2 that hasn't yet been through a v3 manifest-list write has a legitimately nil `FirstRowID` (scanner copies it through at `scanner.go:805`; `inheritRowIDs` is false when the value is nil, `manifest.go:720`).

So a mixed group here is the common v1/v2→v3 migration case, not "invalid metadata or caller-constructed input" — and the drop-lineage branch below fires on a normal workload. The comment this replaced said exactly that. I'd restore the accurate framing: mixed groups are expected during migration, row IDs get assigned at the next v3 manifest-list write, and until then lineage is dropped for the merged output. wdyt?

table/rewrite_data_files.go
```diff
@@ -301,23 +301,17 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
...
+	// should already inherit first_row_id for every task, including files
+	// carried forward across a v2->v3 upgrade, so a mixed group here is
+	// treated conservatively as invalid metadata or caller-constructed input.
 	preserveLineage := tbl.metadata.Version() >= 3 && allTasksHaveRowLineage(group.Tasks)
 	if preserveLineage {
 		scanOpts = append(scanOpts, WithRowLineage())
 	} else if tbl.metadata.Version() >= 3 {
```
laskoviymishka, line 315: Dropping the inline comment on this branch leaves the `if lineageFiles > 0` guard below looking like a bug to a fresh reader — it's asymmetric (`lineageFiles`, not `legacyFiles`), and an all-legacy v3 group falls through silently with no log. I'd keep a condensed comment here explaining the actual action: lineage is dropped for the whole group and a `slog.Warn` is emitted only when at least one source file actually carried lineage, rather than returning an error or splitting the group.

table/rewrite_data_files.go
```diff
@@ -327,7 +321,7 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 			}
 		}
 		if lineageFiles > 0 {
-			slog.Warn("compaction group has mixed row lineage; dropping _row_id on output",
+			slog.Warn("compaction group has unexpected mixed row lineage on v3 table; dropping _row_id on output",
```
laskoviymishka: "on v3 table" is redundant given the `Version() >= 3` guard, and "unexpected" is non-actionable — and, per the comment above, a mixed group isn't actually unexpected during migration. The original string read cleaner; I'd leave it as it was.

table/rewrite_data_files.go
```diff
@@ -390,9 +384,9 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 
 // allTasksHaveRowLineage returns true iff every task in the group has a
 // non-nil FirstRowID — i.e. every source file already carries v3 row lineage.
-// Used to gate the preservation path on compaction: mixed groups (some
-// lineage, some legacy) would otherwise produce per-file invariant
-// violations, so the gate is conservative.
+// Used to gate the preservation path on compaction: mixed groups are treated
+// conservatively because spec-compliant v3 planning should already surface an
```
laskoviymishka: Same spec claim as above, and this godoc now justifies caller policy rather than describing what the helper does. I'd keep the first sentence, note it returns false for an empty slice, and drop the second sentence.

laskoviymishka (changes requested):
Thanks for tightening up the mixed-lineage docs here — that warning path deserves a clearer explanation than it had.

One blocking thing before merge, though: the PR doesn't do what the title and description say. It advertises splitting mixed v3 groups plus two regression tests, but the diff is comment-only (+11/-17, one warning string), runtime behavior is identical to main, and neither `TestExecuteCompactionGroupPreservesLineageSubsetOnMixedTasks` nor `TestExecuteCompactionGroupUsesRowLineage` exists in the repo — `go test -run` on an unknown pattern exits 0 with zero tests, so that's not real coverage.

And the new comment is spec-incorrect: it calls a mixed group "invalid metadata or caller-constructed input" and says v3 planning "should already inherit first_row_id for every task." That's not true across a v2→v3 upgrade — pre-upgrade files carry a legitimately null `first_row_id` until the first v3 manifest-list write assigns IDs lazily, so the drop-lineage branch fires on a normal migration workload. The comment this replaced described that honestly.

Easiest path is to make this an honest docs PR: retitle to something like "clarify mixed-group lineage-drop behavior," fix the description, and restore accurate framing on both the block comment and the `allTasksHaveRowLineage` godoc. Or, if the split is the real goal, land the code plus a genuine mixed-group test. One more thing worth a follow-up (not blocking here): when the warn path fires we silently erase previously-assigned row IDs with no catalog-visible signal — a snapshot-summary property like `compaction.lineage-dropped=true` would give CDC/`_row_id` readers something to detect.

Once the description and the spec framing are sorted, happy to take another pass and approve.

laskoviymishka (approved):
:shipit:

PR #1394, opened 2026-07-04
fix(metadata): validate v3 snapshot sequence numbers

fallintoplace:
## Summary
- In `MetadataBuilder.addSnapshotInternal`, sequence-number monotonicity validation now applies for format versions **>= 2** (not only `== 2`).
- This closes the v3 gap where a later snapshot with a non-increasing sequence number could be accepted.
- Adds regression coverage with `TestV3SequenceNumberCannotDecrease`.

## Testing
- `go test ./table -run 'TestV2SequenceNumberCannotDecrease|TestV3SequenceNumberCannotDecrease' -count=1`
- `go test ./table -run 'Test.*Snapshot' -count=1`
- `go test ./table -count=1`

### DIFF
--- a/table/metadata_builder_internal_test.go
+++ b/table/metadata_builder_internal_test.go
@@ -923,6 +923,47 @@ func TestV2SequenceNumberCannotDecrease(t *testing.T) {
 	require.ErrorContains(t, err, "can't add snapshot with sequence number 0, must be > than last sequence number 1")
 }
 
+func TestV3SequenceNumberCannotDecrease(t *testing.T) {
+	builder := builderWithoutChanges(3)
+	schemaID := 0
+	firstRowID1 := int64(0)
+	addedRows := int64(10)
+
+	snapshot1 := Snapshot{
+		SnapshotID:       1,
+		ParentSnapshotID: nil,
+		SequenceNumber:   0,
+		TimestampMs:      builder.base.LastUpdatedMillis() + 1,
+		ManifestList:     "/snap-1.avro",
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM, the `>= 2` change is the right call and matches the Java reference.

table/metadata_builder_internal_test.go
```diff
@@ -923,6 +923,46 @@ func TestV2SequenceNumberCannotDecrease(t *testing.T) {
...
+		ParentSnapshotID: &parentSnapshotID,
+		SequenceNumber:   0,
+		TimestampMs:      builder.lastUpdatedMS + 1,
+		ManifestList:     "/snap-0.avro",
+		Summary: &Summary{
+			Operation:  OpAppend,
+			Properties: map[string]string{},
+		},
+		SchemaID:   &schemaID,
+		FirstRowID: &firstRowID,
```
zeroshade: nit (cosmetic): `snapshot2` reuses the same `&firstRowID` pointer (value 0) as `snapshot1`, which adds 10 rows — realistically `snapshot2.FirstRowID` would start at 10. Doesn't affect what this test asserts (the sequence-number guard), but a distinct value would read more faithfully.

zeroshade (approved):
LGTM — broadening the sequence-number monotonicity guard from `== 2` to `>= 2` correctly extends it to v3+ (matching the Java reference); v1 and root snapshots are unaffected. One cosmetic test nit inline; not blocking. Thanks!

laskoviymishka (approved):

PR #1395, opened 2026-07-04
fix(codec): validate FileScanTask scan ranges on decode

fallintoplace:
## Summary
- Reject negative Start/Length scan ranges in DecodeFileScanTask.
- Add regression coverage for malformed envelopes with negative Start and Length.

## Testing
- go test ./codec -run 'DecodeFileScanTask|EncodeDecodeFileScanTask' -count=1
- go test ./codec -count=1

### DIFF
--- a/codec/file_scan_task.go
+++ b/codec/file_scan_task.go
@@ -87,6 +87,12 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 	if _, err := fileScanTaskSchema.Decode(data, &envelope); err != nil {
 		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: decode: %w", err)
 	}
+	if envelope.Start < 0 {
+		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: start must be non-negative: %d", envelope.Start)
+	}
+	if envelope.Length < 0 {
+		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: length must be non-negative: %d", envelope.Length)
+	}
 	file, err := DecodeDataFile(envelope.File, spec, schema, version)
 	if err != nil {
 		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: file: %w", err)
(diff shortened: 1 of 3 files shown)


codec/file_scan_task.go
```diff
@@ -87,6 +87,12 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 	if _, err := fileScanTaskSchema.Decode(data, &envelope); err != nil {
 		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: decode: %w", err)
 	}
+	if envelope.Start < 0 {
```
tanmayrauth, line 96: Nice that this sits before DecodeDataFile so a malformed envelope is rejected cheaply. One thought: EncodeFileScanTask copies task.Start/task.Length straight through without the same check, so a task built with a negative range encodes fine and only blows up on decode. Since this is a cross-process boundary, would you consider mirroring these two guards in the encoder so a bad range fails fast at the producer too?

codec/file_scan_task_internal_test.go
```diff
@@ -47,3 +47,38 @@ func TestDecodeFileScanTaskInnerErrorCarriesMarker(t *testing.T) {
 	require.Contains(t, err.Error(), "codec: DecodeFileScanTask: file:",
 		"an inner (primary-file) decode error must carry the function + sub-path marker")
 }
+
```
tanmayrauth, line 50: Minor: since the range check returns before DecodeDataFile is ever reached, the spec/schema fixtures here aren't actually exercised on this path. Fine to leave, but you could drop them or add a one-liner noting the check  intentionally precedes file decode.

tanmayrauth (commented):
Clean defensive check and it's round-trip safe — the encoder only ever emits Start=0/Length=size. Two small things inline.

laskoviymishka (approved):
👍

laskoviymishka:
@fallintoplace i'll wait for @tanmayrauth comments to fix before merging

fallintoplace:
Both comments should be addressed now. It would be great if you can check again. @tanmayrauth

tanmayrauth:
LGTM!

PR #1396, opened 2026-07-04
fix(io): return not-exist error for missing MemFS removals

fallintoplace:
## Summary
- Make MemFS.Remove return a `*fs.PathError` with `ErrNotExist` when the target file path is absent.
- Add regression coverage in `io/mem_test.go` for removing a missing file path.

## Testing
- `go test ./io -count=1`

### DIFF
--- a/io/mem.go
+++ b/io/mem.go
@@ -80,8 +80,13 @@ func (m *MemFS) Open(name string) (File, error) {
 
 func (m *MemFS) Remove(name string) error {
 	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if _, ok := m.files[name]; !ok {
+		return &fs.PathError{Op: "remove", Path: name, Err: fs.ErrNotExist}
+	}
+
 	delete(m.files, name)
-	m.mu.Unlock()
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


io/mem.go
```diff
@@ -80,8 +80,13 @@ func (m *MemFS) Open(name string) (File, error) {
 
 func (m *MemFS) Remove(name string) error {
 	m.mu.Lock()
+	defer m.mu.Unlock()
```
tanmayrauth, line 83: Nice catch switching to `defer m.mu.Unlock()` here — without it the new early return on a missing key would've leaked the lock. And using `Op: "remove"` matches os.Remove's own PathError.Op, so error strings stay consistent between the mem and local backends. This also means  table/orphan_cleanup.go:858's `!os.IsNotExist(rmErr)` guard is now actually exercised under mem://, which it wasn't before. LGTM.

tanmayrauth (commented):
LGTM — this lines MemFS up with LocalFS/os.Remove and with the contract in io.go, and callers already assume it (orphan_cleanup.go guards on os.IsNotExist). One small note inline.

laskoviymishka (approved):
🚢

PR #1397, opened 2026-07-04
fix(manifest): protect FileSequenceNum from aliasing

fallintoplace:
### What changed
- Return a copied pointer from `ManifestEntry.FileSequenceNum()` to avoid exposing internal `manifestEntry` state.
- Added regression coverage for pointer aliasing; mutating the returned pointer no longer mutates manifest-entry data.
- Added tests in `manifest_test.go` for `FileSequenceNum()` copy behavior.

### DIFF
--- a/table/updates.go
+++ b/table/updates.go
@@ -775,6 +775,7 @@ func (u *addEncryptionKeyUpdate) Apply(builder *MetadataBuilder) error {
 	if err := u.EncryptionKey.Validate(); err != nil {
 		return fmt.Errorf("%w: %w", iceberg.ErrInvalidArgument, err)
 	}
+
 	return builder.AddEncryptionKey(u.EncryptionKey)
 }
 
--- a/table/updates.go
+++ b/table/updates.go
@@ -772,6 +772,9 @@ func NewAddEncryptionKeyUpdate(key EncryptionKey) *addEncryptionKeyUpdate {
 }
 
 func (u *addEncryptionKeyUpdate) Apply(builder *MetadataBuilder) error {
+	if err := u.EncryptionKey.Validate(); err != nil {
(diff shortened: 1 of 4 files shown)


table/updates.go
```diff
@@ -772,6 +772,10 @@ func NewAddEncryptionKeyUpdate(key EncryptionKey) *addEncryptionKeyUpdate {
 }
 
 func (u *addEncryptionKeyUpdate) Apply(builder *MetadataBuilder) error {
+	if err := u.EncryptionKey.Validate(); err != nil {
```
tanmayrauth: This validate is redundant — Apply just delegates to AddEncryptionKey, which now validates the same key at metadata.go:1341. Might be worth dropping this one so there's a single source of truth. Minor side effect of keeping it: on a V2 table with an invalid key, Apply reports the "key-id" error first, whereas calling AddEncryptionKey directly reports the "format version 3" error first — both ErrInvalidArgument, so not a real problem, just slightly  inconsistent ordering.

tanmayrauth (approved):
PR looks good and the tests are thorough, just one small nit.

zeroshade (commented):
Heads-up: the title/description say `fix(manifest): protect FileSequenceNum from aliasing`, but the branch here is `fix/encryption-key-validation` and the diff only touches `table/encryption.go`, `table/metadata.go`, and `table/updates_test.go` — this is encryption-key validation, not a manifest/FileSequenceNum change.

It's also a strict subset of #1406: the patches for those three files are byte-for-byte identical, and #1406 additionally wires the `EncryptionKey.Validate()` call into `table/updates.go`'s `Apply()`. Recommend closing this in favor of #1406 to avoid a duplicate-definition conflict if both land. (The actual FileSequenceNum aliasing fix is #1401.)

PR #1398, opened 2026-07-04
fix(conflict_validation): reject invalid isolation-level values

fallintoplace:
## Summary
- reject malformed table conflict isolation-level properties at validation time
- propagate errors from overwrite/row-delta conflict checks
- add regression tests for helper and validator error behavior

## Testing
- go test ./table -count=1


table/conflict_validation.go
```diff
@@ -96,19 +96,23 @@ const (
...
+		return defVal, nil
 	}
+
 	switch IsolationLevel(v) {
 	case IsolationSerializable, IsolationSnapshot:
-		return IsolationLevel(v)
+		return IsolationLevel(v), nil
 	default:
-		return defVal
+		return defVal, fmt.Errorf("%w: %q for property %q", ErrInvalidIsolationLevel, v, key)
```
tanmayrauth, line 125: This rejects anything that isn't exactly lowercase "serializable"/"snapshot" (empty string included). I think Java's fromName is case-insensitive — if so, a config ported from Java/Spark with "SERIALIZABLE" would now fail here. Worth confirming; lowercasing v before the switch would keep parity.

tanmayrauth (commented):
Clean change. One question about the new reject path.

table/conflict_validation.go
```diff
@@ -96,19 +97,23 @@ const (
...
-		return defVal
+		return defVal, nil
 	}
-	switch IsolationLevel(v) {
+
+	switch IsolationLevel(strings.ToLower(v)) {
 	case IsolationSerializable, IsolationSnapshot:
-		return IsolationLevel(v)
+		return IsolationLevel(strings.ToLower(v)), nil
 	default:
```
laskoviymishka, line 124: the empty-string case is the one I'd push back on here. right now `''` (key present, empty value) becomes a hard, non-retryable commit failure — but that's a value some catalogs write when you "clear" a property without deleting the key, and some engines use it as a null/default repr.

the spec only says a missing key falls back to default; it's silent on empty string. i'd keep `''` on the fallback path alongside the absent-key case and only error on non-empty unrecognized values. that also means dropping the new "empty string returns error" test. wdyt?

table/snapshot_producers.go
```diff
@@ -259,7 +259,11 @@ func (of *overwriteFiles) validate(cc *conflictContext) error {
 	if of.base.op == OpDelete {
 		key, defVal = WriteDeleteIsolationLevelKey, WriteDeleteIsolationLevelDefault
 	}
-	if readIsolationLevel(of.base.txn.meta.props, key, defVal) != IsolationSerializable {
+	level, err := readIsolationLevel(of.base.txn.meta.props, key, defVal)
```
laskoviymishka, line 258: this validation only runs once we actually reach `validate()` — and on the first commit attempt `cc` is nil and we bail at the top before ever getting here, so an invalid isolation-level silently commits on the fast path.

the PR title says we reject invalid values, but really we reject them only on certain paths. i'd lift the isolation-level check somewhere unconditional (top of `validate()`, or at commit time before the producer runs) so it fires regardless of file composition or retry state.

table/row_delta.go
```diff
@@ -230,8 +230,11 @@ func (rd *RowDelta) validate(cc *conflictContext) error {
 	}
 
 	if len(eqDeleteFiles) > 0 {
-		level := readIsolationLevel(rd.txn.meta.props,
+		level, err := readIsolationLevel(rd.txn.meta.props,
```
laskoviymishka: same gating issue on this path — the read + validate lives inside `if len(eqDeleteFiles) > 0`, so a RowDelta carrying only positional deletes with an invalid `write.delete.isolation-level` commits without error. pulling the property validation up into a single unconditional step covers this case too.

table/conflict_validation.go
```diff
@@ -96,19 +97,23 @@ const (
 	WriteUpdateIsolationLevelDefault = IsolationSerializable
 )
 
+// ErrInvalidIsolationLevel is returned when a configured isolation level
+// value is invalid.
```
laskoviymishka: could we add a line noting this is deliberately terminal — it doesn't wrap `ErrCommitFailed`, so it exits `doCommit` without retry rather than being treated as a retryable conflict?

every other exported sentinel here documents its retry semantics, and without the note it's easy for the next person to "fix" this by wrapping `ErrCommitFailed` and quietly make a config error retryable.

table/conflict_validation.go
```diff
@@ -96,19 +97,23 @@ const (
...
+func readIsolationLevel(props iceberg.Properties, key string, defVal IsolationLevel) (IsolationLevel, error) {
 	v, ok := props[key]
 	if !ok {
-		return defVal
+		return defVal, nil
 	}
-	switch IsolationLevel(v) {
+
+	switch IsolationLevel(strings.ToLower(v)) {
 	case IsolationSerializable, IsolationSnapshot:
```
laskoviymishka, line 122: minor: we lower-case `v` twice on the match path — once for the switch discriminant, once in the returned value. worth computing it once:

```go
lower := IsolationLevel(strings.ToLower(v))
switch lower {
case IsolationSerializable, IsolationSnapshot:
	return lower, nil
default:
	return defVal, fmt.Errorf("%w: %q for property %q", ErrInvalidIsolationLevel, v, key)
}
```

keeping the original `v` in the error message is the right call — nice for a faithful repro.

table/conflict_validation_test.go
```diff
@@ -28,52 +28,73 @@ import (
...
+			want:    IsolationSnapshot,
+			wantErr: true,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got := readIsolationLevel(tt.props, tt.key, tt.defVal)
+			got, err := readIsolationLevel(tt.props, tt.key, tt.defVal)
+			if tt.wantErr {
```
laskoviymishka, line 191: this early return skips the `assert.Equal(tt.want, got)` below, so `want` on the two error rows is never actually checked — it's dead. if the impl started returning `IsolationSnapshot` instead of `defVal` on the error path, these tests would still pass.

either drop `want` from the error rows, or assert `got` on the error path too (`assert.Equal(t, tt.want, got)` before the return).

table/producer_validate_test.go
```diff
@@ -165,18 +165,34 @@ func TestOverwriteFiles_IsolationKeySplitByOp(t *testing.T) {
...
 		"non-delete ops must consult write.update.isolation-level")
 }
 
+// TestOverwriteFiles_ValidateInvalidIsolationLevel verifies that malformed
+// isolation settings fail the validator before conflict scanning runs.
+func TestOverwriteFiles_ValidateInvalidIsolationLevel(t *testing.T) {
+	cc := newEmptyConflictContext(t)
+	txn := newValidateTestTxn(t, iceberg.Properties{
+		WriteUpdateIsolationLevelKey: "not-a-level",
+	})
```
laskoviymishka, line 190: this only exercises the non-delete branch (update key, `op != OpDelete`). the `op == OpDelete` branch that swaps in `WriteDeleteIsolationLevelKey` never gets an invalid-value test, so a transposition of the two keys wouldn't be caught. worth a sibling case with `of.base.op = OpDelete` and an invalid value under the delete key.

laskoviymishka (changes requested):
Direction here is right — moving from silent fallback to rejecting invalid isolation-level values matches Java, and the mechanics are clean: the error wraps with `%w`, `overwriteFiles.validate` checks it before the SNAPSHOT nil-return, and the `conflictErr` rename reads well.

I'd hold this before merging though, because two design questions decide whether it actually does what the title says.

The empty-string case is the first one. Turning `''` into a hard, non-retryable commit failure changes observable behavior for any table that ever cleared the property to an empty value — that's a real thing catalogs and engines do when they can't delete an individual property. The spec only speaks to missing keys, so I'd keep `''` on the fallback path and only error on non-empty unrecognized values.

The second is that the check is gated behind file composition and retry state — on the first commit attempt `cc` is nil and we return before reaching it, and RowDelta only validates when there are equality deletes. So an invalid level silently commits on those paths. I think this wants to be a single unconditional check (top of `validate()`, or at commit time) rather than lazy.

- empty string: fall back vs. hard error (I lean fall back)
- make the invalid-value check fire on all paths, not just eq-deletes / non-nil `cc`
- a doc note on `ErrInvalidIsolationLevel` that it's terminal and deliberately doesn't wrap `ErrCommitFailed`
- decide whether to also model `write.merge.isolation-level` here or explicitly note it's out of scope for now

The rest is small — a double `ToLower`, the dead `want` fields on the error test rows, and a missing `op == OpDelete` coverage case. Once the two design questions are settled, happy to take another pass and approve.

zeroshade (approved):
LGTM — all the earlier points are addressed: empty-string fallback, the unconditional isolation-level check moved before the `cc == nil` guard in both paths, the terminal-error doc note, single `ToLower`, and `OpDelete` coverage. One small scoping note: `write.merge.isolation-level` isn't explicitly called out as out-of-scope — a one-line note would clarify intent (it silently falls back like before, so there's no regression).

laskoviymishka (approved):
lgtm! but need to fix conflicts

PR #1399, opened 2026-07-04
fix(cli): validate branch and tag retention inputs

fallintoplace:
## Why
The branch/tag create commands previously accepted invalid ref names and non-positive retention values. Those inputs could create metadata updates that should have been rejected up front.

## What Changed
- Reject invalid branch/tag names before updates are built, preventing malformed refs (empty names, whitespace-wrapped names, path separators, `.`/`..`, control characters).
- Reject non-positive retention settings (`--max-ref-age`, `--max-snapshot-age`, `--min-snapshots-to-keep`) so invalid policies are not persisted.
- Add regression coverage for both branch and tag validation paths.

## Testing
`go test ./cmd/iceberg -run 'TestRunBranchCreateRejectsNegativeRetentionValues|TestRunBranchCreateRejectsInvalidRefName|TestRunTagCreateRejectsNegativeMaxRefAge|TestRunTagCreateRejectsInvalidRefName' -count=1`

### DIFF
--- a/cmd/iceberg/branch_tag.go
+++ b/cmd/iceberg/branch_tag.go
@@ -23,6 +23,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
+	"unicode"
 
 	"github.com/apache/iceberg-go/catalog"
 	"github.com/apache/iceberg-go/table"
@@ -50,6 +52,12 @@ func runTag(ctx context.Context, output Output, cat catalog.Catalog, cmd *TagCmd
 func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *BranchCreateCmd) {
 	tbl := loadTable(ctx, output, cat, cmd.TableID)
 	meta := tbl.Metadata()
+	if err := validateRefName(cmd.BranchName); err != nil {
+		output.Error(fmt.Errorf("invalid branch name: %w", err))
+		osExit(1)
+
+		return
+	}
 
 	if _, found := findSnapshotRef(meta, cmd.BranchName); found {
 		output.Error(fmt.Errorf("ref %q already exists", cmd.BranchName))
@@ -80,6 +88,13 @@ func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cm
 			return
 		}
 
+		if d <= 0 {
+			output.Error(errors.New("invalid --max-ref-age: must be greater than 0"))
+			osExit(1)
+
+			return
+		}
+
 		maxRefAgeMs = d.Milliseconds()
 	}
 
@@ -93,11 +108,25 @@ func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cm
 			return
 		}
 
+		if d <= 0 {
+			output.Error(errors.New("invalid --max-snapshot-age: must be greater than 0"))
+			osExit(1)
+
+			return
+		}
+
 		maxSnapshotAgeMs = d.Milliseconds()
 	}
 
 	var minSnapshotsToKeep int
 	if cmd.MinSnapshotsToKeep != nil {
+		if *cmd.MinSnapshotsToKeep <= 0 {
+			output.Error(errors.New("invalid --min-snapshots-to-keep: must be greater than 0"))
+			osExit(1)
+
+			return
+		}
+
 		minSnapshotsToKeep = *cmd.MinSnapshotsToKeep
 	}
 
@@ -137,6 +166,12 @@ func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cm
 func runTagCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *TagCreateCmd) {
 	tbl := loadTable(ctx, output, cat, cmd.TableID)
 	meta := tbl.Metadata()
+	if err := validateRefName(cmd.TagName); err != nil {
+		output.Error(fmt.Errorf("invalid tag name: %w", err))
+		osExit(1)
+
+		return
+	}
 
 	if _, found := findSnapshotRef(meta, cmd.TagName); found {
 		output.Error(fmt.Errorf("ref %q already exists", cmd.TagName))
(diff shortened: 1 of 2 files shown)


cmd/iceberg/branch_tag.go
```diff
@@ -80,6 +88,13 @@ func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cm
 			return
 		}
 
+		if d <= 0 {
```
tanmayrauth, line 92: Good catch — worth noting in the PR description that this isn't redundant with the library's `WithMaxRefAgeMs` guard: `setSnapshotRefUpdate.Apply` only applies etention when the value is > 0, so today a negative `--max-ref-age` is silently  dropped rather than rejected. This surfaces it.

cmd/iceberg/branch_tag.go
```diff
@@ -197,6 +239,26 @@ func runTagCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *
...
+
+	if name == "." || name == ".." {
+		return errors.New("name may not be '.' or '..'")
+	}
+
+	if strings.ContainsAny(name, "/\\") {
+		return errors.New("name may not contain path separators")
+	}
+
+	if strings.IndexFunc(name, func(r rune) bool { return unicode.IsControl(r) }) >= 0 {
```
tanmayrauth: nit: `strings.ContainsFunc(name, unicode.IsControl)` reads a bit cleaner here

cmd/iceberg/branch_tag.go
```diff
@@ -50,6 +52,12 @@ func runTag(ctx context.Context, output Output, cat catalog.Catalog, cmd *TagCmd
 func runBranchCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *BranchCreateCmd) {
 	tbl := loadTable(ctx, output, cat, cmd.TableID)
 	meta := tbl.Metadata()
+	if err := validateRefName(cmd.BranchName); err != nil {
```
tanmayrauth, line 53: Could this move above `loadTable` at line 51? It only reads `cmd.BranchName`, so  running it first lets an invalid name fail before a full catalog load (a network round-trip for REST/Glue). Same for the tag path at line 169.

cmd/iceberg/branch_tag.go
```diff
@@ -197,6 +239,26 @@ func runTagCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *
 	output.RefCreated(result)
 }
 
+func validateRefName(name string) error {
```
tanmayrauth, line 244: This (and the retention checks) only guard the CLI path — `table.SetSnapshotRef` / `NewSetSnapshotRefUpdate` still accept the same inputs, so other iceberg-go consumers get no protection. Fine to keep CLI-scoped for this PR, but if the goal is data integrity the name/retention validation arguably belongs in the library so every caller benefits.

cmd/iceberg/branch_tag.go
```diff
@@ -197,6 +239,26 @@ func runTagCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *
...
+	if strings.TrimSpace(name) != name || name == "" {
+		return errors.New("name must be non-empty and may not contain leading/trailing whitespace")
+	}
+
+	if name == "." || name == ".." {
+		return errors.New("name may not be '.' or '..'")
+	}
+
+	if strings.ContainsAny(name, "/\\") {
```
tanmayrauth: Was the path-separator ban intentional? This makes the CLI stricter than the  iceberg-go `table` layer, the spec, and Java/Spark/PyIceberg, none of which restrict ref-name characters. Slash-style branch names (`audit/foo`, git-flow `feature/x`) are common in WAP workflows and created freely by other engines —
  after this, a branch that's legal in Spark can't be created with the same name here. If the motivation is Spark's `branch_<name>` metadata-selector ambiguity,  worth calling that out in a comment; otherwise this may be over-tight.

tanmayrauth (commented):
Nice fix, the `d <= 0` checks fill a real gap (the update Apply layer silently drops non-positive retention today). One question about how strict the name validation should be, plus a couple of nits. Details inline.

cmd/iceberg/branch_tag_test.go
```diff
@@ -109,6 +110,158 @@ func TestTextOutputRefCreated(t *testing.T) {
...
+func TestRunBranchCreateRejectsNegativeRetentionValues(t *testing.T) {
+	tbl := branchTagTestTable(t)
+	cat := &branchTagCommitCatalog{tbl: tbl}
+	out := &refCreatedCapture{}
+
+	exitCode := captureBranchTagExit(func() {
+		runBranchCreate(context.Background(), out, cat, &BranchCreateCmd{
+			TableID:    "db.events",
+			BranchName: "new-feature",
+			MaxRefAge:  "-1d",
```
zeroshade: `-1d` isn't a valid Go duration (`time.ParseDuration` supports ns/us/ms/s/m/h, not `d`), so this hits the parse-error branch rather than the `d <= 0` guard it's meant to exercise — the assertion passes for the wrong reason. The tag test correctly uses `-12h`; use a valid negative duration like `-1h` here.

cmd/iceberg/branch_tag.go
```diff
@@ -197,6 +241,26 @@ func runTagCreate(ctx context.Context, output Output, cat catalog.Catalog, cmd *
...
```
zeroshade: Rejecting `/` is stricter than the Iceberg spec / Java / Spark / PyIceberg, which allow slash-style ref names (e.g. `audit/foo`) common in WAP workflows. If this is intentional (e.g. Spark selector concerns), a short comment explaining why would help; otherwise consider relaxing it.

zeroshade (changes requested):
A misleading test and one interoperability concern (inline).

zeroshade (approved):
Re-reviewed after the latest commit — both points are addressed:
- The test now uses `-1h` (a valid negative duration) so it actually exercises the `d <= 0` guard instead of failing at parse time.
- The `/` restriction is gone from `validateRefName`, and slash-style ref names (`audit/foo`, `release/1.0`) are now explicitly tested as allowed.

Also nice to see the name validation moved ahead of the table load. LGTM — thanks!

laskoviymishka (approved):
👍

PR #1400, opened 2026-07-04
fix(table): validate parquet file writer properties

fallintoplace:
Parquet writer initialization now validates WriteProps type before constructing writer properties so malformed file-info does not panic.

- Adds safe getWriteProperties helper used by NewFileWriter.
- Returns iceberg.ErrInvalidArgument-compatible errors when write properties are nil or the wrong type.
- Adds regression tests for nil and non-slice write properties.

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -365,7 +365,12 @@ func (p parquetFormat) NewFileWriter(ctx context.Context, fs iceio.WriteFileIO,
 
 	counter := &internal.CountingWriter{W: fw}
 	mem := compute.GetAllocator(ctx)
-	writerProps := parquet.NewWriterProperties(info.WriteProps.([]parquet.WriterProperty)...)
+	writerProps, err := getWriteProperties(info.WriteProps)
+	if err != nil {
+		fw.Close()
+
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):

table/internal/parquet_files.go
```diff
@@ -387,6 +392,19 @@ func (p parquetFormat) NewFileWriter(ctx context.Context, fs iceio.WriteFileIO,
 	}, nil
 }
 
+func getWriteProperties(writeProps any) (*parquet.WriterProperties, error) {
+	if writeProps == nil {
+		return nil, fmt.Errorf("%w: write properties are required", iceberg.ErrInvalidArgument)
```
zeroshade, line 399: nit: since `WriteProps` is typed `any`, a reader might reasonably expect `nil` to mean "use defaults". A brief comment documenting that nil is intentionally treated as an error (fail-fast) rather than defaulting to `parquet.NewWriterProperties()` would avoid confusion.

zeroshade (approved):
LGTM — replacing the bare type assertion with `getWriteProperties` turns a latent panic into a proper `ErrInvalidArgument`, and the writer file handle is closed on the error path. One doc nit inline; not blocking. Thanks!

laskoviymishka (approved):
👍

PR #1401, opened 2026-07-05
fix(manifest): protect FileSequenceNum from aliasing

fallintoplace:
## Why
Manifest callers currently can mutate internal `ManifestEntry` state by writing through the pointer returned from `FileSequenceNum()`. That creates subtle state corruption risk during metadata handling and makes aliasing bugs hard to detect.

## What changed
- Return a copied `int64` pointer from `FileSequenceNum()` instead of exposing the internal pointer directly.
- Keep behavior unchanged for nil sequence values.
- Preserve the existing API shape while preventing external mutation of the manifest entry's internal sequence number.
- Add regression coverage that confirms returned pointer mutation no longer impacts the underlying manifest entry.

## Validation
- `go test ./...` for package-level coverage of the updated tests.
- `golangci-lint` clean on this branch.

## Compatibility
This is a source-compatible behavioral safety fix with no external API change.

### DIFF
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -1941,8 +1941,6 @@ func (m *ManifestTestSuite) TestManifestEntryFileSequenceNumReturnsCopy() {
 	current := entry.FileSequenceNum()
 	m.Require().NotNil(current)
 	m.Assert().Equal(int64(10), *current)
-	m.Assert().NotEqual(int64(20), *current)
-	m.Assert().Equal(entryFileSeqNum, *current)
 }
 
 // equalityIDsSchemaIsInt asserts equality_ids uses Avro "int", not "long".
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -1925,6 +1925,26 @@ func (m *ManifestTestSuite) TestManifestEntryBuilder() {
 	m.Assert().Equal(0, *data.SortOrderID())
 }
 
+func (m *ManifestTestSuite) TestManifestEntryFileSequenceNumReturnsCopy() {
+	entryFileSeqNum := int64(10)
+	entry := &manifestEntry{
+		EntryStatus: EntryStatusADDED,
+		FileSeqNum:  &entryFileSeqNum,
+		Data:        &dataFile{Path: "/data/sample.parquet"},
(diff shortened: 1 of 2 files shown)


manifest_test.go
```diff
@@ -1925,6 +1925,26 @@ func (m *ManifestTestSuite) TestManifestEntryBuilder() {
...
+
+	got := entry.FileSequenceNum()
+	m.Require().NotNil(got)
+	m.Assert().Equal(entryFileSeqNum, *got)
+	*got = 20
+
+	current := entry.FileSequenceNum()
+	m.Require().NotNil(current)
+	m.Assert().Equal(int64(10), *current)
+	m.Assert().NotEqual(int64(20), *current)
```
tanmayrauth: nit: these last two assertions are redundant with the Equal(int64(10), *current) above — after that passes, both NotEqual(20) and Equal(entryFileSeqNum) are already implied. Could drop them to keep the test tight.

tanmayrauth (approved):
LGTM, one small nit.

manifest_test.go
```diff
@@ -1925,6 +1925,25 @@ func (m *ManifestTestSuite) TestManifestEntryBuilder() {
...
+	}
+
+	got := entry.FileSequenceNum()
+	m.Require().NotNil(got)
+	m.Assert().Equal(entryFileSeqNum, *got)
+	*got = 20
+
+	current := entry.FileSequenceNum()
+	m.Require().NotNil(current)
+	m.Assert().NotSame(got, current)
```
zeroshade: nit: `NotSame(got, current)` together with `Equal(int64(10), *current)` is slightly redundant — asserting the re-fetched value is still `10` after mutating `*got` already proves the returned pointer is an independent copy. Harmless either way.

zeroshade (approved):
LGTM — returning a copy of the `*int64` from `FileSequenceNum()` correctly prevents callers from mutating internal `manifestEntry` state, and the nil case is preserved. One minor test nit inline; not blocking. Thanks!

laskoviymishka (approved):
:shipit:

PR #1402, opened 2026-07-05
fix(manifest): clone partition and key metadata inputs

fallintoplace:
## Summary

`ManifestBuilder` previously retained caller-owned partition summaries and key-metadata bytes, allowing later caller mutations to change built manifest metadata.

- deep-copy partition summaries, including bounds and `ContainsNaN` pointers;
- clone key-metadata bytes;
- preserve nil inputs;
- add regression tests that mutate every caller-owned input after building.

## Testing

- `go test .`

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -144,13 +144,48 @@ func (b *ManifestBuilder) DeletedRows(cnt int64) *ManifestBuilder {
 }
 
 func (b *ManifestBuilder) Partitions(p []FieldSummary) *ManifestBuilder {
-	b.m.PartitionList = &p
+	if p == nil {
+		// Preserve the semantic "no partition summaries" state by clearing the
+		// union pointer entirely instead of storing a pointer to a nil slice.
+		b.m.PartitionList = nil
+
+		return b
+	}
+
+	copied := make([]FieldSummary, len(p))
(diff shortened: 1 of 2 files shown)


manifest.go
```diff
@@ -144,13 +144,44 @@ func (b *ManifestBuilder) DeletedRows(cnt int64) *ManifestBuilder {
 }
 
 func (b *ManifestBuilder) Partitions(p []FieldSummary) *ManifestBuilder {
-	b.m.PartitionList = &p
+	if p == nil {
+		b.m.PartitionList = nil
+
```
zeroshade, line 151: nit: worth a one-line comment noting the intentional behavior change — previously `Partitions(nil)` stored a non-nil pointer to a nil slice (`&p`), whereas now it stores `nil` (same for `KeyMetadata(nil)`). It's arguably more correct, but Avro serialization sees a nil-pointer vs. empty-value difference, so a note would help future readers.

zeroshade (approved):
LGTM — the partition `FieldSummary` slice (including `LowerBound`/`UpperBound`/`ContainsNaN`) and key metadata are now deep-copied, closing the caller-aliasing window. One doc nit inline; not blocking. Thanks!

PR #1403, opened 2026-07-05
fix(table): clone row delta snapshot properties

fallintoplace:
## Why
`Transaction.NewRowDelta` used to store `snapshotProps` by reference. If callers mutated the same map after creating a row delta, those changes leaked into snapshot summary metadata after commit.

## Fix
Clone snapshot properties at construction time so the `RowDelta` snapshot metadata is isolated from caller-owned maps.

- Defensive-copy `snapshotProps` in `NewRowDelta`.
- Add a regression test (`TestNewRowDeltaCopiesSnapshotProperties`) that mutates the original map after `NewRowDelta` and verifies the committed snapshot still uses the original values.

## Testing
go test ./table -run TestNewRowDeltaCopiesSnapshotProperties -count=1

### DIFF
--- a/table/row_delta.go
+++ b/table/row_delta.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 
 	"github.com/apache/iceberg-go"
 )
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM. Nice catch on the builder-lifetime hazard — `NewRowDelta` hands the `*RowDelta` back to the caller and commits later, so aliasing the props map was a real leak window, unlike `Append`/`Overwrite` which build+commit in one call and don't need this. The `nil` case stays safe too since `maps.Clone(nil)` is `nil` and the downstream `maps.Copy` no-ops on it. Test covers both the mutated
  key and the added key.

zeroshade (approved):

PR #1404, opened 2026-07-05
fix(types): validate fixed type length

fallintoplace:
## Why
`fixed[N]` schemas should only allow a positive fixed byte length, but the parser and constructor path currently allowed invalid lengths through. That could leak malformed type definitions into metadata and fail later in conversion/encoding paths.

## What changed
- Added strict fixed-length validation for both construction and parsing paths.
- `FixedTypeOf` now rejects non-positive lengths.
- Fixed-type parsing now validates the `fixed(<length>)` declaration using the same checks, so `fixed[0]` and invalid forms are rejected immediately.
- Added regression coverage for malformed/invalid fixed type lengths.

## Validation
- `go test . -run "TestFixedType|TestFixedTypeInvalidParse" -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m`

### DIFF
--- a/types.go
+++ b/types.go
@@ -189,7 +189,10 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
 					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
 				}
 
-				n, _ := strconv.Atoi(matches[1])
+				n, err := strconv.Atoi(matches[1])
+				if err != nil {
+					return fmt.Errorf("%w: invalid fixed length %q: %v", ErrInvalidTypeString, matches[1], err)
+				}
 				t.Type = FixedType{len: n}
 			case strings.HasPrefix(typename, "decimal"):
 				matches := decimalRegex.FindStringSubmatch(typename)
@@ -558,7 +561,21 @@ func (m *MapType) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-func FixedTypeOf(n int) FixedType { return FixedType{len: n} }
+func validateFixedLength(length int) error {
+	if length < 0 {
+		return fmt.Errorf("fixed length must be non-negative, got %d", length)
+	}
+
+	return nil
+}
+
+func FixedTypeOf(n int) FixedType {
+	if err := validateFixedLength(n); err != nil {
+		panic(fmt.Errorf("%w: %s", ErrInvalidArgument, err))
+	}
+
+	return FixedType{len: n}
+}
 
 type FixedType struct {
 	len int
--- a/types_test.go
+++ b/types_test.go
@@ -47,6 +47,7 @@ func TestTypesBasic(t *testing.T) {
 		{"binary", iceberg.PrimitiveTypes.Binary},
 		{"unknown", iceberg.PrimitiveTypes.Unknown},
 		{"variant", iceberg.VariantType{}},
+		{"fixed[0]", iceberg.FixedTypeOf(0)},
 		{"fixed[5]", iceberg.FixedTypeOf(5)},
 		{"decimal(9, 4)", iceberg.DecimalTypeOf(9, 4)},
 	}
@@ -77,6 +78,49 @@ func TestFixedType(t *testing.T) {
 	assert.Equal(t, "fixed[5]", typ.String())
 	assert.True(t, typ.Equals(iceberg.FixedTypeOf(5)))
 	assert.False(t, typ.Equals(iceberg.FixedTypeOf(6)))
+	assert.Equal(t, "fixed[0]", iceberg.FixedTypeOf(0).String())
+
+	t.Run("FixedTypeOf validates length", func(t *testing.T) {
+		assertFixedTypeOfPanicsWithInvalidArgument(t, -1)
+	})
+}
+
+func TestFixedTypeInvalidParse(t *testing.T) {
+	tests := []struct {
+		name string
+		data string
+	}{
+		{
+			name: "malformed suffix",
+			data: `{"id": 1, "name": "f", "type": "fixed[0]junk", "required": true}`,
+		},
+		{
+			name: "overflow",
+			data: `{"id": 1, "name": "f", "type": "fixed[99999999999999999999]", "required": true}`,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var n iceberg.NestedField
+			err := json.Unmarshal([]byte(tt.data), &n)
+			assert.ErrorIs(t, err, iceberg.ErrInvalidTypeString)
+		})
+	}
+}
+
+func assertFixedTypeOfPanicsWithInvalidArgument(t *testing.T, length int) {
+	t.Helper()
+
+	defer func() {
+		r := recover()
+		require.NotNil(t, r)
+		err, ok := r.(error)
+		require.True(t, ok)
+		assert.ErrorIs(t, err, iceberg.ErrInvalidArgument)
+	}()
+
+	iceberg.FixedTypeOf(length)
 }
 
 func TestDecimalType(t *testing.T) {
--- a/literals.go
+++ b/literals.go
@@ -1119,7 +1119,7 @@ func (b *BinaryLiteral) UnmarshalBinary(data []byte) error {
 type FixedLiteral []byte
 
 func (FixedLiteral) Comparator() Comparator[[]byte] { return bytes.Compare }
-func (f FixedLiteral) Type() Type                   { return FixedTypeOf(len(f)) }
+func (f FixedLiteral) Type() Type                   { return FixedType{len: len(f)} }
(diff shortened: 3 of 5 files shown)


types.go
```diff
@@ -558,7 +564,21 @@ func (m *MapType) UnmarshalJSON(b []byte) error {
...
-func FixedTypeOf(n int) FixedType { return FixedType{len: n} }
+func validateFixedLength(length int) error {
+	if length <= 0 {
+		return fmt.Errorf("invalid fixed length %d: must be greater than 0", length)
+	}
+
+	return nil
+}
+
+func FixedTypeOf(n int) FixedType {
```
tanmayrauth, line 572: Any reason to panic here rather than return an error like GeometryTypeOf / GeographyTypeOf do? The two callers (this + arrow_utils.go:374 with dt.ByteWidth) pass runtime-derived lengths, so a panic turns unusual external input into a crash rather than a catchable error. Panic is fine if we're confident those lengths are always validated upstream — just want to confirm that's the case.

tanmayrauth (commented):
The parse-path validation looks right and catching the Atoi error is a nice bonus. One thing worth deciding on: FixedTypeOf now panics, and a couple of callers feed it runtime-derived lengths. Details inline.

types.go
```diff
@@ -558,7 +564,21 @@ func (m *MapType) UnmarshalJSON(b []byte) error {
...
+	if err := validateFixedLength(n); err != nil {
+		panic(fmt.Errorf("%w: %w", ErrInvalidArgument, err))
```
laskoviymishka: I think there's a real crash hiding here. `FixedLiteral.Type()` at literals.go:1124 is `return FixedTypeOf(len(f))`, and `FixedLiteral` is an exported `[]byte` whose `UnmarshalBinary` happily accepts empty input. So a zero-length literal — the zero value, `FixedLiteral(nil)`, or anything from `UnmarshalBinary([]byte{})` — now panics inside `.Type()`, which has no error return, so callers can't catch it.

This is the same failure #1405 fixed for `DecimalLiteral`. I'd mirror it: either change literals.go:1124 to `return FixedType{len: len(f)}` to bypass the validator, or make `UnmarshalBinary` reject empty data with an error so corruption gets caught through the normal channel (I'd lean toward the latter). Either way this needs handling before merge.

types.go
laskoviymishka, line 572: Making a public constructor panic is a real API choice, and I'm not sure it's the one we want here. `table/arrow_utils.go:374` calls `iceberg.FixedTypeOf(dt.ByteWidth)` — `ByteWidth` can be `0` for a degenerate Arrow schema, so that path goes from silently producing a zero-width type to crashing, and it's not updated in this PR.

I'd lean toward `FixedTypeOf(n int) (FixedType, error)`, or a `MustFixedTypeOf` alongside it à la `regexp.MustCompile`, so callers deriving `n` from external data have a channel to handle it. wdyt?

types.go
```diff
@@ -189,7 +189,13 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
 					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
 				}
 
-				n, _ := strconv.Atoi(matches[1])
+				n, err := strconv.Atoi(matches[1])
+				if err != nil {
+					return fmt.Errorf("%w: %s", ErrInvalidTypeString, err)
+				}
+				if err := validateFixedLength(n); err != nil {
```
laskoviymishka: I'm a little wary of gating this on the read path. The spec doesn't require `fixed` length > 0 — Java's `FixedType.ofLength` + `FIXED` regex, PyIceberg's `FixedType.__init__`, and iceberg-rust's `Fixed(u64)` all accept `fixed[0]` with no check. So after this, any table metadata written by one of those clients with a `fixed[0]` column fails to unmarshal here with `ErrInvalidTypeString`, i.e. an unreadable table.

I'd keep the validation on the construction/write path but drop it from `UnmarshalJSON`, unless we've confirmed via the community that the spec intends to prohibit `fixed[0]`. Do we know either way?

types.go
```diff
@@ -189,7 +189,13 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
 					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
 				}
 
-				n, _ := strconv.Atoi(matches[1])
+				n, err := strconv.Atoi(matches[1])
```
laskoviymishka, line 192: Worth noting the regex `(\d+)` already guarantees digits, so this `err` branch is only reachable via integer overflow (e.g. `fixed[99999999999999999999]`) — which is real, and better than the pre-PR behavior of silently mapping to `0`, but it's untested and surfaces a raw Go parse error rather than a domain message.

I'd add a `fixed[99999999999999999999]` case and wrap the error a bit more descriptively.

types.go
```diff
@@ -558,7 +564,21 @@ func (m *MapType) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-func FixedTypeOf(n int) FixedType { return FixedType{len: n} }
+func validateFixedLength(length int) error {
+	if length <= 0 {
+		return fmt.Errorf("invalid fixed length %d: must be greater than 0", length)
```
laskoviymishka: Small thing while we're here: this message has its own colon, so wrapped it reads `invalid argument: invalid fixed length 0: must be greater than 0` — three colon segments. I'd flatten it to `fixed length must be greater than 0, got %d`.

Also the `%w: %w` at line 577 is the only multi-`%w` in the package (everything in exprs.go uses a single `%w` sentinel + string detail), and since the inner error isn't a sentinel the second `%w` buys nothing for `errors.Is`. `panic(fmt.Errorf("%w: fixed length must be greater than 0, got %d", ErrInvalidArgument, n))` would keep it consistent.

types_test.go
```diff
@@ -77,6 +77,39 @@ func TestFixedType(t *testing.T) {
 	assert.Equal(t, "fixed[5]", typ.String())
 	assert.True(t, typ.Equals(iceberg.FixedTypeOf(5)))
 	assert.False(t, typ.Equals(iceberg.FixedTypeOf(6)))
+
+	t.Run("FixedTypeOf validates length", func(t *testing.T) {
+		assert.PanicsWithError(t, "invalid argument: invalid fixed length 0: must be greater than 0", func() {
```
laskoviymishka: Matching the exact panic string makes this brittle — any wording tweak (like the one above) breaks it. I'd assert on the sentinel instead: recover and `assert.ErrorIs(recovered, iceberg.ErrInvalidArgument)`.

types_test.go
```diff
@@ -77,6 +77,39 @@ func TestFixedType(t *testing.T) {
...
+	tests := []struct {
+		name string
+		data string
+	}{
+		{
+			name: "zero length",
+			data: `{"id": 1, "name": "f", "type": "fixed[0]", "required": true}`,
+		},
+		{
+			name: "invalid syntax",
```
laskoviymishka: `fixed[0]junk` fails the bracket regex and returns before ever reaching `validateFixedLength`, so this exercises the pre-existing path, not the new code, and the name is a bit misleading. I'd either rename it (it's really "malformed suffix") or swap in the overflow case from the comment above so we're actually covering the new branches.

laskoviymishka (changes requested):
Nice catch on the underlying bug — `FixedTypeOf` accepting a zero/negative length silently was worth closing. Almost there, but I'd hold this before merging. One blocking thing, plus a design call worth settling.

The blocker: `FixedLiteral.Type()` (literals.go:1124) is `return FixedTypeOf(len(f))`, and `FixedLiteral` is an exported `[]byte` whose `UnmarshalBinary` accepts empty input. So a zero-length literal now panics inside `.Type()`, which has no error return — callers can't recover. This is exactly what #1405 guarded for `DecimalLiteral`; I'd mirror that here (either `FixedType{len: len(f)}` at literals.go:1124, or reject empty data in `UnmarshalBinary`).

The design call: the spec doesn't mandate `fixed` length > 0, and Java/PyIceberg/Rust all accept `fixed[0]`. Gating it in `UnmarshalJSON` means metadata from another client with a `fixed[0]` column becomes unreadable here. I'd keep the validation on construction but leave the read path permissive unless we've confirmed the spec intends to prohibit it. Related: I'd also reconsider panicking from a public constructor — `arrow_utils.go:374` passes `dt.ByteWidth`, which can be `0`, and it isn't updated; a `(FixedType, error)` signature or a `Must` variant would be friendlier.

- handle/test the `strconv.Atoi` overflow case, and give it a domain-level message
- flatten the validation message to avoid the triple-colon wrap, and drop the unusual `%w: %w`
- assert on `ErrInvalidArgument` rather than the exact panic string
- the `fixed[0]junk` case fails the regex before hitting the new code — rename or swap it for the overflow input

Once the panic and the read-path question are sorted, happy to take another pass and approve.

laskoviymishka (approved):
Look good now!

PR #1405, opened 2026-07-05
fix(types): validate decimal parsing

fallintoplace:
## Why
Iceberg schema parsing currently accepted malformed decimal strings and invalid precision/scale combinations, which could surface as confusing errors later in Arrow/JSON/parquet paths.

## What changed
- Tightened decimal type parsing to require a fully anchored `decimal(<precision>,<scale>)` form.
- Added explicit numeric parsing and range validation for precision/scale.
- Enforced the same precision/scale validation in `DecimalTypeOf` and parser-backed creation paths.
- Kept behavior unchanged for valid decimals while rejecting malformed or unsafe inputs early.
- Added regression coverage for:
  - trailing garbage (`decimal(10,2)junk`)
  - negative/zero precision
  - scale greater than precision
  - non-integer precision/scale tokens

## Validation
- `go test . -run TestDecimalType -count=1`
- `go test ./table ./table/substrait -run "TestArrowToIceberg|TestToDecimalLiteralUsesBoundFieldPrecision|TestDecimalTypeInvalidParse" -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m`

### DIFF
--- a/literals_test.go
+++ b/literals_test.go
@@ -389,6 +389,14 @@ func TestDecimalToDecimalConversion(t *testing.T) {
 	assert.ErrorContains(t, err, "could not convert 34.11 to decimal(9, 3)")
 }
 
+func TestDecimalLiteralTypeDoesNotPanicForLargeScale(t *testing.T) {
+	lit := iceberg.DecimalLiteral{Scale: 10, Val: decimal128.FromI64(1234)}
+
+	assert.NotPanics(t, func() {
+		assert.NotNil(t, lit.Type())
+	})
+}
+
 func TestDecimalLiteralConversions(t *testing.T) {
 	n1 := iceberg.Decimal{Val: decimal128.FromI64(1234), Scale: 2}
 	n2 := iceberg.Decimal{Val: decimal128.FromI64(math.MaxInt32 + 1), Scale: 0}
--- a/types.go
+++ b/types.go
@@ -206,7 +206,7 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
 					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
 				}
 				if err := validateDecimalPrecisionScale(prec, scale); err != nil {
-					return fmt.Errorf("%w: %s", ErrInvalidTypeString, err)
+					return fmt.Errorf("%w: %w", ErrInvalidTypeString, err)
 				}
 
 				t.Type = DecimalType{precision: prec, scale: scale}
@@ -599,14 +599,11 @@ func validateDecimalPrecisionScale(precision, scale int) error {
 	if precision <= 0 {
 		return fmt.Errorf("invalid precision %d: must be greater than 0", precision)
 	}
-	if precision >= 40 {
-		return fmt.Errorf("invalid precision %d: must be less than 40", precision)
+	if precision > 38 {
+		return fmt.Errorf("invalid precision %d: must be less than or equal to 38", precision)
 	}
 	if scale < 0 {
 		return fmt.Errorf("invalid scale %d: must be greater than or equal to 0", scale)
-	}
-	if scale > precision {
-		return fmt.Errorf("invalid scale %d: must be less than or equal to precision %d", scale, precision)
 	}
 
 	return nil
--- a/types.go
+++ b/types.go
@@ -34,7 +34,7 @@ import (
 
 var (
 	regexFromBrackets = regexp.MustCompile(`^\w+\[(\d+)\]$`)
-	decimalRegex      = regexp.MustCompile(`decimal\(\s*(\d+)\s*,\s*(\d+)\s*\)`)
+	decimalRegex      = regexp.MustCompile(`^decimal\(\s*(\d+)\s*,\s*(\d+)\s*\)$`)
 	geometryRegex     = regexp.MustCompile(`(?i)^geometry\s*(?:\(\s*([^),]+?)\s*\))?$`)
 	geographyRegex    = regexp.MustCompile(`(?i)^geography\s*(?:\(\s*([^\s,)]+)\s*(?:,\s*(\w+)\s*)?\))?$`)
 )
@@ -197,8 +197,18 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
 					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
 				}
 
-				prec, _ := strconv.Atoi(matches[1])
-				scale, _ := strconv.Atoi(matches[2])
+				prec, err := strconv.Atoi(matches[1])
+				if err != nil {
+					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
+				}
+				scale, err := strconv.Atoi(matches[2])
+				if err != nil {
+					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
+				}
+				if err := validateDecimalPrecisionScale(prec, scale); err != nil {
+					return fmt.Errorf("%w: %s", ErrInvalidTypeString, err)
+				}
+
 				t.Type = DecimalType{precision: prec, scale: scale}
 			// note that geo type names are case insensitive but other type names are case sensitive.
 			// matches java behavior - this behavior is intentional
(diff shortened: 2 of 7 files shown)


types.go
```diff
@@ -578,9 +588,30 @@ func (f FixedType) String() string { return fmt.Sprintf("fixed[%d]", f.len) }
 func (f FixedType) primitive()     {}
 
 func DecimalTypeOf(prec, scale int) DecimalType {
+	if err := validateDecimalPrecisionScale(prec, scale); err != nil {
+		panic(fmt.Errorf("%w: %w", ErrInvalidArgument, err))
```
tanmayrauth, line 592: This panic regresses `DecimalLiteral.Type()`, which returns `DecimalTypeOf(9, d.Scale)` (literals.go:1254 — hardcoded precision 9 per #1028). Any decimal literal with scale > 9 now panics here because scale > precision, and Iceberg allows scale up to 38. I think that's why the tests in this PR had to drop `decimal(38,10)`→`decimal(38,9)` and `DecimalTypeOf(8,9)`→`DecimalTypeOf(8,2)` — those edits are working around the panic rather than the code being safe.
  
  It's also reachable outside substrait: exprs.go:521 does `lit.Type().Equals(b.field.Type)` in the literal-binding path, so filtering on a high-scale decimal column hits the same panic.
  
  Could we avoid panicking on this path? Options: have `DecimalLiteral.Type()` construct the struct directly (as before), or make `DecimalTypeOf` return an error / stay lenient and validate only on the parse path.

types.go
```diff
@@ -578,9 +588,30 @@ func (f FixedType) String() string { return fmt.Sprintf("fixed[%d]", f.len) }
...
+	}
+
 	return DecimalType{precision: prec, scale: scale}
 }
 
+func validateDecimalPrecisionScale(precision, scale int) error {
+	if precision <= 0 {
+		return fmt.Errorf("invalid precision %d: must be greater than 0", precision)
+	}
+	if precision >= 40 {
```
tanmayrauth: The spec (and Java) cap decimal precision at 38 — that's the max for 16-byte decimal128. `>= 40` still admits `decimal(39,x)`, which other engines will reject. Should this be `> 38`?

types.go
```diff
@@ -197,8 +197,18 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
 					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
 				}
 
-				prec, _ := strconv.Atoi(matches[1])
-				scale, _ := strconv.Atoi(matches[2])
+				prec, err := strconv.Atoi(matches[1])
```
tanmayrauth, line 200: nit: with the anchored `^decimal\(\s*(\d+)...\)$` regex the captures are always `\d+`, so `Atoi` here can only fail on integer overflow (e.g. `decimal(99999999999,2)`) — `validateDecimalPrecisionScale` does the real work. Fine to keep, maybe just a comment noting overflow is the only failure path.

tanmayrauth (commented):
The parse-path tightening is solid, but routing DecimalTypeOf through a panic regresses DecimalLiteral.Type() for any scale > 9, which is legal in Iceberg. Details inline.

types.go
```diff
@@ -578,9 +588,30 @@ func (f FixedType) String() string { return fmt.Sprintf("fixed[%d]", f.len) }
...
+	if precision <= 0 {
+		return fmt.Errorf("invalid precision %d: must be greater than 0", precision)
+	}
+	if precision > 38 {
+		return fmt.Errorf("invalid precision %d: must be less than or equal to 38", precision)
+	}
+	if scale < 0 {
+		return fmt.Errorf("invalid scale %d: must be greater than or equal to 0", scale)
+	}
+	if scale > precision {
```
laskoviymishka: I don't think we can enforce `scale > precision` here. The spec only constrains precision (must be 38 or less) — there's no rule tying scale to precision, and Java's `Types.DecimalType`, iceberg-rust, and PyIceberg all accept things like `decimal(5,7)`.

Once this lands, a table written by Java/Spark with scale > precision serializes fine but our `UnmarshalJSON` rejects it, so the table becomes unreadable here even read-only. I'd drop this guard and keep just precision in `(0,38]` and `scale >= 0`. wdyt?

types.go
```diff
@@ -578,9 +588,30 @@ func (f FixedType) String() string { return fmt.Sprintf("fixed[%d]", f.len) }
 func (f FixedType) primitive()     {}
 
 func DecimalTypeOf(prec, scale int) DecimalType {
+	if err := validateDecimalPrecisionScale(prec, scale); err != nil {
```
laskoviymishka, line 591: Panicking here makes `DecimalTypeOf` unsafe on data-derived input. `arrow_utils.go` calls `DecimalTypeOf(dec.GetPrecision(), dec.GetScale())` for incoming Arrow schemas, and Arrow permits scale > precision (`Decimal32Type{Precision:8, Scale:9}` is valid), so reading such a batch now takes down the process instead of returning an error. The `Scale:9→Scale:2` edit in `arrow_utils_test.go` is the tell that this path is live.

I'd have `DecimalTypeOf` return `(DecimalType, error)` so `arrow_utils.go` can propagate, rather than validate-by-panic on library input.

types.go
```diff
@@ -197,8 +197,18 @@ func (t *typeIFace) UnmarshalJSON(b []byte) error {
...
+				prec, err := strconv.Atoi(matches[1])
+				if err != nil {
+					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
+				}
+				scale, err := strconv.Atoi(matches[2])
+				if err != nil {
+					return fmt.Errorf("%w: %s", ErrInvalidTypeString, typename)
+				}
+				if err := validateDecimalPrecisionScale(prec, scale); err != nil {
+					return fmt.Errorf("%w: %s", ErrInvalidTypeString, err)
```
laskoviymishka: Small thing: this interpolates a real error with `%s`, which drops its wrap chain — callers can't `errors.As`/`Is` back to the cause. Since Go 1.20 supports multiple `%w` (and the `DecimalTypeOf` panic just above already uses double-`%w`), I'd make this `%w: %w`.

literals_test.go
```diff
@@ -389,6 +389,15 @@ func TestDecimalToDecimalConversion(t *testing.T) {
 	assert.ErrorContains(t, err, "could not convert 34.11 to decimal(9, 3)")
 }
 
+func TestDecimalLiteralTypeSupportsLargeScale(t *testing.T) {
+	lit := iceberg.DecimalLiteral{Scale: 10, Val: decimal128.FromI64(1234)}
+
+	assert.NotPanics(t, func() {
+		typeStr := lit.Type().String()
+		assert.Equal(t, "decimal(9, 10)", typeStr)
```
laskoviymishka: This pins `decimal(9, 10)` — an invalid type — as a guarantee, but the name frames it as a feature. If #1028 ever gets fixed and `Type()` stops hardcoding precision 9, this test breaks even though the fix is correct.

I'd either drop the exact-string assert or rename it as a known-wart characterization and reference #1028 in the body, so a future maintainer knows it's intentional.

types_test.go
```diff
@@ -86,6 +86,57 @@ func TestDecimalType(t *testing.T) {
...
+			name: "precision zero",
+			data: `{"id": 1, "name": "d", "type": "decimal(0,2)", "required": true}`,
+		},
+		{
+			name: "scale exceeds precision",
+			data: `{"id": 1, "name": "d", "type": "decimal(10,11)", "required": true}`,
+		},
+		{
+			name: "negative scale",
+			data: `{"id": 1, "name": "d", "type": "decimal(10,-1)", "required": true}`,
```
laskoviymishka, line 120: This case doesn't actually reach `validateDecimalPrecisionScale` — the anchored `\d+` in `decimalRegex` rejects the `-` first, so it passes via the regex mismatch, not the `scale < 0` branch. That branch is only reachable through `DecimalTypeOf`, which this test doesn't touch.

Worth a note, or point it at `DecimalTypeOf(10, -1)` if you want to cover the branch it's named for.

laskoviymishka (changes requested):
The extra validation and the anchored regex are a real improvement — rejecting `decimal(10,2)junk` and surfacing the `Atoi` failures instead of swallowing them is the right call.

One blocking thing before merge though: the `scale > precision` guard in `validateDecimalPrecisionScale` is stricter than the spec and every reference client. The spec only constrains precision (38 or less); Java, iceberg-rust, and PyIceberg all accept scale > precision. So a table written by Java/Spark with e.g. `decimal(5,7)` serializes fine but our `UnmarshalJSON` now rejects it, and the same guard turns `arrow_utils.go`'s `DecimalTypeOf` call into a process panic on valid Arrow input. I'd drop that one check and keep precision in `(0,38]` and `scale >= 0`.

Left a few smaller notes inline — the panic-vs-error shape of `DecimalTypeOf`, the `%w` wrapping, and the two tests that pin `decimal(9,10)` / don't quite exercise the negative-scale path.

Drop the `scale > precision` constraint and this is good to land.

laskoviymishka (approved):
:shipit:

PR #1406, opened 2026-07-05
fix(table): validate encryption-key metadata fields

fallintoplace:
## Why
`add-encryption-key` updates accepted incomplete key payloads and could persist invalid metadata (empty key-id or empty encrypted metadata).

## What changed
- Validate `EncryptionKey` whenever an encryption-key add update is applied (both through `MetadataBuilder.AddEncryptionKey` and the update object path).
- Return `ErrInvalidArgument` when required encryption key fields are missing or empty.
- Add coverage for empty `key-id`, empty `encrypted-key-metadata`, and malformed update payloads that decode with missing fields.

## Testing
- `go test ./table -run 'TestAddEncryptionKeyUpdate_Apply_RejectsMissingKeyID|TestAddEncryptionKeyUpdate_Apply_RejectsMissingEncryptedKeyMetadata|TestAddEncryptionKeyUpdate_UnmarshalMissingFields_ApplyRejects' -count=1`

### DIFF
--- a/table/encryption.go
+++ b/table/encryption.go
@@ -18,7 +18,9 @@
 package table
 
 import (
+	"encoding/base64"
 	"errors"
+	"fmt"
 	"maps"
 	"strings"
 )
@@ -31,13 +33,20 @@ type EncryptionKey struct {
 	Properties           map[string]string `json:"properties,omitempty"`
 }
 
-func (e EncryptionKey) Validate() error {
-	if strings.TrimSpace(e.KeyID) == "" {
+func (e EncryptionKey) validate() error {
+	trimmedKeyID := strings.TrimSpace(e.KeyID)
+	if trimmedKeyID == "" {
 		return errors.New("encryption key-id must be non-empty")
 	}
+	if trimmedKeyID != e.KeyID {
+		return errors.New("encryption key-id must not have leading or trailing whitespace")
+	}
 
-	if strings.TrimSpace(e.EncryptedKeyMetadata) == "" {
+	if e.EncryptedKeyMetadata == "" {
 		return errors.New("encrypted key metadata must be non-empty")
+	}
+	if _, err := base64.StdEncoding.DecodeString(e.EncryptedKeyMetadata); err != nil {
+		return fmt.Errorf("encrypted key metadata must be valid base64: %w", err)
 	}
 
 	return nil
--- a/table/encryption.go
+++ b/table/encryption.go
@@ -17,7 +17,11 @@
 
 package table
 
-import "maps"
+import (
+	"fmt"
+	"maps"
+	"strings"
+)
 
 // EncryptionKey represents an encryption key stored in table metadata (V3+).
 type EncryptionKey struct {
@@ -25,6 +29,18 @@ type EncryptionKey struct {
 	EncryptedKeyMetadata string            `json:"encrypted-key-metadata"`
 	EncryptedByID        *string           `json:"encrypted-by-id,omitempty"`
 	Properties           map[string]string `json:"properties,omitempty"`
+}
+
+func (e EncryptionKey) Validate() error {
+	if strings.TrimSpace(e.KeyID) == "" {
+		return fmt.Errorf("encryption key-id must be non-empty")
+	}
+
+	if strings.TrimSpace(e.EncryptedKeyMetadata) == "" {
+		return fmt.Errorf("encrypted key metadata must be non-empty")
+	}
+
+	return nil
 }
 
 func (e EncryptionKey) Equals(other EncryptionKey) bool {
--- a/table/encryption.go
+++ b/table/encryption.go
@@ -42,11 +42,12 @@ func (e EncryptionKey) validate() error {
 		return errors.New("encryption key-id must not have leading or trailing whitespace")
 	}
 
-	if e.EncryptedKeyMetadata == "" {
-		return errors.New("encrypted key metadata must be non-empty")
-	}
-	if _, err := base64.StdEncoding.DecodeString(e.EncryptedKeyMetadata); err != nil {
+	decodedMetadata, err := base64.StdEncoding.DecodeString(e.EncryptedKeyMetadata)
+	if err != nil {
 		return fmt.Errorf("encrypted key metadata must be valid base64: %w", err)
+	}
+	if len(decodedMetadata) == 0 {
+		return errors.New("encrypted key metadata must be non-empty")
 	}
 
 	return nil
--- a/table/encryption.go
+++ b/table/encryption.go
@@ -18,7 +18,7 @@
 package table
 
 import (
-	"fmt"
+	"errors"
 	"maps"
 	"strings"
 )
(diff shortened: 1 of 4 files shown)


table/updates.go
```diff
@@ -772,6 +772,10 @@ func NewAddEncryptionKeyUpdate(key EncryptionKey) *addEncryptionKeyUpdate {
 }
 
 func (u *addEncryptionKeyUpdate) Apply(builder *MetadataBuilder) error {
+	if err := u.EncryptionKey.Validate(); err != nil {
```
tanmayrauth: This Validate() looks redundant — Apply just delegates to builder.AddEncryptionKey on the next line, and that already calls key.Validate() (metadata.go:1341), so an invalid key hits the same ErrInvalidArgument either way. Your TestAddEncryptionKeyUpdate_UnmarshalMissingFields_ApplyRejects still passes through AddEncryptionKey without this check. Any reason to keep both? Either drop this one or add a short  comment on why the update path validates independently, so the two don't drift.

tanmayrauth (commented):
Clean fix, and nice that validation only sits on the write/update path so existing metadata still loads. One redundancy question inline.

zeroshade (approved):
LGTM — `EncryptionKey.Validate()` (non-empty key-id + encrypted-key-metadata) is wired into `AddEncryptionKey`, with good coverage including the JSON-deserialized incomplete-key case. Reads cleanly — validation happens once, in `AddEncryptionKey`. Separately: #1397 appears to be a mislabeled duplicate of this PR (same three-file diff); I've suggested closing it in favor of this one over there.

table/encryption.go
```diff
@@ -27,6 +31,18 @@ type EncryptionKey struct {
 	Properties           map[string]string `json:"properties,omitempty"`
 }
 
+func (e EncryptionKey) Validate() error {
```
laskoviymishka: `Validate()` is exported but nothing outside the package calls it — callers hand keys to `AddEncryptionKey` and let it validate. Every other validation surface in here (`commonMetadata.validate`, `metadataV3.validate`) is unexported, so this widens the public API for something internal, and the rules become a breaking change to touch later.

The errors it returns are also plain `errors.New`, which makes the `fmt.Errorf("%w: %w", iceberg.ErrInvalidArgument, err)` at the call site a bit of a dead end — the second `%w` promises `errors.Is` matchability on a value no caller can reference.

I'd either unexport it to `validate()`, or if there's a reason to keep it public, have it wrap with `ErrInvalidArgument` itself and drop the re-wrap. wdyt?

table/encryption.go
```diff
@@ -27,6 +31,18 @@ type EncryptionKey struct {
...
+	if strings.TrimSpace(e.KeyID) == "" {
+		return errors.New("encryption key-id must be non-empty")
+	}
+
+	if strings.TrimSpace(e.EncryptedKeyMetadata) == "" {
```
laskoviymishka: `TrimSpace` on `EncryptedKeyMetadata` doesn't really buy us anything — this field is a base64 blob, and whitespace is never valid base64, so the check collapses to `== ""` in practice.

If we actually want to validate the field the way the spec describes it (base64-encoded), I'd do a real decode check — `base64.StdEncoding.DecodeString(...)` and reject on error. That catches genuinely malformed values, and it matches iceberg-rust, which hard-errors decoding this field on read — so a table we write with a non-base64 value would be unreadable there.

Separately on `KeyID`: `TrimSpace` there is defensible, but the dedup loop just below compares raw `k.KeyID == key.KeyID`, so `"  key  "` and `" key "` land as distinct entries and you can't `RemoveEncryptionKey` with the trimmed form. If we're going to trim for validation, I'd store the trimmed value too (or just reject padded ids). Whichever we pick, worth a test pinning it. wdyt?

table/updates_test.go
```diff
@@ -884,6 +884,37 @@ func TestAddEncryptionKeyUpdate_Apply_RejectsV2(t *testing.T) {
 	assert.Contains(t, err.Error(), "format version 3")
 }
 
+func TestAddEncryptionKeyUpdate_Apply_RejectsMissingKeyID(t *testing.T) {
```
laskoviymishka, line 887: All three cases go through `NewAddEncryptionKeyUpdate(key).Apply` — none calls `b.AddEncryptionKey` directly, which is the exported, documented entry point the validation actually lives on. If `Apply` ever gets refactored to not route through `AddEncryptionKey`, these still pass while the builder contract silently breaks. I'd add one case hitting `b.AddEncryptionKey(key)` directly.

table/updates_test.go
```diff
@@ -884,6 +884,37 @@ func TestAddEncryptionKeyUpdate_Apply_RejectsV2(t *testing.T) {
...
+
+func TestAddEncryptionKeyUpdate_UnmarshalMissingFields_ApplyRejects(t *testing.T) {
+	data := []byte(`[{"action":"add-encryption-key","encryption-key":{"key-id":"my-key"}}]`)
+
+	var updates Updates
+	require.NoError(t, json.Unmarshal(data, &updates))
+	require.Len(t, updates, 1)
+
+	err := updates[0].Apply(buildFromBaseV3(t))
+	require.ErrorIs(t, err, iceberg.ErrInvalidArgument)
```
laskoviymishka, line 949: The other two tests assert both `ErrorIs` and a `Contains` on the field name — this one only checks `ErrorIs`. If the error ever fires for a different reason (a future version guard, say), this still passes while testing the wrong thing. I'd add `assert.Contains(t, err.Error(), "key-id")` to pin that it's the missing field being rejected.

Minor while we're here: the three new tests use `assert.ErrorIs` where the rest of the file uses `require.ErrorIs` — worth matching.

laskoviymishka (approved):
Nice — this is a clean, focused fix and I don't have any objections to it landing.

A few small things I'd fold in while we're here, none blocking. The main one: `TrimSpace` on the encrypted-key-metadata field doesn't really validate anything, since that field is base64 and whitespace is never valid base64 — the check just collapses to non-empty. If we want to actually validate it per the spec, a base64 decode-check is the real guard, and it keeps us readable in iceberg-rust, which hard-errors on a non-base64 value here.

The rest is polish: `Validate()` is exported for something only used internally (and the `%w: %w` wrap at the call site can't be matched by `errors.Is`), and the tests would be stronger with a direct `b.AddEncryptionKey` case plus a message assertion on the JSON-path test. Specifics are inline.

Not for this PR, but worth a follow-up: the read path (`metadataV3.validate`) doesn't run these checks, so pre-existing invalid metadata still loads.

None of this blocks — fix what makes sense and I'm happy to approve.

zeroshade (changes requested):
Following up — I approved earlier, but a Low-severity edge case turned up on re-review, so switching to request-changes:

**Low — whitespace-only key metadata passes validation** (`table/encryption.go:45`)
`EncryptedKeyMetadata` only checks the raw string for `""`, but Go's base64 decoder ignores `\r`/`\n`, so a value like `"\n"` passes validation while decoding to empty metadata.
*Fix:* store the decoded bytes and reject `len(decoded) == 0`, or reject whitespace-only metadata before decoding.

zeroshade (approved):

PR #1411, opened 2026-07-05
fix(sql): hide namespace exists sentinel property

fallintoplace:
## Motivation

SQL catalogs persist an internal `exists=true` row for namespaces that have no user properties. Returning that row leaks an implementation detail through the public namespace-properties API, while allowing callers to mutate it can make a sentinel-only namespace disappear.

## Changes

- hide the internal `exists` sentinel from `LoadNamespaceProperties`
- reject the reserved key during namespace creation, updates, and removals with `iceberg.ErrInvalidArgument`
- preserve the sentinel internally for namespaces created without user properties
- cover sentinel-only and user-property namespaces on the in-memory and SQLite backends

Removal is intentionally rejected rather than treated as a missing-property no-op because deleting the only sentinel row would remove the namespace itself.

## Testing

- `go test ./catalog/sql`
- full CI

### DIFF
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -225,12 +225,18 @@ func init() {
 
 var _ catalog.Catalog = (*Catalog)(nil)
 
+const namespaceExistsProperty = "exists"
+
 var (
-	minimalNamespaceProps = iceberg.Properties{"exists": "true"}
+	minimalNamespaceProps = iceberg.Properties{namespaceExistsProperty: "true"}
 
 	dialects  = map[SupportedDialect]schema.Dialect{}
 	dialectMx sync.Mutex
 )
+
+func isReservedNamespaceProperty(key string) bool {
+	return key == namespaceExistsProperty
+}
 
 func createDialect(d SupportedDialect) (schema.Dialect, error) {
 	switch d {
@@ -1151,6 +1157,12 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 		return err
 	}
 
+	for key := range props {
+		if isReservedNamespaceProperty(key) {
+			return fmt.Errorf("%w: cannot create namespace with reserved namespace property %q", iceberg.ErrInvalidArgument, key)
+		}
+	}
+
 	_, exists, err := c.resolveNamespaceKey(ctx, namespace)
 	if err != nil {
 		return err
@@ -1248,11 +1260,31 @@ func (c *Catalog) LoadNamespaceProperties(ctx context.Context, namespace table.I
 
 		result := make(iceberg.Properties)
 		for _, p := range props {
+			if isReservedNamespaceProperty(p.PropertyKey) {
+				continue
+			}
 			result[p.PropertyKey] = p.PropertyValue.String
 		}
 
 		return result, nil
 	})
+}
+
+func validateNamespacePropertyUpdates(removals []string, updates iceberg.Properties) error {
+	for key := range updates {
+		if isReservedNamespaceProperty(key) {
+			return fmt.Errorf("%w: cannot update reserved namespace property %q", iceberg.ErrInvalidArgument, key)
+		}
+	}
+
+	for _, key := range removals {
+		if isReservedNamespaceProperty(key) {
+			// Removing the marker would make a namespace without user properties disappear.
+			return fmt.Errorf("%w: cannot remove reserved namespace property %q", iceberg.ErrInvalidArgument, key)
+		}
+	}
+
+	return nil
 }
 
 func (c *Catalog) ListTables(ctx context.Context, namespace table.Identifier) iter.Seq2[table.Identifier, error] {
@@ -1374,6 +1406,9 @@ func getUpdatedPropsAndUpdateSummary(currentProps iceberg.Properties, removals [
 func (c *Catalog) UpdateNamespaceProperties(ctx context.Context, namespace table.Identifier, removals []string, updates iceberg.Properties) (catalog.PropertiesUpdateSummary, error) {
 	var summary catalog.PropertiesUpdateSummary
 	if err := checkValidNamespace(namespace); err != nil {
+		return summary, err
+	}
+	if err := validateNamespacePropertyUpdates(removals, updates); err != nil {
 		return summary, err
 	}
 
--- a/catalog/sql/sql_test.go
+++ b/catalog/sql/sql_test.go
@@ -1907,7 +1907,102 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
 		s.Require().NoError(tt.cat.CreateNamespace(ctx, tt.namespace, nil))
 		props, err := tt.cat.LoadNamespaceProperties(ctx, tt.namespace)
 		s.Require().NoError(err)
-		s.Equal(iceberg.Properties{"exists": "true"}, props)
+		s.Empty(props)
+	}
+}
+
+func (s *SqliteCatalogTestSuite) TestCreateNamespaceRejectsReservedProperty() {
+	tests := []struct {
+		cat       *sqlcat.Catalog
+		namespace table.Identifier
+	}{
+		{s.getCatalogMemory(), table.Identifier{databaseName()}},
+		{s.getCatalogSqlite(), strings.Split(hiearchicalNamespaceName(), ".")},
+	}
+
+	ctx := context.Background()
+	for _, tt := range tests {
+		err := tt.cat.CreateNamespace(ctx, tt.namespace, iceberg.Properties{"exists": "false"})
+		s.ErrorIs(err, iceberg.ErrInvalidArgument)
+		s.ErrorContains(err, "cannot create namespace with reserved namespace property")
+		s.ErrorContains(err, `"exists"`)
+
+		_, err = tt.cat.UpdateNamespaceProperties(ctx, tt.namespace, []string{"exists"}, nil)
+		s.ErrorIs(err, iceberg.ErrInvalidArgument)
+
+		_, err = tt.cat.UpdateNamespaceProperties(ctx, tt.namespace, nil, iceberg.Properties{"exists": "false"})
+		s.ErrorIs(err, iceberg.ErrInvalidArgument)
+
+		exists, err := tt.cat.CheckNamespaceExists(ctx, tt.namespace)
+		s.Require().NoError(err)
+		s.False(exists)
+	}
+}
+
+func (s *SqliteCatalogTestSuite) TestNamespacePropertiesCannotUpdateSentinel() {
+	tests := []struct {
+		name          string
+		cat           *sqlcat.Catalog
+		namespace     table.Identifier
+		createProps   iceberg.Properties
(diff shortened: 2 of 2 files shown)


catalog/sql/sql.go
```diff
@@ -672,13 +677,30 @@ func (c *Catalog) LoadNamespaceProperties(ctx context.Context, namespace table.I
...
 }
 
+func validateNamespacePropertyUpdates(removals []string, updates iceberg.Properties) error {
+	if _, ok := updates[namespaceExistsProperty]; ok {
+		return fmt.Errorf("%w: cannot update reserved namespace property %q", iceberg.ErrInvalidArgument, namespaceExistsProperty)
+	}
+
+	for _, key := range removals {
+		if isReservedNamespaceProperty(key) {
+			return fmt.Errorf("%w: cannot remove reserved namespace property %q", iceberg.ErrInvalidArgument, namespaceExistsProperty)
```
tanmayrauth: Small one: the loop binds `key` but this formats `namespaceExistsProperty`. Same value today since only "exists" is reserved, but if the reserved set ever grows this will name the wrong key in the error — probably want `key` here.

catalog/sql/sql_test.go
```diff
@@ -1129,7 +1129,35 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+		s.Empty(props)
+	}
+}
+
+func (s *SqliteCatalogTestSuite) TestNamespacePropertiesCannotUpdateSentinel() {
+	tests := []struct {
+		cat       *sqlcat.Catalog
+		namespace table.Identifier
+	}{
+		{s.getCatalogMemory(), table.Identifier{databaseName()}},
```
tanmayrauth, line 1919: Since this seeds `{"comment": "baseline"}`, no sentinel row is actually written, so the guard is only exercised at the argument layer. Could you add a case that creates a namespace with empty props (which does insert the `exists=true` row) and then asserts remove/update of `exists` is still rejected and the namespace loads empty? That covers the real-world path where the sentinel is present.

tanmayrauth (commented):
Core change reads cleanly and the tests pass. One thing worth deciding before merge: this reserves `exists` on the read path (LoadNamespaceProperties) and the update path (UpdateNamespaceProperties), but CreateNamespace — untouched here — still inserts whatever the caller passes. So `CreateNamespace(ns, {"exists": "false", ...})` persists a sentinel the new code then permanently hides on read and refuses to update or remove. Not asking you to change  CreateNamespace in this PR necessarily, but it'd be good to confirm whether leaving that path unguarded is intended, since the three paths no longer agree on what "reserved" means. Two small inline notes below.

catalog/sql/sql.go
zeroshade: This formats the constant `namespaceExistsProperty` (always `exists`) instead of the loop variable `key`. They're identical today because `isReservedNamespaceProperty` only matches `exists`, but once the reserved set grows this will report the wrong property name for any reserved-key removal. Use `key` here.

zeroshade (changes requested):
One latent bug in the removal-branch error message (inline), plus a small coverage gap: the new test always seeds the namespace with non-empty user props, so the `exists` sentinel row is never actually persisted. Consider adding a case that creates a namespace with nil/empty properties (which does write the sentinel) and asserts it's hidden by `LoadNamespaceProperties` and rejected on remove — that exercises the sentinel-present path.

catalog/sql/sql.go
```diff
@@ -224,12 +224,17 @@ func init() {
 var _ catalog.Catalog = (*Catalog)(nil)
 
 var (
-	minimalNamespaceProps = iceberg.Properties{"exists": "true"}
+	namespaceExistsProperty = "exists"
```
laskoviymishka: Since `namespaceExistsProperty` is a fixed protocol value we share with Java and PyIceberg, I'd make it a `const` rather than a `var` — nothing should ever reassign it. It has to move out of this block though, since `minimalNamespaceProps` is a map and can't be const:

```go
const namespaceExistsProperty = "exists"

var (
	minimalNamespaceProps = iceberg.Properties{namespaceExistsProperty: "true"}
	...
)
```

catalog/sql/sql.go
```diff
@@ -1111,13 +1116,30 @@ func (c *Catalog) LoadNamespaceProperties(ctx context.Context, namespace table.I
...
+				continue
+			}
 			result[p.PropertyKey] = p.PropertyValue.String
 		}
 
 		return result, nil
 	})
 }
 
+func validateNamespacePropertyUpdates(removals []string, updates iceberg.Properties) error {
```
laskoviymishka, line 1273: We reject the sentinel on update and remove here, but `CreateNamespace` still writes caller props verbatim — nothing guards `exists` on the way in.

So a caller can do `CreateNamespace(ctx, ns, {"exists": "false"})` and then they're stuck: Load filters the key out, and this validation rejects any attempt to update or remove it. The three paths don't agree on what "reserved" means, and that leaves the namespace in a state the public API can't get out of.

I'd guard create with the same `isReservedNamespaceProperty` predicate — either reject reserved keys up front, or strip/overwrite them the way Java does (it always forces `exists=true`). tanmayrauth raised this too; I think it's the one thing worth settling before merge. wdyt?

catalog/sql/sql.go
```diff
@@ -1111,13 +1116,30 @@ func (c *Catalog) LoadNamespaceProperties(ctx context.Context, namespace table.I
...
+	if _, ok := updates[namespaceExistsProperty]; ok {
```
laskoviymishka: The removals loop below goes through `isReservedNamespaceProperty`, but this update check reaches into the map directly. Same result today, but if a second reserved key ever gets added to the predicate this branch silently misses it — I'd range over `updates` and call the helper here too, so there's one source of truth.

catalog/sql/sql.go
```diff
@@ -1111,13 +1116,30 @@ func (c *Catalog) LoadNamespaceProperties(ctx context.Context, namespace table.I
...
+		return fmt.Errorf("%w: cannot update reserved namespace property %q", iceberg.ErrInvalidArgument, namespaceExistsProperty)
+	}
+
+	for _, key := range removals {
+		if isReservedNamespaceProperty(key) {
```
laskoviymishka, line 1281: One thing worth deciding: Java's `removeProperties` doesn't guard `exists` at all — it lets you remove it. And the removals contract here normally routes unknown keys to the Missing set rather than erroring, so a caller who never saw `exists` from Load now gets a hard failure for removing a key they didn't know existed.

Blocking it is arguably safer (removing `exists` on a sentinel-only namespace would orphan it), so I'm fine keeping the guard — but I'd either drop `exists` into a silent no-op to match the removals contract, or keep the error and add a comment that we're intentionally stricter than Java here. wdyt?

catalog/sql/sql.go
```diff
@@ -1241,6 +1263,10 @@ func (c *Catalog) UpdateNamespaceProperties(ctx context.Context, namespace table
 		return summary, err
 	}
 
+	if err := validateNamespacePropertyUpdates(removals, updates); err != nil {
```
laskoviymishka, line 1411: This runs after `LoadNamespaceProperties`, so a bad `exists` in removals costs a DB round-trip before we return the arg error. Argument validation should come before the side effects — I'd move this to the top of `UpdateNamespaceProperties`, ahead of the load.

catalog/sql/sql_test.go
```diff
@@ -1882,7 +1882,71 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+			cat:           s.getCatalogSqlite(),
+			namespace:     strings.Split(hiearchicalNamespaceName(), "."),
+			createProps:   nil,
+			expectedProps: nil,
+		},
+	}
+
+	ctx := context.Background()
+	for _, tt := range tests {
+		tt := tt
```
laskoviymishka: `tt := tt` is a no-op on Go 1.22+ (go.mod is on 1.25) — loop vars are per-iteration now, so this can go.

catalog/sql/sql_test.go
```diff
@@ -1882,7 +1882,71 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+			_, err := tt.cat.UpdateNamespaceProperties(ctx, tt.namespace, []string{"exists"}, nil)
+			s.ErrorIs(err, iceberg.ErrInvalidArgument)
+			s.ErrorContains(err, "cannot remove reserved namespace property")
+			s.ErrorContains(err, `"exists"`)
+
+			_, err = tt.cat.UpdateNamespaceProperties(ctx, tt.namespace, nil, iceberg.Properties{"exists": "false"})
+			s.ErrorIs(err, iceberg.ErrInvalidArgument)
+			s.ErrorContains(err, "cannot update reserved namespace property")
+
+			props, err := tt.cat.LoadNamespaceProperties(ctx, tt.namespace)
```
laskoviymishka, line 1994: Could we also assert the namespace is still intact after these rejected updates? A quick `s.True(tt.cat.CheckNamespaceExists(ctx, tt.namespace))` here would confirm the sentinel row actually survived — right now the test proves the calls fail but not that existence state is preserved.

laskoviymishka (changes requested):
The read-side filter and the update/remove rejection are the right shape: hiding the sentinel on Load and refusing to let callers mutate it directly is exactly what this should do, and the new test covers the backend × user-prop combinations well.

I'd hold this before merging though, and tanmayrauth's point about `CreateNamespace` is the real one. We now reject `exists` on update and remove, but create still writes caller props verbatim — so `CreateNamespace(ctx, ns, {"exists": "false"})` goes straight through, and after that the namespace is stuck: Load filters the key out, and every update or remove that touches it is rejected. The three paths don't agree on what "reserved" means, and the gap is only reachable through the public API. I'd close it by guarding create with the same predicate — reject reserved keys, or strip/overwrite the way Java forces `exists=true`.

A couple of smaller things I'd want before merge:

- guard the sentinel on create the same way we do on update/remove
- decide whether removing `exists` should be a silent no-op (matches Java + the removals contract) or a documented, intentionally-stricter error
- make `namespaceExistsProperty` a const, and use `isReservedNamespaceProperty` in the update branch too so there's one source of truth

The rest is inline. Once the create path lines up with the other two, happy to take another pass.

zeroshade (approved):
Both points from July resolved:

- The reserved-property error now formats the actual `key`, not the `namespaceExistsProperty` constant, so it stays correct as the reserved set grows (`validateNamespacePropertyUpdates`).
- The coverage gap is closed — `TestNamespacePropertiesCannotUpdateSentinel` includes `"… sentinel only"` cases created with `nil` props (which actually persists the `exists` row), asserting it's hidden by `LoadNamespaceProperties`, rejected on remove *and* update, and that the namespace still reports as existing.

Clean refactor into `isReservedNamespaceProperty`/`validateNamespacePropertyUpdates`. LGTM.

(Note: not merging from my side — @laskoviymishka still has an open change request on this one.)

laskoviymishka (approved):

PR #1412, opened 2026-07-05
fix(table): include orphan file sizes in cleanup results

fallintoplace:
## Summary
- Add per-orphan file sizes to table orphan cleanup result as .
- Populate file sizes during scan so  retains size metadata, not just paths.
- Map sizes into CLI output entries and render per-file size in text output.
- Extend  JSON output to always include  and add regression tests for table/CLI.

## Testing
- go test ./...

### DIFF
--- a/cmd/iceberg/clean_orphan_files.go
+++ b/cmd/iceberg/clean_orphan_files.go
@@ -94,14 +94,32 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
 }
 
 func buildCleanOrphanFilesResult(tbl *table.Table, result table.OrphanCleanupResult, dryRun bool) CleanOrphanFilesResult {
-	files := result.OrphanFileLocations
-	if !dryRun {
-		files = result.DeletedFiles
-	}
+	var entries []OrphanFileEntry
+	if dryRun {
+		entries = make([]OrphanFileEntry, 0, len(result.OrphanFiles))
+		for _, f := range result.OrphanFiles {
+			entries = append(entries, OrphanFileEntry{
+				Path:      f.Path,
+				SizeBytes: f.SizeBytes,
+			})
+		}
+	} else {
+		deletedSet := make(map[string]struct{}, len(result.DeletedFiles))
+		for _, f := range result.DeletedFiles {
+			deletedSet[f] = struct{}{}
+		}
 
-	entries := make([]OrphanFileEntry, 0, len(files))
-	for _, f := range files {
-		entries = append(entries, OrphanFileEntry{Path: f})
+		entries = make([]OrphanFileEntry, 0, len(result.DeletedFiles))
+		for _, f := range result.OrphanFiles {
+			if _, ok := deletedSet[f.Path]; !ok {
+				continue
+			}
+
+			entries = append(entries, OrphanFileEntry{
+				Path:      f.Path,
+				SizeBytes: f.SizeBytes,
+			})
+		}
 	}
 
 	return CleanOrphanFilesResult{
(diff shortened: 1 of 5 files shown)


table/orphan_cleanup.go
```diff
@@ -161,11 +161,17 @@ func WithEqualAuthorities(authorities map[string]string) OrphanCleanupOption {
 
 type OrphanCleanupResult struct {
 	OrphanFileLocations []string
+	OrphanFiles         []OrphanFile
```
tanmayrauth: `OrphanFiles []OrphanFile` now carries the path *and* size for every orphan, while `OrphanFileLocations []string` above it carries just the paths — so the paths are duplicated across the two fields. Keeping `OrphanFileLocations` for backward compat is the right call, but could you add a one-line doc comment noting it's retained for compatibility and that `OrphanFiles` is the richer form? Otherwise a future reader has to diff the two to figure out why both
  exist.

cmd/iceberg/clean_orphan_files.go
```diff
@@ -99,9 +99,17 @@ func buildCleanOrphanFilesResult(tbl *table.Table, result table.OrphanCleanupRes
...
+	for _, f := range result.OrphanFiles {
+		orphanSizes[f.Path] = f.SizeBytes
+	}
+
 	entries := make([]OrphanFileEntry, 0, len(files))
 	for _, f := range files {
-		entries = append(entries, OrphanFileEntry{Path: f})
+		entries = append(entries, OrphanFileEntry{
+			Path:      f,
+			SizeBytes: orphanSizes[f],
```
tanmayrauth: `SizeBytes: orphanSizes[f]` — in the non-dry-run branch `files` comes from `result.DeletedFiles`, but this size map is built from `result.OrphanFiles`. Any deleted path not present in `OrphanFiles` would silently render as `0 B` rather than being caught. Today `DeletedFiles ⊆ OrphanFiles` so it's safe, but that's an implicit invariant — worth a short comment, or build the deleted entries by filtering `OrphanFiles` against the deleted set so the size always
  stays attached to its path.

cmd/iceberg/maintenance.go
```diff
@@ -172,7 +172,7 @@ type CleanOrphanFilesResult struct {
 
 type OrphanFileEntry struct {
 	Path      string `json:"path"`
-	SizeBytes int64  `json:"size_bytes,omitempty"`
+	SizeBytes int64  `json:"size_bytes"`
```
tanmayrauth, line 175: Dropping `omitempty` means a genuinely 0-byte orphan now always serializes `"size_bytes":0` instead of omitting the field. I think that's the right move for a stable output schema (and it matches `PartitionStatsEntry.SizeBytes` at line 146, which also has no omitempty) — just flagging it as an observable JSON change for anyone parsing the output.

tanmayrauth (commented):
LGTM, clean change and the table-level test covers the real path. Just a couple of small things around the CLI size lookup and the retained `OrphanFileLocations` field. Details inline.

table/orphan_cleanup.go
zeroshade: nit: `OrphanFiles` and `OrphanFileLocations` carry duplicate path data. A short doc comment noting `OrphanFileLocations` is retained for backward compatibility while `OrphanFiles` is the canonical richer form would help future readers.

cmd/iceberg/clean_orphan_files.go
zeroshade: nit: this size lookup assumes `DeletedFiles` is a subset of `OrphanFiles` (a path absent from the map silently yields `SizeBytes: 0`). True today since both derive from the same scan, but it's an undocumented invariant — filtering `OrphanFiles` by the deleted set would remove the implicit dependency.

zeroshade (approved):
LGTM — per-file orphan sizes are surfaced in both text and JSON output with adequate test coverage. Two nits inline.

PR #1413, opened 2026-07-05
fix(partitions): reject negative PartitionSpec IDs

fallintoplace:
## Summary

- validate `WithSpecID` inputs and reject negative partition spec IDs;
- return `ErrInvalidPartitionSpec` with the invalid value;
- add regression coverage for the negative boundary.

## Testing

- `go test . -run TestPartitionSpecRejectsNegativeSpecID -count=1`

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -210,6 +210,9 @@ func NewPartitionSpecOpts(opts ...PartitionOption) (PartitionSpec, error) {
 
 func WithSpecID(id int) PartitionOption {
 	return func(p *PartitionSpec) error {
+		if id < 0 {
+			return fmt.Errorf("spec id must be non-negative: %d", id)
+		}
 		p.id = id
 
 		return nil
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -176,6 +176,13 @@ func TestPartitionSpecRejectsInvalidBucketTransform(t *testing.T) {
 	require.ErrorContains(t, err, "numBuckets > 0")
 }
 
+func TestPartitionSpecRejectsNegativeSpecID(t *testing.T) {
+	_, err := iceberg.NewPartitionSpecOpts(iceberg.WithSpecID(-1))
+
+	require.ErrorIs(t, err, iceberg.ErrInvalidPartitionSpec)
+	require.ErrorContains(t, err, "spec id must be non-negative: -1")
+}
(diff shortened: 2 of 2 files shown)


partitions.go
```diff
@@ -210,6 +210,9 @@ func NewPartitionSpecOpts(opts ...PartitionOption) (PartitionSpec, error) {
 
 func WithSpecID(id int) PartitionOption {
 	return func(p *PartitionSpec) error {
+		if id < 0 {
+			return fmt.Errorf("%w: spec id must be non-negative: %d", ErrInvalidPartitionSpec, id)
```
tanmayrauth: `return fmt.Errorf("%w: spec id must be non-negative: %d", ErrInvalidPartitionSpec, id)` — this double-wraps the sentinel. NewPartitionSpecOpts already wraps every option error with `%w: ErrInvalidPartitionSpec` (line 203), so the message comes out as "invalid partition spec: invalid partition spec: spec id must be non-negative: -1". The sibling options (AddPartitionFieldByName, addSpecFieldInternal, etc.) all return bare errors for exactly this reason.

tanmayrauth (commented):
One thing on the error wrapping, please check it inline.

partitions.go
zeroshade: `NewPartitionSpecOpts` already wraps option errors with `fmt.Errorf("%w: %w", ErrInvalidPartitionSpec, err)`, so wrapping the sentinel here too double-wraps it — the message becomes `invalid partition spec: invalid partition spec: spec id must be non-negative: -1`. Sibling options return a bare error and let the outer wrapper attach the sentinel; suggest `fmt.Errorf("spec id must be non-negative: %d", id)` here.

zeroshade (changes requested):
One small error-wrapping fix and this is good to go.

zeroshade (approved):
LGTM. WithSpecID now returns an unwrapped error, so NewPartitionSpecOpts wraps ErrInvalidPartitionSpec exactly once (no double-wrap), and the table test asserts ErrorIs plus the clean message. Rejecting negative spec IDs at construction is spec-aligned and matches Java/pyiceberg, which likewise prevent negatives at write time rather than on metadata read. CI green. This supersedes my earlier change request.

PR #1414, opened 2026-07-05
fix(table): guard malformed parquet UUID stats

fallintoplace:
## What changed
- Guard UUID parquet stats aggregation with a panic-safe wrapper in DataFileStatsFromMeta.
- If UUID bounds parsing panics (e.g. malformed fixed bytes), invalidate stats for that column so read planning still succeeds.
- Add regression coverage for malformed UUID min/max bytes and verify column bounds are omitted.

## Validation
- go test ./table/internal -run TestDataFileStatsFromMetaWithMalformedUUIDStats -count=1
- go test ./table/internal

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -505,7 +505,19 @@ type wrappedUUIDStats struct {
 	*metadata.FixedLenByteArrayStatistics
 }
 
+func (w *wrappedUUIDStats) HasMinMax() bool {
+	const uuidByteWidth = 16
+
+	return w.FixedLenByteArrayStatistics.HasMinMax() &&
+		len(w.FixedLenByteArrayStatistics.Min()) == uuidByteWidth &&
+		len(w.FixedLenByteArrayStatistics.Max()) == uuidByteWidth
+}
+
 func (w *wrappedUUIDStats) Min() uuid.UUID {
+	if !w.HasMinMax() {
+		return uuid.Nil
+	}
+
 	uid, err := uuid.FromBytes(w.FixedLenByteArrayStatistics.Min())
 	if err != nil {
 		panic(err)
@@ -515,6 +527,10 @@ func (w *wrappedUUIDStats) Min() uuid.UUID {
 }
 
 func (w *wrappedUUIDStats) Max() uuid.UUID {
+	if !w.HasMinMax() {
+		return uuid.Nil
+	}
+
 	uid, err := uuid.FromBytes(w.FixedLenByteArrayStatistics.Max())
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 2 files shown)


table/internal/parquet_files.go
```diff
@@ -730,6 +736,18 @@ func (p parquetFormat) DataFileStatsFromMeta(meta Metadata, statsCols map[int]St
 	}
 }
 
+func safeUpdateAgg(agg StatsAgg, stats interface{ HasMinMax() bool }) (err error) {
```
tanmayrauth: `safeUpdateAgg` recovers from *any* panic during aggregation, but the actual bug is narrower: `wrappedUUIDStats.Min()`/`Max()` (lines 508/517) panic when `uuid.FromBytes` gets malformed non-16-byte stats. A blanket recover here means a genuine bug in any other aggregator (decimal, string, FLBA) also gets silently swallowed as "invalidate this column" instead of surfacing.
  
  Would you consider fixing it at the source instead — have the UUID wrapper (and by symmetry the FLBA/decimal-FLBA wrappers, which have the same fixed-length assumption) detect wrong-length bytes and signal "no valid bounds" so the column gets invalidated at the same spot as the `!set` / `stats == nil` cases (lines 653/660)? That keeps invalidation scoped to malformed bounds and preserves panics as real-bug signals everywhere else.

partitions.go
```diff
@@ -210,6 +210,9 @@ func NewPartitionSpecOpts(opts ...PartitionOption) (PartitionSpec, error) {
 
 func WithSpecID(id int) PartitionOption {
 	return func(p *PartitionSpec) error {
+		if id < 0 {
```
tanmayrauth: This `id < 0` check (the "Validate partition spec IDs in WithSpecID" commit) looks unrelated to the UUID stats fix — can it go in its own PR? Bundling it makes this one harder to review and revert.
  
  Also, just to confirm: this only fires on the programmatic `NewPartitionSpecOpts`/`BindToSchema` path — `UnmarshalJSON` at partitions.go:436 sets `ps.id` directly and bypasses it, so existing metadata still reads fine. Is there any internal caller that passes a transient negative id before assigning the real one?

tanmayrauth (commented):
The UUID panic guard works and the regression test is solid, but I'd like to sort out a couple of things before approving — the recover() is broader than the actual bug, and there's an unrelated spec-id change bundled in. Details  inline.

table/internal/parquet_files.go
```diff
@@ -730,6 +736,18 @@ func (p parquetFormat) DataFileStatsFromMeta(meta Metadata, statsCols map[int]St
...
+	defer func() {
+		if r := recover(); r != nil {
```
zeroshade: This blanket `recover()` swallows panics from any aggregator (nil deref, type assertion, decimal/FLBA, etc.), turning real bugs into silent column-stat invalidation. Prefer fixing the root cause: have the UUID (and decimal/FLBA) stat wrappers' `Min()`/`Max()` check byte-length and return a no-bounds result rather than panicking.

partitions.go
zeroshade: This negative-spec-id guard is unrelated to the UUID-stats fix and is identical to the change in #1413 (reject negative PartitionSpec IDs). Please drop it from this PR so the two don't conflict — it can land via #1413.

zeroshade (changes requested):
Two items from the earlier review still stand (inline).

zeroshade (approved):
Re-reviewed after the latest commit — both points resolved: the blanket `recover()`/`safeUpdateAgg` is gone, and the UUID/FLBA/decimal stat wrappers now validate byte-length in `HasMinMax()` and return a no-bounds result (invalidating the column and continuing) instead of panicking — exactly the root-cause fix. The unrelated `partitions.go` change is also dropped (now just `parquet_files.go` + test). LGTM — thanks!

PR #1415, opened 2026-07-05
fix(schema): prevent AsStruct aliasing

fallintoplace:
## What changed
- Make `Schema.AsStruct()` return a deep-cloned `StructType` field list instead of exposing internal schema state.
- Add deep-copy helpers for nested struct/list/map types to prevent mutable aliasing through returned nested types.
- Add regression tests covering top-level and nested field aliasing.

This ensures callers cannot mutate schema internals through `Schema.AsStruct()`.

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -257,7 +257,13 @@ func (s *Schema) Type() string { return "struct" }
 
 // AsStruct returns a Struct with the same fields as the schema which can
 // then be used as a Type.
-func (s *Schema) AsStruct() StructType    { return StructType{FieldList: s.fields} }
+func (s *Schema) AsStruct() StructType { return StructType{FieldList: cloneFields(s.fields)} }
+
+// asStructRef returns a read-only view over the schema field slice for
+// internal callers that only traverse fields and must avoid clone-on-read
+// overhead on hot paths.
+func (s *Schema) asStructRef() StructType { return StructType{FieldList: s.fields} }
+
 func (s *Schema) NumFields() int          { return len(s.fields) }
 func (s *Schema) Field(i int) NestedField { return s.fields[i] }
 func (s *Schema) Fields() []NestedField   { return slices.Clone(s.fields) }
@@ -308,6 +314,46 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 		Fields []NestedField `json:"fields"`
 		*Alias
 	}{Type: "struct", Fields: s.fields, Alias: &aliasCopy})
+}
+
+func cloneFields(fields []NestedField) []NestedField {
+	if len(fields) == 0 {
+		return nil
+	}
+
+	cloned := make([]NestedField, len(fields))
+	for i, field := range fields {
+		cloned[i] = field
+		cloned[i].Type = cloneType(field.Type)
+	}
+
+	return cloned
+}
+
+func cloneType(t Type) Type {
+	if t == nil {
+		return nil
+	}
+
+	switch typed := t.(type) {
+	case *StructType:
+		return &StructType{FieldList: cloneFields(typed.FieldList)}
+	case *ListType:
+		cloned := *typed
+		cloned.Element = cloneType(typed.Element)
+
+		return &cloned
+	case *MapType:
+		cloned := *typed
+		cloned.KeyType = cloneType(typed.KeyType)
+		cloned.ValueType = cloneType(typed.ValueType)
+
+		return &cloned
+	default:
+		// Leaf and parameterized non-container types are immutable value objects,
+		// so sharing them between clones is safe.
+		return t
+	}
 }
 
 // FindColumnName returns the name of the column identified by the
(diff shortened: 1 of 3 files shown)


schema.go
```diff
@@ -257,7 +257,7 @@ func (s *Schema) Type() string { return "struct" }
 
 // AsStruct returns a Struct with the same fields as the schema which can
 // then be used as a Type.
-func (s *Schema) AsStruct() StructType    { return StructType{FieldList: s.fields} }
+func (s *Schema) AsStruct() StructType    { return StructType{FieldList: cloneFields(s.fields)} }
```
tanmayrauth: `func (s *Schema) AsStruct() StructType { return StructType{FieldList: cloneFields(s.fields)} }`
  
  This method isn't only a public convenience — it's called on internal hot paths that never mutate the result: the schema visitors (Visit at schema.go:599, PreOrderVisit at schema.go:770, visitStructWithPartner at schema.go:1485) and the metrics evaluator constructors (table/evaluators.go:713, :728, :1247). Those evaluators are built per-manifest/per-file during scan planning (table/scanner.go, arrow_scanner.go:955), and the visitors run on every projection/prune/promote pass. All of them just read FieldList.
  
  With this change every one of them now recursively clones the whole schema tree on each call, which is a meaningful regression for wide/deeply-nested schemas during scans. Could we keep an internal non-cloning path (callers use s.fields directly, or a private asStructRef()) and only clone in the exported AsStruct()? That preserves the public-safety guarantee without taxing the read-only internal callers.

schema.go
```diff
@@ -310,6 +310,44 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
...
+	cloned := make([]NestedField, len(fields))
+	for i, field := range fields {
+		cloned[i] = field
+		cloned[i].Type = cloneType(field.Type)
+	}
+
+	return cloned
+}
+
+func cloneType(t Type) Type {
```
tanmayrauth, line 333: `func cloneType(t Type) Type {`
  
  The default branch returns t unchanged, so leaf/parameterized types (Decimal, Fixed, Geometry, Geography, Variant, ...) stay aliased — only the container types are deep-copied. I think that's intentional since the package treats those as immutable, but a one-line comment saying so would keep the partial clone from reading as a missed case.

schema_test.go
```diff
@@ -219,6 +219,39 @@ func TestNestedFieldToString(t *testing.T) {
...
+}
+
+func TestSchemaAsStructClonesNestedTypes(t *testing.T) {
+	schema := iceberg.NewSchema(0,
+		iceberg.NestedField{
+			ID:   1,
+			Name: "person",
+			Type: &iceberg.StructType{FieldList: []iceberg.NestedField{
+				{ID: 2, Name: "name", Type: iceberg.PrimitiveTypes.String},
+			}},
```
tanmayrauth, line 240: Nice coverage on the struct case. Since cloneType also special-cases *ListType and *MapType (schema.go:337, :342), could you add a nested list and map aliasing test too? Otherwise a future tweak to those branches could regress without a failing test.

tanmayrauth (commented):
Fix is correct and the tests nail the aliasing. My one real concern is that AsStruct is on the scan/visitor hot paths and none of those callers mutate — so this makes them all deep-clone for nothing. Details inline.

schema.go
```diff
@@ -257,7 +257,7 @@ func (s *Schema) Type() string { return "struct" }
 
 // AsStruct returns a Struct with the same fields as the schema which can
 // then be used as a Type.
-func (s *Schema) AsStruct() StructType    { return StructType{FieldList: s.fields} }
+func (s *Schema) AsStruct() StructType    { return StructType{FieldList: cloneFields(s.fields)} }
 func (s *Schema) NumFields() int          { return len(s.fields) }
```
zeroshade, line 267: `AsStruct()` is on the read-only scan critical path (called by schema visitors and evaluator constructors that never mutate the result). Deep-cloning the full field tree on every call is a real regression for wide/nested schemas. Suggest a shared/private `asStructRef()` for internal read-only callers, keeping the deep clone only in the exported `AsStruct()`.

schema.go
```diff
@@ -310,6 +310,44 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
...
+		cloned.Element = cloneType(typed.Element)
+
+		return &cloned
+	case *MapType:
+		cloned := *typed
+		cloned.KeyType = cloneType(typed.KeyType)
+		cloned.ValueType = cloneType(typed.ValueType)
+
+		return &cloned
+	default:
```
zeroshade, line 352: The `default` branch returns `t` unchanged (primitive/Decimal/Fixed/Geometry/Geography/Variant) — likely intentional since these are immutable values, but a one-line comment would keep it from reading as a missed case.

zeroshade (changes requested):
The aliasing fix is right in principle, but the current placement is a scan-path performance regression (inline). Also missing `*ListType`/`*MapType` aliasing tests — the added tests only cover the top-level field list and nested `*StructType`.

zeroshade (approved):
Re-reviewed after the latest commit — the hot-path concern is fully resolved:
- `AsStruct()` still deep-clones for the public/safe path, but the internal read-only visitors now use the new non-cloning `asStructRef()`, and the evaluator constructors no longer call `AsStruct()` at all — so scan planning no longer pays for a full field-tree clone.
- The `cloneType` default branch now has an explanatory comment.
- Dedicated `*ListType`/`*MapType` aliasing tests were added.

LGTM — thanks for the quick turnaround!

PR #1416, opened 2026-07-05
fix(table): make transaction apply atomic

fallintoplace:
## Summary
- Make Transaction.apply stage metadata updates on a cloned metadata builder before mutating transaction state
- Avoid partially applying updates when one update fails (no-op on error)
- Keep requirements list atomic together with metadata mutations
- Ensure metadata file source selection prefers table location, with fallback to last logged metadata entry

## Testing
- go test ./table -run 'TestTransactionApplyKeepsMetadataUnchangedOnUpdateFailure|TestTransactionApply' -count=1


table/metadata.go
```diff
@@ -283,6 +283,59 @@ func MetadataBuilderFromBase(metadata Metadata, currentFileLocation string) (*Me
...
+		formatVersion:        b.formatVersion,
+		uuid:                 b.uuid,
+		loc:                  b.loc,
+		lastUpdatedMS:        b.lastUpdatedMS,
+		lastColumnId:         b.lastColumnId,
+		schemaList:           slices.Clone(b.schemaList),
+		currentSchemaID:      b.currentSchemaID,
+		specs:                slices.Clone(b.specs),
+		defaultSpecID:        b.defaultSpecID,
+		lastPartitionID:      b.lastPartitionID,
```
tanmayrauth: nit: this shallow-copies the pointer while every sibling *int/*int64 (currentSnapshotID, lastSequenceNumber, nextRowID, lastAdded*) goes through cloneMetadataBuilderInt. It's safe today only because addPartitionSpec reassigns   `b.lastPartitionID = &x` rather than writing `*b.lastPartitionID`, so a failed apply can't leak back into t.meta — but it's inconsistent and turns into a real atomicity hole if anything ever mutates the pointee in place. Suggest  `cloneMetadataBuilderInt(b.lastPartitionID)` for consistency.

table/metadata.go
```diff
@@ -283,6 +283,59 @@ func MetadataBuilderFromBase(metadata Metadata, currentFileLocation string) (*Me
...
+
+// clone returns a metadata-builder copy suitable for staging updates.
+// The copy shares only immutable data with the original builder and keeps
+// value types separated so partial update attempts remain atomic.
+func (b *MetadataBuilder) clone() *MetadataBuilder {
+	if b == nil {
+		return nil
+	}
+
+	cloned := &MetadataBuilder{
```
tanmayrauth, line 304: Suggestion: this manually lists all 29 fields, so a newly-added scalar field would silently clone as its zero value with no compiler warning — an easy future bug. Consider `cloned := *b` to auto-carry every value/scalar field, then override just the reference types (slices/maps/pointers) below. That way the only thing to maintain is the deep-copy list, not the full field set.

tanmayrauth (commented):
Thanks, Core fix looks right — staging on a clone and swapping only on full success makes apply a clean no-op on failure.

table/metadata.go
zeroshade: nit: `lastPartitionID` is shallow-copied here while the other `*int`/`*int64` fields go through `cloneMetadataBuilderInt`. Safe today (mutations reassign the pointer rather than write through it), but inconsistent and a latent atomicity hole if that ever changes. The full manual field enumeration is also a bit fragile — `cloned := *b` then overriding only the reference types would be more robust.

zeroshade (approved):
LGTM — staging all updates on a cloned builder before committing gives real atomicity, and the test confirms metadata/reqs are untouched after a mid-batch failure. One consistency nit inline.

laskoviymishka (approved):
LGTM!

PR #1417, opened 2026-07-05
fix(table): validate write defaults under old schema versions

fallintoplace:
## Summary

`checkSchemaCompatibility` now treats `WriteDefault` like `InitialDefault` for non-geometry/geography fields when creating or evolving v1/v2 schemas.

- v1/v2 reject non-null `InitialDefault` and `WriteDefault`, with separate diagnostics.
- geometry/geography defaults remain required to be null.
- compatibility errors print default values, not pointer addresses.
- tests cover direct compatibility checks and `UpdateColumn(... WriteDefault ...)` end to end.

## Compatibility scope

This check runs through `MetadataBuilder.AddSchema`, which is used by new schema construction and schema evolution. `ParseMetadataBytes` unmarshals existing metadata directly and does not call `checkSchemaCompatibility`, so this change does not reject externally-produced v1/v2 metadata during read. The stricter `WriteDefault` validation is intentional for schemas constructed or evolved by this package.

## Testing

- `go test ./table -run 'TestNonSpecialDefaultsRequireV3|TestGeometryGeographyNullOnlyDefaults' -count=1`
- `go test . -count=1`
- `go test ./table -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=10m . ./table/...`

### DIFF
--- a/table/metadata_builder_internal_test.go
+++ b/table/metadata_builder_internal_test.go
@@ -2259,6 +2259,7 @@ func TestNonSpecialDefaultsRequireV3(t *testing.T) {
 			require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
 			require.ErrorContains(t, err, "invalid initial default")
 			require.ErrorContains(t, err, "non-null default")
+			require.ErrorContains(t, err, "non-null default (7)")
 			require.ErrorContains(t, err, "is not supported until v3")
 		})
 
@@ -2279,6 +2280,7 @@ func TestNonSpecialDefaultsRequireV3(t *testing.T) {
 			require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
 			require.ErrorContains(t, err, "invalid write default")
 			require.ErrorContains(t, err, "non-null default")
+			require.ErrorContains(t, err, "non-null default (7)")
 			require.ErrorContains(t, err, "is not supported until v3")
 		})
 
--- a/table/metadata_builder_internal_test.go
+++ b/table/metadata_builder_internal_test.go
@@ -2232,6 +2232,81 @@ func TestGeometryGeographyNullOnlyDefaults(t *testing.T) {
 	}
 }
 
+func TestNonSpecialDefaultsRequireV3(t *testing.T) {
+	testTypes := []struct {
+		name string
+		typ  iceberg.Type
+	}{
+		{"int", iceberg.Int32Type{}},
+		{"string", iceberg.StringType{}},
+	}
+
+	for _, tt := range testTypes {
+		t.Run(tt.name+" v2 with initial default", func(t *testing.T) {
+			defaultValue := 7
+			sc := iceberg.NewSchema(0,
+				iceberg.NestedField{
+					Type:           tt.typ,
+					ID:             1,
+					Name:           "col",
+					Required:       false,
+					InitialDefault: &defaultValue,
+				},
+			)
+
+			err := checkSchemaCompatibility(sc, 2)
+			require.Error(t, err)
+			require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
+			require.ErrorContains(t, err, "invalid initial default")
+			require.ErrorContains(t, err, "non-null default")
+			require.ErrorContains(t, err, "is not supported until v3")
+		})
+
+		t.Run(tt.name+" v2 with write default", func(t *testing.T) {
+			defaultValue := 7
+			sc := iceberg.NewSchema(0,
+				iceberg.NestedField{
+					Type:         tt.typ,
+					ID:           1,
+					Name:         "col",
+					Required:     false,
+					WriteDefault: &defaultValue,
+				},
+			)
+
+			err := checkSchemaCompatibility(sc, 2)
+			require.Error(t, err)
+			require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
+			require.ErrorContains(t, err, "invalid write default")
+			require.ErrorContains(t, err, "non-null default")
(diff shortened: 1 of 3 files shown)


table/metadata_schema_compatibility.go
```diff
@@ -47,7 +47,12 @@ func (e ErrIncompatibleSchema) Error() string {
 					fmt.Fprintf(&problems, "\n- invalid write default for %s: %s columns must default to null", f.ColName, f.Field.Type)
 				}
 			} else {
-				fmt.Fprintf(&problems, "\n- invalid initial default for %s: non-null default (%v) is not supported until v%d", f.ColName, f.Field.InitialDefault, f.InvalidDefault.MinFormatVersion)
+				if f.Field.InitialDefault != nil {
+					fmt.Fprintf(&problems, "\n- invalid initial default for %s: non-null default (%v) is not supported until v%d", f.ColName, f.Field.InitialDefault, f.InvalidDefault.MinFormatVersion)
+				}
+				if f.Field.WriteDefault != nil {
```
zeroshade, line 54: nit: now that the `WriteDefault: field.InitialDefault` initializer is removed, the `InvalidDefault.WriteDefault` struct field is dead — it's never set, and the error text reads `f.Field.WriteDefault` directly. Worth removing that field. Also consider adding a test for the `InitialDefault`-only pre-v3 case.

zeroshade (approved):
LGTM — the compatibility check now fires on a non-null `WriteDefault` (not just `InitialDefault`) for pre-v3 schemas, with separate error lines for each. One dead-code nit inline.

table/metadata_schema_compatibility.go
```diff
@@ -47,7 +47,12 @@ func (e ErrIncompatibleSchema) Error() string {
...
+					fmt.Fprintf(&problems, "\n- invalid write default for %s: non-null default (%v) is not supported until v%d", f.ColName, f.Field.WriteDefault, f.InvalidDefault.MinFormatVersion)
```
laskoviymishka: `InitialDefault`/`WriteDefault` are `any`, and every construction site here (including these new tests and `TestGeometryGeographyNullOnlyDefaults`) sets them to `&defaultValue`. `%v` on a pointer-to-scalar prints the address, so this renders `non-null default (0xc0000140a0)` rather than `(7)` — which defeats the point of the split, and the tests don't catch it since they only `ErrorContains` the fixed substrings.

I'd dereference before formatting (a small helper or `reflect.Indirect`), and while we're here add a `require.ErrorContains(t, err, "(7)")` to one subtest so the value is actually pinned. wdyt?

table/metadata_schema_compatibility.go
```diff
@@ -133,11 +137,11 @@ func checkSchemaCompatibility(sc *iceberg.Schema, formatVersion int) error {
 				})
 			}
 		default:
-			if field.InitialDefault != nil && formatVersion < defaultValuesMinFormatVersion {
+			if (field.InitialDefault != nil || field.WriteDefault != nil) && formatVersion < defaultValuesMinFormatVersion {
```
laskoviymishka, line 160: One thing I want to make sure of: Java's `Schema.checkCompatibility` only gates on `initialDefault()`, not `writeDefault()` — write-default is described as forward-compatible (old writers just fail on the missing field). So this now rejects a v2 schema that Java/PyIceberg would accept (initial-default null, write-default set).

If `checkSchemaCompatibility` only runs on iceberg-go-initiated schema construction/evolution, this is fine and arguably more correct. But if it's also on the read path via `MetadataBuilder.AddSchema`, we could fail to load externally-produced v2 metadata that carries a write-default. Do you know which path this fires on? If it's read-side, I'd want to loosen this to match Java; if it's construction-only, worth a one-line comment noting the intentional divergence (the `// Java: Schema::checkCompatibility` note above becomes stale otherwise).

table/metadata_builder_internal_test.go
```diff
@@ -2232,6 +2232,81 @@ func TestGeometryGeographyNullOnlyDefaults(t *testing.T) {
 	}
 }
 
+func TestNonSpecialDefaultsRequireV3(t *testing.T) {
```
laskoviymishka, line 2235: Coverage here is at the `checkSchemaCompatibility` layer, but the bug this fixes is reached through `UpdateColumn(path, ColumnUpdate{WriteDefault: ...}).Commit()` on a v2 table. Would be worth one test at that layer asserting the error surfaces end-to-end through `AddSchema` — that's the path a real caller hits, and there's currently no `WriteDefault` case in `TestUpdateColumn` at all. Non-blocking, but it'd lock the actual regression.

laskoviymishka (changes requested):
Nice, focused fix — the write-default now gets validated symmetrically with the initial-default for pre-v3 schemas, and dropping the mislabeled `WriteDefault any` field (it was holding `InitialDefault`) is a good cleanup.

Almost there. One thing before merge: the two `non-null default (%v)` lines format the `any` values directly, and those hold a pointer in every test here (`&defaultValue`). `%v` on a `*int` prints the address, so the diagnostic renders `non-null default (0xc0000140a0)` instead of `(7)` for exactly the inputs the new tests build — and the tests pass anyway because they only assert on the fixed substrings. I'd dereference before formatting and pin the value in one subtest assertion.

One question too: gating write-default on pre-v3 is stricter than Java's `checkCompatibility`, which only looks at initial-default. If this runs on the read path it could reject externally-produced v2 metadata carrying a write-default — flagged inline, would like to know which path it fires on.

Once the formatting is sorted and we've settled the read-path question, happy to take another pass.

laskoviymishka (approved):
🚢

Issue #1360, opened 2026-07-03
SQL UDF (function) catalog support: UDF metadata model + REST read-side client


rraulinio:
# Add SQL UDF metadata model and REST read support

Spec: https://iceberg.apache.org/udf-spec/

## Background

Apache Iceberg 1.11.0 introduced a **SQL UDF specification**.

In plain terms, a catalog UDF is a named, reusable SQL routine stored in the catalog. A scalar UDF is a reusable expression, and a UDTF is closer to a parameterized view that returns rows. For example, a catalog can define `add_one(x int)` once, and engines such as Spark or Trino can discover it from the catalog and use the SQL representation written for their dialect.

The metadata design deliberately mirrors Iceberg tables and views: each function is represented by one self-contained, immutable JSON metadata file. The catalog maps the function name to the current metadata file location and updates that mapping with an atomic swap. All overloads of a name live in a single metadata file, each overload has its own version history, and a `definition-log` records the selected definition-version mappings over time so the function can be rolled back without external state.

## What this unlocks

- Go clients can discover and read catalog-managed UDFs from REST catalogs that serve them.
- Go-based catalog servers and tools get a shared, spec-validated model for UDF metadata instead of each reinventing JSON handling, similar to the role the existing `view` package plays for views.

## Proposed decomposition - two PRs

The split keeps review focused: **PR 1 is reviewable purely against the UDF format spec** with no HTTP involved, and **PR 2 is reviewable purely against the REST OpenAPI**. PR 2 depends on PR 1.

- [x] **PR 1 - `udf` metadata model package.** Add a new package named `udf`, matching Java's `org.apache.iceberg.udf` package and the spec title. The package would mirror the structure of the existing `view` package: metadata model (`function-uuid`, `format-version`, `definitions`, `definition-log`, `location`, `properties`, `secure`, `doc`), definitions (`parameters`, `return-type`, `function-type` `udf`/`udtf`, `versions`, `current-version-id`, optional `return-nullable`, optional `doc`, and optional `specific-name`), definition versions (`representations`, `deterministic`, `on-null-input`, `timestamp-ms`), SQL representations (one per dialect per version), an unknown representation fallback for forward-compatible round-tripping, parsing/serialization, validation of the spec invariants, and a metadata builder for writers.

[...]

PR #1418, opened 2026-07-06
Add SQL UDF metadata model package

rraulinio:
## Description

Addressing part 1 from https://github.com/apache/iceberg-go/issues/1360.

This adds a new `udf` package for Apache Iceberg SQL UDF metadata. The package models the UDF metadata JSON format from the Iceberg UDF spec: function UUID, format version, definitions, definition versions, `definition-log`, properties, secure flag, documentation, scalar UDFs and table UDFs. It includes parsing, JSON serialization, validation, semantic equality, cloning and a metadata builder for writers.

Definitions are keyed by the canonical `definition-id` derived from parameter types, including nested list/map/struct signatures such as `int,list<int>,struct<id:int,name:string>`. UDF types are represented separately from schema types because UDF nested type JSON does not require field IDs, while iceberg-go schema parsing expects them. The parser accepts primitive type strings and the spec's nested type JSON objects, ignores extra nested-type fields as required by the spec and rejects nested canonical strings as metadata type encodings.

The implementation validates the important spec invariants: required metadata fields, supported format version, one definition per signature, unique optional `specific-name`, valid function type, UDTF return type as a struct, non-empty definition versions, current version references, required `definition-log`, required representation fields, one SQL representation per dialect, supported `on-null-input` values and invalid nil/null or negative ID/timestamp cases. Unknown representation types are preserved for forward-compatible round-tripping when read from metadata, while writer-created unknown representations without raw JSON are rejected.

This PR is intentionally limited to the metadata model. REST read-side support remains the next step from #1360.

## Tests

- `go test ./udf`
- `go test -race ./udf`
- `go test ./...`
- `go vet ./udf`

### DIFF
--- a/udf/metadata.go
+++ b/udf/metadata.go
@@ -520,8 +520,15 @@ func (d *Definition) validate() error {
 	switch d.FunctionType {
 	case FunctionTypeUDF:
 	case FunctionTypeUDTF:
-		if _, ok := d.ReturnType.(StructType); !ok {
+		st, ok := d.ReturnType.(StructType)
+		if !ok {
 			return fmt.Errorf("%w: definition %q is a udtf but its return-type is not a struct",
+				ErrInvalidUDFMetadata, d.DefinitionID)
+		}
+		// The struct describes the output schema; a table without columns
+		// is meaningless.
+		if len(st.Fields) == 0 {
+			return fmt.Errorf("%w: definition %q is a udtf but its return-type struct has no fields",
 				ErrInvalidUDFMetadata, d.DefinitionID)
 		}
 	case "":
@@ -743,6 +750,9 @@ func (m *metadata) DefinitionByID(definitionID string) (*Definition, bool) {
 
 func (m *metadata) Equals(other Metadata) bool {
 	if other == nil {
+		return false
+	}
+	if o, ok := other.(*metadata); ok && o == nil {
 		return false
 	}
 
--- a/udf/metadata.go
+++ b/udf/metadata.go
@@ -1,0 +1,887 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package udf provides the metadata model for Iceberg SQL UDFs as
+// specified by the Iceberg UDF spec.
+// https://iceberg.apache.org/udf-spec/
+package udf
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"maps"
+	"slices"
+	"strings"
+	"sync"
+
+	"github.com/apache/iceberg-go"
+
+	"github.com/google/uuid"
+)
+
+var (
+	ErrInvalidUDFMetadata              = errors.New("invalid UDF metadata")
+	ErrInvalidUDFMetadataFormatVersion = errors.New("invalid or missing format-version in UDF metadata")
+)
+
+const (
+	SupportedUDFFormatVersion = 1
+
+	initialVersionID = 1
+)
+
+// FunctionType indicates whether a definition is a scalar function (udf)
+// or a table function (udtf).
+type FunctionType string
+
+const (
+	FunctionTypeUDF  FunctionType = "udf"
+	FunctionTypeUDTF FunctionType = "udtf"
+)
+
+// OnNullInput defines how a UDF behaves when any input parameter is NULL.
+type OnNullInput string
+
+const (
+	// OnNullInputCall means the function may handle NULLs internally, so
+	// engines must execute it even if some inputs are NULL. This is the
+	// default behavior when on-null-input is absent.
+	OnNullInputCall OnNullInput = "call"
+	// OnNullInputReturnNull means the function always returns NULL if any
+	// input argument is NULL, allowing engines to skip evaluation.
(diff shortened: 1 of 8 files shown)


rraulinio:
cc @laskoviymishka @tanmayrauth

udf/types.go
```diff
@@ -0,0 +1,360 @@
...
+
+// PrimitiveTypeOf validates s against the Iceberg type system and the UDF
+// spec's rules for type strings (no spaces or quote characters) and returns
+// the corresponding PrimitiveType.
+func PrimitiveTypeOf(s string) (PrimitiveType, error) {
+	if s == "" {
+		return PrimitiveType{}, fmt.Errorf("%w: type string must not be empty", ErrInvalidUDFType)
+	}
+
+	if strings.ContainsFunc(s, unicode.IsSpace) || strings.ContainsAny(s, `"'`) {
```
tanmayrauth, line 67: This rejects any type string with a space, but the canonical Iceberg string for decimal has one — `decimal(P, S)`. That's what the table spec, Java's DecimalType.toString(), PyIceberg, and iceberg-go's own DecimalType.String()/parser all emit and normalize to. Two consequences I confirmed locally:
  
  1. Reading UDF metadata whose param/return type is `decimal(9, 2)` (i.e. anything a conformant engine writes) fails outright — PrimitiveTypeOf rejects the spaced form, and it propagates to nested types like `list<decimal(9, 2)>` too.
  2. Because the definition-id is derived from the verbatim stored string, an iceberg-go author writing `decimal(9,2)` produces a different definition-id than Java/Py (`decimal(9, 2)`) for the same overload, so they can't be correlated.
  
  Since PrimitiveTypeOf already validates by round-tripping through iceberg's NestedField parser, could it derive the stored canonical string from the *parsed* type (one normalization) instead of storing the raw input and rejecting whitespace? That way `decimal(9, 2)` and `decimal(9,2)` collapse to one canonical form and reads succeed. Worth pinning down what the spec declares the canonical definition-id form to be for parameterized types like decimal — but hard-rejecting the spaced form on the read path breaks compat either way.

udf/types.go
```diff
@@ -0,0 +1,360 @@
...
+		return nil
+	case ListType:
+		return validateType(v.Element)
+	case MapType:
+		if err := validateType(v.Key); err != nil {
+			return err
+		}
+
+		return validateType(v.Value)
+	case StructType:
```
tanmayrauth, line 239: Struct field names are checked for non-emptiness but not uniqueness — `struct<a:int,a:string>` validates and round-trips here. Field names should be unique among siblings; worth rejecting duplicates, mirroring the duplicate-parameter-name check you already have at the definition level.

udf/metadata_builder.go
```diff
@@ -0,0 +1,561 @@
...
+
+	uuid_ := b.uuid
+	if uuid_ == uuid.Nil {
+		uuid_ = uuid.New()
+	}
+
+	definitionLog := cloneDefinitionLog(b.definitionLog)
+	if b.selectionChanged {
+		timestampMS := b.logTimestampMS
+		if timestampMS == 0 {
```
tanmayrauth: `Build` treats a log timestamp of 0 as "unset" and substitutes time.Now(), so a writer can't set a deterministic epoch-0 log timestamp. DefinitionVersion.TimestampMS uses -1 as its sentinel and accepts 0 — could the log timestamp use the same -1 convention so the two agree? Minor.

udf/metadata.go
```diff
@@ -0,0 +1,887 @@
...
+func (m *metadata) Doc() string                         { return m.DocValue }
+
+func (m *metadata) DefinitionByID(definitionID string) (*Definition, bool) {
+	def, ok := m.lazyDefinitionsByID()[definitionID]
+
+	return def, ok
+}
+
+func (m *metadata) Equals(other Metadata) bool {
+	if other == nil {
```
tanmayrauth, line 752: This guards an untyped-nil interface, but a typed-nil Metadata (e.g. `(*metadata)(nil)`) slips through and panics at `other.FunctionUUID()` when it dereferences `*m.UUID`. Not reachable through the public API today, but given you're already nil-checking here it may be worth hardening.

udf/metadata.go
```diff
@@ -0,0 +1,887 @@
...
+	if err := validateType(d.ReturnType); err != nil {
+		return fmt.Errorf("%w: definition %q return-type: %w", ErrInvalidUDFMetadata, d.DefinitionID, err)
+	}
+
+	switch d.FunctionType {
+	case FunctionTypeUDF:
+	case FunctionTypeUDTF:
+		if _, ok := d.ReturnType.(StructType); !ok {
+			return fmt.Errorf("%w: definition %q is a udtf but its return-type is not a struct",
+				ErrInvalidUDFMetadata, d.DefinitionID)
```
tanmayrauth, line 526: validateType accepts an empty struct, so a udtf can declare a zero-column output struct and still validate. Worth requiring a non-empty struct here to match the output-schema intent?

tanmayrauth (commented):
Really clean implementation and exhaustive tests. One thing I think blocks though: decimal (and any spaced) type strings don't parse, which breaks cross-engine reads and makes definition-ids disagree with Java/PyIceberg. Plus a struct validation gap and a few nit, all inline.

udf/types.go
rraulinio: 👍

udf/metadata_builder.go
rraulinio: 👍

udf/metadata.go
rraulinio: 👍

udf/metadata.go
rraulinio: 👍

udf/types.go
rraulinio: Thanks for checking this. I agree the table type renderers commonly use `decimal(9, 2)`, but the UDF metadata spec has a stricter serialization rule here. In the UDF spec's Types section, primitive/semi-structured type strings use examples like `decimal(9,2)` and it explicitly says: "Type strings must contain no spaces or quote characters." The Definition ID section also says the canonical string is formatted with no spaces: https://iceberg.apache.org/udf-spec/#types

So for UDF metadata, `decimal(9, 2)` is invalid input per the UDF spec, even though it is a familiar table-schema rendering. Normalizing through iceberg-go's parsed `DecimalType.String()` would actually move this package in the wrong direction, because that returns the spaced table form and would make us write spec-invalid UDF type strings / definition-ids. I kept this strict and test-pinned so definition IDs stay on the UDF spec's canonical no-space form.

zeroshade (approved):

PR #1419, opened 2026-07-06
fix(table): merge deletion vectors on repeated merge-on-read deletes

### LINKED ISSUE #1372: consecutive `Transaction.Delete` calls on the same file create two live deletion vectors
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

When using v3 tables with merge-on-read mode, old DV are not cleaned up.

Iceberg requires at most one deletion vector (DV) per data file per snapshot. A second `Delete()` touching a data file that already has a DV writes a brand-new Puffin/DV file instead of merging into the existing one, and never removes the old one. The commit succeeds, but the table is now invalid — any subsequent scan fails with `can't index multiple deletion vectors for <path>`.

### Minimal reproducer

```go
func TestDoubleDeleteRegistersTwoDVsForSameFile(t *testing.T) {
	ctx := context.Background()
	tbl := newV3MergeOnReadTestTable(t) // format-version=3, write.delete.mode=merge-on-read

	arrowSchema := arrow.NewSchema([]arrow.Field{
		{Name: "id", Type: arrow.PrimitiveTypes.Int64, Nullable: false},
		{Name: "data", Type: arrow.BinaryTypes.String, Nullable: true},
	}, nil)
	data, _ := array.TableFromJSON(memory.DefaultAllocator, arrowSchema, []string{
		`[{"id":1,"data":"a"},{"id":2,"data":"b"},{"id":3,"data":"c"},{"id":4,"data":"d"},{"id":5,"data":"e"}]`,
	})
	tbl, _ = tbl.Append(ctx, array.NewTableReader(data, -1), nil)

	// First delete writes DV #1 against the single data file.
	tbl, _ = tbl.Delete(ctx, iceberg.EqualTo(iceberg.Reference("id"), int64(2)), nil)

	// Second delete SHOULD merge into DV #1. Instead it writes DV #2
	// against the same data file, and DV #1 is never removed.
	tbl, err := tbl.Delete(ctx, iceberg.EqualTo(iceberg.Reference("id"), int64(4)), nil)
	require.NoError(t, err) // commit "succeeds" — the corruption is silent here

	// Scanning the now-invalid table blows up:
	_, itr, _ := tbl.Scan().ToArrowRecords(ctx)
	for _, err := range itr {
		require.NoError(t, err) // fails: "can't index multiple deletion vectors for ..."
	}
}
```

`newV3MergeOnReadTestTable` is a one-line variant of [`newMergeOnReadTestTable`](https://github.com/debugmiller/iceberg-go/blob/da94843/table/mor_delete_pruning_test.go#L75-L96), with `PropertyFormatVersion: "3"` instead of `"2"`.

### Expected vs. actual

- DVs referencing the data file after 2 deletes — expected **1** (merged bitmap); actual **2** (both live)
- Scan after 2 deletes — expected: succeeds, 3 rows remain; actual: errors with `can't index multiple deletion vectors for <path>`

### Root cause

- [`Transaction.classifyFilesForFilteredDeletions`](https://github.com/debugmiller/iceberg-go/blob/da94843/table/transaction.go#L1543-L1663) only walks data-file manifest entries — [line 1606](https://github.com/debugmiller/iceberg-go/blob/da94843/table/transaction.go#L1606) skips anything that isn't `EntryContentData` — so it never looks up a data file's existing DV.
- [`positionDeleteRecordsToDataFilesDV`](https://github.com/debugmiller/iceberg-go/blob/da94843/table/arrow_utils.go#L1845-L1898) always builds a fresh [`dv.NewDVWriter()`](https://github.com/debugmiller/iceberg-go/blob/da94843/table/arrow_utils.go#L1847) with no prior bitmap.
- The new DV is appended via [`updater.appendDeleteFile(f)`](https://github.com/debugmiller/iceberg-go/blob/da94843/table/transaction.go#L1907) without ever calling [`updater.removeDeletionVector(...)`](https://github.com/debugmiller/iceberg-go/blob/da94843/table/snapshot_producers.go#L608-L616) on the superseded one — even though that method already exists and is used correctly by the compaction/`ReplaceFiles` path.
- The read side already guards against the invalid state — [`buildDVIndex`](https://github.com/debugmiller/iceberg-go/blob/da94843/table/scanner.go#L561-L576) rejects a second DV for the same referenced file — so the failure surfaces at scan time, not commit time.

tanmayrauth:
A second merge-on-read DELETE on a data file that already had a deletion vector wrote a second live DV instead of merging, violating the one-DV-per -data-file invariant and failing subsequent scans.

The v3 delete path now folds the existing DV's positions into the new one and supersedes the old DV, mirroring the compaction/ReplaceFiles path.

Fixes #1372

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1851,16 +1851,20 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
 
 		// Seed the writer with any deletion vector the referenced data file
 		// already carries so its previously deleted positions survive into the
-		// merged DV. The read path that produced args.itr already applied these
-		// existing deletes, so the incoming batches hold only the new positions;
-		// without the seed the replacement DV would silently resurrect rows. The
-		// spec allows at most one DV per data file, so the caller supersedes the
-		// old DV once this merged one is written.
+		// merged DV. The scan that produced args.itr does NOT apply the existing
+		// DVs (makePositionDeleteRecordsForFilter builds its FileScanTasks
+		// without DeleteFiles), so the incoming batches may re-report positions
+		// already in the old DV; re-Setting them on top of the seed is a no-op
+		// because the bitmap is idempotent. The seed is what guarantees the old
+		// positions survive — do not remove it, or a second delete would drop
+		// the earlier deletes and resurrect those rows. The spec allows at most
+		// one DV per data file, so the caller supersedes the old DV once this
+		// merged one is written.
 		hasEntries := false
 		for filePath, bitmap := range args.existingDVs {
 			pCtx, ok := partitionContextByFilePath[filePath]
 			if !ok {
-				yield(nil, fmt.Errorf("unexpected missing partition context for path %s", filePath))
+				yield(nil, fmt.Errorf("unexpected missing partition context for existing deletion vector path %s", filePath))
 
 				return
 			}
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1633,6 +1633,11 @@ type recordWritingArgs struct {
 	maxWriteWorkers int
 	clustered       bool
 	factoryOpts     []writerFactoryOption
+	// existingDVs maps a data file path to the positions already recorded in
+	// its current deletion vector. On the v3 DV write path these are folded
+	// into the newly written DV so a data file that already had a DV ends up
+	// with a single merged vector (the caller removes the superseded one).
+	existingDVs map[string]*dv.RoaringPositionBitmap
 }
 
 func recordsToDataFiles(ctx context.Context, rootLocation string, meta *MetadataBuilder, args recordWritingArgs) (ret iter.Seq2[iceberg.DataFile, error]) {
@@ -1844,7 +1849,25 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
 			return metadata.PartitionSpecByID(int(id))
 		})
 
+		// Seed the writer with any deletion vector the referenced data file
+		// already carries so its previously deleted positions survive into the
+		// merged DV. The read path that produced args.itr already applied these
+		// existing deletes, so the incoming batches hold only the new positions;
+		// without the seed the replacement DV would silently resurrect rows. The
+		// spec allows at most one DV per data file, so the caller supersedes the
+		// old DV once this merged one is written.
 		hasEntries := false
+		for filePath, bitmap := range args.existingDVs {
+			pCtx, ok := partitionContextByFilePath[filePath]
+			if !ok {
+				yield(nil, fmt.Errorf("unexpected missing partition context for path %s", filePath))
+
+				return
+			}
+			writer.Load(filePath, bitmap, pCtx.specID, pCtx.partitionData)
+			hasEntries = true
+		}
+
 		for batch, err := range args.itr {
 			if err != nil {
 				yield(nil, err)
--- a/table/dv/dv_writer.go
+++ b/table/dv/dv_writer.go
@@ -128,12 +128,20 @@ func (w *DVWriter) Add(dataFilePath string, positions []int64, specID int32, par
 // file so subsequent Add calls merge new positions into it. The spec permits
 // at most one DV per data file per snapshot, so when a data file that already
 // has a DV receives more deletes the old positions must be carried into the
-// replacement DV (and the old DV superseded) rather than dropped. Load must be
-// called before any Add for the same path; specID and partitionData are
-// captured exactly as Add's first call would, and the supplied bitmap is
-// unioned into the entry. A nil bitmap simply registers the entry so it is
-// re-emitted. Callers retain ownership of the passed bitmap; only its set
-// positions are copied.
+// replacement DV (and the old DV superseded) rather than dropped.
+//
+// Load is intended to be called before any Add for the same path: on the first
+// call for a path it captures specID and partitionData exactly as Add would and
+// unions the supplied bitmap into a fresh entry. If an entry already exists
+// (because Add or Load ran first for this path), Load only unions the bitmap in
+// and the earlier specID/partitionData are kept — the later values are ignored,
+// mirroring Add's first-write-wins behavior. Callers must not pass conflicting
+// partition values across calls for the same path.
+//
+// bitmap must be non-nil; its set positions are copied and the caller retains
+// ownership. (A nil bitmap would register a cardinality-0 entry, which is never
+// useful — collectExistingDVs always supplies a non-nil bitmap read from the
+// existing DV.)
 func (w *DVWriter) Load(dataFilePath string, bitmap *RoaringPositionBitmap, specID int32, partitionData map[int]any) {
 	entry, ok := w.entries[dataFilePath]
 	if !ok {
--- a/table/dv/dv_writer.go
+++ b/table/dv/dv_writer.go
@@ -124,6 +124,31 @@ func (w *DVWriter) Add(dataFilePath string, positions []int64, specID int32, par
 	return nil
 }
 
+// Load seeds the writer with an already-written deletion vector for a data
+// file so subsequent Add calls merge new positions into it. The spec permits
+// at most one DV per data file per snapshot, so when a data file that already
+// has a DV receives more deletes the old positions must be carried into the
+// replacement DV (and the old DV superseded) rather than dropped. Load must be
+// called before any Add for the same path; specID and partitionData are
+// captured exactly as Add's first call would, and the supplied bitmap is
+// unioned into the entry. A nil bitmap simply registers the entry so it is
+// re-emitted. Callers retain ownership of the passed bitmap; only its set
+// positions are copied.
+func (w *DVWriter) Load(dataFilePath string, bitmap *RoaringPositionBitmap, specID int32, partitionData map[int]any) {
+	entry, ok := w.entries[dataFilePath]
+	if !ok {
+		entry = &dvEntry{
+			bitmap:        NewRoaringPositionBitmap(),
+			specID:        specID,
+			partitionData: maps.Clone(partitionData),
+		}
+		w.entries[dataFilePath] = entry
+		w.order = append(w.order, dataFilePath)
+	}
+
+	entry.bitmap.Or(bitmap)
+}
+
 // Flush writes one Puffin file containing one blob per data file, and returns
 // manifest entries ready for RowDelta.AddDeletes(). Each output DataFile
 // carries the partition spec and partition record of the data file it
--- a/table/mor_delete_pruning_test.go
+++ b/table/mor_delete_pruning_test.go
@@ -75,6 +75,14 @@ func TestMergeOnReadDeleteAcrossPrunedRowGroups(t *testing.T) {
 func newMergeOnReadTestTable(t *testing.T) *table.Table {
 	t.Helper()
 
+	return newMergeOnReadTestTableVersion(t, "2")
+}
+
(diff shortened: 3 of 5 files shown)


tanmayrauth:

table/transaction.go
```diff
@@ -1913,9 +1939,50 @@ func (t *Transaction) writePositionDeletesForFiles(ctx context.Context, fs io.IO
...
+		return nil, nil
+	}
+
+	wanted := make(map[string]struct{}, len(files))
+	for _, df := range files {
+		wanted[df.FilePath()] = struct{}{}
+	}
+
+	result := make(map[string]iceberg.DataFile)
+	for df, err := range s.dataFiles(fs, nil) {
```
laskoviymishka: `s.dataFiles(fs, nil)` runs with `discardDeleted=false`, so this pulls in DELETED-status entries too. After a second delete, the snapshot's deleted-entries manifest still carries the superseded old DV as a DELETED entry for the same referenced data file. On a third delete this loop can yield both the live ADDED merged DV and the stale DELETED one for the same path, and `result[*ref] = df` is last-write-wins into a map — with the current manifest concat order (added, then position deletes, then deleted-files) the DELETED ghost lands last and wins. That seeds the new DV from a stale bitmap (resurrecting positions deleted in the 2nd commit), calls `removeDeletionVector` on a DV that no longer exists, and writes a Puffin missing the merged positions.

The guard is already in the PR — `liveDVCount` in the test skips `entry.Status() == EntryStatusDELETED`. I'd give `collectExistingDVs` the same treatment: iterate `s.entries(fs, iceberg.ManifestContentDeletes)` and `continue` on DELETED status, which also narrows the scan to delete manifests instead of walking everything. wdyt?

table/mor_delete_pruning_test.go
```diff
@@ -95,6 +103,79 @@ func newMergeOnReadTestTable(t *testing.T) *table.Table {
...
+
+	// Second delete on the same data file must merge into DV #1, not add a
+	// second live DV.
+	tbl, err = tbl.Delete(ctx, iceberg.EqualTo(iceberg.Reference("id"), int64(4)), nil)
+	require.NoError(t, err)
+
+	require.Equal(t, 1, liveDVCount(t, tbl),
+		"the two deletes must collapse into a single merged deletion vector")
+	assert.Equal(t, []int64{1, 3, 5}, idsInTable(t, tbl),
+		"both id=2 and id=4 must be deleted and no previously deleted row may resurrect")
```
laskoviymishka, line 144: This stops at two deletes, but the ghost-DV bug above only triggers on the third — the second delete is what writes the DELETED entry the third then picks up. Could we add a triple-delete variant (append, delete id=2, delete id=4, delete id=1, assert `liveDVCount == 1` and rows `[3, 5]`)? That's the test that would've caught this.

While we're here, a partitioned / multi-data-file case would exercise the per-file `pCtx` capture in the seed loop, which the single unpartitioned file doesn't touch today.

table/arrow_utils.go
```diff
@@ -1848,7 +1853,25 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
 			return metadata.PartitionSpecByID(int(id))
 		})
 
+		// Seed the writer with any deletion vector the referenced data file
+		// already carries so its previously deleted positions survive into the
+		// merged DV. The read path that produced args.itr already applied these
```
laskoviymishka: This isn't quite right — `makePositionDeleteRecordsForFilter` builds the `FileScanTask` without `DeleteFiles`, so the scan does not apply the existing DVs; the incoming batches can carry already-deleted positions. Correctness still holds, but only because `bitmap.Set` is idempotent, so re-setting old positions on top of the seed is a no-op.

I'd reword to say the scan does not apply existing deletes and the seed is what guarantees the old positions survive. As written, someone later reads this, concludes the seed is redundant, removes it, and the original bug is back.

table/arrow_utils.go
```diff
@@ -1848,7 +1853,25 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
...
 		hasEntries := false
+		for filePath, bitmap := range args.existingDVs {
+			pCtx, ok := partitionContextByFilePath[filePath]
+			if !ok {
+				yield(nil, fmt.Errorf("unexpected missing partition context for path %s", filePath))
+
+				return
+			}
+			writer.Load(filePath, bitmap, pCtx.specID, pCtx.partitionData)
+			hasEntries = true
```
laskoviymishka, line 1872: `hasEntries` is set true for every existing-DV path even when that file gets no new delete records this commit. If we seed a file but `posDeleteRecIter` yields nothing for it, we still flush an identical-replacement Puffin and remove the old DV — churn for no change. Gating the flush on whether records were actually added (or `!bitmap.IsEmpty()`) would avoid rewriting an unchanged DV. Latent, not a correctness bug.

table/transaction.go
```diff
@@ -1913,9 +1939,50 @@ func (t *Transaction) writePositionDeletesForFiles(ctx context.Context, fs io.IO
...
+		if err != nil {
+			return nil, err
```
laskoviymishka: Every other error return in this function wraps with `%w` — this one returns the iterator error bare. `fmt.Errorf("scanning existing deletion vectors: %w", err)` keeps it consistent.

table/transaction.go
```diff
@@ -1891,11 +1892,36 @@ func (t *Transaction) writePositionDeletesForFiles(ctx context.Context, fs io.IO
...
+	// write a second live DV, corrupting the table (buildDVIndex rejects it).
+	var existingDVs map[string]iceberg.DataFile
+	if t.meta.formatVersion >= 3 {
+		existingDVs, err = t.collectExistingDVs(fs, files)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	existingDVBitmaps := make(map[string]*dv.RoaringPositionBitmap, len(existingDVs))
```
laskoviymishka: The `formatVersion >= 3` guard covers the `collectExistingDVs` call, but this alloc + `ReadDV` loop still runs on v1/v2 (harmless — `existingDVs` is nil so the loop is len-0). Moving the `make` and loop inside the v3 block keeps the whole existing-DV machinery behind one guard.

table/arrow_utils.go
```diff
@@ -1848,7 +1853,25 @@ func positionDeleteRecordsToDataFilesDV(ctx context.Context, rootLocation string
...
+		// merged DV. The read path that produced args.itr already applied these
+		// existing deletes, so the incoming batches hold only the new positions;
+		// without the seed the replacement DV would silently resurrect rows. The
+		// spec allows at most one DV per data file, so the caller supersedes the
+		// old DV once this merged one is written.
 		hasEntries := false
+		for filePath, bitmap := range args.existingDVs {
+			pCtx, ok := partitionContextByFilePath[filePath]
+			if !ok {
+				yield(nil, fmt.Errorf("unexpected missing partition context for path %s", filePath))
```
laskoviymishka: This message is identical to the one in the batch path just below, so in logs there's no telling which fired. A distinguishing string for the seed path would save some head-scratching later.

table/dv/dv_writer.go
```diff
@@ -108,6 +108,31 @@ func (w *DVWriter) Add(dataFilePath string, positions []int64, specID int32, par
...
+// file so subsequent Add calls merge new positions into it. The spec permits
+// at most one DV per data file per snapshot, so when a data file that already
+// has a DV receives more deletes the old positions must be carried into the
+// replacement DV (and the old DV superseded) rather than dropped. Load must be
+// called before any Add for the same path; specID and partitionData are
+// captured exactly as Add's first call would, and the supplied bitmap is
+// unioned into the entry. A nil bitmap simply registers the entry so it is
+// re-emitted. Callers retain ownership of the passed bitmap; only its set
+// positions are copied.
+func (w *DVWriter) Load(dataFilePath string, bitmap *RoaringPositionBitmap, specID int32, partitionData map[int]any) {
```
laskoviymishka, line 145: The doc says Load "must be called before any Add for the same path," but nothing enforces that — call it after Add, or twice, and it silently ORs into the live entry and discards the later `specID`/`partitionData`. Idempotent OR makes it harmless today, but the doc promises a precondition the code doesn't hold. I'd either error when the entry already exists, or soften the doc to describe what actually happens.

Also "a nil bitmap simply registers the entry" — `ReadDV` always returns non-nil, so that path never runs, and if it did it'd register a cardinality-0 DV. I'd document that `bitmap` must be non-nil instead.

laskoviymishka (changes requested):
This is a real fix for a real bug, and the seed-then-supersede approach is the right one — fold the existing DV into the new one and remove the old rather than letting a second DV accumulate. The double-delete test and the `liveDVCount` helper are a good foundation.

I'd hold it before merging though. `collectExistingDVs` reads existing DVs with `discardDeleted=false`, so it also picks up DELETED-status entries. After a second delete, the snapshot's deleted-entries manifest still carries the superseded old DV as a DELETED entry against the same data file — so on a third delete the collection can return both the live merged DV and the stale ghost for one path. It's a last-write-wins map, and with the current manifest concat order the ghost wins: we seed from a stale bitmap and resurrect rows that the second delete removed, then try to remove a DV that's already gone. The nice part is the fix is already in this PR — `liveDVCount` skips `EntryStatusDELETED` and `collectExistingDVs` just needs the same guard (switching to `s.entries(fs, iceberg.ManifestContentDeletes)` gets you both the status visibility and a narrower scan).

The matching gap is the test: the current one stops at two deletes, but this only breaks on the third. A triple-delete case would have caught it and belongs alongside the fix.

- skip DELETED-status entries in `collectExistingDVs`
- add a triple-delete test (and ideally a partitioned / multi-file case)
- fix the seed comment in `arrow_utils.go` — the scan does not apply existing DVs, so the comment currently invites removing the seed and reintroducing the bug

The rest (error wrapping, the v1/v2 alloc, the duplicate error message, the Load doc's unenforced precondition) are small and inline. Once the blocker and the test are in, happy to take another pass and approve.

tanmayrauth:
@laskoviymishka Thanks for the thorough pas, all three must-haves are in, plus the inline items.

- Ghost DELETED DV (blocker): collectExistingDVs now iterates s.entries(fs, iceberg.ManifestContentDeletes) and skips EntryStatusDELETED, so the superseded DV can't win the last-write into the result map. This also narrows the scan to delete manifests.
- Triple-delete test: added TestV3MergeOnReadTripleDeleteMergesDeletionVector (append, delete 2/4/1 → liveDVCount == 1, rows [3, 5]). Confirmed it fails on the pre-fix code with exactly the symptom you described — id=4 resurrects ([3, 4, 5]) — and passes with the fix.
- Partitioned case: added TestV3MergeOnReadDoubleDeletePartitionedMergesDeletionVectors, an identity-partitioned table where a single delete touches two data files that each already carry a DV, exercising the per-file pCtx capture in the seed loop.
- Seed comment: reworded to state the scan does not apply existing DVs; correctness holds because bitmap.Set is idempotent, and the seed is what preserves the old positions.
- Also handled: error wrapping, the v1/v2 alloc moved behind the v3 guard, the duplicated error message, and the Load doc.

One I left for a follow-up, the hasEntries churn. A correct fix has to skip both the re-flush and the  removeDeletionVector for a seeded-but-unchanged file, but the Puffin flush is atomic across all files, so per-file gating couples the writer with the removal loop; done wrong it removes an old DV without   re-emitting it — the resurrection bug we just fixed. Since it's no-op churn and not a correctness issue, I'd rather land it separately. Happy to fold it in here if you'd prefer.

laskoviymishka (approved):
nice! ❤️

PR #1421, opened 2026-07-07
feat(metrics): add report and wire types (#1236)

tanmayrauth:
Second PR of the metrics reporting framework. Adds the spec-compliant report types and their JSON wire format: ScanReport, CommitReport, the ScanMetricsResult/CommitMetricsResult metric sets, CounterResult and TimerResult, plus the ReportMetricsRequest envelope with the flattened report-type discriminator.

Metric names, units, and the request shape match Java verbatim and are locked by round-trip interop tests against fixtures in the Java wire format. ScanReport.filter defaults to the always-true expression for v1; unknown metric keys are tolerated.

Related: #1236

### DIFF
--- a/metrics/commit_report.go
+++ b/metrics/commit_report.go
@@ -1,0 +1,81 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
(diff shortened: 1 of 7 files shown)


tanmayrauth:
@laskoviymishka  PTAL when you get chance

zeroshade (approved):

PR #1423, opened 2026-07-08
fix(io): S3 compatibility with GCS HMAC endpoints

### LINKED ISSUE #1422: S3 IO cannot write to GCS HMAC S3-compatible interop (SignatureDoesNotMatch)
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

When the S3 IO backend (`io/gocloud`) is pointed at a non-AWS S3-compatible endpoint that supports the AWS SigV4 HMAC interop API but is **not** Amazon S3 itself — most notably **Google Cloud Storage's S3-compatible interop endpoint (`https://storage.googleapis.com`)** with HMAC keys — every write fails with:

```
operation error S3: PutObject, https response error StatusCode: 403,
RequestID: , HostID: ,
api error SignatureDoesNotMatch: Access denied.
```

(Empty `RequestID`/`HostID` is characteristic of the GCS interop response.)

This affects any iceberg-go user with a REST catalog (Lakekeeper, Polaris, etc.) whose warehouse is on GCS via HMAC Keys.

badalprasadsingh:
## Description

When the S3 IO backend (`io/gocloud`) is pointed at a non-AWS S3-compatible endpoint that supports the AWS SigV4 HMAC interop API but is **not** Amazon S3 itself — most notably **Google Cloud Storage's S3-compatible interop endpoint
(`https://storage.googleapis.com`)** with HMAC keys.

Fixes #1422 

## Proposed Fix

* When `props["s3.endpoint"]` (`io.S3EndpointURL`) is set, configure the S3 client to use `aws.RequestChecksumCalculationWhenRequired` so the SDK doesn't automatically enable CRC32 for requests where it isn't required.
* Add an **Initialize** middleware (`iceberg-go/strip-s3-input-checksum-algorithm`) **before** `AWSChecksum:SetupInputContext` to clear `ChecksumAlgorithm` on `PutObject`, `UploadPart`, and `CreateMultipartUpload`, preventing chunked encoding from being enabled for compatible S3 endpoints.
* Add a **Finalize** middleware (`iceberg-go/strip-gcs-incompatible-signed-headers`) **before** SigV4 signing to remove `Amz-Sdk-Invocation-Id`, `Amz-Sdk-Request`, and `Accept-Encoding` from the signed request, avoiding signature mismatches with GCS S3-compatible endpoints while leaving the AWS S3 code path unchanged.


io/gocloud/s3.go
```diff
@@ -156,6 +158,70 @@ func applyS3TransportTuning(t *http.Transport) {
...
+		return stack.Initialize.Add(m, smithymiddleware.Before)
+	}
+
+	return nil
+}
+
+var gcsIncompatibleSignedHeaders = []string{
+	"Amz-Sdk-Invocation-Id",
+	"Amz-Sdk-Request",
+	"Accept-Encoding",
```
laskoviymishka, line 251: I think there's a read-side regression hiding in the `Accept-Encoding` entry here.

`stripGCSIncompatibleSignedHeaders` is registered before `Signing`, and `Signing` runs on every op — `GetObject`, `ListObjectsV2`, not just writes. So this deletes `Accept-Encoding` on reads too. The SDK installs `DisableAcceptEncodingGzip`, which sets `Accept-Encoding: identity` specifically so Go's `http.Transport` won't auto-negotiate gzip and transparently decompress the body. Deleting the header entirely (rather than just omitting it from the signed set) means `net/http` is free to add `Accept-Encoding: gzip` itself and silently decompress the response — which corrupts the Content-Length-bounded footer/range reads we lean on all over the Parquet/Avro/manifest read path against any endpoint that gzips.

Two ways out: scope this middleware to write ops with the same input-type switch `stripS3InputChecksumAlgorithm` uses, or re-set `Accept-Encoding: identity` after deleting it instead of leaving it absent. The first feels cleaner since the GCS signing problem is a write concern. wdyt?

io/gocloud/s3.go
```diff
@@ -156,6 +158,70 @@ func applyS3TransportTuning(t *http.Transport) {
 // path-style addressing. It defaults to virtual-hosted style for
 // standard AWS S3 and path-style for custom endpoints (e.g. MinIO).
 // The s3.force-virtual-addressing property can override either default.
+func applyS3ClientOptions(endpoint string, props map[string]string) func(*s3.Options) {
+	return func(o *s3.Options) {
+		if endpoint != "" {
```
laskoviymishka, line 202: The trigger being `endpoint != ""` treats every custom endpoint as if it were GCS, and I don't think that holds.

MinIO, R2, Ceph, and LocalStack all support SigV4 request checksums and don't choke on the `Amz-Sdk-*` signed headers, so this quietly disables checksum-on-write for all of them — and it does it by overloading `s3.endpoint`/`AWS_S3_ENDPOINT`, a stable documented property, so every existing MinIO/R2 deployment picks up the new behavior on upgrade with no way to opt out. PyIceberg and iceberg-rust both gate this behind an explicit "S3-compatible mode" knob rather than inferring it from the endpoint being set.

I'd gate the workaround behind either a dedicated property (`s3.compat-mode` / `s3.disable-checksum-headers`, defaulting off) or a `storage.googleapis.com` host check, so we're not changing behavior for backends that don't need it. Which direction do you prefer?

laskoviymishka (changes requested):

iremcaginyurtturk:
Hi @badalprasadsingh. I'm hitting the same GCS HMAC SignatureDoesNotMatch issue (#1422) and verified this branch fixes it end-to-end. Are you planning to continue with the review feedback? If you're short on time, I'm happy to help push it forward. Thanks for the work on this!

badalprasadsingh:
Hey @iremcaginyurtturk, thanks for verifying the fix! Yes, I'll be continuing with the review feedback. I've been heavily short on time over the past few weeks, but I should be able to get back to it soon. Thanks for offering to help—I really appreciate it!

iremcaginyurtturk:
Sounds good, thanks @badalprasadsingh

io/gocloud/s3.go
badalprasadsingh: Done.

Scoped the header stripping to write ops only.

io/gocloud/s3.go
badalprasadsingh: Done.

Went with the dedicated-property route you suggested: added `s3.compat-mode` (default off), so `s3.endpoint`/`AWS_S3_ENDPOINT` behavior is unchanged.

laskoviymishka (approved):
:shipit:

PR #1425, opened 2026-07-08
fix(table): apply optionality updates to list elements and map values in UpdateSchema

### LINKED ISSUE #1424: applyChanges ignores optionality updates for list elements and map values
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

In Iceberg-Go, `applyChanges.List` and `applyChanges.Map` ignore optionality updates registered for list elements and map values.

While type promotions are correctly applied, changes to a list element's `Required` flag or a map value's `Required` flag are silently dropped because `ElementRequired` and `ValueRequired` are always copied from the original schema instead of using the registered updates.

badalprasadsingh:
## Description

`applyChanges.List` and `applyChanges.Map` silently dropped _optionality_ updates registered against a list's element or a map's value. 

Fixes #1424 

## Test

Added `TestApplyChangesNestedOptionalityUpdates`

### DIFF
--- a/table/update_schema.go
+++ b/table/update_schema.go
@@ -830,11 +830,13 @@ func (a *applyChanges) Field(field iceberg.NestedField, fieldResult iceberg.Type
 	return fieldResult
 }
 
-// updatedRequired returns the effective Required flag for fieldID.
-//
-// Note: updateColumn seeds each pending update from a copy of the original
+// updateColumn seeds each pending update from a copy of the original
 // NestedField. As a result, updates that only modify other properties
 // still preserve the original Required value here.
+// 
+// Note: If updateColumn is refactored to lazily populate only the
+// touched fields, this helper must gain an explicit "Required was set"
+// signal or it will silently return the zero-value false.
 func (a *applyChanges) updatedRequired(fieldID int, fallback bool) bool {
 	if update, ok := a.updates[fieldID]; ok {
 		return update.Required
--- a/table/update_schema.go
+++ b/table/update_schema.go
@@ -830,21 +830,29 @@ func (a *applyChanges) Field(field iceberg.NestedField, fieldResult iceberg.Type
 	return fieldResult
 }
 
+// updatedRequired returns the effective Required flag for fieldID.
+//
+// Note: updateColumn seeds each pending update from a copy of the original
+// NestedField. As a result, updates that only modify other properties
+// still preserve the original Required value here.
+func (a *applyChanges) updatedRequired(fieldID int, fallback bool) bool {
+	if update, ok := a.updates[fieldID]; ok {
+		return update.Required
+	}
+
+	return fallback
+}
+
 func (a *applyChanges) List(listType iceberg.ListType, elementResult iceberg.Type) iceberg.Type {
 	elementType := a.Field(listType.ElementField(), elementResult)
 	if elementType == nil {
 		panic(fmt.Sprintf("cannot delete element type from list: %s", elementResult))
 	}
 
-	elementRequired := listType.ElementRequired
-	if update, ok := a.updates[listType.ElementID]; ok {
-		elementRequired = update.Required
-	}
-
 	return &iceberg.ListType{
 		ElementID:       listType.ElementID,
 		Element:         elementType,
-		ElementRequired: elementRequired,
+		ElementRequired: a.updatedRequired(listType.ElementID, listType.ElementRequired),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


badalprasadsingh:
@laskoviymishka, please take a look whenever you are free

table/update_schema.go
```diff
@@ -836,10 +836,15 @@ func (a *applyChanges) List(listType iceberg.ListType, elementResult iceberg.Typ
 		panic(fmt.Sprintf("cannot delete element type from list: %s", elementResult))
 	}
 
+	elementRequired := listType.ElementRequired
+	if update, ok := a.updates[listType.ElementID]; ok {
```
laskoviymishka: this reads `update.Required` whenever any update record exists, which is correct today only because `updateColumn` initializes the update from a copy of the original field — so for a type-promotion-only or doc-only change, `update.Required` happens to equal the original optionality.

It works, but the correctness lives in a detail two functions away rather than here. If `updateColumn` ever switches to lazily populating only the touched fields, this silently starts writing a zero-value `false`. I'd either mirror `Struct()` (which copies the field wholesale) or gate on an explicit "was Required set" signal so this site can tell the difference.

Same shape at line 877 for `mapType.ValueID`. Since `Struct()`, `List()`, and `Map()` now all do this lookup, could we pull it into a small `resolveRequired(fieldID, original)` helper? wdyt?

table/update_schema_test.go
```diff
@@ -1043,6 +1043,192 @@ func TestAddColumnMonotonicFieldIDs(t *testing.T) {
...
+
+		tags, ok := newSchema.FindFieldByName("tags")
+		assert.True(t, ok)
+		listType, ok := tags.Type.(*iceberg.ListType)
+		assert.True(t, ok)
+		assert.False(t, listType.ElementRequired,
+			"list element optionality update must be reflected in the applied schema")
+	})
+
+	t.Run("optional -> required is applied with allowIncompatibleChanges", func(t *testing.T) {
```
laskoviymishka, line 1078: We cover optional→required *with* `allowIncompatibleChanges`, but never assert the rejection when it's off — and that guard (`update_schema.go:463`) is the whole safety contract for tightening nullability on elements/values. Java and Py both test this rejection on nested element/value fields; we test the apply side but not the guard-rail.

The worry is a future refactor that moves or misroutes the check downstream: the struct-field error tests would still pass while list/map silently regressed. A subtest per List/Map like:

```go
_, err := NewUpdateSchema(table.NewTransaction(), true, false).
    UpdateColumn([]string{"tags", "element"}, ColumnUpdate{
        Required: iceberg.Optional[bool]{Val: true, Valid: true},
    }).Apply()
assert.Error(t, err)
assert.Contains(t, err.Error(), "cannot change column nullability from optional to required")
```

would pin it. A doc-only-update subtest (registers an update but leaves `Required.Valid` false) would also lock in the fallback behavior from the comment on `update_schema.go`.

table/update_schema_test.go
```diff
@@ -1043,6 +1043,192 @@ func TestAddColumnMonotonicFieldIDs(t *testing.T) {
...
+		table := New([]string{"id"}, meta, "", nil, nil)
+		newSchema, err := NewUpdateSchema(table.NewTransaction(), true, false).
+			UpdateColumn([]string{"tags", "element"}, ColumnUpdate{
+				Required: iceberg.Optional[bool]{Val: false, Valid: true},
+			}).Apply()
+		assert.NoError(t, err)
+
+		tags, ok := newSchema.FindFieldByName("tags")
+		assert.True(t, ok)
+		listType, ok := tags.Type.(*iceberg.ListType)
```
laskoviymishka, line 1072: `assert.True` doesn't stop the test, so on a regression where `ok` is false we fall through to a nil `listType` and the next line panics on the deref — a confusing stack trace instead of a clean failure message. This repeats in all eight subtests.

I'd switch the two guard checks in each to `require.True(t, ok, ...)` (add the `testify/require` import) so a bad assertion fails cleanly right there.

laskoviymishka (changes requested):
Nice fix, this is a faithful port of the Java/Py `ApplyChanges.list()/map()` pattern, and it closes a real gap where element and map-value optionality updates were being silently dropped. The apply-side coverage is thorough and the interop story is clean.

I'd hold it before merge on one thing: the suite never asserts the guard-rail. optional→required *without* `allowIncompatibleChanges` should be rejected, and right now nothing exercises that for element/value paths — so a future refactor that moved the check downstream could regress list/map while the struct-field tests stayed green. Java and Py both test this rejection on nested fields; I left a suggested subtest inline.

- add a negative test per List/Map: optional→required with `allowIncompatibleChanges` off asserts an error
- use `require.True` (not `assert.True`) for the `ok` guards in the new subtests, so a regression fails cleanly instead of panicking on a nil deref
- optional, but worth a look: the `update.Required` read is correct only by construction — a small guard or a shared helper across `Struct()/List()/Map()` would make that explicit

Once the coverage is in, happy to take another pass and approve.

laskoviymishka (approved):
👍

PR #1427, opened 2026-07-09
fix(table): fix occ rewrite resurrection

twuebi:
This fixes manifest-list replay after optimistic-concurrency retries. On retry, snapshot producers now rebuild parent-dependent manifests against the fresh branch head while reusing already-written added-content manifests. That prevents overwrite/replace retries from resurrecting removed data files or dropping concurrently committed files. The rebuild also refreshes sequence numbers, v3 row-lineage fields, and snapshot summary totals from latest metadata. If a file planned for rewrite is no longer live, replay now aborts so callers can re-plan safely. Regression coverage adds an end-to-end replace-after-conflict scenario plus focused retry-rebuild tests.

### DIFF
--- a/table/occ_scenario_test.go
+++ b/table/occ_scenario_test.go
@@ -451,6 +451,129 @@ func (s *OCCScenarioTestSuite) TestV3MergeAppendRowLineageAfterOCCRetry() {
 	}
 }
 
+// TestReplaceFilesNoResurrectionAfterConflict is a regression test for the OCC
+// retry-replay resurrection bug. When a replace/overwrite commit retries after
+// a conflict, the rebuild must re-apply the file removal against the FRESH
+// parent — not graft attempt-0's stale "own" manifests onto a fresh parent
+// that still lists the removed files. Otherwise the removed files come back
+// alive next to their replacement and their rows are duplicated.
+//
+// Scenario:
+//   - Base: file0 = rows [1,2,3] committed.
+//   - Concurrent peer: appends filePeer = row [4]; catalog head is file0+peer.
+//   - Writer A (stale base = only file0) replaces file0 with a compacted file
+//     holding the same rows [1,2,3]. First attempt conflicts; the retry rebases
+//     onto the fresh head.
+//
+// Correct: live rows = {1,2,3,4} (compacted + peer), file0 dropped → 4 rows.
+// Buggy:   file0 resurrected → rows [1,2,3] duplicated → 7 rows, 3 live files.
+func (s *OCCScenarioTestSuite) TestReplaceFilesNoResurrectionAfterConflict() {
+	props := iceberg.Properties{
+		table.CommitMinRetryWaitMsKey:      "0",
+		table.CommitMaxRetryWaitMsKey:      "0",
+		table.CommitTotalRetryTimeoutMsKey: "60000",
+		table.CommitNumRetriesKey:          "2",
+		// Snapshot isolation lets the replace commit past the concurrent
+		// append the same way compaction's rewrite semantics do; the bug we
+		// are guarding lives in the retry rebuild, not the validator.
+		table.WriteUpdateIsolationLevelKey: string(table.IsolationSnapshot),
+	}
+
+	localFS := func(_ context.Context) (iceio.IO, error) { return iceio.LocalFS{}, nil }
+
+	// Step 1: base table with file0 = rows [1,2,3].
+	baseTbl, cat := s.makeTable(0, props)
+	arrowSc, err := table.SchemaToArrowSchema(baseTbl.Schema(), nil, false, false)
+	s.Require().NoError(err)
+
+	file0Path := s.location + "/data/file0.parquet"
+	writeParquetFile(s.T(), file0Path, arrowSc,
+		`[{"id":1,"val":"a"},{"id":2,"val":"b"},{"id":3,"val":"c"}]`)
+
+	tx := baseTbl.NewTransaction()
+	s.Require().NoError(tx.AddFiles(s.ctx, []string{file0Path}, nil, false))
+	_, err = tx.Commit(s.ctx)
+	s.Require().NoError(err)
+	metaBase := cat.current
+
+	// Step 2: a concurrent peer appends filePeer = row [4] onto metaBase.
+	peerCat := &occScenarioCatalog{current: metaBase, location: s.location}
+	peerTbl := table.New(baseTbl.Identifier(), metaBase,
+		s.location+"/metadata/base.metadata.json", localFS, peerCat)
+
+	peerPath := s.location + "/data/file_peer.parquet"
+	writeParquetFile(s.T(), peerPath, arrowSc, `[{"id":4,"val":"d"}]`)
+
+	ptx := peerTbl.NewTransaction()
+	s.Require().NoError(ptx.AddFiles(s.ctx, []string{peerPath}, nil, false))
+	_, err = ptx.Commit(s.ctx)
+	s.Require().NoError(err)
+	metaPeer := peerCat.current
+
+	// Step 3: Writer A replaces file0 -> compacted, starting from the stale
+	// metaBase; its catalog is at metaPeer and returns one 409 before success.
+	catA := &occScenarioCatalog{current: metaPeer, conflictsLeft: 1, location: s.location}
+	writerA := table.New(baseTbl.Identifier(), metaBase,
+		s.location+"/metadata/base.metadata.json", localFS, catA)
+
+	tasks, err := writerA.Scan().PlanFiles(s.ctx)
+	s.Require().NoError(err)
+	s.Require().Len(tasks, 1, "Writer A's stale base must see exactly file0")
+	oldFile := tasks[0].File
+
+	compactedPath := s.location + "/data/compacted.parquet"
+	writeParquetFile(s.T(), compactedPath, arrowSc,
+		`[{"id":1,"val":"a"},{"id":2,"val":"b"},{"id":3,"val":"c"}]`)
+	newBuilder, err := iceberg.NewDataFileBuilder(
+		*iceberg.UnpartitionedSpec, iceberg.EntryContentData,
+		compactedPath, iceberg.ParquetFile, nil, nil, nil, 3, 512)
+	s.Require().NoError(err)
+	newFile := newBuilder.Build()
+
+	txA := writerA.NewTransaction()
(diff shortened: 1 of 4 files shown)


twuebi:
cc @laskoviymishka

table/snapshot_producers.go
```diff
@@ -812,125 +886,207 @@ func (sp *snapshotProducer) writeAddedManifest(content iceberg.ManifestContent,
...
+	return ok
+}
+
+// checkRemovedFiles scans parent once and (1) fails terminally if any data file
+// this producer intends to rewrite is no longer live — a concurrent writer
+// already removed it, so proceeding would leave our replacement beside the
+// concurrent result and duplicate rows (Java's failMissingDeletePaths) — and
+// (2) returns which removed delete files / deletion vectors are still present so
+// the retry summary does not double-subtract a peer's prior removal.
+func (sp *snapshotProducer) checkRemovedFiles(parent *Snapshot) (*removedFilePresence, error) {
```
laskoviymishka, line 1025: `rebuildFn` calls this unconditionally on every retry, including for `fastAppendFiles`/`mergeAppendFiles`, which never populate any of the removed-file maps. For those producers we now read `parent.Manifests` and iterate every entry on each attempt just to conclude there's nothing to check.

That's a scaling regression precisely where it hurts — under append contention the retry loop turns into repeated full-table manifest scans, where the old path only re-read the parent's manifest list. Quick fix, short-circuit at the top before touching `parent.Manifests`:

```go
if len(sp.deletedFiles) == 0 && len(sp.deletedDeleteFiles) == 0 && len(sp.deletedDVsByRef) == 0 {
    return present, nil
}
```

Longer term it'd be nice to share one parent scan across `checkRemovedFiles`, `deletedEntries`, and `existingManifests` — an overwrite retry currently walks the same entries up to three times — but the early-out is the part I'd want for this PR.

table/snapshot_producers.go
```diff
@@ -812,125 +886,207 @@ func (sp *snapshotProducer) writeAddedManifest(content iceberg.ManifestContent,
...
-		}
+	missing := make([]string, 0, len(missingData))
+	for path := range missingData {
+		missing = append(missing, path)
 	}
+	sort.Strings(missing)
 
-	return own, nil
+	return nil, fmt.Errorf("%w: %d data file(s) to rewrite are no longer on the branch head, e.g. %v",
+		ErrCommitDiverged, len(missing), missing[:min(len(missing), 5)])
```
laskoviymishka, line 1097: This abort — `ErrCommitDiverged` when a rewrite target is already gone from head — is the one behavior in this PR that changes what callers see. Previously they got silent resurrection; now they get a terminal, non-retryable error so they can re-plan. That's the right call and matches Java's `failMissingDeletePaths`, but nothing exercises it — `TestSummaryOnRetry_Skips/SubtractsPresentDeleteFile` cover the delete-file presence gating, not the `missingDataError` branch.

A unit test calling `checkRemovedFiles` with a parent missing one of `sp.deletedFiles` and asserting `errors.Is(err, ErrCommitDiverged)` would lock this down. An end-to-end scenario where the peer removes the exact file writer A is replacing, asserting the commit fails diverged rather than resurrecting, would be even better.

table/snapshot_producers.go
```diff
@@ -812,125 +886,207 @@ func (sp *snapshotProducer) writeAddedManifest(content iceberg.ManifestContent,
...
+			if err != nil {
+				return nil, err
+			}
+			df := entry.DataFile()
+			if m.ManifestContent() == iceberg.ManifestContentData {
+				delete(missingData, df.FilePath())
+
+				continue
+			}
+			if ref := df.ReferencedDataFile(); isDeletionVector(df) && ref != nil {
```
laskoviymishka: Two things on this branch. A deletion vector with a nil `ReferencedDataFile()` falls into the `else` and gets matched against `deletedDeleteFiles` by path — the wrong map for a DV. Low-probability (a v3 DV should always carry a ref) but it silently misclassifies instead of erroring; I'd split `isDeletionVector(df) && ref == nil` into its own case that skips or errors.

The other is a question on the matching itself: presence is keyed off the referenced data file, not the DV's own path. If a peer superseded the same data file's DV with a new one (same ref, different path), we'd see "some DV for this ref still exists" and report our stale DV as present — which `summaryOnRetry` would then count as a live removal for a DV that's no longer physically in the manifest. I couldn't fully trace whether an isolation validator rejects this first. If it doesn't, keying presence off the exact DV path (falling back to ref only for the summary decision) plus a concurrent-DV-replacement test would settle it. wdyt?

table/snapshot_producers.go
```diff
@@ -1115,17 +1271,13 @@ func (sp *snapshotProducer) commit(ctx context.Context) (_ []Update, _ []Require
...
+		// / DVs are gated by present (see checkRemovedFiles) so a peer's prior
+		// removal is not subtracted twice, which would undercount
+		// total-delete-files.
 		if freshParent != nil && freshParent.Summary != nil && capturedSnapshot.Summary != nil {
-			deltaSummary := Summary{
-				Operation:  capturedSnapshot.Summary.Operation,
-				Properties: maps.Clone(capturedSnapshot.Summary.Properties),
-			}
-			if s, sumErr := updateSnapshotSummaries(deltaSummary, freshParent.Summary.Properties); sumErr == nil {
+			if s, sumErr := sp.summaryOnRetry(sp.snapshotProps, freshParent, present); sumErr == nil {
```
laskoviymishka: If `summaryOnRetry` errors, we silently keep the stale attempt-0 summary — `sumErr` is discarded. The old `updateSnapshotSummaries` path swallowed its error too so this isn't a regression, but this PR reworks exactly this code and it's a natural spot to fix. I'd return the error from `rebuildFn` so the commit loop surfaces it rather than committing a summary we know is wrong.

table/snapshot_producers.go
```diff
@@ -812,125 +886,207 @@ func (sp *snapshotProducer) writeAddedManifest(content iceberg.ManifestContent,
...
-	own := make([]iceberg.ManifestFile, 0, len(allManifests))
-	for _, m := range allManifests {
-		if !inherited[m.FilePath()] {
-			own = append(own, m)
-		}
+	missing := make([]string, 0, len(missingData))
+	for path := range missingData {
+		missing = append(missing, path)
 	}
+	sort.Strings(missing)
```
laskoviymishka: We already import `slices` and target Go 1.25, so `slices.Sort(missing)` drops the `sort` import entirely. While we're here, `%v` on a `[]string` prints Go slice syntax (`[path1 path2]`) in a user-facing error — `strings.Join` or `%q` reads cleaner.

table/snapshot_producers.go
```diff
@@ -812,125 +886,207 @@ func (sp *snapshotProducer) writeAddedManifest(content iceberg.ManifestContent,
...
+	return sp.rebaseSummary(delta, previousSummary, props)
+}
+
+// summaryOnRetry recomputes the snapshot summary against the fresh parent for an
+// OCC retry. Additions are always counted; a removal of a delete file / deletion
+// vector is counted only when it is still present on parent (present.counts),
+// because a peer may have already dropped it — replaying the captured removal
+// delta on top of a fresh parent that no longer counts the file undercounts
+// total-delete-files. Removed data files are always counted: a successful retry
+// has already asserted they are present (assertRemovedFilesPresent).
```
laskoviymishka: `assertRemovedFilesPresent` doesn't exist — the function that does this is `checkRemovedFiles`. Worth fixing so the comment doesn't send someone hunting for a symbol that isn't there.

table/occ_scenario_test.go
```diff
@@ -451,6 +451,129 @@ func (s *OCCScenarioTestSuite) TestV3MergeAppendRowLineageAfterOCCRetry() {
...
+
+	ptx := peerTbl.NewTransaction()
+	s.Require().NoError(ptx.AddFiles(s.ctx, []string{peerPath}, nil, false))
+	_, err = ptx.Commit(s.ctx)
+	s.Require().NoError(err)
+	metaPeer := peerCat.current
+
+	// Step 3: Writer A replaces file0 -> compacted, starting from the stale
+	// metaBase; its catalog is at metaPeer and returns one 409 before success.
+	catA := &occScenarioCatalog{current: metaPeer, conflictsLeft: 1, location: s.location}
```
laskoviymishka, line 522: Every OCC scenario here uses `conflictsLeft: 1`, so we only ever exercise a single rebuild. The reused-added-content path is most interesting across more than one retry — that's where orphaned-manifest churn and cumulative removal tracking would show up if the reuse logic were wrong.

A `conflictsLeft: 2` variant (two distinct peer commits) asserting the same no-resurrection invariant would lock in retry idempotency, which is the property the added-content reuse is really claiming. Nice-to-have, not blocking.

laskoviymishka (changes requested):
The core fix here is right, and a good one. Rebuilding the parent-dependent manifests against the fresh head on every retry — rather than grafting attempt-0's "own" manifests onto a fresh parent — is exactly what stops the resurrection, and splitting parent-independent added content (written once, reused) from the parent-dependent tombstones/inherited manifests (recomputed each attempt) is a clean way to express it. `TestReplaceFilesNoResurrectionAfterConflict` is a real regression test, not a tautology: it fails on the old code with file0 back from the dead and passes here.

A nice bonus: the added/reused split also sidesteps a latent `computeOwnManifests` misclassification for merge-append (a merged/inherited manifest could be mistaken for "own"), beyond the titled resurrection. Intentional secondary fix or happy side effect? Either way it's an improvement.

I think this is good to land after a couple of quick fixes — nothing structural:

- Short-circuit `checkRemovedFiles` when there's nothing to remove, so plain appends don't pay a full parent scan on every retry.
- Add a test for the `ErrCommitDiverged` abort branch — it's the one caller-visible semantic change here and it's currently untested.

The rest are small: the `assertRemovedFilesPresent` doc typo, `slices.Sort` over the extra `sort` import, the DV nil-ref branch, and not swallowing the `summaryOnRetry` error. A `conflictsLeft: 2` scenario would be a nice-to-have to cover more than one rebuild. Details inline. Happy to approve on the next pass once the short-circuit and the abort-path test are in.

zeroshade (approved):
LGTM from my side — the retry-replay fix and the checkRemovedFiles fail-closed behavior look correct and the regression coverage is solid. Holding the merge until @laskoviymishka's requested changes are resolved.

laskoviymishka (commented):
💃 👍

laskoviymishka (approved):

PR #1429, opened 2026-07-09
fix(ci): give Trino healthcheck startup grace time

### LINKED ISSUE #1420: CI flake: Spark 3.5 integration setup fails because Trino is marked unhealthy before startup completes
### Apache Iceberg version

None

### Please describe the bug 🐞

Hi! I noticed a flaky failure in the Go integration workflow where the Spark 3.5 setup job sometimes fails before the actual Go integration tests start.

The failing job is:

integration-test (3.5, integration-setup)

Example failed runs:

https://github.com/apache/iceberg-go/actions/runs/28865791807

https://github.com/apache/iceberg-go/actions/runs/28865731963

The failure happens during the integration setup step, while running:

make integration-setup

That target runs Docker Compose with:

docker compose -f internal/recipe/docker-compose.yml up -d --wait

The setup fails with:

container trino is unhealthy

From the logs, this looks timing-related. In one failed run, Docker Compose marked Trino as unhealthy, but Trino finished starting almost immediately afterward and logged:

Server startup completed in 30.09s

======== SERVER STARTED ========

So it seems like Trino is still coming up, but Docker Compose gives up just before it becomes ready.

This does not look related to the Go code change that triggered the run, since the failure happens during Docker setup before the integration tests are executed. The Spark 4.0 integration job also passed in the same workflow run, and nearby CI runs have passed, which makes this look like a flaky setup/readiness issue rather than a deterministic test failure.

**Expected behavior**

The integration setup should wait long enough for Trino to become ready, or use a readiness check that avoids failing when Trino is just slow to start.

**Possible fixes**

A few possible ways to make this more stable:

Increase Trino’s healthcheck/startup budget.

Add an explicit healthcheck/start period for the Trino service.

Pin the trinodb/trino image to a known stable version instead of using the floating latest image.

Add a small retry around the Docker Compose setup step if this is expected to occasionally race.

**Additional context**

The Trino service in internal/recipe/docker-compose.yml currently uses trinodb/trino, and the failure appears to be caused by service readiness timing rather than a failing Go integration test.

fallintoplace:
## Summary
- give the Trino service a startup grace period in `internal/recipe/docker-compose.yml`
- keep `docker compose up -d --wait` from failing while Trino is still booting in CI

## Why
The Spark 3.5 integration setup has been flaking before any Go tests run because Trino can take around 30 seconds to become ready in GitHub Actions. Without a startup grace period, the container's healthcheck can flip to `unhealthy` just before startup completes, which makes `make integration-setup` fail even though Trino comes up moments later.

This keeps the existing readiness flow in place and just gives the healthcheck enough warm-up time to match observed CI startup behavior.

Closes #1420.

## Testing
- `docker compose -f internal/recipe/docker-compose.yml config`

### DIFF
--- a/internal/recipe/docker-compose.yml
+++ b/internal/recipe/docker-compose.yml
@@ -117,6 +117,11 @@ services:
       - CATALOG_MANAGEMENT=dynamic
     ports:
       - 8088:8080
+    healthcheck:
+      # Trino can take ~30s to finish booting in GitHub Actions; without a
+      # startup grace period, docker compose up --wait can mark it unhealthy
+      # just before the server reaches READY.
+      start_period: 45s
 
   fake-gcs-server:
     image: fsouza/fake-gcs-server


laskoviymishka (approved):
thx for that!

PR #1430, opened 2026-07-09
feat(metrics): wire reporter selection through catalog, table, and scan (#1236)

tanmayrauth:
Third PR of the metrics reporting framework. Adds reporter selection and threads it from catalog to scan:

- metrics: a name->factory registry and FromProperties, selecting a reporter via the metrics-reporter-impl property (the Go analogue of Java's reflection-based loader); built-in nop and logging reporters.
- table: a WithMetricsReporter option on New/NewFromLocation (variadic, backward compatible), a MetricsReporter accessor, and a WithReporter scan option; scans inherit the table's reporter.
- catalog/{rest,sql,glue,hive,hadoop} resolve the reporter from properties when loading or creating a table.

Reporting stays opt-in: the default reporter is no-op, so there is no behavior change for existing users. Report emission is added with the scan and commit instrumentation in later PRs.

Related: #1236

### DIFF
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -284,18 +284,15 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 	}
 	// Load the metadata file to get table properties
 	ctx = utils.WithAwsConfig(ctx, c.awsCfg)
-	reporter, err := metrics.FromProperties(c.props)
-	if err != nil {
-		return nil, err
-	}
-	// Read the metadata file
+	// Read the metadata file. The reporter is intentionally not wired here: the
+	// tail c.LoadTable(...) re-resolves it through convertGlueToIceberg, and the
+	// table built below is discarded after constructTableInput.
 	tbl, err := table.NewFromLocation(
 		ctx,
 		[]string{tableName},
 		metadataLocation,
 		io.LoadFSFunc(nil, metadataLocation),
 		c,
-		table.WithMetricsReporter(reporter),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read table metadata from %s: %w", metadataLocation, err)
@@ -841,7 +838,7 @@ func (c *Catalog) convertGlueToIceberg(ctx context.Context, glueTable *types.Tab
 
 	reporter, err := metrics.FromProperties(c.props)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
 
 	icebergTable, err := table.NewFromLocation(
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -150,6 +150,9 @@ type Catalog struct {
 	catalogId *string
 	awsCfg    *aws.Config
 	props     iceberg.Properties
+	// reporter builds and caches the catalog's metrics reporter once, so it is
+	// constructed per-catalog rather than per table load. Released by Close.
+	reporter metrics.CachedReporter
 }
 
 // NewCatalog creates a new instance of glue.Catalog with the given options.
@@ -245,6 +248,12 @@ func (c *Catalog) LoadTable(ctx context.Context, identifier table.Identifier) (*
 func (c *Catalog) CatalogType() catalog.Type {
 	return catalog.Glue
 }
+
+// Close releases the catalog's metrics reporter. The Glue catalog does not own
+// the lifetime of the AWS clients it was configured with, so only the reporter
+// is released. Callers holding a [catalog.Catalog] can reach this via an
+// io.Closer type assertion.
+func (c *Catalog) Close() error { return c.reporter.Close() }
 
 // CreateTable creates a new Iceberg table in the Glue catalog.
 // This function will create the metadata file in S3 using the catalog and table properties,
@@ -836,7 +845,7 @@ func (c *Catalog) convertGlueToIceberg(ctx context.Context, glueTable *types.Tab
 		return nil, fmt.Errorf("missing metadata location for table %s", tableName)
 	}
 
-	reporter, err := metrics.FromProperties(c.props)
+	reporter, err := c.reporter.Get(c.props)
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -33,6 +33,7 @@ import (
 	"github.com/apache/iceberg-go/catalog"
 	"github.com/apache/iceberg-go/catalog/internal"
 	"github.com/apache/iceberg-go/io"
+	"github.com/apache/iceberg-go/metrics"
 	"github.com/apache/iceberg-go/table"
 	"github.com/apache/iceberg-go/utils"
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -283,6 +284,10 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 	}
 	// Load the metadata file to get table properties
 	ctx = utils.WithAwsConfig(ctx, c.awsCfg)
+	reporter, err := metrics.FromProperties(c.props)
+	if err != nil {
+		return nil, err
+	}
 	// Read the metadata file
 	tbl, err := table.NewFromLocation(
 		ctx,
@@ -290,6 +295,7 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 		metadataLocation,
 		io.LoadFSFunc(nil, metadataLocation),
 		c,
+		table.WithMetricsReporter(reporter),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read table metadata from %s: %w", metadataLocation, err)
@@ -833,12 +839,18 @@ func (c *Catalog) convertGlueToIceberg(ctx context.Context, glueTable *types.Tab
 		return nil, fmt.Errorf("missing metadata location for table %s", tableName)
 	}
 
+	reporter, err := metrics.FromProperties(c.props)
+	if err != nil {
+		return nil, err
+	}
+
 	icebergTable, err := table.NewFromLocation(
 		utils.WithAwsConfig(ctx, c.awsCfg),
 		TableIdentifier(database, tableName),
 		metadataLocation,
 		io.LoadFSFunc(nil, metadataLocation),
 		c,
+		table.WithMetricsReporter(reporter),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create iceberg table from location %s: %w", metadataLocation, err)
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -93,7 +93,7 @@ func init() {
 			return nil, err
 		}
 
-		return NewCatalog(WithAwsConfig(awsConfig), WithAwsProperties(AwsProperties(props))), nil
+		return NewCatalog(WithAwsConfig(awsConfig), WithAwsProperties(AwsProperties(props)))
 	}))
 }
 
@@ -165,9 +165,9 @@ type Catalog struct {
 //     ...other transport options...
 // })
 // awsCfg.HTTPClient = client
-// catalog := glue.NewCatalog(glue.WithAwsConfig(awsCfg))
+// catalog, err := glue.NewCatalog(glue.WithAwsConfig(awsCfg))
 
-func NewCatalog(opts ...Option) *Catalog {
+func NewCatalog(opts ...Option) (*Catalog, error) {
 	glueOps := &options{}
 
 	for _, o := range opts {
@@ -185,12 +185,23 @@ func NewCatalog(opts ...Option) *Catalog {
 		catalogId = nil
 	}
 
-	return &Catalog{
+	cat := &Catalog{
 		glueSvc:   glue.NewFromConfig(glueOps.awsConfig),
 		catalogId: catalogId,
 		awsCfg:    &glueOps.awsConfig,
 		props:     iceberg.Properties(glueOps.awsProperties),
 	}
+
+	// Resolve the metrics reporter once, up front. CachedReporter caches the
+	// first result, so building it here means a bad metrics-reporter-impl fails
+	// construction — rather than surfacing later as a spurious CreateTable error
+	// (after the Glue entry is written, via the trailing convertGlueToIceberg) or
+	// as a cached error that wedges every later table op on this catalog.
+	if _, err := cat.reporter.Get(cat.props); err != nil {
+		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
+	}
+
+	return cat, nil
 }
 
 // ListTables returns a list of Iceberg tables in the given Glue database.
@@ -251,23 +262,20 @@ func (c *Catalog) CatalogType() catalog.Type {
 
 // Close releases the catalog's metrics reporter. The Glue catalog does not own
 // the lifetime of the AWS clients it was configured with, so only the reporter
-// is released. Callers holding a [catalog.Catalog] can reach this via an
-// io.Closer type assertion.
+// is released. Callers holding a [catalog.Catalog] can reach this via a
+// [catalog.Closer] type assertion.
 func (c *Catalog) Close() error { return c.reporter.Close() }
+
+var _ catalog.Closer = (*Catalog)(nil)
 
 // CreateTable creates a new Iceberg table in the Glue catalog.
 // This function will create the metadata file in S3 using the catalog and table properties,
 // to determine the bucket and key for the metadata location.
 func (c *Catalog) CreateTable(ctx context.Context, identifier table.Identifier, schema *iceberg.Schema, opts ...catalog.CreateTableOpt) (*table.Table, error) {
-	// Resolve the reporter before mutating the catalog: a bad
-	// metrics-reporter-impl must fail before the Glue entry is created, not
-	// after, which would report a failure for a table that was actually created
-	// (and make the retry hit ErrTableAlreadyExists). CachedReporter caches this
-	// first result, so the trailing LoadTable reuses it.
-	if _, err := c.reporter.Get(c.props); err != nil {
-		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
-	}
-
+	// The reporter is resolved once at construction (see NewCatalog), so a bad
+	// metrics-reporter-impl already failed there — no per-op guard is needed
+	// before mutating the catalog, and the trailing LoadTable reuses the cached
+	// reporter.
 	staged, err := internal.CreateStagedTable(ctx, c.props, c.LoadNamespaceProperties, identifier, schema, opts...)
 	if err != nil {
 		return nil, err
@@ -301,13 +309,6 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 		return nil, err
 	}
 
-	// Resolve the reporter before mutating the catalog, for the same reason as
-	// CreateTable: an invalid reporter must not turn a successful registration
-	// into a reported failure. CachedReporter caches this for the trailing
-	// LoadTable.
-	if _, err := c.reporter.Get(c.props); err != nil {
-		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
-	}
 	// Load the metadata file to get table properties
 	ctx = utils.WithAwsConfig(ctx, c.awsCfg)
 	// Read the metadata file. The reporter is intentionally not wired here: the
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -259,6 +259,15 @@ func (c *Catalog) Close() error { return c.reporter.Close() }
 // This function will create the metadata file in S3 using the catalog and table properties,
 // to determine the bucket and key for the metadata location.
 func (c *Catalog) CreateTable(ctx context.Context, identifier table.Identifier, schema *iceberg.Schema, opts ...catalog.CreateTableOpt) (*table.Table, error) {
+	// Resolve the reporter before mutating the catalog: a bad
+	// metrics-reporter-impl must fail before the Glue entry is created, not
+	// after, which would report a failure for a table that was actually created
+	// (and make the retry hit ErrTableAlreadyExists). CachedReporter caches this
+	// first result, so the trailing LoadTable reuses it.
+	if _, err := c.reporter.Get(c.props); err != nil {
+		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
+	}
+
 	staged, err := internal.CreateStagedTable(ctx, c.props, c.LoadNamespaceProperties, identifier, schema, opts...)
 	if err != nil {
 		return nil, err
@@ -290,6 +299,14 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 	database, tableName, err := identifierToGlueTable(identifier)
 	if err != nil {
 		return nil, err
+	}
+
+	// Resolve the reporter before mutating the catalog, for the same reason as
+	// CreateTable: an invalid reporter must not turn a successful registration
+	// into a reported failure. CachedReporter caches this for the trailing
+	// LoadTable.
+	if _, err := c.reporter.Get(c.props); err != nil {
+		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
 	// Load the metadata file to get table properties
 	ctx = utils.WithAwsConfig(ctx, c.awsCfg)
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -552,7 +552,7 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 
 	reporter, err := metrics.FromProperties(c.props)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
 
 	tbl := table.New(
@@ -579,7 +579,7 @@ func (c *Catalog) loadTable(ctx context.Context, ident table.Identifier) (*table
 
 	reporter, err := metrics.FromProperties(c.props)
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
 
 	tbl, err := table.NewFromLocation(ctx, ident, metaPath, io.LoadFSFunc(c.props, metaPath), c, table.WithMetricsReporter(reporter))
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -244,14 +244,25 @@ func NewCatalog(name, warehouse string, props iceberg.Properties) (*Catalog, err
 		return nil, fmt.Errorf("hadoop catalog: %T does not implement HadoopCatalogFS", filesystem)
 	}
 
-	return &Catalog{
+	cat := &Catalog{
 		name:      name,
 		warehouse: warehouse,
 		isLocal:   isLocal,
 		// filesystem is resolved dynamically from the IO registry based on the warehouse scheme
 		filesystem: hadoopFs,
 		props:      props,
-	}, nil
+	}
+
+	// Resolve the metrics reporter once, up front. CachedReporter caches the
+	// first result, so building it here means a bad metrics-reporter-impl fails
+	// construction — rather than surfacing later as a spurious CreateTable error
+	// after the table is already on disk, or as a cached error that wedges every
+	// later table op on this catalog.
+	if _, err := cat.reporter.Get(props); err != nil {
+		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
+	}
+
+	return cat, nil
 }
 
 func (c *Catalog) CatalogType() catalog.Type {
@@ -259,8 +270,10 @@ func (c *Catalog) CatalogType() catalog.Type {
 }
 
 // Close releases the catalog's metrics reporter. Callers holding a
-// [catalog.Catalog] can reach this via an io.Closer type assertion.
+// [catalog.Catalog] can reach this via a [catalog.Closer] type assertion.
 func (c *Catalog) Close() error { return c.reporter.Close() }
+
+var _ catalog.Closer = (*Catalog)(nil)
 
 // joinPath is a helper that allows paths to be joined as both local filesystem
 // paths or as remote URIs needed for filesystems like blob stores.
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -36,6 +36,7 @@ import (
 	"github.com/apache/iceberg-go/catalog"
 	"github.com/apache/iceberg-go/catalog/internal"
 	"github.com/apache/iceberg-go/io"
+	"github.com/apache/iceberg-go/metrics"
 	"github.com/apache/iceberg-go/table"
 	"github.com/google/uuid"
 )
@@ -549,12 +550,18 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 
 	c.writeVersionHint(ident, version)
 
+	reporter, err := metrics.FromProperties(c.props)
+	if err != nil {
+		return nil, err
+	}
+
 	tbl := table.New(
 		ident,
 		metadata,
 		metaPath,
 		io.LoadFSFunc(c.props, metaPath),
 		c,
+		table.WithMetricsReporter(reporter),
 	)
 
 	return tbl, nil
@@ -570,7 +577,12 @@ func (c *Catalog) loadTable(ctx context.Context, ident table.Identifier) (*table
 		return nil, 0, err
 	}
 
-	tbl, err := table.NewFromLocation(ctx, ident, metaPath, io.LoadFSFunc(c.props, metaPath), c)
+	reporter, err := metrics.FromProperties(c.props)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	tbl, err := table.NewFromLocation(ctx, ident, metaPath, io.LoadFSFunc(c.props, metaPath), c, table.WithMetricsReporter(reporter))
 	if err != nil {
 		return nil, 0, err
 	}
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -495,6 +495,15 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 		return nil, err
 	}
 
+	// Resolve the reporter before any catalog mutation: a bad
+	// metrics-reporter-impl must fail here, not after the metadata file and
+	// version hint are written, which would report a failure for a table that
+	// was actually created (and make the retry hit ErrTableAlreadyExists).
+	reporter, err := c.reporter.Get(c.props)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
+	}
+
 	ns := catalog.NamespaceFromIdent(ident)
 	nsPath := c.namespaceToPath(ns)
 
@@ -556,11 +565,6 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 	}
 
 	c.writeVersionHint(ident, version)
-
-	reporter, err := c.reporter.Get(c.props)
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
-	}
 
 	tbl := table.New(
 		ident,
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -184,6 +184,9 @@ type Catalog struct {
 	isLocal    bool
 	filesystem HadoopCatalogFS
 	props      iceberg.Properties
+	// reporter builds and caches the catalog's metrics reporter once, so it is
+	// constructed per-catalog rather than per table load. Released by Close.
+	reporter metrics.CachedReporter
 }
 
 // NewCatalog creates a new Hadoop catalog rooted at the given warehouse path.
@@ -254,6 +257,10 @@ func NewCatalog(name, warehouse string, props iceberg.Properties) (*Catalog, err
 func (c *Catalog) CatalogType() catalog.Type {
 	return catalog.Hadoop
 }
+
+// Close releases the catalog's metrics reporter. Callers holding a
+// [catalog.Catalog] can reach this via an io.Closer type assertion.
+func (c *Catalog) Close() error { return c.reporter.Close() }
 
 // joinPath is a helper that allows paths to be joined as both local filesystem
 // paths or as remote URIs needed for filesystems like blob stores.
@@ -550,7 +557,7 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 
 	c.writeVersionHint(ident, version)
 
-	reporter, err := metrics.FromProperties(c.props)
+	reporter, err := c.reporter.Get(c.props)
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
@@ -577,7 +584,7 @@ func (c *Catalog) loadTable(ctx context.Context, ident table.Identifier) (*table
 		return nil, 0, err
 	}
 
-	reporter, err := metrics.FromProperties(c.props)
+	reporter, err := c.reporter.Get(c.props)
 	if err != nil {
 		return nil, 0, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -247,7 +247,7 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 		table.WithMetricsReporter(reporter),
 	)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read table metadata from %s: %s", metadataLocation, err)
+		return nil, fmt.Errorf("failed to read table metadata from %s: %w", metadataLocation, err)
 	}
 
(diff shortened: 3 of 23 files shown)


table/table.go
```diff
@@ -157,6 +168,7 @@ func (t *Table) Refresh(ctx context.Context) error {
 	t.fsF = fresh.fsF
 	t.metadataLocation = fresh.metadataLocation
 	t.planner = fresh.planner
+	t.reporter = fresh.reporter
```
laskoviymishka: Refresh unconditionally copies `fresh.reporter`, so any reporter a caller injected via `WithMetricsReporter` gets replaced on the next refresh. Since commit retry loops call Refresh, a user-configured reporter silently reverts to the catalog-derived one (nop, in most setups) after the first cycle.

I'd only inherit the fresh reporter when the caller hasn't set one:

```go
if _, isNop := t.reporter.(metrics.NopReporter); isNop {
    t.reporter = fresh.reporter
}
```

Either that or document the reset explicitly — but either way this needs a test. Right now nothing covers reporter survival across Refresh, and it's the invariant most likely to regress. wdyt?

catalog/rest/rest.go
```diff
@@ -834,12 +835,18 @@ func (r *Catalog) tableFromResponse(_ context.Context, identifier []string, meta
 		fsF = iceio.LoadFSFunc(config, loc)
 	}
 
+	reporter, err := metrics.FromProperties(config)
```
laskoviymishka: Every other catalog resolves the reporter from its own catalog props, but here we're passing the fully-merged `config` — `r.props` plus the table metadata properties plus the server-vended `ret.Config` plus creds. So a `metrics-reporter-impl` coming back from the server, or set as a table property, can shadow what the client configured, and last-layer-wins is hard to reason about. Worse, an unknown name in any of those layers turns LoadTable into a hard client-side error.

I'd resolve from `r.props` here to match the other catalogs:

```go
reporter, err := metrics.FromProperties(r.props)
```

If server-vended reporter selection is actually intended, that's a real design decision worth calling out explicitly rather than falling out of the merge order. wdyt?

metrics/registry_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	t.Run("unknown name is an error", func(t *testing.T) {
+		_, err := FromProperties(map[string]string{ReporterImplKey: "does-not-exist"})
+		assert.Error(t, err)
+	})
+}
+
+func TestRegister(t *testing.T) {
+	t.Run("factory is selectable via FromProperties", func(t *testing.T) {
+		Register("test-custom", func(props map[string]string) (Reporter, error) {
```
laskoviymishka, line 60: This mutates the process-global registry and never cleans up, so `go test -count=2 ./metrics/` panics on the duplicate Register — and the registry can't be exercised in isolation.

I'd make the registry testable: either expose a `newRegistry()` that tests use for a fresh instance, or register with `t.Cleanup(func(){ Deregister(...) })`. The struct-with-methods shape is nicer long-term since it also removes the global-mutable-state smell, but a cleanup hook is the minimal fix. Thoughts?

catalog/glue/glue.go
```diff
@@ -279,13 +280,18 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 	}
 	// Load the metadata file to get table properties
 	ctx = utils.WithAwsConfig(ctx, c.awsCfg)
+	reporter, err := metrics.FromProperties(c.props)
```
laskoviymishka: This reporter block is dead — the temporary table built here via `NewFromLocation` is discarded, and RegisterTable returns `c.LoadTable(...)` at the tail, which re-resolves the reporter through `convertGlueToIceberg`. So we resolve twice and only the second one survives.

I'd drop the `FromProperties` + `WithMetricsReporter` here and let the tail LoadTable own the wiring, matching how sql.CreateTable delegates. As written it's harmless but confusing, and a future `NewFromLocation` path that skips LoadTable would silently lose the reporter.

catalog/hive/hive.go
```diff
@@ -149,12 +150,18 @@ func (c *Catalog) LoadTable(ctx context.Context, identifier table.Identifier) (*
 		return nil, fmt.Errorf("failed to get metadata location: %w", err)
 	}
 
+	reporter, err := metrics.FromProperties(c.opts.props)
```
laskoviymishka: We call `FromProperties` on every load/register across all five catalogs, so each table op builds a fresh reporter instance. Java keeps a single lazily-created reporter per catalog. For the built-in nop/logging reporters this is fine since they're stateless, but a stateful reporter — one holding a connection, buffer, or pool, or meant to aggregate across calls — gets multiplied per table op and siloed per table.

Not blocking for this PR since the built-ins are stateless and there's no Closer path yet, but I'd want us to decide whether reporter lifetime is per-catalog or per-table before third-party stateful reporters show up. A per-catalog cached reporter would match Java. wdyt?

catalog/sql/sql.go
```diff
@@ -434,12 +435,18 @@ func (c *Catalog) LoadTable(ctx context.Context, identifier table.Identifier) (*
 		return nil, fmt.Errorf("%w: %s, metadata location is missing", catalog.ErrNoSuchTable, identifier)
 	}
 
+	reporter, err := metrics.FromProperties(c.props)
```
laskoviymishka: Bare `return nil, err` here drops the operation context — everywhere else in these catalog files we wrap with `fmt.Errorf("failed to ...: %w")`. A user who typos the reporter name gets `metrics: unknown reporter "foo"` with no hint which catalog or table triggered it.

I'd wrap it: `fmt.Errorf("failed to initialize metrics reporter: %w", err)`. Same pattern repeats in the glue/hadoop/hive/rest sites.

catalog/sql/sql_test.go
```diff
@@ -425,6 +426,77 @@ func (s *SqliteCatalogTestSuite) TestCreateTableDefaultSortOrder() {
...
+		cat := newCatMemory(nil)
+		created, _ := createTable(cat)
+		s.IsType(metrics.NopReporter{}, created.MetricsReporter())
+	})
+
+	s.Run("unknown reporter name fails table load", func() {
+		// Create with a well-configured catalog, then reopen the same
+		// on-disk DB with a bad reporter name so the failure is isolated to
+		// LoadTable rather than the create path.
+		good := s.getCatalogSqlite()
```
laskoviymishka: This leans on `s.getCatalogSqlite()`/`s.catalogUri()` sharing on-disk state with the sibling suite setup, which makes the test order-dependent and fragile. I'd create and reopen the DB within the test itself so the unknown-reporter failure is fully self-contained.

metrics/registry.go
```diff
@@ -0,0 +1,89 @@
...
+
+// ReporterImplKey is the catalog/table property that selects a registered
+// reporter by name (e.g. "logging"). It is the Go analogue of Java's
+// metrics-reporter-impl; Go uses a name→factory registry rather than
+// reflection over a class name.
+const ReporterImplKey = "metrics-reporter-impl"
+
+// Built-in reporter names usable as the value of [ReporterImplKey].
+const (
+	reporterNameNop     = "nop"
```
laskoviymishka: Users have to pass `"nop"`/`"logging"` as bare string literals in their property maps, with no compile-time signal if they fumble one. Since `ReporterImplKey` is already exported, I'd export these too — `ReporterNameNop` / `ReporterNameLogging` — so callers get a symbol to reference. wdyt?

laskoviymishka (changes requested):
Nice piece of wiring — the option plumbing through `New`/`NewFromLocation`, the scan inheritance, and the SQL wiring test are all clean, and the name→factory registry is a sensible Go take on Java's `metrics-reporter-impl`.

I'd hold this before merging though. `Refresh` unconditionally copies `fresh.reporter`, so a reporter a caller injected via `WithMetricsReporter` gets silently reverted to the catalog default on the next refresh — and since commit retry loops call `Refresh`, that reverts mid-operation. It's also untested, which is the one invariant here most likely to regress. I'd only inherit `fresh.reporter` when the caller hasn't set one, and add a test that a custom reporter survives a refresh.

The REST catalog also resolves the reporter from the fully-merged config rather than `r.props`, so a server-vended or table-level `metrics-reporter-impl` can shadow the client's choice and last-layer-wins is hard to reason about. And the global registry gets mutated in tests with no cleanup, which panics under `go test -count=2` and makes it untestable in isolation.

Things I'd want settled in this PR before merge:
- `Refresh` no longer drops a caller-set reporter, with a test covering it
- REST resolves from `r.props` (or we explicitly decide server-vended selection is intended)
- the registry is testable in isolation — fresh instance or cleanup hook
- the dead double-resolution in Glue `RegisterTable` is removed

One thing that's not blocking but worth deciding now: reporter lifetime is per-table-op here where Java is per-catalog. Fine for the stateless built-ins, but it'll matter for the first stateful reporter, so I'd rather pin the contract while the surface is small.

Once those are addressed I'm happy to take another pass and approve.

table/table.go
tanmayrauth: Fixed it, refresh now only inherits fresh.reporter when the current one is still the nop default, so a WithMetricsReporter-injected reporter survives the commit retry loop.

catalog/rest/rest.go
tanmayrauth: Agreed, server-vended selection wasn't intended. Changed to resolve from r.props so it matches the other catalogs and the client stays in control; if we ever want servers to drive reporter choice we'll make that an explicit,  documented decision.

metrics/registry_test.go
tanmayrauth: Added a Deregister(name) and TestRegister now does t.Cleanup(func(){ Deregister("test-custom") }). Confirmed go test -count=2 ./metrics/ passes now. Kept the global registry for this PR rather than reshaping into a  struct-with-methods.

catalog/glue/glue.go
tanmayrauth: I will clean it up.

catalog/hive/hive.go
tanmayrauth: Agreed this should be pinned. For this PR I've kept it per-op since the built-ins are stateless and there's no Closer path yet — but the intended shape is a per-catalog cached reporter matching Java, once a stateful reporter  (connection/buffer/pool) shows up. Happy to do the per-catalog caching in this PR as well if needed.

catalog/sql/sql.go
tanmayrauth: Wrapped as fmt.Errorf("failed to initialize metrics reporter: %w", err), and applied the same at the glue/hadoop/hive/rest sites so a typo'd reporter name surfaces with context.

catalog/sql/sql_test.go
tanmayrauth: the subtest now provisions its own temp DB and builds its props from a local base map instead of leaning on s.getCatalogSqlite() / s.catalogUri().

metrics/registry.go
tanmayrauth: Done, exported ReporterNameNop / ReporterNameLogging and used them in the init() registrations, so callers have symbols instead of bare "nop"/"logging" literals.

table/table.go
```diff
@@ -112,11 +114,20 @@ func (t Table) Schema() *iceberg.Schema                      { return t.metadata
...
 func (t Table) SortOrder() SortOrder                         { return t.metadata.SortOrder() }
 func (t Table) Properties() iceberg.Properties               { return t.metadata.Properties() }
-func (t Table) NameMapping() iceberg.NameMapping             { return t.metadata.NameMapping() }
-func (t Table) Location() string                             { return t.metadata.Location() }
-func (t Table) CurrentSnapshot() *Snapshot                   { return t.metadata.CurrentSnapshot() }
-func (t Table) SnapshotByID(id int64) *Snapshot              { return t.metadata.SnapshotByID(id) }
-func (t Table) SnapshotByName(name string) *Snapshot         { return t.metadata.SnapshotByName(name) }
+
+// MetricsReporter returns the table's metrics reporter, never nil.
+func (t Table) MetricsReporter() metrics.Reporter {
```
zeroshade, line 124: The reporter added here isn't propagated on commit. `doCommit` (around `table/table.go:576`) returns `New(t.identifier, newMeta, newLoc, t.fsF, t.cat)` without `WithMetricsReporter(t.MetricsReporter())`, so tables returned from `Transaction.Commit`/`AppendTable` silently fall back to `NopReporter`, and subsequent scans from them stop reporting. `Transaction.Scan` (around `table/transaction.go:2106`) similarly doesn't seed `reporter` the way `Table.Scan` does. Please carry `MetricsReporter()` through both paths.

zeroshade (changes requested):
Nice work wiring configurable reporters through the catalog. A couple of table lifecycle paths still drop the reporter, which undercuts the feature (inline).

table/table.go
tanmayrauth: Good catch, both paths were dropping the reporter. Fixed it:
  
  - doCommit (table/table.go:576) now returns New(..., WithMetricsReporter(t.MetricsReporter())), so tables from Transaction.Commit/AppendTable carry the configured reporter forward and scans off them keep reporting instead of
  falling back to NopReporter.
  - Transaction.Scan (table/transaction.go) now seeds reporter: t.tbl.MetricsReporter(), matching Table.Scan.

  MetricsReporter() is nil-safe (returns NopReporter{} when unset), so both paths stay safe when no reporter was configured.

table/table.go
```diff
@@ -157,6 +168,13 @@ func (t *Table) Refresh(ctx context.Context) error {
 	t.fsF = fresh.fsF
 	t.metadataLocation = fresh.metadataLocation
 	t.planner = fresh.planner
+	// Only inherit the catalog-derived reporter when the caller hasn't set one
+	// of their own. Refresh runs inside commit retry loops, so unconditionally
+	// copying fresh.reporter would silently revert a WithMetricsReporter-injected
+	// reporter to the catalog default mid-operation.
+	if _, isNop := t.reporter.(metrics.NopReporter); isNop {
```
laskoviymishka: This is the one thing I'd tighten before approving. The guard uses the concrete `NopReporter` type as a stand-in for "the caller never set a reporter," and that conflates two different intents.

A caller who explicitly opts out with `WithMetricsReporter(metrics.NopReporter{})` — or passes `metrics.Combine(nil, nil)`, which resolves to `NopReporter{}` — reads as "unset" here and gets silently reverted to the catalog default on the next `Refresh`. Since this is the exact invariant last round was about, I'd rather make it explicit with a was-set flag:

```go
// in WithMetricsReporter, when r != nil:
t.reporter = r
t.reporterSet = true

// in Refresh:
if !t.reporterSet {
    t.reporter = fresh.reporter
}
```

That also drops the type dependency entirely. And I'd add the companion test — `WithMetricsReporter(metrics.NopReporter{})`, refresh against a catalog that hands back a logging reporter, assert the reporter is still `NopReporter{}`. Right now that case would silently fail. wdyt?

catalog/hive/hive.go
```diff
@@ -226,12 +233,18 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
...
 	tbl, err := table.NewFromLocation(
 		ctx,
 		identifier,
 		metadataLocation,
 		io.LoadFSFunc(c.opts.props, metadataLocation),
 		c,
+		table.WithMetricsReporter(reporter),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read table metadata from %s: %s", metadataLocation, err)
```
laskoviymishka: Not from this PR, but it's now sitting right under the new reporter block and the reporter error just above it uses `%w`. I'd switch this `%s` to `%w` so `errors.Is`/`errors.As` can unwrap it. Non-blocking.

table/table.go
```diff
@@ -873,28 +905,48 @@ func (t Table) Scan(opts ...ScanOption) *Scan {
...
 
+// Option configures a [Table] at construction. Options are applied in order
+// after the core fields are set.
+type Option func(*Table)
+
+// WithMetricsReporter sets the metrics reporter for the table; scans created
+// from the table inherit it. A nil reporter is ignored (the table keeps its
+// default no-op reporter).
+func WithMetricsReporter(r metrics.Reporter) Option {
+	return func(t *Table) {
```
laskoviymishka, line 989: `WithReporter` returns the package-level `noopOption` when `r` is nil, but this allocates a live closure that no-ops on nil. Could mirror `WithReporter` with a shared `noopTableOption` so the two idioms match. Non-blocking.

metrics/registry.go
```diff
@@ -0,0 +1,99 @@
...
+// and undo it via t.Cleanup, keeping the process-global registry re-runnable
+// under go test -count=N.
+func Deregister(name string) {
+	registryMu.Lock()
+	defer registryMu.Unlock()
+	delete(registry, name)
+}
+
+// FromProperties builds the reporter named by props[ReporterImplKey]. An absent
+// or empty name yields [NopReporter] (reporting is opt-in), and an unrecognized name is
```
laskoviymishka, line 96: The "reporting is opt-in" choice is reasonable, but it diverges from Java, where an absent `metrics-reporter-impl` defaults to `LoggingMetricsReporter`, not a nop. Folks migrating from Java will expect scan/commit reports to show up by default, so I'd add a line to this godoc noting the divergence is intentional. Non-blocking.

laskoviymishka (changes requested):
Thanks for turning all of this around — the whole round-1 list is in and it's clean. `Refresh` now guards before copying `fresh.reporter` with two tests pinning both directions (`TestRefreshKeepsCallerReporter` / `TestRefreshInheritsCatalogReporterWhenUnset`), REST resolves from `r.props` with a comment explaining why the merged config would otherwise shadow the client's choice, `Deregister` plus the `t.Cleanup` hook make the registry re-runnable under `-count=N`, the dead Glue double-resolution is now a comment, and `doCommit` carries the reporter forward. All of that checks out.

One thing I'd tighten before I approve, and it's a consequence of the round-1 fix rather than something carried over: the `Refresh` guard uses `t.reporter.(metrics.NopReporter)` as the "caller never set one" signal, which conflates an explicit `WithMetricsReporter(NopReporter{})` opt-out (and `Combine()` of nils) with "unset" — those get reverted to the catalog default on refresh. Since this is the exact invariant we were fixing, I left an inline with a `reporterSet` flag that makes it unambiguous, plus the companion test I'd add.

On the table-lifecycle paths that still drop the reporter — I'm with @zeroshade on his thread, same class of gap; `Transaction.StagedTable()` is the one I'd specifically make sure is covered, since it rebuilds via `New(...)` without forwarding the reporter.

Two non-blockers worth deciding while the surface is small, not necessarily this round: the per-op reporter lifetime I flagged last round is still here — still fine for the stateless built-ins, but Java resolves once per catalog, and there's no `Close()` on the `Reporter` interface where Java's is `Closeable`. Since we're stabilizing this as public API now, adding `Close()` later is a breaking change, so I'd rather pin whether it goes in now. And a one-liner in the `FromProperties` godoc noting the nop-default is an intentional divergence from Java's logging default would save the next reader a trip to the source.

Tighten the guard (and its test) and I'm happy to approve — this is basically there.

tanmayrauth:
@laskoviymishka  Thanks for the thorough round-2 pass, everything's in.

For the Refresh guard, I switched to a reporterSet bool on Table that WithMetricsReporter sets when r != nil, so Refresh now guards on if !t.reporterSet instead of the NopReporter type assertion; that stops an explicit  WithMetricsReporter(metrics.NopReporter{}) opt-out (and Combine(nil, nil)) from being conflated with "unset", drops the type dependency entirely, and is pinned by a new TestRefreshKeepsExplicitNopOptOut where an explicit  NopReporter{} refreshed against a catalog handing back a live reporter stays NopReporter{} — the case that failed silently before.

On the table-lifecycle paths, you were right that Transaction.StagedTable() rebuilt via New(...) without forwarding the reporter and reset it to the nop default, so I now pass WithMetricsReporter(t.tbl.MetricsReporter()) through  (covered by TestStagedTableInheritsReporter), which forwards whatever the transaction's table carried since MetricsReporter() is never nil — and if you or @zeroshade know of other New(...) rebuild sites on those paths, point me  at them and I'll sweep them in the same pass.

On Reporter.Close() and per-catalog lifetime I'd like to defer rather than pull it into this PR, but as a conscious decision: all three built-ins (NopReporter, LoggingReporter, InMemoryReporter) are stateless so Close() would be  a no-op we'd be committing to on the interface today, and matching Java's Closeable properly also means moving to resolve-once-per-catalog with deterministic shutdown, which is a larger change than this PR's scope — so since adding Close() error later is a breaking change, I'll open a dedicated follow-up to decide Close() plus per-catalog lifetime together before we call the metrics API stable, though if you'd rather pin Close() now (no-op on the  built-ins) to avoid the break, wdyt?

Finally, the three non-blocking nits are also in: WithMetricsReporter(nil) returns a shared noopTableOption mirroring WithReporter, the Hive metadata-read error now uses %w, and the FromProperties godoc notes the nop default is  an intentional divergence from Java's LoggingMetricsReporter.

table/table.go
```diff
@@ -603,7 +628,7 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
 
 	deleteOldMetadata(fs, t.metadata, newMeta)
 
-	return New(t.identifier, newMeta, newLoc, t.fsF, t.cat), nil
+	return New(t.identifier, newMeta, newLoc, t.fsF, t.cat, WithMetricsReporter(t.MetricsReporter())), nil
```
laskoviymishka: This is the one thing I'd fix before merge, and it's a consequence of the round-2 change that carries the reporter through `doCommit` — good change, but `t.MetricsReporter()` is never nil, so `WithMetricsReporter(t.MetricsReporter())` flips `reporterSet` to true unconditionally. A table that was riding the catalog default now looks caller-set after its first commit, and every subsequent `Refresh` stops inheriting the catalog reporter — the exact invariant `TestRefreshInheritsCatalogReporterWhenUnset` pins, except that test only covers create+Refresh, not create+Commit+Refresh.

I'd propagate the state directly instead of round-tripping through the accessor — a private `withReporterState(t.reporter, t.reporterSet)` option that copies both fields, so a defaulted table stays defaulted across a commit. Same shape at `StagedTable`, though that one's short-lived so it barely matters in practice.

A create+Commit+Refresh test would lock it down. wdyt?

table/metrics_wiring_test.go
```diff
@@ -0,0 +1,161 @@
...
+		func(context.Context) (iceio.IO, error) { return iceio.NewMemFS(), nil }, nil,
+		WithMetricsReporter(rep))
+
+	staged, err := tbl.NewTransaction().StagedTable()
+	require.NoError(t, err)
+	assert.Same(t, rep, staged.MetricsReporter(),
+		"StagedTable must forward the transaction table's reporter")
+}
+
+// on: a reporter injected via WithMetricsReporter must survive a Refresh even
```
laskoviymishka: These two doc comments lost their leading function name — this one starts mid-sentence at "on:", and the one on `TestRefreshInheritsCatalogReporterWhenUnset` (line 150) opens lowercase at "the caller never overrode". Looks like the `// TestX ...` opener got clipped on both. Worth restoring the convention since these are the tests doing the load-bearing work here.

metrics/registry.go
```diff
@@ -0,0 +1,104 @@
...
+		panic("metrics: Register called twice for " + name)
+	}
+	registry[name] = factory
+}
+
+// Deregister removes a previously registered reporter factory. It is a no-op if
+// name is not registered. This exists primarily so tests can register a factory
+// and undo it via t.Cleanup, keeping the process-global registry re-runnable
+// under go test -count=N.
+func Deregister(name string) {
```
laskoviymishka, line 86: Since this is public API and we're stabilizing it now — `Deregister` will happily remove the built-in `nop`/`logging` factories, and because `Register` panics on duplicate, nothing can put them back for the rest of the process. So a stray `Deregister("nop")` permanently breaks `metrics-reporter-impl=nop` everywhere. I'd guard the two built-in names, or at least call the hazard out in the godoc. wdyt?

laskoviymishka (changes requested):
Thanks for turning round 2 around — the `reporterSet` flag is exactly right, the guard now keys off `!t.reporterSet` instead of the `NopReporter` type assertion, and `TestRefreshKeepsExplicitNopOptOut` pins the explicit-opt-out case the old type check would have silently reverted. `StagedTable()` forwards the reporter now with `TestStagedTableInheritsReporter` covering it, and the `FromProperties` godoc calls out the nop-default divergence from Java. That all checks out.

One new thing before I approve, and it's a side effect of the round-2 `doCommit` fix rather than anything carried over: carrying the reporter through `doCommit` via `WithMetricsReporter(t.MetricsReporter())` flips `reporterSet` to true on every commit, because the accessor is never nil. So a table riding the catalog default gets frozen to that default after its first commit and stops inheriting on later refreshes — the same inheritance invariant we've spent two rounds pinning, just on the create+Commit+Refresh path the current test doesn't cover. I left an inline with the fix and the test I'd add.

The other one is what I flagged last round, and it's still the moment to decide: no `Close()` on the `Reporter` interface where Java's is `Closeable`, and the per-op `FromProperties` lifetime versus Java's per-catalog. The concrete case is the first stateful reporter — an HTTP-backed one will need shared-client cleanup, and adding `io.Closer` after this surface ships is a breaking change. I'm happy to take it to a tracked follow-up rather than block on it, but I'd want that to be a deliberate call now, not a default.

Fix the commit-path guard (plus the test) and I'm happy to approve — this is right at the line.

metrics/registry.go
tanmayrauth: Guarded rather than just documented — Deregister is now a no-op for nop and logging, with the reasoning in the godoc (since Register panics on duplicates, a removed built-in could never be restored). I  added TestDeregisterProtectsBuiltins so both stay resolvable after a Deregister attempt.

table/metrics_wiring_test.go
tanmayrauth: Fixed both — restored the // TestRefreshKeepsCallerReporter … and // TestRefreshInheritsCatalogReporterWhenUnset … openers. Clipped in a round-2 edit

table/table.go
tanmayrauth: Done. I went with the withReporterState(t.reporter, t.reporterSet) option exactly as you suggested — copies both fields so it doesn't round-trip through the never-nil accessor — and applied the same at   StagedTable. TestCommitPreservesDefaultedReporterInheritance locks down create+Commit+Refresh; I verified it fails on the old line and passes now.

tanmayrauth:
Both in, and I made the deliberate call on Close() you asked for.

Commit-path guard — fixed, and you were exactly right about the mechanism: MetricsReporter() is never nil, so carrying it through doCommit via WithMetricsReporter(...) flipped reporterSet true on every commit and froze a catalog-defaulted table after its first commit. I replaced it with a private  withReporterState(t.reporter, t.reporterSet) that copies both fields verbatim, applied at both doCommit and StagedTable, so a defaulted table stays defaulted across a commit. TestCommitPreservesDefaultedReporterInheritance pins create+Commit+Refresh — I confirmed it fails against the old line (freezes to nop)  and passes now.

 Close() / per-catalog lifetime — I'd floated deferring this last round; on reflection you're right that it's the moment to decide, and since adding it later is the breaking change, I pinned it now instead of taking the default:
 - Reporter now embeds io.Closer (Java's Closeable shape). Built-ins close as no-ops; the composite closes its children with panic isolation + error joining. The interface surface is settled, so no external implementer eats a break later.
  - I also did the per-catalog half rather than leave it half-solved: added metrics.CachedReporter, which resolves the reporter once and caches it, and wired every catalog (rest, glue, sql, hadoop, hive) to hold one plus a Close() that releases it — hive's existing Close now closes the reporter too. So the first stateful HTTP reporter is built once per catalog and has an owner that closes it, matching Java.

The one thing I deliberately left out is Close() on the catalog.Catalog interface itself — that's a much broader breaking change to a widely-implemented public interface, so it feels like its own decision rather than something to fold in here. Catalogs expose Close() as concrete methods for now (reachable via  an io.Closer assertion, same as hive already did). Happy to take interface-level catalog Close to a follow-up if you'd want it — but the reporter's own lifetime and cleanup are handled in this PR.

zeroshade (changes requested):
(Medium) The metrics reporter is initialized *after* the catalog mutation, so an invalid reporter turns a successful create into a reported failure.

In the Hadoop catalog `CreateTable` (`catalog/hadoop/hadoop.go`), the metadata file is committed (`:554`) and the version hint written (`:558`) before `c.reporter.Get(c.props)` runs (`:560`); if that errors, `CreateTable` returns an error (`:562`) even though the table already exists on disk.

The Glue catalog has the same shape: `CreateTable`/`RegisterTable` create the Glue entry (`catalog/glue/glue.go:272` / `:306`) and only then reach `reporter.Get` via the trailing `LoadTable` (`glue.go:743`).

With an invalid `metrics-reporter-impl`, callers receive an error while the table was actually created, and a retry then hits `ErrTableAlreadyExists` — a confusing partial-success.

Fix: resolve/validate the reporter before any catalog mutation (or once at catalog construction and reuse it) so a reporter error can't make a successful create look like a failure.

catalog/hadoop/hadoop.go
```diff
@@ -549,12 +557,18 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 
 	c.writeVersionHint(ident, version)
 
+	reporter, err := c.reporter.Get(c.props)
```
laskoviymishka: This is the one I'd still hold on — it's zeroshade's ordering thread, and I think it's actually a bit worse than the partial-success framing we started with.

The first `Get(c.props)` fires after `commitMetadataFile` and `writeVersionHint` have already written the table to disk, so an invalid `metrics-reporter-impl` turns a create that physically succeeded into a returned error, and the retry then hits `ErrTableAlreadyExists`. But because `CachedReporter` caches that first error, every later `LoadTable`/`loadTable`/`CreateTable` on the same catalog returns it too — one typo wedges the whole catalog for all table ops until it's reconstructed. Glue has the identical shape through `convertGlueToIceberg`.

`CachedReporter` is built exactly for resolving once up front — I'd call `c.reporter.Get(c.props)` in `NewCatalog` (or validate before the first mutation) so a reporter error can't masquerade as a failed create and can't poison later ops. wdyt?

metrics/cached_reporter.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Close closes the built reporter, if one was ever built, and is a no-op
+// otherwise (including when Get was never called or returned an error). Close is
+// expected at catalog shutdown, after operations have quiesced; it is safe for
+// concurrent use, but a Get racing a Close has no defined ordering.
+func (c *CachedReporter) Close() error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.rep != nil {
```
laskoviymishka: Since the whole point of the io.Closer work is the first stateful (HTTP-backed) reporter, I'd tighten this a touch.

We don't reset `c.rep` after closing, so a second `Close()` re-closes the underlying reporter — the contract says Close is called once, and a user reporter isn't obliged to be idempotent. Niling `c.rep` here makes Close idempotent and also stops a post-Close `Get()` from handing back an already-closed reporter. The same guard is interface-identity, so a factory returning a typed-nil (`(*Custom)(nil)`) would slip past `c.rep != nil` and panic in Close unrecovered — worth a line on the factory contract, or a defensive check. wdyt?

catalog/glue/glue.go
```diff
@@ -241,6 +245,12 @@ func (c *Catalog) CatalogType() catalog.Type {
 	return catalog.Glue
 }
 
+// Close releases the catalog's metrics reporter. The Glue catalog does not own
+// the lifetime of the AWS clients it was configured with, so only the reporter
+// is released. Callers holding a [catalog.Catalog] can reach this via an
+// io.Closer type assertion.
+func (c *Catalog) Close() error { return c.reporter.Close() }
```
laskoviymishka, line 267: The reporter half of the round-2/3 `Close()` ask landed nicely — but this is where the catalog half is still missing: `Close()` is on every concrete catalog, not on `catalog.Catalog`, and `catalog.Load` hands back the interface. So the common caller can't `defer cat.Close()` without a type assertion, and the leak becomes the default outcome for the first stateful reporter.

I don't think adding `Close()` to the interface has to block this PR — you're right that it's a broader breaking change. But I'd like a tracked follow-up with an owner before we merge, either adding it to the interface or a documented `catalog.Closer` optional-interface idiom like `PurgeableTable`. Can we file that?

catalog/hive/hive.go
```diff
@@ -94,8 +98,10 @@ func (c *Catalog) CatalogType() catalog.Type {
 	return catalog.Hive
 }
 
+// Close releases the Hive client and the catalog's metrics reporter, joining
+// any errors so one failing close does not skip the other.
 func (c *Catalog) Close() error {
-	return c.client.Close()
+	return errors.Join(c.client.Close(), c.reporter.Close())
```
laskoviymishka: The comment says one failing close doesn't skip the other, but that only holds for errors — `errors.Join(c.client.Close(), c.reporter.Close())` evaluates both args before the call, so if `c.client.Close()` panics the reporter never closes and leaks. Sequencing into locals first (`clientErr := ...; reporterErr := ...`) keeps the guarantee the comment promises.

tanmayrauth:
@zeroshade  Fixed it, and it turned out to be broader than the two paths you flagged, so I swept every reporter-resolution site while I was in there. Hadoop CreateTable and Glue CreateTable/RegisterTable now resolve the reporter before the metadata/version-hint write and the Glue entry; the same after-mutation shape existed in Hive CreateTable, SQL CreateTable, and REST CreateTable/RegisterTable (all resolved the reporter only through a trailing LoadTable/tableFromResponse), and those are fixed the same way. The paths that were already safe I left alone — Hive RegisterTable resolves before client.CreateTable, Glue/SQL CommitTable resolve during the pre-mutation load of the current table, and Hive/REST CommitTable are metadata-only. I went with resolve-before-mutation everywhere rather than your "once at construction" alternative because glue.NewCatalog and hive.NewCatalogWithClient don't return an error, so a bad metrics-reporter-impl can't be surfaced at construction there; CachedReporter still caches the first Get, so the trailing call just reuses it with no double resolution. Added tests covering create-with-invalid-reporter leaving no partial state (Hadoop/SQL assert the table doesn't exist afterward, Hive asserts the metastore client is never called, REST asserts the create POST is never sent), and rebased on latest main.

tanmayrauth:
@zeroshade @laskoviymishka can you please review it?

laskoviymishka (changes requested):
Round 3 turned around cleanly — the commit-path inheritance fix is exactly right, and the two items that have been open since round 2 both landed. Nice work; this has come a long way.

One thing left, and it's zeroshade's ordering thread rather than a new one from me. The first `c.reporter.Get(c.props)` still runs after the table is committed to disk in Hadoop `CreateTable` (and in Glue via `convertGlueToIceberg`), so an invalid `metrics-reporter-impl` turns a create that physically succeeded into a returned error, and the retry hits `ErrTableAlreadyExists`. The `CachedReporter` refactor actually sharpens this rather than fixing it: the first error gets cached, so one bad name wedges every later `LoadTable`/`CreateTable` on that catalog too. `CachedReporter` is built to resolve once up front — calling `Get` in `NewCatalog` (or validating before the first mutation) closes it for both catalogs. I left the mechanism inline.

Not blocking this round, but I'd like an owner on it before we merge: the `Close()` that landed is on each concrete catalog, not on `catalog.Catalog`, so callers holding the interface from `catalog.Load` can't `defer cat.Close()` without asserting — which is where the leak actually bites the first stateful reporter. A tracked follow-up (interface `Close()` or a documented `catalog.Closer` idiom) is fine. Same for the `internal/utils.go` staged-table paths that build with a nop reporter — harmless today, but worth a note before the emission PR lands.

Everything I asked for last round is in:
- `CachedReporter` gives the reporter a per-catalog lifetime — the per-op-vs-per-catalog thread from round 1 is resolved.
- `Reporter` embeds `io.Closer` with `Close()` on Nop/Logging/InMemory/composite and panic isolation on the composite Close — the Java-Closeable item from round 2/3.
- `withReporterState` on `doCommit` and `StagedTable`, with `TestCommitPreservesDefaultedReporterInheritance` — the "frozen to catalog default after first commit" bug is gone.
- `reporterSet` guard with `TestRefreshKeepsExplicitNopOptOut` still holding the explicit-opt-out case.

Fix the ordering one and get the interface-Close follow-up filed, and I'm happy to approve. Really close now.

catalog/hive/hive.go
tanmayrauth: Fixed, though pure sequencing wouldn't cover it — errors.Join already evaluates both regardless of error, so only a panic skips the second close. Wrapped each in a closeIsolated helper that recovers into an error and joins (same shape as compositeReporter.Close). Pinned by a test where the client panics on  Close and the reporter still closes.

catalog/glue/glue.go
tanmayrauth: Went with the catalog.Closer optional-interface idiom you suggested rather than deferring — documented it PurgeableTable-style with a defer usage example and asserted it on all five catalogs, so catalog.Load callers get Close via one documented assertion. Left it off the Catalog interface itself since that's  the breaking change; happy to file a follow-up if you'd still want the promotion tracked.

metrics/cached_reporter.go
tanmayrauth: Close now nils the reporter after closing, so it's idempotent and a post-Close Get returns NopReporter{} rather than a released reporter. Documented the typed-nil case as a Factory contract (must return a usable reporter or an error) instead of a reflect check. Added tests for both.

catalog/hadoop/hadoop.go
tanmayrauth: Moved resolution to construction. Hadoop resolves in NewCatalog; glue's NewCatalog now returns (*Catalog, error) and resolves there too (updated the registrar, CLI, and callers). A bad name now fails construction instead of a post-write CreateTable, and since CachedReporter resolves once, no cached error can wedge later ops. Test updated to pin construction-time failure.

laskoviymishka (approved):
This is good to go. Approving.

The ordering issue is fixed now. Reporter resolution happens during `NewCatalog` for all five catalogs, before any metadata or version-hint write during create. An invalid `metrics-reporter-impl` now fails before the create starts, rather than returning an error after the table may already have been created. This also removes the risk of a cached reporter error leaving the catalog unusable.

`CachedReporter.Close` is idempotent now as well: it clears `rep` and marks the reporter closed. That gives us sane behavior before we introduce a stateful reporter.

Thanks for fixing every reporter resolution site, not only the two originally called out.

zeroshade (changes requested):
Re-reviewing. The earlier Hadoop/Glue "reporter resolved after mutation" concern looks addressed — thanks. One new issue on the REST path (and CI is currently red):

**Medium — server-vended reporter can break table ops** (`catalog/rest/rest.go:1065`)
REST reporter selection reads `r.props`, which is built from `/v1/config` defaults/overrides as well as client options. A server-vended `metrics-reporter-impl` (e.g. a Java class name, or any unregistered Go name) can make table load/create/register fail even when the client didn't configure metrics reporting — contradicting the nearby comment that server-vended selection shouldn't be an accidental side effect.
*Fix:* keep a separate client-supplied metrics-reporter property set, or strip `metrics-reporter-impl` from server-derived REST config before initializing the cached reporter.

Also: 4 CI checks are currently failing — please take a look.

tanmayrauth:
@zeroshade  the comment was misleading, and you're right that reading from r.props didn't actually avoid server-vended selection. r.props is toProps(ops) after fetchConfig merges the /v1/config defaults and overrides, so a  server-vended metrics-reporter-impl was already folded in — an unregistered name (e.g. a Java class) would fail the client's load/create/register even when the client never configured metrics.

Fixed by keeping a separate client-supplied set: init now captures reporterProps from the client options before fetchConfig merges server config, and reporter resolution (load, create, register) reads reporterProps instead of  r.props. So a server-vended impl — registered or not — can no longer select or break the client's reporter; with no client selection, ops fall back to nop. Added TestServerVendedReporterImplIgnored (server vends does-not-exist   via overrides, client sets nothing → create succeeds with a nop reporter), which fails against the old r.props path.

tanmayrauth:
@zeroshade  can you please take a look when you get chance?

catalog/sql/sql.go
```diff
@@ -359,6 +363,14 @@ func (c *Catalog) CatalogType() catalog.Type {
 	return catalog.SQL
 }
 
+// Close releases the catalog's metrics reporter. The SQL catalog is constructed
+// from a caller-owned database handle, so it does not close that handle; only
+// the reporter is released. Callers holding a [catalog.Catalog] can reach this
+// via a [catalog.Closer] type assertion.
+func (c *Catalog) Close() error { return c.reporter.Close() }
```
zeroshade: Non-blocking (fine to fold into the rebase):

This doc says the SQL catalog is "constructed from a caller-owned database handle" — true for `NewCatalog`, but the `init()` registrar path does its own `sql.Open(driver, uri)` and hands that DB to `NewCatalog`. For a registry-loaded catalog (`catalog.Load("sql", …)`) nobody else holds that handle, so `Close()` — which only releases the reporter — leaks the connection pool that the new `catalog.Closer` contract implies it cleans up.

Suggestion: track ownership (e.g. an `ownsDB` bool set on the registrar path) and close the owned `*sql.DB`/bun handle in `Close()`, while keeping the current non-closing behavior for caller-supplied handles passed to `NewCatalog`.

zeroshade (approved):
Re-reviewed after `75a0077`. Both findings from my earlier reviews are resolved — approving. 🎉

- ✅ **Server-vended reporter** — the reporter now resolves from a client-only `reporterProps` snapshot captured before `/v1/config` defaults/overrides are merged, so a server-vended `metrics-reporter-impl` can't select (or, via an unregistered name, break) a reporter the client never configured. `tableFromResponse`, `CreateTable`, and `RegisterTable` all use it.
- ✅ **Reporter resolved after mutation** — now resolved at catalog construction (`hadoop`/`glue` `NewCatalog`), so a bad reporter fails before any table is written rather than turning a successful create into a reported failure.

The idempotent `Close` + `catalog.Closer` hardening is a nice touch.

Two non-blocking follow-ups for whenever you next push (not gating this approval):

1. **Needs a rebase** — the branch currently conflicts with `main`, which advanced under it across `catalog/{catalog,glue,hive,rest,sql}.go` + tests, so it can't merge as-is.
2. One inline note below on the new `catalog.Closer` contract for the registry-created SQL catalog.

tanmayrauth:
@zeroshade Thanks for the re-review! Both follow-ups handled:
  


  For catalog.Closer contract for the registry-loaded SQL catalog - good catch, I've fixed it. Added an ownsDB flag set only on the registrar path (which does its own sql.Open); Close() now closes that owned handle in addition to releasing the reporter. Caller-supplied handles passed to NewCatalog stay ownsDB=false and are left open as before. Doc comment on Close() updated to describe both cases.

PR #1432, opened 2026-07-10
feat(udf): add SQL UDF REST catalog read support

rraulinio:
## Description

Addressing part 2 from https://github.com/apache/iceberg-go/issues/1360, following the SQL UDF metadata model added in #1418.

This adds read-side SQL UDF support to the REST catalog through a new optional `FunctionCatalog` capability interface. REST catalogs can list functions in a namespace, load a function and check whether a function exists. Loaded functions are represented by a new `udf.UDF` wrapper containing the catalog identifier, parsed UDF metadata and optional metadata location.

The REST implementation supports the two function endpoints currently defined by the Iceberg REST OpenAPI:

- `GET /v1/{prefix}/namespaces/{namespace}/functions`
- `GET /v1/{prefix}/namespaces/{namespace}/functions/{function}`

Function endpoints are not part of the REST spec's assumed default endpoint set, so they are used only when advertised by the server through `ConfigResponse.endpoints`. An unsupported load or existence check returns `ErrEndpointNotSupported`, while an unsupported listing yields no results without issuing a request, matching the existing list behavior for optional endpoints.

`ListFunctions` is a paginated iterator and handles the function-specific `CatalogObjectIdentifier` wire shape: each identifier is a flat array of hierarchy levels, such as `["accounting", "tax", "paid"]`, rather than the `{namespace, name}` object used for tables and views. `LoadFunction` parses the returned metadata using the `udf` package and returns all overloads from the single metadata response. The optional `metadata-location` is preserved when present.

The catalog adds `ErrNoSuchFunction` for `NoSuchFunctionException` responses and validates function identifiers independently from table and view identifiers. `CheckFunctionExists` uses `LoadFunction` because the REST spec does not define a HEAD endpoint for functions; invalid identifiers remain errors, while a server-reported missing function returns `false` without an error.

This PR does not add create, replace or drop operations because standardized REST write endpoints for functions are not currently defined. It also does not resolve overloads or execute UDF SQL bodies; those remain query-engine responsibilities.

The tests cover list and load responses, pagination, early iterator termination, page-size handling, subsequent-page errors, missing namespaces and functions, malformed metadata, identifier validation, GET-based existence checks and endpoint capability negotiation. They also add focused coverage for existing UDF metadata initialization and nil-properties safeguards.

## Tests

- `go test ./...`
- `go test ./catalog/rest -run 'Function|Endpoint' -count=10`
- `go test -race ./catalog ./catalog/rest ./udf -run 'Function|Endpoint|TestUDF|TestBuilderFromBaseNilProperties|TestMetadataInitDefaults' -count=1`
- `go vet ./...`

### DIFF
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -39,7 +39,6 @@ import (
 	"github.com/apache/iceberg-go"
 	iceinternal "github.com/apache/iceberg-go/internal"
 	"github.com/apache/iceberg-go/table"
-	"github.com/apache/iceberg-go/udf"
 )
 
 type Type string
@@ -192,30 +191,6 @@ type TransactionalCatalog interface {
 // remains the single source of truth.
 type PurgeableTable interface {
 	PurgeTable(ctx context.Context, identifier table.Identifier) error
-}
-
-// FunctionCatalog is an optional interface implemented by catalogs that
-// support the Iceberg REST function (SQL UDF) endpoints. The function
-// endpoints are not part of the spec's assumed default endpoint set, so
-// servers must advertise them: unsupported loads report an
-// endpoint-not-supported error and unsupported listings yield no results.
-// Callers should check for this capability via a type assertion:
-//
-//	if fc, ok := cat.(catalog.FunctionCatalog); ok {
-//	    fn, err := fc.LoadFunction(ctx, ident)
-//	}
-type FunctionCatalog interface {
-	// ListFunctions returns the function identifiers under a namespace,
-	// with the returned identifiers containing the information required
-	// to load the function via this catalog.
-	ListFunctions(ctx context.Context, namespace table.Identifier) iter.Seq2[table.Identifier, error]
-	// LoadFunction loads a function from the catalog. All overloaded
-	// definitions are included in the single metadata response.
-	LoadFunction(ctx context.Context, identifier table.Identifier) (*udf.UDF, error)
-	// CheckFunctionExists returns if the function exists. The REST spec
-	// defines no HEAD endpoint for functions, so existence is checked by
-	// loading the function.
-	CheckFunctionExists(ctx context.Context, identifier table.Identifier) (bool, error)
 }
 
 func ToIdentifier(ident ...string) table.Identifier {
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -205,12 +205,18 @@ func ToIdentifier(ident ...string) table.Identifier {
 	return table.Identifier(ident)
 }
 
-func TableNameFromIdent(ident table.Identifier) string {
+// ObjectNameFromIdent returns the name of a catalog object (a table, view,
+// or function), which is the last element of its identifier.
+func ObjectNameFromIdent(ident table.Identifier) string {
 	if len(ident) == 0 {
 		return ""
 	}
 
 	return ident[len(ident)-1]
+}
+
+func TableNameFromIdent(ident table.Identifier) string {
+	return ObjectNameFromIdent(ident)
 }
 
 func NamespaceFromIdent(ident table.Identifier) table.Identifier {
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -39,6 +39,7 @@ import (
 	"github.com/apache/iceberg-go"
 	iceinternal "github.com/apache/iceberg-go/internal"
 	"github.com/apache/iceberg-go/table"
+	"github.com/apache/iceberg-go/udf"
 )
 
 type Type string
@@ -63,6 +64,7 @@ var (
 	ErrNamespaceNotEmpty      = errors.New("namespace is not empty")
 	ErrNoSuchView             = errors.New("view does not exist")
 	ErrViewAlreadyExists      = errors.New("view already exists")
+	ErrNoSuchFunction         = errors.New("function does not exist")
 	ErrEmptyCommitList        = errors.New("commit list must not be empty")
 	ErrMissingIdentifier      = errors.New("every table commit must have a valid identifier")
 )
@@ -192,6 +194,30 @@ type PurgeableTable interface {
 	PurgeTable(ctx context.Context, identifier table.Identifier) error
 }
 
+// FunctionCatalog is an optional interface implemented by catalogs that
+// support the Iceberg REST function (SQL UDF) endpoints. The function
+// endpoints are not part of the spec's assumed default endpoint set, so
+// servers must advertise them: unsupported loads report an
+// endpoint-not-supported error and unsupported listings yield no results.
+// Callers should check for this capability via a type assertion:
+//
+//	if fc, ok := cat.(catalog.FunctionCatalog); ok {
+//	    fn, err := fc.LoadFunction(ctx, ident)
+//	}
+type FunctionCatalog interface {
+	// ListFunctions returns the function identifiers under a namespace,
+	// with the returned identifiers containing the information required
+	// to load the function via this catalog.
+	ListFunctions(ctx context.Context, namespace table.Identifier) iter.Seq2[table.Identifier, error]
+	// LoadFunction loads a function from the catalog. All overloaded
+	// definitions are included in the single metadata response.
+	LoadFunction(ctx context.Context, identifier table.Identifier) (*udf.UDF, error)
+	// CheckFunctionExists returns if the function exists. The REST spec
+	// defines no HEAD endpoint for functions, so existence is checked by
+	// loading the function.
+	CheckFunctionExists(ctx context.Context, identifier table.Identifier) (bool, error)
+}
+
 func ToIdentifier(ident ...string) table.Identifier {
 	if len(ident) == 1 {
 		if ident[0] == "" {
@@ -216,7 +242,7 @@ func NamespaceFromIdent(ident table.Identifier) table.Identifier {
 	return ident[:len(ident)-1]
 }
 
-func validateTableOrViewIdentifier(ident table.Identifier, notFoundErr error) error {
+func validateIdentifier(ident table.Identifier, notFoundErr error) error {
 	if len(ident) < 2 {
 		return fmt.Errorf("%w: missing namespace or invalid identifier %v",
 			notFoundErr, strings.Join(ident, "."))
@@ -241,12 +267,17 @@ func validateTableOrViewIdentifier(ident table.Identifier, notFoundErr error) er
 
 // ValidateTableIdentifier checks that an identifier contains at least one valid namespace level and a table name.
 func ValidateTableIdentifier(ident table.Identifier) error {
-	return validateTableOrViewIdentifier(ident, ErrNoSuchTable)
+	return validateIdentifier(ident, ErrNoSuchTable)
 }
 
 // ValidateViewIdentifier checks that an identifier contains at least one valid namespace level and a view name.
 func ValidateViewIdentifier(ident table.Identifier) error {
-	return validateTableOrViewIdentifier(ident, ErrNoSuchView)
+	return validateIdentifier(ident, ErrNoSuchView)
+}
+
+// ValidateFunctionIdentifier checks that an identifier contains at least one valid namespace level and a function name.
+func ValidateFunctionIdentifier(ident table.Identifier) error {
+	return validateIdentifier(ident, ErrNoSuchFunction)
 }
 
 type CreateTableOpt func(*CreateTableCfg)
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -1028,7 +1028,7 @@ func (r *Catalog) splitIdentForPath(ident table.Identifier) (string, string, err
 		return "", "", err
 	}
 
-	return r.encodeNamespace(catalog.NamespaceFromIdent(ident)), catalog.TableNameFromIdent(ident), nil
+	return r.encodeNamespace(catalog.NamespaceFromIdent(ident)), catalog.ObjectNameFromIdent(ident), nil
 }
 
 func (r *Catalog) splitViewIdentForPath(ident table.Identifier) (string, string, error) {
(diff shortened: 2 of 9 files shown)


rraulinio:
cc @laskoviymishka @tanmayrauth

catalog/catalog.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/apache/iceberg-go"
 	iceinternal "github.com/apache/iceberg-go/internal"
 	"github.com/apache/iceberg-go/table"
+	"github.com/apache/iceberg-go/udf"
```
laskoviymishka: Putting `FunctionCatalog` in the base `catalog` package means `catalog` now imports `udf`, and since `LoadFunction` returns `*udf.UDF`, every consumer of the base package transitively depends on `udf`.

That's also asymmetric with views: there's no `ViewCatalog` interface here — `LoadView`/`ListViews`/`CheckViewExists` are just methods on `*rest.Catalog`. I'd follow that precedent and keep `FunctionCatalog` local to `catalog/rest` (the `var _ catalog.FunctionCatalog` assertion in rest.go goes away with it). If we genuinely need a cross-package interface, I'd at least avoid the concrete `*udf.UDF` return type so the base package stays leaf-independent. wdyt?

catalog/catalog.go
```diff
@@ -192,6 +194,30 @@ type PurgeableTable interface {
 	PurgeTable(ctx context.Context, identifier table.Identifier) error
 }
 
+// FunctionCatalog is an optional interface implemented by catalogs that
+// support the Iceberg REST function (SQL UDF) endpoints. The function
+// endpoints are not part of the spec's assumed default endpoint set, so
+// servers must advertise them: unsupported loads report an
+// endpoint-not-supported error and unsupported listings yield no results.
```
laskoviymishka: This doc promises "unsupported loads report an endpoint-not-supported error," but `ErrEndpointNotSupported` lives in `catalog/rest`, not `catalog`. A caller using the documented type-assertion pattern imports only `catalog` and can't `errors.Is` against a sentinel it can't name.

Either move `ErrEndpointNotSupported` up to `catalog`, or drop the endpoint-negotiation language from this doc and describe it only in the rest.go methods. This one resolves naturally if we go with the "keep it in `catalog/rest`" direction from the layering comment above.

catalog/rest/rest.go
```diff
@@ -2045,3 +2054,145 @@ func (r *Catalog) LoadView(ctx context.Context, identifier table.Identifier) (*v
...
+
+// CheckFunctionExists returns if the function exists. The REST spec defines
+// no HEAD endpoint for functions, so existence is checked by loading the
+// function; if loading is unsupported, its ErrEndpointNotSupported surfaces
+// rather than a bogus "not found".
+func (r *Catalog) CheckFunctionExists(ctx context.Context, identifier table.Identifier) (bool, error) {
+	// Validate first, as the table and view existence checks do: an invalid
+	// identifier surfaces as an error, so ErrNoSuchFunction below can only
+	// mean a server-reported "not found".
+	if err := catalog.ValidateFunctionIdentifier(identifier); err != nil {
```
laskoviymishka: This validation is redundant — `LoadFunction` immediately calls `splitFunctionIdentForPath`, which calls `ValidateFunctionIdentifier` again, so the identifier is validated twice.

The comment is also inaccurate: `CheckTableExists`/`CheckViewExists` don't do an explicit up-front validate, they call `splitIdentForPath`/`splitViewIdentForPath` directly (one pass) and then HEAD/Load. I'd drop the explicit call and the comment here — `TestLoadFunctionInvalidIdentifier` still passes because `LoadFunction` validates first.

catalog/rest/rest_functions_test.go
```diff
@@ -0,0 +1,428 @@
...
+
+	cat, err := rest.NewCatalog(context.Background(), "rest", r.srv.URL, rest.WithOAuthToken(TestToken))
+	r.Require().NoError(err)
+	// Passing in a custom page size through context
+	ctx := cat.SetPageSize(context.Background(), customPageSize)
+
+	var lastErr error
+	functions := make([]table.Identifier, 0)
+	for function, err := range cat.ListFunctions(ctx, catalog.ToIdentifier(namespace)) {
+		functions = append(functions, function)
```
laskoviymishka, line 93: This appends `function` before checking `err`. If the iterator ever yields `(empty-ident, err)`, the empty identifier gets appended before the error is caught — the assertion on the collected slice would then be checking corrupted data.

`TestListFunctionsPaginationErrorOnSubsequentPage` gets this right (checks `err`, breaks, then appends). I'd invert this loop to match — check `err` first, append on success.

catalog/rest/rest.go
```diff
@@ -2045,3 +2054,145 @@ func (r *Catalog) LoadView(ctx context.Context, identifier table.Identifier) (*v
...
+	}
+
+	path, err := endpointLoadFunction.reqPath(ns, fn)
+	if err != nil {
+		return nil, err
+	}
+
+	rsp, err := doGet[loadFunctionResponse](ctx, r.baseURI, path,
+		r.cl, map[int]error{
+			http.StatusNotFound: catalog.ErrNoSuchFunction,
```
laskoviymishka, line 2193: This maps every 404 to `ErrNoSuchFunction`, but the spec's GET load defines two 404 bodies — `NoSuchNamespaceException` and `NoSuchFunctionException`. The status-code-only override ignores `error.type`, so `LoadFunction` on a deleted namespace returns `ErrNoSuchFunction`, and `CheckFunctionExists` then reports `(false, nil)` for a missing namespace.

Java's view/table handlers discriminate on the type string before falling through. This is a pre-existing gap for `LoadView`/`LoadTable` too, so not a blocker for this PR — but worth a look as a consistent follow-up, since the function path is a fresh chance to get it right.

catalog/rest/rest.go
```diff
@@ -2045,3 +2054,145 @@ func (r *Catalog) LoadView(ctx context.Context, identifier table.Identifier) (*v
...
+		})
+	if err != nil {
+		return nil, err
+	}
+
+	metadata, err := udf.ParseMetadataBytes(rsp.RawMetadata)
```
laskoviymishka, line 2215: If a server returns `metadata-location` with no `metadata` key, `RawMetadata` is nil and this surfaces as `"unexpected end of JSON input"`, which is confusing. The spec marks metadata required so only a buggy server hits it, but a `len(rsp.RawMetadata) == 0` guard with a clearer message would be a cheap robustness win. Non-blocking.

catalog/rest/rest_functions_test.go
```diff
@@ -0,0 +1,428 @@
...
+	cat, err := rest.NewCatalog(context.Background(), "rest", r.srv.URL, rest.WithOAuthToken(TestToken))
+	r.Require().NoError(err)
+
+	var lastErr error
+	for _, err := range cat.ListFunctions(context.Background(), table.Identifier{}) {
+		if err != nil {
+			lastErr = err
+		}
+	}
+	r.Error(lastErr)
```
laskoviymishka: `r.Error(lastErr)` passes for any error, including a wrong sentinel — compare `TestListTables404`, which pins `ErrorIs(ErrNoSuchNamespace)`. I'd add an `ErrorIs` here to nail down which error we expect.

Same file has a couple of loops (this one, `TestListFunctions404`) that keep iterating after capturing the error instead of breaking — harmless since the iterator stops after an error, but it reads as if multiple errors could arrive. I'd break after the first for clarity while we're here.

udf/udf.go
```diff
@@ -0,0 +1,58 @@
...
+	return &UDF{
+		identifier:       ident,
+		metadata:         meta,
+		metadataLocation: metadataLocation,
+	}
+}
+
+// Equals reports whether two UDFs have the same identifier, metadata
+// location and metadata.
+func (u UDF) Equals(other UDF) bool {
```
laskoviymishka, line 49: `New` returns `*UDF` but `Equals` takes a value, so callers end up writing `fn.Equals(*same)` (visible in the test). Minor friction — I'd either use pointer receivers/argument to match the constructor, or drop a one-line doc noting the value-copy semantics are intentional.

catalog/rest/rest.go
```diff
@@ -1038,6 +1039,14 @@ func (r *Catalog) splitViewIdentForPath(ident table.Identifier) (string, string,
 	return r.encodeNamespace(catalog.NamespaceFromIdent(ident)), catalog.TableNameFromIdent(ident), nil
 }
 
+func (r *Catalog) splitFunctionIdentForPath(ident table.Identifier) (string, string, error) {
+	if err := catalog.ValidateFunctionIdentifier(ident); err != nil {
+		return "", "", err
+	}
+
+	return r.encodeNamespace(catalog.NamespaceFromIdent(ident)), catalog.TableNameFromIdent(ident), nil
```
laskoviymishka: Using `catalog.TableNameFromIdent` to pull the function name works (it's the last element) but reads oddly — `LoadView` has the same wart. Not for this PR, but a type-agnostic `catalog.ObjectNameFromIdent` would clean up all three call sites.

laskoviymishka (changes requested):
This is a clean, well-scoped read-side implementation, and the test coverage is genuinely broad — pagination, 404s, malformed metadata, and the advertised-but-unsupported endpoint path are all exercised. The `CatalogObjectIdentifier` wire shape and the endpoint negotiation both look right to me.

I'd hold it before merging, though, mostly on one decision: where `FunctionCatalog` lives. Putting it in the base `catalog` package pulls a `catalog` → `udf` import into every consumer of that package, and it's asymmetric with how we model views — there's no `ViewCatalog` interface, `LoadView`/`ListViews` are just methods on `*rest.Catalog`. The interface doc also promises endpoint-not-supported behavior through `ErrEndpointNotSupported`, which lives in `catalog/rest`, so a caller holding only the `catalog`-package interface can't `errors.Is` against it. Both of those dissolve if we keep `FunctionCatalog` local to `catalog/rest` and follow the view precedent, so I'd settle that direction first.

Things I'd like to settle in this PR:

- Decide where `FunctionCatalog` lives (I lean toward `catalog/rest`, matching views), which also resolves the `ErrEndpointNotSupported` doc/location mismatch.
- Drop the redundant explicit validation in `CheckFunctionExists` and fix its comment — the table/view checks don't actually validate up front the way the comment claims.
- Fix `TestListFunctions200` to check the error before appending, matching the pagination-error test.
- Worth a look but not blocking: 404s mapping every case to `ErrNoSuchFunction` (loses namespace-not-found), a nil-metadata guard, and pinning the sentinel in `TestListFunctionsInvalidNamespace`.

Once the placement question and the two concrete bugs are addressed, happy to take another pass and approve.

catalog/catalog.go
rraulinio: Agreed. I moved `FunctionCatalog` into `catalog/rest`, removed the `catalog -> udf` dependency and removed the base-package interface assertion. Callers holding a `catalog.Catalog` can now assert `rest.FunctionCatalog`.

catalog/catalog.go
rraulinio: The endpoint-negotiation behavior is now documented on the REST-local interface and methods, alongside `rest.ErrEndpointNotSupported`.

catalog/rest/rest.go
rraulinio: Fixed. `LoadFunction` and `CheckFunctionExists` now validate the identifier once and share a private post-validation loader. This removes the duplicate validation while preserving the existing behavior that malformed identifiers return an error and only a server-reported missing function maps to `(false, nil)`.

catalog/rest/rest_functions_test.go
rraulinio: Fixed, the iterator error is now checked before appending the returned identifier.

catalog/rest/rest.go
rraulinio: Fixed. `LoadFunction` now inspects the REST error type and maps `NoSuchNamespaceException` to `ErrNoSuchNamespace`, while `NoSuchFunctionException` remains `ErrNoSuchFunction`. Coverage was added for both `LoadFunction` and `CheckFunctionExists`.

catalog/rest/rest.go
rraulinio: Added, a successful response without the required `metadata` field now returns a clear `ErrRESTError`, with regression coverage.

catalog/rest/rest_functions_test.go
rraulinio: Fixed, the invalid-namespace test now pins `ErrNoSuchNamespace`, and the affected iterator loops stop after receiving the first error.

udf/udf.go
rraulinio: Kept the value receiver and argument to match `view.View` and added documentation making the intentional value-copy semantics and pointer-call usage explicit.

catalog/rest/rest.go
rraulinio: Addressed, I added `catalog.ObjectNameFromIdent`, retained `TableNameFromIdent` as a compatibility alias and updated the table, view and function REST path splitters to use the type-neutral helper.

rraulinio:
@laskoviymishka Addressed, PTAL.

laskoviymishka (approved):
Thanks for pushing this to complition! All of the round-1 points are addressed. 

The `FunctionCatalog` interface and the `udf` dependency now live in `catalog/rest`, so the base package stays leaf-independent; the endpoint-negotiation docs and `ErrEndpointNotSupported` sit together on the REST-local interface. `CheckFunctionExists` validates once through a shared post-validation loader, the list test checks the iterator error before appending, the nil-metadata guard returns a clear `ErrRESTError`, and `ObjectNameFromIdent` cleans up all three path splitters.

I took a closer look at the 404 discrimination since it was the one behavioral change: `loadFunction` now rewraps `NoSuchNamespaceException` as `catalog.ErrNoSuchNamespace` and leaves everything else as `ErrNoSuchFunction`, and `TestLoadFunction404Namespace` pins both the load path and the `CheckFunctionExists` propagation. Looks right.

LGTM, this is good to merge 🚢 👍

PR #1434, opened 2026-07-10
feat(compaction): add CollectDeadPositionDeletes for leader-side expunge

twuebi:
Position-delete cleanup during a rewrite must re-check each delete against the full set of rewritten data files, which the per-group CollectSafePositionDeletes cannot do on its own. CollectDeadPositionDeletes walks the snapshot and returns the position deletes the rewrite makes dead: file-scoped ones when their single target is rewritten, partition-scoped ones when no surviving data file in their partition predates them (deletion vectors stay on the per-group path). The file_path/referenced_data_file resolution it shares with isFileScoped is folded into a single referencedDataFilePath helper.

### DIFF
--- a/table/compaction/compaction.go
+++ b/table/compaction/compaction.go
@@ -345,6 +345,9 @@ func fileScopedDeletedRows(t table.FileScanTask) int64 {
 
 // isFileScoped reports whether a delete file applies to exactly one data file.
 // Equality deletes and partition-scoped positional deletes are not file-scoped.
+// Mirrors Java ContentFileUtil.isFileScoped. Stricter than a nil check on
+// referenced_data_file: an empty-string ref counts as unset and falls through
+// to the file_path bounds.
 func isFileScoped(d iceberg.DataFile) bool {
 	if d.ContentType() == iceberg.EntryContentEqDeletes {
 		return false
--- a/table/compaction/compaction.go
+++ b/table/compaction/compaction.go
@@ -344,23 +344,38 @@ func fileScopedDeletedRows(t table.FileScanTask) int64 {
 }
 
 // isFileScoped reports whether a delete file applies to exactly one data file.
-// Mirrors Java ContentFileUtil.isFileScoped/referencedDataFile: a deletion
-// vector or path-scoped positional delete carries a referenced data file
-// either explicitly or, when referenced_data_file is absent (it is optional in
-// V2 and our own scan planning never sets it — see matchDeletesToData), through
-// equal file_path lower/upper bounds. Equality deletes and partition-scoped
-// positional deletes are not file-scoped.
+// Equality deletes and partition-scoped positional deletes are not file-scoped.
 func isFileScoped(d iceberg.DataFile) bool {
 	if d.ContentType() == iceberg.EntryContentEqDeletes {
 		return false
 	}
-	if d.ReferencedDataFile() != nil {
-		return true
+
+	return referencedDataFilePath(d) != ""
+}
+
+// referencedDataFilePath resolves the single data file a delete targets, or ""
+// when it is partition-scoped (no single target). Mirrors Java
+// ContentFileUtil.referencedDataFile: explicit referenced_data_file first, then
+// equal file_path lower/upper bounds — the bounds fallback matters because
+// referenced_data_file is optional in V2 and our own scan planning never sets
+// it (see matchDeletesToData).
+func referencedDataFilePath(d iceberg.DataFile) string {
+	if ref := d.ReferencedDataFile(); ref != nil && *ref != "" {
+		return *ref
 	}
+
 	lower := d.LowerBoundValues()[filePathFieldID]
 	upper := d.UpperBoundValues()[filePathFieldID]
+	if len(lower) == 0 || !bytes.Equal(lower, upper) {
+		return ""
+	}
 
-	return len(lower) > 0 && bytes.Equal(lower, upper)
+	lit, err := iceberg.LiteralFromBytes(iceberg.PrimitiveTypes.String, lower)
+	if err != nil {
+		return ""
+	}
+
+	return lit.(iceberg.TypedLiteral[string]).Value()
 }
 
 // filePathFieldID is the reserved field ID of the file_path column in a
--- a/table/compaction/compaction.go
+++ b/table/compaction/compaction.go
@@ -374,8 +374,12 @@ func referencedDataFilePath(d iceberg.DataFile) string {
 	if err != nil {
 		return ""
 	}
+	s, ok := lit.(iceberg.TypedLiteral[string])
+	if !ok {
+		return ""
+	}
 
-	return lit.(iceberg.TypedLiteral[string]).Value()
+	return s.Value()
 }
 
 // filePathFieldID is the reserved field ID of the file_path column in a
--- a/table/compaction/pos_delete_collect.go
+++ b/table/compaction/pos_delete_collect.go
@@ -1,0 +1,217 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package compaction
+
+import (
+	"context"
+	"slices"
+
+	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
+	"github.com/apache/iceberg-go/table"
+)
+
+// CollectDeadPositionDeletes walks the given snapshot's manifests and returns
+// the position-delete files made dead by a rewrite that removes rewrittenPaths.
+// A dead position delete references only data files the rewrite is removing, so
+// after the rewrite it applies to nothing and is safe to expunge in the same
+// commit.
+//
+// Scope resolution mirrors Java ContentFileUtil.referencedDataFile and
+// DeleteFileIndex:
+//
+//   - File-scoped — explicit referenced_data_file, or equal file_path
+//     lower/upper bounds — is dead iff its single target is in rewrittenPaths.
+//   - Partition-scoped — no single target — is dead iff no surviving
+//     (non-rewritten) data file in the same (specID, partition) has a sequence
+//     number <= the delete's. Such a survivor is one the delete still applies
+//     to, so the delete must be retained for a later dangling-delete pass.
+//
+// Deletion vectors (Puffin position deletes) are intentionally excluded: they
+// are 1:1 with their data file and expunged per-group via
+// [table.CompactionGroupResult.SafeDeletionVectors].
+//
+// rewrittenPaths is the union of every old data file path being replaced across
+// all rewrite groups. Like [CollectDeadEqualityDeletes], the returned files are
+// safe to remove in the same commit that stages the rewrite: a concurrent
+// commit cannot resurrect them, because any concurrent delete touching a
+// rewritten file is rejected by the rewrite validator and any concurrent data
+// file gets a sequence number strictly greater than every preexisting delete.
+//
+// Two-pass design: the first pass walks delete manifests to resolve candidates,
+// deciding file-scoped ones immediately; the more expensive data-manifest walk
+// runs only when a partition-scoped candidate needs a survivor check.
+func CollectDeadPositionDeletes(
+	ctx context.Context,
+	fs iceio.IO,
+	snap *table.Snapshot,
+	rewrittenPaths map[string]struct{},
+) ([]iceberg.DataFile, error) {
+	if snap == nil || len(rewrittenPaths) == 0 {
+		return nil, nil
+	}
+
+	manifests, err := snap.Manifests(fs)
+	if err != nil {
+		return nil, err
+	}
+
+	// First pass: delete manifests, resolving each candidate's expunge scope.
+	candidates, err := collectPositionDeleteCandidates(ctx, fs, manifests)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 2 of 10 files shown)


twuebi:
@laskoviymishka

table/compaction/pos_delete_collect_ext_test.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// TestCollectDeadPositionDeletes drives the full manifest walk: a
+// partition-scoped position delete (no referenced_data_file, no file_path
+// bounds — Spark's default granularity) covering three data files is dead only
+// when every file it covers is rewritten, and retained the moment one is not.
+func TestCollectDeadPositionDeletes(t *testing.T) {
+	ctx := t.Context()
+	fs := iceio.LocalFS{}
+	tbl := newCDCStressTable(t)
```
zeroshade, line 43: nit (test coverage): `newCDCStressTable` uses `UnpartitionedSpec`, so every data file and the delete collapse to the same `partitionBucketKey` (`"0:_"`). That means the partition-scoped survivor check — the subtlest part of `minSurvivorSeqByPartition` / `decideDeadPositionDeletes`, and the reason position deletes key on `(specID, partition)` — is never exercised end-to-end here. (`TestDecideDeadPositionDeletes` covers the decision with hand-built keys, but not key derivation or cross-partition exclusion.)

Consider a partitioned-table case: a partition-scoped delete in partition A that stays alive while a same-partition survivor with `seq <= delete.seq` exists, and becomes dead once the only survivors live in a different partition. That's the behavior that distinguishes this from a naive implementation, so it's worth a regression guard.

table/compaction/compaction.go
```diff
@@ -344,23 +344,38 @@ func fileScopedDeletedRows(t table.FileScanTask) int64 {
...
+		return ""
+	}
+
+	lit, err := iceberg.LiteralFromBytes(iceberg.PrimitiveTypes.String, lower)
+	if err != nil {
+		return ""
+	}
 
-	return len(lower) > 0 && bytes.Equal(lower, upper)
+	return lit.(iceberg.TypedLiteral[string]).Value()
```
zeroshade: nit (defensiveness): this assertion panics if `lit` isn't a string-typed literal. It's safe today — `LiteralFromBytes(PrimitiveTypes.String, …)` always returns one — but since it decodes arbitrary `file_path` bounds from file metadata, the comma-ok form is more robust and consistent with the `err` check just above:

```go
s, ok := lit.(iceberg.TypedLiteral[string])
if !ok {
	return ""
}

return s.Value()
```

table/compaction/pos_delete_collect.go
```diff
@@ -0,0 +1,217 @@
...
+// all rewrite groups. Like [CollectDeadEqualityDeletes], the returned files are
+// safe to remove in the same commit that stages the rewrite: a concurrent
+// commit cannot resurrect them, because any concurrent delete touching a
+// rewritten file is rejected by the rewrite validator and any concurrent data
+// file gets a sequence number strictly greater than every preexisting delete.
+//
+// Two-pass design: the first pass walks delete manifests to resolve candidates,
+// deciding file-scoped ones immediately; the more expensive data-manifest walk
+// runs only when a partition-scoped candidate needs a survivor check.
+func CollectDeadPositionDeletes(
```
zeroshade, line 62: question (not blocking): `CollectDeadPositionDeletes` is exported but has no in-repo caller yet, unlike `CollectDeadEqualityDeletes` which `cmd/iceberg/compact.go` feeds into `RewriteDataFilesOptions.ExtraDeleteFilesToRemove`. The title says "for leader-side expunge" — is that integration a planned follow-up, or is this meant purely as an API for external coordinators (like `CollectSafePositionDeletes`)? Either is fine; a one-line note on the intended consumer in the doc comment would help readers.

table/compaction/compaction.go
```diff
@@ -344,23 +344,38 @@ func fileScopedDeletedRows(t table.FileScanTask) int64 {
...
+}
+
+// referencedDataFilePath resolves the single data file a delete targets, or ""
+// when it is partition-scoped (no single target). Mirrors Java
+// ContentFileUtil.referencedDataFile: explicit referenced_data_file first, then
+// equal file_path lower/upper bounds — the bounds fallback matters because
+// referenced_data_file is optional in V2 and our own scan planning never sets
+// it (see matchDeletesToData).
+func referencedDataFilePath(d iceberg.DataFile) string {
+	if ref := d.ReferencedDataFile(); ref != nil && *ref != "" {
```
zeroshade, line 366: FYI (no action needed): folding `*ref != ""` in here slightly changes `isFileScoped` — a non-nil-but-empty `referenced_data_file` previously counted as file-scoped and now falls through to the `file_path` bounds. This is more correct, only affects the delete-ratio heuristic, and is captured by the new `TestReferencedDataFilePath` "empty referenced_data_file falls through to bounds" case. Noting for the record.

zeroshade (commented):
A few non-blocking nits, posted inline.

Overall this looks correct. I verified the `(specID, partition)` keying for partition-scoped position deletes against Java `DeleteFileIndex` and `format/spec.md`: position deletes are never global (only equality deletes route an unpartitioned spec into `globalDeletes`), and matching requires both spec id and partition values to be equal. So keying on `partitionBucketKey(specID, partition)` here — rather than reusing the eq-delete's global/empty-partition handling — is the right call.

table/compaction/pos_delete_collect_ext_test.go
twuebi: added tests

table/compaction/pos_delete_collect.go
twuebi: Meant as an API for external coordinators, like `CollectSafePositionDeletes`

table/compaction/compaction.go
```diff
@@ -344,23 +344,42 @@ func fileScopedDeletedRows(t table.FileScanTask) int64 {
...
-// positional deletes are not file-scoped.
+// Equality deletes and partition-scoped positional deletes are not file-scoped.
 func isFileScoped(d iceberg.DataFile) bool {
 	if d.ContentType() == iceberg.EntryContentEqDeletes {
 		return false
 	}
-	if d.ReferencedDataFile() != nil {
-		return true
+
+	return referencedDataFilePath(d) != ""
```
laskoviymishka, line 356: heads up that this quietly changes `isFileScoped` for a non-nil empty-string `referenced_data_file`: the old path returned true on any non-nil ref, the new one returns false because `referencedDataFilePath` treats `""` as unset. That's the more correct answer, and our builder rejects empty refs at build anyway, so it only bites foreign writers.

But it's silent, and the Java-parity paragraph that used to explain the bounds fallback is gone from `isFileScoped` entirely. I'd keep the change, move a line of that explanation back onto `isFileScoped`, and add a direct `isFileScoped` case for an empty-string ref (expect false) so the tightening is pinned. wdyt?

table/compaction/pos_delete_collect.go
```diff
@@ -0,0 +1,221 @@
...
+	// First pass: delete manifests, resolving each candidate's expunge scope.
+	candidates, err := collectPositionDeleteCandidates(ctx, fs, manifests)
+	if err != nil {
+		return nil, err
+	}
+
+	// Second pass runs only when a partition-scoped candidate needs a survivor
+	// check; file-scoped ones are decided from rewrittenPaths alone.
+	var minSurvivorSeq map[string]int64
+	if slices.ContainsFunc(candidates, func(c positionDeleteCandidate) bool { return c.fileScopedTarget == "" }) {
```
laskoviymishka: small one — `collectPositionDeleteCandidates` already walks every candidate, so it could hand back a `hasPartitionScoped` bool and save this re-scan (and drop the `slices` import). Not important, just while we're here.

table/compaction/pos_delete_collect.go
```diff
@@ -0,0 +1,221 @@
...
+				return nil, err
+			}
+			df := e.DataFile()
+			if df.ContentType() != iceberg.EntryContentData {
+				continue
+			}
+			if _, rewritten := rewrittenPaths[df.FilePath()]; rewritten {
+				continue
+			}
+			seq := max(e.SequenceNum(), 0)
```
laskoviymishka, line 184: the two passes treat the negative-sequence sentinel differently: the delete side drops candidates with `seq < 0` and has a comment explaining why, while this survivor side silently clamps it to 0 with `max()`. The net effect is correct, but a reader landing on one side without the other could easily read it as a bug. Could we mirror the delete-side comment here so the two sentinel choices read as deliberate?

table/compaction/pos_delete_collect.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+	}
+
+	return dead
+}
+
+// isDeletionVector reports whether df is a deletion vector: a Puffin file with
+// position-delete content. Mirrors table.isDeletionVector, reimplemented here
+// because that predicate is unexported.
+func isDeletionVector(df iceberg.DataFile) bool {
```
laskoviymishka: this is now the third byte-for-byte copy of `table.isDeletionVector` (`scanner.go` and `mor_delete_pruning_test.go` have the others). Since compaction is the second real consumer, I'd lean toward exporting it from the `table` package and dropping the copies. If we'd rather not widen that surface, a `//nolint:dupl` with a TODO would at least mark it as known. wdyt?

table/compaction/pos_delete_collect_ext_test.go
```diff
@@ -0,0 +1,194 @@
...
+	require.NoError(t, rd.Commit(ctx))
+	tbl, err = tx.Commit(ctx)
+	require.NoError(t, err)
+
+	t.Run("all covered files rewritten is dead", func(t *testing.T) {
+		rewritten := pathSet(dataPaths...)
+		dead, err := compaction.CollectDeadPositionDeletes(ctx, fs, tbl.CurrentSnapshot(), rewritten)
+		require.NoError(t, err)
+		require.Len(t, dead, 1, "a partition-scoped delete whose every covered file is rewritten must be expunged")
+	})
```
laskoviymishka, line 80: this asserts the right files come back, but nothing here commits the expunge through `RewriteDataFiles` and re-scans — so the actual integration path (Collect → `ExtraDeleteFilesToRemove` → commit → scan) is never exercised. The eq-delete side has `TestCDCStress` doing exactly that round-trip.

I'd add one test that routes the collected files through `ExtraDeleteFilesToRemove`, commits, and re-scans to prove no deleted row reappears — it'd also immediately catch the `RemovedEqualityDeleteFiles` miscount I raised in the review. Fine to gate under `-short` if it's heavy.

table/compaction/pos_delete_collect_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestReferencedDataFilePath(t *testing.T) {
+	filePathID := filePathFieldID
+
+	t.Run("explicit referenced_data_file", func(t *testing.T) {
+		df := newPosDelete(t, "del-a.parquet").ReferencedDataFile("data-a.parquet").Build()
+		assert.Equal(t, "data-a.parquet", referencedDataFilePath(df))
```
laskoviymishka: nit — these use `assert.Equal` while the rest of both files use `require`. Just for consistency I'd switch to `require.Equal` here.

laskoviymishka (changes requested):
This is a clean, well-reasoned PR. The two-pass design mirrors `CollectDeadEqualityDeletes`, the docstring on `CollectDeadPositionDeletes` actually explains the scope resolution well, and the `(specID, partition)` survivor keying matches Java's `DeleteFileIndex` — the partitioned test covering cross-partition and newer-sequence survivors is exactly the case I'd want pinned.

I'd hold this before merging though, on two things that turn out to be linked.

The docstring says the collected files feed `RewriteDataFilesOptions.ExtraDeleteFilesToRemove`, but that consumer (`rewrite_data_files.go` around line 261) increments `result.RemovedEqualityDeleteFiles` unconditionally for every entry in that slice. So dead *position* deletes routed there get counted as equality deletes — `RemovedEqualityDeleteFiles` is inflated and `RemovedPositionDeleteFiles` never reflects the expunge this PR exists to do. It's a pre-existing bug, but this PR's intended usage is what activates it for the first time. I'd discriminate by content type in that loop (pos-delete entries → `RemovedPositionDeleteFiles`) before wiring this up. The `ExtraDeleteFilesToRemove` docstring at `rewrite_data_files.go:148-156` also becomes misleading in the same breath — it currently says position deletes "are removed automatically and do NOT need to be passed in here," which is no longer the whole story.

Relatedly, neither external test commits the expunge through `RewriteDataFiles` and re-scans — they only assert the returned file set. The eq-delete path has `TestCDCStress` doing the full round-trip, and a mis-expunge here is silent data loss, so I'd really like the same coverage. A round-trip test would also catch the metric miscount above immediately.

- discriminate by content type in the `ExtraDeleteFilesToRemove` removal loop so dead position deletes land in `RemovedPositionDeleteFiles`
- a round-trip test (Collect → `ExtraDeleteFilesToRemove` → `RewriteDataFiles` → re-scan), can gate under `-short`
- pin the `isFileScoped` empty-string behavior change with a doc line and a direct test case
- mirror the sentinel-handling comment onto the survivor side so the two passes read as deliberate

The dedupe of `isDeletionVector` and the `assert`/`require` consistency are just nits. The core logic is right and the keying matches Java — once the metric routing and a round-trip test are in, happy to take another pass and approve.

table/compaction/pos_delete_collect.go
twuebi: went for the export

table/compaction/pos_delete_collect_ext_test.go
twuebi: added the round trip, that was a good call, it surfaced a bug, the commit failed because per-group staging already removes task-attached pos deletes and the collector's output overlaps it by construction. RewriteDataFiles now skips already-staged files and routes metrics by content type. Runs fast enough, no gate needed.

laskoviymishka (approved):
All four things I flagged last round are in, and the round-trip test earned its keep - it surfaced the staged/collected overlap, and the stagedDeleteFiles dedup handles it cleanly (removed and counted once). 

Metrics now route by content type, so a dead position delete lands in RemovedPositionDeleteFiles instead of being miscounted as an equality delete, and TestDeadPositionDeleteRoundTrip pins the whole path end-to-end: id 2 stays deleted and no delete file survives the rewrite. 

Nice bonus exporting IsDeletionVector and dropping the copies.

The isFileScoped empty-string tightening is documented and the referencedDataFilePath cases cover it: a direct `isFileScoped(emptyRef) == false` case would've been the last mile, but it's pinned closely enough that I won't hold on it.

LGTM, approving.

PR #1435, opened 2026-07-10
test(table): guard MetadataBuilder.clone against field drift

laskoviymishka:
MetadataBuilder.clone copies its fields by hand, so a field added to the struct but not to clone() is silently shared between builders. Add a reflection-based test that fills every field and asserts the clone is value-equal yet independently backed, catching both a dropped field and one copied by reference.

Also cover the requirements half of apply's atomicity (a mid-batch update failure must not leak staged requirements), which no existing test exercised, and rename the generic pointer-clone helper from cloneMetadataBuilderInt to clonePtr since it is not int-specific.

### DIFF
--- a/table/metadata_builder_internal_test.go
+++ b/table/metadata_builder_internal_test.go
@@ -2576,9 +2576,11 @@ var sharedCloneFields = map[string]struct{}{
 // it fills every field of MetadataBuilder with non-zero data and asserts the
 // clone is value-equal yet backed by independent storage. A field added to the
 // struct but omitted from clone() fails the value-equal check; a field copied
-// by reference instead of cloned fails the aliasing check. The filler fails the
-// test on any kind it cannot populate, so a new field of an unhandled type
-// cannot slip through as a vacuous pass.
+// by reference instead of cloned fails the aliasing check. The filler recurses
+// into pointers, slices, and maps to give their targets non-zero contents, so a
+// clone() that allocates fresh storage but forgets to copy the underlying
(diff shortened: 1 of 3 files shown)


zeroshade (commented):
Nice guard against clone() field drift. One non-blocking suggestion to make it stronger (inline).

table/metadata_builder_internal_test.go
```diff
@@ -2562,3 +2564,114 @@ func TestSetFormatVersionV2ToV3FromDeserializedMetadata(t *testing.T) {
...
+// test on any kind it cannot populate, so a new field of an unhandled type
+// cannot slip through as a vacuous pass.
+//
+// The aliasing check is one level deep: it verifies each slice/map/pointer
+// field has its own backing storage, not that reference types nested inside
+// elements are independent (clone() shares those by design). A future field
+// nesting references inside an element would need its own dedicated coverage.
+func TestMetadataBuilderCloneCoversAllFields(t *testing.T) {
+	orig := &MetadataBuilder{}
+	fillReferenceFields(t, reflect.ValueOf(orig).Elem())
```
zeroshade, line 2591: Non-blocking: the reference filler leaves pointer targets, slice elements, and map entries at their zero values, so a broken `clone()` that allocates fresh storage but fails to copy the pointee/element *values* could still pass the aliasing check. Populating nested pointers / slice elements / map keys+values with non-zero data (e.g. a small recursive filler for the supported kinds) would close that gap.

zeroshade (approved):
LGTM, thanks!

PR #1436, opened 2026-07-10
fix(hadoop): remove table directory after purging files

fallintoplace:
Hadoop `PurgeTable` deleted the current metadata file through `Table.PurgeFiles` and then called `DropTable`. Because `DropTable` uses loadable metadata to recognize a table, a successful file purge returned `ErrNoSuchTable` and left the table directory behind.

`PurgeTable` now removes the already-validated table path directly after best-effort file cleanup instead of checking the deleted metadata again. Regression coverage creates a table with an externally referenced data file and verifies that purge succeeds, removes the table directory, makes `CheckTableExists` return false, and deletes the external file.

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -881,8 +881,10 @@ func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) e
 		log.Printf("WARNING: failing to purge some files in Hadoop table %s: %v", identifier, purgeErr)
 	}
 
-	// Delete the table directory root from the local storage
-	return c.DropTable(ctx, identifier)
+	// PurgeFiles removes the metadata files that DropTable uses to recognize a
+	// table. The table was already validated by LoadTable above, so remove its
+	// directory directly instead of checking it again after the purge.
+	return c.filesystem.RemoveAll(c.tableToPath(identifier))
 }
 
 func (c *Catalog) RenameTable(_ context.Context, _, _ table.Identifier) (*table.Table, error) {
--- a/catalog/hadoop/hadoop_test.go
+++ b/catalog/hadoop/hadoop_test.go
@@ -1437,6 +1437,47 @@ func (s *HadoopCatalogTestSuite) TestDropTableVerifyCleanup() {
 	s.True(os.IsNotExist(statErr))
 }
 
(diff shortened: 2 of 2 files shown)


catalog/hadoop/hadoop.go
```diff
@@ -881,8 +881,10 @@ func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) e
 		log.Printf("WARNING: failing to purge some files in Hadoop table %s: %v", identifier, purgeErr)
 	}
 
-	// Delete the table directory root from the local storage
-	return c.DropTable(ctx, identifier)
+	// PurgeFiles removes the metadata files that DropTable uses to recognize a
+	// table. The table was already validated by LoadTable above, so remove its
+	// directory directly instead of checking it again after the purge.
+	return c.filesystem.RemoveAll(c.tableToPath(identifier))
```
zeroshade, line 887: This `RemoveAll` unconditionally removes the entire table directory (data files included) after `PurgeFiles`. But `PurgeFiles` intentionally skips data-file deletion when `gc.enabled=false` (see `table/orphan_cleanup.go`), so this regresses that contract — the SQL catalog preserves data in that case (cf. the `gc.enabled=false` test in `catalog/sql/sql_test.go`, and the `PurgeableTable` docs in `catalog/catalog.go`).

Please gate the `RemoveAll` on the loaded table's `gc.enabled` property (when GC is disabled, only remove metadata/empty dirs, not data), and add a Hadoop `PurgeTable` test with `gc.enabled=false` that asserts referenced data files survive.

zeroshade (changes requested):
Thanks for the cleanup here — avoiding the extra metadata round-trip is a nice simplification. One correctness concern before this can go in (inline): `PurgeTable` will now delete data files even when `gc.enabled=false`.

zeroshade:
Heads-up while triaging reviews: #1449 also rewrites the Hadoop `PurgeTable` cleanup, and it additionally gates the final `RemoveAll` on the table's `gc.enabled` property — i.e. it addresses the `gc.enabled=false` data-deletion concern raised in the review here, and adds a `gc.enabled=false` preservation test. The two PRs overlap on the same lines, so they can't both merge as-is. Suggest consolidating on #1449 and closing this (or pulling the `gc.enabled` gate into this PR).

PR #1437, opened 2026-07-10
fix(table): honor URI prefix mismatch modes in orphan cleanup

fallintoplace:
## Summary

`DeleteOrphanFiles` compared full normalized URIs before checking scheme and authority mismatches. A listed file with the same object path as a reference but a non-equivalent prefix could therefore fall through as an orphan.

Orphan detection now indexes both normalized URIs and path components. Exact and configured-equivalent matches remain referenced; same-path prefix conflicts follow the configured policy:

- `ERROR` stops cleanup;
- `IGNORE` keeps the file;
- `DELETE` treats it as an orphan.

Windows drive letters remain part of path comparison keys.

## Testing

- `go test ./table`
- all three modes across scheme and authority mismatches
- Windows cross-drive regression coverage

### DIFF
--- a/table/orphan_cleanup.go
+++ b/table/orphan_cleanup.go
@@ -723,9 +723,13 @@ func normalizeNonURLPath(path string) string {
 // filePathKey returns the path component used to compare listed files with
 // references before applying scheme and authority mismatch policy.
 func filePathKey(file string) string {
-	parsedURL, err := url.Parse(file)
-	if err == nil && (parsedURL.Scheme != "" || parsedURL.Host != "") {
-		return normalizeNonURLPath(parsedURL.Path)
+	// A bare Windows path such as C:/data/file.parquet is parsed as a URL
+	// with scheme "c". Only parse URL-shaped values here so drive letters
+	// remain part of the comparison key.
+	if strings.Contains(file, "://") || strings.HasPrefix(strings.ToLower(file), "file:") {
+		if parsedURL, err := url.Parse(file); err == nil {
+			return normalizeNonURLPath(parsedURL.Path)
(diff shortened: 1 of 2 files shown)


table/orphan_cleanup.go
```diff
@@ -664,6 +698,17 @@ func normalizeNonURLPath(path string) string {
 	return strings.ReplaceAll(normalized, "\\", "/")
 }
 
+// filePathKey returns the path component used to compare listed files with
+// references before applying scheme and authority mismatch policy.
+func filePathKey(file string) string {
+	parsedURL, err := url.Parse(file)
```
zeroshade: `url.Parse` on a bare Windows path like `C:/data/file.parquet` interprets `C` as the scheme, so this returns only `parsedURL.Path` and drops the drive letter — `C:/data/file.parquet` and `D:/data/file.parquet` then map to the same key `/data/file.parquet`. With `PrefixMismatchIgnore` a wrong-drive file can be wrongly treated as referenced; with the default error mode, unrelated drive-local paths can trigger errors. Suggest only URL-parsing strings that contain `://` (or an explicit `file:` scheme) and otherwise returning `normalizeNonURLPath(file)`, plus a Windows drive-letter test case.

zeroshade (changes requested):
The path-key index for prefix-mismatch handling is a good approach. One Windows path edge case to fix (inline).

zeroshade (approved):
Re-reviewed: bare and file: Windows drive paths are now preserved (drive-letter interception in normalizeFilePathWithConfig), addressing my earlier note. Thanks @fallintoplace!

PR #1438, opened 2026-07-10
fix(catalog): validate requirements for missing tables

fallintoplace:
`UpdateAndStageTable` validated commit requirements only when the table already existed. For a missing table it constructed synthetic metadata and applied creation updates without calling `Validate(nil)`, allowing requirements such as table UUID, schema ID, or snapshot reference assertions to be bypassed across the SQL, Glue, and Hive catalogs.

Requirements are now validated before the helper chooses its staging metadata: existing tables validate against current metadata, while missing tables validate against nil. Regression coverage verifies that `AssertTableUUID` rejects an otherwise valid creation commit without writing metadata, while `AssertCreate` still stages the same creation updates successfully.

### DIFF
--- a/catalog/internal/utils.go
+++ b/catalog/internal/utils.go
@@ -202,15 +202,17 @@ func UpdateAndStageTable(ctx context.Context, catprops iceberg.Properties, curre
 	)
 
 	if current != nil {
-		for _, r := range reqs {
-			if err := r.Validate(current.Metadata()); err != nil {
-				return nil, err
-			}
-		}
-
 		baseMeta = current.Metadata()
 		metadataLoc = current.MetadataLocation()
-	} else {
+	}
+
+	for _, r := range reqs {
+		if err := r.Validate(baseMeta); err != nil {
+			return nil, err
+		}
+	}
+
+	if current == nil {
 		var err error
 		baseMeta, err = table.NewMetadata(iceberg.NewSchema(0), nil, table.UnsortedSortOrder, "", nil)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


catalog/internal/utils.go
```diff
@@ -202,15 +202,17 @@ func UpdateAndStageTable(ctx context.Context, catprops iceberg.Properties, curre
...
-			}
-		}
-
 		baseMeta = current.Metadata()
 		metadataLoc = current.MetadataLocation()
-	} else {
+	}
+
+	for _, r := range reqs {
+		if err := r.Validate(baseMeta); err != nil {
```
zeroshade, line 210: This correctly validates every requirement against `baseMeta` even when the table is missing (baseMeta nil) — 👍.

However, `catalog/hadoop/hadoop.go`'s `CommitTable` doesn't use `UpdateAndStageTable`; it still validates requirements only inside `if current != nil` (around `hadoop.go:614`). So a commit carrying e.g. `AssertTableUUID` against a missing table bypasses validation there and proceeds as a create — the same bug this PR fixes elsewhere. Please apply the same ordering in the Hadoop path (validate all reqs against `baseMeta`, which may be nil, before creating empty base metadata) and add a Hadoop regression test for a non-create requirement on a missing table.

zeroshade (changes requested):
The fix to `UpdateAndStageTable` is correct and well-tested. Blocking on one gap: the Hadoop catalog doesn't go through that helper, so its `CommitTable` still has the original bug (inline).

zeroshade:
Heads-up while triaging reviews: #1451 includes this same `UpdateAndStageTable` change and additionally fixes the Hadoop `CommitTable` path to validate requirements for missing tables — the gap raised in the review here — plus adds missing-table requirement tests across Glue/Hadoop/Hive/SQL (verified passing locally). The two PRs overlap on the same lines. Suggest consolidating on #1451 and closing this.

PR #1439, opened 2026-07-10
fix(sql): distinguish dotted namespace components

fallintoplace:
## Summary

- preserve the existing dot-joined representation for ordinary SQL namespaces
- use a versioned JSON representation only when a new namespace component contains a dot or would collide with the encoding marker
- resolve namespace properties from the original identifier components while using the encoded key for distinct default table and view locations
- decode namespaces consistently across tables, views, properties, existence checks, renames, and drops
- keep parent listing bounded in SQL for both legacy and encoded descendants, with escaped `LIKE` wildcards, then verify component ancestry after decoding

## Compatibility policy

Existing rows are not rewritten. A legacy value such as `a.b` keeps the catalog historical interpretation as the two-component namespace `[a, b]`; the catalog does not guess whether an old ambiguous row was originally intended as `[a.b]`. New dotted components use the reversible encoding, so `[a.b]` and `[a, b]` can coexist from this release forward without changing ordinary stored keys. Legacy rows containing the encoding marker fall back to their historical key when the suffix is not a canonical encoding.

## Testing

The regressions create the two previously colliding namespace shapes and verify their properties, default locations, and tables remain separate through create, list, load, rename, and drop. Dotted table and view creation also verifies namespace location properties are resolved from the original components. A directly inserted legacy marker-prefixed row remains usable through listing, property updates, table creation/load, and cleanup. Parent listing covers literal `%`, `_`, and the SQL escape character, including a namespace that would match an unescaped pattern.

- 20 repeated focused namespace test runs
- `go test ./catalog/sql ./catalog/hive ./catalog/glue`
- repository package tests excluding the locally hanging `io/gocloud` package
- `go vet ./catalog/sql ./catalog/hive ./catalog/glue ./catalog/internal`

### DIFF
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -20,6 +20,7 @@ package sql
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"iter"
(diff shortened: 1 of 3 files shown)


catalog/sql/sql.go
```diff
@@ -1009,9 +1042,15 @@ func (c *Catalog) ListNamespaces(ctx context.Context, parent table.Identifier) (
...
-	ret := make([]table.Identifier, len(namespaces))
-	for i, n := range namespaces {
-		ret[i] = strings.Split(n, ".")
+	ret := make([]table.Identifier, 0, len(namespaces))
+	for _, n := range namespaces {
+		ident, err := namespaceFromString(n)
+		if err != nil {
+			return nil, err
+		}
+		if len(parent) == 0 || (len(ident) >= len(parent) && slices.Equal(ident[:len(parent)], parent)) {
```
zeroshade, line 1150: Non-blocking: `ListNamespaces` now drops the SQL `LIKE 'parent%'` narrowing and instead fetches every namespace for the catalog and filters by parent in Go. That's understandable given the new encoding (a JSON-encoded value won't match a dot-prefix LIKE), but for catalogs with many namespaces it becomes a fetch-all + client-side scan. If that turns into a hotspot, an encoded/indexed prefix column or a bounded query could restore server-side filtering.

zeroshade (commented):
The JSON-encoding approach to keep dotted namespace components distinct from hierarchical ones is reasonable, and the round-trip test is a good guard. One performance nit (inline).

zeroshade (changes requested):
A couple of correctness issues remain around the new encoded-namespace handling:

**1. (High) Dotted-namespace table/view location still resolves through dot-joined strings**

`CreateTable` calls `internal.CreateStagedTable(ctx, ..., ident, ...)` (`catalog/sql/sql.go:564`) before the namespace is encoded. `CreateStagedTable` joins the namespace with `"."` (`catalog/internal/utils.go:105`) and `ResolveTableLocation` splits it back on `"."` (`utils.go:148`). For a namespace like `["a.b"]` this resolves properties for `["a","b"]`, so:

- if the hierarchical namespace `a.b` doesn't exist, table creation fails; and
- if it does exist, the dotted namespace inherits the wrong namespace properties and can collide on the same default table location.

`CreateView` has the same issue via `ResolveTableLocation(ctx, "", ns, ...)` (`sql.go:1193`), where `ns` is the dot-joined string.

Please resolve the location from the original `table.Identifier` namespace components rather than a dot-joined string, and ensure the default-location key is distinct for dotted components.

**2. (Medium) Legacy rows beginning with the reserved prefix fail to decode after upgrade**

`namespaceFromString` (`sql.go:80`) treats any value starting with `__iceberg_namespace_v1__:` as JSON and errors if it isn't (`sql.go:86-87`). A pre-existing catalog row whose namespace legitimately began with that literal string (previously permitted) would now fail to list/load after upgrade. Consider falling back to the legacy interpretation on JSON-decode failure, or documenting a migration path.

zeroshade (approved):
Re-reviewed: the dotted-namespace location resolution and legacy-decode concerns are addressed (component-aware property lookup via ResolveTableLocationWithNamespace, distinct encoded storage key, and legacy fallback in namespaceFromString). Thanks @fallintoplace!

PR #1441, opened 2026-07-10
fix(catalog): reject empty table identifiers

fallintoplace:
## Summary

`NamespaceFromIdent` sliced with an upper bound of -1 for an empty identifier, allowing invalid public API input to panic. The shared staging helper also derived the namespace before validating the table identifier.

`NamespaceFromIdent` now returns nil for empty input, and table creation validates before resolving locations or writing metadata. SQL table operations validate at their public boundaries, while Glue and Hive identifier conversion use the shared typed validators. Hive view operations retain `ErrNoSuchView` for malformed view identifiers instead of leaking the table-specific error from their shared shape parser.

Existing error conventions are preserved: missing namespaces during creation return `ErrNoSuchNamespace`; table lookup and mutation operations return `ErrNoSuchTable`; view operations return `ErrNoSuchView`.

## Testing

Regression coverage passes nil, empty, and one-component identifiers through create, load, commit, drop, purge, rename, and existence checks for SQL, Glue, and Hive. Hive create/load/drop/exists view operations also verify the view-specific typed error. Backend mocks receive no calls for rejected identifiers.

- `go test ./catalog/...`
- `go test ./...`

### DIFF
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -897,8 +897,11 @@ func (c *Catalog) getIcebergTable(ctx context.Context, database, tableName strin
 }
 
 func identifierToTableName(identifier table.Identifier) (string, string, error) {
+	if err := catalog.ValidateTableIdentifier(identifier); err != nil {
+		return "", "", err
+	}
 	if len(identifier) != 2 {
-		return "", "", fmt.Errorf("invalid identifier, expected [database, table]: %v", identifier)
+		return "", "", fmt.Errorf("%w: expected [database, table], got %v", catalog.ErrNoSuchTable, identifier)
 	}
 
 	return identifier[0], identifier[1], nil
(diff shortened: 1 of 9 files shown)


catalog/hive/hive.go
```diff
@@ -852,8 +852,11 @@ func (c *Catalog) getIcebergTable(ctx context.Context, database, tableName strin
 }
 
 func identifierToTableName(identifier table.Identifier) (string, string, error) {
+	if err := catalog.ValidateTableIdentifier(identifier); err != nil {
```
zeroshade: `identifierToTableName` now returns `ErrNoSuchTable` on an invalid identifier, but this helper is also used by the Hive view operations (e.g. `LoadView`/`DropView`/`CheckViewExists` at lines ~257 and ~574). Those callers — and their tests — branch on `ErrNoSuchView`, so an invalid *view* identifier will now surface `ErrNoSuchTable` instead. Consider splitting into a view variant (using `ValidateViewIdentifier` and returning `ErrNoSuchView`), or parameterizing the not-found error passed into the validation path.

zeroshade (changes requested):
Rejecting empty identifiers is a good addition. One issue for view operations before merge (inline).

zeroshade (approved):
Re-reviewed: view operations now resolve via identifierToViewName and return ErrNoSuchView, addressing my earlier note. Thanks @fallintoplace!

PR #1442, opened 2026-07-10
fix(hive): lock table drops and renames

fallintoplace:
Hive commits use exclusive HMS table locks, but drop and rename previously bypassed that protocol and could race with metadata publication.

Drop now acquires the table lock and re-reads the table while holding it before deleting the catalog entry. Rename requests source and destination locks atomically in lexical order, then re-reads the source and verifies that its metadata location has not changed before altering the descriptor. Lock acquisition failures and rename revalidation conflicts wrap `table.ErrCommitFailed` so callers can retry safely.

The lock helper now supports sorted, deduplicated multi-table requests while retaining the single-table entry point used by commits. Tests cover component ordering, retryable drop conflicts, rename metadata revalidation, and lock participation in drop and purge paths.

Tests:
- `go test ./...`

### DIFF
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -373,6 +373,16 @@ func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) er
 		return err
 	}
 
+	lock, err := acquireLock(ctx, c.client, database, tableName, c.opts)
+	if err != nil {
+		return fmt.Errorf("%w: failed to acquire lock for %s.%s: %w", table.ErrCommitFailed, database, tableName, err)
+	}
+	defer func() {
+		_ = lock.Release(ctx)
+	}()
+
+	// Re-read after acquiring the lock so drop cannot act on table state that a
+	// concurrent commit changed while lock acquisition was waiting.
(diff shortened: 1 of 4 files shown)


catalog/hive/hive.go
```diff
@@ -373,6 +373,16 @@ func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) er
 		return err
 	}
 
+	lock, err := acquireLock(ctx, c.client, database, tableName, c.opts)
```
zeroshade, line 376: Micro-nit: the exclusive lock is acquired before confirming the table exists, so dropping a nonexistent table still takes and releases a lock (as the updated `TestHiveDropTableNotExists` reflects). Harmless — and lock-before-read is arguably safer for concurrency — just flagging in case you'd prefer to short-circuit obviously-missing tables before locking.

zeroshade (commented):
Locking approach looks solid — sorted multi-locks for deadlock avoidance and the metadata-location recheck after acquiring the lock on rename is a good optimistic-concurrency guard. One micro-nit (inline).

zeroshade (approved):
LGTM, thanks!

PR #1443, opened 2026-07-10
fix: validate retention durations

fallintoplace:
Reject invalid retention and cleanup inputs before any scan or metadata mutation.

This covers negative ages, invalid retain-last values, non-finite day durations, and time.Duration overflow boundaries. Zero age remains supported where the library contract allows it.

Tests: go test ./table ./cmd/iceberg; go test -short ./... (io/gocloud hangs in existing tests).


table/orphan_cleanup.go
```diff
@@ -95,6 +96,9 @@ func WithLocation(location string) OrphanCleanupOption {
 func WithFilesOlderThan(duration time.Duration) OrphanCleanupOption {
 	return func(cfg *orphanCleanupConfig) {
 		cfg.olderThan = duration
+		if duration < 0 && cfg.validationErr == nil {
```
zeroshade, line 99: This rejection changes existing behavior: the integration suite calls `WithFilesOlderThan(-1*time.Hour)` in several places (e.g. `table/orphan_cleanup_integration_test.go:226,273,314,363,403`) to intentionally consider *all* files (including future timestamps), so those tests will now fail at option-validation before scanning. A negative-duration unit test was added, but the integration fixtures using a negative age weren't updated. Please either update those call sites to a non-negative age (and set file mtimes accordingly), or keep an explicit "consider all files" affordance.

zeroshade (changes requested):
The non-negative validation is reasonable, but as written it breaks existing integration tests (inline).

zeroshade (approved):

PR #1444, opened 2026-07-10
fix(table): return defensive metadata getter copies

fallintoplace:
## Summary

Return defensive copies from table metadata getters so callers cannot mutate persisted metadata through returned maps, slices, pointers, or nested values.

The clones cover schemas and defaults, partition specs, snapshots and refs, sort orders, statistics/blob metadata, partition statistics, encryption keys, and properties. Nil slice behavior is preserved.

## Testing

- `go test ./table ./cmd/iceberg`
- regression coverage mutates every returned getter value—including binary defaults, source ID slices, blob metadata, and encryption properties—and verifies the underlying metadata remains unchanged

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1761,15 +1761,15 @@ func (c *commonMetadata) Properties() iceberg.Properties {
 }
 
 func (c *commonMetadata) Statistics() iter.Seq[StatisticsFile] {
-	return slices.Values(c.StatisticsList)
+	return slices.Values(cloneStatisticsFiles(c.StatisticsList))
 }
 
 func (c *commonMetadata) PartitionStatistics() iter.Seq[PartitionStatisticsFile] {
-	return slices.Values(c.PartitionStatsList)
+	return slices.Values(slices.Clone(c.PartitionStatsList))
 }
 
 func (c *commonMetadata) EncryptionKeys() iter.Seq[EncryptionKey] {
-	return slices.Values(c.EncryptionKeyList)
+	return slices.Values(cloneEncryptionKeys(c.EncryptionKeyList))
 }
 
 func cloneSchema(schema *iceberg.Schema) *iceberg.Schema {
@@ -1785,6 +1785,10 @@ func cloneSchema(schema *iceberg.Schema) *iceberg.Schema {
 }
 
 func cloneSchemas(schemas []*iceberg.Schema) []*iceberg.Schema {
+	if schemas == nil {
+		return nil
+	}
+
 	clones := make([]*iceberg.Schema, len(schemas))
 	for i, schema := range schemas {
 		clones[i] = cloneSchema(schema)
@@ -1797,9 +1801,36 @@ func cloneNestedFields(fields []iceberg.NestedField) []iceberg.NestedField {
 	clones := slices.Clone(fields)
 	for i := range clones {
 		clones[i].Type = cloneSchemaType(clones[i].Type)
+		clones[i].InitialDefault = cloneDefault(clones[i].InitialDefault)
+		clones[i].WriteDefault = cloneDefault(clones[i].WriteDefault)
 	}
 
 	return clones
+}
+
+func cloneDefault(value any) any {
+	switch value := value.(type) {
+	case []byte:
+		return slices.Clone(value)
+	case iceberg.BinaryLiteral:
+		return iceberg.BinaryLiteral(slices.Clone([]byte(value)))
+	case iceberg.FixedLiteral:
+		return iceberg.FixedLiteral(slices.Clone([]byte(value)))
+	case []any:
+		clones := make([]any, len(value))
+		for i, item := range value {
+			clones[i] = cloneDefault(item)
+		}
+		return clones
+	case map[string]any:
+		clones := make(map[string]any, len(value))
+		for key, item := range value {
+			clones[key] = cloneDefault(item)
+		}
+		return clones
+	default:
+		return value
+	}
 }
 
 func cloneSchemaType(typ iceberg.Type) iceberg.Type {
@@ -1836,6 +1867,10 @@ func clonePartitionSpec(spec iceberg.PartitionSpec) iceberg.PartitionSpec {
 }
 
 func clonePartitionSpecs(specs []iceberg.PartitionSpec) []iceberg.PartitionSpec {
+	if specs == nil {
+		return nil
+	}
+
 	clones := make([]iceberg.PartitionSpec, len(specs))
 	for i, spec := range specs {
 		clones[i] = clonePartitionSpec(spec)
@@ -1901,6 +1936,10 @@ func cloneSnapshotPtr(snapshot *Snapshot) *Snapshot {
 }
 
 func cloneSnapshots(snapshots []Snapshot) []Snapshot {
+	if snapshots == nil {
+		return nil
+	}
+
 	clones := make([]Snapshot, len(snapshots))
 	for i, snapshot := range snapshots {
 		clones[i] = cloneSnapshot(snapshot)
@@ -1921,9 +1960,56 @@ func cloneSortOrder(order SortOrder) SortOrder {
 }
 
 func cloneSortOrders(orders []SortOrder) []SortOrder {
+	if orders == nil {
+		return nil
+	}
+
 	clones := make([]SortOrder, len(orders))
 	for i, order := range orders {
 		clones[i] = cloneSortOrder(order)
+	}
+
+	return clones
+}
+
+func cloneStatisticsFiles(stats []StatisticsFile) []StatisticsFile {
+	if stats == nil {
+		return nil
+	}
+
+	clones := make([]StatisticsFile, len(stats))
+	for i, stat := range stats {
+		clones[i] = stat
+		if stat.KeyMetadata != nil {
+			value := *stat.KeyMetadata
+			clones[i].KeyMetadata = &value
+		}
+		if stat.BlobMetadata != nil {
+			clones[i].BlobMetadata = make([]BlobMetadata, len(stat.BlobMetadata))
+			for j, blob := range stat.BlobMetadata {
+				clones[i].BlobMetadata[j] = blob
+				clones[i].BlobMetadata[j].Fields = slices.Clone(blob.Fields)
+				clones[i].BlobMetadata[j].Properties = maps.Clone(blob.Properties)
+			}
+		}
+	}
+
+	return clones
+}
+
+func cloneEncryptionKeys(keys []EncryptionKey) []EncryptionKey {
+	if keys == nil {
+		return nil
+	}
+
+	clones := make([]EncryptionKey, len(keys))
+	for i, key := range keys {
+		clones[i] = key
+		if key.EncryptedByID != nil {
+			value := *key.EncryptedByID
+			clones[i].EncryptedByID = &value
+		}
+		clones[i].Properties = maps.Clone(key.Properties)
 	}
 
 	return clones
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1658,18 +1658,18 @@ func (c *commonMetadata) TableUUID() uuid.UUID       { return c.UUID }
 func (c *commonMetadata) Location() string           { return c.Loc }
 func (c *commonMetadata) LastUpdatedMillis() int64   { return c.LastUpdatedMS }
 func (c *commonMetadata) LastColumnID() int          { return c.LastColumnId }
-func (c *commonMetadata) Schemas() []*iceberg.Schema { return slices.Clone(c.SchemaList) }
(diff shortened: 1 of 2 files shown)


table/metadata.go
```diff
@@ -1525,7 +1537,7 @@ func (c *commonMetadata) TableUUID() uuid.UUID       { return c.UUID }
 func (c *commonMetadata) Location() string           { return c.Loc }
 func (c *commonMetadata) LastUpdatedMillis() int64   { return c.LastUpdatedMS }
 func (c *commonMetadata) LastColumnID() int          { return c.LastColumnId }
-func (c *commonMetadata) Schemas() []*iceberg.Schema { return c.SchemaList }
+func (c *commonMetadata) Schemas() []*iceberg.Schema { return slices.Clone(c.SchemaList) }
```
zeroshade: `slices.Clone` copies the outer slice but not the `*iceberg.Schema` pointees, so callers still receive the original `Schema` objects and can mutate exported fields like `ID`/`IdentifierFieldIDs`, corrupting metadata state. To actually protect it, clone each schema (or return immutable copies).

table/metadata.go
```diff
@@ -1536,7 +1548,7 @@ func (c *commonMetadata) CurrentSchema() *iceberg.Schema {
 }
 
 func (c *commonMetadata) PartitionSpecs() []iceberg.PartitionSpec {
-	return c.Specs
+	return slices.Clone(c.Specs)
```
zeroshade: Same shallow-copy issue: `slices.Clone(c.Specs)` copies the `PartitionSpec` structs but they still share their internal `fields` slice and each field's `SourceIDs` backing array, so `specs[i].Field(j).SourceIDs[k] = ...` mutates the original. A deep clone of each spec (fields + SourceIDs, rebuilding derived maps) is needed here — you already do the SourceIDs clone in the builder `clone()` around line 1722.

zeroshade (changes requested):
Good direction on defensive copies for the metadata getters. Two of them are still shallow, which defeats the protection for those types (inline).

table/metadata.go
```diff
@@ -1591,23 +1617,26 @@ func (c *commonMetadata) CurrentSnapshot() *Snapshot {
...
+	return cloneSortOrder(UnsortedSortOrder)
 }
 
 func (c *commonMetadata) DefaultSortOrder() int {
 	return c.DefaultSortOrderID
 }
 
 func (c *commonMetadata) Properties() iceberg.Properties {
-	return c.Props
+	return maps.Clone(c.Props)
```
laskoviymishka, line 1760: While we're clamping down here, the getters just below still hand out shared internal state — `Statistics()`, `PartitionStatistics()` and `EncryptionKeys()` yield value copies of the top-level structs, but `BlobMetadata.Properties`/`Fields` and `EncryptionKey.Properties` are still the internal maps/slices. `metadata.Statistics()...BlobMetadata[0].Properties["x"] = "y"` mutates the real `StatisticsList`.

Same class as what this PR fixes, on the same interface. Either deep-clone these too, or explicitly document which getters guarantee copies and defer the rest — either's fine, but I'd rather not ship an interface where half the getters are safe and half aren't, silently.

table/metadata.go
```diff
@@ -1622,6 +1651,163 @@ func (c *commonMetadata) EncryptionKeys() iter.Seq[EncryptionKey] {
 	return slices.Values(c.EncryptionKeyList)
 }
 
+func cloneSchema(schema *iceberg.Schema) *iceberg.Schema {
+	if schema == nil {
+		return nil
+	}
+
+	return iceberg.NewSchemaWithIdentifiers(
```
laskoviymishka, line 1780: Worth a thought on the hot path: cloning through `NewSchemaWithIdentifiers` runs `init()`, so every clone comes back with cold lazy caches (`idToName`, `nameToID`, etc.). `CurrentSchema()`/`Schemas()` get hit per-file during scan planning, so this is a fair bit of re-derivation and allocation on large tables. (Also `schema.Fields()` already returns a `slices.Clone`, so `cloneNestedFields` clones the top-level slice a second time.)

Not blocking, but I'd consider either caching the clone or shallow-cloning while sharing the atomic cache pointers and copying only the mutable exported fields (`ID`, `IdentifierFieldIDs`). wdyt?

table/metadata.go
```diff
@@ -1622,6 +1651,163 @@ func (c *commonMetadata) EncryptionKeys() iter.Seq[EncryptionKey] {
...
+		schema.ID,
+		slices.Clone(schema.IdentifierFieldIDs),
+		cloneNestedFields(schema.Fields())...,
+	)
+}
+
+func cloneSchemas(schemas []*iceberg.Schema) []*iceberg.Schema {
+	clones := make([]*iceberg.Schema, len(schemas))
```
laskoviymishka, line 1792: One behavior change to watch: `make([]T, len(x))` turns a nil `SchemaList` into a non-nil empty slice, so `Schemas()` now returns `[]` where it used to return `nil` (and marshals to `[]` instead of `null`). Same for `clonePartitionSpecs`/`cloneSortOrders`/`cloneSnapshots`. Any caller using `== nil` to mean "unset", or comparing JSON round-trips, will shift. I'd guard each with `if x == nil { return nil }`.

table/metadata.go
```diff
@@ -1622,6 +1651,163 @@ func (c *commonMetadata) EncryptionKeys() iter.Seq[EncryptionKey] {
...
+		clones[i] = cloneSchema(schema)
+	}
+
+	return clones
+}
+
+func cloneNestedFields(fields []iceberg.NestedField) []iceberg.NestedField {
+	clones := slices.Clone(fields)
+	for i := range clones {
+		clones[i].Type = cloneSchemaType(clones[i].Type)
```
laskoviymishka, line 1803: I think there's still a shallow copy hiding here. We deep-copy `Type` but leave `InitialDefault` and `WriteDefault` as shared `any` values — and those can hold a `[]byte` (BinaryLiteral/FixedLiteral), so a caller can do `schema.Fields()[0].WriteDefault.([]byte)[0] = 0xFF` and reach right back into the original.

This looks like the class zeroshade was pointing at in the shallow-copy thread. I'd add a small `cloneDefault(v any) any` that `slices.Clone`s the `[]byte`-backed cases and passes everything else through, then apply it to both defaults here.

table/metadata.go
```diff
@@ -1622,6 +1651,163 @@ func (c *commonMetadata) EncryptionKeys() iter.Seq[EncryptionKey] {
...
+	default:
+		return typ
+	}
+}
+
+func clonePartitionSpec(spec iceberg.PartitionSpec) iceberg.PartitionSpec {
+	fields := make([]iceberg.PartitionField, spec.NumFields())
+	for i := range fields {
+		fields[i] = spec.Field(i)
+		fields[i].SourceIDs = slices.Clone(fields[i].SourceIDs)
```
laskoviymishka: Minor: this `slices.Clone` is redundant — `NewPartitionSpecID` already clones `SourceIDs` when it builds the field (partitions.go:372), so we clone twice. Can drop the explicit one here.

table/metadata_getters_test.go
```diff
@@ -0,0 +1,126 @@
...
+	schemas := metadata.Schemas()
+	schemas[0].ID = 99
+	schemas[0].IdentifierFieldIDs[0] = 99
+	nestedFields := schemas[0].Fields()
+	nestedFields[0].Name = "mutated"
+	nestedFields[0].Type = iceberg.PrimitiveTypes.String
+
+	partitionSpecs := metadata.PartitionSpecs()
+	partitionField := partitionSpecs[0].Field(0)
+	partitionField.SourceIDs[0] = 99
```
laskoviymishka, line 112: This block doesn't actually guard the `SourceIDs` clone. `PartitionField.SourceIDs` is `json:"-"` (partitions.go:47), so the closing `JSONEq` never sees it change — and `Field(0)` returns a value copy, so `partitionField.Name = "mutated"` is a no-op too. The whole assertion would still pass if we deleted `slices.Clone(fields[i].SourceIDs)` from the clone helper. Same holds for the `PartitionSpec()` / `PartitionSpecByID()` blocks just below.

I'd assert the returned `SourceIDs` is a distinct backing slice directly — mutate it, then read `metadata.Specs[0].Field(0).SourceIDs` back (the test is in package `table`) and require it unchanged — rather than relying on the JSON round-trip.

table/metadata_getters_test.go
```diff
@@ -0,0 +1,126 @@
...
+	refs := metadata.Ref()
+	*refs.MinSnapshotsToKeep = 99
+	for _, ref := range metadata.Refs() {
+		*ref.MaxSnapshotAgeMs = 99
+	}
+
+	sortOrders := metadata.SortOrders()
+	fields := sortOrders[0].Fields()
+	for _, field := range fields {
+		field.SourceIDs[0] = 99
```
laskoviymishka, line 160: Same dead-coverage issue here — `Fields()` yields `SortField` by value, so `field.SourceIDs[0] = 99` mutates a loop-local copy, and `SortField.SourceIDs` is `json:"-"` on top of that. Nothing in this block fails if `cloneSortOrder` is removed. Since we're in package `table`, I'd reach into `sortOrders[0].fields` directly and mutate a serializable field (`Direction`/`NullOrder`), or compare the slice identity against the original.

laskoviymishka (changes requested):
Thanks for this — defending the metadata getters with copies is the right call, and the `SnapshotRef`/`Snapshot`/`Properties` cases look solid.

I'd hold it before merging, though. There's still one shallow copy live: `cloneNestedFields` deep-copies each field's `Type` but leaves `InitialDefault`/`WriteDefault` as shared `any` values, and those can hold a `[]byte` (BinaryLiteral/FixedLiteral) — so a caller can still mutate through `schema.Fields()[n].WriteDefault` back into the original. I think that's the same class zeroshade flagged in the still-open shallow-copy thread, so worth closing them together with a small `cloneDefault` helper.

The other thing is the new test. The `SourceIDs` assertions don't actually exercise the clones they're guarding — `PartitionField.SourceIDs` and `SortField.SourceIDs` are `json:"-"`, so the final `JSONEq` never observes them, and a few of the mutations run against value copies. As written the suite passes whether or not those `SourceIDs` clones exist, which leaves the regression net with a hole right where it counts. I'd assert slice-distinctness directly instead of leaning on the JSON round-trip.

- deep-copy `InitialDefault`/`WriteDefault` in `cloneNestedFields`
- make the `SourceIDs` / SortOrder test assertions actually fail if the clone is removed
- decide `Statistics()`/`PartitionStatistics()`/`EncryptionKeys()` — clone them too, or document them as uncopied and defer

The nil→non-nil slice change and the cold-cache cost on `cloneSchema` are worth a look too, but they're not blockers. Once the copy gap and the test are sorted, happy to take another pass.

zeroshade (approved):
Re-reviewed — the shallow-copy issue I flagged is addressed: `Schemas()` and `PartitionSpecs()` now deep-copy via `cloneSchemas`/`clonePartitionSpecs` (including each field's `SourceIDs`), so callers can't mutate the stored metadata. Thanks @fallintoplace!

zeroshade (commented):
Re-reviewed at head — the correctness issues are addressed (deep-clone of schemas/specs/stats/encryption-keys/defaults incl. byte-backed literals, nil-vs-empty preserved, and the tests now actually guard the clones). Remaining is a non-blocking performance concern below. CI green.

table/metadata.go
```diff
@@ -1709,35 +1735,283 @@ func (c *commonMetadata) CurrentSnapshot() *Snapshot {
...
-	return slices.Values(c.EncryptionKeyList)
+	return slices.Values(cloneEncryptionKeys(c.EncryptionKeyList))
+}
+
+func cloneSchema(schema *iceberg.Schema) *iceberg.Schema {
+	if schema == nil {
+		return nil
+	}
+
+	return iceberg.NewSchemaWithIdentifiers(
```
zeroshade, line 1780: Non-blocking (perf): cloning via `NewSchemaWithIdentifiers` runs `init()`, so every `CurrentSchema()`/`Schemas()` call returns cold `idToName`/`nameToID` caches. These are hit per-file during scan planning — worth a benchmark or a cache-preserving clone before merge.

table/metadata.go
```diff
@@ -1709,35 +1735,283 @@ func (c *commonMetadata) CurrentSnapshot() *Snapshot {
...
+		}
+	default:
+		return typ
+	}
+}
+
+func clonePartitionSpec(spec iceberg.PartitionSpec) iceberg.PartitionSpec {
+	fields := make([]iceberg.PartitionField, spec.NumFields())
+	for i := range fields {
+		fields[i] = spec.Field(i)
```
zeroshade, line 1864: Minor: this `slices.Clone(SourceIDs)` is redundant — `NewPartitionSpecID` already clones `SourceIDs` when building the field.

laskoviymishka (approved):
:shipit:

PR #1447, opened 2026-07-11
feat(encryption): add EncryptionManager, KeyManagementClient interfaces + in-memory KMS

hectar-glitches:
## Scope: 

Landed the first slice of the encryption package per my earlier plan with the interfaces + no-op impl + in-memory KMS. Ready for review. EncryptingFileIO and manifest wiring intentionally deferred to follow-up PRs pending the key_metadata routing decision.

### Public API:

`EncryptionManager`: coordination interface: `NewEncryptedOutputFile(ctx, writer, keyID)` and `NewDecryptedInputFile(ctx, file, keyMetadata)`
`KeyManagementClient`: KMS interface: `WrapKey` / `UnwrapKey` / `SupportsKeyGeneration` / `GenerateKey`
`EncryptionKeyMetadata`:opaque `[]byte` type for the per-file blob stored in `DataFile.KeyMetadata` / `ManifestFile.KeyMetadata` / `StatisticsFile.KeyMetadata`
`EncryptedInputFile` / `EncryptedOutputFile`: file wrappers exposing `KeyMetadata()`; names match the Java `org.apache.iceberg.encryption` types
`NativeEncryptionInputFile` / `NativeEncryptionOutputFile`: extension interfaces returning 'any' for format-native crypto properties (so 'parquet.File{En,De}cryptionProperties' don't force a parquet import into a format-agnostic package)
`PlaintextEncryptionManager`: no-op impl for unencrypted tables; matches Java's `PlaintextEncryptionManager` (silently ignores 'keyID', passes bytes through)
`InMemoryKeyManagementClient`: AES-GCM wrap/unwrap over an in-memory KEK map; documented as testing only
Sentinel errors: `ErrUnknownKeyID`, `ErrInvalidKeyLength`, `ErrCiphertextTooShort`, `ErrAuthenticationFailed`(matches the `Err*` convention used in `io`, `catalog`, root `errors.go`)

### Tests

`PlaintextEncryptionManager` round-trip (data & key metadata pass-through)
`InMemoryKeyManagementClient`: wrap/unwrap round-trip, non-deterministic nonces, tamper detection, cross-key confusion, ciphertext-too-short, KEK overwrite, 'AddKey' input copy isolation, concurrent access via 'errgroup' (with race detector), key generation + round-trip, invalid length paths — all using `errors.Is` and `t.Context()`

### Deferred to follow-ups (per our thread)

`EncryptingFileIO` wrapper — needs the 'key_metadata' routing decision
`encryption.Register` / `LoadKMS' registry for 'encryption.kms-type` / `encryption.kms-impl` catalog properties
Bridge between `table.EncryptionKey.EncryptedKeyMetadata` (base64 'string') and `EncryptionKeyMetadata` (`[]byte`)
Parquet / Manifest / Puffin native encryption wiring
V3-only gating in table format

### Notes
'EncryptionKeyMetadata' is a defined type ('type EncryptionKeyMetadata []byte'), not a type alias. Every call site pulling 'KeyMetadata() []byte' from a manifest or data file will need an explicit conversion ('encryption.EncryptionKeyMetadata(f.KeyMetadata())'). If the group prefers a type alias instead, it's a one-line change here before anything downstream is wired.

'NativeEncryptionInputFile' / 'NativeEncryptionOutputFile' return 'any' for format-native properties. This avoids a parquet import in a format-agnostic package as callers type-assert to '*parquet.FileDecryptionProperties' etc. Happy to discuss alternatives.

'InMemoryKeyManagementClient' uses AES-GCM for key wrapping (matching the Java standard KMS). It is exported (not test-only file) so other packages can use it in their own tests.

### DIFF
--- a/encryption/encryption.go
+++ b/encryption/encryption.go
@@ -1,0 +1,151 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package encryption defines interfaces and utilities for Iceberg table
+// encryption. It provides:
+//
+//   - [EncryptionManager] — the central coordination interface for per-file
+//     envelope encryption and decryption.
+//   - [KeyManagementClient] — the interface for KMS backends (wrap/unwrap DEKs).
+//   - [EncryptedInputFile] / [EncryptedOutputFile] — file wrappers that carry
+//     per-file key metadata alongside the data stream.
+//   - [NativeEncryptionInputFile] / [NativeEncryptionOutputFile] — extensions
+//     for format-native encryption (e.g., Parquet column-level encryption).
+//   - [PlaintextEncryptionManager] — a no-op implementation suitable for
+//     unencrypted tables.
+package encryption
+
+import (
+	"context"
+
+	icebergio "github.com/apache/iceberg-go/io"
+)
+
+// EncryptionKeyMetadata is the opaque per-file key metadata blob embedded in
+// manifest entries (DataFile.KeyMetadata, ManifestFile.KeyMetadata) and
+// statistics files (StatisticsFile.KeyMetadata). The encoding is
+// [EncryptionManager]-defined and typically envelopes the wrapped DEK together
+// with any associated authenticated additional data (AAD).
+type EncryptionKeyMetadata []byte
+
+// EncryptedInputFile is a readable file that carries the per-file encryption
+// key metadata required by an [EncryptionManager] to derive the decryption key.
+type EncryptedInputFile interface {
+	icebergio.File
+	// KeyMetadata returns the opaque per-file encryption key metadata.
+	KeyMetadata() EncryptionKeyMetadata
+}
+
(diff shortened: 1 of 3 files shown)


hectar-glitches:
@rambleraptor first steps

rambleraptor:
@hectar-glitches this looks great! Let's try and divide up work in the issue. I don't want us to be stepping on each other's toes. 

Looks like there was a Docker timeout and the integration test job needs to be re-run.

hectar-glitches:
@rambleraptor I am glad to follow your lead in dividing the work. I'll check the Docker issue.

hectar-glitches:
@rambleraptor rerun the job and it looks good

rambleraptor:
Is this ready for review? Could you take it off draft if it is?

hectar-glitches:
i think it is ready, taking it off draft...

zeroshade (changes requested):
Thanks for getting this started — the interface split looks reasonable. A couple of fail-open issues in the plaintext manager should be addressed before merge:

**Medium — `PlaintextEncryptionManager.NewEncryptedOutputFile` silently ignores `keyID`** (`encryption/encryption.go:126`)

A caller that passes an encryption key ID can successfully write plaintext data with no error. That fails open if encryption is configured but no real manager/KMS is installed.
*Suggested fix:* return an error when `keyID` is non-empty, or make plaintext mode opt-in only for explicitly unencrypted writes.

**Medium — `PlaintextEncryptionManager.NewDecryptedInputFile` accepts key metadata** (`encryption/encryption.go:135`)

It accepts non-empty key metadata and returns the underlying file unchanged, which can silently hand encrypted bytes to readers as if they were plaintext when a real decryption manager is missing.
*Suggested fix:* return an error when `keyMetadata` is non-empty unless plaintext pass-through is explicitly requested via a separate unsafe/test-only path.

hectar-glitches:
@rambleraptor i made `PlaintextEncryptionManager` fail closed

zeroshade (approved):
Re-reviewed the latest revision — the fail-open concerns from my earlier review are fully addressed. `NewEncryptedOutputFile` and `NewDecryptedInputFile` now fail closed (`ErrKeyIDNotSupported` / `ErrKeyMetadataNotSupported`), with focused tests for both paths and solid coverage on the in-memory KMS. Thanks @hectar-glitches!

PR #1448, opened 2026-07-11
fix(table): include statistics in metadata equality

fallintoplace:
## Summary

- compare statistics and partition statistics in common metadata equality
- compare nested blob metadata, key metadata, slices, and maps explicitly
- add metadata equality and SQL catalog commit regressions

## Testing

- go test ./table ./catalog/sql

### DIFF
--- a/catalog/sql/sql_test.go
+++ b/catalog/sql/sql_test.go
@@ -1714,6 +1714,37 @@ func (s *SqliteCatalogTestSuite) TestCommitTable() {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Verified locally: `go vet` clean and tests pass — `TestMetadataEqualsIncludesStatistics` and `TestSqlCatalog/TestCommitTableWithStatisticsUpdate`, plus full `table` and `catalog/sql` packages green. The new `statisticsFileEqual`/`partitionStatisticsFileEqual` correctly close the gap where metadata differing only in statistics compared equal. LGTM.

PR #1449, opened 2026-07-11
fix(hadoop): purge table root after metadata cleanup

fallintoplace:
## Summary

- remove the Hadoop table root directly after purging files instead of calling DropTable after metadata deletion
- preserve data files when gc.enabled=false while removing table metadata
- add local and MinIO-backed purge regressions covering metadata, data, table-root, and CheckTableExists behavior

## Testing

- go test ./catalog/hadoop
- go test -tags integration ./catalog/hadoop -run ^$

### DIFF
--- a/catalog/hadoop/hadoop_minio_integration_test.go
+++ b/catalog/hadoop/hadoop_minio_integration_test.go
@@ -22,6 +22,7 @@ package hadoop_test
 import (
 	"context"
 	"fmt"
+	"io/fs"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


zeroshade:
Note: this overlaps with #1436 — both rewrite the Hadoop `PurgeTable` cleanup. This PR is the more complete version, since it preserves data files when `gc.enabled=false` (which #1436 does not). Suggest closing #1436 in favor of this one.

catalog/hadoop/hadoop_minio_integration_test.go
```diff
@@ -22,6 +22,7 @@ package hadoop_test
 import (
 	"context"
 	"fmt"
+	stdfs "io/fs"
```
zeroshade: Minor: this file imports `io/fs` aliased as `stdfs`, while `hadoop_test.go` in the same package uses the bare `fs` name for the same import. Worth aligning on one alias across the package's test files for consistency.

zeroshade (commented):
The `gc.enabled` gate here is the correct behavior (preserve data files when GC is disabled). One consistency nit inline. Also left a separate comment re: overlap with #1436.

zeroshade (approved):
LGTM, thanks!

PR #1450, opened 2026-07-11
fix(hive): keep catalog properties out of view metadata

fallintoplace:
## Summary

- separate filesystem IO properties from persisted view metadata properties
- keep Hive catalog credentials, endpoints, and connection settings out of view metadata JSON
- add a regression covering fake S3/Hive secrets and custom view properties

## Testing

- go test ./...
- go test -tags integration ./catalog/hive -run ^$

### DIFF
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -310,16 +310,16 @@ func (c *Catalog) CreateView(ctx context.Context, identifier table.Identifier, v
 	defaultNS := catalog.NamespaceFromIdent(identifier)
 	catalogName := "hive"
 
-	// Merge catalog props with view props so io.LoadFS can resolve and view metadata gets user properties.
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Verified locally: `go vet` clean and tests pass — new `TestCreateViewDoesNotPersistCatalogProperties` plus full `catalog/hive` and `view` packages green. Splitting `ioProps` (catalog+view, for `io.LoadFS`) from the persisted `metadataProps` (view-only) correctly keeps catalog secrets (`s3.secret-access-key`, session token, endpoint, etc.) out of view metadata, and `CreateView` stays backward-compatible. LGTM. (Non-blocking follow-up: worth checking whether the SQL/REST view-create paths want the same `CreateViewWithIOProperties` treatment.)

PR #1452, opened 2026-07-11
fix(cli): execute confirmed orphan cleanup plans

fallintoplace:
## Summary

Split orphan cleanup into an explicit plan and execution phase so confirmation applies to the exact files shown to the user.

- capture candidate paths, sizes, and the age cutoff in an immutable cleanup plan;
- show the planned files before confirmation while keeping JSON output to one final document;
- execute only the confirmed plan, without rescanning or adding newly appeared files;
- return the actual deleted-file result and sizes in the final CLI output;
- reject planning-only options when executing an existing plan.

The existing `DeleteOrphanFiles` path uses the same planning and execution flow.

## Testing

- `go test ./table ./cmd/iceberg`
- coordinated plan/execute coverage proves files appearing after planning remain untouched
- CLI coverage verifies preview and final deleted-file output
- real modification-time coverage verifies the age filter

### DIFF
--- a/cmd/iceberg/clean_orphan_files.go
+++ b/cmd/iceberg/clean_orphan_files.go
@@ -40,19 +40,24 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
 
 	opts := []table.OrphanCleanupOption{
 		table.WithFilesOlderThan(olderThan),
-		table.WithDryRun(true),
 	}
 
 	if cmd.Location != "" {
 		opts = append(opts, table.WithLocation(cmd.Location))
 	}
 
-	result, err := tbl.DeleteOrphanFiles(ctx, opts...)
+	plan, err := tbl.PlanOrphanFiles(ctx, opts...)
 	if err != nil {
 		output.Error(fmt.Errorf("orphan file scan failed: %w", err))
 		os.Exit(1)
 	}
 
+	planFiles := plan.Files()
+	result := table.OrphanCleanupResult{
+		OrphanFileLocations: planFiles,
+		OrphanFiles:         plan.OrphanFiles(),
+		TotalSizeBytes:      plan.TotalSizeBytes(),
+	}
 	cliResult := buildCleanOrphanFilesResult(tbl, result, cmd.DryRun)
 
 	if cmd.DryRun {
@@ -61,29 +66,27 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
 		return
 	}
 
-	if len(result.OrphanFileLocations) == 0 {
+	if len(planFiles) == 0 {
 		output.CleanOrphanFilesResult(cliResult)
 
 		return
 	}
 
+	// Text output shows the exact plan before confirmation. JSON output emits
+	// only the final result so one invocation produces one JSON document.
+	if shouldPrintCleanOrphanPreview(output) {
+		previewResult := buildCleanOrphanFilesResult(tbl, result, true)
+		output.CleanOrphanFilesResult(previewResult)
+	}
+
 	prompt := fmt.Sprintf("Delete %d orphan file(s) (%s) from %s?",
-		len(result.OrphanFileLocations), formatBytes(result.TotalSizeBytes), tableIDString(tbl))
+		len(planFiles), formatBytes(plan.TotalSizeBytes()), tableIDString(tbl))
 	if err := confirmAction(prompt, cmd.Yes); err != nil {
 		output.Error(err)
 		os.Exit(1)
 	}
 
-	deleteOpts := []table.OrphanCleanupOption{
-		table.WithFilesOlderThan(olderThan),
-		table.WithDryRun(false),
-	}
-
-	if cmd.Location != "" {
-		deleteOpts = append(deleteOpts, table.WithLocation(cmd.Location))
-	}
-
-	deleteResult, err := tbl.DeleteOrphanFiles(ctx, deleteOpts...)
+	deleteResult, err := tbl.ExecuteOrphanCleanup(ctx, plan)
 	if err != nil {
 		output.Error(fmt.Errorf("orphan file deletion failed: %w", err))
 		os.Exit(1)
@@ -91,6 +94,15 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
 
 	cliResult = buildCleanOrphanFilesResult(tbl, deleteResult, false)
 	output.CleanOrphanFilesResult(cliResult)
+}
+
+func shouldPrintCleanOrphanPreview(output Output) bool {
+	switch output.(type) {
+	case jsonOutput:
+		return false
+	default:
+		return true
+	}
 }
 
 func buildCleanOrphanFilesResult(tbl *table.Table, result table.OrphanCleanupResult, dryRun bool) CleanOrphanFilesResult {
--- a/cmd/iceberg/clean_orphan_files_test.go
+++ b/cmd/iceberg/clean_orphan_files_test.go
@@ -19,14 +19,85 @@ package main
 
 import (
 	"bytes"
+	"context"
+	"io/fs"
 	"os"
 	"testing"
 
+	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/catalog"
+	icebergio "github.com/apache/iceberg-go/io"
 	"github.com/apache/iceberg-go/table"
 	"github.com/pterm/pterm"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+type cleanOrphanCatalog struct {
+	catalog.Catalog
+	table *table.Table
+}
+
+func (c cleanOrphanCatalog) LoadTable(context.Context, table.Identifier) (*table.Table, error) {
+	return c.table, nil
+}
+
+type cleanOrphanOutput struct {
+	Output
+	results []CleanOrphanFilesResult
+	errors  []error
+}
+
+func (o *cleanOrphanOutput) CleanOrphanFilesResult(result CleanOrphanFilesResult) {
+	o.results = append(o.results, result)
+}
+
+func (o *cleanOrphanOutput) Error(err error) {
+	o.errors = append(o.errors, err)
+}
+
+func TestRunCleanOrphanFilesPreviewsPlanBeforeDeletion(t *testing.T) {
+	ctx := context.Background()
+	memFS := icebergio.NewMemFS()
+	location := "mem://preview/table"
+	metadataLocation := location + "/metadata/v1.metadata.json"
+	orphanPath := location + "/data/orphan.parquet"
+
+	meta, err := table.NewMetadata(iceberg.NewSchema(0), nil, table.UnsortedSortOrder, location, nil)
+	require.NoError(t, err)
+	require.NoError(t, memFS.WriteFile(metadataLocation, nil))
+	require.NoError(t, memFS.WriteFile(orphanPath, []byte("orphan")))
+
+	tbl := table.New(
+		table.Identifier{"db", "preview"},
+		meta,
+		metadataLocation,
+		func(context.Context) (icebergio.IO, error) { return memFS, nil },
+		nil,
+	)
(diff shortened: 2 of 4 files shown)


cmd/iceberg/clean_orphan_files.go
```diff
@@ -40,50 +40,46 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
...
 		output.Error(fmt.Errorf("orphan file scan failed: %w", err))
 		os.Exit(1)
 	}
 
-	cliResult := buildCleanOrphanFilesResult(tbl, result, cmd.DryRun)
+	result := table.OrphanCleanupResult{
+		OrphanFileLocations: plan.Files(),
+		TotalSizeBytes:      plan.TotalSizeBytes(),
+	}
+	cliResult := buildCleanOrphanFilesResult(tbl, result, true)
```
zeroshade: Non-blocking: this hardcodes `dryRun=true` for the pre-confirmation result, so a real (non-dry-run) run that finds zero orphans still reports `"dry_run": true` in JSON. Consider using `cmd.DryRun` for the final/no-op result.

cmd/iceberg/clean_orphan_files.go
```diff
@@ -40,50 +40,46 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
...
 	}
 
-	if len(result.OrphanFileLocations) == 0 {
+	if len(plan.Files()) == 0 {
 		output.CleanOrphanFilesResult(cliResult)
 
 		return
 	}
 
+	output.CleanOrphanFilesResult(cliResult)
```
zeroshade: Non-blocking: with `--output json`, emitting this preview and then the post-deletion result later produces two top-level JSON objects on stdout for a single invocation (and an aborted confirmation leaves a success-looking dry-run object behind). Consider keeping the pre-confirmation preview out of the structured output, or emitting a single JSON result at the end.

zeroshade (commented):
Reusing a scanned plan for the confirmed delete is a nice improvement. Two small non-blocking notes on the CLI output (inline).

cmd/iceberg/clean_orphan_files.go
```diff
@@ -40,19 +40,22 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
...
 	}
 
-	result, err := tbl.DeleteOrphanFiles(ctx, opts...)
+	plan, err := tbl.PlanOrphanFiles(ctx, opts...)
 	if err != nil {
 		output.Error(fmt.Errorf("orphan file scan failed: %w", err))
 		os.Exit(1)
 	}
 
+	result := table.OrphanCleanupResult{
```
laskoviymishka, line 56: This hand-built result never populates `OrphanFiles`, and that quietly breaks both the preview and the final output.

`buildCleanOrphanFilesResult` with `dryRun=false` intersects `result.OrphanFiles` against `DeletedFiles` — but `OrphanFiles` is nil here, so the intersection is empty. The text preview prints "No orphan files found." right before we prompt to delete real files, and the final output reuses this same `cliResult` (`deleteResult` from `ExecuteOrphanCleanup` is discarded), so the tool reports `OrphanFileCount: 0` even after it deletes the full set.

The plan already carries the per-file entries privately. I'd add a public accessor — `func (p OrphanCleanupPlan) OrphanFiles() []OrphanFile { return slices.Clone(p.orphanFiles) }` — populate `result.OrphanFiles` from it for the preview, and rebuild `cliResult` from `deleteResult` after `ExecuteOrphanCleanup` before the final output. wdyt?

cmd/iceberg/clean_orphan_files.go
```diff
@@ -61,29 +64,26 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
 		return
 	}
 
-	if len(result.OrphanFileLocations) == 0 {
+	if len(plan.Files()) == 0 {
```
laskoviymishka: `plan.Files()` clones the slice on every call, and we hit it three times in this function (this guard, the prompt, the result build). Minor, but I'd pull it into a local once — or add a `Len() int` on the plan for the count-only checks.

cmd/iceberg/clean_orphan_files.go
```diff
@@ -93,6 +93,15 @@ func runCleanOrphanFiles(ctx context.Context, output Output, cat catalog.Catalog
 	output.CleanOrphanFilesResult(cliResult)
 }
 
+func shouldPrintCleanOrphanPreview(output Output) bool {
+	switch output.(type) {
+	case jsonOutput, *jsonOutput:
```
laskoviymishka: `jsonOutput` is only ever constructed as a value, so the `*jsonOutput` arm is dead outside the test — I'd drop it. Separately, `isMachineReadable` reads a touch clearer than `shouldPrintCleanOrphanPreview`, since that's really the property we're keying on.

cmd/iceberg/clean_orphan_files_test.go
```diff
@@ -171,6 +171,12 @@ func TestBuildCleanOrphanFilesResultDeletedFiltersToDeletedFiles(t *testing.T) {
 	assert.Equal(t, int64(2048), result.OrphanFiles[0].SizeBytes)
 }
 
+func TestShouldPrintCleanOrphanPreview(t *testing.T) {
```
laskoviymishka, line 245: This covers the predicate, but the behavior the PR is actually selling — the text preview listing the exact files before we prompt — has no test. An end-to-end `runCleanOrphanFiles` case that captures text output and asserts the preview contains the orphan paths would have caught the empty-preview bug I flagged in `clean_orphan_files.go`; right now nothing does. Could we add one?

table/orphan_cleanup.go
```diff
@@ -175,12 +175,40 @@ type OrphanFile struct {
...
+	return slices.Clone(p.orphanFileLocations)
+}
+
+// TotalSizeBytes returns the combined size of files in the cleanup plan.
+func (p OrphanCleanupPlan) TotalSizeBytes() int64 {
+	return p.totalSizeBytes
+}
+
+// Cutoff returns the age cutoff used to create the cleanup plan.
+func (p OrphanCleanupPlan) Cutoff() time.Time {
```
laskoviymishka, line 216: `Cutoff()` is only read by the test — no execution path uses it, and `ExecuteOrphanCleanup` doesn't re-check age against it. As a public accessor it reads like an execute-time guard that doesn't exist. I'd either drop it, or godoc it as informational: the plan-time filter, not enforced at execution.

table/orphan_cleanup.go
```diff
@@ -192,23 +220,58 @@ func (t Table) DeleteOrphanFiles(ctx context.Context, opts ...OrphanCleanupOptio
...
+// PlanOrphanFiles identifies orphan files without deleting anything. The
+// returned plan can be shown to a user and passed to ExecuteOrphanCleanup to
+// delete exactly that set.
+func (t Table) PlanOrphanFiles(ctx context.Context, opts ...OrphanCleanupOption) (OrphanCleanupPlan, error) {
+	return t.planOrphanFiles(ctx, newOrphanCleanupConfig(opts...))
+}
+
+// ExecuteOrphanCleanup deletes exactly the files in plan. It does not perform
+// another orphan scan, so files appearing after planning are not included.
+func (t Table) ExecuteOrphanCleanup(ctx context.Context, plan OrphanCleanupPlan, opts ...OrphanCleanupOption) (OrphanCleanupResult, error) {
```
laskoviymishka, line 297: `executeOrphanCleanup` only reads `deleteFunc` and `maxConcurrency` off `cfg`, but `ExecuteOrphanCleanup` accepts the full `OrphanCleanupOption` set — so `WithFilesOlderThan`, `WithLocation`, `WithPrefixMismatchMode` etc. are silently dropped here. Someone calling `ExecuteOrphanCleanup(ctx, plan, WithFilesOlderThan(24*time.Hour))` gets no error and no effect, which is a nasty footgun given the plan's whole point is that its set is already fixed.

I'd narrow this to a smaller option type (just `WithDeleteFunc`/`WithMaxConcurrency`/`WithDryRun`), so the scan-time options aren't even expressible on the execute path. wdyt?

table/orphan_cleanup_test.go
```diff
@@ -86,6 +86,45 @@ func TestOrphanCleanupOptions(t *testing.T) {
...
+
+	tbl := New(
+		[]string{"db", "plan_race"},
+		meta,
+		metadataLocation,
+		func(context.Context) (io.IO, error) { return fs, nil },
+		nil,
+	)
+
+	plan, err := tbl.PlanOrphanFiles(ctx, WithFilesOlderThan(time.Hour))
```
laskoviymishka, line 111: The don't-expand-after-planning behavior this asserts is good and worth having. But `WithFilesOlderThan(time.Hour)` isn't exercising the age filter — MemFS reports a zero `ModTime`, so every file is "older than" any cutoff and the option is a no-op here. If we want real coverage of the age boundary, a separate case with actual mtimes that asserts a recent file gets excluded would do it.

I'd also assert `result.OrphanFiles` here, not just `DeletedFiles` — the path/size pairs are exactly what broke in the CLI.

laskoviymishka (changes requested):
The plan/execute split is the right shape here. Planning once and then deleting exactly that set closes the window between the confirm prompt and the delete, so files that show up after the user confirms can't get swept in — and capturing the cutoff at plan time is the correct instinct.

The blocker is the hand-built `OrphanCleanupResult` in the CLI: it never populates `OrphanFiles`, so `buildCleanOrphanFilesResult` intersects a nil slice and comes up empty. The result is that the text preview prints "No orphan files found." right before we prompt to delete real files, and the final summary reports 0 files deleted even on a successful cleanup (the `deleteResult` from `ExecuteOrphanCleanup` gets discarded). That defeats the "show the exact plan before confirmation" goal the PR is built around. The clean fix is a public `OrphanFiles()` accessor on the plan so the CLI stops reconstructing a lossy result by hand — details inline.

The other thing I'd want addressed is `ExecuteOrphanCleanup` accepting the full option set but silently honoring only `deleteFunc`/`maxConcurrency`/`dryRun`. Passing `WithFilesOlderThan` there does nothing, with no error — a real footgun when the plan is supposed to be immutable.

- populate `OrphanFiles` and rebuild the CLI result from the delete outcome
- narrow (or reject) the options `ExecuteOrphanCleanup` can't honor
- a test for the text preview content, and a real-mtime case so the age filter is actually exercised
- drop or document `Cutoff()`, which nothing enforces at execution

Once those are in, happy to take another pass and approve. I've stayed off the CLI output details since zeroshade's already looking there.

zeroshade (changes requested):
(Medium) The pre-confirmation preview for a non-dry-run cleanup renders zero files.

In `runCleanOrphanFiles` (`cmd/iceberg/clean_orphan_files.go:61`), `cliResult` is built with `dryRun = cmd.DryRun` (i.e. `false` for a real run) from the *plan* result. The "exact plan before confirmation" preview (`clean_orphan_files.go:77-79`) then reuses that result, but `buildCleanOrphanFilesResult` with `dryRun=false` filters entries by `result.DeletedFiles` (`:118-127`) — which is empty before deletion has happened. So the preview always shows 0 orphan files even though `planFiles` is non-empty (guaranteed by the early return at `:69`).

Fix: build the pre-confirmation preview with `dryRun=true` semantics (from `result.OrphanFiles`), and keep the `dryRun=false`/`DeletedFiles` build only for the post-deletion output at `:94`.

zeroshade (approved):
Re-reviewed — my earlier concern is resolved. The pre-confirmation preview now builds with dry-run semantics (the plan's orphan set) instead of the empty post-deletion set, so it shows the real count, and JSON output emits a single document. Thanks @fallintoplace!

zeroshade:
@laskoviymishka mind giving this another look when you have time? The plan/execute split is in place, and the pre-confirmation preview now uses dry-run semantics (it shows the real planned orphan set, and JSON output emits a single document), so the concern I'd raised is resolved. I've approved on my side — a review update from you would unblock the merge. Thanks!

laskoviymishka (approved):
👍

PR #1453, opened 2026-07-11
fix: separate snapshot and reference expiration ages

fallintoplace:
## What

`WithOlderThan` should control snapshot expiration only. Previously it was used as the fallback for `max-ref-age-ms`, so old non-main branches and tags could be removed even when no reference-age policy was configured.

Reference expiration now uses the ref policy, the table `max-ref-age-ms` property, and the specification default. The docs make the separation explicit, and a regression covers an old branch and tag.

## Tests

- `go test ./...`


zeroshade (approved):
Verified locally: `go vet` clean and tests pass — new `TestExpireSnapshotsWithOlderThanDoesNotExpireSnapshotRefs` plus full `table` package green. Correctly separates ref-age from snapshot-age by dropping `cfg.maxSnapshotAgeMs` from the `maxRefAgeMs` fallback, so `WithOlderThan` no longer expires refs. LGTM.

PR #1455, opened 2026-07-11
Unify as-of scan snapshot resolution

fallintoplace:
## What

An as-of timestamp before the first snapshot could make `Scan.Snapshot()` and `Scan.Projection()` fall back to the current snapshot and schema, while `PlanFiles()` reported that no historical snapshot existed.

This adds an error-capable `ResolveSnapshot` method and routes snapshot introspection, schema resolution, manifest planning, and record setup through it. `Snapshot()` remains source-compatible and returns nil when an explicit snapshot cannot be resolved.

## Tests

- `go test ./...`
- Regression coverage for `Snapshot`, `ResolveSnapshot`, `Projection`, `PlanFiles`, and `ToArrowRecords` before the first snapshot.

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -266,9 +266,17 @@ func (scan *Scan) UseRef(name string) (*Scan, error) {
 	return nil, fmt.Errorf("%w: cannot scan unknown ref=%s", iceberg.ErrInvalidArgument, name)
 }
 
-func (scan *Scan) Snapshot() *Snapshot {
+// ResolveSnapshot resolves the snapshot selected by this scan. Live scans use
+// the table's current snapshot; explicit snapshot IDs and as-of timestamps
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified locally: `go vet` clean and tests pass — `TestTable_WithSnapshotAsOf` (all subtests) plus full `table` package green. Centralizing resolution in `ResolveSnapshot()` is a clean dedup with consistent errors. Confirmed the dropped `snap == nil` guard in `effectiveSchema` is safe — the live-scan early return (`snapshotID == nil && asOfTimestamp == nil`) means `snap` is always non-nil past that point — and this also fixes the prior fall-through where an unresolvable as-of silently returned CurrentSnapshot. LGTM.

PR #1461, opened 2026-07-13
feat(rest): Add support for Rename View endpoint

rambleraptor:
We were missing support for `renameView` on the REST Catalog. It looks we started support and then never finished it?

### DIFF
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -2199,6 +2199,47 @@ func (r *Catalog) LoadView(ctx context.Context, identifier table.Identifier) (*v
 	return view.New(identifier, metadata, rsp.MetadataLoc), nil
 }
 
+func (r *Catalog) RenameView(ctx context.Context, from, to table.Identifier) (*view.View, error) {
+	if err := r.endpoints.check(endpointRenameView); err != nil {
+		return nil, err
+	}
+	if err := catalog.ValidateViewIdentifier(from); err != nil {
+		return nil, err
+	}
+	if err := catalog.ValidateViewIdentifier(to); err != nil {
+		return nil, err
+	}
+
+	type payload struct {
+		Source      identifier `json:"source"`
+		Destination identifier `json:"destination"`
+	}
+	src := identifier{
+		Namespace: catalog.NamespaceFromIdent(from),
+		Name:      catalog.TableNameFromIdent(from),
+	}
+	dst := identifier{
+		Namespace: catalog.NamespaceFromIdent(to),
+		Name:      catalog.TableNameFromIdent(to),
+	}
+
+	path, err := endpointRenameView.reqPath()
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = doPost[payload, any](ctx, r.baseURI, path, payload{Source: src, Destination: dst}, r.cl,
+		map[int]error{
+			http.StatusNotFound: catalog.ErrNoSuchView,
+			http.StatusConflict: catalog.ErrViewAlreadyExists,
+		}, allowNoContent())
+	if err != nil {
+		return nil, err
+	}
+
+	return r.LoadView(ctx, to)
+}
+
 // FunctionCatalog is an optional interface for catalogs that support the
 // function (SQL UDF) read endpoints defined by the REST spec: list and load.
 // The function endpoints are not part of the spec's assumed default endpoint
--- a/catalog/rest/rest_integration_test.go
+++ b/catalog/rest/rest_integration_test.go
@@ -304,7 +304,7 @@ func (s *RestIntegrationSuite) TestRenameView() {
 	fromIdent := catalog.ToIdentifier(TestNamespaceIdent, "test-view")
 	toIdent := catalog.ToIdentifier(TestNamespaceIdent, "renamed-view")
 
-	viewRepr := view.NewRepresentation(fmt.Sprintf("SELECT * FROM  %s.%s", TestNamespaceIdent, "test-table"), "trino")
+	viewRepr := view.NewRepresentation(fmt.Sprintf("SELECT * FROM %s.%s", TestNamespaceIdent, "test-table"), "trino")
 	viewVersion, err := view.NewVersion(1, 1, []view.Representation{viewRepr}, table.Identifier{TestNamespaceIdent})
 	s.Require().NoError(err)
 
@@ -317,6 +317,8 @@ func (s *RestIntegrationSuite) TestRenameView() {
 	s.Require().NoError(err)
 	s.Equal(toIdent, renamed.Identifier())
 	s.Equal(created.Metadata().ViewUUID(), renamed.Metadata().ViewUUID())
+	// A rename moves the pointer, it must not rewrite the metadata.
+	s.Equal(created.MetadataLocation(), renamed.MetadataLocation())
 
 	// The old name is gone, the new one exists
 	exists, err := s.cat.CheckViewExists(s.ctx, fromIdent)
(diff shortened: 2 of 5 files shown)


catalog/rest/rest.go
```diff
@@ -2045,3 +2045,44 @@ func (r *Catalog) LoadView(ctx context.Context, identifier table.Identifier) (*v
...
+	}
+
+	path, err := endpointRenameView.reqPath()
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = doPostAllowNoContent[payload, any](ctx, r.baseURI, path, payload{Source: src, Destination: dst}, r.cl,
+		map[int]error{
+			http.StatusNotFound: catalog.ErrNoSuchView,
```
tanmayrauth, line 2233: The rename endpoint returns 404 for two distinct cases, a missing source view AND a missing destination namespace (NoSuchNamespaceException) — and both collapse to ErrNoSuchView here. Impact: a caller that does `errors.Is(err, catalog.ErrNoSuchNamespace)` to auto-create the destination namespace and retry will never fire, and the user gets a misleading "view does not exist" when the real problem is the target namespace. CreateView just above (line 2003) maps 404 to ErrNoSuchNamespace for exactly this reason. Since both are 404 you can't split them on status code alone (you'd need the response `type` field), so this may be an accepted limitation — RenameTable has it too.

tanmayrauth (commented):
Clean port of RenameTable, reads correctly and the endpoint's already wired into viewEndpoints. One question on the 404 mapping, inline.

catalog/rest/rest.go
rambleraptor: I'd advocate for accepting the limitation.

1. It matches what RenameTable does. Consistency feels important here.
2. It's always been unclear to me if the type field is supported by many REST Catalog implementations. Most of those error types aren't standardized by the spec and just float through from the Java implementation if I recall correctly.

tanmayrauth (approved):

catalog/rest/rest_integration_test.go
```diff
@@ -289,6 +289,49 @@ func (s *RestIntegrationSuite) TestUpdateView() {
...
+		tableSchemaSimple, catalog.WithProperties(iceberg.Properties{"foobar": "baz"}),
+		catalog.WithLocation(location))
+	s.Require().NoError(err)
+	s.Require().NotNil(tbl)
+
+	// Create a view to rename
+	fromIdent := catalog.ToIdentifier(TestNamespaceIdent, "test-view")
+	toIdent := catalog.ToIdentifier(TestNamespaceIdent, "renamed-view")
+
+	viewRepr := view.NewRepresentation(fmt.Sprintf("SELECT * FROM  %s.%s", TestNamespaceIdent, "test-table"), "trino")
```
laskoviymishka: tiny one — double space in the format string (`FROM  %s`). I know it's copied from the other view tests, but since we're adding a fresh one we could quietly fix it here. wdyt?

catalog/rest/rest_integration_test.go
```diff
@@ -289,6 +289,49 @@ func (s *RestIntegrationSuite) TestUpdateView() {
...
+
+	created, err := s.cat.CreateView(s.ctx, fromIdent, viewVersion, tableSchemaSimple,
+		catalog.WithViewProperties(iceberg.Properties{"foobar": "baz"}))
+	s.Require().NoError(err)
+
+	// Rename it
+	renamed, err := s.cat.RenameView(s.ctx, fromIdent, toIdent)
+	s.Require().NoError(err)
+	s.Equal(toIdent, renamed.Identifier())
+	s.Equal(created.Metadata().ViewUUID(), renamed.Metadata().ViewUUID())
```
laskoviymishka, line 319: Could we also capture `created.MetadataLocation()` before the rename and assert it's unchanged after? We prove the UUID and name carry over, but "a rename doesn't rewrite metadata" is the property most likely to regress silently, and it's one more line here.

catalog/rest/rest_test.go
```diff
@@ -2660,6 +2660,103 @@ func (r *RestCatalogSuite) TestRegisterView409() {
...
+		r.Equal("source", payload.Source.Name)
+		r.Equal([]string{"example"}, payload.Destination.Namespace)
+		r.Equal("destination", payload.Destination.Name)
+
+		w.WriteHeader(http.StatusNoContent)
+	})
+
+	// Mock the load view endpoint for loading the renamed view.
+	r.mux.HandleFunc("/v1/namespaces/example/views/destination", func(w http.ResponseWriter, req *http.Request) {
+		r.Require().Equal(http.MethodGet, req.Method)
```
laskoviymishka, line 2694: This LoadView leg only checks the method, and the 404/409 handlers below skip headers entirely — the 204 rename handler and the analogous TestRegisterView tests all run the `for k, v := range TestHeaders` loop. I'd add it here and in the two error handlers so we're actually exercising auth on every leg.

catalog/rest/rest_test.go
```diff
@@ -2660,6 +2660,103 @@ func (r *RestCatalogSuite) TestRegisterView409() {
...
+	renamed, err := cat.RenameView(context.Background(), fromIdent, toIdent)
+	r.Require().NoError(err)
+
+	r.Equal(toIdent, renamed.Identifier())
+	r.Equal(metadataLoc, renamed.MetadataLocation())
+	r.Equal(uuid.MustParse("a1b2c3d4-e5f6-7890-1234-567890abcdef"), renamed.Metadata().ViewUUID())
+	r.Equal(exampleViewSQL, renamed.Metadata().CurrentVersion().Representations[0].Sql)
+}
+
+func (r *RestCatalogSuite) TestRenameView404() {
```
laskoviymishka, line 2720: RenameView has two `ValidateViewIdentifier` guards but nothing exercises them — `TestTableAndViewIdentifiersRequireNamespace` covers RenameTable's namespace-absent case and adds nothing for views, so both branches are currently unproven. I'd add the mirror of the RenameTable cases there:

```go
_, err = cat.RenameView(context.Background(), table.Identifier{"source"}, table.Identifier{"example", "destination"})
r.ErrorIs(err, catalog.ErrNoSuchView)

_, err = cat.RenameView(context.Background(), table.Identifier{"example", "source"}, table.Identifier{"destination"})
r.ErrorIs(err, catalog.ErrNoSuchView)
```

This is the one gap I'd want closed before it goes in.

laskoviymishka (changes requested):
This is a clean port of RenameTable — the endpoint path, request body, and the 204/404/409 handling all look right, and the integration test actually round-trips a rename. No objection to merging.

The one thing I'd like before it lands is test coverage for the two `ValidateViewIdentifier` guards. `TestTableAndViewIdentifiersRequireNamespace` exercises those for RenameTable but nothing hits them for RenameView, so both guard branches ship unproven — a quick mirror of the RenameTable cases closes it.

A couple of smaller things worth doing while we're here: the error-path handlers (and the LoadView leg of the 204 test) skip the `TestHeaders` assertion the rest of the view tests run, and the integration test could assert the metadata location is unchanged across the rename.

Two things I noticed but wouldn't block on, both inherited straight from RenameTable: the spec allows a 404 on rename to also mean the target namespace is missing (we map every 404 to `ErrNoSuchView`), and returning the loaded view means a post-rename `LoadView` failure surfaces as a rename error even though the rename committed. Both are fine to leave as follow-ups since you're matching the existing pattern — just flagging them.

Fix the identifier tests and I'm happy to approve.

catalog/rest/rest_integration_test.go
rambleraptor: Yeah, we should do this the right way. Fixed.

catalog/rest/rest_test.go
rambleraptor: There's some potential room for test consolidation here, but I'll just add those checks in for now.

catalog/rest/rest_test.go
rambleraptor: Yeah, this is a great call

laskoviymishka (approved):
All looks good now! 

:shipit:

PR #1463, opened 2026-07-13
feat(transforms): read unknown partition and sort transforms (v3)

rambleraptor:
The v3 spec states that:

```
Partition fields that use an unknown transform can be read by ignoring the partition field for the purpose of filtering data files during scan planning. In v1 and v2, readers should ignore fields with unknown transforms while reading; this behavior is required in v3.
```

This introduces the concept of an UnknownTransform and then makes sure that v3 ignores it for partitioning.

### DIFF
--- a/expr_json.go
+++ b/expr_json.go
@@ -594,6 +594,11 @@ func decodeTerm(raw json.RawMessage) (UnboundTerm, error) {
 		if err != nil {
 			return nil, fmt.Errorf("%w: cannot parse transform term: %s", ErrInvalidArgument, err)
 		}
+		// Unknown transforms are tolerated in partition/sort metadata, but a
+		// filter expression referencing one can't be evaluated.
+		if _, ok := tf.(UnknownTransform); ok {
+			return nil, fmt.Errorf("%w: unknown transform in expression term: %s", ErrInvalidArgument, t.Transform)
+		}
 
 		return NewUnboundTransform(tf, Reference(t.Term)), nil
 	default:
--- a/partitions.go
+++ b/partitions.go
@@ -325,6 +325,14 @@ func validateTransform(transform Transform) error {
 		return t.validateWidth()
 	case *TruncateTransform:
 		return t.validateWidth()
+	case UnknownTransform:
+		// The zero value is constructible from outside the package and would
+		// serialize as "transform": "".
+		if t.String() == "" {
+			return fmt.Errorf("%w: unknown transform has no name", ErrInvalidTransform)
+		}
+
+		return nil
 	default:
 		return nil
 	}
@@ -730,6 +738,11 @@ func GeneratePartitionFieldName(schema *Schema, field PartitionField) (string, e
 
 	transform := field.Transform
 	switch t := transform.(type) {
+	case UnknownTransform:
+		// A generated name would embed the transform's brackets, e.g.
+		// "id_custom_transform[42]". Make the caller supply one.
+		return "", fmt.Errorf("%w: partition field using unknown transform %s must be given an explicit name",
+			ErrInvalidTransform, t)
 	case IdentityTransform:
 		return sourceName, nil
 	case VoidTransform:
--- a/partitions.go
+++ b/partitions.go
@@ -321,6 +321,8 @@ func validateTransform(transform Transform) error {
 		return t.validateWidth()
 	case *TruncateTransform:
 		return t.validateWidth()
+	case UnknownTransform:
+		return fmt.Errorf("%w: cannot write partition spec with unknown transform: %s", ErrInvalidTransform, t.String())
 	default:
 		return nil
 	}
--- a/partitions.go
+++ b/partitions.go
@@ -125,6 +125,10 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 	_, hasSourceID := raw["source-id"]
 	_, hasSourceIDs := raw["source-ids"]
 
+	if tf, ok := raw["transform"]; !ok || string(tf) == "null" {
+		return fmt.Errorf("%w: partition field requires a transform", ErrInvalidTransform)
+	}
+
 	aux := struct {
 		SourceID        int    `json:"source-id"`
 		SourceIDs       []int  `json:"source-ids,omitempty"`
--- a/partitions.go
+++ b/partitions.go
@@ -321,8 +321,6 @@ func validateTransform(transform Transform) error {
 		return t.validateWidth()
 	case *TruncateTransform:
 		return t.validateWidth()
-	case UnknownTransform:
-		return fmt.Errorf("%w: cannot write partition spec with unknown transform: %s", ErrInvalidTransform, t.String())
 	default:
 		return nil
 	}
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -526,6 +526,24 @@ func TestPartitionSpecToPathWithDroppedLeadingSourceColumn(t *testing.T) {
 	assert.Equal(t, "foo=null/bar=7/baz=true", spec.PartitionToPath(record, schema))
 }
 
+// An unknown transform renders the partition value, so two different values
+// must not collapse to the same path -- that path string is the writer key.
+func TestPartitionSpecToPathUnknownTransform(t *testing.T) {
+	schema := iceberg.NewSchema(0,
+		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int32, Required: true})
+
+	unknown, err := iceberg.ParseTransform("custom_transform[42]")
+	require.NoError(t, err)
+
+	spec := iceberg.NewPartitionSpecID(3, iceberg.PartitionField{
+		SourceIDs: []int{1}, FieldID: 1000, Transform: unknown, Name: "custom",
+	})
+
+	assert.Equal(t, "custom=1", spec.PartitionToPath(partitionRecord{int32(1)}, schema))
+	assert.Equal(t, "custom=2", spec.PartitionToPath(partitionRecord{int32(2)}, schema))
+	assert.Equal(t, "custom=null", spec.PartitionToPath(partitionRecord{nil}, schema))
+}
+
 func TestGetPartitionFieldName(t *testing.T) {
 	schema := iceberg.NewSchema(0,
 		iceberg.NestedField{ID: 1, Name: "str", Type: iceberg.PrimitiveTypes.String},
@@ -581,6 +599,25 @@ func TestGetPartitionFieldName(t *testing.T) {
 			assert.Equal(t, test.expectedName, name)
 		})
 	}
+}
+
+// Generating a name would embed the transform's brackets, so an unknown
+// transform needs an explicit one.
+func TestGetPartitionFieldNameUnknownTransform(t *testing.T) {
+	schema := iceberg.NewSchema(0,
+		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int32, Required: true})
+
+	unknown, err := iceberg.ParseTransform("custom_transform[42]")
+	require.NoError(t, err)
+
+	_, err = iceberg.GeneratePartitionFieldName(schema,
+		iceberg.PartitionField{SourceIDs: []int{1}, FieldID: 1000, Transform: unknown})
+	require.ErrorIs(t, err, iceberg.ErrInvalidTransform)
+
+	name, err := iceberg.GeneratePartitionFieldName(schema,
+		iceberg.PartitionField{SourceIDs: []int{1}, FieldID: 1000, Transform: unknown, Name: "custom"})
+	require.NoError(t, err)
+	assert.Equal(t, "custom", name)
 }
 
 func TestPartitionFieldUnmarshalJSON(t *testing.T) {
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -254,6 +254,47 @@ func TestPartitionSpec_MarshalTextRejectsInvalidTruncateTransform(t *testing.T) 
 	require.ErrorContains(t, err, "width > 0")
 }
 
+// A v3 table may use a partition transform this implementation doesn't know.
+// Readers must load it and preserve it on write, ignoring it when filtering.
+func TestPartitionFieldUnknownTransformRoundTrip(t *testing.T) {
+	jsonData := `
+	{
+		"source-id": 1,
+		"field-id": 1000,
+		"transform": "custom_transform[42]",
+		"name": "id_custom"
+	}`
+	var field iceberg.PartitionField
+	err := json.Unmarshal([]byte(jsonData), &field)
+	require.NoError(t, err)
+
+	_, ok := field.Transform.(iceberg.UnknownTransform)
+	require.True(t, ok, "unknown transform should parse to UnknownTransform")
+	assert.Equal(t, "custom_transform[42]", field.Transform.String())
+
+	data, err := json.Marshal(field)
+	require.NoError(t, err)
+	assert.JSONEq(t, jsonData, string(data))
+}
+
+// Writers must not commit a partition spec that uses an unknown transform.
+func TestPartitionSpecRejectsUnknownTransform(t *testing.T) {
+	schema := iceberg.NewSchema(1, iceberg.NestedField{
+		ID:   1,
+		Name: "id",
+		Type: iceberg.PrimitiveTypes.Int32,
+	})
+
+	unknown, err := iceberg.ParseTransform("custom_transform[42]")
+	require.NoError(t, err)
+
+	_, err = iceberg.NewPartitionSpecOpts(
+		iceberg.AddPartitionFieldBySourceID(1, "id_custom", unknown, schema, nil),
+	)
+	require.ErrorIs(t, err, iceberg.ErrInvalidTransform)
+	require.ErrorContains(t, err, "custom_transform[42]")
+}
+
 func TestUnpartitionedWithVoidField(t *testing.T) {
 	spec := iceberg.NewPartitionSpec(iceberg.PartitionField{
 		SourceIDs: []int{3}, FieldID: 1001, Name: "void", Transform: iceberg.VoidTransform{},
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -277,8 +277,9 @@ func TestPartitionFieldUnknownTransformRoundTrip(t *testing.T) {
 	assert.JSONEq(t, jsonData, string(data))
 }
 
-// Writers must not commit a partition spec that uses an unknown transform.
-func TestPartitionSpecRejectsUnknownTransform(t *testing.T) {
+// Building a spec tolerates unknown transforms, matching Java's PartitionSpec
+// builder. Rejection happens later, only when evolving such a spec.
+func TestPartitionSpecAcceptsUnknownTransform(t *testing.T) {
 	schema := iceberg.NewSchema(1, iceberg.NestedField{
 		ID:   1,
 		Name: "id",
@@ -288,11 +289,13 @@ func TestPartitionSpecRejectsUnknownTransform(t *testing.T) {
 	unknown, err := iceberg.ParseTransform("custom_transform[42]")
 	require.NoError(t, err)
 
-	_, err = iceberg.NewPartitionSpecOpts(
+	spec, err := iceberg.NewPartitionSpecOpts(
 		iceberg.AddPartitionFieldBySourceID(1, "id_custom", unknown, schema, nil),
 	)
-	require.ErrorIs(t, err, iceberg.ErrInvalidTransform)
-	require.ErrorContains(t, err, "custom_transform[42]")
+	require.NoError(t, err)
+
+	_, ok := spec.Field(0).Transform.(iceberg.UnknownTransform)
+	assert.True(t, ok)
 }
 
 func TestUnpartitionedWithVoidField(t *testing.T) {
--- a/table/metadata_internal_test.go
+++ b/table/metadata_internal_test.go
@@ -423,6 +423,13 @@ func TestMetadataV3ParsesUnknownTransforms(t *testing.T) {
 	_, ok = sf.Transform.(iceberg.UnknownTransform)
 	assert.True(t, ok, "unknown sort transform should load")
 	assert.Equal(t, "custom_sort[7]", sf.Transform.String())
+
+	// Marshalling back must preserve the unknown transforms verbatim, proving
+	// no write-path guard strips or rewrites them.
+	out, err := json.Marshal(meta)
+	require.NoError(t, err)
+	assert.Contains(t, string(out), "custom_transform[42]")
+	assert.Contains(t, string(out), "custom_sort[7]")
 }
 
 func TestMetadataEqualsIncludesStatistics(t *testing.T) {
--- a/table/metadata_internal_test.go
+++ b/table/metadata_internal_test.go
@@ -400,6 +400,31 @@ func assertMissingLastUpdatedMS(t *testing.T, metadata map[string]any) {
 	assert.ErrorContains(t, err, "last-updated-ms is absent or null")
 }
 
+// A full v3 metadata document must load when its partition spec and sort order
+// use unknown transforms — nothing on the metadata path may trip a guard.
+func TestMetadataV3ParsesUnknownTransforms(t *testing.T) {
+	j := strings.NewReplacer(
+		`{"name": "x", "transform": "identity", "source-id": 1, "field-id": 1000}`,
+		`{"name": "x", "transform": "custom_transform[42]", "source-id": 1, "field-id": 1000}`,
+		`{"transform": "bucket[4]", "source-id": 3, "direction": "desc", "null-order": "nulls-last"}`,
+		`{"transform": "custom_sort[7]", "source-id": 3, "direction": "desc", "null-order": "nulls-last"}`,
+	).Replace(ExampleTableMetadataV3)
+
+	meta, err := ParseMetadataBytes([]byte(j))
+	require.NoError(t, err)
+
+	spec := meta.PartitionSpec()
+	pf := spec.Field(0)
+	_, ok := pf.Transform.(iceberg.UnknownTransform)
+	assert.True(t, ok, "unknown partition transform should load")
+	assert.Equal(t, "custom_transform[42]", pf.Transform.String())
+
+	sf := meta.SortOrder().Field(1)
+	_, ok = sf.Transform.(iceberg.UnknownTransform)
+	assert.True(t, ok, "unknown sort transform should load")
+	assert.Equal(t, "custom_sort[7]", sf.Transform.String())
+}
+
 func TestMetadataEqualsIncludesStatistics(t *testing.T) {
 	builder := builderWithoutChanges(2)
 	base, err := builder.Build()
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -249,6 +249,11 @@ func (s SortOrder) Len() int {
 	return len(s.fields)
 }
 
+// Field returns the sort field at index i.
+func (s SortOrder) Field(i int) SortField {
+	return s.fields[i]
+}
+
 func (s SortOrder) MarshalJSON() ([]byte, error) {
 	type Alias struct {
 		OrderID int         `json:"order-id"`
@@ -333,10 +338,6 @@ func newSortOrder(orderID int, fields []SortField, validateSourceIDs bool) (Sort
 			if err := validateSortSourceIDs(field.SourceIDs); err != nil {
 				return SortOrder{}, fmt.Errorf("%w: sort field at index %d has invalid source IDs: %v",
 					ErrInvalidSortSourceID, idx, err)
-			}
-			if u, ok := field.Transform.(iceberg.UnknownTransform); ok {
-				return SortOrder{}, fmt.Errorf("%w: cannot write sort order with unknown transform: %s",
-					ErrInvalidTransform, u.String())
 			}
 		}
 	}
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -334,6 +334,10 @@ func newSortOrder(orderID int, fields []SortField, validateSourceIDs bool) (Sort
 				return SortOrder{}, fmt.Errorf("%w: sort field at index %d has invalid source IDs: %v",
 					ErrInvalidSortSourceID, idx, err)
 			}
+			if u, ok := field.Transform.(iceberg.UnknownTransform); ok {
+				return SortOrder{}, fmt.Errorf("%w: cannot write sort order with unknown transform: %s",
+					ErrInvalidTransform, u.String())
+			}
 		}
 	}
 
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -146,6 +146,10 @@ func (s *SortField) UnmarshalJSON(b []byte) error {
 		return fmt.Errorf("%w: exactly one of source-id or source-ids is required", ErrInvalidSortSourceID)
 	}
 
+	if tf, ok := raw["transform"]; !ok || string(tf) == "null" {
+		return fmt.Errorf("%w: sort field requires a transform", iceberg.ErrInvalidTransform)
+	}
+
 	aux := struct {
 		SourceID        int           `json:"source-id"`
 		SourceIDs       []int         `json:"source-ids,omitempty"`
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -253,9 +253,11 @@ func (s SortOrder) Len() int {
 	return len(s.fields)
 }
 
-// Field returns the sort field at index i.
+// Field returns a copy of the sort field at index i, like Fields does, so a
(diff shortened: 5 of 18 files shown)


transforms.go
```diff
@@ -230,6 +239,49 @@ func (VoidTransform) Project(string, BoundPredicate) (UnboundPredicate, error) {
...
+func (UnknownTransform) Apply(Optional[Literal]) Optional[Literal] {
+	return Optional[Literal]{}
+}
+
+func (UnknownTransform) ToHumanStr(any) string { return "null" }
+
+func (UnknownTransform) ToHumanStrType(Type, any) string { return "null" }
+
+// Project returns nil so scans don't prune on an unknown partition field.
+func (UnknownTransform) Project(string, BoundPredicate) (UnboundPredicate, error) {
```
tanmayrauth, line 304: This returning nil is the entire mechanism for "ignore the unknown field when filtering" — inclusiveProjection in table/evaluators.go treats a nil projection as AlwaysTrue — but there's no test for it. If a later change makes Project return a non-nil predicate, or the projection loop stops treating nil as AlwaysTrue, unknown partition fields start pruning again and the suite stays green, silently dropping matching rows. Please add a test that builds a spec with an unknown transform and asserts a predicate on that source column plans without pruning.

partitions.go
```diff
@@ -291,6 +291,8 @@ func validateTransform(transform Transform) error {
 		return t.validateNumBuckets()
 	case *BucketTransform:
 		return t.validateNumBuckets()
+	case UnknownTransform:
```
tanmayrauth, line 328: Partition specs reject unknown transforms on write here, but the sort-order path doesn't: newSortOrder (table/sorting.go:309) and CheckCompatibility (table/sorting.go:334) accept them, and CheckCompatibility passes only because UnknownTransform.CanTransform returns true unconditionally. An UnknownTransform is un-evaluatable (Apply returns empty, Project returns nil), so a SortOrder built with one commits an un-enforceable sort into metadata with no guard and no test — downstream engines then read a declared sort order they can't reproduce. Either mirror the partition-spec rejection on the sort-write path, or add a test pinning the intended behavior.

transforms.go
```diff
@@ -230,6 +239,49 @@ func (VoidTransform) Project(string, BoundPredicate) (UnboundPredicate, error) {
...
+
+	return ok && t.name == o.name
+}
+
+// Apply can't be evaluated for an unknown transform.
+func (UnknownTransform) Apply(Optional[Literal]) Optional[Literal] {
+	return Optional[Literal]{}
+}
+
+func (UnknownTransform) ToHumanStr(any) string { return "null" }
```
tanmayrauth: Returning the literal "null" here (and in ToHumanStrType on 278) is indistinguishable from the human rendering of an actual null value, so partition-summary/describe output for an unknown-transform field reads as "null" and an operator can't tell "unknown transform" from "null partition value." A distinct marker (e.g. the transform name) would avoid the ambiguity.

tanmayrauth (commented):
Read path and no-prune behavior check out. One missing test on the core behavior, plus a sort-order write gap and a display nit. Details inline.

partitions.go
rambleraptor: Mirrored the rejection.

laskoviymishka, line 328: This case is right for a direct write, but it also fires on the preserve path. `UpdateSpec.Apply` rebuilds the whole spec through `NewPartitionSpecOpts`, so every existing field — including a passthrough unknown-transform one we loaded from a v3 table — gets re-validated here. That means once a table has an unknown partition field, we can't evolve its spec at all: renaming or dropping an unrelated field fails with `ErrInvalidTransform` at commit.

I'd only validate the fields being newly added, and route the preserved fields through a non-validating construction. Java rejects at the `UpdateSpec` constructor with a clear message rather than an opaque commit-time error, so an early guard there would be friendlier too. A test that evolves a spec on an unknown-transform table would lock this down. wdyt?

partitions_test.go
```diff
@@ -191,6 +191,47 @@ func TestPartitionSpec_MarshalTextRejectsInvalidBucketTransform(t *testing.T) {
 	require.ErrorContains(t, err, "numBuckets > 0")
 }
 
+// A v3 table may use a partition transform this implementation doesn't know.
+// Readers must load it and preserve it on write, ignoring it when filtering.
+func TestPartitionFieldUnknownTransformRoundTrip(t *testing.T) {
```
laskoviymishka, line 259: Nice that this proves the field round-trips. Could we add a sibling test that unmarshals a full table-metadata JSON — partition specs and all — with an unknown transform and asserts it loads clean? That's the case that actually matters for the read goal, and the one thing that proves loading a v3 table doesn't trip `validateTransform` anywhere on the metadata path. The field-level test doesn't exercise that.

table/sorting.go
```diff
@@ -320,6 +320,10 @@ func newSortOrder(orderID int, fields []SortField, validateSourceIDs bool) (Sort
 				return SortOrder{}, fmt.Errorf("%w: sort field at index %d has invalid source IDs: %v",
 					ErrInvalidSortSourceID, idx, err)
 			}
+			if u, ok := field.Transform.(iceberg.UnknownTransform); ok {
```
laskoviymishka: Two things on this guard.

First, it's stricter than the spec and than Java. The spec only forbids writing partition specs with unknown transforms, and Java's `SortOrder` builder doesn't reject unknown sort transforms (`canTransform` stays true). A v3 sort order round-tripped through `NewSortOrder` — say during `UpdateSortOrder` — would get wrongly rejected here. I think this ties into tanmayrauth's sort-order thread, so I want to make sure we actually want to reject on write at all, rather than tolerate it like Java does.

Second, if we do keep it: it's inside the `validateSourceIDs` block, so the JSON unmarshal path (`validateSourceIDs=false`) skips it entirely, and `MetadataBuilder.AddSortOrder` bypasses it too. That's an inconsistent guard. I'd pull it out of the `validateSourceIDs` branch so write and load agree. wdyt?

table/sorting_test.go
```diff
@@ -262,6 +280,33 @@ func TestUnmarshalInvalidSortTransform(t *testing.T) {
 		]
 	}`
 
+	var order table.SortOrder
+	err := json.Unmarshal([]byte(j), &order)
+	require.NoError(t, err)
+	require.Equal(t, 3, order.Len())
+
+	var first table.SortField
+	for i, f := range order.Fields() {
```
laskoviymishka: Tiny thing — this loop only exists to grab field 0. `order.Fields()[0]` (or a `Field(0)` accessor if we add one) reads cleaner.

transforms.go
```diff
@@ -77,7 +84,9 @@ func ParseTransform(s string) (Transform, error) {
 		return HourTransform{}, nil
 	}
 
-	return nil, fmt.Errorf("%w: %s", ErrInvalidTransform, s)
+	// Unknown transform: v3 readers must load these and ignore them when
+	// filtering instead of failing. Keep the original string so it round-trips.
+	return UnknownTransform{name: s}, nil
```
laskoviymishka, line 96: Since `ParseTransform` now keeps the original casing here and `Equals` compares `name` byte-for-byte, `"Custom_V2[3]"` and `"custom_v2[3]"` become two distinct `UnknownTransform`s — the new test even pins that behavior. On a partition spec that'd let two effectively-identical unknown fields coexist. Probably fine in practice, but I'd either normalize the stored name or leave a comment noting the case-sensitivity is intentional for round-trip fidelity.

laskoviymishka (changes requested):
The read side of this is nicely done — `ParseTransform` splitting malformed-known transforms (still error) from genuinely-unknown ones (parse to `UnknownTransform`) is exactly the right line to draw, `Project` returning nil so unknown fields never prune is correct, and the `expr_json` guard rejecting unknown transforms in filter terms is a good catch. The parsing and projection tests are thorough.

I'd hold before merging though — the read tolerance doesn't yet survive contact with the write and evolution paths.

The bigger one: `UpdateSpec.Apply` rebuilds the entire spec through the validating constructor, so any table carrying an unknown partition field can't have its spec evolved at all — renaming or dropping an unrelated field fails at commit. That breaks the load-and-preserve goal the moment someone tries to evolve. Detail's inline; the fix is to only validate newly-added fields.

The other: a table loaded with an unknown partition transform reads fine, but `WriteRecords` on it takes the partitioned path (`IsUnpartitioned` is false), and `UnknownTransform.Apply` returns an empty Optional — so we'd silently write null partition values (or panic in `fileToDataFile`) with no guard at the write entry point. I'd reject unknown partition transforms at `recordsToDataFiles`/`WriteRecords` rather than let it corrupt silently.

tanmayrauth's missing-test point is worth picking up too: a full table-metadata JSON round-trip (not just the field-level one) is the test that actually proves a v3 table loads, and it doubles as the guardrail against `validateTransform` sneaking onto the read path later. And the new sort-order write rejection ties into their sort-order thread — Java doesn't reject unknown sort transforms on write, so I flagged whether we want that guard at all.

- `UpdateSpec.Apply`: validate only new fields, not preserved ones (plus a spec-evolution test on an unknown-transform table)
- Guard `WriteRecords` against unknown partition transforms so we don't write null partitions silently
- Resolve the sort-order write rejection: keep-and-move-it out of the `validateSourceIDs` block, or drop it to match Java
- A full table-metadata JSON round-trip test

Once those are sorted, happy to take another pass and approve.

partitions.go
rambleraptor: Sounds good. I added the early guard + tests.

partitions_test.go
rambleraptor: Great idea for a test. Added it.

table/sorting.go
rambleraptor: Originally, I was thinking that we should be stricter than Java and then could open it up in a follow-up PR. Let's go ahead and tolerate it instead. It's better to stick with the reference implementation than figure out the rejection logic.

transforms.go
rambleraptor: I kept the behavior + added a comment. I double checked Java and it appears to do a byte-for-byte check as well.

rambleraptor:
@laskoviymishka thanks for the review! I've gone ahead and made those changes.

transforms.go
```diff
@@ -230,6 +240,49 @@ func (VoidTransform) Project(string, BoundPredicate) (UnboundPredicate, error) {
...
+func (UnknownTransform) PreservesOrder() bool { return false }
+
+func (t UnknownTransform) Equals(other Transform) bool {
+	o, ok := other.(UnknownTransform)
+
+	return ok && t.name == o.name
+}
+
+// Apply can't be evaluated for an unknown transform.
+func (UnknownTransform) Apply(Optional[Literal]) Optional[Literal] {
```
laskoviymishka, line 287: this is the round-1 write-path concern, still open — flagging it here because `Apply` returning an empty Optional (and `ToHumanStrType` just below returning the transform name) is exactly what lets it corrupt silently.

A table loaded with an unknown partition transform isn't unpartitioned (`IsUnpartitioned` only counts `VoidTransform`), so `WriteRecords` takes the partitioned path, `getRecordPartitions` calls this `Apply`, gets `Valid:false` for every row, and we write null partition values under a `field=<transform-name>` dir — unreadable by Java/PyIceberg. On the `AddFiles` path `fileToDataFile` panics instead.

I'd guard at `recordsToDataFiles`/`WriteRecords`: range the current spec and reject if any field is an `UnknownTransform` before writing anything. That's the spec rule too — writers must not commit against an unknown-transform spec. Once that's in, this is good to land.

table/update_spec.go
```diff
@@ -82,6 +82,12 @@ func NewUpdateSpec(t *Transaction, caseSensitive bool) *UpdateSpec {
 	nameToField := make(map[string]iceberg.PartitionField)
 	partitionSpec := t.tbl.Metadata().PartitionSpec()
 	for _, partitionField := range partitionSpec.Fields() {
+		if _, ok := partitionField.Transform.(iceberg.UnknownTransform); ok {
```
laskoviymishka, line 96: this went the opposite way from my round-1 ask, and I want to make sure that's intentional rather than a slip.

Round 1 I asked to validate only newly-added fields, so a table with an unknown transform on one column could still rename or drop an unrelated one. This rejects any evolution up front — the `UpdateSpec` is dead before a single op is queued.

If it's deliberate parity with Java's `BaseUpdatePartitionSpec`, I'm happy with it — simpler, clearer signal, and I'll take Java's precedent over my earlier ask. Just want to confirm it's the intended final direction. wdyt?

table/metadata_internal_test.go
```diff
@@ -309,6 +309,31 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+	spec := meta.PartitionSpec()
+	pf := spec.Field(0)
+	_, ok := pf.Transform.(iceberg.UnknownTransform)
+	assert.True(t, ok, "unknown partition transform should load")
+	assert.Equal(t, "custom_transform[42]", pf.Transform.String())
+
+	sf := meta.SortOrder().Field(1)
+	_, ok = sf.Transform.(iceberg.UnknownTransform)
+	assert.True(t, ok, "unknown sort transform should load")
+	assert.Equal(t, "custom_sort[7]", sf.Transform.String())
```
laskoviymishka, line 425: this is the full-metadata round-trip I asked for last round, thanks — it proves the load path doesn't trip a guard.

It only goes one direction though: `ParseMetadataBytes` then assert. The field-level `TestPartitionFieldUnknownTransformRoundTrip` covers marshal-back, so serialization itself is tested — but a marshal-back assert here (that the output still contains `custom_transform[42]`/`custom_sort[7]`) is the exact guardrail against `validateTransform` sneaking onto the write path that I was after. Cheap to add while we're here.

transforms_test.go
```diff
@@ -103,6 +93,45 @@ func TestParseTransform(t *testing.T) {
...
+		{"truncate extra token", "truncate_garbage[4]"},
+		{"preserves original case", "Custom_V2[3]"},
+	}
+
+	for _, tt := range unknownTests {
+		t.Run("unknown/"+tt.name, func(t *testing.T) {
+			tr, err := iceberg.ParseTransform(tt.toparse)
+			require.NoError(t, err)
+			_, ok := tr.(iceberg.UnknownTransform)
+			assert.True(t, ok)
```
laskoviymishka: tiny one: `assert.True(t, ok)` is non-fatal, so if `ParseTransform` ever regresses and returns a non-`UnknownTransform`, execution falls through to the bare `u := tr.(iceberg.UnknownTransform)` on line 130 and panics the goroutine instead of failing cleanly. `require.True` here fixes it.

laskoviymishka (changes requested):
This is close. The read side held up from round 1, and most of what I asked for landed — `TestMetadataV3ParsesUnknownTransforms` gives us the full-metadata load, the sort-order path now tolerates unknown transforms and matches Java (`TestUnmarshalUnknownSortTransform`, `TestNewSortOrderAcceptsUnknownTransform`), and `partitionField` propagates the `ParseTransform` error instead of swallowing it. Thanks for working through all of that.

One blocking thing left, and it's the write-path guard from last round — the null-partition one. A table loaded with an unknown partition transform still takes the partitioned write path (`IsUnpartitioned` only counts `VoidTransform`), `UnknownTransform.Apply` returns an empty Optional, and we'd silently write null partition values — or panic in `fileToDataFile` on `AddFiles`. I'd reject unknown partition transforms at `recordsToDataFiles`/`WriteRecords` before any bytes are written; that also matches the spec rule that writers must not commit against an unknown-transform spec. Left the mechanism inline.

One thing I want to flag rather than block on: the evolution guard went the opposite way from my round-1 ask. I asked to validate only new fields so unrelated evolution still works; the diff blanket-rejects any spec evolution on an unknown-transform table via the `NewUpdateSpec` guard. If that's deliberate parity with Java's `BaseUpdatePartitionSpec`, I'm happy to take Java's precedent over my earlier ask — just confirm it's the intended direction. Left a note on it.

So really just:
- the `WriteRecords`/`recordsToDataFiles` guard (the one blocker)
- a yes/no on the blanket-reject direction
- optionally, a marshal-back assert in the metadata round-trip test to close the guardrail

Fix the write guard and confirm the evolution direction and I'm happy to approve.

zeroshade (changes requested):
Thanks for adding unknown-transform round-tripping! A few validation gaps to address:

**Medium — empty transform string accepted** (`transforms.go:88`)
The unknown-transform fallback also accepts an empty string. Since partition/sort JSON unmarshal into a plain string, a missing/`null` `"transform"` becomes `UnknownTransform{name: ""}` instead of failing, so invalid metadata is accepted and round-tripped as `"transform": ""`.
*Fix:* reject empty transform strings; ideally have the JSON unmarshallers require `"transform"` present and non-null before calling `ParseTransform`.

**Medium — malformed reserved transforms fall through as unknown** (`transforms.go:52`)
`bucket[]`, `bucket[-1]`, `truncate[-1]` no longer match the numeric regex and fall through as unknown transforms, weakening validation for known Iceberg syntax.
*Fix:* detect reserved `bucket`/`truncate` forms separately and return `ErrInvalidTransform` when the bracketed argument is missing/negative/non-numeric.

**Medium — partition evolution can introduce unknown transforms** (`table/update_spec.go:85`)
The guard only rejects unknown transforms already in the base spec; callers can still `ParseTransform("custom[42]")` + `AddField`, and `BuildUpdates` will build a spec with an unknown transform whose `Apply` produces null partition values.
*Fix:* reject `UnknownTransform` in `UpdateSpec.AddField`/`partitionField` before appending to `us.adds`.

rambleraptor:
@zeroshade changes made! Thanks for your patience. I've been in-and-out on PTO.

One quick note - It looks like Java treats bucket[-1] as an unknown transform, not an error. I'm treating as an error now, but we can potentially open that up in the future.

table/update_spec.go
rambleraptor: Yeah, should've been more explicit about that. I did that for Java parity. Happy to change it back if you want.

rambleraptor:
@laskoviymishka thanks for your patience on this one! just got back from vacation and made those changes.

zeroshade (changes requested):
Gave this the full read against the v3 spec and the Java/PyIceberg references — the core is genuinely solid; I have one blocking item (`bucket[-1]` handling) plus the CI/rebase mechanics.

**What's right (spec-verified):**
- Read semantics match the spec — v3 *requires* readers to load unknown-transform tables and ignore those fields when filtering. `UnknownTransform.Project()→nil` (⇒ `AlwaysTrue`, no pruning), empty `Apply()`, and verbatim round-trip are exactly right; `TestUnknownTransformProjection` / `TestMetadataV3ParsesUnknownTransforms` pin it.
- Write rejection matches the spec ("writers are not allowed to commit … a spec with an unknown transform") — you guard AddFiles, WriteRecords, `UpdateSpec` (base + `addField`), and filter terms, each with a clear error. Thorough.
- The `fromString`→unknown fallback matches Java (`new UnknownTransform<>(...)`) and PyIceberg (`parse_transform`).

**Blocking — `bucket[-1]` should become `UnknownTransform`, not an error:**
`reservedWidthRegex` makes `bucket[-1]`, `bucket[]`, `bucket[abc]` (and the truncate equivalents) hard errors. Java's width pattern is `(\w+)\[(\d+)\]`, so those simply don't match and fall through to `UnknownTransform`; only `bucket[0]` (matches `\d+`, then fails `Bucket.get`'s `> 0` precondition) is an error. So Java is `bucket[0]`→error, `bucket[-1]`→**unknown**.

This matters because the feature exists to honor the v3 mandate that "all v3 readers are required to read tables with unknown transforms, ignoring them." With `reservedWidthRegex`, iceberg-go would *refuse to load* a table Java writes/reads as an unknown transform — the exact forward-compat failure this PR is meant to prevent. Let's fix it here rather than defer: drop `reservedWidthRegex`, keep the `> 0`/`≤ MaxInt32` validation on the digit-matching path (so `bucket[0]`/`truncate[0]` still error, exactly like Java's precondition), and let any bucket/truncate string that doesn't match the strict `\[(\d+)\]` width fall through to `UnknownTransform`. In the test split that moves `bucket[-1]`/`bucket[]`/`bucket[abc]` from `errorTests` into `unknownTests` (worth a round-trip assertion on one). It also simplifies the parser.

(PyIceberg does raise on `bucket[-1]`, so there's precedent for strictness — but it's the outlier here; Java + the spec's read-anything mandate are what matter for forward compat.)

**Mechanical (needed for CI/merge):**
- CI is red on one lint nit across all 4 jobs: `table/table_test.go:844` — `fmt.Sprintf("%s/partitioned_table/test-unknown.parquet", t.location)` trips `perfsprint`; make it `t.location + "/partitioned_table/test-unknown.parquet"`.
- Needs a rebase — #1508 (parameterized-transform validation + the `newSortOrder` `MarshalText` backstop) and #1527/#1548 have landed since. `UnknownTransform.MarshalText` returns no error, so it stays compatible with the sort-order check, but please confirm after rebase.

Everything else — parse / round-trip / projection / write-rejection / full-v3-metadata coverage — is excellent. Happy to re-review quickly once `bucket[-1]` is treated as unknown and CI is green.

laskoviymishka (approved):
👍

rambleraptor:
@laskoviymishka handled the merge conflicts! thanks for your patience. this should be ready to merge.

laskoviymishka:
Need @zeroshade approval to unblock merge here.

transforms.go
```diff
@@ -230,6 +246,49 @@ func (VoidTransform) Project(string, BoundPredicate) (UnboundPredicate, error) {
...
+	return Optional[Literal]{}
+}
+
+func (t UnknownTransform) ToHumanStr(any) string { return t.name }
+
+func (t UnknownTransform) ToHumanStrType(Type, any) string { return t.name }
```
zeroshade: **Blocking (part of the position-delete issue in the summary).** `ToHumanStrType` and `ToHumanStr` (`:283`) return the transform *name*, so they are constant for a given spec field regardless of the value passed in.

`PartitionSpec.PartitionToPath` (`partitions.go:710`) feeds this straight into the writer key at `table/rolling_data_writer.go:348`. Two genuinely different partition values therefore produce the identical path string, the map lookup returns the writer created for the first one, and the new `partitionValues` are discarded. The same collapse shows up in `table/snapshots.go:459,474`, where `partition-summaries` and `changed-partition-count` come out wrong.

Suggested fix: render the value rather than the transform name. Java's `Transform#toHumanString` default renders the value and `UnknownTransform` does not override it, so matching that behavior also fixes the summary counts and makes it impossible for any future `PartitionToPath` caller to collide.

transforms.go
```diff
@@ -230,6 +246,49 @@ func (VoidTransform) Project(string, BoundPredicate) (UnboundPredicate, error) {
 	return nil, nil
 }
 
+// UnknownTransform is a placeholder for a partition or sort transform that
+// this implementation doesn't recognize. The v3 spec requires readers to load
+// tables that use unknown transforms and to ignore those fields when
+// filtering; writers must not commit a partition spec that uses one.
+type UnknownTransform struct {
```
zeroshade, line 253: Non-blocking: `iceberg.UnknownTransform{}` is a legal composite literal from outside the package — all fields being unexported makes them unsettable, not the struct unconstructible. That yields `name == ""`, which marshals to `"transform": ""` and produces metadata that cannot be read back. `MarshalText` returning no error also bypasses the `newSortOrder` backstop at `table/sorting.go:321`.

Suggested fix: reject empty names in `validateTransform` (`partitions.go:314`), or return an error from `MarshalText` when `name` is empty.

transforms.go
```diff
@@ -230,6 +246,49 @@ func (VoidTransform) Project(string, BoundPredicate) (UnboundPredicate, error) {
...
+
+func (t UnknownTransform) MarshalText() ([]byte, error) {
+	return []byte(t.name), nil
+}
+
+func (t UnknownTransform) String() string { return t.name }
+
+// CanTransform assumes an unknown transform could apply to any type -- the
+// real applicability isn't known.
+func (UnknownTransform) CanTransform(Type) bool { return true }
```
zeroshade, line 273: Non-blocking: because `CanTransform` is unconditionally true, `SortOrder.CheckCompatibility` silently accepts any source type for an unknown transform. That is the right runtime behavior — we genuinely cannot know — but the doc comment reads as if compatibility were verified.

Suggested fix: note on this method that compatibility is *unverifiable* for unknown transforms, not verified.

expr_json.go
```diff
@@ -579,6 +579,11 @@ func decodeTerm(raw json.RawMessage) (UnboundTerm, error) {
 		if err != nil {
 			return nil, fmt.Errorf("%w: cannot parse transform term: %s", ErrInvalidArgument, err)
 		}
+		// Unknown transforms are tolerated in partition/sort metadata, but a
+		// filter expression referencing one can't be evaluated.
+		if _, ok := tf.(UnknownTransform); ok {
```
zeroshade, line 599: This rejection has no test coverage at all. It is a small addition, but it is the only thing standing between an unknown transform and a malformed REST filter payload.

Suggested fix: add a `decodeTerm` case feeding a term whose transform is unknown and assert `ErrInvalidArgument`.

table/sorting.go
```diff
@@ -249,6 +253,11 @@ func (s SortOrder) Len() int {
 	return len(s.fields)
 }
 
+// Field returns the sort field at index i.
+func (s SortOrder) Field(i int) SortField {
```
zeroshade, line 259: Non-blocking: the new `Field(i)` returns `s.fields[i]` raw while `Fields()` (`:242`) clones. `SortField.SourceIDs` is a slice, so a caller can mutate the sort order's internals through the returned value — which breaks exactly the invariant `TestSortOrderReturnsDefensiveCopies` asserts for `Fields()`.

Suggested fix: return `cloneSortField(s.fields[i])`, and document that an out-of-range index panics.

table/update_spec.go
```diff
@@ -86,6 +86,12 @@ func NewUpdateSpec(t *Transaction, caseSensitive bool) *UpdateSpec {
...
```
zeroshade, line 96: Worth surfacing to users: this blanket rejection blocks *all* spec evolution on a table with an unknown transform, including unrelated renames and drops. You confirmed this matches Java and it is the safer default, so no change to the behavior — but the rationale currently lives only in `update_spec_test.go:64`.

Suggested fix: move that reasoning into the `NewUpdateSpec` doc comment, where someone hitting the error will actually find it.

zeroshade (changes requested):
Thanks for taking this on — the read side of v3 unknown-transform tolerance is right, and I traced it end to end rather than taking the summary on faith.

Parsing preserves the original string with casing intact (`transforms.go:96`; only the match key is lowercased at `:58`), so `String()`/`MarshalText()` round-trip verbatim, matching Java's `Transforms.fromString`. Pruning bails out correctly: `UnknownTransform.Project` returns `(nil, nil)` (`transforms.go:288`) and `inclusiveProjection.VisitBound` (`table/evaluators.go:637-643`) skips nil, leaving `AlwaysTrue`. I followed every downstream consumer — `buildManifestEvaluator` (`table/scanner.go:468`) and `buildPartitionEvaluator` (`:483`) both consume the *projected* filter, so an unknown field can never reach manifest-summary or partition-record evaluation, and since there is no strict-projection path in this codebase the "nil must mean `AlwaysFalse` in strict mode" trap doesn't apply. The `bucket[0]` vs `bucket[-1]` split now matches Java exactly (`bucket[0]`/`truncate[0]` and int32 overflow are errors at `transforms.go:62`/`:71`; `bucket[-1]`, `bucket[]`, `bucket[abc]` fall through to `UnknownTransform`) — that matters, because erroring there would refuse a table Java reads fine. Sort orders carry no unknown-transform restriction in the spec, so tolerating them is also correct.

There is one blocking issue, and it is a silent data-correctness bug.

## Blocking: position-delete writes against an unknown-transform spec collapse distinct partitions into one delete file

`table/arrow_utils.go:1988` (`positionDeleteRecordsToDataFiles`) — the guard is missing here.

`checkNoUnknownTransform` covers `filesToDataFiles` (`table/arrow_utils.go:1718`) and `recordsToDataFiles` (`:1864`), but not `positionDeleteRecordsToDataFiles` (`:1988`), which is reached from `table/transaction.go:2131` for `Delete()` on format-version <= 2. `IsUnpartitioned()` only counts `VoidTransform`, so such a spec takes the fanout branch into `table/pos_delete_partitioned_fanout_writer.go:148`.

Combined with `ToHumanStr`/`ToHumanStrType` returning the transform *name* rather than the value (see the inline comment on `transforms.go:285`), the failure is concrete: a v2 table written by Java with an unknown partition transform — loadable only because of this PR — has rows deleted from partitions A and B. `PartitionSpec.PartitionToPath` (`partitions.go:710`) produces the *identical* path string for both, and that string is the writer key at `table/rolling_data_writer.go:348`, which returns the existing writer and discards the new `partitionValues`. One rolling writer is created keyed on A's `partitionData`, and B's position deletes land in a file whose manifest entry claims partition A. Scan planning pairs delete files to data files by partition, so B's deleted rows resurrect and A's live rows may be dropped. Nothing errors and nothing logs.

The spec is explicit here: *"Writers are not allowed to commit files with a partition spec that contains a field with an unknown transform"* — which covers delete files.

**Suggested fix (two parts):**

1. Add the guard to `positionDeleteRecordsToDataFiles` near the `latestMetadata` resolution (~`table/arrow_utils.go:2010`), checking the **historic** specs reachable via `partitionContextByFilePath[*].specID` rather than just `meta.CurrentSpec()` — a table that has evolved past the unknown transform still reaches this path. Guard the v3 DV path too for spec parity; it keys by data-file path so it is not corrupting, but the spec sentence applies equally.
2. Make `ToHumanStrType`/`ToHumanStr` render the value rather than the transform name, so no `PartitionToPath` caller can collide. Java's `Transform#toHumanString` default renders the value and `UnknownTransform` does not override it. Note that `table/snapshots.go:459,474` has the same collapse today (wrong `partition-summaries` and `changed-partition-count`), and that one persists on the v3 DV path even after part 1.

## Non-blocking observations outside the diff

- `table/transaction.go:864` — `AddDataFiles` is not covered by `checkNoUnknownTransform` either; `validateDataFilePartitionData` (`:689`) only checks value shape. Same spec rule, lower risk since the caller supplies the tuple. Suggested fix: apply the same guard.
- `partitions.go:742` — the default branch generates names like `id_custom_transform[42]`, embedding brackets in a field name. Reachable via `AddPartitionFieldBySourceID` with an empty name. Suggested fix: reject an empty name there rather than synthesizing one.
- `table/sort_keys.go:51` — `resolveSortKeys` ignores `field.Transform` and sorts the raw column, then stamps `SortOrderID`. Pre-existing (bucket and truncate have the same problem), but newly reachable through this PR. Worth a follow-up issue rather than a change here.
- Authoring new unknown transforms is rejected at `table/update_spec.go:253` (`AddField`) and `:89` (base spec), but *not* by `NewPartitionSpecOpts`, `table.NewMetadata`, or catalog `CreateTable`. That roughly matches Java — its builder allows it and the write throws — though the spec sentence is broader. Not blocking.

## Test coverage

What is covered is covered well: the parse-level unknown/error split with case preservation and `MarshalText` round-trip (`transforms_test.go`, 14 unknown plus 7 error cases), field-level `assert.JSONEq` round-trip (`partitions_test.go:258`), a full v3 metadata load and marshal-back (`table/metadata_internal_test.go:390`), the pruning bailout (`table/evaluators_test.go:1173`), sort-order load and construction, and write rejection at `AddFiles`, `WriteRecords`, `NewUpdateSpec`, and `AddField`.

The gaps, roughly in priority order:

1. **`Delete()` / position-delete writing on an unknown-transform spec.** This is the gap that let the blocking issue through.
2. **`PartitionToPath` / `ToHumanStrType` for `UnknownTransform`.** A single assertion on two *different* values would have caught the constant-key collapse.
3. **Mixed spec.** `unknownSpec()` (`table/evaluators_test.go:825`) contains only the unknown field. A spec with both `bucket[4](id)` and `custom[42](id)` — bucket still prunes, unknown contributes nothing — is the case that actually pins the behavior.
4. **End-to-end `PlanFiles` / `ToArrowTable`** asserting all rows return under a filter that would prune under identity. Everything today is unit-level.
5. **`TableWritingTestSuite` is instantiated only for formatVersion 1 and 2** (`table/table_test.go:2877-2878`), so `TestAddFilesToUnknownTransformSpecFails` never runs against v3 — the version where this behavior is actually required.
6. **`UnknownTransform.Equals`** is untested. Assert false against a different-named unknown, false against `IdentityTransform{}`, and case sensitivity; `checkForRedundantPartitions` (`partitions.go:327`) and `validatePartitionFields` (`:518`) both depend on it.
7. **Catalog `CreateTable` with an unknown-transform spec.** `TestPartitionSpecAcceptsUnknownTransform` pins that *building* succeeds, but nothing pins the commit outcome.

rambleraptor:
@zeroshade made those changes! I do want to note that I can't enable those v3 tests since they're broken for unrelated reasons. I opened up #1693 to make a quick fix to the test suite.

zeroshade (approved):
Looks good now! Thanks!

PR #1464, opened 2026-07-13
fix(manifest): preserve empty partition list on rewrite

### LINKED ISSUE #1309: `AddFiles` after a prior `Delete` snapshot produces manifest list unreadable by Redshift Spectrum
### Apache Iceberg version

main (development)

### Version

- **Working:** v0.5.0-rc (pseudo-version `v0.5.0-rc0.0.20260302222559-f192c8e504eb`)
- **Broken:** v0.6.0

### Summary

After upgrading from v0.5 to v0.6.0, calling `AddFiles()` on a table that has had a prior `Delete()` snapshot produces a manifest list file that Redshift Spectrum cannot parse:

```
ERROR: Wrong type in Avro file.
Field: partitions. Expected: 12. Got: 7
location: avro_utils.hpp:55
```

The Delete snapshot itself is readable. A standalone `AddFiles` on a fresh table (no prior Delete) is also readable. But when `AddFiles`' fast-append snapshot producer inherits manifests from a prior Delete (overwrite) snapshot into its manifest list, the resulting Avro encoding of the `partitions` field becomes incompatible with Redshift's reader.

Reverting to v0.5 resolves the issue.

### Table Setup

- Glue catalog (AWS)
- Format version: 2
- Unpartitioned table (also reproduces with partitioned)

### Reproduction Steps

**Step 1**: Create table, APPEND data via `AddFiles` → queryable via Redshift ✓

**Step 2**: `Delete()` with a filter (separate transaction) → queryable via Redshift ✓

**Step 3**: `AddFiles()` again (separate transaction, new data) → commit succeeds, but **Redshift returns XX000**

### Code

```go
// Transaction 1: initial append (works fine)
tbl, _ := catalog.LoadTable(ctx, identifier)
txn := tbl.NewTransaction()
txn.AddFiles(ctx, initialFiles, nil, false)
txn.Commit(ctx)

// Transaction 2: delete (works fine, queryable after)
tbl, _ = catalog.LoadTable(ctx, identifier)
txn = tbl.NewTransaction()
txn.Delete(ctx, iceberg.EqualTo(iceberg.Reference("status"), "DELETED"), nil)
txn.Commit(ctx)

// Transaction 3: append again (commit succeeds but Redshift can't read the result)
tbl, _ = catalog.LoadTable(ctx, identifier)
txn = tbl.NewTransaction()
txn.AddFiles(ctx, newFiles, nil, false)
txn.Commit(ctx)  // manifest list produced here is unreadable
```

### Test Matrix

| Scenario                                                                    | Result               |
| --------------------------------------------------------------------------- | -------------------- |
| `AddFiles` on fresh table (no prior Delete)                                 | **PASS** - queryable |
| `Delete` after prior `AddFiles` data (separate transactions)                | **PASS** -queryable  |
| `AddFiles` after prior `Delete` (separate transactions)                     | **FAIL** - XX000     |
| `Delete` + `AddFiles` in same transaction (table has prior `AddFiles` data) | **FAIL** - XX000     |

### Possible Root Cause

I suspect the issue may be in how the [`fastAppend`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L50) snapshot producer (used by [`AddFiles`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L980)) writes the manifest list when inheriting manifests from a prior [`mergeOverwrite`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L59) ([`Delete`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L1318)) snapshot.

[`Delete`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L1318) uses the [`mergeOverwrite`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L59) snapshot producer, which writes manifests that Redshift reads correctly. The subsequent [`AddFiles`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L980) uses [`appendSnapshotProducer`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L157), whose [`existingManifests()`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/snapshot_producers.go#L90) method inherits the parent snapshot's manifests. When the parent is a Delete/overwrite snapshot, the inherited manifests are included in a new manifest list file written by `fastAppend`.

The resulting manifest list Avro file has an encoding of the `partitions` field that Redshift's reader rejects (`Expected: 12. Got: 7`). I believe his only happens when the inherited manifests come from a `mergeOverwrite` producer. Inheriting from another `fastAppend` snapshot works fine.

### Potentially Relevant Changes (v0.5 → v0.6)

- [**PR #871**](https://github.com/apache/iceberg-go/pull/871): Avro library replacement (`hamba/avro` to `twmb/avro`). The OCF writer changed from `ocf.NewEncoderWithSchema` with `FullSchemaMarshaler` to `ocf.NewWriter` with `fileSchema.String()`
- [**PR #869**](https://github.com/apache/iceberg-go/pull/869): `fast-append must inherit all parent manifests unconditionally`. Changed [`fastAppendFiles.existingManifests()`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/snapshot_producers.go#L90) to no longer filter manifests
- [**PR #799**](https://github.com/apache/iceberg-go/pull/799)/[**#806**](https://github.com/apache/iceberg-go/pull/806): `AddFiles` now uses [`appendSnapshotProducer()`](https://github.com/apache/iceberg-go/blob/v0.6.0/table/transaction.go#L157) instead of hardcoded `fastAppend()`

The manifest list schema has a `partitions` field typed as `union[null, array<record>]`. When the fast-append writer serializes a manifest list that includes entries originally produced by the overwrite producer, the Avro schema JSON or binary encoding of this field differs from what Redshift expects (type 12 vs type 7).

### Additional Note

I tried querying via Athena instead of Redshift for one of the Delete+AddFiles same-transaction cases. Athena does not error, but returns **wrong data**. Columns from the appended file come back as NULL/empty even though the underlying parquet files contain values. Rows from the Delete's internal copy-on-write rewrite read correctly, but rows from the `AddFiles` portion do not. This further confirms something is wrong with the manifest metadata produced by `AddFiles` when inheriting from a Delete snapshot.

### Environment

- Go 1.26
- AWS Glue catalog
- Queried via Amazon Redshift Spectrum
- Table format version 2
- Parquet files with embedded Iceberg field IDs

Revanth14:
Fixes #1309

## Summary

- preserve present-but-empty `partitions` arrays when inherited manifest-list entries are rewritten
- keep genuinely absent partition summaries encoded as Avro null
- leave populated partition summaries unchanged
- add unit and end-to-end coverage for `AddFiles` after a copy-on-write `Delete`

The regression occurred because an empty Avro array decoded as a non-nil pointer containing a nil Go slice. Re-encoding that value selected the null union branch, which strict readers such as Redshift Spectrum reject.

## Recovery note

This fix prevents new manifest-list rewrites from emitting null `partitions` values, but it does not repair manifest lists that a pre-fix iceberg-go version already wrote with null. Tables that already hit this issue must expire or recommit the affected snapshots so their manifest lists are rewritten before strict readers can consume them.

## Testing

- `go vet . ./table`
- `go test . -run '^TestManifests$' -count=1`
- `go test ./table -count=1 -timeout=3m`

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -306,6 +306,27 @@ func (m *manifestFile) toV1(v1file *manifestFileV1) {
 	}
 }
 
+// ensurePartitionList normalizes the narrow case behind #1309: an inherited
+// manifest whose partition-field summary list is present but empty. Only an
+// empty-spec manifest has zero summaries, so in practice this is an
+// unpartitioned table. On decode that present-empty array becomes a non-nil
+// pointer over a nil slice, which the manifest-list writer would otherwise
+// re-encode as Avro null for the optional partitions union (field 507) — a
+// value strict readers such as Redshift Spectrum reject. Only that
+// present-but-nil slice is normalized back to an empty array. A genuinely
+// absent value (nil pointer) is left null, which is spec-legal because
+// partitions is optional; populated (partitioned) summaries are non-empty and
+// pass through untouched.
+func ensurePartitionList(partitions **[]FieldSummary) {
+	if *partitions == nil {
+		return
+	}
+	if **partitions == nil {
+		empty := []FieldSummary{}
+		*partitions = &empty
+	}
+}
+
 func (m *manifestFile) Version() int                     { return m.version }
 func (m *manifestFile) FilePath() string                 { return m.Path }
 func (m *manifestFile) Length() int64                    { return m.Len }
@@ -1531,6 +1552,7 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 		var tmp manifestFileV1
 		for _, file := range files {
 			file.(*manifestFile).toV1(&tmp)
+			ensurePartitionList(&tmp.PartitionList)
 			if err := m.writer.Encode(&tmp); err != nil {
 				return err
 			}
@@ -1553,6 +1575,7 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 			}
 
 			wrapped := *(file.(*manifestFile))
+			ensurePartitionList(&wrapped.PartitionList)
 			if m.version == 3 {
 				// Ref: https://github.com/apache/iceberg/blob/ea2071568dc66148b483a82eefedcd2992b435f7/core/src/main/java/org/apache/iceberg/ManifestListWriter.java#L157-L168
 				if wrapped.Content == ManifestContentData {
--- a/table/manifest_list_partitions_regression_test.go
+++ b/table/manifest_list_partitions_regression_test.go
@@ -1,0 +1,213 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table_test
+
+// End-to-end regression test for issue #1309: AddFiles() after a prior Delete()
+// produces a manifest list whose `partitions` field serializes as Avro null,
+// which Redshift Spectrum rejects ("Wrong type in Avro file. Field: partitions.
+// Expected: 12 (union). Got: 7 (null)").
+//
+// The trigger is the inheritance path: Delete() uses the mergeOverwrite producer
+// and the following AddFiles() uses fastAppend, which inherits the delete
+// snapshot's manifests unchanged. When the manifest list is rewritten, an
+// inherited manifest whose `partitions` summaries are present-but-empty (an
+// unpartitioned table has zero field summaries) round-trips as a non-nil pointer
+// to a nil slice, which the writer previously encoded as null instead of a
+// present empty array.
+//
+// The fix (ensurePartitionList in manifest.go) normalizes a present-but-nil
+// summary list back to an empty array so the union resolves to the array branch.
+// This test asserts every record in the final manifest list encodes `partitions`
+// as a present array, never null.
+
+import (
+	"context"
+	"errors"
+	"io"
+	"path/filepath"
+	"testing"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet/pqarrow"
+	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
+	"github.com/apache/iceberg-go/table"
+	"github.com/stretchr/testify/require"
+	"github.com/twmb/avro/ocf"
+)
+
+// manifestListRecordV2 mirrors the version-2 manifest-list entry schema
+// (internal/avro_schemas.go, "manifest_list_file_v2"). Only PartitionList is
+// inspected, but the full record is declared so the reader schema derived from
+// these struct tags matches the writer schema field-for-field. PartitionList is
+// decoded as `any` so a null union member surfaces as a nil interface and a
+// present array surfaces as a non-nil []any.
+type manifestListRecordV2 struct {
+	Path               string `avro:"manifest_path"`
+	Len                int64  `avro:"manifest_length"`
+	SpecID             int32  `avro:"partition_spec_id"`
+	Content            int32  `avro:"content"`
+	SeqNumber          int64  `avro:"sequence_number"`
+	MinSeqNumber       int64  `avro:"min_sequence_number"`
+	AddedSnapshotID    int64  `avro:"added_snapshot_id"`
+	AddedFilesCount    int32  `avro:"added_files_count"`
+	ExistingFilesCount int32  `avro:"existing_files_count"`
+	DeletedFilesCount  int32  `avro:"deleted_files_count"`
+	PartitionList      any    `avro:"partitions"`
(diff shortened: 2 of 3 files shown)


manifest.go
```diff
@@ -1531,6 +1552,7 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 		var tmp manifestFileV1
 		for _, file := range files {
 			file.(*manifestFile).toV1(&tmp)
+			ensurePartitionList(&tmp.PartitionList)
```
tanmayrauth, line 1555: The v1 branch fix here has no test that fails without it. `TestManifestListV1WriterPartitionListEncoding` builds its manifests fresh via `Partitions([]FieldSummary{})`, which stores a non-nil empty slice, so `ensurePartitionList` is a no-op on that input — the test passes even with this line removed. The v1 collapse only reproduces on a decode→rewrite (a read-back empty array decodes to a nil slice), same as the v2 case the e2e test covers. As written, a  future refactor that drops this line reintroduces the null-collapse for v1 manifest-list rewrites with the suite still green. A v1 analogue of  `TestManifestListWriterPreservesEmptyPartitionArrayWhenRewritingInheritedManifest` — read a v1 manifest list back, then rewrite it as v1, would close the gap.

tanmayrauth (commented):
Fix is correct and confirmed against the issue — reverting `ensurePartitionList` makes both the unit and e2e tests fail as expected. One test-gap on the v1 path, details inline.

manifest.go
```diff
@@ -306,6 +306,27 @@ func (m *manifestFile) toV1(v1file *manifestFileV1) {
...
+// empty-spec manifest has zero summaries, so in practice this is an
+// unpartitioned table. On decode that present-empty array becomes a non-nil
+// pointer over a nil slice, which the manifest-list writer would otherwise
+// re-encode as Avro null for the optional partitions union (field 507) — a
+// value strict readers such as Redshift Spectrum reject. Only that
+// present-but-nil slice is normalized back to an empty array. A genuinely
+// absent value (nil pointer) is left null, which is spec-legal because
+// partitions is optional; populated (partitioned) summaries are non-empty and
+// pass through untouched.
+func ensurePartitionList(partitions **[]FieldSummary) {
```
laskoviymishka, line 320: The `**[]FieldSummary` works and it's well-commented, but a value-in/value-out form reads cleaner at both call sites — `tmp.PartitionList = ensurePartitionList(tmp.PartitionList)`. Not blocking, just a preference. Thoughts?

manifest.go
```diff
@@ -306,6 +306,27 @@ func (m *manifestFile) toV1(v1file *manifestFileV1) {
...
+	if *partitions == nil {
+		return
+	}
+	if **partitions == nil {
+		empty := []FieldSummary{}
+		*partitions = &empty
```
laskoviymishka, line 326: I'd add a one-liner that the pointer swap here (rather than `**partitions = empty`) is deliberate. Both `toV1` and the `wrapped := *file` copy share the underlying slice with the source manifest, so swapping the pointer leaves the original untouched — but writing through it would mutate the inherited manifest and break the next commit that reads it. That's subtle enough that someone could "simplify" it and reintroduce aliasing. wdyt?

table/manifest_list_partitions_regression_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/twmb/avro/ocf"
+)
+
+// manifestListRecordV2 mirrors the version-2 manifest-list entry schema
+// (internal/avro_schemas.go, "manifest_list_file_v2"). Only PartitionList is
+// inspected, but the full record is declared so the reader schema derived from
+// these struct tags matches the writer schema field-for-field. PartitionList is
+// decoded as `any` so a null union member surfaces as a nil interface and a
+// present array surfaces as a non-nil []any.
+type manifestListRecordV2 struct {
```
laskoviymishka, line 62: This is a field-for-field duplicate of `manifestListPartitionsRecord` in `manifest_test.go` (different package, so it can't just be shared). Not a blocker, but a schema change now has to touch both or they silently drift — I'd at least put a comment on each pointing at the other. wdyt?

table/manifest_list_partitions_regression_test.go
```diff
@@ -0,0 +1,213 @@
...
+	col := bldr.NewArray()
+	defer col.Release()
+
+	rec := array.NewRecordBatch(arrowSchema, []arrow.Array{col}, int64(len(ids)))
+	defer rec.Release()
+
+	arrTbl := array.NewTableFromRecords(arrowSchema, []arrow.RecordBatch{rec})
+	defer arrTbl.Release()
+
+	fo, err := fs.Create(path)
```
laskoviymishka, line 208: `fo` never gets closed — `pqarrow.WriteTable` closes its own parquet writer but not the underlying `FileWriter`. It's test-only and `t.TempDir()` cleans the dir, so the leak is hidden, but on Windows the subsequent `AddFiles` could read a partially-flushed file. I'd add `defer func() { require.NoError(t, fo.Close()) }()` right after `Create`.

laskoviymishka (approved):
Nice — this is a clean, well-scoped fix for #1309, and the tests are careful about the right thing: they assert the raw Avro union value rather than the normalized `Partitions()` accessor, which would have masked the bug. tanmayrauth already confirmed it's red-green by reverting `ensurePartitionList`, so I'm confident the fix is correct and at the right layer.

No code blockers from me — a few small things before merge.

The one I'd most want addressed is a doc note, not code: the fix stops us emitting new null partitions, but it doesn't heal manifest lists a pre-fix iceberg-go already wrote with null — including the delete snapshot's own manifest list in the #1309 scenario. Anyone already hit by this will need to expire/re-commit those snapshots to recover. Worth a line in the PR description (and maybe the doc comment on `ensurePartitionList`) so that expectation is explicit.

Then a couple of tidy-ups inline: the `writeInt64Parquet` helper leaks its file handle, and I'd add a one-liner explaining why `ensurePartitionList` swaps the pointer instead of writing through it — it's load-bearing for the aliasing and easy to "simplify" into a bug. tanmayrauth's v1-path test gap is worth folding in while we're here too.

Fix those and I'm happy to merge.

zeroshade (approved):
LGTM, thanks!

PR #1465, opened 2026-07-14
feat(table): add history metadata table

laskoviymishka:
Add an inspect subsystem (Table.Inspect) and its first metadata table, History, mirroring the Java, PyIceberg, and Rust clients.

History exposes the snapshot log as an Arrow RecordReader with columns made_current_at (timestamp[us,UTC]), snapshot_id, parent_id, and is_current_ancestor (field IDs 1-4 for cross-client parity). Ancestry is computed by walking the current snapshot's parent chain via the existing AncestorsOf helper (cycle-guarded); rolled-back snapshots are flagged non-ancestors and expired snapshots render a null parent_id. The context is honored during iteration and errors are wrapped.

### DIFF
--- a/table/inspect.go
+++ b/table/inspect.go
@@ -1,0 +1,138 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1466, opened 2026-07-15
fix(sql): keep namespace drops atomic

fallintoplace:
## Summary

- check namespace existence, catalog-object emptiness, and property deletion in one serializable write transaction
- treat views and future object types as namespace contents instead of relying on the table-only `ListTables` result
- recheck source and destination namespaces inside serializable create, commit, and rename transactions so concurrent drops cannot leave orphaned objects
- reload namespace properties inside the serializable update transaction so a concurrent drop cannot recreate a namespace
- retry recognized serialization, deadlock, and database-lock failures with bounded, context-aware backoff
- keep view-only namespaces intact and return the object-neutral `ErrNamespaceNotEmpty` sentinel

## Why

On the V1 SQL schema, `ListTables` intentionally excludes views. `DropNamespace` used that filtered result to decide whether a namespace was empty, so a namespace containing only a view could report a successful drop while the view row—and therefore the namespace—still existed.

The old prechecks also left races with object creation, rename, and namespace property updates. Drop, creators, renames, and property updates now participate in serializable transactions and recheck the relevant namespaces in the same transaction as their writes. Property updates reload the current rows and recompute their update summary on every retry, preventing stale results and namespace resurrection.

Database engines can abort serializable transactions under contention. The retry classifier handles PostgreSQL SQL states, typed MySQL deadlock and lock-timeout errors, SQLite lock errors from the actual driver, and Oracle serialization and deadlock errors. Exhausted retries return the database error directly. The `NewCatalog` documentation also calls out the WAL and busy-timeout configuration expected for concurrent SQLite writers.

## Testing

- view-only namespace regression
- deterministic drop races covering `CreateTable`, `CreateView`, and the create branch of `CommitTable`
- deterministic drop versus namespace-property-update regression
- real SQLite driver lock classification plus PostgreSQL, MySQL, and Oracle retry classification
- retry exhaustion and cancellation coverage
- `go test ./catalog/sql`
- `go test -race ./catalog/sql`
- `go vet ./catalog/sql`

### DIFF
--- a/catalog/sql/drop_namespace_internal_test.go
+++ b/catalog/sql/drop_namespace_internal_test.go
@@ -1,0 +1,384 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package sql
+
+import (
+	"context"
+	dbsql "database/sql"
+	"errors"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/catalog"
+	_ "github.com/apache/iceberg-go/io/gocloud"
+	"github.com/apache/iceberg-go/table"
+	mysql "github.com/go-sql-driver/mysql"
+	"github.com/stretchr/testify/require"
+	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/driver/sqliteshim"
+)
+
+type namespaceRaceContextKey struct{}
+
+type namespaceRaceOperation string
+
+const (
+	dropNamespaceOperation    namespaceRaceOperation = "drop"
+	createObjectOperation     namespaceRaceOperation = "create-object"
+	updatePropertiesOperation namespaceRaceOperation = "update-properties"
+)
+
+type namespaceRaceHook struct {
+	dropAtDelete       chan struct{}
+	createAtInsert     chan struct{}
+	updateAtInsert     chan struct{}
+	releaseDrop        chan struct{}
+	dropDone           chan struct{}
+	releaseUpdate      chan struct{}
+	dropAtDeleteOnce   sync.Once
+	createAtInsertOnce sync.Once
+	updateAtInsertOnce sync.Once
+}
+
+func (h *namespaceRaceHook) BeforeQuery(ctx context.Context, event *bun.QueryEvent) context.Context {
+	operation, _ := ctx.Value(namespaceRaceContextKey{}).(namespaceRaceOperation)
+	query := event.Query
+
+	switch {
+	case operation == dropNamespaceOperation && event.Operation() == "DELETE" && strings.Contains(query, "iceberg_namespace_properties"):
+		h.dropAtDeleteOnce.Do(func() { close(h.dropAtDelete) })
+		<-h.releaseDrop
+	case operation == createObjectOperation && event.Operation() == "INSERT" && strings.Contains(query, "iceberg_tables"):
+		h.createAtInsertOnce.Do(func() { close(h.createAtInsert) })
+		<-h.releaseDrop
+		<-h.dropDone
+	case operation == updatePropertiesOperation && event.Operation() == "INSERT" && strings.Contains(query, "iceberg_namespace_properties"):
+		h.updateAtInsertOnce.Do(func() { close(h.updateAtInsert) })
+		<-h.releaseUpdate
+	}
+
+	return ctx
+}
+
+func (*namespaceRaceHook) AfterQuery(context.Context, *bun.QueryEvent) {}
+
+func newNamespaceRaceCatalog(t *testing.T) *Catalog {
+	t.Helper()
+
+	dbPath := filepath.Join(t.TempDir(), "catalog.db")
+	db, err := dbsql.Open(sqliteshim.ShimName, dbPath)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, db.Close()) })
+	db.SetMaxOpenConns(4)
+	_, err = db.Exec("PRAGMA journal_mode=WAL")
+	require.NoError(t, err)
+
+	warehouse := filepath.Join(t.TempDir(), "warehouse")
+	cat, err := NewCatalog("default", db, SQLite, iceberg.Properties{
+		"warehouse": "file://" + filepath.ToSlash(warehouse),
+	})
+	require.NoError(t, err)
+
+	return cat
+}
+
+type sqlStateTestError string
+
+func (e sqlStateTestError) Error() string    { return string(e) }
+func (e sqlStateTestError) SQLState() string { return string(e) }
+
+func TestRetrySerializableWriteTx(t *testing.T) {
+	t.Run("retries transient errors", func(t *testing.T) {
+		attempts := 0
+		err := retrySerializableWriteTx(context.Background(), func() error {
+			attempts++
+			if attempts == 1 {
+				return sqlStateTestError("40001")
+			}
+
+			return nil
+		})
+		require.NoError(t, err)
+		require.Equal(t, 2, attempts)
+	})
+
+	t.Run("does not retry permanent errors", func(t *testing.T) {
+		attempts := 0
+		permanentErr := errors.New("permanent error")
+		err := retrySerializableWriteTx(context.Background(), func() error {
+			attempts++
+
+			return permanentErr
+		})
+		require.ErrorIs(t, err, permanentErr)
+		require.Equal(t, 1, attempts)
+	})
+
+	t.Run("returns transient error after attempts are exhausted", func(t *testing.T) {
+		attempts := 0
+		transientErr := sqlStateTestError("40001")
+		err := retrySerializableWriteTx(context.Background(), func() error {
+			attempts++
+
+			return transientErr
+		})
+		require.ErrorIs(t, err, transientErr)
+		require.Equal(t, serializableWriteMaxAttempts, attempts)
+	})
+
+	t.Run("stops when context is canceled", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		attempts := 0
+		err := retrySerializableWriteTx(ctx, func() error {
+			attempts++
+
+			return sqlStateTestError("40001")
+		})
+		require.ErrorIs(t, err, context.Canceled)
+		require.Equal(t, 1, attempts)
+	})
+}
+
+func TestRetryableSerializableError(t *testing.T) {
+	tests := []struct {
+		name string
+		err  error
+	}{
+		{name: "postgres serialization", err: sqlStateTestError("40001")},
+		{name: "postgres deadlock", err: sqlStateTestError("40P01")},
+		{name: "mysql deadlock", err: &mysql.MySQLError{Number: 1213}},
+		{name: "mysql lock timeout", err: &mysql.MySQLError{Number: 1205}},
+		{name: "sqlite3 busy", err: errors.New("database is locked")},
+		{name: "oracle serialization", err: errors.New("ORA-08177: can't serialize access for this transaction")},
+		{name: "oracle deadlock", err: errors.New("ORA-00060: deadlock detected while waiting for resource")},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require.True(t, isRetryableSerializableError(tt.err))
+		})
+	}
+
+	require.False(t, isRetryableSerializableError(errors.New("constraint violation")))
+}
+
+func TestRetryableSerializableErrorFromSQLiteDriver(t *testing.T) {
+	dbPath := filepath.Join(t.TempDir(), "locked.db")
+	first, err := dbsql.Open(sqliteshim.ShimName, dbPath)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, first.Close()) })
+	second, err := dbsql.Open(sqliteshim.ShimName, dbPath)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, second.Close()) })
+
+	first.SetMaxOpenConns(1)
+	second.SetMaxOpenConns(1)
+	_, err = first.Exec("PRAGMA busy_timeout=0")
+	require.NoError(t, err)
+	_, err = second.Exec("PRAGMA busy_timeout=0")
+	require.NoError(t, err)
+	_, err = first.Exec("CREATE TABLE locks (id INTEGER PRIMARY KEY)")
+	require.NoError(t, err)
+
+	tx, err := first.Begin()
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = tx.Rollback() })
+	_, err = tx.Exec("INSERT INTO locks VALUES (1)")
+	require.NoError(t, err)
+
+	_, lockErr := second.Exec("INSERT INTO locks VALUES (2)")
+	require.Error(t, lockErr)
+	require.True(t, isRetryableSerializableError(lockErr), "unexpected SQLite lock error type %T: %v", lockErr, lockErr)
+}
+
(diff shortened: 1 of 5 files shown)


catalog/sql/sql.go
```diff
@@ -847,33 +887,33 @@ func (c *Catalog) DropNamespace(ctx context.Context, namespace table.Identifier)
...
-	if !exists {
-		return fmt.Errorf("%w: %s", catalog.ErrNoSuchNamespace, nsToDelete)
-	}
-
-	tbls := make([]table.Identifier, 0)
-	iter := c.ListTables(ctx, namespace)
-
-	for tbl, err := range iter {
-		tbls = append(tbls, tbl)
+	return withSerializableWriteTx(ctx, c.db, func(ctx context.Context, tx bun.Tx) error {
```
tanmayrauth, line 1194: The serializable txn gives you the mutual exclusion, but the failure mode it introduces isn't handled anywhere. On Postgres SERIALIZABLE one txn aborts with SQLSTATE 40001 at commit; on MySQL SERIALIZABLE the two existence SELECTs become LOCK IN SHARE MODE and the read-then-write pair deadlocks (ER_LOCK_DEADLOCK); and on the default production SQLite catalog — which sets neither WAL nor busy_timeout (those are only in the test) — two concurrent writers get SQLITE_BUSY immediately. In all three cases CreateTable/CreateView/DropNamespace/staged-CommitTable return that raw driver error wrapped in none of the catalog sentinels, so a caller doing errors.Is(err,  table.ErrCommitFailed) or ErrNamespaceNotEmpty to retry never matches and just sees an opaque BUSY/deadlock. That's inconsistent with the update path at sql.go:625, which wraps its conflict in table.ErrCommitFailed precisely so Table.doCommit retries. These new serializable paths should either retry on the transient serialization error or wrap it in a recognized sentinel.

catalog/sql/sql.go
```diff
@@ -847,33 +887,33 @@ func (c *Catalog) DropNamespace(ctx context.Context, namespace table.Identifier)
...
 
-		break // there is already at least a table
-	}
+		propsExist, err := tx.NewSelect().Model((*sqlIcebergNamespaceProps)(nil)).
+			ColumnExpr("1").
+			Where("catalog_name = ?", c.name).
+			Where("namespace = ?", nsToDelete).
+			Limit(1).Exists(ctx)
+		if err != nil {
+			return fmt.Errorf("error checking namespace '%s' properties: %w", namespace, err)
```
tanmayrauth: `namespace` here is a table.Identifier ([]string), so '%s' prints `namespace '[foo]'` while the rest of the function uses the joined nsToDelete (`foo`). Use nsToDelete in both messages so the error text is consistent and not bracketed.

tanmayrauth (commented):
Core fix looks right, the direct row count fixes the view-only namespace bug and the in-tx recheck closes the orphan race. One substantive point on how serialization failures surface, plus two nits. Details inline.

zeroshade (changes requested):
Thanks for tightening this up. One remaining gap before merge:

**Medium — namespace property update can resurrect a dropped namespace** (`catalog/sql/sql.go:952`)

`DropNamespace` is now serialized against the table/view creation paths, but namespace property updates still do their existence check *before* a non-serializable write transaction. A concurrent `UpdateNamespaceProperties` (with updates) can read the namespace before the drop, then insert property rows after the drop commits — making both operations report success and resurrecting the namespace.

**Suggested fix:** move `UpdateNamespaceProperties` onto the same serializable transaction pattern and recheck namespace existence inside the transaction before applying deletes/inserts. A race test for drop vs. namespace-property-update would help guard this.

catalog/sql/sql.go
```diff
@@ -293,6 +294,72 @@ func withWriteTx(ctx context.Context, db *bun.DB, fn func(context.Context, bun.T
...
+
+func isRetryableSerializableError(err error) bool {
+	var stateErr interface{ SQLState() string }
+	if errors.As(err, &stateErr) {
+		switch stateErr.SQLState() {
+		case "40001", "40P01":
+			return true
+		}
+	}
+
```
laskoviymishka, line 350: I think this `interface{ Code() int }` branch never matches the real driver. mattn/go-sqlite3 implements `Code() sqlite3.ErrNo` and modernc.org/sqlite (what `sqliteshim` wraps) uses its own named `ErrorCode` — both are `int` underneath, but interface satisfaction is by exact signature, so `Code() ErrNo` doesn't satisfy `Code() int`. `errors.As` falls through and we're left relying on the `database is locked` string match.

The unit test only passes because `sqliteCodeTestError.Code()` returns a bare `int`, so it exercises a shape the real driver never produces. Net effect: the SQLITE_BUSY/LOCKED retry doesn't actually fire for a production SQLite catalog.

I'd switch this to `errors.As` against the concrete driver error type (or drop the `Code()` branch and lean on the string matches, with a comment on why), and add a test that runs a real driver error through `isRetryableSerializableError` so this can't silently regress. wdyt?

catalog/sql/sql.go
```diff
@@ -293,6 +294,72 @@ func withWriteTx(ctx context.Context, db *bun.DB, fn func(context.Context, bun.T
...
+	var sqliteErr interface{ Code() int }
+	if errors.As(err, &sqliteErr) {
+		switch sqliteErr.Code() & 0xff {
+		case 5, 6: // SQLITE_BUSY, SQLITE_LOCKED
+			return true
+		}
+	}
+
+	msg := err.Error()
+
```
laskoviymishka, line 360: These MySQL matches key off the rendered error string (`Error 1213 (...`), which is go-sql-driver/mysql's formatting rather than a stable contract — a driver bump, a proxy, or localized text breaks classification silently.

I'd match on the typed error instead: `errors.As` against `*mysql.MySQLError` and check `Number == 1213 || Number == 1205`, keeping the substring only as a commented fallback.

catalog/sql/sql.go
```diff
@@ -293,6 +294,72 @@ func withWriteTx(ctx context.Context, db *bun.DB, fn func(context.Context, bun.T
...
+	return strings.Contains(msg, "Error 1213 (") ||
+		strings.Contains(msg, "Error 1205 (") ||
+		strings.Contains(msg, "database is locked") ||
+		strings.Contains(msg, "database table is locked") ||
```
laskoviymishka, line 364: While we're matching Oracle errors — under SERIALIZABLE, lock-cycle deadlocks come back as ORA-00060, not ORA-08177, and Oracle kills one of the waiters. That victim is exactly the transaction we'd want to retry, but right now it surfaces as a permanent error.

I'd add `strings.Contains(msg, "ORA-00060")` alongside the ORA-08177 check.

catalog/sql/sql.go
```diff
@@ -293,6 +294,72 @@ func withWriteTx(ctx context.Context, db *bun.DB, fn func(context.Context, bun.T
 	})
 }
 
+const (
```
laskoviymishka, line 304: My concern with this retry budget is that it only helps if the DB actually blocks and then hands back a retryable error. On the production SQLite path it won't: `NewCatalog` takes a pre-configured `*sql.DB` with no WAL and no busy_timeout, so a contended writer gets SQLITE_BUSY back immediately, and three attempts at 10/20ms burn through the budget faster than a single disk round-trip. The race tests only survive because they set `PRAGMA journal_mode=WAL` and `SetMaxOpenConns(4)`, which isn't how the catalog is actually built.

So the retry reads as transparent but for the default SQLite catalog it mostly isn't. I'd either set WAL + a busy_timeout for SQLite in `NewCatalog` (or use `BEGIN IMMEDIATE` instead of SERIALIZABLE there), or document that those pragmas are required for the retry to mean anything. Thoughts?

catalog/sql/sql.go
```diff
@@ -745,6 +852,10 @@ func (c *Catalog) CommitTable(ctx context.Context, ident table.Identifier, reqs
 		return nil
 	})
 	if err != nil {
+		if current == nil && isRetryableSerializableError(err) {
+			return nil, "", fmt.Errorf("%w: %w", table.ErrCommitFailed, err)
```
laskoviymishka: Two things on this wrap. `fmt.Errorf("%w: %w", ...)` makes the result satisfy both `errors.Is(_, table.ErrCommitFailed)` and `errors.Is(_, err)` — so the raw driver serialization error becomes a first-class sentinel, and a driver swap silently changes what callers match. I'd wrap only the sentinel: `fmt.Errorf("%w: %v", table.ErrCommitFailed, err)`.

Separately, handing the exhausted serialization failure back as `ErrCommitFailed` re-enters `table.doCommit`'s retry loop, which re-runs `withSerializableWriteTx` for another 3 inner attempts — up to 9 total. That outer loop exists for CAS-on-metadata conflicts, and on the create branch (`current == nil`) there's no version skew to resolve anyway. I'd return the serialization error directly, or under a distinct sentinel, rather than conflating "someone beat me to the CAS" with "the DB couldn't serialize under load."

catalog/sql/sql.go
```diff
@@ -869,7 +980,15 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
 		return nil, fmt.Errorf("%w: %s", catalog.ErrNoSuchNamespace, toNs)
 	}
 
-	err = withWriteTx(ctx, c.db, func(ctx context.Context, tx bun.Tx) error {
+	err = withSerializableWriteTx(ctx, c.db, func(ctx context.Context, tx bun.Tx) error {
+		namespaceExists, err := c.namespaceExistsInTx(ctx, tx, toNs)
```
laskoviymishka: This rechecks the destination namespace inside the tx, which is the race the PR is closing — but the source side still relies on the UPDATE's RowsAffected, and a concurrent DropNamespace of the *source* namespace only deletes its property rows, not the `iceberg_tables` row. So the UPDATE still matches and we'd leave a table row pointing at a namespace that no longer exists.

Narrower than the to-namespace case, but since we're here I'd add a `namespaceExistsInTx(ctx, tx, fromNs)` check before the UPDATE too.

catalog/sql/drop_namespace_internal_test.go
```diff
@@ -0,0 +1,302 @@
...
+	switch {
+	case operation == dropNamespaceOperation && event.Operation() == "DELETE" && strings.Contains(query, "iceberg_namespace_properties"):
+		h.dropAtDeleteOnce.Do(func() { close(h.dropAtDelete) })
+		<-h.releaseDrop
+	case operation == createViewOperation && event.Operation() == "INSERT" && strings.Contains(query, "iceberg_tables"):
+		h.createAtInsertOnce.Do(func() { close(h.createAtInsert) })
+		<-h.releaseDrop
+		// Give the namespace deletion the first opportunity to commit. A creator
+		// that did not recheck inside its transaction would then insert an orphan.
+		time.Sleep(50 * time.Millisecond)
```
laskoviymishka: This 50ms sleep (plus the 250ms select timeout further down) is the load-bearing synchronization for the create/drop race, and it's wall-clock. On a slow CI runner the drop may not reach commit within 50ms, in which case create wins and we quietly exercise the create-wins path instead of the drop-wins one we actually care about.

The new update-properties test uses a clean release channel to sequence things deterministically — I'd bring the same idea here: a channel closed by the drop goroutine right after its `withSerializableWriteTx` returns, with the creator waiting on that instead of sleeping.

catalog/sql/drop_namespace_internal_test.go
```diff
@@ -0,0 +1,302 @@
...
+	dropErr := <-dropErrCh
+	createErr := <-createErrCh
+	require.False(t, dropErr == nil && createErr == nil, "drop and create must not both report success")
+
+	viewExists, err := cat.CheckViewExists(context.Background(), viewID)
+	require.NoError(t, err)
+	namespaceExists, err := cat.CheckNamespaceExists(context.Background(), namespace)
+	require.NoError(t, err)
+
+	if createErr == nil {
```
laskoviymishka: The `if createErr == nil / else if dropErr == nil` pair has no `else`, so when SQLite serialization makes *both* fail, the test runs no state assertions and passes vacuously — an orphan (view row present, namespace props gone) with both erroring would slip right through.

I'd add the else and assert self-consistency there: namespace exists iff the view exists. If both-fail is an accepted outcome of the retry design, a `t.Logf` of both errors plus that consistency check would still guard it.

catalog/sql/drop_namespace_internal_test.go
```diff
@@ -0,0 +1,302 @@
...
+	require.False(t, isRetryableSerializableError(errors.New("constraint violation")))
+}
+
+func TestDropNamespaceDoesNotRaceWithCreateView(t *testing.T) {
+	cat := newNamespaceRaceCatalog(t)
+
+	namespace := table.Identifier{"race_namespace"}
+	viewID := table.Identifier{"race_namespace", "race_view"}
+	require.NoError(t, cat.CreateNamespace(context.Background(), namespace, nil))
+
```
laskoviymishka, line 193: Good to see the update-properties race covered now. CreateTable and the CommitTable create branch got the same serializable + in-tx recheck treatment as CreateView, though, and neither has a race test — a regression on either wouldn't be caught here.

Could we parameterize this over the create operation (CreateTable / CreateView / CommitTable-create) so all three run against the same drop race? wdyt?

laskoviymishka (changes requested):
Dropping the `iceberg_type` filter fixes the view-only namespace case, and moving the object/property checks inside the serializable transaction is the right change. `TestDropNamespaceRejectsViewOnlyNamespace` covers the original bug.

`UpdateNamespaceProperties` now follows the same serializable transaction + in-transaction recheck pattern, which closes Zeroshade’s resurrection race. The new race test covers that path well. Tanmay’s points are also addressed: serialization failures are retried, and the `DropNamespace` errors now use `nsToDelete`.

I’d still hold the merge because the retry logic has two correctness problems.

The `interface{ Code() int }` check does not match errors from the actual SQLite driver. Its `Code()` method returns the named `ErrNo` / `ErrorCode` type, not a bare `int`, so `SQLITE_BUSY` and `SQLITE_LOCKED` will not be retried in production. The unit test misses this because the fake error returns `int`.

The retry budget also does little for the default SQLite catalog. `NewCatalog` configures neither WAL nor `busy_timeout`, so concurrent writes can fail with `SQLITE_BUSY` immediately. The race tests avoid this only because they enable WAL themselves.

The create/drop race test should also be tightened. It uses a 50 ms sleep and a 250 ms timeout to force ordering, which is likely to become flaky on slower CI. The update-properties test already uses a release channel to control the sequence, so the same pattern should work here. The branch where both operations fail currently has no assertions and can pass without proving anything.

Before merge, I’d address these:

* Match the real SQLite error type when classifying `BUSY` / `LOCKED`, and add a test using an error from the actual driver.
* Decide whether `NewCatalog` should configure WAL or `busy_timeout`, or document that callers must do it.
* Remove the double `%w` wrapping into `ErrCommitFailed`.
* Make the create/drop race deterministic, assert the both-error case, and cover `CreateTable` and the create branch of `CommitTable` as well.

After that I’m happy to review it again.

zeroshade (changes requested):
Re-reviewing the rebased revision — the namespace-drop atomicity looks good now. One new issue:

**Medium — timer drain can block on Go 1.23+** (`catalog/sql/sql.go:322`)
The retry-delay cancellation uses the pre-Go-1.23 timer drain pattern (`if !timer.Stop() { <-timer.C }`). This module targets Go 1.25, where timer channels are synchronous and `Stop` no longer requires draining; if the context wins the select around timer expiry, the drain receive can block forever.
*Fix:* drop the drain for Go 1.23+, or make it non-blocking (`select { case <-timer.C: default: }`).

zeroshade (approved):
Re-reviewed after `6d57c01`. Both findings from my earlier reviews are resolved — approving. 🎉

- ✅ **Namespace resurrection** — `UpdateNamespaceProperties` now runs under `withSerializableWriteTx` with an in-transaction `namespaceExistsInTx` recheck, and the new release-channel race test covers drop-vs-update. The read-before-drop / insert-after-drop window is closed.
- ✅ **Timer drain on Go 1.23+** — the blocking `if !timer.Stop() { <-timer.C }` is gone; `ctx.Done()` now just calls `timer.Stop()` and returns `context.Cause(ctx)`, so an expiring retry delay can't wedge on a synchronous timer channel.

---

@laskoviymishka — flagging for your re-review, since your `CHANGES_REQUESTED` is the remaining gate. From my read, `6d57c01` looks like it addresses the points you raised, but you're the better judge here:

- **SQLite `Code() int` classification** — the typed branch is gone; retryable-lock detection now keys off the stable messages (`database is locked` / `database table is locked`), which hold across both sqliteshim drivers. `ORA-00060` was also added, and MySQL now matches the typed `*mysql.MySQLError` (1205/1213).
- **WAL / `busy_timeout`** — `NewCatalog` now documents that concurrent-writer callers must set these in the DSN, rather than mutating the caller-owned DB.
- **Double `%w`** — collapsed to a single `%w` wrap on `ErrCommitFailed`.
- **Create/drop race test** — the 50 ms sleep is replaced with the same deterministic release-channel pattern as the update test; the both-fail branch now asserts `require.False(dropErr == nil && createErr == nil)` unconditionally; and coverage was extended to `CreateTable` and the `CommitTable` create branch.

I'm not merging over your review — leaving that to your re-review.

laskoviymishka (approved):
:shipit:

PR #1467, opened 2026-07-15
fix(glue): fence commits during table rename

fallintoplace:
## Summary

- conditionally claim the Glue source table by `VersionId` before issuing the service's unconditional delete
- make the claimed source temporarily non-Iceberg so new writers stop, while stale writers fail the optimistic version check
- reconcile ambiguous claim responses before cleanup, restoring a claim that Glue applied despite returning an error
- restore the source conditionally if deletion fails, then remove the destination only after the source is visible again
- allow explicit `DropTable` to remove a rename claim orphaned by process termination while continuing to reject unrelated non-Iceberg tables
- run rollback with a detached, bounded context, preserve rollback failures, and retain pre-existing table parameters

## Why

Glue implements rename as create-destination followed by delete-source, but `DeleteTable` has no version condition. A commit could previously land after the final source read and then be deleted, leaving the destination on the older metadata version.

The conditional claim closes that window: a commit that wins first invalidates the rename's `VersionId`, while a rename that wins first prevents new Iceberg commits from advancing the source before deletion. For transport errors where Glue may have applied `UpdateTable`, recovery reads the raw source state and only removes the destination after the source is known to be visible again.

The claim is durable, so a process can terminate after claiming the source but before delete or rollback. Load, commit, and re-rename remain fenced, while an explicit drop provides an API recovery path for that orphaned claim.

## Testing

- concurrent commit/rename fencing
- source restoration after delete failure
- ambiguous successful-update response recovery and parameter preservation
- orphaned rename-claim drop recovery and non-Iceberg drop rejection
- 10 repeated focused rename/drop test runs
- `go test ./catalog/glue`
- `go test -race ./catalog/glue`
- `go vet ./catalog/glue`

### DIFF
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -26,6 +26,7 @@ import (
 	"maps"
 	"strconv"
 	"strings"
+	"time"
 	_ "unsafe"
 
 	"github.com/apache/iceberg-go"
@@ -55,6 +56,8 @@ const (
 	tableParamTableType                = "table_type"
 	tableParamMetadataLocation         = "metadata_location"
 	tableParamPreviousMetadataLocation = "previous_metadata_location"
+	tableParamRenameToken              = "iceberg.go.rename-token"
+	glueTypeIcebergRenaming            = "ICEBERG_RENAMING"
 
 	// The ID of the Glue Data Catalog where the tables reside. If none is provided, Glue
 	// automatically uses the caller's AWS account ID by default.
@@ -77,6 +80,7 @@ const (
 	icebergFieldIDKey       = "iceberg.field.id"
 	icebergFieldOptionalKey = "iceberg.field.optional"
 	icebergFieldCurrentKey  = "iceberg.field.current"
+	renameCleanupTimeout    = 5 * time.Second
 )
 
 var _ catalog.Catalog = (*Catalog)(nil)
@@ -379,10 +383,15 @@ func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) er
 		return err
 	}
 
-	// Check if the table exists and is an Iceberg table.
-	_, err = c.getTable(ctx, database, tableName)
+	// Use the raw lookup so an explicit drop can recover a source left claimed
+	// by a rename process that terminated before delete or rollback.
+	glueTable, err := c.getRawTable(ctx, database, tableName)
 	if err != nil {
 		return err
+	}
+	tableType := glueTable.Parameters[tableParamTableType]
+	if !strings.EqualFold(tableType, glueTypeIceberg) && tableType != glueTypeIcebergRenaming {
+		return fmt.Errorf("table %s.%s is not an iceberg table", database, tableName)
 	}
 
 	params := &glue.DeleteTableInput{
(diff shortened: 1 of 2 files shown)


catalog/glue/glue.go
```diff
@@ -671,6 +754,23 @@ func (c *Catalog) ListNamespaces(ctx context.Context, parent table.Identifier) (
...
 // GetTable loads a table from the Glue Catalog using the given database and table name.
 func (c *Catalog) getTable(ctx context.Context, database, tableName string) (*types.Table, error) {
+	tbl, err := c.getRawTable(ctx, database, tableName)
+	if err != nil {
+		return nil, err
+	}
+
+	tableType := tbl.Parameters[tableParamTableType]
+	if tableType == glueTypeIcebergRenaming {
+		return nil, fmt.Errorf("%w: table %s.%s is being renamed", table.ErrCommitFailed, database, tableName)
```
tanmayrauth, line 766: This makes getTable reject any ICEBERG_RENAMING table, and every public op goes through getTable — LoadTable (236), CommitTable (317), DropTable (387), and RenameTable's own source fetch (446). So if the process dies between the successful claim (line 472) and the delete/rollback, the source is stuck in ICEBERG_RENAMING and nothing — load, commit, drop, or re-rename — can recover it; the only fix is manual Glue surgery. That's a regression from the old path, which never mutated the source, so a crash always left it usable. Worst case: claim succeeds, pod is killed before DeleteTable, and on restart DropTable(from) returns "is being renamed" with no way to clean up. Suggest DropTable go through getRawTable and be allowed to delete a claimed source so there's an in-API recovery path.

catalog/glue/glue.go
```diff
@@ -443,37 +447,48 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
...
-	// Revalidate the source table immediately before delete to narrow the
-	// rename race window. Glue does not offer a conditional DeleteTable, so
-	// this cannot eliminate the race if a commit lands after this read.
-	currentFromGlueTable, err := c.getTable(ctx, fromDatabase, fromTable)
+	// Claim the source with a conditional update before issuing Glue's
+	// unconditional delete. Changing table_type makes new Iceberg writers refuse
+	// the source, while VersionId rejects writers that loaded the previous version.
+	renameToken := fmt.Sprintf("%s.%s@%s", toDatabase, toTable, aws.ToString(fromGlueTable.VersionId))
+	claimInput := glueTableInput(fromTable, fromGlueTable)
+	claimInput.Parameters[tableParamTableType] = glueTypeIcebergRenaming
```
tanmayrauth, line 474: This marker is durable and there's no self-healing for it — the rename-token embeds only dest+VersionId, no timestamp, so a stale claim from a crashed rename is indistinguishable from a live one. Worth either stamping a timestamp so a recovery path can age out stale claims, or documenting that operators must clear the marker manually.

catalog/glue/glue.go
```diff
@@ -443,37 +447,48 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch the table %s.%s: %w", fromDatabase, fromTable, err)
 	}
+	if !strings.EqualFold(fromGlueTable.Parameters[tableParamTableType], glueTypeIceberg) {
```
tanmayrauth: getTable at line 446 already errors on non-iceberg (and RENAMING) tables, so this branch is unreachable. Not harmful, but it reads as if getTable can return non-iceberg tables — drop it or add a comment.

tanmayrauth (commented):
The VersionId-conditional claim closes the commit-loss race cleanly. One blocker: an orphaned ICEBERG_RENAMING claim can't be recovered through the API. Details inline.

zeroshade (approved):
LGTM, thanks!

PR #1468, opened 2026-07-15
fix(hive): clean up pending locks after cancellation

fallintoplace:
## Summary

- defer pending-lock cleanup from the point where Hive returns a lock ID until ownership is transferred to `HiveLock`
- detach cleanup from caller cancellation while preserving context values, with a five-second timeout
- join unlock failures with the primary acquisition error instead of discarding them

## Why

The cancellation path passed an already-canceled context to `Unlock`, so a real metastore client could reject cleanup without sending it and leave the pending lock behind. Check failures and retry exhaustion also discarded cleanup errors.

## Testing

- `go test ./catalog/hive`
- `go vet ./catalog/hive`
- repository package suite excluding the locally hanging `io/gocloud` test binary

### DIFF
--- a/catalog/hive/lock.go
+++ b/catalog/hive/lock.go
@@ -32,6 +32,8 @@ import (
 // ErrLockAcquisitionFailed is returned when a lock cannot be acquired after all retries.
 var ErrLockAcquisitionFailed = errors.New("failed to acquire lock")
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1470, opened 2026-07-15
fix(table): avoid mutating binary upper bounds

fallintoplace:
## Summary

- clone binary prefixes before incrementing truncated upper bounds
- return an owned copy when truncation is unnecessary
- verify repeated truncation is stable and never aliases or mutates the input

## Why

`TruncateUpperBoundBinary` sliced directly into the caller's byte array and incremented that shared prefix. Statistics aggregation could therefore mutate its stored maximum while merely calculating an upper bound, making later calls and comparisons depend on call order.

## Testing

- `go test ./table/internal -run TestTruncateUpperBoundBinary -count=20`
- `go test ./table/internal ./table`
- `go vet ./table/internal ./table`

### DIFF
--- a/table/internal/utils.go
+++ b/table/internal/utils.go
@@ -18,7 +18,6 @@
 package internal
 
 import (
-	"bytes"
 	"container/heap"
 	"context"
 	"encoding/binary"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1471, opened 2026-07-15
fix(puffin): clone blob metadata inputs

fallintoplace:
## Summary

- clone blob field IDs and properties before validation and storage
- return separate metadata copies so callers cannot mutate the pending footer
- verify mutations through both the input and returned metadata leave serialized metadata unchanged

## Why

`AddBlob` retained caller-owned slices and maps after validating them. A caller could mutate either the input or returned `BlobMetadata` before `Finish`, causing the writer to serialize values that never passed validation.

## Testing

- focused ownership and writer validation tests repeated 20 times
- `go test ./puffin`
- `go vet ./puffin`

### DIFF
--- a/puffin/puffin_test.go
+++ b/puffin/puffin_test.go
@@ -134,6 +134,30 @@ func TestRoundTrip(t *testing.T) {
 	assert.Equal(t, blob2Data, allBlobs[1].Data)
 }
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1472, opened 2026-07-15
fix(io): preprocess names consistently for removal

fallintoplace:
## Summary

- centralize wrapped filesystem filename preprocessing and root normalization
- apply the same transformation to `Open`, `ReadFile`, and `Remove`
- verify all operations address the identical underlying filename

## Why

`FSPreProcName` promised to preprocess filenames before forwarding them, but `Remove` bypassed the configured function. The wrapper could consequently read one transformed path and attempt to delete a different original URI or absolute path.

## Testing

- filename transformation regression repeated 20 times
- `go test ./io`
- `go vet ./io`

### DIFF
--- a/io/io.go
+++ b/io/io.go
@@ -241,16 +241,12 @@ type readFileFS struct {
 }
 
 func (r readFileFS) ReadFile(name string) ([]byte, error) {
-	if r.preProcessName != nil {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1473, opened 2026-07-15
fix(cli): fail closed on explicit config errors

fallintoplace:
Explicit config failures currently fall back to CLI defaults, which is risky for commands that mutate catalog state.

This adds an error-reporting config loader while retaining the existing best-effort helper for compatibility. The CLI now exits before catalog initialization when an explicit config cannot be read, YAML is malformed, or the requested catalog is absent. A missing implicit home config remains optional. Diagnostics include the resolved config path.

Regression coverage includes implicit and explicit missing files, read failures, malformed YAML, absent catalog selection, valid config merging, and the CLI exit path.

### DIFF
--- a/cmd/iceberg/args_test.go
+++ b/cmd/iceberg/args_test.go
@@ -18,6 +18,9 @@
 package main
 
 import (
+	"os"
+	"os/exec"
+	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1474, opened 2026-07-15
fix(cli): resolve sort fields by schema name

fallintoplace:
The CLI sort-order parser currently ignores each requested column name and assigns source IDs from the argument position. That silently creates incorrect sort orders for schemas whose field IDs are not 1, 2, 3, and so on.

This passes the table schema into sort-order parsing, resolves top-level names and nested paths to their Iceberg field IDs, rejects unknown or duplicate fields and malformed components, and checks the resulting sort order against the schema before table creation.

Tests use non-sequential field IDs and assert the exact source IDs emitted in a REST create-table request.


zeroshade (approved):
LGTM, thanks!

PR #1469, opened 2026-07-15
feat(rest): implement WaitForPlan poller for scan planning

Revanth14:
## Summary

Implements `Catalog.WaitForPlan` as part of [#1178](https://github.com/apache/iceberg-go/issues/1178), building on the merged scan-planning client methods (#1324).

The poller waits for an asynchronously submitted REST scan plan to complete and returns the completed planning result (task envelope and storage credentials).

## Changes

- Polls `FetchPlanningResult` using decorrelated-jitter backoff.
- Handles completed, submitted, failed, cancelled, and expired plans.
- Retries transient GET responses: `408`, `429`, `500`, `502`, `503`, and `504`.
- Supports integer and HTTP-date `Retry-After` headers.
- Uses the caller's context deadline as the primary timeout.
- Adds `MaxRetries` and `ErrPlanPollExhausted`.
- Cancels abandoned plans on context cancellation or retry exhaustion.
- Bounds cleanup using `CancelGracePeriod`.
- Forwards access-delegation headers while polling.
- Handles failed responses with missing or malformed error details.
- Threads HTTP status and `Retry-After` through the shared REST error type (`errorResponse`/`handleNon200`) so the poller can apply retry policy without re-plumbing each method. This is the only change outside the scan-planning surface.
- Adds overflow-safe backoff calculations and empty plan-ID validation.

When the context has a deadline, the default retry cap is disabled so polling continues until that deadline.

This PR implements the wait/poll phase only. Scan-task **payload decoding** (`RESTFileScanTask`/`RESTDeleteFile`), `plan-tasks` fan-out, and wiring `PlanFiles` end-to-end remain follow-up work.

## Testing

All pass, including under the race detector:

```text
go test ./catalog/rest
go test -race ./catalog/rest
go vet ./catalog/rest

PR #1475, opened 2026-07-15
test(rest): fix flaky TestWaitForPlanBoundsSlowServerSideCancel

zeroshade:
## What

`TestWaitForPlanBoundsSlowServerSideCancel` (added in #1469) is flaky. It failed on `main` in the `ubuntu-latest go1.25.8` cell of the `Go` workflow (run 29428513100 on commit 3648ea3) while passing on every other commit and matrix cell.

## Root cause

`WaitForPlan` issues the server-side cancel (`DELETE .../plan/plan-1`) on a context **detached** from the caller (bounded by `CancelGracePeriod`). When the caller's context is cancelled, `WaitForPlan` returns `context.Canceled` immediately, while the detached goroutine issues the `DELETE` (which sets `cancelStarted`) concurrently.

The test asserted `cancelStarted.Load()` **immediately** after `WaitForPlan` returned, so on a slower runner the assertion could execute before the detached cancel had been attempted. This is a race in the test, not in the product code.

## Fix

Poll the flag with `assert.Eventually` (1s budget, 5ms tick) instead of reading it exactly once, and add a comment explaining why so it isn't "simplified" back.

## Verification

- `go test ./catalog/rest/ -run TestWaitForPlan -race -count=50` → `ok`
- full `catalog/rest` package (`-count=1`) → `ok`
- `gofmt` / `go vet` clean

### DIFF
--- a/catalog/rest/wait_for_plan_test.go
+++ b/catalog/rest/wait_for_plan_test.go
@@ -452,7 +452,11 @@ func TestWaitForPlanBoundsSlowServerSideCancel(t *testing.T) {
 	select {
 	case err := <-done:
 		require.ErrorIs(t, err, context.Canceled)
-		assert.True(t, cancelStarted.Load(), "expected the server-side cancel to be attempted")
+		// The server-side cancel runs on a context detached from the caller, so it
+		// may be issued slightly after WaitForPlan returns. Poll instead of reading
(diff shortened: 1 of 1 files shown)


laskoviymishka (approved):
👍

PR #1476, opened 2026-07-15
feat(table): add snapshots metadata table

laskoviymishka:
Second inspect table on the Table.Inspect hub. One row per snapshot with committed_at, snapshot_id, parent_id, operation, manifest_list, and summary (field IDs 1-8, matching Java/PyIceberg/Rust); null operation/summary/ manifest_list when absent. Also adds allocator injection, exported HistorySchema/SnapshotsSchema, and a shared single-batch reader helper.

### DIFF
--- a/table/inspect.go
+++ b/table/inspect.go
@@ -34,12 +34,34 @@ import (
 // Obtain one via Table.Inspect. Each method returns an array.RecordReader that
 // the caller is responsible for releasing.
 type InspectTable struct {
-	tbl Table
+	tbl   Table
+	alloc memory.Allocator
 }
 
-// Inspect returns an InspectTable for reading this table's metadata tables.
-func (t Table) Inspect() InspectTable {
-	return InspectTable{tbl: t}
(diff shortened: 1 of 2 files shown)


zeroshade (changes requested):
Thanks! One correctness issue before merge:

**Medium — snapshots metadata table omits the `operation` summary entry** (`table/inspect.go:215`)

The `summary` map only iterates `snap.Summary.Properties`, so it omits the snapshot `operation`. Since `Summary.MarshalJSON` serializes `Operation` back into the metadata summary map, the inspect table no longer faithfully exposes the stored snapshot summary — it returns an empty map for summaries that contain only `"operation"`.

**Suggested fix:** include `operation` in the `summary` map (e.g. append `"operation"` / `string(snap.Summary.Operation)` before appending `Properties`) and update the snapshot test expectation accordingly.

zeroshade (approved):
Re-reviewed: the summary map now folds in the operation key (table/inspect.go), addressing my earlier note. Thanks @laskoviymishka!

PR #1477, opened 2026-07-15
fix(catalog/rest): ScanPlan requests + responses

rambleraptor:
Part of #1178 

This allows catalog.PlanFiles to send the Plan requests and parse the responses. This brings together the work that @Revanth14 has been doing with my expression parsing work.

I also changed the expression parsing interface, since we already have JSON parsing interfaces for expressions.


catalog/rest/scan_planning.go
```diff
@@ -187,9 +189,180 @@ func (r *Catalog) SupportsRemoteScanPlanning() bool {
...
+	}
+
+	return nil
+}
+
+// planTableScanRequestFrom builds the planTableScan request body from the
+// planner-level request, serializing the row filter to ExpressionParser JSON. A
+// nil or always-true filter is left off so the server plans without one.
+func planTableScanRequestFrom(req table.ScanPlanningRequest) (PlanTableScanRequest, error) {
+	out := PlanTableScanRequest{
```
tanmayrauth, line 349: Neither the initial plan request here nor the poll at line 223 (WaitForPlanOptions{}) sets AccessDelegation, so X-Iceberg-Access-Delegation is never sent. A spec-compliant server vends storage credentials only when asked, so completed.StorageCredentials comes back empty, planIOFromCredentials returns nil, and the scan silently falls back to the table's ambient FileIO — for a table with no ambient creds that's a read failure at scan time. TestPlanFilesSurfacesVendedCredentials passes only because the mock vends unconditionally, and ScanPlanningRequest has no field to carry delegation, so a caller can't thread it either.  This is the fix that has to land first — the planScopedIO path below is unreachable against a real server until it does.

catalog/rest/scan_planning.go
```diff
@@ -187,9 +189,180 @@ func (r *Catalog) SupportsRemoteScanPlanning() bool {
...
+
+func (p *planScopedIO) Load(ctx context.Context) (iceio.IO, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.cached != nil {
+		return p.cached, nil
+	}
+
+	fs, err := iceio.LoadFS(ctx, resolveStorageCredentials(p.creds, p.location), p.location)
```
tanmayrauth: This builds the plan FileIO once and caches it until Close with no expiry handling, whereas the normal table path (rest.go:1044) reads vended creds through vendedCredentialRefresher, which tracks the credential expiry keys parseCredentialExpiry already knows about and re-fetches on TTL. A scan that outlives the vended credential's TTL — the long fanout plans that make server-side planning worthwhile — starts getting raw storage 403s on later ReadTasks reads with no diagnosable "expired" error. Plan creds can't be refreshed via the table-cred endpoint, so this needs a deliberate call: at least reuse parseCredentialExpiry to surface an expiry error, and ideally don't stand up a second vended-cred IO path parallel to vendedCredentialRefresher.

catalog/rest/scan_planning.go
```diff
@@ -187,9 +189,180 @@ func (r *Catalog) SupportsRemoteScanPlanning() bool {
...
+	}
+
+	tasks, err := remoteScanTasks(files, deletes)
+	if err != nil {
+		return table.ScanPlanningResult{}, err
+	}
+
+	return table.ScanPlanningResult{
+		Tasks: tasks,
+		IO:    planIOFromCredentials(completed.StorageCredentials, req.MetadataLocation),
```
tanmayrauth: This resolves the plan's credential array against the metadata location and builds a single FileIO for the scan. That matches how the rest of iceberg-go handles vended creds (loadTable resolves against MetadataLoc too), so it's not a regression — but plan.storage-credentials is an array of prefix-scoped creds, and resolving once against the metadata location means a plan whose data/delete files sit under a different prefix picks the wrong or no credential once the decoder routes reads through this IO. Fine for a single-bucket table where data lives under the table root; worth settling the PlanIO shape before the
  decoder makes it load-bearing.

catalog/rest/scan_planning.go
```diff
@@ -187,9 +189,180 @@ func (r *Catalog) SupportsRemoteScanPlanning() bool {
...
+		SnapshotID:        req.SnapshotID,
+		Select:            req.SelectedFields,
+		MinRowsRequested:  req.MinRowsRequested,
+		CaseSensitive:     req.CaseSensitive,
+		UseSnapshotSchema: req.UseSnapshotSchema,
+		StatsFields:       req.StatsFields,
+	}
+
+	if req.RowFilter != nil && !req.RowFilter.Equals(iceberg.AlwaysTrue{}) {
+		filter, err := json.Marshal(req.RowFilter)
```
tanmayrauth: Not a bug today — all BooleanExpression types have MarshalJSON and predicateType errors on unknown ops. But moving from a dedicated MarshalExpressionJSON entry point to bare json.Marshal drops the one place that guaranteed an expression is REST-serializable: marshalChildren recurses through json.Marshal on the interface, so a future expression type added without a MarshalJSON won't fail to compile and won't error — it silently emits struct-default JSON and the server gets a malformed filter. Marshaling through a helper that asserts the concrete type is known would keep that a loud error rather than a silent wrong
  filter.

tanmayrauth (commented):
Mechanics look right; the open questions are all in the plan-scoped IO / credential design. Details inline.

zeroshade (changes requested):
Thanks! One issue to address before merge:

**Medium — plan-scoped IO drops catalog/table IO properties** (`catalog/rest/scan_planning.go:322`)

The plan-scoped IO is built from only `resolveStorageCredentials(p.creds, p.location)`, dropping the catalog/table IO properties that the normal table FileIO was created with. When `ReadTasks` uses this PlanIO instead of `scan.ioF`, remote scans with vended credentials can lose required settings such as custom S3 endpoints, path-style flags, regions, warehouse fallback, or provider-specific config.

**Suggested fix:** store the base IO properties on `planScopedIO` and load with a clone of those properties overlaid with the matching vended credential config.

zeroshade (changes requested):
Re-reviewing. The plan-scoped IO issue from my earlier review still stands, plus a second serialization gap:

**Medium — plan-scoped IO drops catalog/table properties** (`catalog/rest/scan_planning.go:322`)
Plan-scoped IO is loaded only from the resolved vended-credential config, dropping the catalog/table IO properties used by the normal table FileIO. Remote scans with vended credentials can lose custom endpoints, path-style flags, regions, warehouse fallback, or provider-specific config.
*Fix:* carry the base IO properties into `planScopedIO` and load with a cloned base config overlaid with the vended credential config.

**Medium — unbound filter marshaling fails for timestamp predicates** (`catalog/rest/scan_planning.go:362`)
`planTableScanRequestFrom` marshals the unbound row filter directly. Some expressions (notably timestamp/timestamptz predicates) need the table schema to choose the correct ExpressionParser literal form and fail to serialize, even though local planning handles them after binding.
*Fix:* bind the filter against the effective planning schema (with case-sensitivity) before marshaling; add timestamp/timestamptz coverage.

catalog/rest/scan_planning.go
rambleraptor: I believe that @zeroshade's copy suggestion will fix this.

catalog/rest/scan_planning.go
rambleraptor: I set a TODO to help determine this in a later phase if that works.

rambleraptor:
@zeroshade made those changes! Let me know if the JSON marshaling looks wonky. I'm trying to make sure we can use standard json.Marshal / json.Unmarshal whenever possible.

catalog/rest/scan_planning.go
```diff
@@ -187,9 +190,208 @@ func (r *Catalog) SupportsRemoteScanPlanning() bool {
...
+}
+
+// planIOBaseProps rebuilds the props the table's own FileIO was built with, so a
+// plan-scoped IO keeps settings like a custom S3 endpoint or region. loadTable's
+// config block never reaches the planner seam, so it isn't included.
+func (r *Catalog) planIOBaseProps(meta table.ScanPlanningMetadata) iceberg.Properties {
+	props := make(iceberg.Properties, len(r.props))
+	maps.Copy(props, r.props)
+	if meta != nil {
+		maps.Copy(props, meta.Properties())
```
tanmayrauth, line 256: This overlay carries catalog props + table metadata props, but the normal table FileIO is built from r.props + Metadata.Properties() + ret.Config + vended creds (rest.go:1256-1260). ret.Config is the loadTable `config` block —  merged last-but-one so it overrides catalog/table props — and it's the standard place a REST server returns s3.endpoint, s3.path-style-access, and client.region, especially for non-AWS / vended-cred setups. 

Dropping it means those are the settings most likely to go missing: a MinIO-backed table whose server returns s3.endpoint + s3.path-style-access in the config block gets them on the normal FileIO but not on the plan-scoped IO, so vended-cred ReadTasks point at the AWS-default host and fail — unless the plan creds happen to re-carry that config, which isn't guaranteed for endpoint/path-style/region. This is the same class of property drop flagged earlier; carrying r.props + table props doesn't cover it because the override source is ret.Config. It can't be fixed with another maps.Copy here — ScanPlanningRequest/ScanPlanningMetadata never expose ret.Config, so the seam itself needs to carry the effective IO props (config block included) for the overlay to actually preserve endpoint/region/path-style.

catalog/rest/scan_planning.go
```diff
@@ -187,9 +190,208 @@ func (r *Catalog) SupportsRemoteScanPlanning() bool {
...
+		caseSensitive = *req.CaseSensitive
+	}
+
+	// Snapshot-schema selection (UseSnapshotSchema) is OQ4, deferred; bind current.
+	bound, err := iceberg.BindExpr(req.Metadata.CurrentSchema(), req.RowFilter, caseSensitive)
+	if err != nil {
+		return nil, fmt.Errorf("%w: binding scan filter: %s", iceberg.ErrInvalidArgument, err)
+	}
+
+	filter, err := json.Marshal(bound)
```
tanmayrauth, line 389: Binding before marshal fixed the timestamp case but regressed nested-field refs. boundRef.MarshalJSON (expr_json.go:276) emits the leaf name, so EqualTo(Reference("location.lat"), 45.0) now serializes as
  {"type":"eq","term":"lat",...} where the old unbound path emitted "location.lat". The server then resolves the wrong column (or a different top-level "lat") → wrong files planned / silent wrong results. Suggest binding only for literal typing and serializing the term from the full path  (schema.FindColumnName(field.ID)), plus a nested-field test.

tanmayrauth (commented):
Two things from the latest commit: binding the filter fixed timestamps but regressed nested-field refs, and the base-props overlay still drops the loadTable config block, so the endpoints/regions it was meant to preserve can still go missing. Both are latent behind SupportsRemoteScanPlanning()=false today — worth settling before the decoder makes this path live.

catalog/rest/vended_creds.go
```diff
@@ -108,14 +114,20 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
...
-	if v.cachedIO != nil && !v.now().After(v.expiresAt) {
+	if v.cachedIO != nil && !v.expired() {
 		return v.cachedIO, nil
 	}
 
 	var config iceberg.Properties
-	if v.cachedIO == nil {
+	switch {
+	case v.cachedIO == nil:
 		config = v.props
```
zeroshade, line 124: Non-blocking nit (fine as a follow-up — not gating the approval):

The `ErrVendedCredentialsExpired` fail-loud path and the `expired()` guard above both only fire when `cachedIO != nil`. On the **first** `Load` we take this branch (`cachedIO == nil`), set `config = v.props`, and build the FileIO via `iceio.LoadFS` below **without** checking whether the creds are already past `expiresAt`. So a plan whose creds expire before the first read still gets a working-looking IO whose reads 403 — the exact opaque failure this sentinel was added to prevent.

It's narrow (needs creds already expired at first use) and non-regressive, so it's a nit rather than a blocker. To close it: when `fetchCreds == nil && cachedIO == nil`, compute the expiry from `config` and return `ErrVendedCredentialsExpired` if it's already past, before `LoadFS`.

zeroshade (approved):
Re-reviewed after `e2e1ca4` ("PR comments") — both findings from my earlier reviews are resolved, with tests. Approving. 🎉

- ✅ **Plan-scoped IO now carries the table's base props.** `planIOBaseProps` merges catalog + table-metadata properties and overlays the vended creds on top (vended wins on collision), so a custom S3 endpoint / region / path-style / warehouse fallback survives into the plan-scoped FileIO.
- ✅ **Row filter is bound before marshaling.** `marshalScanFilter` binds against the current schema (respecting case-sensitivity) before encoding, so timestamp/timestamptz predicates serialize correctly — covered by the new "timestamp predicate" test.

The extra `ErrVendedCredentialsExpired` fail-loud path + never-expires semantics are a nice touch.

Left one **non-blocking** nit inline on the first-load expiry path — not gating this approval.

catalog/rest/vended_creds.go
rambleraptor: Thanks for the nit @zeroshade. I was OOO last week and just getting around to it. Fixed now.

zeroshade (approved):
Re-confirming approval — the follow-up commit (`7389691`) cleanly closes the non-blocking nit from my last pass: on the first `Load` with plan-scoped creds (`fetchCreds == nil`), it now checks `parseCredentialExpiry(config)` and fails loud with `ErrVendedCredentialsExpired` before handing back an IO whose reads would 403. Nice refactor of the message into `expiredError`, and `TestPlanScopedIOAlreadyExpiredCredentials` locks it in. Good to merge.

PR #1478, opened 2026-07-15
 feat(table): bounds for shredded variant fields

### LINKED ISSUE #988: feat(table): skip min/max stats on the variant column itself, allow stats on shredded children
Parent: #589

Variants don't have a natural total order, so emitting min/max for the parent variant column is meaningless. Java's current posture: skip min/max stats for the variant column itself but emit stats on shredded typed columns (which are typed primitives at write time, so the existing column-stats path covers them for free once #987 lands).

Mirror that here: the stats writer in `table/internal/parquet_files.go` should special-case `VariantType` to skip its own stats but allow the typed_value subtree to carry stats normally. The carve-out is one branch in the writer plus an assertion test that no min/max appears for the parent variant column on round-trip, while min/max is present for shredded children when typed.

Java reference: [apache/iceberg#10498](https://github.com/apache/iceberg/pull/10498) (variant stats handling). Depends on #987 because shredded children only exist once shredding is wired through the writer.

nssalian:
Closes [#988](https://github.com/apache/iceberg-go/issues/988).

Writes per-field min/max bounds for shredded variant columns. Each shredded `typed_value` field's bounds are serialized as a variant object keyed by its normalized JSON path and stored under the parent variant field id, matching Java's `ParquetMetrics.variant()`. The parent variant column still emits no scalar min/max. Builds on the shredded writer [#1351](https://github.com/apache/iceberg-go/pull/1351).

## Changes

- `collectVariantBounds` walks each shredded variant's `typed_value` subtree, reads each primitive leaf's Parquet min/max, and serializes the lower/upper bounds as a variant object (spec "Bounds for Variant"). Keys are RFC-9535 normalized JSON paths, matching Java's `PathUtil`.
- A field's bound is dropped when its residual `value` column holds a non-null, non-variant-null value (partial shred / mixed type), and when min or max is NaN. Both match Java.
- Arrays contribute no bounds (Java parity).
- `DataFileStatsFromMeta` takes the inferred arrow schema to locate shredded leaves; `ToDataFile` merges the bounds into `lower_bounds`/`upper_bounds`. The add-files path passes no schema and does not recompute.

## Notes

- Integer bounds use arrow-go's minimal-width `AppendInt`, not Java's fixed INT32/INT64. Both are valid variant encodings and read to the same integer, so bounds are read-equivalent across implementations though not byte-identical.
- Java also emits variant field value/null counts; iceberg-go still omits those (pre-dates this PR, out of scope). Tracking as a follow-up.
- The residual-value invalidation matches Java's `MetricsVariantVisitor.value()` rule, including the no-usable-stats case hardened in [apache/iceberg#16585](https://github.com/apache/iceberg/pull/16585).

## Testing

- Unit: path escaping incl. control chars, leaf enumeration (nested object, narrow int, array-skip, root scalar), bounds serialization, NaN guard.
- Write path: byte-exact bounds for all 14 shredded scalar types; nested object; a null-coexisting field keeps its bound; a mixed-type field is dropped.
- Spark 4 integration: a Go-written shredded table carries variant child bounds and Spark reads it back.


zeroshade (changes requested):
Thanks for this! One correctness issue I'd like to see addressed before merge:

**High — variant child bounds can be emitted from incomplete row-group stats** (`table/internal/parquet_files.go:1058`)

Variant child bounds are still emitted when a typed child column is missing statistics in one or more row groups. The code `continue`s on `!set`, missing stats, or missing min/max, so an aggregator built from the other row groups can produce lower/upper bounds that cover only part of the file. Those incorrect bounds can make planning prune files that actually contain matching rows.

**Suggested fix:** track invalid typed paths the way the existing top-level metrics logic does — if stats are unavailable for a row group that has non-null values, invalidate that leaf and omit its variant bound. Only skip safely when the row group is provably all-null for that leaf.

zeroshade (changes requested):
Re-reviewing the latest revision. The variant-bounds correctness issue from my earlier review is still present, and there's a closely related second case:

**High — partial variant child bounds from incomplete row-group stats** (`table/internal/parquet_files.go:1042`)
Typed variant child bounds are accumulated only from row groups that have usable min/max stats. When a typed child has non-null values in a row group but stats are unset/unavailable, the code `continue`s and can emit a parent variant bound covering only part of the file — readers may then prune files that actually contain matching rows.
*Fix:* track an invalid state per typed leaf (mirroring the existing top-level metric invalidation); only skip a row group when the leaf is provably all-null, otherwise omit that leaf's bound.

**High — nested leaves drop ancestor residual `value` columns** (`table/internal/variant_bounds.go:72`)
Nested variant leaves only remember the leaf's own residual `value` column. For a shredded path like `payload.location.latitude`, rows where `location` is a non-object live in `payload.typed_value.location.value`, but the `latitude` bound isn't invalidated, so it again describes only the object-shaped subset.
*Fix:* carry all ancestor residual value paths per leaf and suppress the leaf bound if any ancestor/leaf residual column has non-null, non-variant-null values.

nssalian:
Thanks for the review @zeroshade. I haven't had a chance to make the changes yet since my initial commit at the start of the PR. I'll follow up with the changes in upcoming commits.

nssalian:
Both the review items are addressed along with a few additional changes:
- Incomplete row-group stats: a typed leaf with non-null values but no usable min/max is now invalidated and its bound dropped, only a provably all-null row group is skipped. 
- Ancestor residuals: each leaf carries its full ancestor residual value chain and drops its bound if any ancestor/self 
residual holds non-null data. 
- I added new regression tests for both
- I also fixed a mislabeled comment in DataFileFromMetadata and added the missing bound-correctness test on that recompute path. 
- Rebased/merged onto latest main.

table/internal/parquet_files.go
```diff
@@ -989,15 +964,195 @@ func (p parquetFormat) DataFileStatsFromMeta(meta Metadata, statsCols map[int]St
 		return ok
 	})
 
+	vlo, vup := p.collectVariantBounds(pqmeta, arrowSchema, colMapping)
```
zeroshade: **Medium — variant bounds bypass the configured metrics mode**

`collectVariantBounds` is invoked unconditionally here and isn't passed `statsCols`, so it can't consult the parent variant column's `MetricsMode`. The normal-column path honors it — `MetricModeNone` skips the column entirely (`DataFileStatsFromMeta`, ~L888-890) and `MetricModeCounts` skips min/max (~L919) — but variant child lower/upper values are still written even when `write.metadata.metrics.default=none`, `=counts`, or `write.metadata.metrics.column.<name>=none` is configured, exposing variant child data values in metadata that the user explicitly disabled.

Suggested fix: thread `statsCols` into `collectVariantBounds` and, per parent variant field id, only emit bounds when the parent's mode permits them (skip `None`/`Counts`). The bounds themselves are correct — this is a metrics-config / info-exposure gap, not a wrong-bounds bug.

zeroshade (commented):
Re-reviewed the latest revision. Both High correctness issues from my earlier reviews — incomplete row-group stats invalidation and ancestor residual `value` tracking for nested leaves — are now correctly handled by the per-leaf `invalid` flag and the `valuePaths` change. Thanks for addressing those. One remaining item inline (Medium).

table/internal/parquet_files.go
nssalian: Nice catch. I fixed it. Turned out the gate alone wouldn't have done anything - `Variant()` was returning an empty collector, so variant fields never had a stats entry to check the mode against. 
Couple of parts in my commit:
1. `Variant()` now populates the collector with the resolved mode (default + per-column override), sharing a `resolveColumnMetricsMode` helper with Primitive().
2. `collectVariantBounds` skips the parent variant field when it's none/counts (or missing) - same as Java's 
 `ParquetMetrics.variant()`. Per-column still wins over the default, so `column.<name>=full` emits bounds even under a global none. 

I added unit + e2e tests for all the mode combos. PTAL

table/internal/parquet_files.go
```diff
@@ -989,15 +964,199 @@ func (p parquetFormat) DataFileStatsFromMeta(meta Metadata, statsCols map[int]St
...
+				continue
+			}
+			set, serr := cc.StatsSet()
+			st, terr := cc.Statistics()
+			switch {
+			case serr != nil || !set || terr != nil || st == nil:
+				e.invalid = true
+			case st.HasMinMax():
+				if e.agg == nil {
+					agg, aerr := p.createStatsAgg(e.leaf.icebergType, st.Type().String(), 0)
```
zeroshade, line 1104: Non-blocking: this passes `truncLen=0`, so a variant column configured with `truncate(N)` still writes full-length string/binary child bounds, whereas the primitive path honors the mode via `createStatsAgg(..., statsCol.Mode.Len)` (parquet_files.go:933). Full bounds are valid and read-equivalent, so this isn't a blocker — but it does mean `truncate(N)` is silently ignored for shredded variant leaves. Worth either threading the parent's truncate length through here, or documenting it next to the value/null-count follow-up so the metrics-mode semantics stay consistent. Was skipping truncation for variant child bounds intentional?

zeroshade (approved):
Re-reviewed the latest revision (49213e2a). The metrics-mode gap from my previous pass is now handled correctly: `Variant()` emits a `StatisticsCollector` carrying the resolved mode (via the shared `resolveColumnMetricsMode` helper with `Primitive()`), and `collectVariantBounds` now skips `none`/`counts` parents (and missing entries), with `TestCollectVariantBoundsMetricsMode` locking the behavior in. Together with the earlier incomplete-row-group invalidation and ancestor-residual `value` tracking, all of my correctness concerns are resolved and the test coverage is thorough.

LGTM — approving. One non-blocking nit inline about `truncate(N)` not being applied to variant child bounds; fine to handle as a follow-up.

table/internal/parquet_files.go
nssalian: I fixed this. Variant string/binary leaf bounds now truncate to a fixed 16, matching what Java's [ParquetVariantUtil(https://github.com/apache/iceberg/blob/main/parquet/src/main/java/org/apache/iceberg/parquet/ParquetVariantUtil.java#L347-L375) actually does: fixed length, independent of the column's configured metrics length, applied even under `full`. Threading the parent's `truncate(N)` through would've diverged from Java, so I didn't. Added string + binary tests for it.

One thing I found while matching it: there's an upper-bound bug in Java's variant binary path that can emit a bound below the true max (wrong pruning for binary variant fields > 16 bytes). Go does the correct thing here - the one spot it intentionally differs from Java today. I'll file that against Java and link it.

table/internal/variant_bounds.go
```diff
@@ -0,0 +1,368 @@
...
+type variantFieldBound struct {
+	jsonPath     string
+	icebergType  iceberg.PrimitiveType
+	lower, upper iceberg.Literal
+}
+
+const variantBoundTruncateLen = 16
+
+// truncateVariantBound truncates string/binary leaf bounds to variantBoundTruncateLen; a nil upper drops that bound.
+func truncateVariantBound(it iceberg.PrimitiveType, lo, hi iceberg.Literal) (iceberg.Literal, iceberg.Literal) {
```
zeroshade: This truncates to a hardcoded `variantBoundTruncateLen = 16`, but the primitive path derives its length from the column's metrics mode — `createStatsAgg(..., statsCol.Mode.Len)` (parquet_files.go:675). So for a column configured `truncate(8)`, `truncate(32)`, or `full`, primitive bounds honor that setting while these variant child bounds stay capped at 16.

Could you thread the parent variant column's mode through instead? `collectVariantBounds` already has `statsCols[parentID].Mode` in scope at the call site (parquet_files.go:1132), so passing the mode (or its `Len`) into `truncateVariantBound` and applying:
- `Truncate` ⇒ use `Mode.Len`
- `Full` ⇒ no truncation (return `lo`/`hi` unchanged)

would make variant child bounds consistent with primitive columns. Default `truncate(16)` behavior is unchanged.

zeroshade (changes requested):
Thanks for the quick turnaround on the truncate fix, @nssalian — reusing `TruncateUpperBoundText`/`TruncateUpperBoundBinary` and dropping an upper bound that can't be truncated up is exactly right, and it's correctness-safe.

One refinement before I merge: it truncates to a hardcoded `16` rather than the column's configured metrics length, so it diverges from primitive columns for any non-default config (details inline). Honoring the parent's `Mode.Len` (and skipping truncation entirely for `full`) closes the loop on my earlier nit and keeps variant child bounds consistent with top-level columns. That's the last item from me — happy to merge as soon as it's in.

table/internal/variant_bounds.go
nssalian: Done - threaded the parent column's metrics mode through, so variant string/binary child bounds match the primitive path: `truncate(N)` clips to N, `full` leaves them untruncated, default `truncate(16)` unchanged. Tests cover `truncate(8)`, `full`, and binary.

One heads-up: Java (`ParquetVariantUtil`) still hardcodes 16 regardless of mode,  so Go and Java bounds differ in length under non-default modes (both accurate). There was a binary round-down bug is gone now that [#16880](https://github.com/apache/iceberg/pull/16880) merged, so binary bounds otherwise agree. I'll clean up this in Java in a follow-up PR, but the Go side is correct now and ready for review.

zeroshade (approved):
Re-reviewed `473a6fda`. The mode threading is exactly what I was after: `truncateVariantBound` now takes the parent column's `Mode.Len` (parquet_files.go:1132), so `truncate(N)` clips to N, `full` (Len 0) leaves child bounds untruncated, and the default `truncate(16)` is unchanged — consistent with the primitive path. Test coverage is solid too (full-unchanged, truncate-and-increment upper, all-max upper dropped, and the nil-upper omitted from the serialized object).

Thanks for the quick iterations, @nssalian — LGTM.

PR #1479, opened 2026-07-15
refactor(table): centralize transaction metadata checks

### LINKED ISSUE #1431: refactor(table): centralize transaction metadata initialization checks
## Context

PR #1388 added initialization error handling to transactions. It also exposed a maintainability problem: many transaction entry points access `t.meta` directly, and each one has to remember to call `ensureInitialized()` first. The original review missed `UpdateSchema` and `UpdateSpec`, which is exactly the kind of omission this issue is about.

## Proposal

Add one internal accessor for transaction metadata and use it anywhere transaction code needs the metadata builder:

```go
func (t *Transaction) txnMeta() (*MetadataBuilder, error) {
	if err := t.ensureInitialized(); err != nil {
		return nil, err
	}

	return t.meta, nil
}
```

Callers should obtain the builder through this accessor before reading or mutating transaction metadata:

```go
meta, err := t.txnMeta()
if err != nil {
	return err
}

currentSchema := meta.CurrentSchema()
```

The nil transaction case should return a normal error instead of falling through to a nil dereference.

## Scope

- Audit transaction methods and replace direct metadata access with `txnMeta()`.
- Update `NewUpdateSpec` and `NewUpdateSchema` to use the same path.
- Keep the existing public transaction constructors and their error behavior unchanged.
- Add regression coverage for nil transactions, broken transaction initialization, and representative transaction entry points.

## Acceptance criteria

- Metadata access from transaction entry points consistently passes through the checked accessor.
- Broken or nil transactions return an error instead of panicking.
- Existing valid transaction behavior is unchanged.
- The regression tests cover the paths that previously missed `ensureInitialized()`.

Related: #1388

Revanth14:
Closes #1431.

## Summary

- Add `txnMeta()` as the canonical checked accessor for transaction metadata.
- Route transaction entry points, builders, and internal helpers through the accessor.
- Return initialization errors for nil and broken transactions instead of panicking.
- Preserve fresh metadata reads after staged property updates rebuild the metadata builder.
- Add regression coverage for representative transaction operations.

## Testing

- `go test -count=1 ./table`
- `go test ./...`

### DIFF
--- a/table/addfiles_merge_regression_test.go
+++ b/table/addfiles_merge_regression_test.go
@@ -32,7 +32,7 @@ package table_test
 //
 // to:
 //
-//	updater := t.appendSnapshotProducer(fs, snapshotProps)
+//	updater := t.appendSnapshotProducer(meta, fs, snapshotProps)
 
 import (
 	"context"
--- a/table/commit.go
+++ b/table/commit.go
@@ -54,10 +54,14 @@ type TableCommit struct {
 // commit endpoint returns 204 No Content (no metadata), callers must
 // LoadTable after a successful CommitTransaction if they need updated state.
 func (t *Transaction) TableCommit() (TableCommit, error) {
+	if err := t.checkNotNil(); err != nil {
+		return TableCommit{}, err
+	}
 	t.mx.Lock()
 	defer t.mx.Unlock()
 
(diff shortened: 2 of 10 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1480, opened 2026-07-15
fix(promotion): Require equal scale in decimal type promotion

rambleraptor:
When promoting decimals, the spec states that you can only widen precision. iceberg-go currently allows widening scale + precision.

We should match the spec + the other implementations.


rambleraptor:
@zeroshade mind taking a look? It looks like you wrote this area originally + I want to make sure that I'm matching your vision

zeroshade (approved):
LGTM, thanks!

PR #1481, opened 2026-07-16
fix(catalog): compare multi-table identifiers by component

fallintoplace:
## Summary

- compare structured table identifiers component by component when detecting duplicates
- keep dot-joined names only for human-readable errors
- clone identifiers retained for duplicate checks and post-commit reloads

## Problem

Multi-table transactions flattened identifiers with `strings.Join(identifier, ".")`. Distinct valid identifiers such as `["a.b", "c"]` and `["a", "b.c"]` both became `a.b.c`, so adding the second table incorrectly returned a duplicate-table error.

## Fix

Duplicate detection now compares the identifier slices directly with `slices.Equal`. The regression commits both formerly colliding tables and still verifies that an exact duplicate is rejected.

## Testing

- `go test ./catalog`
- `go test -race ./catalog`
- `go vet ./catalog`

### DIFF
--- a/catalog/multi_table_transaction.go
+++ b/catalog/multi_table_transaction.go
@@ -48,12 +48,12 @@ import (
 //
 //	err = mtx.Commit(ctx)
 type MultiTableTransaction struct {
-	cat       TransactionalCatalog
-	loader    Catalog
-	txns      []*table.Transaction
-	idents    []string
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1482, opened 2026-07-16
fix(table): avoid panic pruning an empty snapshot log

fallintoplace:
## Summary

- avoid indexing the rebuilt snapshot log when it has no entries
- preserve an accepted empty log while retaining the current snapshot
- continue requiring the final entry of every nonempty log to reference the current snapshot

## Problem

Metadata can contain a current snapshot and an empty `snapshot-log`. When `RemoveSnapshots` removed a different snapshot, log maintenance rebuilt the empty log and unconditionally indexed its last entry, causing a slice-bounds panic.

## Fix

The current-snapshot consistency check now runs only when the rebuilt log contains an entry. No history is synthesized, and validation for nonempty logs is unchanged.

## Testing

- `go test ./table`
- targeted regression tests under `go test -race`
- `go vet ./table`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1073,7 +1073,7 @@ func (b *MetadataBuilder) updateSnapshotLog() error {
 				newSnapsLog = make([]SnapshotLogEntry, 0, len(b.snapshotLog)-len(newSnapsLog))
 			}
 		}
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1483, opened 2026-07-16
fix(hadoop): recognize UUID metadata versions above 99999

fallintoplace:
## Summary

- accept UUID-style metadata version prefixes with five or more digits
- retain the existing five-digit minimum and checked integer conversion
- cover the 99,999 to 100,000 boundary, gzip metadata, larger versions, short prefixes, and overflow

## Problem

Java formats these metadata versions with `%05d` and PyIceberg uses `:05d`; both specify a minimum width, not a maximum. Version 100,000 therefore produces a six-digit filename. The previous exact-five-digit regex ignored that file and could make the Hadoop catalog select stale metadata.

## Fix

The UUID filename pattern now accepts `{5,}` digits. Discovery still rejects short nonstandard prefixes and versions that overflow `int`, and numeric ordering continues to select the latest file.

## Testing

- `go test ./catalog/hadoop`
- `go test -race ./catalog/hadoop`
- `go vet ./catalog/hadoop`


zeroshade (approved):
LGTM, thanks!

PR #1484, opened 2026-07-16
fix(rest): stop pagination on repeated page tokens

fallintoplace:
## Summary

- route namespace, table, view, and function listings through one pagination loop
- detect repeated nonempty page tokens before issuing another request
- preserve iterator results and early termination behavior

## Problem

REST listing loops stopped only when `next-page-token` was empty. A server returning `A → A` or `A → B → A` caused the client to request pages indefinitely and repeatedly yield duplicate objects.

## Fix

The shared pagination loop records every next-page token before requesting it. If a later response repeats a token, pagination stops and returns an error wrapping `ErrRESTError`. The opaque token itself is not copied into the error.

The end-to-end regression exercises `ListTables` with an `A → B → A` cycle, verifies all three received pages are delivered, and fails if a fourth request occurs. Existing pagination tests continue to cover normal multi-page results, subsequent-page failures, and callers stopping iteration early.

## Testing

- `go test ./catalog/rest`
- `go test -race ./catalog/rest`
- `go vet ./catalog/rest`

### DIFF
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -1079,29 +1079,46 @@ func (r *Catalog) fetchTableCreds(ctx context.Context, ident []string, location 
 	return resolveStorageCredentials(ret.StorageCredentials, location), nil
 }
 
+type identifierPageFetcher func(pageToken string) ([]table.Identifier, string, error)
+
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1485, opened 2026-07-16
fix(io): avoid deadlock in MemFS WalkDir callbacks

fallintoplace:
## Summary

- snapshot matching MemFS paths and sizes under the read lock
- release the lock before building entries and invoking `WalkDir` callbacks
- cover callbacks that remove visited files without hanging

## Problem

`MemFS.WalkDir` invoked callbacks while holding `RLock`. A callback calling `Remove` or `WriteFile` then waited for the write lock, but `WalkDir` could not release its read lock until that callback returned.

## Fix

Treat each walk as a point-in-time snapshot of matching paths and sizes. Root normalization and callback entry construction happen outside the critical section, keeping the write-blocking window limited to reading the in-memory map. Existing filtering, metadata, map iteration order, and callback error behavior remain unchanged.

## Testing

- `go test ./io -run TestMemIO_WalkDir -count=20`
- `go test ./io`
- `go test -race ./io`
- `go vet ./io`

### DIFF
--- a/io/mem.go
+++ b/io/mem.go
@@ -106,16 +106,29 @@ func (m *MemFS) WriteFile(name string, content []byte) error {
 }
 
 func (m *MemFS) WalkDir(root string, fn fs.WalkDirFunc) error {
-	m.mu.RLock()
-	defer m.mu.RUnlock()
+	type walkEntry struct {
+		path string
+		size int64
+	}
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, thanks!

PR #1486, opened 2026-07-17
fix(hive): synchronize HMS table metadata on commit

fallintoplace:
## Problem

Hive commits only advanced `metadata_location`. The HMS storage descriptor and table properties could stay stale after schema, location, or property updates. The update helper also reused the original parameters map.

## Changes

- refresh HMS columns and location from the committed Iceberg metadata
- apply property additions, changes, and removals while preserving unrelated HMS parameters
- maintain current and previous metadata pointers and reserved Iceberg parameters
- copy the parameters map and storage descriptor before updating them

## Testing

Added coverage for schema and location changes, property synchronization, reserved parameters, preserved HMS fields, and input immutability.

- `go test ./catalog/hive`

### DIFF
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -582,7 +582,12 @@ func (c *Catalog) CommitTable(ctx context.Context, identifier table.Identifier, 
 	}
 
 	if current != nil {
-		updatedHiveTbl := updateHiveTableForCommit(currentHiveTbl, staged.MetadataLocation())
+		updatedHiveTbl := updateHiveTableForCommit(
+			currentHiveTbl,
+			current.Metadata(),
+			staged.Metadata(),
+			staged.MetadataLocation(),
+		)
 
 		if err := c.client.AlterTable(ctx, database, tableName, updatedHiveTbl); err != nil {
 			return nil, "", fmt.Errorf("failed to commit table %s.%s: %w", database, tableName, err)
--- a/catalog/hive/hive_test.go
+++ b/catalog/hive/hive_test.go
@@ -715,6 +715,75 @@ func TestHiveCommitTableValidatesRequirementsForMissingTable(t *testing.T) {
 	mockClient.AssertExpectations(t)
 }
 
+func TestHiveCommitTableSynchronizesHMSMetadata(t *testing.T) {
+	ctx := context.Background()
+	oldLocation := "file://" + filepath.Join(t.TempDir(), "old-location")
+	newLocation := "file://" + filepath.Join(t.TempDir(), "new-location")
+	oldMetadataLocation := oldLocation + "/metadata/v1.metadata.json"
+	currentProps := iceberg.Properties{
+		"changed":    "old",
+		"removed":    "value",
+		GCEnabledKey: "false",
+	}
+	current, err := table.NewMetadata(
+		testSchema, iceberg.UnpartitionedSpec, table.UnsortedSortOrder,
+		oldLocation, currentProps,
+	)
+	require.NoError(t, err)
+	require.NoError(t, cataloginternal.WriteTableMetadata(
+		current, iceio.LocalFS{}, oldMetadataLocation, table.MetadataCompressionCodecNone,
+	))
+
+	existing := constructHiveTable(
+		"test_database", "test_table", oldLocation, oldMetadataLocation, testSchema, currentProps,
+	)
+	existing.Parameters["hms-owned"] = "keep"
+	newSchema := iceberg.NewSchema(
+		1,
+		iceberg.NestedField{ID: 1, Name: "renamed", Type: iceberg.PrimitiveTypes.String},
+	)
+	mockClient := &mockHiveClient{}
+	expectImmediateTableLock(mockClient, 1)
+	mockClient.On("GetTable", mock.Anything, "test_database", "test_table").
+		Return(existing, nil).Once()
+	var altered *hive_metastore.Table
+	mockClient.On("AlterTable", mock.Anything, "test_database", "test_table", mock.AnythingOfType("*hive_metastore.Table")).
+		Run(func(args mock.Arguments) {
+			altered = args.Get(3).(*hive_metastore.Table)
+		}).Return(nil).Once()
+	cat := NewCatalogWithClient(mockClient, iceberg.Properties{})
+
+	_, metadataLocation, err := cat.CommitTable(
+		ctx,
+		TableIdentifier("test_database", "test_table"),
+		nil,
+		[]table.Update{
+			table.NewAddSchemaUpdate(newSchema),
+			table.NewSetCurrentSchemaUpdate(newSchema.ID),
+			table.NewSetLocationUpdate(newLocation),
+			table.NewSetPropertiesUpdate(iceberg.Properties{
+				"changed":    "new",
+				"added":      "value",
+				GCEnabledKey: "true",
+			}),
+			table.NewRemovePropertiesUpdate([]string{"removed"}),
+		},
+	)
+
+	require.NoError(t, err)
+	require.NotNil(t, altered)
+	require.Equal(t, metadataLocation, altered.Parameters[MetadataLocationKey])
+	require.Equal(t, oldMetadataLocation, altered.Parameters[PreviousMetadataLocationKey])
+	require.Equal(t, "new", altered.Parameters["changed"])
+	require.Equal(t, "value", altered.Parameters["added"])
+	require.NotContains(t, altered.Parameters, "removed")
+	require.Equal(t, "keep", altered.Parameters["hms-owned"])
+	require.Equal(t, "true", altered.Parameters[ExternalTablePurgeKey])
+	require.Equal(t, newLocation, altered.Sd.Location)
+	require.Equal(t, schemaToHiveColumns(newSchema), altered.Sd.Cols)
+	mockClient.AssertExpectations(t)
+}
+
 func TestHiveDropTableLockFailureIsRetryable(t *testing.T) {
 	mockClient := &mockHiveClient{}
 	mockClient.On("Lock", mock.Anything, mock.AnythingOfType("*hive_metastore.LockRequest")).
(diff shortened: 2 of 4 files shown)


tanmayrauth (approved):
LGTM, thanks for the fix.

catalog/hive/hive.go
```diff
@@ -532,7 +532,14 @@ func (c *Catalog) CommitTable(ctx context.Context, identifier table.Identifier,
 	}
 
 	if current != nil {
-		updatedHiveTbl := updateHiveTableForCommit(currentHiveTbl, staged.MetadataLocation())
+		updatedHiveTbl := updateHiveTableForCommit(
+			currentHiveTbl,
+			current.Properties(),
+			staged.Properties(),
```
laskoviymishka: six positional args with two adjacent `iceberg.Properties` (`currentProps`, `newProps`) is easy to transpose silently — swap them and the diff logic quietly does the wrong thing.

Since there's a single call site, passing the current/staged `Metadata` (or a small delta struct) instead of deconstructed fields would make the swap impossible. Not blocking, but worth doing while it's fresh.

catalog/hive/schema.go
```diff
@@ -160,22 +161,51 @@ func constructHiveTable(dbName, tableName, location, metadataLocation string, sc
...
-	// Update parameters
+	updated.Parameters = maps.Clone(existing.Parameters)
 	if updated.Parameters == nil {
 		updated.Parameters = make(map[string]string)
 	}
 
-	// Store previous metadata location
-	if oldLocation, ok := updated.Parameters[MetadataLocationKey]; ok {
+	for key := range currentProps {
+		delete(updated.Parameters, key)
```
laskoviymishka, line 208: this assumes the HMS `Parameters` we're deleting are exactly the Iceberg user-properties we last wrote — but old commits only ever stamped `metadata_location`, so any property a table already had in HMS (or anything an operator set directly that happens to collide with an Iceberg key name) gets silently dropped on the first commit after this lands.

I'd at minimum document that invariant right here as a known limitation, or gate the delete on an allow-list of Iceberg key prefixes so we're not reaching into keys we didn't write. wdyt?

catalog/hive/schema.go
```diff
@@ -160,22 +161,51 @@ func constructHiveTable(dbName, tableName, location, metadataLocation string, sc
...
+	}
+	for key, value := range newProps {
+		updated.Parameters[key] = value
+	}
+	delete(updated.Parameters, PreviousMetadataLocationKey)
+
+	if oldLocation, ok := existing.Parameters[MetadataLocationKey]; ok {
```
laskoviymishka, line 220: nice that this reads `existing.Parameters` and not `updated.Parameters` — if `newProps` ever carried `metadata_location`, reading the mutated map would capture the wrong previous location. It's subtle enough that a one-line comment on why would save the next person from "simplifying" it back.

catalog/hive/schema.go
```diff
@@ -160,22 +161,51 @@ func constructHiveTable(dbName, tableName, location, metadataLocation string, sc
...
 		updated.Parameters[PreviousMetadataLocationKey] = oldLocation
 	}
 
-	// Set new metadata location
+	updated.Parameters[TableTypeKey] = TableTypeIceberg
```
laskoviymishka, line 224: Java's `HMSTablePropertyHelper` translates `gc.enabled` → `external.table.purge` when it syncs (`ICEBERG_TO_HMS_TRANSLATION`). We copy `gc.enabled` through verbatim but never write `external.table.purge`, so Hive's own `DROP TABLE` won't honour the GC intent — `gc.enabled=false` won't stop the purge, and `gc.enabled=true` won't trigger it.

This was inert before because commits only touched `metadata_location`; now that we sync every property on commit it's live. I'd mirror `gc.enabled` into `external.table.purge` here, and match the same in `constructHiveTable` so both paths agree.

catalog/hive/schema.go
```diff
@@ -160,22 +161,51 @@ func constructHiveTable(dbName, tableName, location, metadataLocation string, sc
...
 	updated.Parameters[MetadataLocationKey] = newMetadataLocation
+	updated.Parameters[ExternalKey] = "TRUE"
+	updated.Parameters["storage_handler"] = "org.apache.iceberg.mr.hive.HiveIcebergStorageHandler"
```
laskoviymishka: `storage_handler` and the handler class are string literals here and again in `constructHiveTable` (and now the test) — every other reserved key has a named constant in `options.go`. I'd pull `StorageHandlerKey` / `IcebergStorageHandler` out and use them in all three spots.

catalog/hive/schema.go
```diff
@@ -160,22 +161,51 @@ func constructHiveTable(dbName, tableName, location, metadataLocation string, sc
...
+
+	if existing.Sd == nil {
+		updated.Sd = constructHiveTable(
```
laskoviymishka: we call `constructHiveTable` here only to pull its `.Sd`, but `.Sd` is derived entirely from location and schema — `newProps` doesn't affect it, so passing it reads as if it does. A small `buildIcebergSD(location, schema)` helper called from both spots would be clearer.

Also nothing covers this branch — `existing.Sd == nil` is never hit by the test. wdyt about a case that constructs `existing` with a nil `Sd`?

catalog/hive/schema.go
```diff
@@ -160,22 +161,51 @@ func constructHiveTable(dbName, tableName, location, metadataLocation string, sc
...
+			existing.DbName,
+			existing.TableName,
+			newLocation,
+			newMetadataLocation,
+			newSchema,
+			newProps,
+		).Sd
+	} else {
+		storageDescriptor := *existing.Sd
```
laskoviymishka: `storageDescriptor := *existing.Sd` is a shallow copy — `SerdeInfo` (a `*SerDeInfo`) and the SD-level `Parameters` map stay aliased to `existing.Sd`. It's inert today since we only replace `Cols` and `Location` and discard `existing` after `AlterTable`, but a real HMS response populates `SerdeInfo.Parameters`, so this is a footgun waiting on the next edit.

I'd clone `SerdeInfo` (and its `Parameters` map) when it's non-nil, consistent with the `maps.Clone` we already do for the table `Parameters`.

catalog/hive/schema.go
```diff
@@ -160,22 +161,51 @@ func constructHiveTable(dbName, tableName, location, metadataLocation string, sc
...
+		storageDescriptor.Cols = schemaToHiveColumns(newSchema)
```
laskoviymishka: `schemaToHiveColumns` calls `newSchema.Fields()` straight off the pointer, so a nil `newSchema` panics here (and in the `constructHiveTable` branch above). `staged.Schema()` is non-nil in practice so this is latent, but the six-arg signature makes a nil mis-call easy. A nil guard or a documented precondition would be enough.

catalog/hive/hive_test.go
```diff
@@ -888,6 +888,62 @@ func TestConstructHiveTablePreservesReservedParameters(t *testing.T) {
 	assert.Equal("gzip", hiveTbl.Parameters["write.metadata.compression"])
 }
 
+func TestUpdateHiveTableForCommitSynchronizesMetadata(t *testing.T) {
```
laskoviymishka, line 1096: the test drives `updateHiveTableForCommit` directly, which is great for the merge logic, but nothing exercises the real path through `CommitTable` — that `current.Properties()` maps to `currentProps` and `staged.Properties()` to `newProps`, and that `AlterTable` actually receives the merged table.

Given the arg list grew to six positional params, a mock `CommitTable` test asserting the `AlterTable` payload has the new columns and merged params would lock the wiring down. Right now a swapped-arg regression would sail through.

catalog/hive/hive_test.go
```diff
@@ -888,6 +888,62 @@ func TestConstructHiveTablePreservesReservedParameters(t *testing.T) {
...
+	require.Equal(t, TableTypeIceberg, updated.Parameters[TableTypeKey])
+	require.Equal(t, "TRUE", updated.Parameters[ExternalKey])
+	require.Equal(t, "org.apache.iceberg.mr.hive.HiveIcebergStorageHandler", updated.Parameters["storage_handler"])
+	require.Equal(t, "s3://bucket/old-location/metadata/v1.metadata.json", updated.Parameters[PreviousMetadataLocationKey])
+	require.Equal(t, "s3://bucket/new-location/metadata/v2.metadata.json", updated.Parameters[MetadataLocationKey])
+	require.Equal(t, "s3://bucket/new-location", updated.Sd.Location)
+	require.Equal(t, schemaToHiveColumns(newSchema), updated.Sd.Cols)
+	require.Equal(t, originalStorageDescriptor.InputFormat, updated.Sd.InputFormat)
+	require.Equal(t, originalStorageDescriptor.OutputFormat, updated.Sd.OutputFormat)
+	require.Equal(t, originalStorageDescriptor.SerdeInfo, updated.Sd.SerdeInfo)
```
laskoviymishka, line 1164: this passes whether `SerdeInfo` is a fresh copy or the same pointer — `require.Equal` is a deep-equal, and here it's literally the same pointer, so it doesn't prove the independence the copy is supposed to give us.

To actually assert it, mutate `updated.Sd.SerdeInfo.Parameters` after the call and check `existing.Sd.SerdeInfo.Parameters` is untouched (or at least assert the pointers differ). Pairs with the shallow-copy note on the SD.

laskoviymishka (changes requested):
Thanks for tackling this — syncing the full HMS table state on commit rather than just `metadata_location` is the right fix, and the new test covers the merge path well.

I'd hold it before merging, mostly because syncing every property on every commit makes two things that were dormant suddenly matter.

The property diff deletes all of `currentProps` from HMS before applying `newProps`. That's correct only if HMS's user-properties were last written by this code, but old commits only ever stamped `metadata_location` — so any property already in HMS, or one an operator set directly that happens to share a key name with an Iceberg property, gets silently dropped on the first commit after this lands. I'd either document that invariant right at the delete loop or gate it on an allow-list of Iceberg key prefixes.

Separately, `gc.enabled` is synced through verbatim but never translated to `external.table.purge` the way Java's `HMSTablePropertyHelper` does. Before this PR that was inert; now that we push all properties on commit, a table with `gc.enabled=false` won't get `external.table.purge=false` in HMS, so Hive's own `DROP TABLE` won't honour the GC intent either way.

- a test through the real `CommitTable` path (not just the helper directly) — the arg list grew to six positional params, two of them adjacent `Properties`, so a swap would currently go unnoticed
- the SD copy is shallow (`SerdeInfo` and its `Parameters` stay aliased), and the `SerdeInfo` assertion passes by identity so it doesn't catch it
- nil `newSchema` panics in `schemaToHiveColumns`, and the nil-`Sd` branch is untested

laskoviymishka (approved):
🚢 👍

PR #1487, opened 2026-07-17
fix(glue): apply catalog properties when loading metadata

fallintoplace:
## Problem

Glue loaded table metadata with a nil property map in both registration and normal table loading. Filesystem implementations that depend on catalog configuration could therefore fail even though the catalog was configured correctly.

## Changes

Pass the Glue catalog properties to the filesystem factory used by `RegisterTable` and `convertGlueToIceberg`.

## Testing

Added a property-dependent custom filesystem and exercised both registration and normal loading.

- `go test ./catalog/glue`

### DIFF
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -288,7 +288,7 @@ func (c *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 		ctx,
 		[]string{tableName},
 		metadataLocation,
-		io.LoadFSFunc(nil, metadataLocation),
+		io.LoadFSFunc(c.props, metadataLocation),
 		c,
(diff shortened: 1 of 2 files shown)


laskoviymishka (approved):
🚢 👍

iremcaginyurtturk:
Hi, thanks for the fix! Would it be possible to get this merged soon? 🙏

fallintoplace:
@iremcaginyurtturk thank you for the comment. Normally this doesn't take too long.

fallintoplace:
@iremcaginyurtturk it is now merged.

iremcaginyurtturk:
@fallintoplace Thanks for informing!

PR #1488, opened 2026-07-17
fix(rest): reject nil view versions

fallintoplace:
## Problem

`CreateView` dereferenced the supplied view version without checking it, so a nil version caused a panic instead of a normal validation error.

## Changes

Reject nil view versions with `iceberg.ErrInvalidArgument` before copying or modifying the version.

## Testing

Added a regression test that passes a nil version and verifies the returned sentinel error.

- `go test ./catalog/rest`


laskoviymishka (approved):
Nice smol one!

PR #1489, opened 2026-07-17
fix(sql): reject table-view name collisions

fallintoplace:
## Problem

SQL tables and views share the same catalog key, but creation only checked for an existing object of the same kind. A table could write metadata where a view already existed, or vice versa, before the database rejected the catalog insert. A concurrent creator could also win after the prechecks and leave the losing request's metadata file behind.

## Changes

- reject table creation when the identifier is already registered as either a table or a view
- reject view creation when the identifier is already registered as either a view or a table
- return the matching catalog sentinel for conflicts found by the prechecks
- check and document table existence as catalog registration rather than metadata-file reachability
- remove newly written metadata when a catalog insert fails, using a non-cancelled cleanup context and the catalog filesystem properties

The prechecks make ordinary duplicate and cross-kind conflicts deterministic. A genuinely concurrent creator can still win between the precheck and insert, so the losing request keeps the database insert error; cleanup prevents that race from leaking its metadata file.

## Testing

- both table/view collision directions with unchanged metadata directories
- duplicate-table creation returns `ErrTableAlreadyExists`
- forced catalog insert failures remove table and view metadata
- `go test ./catalog/sql`
- `go test -race ./catalog/sql`
- `go vet ./catalog/sql`

### DIFF
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -135,7 +135,8 @@ type Catalog interface {
 	// RenameTable tells the catalog to rename a given table by the identifiers
 	// provided, and then loads and returns the destination table
 	RenameTable(ctx context.Context, from, to table.Identifier) (*table.Table, error)
-	// CheckTableExists returns if the table exists
+	// CheckTableExists reports whether the identifier is registered as a table in
+	// the catalog. It does not validate that the table metadata file is readable.
 	CheckTableExists(ctx context.Context, identifier table.Identifier) (bool, error)
 	// ListNamespaces returns the list of available namespaces, optionally filtering by a
 	// parent namespace
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -135,8 +135,7 @@ type Catalog interface {
 	// RenameTable tells the catalog to rename a given table by the identifiers
 	// provided, and then loads and returns the destination table
 	RenameTable(ctx context.Context, from, to table.Identifier) (*table.Table, error)
-	// CheckTableExists reports whether the identifier is registered as a table in
-	// the catalog. It does not validate that the table metadata file is readable.
+	// CheckTableExists reports whether the table exists.
 	CheckTableExists(ctx context.Context, identifier table.Identifier) (bool, error)
 	// ListNamespaces returns the list of available namespaces, optionally filtering by a
 	// parent namespace
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -770,7 +770,7 @@ func (c *Catalog) checkIdentifierAvailable(
 	}
 
 	identifierName := strings.Join(identifier, ".")
-	if !c.isV0() && entry.IcebergType.String == ViewType {
+	if !c.isV0() && entry.IcebergType.Valid && entry.IcebergType.String == ViewType {
 		return fmt.Errorf("%w: %s", catalog.ErrViewAlreadyExists, identifierName)
 	}
 
@@ -846,9 +846,11 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 		return nil
 	})
 	if err != nil {
-		if collisionErr := c.checkIdentifierAvailableInCatalog(ctx, ns, tblIdent, ident); errors.Is(collisionErr, catalog.ErrTableAlreadyExists) ||
-			errors.Is(collisionErr, catalog.ErrViewAlreadyExists) {
-			err = collisionErr
+		if !errors.Is(err, catalog.ErrTableAlreadyExists) && !errors.Is(err, catalog.ErrViewAlreadyExists) {
+			if collisionErr := c.checkIdentifierAvailableInCatalog(ctx, ns, tblIdent, ident); errors.Is(collisionErr, catalog.ErrTableAlreadyExists) ||
+				errors.Is(collisionErr, catalog.ErrViewAlreadyExists) {
+				err = collisionErr
+			}
 		}
 		if cleanupErr := removeUncommittedMetadata(ctx, staged.MetadataLocation(), staged.FS); cleanupErr != nil {
 			err = errors.Join(err, cleanupErr)
@@ -1597,9 +1599,11 @@ func (c *Catalog) CreateView(ctx context.Context, identifier table.Identifier, s
 		return nil
 	})
 	if err != nil {
-		if collisionErr := c.checkIdentifierAvailableInCatalog(ctx, ns, viewIdent, identifier); errors.Is(collisionErr, catalog.ErrTableAlreadyExists) ||
-			errors.Is(collisionErr, catalog.ErrViewAlreadyExists) {
-			err = collisionErr
+		if !errors.Is(err, catalog.ErrTableAlreadyExists) && !errors.Is(err, catalog.ErrViewAlreadyExists) {
+			if collisionErr := c.checkIdentifierAvailableInCatalog(ctx, ns, viewIdent, identifier); errors.Is(collisionErr, catalog.ErrTableAlreadyExists) ||
+				errors.Is(collisionErr, catalog.ErrViewAlreadyExists) {
+				err = collisionErr
+			}
 		}
 		loadFS := io.LoadFSFunc(c.props, metadataLocation)
 		if cleanupErr := removeUncommittedMetadata(ctx, metadataLocation, loadFS); cleanupErr != nil {
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -324,7 +324,6 @@ func retrySerializableWriteTx(ctx context.Context, run func() error) error {
 		if attempt == serializableWriteMaxAttempts-1 {
 			break
 		}
-
 		delay := serializableWriteRetryDelay << attempt
 		timer := time.NewTimer(delay)
 		select {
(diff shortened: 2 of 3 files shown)


catalog/sql/sql.go
```diff
@@ -1285,6 +1333,12 @@ func (c *Catalog) CreateView(ctx context.Context, identifier table.Identifier, s
 
 		return nil
 	})
+	if err != nil {
+		loadFS := io.LoadFSFunc(props, metadataLocation)
```
laskoviymishka: This cleanup won't run on any backend that keeps credentials in the catalog config.

`props` here is the caller's view-level props, which is nil in the common case — the S3/GCS/ADLS credentials live in `c.props`, which is what `LoadView` and `CheckTableExists` use. So `io.LoadFSFunc(props, ...)` fails to authenticate, `removeUncommittedMetadata` returns an error that just gets joined and returned, and the orphaned metadata file is exactly what stays behind. The table path already gets this right via `staged.FS` (built from `c.props`).

I'd switch this to `io.LoadFSFunc(c.props, metadataLocation)`.

catalog/sql/sql_test.go
```diff
@@ -1972,6 +1972,95 @@ func (s *SqliteCatalogTestSuite) TestCreateView() {
...
+}
+
+func (s *SqliteCatalogTestSuite) TestCreateViewConflictsWithTable() {
+	ctx := context.Background()
+	db := s.getCatalogSqlite()
+	identifier := s.randomTableIdentifier()
+	s.Require().NoError(db.CreateNamespace(ctx, catalog.NamespaceFromIdent(identifier), nil))
+	tbl, err := db.CreateTable(ctx, identifier, tableSchemaNested)
+	s.Require().NoError(err)
+	s.Require().NoError(os.Remove(strings.TrimPrefix(tbl.MetadataLocation(), "file://")))
```
laskoviymishka: This `os.Remove` makes the file-count assertion prove nothing.

After deleting the metadata file, `before` is empty, `CreateView` bails at the `CheckTableExists` guard before it ever writes metadata, so `after` is also empty and `s.Len(after, len(before))` is just `0 == 0`. It never touches the cleanup path, and it leaves the catalog in a state that can't occur in practice (row present, file gone).

I'd drop the `os.Remove`, take `before` right after `CreateTable` (one file present), then assert `after` still equals `before` after the failed `CreateView`.

catalog/sql/sql.go
```diff
@@ -627,6 +640,14 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 		return nil, fmt.Errorf("%w: %s", catalog.ErrNoSuchNamespace, ns)
 	}
 
+	viewExists, err := c.CheckViewExists(ctx, ident)
```
laskoviymishka: The symmetry with `CreateView` isn't quite complete. `CreateView` now returns typed sentinels for both collisions, but `CreateTable` only prechecks the view case — a duplicate *table* still falls through to the raw DB unique-constraint error rather than `ErrTableAlreadyExists`, so callers switching on the error for idempotent-create or "already exists" handling will misclassify it.

I'd add a `CheckTableExists` precheck right after this block. Worth a test for the real duplicate-table case too — the trigger test forces every insert to fail, which is a different path. wdyt?

catalog/sql/sql.go
```diff
@@ -919,16 +944,31 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
 }
 
 func (c *Catalog) CheckTableExists(ctx context.Context, identifier table.Identifier) (bool, error) {
-	_, err := c.LoadTable(ctx, identifier)
+	ns, err := c.namespaceKey(ctx, catalog.NamespaceFromIdent(identifier))
```
laskoviymishka, line 1202: This quietly changes the contract of a public interface method. Before, it went through `LoadTable`, so a table with a missing or unreadable metadata file returned false; now it's a catalog-row `EXISTS` check that returns true whenever the row is present, regardless of metadata reachability.

That's the right behavior for collision detection and matches Java's `JdbcUtil.tableExists`, but it's a behavioral break for anyone relying on the old reachability semantics. I'd add a one-line doc comment here (and on the interface in `catalog/catalog.go`) noting it checks catalog-registry existence, not metadata-file validity.

laskoviymishka (changes requested):
Nice fix — rejecting table/view name collisions closes a real correctness gap, and cleaning up the orphaned metadata file when the catalog insert fails is the right call. I'd hold it before merging though, mostly on one thing.

The view cleanup uses the caller-supplied `props` to build the FS (`io.LoadFSFunc(props, metadataLocation)`), but everywhere else — `LoadView`, `CheckTableExists`, and the table cleanup via `staged.FS` — reads credentials from `c.props`. `props` is nil in the common case, so on any backend that keeps object-store credentials in the catalog config (S3/GCS/ADLS) the FS won't authenticate, `removeUncommittedMetadata` fails, and the error just gets joined and returned — the leaked file stays. That's the exact scenario this PR is meant to prevent, so I'd switch it to `c.props`.

- View cleanup should use `c.props`, not the caller `props`.
- `TestCreateViewConflictsWithTable` is vacuous as written — the `os.Remove` makes the file-count check `0 == 0` and never exercises cleanup. Drop the remove and snapshot after `CreateTable`.
- `CreateView` returns typed sentinels for both collision kinds, but `CreateTable` still returns a raw DB error for a duplicate table — worth mirroring with a `CheckTableExists` precheck (and a test).
- `CheckTableExists` now checks the catalog row rather than metadata reachability — correct, but it's a public-interface change that deserves a doc comment.

One framing note: the description says it "returns the matching catalog sentinel before writing new metadata," but that only holds for the precheck path — a genuine concurrent creator still races past the prechecks and hits the raw DB error. The cleanup shrinks the blast radius (no orphan file) rather than closing the race, so I'd soften that wording a touch.

Once the props and the test are sorted, happy to take another pass and approve.

catalog/sql/sql.go
```diff
@@ -667,6 +680,22 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 		return nil, fmt.Errorf("%w: %s", catalog.ErrNoSuchNamespace, ns)
 	}
 
+	viewExists, err := c.CheckViewExists(ctx, ident)
```
zeroshade: TOCTOU: `CheckViewExists`/`CheckTableExists` run in separate read transactions before the write-tx insert below. A concurrent `CreateView` can pass this check and race the insert; consider rechecking inside the same serializable write transaction (cf. #1466).

catalog/sql/sql.go
```diff
@@ -1329,7 +1379,15 @@ func (c *Catalog) CreateView(ctx context.Context, identifier table.Identifier, s
 		return err
```
zeroshade: (re: catalog/sql/sql.go:1377) Same TOCTOU on the `CreateView` path — a raced insert failure surfaces a raw dialect unique-constraint error rather than the intended catalog sentinel, so error semantics differ across sqlite/postgres/mysql.

zeroshade (changes requested):
Both-direction rejection is correct for the non-raced case (nice work), but the existence checks aren't atomic with the insert — they run in separate read transactions before the `withWriteTx` insert. Under concurrent create, the loser can pass both prechecks then hit the DB unique constraint and surface a raw dialect error instead of `ErrTableAlreadyExists`/`ErrViewAlreadyExists`. Recommend the serializable write-tx + in-tx recheck pattern from #1466 (same file), plus a concurrency test. CI is green.

catalog/catalog.go
```diff
@@ -135,7 +135,8 @@ type Catalog interface {
 	// RenameTable tells the catalog to rename a given table by the identifiers
 	// provided, and then loads and returns the destination table
 	RenameTable(ctx context.Context, from, to table.Identifier) (*table.Table, error)
-	// CheckTableExists returns if the table exists
+	// CheckTableExists reports whether the identifier is registered as a table in
+	// the catalog. It does not validate that the table metadata file is readable.
```
laskoviymishka: This reads as an interface-wide guarantee, but the REST impl falls back to `LoadTable` when the server has no HEAD endpoint, which does fetch and parse the metadata file.

Since the SQL impl already carries its own doc comment saying this just below, I'd scope the strict wording there and keep the interface doc looser here. wdyt?

catalog/sql/sql.go
```diff
@@ -736,6 +735,56 @@ func checkValidNamespace(ident table.Identifier) error {
...
+	err := query.Scan(ctx)
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("error checking identifier availability: %w", err)
+	}
+
+	identifierName := strings.Join(identifier, ".")
+	if !c.isV0() && entry.IcebergType.String == ViewType {
```
laskoviymishka: This is correct today — a NULL `iceberg_type` on a legacy row gives `Valid:false, String:""`, and `"" == ViewType` is false, so we fall through to the table sentinel. But it leans on the zero-value of `String` when `Valid` is false, and it's the only NullString read in this file that doesn't gate on `.Valid` first.

I'd add the explicit check so a future driver that puts garbage in `String` on a NULL column can't misclassify a legacy table row as a view:

```go
if !c.isV0() && entry.IcebergType.Valid && entry.IcebergType.String == ViewType {
```

catalog/sql/sql.go
```diff
@@ -793,6 +846,14 @@ func (c *Catalog) CreateTable(ctx context.Context, ident table.Identifier, sc *i
 		return nil
 	})
 	if err != nil {
+		if collisionErr := c.checkIdentifierAvailableInCatalog(ctx, ns, tblIdent, ident); errors.Is(collisionErr, catalog.ErrTableAlreadyExists) ||
```
laskoviymishka: When the in-tx `checkIdentifierAvailable` already returned a sentinel, this opens a second read tx and re-derives the same collision — an extra round-trip on the contention-free duplicate case, which is the common one. I'd only recheck when the error isn't already one of ours:

```go
if !errors.Is(err, catalog.ErrTableAlreadyExists) && !errors.Is(err, catalog.ErrViewAlreadyExists) {
    if collisionErr := c.checkIdentifierAvailableInCatalog(ctx, ns, tblIdent, ident); ... {
        err = collisionErr
    }
}
```

Same thing over in `CreateView`.

catalog/sql/sql_test.go
```diff
@@ -2123,6 +2123,135 @@ func (s *SqliteCatalogTestSuite) TestCreateView() {
...
+		results <- createResult{objectType: sqlcat.TableType, err: err}
+	}()
+	go func() {
+		<-start
+		err := db.CreateView(ctx, identifier, tableSchemaNested, "SELECT 1", nil)
+		results <- createResult{objectType: sqlcat.ViewType, err: err}
+	}()
+	close(start)
+
+	first, second := <-results, <-results
```
laskoviymishka, line 2194: If both creates happen to fail, the swap leaves `first.err` non-nil and this just blows up here with no hint about what actually broke. Worth asserting exactly one of the two succeeded (or at least a message on this line) so a regression points at the real failure. wdyt?

laskoviymishka (approved):
LGTM!

Left couple nit's inline.

catalog/catalog.go
fallintoplace: Thanks, these all make sense. I’ve applied them in c8c0977

zeroshade (approved):
Thanks for the thorough follow-up — this resolves everything from the last pass:

- **Atomicity fixed.** Both `CreateTable` and `CreateView` now do the existence recheck (`checkIdentifierAvailable`) *inside* the `withSerializableWriteTx` transaction, right before the insert — so the check is atomic with the write.
- **Portable error semantics.** The re-query (`checkIdentifierAvailableInCatalog`) on write-tx failure translates a raw dialect unique-constraint error into `ErrTableAlreadyExists`/`ErrViewAlreadyExists`, so a true race surfaces the catalog sentinel consistently across sqlite/postgres/mysql rather than a dialect-specific string.
- **Symmetry.** A duplicate *table* now returns `ErrTableAlreadyExists` (not the raw "failed to create table"); `TestCreateDuplicatedTable` asserts the sentinel via `ErrorIs`.
- **No orphaned metadata.** `removeUncommittedMetadata` cleans up on failure (table via `staged.FS`, view via `c.props`), covered by `TestCreateRemovesMetadataIfCatalogInsertFails`.
- **Concurrency guard.** `TestConcurrentTableViewCollisionReturnsCatalogSentinel` (WAL + racing create) is exactly the regression test I asked for.

LGTM.

PR #1490, opened 2026-07-17
fix(glue): accept empty root namespace identifiers

fallintoplace:
## Problem

Glue treated an allocated empty identifier differently from nil when listing the root namespace. Callers using `table.Identifier{}` received a hierarchical-namespace error even though it represents the same root.

## Changes

Use identifier length to distinguish the root from unsupported non-empty hierarchical namespaces.

## Testing

Added coverage for an allocated empty root identifier.

- `go test ./catalog/glue`


laskoviymishka (approved):
:shipit:

tanmayrauth (approved):

PR #1491, opened 2026-07-17
fix(hadoop): validate metadata before dropping tables

fallintoplace:
## Problem

`DropTable` considered a directory to be a table based only on the presence of a metadata-shaped filename, then recursively deleted the directory. A partial or corrupt metadata file could therefore authorize a destructive drop.

## Changes

Load and parse the current table metadata before removing the table directory. Invalid metadata now returns an error and leaves the directory untouched.

The existing successful-drop tests now create real table metadata rather than placeholder JSON.

## Testing

Added a regression test with invalid metadata and a marker file that must survive the failed drop.

- `go test ./catalog/hadoop`

### DIFF
--- a/catalog/hadoop/hadoop.go
+++ b/catalog/hadoop/hadoop.go
@@ -887,19 +887,12 @@ func (c *Catalog) ListTables(_ context.Context, ns table.Identifier) iter.Seq2[t
 	}
 }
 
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM. DropTable now loads and validates table metadata before RemoveAll, so bogus or non-Iceberg paths are not purged; valid-table purge behavior is preserved. CI green.

PR #1492, opened 2026-07-17
feat(rest): decode REST scan-task payloads into FileScanTasks

Revanth14:
## Summary

Implements REST scan-task payload decoding as part of [#1178](https://github.com/apache/iceberg-go/issues/1178).

This converts REST `ScanTasks` responses into `table.FileScanTask` values, including data files, delete files, metrics, partitions, bounds, and residual filters.

## Changes

- Adds REST wire types for content files, count maps, bounds, data files, and delete files. These (`RESTContentFile`/`RESTDataFile`/`RESTFileScanTask`/`RESTDeleteFile`) are exported as part of the existing `ScanTasks` response surface introduced as stubs in #1324.
- Adds `DecodeScanTasks` with metadata-aware partition and residual decoding.
- Decodes partition values using their partition transform result types.
- Decodes lower and upper bounds from Java-compatible hexadecimal binary values.
- Resolves envelope-local delete-file references.
- Classifies positional deletes, equality deletes, and Puffin deletion vectors.
- Supports and validates Java's optional `referenced-data-file` field, while deriving the target for deletion vectors when omitted.
- Validates malformed metrics, references, content variants, file sizes, record counts, split offsets, and equality IDs.
- Adds `Residual` to `table.FileScanTask`.
- Extends the existing `FileScanTask` codec with an optional framed residual extension:
  - New decoders can read legacy payloads.
  - Existing decoders continue reading the unchanged Avro envelope prefix.

Bounds fixtures mirror Java's authoritative `TestContentFileParser`, where values are hexadecimal encodings of raw Iceberg binary values, for example:

    "lower-bounds":{"keys":[3,4],"values":["01000000","02000000"]}

## Scope

This PR implements decoding only. It does not add plan-task fanout, `PlanFiles` orchestration, storage-credential handling, capability enablement, or the fake REST planning server.

`DecodeScanTasks` is a standalone building block with no production caller in this PR; `PlanFiles` will invoke it in a follow-up.

`ReadTasks` currently continues applying the Scan's original row filter and does not consume per-task residuals. This preserves correctness for scans using their originating `Scan`, but applying task-specific residuals remains follow-up work before decoded tasks can be consumed independently.

Delete-file references are scoped to each `ScanTasks` response, so callers must decode each response before combining results.

## Testing

    go test ./...
    go test -race ./catalog/rest ./codec ./table
    go vet ./catalog/rest ./codec ./table
    git diff --check

### DIFF
--- a/catalog/rest/scan_task_decoder.go
+++ b/catalog/rest/scan_task_decoder.go
@@ -135,8 +135,8 @@ func DecodeScanTasks(
 		return nil, nil
 	}
 
-	// Decode every delete even if the server did not reference it. This catches
-	// a malformed response rather than silently ignoring corrupt wire data.
+	// Decode every delete before resolving references. This validates all wire
+	// entries, including an unreferenced entry that will be rejected below.
 	deletes := make([]iceberg.DataFile, len(wire.DeleteFiles))
 	for i := range wire.DeleteFiles {
 		decoded, err := decodeRESTDeleteFile(wire.DeleteFiles[i], metadata, "")
@@ -148,6 +148,7 @@ func DecodeScanTasks(
 
 	out := make([]table.FileScanTask, 0, len(wire.FileScanTasks))
 	dvOwners := make(map[int]string)
+	referencedDeletes := make([]bool, len(deletes))
 	for i := range wire.FileScanTasks {
 		wireTask := wire.FileScanTasks[i]
 		if wireTask.DataFile == nil {
@@ -187,6 +188,7 @@ func DecodeScanTasks(
 					ErrRESTError, i, ref)
 			}
 			seenRefs[ref] = struct{}{}
+			referencedDeletes[ref] = true
 
 			wireDelete := wire.DeleteFiles[ref]
 			if wireDelete.ReferencedDataFile != nil && *wireDelete.ReferencedDataFile != dataFile.FilePath() {
@@ -226,6 +228,14 @@ func DecodeScanTasks(
 		}
 
 		out = append(out, task)
+	}
+
+	for i, referenced := range referencedDeletes {
+		if !referenced {
+			return nil, fmt.Errorf(
+				"%w: decoding scan tasks: delete-files[%d] is not referenced by any file scan task",
+				ErrRESTError, i)
+		}
 	}
 
 	return out, nil
--- a/catalog/rest/scan_task_decoder.go
+++ b/catalog/rest/scan_task_decoder.go
@@ -385,13 +385,11 @@ func contentFileBuilder(
 	if wire.Content != wantContent {
 		return nil, "", fmt.Errorf("content is %q, want %q", wire.Content, wantContent)
 	}
-	// record-count and file-size may be 0 per the wire contract (e.g. an empty
-	// data file); only reject the negative values a conformant server never sends.
-	if wire.RecordCount < 0 {
-		return nil, "", fmt.Errorf("record-count must be non-negative: %d", wire.RecordCount)
+	if wire.RecordCount <= 0 {
+		return nil, "", fmt.Errorf("record-count must be positive: %d", wire.RecordCount)
 	}
-	if wire.FileSizeInBytes < 0 {
-		return nil, "", fmt.Errorf("file-size-in-bytes must be non-negative: %d", wire.FileSizeInBytes)
+	if wire.FileSizeInBytes <= 0 {
+		return nil, "", fmt.Errorf("file-size-in-bytes must be positive: %d", wire.FileSizeInBytes)
 	}
 
 	spec := metadata.PartitionSpecByID(wire.SpecID)
--- a/catalog/rest/scan_task_decoder.go
+++ b/catalog/rest/scan_task_decoder.go
@@ -1,0 +1,718 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package rest
+
+import (
+	"bytes"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math"
+	"slices"
+
+	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/table"
+	"github.com/twmb/avro/atype"
+)
+
+// The exported REST wire types below are intentional: the low-level planning
+// methods expose ScanTasks directly, while DecodeScanTasks is the boundary for
+// callers that want table.FileScanTask domain objects.
+
+// RESTCountMap is the REST column-id-to-count representation. Keys and values
+// are parallel arrays because JSON object keys cannot preserve integer IDs.
+type RESTCountMap struct {
+	Keys   []int   `json:"keys"`
+	Values []int64 `json:"values"`
+}
+
+// RESTValueMap is the REST column-id-to-binary representation used for lower
+// and upper bounds. Java's ContentFileParser encodes each raw Iceberg binary
+// value as a hexadecimal string; these are not typed JSON partition values.
+type RESTValueMap struct {
+	Keys   []int    `json:"keys"`
+	Values []string `json:"values"`
+}
+
+// RESTContentFile contains the fields common to REST data and delete files.
+// Partition values remain raw until DecodeScanTasks resolves spec-id and the
+// result type of each partition transform.
+type RESTContentFile struct {
+	SpecID          int               `json:"spec-id"`
+	Partition       []json.RawMessage `json:"partition"`
+	Content         string            `json:"content"`
+	FilePath        string            `json:"file-path"`
+	FileFormat      string            `json:"file-format"`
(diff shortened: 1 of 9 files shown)


tanmayrauth (approved):

catalog/rest/scan_task_decoder.go
```diff
@@ -0,0 +1,718 @@
...
+		return nil, "", fmt.Errorf("content is %q, want %q", wire.Content, wantContent)
+	}
+	// record-count and file-size may be 0 per the wire contract (e.g. an empty
+	// data file); only reject the negative values a conformant server never sends.
+	if wire.RecordCount < 0 {
+		return nil, "", fmt.Errorf("record-count must be non-negative: %d", wire.RecordCount)
+	}
+	if wire.FileSizeInBytes < 0 {
+		return nil, "", fmt.Errorf("file-size-in-bytes must be non-negative: %d", wire.FileSizeInBytes)
+	}
```
dttung2905, line 403: This allows `>= 0`  and passed into `iceberg.NewDataFileBuilder` in line 410 and then the Builder reject `<=0` case
https://github.com/apache/iceberg-go/blob/116e58c1432eb06a730054428f2fc8c0b8a00123/manifest.go#L2352-L2358

I think `0` pass the check here but fail later with a confusing `invalid argument`. Can we reject `0` here early with a clearer error ?

Revanth14: Good catch. Although the wire fields do not declare a positive minimum, NewDataFileBuilder requires both values to be greater than zero. I’ll validate that at the decoding boundary and add zero-value test cases so malformed responses return a clear field-specific error.

catalog/rest/scan_task_decoder.go
```diff
@@ -0,0 +1,716 @@
...
+		if wire.Values[i] < 0 {
+			return nil, fmt.Errorf("%s[%d] for field ID %d is negative", name, i, key)
+		}
+		out[key] = wire.Values[i]
+	}
+
+	return out, nil
+}
+
+func decodeValueMap(name string, wire *RESTValueMap) (map[int][]byte, error) {
```
zeroshade, line 561: Non-blocking / parity note: bounds are stored as raw bytes without validating the field ID against the schema or that the bytes decode for the field's type. That's actually consistent with the manifest/Avro path — it stores `lower_bounds`/`upper_bounds` as `map[int][]byte`, defers to `LiteralFromBytes` at evaluation time, and doesn't validate field IDs either. So I would *not* add eager validation here (it'd also reject the `Fixed`-length padding that `LiteralFromBytes` intentionally tolerates). Flagging only so the parity is a conscious choice rather than an oversight.

catalog/rest/scan_task_decoder.go
```diff
@@ -0,0 +1,716 @@
...
+		fixedSizes,
+		wire.RecordCount,
+		wire.FileSizeInBytes,
+	)
+	if err != nil {
+		return nil, "", err
+	}
+
+	if wire.KeyMetadata != nil {
+		key, err := hex.DecodeString(*wire.KeyMetadata)
```
zeroshade, line 434: Question (non-blocking): `key-metadata` is hex-decoded here. Bounds/binary hex parity with Java's `ContentFileParser` is well covered, but I didn't spot a `key-metadata` parity fixture. Could you confirm Java emits `key-metadata` as a hex string (not base64), ideally with a small fixture? Just closing the last encoding unknown.

catalog/rest/scan_task_decoder.go
```diff
@@ -0,0 +1,716 @@
...
+		if len(wire.DeleteFiles) != 0 {
+			return nil, fmt.Errorf("%w: decoding scan tasks: %d delete files have no file scan tasks", ErrRESTError, len(wire.DeleteFiles))
+		}
+
+		return nil, nil
+	}
+
+	// Decode every delete even if the server did not reference it. This catches
+	// a malformed response rather than silently ignoring corrupt wire data.
+	deletes := make([]iceberg.DataFile, len(wire.DeleteFiles))
```
zeroshade, line 140: Question (non-blocking): decoding every delete even when no task references it (to catch corrupt wire) is a nice touch. But a well-formed yet unreferenced delete is then silently dropped rather than surfaced. Is silent dropping the intended contract, or should an unreferenced delete be an error? Curious whether Java treats a delete with no referencing task as malformed.

catalog/rest/scan_task_decoder_test.go
```diff
@@ -0,0 +1,544 @@
...
+	wire.DeleteFiles[0].ContentSizeInBytes = int64Ptr(20)
+
+	tasks, err := DecodeScanTasks(wire, metadata, metadata.schema, nil)
+	require.NoError(t, err)
+	require.Len(t, tasks, 1)
+	require.Len(t, tasks[0].DeletionVectorFiles, 1)
+	assert.Equal(t, stringPtr("s3://bucket/table/data.parquet"), tasks[0].DeletionVectorFiles[0].ReferencedDataFile())
+}
+
+func TestRESTValueMapMatchesJavaContentFileParser(t *testing.T) {
```
zeroshade, line 203: Non-blocking: this asserts bounds round-trip to raw bytes (correct, and matches how they're stored). Since these vectors mirror Java's `TestContentFileParser`, a `LiteralFromBytes(field.Type, ...)` assertion for a couple of representative types (e.g. int + decimal + timestamp) would additionally pin endianness/type-dispatch. Low priority since it exercises the shared decode path rather than this PR's code.

zeroshade (commented):
Did a thorough pass over the decoder, the codec extension, and the tests — this is in good shape.

- Validation is careful and matches the existing manifest/Avro contract: `record-count`/`file-size` are rejected when `<= 0` (same as `NewDataFileBuilder`), and bounds are stored as raw `map[int][]byte` with typed decode deferred to `LiteralFromBytes` at evaluation time — exactly how manifest reads behave.
- Partition decode uses `field.Transform.ResultType(sourceType)`, consistent with `table/internal/utils.go`, and covers every primitive wire type (incl. hex for fixed/binary).
- The residual codec extension is genuinely backward/forward-compatible: a byte-stable Avro envelope + magic/uvarint trailer, an empty trailer decodes to a nil residual, and the `fileScanTaskShape` compile-time drift guard is a nice touch.
- Malformed-input coverage is excellent (20+ negative cases, plus codec error-marker/negative-range tests).

A few non-blocking nits and two questions inline — nothing that must change, mostly parity confirmations. Happy to approve once you've had a look at the two questions (key-metadata encoding + unreferenced-delete handling).

catalog/rest/scan_task_decoder.go
Revanth14: Agreed, this is intentional. The decoder keeps bounds as raw bytes and defers type interpretation to `LiteralFromBytes`, matching the manifest/Avro path and preserving tolerated `Fixed` padding.

I added int, decimal, and timestamp assertions to pin the binary encoding without introducing eager validation. Thanks for flagging this!

catalog/rest/scan_task_decoder.go
Revanth14: Confirmed Java's `ContentFileParser` passes `key-metadata` through `SingleValueParser` as a binary value, whose JSON representation uses base16 encoding.

I added a focused parity test using the exact value from Java's `TestContentFileParser` fixture. It verifies that the 32-character hex string decodes to 16 bytes, which also distinguishes the encoding from base64. Thanks for catching this gap!

catalog/rest/scan_task_decoder.go
Revanth14: I opted to treat an unreferenced delete as malformed. 
To answer the Java question directly: no, Java's `PlanTableScanResponse.validate()` only rejects `delete-files` when there are no file scan tasks at all ("deleteFiles should only be returned with fileScanTasks that reference them"); with tasks present, an unreferenced delete is silently ignored at parse time. So this is intentionally stricter than Java -  the REST spec describes `delete-files` as "Delete files referenced by file scan tasks", and this enforces the contract Java's own validation message states, per-entry rather than only in the degenerate case.

The decoder now tracks referenced indices and returns a field-specific error for any unused `delete-files` entry, with a test covering it. Happy to relax to Java's ignore-silently behavior if you'd rather match parity here.

catalog/rest/scan_task_decoder_test.go
Revanth14: Added a focused `LiteralFromBytes` test covering int, decimal, and timestamp bounds. It pins the expected little-endian encoding for int/timestamp and two's-complement big-endian encoding for decimal, while keeping the decoder's output as raw bytes. Thanks for the suggestion!

Revanth14:
> Did a thorough pass over the decoder, the codec extension, and the tests — this is in good shape.
> 
> * Validation is careful and matches the existing manifest/Avro contract: `record-count`/`file-size` are rejected when `<= 0` (same as `NewDataFileBuilder`), and bounds are stored as raw `map[int][]byte` with typed decode deferred to `LiteralFromBytes` at evaluation time — exactly how manifest reads behave.
> * Partition decode uses `field.Transform.ResultType(sourceType)`, consistent with `table/internal/utils.go`, and covers every primitive wire type (incl. hex for fixed/binary).
> * The residual codec extension is genuinely backward/forward-compatible: a byte-stable Avro envelope + magic/uvarint trailer, an empty trailer decodes to a nil residual, and the `fileScanTaskShape` compile-time drift guard is a nice touch.
> * Malformed-input coverage is excellent (20+ negative cases, plus codec error-marker/negative-range tests).
> 
> A few non-blocking nits and two questions inline — nothing that must change, mostly parity confirmations. Happy to approve once you've had a look at the two questions (key-metadata encoding + unreferenced-delete handling).

Thanks for the thorough review!

I've addressed the two questions and the additional bounds-test suggestion:

- Confirmed Java encodes `key-metadata` using base16 and added a focused fixture matching `TestContentFileParser`.
- Unreferenced delete files are now rejected with a field-specific error, with test coverage intentionally stricter than Java, which silently ignores them at parse time (details in the inline thread).
- Added `LiteralFromBytes` assertions for int, decimal, and timestamp bounds to pin endianness and type dispatch.

zeroshade (approved):
Looks good! Thanks!

Issue #1289, opened 2026-06-23
Encryption Support


rambleraptor:
### Feature Request / Improvement

iceberg-go will read the `encryption-keys` metadata, but it doesn't support encrypting / decrypting data using that metadata.

I think the biggest part of this will be figuring out the proper interfaces. There will many different implementations of these interfaces over time.

I'd love to hear everyone's idea on the basic design here:

# Interfaces 

- [ ] Encryption interfaces + Plaintext manager — EncryptionManager, EncryptedInputFile/EncryptedOutputFile, EncryptionKeyMetadata, NativeEncryption*, and the no-op PlaintextEncryptionManager. 
- [ ] KMS client interface — KeyManagementClient (wrapKey/unwrapKey/supportsKeyGeneration/generateKey), plus a test/in-memory impl and encryption.kms-type/encryption.kms-impl catalog-property wiring.
- [ ] StandardEncryptionManager — envelope KEK/DEK logic, local DEK generation, key cache, rotation/timestamp tracking, encryption.key-id + encryption.data-key-length table properties. Consumes the existing encryption-keys metadata array.

# Integration
- [ ] EncryptingFileIO wrapper — slot encryption into the io.IO layer: decrypt on Open of Data/Manifest/ManifestList/Statistics files based on key_metadata, encrypt on write. 
- [ ] Parquet native encryption for data/delete files — wire FileEncryptionProperties/FileDecryptionProperties from NativeEncryptionOutputFile into table/internal/parquet_files.go (write) and the scan read path.
- [ ] Manifest + manifest-list encryption — whole-file AES-GCM using the existing key_metadata fields, including the manifest-list double-encryption (KEK + timestamp AAD) on the write/read planners.
- [ ] Puffin blob / statistics-file encryption — encrypt DV and stats blobs; populate StatisticsFile.KeyMetadata.

# Testing
- [ ] Format-version compatibility + end-to-end + cross-client tests — reject encryption props on v1, gate behavior, and round-trip a Java-produced encrypted table end-to-end (the equivalent of the DV cross-client compliance PR #1131).

laskoviymishka:
That's a cool feature! We totally should do this!
I would like to dive a bit deeper here with a design (how this would be setted from user perspective).

hectar-glitches:
[...]

PR #1493, opened 2026-07-17
feat(encryption): add KMS catalog-property registry

hectar-glitches:
## What

Adds a named-factory registry for `KeyManagementClient` implementations, selected via the `encryption.kms-type` / `encryption.kms-impl` catalog properties. This completes the second half of the "KMS client interface"
checklist item from #1289 (the interface + in-memory impl landed in #1447).

Mirrors the existing `io.Register`/`io.SchemeFactory` scheme registry pattern (`io/registry.go`) rather than introducing a new convention.

## Details

- `RegisterKMS(name, factory)` / `UnregisterKMS(name)` / `GetRegisteredKMSNames()`: a mutex-guarded named registry for `KMSFactory` implementations.
- `LoadKeyManagementClient(props)`: resolves `encryption.kms-type` first,
  falling back to `encryption.kms-impl`: returning an error wrapping the new
  `ErrKMSTypeNotFound` if neither is set or the name isn't registered.
- Registers `"memory"` to `InMemoryKeyManagementClient` as the built-in option in `init()`.

## Not in scope

- `StandardEncryptionManager` (envelope KEK/DEK logic).

## Testing

- New unit tests in `encryption/kms_registry_test.go` covering: resolution via each property, `kms-type` taking precedence when both are set, missing/unregistered names, duplicate/nil-factory registration panics, and factory-error propagation.

### DIFF
--- a/encryption/kms_registry.go
+++ b/encryption/kms_registry.go
@@ -25,11 +25,11 @@ import (
 	"sync"
 )
 
-// KMSTypeKey is the catalog/table property that selects a
-// [KeyManagementClient] implementation registered via [RegisterKMS] (e.g.
-// "memory"). It is a registered short name, not a fully-qualified class
-// name.
-const KMSTypeKey = "encryption.kms-type"
+// KMSTypeKey is the catalog property that selects a [KeyManagementClient]
+// implementation registered via [RegisterKMS] (e.g. "memory"). It is a
+// registered short name, not a fully-qualified class name.
+
+const KMSTypeKey = "kms-type"
 
 // ErrKMSTypeNotFound is returned by [LoadKeyManagementClient] when the
 // requested kmsType has not been registered via [RegisterKMS].
--- a/encryption/kms_registry.go
+++ b/encryption/kms_registry.go
@@ -25,19 +25,11 @@ import (
 	"sync"
 )
 
-// Catalog/table property keys used to select and configure a KeyManagementClient.
-const (
-	// KMSTypeKey selects a built-in KeyManagementClient implementation
-	// registered via [RegisterKMS] (e.g. "memory").
-	KMSTypeKey = "encryption.kms-type"
-
-	// KMSImplKey selects a custom KeyManagementClient implementation by a
-	// name registered via [RegisterKMS]. It is an alternative to
-	// [KMSTypeKey] for vendors that ship their own KMS integration under a
-	// distinct name; the two properties share the same registry and
-	// resolution mechanism.
-	KMSImplKey = "encryption.kms-impl"
-)
+// KMSTypeKey is the catalog/table property that selects a
+// [KeyManagementClient] implementation registered via [RegisterKMS] (e.g.
+// "memory"). It is a registered short name, not a fully-qualified class
+// name.
+const KMSTypeKey = "encryption.kms-type"
 
 // ErrKMSTypeNotFound is returned by [LoadKeyManagementClient] when the
 // requested kmsType has not been registered via [RegisterKMS].
@@ -55,8 +47,8 @@ var (
 )
 
 // RegisterKMS adds a new named [KMSFactory] to the registry so it can later
-// be selected via the [KMSTypeKey] or [KMSImplKey] catalog property. It
-// panics if factory is nil or if name is already registered.
+// be selected via the [KMSTypeKey] catalog property. It panics if factory is
+// nil or if name is already registered.
 func RegisterKMS(name string, factory KMSFactory) {
 	if factory == nil {
 		panic("encryption: RegisterKMS factory is nil")
@@ -88,18 +80,13 @@ func GetRegisteredKMSNames() []string {
 }
 
 // LoadKeyManagementClient resolves and constructs a [KeyManagementClient]
-// from props. It looks up the KMS name under [KMSTypeKey] first, falling
-// back to [KMSImplKey] if present; it returns an error wrapping
-// [ErrKMSTypeNotFound] if neither property is set or the named KMS has not
-// been registered via [RegisterKMS].
+// from props. It looks up the KMS name under [KMSTypeKey]; it returns an
+// error wrapping [ErrKMSTypeNotFound] if the property is unset or the named
+// KMS has not been registered via [RegisterKMS].
 func LoadKeyManagementClient(props map[string]string) (KeyManagementClient, error) {
 	name := props[KMSTypeKey]
 	if name == "" {
-		name = props[KMSImplKey]
-	}
-
-	if name == "" {
-		return nil, fmt.Errorf("%w: neither %q nor %q is set", ErrKMSTypeNotFound, KMSTypeKey, KMSImplKey)
+		return nil, fmt.Errorf("%w: %q is not set", ErrKMSTypeNotFound, KMSTypeKey)
 	}
 
 	kmsRegMutex.RLock()
--- a/encryption/kms_registry.go
+++ b/encryption/kms_registry.go
@@ -1,0 +1,120 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


encryption/kms_registry.go
```diff
@@ -0,0 +1,120 @@
...
+
+	kmsRegMutex.RLock()
+	factory, ok := kmsRegistry[name]
+	kmsRegMutex.RUnlock()
+
+	if !ok {
+		return nil, fmt.Errorf("%w: %q", ErrKMSTypeNotFound, name)
+	}
+
+	return factory(props)
```
tanmayrauth, line 104: Nothing verifies props actually reaches the factory here — every test registers a factory that ignores its argument. The whole point of KMSFactory(props) is that real KMS factories read key  material/region/endpoints out of props, so a future cleanup that filtered the map or passed nil before this call would compile, pass the whole suite, and silently hand an AWS/GCP factory an empty map — a factory reading encryption.kms.region would build a wrong-region client or fail at construction, uncaught. Please add a test that registers a factory recording its received props and assert it equals the input (include a sentinel key like encryption.kms.region).

encryption/kms_registry.go
```diff
@@ -0,0 +1,120 @@
...
+	// KMSTypeKey selects a built-in KeyManagementClient implementation
+	// registered via [RegisterKMS] (e.g. "memory").
+	KMSTypeKey = "encryption.kms-type"
+
+	// KMSImplKey selects a custom KeyManagementClient implementation by a
+	// name registered via [RegisterKMS]. It is an alternative to
+	// [KMSTypeKey] for vendors that ship their own KMS integration under a
+	// distinct name; the two properties share the same registry and
+	// resolution mechanism.
+	KMSImplKey = "encryption.kms-impl"
```
tanmayrauth: kms-type and kms-impl resolve the same registry by the same short-name lookup, so they're functionally identical — and that's novel here: nothing else in the repo uses a -impl property, and the precedent (catalog.Load reads a single "type"; the io scheme registry) is one short key. In Iceberg generally -impl means a fully-qualified implementation, distinct from the short -type alias, so someone migrating a config that sets encryption.kms-impl=com.acme.AcmeKmsClient gets ErrKMSTypeNotFound because Go looks it up as a registered short name. Since nothing consumes this yet, either drop kms-impl and keep only kms-type (matching the repo's single-key precedent), or keep both and make the doc explicit that kms-impl is a registered short name, not an FQCN — the "a custom implementation" wording at 34-35 currently reads like the Java FQCN meaning.

encryption/kms_registry.go
```diff
@@ -0,0 +1,120 @@
...
+	return slices.Collect(maps.Keys(kmsRegistry))
+}
+
+// LoadKeyManagementClient resolves and constructs a [KeyManagementClient]
+// from props. It looks up the KMS name under [KMSTypeKey] first, falling
+// back to [KMSImplKey] if present; it returns an error wrapping
+// [ErrKMSTypeNotFound] if neither property is set or the named KMS has not
+// been registered via [RegisterKMS].
+func LoadKeyManagementClient(props map[string]string) (KeyManagementClient, error) {
+	name := props[KMSTypeKey]
```
tanmayrauth, line 91: The doc says it falls back to KMSImplKey "if present," but the code only falls back when kms-type is empty, not when it's set to an unregistered name — so kms-type=typo + kms-impl=memory errors instead of loading memory. The  behavior's defensible (an explicit type shouldn't silently swallow a typo), but the doc reads otherwise. Tighten it to "falls back only when KMSTypeKey is unset," and consider adding that set-but-unregistered-type + valid-impl case  to the precedence test at kms_registry_test.go:44.

tanmayrauth (commented):
Registry mechanics look right and mirror io.Register faithfully. One test-gap plus a couple of design calls worth settling now while nothing consumes the API yet. Details inline.

encryption/kms_registry.go
hectar-glitches: added some tests to register a factory that records the props map it receives and asserts it's identical to what was passed to LoadKeyManagementClient

encryption/kms_registry.go
hectar-glitches: dropped kms-imply entirely leaving just KMSTypeKey

encryption/kms_registry.go
hectar-glitches: will tighten

encryption/kms_registry.go
```diff
@@ -0,0 +1,107 @@
...
+	"maps"
+	"slices"
+	"sync"
+)
+
+// KMSTypeKey is the catalog/table property that selects a
+// [KeyManagementClient] implementation registered via [RegisterKMS] (e.g.
+// "memory"). It is a registered short name, not a fully-qualified class
+// name.
+const KMSTypeKey = "encryption.kms-type"
```
zeroshade: is this already part of the iceberg spec?

zeroshade (approved):

tanmayrauth (approved):

encryption/kms_registry.go
hectar-glitches: no, this isn't part of the Iceberg spec. I checked the spec and configuration docs: the only standardized encryption properties are "encryption.key-id" and "encryption.data-key-length". "encryption.kms-type" is an iceberg-go-specific mechanism for selecting a registered "KeyManagementClient" implementation at runtime so there's no cross-implementation compatibility concern here since it only affects how this Go library wires up its KMS client, not anything written to table metadata. 

Happy to rename it to something less spec-adjacent (e.g. a Go-specific prefix) if that's clearer

zeroshade: let's update the docs to explicitly point out this is *not* a cross-implementation available property. Is there a desire to get this into the spec? Is there an existing proposal for KMS registration for the spec?

hectar-glitches:
@zeroshade is this okay to merge so i can continue?

encryption/kms_registry.go
```diff
@@ -0,0 +1,107 @@
...
+	"maps"
+	"slices"
+	"sync"
+)
+
+// KMSTypeKey is the catalog property that selects a [KeyManagementClient]
+// implementation registered via [RegisterKMS] (e.g. "memory"). It is a
+// registered short name, not a fully-qualified class name.
+
+const KMSTypeKey = "kms-type"
```
zeroshade, line 36: Doc request from last round: please note explicitly (here on `KMSTypeKey`) that `kms-type` is an iceberg-go-specific mechanism and NOT a standardized/cross-implementation Iceberg spec property, so users don't assume portability.

zeroshade (commented):
Re-reviewed at head — dropping `kms-impl` (resolving the typo-fallback ambiguity) and the props-reach-factory test both look good. The one item still open from my last pass is the docs request below (non-blocking). Also still worth answering: is there interest in standardizing KMS registration in the Iceberg spec, or an existing proposal? CI green.

encryption/kms_registry.go
hectar-glitches: done. Added an explicit note on "KMSTypeKey" making clear it's an iceberg-go-specific mechanism

hectar-glitches:
On standardizing KMS registration in the spec, I don't see an existing proposal for a cross-implementation KMS client registration mechanism. The spec's encryption properties ("encryption.key-id", "encryption.data-key-length") describe what key material to use, not how to obtain it from a KMS, so client wiring has been left to each implementation to solve on its own.

zeroshade (approved):
Docs now make the portability caveat explicit on `KMSTypeKey` ("iceberg-go-specific… not a standardized Iceberg spec property… will not be recognized by other implementations") — that was my only actionable ask, thanks. And thanks for chasing down the spec question; agreed that KMS *client wiring* is out of scope for the spec today (it covers key material via `encryption.key-id` etc., not acquisition).

Registry itself is clean: `sync.RWMutex`, panic-on-duplicate/nil matching the `database/sql` driver pattern, full props threaded to the factory, and good coverage (duplicate/nil panics, unregistered, factory error, props pass-through). LGTM.

PR #1501, opened 2026-07-21
  feat(metrics): instrument scan planning and emit ScanReport (#1236)

tanmayrauth:
PR 4: Instruments local scan planning to gather scan metrics and emit a ScanReport through the scan's configured reporter.

  Builds on the reporter plumbing already merged in #1430 (Scan.Reporter(), catalog/table/scan wiring). No new public API — metrics-reporter-impl / WithMetricsReporter remain the entry points.

  Changes

  - PlanFiles now times local planning and, on success, emits a ScanReport via scan.Reporter().Report(...). Split into PlanFiles (mode dispatch + emission) and planFilesLocal (the local-planning body).
  - Manifest counts (total / scanned / skipped, split by data vs delete content) gathered in the single-goroutine manifest-filter loop.
  - Result-scoped metrics (result-data-files, result-delete-files, positional/equality/DV counts, total-file-size-in-bytes, total-delete-file-size-in-bytes) derived from the planned tasks — delete files counted once by path, and
  DV-suppressed positional deletes excluded, so they stay consistent with result-data-files.
  - Schema/filter/projected-fields resolved best-effort; report building never fails a scan.

  Behavior / scope

  - Opt-in, zero-overhead default: with the no-op reporter (the default) no report is built — the instrumented path costs only a few integer increments.
  - Remote planning reports its own metrics server-side and does not emit a local report.
  - Race-safe: all counters written from a single goroutine (manifest counts before the read barrier, result/delete counts after); verified under -race.
  - Omit, don't fabricate: scanned/skipped reflect partition-spec pruning; per-entry skipped-data-files / skipped-delete-files and indexed-delete-files are left unset (deferred) rather than reported as zero.

  Testing

  go test ./table/... ./metrics/... and -race on the scan/delete/DV planning suites. New tests: report assembly, projected-field resolution, result-scoped delete dedup (TestApplyResultDeleteMetrics), emit-on-local, no-emit-on-remote,
  and no-op-reporter safety.
  
  Related: #1236

### DIFF
--- a/table/scan_metrics.go
+++ b/table/scan_metrics.go
@@ -1,0 +1,172 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+import (
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/metrics"
+)
+
+// scanMetricsAccumulator gathers scan-planning counts. Every field is written
+// from a single goroutine — the manifest counts in
+// fetchPartitionSpecFilteredManifestsWithSchema and the result/delete counts in
+// planFilesLocal after collectManifestEntriesWithSchema's concurrency barrier —
+// so plain integers are race-free; no field is touched from the concurrent
+// manifest workers.
+//
+// The scanned/skipped manifest counts measure partition-spec pruning (the
+// filter applied while listing manifests). The per-entry skipped-data-files /
+// skipped-delete-files counts (which would be incremented inside the concurrent
+// openManifest loop, and would use atomics) and indexed-delete-files are left
+// for a follow-up and omitted rather than reported as zero.
+type scanMetricsAccumulator struct {
+	totalDataManifests     int64
+	totalDeleteManifests   int64
+	scannedDataManifests   int64
+	scannedDeleteManifests int64
+	skippedDataManifests   int64
+	skippedDeleteManifests int64
+
+	resultDataFiles       int64
+	resultDeleteFiles     int64
+	totalFileSize         int64
+	totalDeleteFileSize   int64
+	positionalDeleteFiles int64
+	equalityDeleteFiles   int64
+	dvs                   int64
+}
+
+func counterCount(v int64) *metrics.CounterResult {
+	return metrics.NewCounterResult(metrics.UnitCount, v)
+}
+
+func counterBytes(v int64) *metrics.CounterResult {
+	return metrics.NewCounterResult(metrics.UnitBytes, v)
+}
+
+// applyResultDeleteMetrics derives the result-scoped delete-file metrics from
+// the planned tasks. Each delete file is counted once by path across all the
+// data files it applies to, split by kind (positional, equality, deletion
+// vector); positional deletes suppressed by a deletion vector never appear on a
+// task, so they are excluded. This keeps the counts consistent with
+// result-data-files and total-file-size-in-bytes.
+func (acc *scanMetricsAccumulator) applyResultDeleteMetrics(tasks []FileScanTask) {
+	posDeletes := make(map[string]int64)
+	eqDeletes := make(map[string]int64)
+	dvs := make(map[string]int64)
+	for _, t := range tasks {
+		for _, df := range t.DeleteFiles {
+			posDeletes[df.FilePath()] = df.FileSizeBytes()
+		}
+		for _, df := range t.EqualityDeleteFiles {
+			eqDeletes[df.FilePath()] = df.FileSizeBytes()
+		}
+		for _, df := range t.DeletionVectorFiles {
+			dvs[df.FilePath()] = df.FileSizeBytes()
+		}
+	}
+
+	acc.positionalDeleteFiles = int64(len(posDeletes))
+	acc.equalityDeleteFiles = int64(len(eqDeletes))
+	acc.dvs = int64(len(dvs))
+	acc.resultDeleteFiles = acc.positionalDeleteFiles + acc.equalityDeleteFiles + acc.dvs
+	for _, deletes := range []map[string]int64{posDeletes, eqDeletes, dvs} {
+		for _, size := range deletes {
+			acc.totalDeleteFileSize += size
+		}
+	}
+}
+
+// buildScanReport assembles the ScanReport for a completed planning operation.
+// Only the metrics that are actually measured are populated; unmeasured ones
+// (e.g. skipped-data-files, indexed-delete-files) are left unset and omitted.
+func (scan *Scan) buildScanReport(acc *scanMetricsAccumulator, planning time.Duration) metrics.ScanReport {
+	// Resolve the schema the scan actually planned against (the snapshot schema
+	// for snapshot/as-of scans, the current schema otherwise). Best-effort: a
+	// report must never fail the scan, so fall back to the current schema.
+	schema, err := scan.effectiveSchema()
+	if err != nil || schema == nil {
+		schema = scan.metadata.CurrentSchema()
+	}
+
+	ids, names := scan.projectedFields(schema)
+
+	var snapshotID int64
+	if snap := scan.Snapshot(); snap != nil {
+		snapshotID = snap.SnapshotID
+	}
+
+	return metrics.ScanReport{
+		TableName:           strings.Join(scan.identifier, "."),
+		SnapshotID:          snapshotID,
+		SchemaID:            schema.ID,
+		ProjectedFieldIDs:   ids,
+		ProjectedFieldNames: names,
+		Metrics: metrics.ScanMetricsResult{
+			TotalPlanningDuration:      metrics.NewNanosTimerResult(1, planning.Nanoseconds()),
+			ResultDataFiles:            counterCount(acc.resultDataFiles),
+			ResultDeleteFiles:          counterCount(acc.resultDeleteFiles),
+			TotalDataManifests:         counterCount(acc.totalDataManifests),
+			TotalDeleteManifests:       counterCount(acc.totalDeleteManifests),
+			ScannedDataManifests:       counterCount(acc.scannedDataManifests),
+			ScannedDeleteManifests:     counterCount(acc.scannedDeleteManifests),
+			SkippedDataManifests:       counterCount(acc.skippedDataManifests),
+			SkippedDeleteManifests:     counterCount(acc.skippedDeleteManifests),
+			TotalFileSizeInBytes:       counterBytes(acc.totalFileSize),
+			TotalDeleteFileSizeInBytes: counterBytes(acc.totalDeleteFileSize),
+			EqualityDeleteFiles:        counterCount(acc.equalityDeleteFiles),
+			PositionalDeleteFiles:      counterCount(acc.positionalDeleteFiles),
+			DVs:                        counterCount(acc.dvs),
+		},
+	}
+}
+
+// projectedFields resolves the scan's selected fields to their ids and names
+// against schema. A "*" (or empty) selection projects all top-level fields.
+// Unresolvable names are skipped.
+func (scan *Scan) projectedFields(schema *iceberg.Schema) ([]int, []string) {
+	if len(scan.selectedFields) == 0 || slices.Contains(scan.selectedFields, "*") {
+		fields := schema.Fields()
+		ids := make([]int, len(fields))
+		names := make([]string, len(fields))
+		for i, f := range fields {
+			ids[i], names[i] = f.ID, f.Name
+		}
+
+		return ids, names
+	}
+
+	ids := make([]int, 0, len(scan.selectedFields))
+	names := make([]string, 0, len(scan.selectedFields))
+	for _, name := range scan.selectedFields {
+		nf, ok := schema.FindFieldByName(name)
+		if !ok && !scan.caseSensitive {
+			nf, ok = schema.FindFieldByNameCaseInsensitive(name)
+		}
+		if ok {
+			ids = append(ids, nf.ID)
+			names = append(names, nf.Name)
+		}
+	}
+
+	return ids, names
+}
--- a/table/scan_metrics.go
+++ b/table/scan_metrics.go
@@ -65,59 +65,35 @@ func counterBytes(v int64) *metrics.CounterResult {
 }
 
 // applyResultDeleteMetrics derives the result-scoped delete-file metrics from
-// the planned tasks. Each delete file is counted once by path across all the
-// data files it applies to, split by kind (positional, equality, deletion
-// vector); positional deletes suppressed by a deletion vector never appear on a
-// task, so they are excluded. This keeps the counts consistent with
-// result-data-files and total-file-size-in-bytes.
+// the planned tasks. Following Java's ScanMetricsUtil.fileTask, every delete
+// file is counted once per data-file task it applies to — a delete covering N
+// data files is counted N times — with no cross-task dedup, so these
+// Java-parity field names stay directly comparable with what Java consumers
+// report. Sizes accumulate the same way.
 //
-// Parity note: Java's ScanMetricsUtil.fileTask counts result-delete-files per
-// data-file task with no cross-task dedup, so a positional delete covering 5
-// data files counts as 5 there and once here. Counting once by path is the
-// deliberate choice for this result-scoped metric; keep it if touching this.
-//
-// Deletion vectors are keyed by the data file they reference, not by their
-// Puffin file path: a single Puffin file holds one DV blob per data file (see
-// DVWriter.Flush), so all DVs from one flush share a FilePath() and would
-// collapse into one entry if keyed by it. Their size is the per-blob
-// content_size_in_bytes, not the whole Puffin file size, matching Java's
-// ScanTaskUtil.contentSizeInBytes.
+// A deletion vector's size is its per-blob content_size_in_bytes (Java's
+// ScanTaskUtil.contentSizeInBytes), not the size of the Puffin file that holds
+// it: DVWriter.Flush packs one DV blob per data file into a single Puffin file,
+// so the Puffin length would over-count.
 func (acc *scanMetricsAccumulator) applyResultDeleteMetrics(tasks []FileScanTask) {
-	posDeletes := make(map[string]int64)
-	eqDeletes := make(map[string]int64)
-	dvByRef := make(map[string]int64)
 	for _, t := range tasks {
 		for _, df := range t.DeleteFiles {
-			posDeletes[df.FilePath()] = df.FileSizeBytes()
+			acc.positionalDeleteFiles++
+			acc.totalDeleteFileSize += df.FileSizeBytes()
 		}
 		for _, df := range t.EqualityDeleteFiles {
-			eqDeletes[df.FilePath()] = df.FileSizeBytes()
+			acc.equalityDeleteFiles++
+			acc.totalDeleteFileSize += df.FileSizeBytes()
 		}
 		for _, df := range t.DeletionVectorFiles {
-			// Key by the referenced data file so multiple DVs sharing one
-			// Puffin path stay distinct; fall back to the Puffin path if the
-			// reference is somehow unset. Size is the per-blob content size.
-			key := df.FilePath()
-			if ref := df.ReferencedDataFile(); ref != nil {
-				key = *ref
-			}
-			var size int64
+			acc.dvs++
 			if csb := df.ContentSizeInBytes(); csb != nil {
-				size = *csb
+				acc.totalDeleteFileSize += *csb
 			}
-			dvByRef[key] = size
 		}
 	}
 
-	acc.positionalDeleteFiles = int64(len(posDeletes))
-	acc.equalityDeleteFiles = int64(len(eqDeletes))
-	acc.dvs = int64(len(dvByRef))
 	acc.resultDeleteFiles = acc.positionalDeleteFiles + acc.equalityDeleteFiles + acc.dvs
-	for _, deletes := range []map[string]int64{posDeletes, eqDeletes, dvByRef} {
-		for _, size := range deletes {
-			acc.totalDeleteFileSize += size
-		}
-	}
 }
 
 // buildScanReport assembles the ScanReport for a completed planning operation.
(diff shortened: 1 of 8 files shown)


tanmayrauth:
@zeroshade @laskoviymishka Can you please take a look when you get chance?

table/scan_metrics.go
```diff
@@ -0,0 +1,172 @@
...
+
+	ids, names := scan.projectedFields(schema)
+
+	var snapshotID int64
+	if snap := scan.Snapshot(); snap != nil {
+		snapshotID = snap.SnapshotID
+	}
+
+	return metrics.ScanReport{
+		TableName:           strings.Join(scan.identifier, "."),
```
laskoviymishka, line 130: `buildScanReport` never sets `Filter`, so the `MarshalJSON` fallback emits `"filter": true` unconditionally — a scan with a real `rowFilter` still reports `filter: true`, which is misleading once these land in a REST catalog's metrics table.

The infra is already there: every concrete `BooleanExpression` has `MarshalJSON` (expr_json.go). I'd set `Filter` from `scan.rowFilter` here, falling back to always-true on marshal error. Java sanitizes literals via `ExpressionUtil` first and we don't have an equivalent yet, so I'm fine leaving sanitization as a follow-up — but the field needs to reflect the actual filter.

table/scan_metrics.go
```diff
@@ -0,0 +1,172 @@
...
+	dvs := make(map[string]int64)
+	for _, t := range tasks {
+		for _, df := range t.DeleteFiles {
+			posDeletes[df.FilePath()] = df.FileSizeBytes()
+		}
+		for _, df := range t.EqualityDeleteFiles {
+			eqDeletes[df.FilePath()] = df.FileSizeBytes()
+		}
+		for _, df := range t.DeletionVectorFiles {
+			dvs[df.FilePath()] = df.FileSizeBytes()
```
laskoviymishka: I think there's a real undercount here for deletion vectors. `DVWriter.Flush` writes one Puffin file per flush with a DV blob per data file, so every DV `DataFile` from a flush shares the same `FilePath()` (the puffin location) and differs only by `ReferencedDataFile()`. Keying `dvs` by `FilePath()` collapses N DVs into 1, and `resultDeleteFiles` comes out off by N-1. Java keys `dvByPath` by `referencedDataFile()`, not location — I'd key the DV map by `df.ReferencedDataFile()`.

Same map, second bug: the size is `df.FileSizeBytes()`, which is the whole puffin file, not the per-blob DV length. Java's `ScanTaskUtil.contentSizeInBytes` uses `contentSizeInBytes()` for DVs. `DataFile` already carries `ContentSizeInBytes()` (populated in dv_writer.go), so I'd use that (nil-guarded) for the DV size.

While we're here — worth renaming the local `dvs` map, since it shadows the `acc.dvs` field and the other two are `posDeletes`/`eqDeletes` for exactly that reason. And `TestApplyResultDeleteMetrics` should grow a case with two DVs sharing one puffin path so this stays covered once fixed.

table/scanner.go
```diff
@@ -812,6 +835,30 @@ func (scan *Scan) PlanFiles(ctx context.Context) ([]FileScanTask, error) {
...
+		return nil, err
+	}
+
+	// Building the report is skipped for the no-op reporter — the opt-in default
+	// set by Table.Scan, and the nil-reporter case for scans constructed directly
+	// in tests (see Reporter, which maps nil to NopReporter). A NopReporter
+	// discards the report, so assembling one would be pure overhead.
+	rep := scan.Reporter()
+	if _, isNop := rep.(metrics.NopReporter); !isNop {
+		rep.Report(ctx, scan.buildScanReport(&acc, time.Since(start)))
```
laskoviymishka: `planFilesLocal` returns `(nil, nil)` when there's no snapshot or every manifest is pruned, but we still emit a report here with `SnapshotID: 0` — a zero-value int64, not a real snapshot id. Java doesn't emit a scan report in the no-snapshot case at all.

I'd guard on `scan.Snapshot() == nil` before emitting so we don't push bogus zero-snapshot records into the reporter. wdyt?

table/scanner.go
```diff
@@ -653,10 +654,13 @@ func (scan *Scan) fetchPartitionSpecFilteredManifests(ctx context.Context) ([]ic
 		return nil, err
 	}
```
laskoviymishka, line 656: This wrapper allocates a fresh `&scanMetricsAccumulator{}` and throws it away — the manifest counts it records go nowhere. Fine today since this path has no reporter behind it, but it's a quiet trap: a future caller wiring this path to a report would silently get zeroed manifest counters with no hint why.

I'd either drop the accumulator param from this path, or leave a comment making it explicit that the counts are intentionally discarded here.

table/scan_metrics.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// buildScanReport assembles the ScanReport for a completed planning operation.
+// Only the metrics that are actually measured are populated; unmeasured ones
+// (e.g. skipped-data-files, indexed-delete-files) are left unset and omitted.
+func (scan *Scan) buildScanReport(acc *scanMetricsAccumulator, planning time.Duration) metrics.ScanReport {
+	// Resolve the schema the scan actually planned against (the snapshot schema
+	// for snapshot/as-of scans, the current schema otherwise). Best-effort: a
+	// report must never fail the scan, so fall back to the current schema.
+	schema, err := scan.effectiveSchema()
```
laskoviymishka: `buildScanReport` re-resolves the schema with `effectiveSchema()` and then silently falls back to `CurrentSchema()` on error — but `planFilesLocal` already resolved the schema it actually planned against. If the two disagree (or the fallback fires), the report describes a different schema than the one used for planning.

I'd thread the planning schema through into `buildScanReport` instead of re-resolving. It also drops a redundant `Schemas()` scan per `PlanFiles`.

table/scan_metrics_test.go
```diff
@@ -0,0 +1,187 @@
...
+			File:                &mockDataFile{path: "s3://b/data-2.parquet", filesize: 2000},
+			DeleteFiles:         []iceberg.DataFile{posA},
+			EqualityDeleteFiles: []iceberg.DataFile{eq},
+			DeletionVectorFiles: []iceberg.DataFile{dv},
+		},
+	}
+
+	var acc scanMetricsAccumulator
+	acc.applyResultDeleteMetrics(tasks)
+
```
laskoviymishka, line 171: This test scans empty metadata, so every manifest counter is zero — which means the scanned/skipped increment logic in `fetchPartitionSpecFilteredManifestsWithSchema` has no real coverage. A swapped scanned/skipped or data/delete branch would sail through green.

Could we add a case with a real snapshot that has manifests and assert `scanned + skipped == total` (and that delete manifests land in the delete counters)? That's the part most likely to regress silently.

table/scanner.go
```diff
@@ -812,6 +835,30 @@ func (scan *Scan) PlanFiles(ctx context.Context) ([]FileScanTask, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Building the report is skipped for the no-op reporter — the opt-in default
+	// set by Table.Scan, and the nil-reporter case for scans constructed directly
+	// in tests (see Reporter, which maps nil to NopReporter). A NopReporter
+	// discards the report, so assembling one would be pure overhead.
+	rep := scan.Reporter()
+	if _, isNop := rep.(metrics.NopReporter); !isNop {
```
laskoviymishka: This guard only catches a bare `metrics.NopReporter` — `Combine(NopReporter{})` returns a `*compositeReporter` and slips right through, so the comment above about avoiding "pure overhead" isn't quite true for a wrapped nop. Minor, but I'd either add a small `isNop`-style interface the composite can answer, or soften the comment so it doesn't over-promise.

table/scanner.go
```diff
@@ -812,6 +835,30 @@ func (scan *Scan) PlanFiles(ctx context.Context) ([]FileScanTask, error) {
 		return nil, fmt.Errorf("%w: unknown scan planning mode %q", iceberg.ErrInvalidArgument, scan.planningMode)
 	}
 
+	start := time.Now()
```
laskoviymishka, line 842: `start` is captured here, but `time.Since(start)` is computed inside the `Report` call below, so `TotalPlanningDuration` also includes `buildScanReport`'s schema re-resolution and projection. I'd snap the elapsed time right after `planFilesLocal` returns and pass that in, so the duration reflects just planning.

table/scan_metrics.go
```diff
@@ -0,0 +1,172 @@
...
+	return metrics.NewCounterResult(metrics.UnitBytes, v)
+}
+
+// applyResultDeleteMetrics derives the result-scoped delete-file metrics from
+// the planned tasks. Each delete file is counted once by path across all the
+// data files it applies to, split by kind (positional, equality, deletion
+// vector); positional deletes suppressed by a deletion vector never appear on a
+// task, so they are excluded. This keeps the counts consistent with
+// result-data-files and total-file-size-in-bytes.
+func (acc *scanMetricsAccumulator) applyResultDeleteMetrics(tasks []FileScanTask) {
```
laskoviymishka, line 78: One parity note: Java's `ScanMetricsUtil.fileTask` counts `resultDeleteFiles` per data-file task with no cross-task dedup, so a positional delete covering 5 data files counts 5 there and 1 here. The count-once-by-path choice is defensible as a result-scoped metric, but since it diverges from Java I'd call it out in this comment so the next person doesn't "fix" it into a mismatch.

table/scanner.go
```diff
@@ -812,6 +835,30 @@ func (scan *Scan) PlanFiles(ctx context.Context) ([]FileScanTask, error) {
...
+	}
+
+	return results, nil
+}
+
+// planFilesLocal performs local scan planning: it reads and filters the
+// snapshot's manifests, builds the matching FileScanTasks, and records planning
+// metrics into acc for the caller to report.
+func (scan *Scan) planFilesLocal(ctx context.Context, acc *scanMetricsAccumulator) ([]FileScanTask, error) {
 	scan.planIO = nil
```
laskoviymishka, line 889: Two small things on `planFilesLocal`: it mutates `scan.planIO = nil` as a side effect that the doc comment doesn't mention, and it returns `(nil, nil)` rather than `([]FileScanTask{}, nil)` on empty manifests. Neither is wrong, but I'd note the `planIO` reset in the comment and pick nil-vs-empty deliberately so callers know which to expect.

laskoviymishka (changes requested):
The `PlanFiles` / `planFilesLocal` split looks good, and `scanMetricsAccumulator` keeps the counting logic easy to follow.

I don’t think this is ready to merge yet because two metrics are currently misleading.

The first is `filter`. `buildScanReport` never sets it, so every report serializes as `filter: true`, even when the scan has a real `rowFilter`. The expression `MarshalJSON` support is already there, so this should be straightforward to wire through. As it stands, a REST catalog would store incorrect scan records.

The second is deletion-vector counting in `applyResultDeleteMetrics`. Iceberg-go writes one Puffin file per flush, with a separate DV blob for each data file. Using `FilePath()` as the map key collapses all DVs in the same Puffin file into one entry, so both `dvs` and `resultDeleteFiles` are undercounted. It also uses the full Puffin file size rather than the blob size.

The map should be keyed by `ReferencedDataFile()`, and the size should come from `ContentSizeInBytes()`, matching Java. A test with two DVs in the same Puffin file would catch this.

A few other things I’d fix here:

* Set `ScanReport.Filter` from `scan.rowFilter`. Sanitization can follow separately.
* Skip report emission when there is no snapshot instead of sending `SnapshotID: 0`.
* Pass the planning schema into `buildScanReport` rather than resolving it again.
* Add a real-snapshot test that verifies the manifest counters are non-zero.

The remaining comments are minor. Once the filter and DV accounting are fixed, I’ll review it again.

tanmayrauth:
Thanks for the thorough review, @laskoviymishka, all issues have been addressed. buildScanReport now sets Filter from scan.rowFilter (serialized via the expression MarshalJSON, falling back to always-true on marshal error; literal  sanitization left as a follow-up). The DV accounting now keys by ReferencedDataFile() and sizes by  ContentSizeInBytes() so multiple DVs sharing one Puffin file no longer collapse, and the local map is renamed to dvByRef to stop shadowing acc.dvs; added a test with two DVs in one Puffin path. 

Report emission is now skipped when scan.Snapshot() == nil so we don't push zero-snapshot records. The planning schema is resolved once in PlanFiles and threaded into both planFilesLocal and buildScanReport (dropping the redundant Schemas() scan and the silent fallback), and TotalPlanningDuration is now snapped right after planning returns, before report assembly. 

Added a real-snapshot test that exercises partition pruning and asserts scanned + skipped == total with delete manifests landing in the delete counters. For the minor items: the Combine(NopReporter{}) case is now handled via a new metrics.IsNop (an unexported isNop() the composite answers); the throwaway-accumulator wrapper, the Java per-task resultDeleteFiles divergence,  and planFilesLocal's planIO reset + nil-vs-empty return are all documented. go build, go vet, gofmt, and the table/metrics suites (incl. -race) pass.

table/scan_metrics.go
```diff
@@ -0,0 +1,202 @@
...
+// DVWriter.Flush), so all DVs from one flush share a FilePath() and would
+// collapse into one entry if keyed by it. Their size is the per-blob
+// content_size_in_bytes, not the whole Puffin file size, matching Java's
+// ScanTaskUtil.contentSizeInBytes.
+func (acc *scanMetricsAccumulator) applyResultDeleteMetrics(tasks []FileScanTask) {
+	posDeletes := make(map[string]int64)
+	eqDeletes := make(map[string]int64)
+	dvByRef := make(map[string]int64)
+	for _, t := range tasks {
+		for _, df := range t.DeleteFiles {
```
zeroshade, line 80: `result-delete-files`/`total-delete-file-size-in-bytes` are deduped by delete path. Java's `ScanMetricsUtil.fileTask` increments per planned data-file task (a delete applying to N files counts N times). Since these are Java-parity field names, consumers compare them directly — either match per-task accumulation or use different names.

table/scan_metrics.go
```diff
@@ -0,0 +1,202 @@
...
+	if snap := scan.Snapshot(); snap != nil {
+		snapshotID = snap.SnapshotID
+	}
+
+	// Serialize the actual row filter as Expression JSON. On a marshal error we
+	// leave Filter unset so ScanReport.MarshalJSON falls back to always-true
+	// rather than failing the scan; literal sanitization (Java's ExpressionUtil)
+	// is a follow-up.
+	var filter json.RawMessage
+	if scan.rowFilter != nil {
```
zeroshade, line 121: `ScanReport.Filter` serializes the raw rowFilter incl. literal predicate values. Java emits `ExpressionUtil.sanitize(...)`. This diverges from Java and can leak user predicate literals to reporters/REST metrics sinks — sanitize before emit.

table/scan_metrics.go
```diff
@@ -0,0 +1,202 @@
...
+			DVs:                        counterCount(acc.dvs),
+		},
+	}
+}
+
+// projectedFields resolves the scan's selected fields to their ids and names
+// against schema. A "*" (or empty) selection projects all top-level fields.
+// Unresolvable names are skipped.
+func (scan *Scan) projectedFields(schema *iceberg.Schema) ([]int, []string) {
+	if len(scan.selectedFields) == 0 || slices.Contains(scan.selectedFields, "*") {
```
zeroshade: `projectedFields` returns only top-level/explicit names. Java uses `TypeUtil.getProjectedIds` + `findColumnName`, including nested field ids/names. Nested-schema scans will under-report `projected-field-ids`/`-names` (current tests are flat only).

zeroshade (changes requested):
Plumbing, no-op fast path, local-only emission, timer placement and counter concurrency are all sound. Holding on Java/spec parity of the emitted `ScanReport`: delete-file result metrics, unsanitized filter literals, and nested projected fields. CI green.

tanmayrauth:
@zeroshade Thanks for the review, addressed all three in the latest commit: delete-file result metrics now follow Java's ScanMetricsUtil.fileTask per-task accumulation (a delete applying to N tasks counts N times, no cross-path dedup) so the Java-parity field names stay directly comparable, while DV sizing keeps using per-blob content_size_in_bytes; the emitted ScanReport.Filter is now run through a new iceberg.SanitizeExpression before serialization so predicate
  literals are masked (structure, column refs, and ops preserved, IN/NOT IN keep their arity) and never leak to reporters/REST sinks — it masks values rather than replicating Java's typed placeholders, which I noted in the doc
  comment; and projectedFields now resolves the actual projected schema and maps every id (including nested) to its full dotted name via Schema.FindColumnName, matching TypeUtil.getProjectedIds + findColumnName, with new tests
  covering nested schemas, filter redaction, and the per-task delete counts.

table/scan_metrics.go
```diff
@@ -0,0 +1,188 @@
...
+
+// projectedFields resolves the scan's projection to the field ids and dotted
+// column names it reads, matching Java's TypeUtil.getProjectedIds +
+// Schema.findColumnName: nested fields are reported by id and full dotted path,
+// not just top-level columns. schema is the schema the scan planned against and
+// is used to resolve names; ids are returned sorted for deterministic output.
+// Ids not resolvable to a name in schema (e.g. reserved row-lineage columns) are
+// skipped so ids and names stay aligned.
+func (scan *Scan) projectedFields(schema *iceberg.Schema) ([]int, []string) {
+	projected, err := scan.Projection()
```
laskoviymishka: threading the schema through `buildScanReport` is exactly what I wanted, thanks — but `projectedFields` still re-derives the ids from `scan.Projection()` and only uses the schema param for the names.

that leaves a seam: if `Projection()` fails for a reason planning didn't hit (the row-lineage version check, say), we silently return nil, nil and emit a report with no projected fields even though planning succeeded. since we already have the exact schema the scan planned against, I'd walk that directly (`IndexByID(schema)` / iterate its fields) and drop the `scan.Projection()` call. wdyt?

table/scan_metrics_test.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	sr := scan.buildScanReport(&scanMetricsAccumulator{}, meta.CurrentSchema(), time.Millisecond)
+	require.NotNil(t, sr.Filter)
+	assert.NotContains(t, string(sr.Filter), "secret-value", "literal must not leak into the report")
+	assert.Contains(t, string(sr.Filter), "data", "column reference is preserved")
+	assert.Contains(t, string(sr.Filter), "(redacted)")
+}
+
+func TestPlanFilesNoSnapshotEmitsNoReport(t *testing.T) {
```
laskoviymishka, line 246: the real-snapshot counter test (`TestFetchManifestCountersWithRealSnapshot`) is in and covers the manifest accounting nicely — the one path still untested is the emission happy path.

nothing runs `PlanFiles` against a real snapshot with a live reporter and asserts the emitted report carries the right `SnapshotID` and non-zero counters. `buildScanReport` is only unit-tested with a no-snapshot scan, so the `snapshotID = snap.SnapshotID` branch never actually runs, and a regression in the `Snapshot() != nil` gate or the id population would be invisible. could we add a positive sibling to this test that plans through a real snapshot with an `InMemoryReporter` and asserts `Reports()` has one report with the expected id?

visitors.go
```diff
@@ -520,3 +520,84 @@ func (c columnNameTranslator) VisitBound(pred BoundPredicate) BooleanExpression
...
+// by an opaque placeholder while preserving the boolean structure, column
+// references, and operations. It lets a filter be emitted somewhere untrusted
+// (e.g. a metrics ScanReport shipped to a REST sink) without leaking the literal
+// values a user scanned with, mirroring the intent of Java's ExpressionUtil.sanitize.
+//
+// Unlike Java, which substitutes type-shaped placeholders (e.g. "(2-digit-int)"),
+// every literal is masked with the same marker; the goal here is to withhold the
+// values, not to preserve their shape. Set predicates (IN / NOT IN) keep their
+// arity so the operation is not misrepresented, but the members are masked.
+func SanitizeExpression(expr BooleanExpression) (BooleanExpression, error) {
```
laskoviymishka, line 541: not blocking — sanitizing before the filter ships is great, and further than I asked for in round 1.

now that `SanitizeExpression` is exported, the `(redacted)` marker is effectively a public contract, and it diverges from Java's type-shaped placeholders, so a catalog can't correlate these filters across clients. the doc comment already flags it, which is the right call — I'd just have the godoc say the guarantee is "no user value leaks," not structural parity with Java, so we're free to change the format later without it being a breaking change. a `NOT IN` case in the test wouldn't hurt either, since it rides the same set-predicate branch as `IN`.

laskoviymishka (approved):
LGTM! thanks for grinding through all of it.

Two small things left, and neither blocks merge for me. `projectedFields` still re-derives the ids from `scan.Projection()` rather than the schema we now thread in, which quietly reopens the nil-projection path we were trying to close; since we already have the planned schema, I'd walk it directly. And there's no positive test that runs `PlanFiles` against a real snapshot and asserts the emitted report carries the right `SnapshotID` and non-zero counters — `buildScanReport` is only exercised with a no-snapshot scan, so the id-population branch never runs.

Everything I asked for in round 1 is in:
- `ScanReport.Filter` set from the row filter, and sanitized via `SanitizeExpression` (which I'd said could follow separately — nice to have it already).
- `applyResultDeleteMetrics` counts per task matching Java, sizes DVs by `ContentSizeInBytes()`, and `TestApplyResultDeleteMetricsDVsShareOnePuffin` covers two DVs in one Puffin file.
- emission gated on `scan.Snapshot() != nil`, so no more `SnapshotID: 0`.
- the planning schema threaded into `buildScanReport` rather than re-resolved.
- `TestFetchManifestCountersWithRealSnapshot` asserting non-zero manifest counters against a real snapshot.

zeroshade's parity items — delete-file metrics, filter-literal sanitization, nested projected fields — all look handled here too. Sort out those two and I'm happy to approve.

tanmayrauth:
Thanks for the review @laskoviymishka. 

On projectedFields: rather than walking the threaded schema directly, I resolve the projection once in PlanFiles and thread the projected schema into  buildScanReport, then walk that (dropping the second scan.Projection() call). The reason for the extra step is that the schema we thread for planning is the full effective schema; walking it directly would report every column as  projected instead of the selected subset (diverging from Java's scan.schema()), so threading the projected schema is what actually closes the nil-projection seam without misreporting the fields. On the test: added TestPlanFilesEmitsReportForRealSnapshot, a positive sibling that plans a real snapshot through an InMemoryReporter and asserts one report with the expected SnapshotID and non-zero counters, so the Snapshot() != nil gate and id-population branch are now exercised. Also took the non-blocking SanitizeExpression note — reworded the godoc so the only guarantee is "no user value leaks" (placeholder format explicitly non-contractual) and added a NOT IN case alongside the IN test.

zeroshade (approved):
All three parity gaps from the last pass are closed, and each is now pinned by a test — thanks for the careful Java cross-referencing:

- **Delete-file metrics** accumulate per data-file task (`applyResultDeleteMetrics`), matching `ScanMetricsUtil.fileTask` — a delete covering N tasks counts N times, no cross-path dedup — split across positional/equality/DV, with DV size taken from `content_size_in_bytes` rather than the Puffin length. Covered by `TestApplyResultDeleteMetrics` + `…DVsShareOnePuffin`.
- **Filter literals** are masked through the new `iceberg.SanitizeExpression` before serialization: structure, column refs and ops preserved, IN/NOT-IN keep their arity (distinct masked members so they don't collapse to EQ), values withheld, with a safe fallback on marshal error. `TestSanitizeExpression` asserts no-leak + arity + structure; `TestBuildScanReportSanitizesFilter` covers the emit path.
- **Projected fields** resolve via `IndexByID` + `FindColumnName` over the threaded projected schema, so nested fields are reported by id and full dotted path — `TestProjectedFieldsNested`.

Diverging from Java's typed placeholders (single `(redacted)` marker) is the right call and it's documented. LGTM.

PR #1503, opened 2026-07-21
fix(io/gcs): authenticate the GCS client with supplied credentials

iremcaginyurtturk:
## Problem

`createGCSBucket` (`io/gocloud/gcs.go`) builds the authenticated HTTP client from Application Default Credentials — or an anonymous client when ADC is absent — and ignores the credentials parsed from the `gcs.jsonkey` (`GCSJSONKey`) and `gcs.keypath` (`GCSKeyPath`) properties. `ParseGCSConfig` turns those into client options, but they are passed to `gcsblob.OpenBucket` alongside the pre-built client, whose transport takes precedence — so the explicit credentials never authenticate requests.

**Effect:** with a catalog configured with explicit GCS service-account credentials but **no** `GOOGLE_APPLICATION_CREDENTIALS` in the environment, GCS reads/writes go out **anonymously** and fail with `HTTP 403 AccessDenied`. The credentials only take effect if the same account is also exported via ADC.

## Fix

Resolve credentials from `gcs.jsonkey` / `gcs.keypath` first (via the non-deprecated `CredentialsFromJSONWithType`, defaulting to a service-account key and honouring `gcs.credtype`) and build the client's token source from them. Fall back to ADC (which may be nil → anonymous client) only when neither property is set.

### DIFF
--- a/io/gocloud/gcs.go
+++ b/io/gocloud/gcs.go
@@ -19,7 +19,9 @@ package gocloud
 
 import (
 	"context"
+	"fmt"
 	"net/url"
+	"os"
 	"strconv"
 
 	"cloud.google.com/go/storage"
@@ -28,6 +30,7 @@ import (
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/gcsblob"
 	"gocloud.dev/gcp"
+	"golang.org/x/oauth2/google"
 	"google.golang.org/api/option"
 )
 
@@ -67,15 +70,58 @@ func ParseGCSConfig(props map[string]string) *gcsblob.Options {
 	}
 }
 
+// gcsScope grants read/write access to GCS objects.
+const gcsScope = "https://www.googleapis.com/auth/devstorage.read_write"
+
+// gcsCredentials builds credentials from the gcs.jsonkey / gcs.keypath
+// properties (defaulting to a service-account key, honouring gcs.credtype),
+// falling back to Application Default Credentials when neither is set.
+func gcsCredentials(ctx context.Context, props map[string]string) (*google.Credentials, error) {
+	credType := google.ServiceAccount
+	if v := props[io.GCSCredType]; v != "" {
+		credType = google.CredentialsType(v)
+	}
+
+	if jsonKey := props[io.GCSJSONKey]; jsonKey != "" {
+		creds, err := google.CredentialsFromJSONWithType(ctx, []byte(jsonKey), credType, gcsScope)
+		if err != nil {
+			return nil, fmt.Errorf("gcs: parsing %s: %w", io.GCSJSONKey, err)
+		}
+
+		return creds, nil
+	}
+
+	if path := props[io.GCSKeyPath]; path != "" {
+		data, err := os.ReadFile(path)
+		if err != nil {
+			return nil, fmt.Errorf("gcs: reading %s %q: %w", io.GCSKeyPath, path, err)
+		}
+		creds, err := google.CredentialsFromJSONWithType(ctx, data, credType, gcsScope)
+		if err != nil {
+			return nil, fmt.Errorf("gcs: parsing credentials file %q: %w", path, err)
+		}
+
+		return creds, nil
+	}
+
+	creds, _ := gcp.DefaultCredentials(ctx)
+
+	return creds, nil
+}
+
 // Construct a GCS bucket from a URL
 func createGCSBucket(ctx context.Context, parsed *url.URL, props map[string]string) (*blob.Bucket, error) {
 	gcscfg := ParseGCSConfig(props)
-	creds, _ := gcp.DefaultCredentials(ctx)
+
+	creds, err := gcsCredentials(ctx, props)
+	if err != nil {
+		return nil, err
+	}
+
 	var client *gcp.HTTPClient
 	if creds == nil {
 		client = gcp.NewAnonymousHTTPClient(gcp.DefaultTransport())
 	} else {
-		var err error
 		client, err = gcp.NewHTTPClient(
 			gcp.DefaultTransport(),
 			gcp.CredentialsTokenSource(creds))
--- a/io/gocloud/gcs.go
+++ b/io/gocloud/gcs.go
@@ -34,30 +34,31 @@ import (
 	"google.golang.org/api/option"
 )
 
-var allowedGCSCredTypes = map[string]option.CredentialsType{
-	"service_account":              option.ServiceAccount,
-	"authorized_user":              option.AuthorizedUser,
-	"impersonated_service_account": option.ImpersonatedServiceAccount,
-	"external_account":             option.ExternalAccount,
+var allowedGCSCredTypes = map[string]struct{}{
+	"service_account":              {},
+	"authorized_user":              {},
+	"impersonated_service_account": {},
+	"external_account":             {},
 }
 
-// ParseGCSConfig parses GCS properties and returns a configuration.
+func resolveGCSCredType(props map[string]string) (string, bool) {
+	v := props[io.GCSCredType]
+	if v == "" {
+		return "", false
+	}
+	if _, ok := allowedGCSCredTypes[v]; !ok {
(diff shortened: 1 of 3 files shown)


io/gocloud/gcs.go
```diff
@@ -67,15 +70,58 @@ func ParseGCSConfig(props map[string]string) *gcsblob.Options {
...
 }
 
+// gcsScope grants read/write access to GCS objects.
+const gcsScope = "https://www.googleapis.com/auth/devstorage.read_write"
+
+// gcsCredentials builds credentials from the gcs.jsonkey / gcs.keypath
+// properties (defaulting to a service-account key, honouring gcs.credtype),
+// falling back to Application Default Credentials when neither is set.
+func gcsCredentials(ctx context.Context, props map[string]string) (*google.Credentials, error) {
+	credType := google.ServiceAccount
```
tanmayrauth, line 105: Defaulting to service_account means a gcs.jsonkey / gcs.keypath holding an authorized_user, external_account, or impersonated_service_account credential fails unless gcs.credtype is also set — CredentialsFromJSONWithType enforces an exact type match. With gcs.jsonkey = an authorized_user JSON and no gcs.credtype you get:
  
    gcs: parsing gcs.jsonkey: google: expected credential type "service_account", found "authorized_user"
    
  Not a regression (those keys were ignored before too), but the error doesn't hint that gcs.credtype is the fix. Either document that gcs.credtype is required for non-service-account keys, or wrap the mismatch error with a hint naming gcs.credtype.

io/gocloud/gcs.go
```diff
@@ -67,15 +70,58 @@ func ParseGCSConfig(props map[string]string) *gcsblob.Options {
...
+	if v := props[io.GCSCredType]; v != "" {
+		credType = google.CredentialsType(v)
```
tanmayrauth, line 107: This casts gcs.credtype raw with no validation, while ParseGCSConfig validates it against allowedGCSCredTypes and silently drops unknown values — two parsers for one property, over two different type spaces (option.CredentialsType vs google.CredentialsType). A typo'd credtype here surfaces as the confusing "expected credential type X, found Y" instead of being ignored. Normalize/validate gcs.credtype once and share it between both paths.

io/gocloud/gcs.go
```diff
@@ -67,15 +70,58 @@ func ParseGCSConfig(props map[string]string) *gcsblob.Options {
...
+	}
+
+	if jsonKey := props[io.GCSJSONKey]; jsonKey != "" {
+		creds, err := google.CredentialsFromJSONWithType(ctx, []byte(jsonKey), credType, gcsScope)
```
tanmayrauth: ParseGCSConfig still appends WithAuthCredentialsJSON/WithAuthCredentialsFile into gcscfg.ClientOptions, but those are dead now: OpenBucket wraps the pre-built client as WithHTTPClient and the api transport (dial.go:44) returns that client and ignores credential options. Harmless today but misleading — anyone reading ParseGCSConfig will think jsonkey/keypath authenticate through it when gcsCredentials is the only thing that matters. Drop the credential options from ParseGCSConfig, or add a comment noting they're superseded.

io/gocloud/gcs_test.go
```diff
@@ -45,3 +52,37 @@ func TestParseGCSConfigUseJSONAPI(t *testing.T) {
 		assert.Len(t, cfg.ClientOptions, 0)
 	})
 }
+
+// Inline JSON key (gcs.jsonkey) yields explicit credentials, not ADC.
+func TestGCSCredentialsFromInlineJSONKey(t *testing.T) {
+	creds, err := gcsCredentials(context.Background(), map[string]string{
+		io.GCSJSONKey:  authorizedUserJSON,
+		io.GCSCredType: "authorized_user",
```
tanmayrauth, line 66: Both credential tests always pass gcs.credtype explicitly, so neither exercises the default (service_account) path — which is exactly where a non-SA key fails. Add a case with a service_account-typed JSON and no gcs.credtype (should succeed) and one with an authorized_user JSON and no gcs.credtype (locks the current behavior) so the default path is covered.

tanmayrauth (commented):
Core fix is right — putting the token source on the client is what makes gcs.jsonkey/gcs.keypath actually authenticate, and the service_account path now works. A couple of usability/test nits. Details inline.

io/gocloud/gcs.go
```diff
@@ -67,18 +71,60 @@ func ParseGCSConfig(props map[string]string) *gcsblob.Options {
...
+		creds, err := google.CredentialsFromJSONWithType(ctx, data, credType, gcsScope)
+		if err != nil {
+			return nil, fmt.Errorf("gcs: parsing %s: set %s if the key is not a service account: %w", source, io.GCSCredType, err)
+		}
+
+		return creds, nil
+	}
+
+	if jsonKey := props[io.GCSJSONKey]; jsonKey != "" {
+		return parse([]byte(jsonKey), io.GCSJSONKey)
```
zeroshade, line 122: Only `gcs.jsonkey`/`gcs.keypath` are handled before falling back to ADC. Iceberg/PyIceberg and REST-vended creds use `gcs.oauth2.token` (+ `gcs.oauth2.token-expires-at`, already recognized here); that token is never turned into an `option.WithTokenSource`, so a vended GCS token is silently dropped.

io/gocloud/gcs.go
```diff
@@ -67,18 +71,60 @@ func ParseGCSConfig(props map[string]string) *gcsblob.Options {
...
+
+	creds, err := gcsCredentials(ctx, props)
+	if err != nil {
+		return nil, err
+	}
+
 	var client *gcp.HTTPClient
 	if creds == nil {
 		client = gcp.NewAnonymousHTTPClient(gcp.DefaultTransport())
 	} else {
```
zeroshade, line 151: No explicit anonymous/no-auth mode: with `gcs.no-auth=true` and ADC present, this still builds an authenticated client. Consider honoring `gcs.no-auth` explicitly via `option.WithoutAuthentication`.

io/gocloud/gcs.go
```diff
@@ -28,34 +30,36 @@ import (
...
+		return "", false
+	}
+
+	return v, true
+}
+
+// ParseGCSConfig parses non-credential GCS blob options; credentials are set on
+// the client by gcsCredentials, which supersedes any credential option here.
 func ParseGCSConfig(props map[string]string) *gcsblob.Options {
 	var o []option.ClientOption
```
zeroshade, line 71: Endpoint override keys off `gcs.endpoint`; Java/PyIceberg use `gcs.service.host`. Consider accepting the standard name as an alias or documenting the difference.

zeroshade (changes requested):
Good fix for the jsonkey/keypath path, and ADC fallback is preserved. But “supplied credentials” isn't complete: REST-vended `gcs.oauth2.token` and explicit `gcs.no-auth` are still ignored, so a vended token silently falls through to ADC/anonymous. Minor: `gcs.endpoint` vs Iceberg's `gcs.service.host` naming. CI green.

iremcaginyurtturk:
@zeroshade  Thanks, addressed all three in 76f3732

zeroshade (approved):
All three from the last pass are handled, each with a test:

- Vended `gcs.oauth2.token` is now turned into a `StaticTokenSource` (with `-expires-at`) instead of being dropped → `TestGCSCredentialsFromOAuthToken`.
- `gcs.no-auth=true` short-circuits to an anonymous client *before* ADC, so it wins even when ADC is present → `TestGCSCredentialsNoAuth`.
- `gcsEndpoint` prefers the standard `gcs.service.host`, falling back to `gcs.endpoint` → `TestGCSEndpointPrefersServiceHost`.

Bonus: a missing `gcs.keypath` is now a hard error rather than a silent ADC fallback (`TestGCSCredentialsMissingKeyPath`), and the `gcs.credtype` hint on non-service-account keys is a nice touch. LGTM.

PR #1505, opened 2026-07-22
test(rest): add Java scan planning integration suite

### LINKED ISSUE #1500: REST scan planning: Java iceberg-rest-fixture integration suite (Phase 6 follow-up)
Follow-up to #1494 (part of #1178, Phase 6).

The client-conformance fake added in #1494 uses hand-authored JSON fixtures. They are written to match Java's `ContentFileParser` conventions (typed partition values, hex-encoded binary bounds), but nothing automatically verifies they match real Java server output.

## Goal

Add an integration suite, gated by `RUN_INTEGRATION_TESTS=1` (matching the existing integration-test convention), that runs the REST scan-planning client against the Java `apache/iceberg-rest-fixture` Docker image for interop confidence.

## Notes

- Mirror the existing `make integration-rest` / docker-compose setup.
- Confirms wire-format fidelity end-to-end (endpoint paths, JSON shapes, error types, storage credentials) against a real server, complementing the unit-level fake from #1494.

Revanth14:
Closes #1500.

## Summary

Add integration coverage for REST scan planning against the real Java `apache/iceberg-rest-fixture`, complementing the deterministic fake server added in #1494.

The tests exercise the existing low-level REST planning client and independently validate the Java response wire format.

## Changes

- Pin the Java REST fixture to a reproducible scan-planning-capable image digest built from `apache/iceberg@b0df3ca`.
  - The existing `1.10.1` fixture predates runtime support for the scan-planning routes.
- Configure the fixture to return plan-scoped storage credentials.
- Add integration tests gated by `RUN_INTEGRATION_TESTS=1`.
- Validate:
  - Planning capability advertisement.
  - Plan submission and cancellation endpoints.
  - Access-delegation request headers.
  - Snapshot, projection, filter, statistics, and planning options accepted by Java.
  - Typed JSON partition values.
  - Hex-encoded Iceberg binary bounds.
  - Residual-expression JSON.
  - Plan-scoped storage credentials.
  - Missing-table, expired-plan, and missing-plan-task error mappings.
- Share independently defined wire-fixture types between the fake-server and Java integration tests instead of decoding through production REST response structs.

## Scope

This change is independent of #1492 and #1477. It tests the low-level REST client already available on `main` and does not add production `Catalog.PlanFiles` orchestration or domain-level `FileScanTask` decoding.

The published Java fixture currently produces synchronous inline plans and does not expose configuration for asynchronous completion or plan-task fan-out. Successful asynchronous polling and nested fan-out remain covered deterministically by the fake server from #1494.

End-to-end coverage through `Scan.PlanFiles()` should be added after #1492 and #1477 land.

## Verification

- `make integration-rest`
- `go test ./...`
- `go test -race ./catalog/rest/...`
- `RUN_INTEGRATION_TESTS=1 go test -race -tags=integration -run '^TestRestIntegration/TestScanPlanningJava' ./catalog/rest`
- `go vet ./...`
- `go vet -tags=integration ./catalog/rest`
- `docker compose -f internal/recipe/docker-compose.yml config --quiet`
- `git diff --check`


catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,321 @@
...
+	planID := *response.PlanID
+	planCancelled := false
+	s.T().Cleanup(func() {
+		if planCancelled {
+			return
+		}
+
+		_ = cancelPlanningWithTimeout(planningCatalog, ident, planID)
+	})
+	s.Empty(response.PlanTasks)
```
zeroshade, line 97: Non-blocking: this asserts `PlanTasks` is empty and only `Len(FileScanTasks, 1)`, so successful Java `plan-tasks` fanout / `FetchScanTasks` traversal isn't exercised against the real server. Consider scoping the name/description to “synchronous” or adding a fanout case if the fixture can produce it.

catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,321 @@
...
+	s.Require().Len(response.FileScanTasks, 1)
```
zeroshade, line 98: Non-blocking: the Go-decoded task payload isn't asserted beyond length. Once #1492's decoder is in the base, add decoded-field assertions so a decoder regression can't pass while the raw-JSON checks succeed.

zeroshade (commented):
Nice addition — a real Java REST fixture check for synchronous scan planning, with safe CI gating (`//go:build integration`, digest-pinned fixture) and deterministic assertions/cleanup. Two non-blocking suggestions below. It's a strong synchronous-plan/wire-compat check rather than a full scan-planning suite (no successful plan-task fanout, and decoded task contents aren't asserted — the latter depends on #1492's decoder being in the base). CI green.

Makefile
```diff
@@ -67,7 +67,7 @@ integration-io:
 	go test -tags=integration -v ./io/...
 
 integration-rest:
-	go test -tags=integration -v -run="^TestRestIntegration$$" ./catalog/rest
+	RUN_INTEGRATION_TESTS=1 go test -tags=integration -v -run="^TestRestIntegration$$" ./catalog/rest
```
laskoviymishka: `integration-rest` is the main CI target for the whole REST suite, so setting `RUN_INTEGRATION_TESTS=1` here opts every run into the Java scan-planning tests — not just this one — and those depend on the new pre-release image, table creation, parquet writes to MinIO, and credential vending all lining up.

I'd pull the scan-planning tests into their own target (`integration-rest-scan-planning`) and set the env var there, leaving `integration-rest` on the stable path. wdyt?

catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,321 @@
...
+	s.Equal(dataPath, task.DataFile.FilePath)
+	s.Require().Len(task.DataFile.Partition, 1)
+	s.Equal("17", string(task.DataFile.Partition[0]),
+		"Java partitions must remain typed JSON values, not binary-bound strings")
+	s.JSONEq(`{"type":"eq","term":"foo","value":"hello"}`, string(task.ResidualFilter))
+	s.Empty(task.DeleteFileReferences)
+
+	lowerBounds := valueMapByFieldID(s, task.DataFile.LowerBounds)
+	upperBounds := valueMapByFieldID(s, task.DataFile.UpperBounds)
+	s.ElementsMatch([]int{1, 2}, task.DataFile.LowerBounds.Keys)
```
laskoviymishka: These key checks are non-fatal, but the hex assertions right below them index `lowerBounds[1]` and `lowerBounds[2]`. If Java ever returns different field IDs, `ElementsMatch` fails, the test keeps running, and then `s.Equal("68656C6C6F", lowerBounds[1])` compares against a zero-value `""` — so the failure you actually see is about hex encoding when the real problem is the field IDs.

I'd make these `s.Require().ElementsMatch` so the test stops at the true cause.

catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,321 @@
...
+func (s *RestIntegrationSuite) requireScanPlanningIntegration() {
+	s.T().Helper()
+	if os.Getenv(runIntegrationTestsEnv) != "1" {
+		s.T().Skipf("set %s=1 to run Java REST scan-planning integration tests", runIntegrationTestsEnv)
+	}
+}
+
+func (s *RestIntegrationSuite) requireJavaScanPlanningCapabilities(cat *rest.Catalog) {
+	s.T().Helper()
+	s.Require().True(cat.SupportsPlanTableScan(), "Java fixture must advertise planTableScan")
```
laskoviymishka: If someone opts in with `RUN_INTEGRATION_TESTS=1` but happens to have the old released image up, this `s.Require().True` fails hard instead of skipping — which reads as a broken test rather than "wrong image". `requireScanPlanningIntegration` already skips cleanly; I'd have this one `Skipf` too when the server doesn't advertise the endpoints, so the two guards behave consistently.

catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,321 @@
...
+	upperBounds := valueMapByFieldID(s, task.DataFile.UpperBounds)
+	s.ElementsMatch([]int{1, 2}, task.DataFile.LowerBounds.Keys)
+	s.ElementsMatch([]int{1, 2}, task.DataFile.UpperBounds.Keys)
+	s.Equal("68656C6C6F", lowerBounds[1], "Java string lower bounds must be uppercase hexadecimal")
+	s.Equal("68656C6C6F", upperBounds[1], "Java string upper bounds must be uppercase hexadecimal")
+	s.Equal("11000000", lowerBounds[2], "Java int lower bounds must use Iceberg little-endian binary")
+	s.Equal("11000000", upperBounds[2], "Java int upper bounds must use Iceberg little-endian binary")
+
+	s.Require().Len(raw.StorageCredentials, 1)
+	s.Equal("gcp", raw.StorageCredentials[0].Prefix)
```
laskoviymishka, line 138: Worth a one-line comment here: the fixture vends these creds under prefix `"gcp"`, but GCSFileIO (Java and Go alike) only matches prefixes starting with `"gs"`, so these creds wouldn't actually resolve for a real `gs://` path today. That's a known fixture quirk, not something this PR introduces — pinning `"gcp"` is fine as a canary that'll trip when the fixture is fixed, but a comment saves the next reader from reading it as a bug and "correcting" it to `"gs"`.

catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+func (s *RestIntegrationSuite) assertJavaPlanningCredentials(credentials []rest.StorageCredential) {
+	s.T().Helper()
+	s.Require().Len(credentials, 1)
+	s.Equal("gcp", credentials[0].Prefix)
+	s.Equal(integrationGCSToken, credentials[0].Config[integrationGCSTokenKey])
+}
+
+func valueMapByFieldID(s *RestIntegrationSuite, valueMap rawValueMapFixture) map[int]string {
```
laskoviymishka: Small consistency thing: every other helper in this file is a method on `*RestIntegrationSuite` (`assertJavaPlanningCredentials` right below is), but this one takes the suite as its first arg. I'd make it a method — `func (s *RestIntegrationSuite) valueMapByFieldID(valueMap rawValueMapFixture) map[int]string` — so call sites read `s.valueMapByFieldID(...)`.

catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,321 @@
...
+type scanPlanningCaptureTransport struct {
+	base *http.Transport
+
+	mu            sync.Mutex
+	planRequests  []http.Header
+	planResponses [][]byte
+}
+
+func newScanPlanningCaptureTransport() *scanPlanningCaptureTransport {
+	return &scanPlanningCaptureTransport{base: http.DefaultTransport.(*http.Transport).Clone()}
```
laskoviymishka: `http.DefaultTransport.(*http.Transport)` is unguarded — if anything in the test binary has swapped `DefaultTransport` for a non-`*http.Transport`, this panics with no test-failure record and takes the whole binary down with it.

Since this constructor doesn't have `s`/`t` handy, the cleanest fix is to build a fresh `&http.Transport{}` instead of cloning the default — you sidestep the assertion entirely, and the capture transport doesn't really need `DefaultTransport`'s config.

internal/recipe/docker-compose.yml
```diff
@@ -46,7 +46,11 @@ services:
       retries: 12
       start_period: 30s
   rest:
-    image: apache/iceberg-rest-fixture:1.10.1
+    # 1.10.1 predates the fixture's scan-planning implementation. Pin the
+    # multi-platform image built from apache/iceberg@b0df3ca so the Java wire
+    # compatibility tests are reproducible even though the released tag is not
+    # available yet.
+    image: apache/iceberg-rest-fixture:latest@sha256:db8de90b5b7693d4ac334c336f91d9bbe320d7b19f4f514d26de84cdfbcbfe8d
```
laskoviymishka, line 67: This is the piece I'd most want to revisit: the shared `rest` service moves from released 1.10.1 to a digest-pinned build off an unreleased `apache/iceberg@b0df3ca`, so every existing REST integration test now runs against a pre-release image. Two risks — a regression in that image shows up as a failure in some unrelated test with no hint it was the bump, and a `:latest@sha256` pin with no release tag keeping it alive could get GC'd on Docker Hub and break CI looking like a flaky pull.

I'd run the scan-planning tests against a second scan-planning-capable service on its own port and leave `rest` on the released tag. If that's too heavy for now, at minimum a tracking issue to move to the release tag once 1.11.0 ships, plus a TODO here. wdyt?

internal/recipe/docker-compose.yml
```diff
@@ -59,6 +63,10 @@ services:
       - CATALOG_WAREHOUSE=s3://warehouse/
       - CATALOG_IO__IMPL=org.apache.iceberg.aws.s3.S3FileIO
       - CATALOG_S3_ENDPOINT=http://minio:9000
+      - CATALOG_INCLUDE__CREDENTIALS=true
```
laskoviymishka, line 80: With include-credentials on, the fixture now injects `gcs.oauth2.token` into every `LoadTableResult`, so the shared fixture state isn't minimal anymore. Go ignores the property today, but any future test asserting an exact table-config map would trip on it. This is the same thread as isolating the scan-planning fixture above — if that service gets split out, this setting rides along with it and stops touching the shared tests.

laskoviymishka (changes requested):
Really nice suite. The `//go:build integration` gating keeps this out of the default build, the capture transport that lets you assert the decoded response and the raw Java wire in a single pass is a clean design, and keeping the wire types independent from the production structs — so both sides can't quietly agree on a wrong encoding — is well-justified in the comment. The typed-partition and hex-bounds assertions are exactly the cross-client contracts worth pinning.

I'd hold on one thing before merge, and it isn't the test logic — it's blast radius. Moving the shared `rest` service to an unreleased digest-pinned image and adding `RUN_INTEGRATION_TESTS=1` to `integration-rest` means every existing REST integration test now runs against a pre-release image on the main CI target. If that image regresses on anything unrelated to scan planning, it shows up as a confusing failure somewhere else entirely, and a `:latest@sha256` pin with no release tag keeping it alive can get GC'd out from under CI. I'd isolate the scan-planning tests onto their own service/target and leave the shared `rest` path on the released tag.

The rest is test-robustness polish: a couple of non-fatal assertions that report the wrong cause on failure, one unguarded type assertion that could panic the whole binary, and a helper that breaks the method convention of the file. Details inline.

- isolate the scan-planning image/target from the shared REST integration suite (or a tracking issue + TODO to move to the release tag once 1.11.0 ships)
- `Require().ElementsMatch` before the bound-value lookups so a field-ID mismatch doesn't masquerade as a hex-encoding failure
- guard or replace the `http.DefaultTransport` type assertion so a swapped transport can't panic the binary

zeroshade (approved):
The scope rename to "synchronous planning" resolves my main note — the test is now honestly labeled for what it exercises (synchronous plan / wire-compat), rather than implying full plan-task fanout. The decoded-task-field assertions remain a sensible follow-up once #1492's decoder is in the base. Nice safe CI gating (`//go:build integration` + `RUN_INTEGRATION_TESTS=1`, digest-pinned fixture). LGTM.

Makefile
Revanth14: Good point, agreed. I moved the Java scan-planning tests into a separate `integration-rest-scan-planning` target and scoped `RUN_INTEGRATION_TESTS=1` to that target. `integration-rest` remains on the stable REST/conformance path.

I also separated the scan-planning fixture onto port `8182` so it doesn’t affect the released REST fixture. Thanks!

catalog/rest/scan_planning_integration_test.go
Revanth14: Good catch. I changed both key checks to `s.Require().ElementsMatch`, so the test now stops immediately if Java returns unexpected field IDs instead of continuing with misleading hex-encoding failures.

catalog/rest/scan_planning_integration_test.go
Revanth14: Agreed. I updated the capability guard to call `Skip` when the server doesn’t advertise the required scan-planning endpoints. This now behaves consistently with `requireScanPlanningIntegration` and reports a mismatched fixture image as a skipped test rather than a failure.

catalog/rest/scan_planning_integration_test.go
Revanth14: Good point. I added a comment explaining that `"gcp"` is a known fixture quirk, why it does not resolve for real `gs://` paths, and that the assertion intentionally acts as a canary for a future fixture fix.

catalog/rest/scan_planning_integration_test.go
Revanth14: Agreed. I converted `valueMapByFieldID` into a method on `*RestIntegrationSuite` and updated the call sites to use `s.valueMapByFieldID(...)`.

catalog/rest/scan_planning_integration_test.go
Revanth14: I replaced the unchecked `http.DefaultTransport.(*http.Transport)` assertion with a fresh `&http.Transport{}`, since this capture transport doesn’t depend on the default transport’s configuration.

internal/recipe/docker-compose.yml
Revanth14: Agreed. I split the fixtures into two services:

- `rest` remains pinned to the released `1.10.1` image on port `8181`.
- `rest-scan-planning` uses the digest-pinned scan-planning image on port `8182`.

The regular REST integration tests therefore stay on the stable released fixture, while only the dedicated scan-planning tests depend on the pre-release image.

internal/recipe/docker-compose.yml
Revanth14: Agreed. After splitting the fixtures, `CATALOG_INCLUDE__CREDENTIALS=true` now applies only to the dedicated `rest-scan-planning` service. The shared `rest` fixture remains unchanged, so regular REST integration tests no longer receive the additional credential configuration.

Revanth14:
> Really nice suite. The `//go:build integration` gating keeps this out of the default build, the capture transport that lets you assert the decoded response and the raw Java wire in a single pass is a clean design, and keeping the wire types independent from the production structs — so both sides can't quietly agree on a wrong encoding — is well-justified in the comment. The typed-partition and hex-bounds assertions are exactly the cross-client contracts worth pinning.
> 
> I'd hold on one thing before merge, and it isn't the test logic — it's blast radius. Moving the shared `rest` service to an unreleased digest-pinned image and adding `RUN_INTEGRATION_TESTS=1` to `integration-rest` means every existing REST integration test now runs against a pre-release image on the main CI target. If that image regresses on anything unrelated to scan planning, it shows up as a confusing failure somewhere else entirely, and a `:latest@sha256` pin with no release tag keeping it alive can get GC'd out from under CI. I'd isolate the scan-planning tests onto their own service/target and leave the shared `rest` path on the released tag.
> 
> The rest is test-robustness polish: a couple of non-fatal assertions that report the wrong cause on failure, one unguarded type assertion that could panic the whole binary, and a helper that breaks the method convention of the file. Details inline.
> 
> A few things I'd want before merge:
> 
> * isolate the scan-planning image/target from the shared REST integration suite (or a tracking issue + TODO to move to the release tag once 1.11.0 ships)
> * `Require().ElementsMatch` before the bound-value lookups so a field-ID mismatch doesn't masquerade as a hex-encoding failure
> * guard or replace the `http.DefaultTransport` type assertion so a swapped transport can't panic the binary
> 
> Once those are addressed, happy to take another pass and approve.

Thanks for the thorough review. I’ve addressed all the requested changes, including isolating the scan-planning fixture/target and tightening the test assertions and transport setup. 
Please take a look when you got some time @laskoviymishka

zeroshade (approved):
Re-reviewed the fixture-isolation delta: the dedicated  compose service on 8182, catalog-scoped helpers, and skipping (rather than failing) when the fixture image lacks scan-planning capabilities all look good. Thanks for the quick turnaround!

zeroshade:
@laskoviymishka can you take a look here for your requested changes review?

Makefile
```diff
@@ -58,7 +58,7 @@ integration-env:
 	@echo "export SPARK_CONTAINER_ID=$$(docker ps -qf 'name=spark-iceberg')"
 	@echo "export DOCKER_API_VERSION=$$(docker version -f '{{.Server.APIVersion}}')"
 
-integration-test: integration-scanner integration-io integration-rest integration-spark integration-hive integration-hadoop
+integration-test: integration-scanner integration-io integration-rest integration-rest-scan-planning integration-spark integration-hive integration-hadoop
```
laskoviymishka, line 64: The isolation is exactly what I wanted. One thing I'd tidy in a follow-up, not blocking: pulling `integration-rest-scan-planning` into the `integration-test` deps means a plain `make integration-test` pulls and runs the pre-release image again, since the target sets `RUN_INTEGRATION_TESTS=1` itself. I'd keep it opt-in (out of the umbrella deps, run explicitly), or note in a comment that the inclusion is deliberate. Fine to leave for later.

internal/recipe/docker-compose.yml
```diff
@@ -59,6 +59,28 @@ services:
       - CATALOG_WAREHOUSE=s3://warehouse/
       - CATALOG_IO__IMPL=org.apache.iceberg.aws.s3.S3FileIO
       - CATALOG_S3_ENDPOINT=http://minio:9000
+  rest-scan-planning:
```
laskoviymishka, line 62: Non-blocking follow-up: this service has no healthcheck, so `docker compose up --wait` calls it ready as soon as the container starts rather than when the fixture is actually serving. It gets probed immediately by the first test body, so a `curl -f http://localhost:8181/v1/config` check with a `start_period` would head off a connection-refused flake on slow CI.

internal/recipe/docker-compose.yml
```diff
@@ -59,6 +59,28 @@ services:
...
+    # 1.10.1 predates the fixture's scan-planning implementation. Pin the
+    # multi-platform image built from apache/iceberg@b0df3ca so the Java wire
+    # compatibility tests are reproducible even though the released tag is not
+    # available yet.
+    image: apache/iceberg-rest-fixture:latest@sha256:db8de90b5b7693d4ac334c336f91d9bbe320d7b19f4f514d26de84cdfbcbfe8d
```
laskoviymishka, line 67: The digest pin itself is right. For a follow-up: `latest@sha256` means once 1.11.0 ships and `:latest` moves, nothing keeps this digest alive and DockerHub can GC it, at which point CI pulls 404. A `# TODO` to move to `apache/iceberg-rest-fixture:1.11.0` once it's released, plus a tracking issue, would close out the last of the pre-release thread. Not a blocker.

catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+func (s *RestIntegrationSuite) ensureNamespaceIn(cat *rest.Catalog) {
+	s.T().Helper()
+	exists, err := cat.CheckNamespaceExists(s.ctx, catalog.ToIdentifier(TestNamespaceIdent))
+	s.Require().NoError(err)
+	if exists {
+		s.Require().NoError(cat.DropNamespace(s.ctx, catalog.ToIdentifier(TestNamespaceIdent)))
+	}
+
+	s.NoError(cat.CreateNamespace(s.ctx, catalog.ToIdentifier(TestNamespaceIdent),
```
laskoviymishka, line 217: Small robustness nit for a follow-up: `s.NoError` here is non-fatal, so if `CreateNamespace` fails the run keeps going into `createScanPlanningTable` and dies on a namespace-not-found from `CreateTable`, pointing at the wrong line. `s.Require().NoError` would report the setup failure directly.

catalog/rest/scan_planning_integration_test.go
```diff
@@ -0,0 +1,354 @@
...
+	arrowSchema, err := table.SchemaToArrowSchema(tableSchemaSimple, nil, false, false)
+	s.Require().NoError(err)
+	arrowTable, err := array.TableFromJSON(memory.DefaultAllocator, arrowSchema,
+		[]string{`[{"foo":"hello","bar":17,"baz":true}]`})
+	s.Require().NoError(err)
+	defer arrowTable.Release()
+
+	dataPath, err := url.JoinPath(tbl.Location(), "data", "bar=17", "data.parquet")
+	s.Require().NoError(err)
+	file, err := mustFS(s.T(), tbl).(iceio.WriteFileIO).Create(dataPath)
```
laskoviymishka, line 245: Same class, also fine as a follow-up: the bare `.(iceio.WriteFileIO)` panics the whole test binary with an interface-conversion error if the FS doesn't implement it, and against a live Docker FileIO a misconfigured endpoint can hand back a read-only impl. Comma-ok with `s.Require().True(ok, "table filesystem must implement WriteFileIO")` fails it as a test instead.

laskoviymishka (approved):
This came together nicely, LGTM. Everything I flagged last round landed: the dedicated `rest-scan-planning` service on 8182 with its own target, the shared `rest` / `integration-rest` path back on the released image, `ElementsMatch` fatal and ahead of the bound lookups, and the `DefaultTransport` assertion replaced with a concrete transport. Thanks for the quick turnaround.

Nothing here blocks. I left a few notes inline, all non-blocking and fine as follow-ups: keeping the new `integration-rest-scan-planning` target opt-in out of the umbrella `integration-test` chain, a `# TODO` plus tracking issue to move off the `latest@sha256` pre-release pin once 1.11.0 ships, a healthcheck on the new service so it doesn't flake on slow CI, and two small test-robustness tidies (a non-fatal `s.NoError` on `CreateNamespace`, and an unguarded `.(iceio.WriteFileIO)` that would panic the binary). Happy to see any of them in a later PR.

laskoviymishka:
@Revanth14  - to merge this need to resolve conflict ;)

Revanth14:
> @Revanth14 - to merge this need to resolve conflict ;)

done and resolved.
Thank you

laskoviymishka:
Nice!

PR #1506, opened 2026-07-22
fix(io/s3): let explicit s3.* credentials override the context AWS config

iremcaginyurtturk:
## Problem

`createS3Bucket` (`io/gocloud/s3.go`) resolves the AWS config by preferring an ambient config injected into the context via `utils.WithAwsConfig` — for example the Glue catalog's own AWS config — and only falls back to `ParseAWSConfig(props)` when no context config exists. The **endpoint** is always read from the properties (`s3.endpoint`), but the **credentials** are whatever the context config carries.

So when a catalog whose API credentials are AWS (Glue) writes to a bucket reached via a custom `s3.endpoint` whose credentials differ — a local MinIO, or a cross-account / delegated-role S3 target — the metadata read-back authenticates with the catalog's AWS credentials instead of the explicitly supplied `s3.access-key-id`, and fails:

```
S3: GetObject ... api error InvalidAccessKeyId: The Access Key Id you provided does not exist in our records.
```

`s3.access-key-id` / `s3.secret-access-key` / `s3.session-token` are silently ignored whenever a context config is present (the endpoint is honoured, so the request reaches the right store — with the wrong credentials).

## Fix

Extract the resolution into `resolveS3AWSConfig`. When explicit `s3.*` credentials are present, layer a static credentials provider built from them over the resolved config — on a copy, so the shared context config is not mutated. Precedence becomes **explicit props > context config > default chain**, matching how `s3.endpoint` already behaves. The common case (Glue + AWS S3, where the storage creds *are* the catalog creds) is unchanged.

### DIFF
--- a/io/gocloud/s3.go
+++ b/io/gocloud/s3.go
@@ -54,7 +54,8 @@ func ParseAWSConfig(ctx context.Context, props map[string]string) (*aws.Config, 
 
 	if tok, ok := props["token"]; ok {
 		opts = append(opts, config.WithBearerAuthTokenProvider(
-			&bearer.StaticTokenProvider{Token: bearer.Token{Value: tok}}))
+			&bearer.StaticTokenProvider{Token: bearer.Token{Value: tok}},
+		))
 	}
 
 	if region, ok := props[io.S3Region]; ok {
@@ -67,7 +68,8 @@ func ParseAWSConfig(ctx context.Context, props map[string]string) (*aws.Config, 
 	token := props[io.S3SessionToken]
 	if accessKey != "" || secretAccessKey != "" || token != "" {
 		opts = append(opts, config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(
-			props[io.S3AccessKeyID], props[io.S3SecretAccessKey], props[io.S3SessionToken])))
+			props[io.S3AccessKeyID], props[io.S3SecretAccessKey], props[io.S3SessionToken],
+		)))
 	}
 
 	if proxy, ok := props[io.S3ProxyURI]; ok {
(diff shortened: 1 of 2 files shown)


io/gocloud/s3.go
```diff
@@ -167,18 +169,35 @@ func resolveUsePathStyle(endpoint string, props map[string]string) bool {
...
-		if err != nil {
-			return nil, err
-		}
+	} else if awscfg, err = ParseAWSConfig(ctx, props); err != nil {
+		return nil, err
+	}
+
+	// Copy before overriding so the shared context config stays untouched.
+	if props[io.S3AccessKeyID] != "" || props[io.S3SecretAccessKey] != "" {
+		cfg := *awscfg
```
tanmayrauth: since you're introducing the copy-to-avoid-mutation pattern here, you could hoist it so it also covers the `awscfg.HTTPClient = newS3BuildableClient()` assignment in createS3Bucket. Today that copy only happens on the props-creds branch, so in the no-props branch the HTTPClient assignment still mutates
  the shared context config (pre-existing, and only when it's nil, so low impact) — hoisting the copy would fully realize the "don't mutate the shared config" intent. Not a blocker.

tanmayrauth (commented):
LGTM — the credential override is correct, done on a copy so the shared context config isn't mutated, and the precedence (explicit props > context config > default chain) matches how the endpoint already behaves. Tests cover the override, no-override, and no-context cases. One optional nice-to-have inline; nothing blocking.

io/gocloud/s3.go
```diff
@@ -167,18 +169,35 @@ func resolveUsePathStyle(endpoint string, props map[string]string) bool {
...
-		}
+		base = v
+	} else if base, err = ParseAWSConfig(ctx, props); err != nil {
+		return nil, err
+	}
+
+	// Always copy so neither the credential override nor the caller's later
+	// mutations (e.g. HTTPClient) touch a shared context config.
+	cfg := *base
+	if props[io.S3AccessKeyID] != "" || props[io.S3SecretAccessKey] != "" {
```
zeroshade: Partial credential override clobbers unset fields: setting only `s3.access-key-id` installs a static provider with an empty secret; `s3.session-token`-only is ignored (not in the condition). Unset credential keys should preserve the ambient/context credentials.

io/gocloud/s3.go
```diff
@@ -167,18 +169,35 @@ func resolveUsePathStyle(endpoint string, props map[string]string) bool {
...
+// resolveS3AWSConfig returns the AWS config for the S3 FileIO, preferring an
+// ambient context config but letting explicit s3.* credentials override it.
+func resolveS3AWSConfig(ctx context.Context, props map[string]string) (*aws.Config, error) {
 	var (
-		awscfg *aws.Config
-		err    error
+		base *aws.Config
+		err  error
 	)
 	if v := utils.GetAwsConfig(ctx); v != nil {
```
zeroshade, line 182: When a context AWS config is present, `ParseAWSConfig` is skipped, so explicit `s3.region`/`client.region` and other parsed config props don't override the context config. Endpoint/path-style are applied later, but region precedence is wrong for context-backed setups.

zeroshade (changes requested):
The copy-before-mutate fix is good and the full access-key + secret override now works without mutating the context config. But partial overrides clobber unset fields, and explicit region/config props are skipped when a context AWS config is present, so “explicit key wins, unset key falls through” isn't fully satisfied. CI green.

iremcaginyurtturk:
Thanks, addressed the reviews in 9a8595b

zeroshade (approved):
Both issues from the last pass are resolved, and the new `resolveS3AWSConfig` reads cleanly:

- **Region precedence:** explicit `s3.region`/`s3.client.region` now override the region even when a context config is present (applied post-copy, independent of source) → `explicit region overrides ctx config`.
- **Partial credentials:** a static provider is installed only when access-key *and* secret are both set; a partial set falls through to the context/default chain instead of clobbering with a blank secret → `partial props creds fall through`.
- Config is always copied before mutation, and the `never mutates the shared ctx config` subtest guards the `HTTPClient` assignment.

LGTM. Optional/non-blocking: `cfg := *base` is a shallow copy — safe today since nothing appends to `APIOptions`/`ConfigSources`, but `base.Copy()` would isolate those slices too and matches the "never touch the shared config" intent.

PR #1507, opened 2026-07-22
fix(hive): derive namespace location from the warehouse

iremcaginyurtturk:
## Problem

`CreateNamespace` (`catalog/hive/hive.go`) sets the Hive database `LocationUri` only from an explicit `location` / `Location` property. When none is provided — the common case (e.g. a client calls `CreateNamespace` with empty properties) — and the metastore has no default warehouse dir configured, the metastore is handed an empty location and rejects it:

```
MetaException: java.lang.IllegalArgumentException: Can not create a Path from an empty string
```

The catalog already carries the `warehouse` option, but `CreateNamespace` never uses it for the namespace location. (pyiceberg's Hive catalog derives the database location from the warehouse.)

## Fix

When no explicit location is given, derive `<warehouse>/<db>.db` — the metastore's conventional database location — from the catalog's `warehouse`. When no warehouse is configured the location stays empty, preserving the previous behavior (no regression for setups that rely on the metastore's own default warehouse dir).

### DIFF
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -777,16 +777,6 @@ func (c *Catalog) ListNamespaces(ctx context.Context, parent table.Identifier) (
 }
 
 // CreateNamespace creates a new namespace in the catalog.
-// defaultNamespaceLocation derives <warehouse>/<db>.db, the metastore's
-// conventional database location, or "" when no warehouse is configured.
-func defaultNamespaceLocation(warehouse, database string) string {
-	if warehouse == "" {
-		return ""
-	}
-
-	return strings.TrimRight(warehouse, "/") + "/" + database + ".db"
-}
-
 func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifier, props iceberg.Properties) error {
 	database, err := identifierToDatabase(namespace)
 	if err != nil {
@@ -824,6 +814,16 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 	}
 
 	return nil
+}
+
+// defaultNamespaceLocation derives <warehouse>/<db>.db, the metastore's
+// conventional database location, or "" when no warehouse is configured.
+func defaultNamespaceLocation(warehouse, database string) string {
+	if warehouse == "" {
+		return ""
+	}
+
+	return strings.TrimSuffix(warehouse, "/") + "/" + database + ".db"
 }
 
 // DropNamespace drops a namespace from the catalog.
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -368,7 +368,8 @@ func (c *Catalog) CreateView(ctx context.Context, identifier table.Identifier, v
 	}
 
 	createdView, err := view.CreateViewWithIOProperties(
-		ctx, catalogName, identifier, freshSchema, viewSQL, defaultNS, loc, ioProps, cfg.Properties)
+		ctx, catalogName, identifier, freshSchema, viewSQL, defaultNS, loc, ioProps, cfg.Properties,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -776,6 +777,16 @@ func (c *Catalog) ListNamespaces(ctx context.Context, parent table.Identifier) (
 }
 
 // CreateNamespace creates a new namespace in the catalog.
+// defaultNamespaceLocation derives <warehouse>/<db>.db, the metastore's
+// conventional database location, or "" when no warehouse is configured.
+func defaultNamespaceLocation(warehouse, database string) string {
+	if warehouse == "" {
+		return ""
+	}
+
+	return strings.TrimRight(warehouse, "/") + "/" + database + ".db"
+}
+
 func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifier, props iceberg.Properties) error {
 	database, err := identifierToDatabase(namespace)
 	if err != nil {
@@ -796,6 +807,12 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 		default:
 			db.Parameters[k] = v
 		}
+	}
+
+	// Derive a location from the warehouse when none is given, so the metastore
+	// is not handed an empty path (which it rejects).
+	if db.LocationUri == "" {
+		db.LocationUri = defaultNamespaceLocation(c.opts.Warehouse, database)
 	}
 
 	if err := c.client.CreateDatabase(ctx, db); err != nil {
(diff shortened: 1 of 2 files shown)


catalog/hive/hive.go
```diff
@@ -776,6 +777,16 @@ func (c *Catalog) ListNamespaces(ctx context.Context, parent table.Identifier) (
 }
 
 // CreateNamespace creates a new namespace in the catalog.
+// defaultNamespaceLocation derives <warehouse>/<db>.db, the metastore's
```
tanmayrauth: The helper landed between this doc comment and the function, so `// CreateNamespace creates a new namespace...` now documents `defaultNamespaceLocation` and `CreateNamespace` is left undocumented — `go doc` and linters will render it on the wrong symbol. Move  `defaultNamespaceLocation` below `CreateNamespace`, or move this comment line down so it sits directly above `func (c *Catalog) CreateNamespace`.

catalog/hive/hive_test.go
```diff
@@ -399,6 +399,28 @@ func TestHiveCreateNamespace(t *testing.T) {
 	mockClient.AssertExpectations(t)
 }
 
+func TestDefaultNamespaceLocation(t *testing.T) {
+	require.Equal(t, "s3://warehouse/db.db", defaultNamespaceLocation("s3://warehouse", "db"))
+	require.Equal(t, "s3://warehouse/db.db", defaultNamespaceLocation("s3://warehouse/", "db"))
+	require.Empty(t, defaultNamespaceLocation("", "db"))
+}
+
+func TestHiveCreateNamespaceDerivesLocationFromWarehouse(t *testing.T) {
```
tanmayrauth, line 422: Nice to also cover precedence: warehouse configured *and* an explicit `location` prop, asserting the explicit value wins. Today the only explicit-location test runs on a catalog with no warehouse, so the `if db.LocationUri == ""` branch is never exercised with a non-empty warehouse — a refactor that moved the
  derivation ahead of the props loop would override a user's explicit location and no test would catch it.

tanmayrauth (commented):
Fix looks correct. Just a few nits, details inline.

catalog/hive/hive.go
laskoviymishka: There's no blank line between the two comment groups, so godoc attaches the whole block to `defaultNamespaceLocation` — which means `CreateNamespace` below loses its doc and the unexported helper ends up documented as "creates a new namespace in the catalog."

I'd move the `// CreateNamespace ...` line back down to sit directly above `func CreateNamespace`, and give the helper its own standalone comment.

catalog/hive/hive.go
```diff
@@ -776,6 +777,16 @@ func (c *Catalog) ListNamespaces(ctx context.Context, parent table.Identifier) (
...
+// conventional database location, or "" when no warehouse is configured.
+func defaultNamespaceLocation(warehouse, database string) string {
+	if warehouse == "" {
+		return ""
+	}
+
+	return strings.TrimRight(warehouse, "/") + "/" + database + ".db"
```
laskoviymishka: `TrimRight(warehouse, "/")` treats the arg as a cutset, so it strips every trailing slash rather than one. Mostly harmless, but two edges leak through: `warehouse == "/"` trims to `""` and we hand the metastore `/db.db`, and `file:///` trims to `file:` so we'd derive `file:/db.db` where Java's stripTrailingSlash stops at the scheme and keeps `file:///db.db`.

`strings.TrimSuffix(warehouse, "/")` sidesteps both and matches what hive.go:340 already does for the table location. I know hadoop.go uses `TrimRight` too, so this isn't a new convention break — just a sharper edge than that path hits. wdyt?

catalog/hive/hive_test.go
```diff
@@ -399,6 +399,28 @@ func TestHiveCreateNamespace(t *testing.T) {
 	mockClient.AssertExpectations(t)
 }
 
+func TestDefaultNamespaceLocation(t *testing.T) {
```
laskoviymishka, line 402: Could we make this table-driven with `t.Run`? Right now it's three bare `require.Equal`s, so a failure only points at the function, not the case — and it's the one test in the file not using the `require.New(t)` style.

While we're restructuring, a `file:///` case and a double-trailing-slash input would be worth pinning down given the `TrimRight` thread above, so the behavior is locked in whichever way we land it.

catalog/hive/hive_test.go
laskoviymishka, line 422: This covers the happy path nicely. Two guard cases aren't exercised though: no warehouse + no location prop (should preserve empty `LocationUri` — the no-regression case), and warehouse set + explicit location prop (explicit should still win). Both are easy to invert accidentally, so I'd add them alongside this one.

catalog/hive/hive.go
```diff
@@ -798,6 +809,12 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 		}
 	}
 
+	// Derive a location from the warehouse when none is given, so the metastore
+	// is not handed an empty path (which it rejects).
+	if db.LocationUri == "" {
```
laskoviymishka, line 804: Small one: if a caller passes `location: ""` explicitly, the props loop sets it empty and this guard then overwrites it with the derived path, so "explicit empty" and "absent" become indistinguishable. Probably fine for the metastore's sake, but if we want to preserve intent we'd have to track whether the key was present. Non-blocking.

laskoviymishka (changes requested):
Nice catch on the empty-`LocationUri` rejection: deriving `<warehouse>/<db>.db` is the right fix, and keeping the `warehouse == ""` guard so unconfigured catalogs still pass through empty is exactly the behavior we want. Agreed with tanmayrauth that the logic is correct.

One thing before merge: the doc comment for `defaultNamespaceLocation` runs straight into `CreateNamespace`'s with no blank line, so godoc hands the whole block to the helper and `CreateNamespace` loses its doc. A blank line, plus moving the `// CreateNamespace` line back above its func, fixes it.

The only other thing I'd actually look at is the `TrimRight(warehouse, "/")` cutset — it strips every trailing slash, which turns `file:///` into `file:` (Java keeps the scheme) and `/` into empty. `TrimSuffix` handles both and matches what we already do a few lines down. Not strictly blocking, but worth a decision.

Everything else is a nit — left them inline. Once the doc comment's sorted, happy to take another pass and approve.

catalog/hive/hive.go
iremcaginyurtturk: Agreed. Switched to strings.TrimSuffix(warehouse, "/") so it trims a single trailing slash and preserves the scheme (file:/// → file:///db.db), matching what hive.go:340 does for the table location.

laskoviymishka (approved):
LGTM now!

Thanks for the contribution!

laskoviymishka:
@iremcaginyurtturk welcome to contributors gang!

iremcaginyurtturk:
@laskoviymishka  thank you 😄

PR #1508, opened 2026-07-22
fix(transforms): validate parameterized transform values

fallintoplace:
## Summary

- validate directly constructed bucket counts and truncate widths before serialization or execution
- reject parameter values outside the parser-supported range of 1 through `math.MaxInt32`
- validate partition specs and sort orders eagerly when they are constructed

## Why

`ParseTransform` rejects invalid parameter values, but callers can construct the exported transform structs directly. A zero truncate width reaches integer modulo and panics, while values such as `1 << 32` narrow to zero in the int32 paths and can panic for both bucket and truncate transforms.

Partition specs and sort orders now reject these invalid transforms during construction instead of deferring the failure until metadata serialization.

## Testing

- `go test ./...`
- `go test -race ./codec/... ./table/...`
- focused transform, partition-spec, and sort-order regression tests

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -294,6 +294,10 @@ func validateTransform(transform Transform) error {
 		return t.validateNumBuckets()
 	case *BucketTransform:
 		return t.validateNumBuckets()
+	case TruncateTransform:
+		return t.validateWidth()
+	case *TruncateTransform:
+		return t.validateWidth()
 	default:
 		return nil
 	}
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -183,6 +183,21 @@ func TestPartitionSpecRejectsNegativeSpecID(t *testing.T) {
 	require.ErrorContains(t, err, "spec id must be non-negative: -1")
 }
 
+func TestPartitionSpecRejectsInvalidTruncateTransform(t *testing.T) {
+	schema := iceberg.NewSchema(1, iceberg.NestedField{
+		ID:   1,
+		Name: "id",
+		Type: iceberg.PrimitiveTypes.Int32,
+	})
+
+	_, err := iceberg.NewPartitionSpecOpts(
+		iceberg.AddPartitionFieldBySourceID(1, "id_truncate", iceberg.TruncateTransform{Width: 0}, schema, nil),
+	)
+
+	require.ErrorIs(t, err, iceberg.ErrInvalidArgument)
+	require.ErrorContains(t, err, "width > 0")
+}
+
 func TestPartitionSpec_MarshalTextRejectsInvalidBucketTransform(t *testing.T) {
 	spec := iceberg.NewPartitionSpecID(3,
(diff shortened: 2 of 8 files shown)


partitions.go
```diff
@@ -291,6 +291,10 @@ func validateTransform(transform Transform) error {
 		return t.validateNumBuckets()
 	case *BucketTransform:
 		return t.validateNumBuckets()
+	case TruncateTransform:
```
tanmayrauth, line 297: Partition specs now validate transform bounds eagerly here, but sort orders don't get the same treatment - `newSortOrder` in table/sorting.go only checks `Transform == nil` and `CanTransform`, not the numeric bounds. So a `SortField` with `TruncateTransform{Width: 0}` is accepted into an in-memory SortOrder and  only fails later when the metadata is marshalled through the `MarshalText` backstop. It can't corrupt metadata, but it's a confusing late failure and an inconsistency with the partition path this PR tightens. Adding the same `validateWidth`/`validateNumBuckets` check in `newSortOrder` would make both paths fail  eagerly at construction.

tanmayrauth (commented):
Nice PR, one non-blocking consistency note inline.

zeroshade (changes requested):
The parse / partition-spec / sort-order / Apply paths validate `bucket[N]`/`truncate[W]` nicely. One construction path is still unguarded: `UpdateSpec.AddField` reparses the transform and drops the error, so an explicit `BucketTransform{0}`/`TruncateTransform{0}` becomes a nil transform instead of a clear error. Please validate there and add a test. CI green.

---
- `table/update_spec.go:378` — `transform, _ := iceberg.ParseTransform(key.Transform)` ignores the parse error. Since `addField` only checks `CanTransform` (which doesn't validate NumBuckets/Width), `AddField(BucketTransform{NumBuckets:0})` stringifies to `bucket[0]`, fails to parse, and proceeds with a nil transform. Propagate the error (e.g. `ErrInvalidArgument`).

zeroshade (approved):
The unguarded construction path I flagged is closed, plus you extended validation further than I asked:

- `UpdateSpec.partitionField` now propagates the `ParseTransform` error (`ErrInvalidArgument: invalid partition transform`) instead of dropping it into a nil transform → covered by the `AddField` bucket/truncate `[0]` subtests.
- `newSortOrder` validates parameterized transforms via the `TextMarshaler` backstop, closing @tanmayrauth's sort-order gap → `TestNewSortOrderRejectsInvalidParameterizedTransform`.
- Bonus: `TruncateTransform.validateWidth` added and wired through `MarshalText`/`Transformer`/`Project`/`validateTransform` for parity with bucket, and both transforms now reject `> MaxInt32` (with a 32-bit-safe `overflowingInt32TransformParameter` helper).

Solid, comprehensive coverage. LGTM.

PR #1509, opened 2026-07-22
fix(manifest): encode decimal partitions with declared precision

### LINKED ISSUE #1027: Bug(manifest): convertDecimalValue uses string length instead of declared precision for Avro fixed-size encoding

tanmayrauth:
Size the Avro fixed[N] buffer from the column's declared precision instead of len(dec.String()), which produced wrong-width buffers and made the encoder reject decimal partition values.

getFieldIDMap now returns a combined scale+precision map keyed by field id, threading precision through the write path and keeping scale on the read path.

Fixes #1027

### DIFF
--- a/data_file_codec.go
+++ b/data_file_codec.go
@@ -107,7 +107,7 @@ func (d *dataFile) MarshalAvroEntry(spec PartitionSpec, schema *Schema, version 
 		return nil, err
 	}
 	clone := cloneDataFileAvroFields(d)
-	clone.PartitionData = avroEncodePartitionData(d.Partition(), maps.nameToID, maps.idToType)
+	clone.PartitionData = avroEncodePartitionData(d.Partition(), maps.nameToID, maps.idToType, maps.idToDecimal)
 
 	return s.Encode(newEncodeEntry(version, clone))
 }
@@ -139,7 +139,7 @@ func unmarshalAvroDataFileEntry(data []byte, spec PartitionSpec, schema *Schema,
 	df.specID = int32(spec.ID())
 	df.fieldNameToID = maps.nameToID
 	df.fieldIDToLogicalType = maps.idToType
-	df.fieldIDToFixedSize = maps.idToFixedSize
+	df.fieldIDToDecimal = maps.idToDecimal
 
 	return df, nil
 }
@@ -205,8 +205,8 @@ func cloneDataFileAvroFields(src *dataFile) *dataFile {
 // iceberg-typed values like Date or Decimal) into the name-keyed
 // avro-friendly map the manifest-entry schema expects. Idempotent:
 // values already in primitive form pass through unchanged.
-func avroEncodePartitionData(idKeyed map[int]any, nameToID map[string]int, logicalTypes map[int]string) map[string]any {
-	converted := avroPartitionData(idKeyed, logicalTypes)
+func avroEncodePartitionData(idKeyed map[int]any, nameToID map[string]int, logicalTypes map[int]string, decimals map[int]decimalMeta) map[string]any {
+	converted := avroPartitionData(idKeyed, logicalTypes, decimals)
 	out := make(map[string]any, len(converted))
 	for name, id := range nameToID {
 		if v, ok := converted[id]; ok {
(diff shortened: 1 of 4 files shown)


tanmayrauth:

zeroshade (changes requested):
Heads up: this overlaps directly with #1355 (same functions, same bug). It correctly sources width from the declared precision, but still zero-pads negatives and silently truncates oversized values — the two correctness bugs #1355 already fixes (two's-complement sign extension + error-on-overflow). Suggest closing this in favor of #1355, or adopting its byte-fitting/error path. CI green.

---
- `manifest.go:1951` — `padOrTruncateBytes` pads short decimal slices with `0x00`. Negative decimals need two's-complement sign extension with `0xff` (e.g. unscaled -1 for decimal(10,2) must be `ff ff ff ff ff`, not `00 00 00 00 ff`). See #1355.
- `manifest.go:1952` — Oversized slices are truncated with `bytes[len-size:]` and no error. Values exceeding the declared fixed size must be rejected, not silently rewritten to a different value. The current `avroEncodePartitionData`/`convertDecimalValue` signatures can't propagate that error — cf. #1355's approach.

PR #1510, opened 2026-07-22
fix(table): cross-check deletion vector cardinality against manifest record_count

### LINKED ISSUE #1058: feat(table/dv): cross-validate manifest record_count against blob cardinality property

tanmayrauth:
ReadDV now validates the decoded bitmap against the manifest entry's record_count (field 103) and cross-checks it with the puffin blob's cardinality property when present. record_count is a required non-nullable long, so it always participates: a manifest zero against a nonzero puffin property is a real disagreement, not an absent value, and fails fast.

Blobs missing the puffin property are still validated against record_count and accepted with a warning. Adds a hand-built fixture to exercise the missing-property branch, which the puffin writer otherwise forbids.

Closes #1058

### DIFF
--- a/table/dv/deletion_vector.go
+++ b/table/dv/deletion_vector.go
@@ -132,22 +132,24 @@ func SerializeDV(bitmap *RoaringPositionBitmap) ([]byte, error) {
 // ReadDV reads a deletion vector from a puffin file using the manifest entry metadata.
 // ContentOffset and ContentSizeInBytes must be set on the DataFile (required by v3 spec).
 //
-// When the puffin blob carries a `cardinality` property (spec-mandated for
-// deletion-vector-v1), its value is parsed and used to validate the decoded
-// bitmap — this catches truncated or partially-overwritten blobs whose CRC
-// still validates over the bytes that are present.
+// The decoded bitmap's cardinality is cross-validated against two independent
+// sources, so a truncated or partially-overwritten blob whose CRC still
+// validates over the bytes that are present is rejected:
 //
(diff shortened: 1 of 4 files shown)


tanmayrauth:

zeroshade (approved):
LGTM. Cross-checks the DV manifest record_count against the Puffin blob cardinality and the decoded bitmap count, tolerating an absent property for reader compatibility. Aligns with the V3 DV/Puffin spec. CI green.

PR #1511, opened 2026-07-22
feat(table): aggregate geo bounds across data files at manifest level

### LINKED ISSUE #993: feat(manifest): aggregate geo bounds at manifest level
Parent: #989

Depends on #992. The manifest writer aggregates min/max per column from each data file's column stats; geo needs the same aggregation but the typed wrapper is a 4D bbox rather than a scalar. Plumb a `GeoBounds` value through `LowerBounds` / `UpperBounds` so the existing `BoundsEvaluator` sees geo bounds via the same interface it already speaks for primitives.

Tests round-trip a manifest containing geo data files and verify bounds aggregate correctly across multiple files in the same partition.

tanmayrauth:
Add GeoBoundsAggregator to combine per-file geometry bounds into one bounding box via per-dimension min/max, never a scalar byte compare, so the unorderable geo bounds stay honest. decodeGeoBound inverts the single-value serialization; the same omit-on-ambiguity rule for Z/M applies across files, and geography files contribute nothing.

Wiring the aggregate into scan-time pruning lands in a follow-up PR.

Fix: #993

### DIFF
--- a/table/internal/geo_codec.go
+++ b/table/internal/geo_codec.go
@@ -19,6 +19,7 @@ package internal
 
 import (
 	"encoding/binary"
+	"fmt"
 	"math"
 
 	"github.com/apache/iceberg-go"
@@ -160,15 +161,22 @@ func (a *geoBoundsAccumulator) layout() (geom.Layout, bool) {
 	hasZ := a.has[geoDimZ] && a.zGeoms == a.geoms
 	hasM := a.has[geoDimM] && a.mGeoms == a.geoms
 
+	return geoPointLayout(hasZ, hasM), true
+}
+
+// geoPointLayout selects the bound point layout implied by which optional
+// dimensions are present. X and Y are always assumed present; callers gate on
+// that before calling.
+func geoPointLayout(hasZ, hasM bool) geom.Layout {
 	switch {
 	case hasZ && hasM:
-		return geom.XYZM, true
+		return geom.XYZM
 	case hasZ:
-		return geom.XYZ, true
+		return geom.XYZ
 	case hasM:
-		return geom.XYM, true
+		return geom.XYM
 	default:
-		return geom.XY, true
+		return geom.XY
 	}
 }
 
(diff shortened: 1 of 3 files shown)


tanmayrauth:

table/internal/geo_codec.go
```diff
@@ -266,3 +274,166 @@ func (g *geoStatsAgg) Update(interface{ HasMinMax() bool }) {}
...
+		return fmt.Errorf("%w: geo upper bound must be 16, 24, or 32 bytes, got %d",
+			iceberg.ErrInvalidBinSerialization, len(upper))
+	}
+	if loLayout != hiLayout {
+		return fmt.Errorf("%w: geo lower/upper bounds have mismatched layouts (%v vs %v)",
+			iceberg.ErrInvalidBinSerialization, loLayout, hiLayout)
+	}
+
+	g.n++
+	g.update(geoDimX, lo[geoDimX], hi[geoDimX])
```
zeroshade, line 420: `GeoBoundsAggregator.Add` always merges X with scalar min/max. Iceberg geography allows `lower_x > upper_x` to denote an antimeridian-crossing box; merging a wrapped file `[170,-170]` with `[10,20]` yields `[10,20]`, dropping the wrapped range. This exported aggregator has no type input to distinguish geometry vs geography — either refuse geography or implement interval-union for longitude.

zeroshade (changes requested):
Geometry XY/XYZ/XYM min/max aggregation and the V3 single-value bound encoding look correct. Blocking issue: the aggregator is type-blind and mis-merges geography bounds that cross the antimeridian (for geography, `lower_x > upper_x` is valid/meaningful). Either make it geometry-only by API and reject geography, or implement longitude interval-union semantics. CI green.

tanmayrauth:
@zeroshade  Good catch. Went geometry-only rather than interval-union, since it matches this PR — the per-file accumulator already omits geography bounds, so there's nothing well-formed to aggregate yet.

  Added NewGeoBoundsAggregator(isGeography bool): Add now rejects non-empty geography bounds with ErrNotImplemented, while empty bounds (what iceberg-go emits for geography) stay a no-op. Docs + tests updated.

  Happy to switch to longitude interval-union if you'd rather geography aggregate correctly now.

tanmayrauth:
@zeroshade  can you please review it when you get chance?

zeroshade (approved):
The blocking antimeridian issue is resolved. Going geometry-only (`NewGeoBoundsAggregator(isGeography)`) is the right scope call here — since the per-file accumulator doesn't emit geography bounds yet, there's nothing well-formed to fold, and refusing non-empty geography bounds with a clear `ErrNotImplemented` (rather than silently unwrapping a `lower_x > upper_x` box) is exactly right. The doc comments citing spec Appendix D and `TestGeoBoundsAggregatorRejectsGeography` nail it down.

Agreed on deferring longitude interval-union to a follow-up for when geography bounds are actually produced. LGTM.

PR #1512, opened 2026-07-23
fix(table): validate snapshot references

fallintoplace:
## Summary

- require `snapshot-id` and `type` when decoding snapshot references
- reject non-positive retention values
- reject branch-only retention settings on tags during decoding and metadata updates

## Why

The Java `SnapshotRef` builder and parser enforce these invariants, but Go only checked the reference type while decoding. Programmatic metadata updates also allowed tag references with `min-snapshots-to-keep` or `max-snapshot-age-ms`, producing metadata Java rejects.

## Testing

- `go test ./...`
- `go test -race ./table`
- `go vet ./...`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -908,6 +908,9 @@ func (b *MetadataBuilder) SetSnapshotRef(
 			return fmt.Errorf("%w: invalid snapshot ref option: %w", iceberg.ErrInvalidArgument, err)
 		}
 	}
+	if err := ref.validate(); err != nil {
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
LGTM. Validates snapshot refs on both the parse and builder paths: dangling snapshot-id, invalid ref type/retention, and main vs current-snapshot-id mismatches are rejected, and no-ref metadata still parses. Good coverage; CI green.

PR #1513, opened 2026-07-23
fix(table): tolerate missing snapshot summary operation

fallintoplace:
## Summary

- default a non-empty snapshot summary without `operation` to `overwrite`
- preserve the remaining summary properties

## Why

Java uses this fallback for compatibility with older snapshot metadata. Go returned `ErrMissingOperation` instead, so it could not load metadata that Java accepts.

## Testing

- `go test ./...`
- `go test -race ./table`
- `go vet ./...`

### DIFF
--- a/table/snapshots.go
+++ b/table/snapshots.go
@@ -222,7 +222,15 @@ func (s *Summary) UnmarshalJSON(b []byte) (err error) {
 
 	op, ok := alias[operationKey]
 	if !ok {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Tolerates snapshots that omit the summary operation: non-empty summaries default to overwrite, empty summaries stay empty, and invalid explicit operations still error. CI green.

PR #1514, opened 2026-07-23
fix(partitions): assign missing partition field IDs

fallintoplace:
## Summary

- assign missing partition field IDs after the greatest of 999, `last-partition-id`, or any explicit field ID
- preserve explicit IDs and assign only the missing fields
- allocate IDs across all specs in table metadata so generated IDs cannot collide between specs
- update `last-partition-id` when it is present and reject explicit null field IDs

## Why

Older partition spec JSON may omit field IDs. Java assigns those IDs from `PARTITION_DATA_ID_START`, while Go decoded every missing ID as zero. That loses field identity and can make partition fields collide.

The metadata parser needs table-wide context because partition field IDs are shared across specs. Standalone partition specs use the same allocation rule within the spec.

## Testing

- `go test -count=1 . ./table`
- full CI

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -425,15 +425,38 @@ func (ps PartitionSpec) MarshalJSON() ([]byte, error) {
 
 func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
 	aux := struct {
-		ID     int              `json:"spec-id"`
-		Fields []PartitionField `json:"fields"`
-	}{ID: ps.id, Fields: ps.fields}
+		ID     int               `json:"spec-id"`
+		Fields []json.RawMessage `json:"fields"`
+	}{ID: ps.id}
 
 	if err := json.Unmarshal(b, &aux); err != nil {
 		return err
 	}
 
-	ps.id, ps.fields = aux.ID, aux.Fields
+	fields := make([]PartitionField, len(aux.Fields))
+	for i, rawField := range aux.Fields {
+		var keys map[string]json.RawMessage
+		if err := json.Unmarshal(rawField, &keys); err != nil {
+			return err
+		}
+		if rawFieldID, ok := keys["field-id"]; ok {
+			var fieldID *int
+			if err := json.Unmarshal(rawFieldID, &fieldID); err != nil {
+				return fmt.Errorf("%w: invalid partition field ID: %w", ErrInvalidPartitionSpec, err)
+			}
+			if fieldID == nil {
+				return fmt.Errorf("%w: partition field ID cannot be null", ErrInvalidPartitionSpec)
+			}
+		}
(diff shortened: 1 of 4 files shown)


partitions.go
```diff
@@ -422,15 +422,47 @@ func (ps PartitionSpec) MarshalJSON() ([]byte, error) {
...
+				return fmt.Errorf("%w: partition field ID cannot be null", ErrInvalidPartitionSpec)
+			}
+			fieldIDCount++
+		}
+		if err := json.Unmarshal(rawField, &fields[i]); err != nil {
+			return err
+		}
+	}
+
+	if fieldIDCount != 0 && fieldIDCount != len(fields) {
```
zeroshade: Mixed set/unset partition field ids are rejected here. Expected behavior is to preserve existing ids and assign fresh ids only to the missing fields.

partitions.go
```diff
@@ -422,15 +422,47 @@ func (ps PartitionSpec) MarshalJSON() ([]byte, error) {
...
+		return fmt.Errorf("%w: cannot parse spec with missing field IDs: %d missing of %d fields",
+			ErrInvalidPartitionSpec, len(fields)-fieldIDCount, len(fields))
+	}
+	if fieldIDCount == 0 {
```
zeroshade: All-missing assigns `PartitionDataIDStart + i` per spec. With no table-level `last-partition-id`/existing-id context, multiple specs parsed from the same metadata each assign 1000, 1001, … and collide across the table-wide partition-field-id space (also watch for collision with an existing field already using 1000/1001).

zeroshade (changes requested):
Base-1000 assignment for the all-missing single-spec case is correct, but two gaps remain: mixed set/unset fields are rejected instead of assigning ids only to the missing ones, and ids are allocated per-spec with no table-wide context, so multiple specs can collide on 1000/1001. CI green.

zeroshade (approved):
Both gaps are fixed, and moving the normalization up to `ParseMetadataBytes` (`assignMissingPartitionFieldIDs`) is the right call — it gives exactly the table-wide context that was missing:

- **Mixed set/unset:** existing ids are preserved and fresh ids are assigned only to fields without one → `TestDeserializePartitionSpecWithPartiallyMissingFieldIDs`, and the direct-unmarshal path assigns after the existing max (`…AssignsAfterExistingFieldIDs`).
- **Cross-spec collisions:** `lastAssignedID` seeds from `last-partition-id` and the max of every existing field id across all specs, then increments monotonically table-wide → `TestAssignMissingPartitionFieldIDsAcrossSpecs` (two specs → 1004, 1005, `last-partition-id` bumped).
- Nice extra: explicit `"field-id": null` is now a hard `ErrInvalidPartitionSpec` rather than silently treated as missing.

LGTM.

PR #1515, opened 2026-07-23
fix(table): avoid mutating sort fields during marshaling

fallintoplace:
## Summary

- compute default sort direction and null order in local variables
- preserve the existing JSON output without changing the `SortField` receiver
- use a value receiver so both `SortField` values and pointers use the custom encoder

## Why

`SortField.MarshalJSON` wrote default values back into the field being serialized. Marshaling should be read-only; mutating caller state is surprising and can race when a field is serialized concurrently. The pointer receiver also meant marshaling a `SortField` value bypassed the custom JSON shape entirely.

## Testing

- `go test ./...`
- `go test -race ./table`
- `go vet ./...`

### DIFF
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -98,16 +98,18 @@ func (s *SortField) String() string {
 	return fmt.Sprintf("%s(%d) %s %s", s.Transform, s.SourceID(), s.Direction, s.NullOrder)
 }
 
-func (s *SortField) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. SortField.MarshalJSON now uses a value receiver and local defaults, so marshaling no longer mutates the source Direction/NullOrder; emitted JSON is unchanged and covered by a regression test. CI green.

PR #1516, opened 2026-07-23
fix(table): make summary equality nil-safe

fallintoplace:
## Summary

- handle nil snapshot summaries symmetrically in `Summary.Equals`
- cover direct summary comparison and `Snapshot.Equals` in both operand orders

## Why

`Summary.Equals` handled a non-nil receiver with a nil argument, but not the reverse. Comparing a snapshot without a summary to one with a summary dereferenced the nil receiver and panicked. Equality should return false regardless of operand order.

## Testing

- `go test ./...`
- `go test -race ./table`
- `go vet ./...`

### DIFF
--- a/table/snapshots.go
+++ b/table/snapshots.go
@@ -199,7 +199,7 @@ func (s *Summary) Equals(other *Summary) bool {
 		return true
 	}
 
-	if s != nil && other == nil {
+	if s == nil || other == nil {
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Summary.Equals is now nil-safe (nil/nil true, nil vs non-nil false before deref); the Snapshot.Equals path benefits. Regression tests added. CI green.

PR #1517, opened 2026-07-23
fix(table): honor parquet row group byte target

fallintoplace:
## Summary

* honor `write.parquet.row-group-size-bytes` when writing buffered Parquet row groups
* start a new row group before the next non-empty batch after the byte target is reached
* avoid creating an empty trailing row group when the file is closed
* reject invalid and non-positive target values, matching Iceberg Java behavior
* preserve `write.parquet.row-group-limit` as an independent row-count ceiling

## Why

`write.parquet.row-group-size-bytes` is exported and documented, but the Go Parquet writer did not read it. Changing the property therefore had no effect on generated row groups.

Arrow Go v18.7 provides `RowGroupTotalBytesWritten` and `NewBufferedRowGroupChecked`, allowing the writer to apply the configured uncompressed byte target safely.

Rotation happens before the next non-empty batch instead of immediately after reaching the target. This prevents closing the writer from producing an empty trailing row group.

## Testing

* added coverage for the default and configured byte targets
* added coverage for malformed, zero, and negative values
* added a Parquet metadata regression test verifying that:

  * a small target creates multiple row groups
  * each row group contains data
  * no empty trailing row group is produced
* `git diff --check`

### DIFF
--- a/table/internal/interfaces.go
+++ b/table/internal/interfaces.go
@@ -120,6 +120,7 @@ type WriteFileInfo struct {
 	FileName         string
 	StatsCols        map[int]StatisticsCollector
 	WriteProps       any
+	RowGroupBytes    int64
 	Content          iceberg.ManifestEntryContent
 	EqualityFieldIDs []int
 	SortOrderID      int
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
LGTM. Parses write.parquet.row-group-size-bytes (128 MiB default, rejects non-positive) and wires it through both write paths as a buffered byte budget, kept distinct from the row-count limit. CI green.

PR #1518, opened 2026-07-23
fix(table): reject null position deletes

fallintoplace:
## Motivation

The Iceberg position delete schema requires `pos` values to be non-null. The Arrow scanner currently reads the raw `Int64Values` backing slice without checking the validity bitmap, so a malformed delete file containing a null position may be interpreted as a real row position instead of returning an invalid-schema error.

## Changes

- Reject null values in position delete `pos` chunks with `iceberg.ErrInvalidSchema`.
- Add a regression test covering a nullable position array and verifying the error contract.

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -451,6 +451,9 @@ func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_
 
 	filePathCol := tbl.Column(tbl.Schema().FieldIndices("file_path")[0]).Data()
 	posCol := tbl.Column(tbl.Schema().FieldIndices("pos")[0]).Data()
+	if posCol.NullN() > 0 {
+		return nil, fmt.Errorf("%w: null pos in position delete file", iceberg.ErrInvalidSchema)
+	}
 
 	return groupPosDeletesByFilePath(ctx, filePathCol, posCol)
 }
--- a/table/arrow_scanner_posdelete_regression_test.go
+++ b/table/arrow_scanner_posdelete_regression_test.go
@@ -24,7 +24,10 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/compute"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/pqarrow"
 	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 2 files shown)


table/arrow_scanner.go
```diff
@@ -408,6 +408,10 @@ func collectPosDeletePositions(positionalDeletes positionDeletes) (set[int64], e
 				return nil, fmt.Errorf("%w: unsupported pos chunk array type %T in position delete file",
 					iceberg.ErrInvalidSchema, arr)
 			}
+			if posArr.NullN() > 0 {
```
laskoviymishka: The guard itself is correct, but I'm not sure it's reachable on the real path. By the time we hit `collectPosDeletePositions`, `readDeletes` has already pushed `posCol` through `groupPosDeletesByFilePath`, which runs `compute.Filter(..., DefaultFilterOptions())` — and I'm genuinely unsure whether arrow-go's default `NullSelectionBehavior` preserves a null-pos row here or silently drops it before it ever reaches this loop.

Rather than hang correctness on that, I'd hoist the check up into `readDeletes` where `posCol` is first extracted, mirroring the `filePathCol.NullN() > 0` guard already at line 304. Then it fires before Filter regardless of the semantics. wdyt?

table/arrow_scanner_posdelete_regression_test.go
```diff
@@ -265,6 +265,27 @@ func TestCollectPosDeletePositionsRejectsUnsupportedPosType(t *testing.T) {
...
+	bldr.AppendNull()
+	bldr.Append(2)
+	posArr := bldr.NewInt64Array()
+	bldr.Release()
+	defer posArr.Release()
+
+	posCol := arrow.NewChunked(arrow.PrimitiveTypes.Int64, []arrow.Array{posArr})
+	defer posCol.Release()
+
+	deletes, err := collectPosDeletePositions(positionDeletes{posCol})
```
laskoviymishka: This calls `collectPosDeletePositions` directly, so it proves the guard in isolation but never exercises the actual path through `readDeletes` → `groupPosDeletesByFilePath`. That's exactly where the reachability question lives, so as written the test stays green even if a null pos gets silently dropped upstream.

A test that builds a delete table with a null-pos row and drives it through `readDeletes` would lock the end-to-end contract and settle the Filter question either way.

table/arrow_scanner_posdelete_regression_test.go
```diff
@@ -265,6 +265,27 @@ func TestCollectPosDeletePositionsRejectsUnsupportedPosType(t *testing.T) {
...
+	require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
+	assert.Nil(t, deletes)
+	assert.Contains(t, err.Error(), "null pos")
```
laskoviymishka: Minor, but `"null pos"` is short enough that an unrelated future message could match it and give a false pass. The sibling assertions all pin a longer fragment, so I'd widen this to `"null pos in position delete file"` to match.

laskoviymishka (changes requested):
Nice catch on this one, a null pos slipping through as a raw `0` and quietly deleting row 0 is exactly the kind of silent corruption worth guarding, and `ErrInvalidSchema` is the right call.

One thing I'd want to sort out before merge though. I'm not sure the guard is actually reachable on the real read path: `readDeletes` sends `posCol` through `groupPosDeletesByFilePath` → `compute.Filter` before `collectPosDeletePositions` gets it, and I can't convince myself whether a null-pos row survives that Filter or gets dropped first. The regression test doesn't settle it either, since it calls `collectPosDeletePositions` directly and never touches that path. I'd hoist the null check up into `readDeletes` right next to the existing `file_path` guard (line 304) so it fires regardless, and add a test that drives a null-pos delete file through `readDeletes` end to end.

Once that's in, happy to take another pass.

table/arrow_scanner_posdelete_regression_test.go
fallintoplace: Thank you, I have fixed these issues in [bd3ba42](https://github.com/apache/iceberg-go/pull/1518/commits/bd3ba42d1d3bb78337248701a1f1a6641d8637af).

laskoviymishka (approved):
👍

PR #1519, opened 2026-07-23
fix(glue): map fixed types to binary

fallintoplace:
## Summary

- map Iceberg fixed types to Glue `binary` instead of the unsupported parameterized `binary(n)` form
- cover fixed types nested inside container types
- align Glue schema conversion with Iceberg Java and PyIceberg

## Testing

- `go test ./catalog/glue`

### DIFF
--- a/catalog/glue/schema.go
+++ b/catalog/glue/schema.go
@@ -120,7 +120,7 @@ func icebergTypeToGlueType(typ iceberg.Type) string {
 	case iceberg.DecimalType:
 		return fmt.Sprintf("decimal(%d,%d)", t.Precision(), t.Scale())
 	case iceberg.FixedType:
-		return fmt.Sprintf("binary(%d)", t.Len())
+		return "binary"
 	case *iceberg.StructType:
 		// For struct types, create a struct<field1:type1,field2:type2,...> representation
 		var fieldStrings []string
--- a/catalog/glue/schema_test.go
+++ b/catalog/glue/schema_test.go
@@ -101,7 +101,16 @@ func TestIcebergTypeToGlueType(t *testing.T) {
 		{
 			name:     "fixed type",
 			input:    iceberg.FixedTypeOf(16),
-			expected: "binary(16)",
+			expected: "binary",
+		},
+		{
+			name: "nested fixed type",
+			input: &iceberg.ListType{
+				ElementID:       1,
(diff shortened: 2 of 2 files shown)


catalog/glue/schema.go
```diff
@@ -120,7 +120,7 @@ func icebergTypeToGlueType(typ iceberg.Type) string {
 	case iceberg.DecimalType:
 		return fmt.Sprintf("decimal(%d,%d)", t.Precision(), t.Scale())
 	case iceberg.FixedType:
-		return fmt.Sprintf("binary(%d)", t.Len())
+		return "binary"
```
laskoviymishka, line 123: This is the right call — Glue/Hive `BINARY` has no length parameter, so plain `binary` is what both Java's `IcebergToGlueConverter` and PyIceberg emit.

I'd drop a one-line comment noting the length is intentionally discarded, otherwise someone's liable to "fix" this back to `binary(%d)` and reintroduce the bug.

```suggestion
		// Glue/Hive BINARY has no length parameter; fixed[n] maps to plain binary.
		return "binary"
```

catalog/glue/schema_test.go
```diff
@@ -101,7 +101,16 @@ func TestIcebergTypeToGlueType(t *testing.T) {
...
+			expected: "binary",
+		},
+		{
+			name: "nested fixed type",
+			input: &iceberg.ListType{
+				ElementID:       1,
+				Element:         iceberg.FixedTypeOf(16),
+				ElementRequired: true,
+			},
+			expected: "array<binary>",
```
laskoviymishka, line 113: Good to have a nested case here. One thought: `array<binary>` mostly re-exercises the list-wrapping path that `TestNestedListType` already covers, and `FixedType` is a leaf with a single return, so this isn't closing a real branch.

A `struct<fingerprint:binary>` or `map<string,binary>` case would be more representative of real schemas and document the rendered string across the container positions. wdyt?

laskoviymishka (approved):
Nice catch! 

`binary(16)` was never a valid Hive/Glue type string (Glue's type system is Hive-based, and Hive `BINARY` takes no length parameter), and dropping to plain `binary` matches both Java's `IcebergToGlueConverter` and PyIceberg's `GLUE_PRIMITIVE_TYPES` exactly.

left couple nits inline.

PR #1520, opened 2026-07-23
fix(glue): reject dropping non-empty namespaces

fallintoplace:
## Summary

- check a Glue database for tables before dropping its namespace
- return `catalog.ErrNamespaceNotEmpty` when any table exists
- avoid deleting the database when Glue cannot determine whether it is empty

The check includes both Iceberg and non-Iceberg tables because deleting a Glue database makes all tables in that database inaccessible.

## Testing

- `go test ./catalog/glue`

### DIFF
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -685,16 +685,23 @@ func (c *Catalog) DropNamespace(ctx context.Context, namespace table.Identifier)
 		return err
 	}
 
+	// Glue has no conditional database delete, so a concurrent table creation can still race this check.
 	tables, err := c.glueSvc.GetTables(ctx, &glue.GetTablesInput{
 		CatalogId:    c.catalogId,
 		DatabaseName: aws.String(databaseName),
 		MaxResults:   aws.Int32(1),
 	})
 	if err != nil {
-		return fmt.Errorf("failed to check namespace %s for tables: %w", databaseName, err)
+		return fmt.Errorf("failed to list tables in namespace %s: %w", databaseName, err)
 	}
 	if len(tables.TableList) > 0 {
-		return fmt.Errorf("%w: %s", catalog.ErrNamespaceNotEmpty, databaseName)
+		tableType := "non-Iceberg"
+		if strings.EqualFold(tables.TableList[0].Parameters[tableParamTableType], glueTypeIceberg) {
+			tableType = "Iceberg"
+		}
+
+		return fmt.Errorf("%w: cannot drop namespace %s because it still contains %s tables",
+			catalog.ErrNamespaceNotEmpty, databaseName, tableType)
 	}
 
 	params := &glue.DeleteDatabaseInput{CatalogId: c.catalogId, Name: aws.String(databaseName)}
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -685,6 +685,18 @@ func (c *Catalog) DropNamespace(ctx context.Context, namespace table.Identifier)
 		return err
 	}
 
+	tables, err := c.glueSvc.GetTables(ctx, &glue.GetTablesInput{
+		CatalogId:    c.catalogId,
+		DatabaseName: aws.String(databaseName),
+		MaxResults:   aws.Int32(1),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to check namespace %s for tables: %w", databaseName, err)
+	}
+	if len(tables.TableList) > 0 {
+		return fmt.Errorf("%w: %s", catalog.ErrNamespaceNotEmpty, databaseName)
+	}
+
 	params := &glue.DeleteDatabaseInput{CatalogId: c.catalogId, Name: aws.String(databaseName)}
 	_, err = c.glueSvc.DeleteDatabase(ctx, params)
 	if err != nil {
--- a/catalog/glue/glue_test.go
+++ b/catalog/glue/glue_test.go
@@ -892,8 +892,6 @@ func TestGlueCreateNamespace(t *testing.T) {
 }
 
 func TestGlueDropNamespace(t *testing.T) {
-	assert := require.New(t)
-
 	mockGlueSvc := &mockGlueClient{}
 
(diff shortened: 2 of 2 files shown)


catalog/glue/glue_test.go
```diff
@@ -907,6 +905,11 @@ func TestGlueDropNamespace(t *testing.T) {
 		},
 	}, nil).Once()
 
+	mockGlueSvc.On("GetTables", mock.Anything, &glue.GetTablesInput{
```
laskoviymishka, line 910: All three new tests build the catalog as `&Catalog{glueSvc: mockGlueSvc}`, so `c.catalogId` is nil and none of the `GetTables` mock expectations pin `CatalogId`. The production call sets `CatalogId: c.catalogId` correctly — but it means the cross-account path is entirely uncovered, and a refactor that dropped the field would still go green.

Could we add a case that constructs the catalog with a non-nil `catalogId` and asserts it flows into the `GetTables` input? wdyt?

catalog/glue/glue.go
```diff
@@ -685,6 +685,18 @@ func (c *Catalog) DropNamespace(ctx context.Context, namespace table.Identifier)
...
 
+	tables, err := c.glueSvc.GetTables(ctx, &glue.GetTablesInput{
+		CatalogId:    c.catalogId,
+		DatabaseName: aws.String(databaseName),
+		MaxResults:   aws.Int32(1),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to check namespace %s for tables: %w", databaseName, err)
+	}
+	if len(tables.TableList) > 0 {
```
laskoviymishka, line 697: This is a best-effort guard — Glue has no conditional delete, so there's still a window where a concurrent `CreateTable` lands between this check and `DeleteDatabase`. That's unavoidable here and I'm fine shipping it, but I'd drop a one-line comment noting the check is best-effort and can race, so the next reader doesn't assume it's airtight.

catalog/glue/glue.go
```diff
@@ -685,6 +685,18 @@ func (c *Catalog) DropNamespace(ctx context.Context, namespace table.Identifier)
...
+		return fmt.Errorf("%w: %s", catalog.ErrNamespaceNotEmpty, databaseName)
```
laskoviymishka: Java's `GlueCatalog` and PyIceberg both distinguish the message here — "still contains Iceberg tables" vs "non-Iceberg tables" — by reading the table's `table_type` parameter. Programmatic callers key off `errors.Is(ErrNamespaceNotEmpty)` either way, so this is purely about the human-readable signal.

Worth folding in while we're here, or leave it? wdyt?

catalog/glue/glue.go
```diff
@@ -685,6 +685,18 @@ func (c *Catalog) DropNamespace(ctx context.Context, namespace table.Identifier)
 		return err
 	}
 
+	tables, err := c.glueSvc.GetTables(ctx, &glue.GetTablesInput{
+		CatalogId:    c.catalogId,
+		DatabaseName: aws.String(databaseName),
+		MaxResults:   aws.Int32(1),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to check namespace %s for tables: %w", databaseName, err)
```
laskoviymishka: Small consistency thing: the sibling errors in this file follow "failed to <verb> <noun>: %w" — e.g. "failed to list tables in namespace %s". I'd match that here, something like `failed to list tables in namespace %s: %w`.

catalog/glue/glue_test.go
```diff
@@ -916,7 +919,60 @@ func TestGlueDropNamespace(t *testing.T) {
...
+			}, mock.Anything).Return(&glue.GetDatabaseOutput{
+				Database: &types.Database{Name: aws.String("test_namespace")},
+			}, nil).Once()
+			mockGlueSvc.On("GetTables", mock.Anything, &glue.GetTablesInput{
+				DatabaseName: aws.String("test_namespace"),
+				MaxResults:   aws.Int32(1),
+			}, mock.Anything).Return(&glue.GetTablesOutput{
+				TableList: []types.Table{{
+					Name:       aws.String("test_table"),
+					Parameters: map[string]string{tableParamTableType: tableType},
```
laskoviymishka: `Parameters` here isn't read by anything — production just does `len(tables.TableList) > 0` with no type filtering, so the fixture reads as if the table type matters when it doesn't. I'd either drop it, or keep it and wire up the Iceberg-vs-non-Iceberg message distinction I flagged over in glue.go so it's actually load-bearing. Either is fine, just not the middle state.

catalog/glue/glue_test.go
```diff
@@ -916,7 +919,60 @@ func TestGlueDropNamespace(t *testing.T) {
 	}
 
 	err := glueCatalog.DropNamespace(context.TODO(), DatabaseIdentifier("test_namespace"))
-	assert.NoError(err)
+	require.NoError(t, err)
+	mockGlueSvc.AssertExpectations(t)
+}
+
+func TestGlueDropNamespaceNotEmpty(t *testing.T) {
+	for _, tableType := range []string{glueTypeIceberg, "EXTERNAL_TABLE"} {
```
laskoviymishka: `glueTypeIceberg` is a named constant sitting right next to a raw `"EXTERNAL_TABLE"`. If there's a matching constant for the external type I'd use it here for consistency; if there isn't one, no big deal.

laskoviymishka (approved):
The fix itself is the right call: gating `DeleteDatabase` on a `GetTables` emptiness check closes a genuine data-loss window (today Glue will happily drop a database out from under tables that Java/PyIceberg/Rust can still see) and lines Glue up with how Hive, SQL, and Hadoop already behave, plus the Java/PyIceberg contract. The `MaxResults=1` check is the efficient way to do it, and the three new tests cover the paths that matter.

Nothing here blocks — just a few small things I'd tidy before it merged, none of them hard, will keep it soak for a day before merging.

zeroshade (approved):
LGTM. Rejects DropNamespace on a non-empty Glue database (maps to ErrNamespaceNotEmpty) via an efficient existence probe, and distinguishes Iceberg vs non-Iceberg entries. CI green.

PR #1521, opened 2026-07-23
fix(glue): use canonical comment property for descriptions

fallintoplace:
## Summary

Glue uses `comment` as the canonical property for namespace and table descriptions, while older catalogs may still provide the legacy `Description` parameter.

This change:

- writes canonical `comment` properties without leaking description fields into Glue parameters;
- accepts legacy `Description` values when loading namespaces;
- gives canonical `comment` values deterministic precedence over legacy values;
- preserves native Glue descriptions as the highest-priority namespace value;
- applies the same compatibility behavior to namespace and table creation.

## Testing

- `go test ./catalog/glue`
- `go test ./catalog/...`
- `git diff --check`

### DIFF
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -973,6 +973,8 @@ func constructParameters(staged *table.Table, previousGlueTable *types.Table) ma
 
 	maps.Copy(parameters, staged.Properties())
 	delete(parameters, tableParamPreviousMetadataLocation)
+	delete(parameters, PropsKeyDescription)
+	delete(parameters, legacyPropsKeyDescription)
 
 	if previousGlueTable != nil {
 		if previousMetadataLocation, ok := previousGlueTable.Parameters[tableParamMetadataLocation]; ok {
--- a/catalog/glue/glue.go
+++ b/catalog/glue/glue.go
@@ -50,8 +50,9 @@ const (
 	glueTableType   = "EXTERNAL_TABLE"
 
 	// property keys
-	PropsKeyLocation    = "location"
-	PropsKeyDescription = "Description"
+	PropsKeyLocation          = "location"
+	PropsKeyDescription       = "comment"
+	legacyPropsKeyDescription = "Description"
 
 	// glue table parameter keys
 	tableParamTableType                = "table_type"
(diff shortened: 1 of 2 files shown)


catalog/glue/glue.go
```diff
@@ -970,7 +977,7 @@ func constructTableInput(tableName string, staged *table.Table, previousGlueTabl
 		},
 	}
 
-	if comment, ok := staged.Properties()[PropsKeyDescription]; ok {
+	if comment, ok := descriptionProperty(staged.Properties()); ok {
```
zeroshade, line 1001: This writes the description to `TableInput.Description`; ensure the same value is not also retained in table parameters (see the `constructParameters` note).

catalog/glue/glue_test.go
```diff
@@ -280,6 +280,61 @@ func TestGlueConstructParametersPreservesReservedParameters(t *testing.T) {
...
+			require.NoError(t, err)
+			staged := table.New(
+				TableIdentifier("test_database", "test_table"),
+				metadata,
+				"s3://test-bucket/test_table/metadata/v1.metadata.json",
+				nil,
+				nil,
+			)
+
+			tableInput := constructTableInput("test_table", staged, nil)
```
zeroshade, line 332: Add assertions (like the DB tests) that table parameters omit `comment`/`Description`, to lock in a single source of truth.

zeroshade (changes requested):
The canonical `comment` key choice is right and matches Hive/SQL/REST. But the table path still double-stores the description — it's written to both `TableInput.Description` and the Glue table parameters, which is exactly the drift this PR is meant to avoid. The database path already filters both keys; please do the same for table parameters and add assertions mirroring the DB no-parameter checks. Overlaps mechanically with #1522 (same area) — coordinate the rebase. CI green.

---
- `catalog/glue/glue.go:955` — `constructParameters` copies all staged props into Glue table parameters, including `comment` and the legacy `Description`. Combined with the `TableInput.Description` write below, the description is stored twice and can drift. `constructDatabaseInput` already filters both keys — mirror that here.

zeroshade (approved):

PR #1522, opened 2026-07-23
fix(glue): preserve existing column comments

fallintoplace:
## Problem

Glue table updates rebuild storage descriptor columns from Iceberg metadata. When a field has no Iceberg documentation, this replaces an existing Glue column comment with an empty value during otherwise unrelated commits.

## Changes

- preserve existing Glue column comments by column name when the Iceberg field doc is empty
- keep non-empty Iceberg field docs authoritative
- use the first existing comment when old Glue columns contain duplicate names
- safely handle tables without an existing storage descriptor

## Testing

- `go test ./...`

### DIFF
--- a/catalog/glue/glue_test.go
+++ b/catalog/glue/glue_test.go
@@ -337,6 +337,53 @@ func TestGlueConstructDescriptionCompatibility(t *testing.T) {
 	}
 }
 
+func TestGlueConstructTableInputPreservesExistingColumnComments(t *testing.T) {
+	metadata, err := table.NewMetadata(
+		testSchema,
+		nil,
+		table.UnsortedSortOrder,
+		"s3://test-bucket/test_table",
+		nil,
+	)
+	require.NoError(t, err)
+	staged := table.New(
+		TableIdentifier("test_database", "test_table"),
+		metadata,
+		"s3://test-bucket/test_table/metadata/v2.metadata.json",
+		nil,
+		nil,
+	)
+	previous := &types.Table{
+		StorageDescriptor: &types.StorageDescriptor{
+			Columns: []types.Column{
+				{
+					Name:       aws.String("foo"),
+					Comment:    aws.String("existing comment"),
+					Parameters: map[string]string{icebergFieldIDKey: "1"},
+				},
+			},
+		},
+	}
+
+	input := constructTableInput("test_table", staged, previous)
+	columns := make(map[string]types.Column, len(input.StorageDescriptor.Columns))
+	for _, column := range input.StorageDescriptor.Columns {
+		columns[aws.ToString(column.Name)] = column
+	}
+
+	require.Equal(t, "existing comment", aws.ToString(columns["foo"].Comment))
+	require.Nil(t, columns["bar"].Comment)
+	require.Nil(t, columns["baz"].Comment)
+
+	withoutStorageDescriptor := &types.Table{}
+	for _, previous := range []*types.Table{nil, withoutStorageDescriptor} {
+		input := constructTableInput("test_table", staged, previous)
+		for _, column := range input.StorageDescriptor.Columns {
+			require.Nil(t, column.Comment)
+		}
+	}
+}
+
 func TestGlueGetTableCaseInsensitive(t *testing.T) {
 	assert := require.New(t)
 
--- a/catalog/glue/schema.go
+++ b/catalog/glue/schema.go
@@ -30,7 +30,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/glue/types"
 )
 
-func schemasToGlueColumns(metadata table.Metadata) []types.Column {
+func schemasToGlueColumns(metadata table.Metadata, existingColumns []types.Column) []types.Column {
 	results := make(map[string]types.Column)
 
 	for _, field := range schemaToGlueColumns(metadata.CurrentSchema(), true) {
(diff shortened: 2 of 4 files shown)


catalog/glue/schema.go
```diff
@@ -49,6 +49,27 @@ func schemasToGlueColumns(metadata table.Metadata) []types.Column {
...
+		if column.Comment == nil {
+			continue
+		}
+		if _, ok := existingComments[name]; !ok {
+			existingComments[name] = column.Comment
+		}
+	}
+
+	for name, column := range results {
+		if aws.ToString(column.Comment) != "" {
```
laskoviymishka: This closes the clobbering bug but opens its mirror: once a Glue comment exists, there's no way to clear it. A user who sets `Doc` then clears it produces `Comment = aws.String("")` from `fieldToGlueColumn`, this loop sees empty and restores the old comment, so the clear is dropped forever.

Root cause is `fieldToGlueColumn` writing `aws.String(field.Doc)` unconditionally — `nil` and empty-string become indistinguishable downstream. I'd have it emit `nil` when `Doc == ""`, then gate this loop on `column.Comment == nil` rather than the empty-string check. That distinguishes "doc never set, defer to Glue" from "explicitly cleared." Thoughts?

catalog/glue/schema.go
```diff
@@ -49,6 +49,27 @@ func schemasToGlueColumns(metadata table.Metadata) []types.Column {
 		}
 	}
 
+	existingComments := make(map[string]*string)
+	for _, column := range existingColumns {
+		name := aws.ToString(column.Name)
```
laskoviymishka: Keying on column name means a rename or drop+re-add carries a stale comment onto a semantically different field — rename field ID=5 `address`, add a new field ID=7 also named `address`, and the new one inherits the old comment on the next commit.

Both sides already carry `iceberg.field.id` in `Parameters`, so I'd key `existingComments` off that instead of `Name`. It's strictly more correct and it also removes the duplicate-name ambiguity for free. Java's converter matches on name too so this isn't a regression, but I'd rather not inherit the limitation. wdyt?

catalog/glue/schema.go
```diff
@@ -49,6 +49,27 @@ func schemasToGlueColumns(metadata table.Metadata) []types.Column {
...
+		if column.Comment == nil {
+			continue
+		}
+		if _, ok := existingComments[name]; !ok {
+			existingComments[name] = column.Comment
```
laskoviymishka: Small thing: this aliases the `*string` straight from `previousGlueTable` into the output column, so the returned `TableInput` shares memory with the previous table. Strings are immutable so it's harmless today, but I'd copy it (`aws.String(aws.ToString(column.Comment))`) to keep the output independent of the previous table's lifetime.

catalog/glue/glue_test.go
```diff
@@ -280,6 +280,35 @@ func TestGlueConstructParametersPreservesReservedParameters(t *testing.T) {
...
+		StorageDescriptor: &types.StorageDescriptor{
+			Columns: []types.Column{
+				{Name: aws.String("foo"), Comment: aws.String("existing comment")},
+			},
+		},
+	}
+
+	input := constructTableInput("test_table", staged, previous)
+
+	require.Equal(t, "existing comment", aws.ToString(input.StorageDescriptor.Columns[0].Comment))
```
laskoviymishka: This only checks `foo`. `testSchema` also has `bar`/`baz` with no matching previous column, so I'd assert those come back with empty comments too — that's what catches anything spurious getting back-filled.

`constructTableInput` now branches on `previousGlueTable` being nil (the create path), and that branch is only exercised implicitly. A case passing `nil` previous would be worth having so the refactor's new nil-guard is actually covered.

catalog/glue/schema_test.go
```diff
@@ -448,6 +448,29 @@ func TestSchemasToGlueColumns(t *testing.T) {
...
 }
+
+func TestSchemasToGlueColumnsPreservesExistingComments(t *testing.T) {
+	schema := iceberg.NewSchema(0,
+		iceberg.NestedField{ID: 1, Name: "preserved", Type: iceberg.PrimitiveTypes.String},
+		iceberg.NestedField{ID: 2, Name: "overridden", Type: iceberg.PrimitiveTypes.String, Doc: "iceberg comment"},
+		iceberg.NestedField{ID: 3, Name: "missing", Type: iceberg.PrimitiveTypes.String},
+	)
+	metadata, err := table.NewMetadata(schema, nil, table.SortOrder{}, "s3://example/path", nil)
+	assert.NoError(t, err)
```
laskoviymishka: `assert.NoError` keeps going on failure, and `NewMetadata` returns a nil `Metadata` interface on error — so a failure here turns into a nil-deref panic in `CurrentSchema()` instead of a clean assertion failure. I'd use `require.NoError`. The pre-existing `TestSchemasToGlueColumns` above has the same pattern, worth fixing while we're here.

catalog/glue/schema_test.go
```diff
@@ -448,6 +448,29 @@ func TestSchemasToGlueColumns(t *testing.T) {
...
+
+	existingColumns := []types.Column{
+		{Name: aws.String("preserved"), Comment: aws.String("existing comment")},
+		{Name: aws.String("preserved"), Comment: aws.String("duplicate comment")},
+		{Name: aws.String("overridden"), Comment: aws.String("existing comment")},
+		{Name: aws.String("missing")},
+	}
+
+	columns := schemasToGlueColumns(metadata, existingColumns)
```
laskoviymishka, line 498: This covers preserve/override/missing but not the case that actually gates the design discussion — a field whose `Doc` was non-empty and is now `""`. Whatever we land on for the clear semantics, I'd add a case here so the decision is pinned down and can't silently regress.

catalog/glue/schema_test.go
```diff
@@ -448,6 +448,29 @@ func TestSchemasToGlueColumns(t *testing.T) {
...
+
+	assert.Equal(t, "existing comment", aws.ToString(columns[0].Comment))
```
laskoviymishka: These assert by positional index, which leans on the sort-by-field-ID order holding. A named lookup would make the intent explicit and survive a reorder.

laskoviymishka (changes requested):
Nice catch on the clobbering bug, a field with an empty `Doc` was writing `aws.String("")` straight over whatever comment already lived in Glue, and preserving the existing comment is the right fix.

I'd hold before merging on one thing though: as written we've traded one version of the bug for its mirror. Once a Glue comment exists, the early-continue on empty means there's no longer any way to clear it — set a doc, then clear it, and the old comment gets silently restored forever. The clean fix is one level down in `fieldToGlueColumn`: emit `nil` for the `Comment` when `Doc == ""` instead of `aws.String("")`, then gate the back-fill loop on `column.Comment == nil`. That's what lets us tell "never set, defer to Glue" apart from "explicitly cleared."

The other thing I'd like us to at least decide on is the name-based matching — keying preservation on column name drags a stale comment onto a different field after a rename or drop+re-add. Both columns already carry `iceberg.field.id` in their `Parameters`, so keying off that is strictly more correct. Java's converter has the same name-based limitation so I don't treat it as a hard blocker, but I'd rather not inherit it if the field-ID key is right there.

- in `fieldToGlueColumn`, emit `nil` rather than `aws.String("")` for an empty doc, and gate the back-fill on `nil` so an intentional clear actually clears
- key preservation by field ID instead of name (or consciously accept Java's name-based limitation)
- add a test for the "doc cleared to empty" case, and tighten the new tests (`require.NoError`, assert the unmatched columns stay clean, cover the nil-previous path)

Once the clear case is sorted, happy to take another pass and approve.

laskoviymishka (approved):
:shipit:

laskoviymishka:
@fallintoplace need to resolve conflicts to merge

zeroshade (approved):
Right shape now — nil vs `""` distinguishes "never documented, defer to Glue" from "explicitly cleared", and keying the back-fill on `iceberg.field.id` avoids dragging a stale comment onto a renamed/re-added field. Nil-previous / nil-StorageDescriptor is handled, and only the `UpdateTable` path passes a previous table so create/register can't accidentally back-fill. Coverage is real: reverting `schema.go` fails `TestGlueConstructTableInputPreservesExistingColumnComments`, `TestSchemasToGlueColumnsPreservesExistingCommentsByFieldID`, and `TestFieldToGlueColumn/field_without_doc`; the clear case is pinned by `TestSchemasToGlueColumnsClearsPreviousIcebergComment`.

Non-blocking: the "cleared" inference reads `metadata.Schemas()`, so if the schema that carried the doc isn't retained, a cleared comment could resurrect from Glue — worth a line in the comment above `clearedComments`. Also the PR description still says preservation is "by column name"; it's by field ID now.

PR #1524, opened 2026-07-24
fix(hadoop): return fully qualified child namespaces

fallintoplace:
## Motivation

For a nested namespace such as `["a", "child"]`, `ListNamespaces(ctx, ["a"])` currently returns only `["child"]`. The truncated identifier cannot be passed back to catalog methods such as `CheckNamespaceExists`, `LoadNamespaceProperties`, or `ListTables` because those methods interpret it as a root namespace.

## Changes

- Build listed child identifiers by appending the directory name to the requested parent.
- Create nested namespaces through the catalog API in the regression test.
- Verify the exact fully qualified results and round-trip every returned identifier through `CheckNamespaceExists`.

## Testing

- `go test ./...`
- `golangci-lint run ./catalog/hadoop`

### DIFF
--- a/catalog/hadoop/hadoop_test.go
+++ b/catalog/hadoop/hadoop_test.go
@@ -1062,17 +1062,19 @@ func (s *HadoopCatalogTestSuite) TestListNamespacesEmpty() {
 }
 
 func (s *HadoopCatalogTestSuite) TestListNamespacesNested() {
-	parentDir := filepath.Join(s.warehouse, "a")
-	s.Require().NoError(os.MkdirAll(filepath.Join(parentDir, "child1"), 0o755))
-	s.Require().NoError(os.MkdirAll(filepath.Join(parentDir, "child2"), 0o755))
+	ctx := context.Background()
+	parent := table.Identifier{"a"}
+	expected := []table.Identifier{{"a", "child1"}, {"a", "child2"}}
+	s.Require().NoError(s.cat.CreateNamespace(ctx, parent, nil))
+	for _, namespace := range expected {
+		s.Require().NoError(s.cat.CreateNamespace(ctx, namespace, nil))
+	}
 
-	namespaces, err := s.cat.ListNamespaces(context.Background(), []string{"a"})
+	namespaces, err := s.cat.ListNamespaces(ctx, parent)
 	s.Require().NoError(err)
-	s.Len(namespaces, 2)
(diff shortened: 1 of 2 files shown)


catalog/hadoop/hadoop_test.go
```diff
@@ -1062,15 +1062,22 @@ func (s *HadoopCatalogTestSuite) TestListNamespacesEmpty() {
 }
 
 func (s *HadoopCatalogTestSuite) TestListNamespacesNested() {
-	parentDir := filepath.Join(s.warehouse, "a")
-	s.Require().NoError(os.MkdirAll(filepath.Join(parentDir, "child1"), 0o755))
-	s.Require().NoError(os.MkdirAll(filepath.Join(parentDir, "child2"), 0o755))
+	ctx := context.Background()
+	parent := table.Identifier{"a"}
+	expected := []table.Identifier{{"a", "child1"}, {"a", "child2"}}
```
tanmayrauth, line 1067: This only exercises a depth-1 parent, so it can't tell "prepend the full parent" apart from "prepend just the last parent component" — both yield ["a","child1"]. A future change that qualified with parent[len-1] instead of the whole slice would keep this test green while breaking any 2+level namespace (listing ["a","child1"] would return ["child1","grandchild"] instead of ["a","child1","grandchild"], and the CheckNamespaceExists round-trip would then fail against the real tree). Add a two-level case: create ["a","child1","grandchild"], then assert ListNamespaces(ctx, ["a","child1"]) returns exactly [["a","child1","grandchild"]] and round-trips through CheckNamespaceExists. That ocks down the full-prefix behavior the fix on hadoop.go:1064 introduces.

tanmayrauth (commented):
Fix is right and matches how the SQL catalog and this file's own ListTables already qualify identifiers. One test-gap worth closing. Details inline.

zeroshade (approved):
LGTM. Qualifying against the requested parent matches the SQL catalog and this file's own `ListTables`, and `slices.Clone` is the right detail — a plain `append(parent, …)` would let successive children stomp a shared backing array. `TestListNamespacesNested` now round-trips `["a","child1","grandchild"]` and asserts the exact result of `ListNamespaces(["a","child1"])`, so a last-component-only regression can't return.

Non-blocking: `slices.Concat(parent, table.Identifier{d.Name()})` is one allocation instead of clone-then-grow. Separately (not this PR), Hadoop lists only direct children while SQL lists all descendants — worth reconciling eventually.

PR #1525, opened 2026-07-24
fix(view): reject null and duplicate metadata entries

fallintoplace:
## Motivation

View metadata validation dereferences every schema and version entry without first checking for `null`, so malformed metadata can panic during parsing instead of returning `ErrInvalidViewMetadata`.

Duplicate schema or version IDs are also accepted. The metadata keeps all duplicate entries in the public lists while its ID indexes retain only one, making `CurrentSchema` and `CurrentVersion` depend on array order.

## Changes

- Reject null schema and version entries before any validation dereferences them.
- Reject duplicate `schema-id` and `version-id` values before constructing or using ID lookups.
- Add regression cases for null and duplicate schemas and versions, including an explicit no-panic assertion.

## Testing

- `go test ./...`
- `golangci-lint run ./view`

### DIFF
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -435,7 +435,7 @@ func (m *metadata) validate() error {
 		return fmt.Errorf("%w: at least one schema is required", ErrInvalidViewMetadata)
 	}
 
-	if err := m.checkSchemaAndVersionIDs(); err != nil {
+	if err := m.checkSchemaAndVersionEntries(); err != nil {
 		return err
 	}
 
@@ -454,7 +454,7 @@ func (m *metadata) validate() error {
 	return nil
 }
 
-func (m *metadata) checkSchemaAndVersionIDs() error {
+func (m *metadata) checkSchemaAndVersionEntries() error {
 	schemaIDs := make(map[int]struct{}, len(m.SchemaList))
 	for i, schema := range m.SchemaList {
 		if schema == nil {
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -435,6 +435,10 @@ func (m *metadata) validate() error {
 		return fmt.Errorf("%w: at least one schema is required", ErrInvalidViewMetadata)
 	}
 
+	if err := m.checkSchemaAndVersionIDs(); err != nil {
+		return err
+	}
+
 	if err := m.checkCurrentVersionExists(); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


view/metadata.go
```diff
@@ -450,6 +454,32 @@ func (m *metadata) validate() error {
 	return nil
 }
 
+func (m *metadata) checkSchemaAndVersionEntries() error {
```
tanmayrauth, line 457: This fixes the view path, but the identical latent panic is still live on the table metadata read path: commonMetadata.checkSchemas() (table/metadata.go:2056) ranges over SchemaList and dereferences s.ID with no nil guard, and it runs during ParseMetadataBytes — so a table metadata doc with "schemas":[null] panics with a nil-pointer deref instead of returning ErrInvalidMetadata (confirmed locally). Table metadata also never rejects duplicate schema-ids, so  it silently last-wins through the same indexBy pattern this PR is fixing for views, whereas Java throws for both (versionsById()/schemasById() use ImmutableMap.builder()). Not asking to expand this PR — but a follow-up applying the same null + duplicate guard to checkSchemas (and the partition-spec/sort-order equivalents) would keep the two metadata paths from diverging.

view/metadata.go
tanmayrauth, line 494: Minor: now that checkSchemaAndVersionEntries runs first and guarantees non-nil, unique schema ids, this function rebuilds a schema-id set that init() already materializes as lazySchemasByID(). Could drop the local map and use m.lazySchemasByID() to avoid a third pass over SchemaList. Cosmetic, not blocking.

tanmayrauth (commented):
Change looks right and matches Java (ImmutableMap.builder() rejects dup schema/version ids there too). One follow-up worth flagging: the same latent panic exists on the table metadata path. Details inline.

zeroshade (approved):
LGTM. Ordering is right: `checkSchemaAndVersionEntries` runs before anything dereferences entries (`init()` only installs `sync.OnceValue` indexes), and parse (`UnmarshalJSON` → `validate`) is the correct layer since the builder can't produce duplicates (`addSchema`/`addVersion` reuse IDs). Error type matches the package. `TestParseMetadataRejectsNullAndDuplicateEntries` fails on main for all four cases (nil-pointer panic on the nulls, nil error on the duplicates). Nice cleanup reusing `lazySchemasByID()` too.

Non-blocking: the same null/duplicate gap exists on the table path (`commonMetadata.checkSchemas`, spec/sort-order equivalents) — worth a follow-up so the two don't diverge.

PR #1526, opened 2026-07-24
fix(udf): protect metadata and identifiers from mutation

fallintoplace:
## Motivation

UDF metadata is documented as immutable, but its getters expose internal maps, slices, pointers, nested type fields, and raw JSON buffers. Mutating a definition returned by `Definitions` or `DefinitionByID` can even leave the lazy definition index keyed by the old ID while the stored definition contains a new one.

Table metadata getters already return defensive copies following #1444. Applying the same boundary to UDF metadata keeps the metadata APIs consistent rather than leaving UDF values externally mutable.

The loaded UDF identifier has the same aliasing problem at construction and access.

## Changes

- Return defensive copies from definition, definition-log, property, and raw-representation getters.
- Deep-copy definition versions, nested UDF types, and unknown representation JSON.
- Make cloning part of the sealed type and representation contracts so new implementations must define copy behavior.
- Clone identifiers both when constructing a UDF and when returning its identifier.
- Document the copy guarantees on metadata getters and add mutation regression coverage.

## Testing

- `go test ./...`
- `golangci-lint run ./udf`

### DIFF
--- a/udf/metadata.go
+++ b/udf/metadata.go
@@ -80,6 +80,7 @@ type Representation interface {
 	// isRepresentation seals the interface to the implementations in
 	// this package.
 	isRepresentation()
+	clone() Representation
 }
 
 // SQLRepresentation stores the function body as a SQL expression in a
@@ -105,6 +106,7 @@ func NewSQLRepresentation(dialect, sql string) (SQLRepresentation, error) {
 
 func (SQLRepresentation) RepresentationType() string { return "sql" }
 func (SQLRepresentation) isRepresentation()          {}
+func (r SQLRepresentation) clone() Representation    { return r }
 
 func (r SQLRepresentation) MarshalJSON() ([]byte, error) {
 	return json.Marshal(struct {
@@ -122,11 +124,16 @@ type UnknownRepresentation struct {
 	raw json.RawMessage
 }
 
-// Raw returns the representation's original JSON.
+// Raw returns a copy of the representation's original JSON.
 func (r UnknownRepresentation) Raw() json.RawMessage { return slices.Clone(r.raw) }
 
 func (r UnknownRepresentation) RepresentationType() string { return r.TypeName }
 func (UnknownRepresentation) isRepresentation()            {}
+func (r UnknownRepresentation) clone() Representation {
+	r.raw = slices.Clone(r.raw)
+
+	return r
+}
 
 func (r UnknownRepresentation) MarshalJSON() ([]byte, error) {
 	return slices.Clone(r.raw), nil
@@ -148,14 +155,10 @@ func representationsEqual(a, b Representation) bool {
 }
 
 func cloneRepresentations(representations []Representation) []Representation {
-	cloned := make([]Representation, len(representations))
+	cloned := slices.Clone(representations)
 	for i, repr := range representations {
-		switch r := repr.(type) {
-		case UnknownRepresentation:
-			r.raw = slices.Clone(r.raw)
-			cloned[i] = r
-		default:
-			cloned[i] = r
+		if repr != nil {
+			cloned[i] = repr.clone()
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


udf/metadata.go
```diff
@@ -733,19 +755,24 @@ type metadata struct {
...
-func (m *metadata) Location() string                    { return m.Loc }
-func (m *metadata) Definitions() []*Definition          { return m.DefinitionList }
-func (m *metadata) DefinitionLog() []DefinitionLogEntry { return m.DefinitionLogList }
-func (m *metadata) Properties() iceberg.Properties      { return m.Props }
-func (m *metadata) Secure() bool                        { return m.SecureValue }
-func (m *metadata) Doc() string                         { return m.DocValue }
+func (m *metadata) FormatVersion() int         { return m.FormatVersionValue }
+func (m *metadata) FunctionUUID() uuid.UUID    { return *m.UUID }
+func (m *metadata) Location() string           { return m.Loc }
+func (m *metadata) Definitions() []*Definition { return cloneSlice(m.DefinitionList) }
```
tanmayrauth, line 761: This makes UDF metadata copy-on-read, which no other metadata type does — commonMetadata returns its internals raw (Schemas() → c.SchemaList, Snapshots() → c.SnapshotList, Properties() → c.Props at table/metadata.go:1646/1685/1727). The repo's contract is immutability-by-convention with the builder as the only isolation point, and the UDF builder already honors that: AddDefinition and AddDefinitionVersion clone their inputs (metadata_builder.go:363, :423) and MetadataBuilderFromBase deep-copies the base.
  
  Recommend dropping the read-path clones and matching the convention. Copy-on-read doesn't scale to TableMetadata — those accessors sit on read/planning paths and would start allocating per call — so keeping it here just leaves UDF a permanent outlier for a safety property nothing else in the repo asks for. If the actual goal is hard immutability, the Iceberg-idiomatic route is unexported fields + constructors like Java's ViewMetadata/TableMetadata, not per-getter copies, but that's a separate change.

udf/metadata.go
```diff
@@ -733,19 +755,24 @@ type metadata struct {
...
+func (m *metadata) DefinitionLog() []DefinitionLogEntry {
+	return cloneDefinitionLog(m.DefinitionLogList)
+}
+func (m *metadata) Properties() iceberg.Properties { return maps.Clone(m.Props) }
+func (m *metadata) Secure() bool                   { return m.SecureValue }
+func (m *metadata) Doc() string                    { return m.DocValue }
 
 func (m *metadata) DefinitionByID(definitionID string) (*Definition, bool) {
```
tanmayrauth, line 769: This is the getter that actually needed fixing: lazyDefinitionsByID caches the same *Definition pointers keyed by ID, so a caller mutating def.DefinitionID on the returned pointer desyncs the map (index keyed by the old ID, value holding the new one). That real bug is fixable here alone — return def.Clone() in this method, or stop handing out the indexed pointer — without making   Definitions()/DefinitionLog()/Properties() copy on every call. Keep this narrow fix and the sweeping copy-on-read in the other getters is solving a problem the convention already covers.

tanmayrauth (commented):
Identifier handling is right. The one thing to settle before this lands: this makes UDF metadata copy-on-read, which no other metadata type in the repo does — recommend matching the house convention instead. Details inline.

udf/metadata.go
fallintoplace: I just have another look. Could this comment be based on stale code? https://github.com/apache/iceberg-go/pull/1444 changed the table metadata getters to return defensive copies, including Schemas, Snapshots, and Properties, and that commit is already in this branch’s history. It looks like the UDF getters here now match the convention on main.

zeroshade (approved):
LGTM. Checked the copies are deep, not header-deep: `Definitions()` → `Definition.Clone` covers parameter types, `ReturnType`, `Versions` → `cloneRepresentations` (incl. `UnknownRepresentation.raw`) and `ReturnNullable`; `cloneDefinitionLog` clones nested `DefinitionVersions`. `DefinitionByID` returning `Clone()` fixes a real bug — mutating the indexed pointer's `DefinitionID` previously desynced `lazyDefinitionsByID`. Adding `clone()` to the sealed `Type`/`Representation` interfaces isn't an external break. `TestMetadataGettersReturnDefensiveCopies`, `TestDefinitionCopiesNestedTypes`, `TestNewDefinitionCopiesInputTypes`, `TestUnknownRepresentationRawReturnsCopy` all fail when reverted. Copy-on-read is the house convention post-#1444, so this matches rather than diverges.

Non-blocking: `NewDefinition` and `Definition.Clone` duplicate the parameter-clone loop — a `cloneParameters` helper would DRY it.

PR #1527, opened 2026-07-25
fix(partitions): retain dropped-source partition fields

### LINKED ISSUE #1205: PartitionType should retain dropped-source partition fields as UnknownType
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

> Follow-up to #1190 / #1191 (requested in [this review comment](https://github.com/apache/iceberg-go/pull/1191#issuecomment-4694968108)). This is a **consistency / latent-footgun** fix, not a live wrong-results bug; #1191 already fixed the one path that miscompared. Filing separately because the change has real blast radius.

## Problem

`PartitionSpec.PartitionType(schema)` **compacts** its result: a partition field whose source column is absent from `schema` (e.g. the column was dropped) is skipped rather than retained.

This diverges from Java Iceberg, whose `PartitionSpec.partitionType()` **retains every field**, substituting `UnknownType` when the source column is gone:

```java
// When the source field has been dropped we cannot determine the type
if (sourceType == null) {
  resultType = Types.UnknownType.get();
}
```

Because manifest partition summaries (`ManifestFile.Partitions()` → `[]FieldSummary`) are **positional against the full partition spec**, the compacted struct cannot be used to index them. #1191 worked around this by adding a parallel `manifestPartitionFields` helper in `table/evaluators.go` that retains every field with an `UnknownType` placeholder, leaving the codebase with **two notions of "partition type"** (compacted vs. full). Any future positional consumer that reaches for `PartitionType()` will silently reintroduce the #1190 class of bug.

## Why This Matters

- **Single source of truth.** One method should define a spec's partition struct; the `manifestPartitionFields` duplicate exists only to compensate for `PartitionType()`'s compaction.
- **Reference parity.** Java retains dropped-source fields as `UnknownType`; matching it removes a subtle behavioral divergence.
- **Footgun prevention.** The compaction is an implicit trap for positional callers, not a documented contract. (#1191 adds a doc note to `PartitionType()` warning against positional use as an interim guard.)

## Expected Behavior

`PartitionType(schema)` retains **all** partition spec fields in order, using `UnknownType` for any field whose source column is missing from `schema`, matching Java. `manifestPartitionFields` is then deleted and `newManifestEvaluator` calls `PartitionType()` directly.

## Blast Radius (call sites to audit)

`PartitionType()` (and the positional `newPartitionRecord`/`GetPartitionRecord` builder that keys values by `FieldID` over `FieldList`) is consumed by:

**Write / encode paths** - in practice always invoked with a schema where every partition source is present, so compaction is a no-op today; each must nonetheless tolerate an `UnknownType` placeholder after the change:
- `manifest.go`- partition `fieldStats` building, and `partitionTypeToAvroSchema(...)` (**avro mapping for `UnknownType` must be confirmed**)
- `data_file_codec.go` - partition encoding
- `partitions.go` - `PartitionToPath`
- `table/snapshots.go` - `GetPartitionRecord` for produced data/delete files
- `table/partitioned_fanout_writer.go`, `table/pos_delete_partitioned_fanout_writer.go`

**Read / scan path** - internally consistent today (schema + record built from the *same* compacted list), but should keep working once placeholders appear:
- `table/scanner.go` - `buildPartitionEvaluator` (`partSchema` + `GetPartitionRecord`); the projected partition filter never references a dropped source, so an unreferenced `UnknownType` placeholder must remain harmless to `ExpressionEvaluator` binding.

## Proposed Fix

1. Change `PartitionType()` to emit an `UnknownType` placeholder (instead of `continue`) for fields whose source column is absent.
2. Remove `manifestPartitionFields`; have `newManifestEvaluator` use `PartitionType()`.
3. Audit the consumers above - primarily that avro partition-schema conversion and the data-file codec tolerate/guard `UnknownType`, and that evaluator binding ignores unreferenced placeholder fields.
4. Tests: dropped-source partition field round-trips positionally through manifest evaluation (the #1191 regression test), plus avro/codec encode and scan-evaluator coverage with a dropped source present.

fallintoplace:
## Problem

`PartitionSpec.PartitionType(schema)` compacted away fields whose source columns were no longer present. Manifest partition summaries are positional against the complete partition spec, so the compacted type was unsafe for positional consumers.

#1191 fixed the active manifest-pruning bug with a separate `manifestPartitionFields` implementation. That left two definitions of a partition type and made the same positional bug easy to reintroduce elsewhere.

## Fix

Make `PartitionType` retain every spec field in order and use it as the single source of truth.

When a source column is missing, `UnknownType` is passed to the transform:

- source-dependent transforms such as identity, truncate, and void resolve to `UnknownType`
- fixed-result transforms such as bucket and time transforms retain their known result type

This matches [current Java Iceberg behavior](https://github.com/apache/iceberg/blob/470f4642bdda336bac5372c79c762a2654dfe3f3/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L188-L200), which is slightly more precise than degrading every dropped-source transform to unknown.

The supporting audit also:

- removes `manifestPartitionFields` and uses `PartitionType` for manifest evaluation
- maps unknown partition fields to Avro `null`
- detects unknown field IDs from the generated entry schema and normalizes stale historical values to null during encoding
- records null presence without inventing lower or upper bounds for unknown fields
- preserves full field positions in partition paths and fanout records

## Safety

Projected filters cannot reference a dropped source ID, so unknown placeholders remain unreferenced during expression binding.

A rewritten historical entry may still carry the old non-null partition value. Its source type is no longer available, so that value cannot be encoded or compared safely. Normalizing only schema-identified unknown fields to null keeps the entry writable and positionally correct; fixed-result fields continue to encode with their known types.

## Tests

Coverage includes:

- source-dependent and fixed-result type resolution
- positional manifest pruning with dropped fields
- Avro schema conversion
- data-file codec round-trip from a stale non-null value to null
- manifest summary construction
- partition paths and fanout partition records

Validation:

- `go test -count=1 ./...`

Fixes #1205

### DIFF
--- a/data_file_codec.go
+++ b/data_file_codec.go
@@ -107,7 +107,7 @@ func (d *dataFile) MarshalAvroEntry(spec PartitionSpec, schema *Schema, version 
 		return nil, err
 	}
 	clone := cloneDataFileAvroFields(d)
-	partitionData, err := avroEncodePartitionData(d.Partition(), maps.nameToID, maps.idToType, maps.idToFixedSize)
+	partitionData, err := avroEncodePartitionData(d.Partition(), maps)
 	if err != nil {
 		return nil, err
 	}
@@ -209,13 +209,20 @@ func cloneDataFileAvroFields(src *dataFile) *dataFile {
 // iceberg-typed values like Date or Decimal) into the name-keyed
 // avro-friendly map the manifest-entry schema expects. Idempotent:
 // values already in primitive form pass through unchanged.
-func avroEncodePartitionData(idKeyed map[int]any, nameToID map[string]int, logicalTypes map[int]string, fixedSizes map[int]int) (map[string]any, error) {
-	converted, err := avroPartitionData(idKeyed, logicalTypes, fixedSizes)
+func avroEncodePartitionData(idKeyed map[int]any, fields dataFileFieldMaps) (map[string]any, error) {
+	converted, err := avroPartitionData(idKeyed, fields.idToType, fields.idToFixedSize)
 	if err != nil {
 		return nil, err
 	}
 	out := make(map[string]any, len(converted))
-	for name, id := range nameToID {
+	for name, id := range fields.nameToID {
+		if _, unknown := fields.unknownFieldIDs[id]; unknown {
+			// UnknownType has no value representation and is encoded as Avro
+			// null. Normalize historical values whose source type was dropped.
+			out[name] = nil
+
+			continue
+		}
 		if v, ok := converted[id]; ok {
 			out[name] = v
 		}
@@ -229,6 +236,7 @@ type dataFileFieldMaps struct {
 	idToType         map[int]string
 	idToFixedSize    map[int]int
 	idToDecimalScale map[int]int
+	unknownFieldIDs  map[int]struct{}
 }
 
 // dataFileSchemaCacheKey identifies a cached avro schema by the
--- a/data_file_codec_test.go
+++ b/data_file_codec_test.go
@@ -66,6 +66,33 @@ func TestManifestEntrySchemaForCachesByPartitionAvroFingerprint(t *testing.T) {
 		"identical (partition type, version) inputs must reuse the cached schema")
 }
 
+func TestDataFileCodecWithDroppedPartitionSource(t *testing.T) {
+	spec := NewPartitionSpec(
+		PartitionField{SourceIDs: []int{1}, FieldID: 1000, Name: "dropped", Transform: IdentityTransform{}},
+		PartitionField{SourceIDs: []int{2}, FieldID: 1001, Name: "bucketed", Transform: BucketTransform{NumBuckets: 8}},
+	)
+	schema := NewSchema(0)
+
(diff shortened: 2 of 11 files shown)


zeroshade (approved):
Approach matches upstream Java (`PartitionSpec.resultType()` substitutes `UnknownType` for a dropped source), and collapsing `manifestPartitionFields` back into `PartitionType` removes the duplicate #1191 had to add. Reproduced the pre-fix bug: a 2-field spec with a dropped first source emitted a single `FieldSummary` and dropped the partition value; now both summaries are emitted positionally and the entry round-trips as `{1000: nil, 1001: 3}`. Tests cover type resolution, avro null conversion, codec round-trip and summary construction.

Non-blocking: `partitioned_fanout_writer.go:256` now silently `continue`s on a missing source on the *current* spec — worth a one-line comment that this yields a null partition value rather than an error. And a `NewManifestWriter → ReadManifest` test for the `unknownFieldIDs` plumbing would pin it directly (verified working manually). Worth a changelog note that rewriting a historical entry normalizes a dropped identity/truncate value to null.

PR #1528, opened 2026-07-25
fix(table): safely match binary equality-delete partitions

### LINKED ISSUE #1349: fix(table): avoid panic when matching equality-delete partitions
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

# Equality-delete scan planning can panic when partition values are non-comparable

## Summary

`table/scanner.go` performs direct `!=` comparisons on partition values during equality-delete scan planning. While this works for comparable Go types, it panics when partition values are non-comparable (for example, slices).

As a result, scan planning can crash instead of returning a normal error.

---

## Impact

- Equality-delete scan planning can panic unexpectedly.
- Any scan path that calls `partitionsMatch` is affected.
- Causes a runtime panic (`comparing uncomparable type`) instead of graceful error handling.
- This is a correctness and stability issue.

---

## Affected Location

```
table/scanner.go
Lines: 547-558
```

---

## Root Cause

The implementation compares partition values using Go's `!=` operator.

Example:

```go
if left != right {
    ...
}
```

However, Go only allows `==` and `!=` on **comparable** types.

Examples of non-comparable types include:

- `[]T` (slices)
- `map[K]V`
- functions

Attempting to compare these values results in a runtime panic:

```
panic: runtime error: comparing uncomparable type []string
```

---

## Expected Behavior

Partition matching should safely handle every supported partition value type.

For non-comparable values, the scanner should:

- use an appropriate typed comparison, or
- detect unsupported/non-comparable values and return a normal error.

Scan planning should never panic because of partition value comparison.

---

## Actual Behavior

When a partition value contains a non-comparable Go type, equality-delete planning crashes with a runtime panic similar to:

```
panic: runtime error: comparing uncomparable type
```

instead of returning an error.

---

fallintoplace:
## Problem

Equality-delete planning compares data-file and delete-file partition maps before attaching a delete to a scan task. The maps store values as `any`, and the previous direct `!=` comparison panicked when a supported partition value was non-comparable in Go.

Binary and fixed partition values are represented as `[]byte`, so a valid table could fail during scan planning with `comparing uncomparable type []uint8`.

## Fix

Compare the maps with `maps.EqualFunc` and `reflect.DeepEqual`.

This keeps the existing matching contract:

- both maps must contain the same field IDs
- values must have the same concrete type and value
- independently allocated byte slices with the same bytes match
- different binary values or type mismatches do not match

It also remains safe if another non-comparable value reaches a custom `DataFile` implementation. Comparable primitives retain their previous equality semantics, including NaN values remaining unequal.

## Tests

The tests cover the comparison helper and the real equality-delete matching path, including:

- equal and different byte slices
- binary versus string values
- comparable multi-field partitions
- different partition field IDs
- selecting only the matching delete file without panicking

Validation:

- `go test -count=1 ./...`

Fixes #1349

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -22,7 +22,9 @@ import (
 	"context"
 	"fmt"
 	"iter"
+	"maps"
 	"math"
+	"reflect"
 	"slices"
 	"strings"
 	"sync"
@@ -589,17 +591,7 @@ func matchEqualityDeletesToData(dataEntry iceberg.ManifestEntry, eqDeleteEntries
 }
 
 func partitionsMatch(a, b map[int]any) bool {
-	if len(a) != len(b) {
-		return false
-	}
-
-	for k, v := range a {
-		if bv, ok := b[k]; !ok || bv != v {
-			return false
-		}
-	}
-
-	return true
+	return maps.EqualFunc(a, b, reflect.DeepEqual)
 }
 
 // buildDVIndex indexes deletion vectors by the data file path they reference.
--- a/table/scanner_internal_test.go
+++ b/table/scanner_internal_test.go
@@ -53,6 +53,78 @@ func newDeleteManifest(minSeqNum int64) iceberg.ManifestFile {
 		Build()
 }
 
+func TestPartitionsMatchHandlesBinaryValues(t *testing.T) {
+	tests := []struct {
+		name  string
+		left  map[int]any
+		right map[int]any
+		match bool
+	}{
+		{
+			name:  "equal binary values",
+			left:  map[int]any{1000: []byte{0xde, 0xad}},
+			right: map[int]any{1000: append([]byte(nil), 0xde, 0xad)},
+			match: true,
+		},
+		{
+			name:  "different binary values",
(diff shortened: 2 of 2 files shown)


tanmayrauth (approved):
LGTM, thanks for raising this PR!

zeroshade (approved):
Correct fix for #1349. `maps.EqualFunc` + `reflect.DeepEqual` keeps the existing contract (same field IDs, type-sensitive equality, NaN still unequal) while making binary/fixed partitions compare by content, and it lines up with the compaction-side mirror (`partitionMatchKey` uses `%v`), so reader and compaction agree on which eq-deletes apply. Both `TestPartitionsMatchHandlesBinaryValues` and `TestMatchEqualityDeletesToDataHandlesBinaryPartitions` panic with `comparing uncomparable type []uint8` on the old code.

Non-blocking: this runs per (data file × eq-delete) pair, so if it ever shows up in a profile a `case []byte: bytes.Equal(...)` fast path with `DeepEqual` fallback is the cheap follow-up.

PR #1529, opened 2026-07-25
fix(transforms): bucket timestamp nanos at microsecond precision

### LINKED ISSUE #1498: BucketTransform hashes timestamp_ns values at nanosecond precision
### Apache Iceberg version

main (development)

## Summary

`BucketTransform` hashes `timestamp_ns` and `timestamptz_ns` values using their raw nanosecond representation.

The Iceberg specification requires nanosecond timestamps to be converted to microseconds before hashing. This ensures equivalent microsecond and nanosecond timestamps produce the same bucket.

## Reproduction

```go
package main

import (
	"fmt"
	"time"

	iceberg "github.com/apache/iceberg-go"
)

func apply(transform iceberg.BucketTransform, value iceberg.Literal) iceberg.Literal {
	return transform.Apply(iceberg.Optional[iceberg.Literal]{
		Valid: true,
		Val:   value,
	}).Val
}

func main() {
	transform := iceberg.BucketTransform{NumBuckets: 16}
	value := time.Date(2017, 11, 16, 22, 31, 8, 1_001, time.UTC)

	micros := iceberg.Timestamp(value.UnixMicro())
	nanos := iceberg.TimestampNano(value.UnixNano())

	fmt.Printf("timestamp:    %s\n", apply(transform, iceberg.NewLiteral(micros)))
	fmt.Printf("timestamp_ns: %s\n", apply(transform, iceberg.NewLiteral(nanos)))
}
```

Run from the repository root:

```shell
go run /tmp/iceberg_timestamp_ns_bucket_repro.go
```

## Actual behavior

```text
timestamp:    6
timestamp_ns: 13
```

The same instant is assigned to different buckets. The mismatch also affects equality-predicate projection.

## Expected behavior

Both representations should produce bucket `6`.

Per the Iceberg specification, `timestamp_ns` and `timestamptz_ns` must be converted to microsecond precision before applying the Murmur3 hash.

## Impact

This can create incompatible partition values between iceberg-go and other Iceberg implementations:

- Go can prune conforming files that contain matching rows.
- Other implementations can prune files written by Go.
- Go-written `bucket[N]` partitions on nanosecond timestamps may be incompatible with Java Iceberg and PyIceberg.

This can result in silent missing query results rather than a visible error.

## Likely cause

`BucketTransform.Apply` and `BucketTransform.Transformer` currently pass raw nanoseconds to `hashHelperInt`.

The nanosecond value should first be converted to microseconds using floor division, including for pre-epoch timestamps.

fallintoplace:
## Problem

Iceberg bucket transforms hash timestamps as their microsecond integer representation. `timestamp_ns` and `timestamptz_ns` were instead hashing the raw nanosecond value, so the same instant could land in a different bucket depending on its timestamp precision.

That breaks partition compatibility with other Iceberg implementations and can make equality projection prune the wrong files.

## Fix

Convert nanoseconds to microseconds before Murmur3 hashing, and use the same helper from both `Apply` and `Transformer`. Predicate projection delegates to those paths, so all three entry points now agree.

The conversion uses floor division. This matters before the Unix epoch: `-1ns` belongs to microsecond `-1`, while ordinary Go integer division would truncate it to `0`.

This follows the [Iceberg Appendix B hash requirements](https://iceberg.apache.org/spec/#appendix-b-32-bit-hash-requirements) and mirrors [Java Iceberg’s timestamp-nano bucket implementation](https://github.com/apache/iceberg/blob/470f4642bdda336bac5372c79c762a2654dfe3f3/api/src/main/java/org/apache/iceberg/transforms/Bucket.java#L235-L247).

## Tests

Coverage verifies:

- `timestamp_ns` and `timestamptz_ns`
- `Apply`, `Transformer`, and equality projection
- positive and sub-microsecond values
- pre-epoch values and the negative microsecond boundary
- the Appendix B example that must produce bucket `6` for 16 buckets

Validation:

- `go test -count=1 ./...`

Fixes #1498

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -291,6 +291,12 @@ func hashHelperInt[T ~int32 | ~int64](v any) uint32 {
 	return murmur3.Sum32(b)
 }
 
+func hashTimestampNano(v any) uint32 {
+	micros := internal.FloorDiv(int64(v.(TimestampNano)), int64(time.Microsecond))
+
+	return hashHelperInt[int64](micros)
+}
+
 func (t BucketTransform) Equals(other Transform) bool {
 	rhs, ok := other.(BucketTransform)
 	if !ok {
@@ -330,7 +336,7 @@ func (t BucketTransform) Apply(value Optional[Literal]) Optional[Literal] {
 	case TimestampLiteral:
 		hash = hashHelperInt[int64](int64(v))
 	case TimestampNsLiteral:
-		hash = hashHelperInt[int64](int64(v))
+		hash = hashTimestampNano(TimestampNano(v))
 	default:
 		return Optional[Literal]{}
 	}
@@ -364,7 +370,7 @@ func (t BucketTransform) Transformer(src Type) func(any) Optional[int32] {
 	case TimestampTzType:
 		h = hashHelperInt[Timestamp]
 	case TimestampNsType, TimestampTzNsType:
-		h = hashHelperInt[TimestampNano]
+		h = hashTimestampNano
 	case DecimalType:
 		h = func(v any) uint32 {
 			b, _ := DecimalLiteral(v.(Decimal)).MarshalBinary()
--- a/transforms_test.go
+++ b/transforms_test.go
@@ -624,12 +624,23 @@ func TestYearMonthTransformNanoseconds(t *testing.T) {
 
 func TestBucketTransformTimestampNanoseconds(t *testing.T) {
 	transform := iceberg.BucketTransform{NumBuckets: 16}
+	specBucket := int32(6)
 	values := []struct {
-		name string
-		ts   iceberg.TimestampNano
+		name       string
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
Verified. Floor-division to micros before hashing is right, and `FloorDiv` is the correct primitive — `-1ns` must map to micro `-1`, which plain Go truncation would turn into `0`. Hash-site coverage is complete: `Apply` and `Transformer` are the only two, and `Project` / `partitioned_fanout_writer.go:285` route through them. The test cross-checks each ns result against the equivalent micros literal plus the Appendix B anchor (bucket 6); it fails on main for the sub-microsecond, negative-boundary and Appendix-B cases across both ns types and all three entry points.

Changelog note: this changes bucket assignment for existing ns-partitioned tables written by iceberg-go — which is the point (the old values were non-spec and incompatible with Java).

PR #1530, opened 2026-07-25
fix(transforms): avoid hour string overflow

fallintoplace:
## Summary

- convert hour ordinals through Unix seconds instead of time.Duration
- avoid duration overflow for valid dates outside roughly 1677 to 2262
- cover years before and after that range

This preserves the direct epoch-hour behavior used by the Java implementation.

## Testing

- go test . -count=1

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -1097,7 +1097,8 @@ func (HourTransform) Apply(value Optional[Literal]) (out Optional[Literal]) {
 func (HourTransform) ToHumanStr(val any) string {
 	switch v := val.(type) {
 	case int32:
-		tm := epochTM.Add(time.Duration(v) * time.Hour)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Correct. `time.Duration(v) * time.Hour` overflows int64 nanoseconds past ~±2562047 hours, so hour ordinals outside ~1677–2262 silently wrapped. Going through `time.Unix(v*3600, 0)` removes the ceiling — worst case `MaxInt32 * 3600 ≈ 7.7e12`s is nowhere near int64 range. The two new `TestToHumanString` rows fail on main with wrapped output (`4645896` → `1915-06-14-00`). No other `time.Duration(...)` conversion remains in transforms.go; Year/Month/Day use `AddDate`. LGTM.

PR #1531, opened 2026-07-25
fix(table): reject invalid schema entries

fallintoplace:
## Summary

- reject null entries in the table metadata schemas list
- reject duplicate schema IDs, including duplicates after the current schema
- return ErrInvalidMetadata instead of allowing a nil dereference

This matches the non-null, unique schema indexing used by the Java implementation.

## Testing

- go test ./table -count=1

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2054,11 +2054,23 @@ func (c *commonMetadata) preValidate() {
 }
 
 func (c *commonMetadata) checkSchemas() error {
-	// check that current-schema-id is present in schemas
-	for _, s := range c.SchemaList {
-		if s.ID == c.CurrentSchemaID {
-			return nil
+	seen := make(map[int]struct{}, len(c.SchemaList))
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Correct and minimal. Confirmed the crash it fixes: `"schemas":[null]` makes `checkSortOrders` → `commonMetadata.CurrentSchema()` deref a nil `*Schema` and panic out of `ParseMetadataBytes`; reverting the source hunk makes the `null schema` subtest panic. Duplicate schema IDs were previously accepted silently, and rejecting them matches Java (`indexSchemas` uses an `ImmutableMap.Builder`). `checkSchemas` running first means the nil guard covers the deref paths, and `MetadataBuilder.AddSchema` reuses IDs so nothing we write trips the check.

Non-blocking: the same duplicate-ID check is absent for `partition-specs`/`sort-orders` — same class of corruption, fine as a follow-up.

PR #1532, opened 2026-07-25
fix(puffin): return defensive metadata copies

fallintoplace:
## Summary

- return independent blob and file property metadata from Reader
- deep-copy nested fields, properties, and compression codec pointers
- protect cached footer metadata through ReadBlob and ReadAllBlobs results

This matches the immutable metadata exposure in the Java implementation and the borrowed immutable accessors in Rust.

## Testing

- go test ./puffin -count=1

### DIFF
--- a/puffin/puffin_reader.go
+++ b/puffin/puffin_reader.go
@@ -24,6 +24,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
+	"slices"
 	"sort"
 )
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Copy depth checks out. `BlobMetadata` has exactly three reference fields — `Fields`, `Properties`, `CompressionCodec` — and `cloneBlobMetadata` handles all three. Exposure coverage is complete: `Blobs`, `Properties`, `ReadBlob`, `ReadAllBlobs` are the only paths that hand out footer-owned metadata (`ReadBlobByMetadata` returns bytes from caller-supplied metadata; there's no `Footer()` accessor). `TestReaderReturnsMetadataCopies` fails on main and mirrors the merged `TestAddBlobCopiesMetadataInputs`.

Non-blocking: the `CompressionCodec` clone branch is currently unreachable (writer never sets a codec, reads reject non-empty), so it's untested — fine as forward-looking, just flagging.

PR #1533, opened 2026-07-25
fix(literals): reject timestamp nanos overflow

fallintoplace:
## Summary

- check microsecond timestamp bounds before converting to nanoseconds
- return ErrBadCast instead of silently wrapping int64 values
- cover both timestamp_ns variants and both overflow directions

This follows the checked multiplication used by the Java implementation.

## Testing

- go test . -count=1

### DIFF
--- a/literals.go
+++ b/literals.go
@@ -824,9 +824,11 @@ func (t TimestampLiteral) To(typ Type) (Literal, error) {
 		return t, nil
 	case TimestampTzType:
 		return t, nil
-	case TimestampNsType:
-		return TimestampNsLiteral(Timestamp(t).ToNanos()), nil
-	case TimestampTzNsType:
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Bounds are exactly right, which is the part worth checking. `MaxInt64/1000 = 9223372036854775` scales to `…775000`; Go truncates toward zero so `MinInt64/1000` scales to `-…775000` — both in range, and one step beyond genuinely wraps (verified `(MinInt64/1000 - 1)*1000` lands on `+9223372036854775616`). So the guard rejects nothing valid and accepts nothing that overflows. `Timestamp.ToNanos` has a single call site, so there's no second unguarded path. `TestTimestampLiteralToNanosRejectsOverflow` asserts both overflow directions *and* that the in-range boundaries still convert. LGTM.

PR #1534, opened 2026-07-25
fix(schema): protect field getters from mutation

fallintoplace:
## Summary

Return independent values from exported schema field getters.

## Why

Field and Fields accessors exposed nested types and cached slices directly. Callers could mutate a returned value and change later schema lookups.

## What changed

Nested types are cloned for Field and Fields. FindFieldByID and FlatFields return independent values, while cached schema indexes remain private to internal read-only lookup paths. Other aliasing surfaces are out of scope for this PR.

## Tests

- go test . -count=1
- Verified the repository packages covered by this change

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -339,8 +339,37 @@ func cloneFields(fields []NestedField) []NestedField {
 
 func cloneField(field NestedField) NestedField {
 	field.Type = cloneType(field.Type)
+	field.InitialDefault = cloneDefault(field.InitialDefault)
+	field.WriteDefault = cloneDefault(field.WriteDefault)
 
 	return field
+}
+
+func cloneDefault(value any) any {
+	switch value := value.(type) {
+	case []byte:
+		return slices.Clone(value)
+	case BinaryLiteral:
+		return BinaryLiteral(slices.Clone([]byte(value)))
+	case FixedLiteral:
+		return FixedLiteral(slices.Clone([]byte(value)))
+	case []any:
+		cloned := make([]any, len(value))
+		for i, item := range value {
+			cloned[i] = cloneDefault(item)
+		}
+
+		return cloned
+	case map[string]any:
+		cloned := make(map[string]any, len(value))
+		for key, item := range value {
+			cloned[key] = cloneDefault(item)
+		}
+
+		return cloned
+	default:
+		return value
+	}
 }
 
 func cloneType(t Type) Type {
--- a/schema.go
+++ b/schema.go
@@ -181,7 +181,13 @@ func (s *Schema) FlatFields() (iter.Seq[NestedField], error) {
 		return nil, err
 	}
 
-	return maps.Values(fields), nil
+	return func(yield func(NestedField) bool) {
+		for field := range maps.Values(fields) {
+			if !yield(cloneField(field)) {
+				return
+			}
+		}
+	}, nil
 }
 
 func (s *Schema) lazyIDToField() (map[int]NestedField, error) {
@@ -267,8 +273,8 @@ func (s *Schema) AsStruct() StructType { return StructType{FieldList: cloneField
 func (s *Schema) asStructRef() StructType { return StructType{FieldList: s.fields} }
 
 func (s *Schema) NumFields() int          { return len(s.fields) }
-func (s *Schema) Field(i int) NestedField { return s.fields[i] }
-func (s *Schema) Fields() []NestedField   { return slices.Clone(s.fields) }
+func (s *Schema) Field(i int) NestedField { return cloneField(s.fields[i]) }
+func (s *Schema) Fields() []NestedField   { return cloneFields(s.fields) }
 func (s *Schema) FieldIDs() []int {
 	idx, _ := s.lazyNameToID()
 
@@ -325,11 +331,16 @@ func cloneFields(fields []NestedField) []NestedField {
 
 	cloned := make([]NestedField, len(fields))
 	for i, field := range fields {
-		cloned[i] = field
-		cloned[i].Type = cloneType(field.Type)
+		cloned[i] = cloneField(field)
 	}
 
 	return cloned
+}
+
+func cloneField(field NestedField) NestedField {
+	field.Type = cloneType(field.Type)
+
+	return field
 }
 
 func cloneType(t Type) Type {
@@ -409,6 +420,9 @@ func (s *Schema) FindFieldByNameCaseInsensitive(name string) (NestedField, bool)
 func (s *Schema) FindFieldByID(id int) (NestedField, bool) {
 	idx, _ := s.lazyIDToField()
 	f, ok := idx[id]
+	if ok {
+		f = cloneField(f)
+	}
 
 	return f, ok
 }
--- a/schema.go
+++ b/schema.go
@@ -340,11 +340,15 @@ func cloneFields(fields []NestedField) []NestedField {
 }
 
 func cloneField(field NestedField) NestedField {
-	field.Type = cloneType(field.Type)
-	field.InitialDefault = cloneDefault(field.InitialDefault)
-	field.WriteDefault = cloneDefault(field.WriteDefault)
-
-	return field
+	return NestedField{
+		ID:             field.ID,
+		Name:           field.Name,
+		Type:           cloneType(field.Type),
+		Required:       field.Required,
+		Doc:            field.Doc,
+		InitialDefault: cloneDefault(field.InitialDefault),
+		WriteDefault:   cloneDefault(field.WriteDefault),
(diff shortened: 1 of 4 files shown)


zeroshade (commented):
Correctness is right and the coverage is real — reverting the source makes `TestSchemaFieldGettersReturnDefensiveCopies` fail on the `WriteDefault` leak, and `FindFieldByName`/`FindTypeByID`/etc. inherit the clone by delegating to `FindFieldByID`; `cloneDefault` matches the existing `table.cloneDefault`.

My concern is cost on read paths. `FindFieldByID` now deep-clones (nested `Type` + `InitialDefault`/`WriteDefault`) on every call, and `arrowAccessor.FieldPartner` (`arrow_utils.go:754`) + `castIfNeeded` (`:937`) call it for every field of every record batch out of `ToRequestedSchema`, needing only `Name`/`Type`. On a struct field with ~50 children I measured `FindFieldByID` go from ~7ns/0 allocs to ~1.2µs/4.9KB/2 allocs — a per-field-per-batch O(subtree) copy on the write path. Please add an internal non-cloning lookup for those call sites (the codebase already has this pattern with `asStructRef`), or post numbers showing it doesn't matter.

Also: the same aliasing hole is still open on exported `IndexByID` and `Schema.NameMapping()` — belong in the same pass or a follow-up. Not blocking correctness, but I'd like the hot-path cost addressed before merge.

schema.go
```diff
@@ -323,13 +331,47 @@ func cloneFields(fields []NestedField) []NestedField {
...
 	for i, field := range fields {
-		cloned[i] = field
-		cloned[i].Type = cloneType(field.Type)
+		cloned[i] = cloneField(field)
 	}
 
 	return cloned
 }
 
+func cloneField(field NestedField) NestedField {
```
laskoviymishka, line 342: This mutate-a-copy-and-return shape has a small trap: `cloneField(f)` with the result dropped compiles fine and silently skips the clone, which is the exact leak we're closing here. A struct-literal return would make an omitted assignment impossible to write, and let the compiler catch a forgotten field while we're at it. Not blocking — wdyt?

schema.go
```diff
@@ -323,13 +331,47 @@ func cloneFields(fields []NestedField) []NestedField {
...
+	field.Type = cloneType(field.Type)
+	field.InitialDefault = cloneDefault(field.InitialDefault)
+	field.WriteDefault = cloneDefault(field.WriteDefault)
+
+	return field
+}
+
+func cloneDefault(value any) any {
```
laskoviymishka, line 354: This is now the third byte-for-byte copy of `cloneDefault` (`table/metadata.go` and `view/metadata.go` have the others). Not asking to fix it in this PR, but could we file a follow-up to hoist `cloneDefault`/`cloneField`/`cloneType` into one place? Right now a new default type has to be handled in three files or it silently aliases in two of them.

schema.go
```diff
@@ -323,13 +331,47 @@ func cloneFields(fields []NestedField) []NestedField {
...
+
+		return cloned
+	case map[string]any:
+		cloned := make(map[string]any, len(value))
+		for key, item := range value {
+			cloned[key] = cloneDefault(item)
+		}
+
+		return cloned
+	default:
```
laskoviymishka, line 376: The default branch is correct today — everything that reaches it (int64, float64, string, bool, uuid.UUID, Decimal) is a value type, so the shallow return can't alias. That's a real invariant though, and it's invisible here. Could we add a short comment listing the expected types and noting each must be a value type or immutable reference? A future default backed by a pointer would otherwise slip through silently.

schema.go
```diff
@@ -405,6 +447,18 @@ func (s *Schema) FindFieldByNameCaseInsensitive(name string) (NestedField, bool)
...
+		f = cloneField(f)
+	}
+
+	return f, ok
+}
+
+// FindFieldByIDRef returns a schema-owned field without cloning it. It is limited
+// to internal callers that only read the result and need to avoid clone-on-read
+// overhead on hot paths.
+func (s *Schema) FindFieldByIDRef(id int, _ internal.SchemaRef) (NestedField, bool) {
```
laskoviymishka, line 471: This is the one surface that deliberately hands back a live reference, so I'd make the contract explicit. The returned `NestedField` is a value copy, but `f.Type` is the same `*StructType`/`*ListType`/`*MapType` pointer the schema owns, so `f.Type.(*StructType).FieldList[0].Name = "x"` reaches straight through and corrupts the live schema.

The two callers today (`FieldPartner`, `castIfNeeded`) are read-only so they're fine, but the doc is the only thing guarding this, and right now it says "limited to internal callers" without saying the Type tree is shared. I'd spell it out: the returned field shares its Type pointer with the schema, and callers must treat `Type` and anything reachable from it as read-only.

While we're here — the `Ref` suffix doesn't really signal "skips clone-on-read" to a reader, but that's a naming taste call; the doc is the important part.

schema_test.go
```diff
@@ -318,6 +320,87 @@ func TestSchemaAsStructClonesNestedMapTypes(t *testing.T) {
...
+	})
+	t.Run("FlatFields", func(t *testing.T) {
+		fields, err := schema.FlatFields()
+		require.NoError(t, err)
+		byID := make(map[int]iceberg.NestedField)
+		for field := range fields {
+			byID[field.ID] = field
+		}
+		assertIndependent(t, byID[1], byID[3])
+	})
```
laskoviymishka, line 406: `FindFieldByName`, `FindFieldByNameCaseInsensitive`, and `FindTypeByID` all get their clone transitively by delegating to `FindFieldByID`, so they're correct today. But nothing exercises them directly — a refactor that short-circuits the delegation would still pass this test. Since `assertIndependent` already does the work, a couple more subtests through those finders would lock the contract cheaply. wdyt?

schema_test.go
```diff
@@ -318,6 +320,87 @@ func TestSchemaAsStructClonesNestedMapTypes(t *testing.T) {
...
+		}
+	}
+	schema := iceberg.NewSchema(0, iceberg.NestedField{
+		ID: 1, Name: "parent", Type: &iceberg.StructType{FieldList: children},
+	})
+	_, ok := schema.FindFieldByIDRef(1, internal.SchemaRef{})
+	require.True(t, ok)
+
+	var field iceberg.NestedField
+	assert.Zero(t, testing.AllocsPerRun(100, func() {
```
laskoviymishka, line 423: Two small things on this test. The zero-alloc check leans on `field` being read outside the lambda to stay live — a future escape-analysis change could elide the assignment and pass spuriously; `runtime.KeepAlive(field)` inside the closure pins it. And since this is the one path that intentionally aliases, a companion assertion that a write through the returned ref does reach the schema would document that contract right next to the defensive-copy test. Both optional.

laskoviymishka (changes requested):
Nice work — the correctness fix here is real (reverting the getters makes `TestSchemaFieldGettersReturnDefensiveCopies` fail on the `WriteDefault` leak), the defensive-copy coverage is genuine, and `cloneDefault` covers every default shape we actually produce and lines up with the existing `table.cloneDefault`. The response to Matt's hot-path concern also looks right to me — `FindFieldByIDRef` plus switching `FieldPartner`/`castIfNeeded` onto it is the clean way to keep the write path zero-alloc.

I'd hold before merge though, mostly around one thing: the "completes immutable-schema behavior" framing isn't quite true yet. The per-getter paths are locked down, but `IndexByID` still returns fields whose `.Type` aliases the live tree, `NameMapping()` hands back the memoized slice directly, and `IdentifierFieldIDs` is a public mutable slice. Matt already flagged the `IndexByID`/`NameMapping` side as same-pass-or-follow-up — I don't think all of it has to land here, but I'd like us to pick explicitly: which holes close in this PR, and which get a tracked issue, so we're not merging under a "complete" banner that has known gaps.

The other thing I'd tighten is `FindFieldByIDRef` itself, since it's the new surface that deliberately hands out a live reference. Right now the doc comment is the only guard against a future internal caller mutating through `f.Type`, and it doesn't actually say the Type tree is shared. Worth making that contract explicit (details inline).

- Document the `FindFieldByIDRef` contract — the returned Type tree is shared, treat it as read-only.
- Decide this-PR vs. tracked follow-up for the `IndexByID` / `NameMapping` / `IdentifierFieldIDs` holes, and adjust the description if they're deferred.
- Document the `cloneDefault` default-branch invariant (everything reaching it is a value type).

The rest is small — a `cloneField` idiom nit, the three-way `cloneDefault` duplication (follow-up), and a couple of test tightenings. Once the contract doc and the scope call are settled, happy to take another pass and approve.

laskoviymishka (approved):

PR #1537, opened 2026-07-25
fix(table): emit Last sentinel for scan tasks with no record batches

### LINKED ISSUE #1536: Filtered scan silently drops all rows queued behind a task whose row groups are fully pruned
### Apache Iceberg version

main (also reproduced on v0.5.0 and v0.6.0)

### Please describe the bug

A filtered table scan can silently return zero rows (no error) when a scan task earlier in plan order produces no record batches — for example when every row group of its file is pruned by parquet row-group statistics.

**Root cause.** `createIterator` (table/arrow_scanner.go) sequences per-task records and only releases task N+1's records after task N emitted a record with `Last=true`. The `dropFile` path emits an empty sentinel batch for statically skipped files, but `processRecords` only sends records when `prev != nil` — a task whose reader yields no batches emits nothing at all. The first such task stalls the sequenced channel; when the channel closes, the iterator's deferred drain releases every queued record from the tasks behind it without yielding them.

**Trigger conditions.** ≥2 planned files where an earlier-ordered file (a) survives file-level pruning (e.g. its manifest entry has no column bounds — common for files registered via `NewDataFileBuilder`/`AddDataFiles` by external committers) and (b) has all of its row groups eliminated by row-group statistics under the scan filter.

**Real-world impact.** We hit this in production reading a table populated by a 2-minute micro-batch committer plus AWS Glue compaction: any time-windowed filtered read returned 0 rows while Athena/Trino returned millions from the same snapshot. Nothing fails — the data is just missing.

**Reproduction.** The linked PR adds `TestScanSurvivesFullyPrunedTask`: two statless data files registered in one commit, a filter that fully prunes the first file's row groups; before the fix the second file's rows are silently dropped (`expected: []int32{1000, 1001, 1002}, actual: nil`), and a checked allocator additionally reports the stalled record leak.

**Fix.** Emit the same empty sentinel batch the `dropFile` path uses when a task's reader produced no records. PR incoming.

### Willingness to contribute

- [x] I can contribute a fix for this bug independently

hankivstmb:
Closes #1536.

## What

A filtered scan silently returns zero rows (no error) when a scan task earlier in plan order produces no record batches — e.g. when every row group of its file is pruned by parquet row-group statistics.

`createIterator` sequences per-task records and only releases task N+1's records after task N emitted a record with `Last=true`. The `dropFile` path already emits an empty sentinel batch for statically skipped files, but a task whose reader yields no batches emitted nothing: the first such task stalls the sequenced channel, and on close the deferred drain releases every queued record from later tasks without yielding them.

## Fix

Emit the same empty sentinel batch the `dropFile` path uses when `processRecords` finishes with no records produced.

## Test

`TestScanSurvivesFullyPrunedTask`: two statless data files (registered via `NewDataFileBuilder`/`AddDataFiles` in one commit, so plan order is deterministic and file-level pruning cannot eliminate either), scanned with a filter that fully prunes the first file's row groups.

- Before: `expected: []int32{1000, 1001, 1002}, actual: nil` — the second file's rows are silently dropped; a checked allocator additionally reports the stalled record leak.
- After: the second file's rows are returned; allocator clean.

Found in production: a table populated by a micro-batch committer plus AWS Glue compaction returned 0 rows for every windowed filtered read in iceberg-go while Athena/Trino returned millions from the same snapshot.

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -1009,6 +1009,20 @@ func (as *arrowScan) processRecords(
 		out <- enumeratedRecord{Record: internal.Enumerated[arrow.RecordBatch]{
 			Value: prev, Index: idx, Last: true,
 		}, Task: task}
+	} else {
+		// The reader produced no batches (e.g. every row group was pruned by
+		// stats). The sequenced channel in createIterator still needs this
+		// task's Last record to release the following tasks' records, so emit
+		// an empty batch just like the dropFile path does; without it every
+		// record queued behind this task is silently discarded on close.
+		var emptySchema *arrow.Schema
+		emptySchema, err = SchemaToArrowSchema(as.projectedSchema, nil, false, as.useLargeTypes)
+		if err != nil {
+			return err
+		}
+		out <- enumeratedRecord{Record: internal.Enumerated[arrow.RecordBatch]{
+			Value: array.NewRecordBatch(emptySchema, nil, 0), Index: idx, Last: true,
+		}, Task: task}
 	}
 
 	if recRdr.Err() != nil && recRdr.Err() != io.EOF {
--- a/table/empty_scan_task_regression_test.go
+++ b/table/empty_scan_task_regression_test.go
@@ -1,0 +1,181 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table_test
+
+// Regression test for: filtered scans silently returning zero rows when an
+// earlier-ordered scan task produces no record batches.
+//
+// createIterator sequences per-task records and only releases task N+1's
+// records after task N emitted a record with Last=true. A task whose file is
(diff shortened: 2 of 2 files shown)


tanmayrauth (commented):
LGTM. One nit: when the reader errors on its first Next() rather than genuinely yielding zero batches, this branch still allocates and emits the sentinel before the recRdr.Err(), so an errored task emits a sentinel *and* the deferred error record. Harmless (the error wins in both the sequencer and the iterator, and the sentinel is drained on close), just a wasted allocation on the error path — could gate  the emit on recRdr.Err()==nil.

zeroshade (approved):
Verified locally: the test fails with only `table/arrow_scanner.go` reverted (`expected []int32{1000,1001,1002}, actual nil`, plus a 64-byte allocator leak) and passes with the fix. Root cause and fix are right — the sequencer's `prev.Record.Last && next.Record.Index == 0` gate requires every task to emit a terminal record, `idx` is still 0, exactly one `Last` per task, and `createIterator` already drops 0-row batches so the sentinel never surfaces. Matching the existing `dropFile` sentinel is the right minimal shape.

Non-blocking: (1) the sentinel is emitted before the `recRdr.Err()` check, so a reader failing on first `Next()` emits a sentinel *and* an error record (harmless, wasted alloc). (2) the sentinel schema is hardcoded to `as.projectedSchema` but `processRecords` is shared with `producePosDeletesFromTask` (schema `PositionalDeleteSchema`) — invisible today only because 0-row batches are dropped.

zeroshade:
@hankivstmb heads-up — main is currently failing to compile from a merge collision between this PR and #1426. This PR added the empty-batch sentinel at table/arrow_scanner.go:1027 using internal.Enumerated[...], while #1426 (merged around the same time) renamed that file's table/internal import to the alias tblutils. Neither diff conflicted textually, but the merged file now has 'undefined: internal', which fails the Go and Audit-and-Verify workflows on main (cascading to codec via the table import). One-line fix at arrow_scanner.go:1027: change internal.Enumerated[arrow.RecordBatch]{ to tblutils.Enumerated[arrow.RecordBatch]{ . Could you open a quick fix PR? Merge-ordering on our side, sorry about that — happy to fast-track the review.

PR #1538, opened 2026-07-25
fix(table): return defensive sort field copies

fallintoplace:
## What changed

- Copy the SortField slice and each SourceIDs slice in NewSortOrder.
- Yield fresh SourceIDs slices from SortOrder.Fields.
- Preserve the existing validation and serialization behavior.

## Why

SortOrder is expected to remain stable after construction and validation. Previously it retained the caller-owned field slice, while Fields returned values whose SourceIDs still shared internal backing arrays. Mutating either value could silently change later serialization and compatibility checks.

The regression test mutates both constructor inputs and iterator results, then reads the order again to verify that its fields remain unchanged.

## Testing

- go test ./table
- go vet ./table

### DIFF
--- a/table/sorting.go
+++ b/table/sorting.go
@@ -341,6 +341,18 @@ func newSortOrder(orderID int, fields []SortField, validateSourceIDs bool) (Sort
 
 func cloneSortField(field SortField) SortField {
 	field.SourceIDs = slices.Clone(field.SourceIDs)
+	switch transform := field.Transform.(type) {
+	case *iceberg.BucketTransform:
+		if transform != nil {
+			cloned := *transform
+			field.Transform = &cloned
+		}
+	case *iceberg.TruncateTransform:
+		if transform != nil {
+			cloned := *transform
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approving. The read-path leak was the one that mattered — `Fields()` handed out `SortField`s whose `SourceIDs` aliased the validated order — and copying in `newSortOrder` covers `UnmarshalJSON` too. `fields` is unexported, `Fields()` is the only accessor, and `AssignFreshSortOrderIDsWithID` builds its own `SourceIDs`, so no external mutation path remains. `TestSortOrderReturnsDefensiveCopies` fails on all three vectors when reverted.

Non-blocking: the transform clone is a type switch over `*BucketTransform`/`*TruncateTransform` — exhaustive today, but add a line on `cloneSortField` noting it, or it silently misses a future pointer transform. Also `PartitionSpec.Fields()` still yields internal `SourceIDs` — same hole one type over, good follow-up.

PR #1539, opened 2026-07-25
fix(hive): cap lock backoff without overflow

fallintoplace:
## What changed

Replace floating-point exponential backoff calculation with bounded integer arithmetic. The calculation now clamps to LockMaxWaitTime before either the shift count or time.Duration can overflow.

## Why

Lock retry counts are configurable. At sufficiently large attempts, converting the floating-point result back to time.Duration produced a negative duration. time.After then returned immediately, bypassing the configured maximum wait and causing rapid lock polling.

Values below the overflow boundary retain exact exponential behavior, while larger attempts saturate at the configured maximum.

## Testing

- go test ./catalog/hive
- go vet ./catalog/hive
- boundary coverage at attempts 62 and 63 using the maximum time.Duration

### DIFF
--- a/catalog/hive/lock.go
+++ b/catalog/hive/lock.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approving — the guard is exact, not just conservative: `minWait > maxWait>>attempt` is precisely `minWait<<attempt > maxWait`, so nothing clamps early and nothing overflows. Boundary checks out (`minWait=1ns, maxWait=MaxInt64`: attempt 62 → `1<<62`, 63 → `maxWait`), and dropping `math.Pow` removes the float precision loss.

Note the regression test is arch-sensitive: out-of-range float→int64 is implementation-defined, so the old bug reproduces on amd64 (CI) but saturates on arm64. Non-blocking: `attempt <= 0 || minWait <= 0` returns `minWait` verbatim, so a negative `lock-check-min-wait-time` still busy-polls — `max(minWait, 0)` closes it while you're here.

PR #1540, opened 2026-07-25
fix(partitions): return defensive field copies

fallintoplace:
## What changed

- Clone SourceIDs for fields returned by PartitionSpec.Field and PartitionSpec.Fields.
- Deep-copy fields returned by FieldsBySourceID instead of cloning only the outer slice.
- Preserve a nil result when no fields match a source ID.
- Reuse the same field-copy helper during construction.

## Why

PartitionField is returned by value, but its SourceIDs slice previously shared backing storage with the spec. A caller could mutate a supposedly read-only result, alter later serialization, and leave the cached source ID lookup keyed by an outdated value.

The regression test mutates results from all three getter paths and verifies that the spec and lookup remain unchanged.

## Testing

- go test .
- go vet .

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -370,10 +370,7 @@ func NewPartitionSpec(fields ...PartitionField) PartitionSpec {
 func NewPartitionSpecID(id int, fields ...PartitionField) PartitionSpec {
 	fieldCopies := make([]PartitionField, len(fields))
 	for i, field := range fields {
-		fieldCopies[i] = field
-		if field.SourceIDs != nil {
-			fieldCopies[i].SourceIDs = slices.Clone(field.SourceIDs)
-		}
+		fieldCopies[i] = clonePartitionField(field)
 	}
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approve. Reverting the source makes `TestPartitionSpecFieldGettersReturnCopies` fail on both leaks (`SourceIDs` → `[97 2]`, `NumBuckets` → 256), and it covers the constructor plus all three getters. `clonePartitionField` is complete for today's transforms (`BucketTransform`/`TruncateTransform` are the only stateful ones; value transforms are copies already), and the unexported `escapedName` survives because `initialize()` fills it before `sourceIdToFields` is built. `FieldsBySourceID` keeps its nil-vs-empty contract. Bonus: this also closes a hole in `table.clonePartitionSpec`, which built its clones from `spec.Field(i)` and shared `*BucketTransform` pointers across metadata copies.

Non-blocking: the transform type switch fails open on a future pointer transform — a short comment on `clonePartitionField` keeps it honest.

PR #1541, opened 2026-07-25
fix(view): return defensive metadata copies

fallintoplace:
## What changed

- Deep-copy versions returned by CurrentVersion and Versions, including summaries, representations, and default namespaces.
- Deep-copy schemas returned by CurrentSchema, Schemas, and SchemasByID.
- Copy nested struct, list, and map types, identifier IDs, and mutable default values.
- Clone version logs and properties.
- Keep CurrentSchemaID on an internal lookup path to avoid an unnecessary defensive copy.

## Why

View metadata getters previously exposed internal pointers, maps, and slices. Callers could mutate persisted metadata without using the metadata builder. SchemasByID cloned only the map, so mutating a returned schema could also invalidate the cached schema lookup.

Regression coverage mutates every affected getter and separately verifies nested schema types, byte-slice defaults, and binary literal defaults.

## Testing

- go test ./view
- go vet ./view

### DIFF
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -365,10 +365,15 @@ func (m *metadata) Equals(other Metadata) bool {
 func (m *metadata) FormatVersion() int         { return m.FormatVersionValue }
 func (m *metadata) ViewUUID() uuid.UUID        { return *m.UUID }
 func (m *metadata) Location() string           { return m.Loc }
-func (m *metadata) Versions() []*Version       { return m.VersionList }
-func (m *metadata) Schemas() []*iceberg.Schema { return m.SchemaList }
+func (m *metadata) Versions() []*Version       { return cloneSlice(m.VersionList) }
+func (m *metadata) Schemas() []*iceberg.Schema { return cloneSchemas(m.SchemaList) }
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM. All seven mutable-returning getters are covered (`Versions`, `Schemas`, `SchemasByID`, `CurrentVersion`, `CurrentSchema`, `VersionLog`, `Properties`), and the copies are deep — `Version.Clone` handles `Summary`/`Representations`/`DefaultNamespace`, and `cloneSchema` → `cloneNestedFields` → `cloneSchemaType`/`cloneDefault` covers nested struct/list/map plus `[]byte`/`Binary`/`Fixed` defaults. The old `SchemasByID` map-only clone leaked pointers into the `sync.OnceValue` index — now fixed. `TestMetadataGettersReturnDefensiveCopies` and `TestCloneSchemaCopiesNestedValues` fail when reverted; helpers match the table equivalents from #1444.

Non-blocking: those ~90 lines of clone helpers are now duplicated verbatim in `table/` and `view/` — hoist into `internal` in a follow-up. And `metadata_builder_test.go:328` drops the schema-ID check (`Schema.Equals` ignores ID) — please also assert the IDs.

PR #1542, opened 2026-07-25
fix(view): report metadata writer close errors

fallintoplace:
## What changed

Move metadata writing into a helper with a named error result. CreateView now returns a View only after the metadata writer closes successfully.

## Why

The previous code deferred CheckedClose against a local error variable and then returned explicit values. Updating that local variable during the deferred close could not change the error already selected for return. Filesystems that flush or finalize writes during Close could therefore report a failure while CreateView returned success.

The regression test exercises the public CreateView path with a registered filesystem whose writer fails during Close. It verifies that the error is returned and no View is produced.

## Testing

- go test ./view
- go vet ./view

### DIFF
--- a/view/view.go
+++ b/view/view.go
@@ -193,15 +193,23 @@ func createView(
 		return nil, errors.New("filesystem IO does not support writing")
 	}
 
-	out, err := wfs.Create(metadataLocation)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create view metadata file: %w", err)
-	}
-	defer internal.CheckedClose(out, &err)
-
-	if _, err := out.Write(viewMetadataBytes); err != nil {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Diagnosis is correct: the old `defer internal.CheckedClose(out, &err)` wrote to a local `err` after the return values were already chosen, so a close failure was dropped and `CreateView` returned a `View` anyway. Extracting `writeViewMetadata` with a named result is the minimal fix, and the `View` is built only after the helper succeeds. `CheckedClose` uses `errors.Join` (no masking) and `out` is closed exactly once. `TestCreateViewReturnsMetadataCloseError` fails on main.

Non-blocking: `require.EqualError` pins the exact mock message — `require.ErrorContains` would survive a future wrap.

PR #1543, opened 2026-07-25
fix(view): protect identifiers from mutation

fallintoplace:
## What changed

Copy view identifiers when constructing a View and when returning Identifier.

## Why

View identifiers are slices. New previously retained the caller-owned slice, and Identifier returned that same backing array. Either path allowed external mutation to change the identity of an existing View and affect equality or catalog behavior.

The regression test mutates both the constructor input and the getter result.

## Testing

- go test ./view
- go vet ./view
- go test ./...

### DIFF
--- a/view/view_test.go
+++ b/view/view_test.go
@@ -145,3 +145,15 @@ func TestCreateViewReturnsMetadataCloseError(t *testing.T) {
 	require.EqualError(t, err, "error on close")
 	require.Nil(t, createdView)
 }
+
+func (t *ViewTestSuite) TestIdentifierReturnsDefensiveCopy() {
+	identifier := []string{"namespace", "view"}
+	vw := New(identifier, nil, "metadata.json")
+
+	identifier[0] = "changed-input"
+	t.Equal([]string{"namespace", "view"}, vw.Identifier())
+
+	returned := vw.Identifier()
+	returned[1] = "changed-output"
+	t.Equal([]string{"namespace", "view"}, vw.Identifier())
+}
--- a/view/view.go
+++ b/view/view.go
@@ -44,7 +44,7 @@ func (t View) Equals(other View) bool {
 		t.metadata.Equals(other.metadata)
 }
 
-func (t View) Identifier() table.Identifier     { return t.identifier }
+func (t View) Identifier() table.Identifier     { return slices.Clone(t.identifier) }
 func (t View) Metadata() Metadata               { return t.metadata }
 func (t View) MetadataLocation() string         { return t.metadataLocation }
 func (t View) CurrentVersion() *Version         { return t.metadata.CurrentVersion() }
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
LGTM. Both aliasing paths are covered (constructor input and `Identifier()` result), mirroring `table.Table` on main (`slices.Clone` in both `New` and `Identifier()`), so it follows the existing convention. `Equals` reads the private field, so nothing else leaks the backing array. `TestIdentifierReturnsDefensiveCopy` fails when reverted and asserts both mutation directions. Minimal, no notes.

zeroshade:
Approved and good to go — this just needs a rebase on main. #1542 landed in view/view.go just ahead of it, so GitHub now flags a conflict. Could you rebase, @fallintoplace? I will merge once it is green.

view/view_test.go
```diff
@@ -145,3 +145,15 @@ func TestCreateViewReturnsMetadataCloseError(t *testing.T) {
 	require.EqualError(t, err, "error on close")
 	require.Nil(t, createdView)
 }
+
+func (t *ViewTestSuite) TestIdentifierReturnsDefensiveCopy() {
+	identifier := []string{"namespace", "view"}
+	vw := New(identifier, nil, "metadata.json")
```
laskoviymishka, line 151: the `nil` `Metadata` is safe today since we only touch `Identifier()`, but it's a sharp edge — any future assertion that reaches into the metadata (equality, `CurrentVersion()`) nil-derefs with an opaque panic.

I'd either build a minimal `Metadata` or drop a one-line comment that nil is intentional because this test only exercises identifier isolation. wdyt?

view/view_test.go
```diff
@@ -145,3 +145,15 @@ func TestCreateViewReturnsMetadataCloseError(t *testing.T) {
 	require.EqualError(t, err, "error on close")
 	require.Nil(t, createdView)
 }
+
+func (t *ViewTestSuite) TestIdentifierReturnsDefensiveCopy() {
```
laskoviymishka, line 149: tiny thing while we're here — this suite method landed after the standalone `TestCreateViewReturnsMetadataCloseError` func. Testify wires it up fine either way, but it'd read cleaner grouped with the other `*ViewTestSuite` methods above.

laskoviymishka (approved):
Agreed with the approval here — bringing `view.View` in line with `table.Table`'s `slices.Clone` on both `New` and `Identifier()` closes the last constructor that was aliasing the caller's slice, and the test asserting both directions is the right proof.

One small thing, not blocking: the new test passes `nil` for `Metadata`. It's safe as written since we only call `Identifier()`, but it's a sharp edge for anyone who extends the test later — a method call into the metadata would nil-deref. I'd either construct a minimal `Metadata` or leave a one-line note that nil is intentional here.

Otherwise this is good to merge.

PR #1544, opened 2026-07-25
fix(manifest): return defensive metadata copies

fallintoplace:
## What changed

- Clone key metadata returned by ManifestFile.
- Deep-copy partition summaries, including bounds and nullable ContainsNaN values.
- Copy the nullable first row ID before returning it.
- Reuse the partition summary copy helper for builder inputs and getter outputs.

## Why

ManifestFile getters exposed slices and pointers owned by decoded manifest-list entries. A caller could mutate key metadata, partition bounds, summary flags, or first row IDs and silently change later planning or manifest rewrite behavior.

The regression test mutates every affected getter and verifies that subsequent reads remain unchanged.

## Testing

- go test .
- go vet .
- go test ./...

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -152,8 +152,19 @@ func (b *ManifestBuilder) Partitions(p []FieldSummary) *ManifestBuilder {
 		return b
 	}
 
-	copied := make([]FieldSummary, len(p))
-	for i, partition := range p {
+	copied := cloneFieldSummaries(p)
+	b.m.PartitionList = &copied
+
+	return b
+}
+
+func cloneFieldSummaries(partitions []FieldSummary) []FieldSummary {
+	if partitions == nil {
+		return nil
+	}
+
+	copied := make([]FieldSummary, len(partitions))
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. The copy is deep enough — `cloneFieldSummaries` covers bounds plus the nullable `ContainsNaN`, i.e. every mutable member of `FieldSummary` — and `KeyMetadata`/`Partitions`/`FirstRowID` are the only `ManifestFile` methods that expose a reference, so the surface is fully covered. Reusing the helper for the builder keeps `TestManifestBuilderPartitionsCopiesInput` honest, and `TestManifestFileGettersReturnDefensiveCopies` fails on all three getters when reverted.

Non-blocking follow-ups: `dataFile` in the same file still returns internals (`SplitOffsets`, `EqualityFieldIDs`, `Partition`, stats maps, …) — bigger surface, worth not leaving the two halves divergent. And `manifest.go:802-805` now calls `FirstRowID()` twice; hoist into a local.

PR #1545, opened 2026-07-25
fix(io): reject invalid MemFS offsets

fallintoplace:
## What changed

Validate negative ReadAt offsets and unknown Seek whence values in MemFS. Both cases now return a PathError wrapping fs.ErrInvalid, and a rejected seek leaves the current position unchanged.

## Why

ReadAt previously sliced with a negative index and panicked. Seek accepted any unknown whence value as an absolute position of zero, silently moving the file cursor instead of reporting invalid input.

The regression test covers both public file operations and verifies that an invalid seek does not alter the cursor.

## Testing

- go test ./io
- go vet ./io
- go test ./...

### DIFF
--- a/io/mem.go
+++ b/io/mem.go
@@ -165,6 +165,8 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
 		abs = f.pos + offset
 	case io.SeekEnd:
 		abs = int64(len(f.data)) + offset
+	default:
+		return 0, &fs.PathError{Op: "seek", Path: f.name, Err: fs.ErrInvalid}
 	}
 	if abs < 0 {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Both guards are correct and neither rejects a valid read. Confirmed the negative `ReadAt` was a real panic, not a wrong result — on main the new test dies with `slice bounds out of range [-1:]` at `mem.go:181`. Validation is appropriately narrow: the `off >= len(data)` EOF, partial-read `io.EOF`, and seek-past-end all keep working, and returning before `f.pos = abs` is what leaves the cursor intact — the test pins that with a `SeekCurrent` read-back. LGTM.

Minor nit: stdlib uses `Op: "read"` for `ReadAt` `PathError`s rather than `"readat"`.

PR #1546, opened 2026-07-25
fix(exprs): protect bound literal sets from mutation

fallintoplace:
## Summary

Return detached literal sets from bound predicates and keep the scan-planning fast path allocation-free.

## Why

Set has a mutating Add method. Returning the internal set let callers and expression visitors change an already-bound predicate, which could affect later evaluation and projection results.

## What changed

- BoundSetPredicate.Literals returns a fresh mutable set with cloned byte-backed literals.
- The public VisitBoundPredicate helper gives visitors detached sets.
- Trusted Iceberg-Go evaluators use an internal capability-gated borrowed path for zero-copy visits and treat those sets as read-only.

## Tests

- Covers binary, fixed, geometry, and geography literal ownership
- Covers mutation through Literals and a custom visitor
- go test .
- go test ./table
- go vet .
- go test ./...

### DIFF
--- a/bound_set_literal_view.go
+++ b/bound_set_literal_view.go
@@ -1,0 +1,91 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package iceberg
+
+import (
+	"fmt"
+	"slices"
+)
+
+type readOnlyLiteralSet struct {
+	Set[Literal]
+}
+
+func (readOnlyLiteralSet) Add(...Literal) {
+	panic(fmt.Errorf("%w: cannot add to read-only literal set", ErrInvalidArgument))
+}
+
+func (s readOnlyLiteralSet) Equals(other Set[Literal]) bool {
+	if rhs, ok := other.(readOnlyLiteralSet); ok {
+		other = rhs.Set
+	}
+
+	return s.Set.Equals(other)
+}
+
+func cloneBoundLiteral(lit Literal) Literal {
+	switch lit := lit.(type) {
+	case BinaryLiteral:
+		return BinaryLiteral(slices.Clone(lit))
+	case FixedLiteral:
+		return FixedLiteral(slices.Clone(lit))
+	case GeoLiteral:
+		lit.val = slices.Clone(lit.val)
+
+		return lit
+	default:
+		return lit
+	}
+}
+
+func (s readOnlyLiteralSet) Members() []Literal {
+	members := s.Set.Members()
+	for i, literal := range members {
+		members[i] = cloneBoundLiteral(literal)
+	}
+
+	return members
+}
+
+func (s readOnlyLiteralSet) All(fn func(Literal) bool) bool {
+	return s.Set.All(func(literal Literal) bool {
+		return fn(cloneBoundLiteral(literal))
+	})
+}
+
+type boundSetLiteralRef interface {
+	boundSetLiteralsRef() Set[Literal]
+}
+
+func (bsp *boundSetPredicate[T]) boundSetLiteralsRef() Set[Literal] {
+	return bsp.lits
+}
+
+func boundSetLiteralsForVisit(predicate BoundPredicate) Set[Literal] {
+	if ref, ok := predicate.(boundSetLiteralRef); ok {
+		return ref.boundSetLiteralsRef()
+	}
+
+	setPredicate, ok := predicate.(BoundSetPredicate)
+	if !ok {
+		panic(fmt.Errorf("%w: %s predicate %T does not implement BoundSetPredicate",
+			ErrNotImplemented, predicate.Op(), predicate))
+	}
+
+	return setPredicate.Literals()
+}
--- a/bound_set_literal_view.go
+++ b/bound_set_literal_view.go
@@ -22,22 +22,6 @@ import (
 	"slices"
 )
 
-type readOnlyLiteralSet struct {
-	Set[Literal]
-}
-
-func (readOnlyLiteralSet) Add(...Literal) {
-	panic(fmt.Errorf("%w: cannot add to read-only literal set", ErrInvalidArgument))
-}
-
-func (s readOnlyLiteralSet) Equals(other Set[Literal]) bool {
-	if rhs, ok := other.(readOnlyLiteralSet); ok {
-		other = rhs.Set
-	}
-
-	return s.Set.Equals(other)
-}
-
 func cloneBoundLiteral(lit Literal) Literal {
 	switch lit := lit.(type) {
 	case BinaryLiteral:
@@ -53,19 +37,13 @@ func cloneBoundLiteral(lit Literal) Literal {
 	}
 }
 
-func (s readOnlyLiteralSet) Members() []Literal {
-	members := s.Set.Members()
-	for i, literal := range members {
-		members[i] = cloneBoundLiteral(literal)
+func cloneBoundLiteralSet(lits Set[Literal]) Set[Literal] {
+	cloned := newLiteralSet()
+	for _, literal := range lits.Members() {
+		cloned.Add(cloneBoundLiteral(literal))
 	}
 
-	return members
-}
-
-func (s readOnlyLiteralSet) All(fn func(Literal) bool) bool {
-	return s.Set.All(func(literal Literal) bool {
-		return fn(cloneBoundLiteral(literal))
-	})
+	return cloned
 }
 
 type boundSetLiteralRef interface {
--- a/bound_set_literal_view_test.go
+++ b/bound_set_literal_view_test.go
@@ -27,7 +27,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestBoundSetPredicateLiteralViewClonesMutableMembers(t *testing.T) {
+func TestBoundSetPredicateLiteralsCloneMutableMembers(t *testing.T) {
 	t.Parallel()
 
 	geometry, err := iceberg.GeometryTypeOf("srid:4326")
@@ -85,7 +85,7 @@ func TestBoundSetPredicateLiteralViewClonesMutableMembers(t *testing.T) {
 				value[0] ^= 0xff
 			}
 			for _, expected := range expectedLiterals {
-				assert.True(t, literals.Contains(expected))
+				assert.True(t, bound.(iceberg.BoundSetPredicate).Literals().Contains(expected))
 			}
 
 			assert.True(t, literals.All(func(literal iceberg.Literal) bool {
@@ -99,7 +99,7 @@ func TestBoundSetPredicateLiteralViewClonesMutableMembers(t *testing.T) {
 				return true
 			}))
 			for _, expected := range expectedLiterals {
-				assert.True(t, literals.Contains(expected))
+				assert.True(t, bound.(iceberg.BoundSetPredicate).Literals().Contains(expected))
 			}
 		})
 	}
--- a/bound_set_literal_view_test.go
+++ b/bound_set_literal_view_test.go
@@ -1,0 +1,106 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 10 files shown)


zeroshade (changes requested):
The hole is real and the fix is functionally correct — `Set[Literal]` exposes `Add`, so `Literals()` was handing callers a writable handle on a bound predicate's state, and both new subtests fail on main. The `members[i] = …` rewrite is safe since `literalSet.Members()` already returns a fresh slice.

My concern is where the cost lands. `Literals()` isn't a cold accessor — `visitors.go:116/118` calls it inside `visitBoundPredicate`, and `inclusiveMetricsEval.Eval` runs `VisitExpr` once per `DataFile` (`table/evaluators.go:793,809`), with `manifestEvalVisitor` per manifest and `bloomPredicateCollector` per row group on top. Benchmarked on a 50-literal binary `IN`: base `1.09 ns/op, 0 allocs` → this PR `~5548 ns/op, 8968 B, 210 allocs`. On a 50k-file scan that's ~275ms and ~450MB of extra garbage per `IN` predicate — a pure regression on the scan-planning hot path for a hardening change.

Suggest splitting the two concerns: (1) `Add`-based mutation is the real issue and is cheap — a read-only view (or a set whose `Add` is rejected) is one small alloc, not 210; (2) mutating the bytes inside a `Binary`/`Fixed`/`GeoLiteral` should be prevented by cloning **once** in `createBoundSetPredicate` while building `typedSet`, so the bound predicate owns its bytes for its lifetime at zero per-call cost. Either or both gets the same isolation with a hot path that still doesn't allocate.

Also, a test gap: only the `BinaryLiteral` branch of the clone switch is exercised; `FixedLiteral` and `GeoLiteral` (the one with a hand-rolled field assignment) aren't. Happy to re-review promptly once the copy moves off the per-file path.

exprs.go
```diff
@@ -1057,7 +1062,38 @@ func (bsp *boundSetPredicate[T]) AsUnbound(r Reference, lits []Literal) UnboundP
...
+
+type readOnlyLiteralSet struct {
+	Set[Literal]
+}
+
+func (readOnlyLiteralSet) Add(...Literal) {
+	panic("cannot add to read-only literal set")
+}
+
+func (s readOnlyLiteralSet) Equals(other Set[Literal]) bool {
```
laskoviymishka: This override makes `readOnlyLiteralSet.Equals(raw)` symmetric, but only from this side. The raw `literalSet.Equals` still asserts `other.(literalSet)` and returns false when it gets a `readOnlyLiteralSet`.

So now that `Literals()` hands back the wrapper, any external caller doing `mySet.Equals(pred.Literals())` gets false even for identical contents, while the reverse direction is fine — an asymmetric-equality trap for anything using these as dedup/cache/equivalence keys. I'd mirror this unwrap in the raw `literalSet.Equals` too so both directions agree. wdyt?

bound_set_literal_view.go
```diff
@@ -0,0 +1,49 @@
...
+
+type boundSetLiteralRef interface {
+	boundSetLiteralsRef() Set[Literal]
+}
+
+func (bsp *boundSetPredicate[T]) boundSetLiteralsRef() Set[Literal] {
+	return bsp.lits
+}
+
+func boundSetLiteralsForVisit(predicate BoundPredicate) Set[Literal] {
```
laskoviymishka, line 57: This is the clever bit that keeps the hot path zero-alloc, and for our built-in visitors it's safe since they only ever call `Contains`/`Len`/`All`. But it also hands the raw, mutable `bsp.lits` straight to every `VisitIn`/`VisitNotIn` — including third-party `BoundBooleanExprVisitor` impls, who have no signal that calling `Add` on the `Set[Literal]` they were given mutates the predicate's internal state and aliases the cloned bytes.

So the wrapper closes the hole for `Literals()` callers and the bypass reopens it for visitor callers. I'd either narrow the bypass to the read-only surface the built-ins actually need, or document on `BoundBooleanExprVisitor.VisitIn`/`VisitNotIn` that the `Set[Literal]` must be treated as read-only. Leaning toward the doc contract since it's cheap — how do you want to frame it?

visitors.go
```diff
@@ -134,9 +134,9 @@ func visitBoolExpr[T any](e BooleanExpression, visitor BooleanExprVisitor[T]) T
 func VisitBoundPredicate[T any](e BoundPredicate, visitor BoundBooleanExprVisitor[T]) T {
 	switch e.Op() {
 	case OpIn:
-		return visitor.VisitIn(e.Term(), e.(BoundSetPredicate).Literals())
+		return visitor.VisitIn(e.Term(), boundSetLiteralsForVisit(e))
```
laskoviymishka: `VisitBoundPredicate` routes through the fast path now, but `columnNameTranslator.VisitBound` and `sanitizeVisitor.VisitBound` have their own type switch and still call `p.Literals().Members()` directly, so they clone every byte-backed literal. `TranslateColumnNames` runs per data file on schema-evolution scans, so that's the same per-file O(N) clone regression from last round, just on a different path than the metrics eval that was benchmarked.

Either thread `boundSetLiteralsForVisit` through here too, or if the clone is genuinely required because the `AsUnbound` recipient may mutate the returned slice, I'd drop a comment saying so. wdyt?

exprs.go
```diff
@@ -1057,7 +1062,38 @@ func (bsp *boundSetPredicate[T]) AsUnbound(r Reference, lits []Literal) UnboundP
...
-	return bsp.lits
+	return bsp.literalView
+}
+
+type readOnlyLiteralSet struct {
+	Set[Literal]
+}
+
+func (readOnlyLiteralSet) Add(...Literal) {
+	panic("cannot add to read-only literal set")
```
laskoviymishka: Every other deliberate panic in this package goes through `panic(fmt.Errorf("%w: ...", ErrXxx))`, so `VisitExpr`'s recover can type-check it and `errors.Is` works. This one's a bare string, so a recover lands in `case string` and loses the error type. I'd make it a package sentinel (`fmt.Errorf("%w: cannot add to read-only literal set", ErrInvalidArgument)`) and panic that, then switch the test to `PanicsWithError`.

bound_set_literal_view.go
```diff
@@ -0,0 +1,49 @@
...
+	if ref, ok := predicate.(boundSetLiteralRef); ok {
+		return ref.boundSetLiteralsRef()
+	}
+
+	return predicate.(BoundSetPredicate).Literals()
```
laskoviymishka: The fallback assertion is unguarded, so a custom `BoundPredicate` carrying `OpIn`/`OpNotIn` without implementing `BoundSetPredicate` panics here with a bare `interface conversion: ...`. It can only reach here via `VisitBoundPredicate`, which is recoverable, but the message is opaque. I'd go comma-ok and panic with a wrapped `ErrNotImplemented` naming the op and concrete type.

bound_set_literal_view_test.go
```diff
@@ -0,0 +1,100 @@
...
+	geometry, err := iceberg.GeometryTypeOf("srid:4326")
+	require.NoError(t, err)
+
+	tests := []struct {
+		name string
+		typ  iceberg.Type
+	}{
+		{name: "binary", typ: iceberg.PrimitiveTypes.Binary},
+		{name: "fixed", typ: iceberg.FixedTypeOf(16)},
+		{name: "geometry", typ: geometry},
```
laskoviymishka, line 44: `GeographyType` binds to `[]byte` and produces a `GeoLiteral` through the exact same clone branch as geometry, but neither this table nor the sibling in `exprs_test.go` exercises it. Low risk since it's the same path, but a `{name: "geography", typ: iceberg.GeographyTypeOf("srid:4326")}` row confirms the wiring end to end. Worth adding while we're here.

bound_set_literal_view_test.go
```diff
@@ -0,0 +1,100 @@
...
+				require.True(t, ok)
+				value[0] ^= 0xff
+			}
+			for _, expected := range expectedLiterals {
+				assert.True(t, literals.Contains(expected))
+			}
+
+			assert.True(t, literals.All(func(literal iceberg.Literal) bool {
+				value, ok := literal.Any().([]byte)
+				require.True(t, ok)
```
laskoviymishka: `require` inside the `All` callback calls `t.FailNow()` → `runtime.Goexit()`; if `All` ever dispatched the callback on another goroutine, that would exit the wrong goroutine and the failure wouldn't surface. `assert.True` is the safe call inside a callback.

exprs_test.go
```diff
@@ -472,6 +473,88 @@ func TestInNotInSimplifications(t *testing.T) {
...
+		}
+	})
+
+	t.Run("access bound literals without allocating", func(t *testing.T) {
+		isin := iceberg.IsIn(iceberg.Reference("foo"), "hello", "world")
+		bound, err := isin.(iceberg.UnboundPredicate).Bind(tableSchemaSimple, true)
+		require.NoError(t, err)
+		bsp := bound.(iceberg.BoundSetPredicate)
+
+		assert.Zero(t, testing.AllocsPerRun(100, func() {
```
laskoviymishka: This asserts `bsp.Literals()` doesn't allocate, which is trivially true since `literalView` is pre-stored — it doesn't cover the thing that actually matters, which is `boundSetLiteralsForVisit(pred).Contains(val)` staying zero-alloc end to end. I'd point `AllocsPerRun` at a `VisitBoundPredicate` call with a small `BoundBooleanExprVisitor` stub that calls `Contains` in `VisitIn`, so the test guards the regression that was actually flagged. wdyt?

bound_set_literal_view.go
```diff
@@ -0,0 +1,49 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package iceberg
+
+func (s readOnlyLiteralSet) Members() []Literal {
```
laskoviymishka: Tiny one — `readOnlyLiteralSet` is defined in `exprs.go` (struct, `Add`, `Equals`) but its `Members`/`All` and the `boundSetLiteralRef` mechanism live here, and the split doesn't track an obvious boundary. I'd consolidate the type plus `cloneBoundLiteral` into one file so it reads in one place. Non-blocking.

laskoviymishka (changes requested):
the read-only view plus cloning once in `createBoundSetPredicate` is the right shape here, and it directly addresses the regression from the last round — the clone is off the `inclusiveMetricsEval` path now that `VisitIn`/`VisitNotIn` go through `boundSetLiteralsForVisit`. Good revision.

I'd hold it before merging though, on one conceptual thing and one concrete bug.

The fast path is what keeps the hot path zero-alloc, but it also hands the raw, mutable `bsp.lits` straight to every `VisitIn`/`VisitNotIn` — including third-party `BoundBooleanExprVisitor` implementations. Our built-in visitors only ever read (`Contains`/`Len`/`All`), so they're fine, but a downstream visitor has no signal that calling `Add` on the `Set[Literal]` it was handed mutates the predicate's internal state. So the wrapper closes the hole for `Literals()` callers and the bypass reopens it for visitor callers. I'd either narrow the bypass to the read-only surface the built-ins need, or document on `VisitIn`/`VisitNotIn` that the set must be treated as read-only.

The concrete one: `readOnlyLiteralSet.Equals` only unwraps from its own side. The raw `literalSet.Equals` still asserts `other.(literalSet)` and returns false for a `readOnlyLiteralSet`, so now that `Literals()` returns the wrapper, `rawSet.Equals(pred.Literals())` is false even for identical contents — an asymmetric-equality trap for anything using these as dedup or cache keys. Mirroring the unwrap in the raw `Equals` fixes it.

The rest — the per-file clone still in `columnNameTranslator`/`sanitizeVisitor`, the bare-string panic, geography test coverage, and the zero-alloc test only covering the accessor rather than the visit path — I've left inline.

Once the visitor exposure and `Equals` are sorted I'm happy to take another pass promptly.

laskoviymishka (approved):
👍 now!

zeroshade (approved):
Re-reviewed the "detach public bound literal sets" revision. The performance regression from my
first round is resolved and the design is clean:

- The per-file clone is off the hot path: the metrics evaluators (`manifestEvalVisitor`,
  `inclusiveMetricsEval`, `strictMetricsEval`, `bloomPredicateCollector`) and `exprEvaluator` now
  dispatch through `VisitBoundPredicateRef`/borrowed, which returns the predicate's own `bsp.lits`
  with no allocation. That was my core concern — `Literals()` cloning per `DataFile` on scan planning.
- Bytes are cloned once at bind time in `createBoundSetPredicate` (`cloneBoundLiteral`), so the bound
  predicate owns its bytes for its lifetime. Right place for it.
- `Literals()` returns a fully detached clone, so external callers can't mutate internal state via
  `Add` or by writing into `[]byte`-backed literals. Since it's a plain `literalSet` again (no
  wrapper), the asymmetric-`Equals` trap from the previous revision is gone.
- The borrow path is gated by `internal.BoundPredicateRef`, which external modules can't construct,
  so third-party visitors are forced onto the detached `VisitBoundPredicate` path. Clean way to keep
  the zero-copy path in-module only.
- Coverage now hits the `FixedLiteral` and `GeoLiteral` clone branches plus external-visitor detachment.

Two non-blocking notes:

1. Nothing guards the zero-alloc property that fixes the regression — a future change could silently
   route the hot path back through the cloning `Literals()`. Worth an `AllocsPerRun` test asserting
   `VisitBoundPredicateRef` with a `Contains`-in-`VisitIn` stub stays at 0 allocs; that's also what
   would have caught the original regression.
2. `Literals()` now clones, so the transform-projection sites still on `p.Literals().Members()`
   (`transforms.go:196`, `1156`, and `setApplyTransform` at `1282`) do a per-projection clone they
   didn't before. These run per partition spec during planning, not per file, so it's not a hot-path
   regression — but routing them through `boundSetLiteralsForVisit` too would keep the clone strictly
   for external callers and is a touch cheaper (`setApplyTransform` overwrites each element anyway).

Approving — the blocking regression is fixed and CI is green.

PR #1547, opened 2026-07-25
fix(name_mapping): isolate updated mappings

fallintoplace:
## What changed

Clone mapped field IDs and name slices while building the result of UpdateNameMapping. Fields without IDs now retain the independently visited child results as well.

## Why

The update visitor reused FieldID pointers from the input mapping. Mutating the returned mapping could therefore change the original. Appending a renamed field to a name slice with spare capacity could also write into the input backing array during the update itself.

The regression test covers both alias paths across parent and nested fields.

## Testing

- go test .
- go vet .
- go test ./...

### DIFF
--- a/name_mapping.go
+++ b/name_mapping.go
@@ -182,16 +182,20 @@ func (u *updateNameMappingVisitor) Fields(st []MappedField, fieldResults []Mappe
 
 func (u *updateNameMappingVisitor) Field(field MappedField, fieldResult []MappedField) MappedField {
 	if field.FieldID == nil {
-		return field
+		return MappedField{
+			Names:  slices.Clone(field.Names),
+			Fields: fieldResult,
+		}
 	}
 
-	fieldNames := field.Names
+	fieldID := *field.FieldID
+	fieldNames := slices.Clone(field.Names)
 	if update, exists := u.updates[*field.FieldID]; exists && !slices.Contains(fieldNames, update.Name) {
 		fieldNames = append(fieldNames, update.Name)
 	}
 
 	return MappedField{
-		FieldID: field.FieldID,
+		FieldID: &fieldID,
 		Names:   fieldNames,
 		Fields:  u.addNewFields(fieldResult, *field.FieldID),
 	}
--- a/name_mapping.go
+++ b/name_mapping.go
@@ -204,7 +204,8 @@ func (u *updateNameMappingVisitor) Field(field MappedField, fieldResult []Mapped
 func (u *updateNameMappingVisitor) removeReassignedNames(field MappedField, assignments map[string]int) *MappedField {
 	removedNames := make(map[string]struct{})
 	for _, name := range field.Names {
-		if assignedID, exists := assignments[name]; exists && assignedID != *field.FieldID {
+		assignedID, exists := assignments[name]
+		if exists && (field.FieldID == nil || assignedID != *field.FieldID) {
 			removedNames[name] = struct{}{}
 		}
 	}
--- a/name_mapping_test.go
+++ b/name_mapping_test.go
@@ -184,6 +184,38 @@ func TestUpdateNameMapping(t *testing.T) {
 		assert.Equal(t, originalMapping, result)
 	})
 
+	t.Run("result does not alias the original", func(t *testing.T) {
+		fieldID, childID := 1, 2
+		parentNames := make([]string, 2)
+		parentNames[0] = "parent"
+		original := iceberg.NameMapping{{
+			FieldID: &fieldID,
+			Names:   parentNames[:1],
+			Fields: []iceberg.MappedField{{
+				FieldID: &childID,
+				Names:   []string{"child"},
+			}},
+		}}
+
+		result, err := iceberg.UpdateNameMapping(
+			original,
+			map[int]iceberg.NestedField{1: {ID: 1, Name: "renamed-parent"}},
+			map[int][]iceberg.NestedField{},
+		)
+		require.NoError(t, err)
+		assert.Empty(t, parentNames[1])
+
+		*result[0].FieldID = 10
+		result[0].Names[0] = "changed-parent"
+		*result[0].Fields[0].FieldID = 20
+		result[0].Fields[0].Names[0] = "changed-child"
+
+		assert.Equal(t, 1, *original[0].FieldID)
+		assert.Equal(t, []string{"parent"}, original[0].Names)
+		assert.Equal(t, 2, *original[0].Fields[0].FieldID)
+		assert.Equal(t, []string{"child"}, original[0].Fields[0].Names)
+	})
+
 	t.Run("update mapping with updates and adds", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


zeroshade (commented):
The isolation fix itself is right and the test is well constructed — `parentNames[:1]` with spare capacity pins the actual corruption (the `append` writing into the caller's backing array), and the `*int` field IDs are no longer shared; reverting the source fails the new subtest on both.

Two things before merge:
1. The `FieldID == nil` branch now returns `fieldResult` instead of the original `field`, which changes output, not just ownership: for `[anon] -> ?([child] -> 2)` with `updates{2: child_renamed}`, main yields `([child] -> 2)` and this branch yields `([child, child_renamed] -> 2)`. That's a fix (renames under an anonymous parent were silently dropped), but it's untested — please add a case so a regression can't slip back.
2. Adjacent gap in the same path: `removeReassignedNames` (`name_mapping.go:207`) dereferences `*field.FieldID` unconditionally, so an anonymous field whose name collides with a rename target makes `UpdateNameMapping` return a nil-pointer error (reproduced on this head). Pre-existing, but it's a three-line guard in code you're already touching.

name_mapping.go
```diff
@@ -182,16 +182,20 @@ func (u *updateNameMappingVisitor) Fields(st []MappedField, fieldResults []Mappe
 
 func (u *updateNameMappingVisitor) Field(field MappedField, fieldResult []MappedField) MappedField {
 	if field.FieldID == nil {
-		return field
+		return MappedField{
```
laskoviymishka, line 185: This branch quietly does two things now: it isolates `Names`, and it swaps `field.Fields` for `fieldResult`, which propagates child visit results the old `return field` silently dropped. The second part is a real fix, renames under an anonymous parent used to be lost, and it looks right.

Only thing worth confirming: PyIceberg's `_UpdateMapping.field()` explicitly returns the original field unchanged when `field_id is None`, so this diverges from that reference. I'd confirm the propagation is intended and drop a one-line comment, since it's not obvious from the diff that this changes output. wdyt?

name_mapping.go
```diff
@@ -182,16 +182,20 @@ func (u *updateNameMappingVisitor) Fields(st []MappedField, fieldResults []Mappe
...
+			Names:  slices.Clone(field.Names),
+			Fields: fieldResult,
+		}
 	}
 
-	fieldNames := field.Names
+	fieldID := *field.FieldID
```
laskoviymishka, line 191: Now that `fieldID` holds the owned copy, the two remaining `*field.FieldID` derefs (the `u.updates` lookup just below and the `addNewFields` call) could just use `fieldID`. Same value, keeps the "own a fresh int" intent unambiguous. Fine as a follow-up.

name_mapping.go
```diff
@@ -200,7 +204,8 @@ func (u *updateNameMappingVisitor) Field(field MappedField, fieldResult []Mapped
 func (u *updateNameMappingVisitor) removeReassignedNames(field MappedField, assignments map[string]int) *MappedField {
 	removedNames := make(map[string]struct{})
 	for _, name := range field.Names {
-		if assignedID, exists := assignments[name]; exists && assignedID != *field.FieldID {
+		assignedID, exists := assignments[name]
+		if exists && (field.FieldID == nil || assignedID != *field.FieldID) {
```
laskoviymishka, line 208: The nil-guard here is the right fix. While we're in this function though, it's the one spot the isolation change doesn't reach: the return a few lines down builds a fresh `MappedField` but copies `field.FieldID` by pointer and assigns `field.Fields` uncloned, so a caller that mutated the result could reach back into the original.

It's latent today since every call site passes an already-visited copy, but it's inconsistent with what the rest of the PR sets up. I'd clone both here too. wdyt?

name_mapping_test.go
```diff
@@ -184,6 +184,91 @@ func TestUpdateNameMapping(t *testing.T) {
 		assert.Equal(t, originalMapping, result)
 	})
 
+	t.Run("result does not alias the original", func(t *testing.T) {
+		fieldID, childID, anonymousChildID := 1, 2, 3
```
laskoviymishka, line 188: The "remove reassigned name" subtest just below uses `makeID(1)` for this. Could use `makeID(1)`/`makeID(2)`/`makeID(3)` inline here too instead of the three named vars, to match the other subtests. Follow-up whenever.

name_mapping_test.go
```diff
@@ -184,6 +184,91 @@ func TestUpdateNameMapping(t *testing.T) {
...
+			},
+		}
+
+		result, err := iceberg.UpdateNameMapping(
+			original,
+			map[int]iceberg.NestedField{1: {ID: 1, Name: "renamed-parent"}},
+			map[int][]iceberg.NestedField{},
+		)
+		require.NoError(t, err)
+		assert.Empty(t, parentNames[1])
```
laskoviymishka, line 215: Two small robustness follow-ups in this block. `assert.Empty(t, parentNames[1])` is the assertion that proves `slices.Clone` happened, and if it ever fails the `*result[0].FieldID` derefs right below panic instead of reporting cleanly, so I'd make it `require`. Same reason to add a `require.NotNil` on the result FieldIDs before mutating them, so a regression that returned a nil FieldID fails the test rather than crashing it.

name_mapping_test.go
```diff
@@ -184,6 +184,91 @@ func TestUpdateNameMapping(t *testing.T) {
...
+
+		*result[0].FieldID = 10
```
laskoviymishka, line 217: This proves element-level isolation (cloned child FieldID and Names), which is great, but it doesn't catch `Fields` slice-header aliasing: appending to `result[0].Fields` with spare capacity could still reach into the original. That's the same class of bug this PR fixes on `Names`, and it's the gap that would let the `removeReassignedNames` point above slip through. Worth an append-based assertion, or sizing an inner `Fields` slice with spare capacity so the shared-backing path is detectable.

laskoviymishka (approved):
This is in good shape and good to merge from my side. The aliasing fix is the right root cause, and the new "result does not alias the original" test is nicely constructed. The head already picked up the nil-guard in `removeReassignedNames` and the anonymous-parent coverage, so nothing here blocks.

A few follow-ups worth doing, either now or after merge:

`removeReassignedNames` is the single path the isolation fix doesn't actually reach. It builds a fresh `MappedField` but copies `field.FieldID` by pointer and returns `field.Fields` uncloned, which is exactly the aliasing this PR closes everywhere else. It's latent today since every call site hands it an already-visited copy, so it's not a bug you can trigger, just an inconsistency with the goal that's cheap to close while the code is open.

The `FieldID == nil` branch is worth a second look too. Swapping `field.Fields` for `fieldResult` is a real fix (renames under an anonymous parent used to be dropped), but it also diverges from PyIceberg, which returns the anonymous field untouched. I'd confirm that's intended and leave a one-line comment, since the diff doesn't make it obvious that output changed.

The rest are small follow-ups on the tests. None of it blocks.

PR #1548, opened 2026-07-25
fix(table): reject reserved metadata field IDs before reassignment

### LINKED ISSUE #1107: Bug(schema): AssignFreshSchemaIDs Java Divergance

tanmayrauth:
NewMetadata reassigns schema IDs before the compatibility check runs, so a user-supplied reserved ID (e.g. _row_id) was silently remapped to a fresh value and later resolved to the wrong physical column against Java-written files. Validate the user schema against the full reserved range before reassignIDs. The AddSchema check stays narrow so internal position-delete schemas (file_path/pos) still pass.

Fixes: #1107

### DIFF
--- a/metadata_columns.go
+++ b/metadata_columns.go
@@ -17,7 +17,23 @@
 
 package iceberg
 
-import "slices"
+import (
+	"math"
+	"slices"
+)
+
+// MaxStructFieldID is the largest field ID a user-supplied schema may assign.
+// Field IDs greater than this are reserved by the spec for metadata columns
+// (_row_id, _last_updated_sequence_number, _file, _pos, _deleted, _spec_id,
+// _partition, ...). Java: TypeUtil.MAX_STRUCT_FIELD_ID (Integer.MAX_VALUE - 200).
+const MaxStructFieldID = math.MaxInt32 - 200
+
+// IsReservedFieldID reports whether fieldID falls in the range the spec reserves
+// for metadata columns. User schemas must not assign IDs in this range; table
+// schema field IDs must not exceed MaxStructFieldID.
+func IsReservedFieldID(fieldID int) bool {
+	return fieldID > MaxStructFieldID
+}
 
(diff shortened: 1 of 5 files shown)


tanmayrauth:
@laskoviymishka @zeroshade  can you please review it when you get chance?

zeroshade (approved):
Boundary is right: the spec's Reserved Field IDs section says tables "must not use field ids greater than 2147483447", so `id > MaxInt32-200` is exact — `MaxStructFieldID` allowed, `+1` rejected (probed both on this branch). Placing the check before `reassignIDs` is correct, and load paths (`ParseMetadata*`) are untouched so Java-written tables carrying reserved IDs still load. `-1` placeholders still create fine, duplicate IDs already panic in `NewSchemaWithIdentifiers`, and every local create path funnels through `NewMetadata`. Tests cover nested struct/list/map plus the end-to-end #1107 regression.

Non-blocking: (1) `metadata_columns.go` cites `Java: TypeUtil.MAX_STRUCT_FIELD_ID`, which doesn't exist in apache/iceberg — please cite the spec's Reserved Field IDs section instead. (2) This closes `NewMetadata` but `AssignFreshSchemaIDs`/`MetadataBuilder.AddSchema` still admit non-lineage reserved IDs — worth noting in the issue that it's narrowed, not fully closed.

PR #1555, opened 2026-07-27
fix(table): invalidate shredded variant bounds only on the field's own residual

nssalian:
`collectVariantBounds` dropped a shredded field's bound when any ancestor object's residual `value` column was non-null. Now it drops only when the field's own `value` residual is non-null.

Per the Iceberg spec [Bounds for Variant](https://github.com/apache/iceberg/blob/main/format/spec.md), a field's bounds must be accurate for its non-null values. A row whose ancestor is a scalar has the descendant field absent (null), so it is outside the bound and cannot invalidate it. This matches Java `ParquetMetrics.object()`/`value()`, where the object residual is trusted and only the field's own residual invalidates its bound.

Serialization unchanged; only which fields get a bound. Follow-up to [#1478](https://github.com/apache/iceberg-go/pull/1478).

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -1099,8 +1099,8 @@ func (p parquetFormat) collectVariantBounds(meta *metadata.FileMetaData, arrowSc
 		}
 		for _, lf := range enumerateVariantLeaves([]string{f.Name}, vt.TypedValue()) {
 			byTyped[lf.typedPath] = &leafAgg{leaf: lf, parentID: parentID}
-			for _, vp := range lf.valuePaths {
-				residualParent[vp] = struct{}{}
+			if lf.ownResidual != "" {
+				residualParent[lf.ownResidual] = struct{}{}
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Semantics check out against both the spec and Java. VariantShredding.md: the `value` column of a partially shredded object must never contain fields represented by `typed_value` columns, so a non-null ancestor residual means the shredded descendant is absent for that row and outside "accurate for all non-null values". Java's `MetricsVariantVisitor.object()` discards its own `valueResult` and `value()` invalidates only on that level's own residual — your `ownResidual` scoping (including "no `value` column ⇒ trust typed") is a 1:1 match. `TestCollectVariantBoundsNestedAncestorResidual`'s three-way rewrite is the right test; the ancestor-only-non-null case can't pass under the old code.

Non-blocking: no case covers a shredded leaf with `typed_value` but no `value` child (`ownResidual == ""`), whose behavior also changed to "always trusted" — worth a case so a refactor can't silently flip it.

PR #1558, opened 2026-07-27
fix(table): honor initial defaults in read filters

### LINKED ISSUE #1557: bug(table): filters on evolved columns ignore initial-default for old files
### Apache Iceberg version

`main` at `9097100f4aa3bae419644e3b1f25d3fe29ce8fb7`. The affected code is also present in v0.6.0.

### Please describe the bug

A filtered Arrow scan can silently drop rows written before a column was added, even when that column's `InitialDefault` satisfies the filter.

For example, start with a v3 table containing two rows, add optional `new_col int` with an initial default of `42`, and scan the old file:

- An unfiltered projection returns both rows with `new_col = 42`, as expected.
- A scan filtered by `new_col == 42` returns no rows and no error.
- `PlanFiles` retains the old data file, so the loss occurs during read-time filtering rather than file planning.

Expected: both old rows are returned because their projected value is `42`.

Actual: the result is empty.

The following package-local regression test reproduces the problem using the existing `buildV3TableWithRows` helper:

```go
func collectInitialDefaultScan(
	t *testing.T,
	scan *Scan,
	tasks []FileScanTask,
) (ids []int64, defaults []int32) {
	t.Helper()

	_, records, err := scan.ReadTasks(t.Context(), tasks)
	require.NoError(t, err)

	for rec, recErr := range records {
		require.NoError(t, recErr)
		require.NotNil(t, rec)
		if rec.NumRows() > 0 {
			idCol := rec.Column(0).(*array.Int64)
			defaultCol := rec.Column(1).(*array.Int32)
			for i := range int(rec.NumRows()) {
				ids = append(ids, idCol.Value(i))
				defaults = append(defaults, defaultCol.Value(i))
			}
		}
		rec.Release()
	}

	return ids, defaults
}

func TestInitialDefaultFilterDropsOldFileRepro(t *testing.T) {
	tbl := buildV3TableWithRows(t, `[{"id":1,"data":"a"},{"id":2,"data":"b"}]`)

	txn := tbl.NewTransaction()
	require.NoError(t, txn.UpdateSchema(true, false).
		AddColumn(
			[]string{"new_col"},
			iceberg.PrimitiveTypes.Int32,
			"",
			false,
			iceberg.Int32Literal(42),
		).
		Commit())
	var err error
	tbl, err = txn.Commit(t.Context())
	require.NoError(t, err)

	control := tbl.Scan(WithSelectedFields("id", "new_col"))
	controlTasks, err := control.PlanFiles(t.Context())
	require.NoError(t, err)
	require.Len(t, controlTasks, 1)
	controlIDs, controlDefaults := collectInitialDefaultScan(t, control, controlTasks)
	require.Equal(t, []int64{1, 2}, controlIDs)
	require.Equal(t, []int32{42, 42}, controlDefaults)

	matching := tbl.Scan(
		WithSelectedFields("id", "new_col"),
		WithRowFilter(iceberg.EqualTo(iceberg.Reference("new_col"), int32(42))),
	)
	matchingTasks, err := matching.PlanFiles(t.Context())
	require.NoError(t, err)
	require.Len(t, matchingTasks, 1, "planning retains the old file")
	matchingIDs, matchingDefaults := collectInitialDefaultScan(t, matching, matchingTasks)

	t.Logf("control IDs=%v defaults=%v", controlIDs, controlDefaults)
	t.Logf("matching-filter IDs=%v defaults=%v", matchingIDs, matchingDefaults)
	require.Equal(t, controlIDs, matchingIDs,
		"every old row projects new_col=42 and must satisfy new_col == 42")
	require.Equal(t, controlDefaults, matchingDefaults)
}
```

Required imports:

```go
import (
	"testing"

	"github.com/apache/arrow-go/v18/arrow/array"
)
```

Run:

```text
go test ./table -run '^TestInitialDefaultFilterDropsOldFileRepro$' -count=2 -v
```

Observed on both runs:

```text
control IDs=[1 2] defaults=[42 42]
matching-filter IDs=[] defaults=[]
```

### Root cause

[`TranslateColumnNames`](https://github.com/apache/iceberg-go/blob/9097100f4aa3bae419644e3b1f25d3fe29ce8fb7/visitors.go#L499-L508) treats a field missing from the file schema as null: it folds `IsNull` to true and every other predicate to `AlwaysFalse`, without consulting the evolved field's `InitialDefault`.

[`getRecordFilter`](https://github.com/apache/iceberg-go/blob/9097100f4aa3bae419644e3b1f25d3fe29ce8fb7/table/arrow_scanner.go#L728-L740) turns that `AlwaysFalse` into `dropFile = true`, and this happens [before projection](https://github.com/apache/iceberg-go/blob/9097100f4aa3bae419644e3b1f25d3fe29ce8fb7/table/arrow_scanner.go#L1113-L1139). Projection later [materializes the correct `InitialDefault`](https://github.com/apache/iceberg-go/blob/9097100f4aa3bae419644e3b1f25d3fe29ce8fb7/table/arrow_utils.go#L1090-L1096), but the file has already been discarded.

### Suggested fix

When translating a bound predicate for a field absent from the file schema, evaluate the predicate against the field's `InitialDefault` and fold it to the corresponding constant result. Preserve the existing null behavior only when no non-null initial default applies.

A regression matrix should cover matching and non-matching predicates, null predicates, and both old and newly written files.

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently

mattfaltyn:
## Summary

- evaluate predicates against a top-level field's `initial-default` when the field is absent from an older file schema
- preserve existing null behavior for fields without defaults, nested-field semantics beneath nullable parents, and BBox handling
- use one hex-first binary/fixed default codec for filter translation and Arrow projection, while retaining read compatibility with legacy iceberg-go v0.6.0 base64 metadata
- serialize newly written binary/fixed initial and write defaults as spec-compliant lowercase hex
- conservatively keep files matchable when invalid metadata supplies a non-null geometry/geography default instead of aborting the scan

Fixes #1557.

## Testing

- `go test ./... -count=1`
- `go test -race -count=1 ./codec/... ./table/...`
- focused translator, byte-codec, JSON serialization, Arrow projection, and scan regressions
- `golangci-lint run --timeout=10m` (repository-pinned v2.8.0)
- `GOOS=linux GOARCH=s390x CGO_ENABLED=0 go build ./...`
- `git diff --check origin/main...HEAD`

Coverage includes boolean and temporal defaults, spec hex and legacy base64 byte defaults, fixed-width disambiguation, geometry/geography fallback, and filter/projection byte consistency. GitHub unit, integration (Spark 3.5 and 4.0), audit, security, license, website, and cross-compile checks are green.


visitors.go
```diff
@@ -496,27 +504,92 @@ func (columnNameTranslator) VisitUnbound(pred UnboundPredicate) BooleanExpressio
...
+	case BoundLiteralPredicate:
+		return p.AsUnbound(ref, p.Literal())
+	case BoundSetPredicate:
+		return p.AsUnbound(ref, p.Literals().Members())
+	default:
+		panic(fmt.Errorf("%w: unsupported predicate: %s", ErrNotImplemented, pred))
+	}
+}
+
+func initialDefaultLiteral(field NestedField) (Literal, error) {
```
tanmayrauth, line 610: This decoder duplicates the initial-default decode logic in table/arrow_utils.go:822 (defaultToScalar) — base64 for binary/fixed, Decimal/string for decimal, etc. Both paths must decode a given initial-default to the same value:  this one feeds the filter, defaultToScalar feeds the column the scan materializes. If they ever diverge (e.g. one switches binary from base64 to hex), a filter on a missing-column-with-default silently returns wrong rows — matching  rows pruned, or non-matching rows returned — with no error. They agree today, but there's no shared helper and no test pinning the equivalence. A full dedupe is awkward since `table` imports `iceberg` and not the reverse, so a consistency test (for each type, assert initialDefaultLiteral(field).Any() equals what defaultToScalar produces) is the cheap way to lock it in.

table/scanner_internal_test.go
```diff
@@ -969,3 +969,72 @@ func TestProjectionV3SchemaAlreadyHasRowID(t *testing.T) {
 		}
 	})
 }
+
```
tanmayrauth, line 1152: This end-to-end scan test only covers an Int32 default. The binary/decimal/uuid cases live in TestTranslateColumnNamesMissingFieldInitialDefault, which stops at the TranslateColumnNames layer and never runs the scan — so the filter-vs-fill consistency (the failure mode above) isn't verified end-to-end for exactly the types most prone to diverge. Adding a binary and a decimal column here would exercise both the filter path and defaultToScalar in the  same test and guard against silent divergence.

visitors_test.go
```diff
@@ -233,6 +233,119 @@ func TestAlwaysExprBinding(t *testing.T) {
 	}
 }
 
+func TestTranslateColumnNamesMissingFieldInitialDefault(t *testing.T) {
```
tanmayrauth, line 249: No coverage for date/time/timestamp defaults, which take the json.Marshal → decodeValue path (visitors.go:550). A default written by iceberg-go arrives numeric and resolves via  Int64Literal.To(DateType); one written by another engine arrives as an ISO string ("2017-11-16") and resolves via StringLiteral.To(DateType). Both are realistic and neither is tested — a temporal default that decodes wrong would mis-prune old files. Worth a date or timestamp case  here.

visitors.go
```diff
@@ -496,27 +504,92 @@ func (columnNameTranslator) VisitUnbound(pred UnboundPredicate) BooleanExpressio
...
+			return AlwaysFalse{}
+		}
+
+		eval, err := ExpressionEvaluator(NewSchema(0, field),
+			unbindPredicate(pred, Reference(field.Name)), true)
+		if err != nil {
+			panic(err)
+		}
+
+		lit, err := initialDefaultLiteral(field)
```
tanmayrauth, line 684: When a default fails to decode, this panics with a bare, context-free error (same for the ExpressionEvaluator panic above and the eval panic below) — VisitExpr recovers it into the error that aborts the scan at arrow_scanner.go:734, so the user sees a raw decode failure with no hint which column or default triggered it. Wrap it with the field, e.g. panic(fmt.Errorf("initial-default for column %q (id %d): %w", field.Name, field.ID, err)), and add the same context to the bare returns in initialDefaultLiteral (the base64, json.Marshal, and decodeValue paths). Aborting on an undecodable default is the right behavior — this is just about making the failure diagnosable.

tanmayrauth (commented):
Core fix looks right and lines up with the arrow fill path. A maintainability note on the duplicated decoder, two test-gaps, and a small error-context nit. Details inline.

zeroshade (changes requested):
Logic and shape are right — this mirrors PyIceberg's `_ColumnNameTranslator`, keeps the no-default behavior intact, and `panic(err)` → `VisitExpr` recover is the existing convention in this file. `unbindPredicate` dedup is a nice side effect, and the tests are real (reverting `visitors.go` fails both `TestTranslateColumnNamesMissingFieldInitialDefault` and `TestArrowScanFiltersMissingColumnInitialDefault`).

Blocking on the binary/geo decoding:

1. `initialDefaultLiteral` base64-decodes `binary`/`fixed` defaults, but spec Appendix D and Java's `SingleValueParser` use **hex** (base16). With a Java/Spark-written `"initial-default": "000102ff"` the base64 decode *succeeds* into garbage, so `EqualTo(col, X'000102ff')` translates to `AlwaysFalse()` and `X'd34d35d367df'` to `AlwaysTrue()` — silently wrong rows, no error. The new `"matching binary metadata default": "AQID"` case cements the deviation rather than catching it.
2. `geometry`/`geography` defaults are WKT strings per Appendix D (`POINT (30 10)`), not bytes — base64-decoding that fails (`illegal base64 data at input byte 5`), so a valid table hard-errors on any filter over that column.

This file already has the spec-correct decoder: `convertValue` in `expr_json.go` hex-decodes binary/fixed (with a comment saying exactly that), reachable via the `decodeValue` you're already calling. Simplest fix is to drop the binary/fixed/geo base64 special case and let `decodeValue` handle the JSON forms (natives are stored via `defaultValue.Any()`). The base64 assumption comes from `defaultToScalar` (table/arrow_utils.go), which has the same bug on the projection path — please don't propagate it; a follow-up to converge `defaultToScalar` on hex would be good.

Also needs a rebase — `visitors.go` now conflicts with `SanitizeExpression` on main, whose `sanitizeVisitor.VisitUnbound` repeats the unbind switch you factored out (reuse `unbindPredicate`).

visitors_test.go
mattfaltyn: Agreed. I added date cases for a numeric Iceberg Go default and an ISO metadata value. Both verify that the missing-column filter folds correctly.

table/scanner_internal_test.go
mattfaltyn: Good point. I added a decimal default to the end-to-end scan so filtering and materialization run together. Binary stays in focused translation coverage because the projection decoder has a separate bug.

visitors.go
mattfaltyn: Thank you. Errors from evaluator construction, default decoding and evaluation now include the column name and field ID. An invalid hex test verifies that context.

mattfaltyn:
Thank you for the precise spec correction. The fixes are now pushed at `e7b72dc`. Binary and fixed metadata use the existing hex decoder. Native byte defaults stay typed and geo defaults remain null-only. I merged current main without rewriting reviewed history and preserved sanitizer masking through `unbindPredicate`. I also limited default folding to top-level fields because nested values remain row-dependent beneath nullable parents. The projection decoder stays out of scope. I will merge main again after #1564 lands because current main does not compile without its unrelated one-line fix.

visitors.go
```diff
@@ -496,29 +503,83 @@ func (columnNameTranslator) VisitUnbound(pred UnboundPredicate) BooleanExpressio
...
+		return p.AsUnbound(ref, p.Literal())
+	case BoundSetPredicate:
+		return p.AsUnbound(ref, p.Literals().Members())
+	default:
+		panic(fmt.Errorf("%w: unsupported predicate: %s", ErrNotImplemented, pred))
+	}
+}
+
+func initialDefaultLiteral(field NestedField) (Literal, error) {
+	switch field.Type.(type) {
```
laskoviymishka: Taking geo out of here was right for the base64 problem, but nothing picks it up now. `StringLiteral.To` has no `Geometry`/`Geography` arm, and `convertValue` in `expr_json.go` doesn't either, so a WKT default like `"POINT (30 10)"` goes `json.Marshal` to `decodeValue` to `ErrBadCast`, and the recovered panic turns into a failed `TranslateColumnNames` for the whole scan.

That's a step back from where we started. Before this PR the column folded to `AlwaysFalse`/`AlwaysTrue` and the read went through; now it errors out. Reachability is narrow since Java's `SingleValueParser` doesn't emit geo initial-defaults today, so it takes a non-Java V3 writer to trigger, but I'd still not ship the hard error.

Simplest thing I see is keeping the old conservative behavior for types we can't decode: if `initialDefaultLiteral` can't produce a literal for geo, treat the column as "could match" instead of panicking. wdyt?

visitors.go
```diff
@@ -496,29 +503,83 @@ func (columnNameTranslator) VisitUnbound(pred UnboundPredicate) BooleanExpressio
...
+	case BinaryType, FixedType:
+		if val, ok := field.InitialDefault.([]byte); ok {
```
laskoviymishka, line 613: This does the right thing, but not for the reason it looks like. `InitialDefault` on a JSON-loaded schema is always a Go `string`, so the `[]byte` assertion misses and we fall through to `json.Marshal` + `decodeValue`, which hex-decodes. That's exactly the behavior we wanted, and the `"000102ff"` / `"010203"` cases lock it in.

My worry is the next reader. Nothing says the fallthrough is load-bearing, so adding an `else { return error }` looks like obvious tidying and would break every metadata-sourced binary default. A line like `// metadata defaults arrive as hex strings; fall through to decodeValue` would keep it safe.

visitors.go
```diff
@@ -496,29 +503,83 @@ func (columnNameTranslator) VisitUnbound(pred UnboundPredicate) BooleanExpressio
...
 	fileColName, found := c.fileSchema.FindColumnName(pred.Term().Ref().Field().ID)
 	if !found {
 		// in the case of schema evolution, the column might not be present
 		// in the file schema when reading older data
-		if pred.Op() == OpIsNull {
+		field := pred.Ref().Field()
+		// A nested field can still be null when an optional parent is null, so
+		// its default is not a file-wide constant. Preserve the existing
+		// missing-column behavior until translation has row-level parent state.
+		if field.InitialDefault == nil || len(pred.Ref().PosPath()) > 1 {
```
laskoviymishka, line 660: Appreciate the comment, it states the limitation clearly, and this leaves nested fields where they already were instead of regressing them. Fine by me for this PR.

Worth noting for later: when every ancestor is required the child default *is* a file-wide constant, so `WHERE addr.country = 'US'` over a `REQUIRED STRUCT addr` could fold safely and this bails out anyway. Narrowing the guard to "bail only if some ancestor is optional" would recover that. Happy for it to be a follow-up.

visitors.go
```diff
@@ -496,29 +503,83 @@ func (columnNameTranslator) VisitUnbound(pred UnboundPredicate) BooleanExpressio
...
+			return fmt.Errorf("initial-default for column %q (id %d): %w",
+				field.Name, field.ID, err)
+		}
+		eval, err := ExpressionEvaluator(NewSchema(0, field),
+			unbindPredicate(pred, Reference(field.Name)), true)
+		if err != nil {
+			panic(withContext(err))
+		}
+
+		lit, err := initialDefaultLiteral(field)
```
laskoviymishka, line 684: Circling back on the `defaultToScalar` note I called a follow-up last round: I think it got more urgent, not less. With the filter path on hex, a Java-written `"initial-default": "000102ff"` folds the predicate to `AlwaysTrue` correctly here, and then `defaultToScalar` in `table/arrow_utils.go` base64-decodes the same string to fill the column. `000102ff` is entirely base64-alphabet characters, so it doesn't even error, it decodes into six garbage bytes.

So the scan returns rows whose binary column doesn't match the predicate that admitted them, with no error anywhere. Nothing this PR broke, but this PR is what makes it observable. Your call whether converging `defaultToScalar` on hex rides along here or gets filed, but I wouldn't let it sit.

table/scanner_internal_test.go
```diff
@@ -1148,3 +1149,105 @@ func TestProjectionV3SchemaAlreadyHasRowID(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			scan := tbl.Scan(
+				WithSelectedFields("id", "new_col"),
+				WithRowFilter(tt.filter),
+			)
+			tasks, err := scan.PlanFiles(t.Context())
+			require.NoError(t, err)
+			require.Len(t, tasks, 1, "manifest planning must retain the old file")
```
laskoviymishka, line 1277: I don't think this assertion can fail. When `new_col` is missing from the file there are no stats for it, so the inclusive-metrics evaluator falls through to `rowsMightMatch` and the file passes whether or not the fold was right. It can't distinguish correct pruning from absent stats.

Same shape one level down: for the `expected: 0` cases the fold to `AlwaysFalse` drops the file, `ToArrowTable` returns zero chunks, and the `== 42` loop never executes. The default-fill check is only really exercised by the two `expected: 2` cases.

If we want the planning-level claim to bite, asserting on the translated expression or the record filter rather than task count would do it.

laskoviymishka (changes requested):
Nice progress. `sanitizeVisitor.VisitBound` delegating through `unbindPredicate` is exactly what I was after, and the binary/fixed path is on hex now with `"000102ff"` / `"010203"` in the test matrix holding it there. Items 1 and 3 from last round are done.

Item 2 is still open, and I think it went the wrong direction. Deleting the geo special case removed the base64 hard-error, but nothing picked up the WKT form: `StringLiteral.To` has no `Geometry`/`Geography` case and neither does `convertValue`, so `"POINT (30 10)"` falls through to `ErrBadCast` and the recovered panic fails `TranslateColumnNames` for the whole scan. Pre-PR that column quietly folded to `AlwaysFalse`/`AlwaysTrue` and the read succeeded, so this is a new failure mode rather than a leftover one. It takes a non-Java V3 writer to reach, since `SingleValueParser` doesn't emit geo initial-defaults today, so I'm not worried about it in the field. I'd still rather have a conservative fallback (undecodable default means "could match") than a panic.

Separately, and to be clear this one is not on you: CI is red on every Go job with `table/arrow_scanner.go:1027:35: undefined: internal`. That came in with #1537 and was already fixed on main by #1564; your merge base just predates the fix. A rebase onto main picks it up and should clear the conflict state too. One consequence worth flagging: because `table` never compiles, `TestArrowScanFiltersMissingColumnInitialDefault` has never actually executed. The `iceberg` package builds fine, so the `visitors_test.go` matrix is genuinely passing.

On `defaultToScalar`, I called it a follow-up last round. Now that the filter path is on hex I'd bump it up: a Java-written `"000102ff"` folds correctly here and then decodes into six garbage bytes on the projection side, because that string happens to be valid base64. Rows come back that don't match the predicate that admitted them. Your call whether it rides along here or gets filed, but it shouldn't sit.

What I'd want before merge:

- rebase onto main so the branch builds and the new `table` test actually runs
- geo defaults stop hard-erroring the scan
- coverage for at least one geo default, plus timestamp and boolean, and one binary/fixed case through the Arrow projection path (the end-to-end test only covers Int32 and Decimal today, which is exactly why the `defaultToScalar` mismatch is invisible)

The rest is inline and mostly small. This is close, so once the rebase and the geo path are sorted I'm happy to take another pass.

zeroshade (changes requested):
Re-reviewed. Two of my three original blockers are resolved:

- Binary/fixed defaults decode as hex now, not base64 — `initialDefaultLiteral` routes native
  `[]byte` through `BinaryLiteral.To` and metadata strings through `decodeValue` (hex), with
  `"000102ff"`/`"010203"` locking it in. (my item 1) ✔
- `sanitizeVisitor.VisitBound` delegates through `unbindPredicate`, reusing the factored-out
  switch instead of duplicating it. (my item 3) ✔

Still blocking:

1. **Geometry/geography defaults hard-error the scan** (my item 2, still open — aligns with
   @laskoviymishka's Aug 5 note). A top-level geo column with a WKT initial-default like
   `"POINT (30 10)"` has no `Geometry`/`Geography` arm in `decodeValue`, so `initialDefaultLiteral`
   returns `ErrBadCast`, the recovered panic fails `TranslateColumnNames`, and the whole scan errors.
   Pre-PR that column folded to `AlwaysFalse`/`AlwaysTrue` and the read succeeded — so this is a new
   failure mode, not a leftover one. Narrow reachability (needs a non-Java V3 writer; Java's
   `SingleValueParser` doesn't emit geo initial-defaults), but I'd keep the conservative
   "undecodable default ⇒ could match" behavior instead of panicking, and add a geo case to
   `TestTranslateColumnNamesMissingFieldInitialDefault`.

2. **Rebase onto main.** The branch is conflicting and every Go job is red on
   `table/arrow_scanner.go:1027: undefined: internal` — a pre-#1564 merge base. Because `table`
   never compiles, the new end-to-end test `TestArrowScanFiltersMissingColumnInitialDefault` has
   never actually run. Please rebase so the branch builds and that test executes in CI.

3. **`defaultToScalar` base64/hex divergence is now observable** (I called it a follow-up last
   round; it's more urgent now). With the filter path on hex, a Java-written `"000102ff"` folds the
   predicate correctly, but `defaultToScalar` in `table/arrow_utils.go` still base64-decodes the same
   string into six garbage bytes — so the scan materializes a value that doesn't match the predicate
   that admitted the file, silently and with no error. The end-to-end test only covers Int32 and
   Decimal, which is why this stays invisible. Either converge `defaultToScalar` on hex here, or file
   a linked follow-up and add a binary-default case through the Arrow projection path.

Happy to re-review promptly once the rebase lands and the geo path stops hard-erroring.

visitors.go
mattfaltyn: Implemented in 97a0c39. Missing top-level geometry/geography fields with non-null defaults now fail open to AlwaysTrue before decoding, so invalid WKT defaults cannot abort translation. Added geometry IsNull and geography NotNull regression cases. The spec still requires these defaults to be null; this is conservative interoperability only.

table/scanner_internal_test.go
mattfaltyn: Agreed on the limitation. I retained PlanFiles only as a stage-separation control showing manifest planning keeps the old file; it is not the proof of folding. The translator matrix directly asserts AlwaysTrue/AlwaysFalse, the scan result counts exercise matching and mismatching paths, and the new binary consistency test jointly checks translation and Arrow materialization.

visitors.go
mattfaltyn: Fixed in 97a0c39. Filter translation and defaultToScalar now use the same hex-first byte-default decoder, NestedField writes new binary/fixed defaults as lowercase hex, and legacy v0.6.0 base64 remains readable. TestInitialDefaultFilterProjectionConsistency verifies that the bytes admitted by the translated predicate are exactly the bytes Arrow materializes.

visitors.go
mattfaltyn: Agreed. I kept the conservative top-level-only guard unchanged in this PR. Folding nested defaults beneath all-required ancestors is a valid optimization, but it is intentionally left out of this correctness fix.

visitors.go
mattfaltyn: Addressed in 97a0c39. Metadata binary/fixed strings now take an explicit shared decoder path instead of relying on fallthrough. The comment records that spec metadata is hex and that legacy iceberg-go v0.6.0 base64 remains readable; hex wins when ambiguous.

mattfaltyn:
Ready for re-review @zeroshade @laskoviymishka.

I merged current main without rewriting reviewed history and preserved the newer BBox handling. The remaining review items are addressed in 97a0c39:

- top-level geo defaults fail open during missing-column translation instead of aborting the scan
- binary/fixed defaults use one hex-first codec across filtering and Arrow projection, with legacy iceberg-go v0.6.0 base64 compatibility
- new metadata writes serialize binary/fixed defaults as spec-compliant lowercase hex
- filter/projection consistency, timestamp, boolean, geo, legacy base64, and empty/fixed-width cases are covered
- nested required-parent folding remains intentionally out of scope

Local validation passed: go test ./..., go test -race ./codec/... ./table/..., golangci-lint v2.8.0, Linux/s390x cross-build, focused regressions, and git diff --check. All GitHub checks, including Spark 3.5 and 4.0 integrations, are green.

I replied to each open inline thread and left reviewer-owned threads unresolved.

laskoviymishka (approved):
LGTM now!

zeroshade (approved):
All three of my blockers are resolved: geo defaults fail open ahead of decoding instead of aborting the scan, filter translation and Arrow projection now share the single hex-first byte-default codec (with the consistency test locking the two paths together), and newly written binary/fixed defaults serialize as spec-compliant lowercase hex while v0.6.0 base64 metadata stays readable. The decode-error context and the widened type matrix from the earlier rounds round it out nicely. Thanks for the persistence across the rounds — this is a thorough fix.

---

zeroshade:
@mattfaltyn This PR is fully approved, but today's batch of merges to `main` (several touched `types.go`/`visitors.go`) introduced a conflict. A rebase should be quick — once CI is green again it can merge immediately.

mattfaltyn:
> @mattfaltyn This PR is fully approved, but today's batch of merges to `main` (several touched `types.go`/`visitors.go`) introduced a conflict. A rebase should be quick — once CI is green again it can merge immediately.
> 
> _Posted by an AI-assisted tool on behalf of maintainer `zeroshade`._

Thank you @zeroshade - rebased!

PR #1559, opened 2026-07-27
fix(io): strip GCS-incompatible SDK headers on reads too

iremcaginyurtturk:
## Problem

The `strip-gcs-incompatible-signed-headers` compat-mode middleware (added in #1423) removes `Amz-Sdk-Invocation-Id`, `Amz-Sdk-Request`, and `Accept-Encoding` from the SigV4 signed set only for **write** operations (`PutObject`/`UploadPart`/`CreateMultipartUpload`).

**Read** operations still send those headers signed. Against Google Cloud Storage's S3-interop endpoint (`storage.googleapis.com` + HMAC keys), a `GetObject`/`HeadObject` — e.g. loading a table's `metadata.json` during a table-exists check — is rejected with:

```
operation error S3: GetObject, StatusCode: 403 ... SignatureDoesNotMatch
```

So writes succeed but any subsequent read fails, making GCS S3-interop unusable end to end.

## Fix

- Strip `Amz-Sdk-Invocation-Id` / `Amz-Sdk-Request` / `Accept-Encoding` before signing on **every** operation (reads included), so none of them are in the SigV4 signed set.
- Re-add `Accept-Encoding: identity` in a middleware inserted **after** the Signing step, so it is present on the wire but **not** part of the signature. This keeps GCS happy while still suppressing the HTTP transport's transparent gzip, so read responses aren't silently decompressed (which would drop `Content-Length`).

Only affects the S3-compat path (`s3.compat-mode=true`); the normal AWS S3 path is unchanged.

### DIFF
--- a/io/gocloud/s3.go
+++ b/io/gocloud/s3.go
@@ -31,7 +31,6 @@ import (
 	"github.com/apache/iceberg-go/io"
 	"github.com/apache/iceberg-go/utils"
 	"github.com/aws/aws-sdk-go-v2/aws"
-	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Correct fix, and the test is the right one: `TestCompatModeGetObjectStripsSignedHeaders` fails on the parent commit (`SignedHeaders=accept-encoding;amz-sdk-invocation-id;amz-sdk-request;…`) and passes here, while still asserting `Accept-Encoding: identity` reaches the wire — that pins both halves of the strip/restore pair. Still gated on `s3.compat-mode`, so the normal AWS path is unaffected, and `strip` stays after the retry middleware that sets `amz-sdk-request`. `go test ./io/...` green.

One nit worth fixing before merge: the `restore` fallback uses `stack.Finalize.Add(restore, smithymiddleware.Before)`, which inserts at index 0 of Finalize — i.e. *before* signing — silently re-signing the `Accept-Encoding` you just removed. Use `smithymiddleware.After`, or just return the `Insert` error. Minor: restore now sets `Accept-Encoding: identity` on writes too (previously deleted outright) — harmless, but the comment frames it as read-path only.

PR #1561, opened 2026-07-27
fix(catalog): add five-minute prefetch buffer for vended credentials

### LINKED ISSUE #1560: REST: Vended credentials reused up to exact expiry with no prefetch buffer
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

The REST catalog's vended-credential reuses the cached `io.IO/credentials` until the exact expiration timestamp.

For long-running pipelines, there should be a safety buffer. 

When a request is issued just before expiry, in-flight latency can push the actual `S3/GCS/ADLS` call past the token's expiry, causing intermittent failures.

## Proposed Solution

Iceberg-Java enforces a [5-minute prefetch](https://github.com/apache/iceberg/blob/60975ecd8d9352cfe62448dd59915b7d7bd57011/aws/src/main/java/org/apache/iceberg/aws/s3/VendedCredentialsProvider.java#L133) buffer for proactively refreshing before expiration. We can also do something similar.

badalprasadsingh:
## Description

Vended storage credentials were reused until their exact expiry, with no safety margin. Under in-flight request latency the underlying `S3/GCS/ADLS` call could land after expiry, causing intermittent errors and aborted transactions.

Fixes #1560 

The solution in this PR aligns the Go client with Iceberg-Java's [5-minute prefetch buffer](https://github.com/apache/iceberg/blob/60975ecd8d9352cfe62448dd59915b7d7bd57011/aws/src/main/java/org/apache/iceberg/aws/s3/VendedCredentialsProvider.java#L133) from the expiration time and refreshes proactively.


zeroshade (changes requested):
Intent is right and the arithmetic on your base is correct — reuse iff `now <= expiresAt - 5m`, the boundary reuses the cache, and a refreshed `expiresAt` can't retrigger itself for normal TTLs. `TestVendedCredsRefreshTriggeredWithinExpiryBuffer` genuinely covers it (fails if I restore `!v.now().After(v.expiresAt)`).

Blocking on the rebase, because `main` moved this code (#1477 just merged) and the naive resolution is broken:

- `loadFS` now delegates to `expired()`: `!v.expiresAt.IsZero() && v.now().After(v.expiresAt)`, where a **zero `expiresAt` deliberately means "never expires"** (plan-scoped creds — see `expiresAtFromConfig`). Folding the buffer in as `now.After(expiresAt.Add(-5m))` makes zero-time creds permanently expired; I confirmed `TestPlanScopedIOCredentialsWithoutExpiry` then fails with `vended storage credentials expired: … expired at 0001-01-01T00:00:00Z`.
- There's also a plan-scoped branch where `fetchCreds == nil` returns `ErrVendedCredentialsExpired`. The buffer must not feed that decision — those creds can't be renewed, so refusing them 5 minutes early turns usable creds into a hard error. Keep `expired()` as the hard boundary and add a separate `shouldRefresh()`/`needsRefresh()` used only where `fetchCreds != nil`.

One more while you're in here: clamp the buffer against the credential's own lifetime. A server vending a sub-5-minute TTL makes every `Table.FS()` call issue a `fetchCreds` REST round trip (serialized behind the semaphore). Java treats `expiresAt - 5m` as *prefetch* time with `staleTime = expiresAt` (keeps serving the cache), so something like `buffer = min(5m, lifetime/2)` keeps the proactive behavior without the fetch storm.

badalprasadsingh:
You were absolutely right about the credential lifetime comment, @zeroshade! I've merged `main` into this branch and addressed the review comment.

Hopefully, we can get this merged once the `CI` is fixed.

catalog/rest/vended_creds.go
```diff
@@ -114,7 +116,7 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 	}
 	defer v.mu.Release(1)
 
-	if v.cachedIO != nil && !v.expired() {
+	if v.cachedIO != nil && !v.needsRenewal() {
```
laskoviymishka, line 119: The goroutine that trips `shouldRefresh()` here holds the semaphore for the whole `fetchCreds` round-trip, so under a high-fan-out scan every other caller blocks on a credential that's still valid for up to five more minutes. Java's `CachedSupplier` serves the cached value at prefetch time and does the refresh in the background — only `staleTime` (hard expiry) actually blocks callers.

I don't think we need to port the background-refresh machinery, but I'd like us to pick deliberately: either accept the serialized blocking refresh at prefetch (simpler, no goroutine to manage) or serve the still-valid cache and let one goroutine refresh behind it. Either's fine, but it should be a choice rather than a side effect.

Whichever way we go, a concurrent test would help pin it down — N goroutines arriving inside the buffer window, asserting `fetchCreds` fires at most once. `TestVendedCredsConcurrentAccess` only covers the initial-load path today. wdyt?

catalog/rest/vended_creds.go
```diff
@@ -154,6 +156,7 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 
 	v.cachedIO = newIO
 	v.expiresAt = v.expiresAtFromConfig(config)
+	v.issuedAt = v.now()
```
laskoviymishka, line 159: `issuedAt` is the moment of first local use, not when the server issued the credential. For a cred that arrives in `props` and sits there a while before the first `loadFS`, the lifetime we compute is short, so `min(5m, lifetime/2)` clamps the buffer well below 5m — a 2m token idle for 90s gives lifetime 30s, buffer 15s.

That's still strictly better than no buffer, and you can't really squeeze 5m of headroom out of a token with 30s left. But it does mean the "5-minute prefetch" the title promises silently degrades for pre-aged creds — which is the original bug in miniature. I'd add a comment saying so, or read a server-side issued-at field if the REST response carries one. wdyt?

catalog/rest/vended_creds.go
```diff
@@ -163,12 +166,45 @@ func (v *vendedCredentialRefresher) expiredError(at time.Time) error {
 		ErrVendedCredentialsExpired, v.location, at.Format(time.RFC3339))
 }
 
-// expired reports whether the cached IO's credentials are past their expiry. A
+func (v *vendedCredentialRefresher) needsRenewal() bool {
```
laskoviymishka, line 169: Worth a doc line here: `needsRenewal()` returning true means two different things depending on the branch. For renewable creds (`fetchCreds != nil`) it's "we're in the prefetch window, go refresh"; for plan-scoped creds it's "hard expired, return the error." It reads cleanly from `loadFS`, but a one-liner spelling out the dual meaning would save the next reader the dispatch trace.

catalog/rest/vended_creds.go
```diff
@@ -163,12 +166,45 @@ func (v *vendedCredentialRefresher) expiredError(at time.Time) error {
...
+	if v.fetchCreds != nil {
+		return v.shouldRefresh()
+	}
+
+	return v.expired()
+}
+
+// expired reports whether the cached IO's credentials are past their expiry, with no safety buffer. A
```
laskoviymishka, line 177: This line's at 84 chars now, and wedging `, with no safety buffer` mid-sentence leaves the `A` dangling at the end — godoc renders the run-on as the summary, and it'll likely trip lll/gocritic. I'd split it:

```go
// expired reports whether the cached IO's credentials are past their expiry.
// Unlike needsRenewal, it applies no prefetch safety buffer.
// A zero expiresAt means "never expires" — see expiresAtFromConfig.
```

catalog/rest/vended_creds.go
```diff
@@ -163,12 +166,45 @@ func (v *vendedCredentialRefresher) expiredError(at time.Time) error {
...
+	buffer := defaultVendedCredentialsExpiryBuffer
+	if v.issuedAt.IsZero() {
+		return buffer
+	}
+
+	half := v.expiresAt.Sub(v.issuedAt) / 2
+	if half < 0 {
+		half = 0
+	}
+	if half < buffer {
```
laskoviymishka, line 201: Minor: this manual clamp can just use the builtin, which the package already leans on (`min(sleep, maxDelay)` in `scan_planning.go`):

```go
buffer = min(buffer, half)
```

gocritic's minMax will probably flag the `if` form anyway.

catalog/rest/vended_creds_test.go
```diff
@@ -336,6 +340,220 @@ func TestVendedCredsServerExpiryUsedOnRefresh(t *testing.T) {
...
+
+			r := newTestRefresher(func(ctx context.Context, ident []string) (iceberg.Properties, error) {
+				callCount.Add(1)
+
+				return iceberg.Properties{}, nil
+			})
+			r.nowFunc = func() time.Time { return now }
+
+			r.cachedIO = iceio.LocalFS{}
+			r.expiresAt = now.Add(tc.expiresIn)
```
laskoviymishka, line 405: This whole table leaves `r.issuedAt` at zero, so `refreshBuffer()` always takes the `issuedAt.IsZero()` early return and the boundary lands at a flat 5m. In production `issuedAt` is set after the first `loadFS`, so a 5m-TTL token clamps to `min(5m, 2.5m) = 2.5m` and the reuse/refresh boundary sits at 2.5m, not the 5m this case asserts.

The clamp path itself is covered in `TestVendedCredsRefreshBuffer`, so it's not a hole in coverage overall — just this table testing the zero-`issuedAt` boundary rather than the production one. A parallel sub-test seeding `r.issuedAt = now` with a 10m TTL would exercise the real boundary.

catalog/rest/vended_creds_test.go
```diff
@@ -336,6 +340,220 @@ func TestVendedCredsServerExpiryUsedOnRefresh(t *testing.T) {
...
+		props:    iceberg.Properties{},
+		nowFunc:  func() time.Time { return now },
+	}
+	r.cachedIO = iceio.LocalFS{}
+	r.expiresAt = now.Add(defaultVendedCredentialsExpiryBuffer - time.Minute)
+
+	got, err := r.loadFS(context.Background())
+	require.NoError(t, err,
+		"plan-scoped creds still within their hard expiry must not be refused early")
+	assert.Equal(t, r.cachedIO, got)
```
laskoviymishka, line 438: Since you mutate `nowFunc` and call `loadFS` again right after this, I'd make this `require.Equal` — if it ever fires, the non-fatal assert lets the second call run on state you didn't expect and muddies the failure. `LocalFS{}` is a value type, so `Equal` is the right comparison here.

catalog/rest/vended_creds_test.go
```diff
@@ -336,6 +340,220 @@ func TestVendedCredsServerExpiryUsedOnRefresh(t *testing.T) {
...
+	// Short-lived (2m) renewable token.
+	// Buffer : lifetime/2 (1m), so a just-issued token is served from cache instead of re-fetched at once.
+	shortTTL := 2 * time.Minute
+
+	var callCount atomic.Int32
+	r := newTestRefresher(func(ctx context.Context, ident []string) (iceberg.Properties, error) {
+		callCount.Add(1)
+
+		return iceberg.Properties{
+			keyS3TokenExpiresAtMs: strconv.FormatInt(now.Add(shortTTL).UnixMilli(), 10),
```
laskoviymishka, line 459: `fetchCreds` returns `now.Add(shortTTL)` off the outer fixed `now`, so the refreshed token expires at the original wall time no matter when it's fetched. After the refresh at T+90s the new cred has ~30s of life and a 15s buffer, so a third `loadFS` at T+105s would immediately re-trigger — invisible here only because there's no third call.

`TestVendedCredsIssuedAtUpdatedOnRefreshPreventsSelfRetrigger` already does this right with `clock().Add(...)`. I'd mirror that here (`r.nowFunc().Add(shortTTL)`) and add a third call asserting no immediate re-fetch, so the test reflects a real server minting `fetch_time + TTL`.

laskoviymishka (approved):
This is in good shape. 

Nothing here blocks merge for me, but there are a couple of things I'd like to look at

The one I keep coming back to is prefetch under concurrency. The goroutine that trips `shouldRefresh()` holds the semaphore across the whole `fetchCreds` round-trip, so every other caller blocks on a credential that's still valid for up to five more minutes. Java's `CachedSupplier` serves the cached value at prefetch time and refreshes in the background — only hard expiry blocks. I don't think we need to port the background-refresh machinery, but since this PR is chasing Java parity I'd like us to decide deliberately whether the serialized blocking refresh is what we want. Details inline.

The other is that `issuedAt` is set to local `now()` at first use, not the server's issue time. For a short-lived cred that's been sitting in `props` for a while, the lifetime we compute — and so the clamped buffer — can end up well under the five minutes the title promises. It's still strictly better than no buffer, but worth a comment so nobody assumes full 5m headroom.

The rest is doc/style polish and a few test-realism gaps, all inline, all follow ups.

zeroshade (approved):
just needs the conflicts resolved!

PR #1563, opened 2026-07-27
fix: preserve current logical schema column order in schemasToGlueColumns

### LINKED ISSUE #1562: Glue: schemasToGlueColumns reorders columns by field ID
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

When syncing an iceberg schema to a Glue table, `schemasToGlueColumns` collects columns into a Go map and sorts them by Iceberg Field ID. This produces a physical Glue column order that differs from the table's current logical schema order whenever columns have been _reordered_, _dropped_, or _added_ out of ID sequence.

We should not be tampering with the order of the columns. The current schema's logical column order should be preserved, then append the historical columns, with **no sorting**. A behavior which is similar to [Iceberg-Java](https://github.com/apache/iceberg/blob/60975ecd8d9352cfe62448dd59915b7d7bd57011/aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java#L346-L366).

badalprasadsingh:
## Description

Fixes #1562, column ordering in `schemasToGlueColumns`. This aligns with [Iceberg-Java's](https://github.com/apache/iceberg/blob/60975ecd8d9352cfe62448dd59915b7d7bd57011/aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java#L346-L366) behavior : emit current-schema columns in logical order first, then append historical columns, with **no sorting.** Added a test for it.

Don't think this should affect backward compatibility.


zeroshade (approved):
Correct and minimal in isolation. Name-keyed dedupe with current-schema order first and historical columns appended is exactly PyIceberg's `_to_columns` (insertion-ordered dict) and Java's `IcebergToGlueConverter`; dropping the `icebergFieldIDKey` sort is right since field IDs never encoded logical order. `TestSchemasToGlueColumnsPreservesLogicalOrder` fails without the change; the existing `TestSchemasToGlueColumns` (asserts full `types.Column` incl. `Comment`) still passes, and `fieldToGlueColumn` is untouched so comments/params aren't dropped.

Heads-up before merge: #1522 just merged and also rewrote `schemasToGlueColumns` (field-ID-keyed comment back-fill). Please rebase on main and confirm the comment back-fill still holds under the new name-keyed ordering — the two changes overlap in the same function, so they should be reconciled together rather than merged blind. Holding my side of the merge until then.

zeroshade:
Approved — but #1522 merged just ahead of this and also rewrote schemasToGlueColumns (it adds field-ID-keyed comment back-fill). Could you rebase on main, @badalprasadsingh, and confirm the comment back-fill still holds under the new name-keyed ordering? The two overlap in that one function, so they should be reconciled together. Happy to merge once rebased and green.

badalprasadsingh:
Thanks @zeroshade! I've rebased onto `main` and addressed the overlap with #1522. Looks good now. 

We can merge this once the `CI` passes.

catalog/glue/schema_test.go
```diff
@@ -534,3 +534,47 @@ func TestSchemasToGlueColumnsClearsPreviousIcebergComment(t *testing.T) {
...
+
+	// "address" is dropped, "email" is added and the remaining columns are reordered
+	current := iceberg.NewSchema(1,
+		iceberg.NestedField{ID: 4, Name: "email", Type: iceberg.StringType{}, Required: false},
+		iceberg.NestedField{ID: 2, Name: "name", Type: iceberg.StringType{}, Required: true},
+		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.Int64Type{}, Required: true},
+	)
+
+	metadata, err := table.NewMetadata(original, nil, table.SortOrder{}, "s3://example/path", nil)
+	assert.NoError(t, err)
```
laskoviymishka, line 553: I'd switch the five setup steps here (NewMetadata through Build) to `require.NoError`. With `assert`, a failure in `AddSchema` or `Build` doesn't stop the test — it falls through to `schemasToGlueColumns` with nil/invalid metadata and panics, so you get a confusing stack trace instead of a clean failure at the step that actually broke.

catalog/glue/schema_test.go
```diff
@@ -534,3 +534,47 @@ func TestSchemasToGlueColumnsClearsPreviousIcebergComment(t *testing.T) {
...
+
+	columns := schemasToGlueColumns(metadata, nil)
+
+	names := make([]string, len(columns))
+	for i, c := range columns {
+		names[i] = aws.ToString(c.Name)
+	}
+
+	assert.Equal(t, []string{"email", "name", "id", "address"}, names, "Glue column order must follow the current logical schema")
+	assert.Equal(t, "false", columns[3].Parameters[icebergFieldCurrentKey])
```
laskoviymishka, line 575: Same robustness thing here — if a regression returns fewer columns, `columns[3]` (and the `range 3` loop just below) panics index-out-of-range instead of failing cleanly. I'd either promote the names check on the line above to `require.Equal`, or drop a `require.Len(t, columns, 4)` in before this line so the count is guaranteed first.

catalog/glue/schema.go
```diff
@@ -43,9 +54,7 @@ func schemasToGlueColumns(metadata table.Metadata, existingColumns []types.Colum
 		}
 
 		for _, field := range schemaToGlueColumns(schema, false) {
-			if _, ok := results[aws.ToString(field.Name)]; !ok {
-				results[aws.ToString(field.Name)] = field
-			}
+			addColumnWithDedupe(field)
```
laskoviymishka, line 57: Not a bug today — `commonMetadata.Schemas()` returns insertion order, so the historical column that wins dedupe here is stable and matches Java. What I'd flag is that the `Metadata` interface doesn't promise any iteration order, so this leans on an implementation detail — a future `Schemas()` that reorders would silently change which historical column lands here.

A one-line comment noting we assume insertion order (or sorting by schema-ID before the loop) would make that explicit. Not blocking.

catalog/glue/schema.go
```diff
@@ -81,33 +90,22 @@ func schemasToGlueColumns(metadata table.Metadata, existingColumns []types.Colum
...
 
-	for name, column := range results {
-		if column.Comment != nil {
+	for i := range columns {
+		if columns[i].Comment != nil {
 			continue
 		}
 
-		fieldID := column.Parameters[icebergFieldIDKey]
+		fieldID := columns[i].Parameters[icebergFieldIDKey]
```
laskoviymishka, line 98: Clearly pre-existing and out of scope here — flagging only because this PR is the reconciliation point for the back-fill. This loop keys off `icebergFieldIDKey`, whereas Java keys existing comments by column name, so on a rename we carry the old comment forward by field ID while Java drops it. Came in with #1522, not this PR — a short comment on why we key by field ID (or a follow-up to align with Java) would be enough. Not blocking.

catalog/glue/schema_test.go
```diff
@@ -534,3 +534,47 @@ func TestSchemasToGlueColumnsClearsPreviousIcebergComment(t *testing.T) {
...
+
+	for i := range 3 {
+		assert.Equal(t, "true", columns[i].Parameters[icebergFieldCurrentKey])
+	}
+}
```
laskoviymishka, line 580: This new test covers ordering well. Since the back-fill loop now applies to appended historical columns too, I'd add a sibling case: a historical schema column with a comment that the current schema drops, then assert the appended tail column gets its comment restored. That's exactly the reconciliation @zeroshade asked to confirm, turned into a regression guard rather than a manual check.

laskoviymishka (approved):
Agreed with the read here, name-keyed dedupe with the current schema's fields first and historical columns appended is exactly PyIceberg's `_to_columns` and Java's `IcebergToGlueConverter`, and dropping the `icebergFieldIDKey` sort is right since field IDs never encoded logical order to begin with.

On the #1522 reconciliation heads-up -- I dug into the overlap and I think it holds. The comment back-fill build logic (`clearedComments`/`existingComments`) is untouched; the only thing this PR changed there is the final apply loop, which went from `for name, column := range results` (random map order, value-copy write-back) to `for i := range columns` (deterministic slice iteration, in-place mutation). That's actually strictly safer than what #1522 left in (no map-order nondeterminism and no copy-back pitfall) so the two reconcile cleanly rather than fighting.

What's left is small and all on the new test: the five setup steps use `assert.NoError`, so a builder failure falls through to a nil-metadata panic instead of failing at the step that broke — I'd make those `require.NoError`, and guard the `columns[3]`/`range 3` indexing with a `require.Len` so a regression fails cleanly instead of panicking. And since this change is the check-in point for back-fill under the new ordering, a case that exercises comment restoration on an appended historical column would turn that reconciliation into a regression guard rather than a manual check.

None of that touches the substance, so I'll leave the merge-hold with @zeroshade, this is just test cleanup.

PR #1566, opened 2026-07-28
fix(cli): report successful drop operation in cli

jx2lee:
## Summary
- Priunt success message after dropping a namespace/table
- Printt distinct success message when `drop table --purge`

## Testing
did not add test code (no exists create/rename cmd)

- `go test ./cmd/iceberg` (fails existing build error in `table/arrow_scanner.go:1027:35: undefined: internal`)

### DIFF
--- a/cmd/iceberg/main.go
+++ b/cmd/iceberg/main.go
@@ -563,6 +563,8 @@ func runDrop(ctx context.Context, output Output, cat catalog.Catalog, cmd *DropC
 		if err != nil {
 			output.Error(err)
 			osExit(1)
+
+			return
 		}
 
 		output.Text("Namespace " + cmd.Namespace.Identifier + " dropped successfully")
@@ -574,12 +576,14 @@ func runDrop(ctx context.Context, output Output, cat catalog.Catalog, cmd *DropC
 			if !ok {
 				output.Error(fmt.Errorf("catalog %s does not support purge", cat.CatalogType()))
 				osExit(1)
+
 				return
 			}
 
 			if err := purger.PurgeTable(ctx, ident); err != nil {
 				output.Error(err)
 				osExit(1)
+
 				return
 			}
 
@@ -591,6 +595,8 @@ func runDrop(ctx context.Context, output Output, cat catalog.Catalog, cmd *DropC
 		if err := cat.DropTable(ctx, ident); err != nil {
 			output.Error(err)
 			osExit(1)
+
+			return
 		}
 
 		output.Text("Table " + cmd.Table.Identifier + " dropped successfully")
--- a/cmd/iceberg/main.go
+++ b/cmd/iceberg/main.go
@@ -564,23 +564,33 @@ func runDrop(ctx context.Context, output Output, cat catalog.Catalog, cmd *DropC
 			output.Error(err)
 			osExit(1)
 		}
+
+		output.Text("Namespace " + cmd.Namespace.Identifier + " dropped successfully")
 	case cmd.Table != nil:
 		ident := catalog.ToIdentifier(cmd.Table.Identifier)
-		var err error
+
 		if cmd.Table.Purge {
-			if purger, ok := cat.(catalog.PurgeableTable); ok {
-				err = purger.PurgeTable(ctx, ident)
-			} else {
+			purger, ok := cat.(catalog.PurgeableTable)
+			if !ok {
 				output.Error(fmt.Errorf("catalog %s does not support purge", cat.CatalogType()))
 				osExit(1)
 			}
-		} else {
-			err = cat.DropTable(ctx, ident)
+
+			if err := purger.PurgeTable(ctx, ident); err != nil {
+				output.Error(err)
+				osExit(1)
+			}
+
+			output.Text("Table " + cmd.Table.Identifier + " purged successfully")
+
+			return
 		}
-		if err != nil {
+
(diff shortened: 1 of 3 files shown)


jx2lee:
CI failure isn’t caused by my changes. 🥲

cmd/iceberg/main.go
```diff
@@ -564,23 +564,33 @@ func runDrop(ctx context.Context, output Output, cat catalog.Catalog, cmd *DropC
...
+
+			return
 		}
-		if err != nil {
+
+		if err := cat.DropTable(ctx, ident); err != nil {
 			output.Error(err)
 			osExit(1)
 		}
+		output.Text("Table " + cmd.Table.Identifier + " dropped successfully")
```
laskoviymishka, line 602: There's no blank line between the `DropTable` error block and this `output.Text`, and nlreturn (on in `.golangci.yaml`) will flag it. Every other site in this function has the blank line — the namespace and purge branches both do — so this'll fail CI as-is. Just need a blank line in before line 593.

cmd/iceberg/main.go
```diff
@@ -564,23 +564,33 @@ func runDrop(ctx context.Context, output Output, cat catalog.Catalog, cmd *DropC
...
 	case cmd.Table != nil:
 		ident := catalog.ToIdentifier(cmd.Table.Identifier)
-		var err error
+
 		if cmd.Table.Purge {
-			if purger, ok := cat.(catalog.PurgeableTable); ok {
-				err = purger.PurgeTable(ctx, ident)
-			} else {
+			purger, ok := cat.(catalog.PurgeableTable)
+			if !ok {
```
laskoviymishka, line 576: The refactor to an early-return guard dropped the structural guarantee the old if/else had here. `purger` is a nil interface when `!ok`, and the only reason the `PurgeTable` call below doesn't nil-deref is that `osExit` never returns.

That's a fragile invariant to lean on — the day someone swaps in an `osExit` double that returns, this deadfalls. I'd add an explicit `return` right after `osExit(1)`, matching what the success branch already does a few lines down. wdyt?

table/arrow_scanner.go
```diff
@@ -1024,7 +1024,7 @@ func (as *arrowScan) processRecords(
 		if err != nil {
 			return err
 		}
-		out <- enumeratedRecord{Record: internal.Enumerated[arrow.RecordBatch]{
+		out <- enumeratedRecord{Record: tblutils.Enumerated[arrow.RecordBatch]{
```
laskoviymishka, line 1027: To close the loop on your PR note — this fix is correct. `tblutils` is imported at line 38 and used throughout the file, so this compiles cleanly; it was just a stale unaliased `internal.Enumerated` the rest of the file had already migrated off.

The only thing I'd change is scope: it's unrelated to the CLI messages and makes the PR non-atomic, which hurts bisection. I'd pull it into its own one-line PR against the scan-planning work so this one stays focused. Not blocking the fix itself.

laskoviymishka (changes requested):
Nice cleanup, the purge branch reads a lot better as an early-return guard than the old nested if/else, and the new messages line up with the "created successfully" / "Renamed table from..." wording we already use.

I'd hold this before merging though, on two things.

First, the blank line: nlreturn is on in `.golangci.yaml`, and there's no blank line between the `DropTable` error block and the `output.Text` right after it (main.go:593). Every other site in this function has one, so CI will fail on it as-is.

Second, none of the new messages are actually asserted anywhere. `errCapture` embeds `textOutput` and drops the text, so all three success paths — table, purge, namespace — emit to stdout and nothing checks them. Since the whole point of the PR is those messages, I'd add a `lastText` capture to `errCapture` and assert the exact string in the drop and purge tests, plus a `TestRunDropNamespace` (that path has no test at all today).

One more while we're here: the `arrow_scanner.go` change is a real fix — I checked, `tblutils` is imported at line 38 and it compiles cleanly — but it's unrelated to the CLI work and makes this non-atomic. I'd pull it into its own one-line PR against the scan-planning change so bisection stays clean.

jx2lee:
@laskoviymishka Thanks for the detailed review.

I updated the PR to add the blank line for nlreturn, added explicit returns after osExit in the purge path, and updated errCapture to capture Text output so the drop and purge success messages are now asserted.

Also, removed the unrelated arrow_scanner.go change from this PR and split it out into #1567 and added coverage for the drop namespace success message.

```bash
❯ gss
 M table/arrow_scanner.go
?? .worktree/
❯ go test ./cmd/iceberg -run '^(TestRunDropTable|TestRunDropTablePurgeSupported|TestRunDropNamespace)$'
ok  	github.com/apache/iceberg-go/cmd/iceberg	(cached)
```

 After modifying `arrow_scanner.go`, the compilation was successful, and the test results are normal.

laskoviymishka (approved):
LGTM!

I'll take a look on red-ci, in a meantime.

Thanks for contributions!

laskoviymishka:
@jx2lee -- linter need to be fixed, but PR-labelr is unrelated to your changes

jx2lee:
> @jx2lee -- linter need to be fixed, but PR-labelr is unrelated to your changes

@laskoviymishka rebase upstream main, and re-fixed nlreturn. ci is running, expected to success(hope..)

laskoviymishka:
@jx2lee - welcome to contributors gang!

PR #1567, opened 2026-07-28
fix(table): use table internal Enumerated in arrow scanner

jx2lee:
## What changed

  Use `tblutils.Enumerated` when emitting the empty final record batch from `arrowScan.processRecords`. [Refer this comment](https://github.com/apache/iceberg-go/pull/1566#discussion_r3666328662).

  ## Why

The empty-batch sentinel path should use the same `table/internal` `Enumerated` type as the rest of the arrow scanner. Referring to `internal.Enumerated` uses the wrong package qualifier and breaks the table package build.

## Testing

 - `go test ./table`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -1024,7 +1024,7 @@ func (as *arrowScan) processRecords(
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


jx2lee:
closed this by duplicated #1564.

PR #1568, opened 2026-07-28
feat(expr): Add bbox predicate evaluation for geo pruning

### LINKED ISSUE #994: feat(expr): bbox predicate evaluation for geo pruning
Parent: #989

Depends on #993. The spec only requires bbox-based pruning, not full geometric predicate evaluation — actual `ST_Intersects` / `ST_Within` is a query-engine concern. iceberg-go just needs `BBoxIntersects` plumbed through `exprs.go` and `BoundsEvaluator` so partition pruning + manifest pruning can drop files whose bbox doesn't intersect the query bbox.

Tests cover `BoundsEvaluator` with bbox predicates plus an integration test asserting a query with a non-intersecting bbox prunes all data files. Spec: [Iceberg primitive types](https://iceberg.apache.org/spec/#primitive-types).

C-Loftus:
Closes #994 

- Adds a BBoxIntersects geospatial predicate with public `BBox` and `IntersectsWithBBox` APIs.
- Wires bbox predicates through binding, visitor dispatch, column-name translation, and operation string generation.
- Implements inclusive manifest/data-file pruning by intersecting query boxes with stored Iceberg geospatial X/Y bounds.
    - Could use confirmation from a maintainer that just X/Y is appropriate. I didn't add any logic for Z/M since I presumed there is some more sophisticated logic that would be needed for that.  
- Adds tests covering predicate binding, metrics pruning, scan planning behavior, and geo-bound decoding helpers.


C-Loftus:
@laskoviymishka I will close this since it appears that tanmayrauth may have been working parallel on the same issue in https://github.com/apache/iceberg-go/pull/1597. I am assuming that was just since he was working on the other related/parent issues to it, but if there is anything I should do in the future beyond claiming the issue at https://github.com/apache/iceberg-go/issues/994 let me know. No problem on this just want to ensure I am on the same page for best practices for contribution 😃  Thanks!

PR #1570, opened 2026-07-28
ci: fix labeler config YAML indentation

laskoviymishka:
actions/labeler v6.1.0 parses .github/labeler.yml with js-yaml warnings promoted to errors. The multi-line flow-style sequence had its closing bracket indented less than its items, which js-yaml now rejects as "deficient indentation", failing the triage job on every open PR.

Convert the flow-style list to a block-style list, which parses cleanly and produces the identical config structure.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -18,12 +18,11 @@
 # Pull request labeler Github Action Configuration: https://github.com/marketplace/actions/labeler
 INFRA:
   - changed-files:
-    - any-glob-to-any-file: [
-        ".github/workflows/**/*",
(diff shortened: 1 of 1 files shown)


iremcaginyurtturk:
Thanks for picking this up @laskoviymishka! Just a heads-up that this duplicates #1565, which I opened earlier today with the same fix 😁 Happy to close mine if you’d rather go with this one

laskoviymishka:
i see, make sense.

will close in favor of #1565

iremcaginyurtturk:
Thanks @laskoviymishka! 🙏

PR #1571, opened 2026-07-28
fix(sql): validate view identifiers at API boundaries

fallintoplace:
## Summary

Validate view identifiers before SQL catalog operations do any database or filesystem work.

## Why

SQL view operations skipped the shared identifier validation used by other catalogs. Names such as .. or nested/view could reach location resolution, where path normalization changes their meaning.

## What changed

Create, load, drop, and existence checks now validate view identifiers before resolving locations or accessing storage.

## Tests

- go test ./catalog/sql
- go vet ./catalog/sql

### DIFF
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1233,10 +1233,6 @@ func (c *Catalog) CheckTableExists(ctx context.Context, identifier table.Identif
 }
 
 func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifier, props iceberg.Properties) error {
-	if err := catalog.ValidateNamespaceIdentifier(namespace); err != nil {
-		return err
-	}
-
 	for key := range props {
 		if isReservedNamespaceProperty(key) {
 			return fmt.Errorf("%w: cannot create namespace with reserved namespace property %q", iceberg.ErrInvalidArgument, key)
@@ -1676,10 +1672,6 @@ func (c *Catalog) listViewsAll(ctx context.Context, namespace table.Identifier) 
 		return nil, nil
 	}
 
-	if err := checkValidNamespace(namespace); err != nil {
-		return nil, err
-	}
-
 	ns := namespaceToString(namespace)
 	if len(namespace) > 0 {
 		var exists bool
@@ -1723,10 +1715,6 @@ func (c *Catalog) listViewsAll(ctx context.Context, namespace table.Identifier) 
 func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) error {
 	if c.isV0() {
 		return errViewsUnsupportedOnV0
-	}
-
-	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
-		return err
 	}
 
 	ns, err := c.namespaceKey(ctx, catalog.NamespaceFromIdent(identifier))
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1721,6 +1721,10 @@ func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) err
 		return errViewsUnsupportedOnV0
 	}
 
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
+		return err
+	}
+
 	ns, err := c.namespaceKey(ctx, catalog.NamespaceFromIdent(identifier))
 	if err != nil {
 		return err
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1581,6 +1581,10 @@ func (c *Catalog) CheckNamespaceExists(ctx context.Context, namespace table.Iden
 
 // CreateView creates a new view in the catalog.
 func (c *Catalog) CreateView(ctx context.Context, identifier table.Identifier, schema *iceberg.Schema, viewSQL string, props iceberg.Properties) error {
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
+		return err
+	}
+
 	if c.isV0() {
 		return errViewsUnsupportedOnV0
 	}
@@ -1713,6 +1717,10 @@ func (c *Catalog) listViewsAll(ctx context.Context, namespace table.Identifier) 
 
 // DropView deletes a view from the catalog.
 func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) error {
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
+		return err
+	}
+
 	if c.isV0() {
 		return errViewsUnsupportedOnV0
 	}
@@ -1788,6 +1796,10 @@ func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) err
 
 // CheckViewExists returns true if a view exists in the catalog.
 func (c *Catalog) CheckViewExists(ctx context.Context, identifier table.Identifier) (bool, error) {
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
+		return false, err
+	}
+
 	if c.isV0() {
 		return false, nil
 	}
@@ -1814,6 +1826,10 @@ func (c *Catalog) CheckViewExists(ctx context.Context, identifier table.Identifi
 
 // LoadView loads a view from the catalog.
 func (c *Catalog) LoadView(ctx context.Context, identifier table.Identifier) (view.Metadata, error) {
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
+		return nil, err
+	}
+
 	if c.isV0() {
 		return nil, errViewsUnsupportedOnV0
 	}
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1721,10 +1721,6 @@ func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) err
 		return errViewsUnsupportedOnV0
 	}
 
-	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
-		return err
-	}
-
 	ns, err := c.namespaceKey(ctx, catalog.NamespaceFromIdent(identifier))
 	if err != nil {
 		return err
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1672,6 +1672,12 @@ func (c *Catalog) ListViews(ctx context.Context, namespace table.Identifier) ite
 }
 
 func (c *Catalog) listViewsAll(ctx context.Context, namespace table.Identifier) ([]table.Identifier, error) {
+	if len(namespace) > 0 {
+		if err := catalog.ValidateNamespaceIdentifier(namespace); err != nil {
+			return nil, err
+		}
+	}
+
 	if c.isV0() {
 		return nil, nil
 	}
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1233,6 +1233,10 @@ func (c *Catalog) CheckTableExists(ctx context.Context, identifier table.Identif
 }
 
 func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifier, props iceberg.Properties) error {
+	if err := checkValidNamespace(namespace); err != nil {
+		return err
+	}
+
 	for key := range props {
 		if isReservedNamespaceProperty(key) {
 			return fmt.Errorf("%w: cannot create namespace with reserved namespace property %q", iceberg.ErrInvalidArgument, key)
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1233,7 +1233,7 @@ func (c *Catalog) CheckTableExists(ctx context.Context, identifier table.Identif
 }
 
 func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifier, props iceberg.Properties) error {
-	if err := checkValidNamespace(namespace); err != nil {
+	if err := catalog.ValidateNamespaceIdentifier(namespace); err != nil {
 		return err
 	}
 
@@ -1581,12 +1581,12 @@ func (c *Catalog) CheckNamespaceExists(ctx context.Context, namespace table.Iden
 
 // CreateView creates a new view in the catalog.
 func (c *Catalog) CreateView(ctx context.Context, identifier table.Identifier, schema *iceberg.Schema, viewSQL string, props iceberg.Properties) error {
-	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
-		return err
-	}
-
 	if c.isV0() {
 		return errViewsUnsupportedOnV0
+	}
+
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
+		return err
 	}
 
 	nsIdent := catalog.NamespaceFromIdent(identifier)
@@ -1672,14 +1672,12 @@ func (c *Catalog) ListViews(ctx context.Context, namespace table.Identifier) ite
 }
 
 func (c *Catalog) listViewsAll(ctx context.Context, namespace table.Identifier) ([]table.Identifier, error) {
-	if len(namespace) > 0 {
-		if err := catalog.ValidateNamespaceIdentifier(namespace); err != nil {
-			return nil, err
-		}
-	}
-
 	if c.isV0() {
 		return nil, nil
+	}
+
+	if err := checkValidNamespace(namespace); err != nil {
+		return nil, err
 	}
 
 	ns := namespaceToString(namespace)
@@ -1723,12 +1721,12 @@ func (c *Catalog) listViewsAll(ctx context.Context, namespace table.Identifier) 
 
 // DropView deletes a view from the catalog.
 func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) error {
-	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
-		return err
-	}
-
 	if c.isV0() {
 		return errViewsUnsupportedOnV0
+	}
+
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
+		return err
 	}
 
 	ns, err := c.namespaceKey(ctx, catalog.NamespaceFromIdent(identifier))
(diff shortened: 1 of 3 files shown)


catalog/sql/sql.go
```diff
@@ -1668,6 +1672,12 @@ func (c *Catalog) ListViews(ctx context.Context, namespace table.Identifier) ite
 }
 
 func (c *Catalog) listViewsAll(ctx context.Context, namespace table.Identifier) ([]table.Identifier, error) {
+	if len(namespace) > 0 {
+		if err := catalog.ValidateNamespaceIdentifier(namespace); err != nil {
```
laskoviymishka: I think this makes `ListViews` stricter than the catalog it's reading from.

`checkValidNamespace` is what `CreateNamespace`, `DropNamespace`, and `LoadNamespaceProperties` all use, and it only checks `len(ident) >= 1`, never the components. So `CreateNamespace(ctx, []string{".."}, nil)` succeeds today, the row lands in the DB, and `ListTables` on it still works since `listTablesAll` goes straight to `resolveNamespaceKey`. After this change `ListViews` rejects that same namespace with `ErrNoSuchNamespace` before it ever reaches the lookup. Same namespace, two different answers depending on which list call we make.

I'd rather push the component check down into `checkValidNamespace` so the write path refuses to create those names in the first place and both list paths see the same universe. If tightening create feels out of scope here, I'd drop this block and let `resolveNamespaceKey`'s not-found path handle it, matching `listTablesAll`. Either is defensible, but I don't think we want the split. wdyt?

catalog/sql/sql.go
```diff
@@ -1788,6 +1802,10 @@ func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) err
 
 // CheckViewExists returns true if a view exists in the catalog.
 func (c *Catalog) CheckViewExists(ctx context.Context, identifier table.Identifier) (bool, error) {
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
```
laskoviymishka: This one shifts V0 behavior. `CheckViewExists` used to return `(false, nil)` for any identifier on a V0 catalog, and now an invalid one gets an error instead. `ListViews` has the same shape, where `(nil, nil)` becomes `ErrNoSuchNamespace`. For `CreateView`, `DropView`, and `LoadView` it's only a swap of which error comes back, but for these two we're turning a clean negative into a failure.

I'd move the validation below the `isV0()` guard in all five. That also matches how `ValidateTableIdentifier` sits after the early exits in `LoadTable` and `CommitTable`.

One thing to flag: if we do that, `TestViewOperationsRejectInvalidIdentifiers` stops passing, because the zero-value catalog it uses is V0. Separate note on that below.

catalog/sql/sql_test.go
```diff
@@ -2218,6 +2218,43 @@ func (s *SqliteCatalogTestSuite) TestCreateView() {
...
+		{name: "empty", identifier: table.Identifier{}},
+		{name: "missing namespace", identifier: table.Identifier{"view"}},
+		{name: "empty name", identifier: table.Identifier{"ns", ""}},
+		{name: "dot name", identifier: table.Identifier{"ns", "."}},
+		{name: "parent name", identifier: table.Identifier{"ns", ".."}},
+		{name: "path separator", identifier: table.Identifier{"ns", "nested/view"}},
+		{name: "control character", identifier: table.Identifier{"ns", "view\nname"}},
+	}
+
+	cat := &sqlcat.Catalog{}
```
laskoviymishka: `&sqlcat.Catalog{}` leaves `schemaVersion` at its zero value, which is V0, so this test only passes because the new validation sits above the `isV0()` guard. It's asserting guard ordering rather than the behavior we actually care about.

I'd stand up a V1 in-memory SQLite catalog here instead, so the assertions hold whichever side of `isV0()` the validation ends up on.

catalog/sql/sql_test.go
```diff
@@ -2218,6 +2218,43 @@ func (s *SqliteCatalogTestSuite) TestCreateView() {
...
+			assert.ErrorIs(t, err, catalog.ErrNoSuchView)
+
+			err = cat.DropView(context.Background(), test.identifier)
+			assert.ErrorIs(t, err, catalog.ErrNoSuchView)
+
+			_, err = cat.CheckViewExists(context.Background(), test.identifier)
+			assert.ErrorIs(t, err, catalog.ErrNoSuchView)
+
+			_, err = cat.LoadView(context.Background(), test.identifier)
+			assert.ErrorIs(t, err, catalog.ErrNoSuchView)
```
laskoviymishka, line 2281: `ListViews` is the one view op missing from this table, and it's also the one with its own validation call (`ValidateNamespaceIdentifier` rather than `ValidateViewIdentifier`). Its only coverage right now is the integration assertion down in `TestListViews`. I'd add a case here so a regression in `listViewsAll` shows up in the fast test too.

catalog/sql/sql_test.go
```diff
@@ -2469,6 +2506,13 @@ func (s *SqliteCatalogTestSuite) TestListViews() {
 
 		break
 	}
+
+	viewsIter = db.ListViews(context.Background(), []string{".."})
+	for _, err := range viewsIter {
```
laskoviymishka, line 2587: This can pass without asserting anything. If `ListViews` ever returns an empty iterator instead of one that yields the error, the loop body never runs, `s.ErrorIs` never fires, and the suite still goes green. It works today because the error path yields exactly one element, but that's an implicit contract we aren't checking here.

I'd pull the error out of the loop and assert after it:

```go
var gotErr error
yielded := false
for _, err := range db.ListViews(context.Background(), []string{".."}) {
	yielded, gotErr = true, err

	break
}
s.True(yielded, "invalid namespace must yield at least one element")
s.ErrorIs(gotErr, catalog.ErrNoSuchNamespace)
```

catalog/catalog.go
```diff
@@ -272,6 +267,24 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
...
+
+	return validateIdentifierComponents(ident, notFoundErr)
+}
+
+// ValidateNamespaceIdentifier checks that an identifier contains at least one valid namespace level.
+func ValidateNamespaceIdentifier(ident table.Identifier) error {
+	if len(ident) < 1 {
+		return fmt.Errorf("%w: empty namespace identifier", ErrNoSuchNamespace)
+	}
+
```
laskoviymishka, line 283: Since this is exported now, I'd note in the doc comment that it's stricter than the length-only `checkValidNamespace` the SQL and REST catalogs use internally. We end up with two namespace validators in tree with different semantics and nothing pointing that out, so the next catalog implementation just picks whichever one it happens to find first.

Not a blocker on its own, though it matters more if we keep the asymmetry (see my note on `listViewsAll`).

laskoviymishka (changes requested):
the refactor here is clean. Splitting `validateIdentifierComponents` out of `validateIdentifier` was the right seam, and the five view entry points look complete, nothing missed.

The `ValidateNamespaceIdentifier` call in `listViewsAll` is stricter than anything else in this catalog. `checkValidNamespace`, which guards `CreateNamespace` and friends, only checks length and never looks at components, so a namespace like `{".."}` can be created today, still lists fine through `ListTables`, and now gets rejected by `ListViews` before the lookup even happens. I'd push the component check into `checkValidNamespace` so the create path refuses those names too, or drop it from `listViewsAll` and let the not-found path handle it. Either way I don't think we want `ListTables` and `ListViews` disagreeing about which namespaces exist.

The other thing is guard ordering. Validation landing above `isV0()` turns `CheckViewExists` on a V0 catalog from `(false, nil)` into an error, and `ListViews` from empty into `ErrNoSuchNamespace`. `TestViewOperationsRejectInvalidIdentifiers` doesn't catch that because `&sqlcat.Catalog{}` is itself V0, so the test passes precisely because of the ordering it should be independent of.

A few things I'd ask for before merge:

- pick one strictness level for namespace identifiers and apply it on both the read and the write path
- move the validation below the `isV0()` guard so V0 semantics don't shift
- back `TestViewOperationsRejectInvalidIdentifiers` with a real V1 catalog, and add a `ListViews` case to it
- fix the `for ... { s.ErrorIs(...); break }` block in `TestListViews` so it can't pass without asserting

catalog/sql/sql.go
```diff
@@ -1672,6 +1676,10 @@ func (c *Catalog) listViewsAll(ctx context.Context, namespace table.Identifier)
 		return nil, nil
 	}
 
+	if err := checkValidNamespace(namespace); err != nil {
```
zeroshade: **Blocking — this guard is not the validator the PR is about, and the test covering it passes for an unrelated reason.**

`checkValidNamespace` here is the *local* helper at `sql.go:736`, which only checks `len(ident) < 1`. It is not `catalog.ValidateNamespaceIdentifier`, so no component rules — no empty levels, no `.`/`..`, no control characters — are applied on this path.

The `{".."}` test case passes off the pre-existing not-found path, not off this guard: `..` has length 1, so `checkValidNamespace` returns nil, and the error comes from `resolveNamespaceKey` below failing to find the namespace. **The test stays green if you delete this guard entirely** — worth confirming locally, it's a quick check.

What the line *does* change is real but unintended: `ListViews(ctx, nil)` now errors where it previously listed root-namespace views. That also diverges from `listTablesAll` (`sql.go:1387`), which has no such guard, so listing tables and listing views in the same catalog now behave differently for a nil namespace.

**Suggested fix:** either use `catalog.ValidateNamespaceIdentifier` here *and* in `listTablesAll` so the two agree and the guard actually validates something, or drop this line and let the not-found path handle it as before. Either way the `{".."}` test needs an assertion that fails when the guard is removed.

catalog/sql/sql.go
```diff
@@ -1717,6 +1725,10 @@ func (c *Catalog) DropView(ctx context.Context, identifier table.Identifier) err
 		return errViewsUnsupportedOnV0
 	}
 
+	if err := catalog.ValidateViewIdentifier(identifier); err != nil {
```
zeroshade: **Blocking — this can make existing rows permanently unremovable through the API.**

`CreateView` previously had no validation at all, so views with identifiers like `a/b`, `.`, or embedded control characters can already exist in `iceberg_tables` in deployed catalogs. Adding validation to `DropView` means those rows now fail validation on the way *out* — the only API path that could remove them. The user is left with a view they can neither load nor drop, recoverable only by direct SQL against the catalog tables.

This is also asymmetric with tables, which have always been validated on create via `createStagedTable` (`catalog/internal/utils.go:128`), so no comparable pool of invalid table rows can exist.

**Suggested fix:** exempt `DropView` from component validation — keeping only a length/shape check sufficient to build the query — so a bad row can always be cleaned up. If you'd rather keep the guard for consistency, the PR needs a documented migration note telling operators how to remove pre-existing invalid views.

catalog/catalog.go
```diff
@@ -272,6 +267,31 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
...
+	if len(ident) < 2 {
+		return fmt.Errorf("%w: missing namespace or invalid identifier %v",
+			notFoundErr, strings.Join(ident, "."))
+	}
+
+	return validateIdentifierComponents(ident, notFoundErr)
+}
+
+// ValidateNamespaceIdentifier checks that an identifier contains at least one valid namespace level.
+func ValidateNamespaceIdentifier(ident table.Identifier) error {
```
zeroshade, line 279: **Blocking as a coordination issue rather than a code defect:** this exact function is also added by #1614, at this same location, with an identical body.

The godoc differs, and #1614's is the accurate one. This comment says the function "checks that an identifier contains at least one valid namespace level," but the implementation never defines or checks what makes a level *valid* — it rejects only an empty identifier and NUL bytes. #1614 says so directly ("no null characters ... other rules intentionally left to the implementation"), which matches what the code does.

**Suggested fix:** rebase onto #1614 and drop this copy. It isn't view-related, so removing it makes this PR match its own title, and it avoids a merge conflict where the two godocs would have to be reconciled by hand.

catalog/catalog.go
```diff
@@ -249,12 +249,7 @@ func NamespaceFromIdent(ident table.Identifier) table.Identifier {
 	return ident[:len(ident)-1]
 }
 
-func validateIdentifier(ident table.Identifier, notFoundErr error) error {
-	if len(ident) < 2 {
-		return fmt.Errorf("%w: missing namespace or invalid identifier %v",
-			notFoundErr, strings.Join(ident, "."))
-	}
-
+func validateIdentifierComponents(ident table.Identifier, notFoundErr error) error {
```
zeroshade: This extraction leaves `validateIdentifierComponents` with exactly one caller — `validateIdentifier` on line 276 — and it was previously inline there. The new `ValidateNamespaceIdentifier` deliberately does *not* use it, since namespaces are allowed component shapes that tables and views are not.

So the extraction is scaffolding from an earlier design in which namespaces reused the component rules, and nothing in the shipped code depends on it. #1614 carries the same leftover, and reusing these rules for namespaces is specifically what the maintainer pushed back on there.

**Suggested fix:** drop the extraction and inline the loop back into `validateIdentifier`. It removes a package-level function that reads as shared infrastructure but has no second consumer, and shrinks the conflict surface with #1614.

catalog/sql/sql.go
```diff
@@ -1233,7 +1233,7 @@ func (c *Catalog) CheckTableExists(ctx context.Context, identifier table.Identif
 }
 
 func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifier, props iceberg.Properties) error {
-	if err := checkValidNamespace(namespace); err != nil {
+	if err := catalog.ValidateNamespaceIdentifier(namespace); err != nil {
```
zeroshade: This rewire is out of scope for a "validate view identifiers" PR — `CreateNamespace` is neither a view API nor a boundary this PR's title covers — and #1614 makes the same change to the REST catalog with a broader contract.

It's also only half the story on its own: `DropNamespace` (`sql.go:1288`), and the load/update-properties paths stay on the length-only `checkValidNamespace` (`sql.go:736`), so after this the SQL catalog validates namespace identifiers on create but not on any other namespace operation.

**Suggested fix:** drop this hunk and let the namespace contract be settled in #1614 in one place, for all operations. If you'd rather keep it here, apply it to every namespace entry point in this file rather than just `CreateNamespace`, so the catalog has one consistent rule.

zeroshade (changes requested):
Thanks for this. The core observation is right: the SQL catalog's view entry points were the one family of APIs with no identifier validation at all, while tables have always been validated via `createStagedTable` (`catalog/internal/utils.go:128`). Adding `catalog.ValidateViewIdentifier` at `CreateView` (`sql.go:1588`), `DropView` (`:1728`), `CheckViewExists` (`:1807`), and `LoadView` (`:1837`) is the right set of boundaries, and placing every guard *below* the `isV0()` check is a detail that's easy to get backwards — you got it right in all of them, so a v0 catalog still returns `errViewsUnsupportedOnV0` rather than a validation error.

Three things block, and one of them is coordination rather than code.

## Coordinate with #1614 before iterating further

**#1571 and #1614 conflict textually and semantically.** Both add an exported `ValidateNamespaceIdentifier` and both extract `validateIdentifierComponents` into `catalog/catalog.go` at the same location. The function bodies are identical; the godoc differs, and #1614's is the accurate one.

They also encode *opposite enforcement contracts* for the same rule:

- **#1614** applies the namespace check on every REST namespace path, including reads — all eight call sites.
- **#1571** applies it in the SQL catalog to `CreateNamespace` only (`sql.go:1236`), leaving `DropNamespace`, `LoadNamespaceProperties`, and `UpdateNamespaceProperties` on the length-only local `checkValidNamespace` (`sql.go:736`).

If both merge as-is, SQL enforces write-side only while REST enforces everywhere — for the same identifier rule, with no documentation of the difference.

**Suggested resolution:** rebase this PR onto #1614 and drop its copy of `ValidateNamespaceIdentifier`, the `validateIdentifierComponents` extraction, and the `CreateNamespace` rewire. None of those three are view-related, and none are unique to this PR. That leaves #1571 as a clean, focused "validate view identifiers" change, which is what the title promises. Whichever contract the two of you settle on — enforce-everywhere or enforce-on-write — should then be stated somewhere durable, because after both land it is not inferable from either PR alone.

One more thing worth noting while the two are being reconciled: after both merge, a namespace `{".."}` is creatable via both REST and SQL, yet no table or view can ever live in it — `ValidateTableIdentifier` and `ValidateViewIdentifier` (`catalog/catalog.go:296`, `:301`) reject `..` components. That's pre-existing, but #1614 formalizes the namespace rule and is the natural place to note the divergence.

## Tests

The four `ValidateViewIdentifier` entry points are genuinely complete and, as noted above, correctly placed below the `isV0()` guards. Four gaps:

1. **No `ListViews` case actually reaches validation.** The `{".."}` test passes off the pre-existing not-found path — see the inline note on `sql.go:1679`; the test stays green if you delete the guard entirely.
2. **Nothing covers the empty-namespace behavior change.** `ListViews(ctx, nil)` now returns an error where it previously listed root-namespace views. That's the one observable behavior change the guard actually makes, and it is untested.
3. **No positive control.** Nothing asserts that a valid *nested* view identifier still works end to end through the newly guarded paths — worth having so a future tightening of `ValidateViewIdentifier` can't silently break legitimate identifiers.
4. **Nothing pins the `CreateNamespace` swap** at `sql.go:1236` (which I'd suggest dropping from this PR anyway).

laskoviymishka (approved):
:shipit:

zeroshade (changes requested):
The post-#1614 slimming is exactly right — this is now the clean, view-only PR the title promises, with the guards correctly placed below the `isV0()` checks and the nested-view positive control and non-vacuous ListViews assertions closing the earlier test gaps.

One blocker from my last round is still open, and it's the operational one: `DropView` validating on the way out makes legacy rows unremovable. `CreateView` shipped without validation, so deployed SQL catalogs can already hold views with identifiers this PR now rejects — and `DropView` was the only API that could clean them up. As written, such a row can be neither loaded nor dropped (and the failure reads `ErrNoSuchView` for a row that exists), leaving raw SQL as the only recovery. Please drop the guard from `DropView` — deleting by exact identifier match is safe regardless of the identifier's shape, and it matches Java's `JdbcCatalog` — and add a test that a pre-validation invalid row can still be dropped. With that, this is ready.

---

zeroshade (approved):
Exactly the resolution requested: `DropView` no longer validates, so deleting by exact identifier match works regardless of the identifier's shape — and `TestDropViewRemovesLegacyInvalidView` proves it end-to-end by inserting a raw legacy `nested/view` row and dropping it through the API with a row-count check. Create/CheckExists/Load keep their guards below the `isV0()` checks. With the #1614 overlap already excised, this is now a tight, complete change. Thanks for the fast turnaround.

---

PR #1572, opened 2026-07-28
fix(hive): propagate object-loading failures during listing

fallintoplace:
## What changed

Hive table and view listing now skips only objects that disappear between `GetTables` and `GetTable`. Other metastore errors are returned through the iterator and stop the listing.

## Why

Previously every `GetTable` error was ignored. Connection, authorization, cancellation, and other operational failures therefore looked like successful but incomplete listings.

Regression coverage verifies both concurrent deletion and operational failure behavior for tables and views.

## Testing

- `go test ./catalog/hive`
- `go vet ./catalog/hive`

### DIFF
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -154,7 +154,7 @@ func (c *Catalog) ListTables(ctx context.Context, namespace table.Identifier) it
 					continue
 				}
 
-				yield(nil, fmt.Errorf("failed to load table %s.%s while listing: %w", database, tableName, err))
+				yield(table.Identifier{}, fmt.Errorf("failed to load table %s.%s while listing: %w", database, tableName, err))
 
 				return
 			}
@@ -667,7 +667,7 @@ func (c *Catalog) ListViews(ctx context.Context, namespace table.Identifier) ite
 					continue
 				}
 
-				yield(nil, fmt.Errorf("failed to load view %s.%s while listing: %w", database, viewName, err))
+				yield(table.Identifier{}, fmt.Errorf("failed to load view %s.%s while listing: %w", database, viewName, err))
 
 				return
 			}
@@ -1020,11 +1020,12 @@ func isNoSuchObjectError(err error) bool {
 		return false
 	}
 
-	errStr := err.Error()
+	var noSuchObjectErr *hive_metastore.NoSuchObjectException
+	if errors.As(err, &noSuchObjectErr) {
+		return true
+	}
 
-	return strings.Contains(errStr, "NoSuchObjectException") ||
-		strings.Contains(errStr, "not found") ||
-		strings.Contains(errStr, "does not exist")
+	return strings.Contains(err.Error(), "NoSuchObjectException")
 }
 
 func isAlreadyExistsError(err error) bool {
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -150,7 +150,13 @@ func (c *Catalog) ListTables(ctx context.Context, namespace table.Identifier) it
 		for _, tableName := range tableNames {
 			tbl, err := c.client.GetTable(ctx, database, tableName)
 			if err != nil {
-				continue
+				if isNoSuchObjectError(err) {
+					continue
+				}
+
+				yield(nil, fmt.Errorf("failed to load table %s.%s while listing: %w", database, tableName, err))
+
+				return
 			}
 			if isIcebergTable(tbl) {
 				if !yield(TableIdentifier(database, tableName), nil) {
@@ -657,8 +663,13 @@ func (c *Catalog) ListViews(ctx context.Context, namespace table.Identifier) ite
 		for _, viewName := range viewNames {
 			tbl, err := c.client.GetTable(ctx, database, viewName)
 			if err != nil {
-				// Skip tables we fail to load, mirroring table listing behavior.
-				continue
+				if isNoSuchObjectError(err) {
+					continue
+				}
+
+				yield(nil, fmt.Errorf("failed to load view %s.%s while listing: %w", database, viewName, err))
+
+				return
 			}
 			if isIcebergView(tbl) {
 				if !yield(TableIdentifier(database, viewName), nil) {
--- a/catalog/hive/hive_test.go
+++ b/catalog/hive/hive_test.go
@@ -308,6 +308,7 @@ func TestHiveListTablesHandlesObjectLoadingErrors(t *testing.T) {
 			if err != nil {
 				require.Equal(t, table.Identifier{}, identifier)
 				gotErr = err
+
 				break
 			}
 		}
@@ -1531,6 +1532,7 @@ func TestHiveListViewsHandlesObjectLoadingErrors(t *testing.T) {
 			if err != nil {
 				require.Equal(t, table.Identifier{}, identifier)
 				gotErr = err
+
 				break
 			}
 		}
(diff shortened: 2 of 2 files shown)


catalog/hive/hive.go
```diff
@@ -150,7 +150,13 @@ func (c *Catalog) ListTables(ctx context.Context, namespace table.Identifier) it
 		for _, tableName := range tableNames {
 			tbl, err := c.client.GetTable(ctx, database, tableName)
 			if err != nil {
-				continue
+				if isNoSuchObjectError(err) {
```
laskoviymishka, line 153: `isNoSuchObjectError` is now the whole ballgame here: it's the only thing standing between "skip a concurrently deleted table" and "abort the listing with a real error." The problem is it matches on the substrings `"not found"` and `"does not exist"`, and plenty of operational failures carry those (a DNS miss like `"metastore host not found"`, an auth `"credentials not found"`, or anything wrapping our own `catalog.ErrNoSuchTable`, whose message is `"table does not exist"`). Each of those gets classified as a concurrent deletion and silently skipped, which is the exact bug this PR is trying to kill.

In the `GetTable` path gohive only ever returns typed Thrift exceptions, so I'd lead with a type check and let that carry the decision:

```go
var nsoe *hive_metastore.NoSuchObjectException
if errors.As(err, &nsoe) {
    continue
}
```

If we want to keep a string fallback for other callers I'd narrow it to just `"NoSuchObjectException"` and drop the two broad ones. Same gate applies on the `ListViews` side. wdyt?

catalog/hive/hive.go
```diff
@@ -150,7 +150,13 @@ func (c *Catalog) ListTables(ctx context.Context, namespace table.Identifier) it
...
+					continue
+				}
+
+				yield(nil, fmt.Errorf("failed to load table %s.%s while listing: %w", database, tableName, err))
```
laskoviymishka: Small consistency thing: Glue and REST both yield `table.Identifier{}` on the error path and here we yield `nil`. Callers shouldn't be dereferencing the identifier after a non-nil error anyway so it's harmless, but I'd match the other catalogs.

catalog/hive/hive_test.go
```diff
@@ -280,6 +280,39 @@ func TestHiveListTablesEmpty(t *testing.T) {
 	mockClient.AssertExpectations(t)
 }
 
+func TestHiveListTablesHandlesObjectLoadingErrors(t *testing.T) {
```
laskoviymishka, line 283: The two subtests are each single-purpose: one only skips, one only propagates. The case I'd actually worry about is the mixed one, since that's where the new branch ordering can go wrong: a listing where a NoSuchObject entry and an operational-failure entry both appear, plus a good one after. I'd add a subtest with `[deleted (skip), broken (propagate), good]` and assert the broken error surfaces while the good one's `GetTable` is never reached. A cancelled-context case that propagates would be a nice one to have too. (Same goes for the `ListViews` test.)

catalog/hive/hive_test.go
```diff
@@ -280,6 +280,39 @@ func TestHiveListTablesEmpty(t *testing.T) {
...
+	t.Run("skips concurrently deleted tables", func(t *testing.T) {
+		mockClient := &mockHiveClient{}
+		mockClient.On("GetTables", mock.Anything, "test_database", "*").Return([]string{"deleted", "test_table"}, nil).Once()
+		mockClient.On("GetTable", mock.Anything, "test_database", "deleted").Return(nil, errNoSuchObject).Once()
+		mockClient.On("GetTable", mock.Anything, "test_database", "test_table").Return(testIcebergHiveTable1, nil).Once()
+
+		var tables []table.Identifier
+		for identifier, err := range NewCatalogWithClient(mockClient, nil).ListTables(context.Background(), DatabaseIdentifier("test_database")) {
```
laskoviymishka: Every other test in this file passes `iceberg.Properties{}` here; these new ones pass `nil`. If `ApplyProperties` ever iterates the map without a nil guard that's a setup panic masking the thing under test. I'd pass `iceberg.Properties{}` to stay consistent.

catalog/hive/hive_test.go
```diff
@@ -280,6 +280,39 @@ func TestHiveListTablesEmpty(t *testing.T) {
...
+
+	t.Run("propagates operational failures", func(t *testing.T) {
+		loadErr := errors.New("metastore unavailable")
+		mockClient := &mockHiveClient{}
+		mockClient.On("GetTables", mock.Anything, "test_database", "*").Return([]string{"broken", "test_table"}, nil).Once()
+		mockClient.On("GetTable", mock.Anything, "test_database", "broken").Return(nil, loadErr).Once()
+
+		var gotErr error
+		for _, err := range NewCatalogWithClient(mockClient, nil).ListTables(context.Background(), DatabaseIdentifier("test_database")) {
+			gotErr = err
```
laskoviymishka: This loop keeps going after the error and only holds the last one, so it leans on the iterator stopping itself rather than checking that it does. I'd break on the first error like a real consumer would, and while we're unpacking both values, pin that the identifier is nil on the error yield:

```go
for ident, err := range NewCatalogWithClient(...).ListTables(...) {
    if err != nil {
        require.Nil(t, ident)
        gotErr = err
        break
    }
}
```

catalog/hive/hive_test.go
```diff
@@ -280,6 +280,39 @@ func TestHiveListTablesEmpty(t *testing.T) {
...
+		}
+		require.ErrorIs(t, gotErr, loadErr)
+		require.ErrorContains(t, gotErr, "failed to load table test_database.broken while listing")
+		mockClient.AssertNotCalled(t, "GetTable", mock.Anything, "test_database", "test_table")
```
laskoviymishka: `AssertNotCalled` with those exact args doesn't really check what we want: there's no `On` registered for `"test_table"`, so any call would already blow up as unexpected. It also doesn't prove iteration actually stopped. I'd swap it for `mockClient.AssertNumberOfCalls(t, "GetTable", 1)`, which pins that we stopped after the broken entry.

laskoviymishka (changes requested):
This is a good catch and the right shape overall. Swallowing every `GetTable` failure during listing meant a flaky metastore could hand back a silently truncated list that looked complete, and splitting concurrent-deletion (skip) from real failures (propagate) is exactly the fix.

Almost there. The one blocking thing before merge is the gate the whole fix now hangs on: `isNoSuchObjectError`. It matches `"not found"` and `"does not exist"` as substrings, and a fair number of operational errors carry those, including anything that wraps our own `catalog.ErrNoSuchTable` (`"table does not exist"`). Those get read as concurrent deletions and skipped, which quietly reopens the exact hole this PR closes. In the `GetTable` path gohive only returns typed Thrift exceptions, so an `errors.As` check against `*hive_metastore.NoSuchObjectException` would make the decision robust; specifics are inline.

The tests exercise both branches, which is great. A couple are softer than they look though: the propagate loop doesn't break on the first error, `AssertNotCalled` with exact args isn't really asserting iteration stopped, and there's no mixed skip-then-propagate case, which is where the new branch ordering is most likely to bite. Those are inline too.

One optional aside while we're here: `ListTables` still doesn't check namespace existence the way `ListViews` does, so the two diverge on a missing namespace. Pre-existing and fine to leave for a follow-up.

Once the discriminator is tightened, happy to take another pass.

laskoviymishka (approved):

PR #1573, opened 2026-07-28
fix(hive): preserve RegisterTable creation errors

fallintoplace:
## What changed

Wrap the Hive metastore `CreateTable` failure returned by `RegisterTable` with `%w`.

## Why

The previous `%s` formatting converted the cause to text, preventing callers from detecting context, transport, or typed metastore errors with `errors.Is` and `errors.As`.

The regression test verifies that the original creation error remains discoverable.

## Testing

- `go test ./catalog/hive`
- `go vet ./catalog/hive`


laskoviymishka (approved):
nice small improvement!

PR #1575, opened 2026-07-28
fix(manifest): protect DataFile metadata from external mutation

fallintoplace:
## Summary

Return defensive copies from DataFile getters and protect mutable builder inputs.

## Why

Getters exposed cached maps, slices, byte buffers, and pointers directly. Callers could mutate decoded or built data files through an accessor, including partition and bounds data used during planning.

## What changed

Collection and optional-value getters now return independent values. Byte bounds and binary partition values are deep-copied, and builder inputs with mutable storage are cloned.

## Tests

- go test .
- go vet .


data_file_refs.go
```diff
@@ -0,0 +1,35 @@
...
+func (d *dataFile) DataFileStatsRef(_ internal.DataFileRef) (
+	valueCounts map[int]int64,
+	nullCounts map[int]int64,
+	nanCounts map[int]int64,
+	lowerBounds map[int][]byte,
+	upperBounds map[int][]byte,
+) {
+	d.initColumnStatsData()
+
+	return d.valCntMap, d.nullCntMap, d.nanCntMap, d.lowerBoundMap, d.upperBoundMap
```
laskoviymishka, line 45: This is the one I'd want settled before merge.

The PR is about protecting DataFile state from callers, but the `[]byte` values in the `lowerBounds`/`upperBounds` returned here are the same backing arrays as `d.lowerBoundMap`/`d.upperBoundMap` — a caller doing `lowerBounds[id][0] = x` writes straight through into cached state, and since `initColumnStatsData` is a `sync.Once` it never rebuilds.

Both current callers are read-only so there's no live bug, but "zero-copy access to immutable DataFile state" claims more than the code delivers — the maps are shared and the byte values are mutable. At minimum, document on this method that callers must treat the returned slices as read-only, so the next caller wiring into this path doesn't corrupt bounds by accident.

manifest.go
```diff
@@ -2201,7 +2202,45 @@ func (d *dataFile) FileFormat() FileFormat            { return d.Format }
...
+}
+
+func clonePartitionMap(src map[int]any) map[int]any {
+	if src == nil {
+		return nil
+	}
+
+	out := maps.Clone(src)
+	for id, value := range out {
+		if bytes, ok := value.([]byte); ok {
```
laskoviymishka, line 2363: `clonePartitionMap` only deep-copies `[]byte` — every other `any` is carried through by the shallow `maps.Clone`, so the isolation guarantee is really "`[]byte` only".

For Avro-decoded files that's fine, since `convertAvroValueToIcebergType` only produces value types. But `NewDataFileBuilder` takes a caller-supplied `map[int]any`, and nothing stops a caller putting a `*big.Rat` or a nested map in there — those still alias after the clone, so the stated protection doesn't hold on the builder path. I'd document that isolation only holds for the decoder's concrete types, or handle the unsupported case explicitly.

manifest.go
```diff
@@ -2201,7 +2202,45 @@ func (d *dataFile) FileFormat() FileFormat            { return d.Format }
 func (d *dataFile) Partition() map[int]any {
 	d.initPartitionData()
 
-	return d.fieldIDToPartitionData
+	return clonePartitionMap(d.fieldIDToPartitionData)
```
laskoviymishka, line 2350: `Partition()` is on the same per-file scan path as the stats getters (partition pruning calls it once per file, plus once per equality-delete/data pair) and now clones unconditionally, while stats get the `DataFileStatsRef` fast path — so the optimization is asymmetric.

More to the point: the escape hatch adds a new internal package, a token type, an interface, and a runtime assertion to avoid five map copies per eval, and there's no benchmark showing that copy was ever hot. This PR needs a benchmark that demonstrates the saving (e.g. `BenchmarkInclusiveMetricsEval` with and without the ref path) before we take on this machinery — if it doesn't move the numbers, the escape hatch should come out and we just copy. Either way the result has to reconcile with `Partition()` copying unconditionally on the same path.

laskoviymishka (changes requested):
The defensive copying looks thorough, and the isolation test is a good way to catch the aliasing cases in the getters.

I’d still hold the merge over `DataFileStatsRef`.

The PR is trying to make `DataFile` state immutable to callers, but `DataFileStatsRef` returns `map[int][]byte` with slices that still share backing arrays with the cached maps. An in-package caller can modify those bytes and corrupt the bounds. Since `initColumnStatsData` is guarded by `sync.Once`, that corruption then stays in place.

The current evaluators only read from the map, so this is not an active bug today, but the “zero-copy access to immutable DataFile state” comment is stronger than what the API guarantees. At minimum, the returned slices should be documented as read-only.

I’m also not convinced the escape hatch is worth the extra machinery yet. It adds a package, token type, interface, and runtime assertion to avoid five map copies per evaluation, but there is no benchmark showing those copies matter. `Partition()` is on the same per-file scan path and now always copies, without a similar fast path. That makes the optimization look both unproven and inconsistent.

I’d settle the following before merge:

* Document, or enforce, that the byte slices returned by `DataFileStatsRef` are read-only.
* Add a benchmark for the stats path, or explain why the escape hatch is needed. It would also be worth deciding whether `Partition()` needs the same treatment.
* Document that `clonePartitionMap` only guarantees deep isolation for the concrete types produced by the decoder, since the builder accepts arbitrary `map[int]any` values.
* Remove the second clone of binary partition values in `NewDataFileBuilder`.
* Tighten the zero-allocation test with `InDelta` and a correctness assertion inside the closure.
* Consider adding the Avro decode path to the isolation test.

Once those are sorted, I’ll take another look.

manifest_test.go
```diff
@@ -2969,6 +2969,82 @@ func (m *ManifestTestSuite) TestManifestEntryPresentEmptyListSurvivesRewrite() {
 		"present-empty column_sizes must survive a decode -> re-encode rewrite as a present array")
 }
 
+func (m *ManifestTestSuite) TestDataFileMetadataIsIsolatedFromExternalMutation() {
```
laskoviymishka, line 3364: This test only covers the builder path. The round-1 ask was to exercise the Avro decode path too, and that's still the gap I care about most.

Here's why it's more than coverage for its own sake. `initPartitionData` rebuilds `fieldIDToPartitionData` from `PartitionData`, and `convertAvroValueToIcebergType` returns `[]byte` as-is, so after a decode the internal `fieldIDToPartitionData` shares backing arrays with the decoder's buffers. `Partition()` clones on the way out, so nothing leaks to callers today, but the internal state is aliased and nothing proves it stays benign.

Can we round-trip a manifest here (encode, decode, mutate the source bytes) and assert the decoded file's `Partition()` and bounds are unchanged? That closes the one path this test doesn't touch.

manifest.go
laskoviymishka, line 2350: The benchmark answered my round-1 doubt about whether the stats fast path is worth it, thanks for adding it. The half that's still open is the consistency call: `Partition()` is on the same per-file scan path and now clones on every call, and `ColumnSizes`/`DistinctValueCounts` sit outside the borrowed view too.

I'm not asking to widen the escape hatch, just to make the scope deliberate. Either extend the borrowed view (or add a `Partition` benchmark showing the copy is cheap), or drop a line on why the partition map and those two stat maps are intentionally left on the copying path. wdyt?

table/data_file_stats_ref_test.go
```diff
@@ -0,0 +1,167 @@
...
+	require.Implements(t, (*dataFileStatsRefer)(nil), file)
+
+	valueCounts, nullCounts, nanCounts, lowerBounds, upperBounds := dataFileStats(file)
+	assert.Equal(t, map[int]int64{1: 2}, valueCounts)
+	assert.Equal(t, map[int]int64{1: 0}, nullCounts)
+	assert.Equal(t, map[int]int64{1: 0}, nanCounts)
+	assert.Equal(t, map[int][]byte{1: {1, 2}}, lowerBounds)
+	assert.Equal(t, map[int][]byte{1: {3, 4}}, upperBounds)
+
+	assert.Zero(t, testing.AllocsPerRun(100, func() {
```
laskoviymishka: This still passes if `dataFileStats` regresses to returning five nil maps. The closure discards every return, so it only proves "zero allocs", not "zero allocs and correct". Round-1 asked to tighten this.

Capture a value and assert it, and `InDelta` avoids a flake if a GC cycle sneaks a fractional alloc in:

```go
var vc map[int]int64
allocs := testing.AllocsPerRun(100, func() {
    vc, _, _, _, _ = dataFileStats(file)
})
assert.InDelta(t, 0.0, allocs, 0.5)
assert.Equal(t, map[int]int64{1: 2}, vc)
```

laskoviymishka (changes requested):
Nice progress. The benchmark is the big one for me: I pushed back on the escape hatch last round as unproven, and `BenchmarkInclusiveMetricsEvalDataFileStats` is exactly what was missing, so I'm no longer worried about the machinery.

First, a correction on my side. I asked to drop the "second clone" of the partition bytes in `NewDataFileBuilder`, and I misread it. `d.PartitionData` (string-keyed, for the Avro encode) and `d.fieldIDToPartitionData` (int-keyed, for eval) are two separate maps that each need their own copy, so those clones aren't redundant. Withdrawing that one.

The thing I'd still like closed before merge is the decode path. The isolation test only exercises the builder path, and looking at decode more closely there's a concrete reason to test it: `initPartitionData` rebuilds `fieldIDToPartitionData` from `PartitionData`, and `convertAvroValueToIcebergType` returns `[]byte` as-is, so after a decode the internal partition map shares backing arrays with the decoder's buffers. `Partition()` clones on the way out so nothing leaks to callers, but I'd like a round-trip test (encode, decode, mutate the source bytes, assert unchanged) to prove that internal aliasing stays benign.

Everything else I asked for is in:

* `DataFileStatsRef`'s returned maps and byte slices are now documented read-only (`data_file_refs.go`).
* the stats-path benchmark is there, and it justifies the fast path.
* `clonePartitionMap`'s isolation limits are documented on `NewDataFileBuilder`.

Two smaller things left, both inline: `Partition()` (and `ColumnSizes`/`DistinctValueCounts`) still clone on every call with no benchmark or note on why they sit outside the borrowed view, and the zero-alloc test discards its results so it only proves "zero allocs", not "zero allocs and correct".

Close the decode-path gap and make the `Partition()` call, and I'm happy to approve.

manifest_test.go
fallintoplace: I traced this through twmb/avro. Its `[]byte` decoder allocates a new slice and copies the bytes [here](https://github.com/twmb/avro/blob/v1.7.2/deser.go#L343-L357), so the decoded value does not alias the encoded input or reader buffer.

I don’t think mutating source bytes tests a real failure mode here.

fallintoplace: I added the round-trip test anyway to lock down the ownership behavior across the Avro decode path.

table/scanner.go
```diff
@@ -182,7 +182,7 @@ func newPartitionRecord(partitionData map[int]any, partitionType *iceberg.Struct
 // GetPartitionRecord converts a DataFile's partition map into a positional
 // record ordered by the fields of the given partition struct type.
 func GetPartitionRecord(dataFile iceberg.DataFile, partitionType *iceberg.StructType) iceberg.StructLike {
-	return newPartitionRecord(dataFile.Partition(), partitionType)
+	return newPartitionRecord(dataFilePartition(dataFile), partitionType)
```
zeroshade, line 190: **Blocking — this re-opens, through an exported API, the exact mutation hole the PR is titled to close.**

`GetPartitionRecord` (line 184) is exported and used outside this repo. It calls `dataFilePartition(dataFile)`, which for a concrete `*dataFile` returns `d.fieldIDToPartitionData` **uncopied** (`data_file_refs.go:48-51`). `newPartitionRecord` (`scanner.go:173`) copies the *values* into a `[]any`, but for a `[]byte` partition value that copies the slice header, not the bytes — so `partitionRecord.Get(pos)` (`scanner.go:127`) hands the caller the DataFile's live `[]byte`. `iceberg.StructLike` is fully public with both `Get` and `Set` (`utils.go:47-56`).

Concrete failure:

```go
table.GetPartitionRecord(df, partType).Get(0).([]byte)[0] = 0xff
```

permanently corrupts `d.fieldIDToPartitionData`. `initPartitionData` is `sync.Once`-guarded (`manifest.go:2076`, `manifest.go:2080-2092`), so the map is never rebuilt and there is no recovery.

**The corruption reaches disk.** `ManifestWriter.Add` calls `entry.Data.Partition()` (`manifest.go:1462`), which clones the already-corrupted map and writes it as that entry's partition tuple. `MarshalAvroEntry` does the same at `data_file_codec.go:110`. A wrong partition tuple in a manifest means files that no longer prune correctly.

This is not a regression — pre-PR, `Partition()` returned the internal map too. The problem is that the PR closes the `Partition()` hole and leaves the equivalent one open one function away, on the public surface. The asymmetry matters: `dataFileStats`' two in-package consumers are read-only visitors you control, while `GetPartitionRecord`'s consumers are unbounded.

**Suggested fix:** either revert this line to `dataFile.Partition()` and keep `dataFilePartition` for the internal-only sites at `scanner.go:568` and `:581`, or have `newPartitionRecord` clone `[]byte` values as it fills the slice. The second costs one clone per *partition field* (typically 1-3), not per column, so it stays cheap while making the exported API honest.

table/data_file_stats_ref.go
```diff
@@ -0,0 +1,60 @@
...
+		lowerBounds map[int][]byte,
+		upperBounds map[int][]byte,
+	)
+}
+
+type dataFilePartitionRefer interface {
+	DataFilePartitionRef(internal.DataFileRef) map[int]any
+}
+
+func dataFileStats(file iceberg.DataFile) (
```
zeroshade, line 29: The read-only contract is documented thoroughly, but on the function nobody calls. `data_file_refs.go:22-30` spells out the aliasing rules on `(*dataFile).DataFileStatsRef` — which no in-package caller invokes directly. `dataFileStats`, which is what callers actually use, has no doc comment at all.

That matters more than usual here because ownership of the return value silently depends on the argument's *dynamic* type: borrowed when it's a `*dataFile`, an independent copy for any other `DataFile` implementation. A future caller in package `table` who writes into `ev.lowerBounds` would pass every test that uses a wrapper type and corrupt cached state in production.

**Suggested fix:** move (or duplicate) the ownership contract onto `dataFileStats` and `dataFilePartition`, and say explicitly that the caller must not mutate or retain the returned maps regardless of which branch is taken.

table/data_file_stats_ref.go
```diff
@@ -0,0 +1,60 @@
...
+) {
+	if ref, ok := file.(dataFileStatsRefer); ok {
+		return ref.DataFileStatsRef(internal.DataFileRef{})
+	}
+
+	return file.ValueCounts(), file.NullValueCounts(), file.NaNValueCounts(),
+		file.LowerBoundValues(), file.UpperBoundValues()
+}
+
+func dataFilePartition(file iceberg.DataFile) map[int]any {
```
zeroshade, line 43: Same missing contract as `dataFileStats` above, and it applies more sharply here: this is the function feeding `GetPartitionRecord` at `scanner.go:185`, so the "must not mutate or retain" rule is the only thing standing between an exported API and permanent corruption of `d.fieldIDToPartitionData`.

**Suggested fix:** document the borrowed-vs-copied return contract here, and note that the `*dataFile` branch aliases state protected by a `sync.Once` (`manifest.go:2080`), so a violation is unrecoverable rather than merely stale.

manifest.go
```diff
@@ -2201,7 +2202,45 @@ func (d *dataFile) FileFormat() FileFormat            { return d.Format }
...
+}
+
+func clonePartitionMap(src map[int]any) map[int]any {
```
zeroshade, line 2353: The `[]byte`-only deep copy is correct, but only because of a non-obvious property of the upstream conversion, and nothing here records that.

I checked every type `convertAvroValueToIcebergType` can produce: `Date`/`Time`/`Timestamp`/`TimestampNano` are integer types, `DecimalLiteral.Val` holds a `decimal128.Num` by value, and `uuid.UUID` is a `[16]byte` array — so `[]byte` really is the only reference-typed value that can appear in this map.

**Suggested fix:** state that in the function's own comment. Without it, the next person to add a partition value type has no signal that this function needs updating, and the failure would be silent aliasing rather than a compile error.

manifest.go
```diff
@@ -2201,7 +2202,45 @@ func (d *dataFile) FileFormat() FileFormat            { return d.Format }
...
+	for id, value := range out {
+		if bytes, ok := value.([]byte); ok {
+			out[id] = slices.Clone(bytes)
+		}
+	}
+
+	return out
+}
+
+func cloneByteMap(src map[int][]byte) map[int][]byte {
```
zeroshade, line 2371: **Highest-value missing test.** Nothing currently locks nil-vs-present-empty through the new clone helpers, and this repo cares about that distinction specifically — see `manifest_test.go:2920` and the assertions at `:2966-2969` requiring that `column_sizes: []` survive a decode → re-encode rewrite as a present array rather than an Avro null (issue #1309).

Today it still works, but only incidentally: `maps.Clone` preserves non-nil for an empty map, and `cloneByteMap`/`clonePartitionMap` nil-check *before* the `make`.

Given the allocation pressure this PR introduces, `if len(src) == 0 { return nil }` is exactly the optimization someone applies to `cloneByteMap` next — and it would silently collapse present-empty `lower_bounds`/`upper_bounds`/`column_sizes` to null on every manifest rewrite, with no failing test anywhere.

**Suggested fix:** for each cloned getter, assert all three states — absent → nil, present-empty → non-nil empty, populated → equal — after a decode → getter → builder → re-encode round trip. That pins the invariant at the layer where the optimization would be applied.

manifest.go
```diff
@@ -2476,7 +2521,7 @@ func NewDataFileBuilder(
 			RecordCount:            recordCount,
 			FileSize:               fileSize,
 			specID:                 int32(spec.id),
-			fieldIDToPartitionData: fieldIDToPartitionData,
+			fieldIDToPartitionData: clonePartitionMap(fieldIDToPartitionData),
 			fieldNameToID:          fieldNameToID,
 			fieldIDToLogicalType:   fieldIDToLogicalType,
```
zeroshade: `fieldIDToPartitionData` is cloned two lines up, but `fieldIDToLogicalType` is retained by reference from the caller — inconsistent with the goal stated in the PR title.

This is harmless today, but only by accident: the builder path never reaches `convertAvroValueToIcebergType`, because the `len(d.fieldIDToPartitionData) < len(d.PartitionData)` guard at `manifest.go:2082` is always false for a builder-constructed file. So the map the caller can still mutate is never read. That's an accident of the current control flow, not a design.

**Suggested fix:** clone it alongside the partition map, or add a comment explaining why it is deliberately shared.

manifest.go
```diff
@@ -2534,35 +2579,38 @@ func (b *DataFileBuilder) DistinctValueCounts(counts map[int]int64) *DataFileBui
 
 // LowerBoundValues sets the lower bound values for the data file.
 func (b *DataFileBuilder) LowerBoundValues(bounds map[int][]byte) *DataFileBuilder {
-	b.d.LowerBounds = mapToAvroColMap(bounds)
+	b.d.LowerBounds = mapToAvroColMap(cloneByteMap(bounds))
```
zeroshade: `mapToAvroColMap(cloneByteMap(bounds))` builds a throwaway map that is immediately flattened into a slice, so the intermediate map allocation and its bucket growth are pure waste. Same on line 2589 for upper bounds.

**Suggested fix:** inline the `slices.Clone` into the flatten loop inside `mapToAvroColMap` (or a `mapToAvroColMapCloned` variant) and drop one map allocation per builder call in each direction. Small, but this is on the write path for every data file.

data_file_refs.go
```diff
@@ -0,0 +1,52 @@
...
+
+import "github.com/apache/iceberg-go/internal"
+
+// DataFileStatsRef returns statistics maps owned by the data file. The token
+// restricts this zero-copy accessor to trusted in-module callers; the public
+// DataFile getters continue returning defensive copies. The returned maps and
+// every byte slice reachable through the bounds maps alias the DataFile and
+// must not be mutated or retained beyond the current evaluation.
+//
+// This view intentionally contains only the maps consumed by metric
```
zeroshade, line 30: "intentionally contains only the maps consumed by metric evaluators" reads as though the set were exhaustive, but `ColumnSizes` and `DistinctValueCounts` are also column-stat maps and they are left on the copying path with no stated rationale.

**Suggested fix:** either say why those two are excluded (I assume: no evaluator reads them, so no hot path justifies the extra surface), or fold them in. As written, the next person to touch this can't tell whether the omission is deliberate or an oversight.

manifest_test.go
```diff
@@ -2969,6 +2969,127 @@ func (m *ManifestTestSuite) TestManifestEntryPresentEmptyListSurvivesRewrite() {
...
+		ReferencedDataFile("data.parquet").
+		ContentOffset(5).
+		ContentSizeInBytes(6).
+		Build()
+
+	partition[0], lower[1][0], upper[1][0], key[0], splits[0], equalityIDs[0] = 0xff, 0xff, 0xff, 0xff, 99, 99
+	partitionData[1000] = []byte{0xff}
+	columnSizes[1], valueCounts[1], nullCounts[1], nanCounts[1], distinctCounts[1] = 99, 99, 99, 99, 99
+
+	m.Equal([]byte{0x01, 0x02}, builder.d.PartitionData["part"])
```
zeroshade: This asserts on `builder.d.PartitionData["part"]`, reaching into an unexported field of the type under test. That couples the test to the internal representation and, more importantly, doesn't exercise the path a caller actually takes.

**Suggested fix:** assert through `Build()` followed by `Partition()`. That tests the same isolation property via the public surface and would keep passing across an internal refactor — while also failing if the *getter* stops copying, which is what the PR is really about.

manifest_test.go
```diff
@@ -2969,6 +2969,127 @@ func (m *ManifestTestSuite) TestManifestEntryPresentEmptyListSurvivesRewrite() {
...
+	m.Equal([]int64{10, 20}, dataFile.SplitOffsets())
+	m.Equal([]int{1, 2}, dataFile.EqualityFieldIDs())
+	m.Equal(3, *dataFile.SortOrderID())
+	m.Equal(int64(4), *dataFile.FirstRowID())
+	m.Equal("data.parquet", *dataFile.ReferencedDataFile())
+	m.Equal(int64(5), *dataFile.ContentOffset())
+	m.Equal(int64(6), *dataFile.ContentSizeInBytes())
+}
+
+func (m *ManifestTestSuite) TestDecodedDataFileMetadataIsIsolatedFromSourceAndGetterMutation() {
```
zeroshade, line 3441: Good test — mutating the encoded buffer in place after decode is a genuinely convincing way to prove the decoder doesn't alias the source. Two gaps worth closing.

Decode-path isolation here covers only `Partition`, `LowerBoundValues`, and `UpperBoundValues`. `ColumnSizes`, `ValueCounts`, `NullValueCounts`, `NaNValueCounts`, `DistinctValueCounts`, `KeyMetadata`, `SplitOffsets`, `EqualityFieldIDs`, and all five `clonePointer` getters are exercised on the builder path only — so a decode-path aliasing bug in any of them would go unnoticed.

**Suggested fix:** extend the post-mutation assertions to the remaining getters. The scaffolding is already here; it's mostly additional assertion lines.

table/data_file_stats_ref_test.go
```diff
@@ -0,0 +1,268 @@
...
+	assert.Equal(t, "partition", partition[1000])
+}
+
+func TestDataFilePartitionFallsBackToPublicGetter(t *testing.T) {
+	file := &publicPartitionDataFile{DataFile: testDataFileWithStats(t)}
+
+	assert.Equal(t, map[int]any{1000: "partition"}, dataFilePartition(file))
+}
+
+func BenchmarkInclusiveMetricsEvalDataFileStats(b *testing.B) {
```
zeroshade: This benchmark is the justification for the whole token mechanism, and it understates its own case in one way while being unfair in another.

**It uses a single-column file** (lines 167-170). Copy cost is dominated by the `2N` byte-slice clones, so one column understates the saving by roughly two orders of magnitude for a realistic table. Sweeping 1 / 50 / 500 columns would *strengthen* the argument for the fast path — while simultaneously making the cost at the ~20 unmitigated call sites impossible to overlook.

**Only the control arm is instrumented.** The "public defensive copies" arm wraps the file in `publicStatsDataFile` (line 189), whose getters each do `f.getterCalls++` (lines 43, 49, 55, 61, 67) inside the benchmarked loop, plus an extra interface hop. The "borrowed stats" arm has neither. So some of the measured delta is instrumentation, not copying.

**Suggested fix:** parameterize over column count, and benchmark against a plain non-instrumented wrapper — keep `publicStatsDataFile` for the correctness tests where the call counting is the point.

zeroshade (changes requested):
Thanks for this — the diagnosis is right and two parts of the fix are unambiguously correct and should survive whatever happens to the rest.

What's genuinely sound:

- **Builder-side cloning is the valuable half.** `NewDataFileBuilder` cloning `fieldIDToPartitionData` (`manifest.go:2524`) and the setters cloning caller input closes a real aliasing bug — a caller could previously mutate the map they passed in and silently change a built `DataFile`.
- **The escape hatch is a clever design.** `internal.DataFileRef` as an unforgeable token (`internal/data_file_ref.go:20-23`) is the right Go idiom for "trusted in-module callers only" — the internal-package rule genuinely prevents external construction, and it keeps `inclusiveMetricsEval` allocation-free without exporting anything.
- **The clone helpers are correct and nil-safe.** `clonePartitionMap` (`manifest.go:2208`) deep-copies exactly the one type that needs it, and both helpers preserve nil-vs-non-nil.
- **The performance concern being defended is real,** not speculative. Pre-PR, `inclusiveMetricsEval.Eval` did zero allocations for stats. Naive copying is 5 map allocs + 2N `[]byte` allocs per data file. On a 100-column table across 1M manifest entries that is roughly 2e8 added allocations in scan planning. Protecting that path was the right instinct.

Two things block, and the first is a correctness issue rather than a performance one.

## 1. The exported API hole at `table/scanner.go:185` (blocking)

Detailed inline. Short version: the PR closes the `Partition()` mutation hole and then keeps the equivalent hole open one function away, through `GetPartitionRecord` — which unlike `dataFileStats`' two in-package read-only consumers has an unbounded external audience. Corruption there is permanent (`sync.Once` at `manifest.go:2076`/`2080`) and propagates to disk via `ManifestWriter.Add` (`manifest.go:1462`).

## 2. The fast path covers 5 of ~25 call sites (blocking)

The escape hatch is applied at exactly five places — `table/evaluators.go:777`, `table/evaluators.go:1246`, `table/scanner.go:185`, `table/scanner.go:568`, `table/scanner.go:581`. Every other `DataFile` metadata read in the repo is now strictly slower than before this PR, and several are on hotter paths than the one that got optimized. None of these lines are touched by the diff, so they need listing here rather than inline.

**The two worst, both more expensive per call than the metrics-eval path the machinery was built to protect:**

- `table/conflict_validation.go:699-700` and `table/compaction/compaction.go:370-371` — `df.LowerBoundValues()[filePathFieldID]` deep-clones the *entire* bounds map (one map alloc plus one `slices.Clone` per column) to read a single key, and then does it again on the next line for upper bounds. On a 500-column table that is ~1000 allocations per delete file per pass. `conflict_validation.go:663` sits inside `forEachAddedEntry`, i.e. once per added delete entry across every concurrent snapshot.
- `table/snapshots.go:458` and `table/snapshots.go:473` — `if len(df.Partition()) > 0` deep-clones the whole partition map purely to test emptiness, and then immediately calls `GetPartitionRecord`, which borrows. Once per data file per commit. Should be `len(dataFilePartition(df)) > 0`.

**Remaining `Partition()` sites:** `data_file_codec.go:110`, `manifest.go:1462` (manifest *write* hot path, every entry), `table/transaction.go:690`, `table/transaction.go:2102`, `table/conflict_validation.go:519`, `:642`, `:663`, `table/compaction/compaction.go:216`, `table/compaction/pos_delete_collect.go:147`, `:185`, `table/compaction/eq_delete_collect.go:108`, `table/compaction/eq_delete_decision.go:156`.

**Remaining bounds sites:** `table/data_file_meta.go:355`, `:359`.

**`clonePointer` now allocates per call** at: `table/arrow_scanner.go:252` (two heap allocs per `sameDVBlob` comparison, inside a dedup loop, only to dereference both immediately), `table/scanner.go:604`, `:937`, `table/scan_metrics.go:90`, `table/transaction.go:1116`, `:1164`, `:1968`, `:2146`, `:2201`, `table/snapshot_producers.go:640`, `:659`, `:1043`, `:1107`, `table/conflict_validation.go:695`, `table/orphan_cleanup.go:513`, `table/rewrite_data_files.go:580`, `table/compaction/compaction.go:366`, `table/row_delta.go:231`.

**This is structural, not an oversight you can patch by adding call sites.** `dataFileStats` and `dataFilePartition` are unexported in package `table`, so `table/compaction` (7 sites above) cannot reach them, and package `iceberg` (`manifest.go:1462`, `data_file_codec.go:110`) cannot import `table` at all. As written the fast path is unreachable from the two groups that need it most. `manifest.go:1462` and `data_file_codec.go:110` are in-package and can just use `d.initPartitionData()` + `d.fieldIDToPartitionData` directly; `table/compaction` cannot.

**Suggested fix — three options, in the order I'd weigh them:**

1. **Copy-on-construction only.** `DataFile` is immutable after decode or build. Cloning the caller's input maps in `NewDataFileBuilder` and the setters — which this PR already does, and is the genuinely valuable half — removes the input-aliasing bug at *zero* per-call cost. Keep the getters zero-copy and document them read-only. That is the stance `net/http.Header` takes, and it is what both PyIceberg and Java Iceberg do for manifest metrics.
2. **If getters must copy, move the escape hatch into `internal`** (exported there, e.g. `internal.DataFileStats(file)`) rather than unexported in `table`, so `table/compaction` and package `iceberg` can use it too. Then extend it to the sites above.
3. `SortOrderID`/`FirstRowID`/`ContentOffset`/`ContentSizeInBytes`/`ReferencedDataFile` are the cheapest to fix and the most called — a `(T, bool)` accessor buys everything `clonePointer` does with no allocation at all.

## Silent behavioral breaks that need a CHANGELOG line

No exported signature changes, `DataFile` (`manifest.go:2671`) is untouched, and external implementations are unaffected thanks to the fallback at `table/data_file_stats_ref.go:46`. But two runtime behaviors change with no compile error anywhere:

1. **Mutations through getters no longer take effect.** `df.LowerBoundValues()[id] = b` or `df.Partition()[id] = v`, used downstream to patch metadata before a manifest rewrite, is now a silent no-op. That is the intended fix, but it lands as a runtime behavior change in a patch-shaped PR.
2. **Pointer identity is no longer stable.** `clonePointer` (`manifest.go:2236`) makes `df.FirstRowID() != df.FirstRowID()`, and `a.ContentOffset() == b.ContentOffset()` is now always false. Nothing in-repo relies on this (`arrow_scanner.go:252` dereferences), but downstream identity checks or cached pointers break silently.

## Test coverage

Covered well: `manifest_test.go:2972` proves builder inputs are isolated for all ten mutable fields and that returned values are independent across calls; `manifest_test.go:3049` proves the Avro decoder doesn't alias the source buffer; `table/data_file_stats_ref_test.go:103` and `:134` verify the fast path is taken and allocation-free, with `:122` and `:148` covering the fallback.

Most valuable additions, in priority order:

1. **nil vs present-empty through the new clone helpers** — detailed inline on `manifest.go:2223`. This is the one I'd insist on.
2. **A test pinning the `GetPartitionRecord` contract in either direction** (`table/scanner.go:184`) — whichever way blocking issue 1 is resolved, the only externally reachable mutation path deserves a test.
3. **A second `dataFileStats` call after an evaluator has run still returns the original bounds.** Cheap insurance, given that the `sync.Once` at `manifest.go:2094` makes any read-only-contract violation permanent.
4. Decode-path isolation currently exercises only `Partition`/`LowerBoundValues`/`UpperBoundValues`; the other cloned getters are builder-path only. Details inline on `manifest_test.go:3049`.
5. No benchmark covers any of the ~20 regressed sites — at minimum `conflict_validation.go:699` and `snapshots.go:458`.

## Bottom line

Resolve the exported-API hole at `table/scanner.go:185` first — that one is a data-corruption path, not a style question. Then either extend the fast path to the regressed sites, which requires moving it into `internal` since `table/compaction` and package `iceberg` cannot reach it where it lives, or drop the token machinery in favour of copy-on-construction only. The nil-vs-present-empty assertions should land before merge either way.

zeroshade (approved):
This revision resolves both structural blockers the right way. `GetPartitionRecord` cloning binary values closes the last externally-reachable corruption path against the once-cached partition map while keeping internal reads borrowed, and moving the token-gated escape hatch into `internal` — with borrowed accessors for stats, bounds, partition, referenced-file, and all five pointer fields, each falling back to public getters for external implementations — makes the fast path reachable everywhere it was needed: all 56 sites, including `table/compaction` and package `iceberg`, are on it. The nil-vs-present-empty round-trip suite and the two benchmarks lock in both the semantics and the allocation contract. This is now the best of both worlds: input-aliasing closed at construction, hot paths allocation-free, and the read-only borrow contract explicit at every boundary.

One thing before merge, docs-only: please add the two silent behavior changes to the PR description so they reach release notes — mutations through getters are now deliberate no-ops, and pointer-valued getters no longer return identity-stable pointers.

---

laskoviymishka (approved):
That's quite a big one, LGTM! Nice job!

laskoviymishka:
@fallintoplace - conflicts and then good to merge

PR #1577, opened 2026-07-28
fix(io): wrap blob Open failures as path errors

fallintoplace:
## What changed

Wrap blob reader failures from `BlobFileIO.Open` with the existing filesystem error adapter and retain the caller-provided path for preprocessing, validation, and backend failures.

## Why

Backend errors were returned directly, contrary to the `IO.Open` contract. Missing objects did not consistently expose `fs.ErrNotExist` through a `*fs.PathError`, and preprocessing could replace the public path with an empty or internal value.

## Testing

- `go test ./io/gocloud -run 'TestBlobFileIOOpenMissingReturnsPathError|TestDefaultKeyExtractor'`
- `go vet ./io/gocloud`

### DIFF
--- a/io/gocloud/blob.go
+++ b/io/gocloud/blob.go
@@ -249,20 +249,21 @@ func directoryName(key string) string {
 }
 
 func (bfs *BlobFileIO) Open(path string) (icebergio.File, error) {
+	originalPath := path
 	var err error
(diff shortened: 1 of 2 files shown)


laskoviymishka (approved):
🛳️

PR #1578, opened 2026-07-28
fix(io): parse local file URIs consistently

fallintoplace:
## What changed

Centralize LocalFS path conversion with URI-aware parsing and use it across reads, writes, directory operations, metadata operations, removal, and renames.

The parser accepts plain native paths, `file:///path`, `file:/path`, and localhost file URIs. It rejects non-file schemes and non-local authorities instead of turning them into surprising relative paths. Native volume paths remain untouched.

## Why

Blindly trimming only `file://` left valid single-slash file URIs unusable and converted file URI authorities into relative filesystem paths.

## Testing

- `go test ./io`
- `go vet ./io`

### DIFF
--- a/io/local.go
+++ b/io/local.go
@@ -35,15 +35,19 @@ func localPath(name string) (string, error) {
 		return name, nil
 	}
 
+	schemeEnd := strings.IndexByte(name, ':')
+	firstSeparator := strings.IndexAny(name, `/\`)
+	if schemeEnd < 0 || (firstSeparator >= 0 && firstSeparator < schemeEnd) {
+		return name, nil
+	}
+	scheme := name[:schemeEnd]
+	if !strings.EqualFold(scheme, "file") {
+		return "", fmt.Errorf("unsupported local filesystem scheme %q", scheme)
+	}
+
 	parsed, err := url.Parse(name)
 	if err != nil {
 		return "", fmt.Errorf("invalid local file path %q: %w", name, err)
-	}
-	if parsed.Scheme == "" {
-		return name, nil
-	}
-	if !strings.EqualFold(parsed.Scheme, "file") {
-		return "", fmt.Errorf("unsupported local filesystem scheme %q", parsed.Scheme)
 	}
 	if parsed.Host != "" && !strings.EqualFold(parsed.Host, "localhost") {
 		return "", fmt.Errorf("unsupported file URI authority %q", parsed.Host)
--- a/io/local.go
+++ b/io/local.go
@@ -18,7 +18,9 @@
 package io
 
 import (
+	"fmt"
 	"io/fs"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -28,16 +30,54 @@ import (
 // the local file system.
 type LocalFS struct{}
 
+func localPath(name string) (string, error) {
+	if filepath.VolumeName(name) != "" {
+		return name, nil
+	}
+
+	parsed, err := url.Parse(name)
+	if err != nil {
+		return "", fmt.Errorf("invalid local file path %q: %w", name, err)
+	}
+	if parsed.Scheme == "" {
+		return name, nil
+	}
+	if !strings.EqualFold(parsed.Scheme, "file") {
+		return "", fmt.Errorf("unsupported local filesystem scheme %q", parsed.Scheme)
+	}
+	if parsed.Host != "" && !strings.EqualFold(parsed.Host, "localhost") {
+		return "", fmt.Errorf("unsupported file URI authority %q", parsed.Host)
+	}
+	if parsed.Opaque != "" {
+		return parsed.Opaque, nil
+	}
+
+	return parsed.Path, nil
+}
+
 func (LocalFS) Open(name string) (File, error) {
-	return os.Open(strings.TrimPrefix(name, "file://"))
+	path, err := localPath(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return os.Open(path)
 }
 
 func (LocalFS) ReadFile(name string) ([]byte, error) {
-	return os.ReadFile(strings.TrimPrefix(name, "file://"))
+	path, err := localPath(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return os.ReadFile(path)
 }
 
 func (LocalFS) Create(name string) (FileWriter, error) {
-	filename := strings.TrimPrefix(name, "file://")
+	filename, err := localPath(name)
+	if err != nil {
+		return nil, err
+	}
 	if err := os.MkdirAll(filepath.Dir(filename), 0o755); err != nil {
 		return nil, err
 	}
@@ -46,7 +86,10 @@ func (LocalFS) Create(name string) (FileWriter, error) {
 }
 
 func (LocalFS) WriteFile(name string, content []byte) error {
-	filename := strings.TrimPrefix(name, "file://")
+	filename, err := localPath(name)
+	if err != nil {
+		return err
+	}
 	if err := os.MkdirAll(filepath.Dir(filename), 0o755); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


io/local.go
```diff
@@ -28,16 +30,76 @@ import (
...
+
+	parsed, err := url.Parse(name)
+	if err != nil {
+		return "", fmt.Errorf("invalid local file path %q: %w", name, err)
+	}
+	if parsed.Host != "" && !strings.EqualFold(parsed.Host, "localhost") {
+		return "", fmt.Errorf("unsupported file URI authority %q", parsed.Host)
+	}
+	if parsed.Opaque != "" {
+		return filepath.FromSlash(parsed.Opaque), nil
```
laskoviymishka: The two file-URI forms decode differently here. `file:///path/to/my%20file.parquet` goes through `parsed.Path`, which `url.Parse` percent-decodes, so we get `/path/to/my file.parquet` (a real space). But `file:/path/to/my%20file.parquet` (single slash, no authority) lands in `parsed.Opaque`, which is raw, so we hand back the literal `%20`. Same file per RFC 8089, two different OS paths out of `localPath`.

At minimum the opaque branch should match the path branch: `url.PathUnescape(parsed.Opaque)` and handle the error. The bigger question of whether we should be decoding at all (a raw OS path with a literal `%` that got concatenated into a `file://` URL now resolves somewhere else) I've put in the top-level comment, but the two forms disagreeing with each other is a bug either way.

io/local.go
```diff
@@ -28,16 +30,76 @@ import (
...
+	}
+	if !strings.EqualFold(scheme, "file") {
+		return "", fmt.Errorf("unsupported local filesystem scheme %q", scheme)
+	}
+
+	parsed, err := url.Parse(name)
+	if err != nil {
+		return "", fmt.Errorf("invalid local file path %q: %w", name, err)
+	}
+	if parsed.Host != "" && !strings.EqualFold(parsed.Host, "localhost") {
```
laskoviymishka, line 59: I think this regresses Windows file URIs. On Windows, `file://C:/warehouse` parses with `parsed.Host == "C"`, so it hits this branch and errors out, where the old `TrimPrefix` gave `C:/warehouse`, which `os.Open` handles fine. And `file:///C:/warehouse` parses to `parsed.Path == "/C:/warehouse"`, and that leading slash before the drive letter isn't a valid Windows path either.

The new `TestLocalFSParsesFileURIs` builds `"file://" + filepath.ToSlash(path)`, so on a Windows runner it'd construct `file://C:/...` and fail on this exact line. I'd special-case a single-letter drive host (rebuild as `host + ":" + path`) and strip the leading slash before a drive letter. Do we run CI on Windows for this package?

io/local_test.go
```diff
@@ -93,32 +93,112 @@ func TestLocalFSWriteFileCreatesParentDirectories(t *testing.T) {
...
 	} {
 		t.Run(tt.name, func(t *testing.T) {
 			require.NoError(t, LocalFS{}.WriteFile(tt.path, content))
 
-			got, err := os.ReadFile(strings.TrimPrefix(tt.path, "file://"))
+			got, err := os.ReadFile(tt.readPath)
 			require.NoError(t, err)
 			assert.Equal(t, content, got)
 		})
 	}
```
laskoviymishka, line 123: Good test to have, but it only exercises `ReadFile` and only over `TempDir` paths, which never contain percent-encodeable characters, so the opaque-vs-path decode gap ships untested.

I'd add a case that writes a file whose name has a literal space, then reads it back through both `file:/...` and `file:///...` (percent-encoded), asserting they resolve to the same file. And since every write method now routes through `localPath` too, at least one `WriteFile`/`Create` round-trip through a `file://` URI would catch a wrong decoded path before it silently writes to the wrong place.

io/local.go
```diff
@@ -28,16 +30,76 @@ import (
...
+	scheme := name[:schemeEnd]
+	// A colon is valid in a native POSIX filename. Treat non-file names as
+	// URIs only when they have an authority delimiter, preserving paths such as
+	// "partition:2026/data.parquet" while still rejecting s3:// and similar
+	// locations passed directly to LocalFS.
+	if !strings.EqualFold(scheme, "file") && !strings.HasPrefix(name[schemeEnd+1:], "//") {
+		return name, nil
+	}
+	if !strings.EqualFold(scheme, "file") {
+		return "", fmt.Errorf("unsupported local filesystem scheme %q", scheme)
```
laskoviymishka, line 52: Not a blocker, but while we're here: this error and the authority error just below are plain `fmt.Errorf` strings, so a caller doing `errors.Is(err, fs.ErrNotExist)` to spot a missing path now gets a different answer than the old `TrimPrefix` path did, and there's no sentinel to match on if catalog code wants to detect "this isn't a local path, fall back" without string-matching.

If we expect anyone to branch on these, I'd define a sentinel and wrap with `%w`; if not, a one-line comment that the string is the contract is enough. wdyt?

io/local.go
```diff
@@ -55,42 +120,92 @@ func (LocalFS) WriteFile(name string, content []byte) error {
...
 
-func (LocalFS) Mkdir(path string) error {
-	return os.Mkdir(strings.TrimPrefix(path, "file://"), 0o755)
+func (LocalFS) Mkdir(name string) error {
+	path, err := localPath(name)
+	if err != nil {
+		return err
+	}
+
+	return os.Mkdir(path, 0o755)
```
laskoviymishka, line 188: Small readability thing: this shadows the `oldpath` param with `:=`, then the next line reuses `err` with `=`, which is easy to misread. Every other method in the file uses distinct names, so I'd do the same here:

```go
src, err := localPath(oldpath)
if err != nil {
    return err
}
dst, err := localPath(newpath)
if err != nil {
    return err
}

return os.Rename(src, dst)
```

Same pattern in `RenameNoReplace` just below.

io/local_test.go
```diff
@@ -93,32 +93,112 @@ func TestLocalFSWriteFileCreatesParentDirectories(t *testing.T) {
...
 }
 
+func TestLocalFSParsesFileURIs(t *testing.T) {
+	t.Parallel()
+
+	dir := t.TempDir()
+	path := filepath.Join(dir, "metadata.json")
```
laskoviymishka, line 130: Wrapping each iteration in `t.Run(name, ...)` would tell us which URI form broke when one fails. Right now a failure just points at the outer test, which matters more once the Windows forms are in the mix. `TestLocalFSWriteFileCreatesParentDirectories` just above already does this.

laskoviymishka (changes requested):
Nice cleanup. Folding the twelve scattered `strings.TrimPrefix(name, "file://")` calls into one `localPath` helper is the right move, and handling `file:///`, `file:/`, and the `localhost` authority properly is a genuine step up from blind prefix-stripping.

The main thing is that the two file-URI forms don't resolve consistently. `file:///path/to/my%20file.parquet` goes through `parsed.Path`, which `url.Parse` percent-decodes to a real space, while `file:/path/to/my%20file.parquet` lands in `parsed.Opaque` and comes back with the literal `%20`. RFC 8089 treats those as the same file, so `localPath` handing back two different OS paths is a bug regardless of which behavior we want. That raises the bigger question I'd like us to decide explicitly: do we want to percent-decode file URIs at all? Decoding `file://` is arguably the more correct reading, but a lot of the callers in this repo build URIs by naive `"file://" + rawOsPath` concatenation, so a raw OS path that happens to contain a literal `%` sequence would now resolve somewhere else. Either we commit to "file URIs must be percent-encoded per RFC" and decode both forms the same way, or we pass paths through untouched, but it should be one rule, stated, and tested.

The other blocker is Windows. `file://C:/warehouse` parses with `Host == "C"` and gets rejected by the authority check, and `file:///C:/warehouse` comes back as `/C:/warehouse` with a leading slash. Neither is a usable Windows path, where the old `TrimPrefix` produced `C:/warehouse`. The new `TestLocalFSParsesFileURIs` builds `"file://" + filepath.ToSlash(path)`, so it'd fail on a Windows runner as written.

- `file:/` and `file:///` resolve to the same OS path for percent-encoded content (decode the opaque branch, or reject it)
- a stated rule on whether we decode file URIs at all, with a test pinning it: a filename with a literal space, read back through both forms
- Windows `file://C:/...` and `file:///C:/...` resolve correctly, and the new test passes on Windows
- at least one write-path round-trip (`WriteFile`/`Create`) through a `file://` URI, since all twelve methods route through `localPath` now

laskoviymishka (approved):

PR #1579, opened 2026-07-28
fix(rest): surface failures refreshing expired credentials

fallintoplace:
## What changed

Return credential-fetch and filesystem-construction failures when the cached vended credentials are already expired. A cached filesystem may still be reused when a proactive refresh fails before hard expiry.

## Why

The expired path previously returned a known-expired client with a nil error. The subsequent storage request then failed with an authentication error while the actual refresh failure was lost.

This distinction remains compatible with proactive refresh behavior: still-valid credentials can degrade gracefully, while hard-expired credentials fail at the refresh boundary with the actionable cause.

## Testing

- `go test ./catalog/rest`
- `go vet ./catalog/rest`

### DIFF
--- a/catalog/rest/vended_creds.go
+++ b/catalog/rest/vended_creds.go
@@ -136,10 +136,6 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 	default:
 		freshCreds, err := v.fetchCreds(ctx, v.identifier)
 		if err != nil {
-			if !v.expired() {
-				return v.cachedIO, nil
-			}
-
 			return nil, fmt.Errorf("refresh vended credentials for %s: %w", v.location, err)
 		}
 
@@ -151,9 +147,6 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 	if err != nil {
 		if v.cachedIO == nil {
 			return nil, err
-		}
-		if !v.expired() {
-			return v.cachedIO, nil
 		}
 
 		return nil, fmt.Errorf("load filesystem with refreshed credentials for %s: %w", v.location, err)
--- a/catalog/rest/vended_creds.go
+++ b/catalog/rest/vended_creds.go
@@ -136,7 +136,11 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 	default:
 		freshCreds, err := v.fetchCreds(ctx, v.identifier)
 		if err != nil {
-			return v.cachedIO, nil
+			if !v.expired() {
+				return v.cachedIO, nil
+			}
+
+			return nil, fmt.Errorf("refresh vended credentials for %s: %w", v.location, err)
 		}
 
 		config = maps.Clone(v.props)
@@ -145,11 +149,11 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 
 	newIO, err := iceio.LoadFS(ctx, config, v.location)
 	if err != nil {
-		if v.cachedIO != nil {
+		if v.cachedIO != nil && !v.expired() {
 			return v.cachedIO, nil
 		}
 
-		return nil, err
+		return nil, fmt.Errorf("load filesystem with refreshed credentials for %s: %w", v.location, err)
 	}
 
 	v.cachedIO = newIO
--- a/catalog/rest/vended_creds.go
+++ b/catalog/rest/vended_creds.go
@@ -149,7 +149,10 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 
 	newIO, err := iceio.LoadFS(ctx, config, v.location)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


catalog/rest/vended_creds.go
```diff
@@ -136,7 +136,11 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 	default:
 		freshCreds, err := v.fetchCreds(ctx, v.identifier)
 		if err != nil {
-			return v.cachedIO, nil
+			if !v.expired() {
```
laskoviymishka: the `!v.expired()` guard here is dead code, we can't reach it in a not-expired state.

the fast path at the top (`if v.cachedIO != nil && !v.expired()`) already returned for everything that isn't expired, and we only reach `default:` when `cachedIO != nil` and `fetchCreds != nil`. the semaphore's held the whole way through, so `expiresAt` can't move and the clock only advances. by the time we're in here `v.expired()` is always true, so `return v.cachedIO, nil` never fires.

so the real question is the post-expiry policy. if it's "once we're past expiry any refresh failure is terminal," I'd drop the guard and propagate unconditionally. that's also closest to Java, which surfaces all refresh errors through `defaultErrorHandler()` with no degradation. if instead we want to keep serving cached creds through a refresh blip while they're still valid, that's a proactive-refresh window and it has to live in `expiresAt` (a pre-expiry threshold, or a `shouldProactivelyRefresh()` predicate), not a guard inside `default:` that can only run after we're already expired.

which policy are you after here? wdyt?

catalog/rest/vended_creds.go
```diff
@@ -145,11 +149,14 @@ func (v *vendedCredentialRefresher) loadFS(ctx context.Context) (iceio.IO, error
 
 	newIO, err := iceio.LoadFS(ctx, config, v.location)
 	if err != nil {
-		if v.cachedIO != nil {
+		if v.cachedIO == nil {
+			return nil, err
+		}
+		if !v.expired() {
```
laskoviymishka: same dead guard as line 139. after the `cachedIO == nil` early return, `cachedIO != nil` is guaranteed, and in `default:` we're always expired, so this `return v.cachedIO, nil` can't fire either. whichever way the policy above lands, this branch should move the same way.

catalog/rest/vended_creds_test.go
```diff
@@ -153,7 +154,7 @@ func TestVendedCredsConcurrentAccess(t *testing.T) {
 	assert.NotNil(t, r.cachedIO, "cachedIO should be set after initial load")
 }
 
-func TestVendedCredsGracefulDegradation(t *testing.T) {
+func TestVendedCredsReturnsRefreshFailureForExpiredCredentials(t *testing.T) {
```
laskoviymishka, line 157: repurposing this test to assert the error path means we've dropped the only coverage of the degradation contract: refresh fails, cached IO comes back, no error.

whichever behavior we settle on, I'd want it covered. a case with `expiresAt` in the future plus a failing `fetchCreds`, asserting cached IO returns with no error, would lock the contract and, handily, immediately show whether that `!v.expired()` branch is actually reachable.

catalog/rest/vended_creds_test.go
```diff
@@ -185,6 +205,7 @@ func TestVendedCredsErrorWhenInitialLoadFails(t *testing.T) {
 
 	got, err := r.loadFS(context.Background())
 	require.Error(t, err)
+	assert.NotContains(t, err.Error(), "refreshed credentials")
```
laskoviymishka, line 208: minor, not blocking: this matches on a substring of the message, so it'll silently pass if that wrapper string ever changes. a sentinel error plus `errors.Is` would be sturdier if you feel like it.

laskoviymishka (changes requested):
Nice direction here. Propagating refresh errors instead of silently handing back stale cached IO matches how Java's `VendedCredentialsProvider` surfaces everything through `defaultErrorHandler()`, and splitting out the `cachedIO == nil` initial-load case reads cleaner than the old shape.

Almost there, one blocking thing before merge: both `!v.expired()` guards inside `default:` are dead code. The fast path at the top of `loadFS` already returns for anything not-expired, and `default:` is only reachable once we're past expiry (the semaphore's held throughout, so the clock only advances), which means `v.expired()` is always true in there and those `return v.cachedIO, nil` degradation paths never fire. I traced it inline. It's really a policy call: either post-expiry failures are terminal (drop the guards, propagate unconditionally, closest to Java), or we want a proactive-refresh window, in which case it has to live in `expiresAt` rather than a guard that only runs after expiry. Asked which you're after inline.

The rename of `TestVendedCredsGracefulDegradation` also dropped the only coverage of the degrade-on-failure contract, so whichever way the policy lands, I'd add a test back for it.

Aside, not this PR: the flat `adls.sas-token-expires-at-ms` key won't match Java's per-account `...-ms.<account>` suffix, so ADLS configs quietly fall back to the 60-minute default TTL. Pre-existing, just noting it since this PR is the first to really lean on that expiry path.

Once the policy question's settled and the degradation test is back, happy to take another pass.

laskoviymishka (approved):

PR #1580, opened 2026-07-28
fix(puffin): reject trailing data in footer payloads

fallintoplace:
## What changed

After decoding the Puffin footer object, require the bounded footer payload reader to contain only JSON whitespace before EOF.

## Why

A single `json.Decoder.Decode` call accepted the first JSON value and silently ignored a second value or non-whitespace garbage, even though the trailer declared those bytes as part of the footer payload.

Unknown fields within the footer object remain allowed for forward compatibility.

## Testing

- `go test ./puffin`
- `go vet ./puffin`

### DIFF
--- a/puffin/puffin_reader.go
+++ b/puffin/puffin_reader.go
@@ -230,6 +230,9 @@ func (r *Reader) readFooter() error {
 		return fmt.Errorf("puffin: decode footer JSON: %w", err)
 	}
 
+	// FooterPayloadSize defines a single JSON footer object. Reject trailing
+	// content deliberately, even though some other Iceberg implementations
+	// accept padding or additional values inside the footer payload.
 	if decoder.More() {
 		return errors.New("puffin: unexpected content after footer JSON")
 	}
--- a/puffin/puffin_reader.go
+++ b/puffin/puffin_reader.go
@@ -230,10 +230,11 @@ func (r *Reader) readFooter() error {
 		return fmt.Errorf("puffin: decode footer JSON: %w", err)
 	}
 
-	if _, err := decoder.Token(); err == nil {
-		return errors.New("puffin: footer contains multiple JSON values")
-	} else if !errors.Is(err, io.EOF) {
-		return fmt.Errorf("puffin: trailing data after footer JSON: %w", err)
+	if decoder.More() {
+		return errors.New("puffin: unexpected content after footer JSON")
+	}
+	if _, err := decoder.Token(); !errors.Is(err, io.EOF) {
+		return errors.New("puffin: unexpected content after footer JSON")
 	}
 
 	// Validate blob metadata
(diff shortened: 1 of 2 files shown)


puffin/puffin_reader.go
```diff
@@ -224,11 +224,18 @@ func (r *Reader) readFooter() error {
...
 
 	payloadReader := io.NewSectionReader(r.r, footerStart+MagicSize, payloadSize)
+	decoder := json.NewDecoder(payloadReader)
 	var footer Footer
-	if err := json.NewDecoder(payloadReader).Decode(&footer); err != nil {
+	if err := decoder.Decode(&footer); err != nil {
 		return fmt.Errorf("puffin: decode footer JSON: %w", err)
 	}
 
+	if _, err := decoder.Token(); err == nil {
```
laskoviymishka: Could we use `decoder.More()` here instead of `Token()`? It collapses this into a single check — `More()` is true when there's any non-whitespace left in the payload and false at a clean EOF, so we drop the three-way `err == nil` / `io.EOF` / else branch entirely.

It'd also make the messages honest. Right now `{"blobs":[]}42` (a bare scalar) and a truncated `{"blobs":[]} {` both report "multiple JSON values", even though neither is a complete second value — with `More()` it's one path and one message like `unexpected content after footer JSON`. wdyt?

puffin/puffin_test.go
```diff
@@ -580,6 +590,33 @@ func TestReaderInvalidFile(t *testing.T) {
 	})
 }
 
+func TestReaderRejectsTrailingFooterData(t *testing.T) {
+	t.Parallel()
```
laskoviymishka, line 594: Small one — only the outer test calls `t.Parallel()`, so the table rows still run serially inside it. And the sibling `TestReaderInvalidFile` doesn't call it at all, so we've got two patterns in the same file now. I'd either add `t.Parallel()` inside each `t.Run` or drop it from the outer test to match — either's fine.

laskoviymishka (changes requested):
Nice, this is a clean hardening fix — hoisting the decoder so we can probe for a second token after `Decode`, and bounding it on the `SectionReader` so `io.EOF` genuinely means "end of footer payload" rather than end of stream, is exactly the right shape. The test table exercises both the accept and reject cases with real assertions, which I appreciate.

One thing I'd settle before merging though: this makes us stricter than Java and PyIceberg. Both silently accept trailing bytes in the footer payload (Jackson and Pydantic defaults), and the Puffin spec is actually silent on whether the `FooterPayloadSize` region has to be exactly one JSON object — it says the payload "is a single FileMetadata object" but doesn't forbid padding after it. So a third-party writer that pads within `FooterPayloadSize` would get rejected by us and accepted by Java/Py. In practice the risk is low since Java's writer emits exact bytes, and iceberg-rust is strict here too, so there's precedent — I'm not against the rejection. But I'd want it to be a conscious, documented choice: a short comment at the check noting we deliberately reject trailing content even though the spec doesn't require it, so a future reader knows it's intentional and not an oversight.

Settle that, take a look at the `More()` suggestion inline, and this is good to land.

laskoviymishka (approved):

PR #1581, opened 2026-07-28
fix(table): canonicalize partition values used as map keys

fallintoplace:
## What changed

Canonicalize partition values used internally for fanout and clustered-writer lookup. Binary and fixed values use a distinct string-backed key, and float NaNs use stable typed sentinels.

Original partition values remain unchanged for paths and metadata. Byte values are cloned when a partition is first recorded so metadata does not retain Arrow-owned storage. Fixed-size Arrow binary values are also converted to Iceberg literals before grouping.

## Why

Binary and fixed partition values are byte slices, which panic when used as Go map keys. Clustered partition equality could panic for the same reason, while repeated NaNs did not compare equal and could split one logical partition.

## Testing

Coverage exercises binary, fixed, and NaN grouping; source-buffer mutation; clustered equality; and closed-partition tracking.

- `go test ./table`
- `go vet ./table`

### DIFF
--- a/table/partitioned_fanout_writer.go
+++ b/table/partitioned_fanout_writer.go
@@ -55,6 +55,7 @@ type partitionInfo struct {
 
 type partitionFieldInfo struct {
 	sourceField iceberg.PartitionField
+	sourceName  string
 	fieldID     int
 	sourceType  iceberg.Type
 }
@@ -330,6 +331,7 @@ func getRecordPartitions(spec iceberg.PartitionSpec, schema *iceberg.Schema, rec
 		partitionColumns[i] = record.Column(colIndices[0])
 		partitionFieldsInfo[i] = partitionFieldInfo{
 			sourceField: sourceField,
+			sourceName:  colName,
 			fieldID:     sourceField.FieldID,
 			sourceType:  sourceType,
 		}
@@ -343,7 +345,14 @@ func getRecordPartitions(spec iceberg.PartitionSpec, schema *iceberg.Schema, rec
 				sourceField := fieldInfo.sourceField
 				val, err := getArrowValueAsIcebergLiteral(col, int(row), fieldInfo.sourceType)
 				if err != nil {
-					return nil, fmt.Errorf("failed to get arrow values as iceberg literal: %w", err)
+					return nil, fmt.Errorf(
+						"failed to convert source column %q (field ID %d) from Arrow type %s to Iceberg type %s: %w",
+						fieldInfo.sourceName,
+						sourceField.SourceID(),
+						col.DataType(),
+						fieldInfo.sourceType,
+						err,
+					)
 				}
 
 				transformedLiteral := sourceField.Transform.Apply(iceberg.Optional[iceberg.Literal]{Valid: true, Val: val})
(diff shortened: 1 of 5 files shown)


table/partitioned_fanout_writer.go
```diff
@@ -500,6 +547,8 @@ func getArrowValueAsIcebergLiteral(column arrow.Array, row int, sourceType icebe
 	case *array.LargeBinary:
 
 		return iceberg.NewLiteral(arr.Value(row)), nil
+	case *array.FixedSizeBinary:
```
laskoviymishka, line 577: The other cases here — including the `LargeBinary` one right above — have a blank line between the `case` label and the `return`. This one doesn't, so `nlreturn` will fail CI. Add the blank line to match.

table/partitioned_fanout_writer.go
```diff
@@ -500,6 +547,8 @@ func getArrowValueAsIcebergLiteral(column arrow.Array, row int, sourceType icebe
...
+		return iceberg.NewLiteral(arr.Value(row)).To(sourceType)
```
laskoviymishka, line 584: `NewLiteral([]byte)` gives a `BinaryLiteral`, and `.To` only succeeds for `Fixed` (matching length), `Binary`, or `UUID`. A `FixedSizeBinary(4)` against an Iceberg `fixed[3]`, or any other `sourceType`, comes back as a bare `ErrBadCast` that propagates up and aborts the whole write with no column or field ID to go on.

It fails loudly rather than corrupting data, but the diagnostic is rough. I'd assert `sourceType` is Fixed/Binary/UUID up front, or read `ByteWidth` off the Arrow type and build the `FixedLiteral` directly when the widths agree, so a mismatch names the column.

table/partitioned_fanout_writer_test.go
```diff
@@ -282,6 +289,66 @@ func (s *FanoutWriterTestSuite) TestIdentityTransform() {
...
+
+			icebergSchema := iceberg.NewSchema(1, iceberg.NestedField{ID: 1, Name: "part", Type: test.icebergType})
+			spec := iceberg.NewPartitionSpec(iceberg.PartitionField{
+				SourceIDs: []int{1}, FieldID: 1000, Name: "part", Transform: iceberg.IdentityTransform{},
+			})
+
+			partitions, err := getRecordPartitions(spec, icebergSchema, record)
+			s.Require().NoError(err)
+			s.Require().Len(partitions, 2)
+			switch values := record.Column(0).(type) {
```
laskoviymishka, line 316: This is meant to prove the stored key is a clone rather than an alias of the Arrow buffer, but the proof hinges on the mutation taking effect. Arrow-Go doesn't guarantee `Value(row)` hands back writable memory — if a release returns a read-only or copied slice, the mutation no-ops, both keys stay `{1,2,3,4}`, and the test passes whether or not `clonePartitionValue` exists. And with no `default` arm, a column that's neither `Binary` nor `FixedSizeBinary` skips the mutation silently and the assertions pass vacuously.

I'd prove the clone structurally — compare `unsafe.SliceData(partitionRec[0].([]byte))` against the Arrow buffer pointer, or `Release` the record before reading `partitionRec` — and add a `default: s.FailNow(...)`.

laskoviymishka (changes requested):
The canonicalization approach is sound, and collapsing NaN payloads per float width matches Java, so no concern on the core design. A few things before merge:

- the `FixedSizeBinary` case is missing the blank line `nlreturn` wants, so CI fails as-is
- that same case leans on `BinaryLiteral.To`'s length check — a width or type mismatch aborts the whole write with a bare `ErrBadCast` that names no column
- the buffer-mutation test can pass without exercising the clone: it depends on the Arrow buffer being writable and has no `default` arm

Details inline.

laskoviymishka (approved):

PR #1582, opened 2026-07-28
fix(table): calculate snapshot summary totals as int64

fallintoplace:
## What changed

Calculate snapshot summary totals with `GetInt64` and format them with `strconv.FormatInt`.

## Why

Snapshot totals such as record counts and file sizes were parsed through platform-sized `int`. On 32-bit systems, ordinary Iceberg totals above `math.MaxInt32` could truncate or wrap.

The regression test adds values above the 32-bit range and verifies their exact totals.

## Testing

- `go test ./table`
- `go vet ./table`

### DIFF
--- a/table/snapshots_internal_test.go
+++ b/table/snapshots_internal_test.go
@@ -217,3 +217,18 @@ func TestUpdateSnapshotSummariesUnsupportedOperation(t *testing.T) {
 	_, err := updateSnapshotSummaries(Summary{Operation: Operation("scan")}, nil)
 	assert.ErrorIs(t, err, iceberg.ErrNotImplemented)
 }
+
+func TestUpdateSnapshotSummariesPreservesLargeTotals(t *testing.T) {
+	t.Parallel()
+
+	result, err := updateSnapshotSummaries(Summary{Operation: OpAppend, Properties: iceberg.Properties{
+		addedRecordsKey:  "1000000000",
+		addedFileSizeKey: "1000000000",
+	}}, iceberg.Properties{
+		totalRecordsKey:  "3000000000",
+		totalFileSizeKey: "8000000000",
+	})
+	require.NoError(t, err)
+	assert.Equal(t, "4000000000", result.Properties[totalRecordsKey])
+	assert.Equal(t, "9000000000", result.Properties[totalFileSizeKey])
+}
--- a/table/snapshots.go
+++ b/table/snapshots.go
@@ -530,12 +530,12 @@ func updateSnapshotSummaries(sum Summary, previous iceberg.Properties) (Summary,
 	}
 
 	updateTotals := func(totalProp, addedProp, removedProp string) {
-		newTotal := previous.GetInt(totalProp, 0)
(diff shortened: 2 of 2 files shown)


tanmayrauth (approved):
LGTM!

table/snapshots_internal_test.go
```diff
@@ -217,3 +217,18 @@ func TestUpdateSnapshotSummariesUnsupportedOperation(t *testing.T) {
 	_, err := updateSnapshotSummaries(Summary{Operation: Operation("scan")}, nil)
 	assert.ErrorIs(t, err, iceberg.ErrNotImplemented)
 }
+
+func TestUpdateSnapshotSummariesPreservesLargeTotals(t *testing.T) {
+	t.Parallel()
```
laskoviymishka, line 222: small thing — this is the only test in the file that opts into `t.Parallel()`; the siblings (`TestMergeSnapshotSummaries`, `TestSnapshotSummaryCollector`, ...) all run sequentially. It's a pure-function test so there's no race concern either way, but the lone opt-in is a little inconsistent. I'd just drop it to match the file, wdyt?

table/snapshots_internal_test.go
```diff
@@ -217,3 +217,18 @@ func TestUpdateSnapshotSummariesUnsupportedOperation(t *testing.T) {
...
+
+	result, err := updateSnapshotSummaries(Summary{Operation: OpAppend, Properties: iceberg.Properties{
+		addedRecordsKey:  "1000000000",
+		addedFileSizeKey: "1000000000",
+	}}, iceberg.Properties{
+		totalRecordsKey:  "3000000000",
+		totalFileSizeKey: "8000000000",
+	})
+	require.NoError(t, err)
+	assert.Equal(t, "4000000000", result.Properties[totalRecordsKey])
```
laskoviymishka, line 232: the fix is right, but this only exercises 2 of the 6 total paths (records + file-size, addition only) and asserts two individual keys. I'd add a case that removes a large value from a large starting total, and cover one of the file-count totals too — the three `GetInt64` reads are independent, so a partial reversion of just the `removed` one wouldn't be caught here. Asserting the full result `Properties` map (the way `TestMergeSnapshotSummaries` does) would also catch a regression that clobbers the other totals.

Worth knowing too: this proves the int64 math is correct, but it can't guard against a revert to `GetInt` — the truncation only happens on a 32-bit host, and CI only runs tests on amd64/arm64 (s390x just cross-compiles). So on every runner we actually have, the old `GetInt` path would pass this test too. A one-line comment saying so, or a build-tagged (386||arm) assertion, would make the intent clear.

laskoviymishka (approved):
Agreed with @tanmayrauth's LGTM, for an 8-line widening this is about as safe as it gets, and it's the right call: `GetInt64` + `FormatInt` matches Java's `long` totals, PyIceberg's arbitrary-precision ints, and the spec's 64-bit record/byte counts. I checked and this is the only production call site doing the running-total math, so nothing sibling got missed.

Left a couple of inline notes for future

PR #1583, opened 2026-07-28
fix(table): validate position-delete columns before access

fallintoplace:
## What changed

Validate that a position-delete table contains exactly one `file_path` column and exactly one `pos` column before indexing either field.

## Why

The scanner indexed the first result from `FieldIndices` without checking its length. Missing required columns therefore caused an index-out-of-range panic, while duplicate columns were ambiguous.

Regression coverage includes valid schemas, the optional `row` column, each missing-column case, neither column, and duplicate `pos` columns.

## Testing

- `go test ./table`
- `go vet ./table`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -449,13 +449,40 @@ func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_
 	}
 	defer tbl.Release()
 
-	filePathCol := tbl.Column(tbl.Schema().FieldIndices("file_path")[0]).Data()
-	posCol := tbl.Column(tbl.Schema().FieldIndices("pos")[0]).Data()
+	filePathIndex, posIndex, err := positionDeleteColumnIndices(tbl.Schema())
+	if err != nil {
+		return nil, err
+	}
+	filePathCol := tbl.Column(filePathIndex).Data()
+	posCol := tbl.Column(posIndex).Data()
 	if posCol.NullN() > 0 {
 		return nil, fmt.Errorf("%w: null pos in position delete file", iceberg.ErrInvalidSchema)
 	}
 
 	return groupPosDeletesByFilePath(ctx, filePathCol, posCol)
+}
+
+func positionDeleteColumnIndices(schema *arrow.Schema) (int, int, error) {
+	requiredColumn := func(name string) (int, error) {
+		indices := schema.FieldIndices(name)
+		if len(indices) != 1 {
+			return 0, fmt.Errorf("%w: position delete file must contain exactly one %q column, found %d",
+				iceberg.ErrInvalidSchema, name, len(indices))
+		}
+
+		return indices[0], nil
+	}
+
+	filePathIndex, err := requiredColumn("file_path")
+	if err != nil {
+		return 0, 0, err
+	}
+	posIndex, err := requiredColumn("pos")
+	if err != nil {
+		return 0, 0, err
+	}
+
+	return filePathIndex, posIndex, nil
 }
 
 type set[T comparable] map[T]struct{}
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -463,6 +463,8 @@ func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_
 }
 
 func positionDeleteColumnIndices(schema *arrow.Schema) (int, int, error) {
+	// Position-delete columns are resolved by their spec-defined names because
+	// Arrow schemas read from external files do not always retain Iceberg IDs.
 	requiredColumn := func(name string) (int, error) {
 		indices := schema.FieldIndices(name)
 		if len(indices) != 1 {
--- a/table/arrow_scanner_posdelete_regression_test.go
+++ b/table/arrow_scanner_posdelete_regression_test.go
@@ -54,6 +54,7 @@ func TestPositionDeleteColumnIndices(t *testing.T) {
 			if test.wantErr != "" {
 				require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
 				require.ErrorContains(t, err, test.wantErr)
+
 				return
 			}
 
--- a/table/arrow_scanner_posdelete_regression_test.go
+++ b/table/arrow_scanner_posdelete_regression_test.go
@@ -32,6 +32,38 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestPositionDeleteColumnIndices(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		fields  []arrow.Field
+		wantErr string
+	}{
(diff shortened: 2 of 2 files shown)


table/arrow_scanner_posdelete_regression_test.go
```diff
@@ -32,6 +32,39 @@ import (
...
+		{name: "valid with row", fields: []arrow.Field{{Name: "file_path", Type: arrow.BinaryTypes.String}, {Name: "pos", Type: arrow.PrimitiveTypes.Int64}, {Name: "row", Type: arrow.BinaryTypes.String}}},
+		{name: "missing file_path", fields: []arrow.Field{{Name: "pos", Type: arrow.PrimitiveTypes.Int64}}, wantErr: `exactly one "file_path" column, found 0`},
+		{name: "missing pos", fields: []arrow.Field{{Name: "file_path", Type: arrow.BinaryTypes.String}}, wantErr: `exactly one "pos" column, found 0`},
+		{name: "missing both", fields: nil, wantErr: `exactly one "file_path" column, found 0`},
+		{name: "duplicate pos", fields: []arrow.Field{{Name: "file_path", Type: arrow.BinaryTypes.String}, {Name: "pos", Type: arrow.PrimitiveTypes.Int64}, {Name: "pos", Type: arrow.PrimitiveTypes.Int64}}, wantErr: `exactly one "pos" column, found 2`},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			filePathIndex, posIndex, err := positionDeleteColumnIndices(arrow.NewSchema(test.fields, nil))
```
laskoviymishka, line 57: This is the part I'd firm up before merge. The test calls `positionDeleteColumnIndices` directly, so it proves the helper is correct but never exercises the panic path this PR is actually fixing — the lookup inside `readDeletes`.

`TestReadDeletesRejectsNullPos` already has the exact infra we need (in-memory Parquet via `pqarrow.WriteTable` + `iceio.NewMemFS`). I'd add one end-to-end case there: write a delete file whose schema drops `file_path` (or `pos`), call `readDeletes`, and assert `errors.Is(err, ErrInvalidSchema)`. That locks the fix in at the real entry point — right now if the helper ever got inlined or bypassed, nothing here would catch it.

table/arrow_scanner.go
```diff
@@ -449,15 +449,42 @@ func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_
...
 	if posCol.NullN() > 0 {
 		return nil, fmt.Errorf("%w: null pos in position delete file", iceberg.ErrInvalidSchema)
 	}
 
 	return groupPosDeletesByFilePath(ctx, filePathCol, posCol)
 }
 
+func positionDeleteColumnIndices(schema *arrow.Schema) (int, int, error) {
+	requiredColumn := func(name string) (int, error) {
+		indices := schema.FieldIndices(name)
```
laskoviymishka, line 469: Not blocking, and this predates your change — the old `[0]` lookup used `FieldIndices` too.

While we're adding validation here though: `file_path` and `pos` are reserved metadata columns with spec-assigned field IDs (2147483546 / 2147483545), and this same file already resolves lineage columns by reserved ID via `fieldIndexByID` rather than by name. A compliant writer that keeps the field IDs but names the columns differently would hit `ErrInvalidSchema` here instead of being read. Switching these two lookups to `fieldIndexByID` would make it consistent with the rest of the scanner. I'm fine leaving it name-based for this PR — if we do, a one-line comment noting the deliberate name-over-ID choice would help the next reader. wdyt?

table/arrow_scanner_posdelete_regression_test.go
```diff
@@ -32,6 +32,39 @@ import (
...
+		name    string
+		fields  []arrow.Field
+		wantErr string
+	}{
+		{name: "valid", fields: []arrow.Field{{Name: "file_path", Type: arrow.BinaryTypes.String}, {Name: "pos", Type: arrow.PrimitiveTypes.Int64}}},
+		{name: "valid with row", fields: []arrow.Field{{Name: "file_path", Type: arrow.BinaryTypes.String}, {Name: "pos", Type: arrow.PrimitiveTypes.Int64}, {Name: "row", Type: arrow.BinaryTypes.String}}},
+		{name: "missing file_path", fields: []arrow.Field{{Name: "pos", Type: arrow.PrimitiveTypes.Int64}}, wantErr: `exactly one "file_path" column, found 0`},
+		{name: "missing pos", fields: []arrow.Field{{Name: "file_path", Type: arrow.BinaryTypes.String}}, wantErr: `exactly one "pos" column, found 0`},
+		{name: "missing both", fields: nil, wantErr: `exactly one "file_path" column, found 0`},
+		{name: "duplicate pos", fields: []arrow.Field{{Name: "file_path", Type: arrow.BinaryTypes.String}, {Name: "pos", Type: arrow.PrimitiveTypes.Int64}, {Name: "pos", Type: arrow.PrimitiveTypes.Int64}}, wantErr: `exactly one "pos" column, found 2`},
```
laskoviymishka, line 51: `requiredColumn` handles `file_path` and `pos` symmetrically, but only the `pos` duplicate is covered. I'd add the mirror case — `{file_path, file_path, pos}` expecting `exactly one "file_path" column, found 2` — so both directions are pinned.

table/arrow_scanner_posdelete_regression_test.go
```diff
@@ -32,6 +32,39 @@ import (
...
+			filePathIndex, posIndex, err := positionDeleteColumnIndices(arrow.NewSchema(test.fields, nil))
+			if test.wantErr != "" {
+				require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
+				require.ErrorContains(t, err, test.wantErr)
+
+				return
+			}
+
+			require.NoError(t, err)
+			assert.Equal(t, 0, filePathIndex)
```
laskoviymishka: Both valid cases assert `filePathIndex == 0` / `posIndex == 1` unconditionally, so this only really checks the fixed ordering. A case with `pos` before `file_path` (legal for Arrow) with the expected indices parameterized per test would prove the lookup resolves by name rather than by position. Minor.

laskoviymishka (changes requested):
Almost there — the fix itself is right. The old `FieldIndices("file_path")[0]` would panic on a delete file missing `file_path` or `pos`, and returning `ErrInvalidSchema` instead is the correct call. The schema comes off an untrusted Parquet file in object storage, so validating it is worth doing, and the helper plus its table test read cleanly.

The gap before merge: the test only calls `positionDeleteColumnIndices` directly, so it never drives the panic path through `readDeletes` — the thing this actually fixes. `TestReadDeletesRejectsNullPos` already has the in-memory Parquet setup for this; a case there that feeds a delete file with a missing column through `readDeletes` and asserts `errors.Is(err, ErrInvalidSchema)` would pin the fix at the real entry point.

The rest is minor and inline: a symmetric `duplicate file_path` case, the hardcoded 0/1 index assertions, and an optional thought about resolving these reserved columns by field ID the way the lineage columns already are. Add that end-to-end case and this is good to go.

laskoviymishka (approved):

PR #1584, opened 2026-07-28
fix(io/gcs): propagate application-default credential errors

fallintoplace:
## Summary

Return application-default credential discovery errors from GCS configuration.

## Why

ADC errors were discarded and a nil credential result was treated as anonymous access. Misconfigured credentials could then fail later with an unrelated storage authorization error.

## What changed

Credential discovery errors now include source context. Anonymous access remains available explicitly through gcs.no-auth, which still takes precedence.

## Tests

- go test ./io/gocloud -run TestGCSCredentials
- go vet ./io/gocloud

### DIFF
--- a/io/gocloud/gcs.go
+++ b/io/gocloud/gcs.go
@@ -131,7 +131,10 @@ func gcsCredentials(ctx context.Context, props map[string]string) (*google.Crede
 		return parse(data, fmt.Sprintf("%s %q", io.GCSKeyPath, path))
 	}
 
-	creds, _ := gcp.DefaultCredentials(ctx)
+	creds, err := gcp.DefaultCredentials(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("gcs: loading application default credentials: %w", err)
+	}
 
 	return creds, nil
 }
--- a/io/gocloud/gcs.go
+++ b/io/gocloud/gcs.go
@@ -132,7 +132,7 @@ func gcsCredentials(ctx context.Context, props map[string]string) (*google.Crede
 	}
 
 	creds, err := gcp.DefaultCredentials(ctx)
-	if err != nil {
+	if err != nil && os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") != "" {
 		return nil, fmt.Errorf("gcs: loading application default credentials: %w", err)
 	}
 
--- a/io/gocloud/gcs.go
+++ b/io/gocloud/gcs.go
@@ -132,7 +132,7 @@ func gcsCredentials(ctx context.Context, props map[string]string) (*google.Crede
 	}
 
 	creds, err := gcp.DefaultCredentials(ctx)
-	if err != nil && os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") != "" {
+	if err != nil {
 		return nil, fmt.Errorf("gcs: loading application default credentials: %w", err)
 	}
 
--- a/io/gocloud/gcs.go
+++ b/io/gocloud/gcs.go
@@ -86,6 +86,14 @@ func ParseGCSConfig(props map[string]string) *gcsblob.Options {
 const gcsScope = "https://www.googleapis.com/auth/devstorage.read_write"
 
 func gcsCredentials(ctx context.Context, props map[string]string) (*google.Credentials, error) {
+	return gcsCredentialsWithDefault(ctx, props, gcp.DefaultCredentials)
+}
+
+func gcsCredentialsWithDefault(
+	ctx context.Context,
+	props map[string]string,
+	defaultCredentials func(context.Context) (*google.Credentials, error),
+) (*google.Credentials, error) {
 	// Explicit no-auth wins over everything: use an anonymous client even when
 	// ADC is available.
 	if noAuth, err := strconv.ParseBool(props[io.GCSNoAuth]); err == nil && noAuth {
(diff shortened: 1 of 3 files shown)


tanmayrauth (approved):
LGTM!

io/gocloud/gcs.go
```diff
@@ -131,7 +131,10 @@ func gcsCredentials(ctx context.Context, props map[string]string) (*google.Crede
 		return parse(data, fmt.Sprintf("%s %q", io.GCSKeyPath, path))
 	}
 
-	creds, _ := gcp.DefaultCredentials(ctx)
+	creds, err := gcp.DefaultCredentials(ctx)
```
laskoviymishka: agreed that discarding the ADC error here was a real bug — silently returning nil creds and falling through to anonymous access hides genuine auth failures.

the thing I'd want sorted before merge is that this also flips behavior for anyone who was relying on that silent fallback: public buckets, the fake-gcs emulator, or any non-GCP env without ADC used to quietly get an anonymous client, and now they hard-fail the moment they open a `gs://` path. Java defers ADC lazily via `StorageOptions` and never fails eagerly at FileIO init, so we'd be diverging from the Java client here.

two ways I could see going — preserve the old `nil, nil` → anonymous semantics when `DefaultCredentials` fails, or keep the eager error but make it self-documenting so the migration path is obvious:

```go
return nil, fmt.Errorf("gcs: no credentials found; set %s=true for anonymous/public bucket access: %w", io.GCSNoAuth, err)
```

I lean toward the second since failing loud on a real auth error is the better default, but either way I'd call this out as a breaking change in the PR description. wdyt?

io/gocloud/gcs.go
```diff
@@ -131,7 +131,10 @@ func gcsCredentials(ctx context.Context, props map[string]string) (*google.Crede
...
+	if err != nil {
```
laskoviymishka, line 143: small defensive one while we're here: if `DefaultCredentials` ever returns `(nil, nil)`, we fall straight back to the anonymous path — the exact silent fallback this PR is removing. gocloud's `gcp` doesn't do that today, but the contract isn't enforced at the call site. a `if creds == nil` guard before the return would close it off.

io/gocloud/gcs_integration_test.go
```diff
@@ -115,6 +115,7 @@ func (s *GCSIOTestSuite) TestGCSWarehouse() {
 		"type":            "sql",
 		"warehouse":       fmt.Sprintf("gs://%s/iceberg/", gcsBucketName),
 		io.GCSEndpoint:    fmt.Sprintf("http://%s/", gcsEndpoint),
+		io.GCSNoAuth:      "true",
```
laskoviymishka: this is the right fix for the emulator — fake-gcs-server serves no real ADC, so without no-auth the suite would now hard-fail in CI. worth a one-line comment saying exactly that, though; otherwise it reads as an unexplained flag and someone may drop it later thinking it's redundant.

io/gocloud/gcs_test.go
```diff
@@ -142,6 +142,19 @@ func TestGCSCredentialsNoAuth(t *testing.T) {
 	assert.Nil(t, creds)
 }
 
+func TestGCSCredentialsPropagatesADCError(t *testing.T) {
+	t.Setenv("GOOGLE_APPLICATION_CREDENTIALS", filepath.Join(t.TempDir(), "missing.json"))
```
laskoviymishka: good case to have. it forces ADC to fail at the env-var step, but it doesn't cover the more common dev failure — no env var set and no creds anywhere (fresh machine, container with nothing mounted), where ADC falls through the well-known-file and metadata-server lookups before failing. a sub-case with `t.Setenv("GOOGLE_APPLICATION_CREDENTIALS", "")` would exercise that path too.

io/gocloud/gcs_test.go
```diff
@@ -142,6 +142,19 @@ func TestGCSCredentialsNoAuth(t *testing.T) {
...
+
+	creds, err := gcsCredentials(context.Background(), nil)
+	require.Error(t, err)
```
laskoviymishka: `require.ErrorContains` already implies the error is non-nil, so the `require.Error` above it is redundant — I'd drop it.

laskoviymishka (changes requested):
Nice catch on the discarded error — `creds, _ := gcp.DefaultCredentials(ctx)` was swallowing real ADC failures and quietly falling back to anonymous access, so propagating it is the right instinct.

I'd hold before merge on one thing, though. This is a behavioral change, not only an error fix: anyone relying on the silent anonymous fallback — public buckets, the fake-gcs emulator, any non-GCP env without ADC — now hard-fails the moment they open a `gs://` path, where they used to get an anonymous client. Java defers ADC lazily and never fails eagerly at FileIO init, so we'd also be diverging from it. I'd either preserve the old `nil, nil` → anonymous semantics or keep the eager error but make it self-documenting by naming `gcs.no-auth` in the message, and call this out as breaking in the description. I lean toward failing loud with the better message.

Left a few minors inline too. Once the compat question's settled, happy to take another pass.

laskoviymishka (approved):

io/gocloud/gcs_test.go
```diff
@@ -142,6 +144,40 @@ func TestGCSCredentialsNoAuth(t *testing.T) {
...
+	})
+
+	t.Run("nil credentials", func(t *testing.T) {
+		creds, err := gcsCredentialsWithDefault(context.Background(), nil,
+			func(context.Context) (*google.Credentials, error) { return nil, nil })
+		require.ErrorContains(t, err, "gcs: no credentials found")
+		assert.Nil(t, creds)
+	})
+
+	creds, err := gcsCredentials(context.Background(), map[string]string{io.GCSNoAuth: "true"})
```
zeroshade, line 189: Non-blocking: this repeats the existing `TestGCSCredentialsNoAuth` assertion. Consider calling `gcsCredentialsWithDefault` here with `gcs.no-auth=true` and a `defaultCredentials` function that panics; that would prove the anonymous path short-circuits ADC lookup, which is the more valuable inverse property.

zeroshade (commented):
This change is correct: I verified that ADC failures are genuinely propagated, the error path cannot leak a bucket name or credential material, and the explicit `gcs.no-auth` path still constructs an anonymous client.

The main non-blocking compatibility note is that this is a real breaking change. Missing ADC previously fell through to an anonymous client; public-bucket and fake-GCS/emulator users now fail unless they set `gcs.no-auth`. The integration test's new `io.GCSNoAuth` setting at `io/gocloud/gcs_integration_test.go:118-120` demonstrates the behavior change in-repo. Consider adding a release-note or breaking-change line. No `gcs.*` property is currently documented elsewhere in the repository, so the error text is the only discoverability path for the opt-out.

Worth a follow-up at `io/gocloud/gcs.go:157-160`: a small `createGCSBucket`-level test could pin that the new ADC error reaches the production caller, rather than only testing the credential helper directly.

PR #1585, opened 2026-07-28
fix(literals): reject oversized fixed-width values

fallintoplace:
## What changed

Continue padding shortened fixed-width bounds for compatibility, but reject inputs larger than the declared fixed width with `ErrInvalidBinSerialization`.

## Why

The previous `len != width` branch copied every mismatched value into a width-sized buffer. Oversized values were silently truncated into a different valid value, which could corrupt bounds used for pruning.

Regression coverage includes exact, shortened, empty, one-byte oversized, and substantially oversized inputs, with expected and actual lengths asserted for oversize errors.

## Testing

- `go test .`
- `go vet .`

### DIFF
--- a/literals.go
+++ b/literals.go
@@ -201,8 +201,8 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
 			copy(padded, data)
 			data = padded
 		} else if len(data) > t.Len() {
-			return nil, fmt.Errorf("%w: fixed[%d] value has %d bytes",
-				ErrInvalidBinSerialization, t.Len(), len(data))
+			return nil, fmt.Errorf("%w: %w: fixed[%d] value has %d bytes",
+				ErrInvalidBinSerialization, ErrInvalidFixedLength, t.Len(), len(data))
 		}
 		var v FixedLiteral
 		err := v.UnmarshalBinary(data)
--- a/literals.go
+++ b/literals.go
@@ -193,13 +193,16 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
 
 		return GeoLiteral{val: data, typ: typ}, nil
 	case FixedType:
-		if len(data) != t.Len() {
+		if len(data) < t.Len() {
 			// looks like some writers will write a prefix of the fixed length
 			// for lower/upper bounds instead of the full length. so let's pad
 			// it out to the full length if unpacking a fixed length literal
 			padded := make([]byte, t.Len())
 			copy(padded, data)
 			data = padded
+		} else if len(data) > t.Len() {
+			return nil, fmt.Errorf("%w: fixed[%d] value has %d bytes",
+				ErrInvalidBinSerialization, t.Len(), len(data))
 		}
 		var v FixedLiteral
 		err := v.UnmarshalBinary(data)
--- a/literals_test.go
+++ b/literals_test.go
@@ -19,7 +19,6 @@ package iceberg_test
 
 import (
 	"encoding/binary"
-	"fmt"
 	"math"
 	"strconv"
 	"testing"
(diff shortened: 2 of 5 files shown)


literals.go
```diff
@@ -193,13 +193,16 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
...
-		if len(data) != t.Len() {
+		if len(data) < t.Len() {
 			// looks like some writers will write a prefix of the fixed length
 			// for lower/upper bounds instead of the full length. so let's pad
 			// it out to the full length if unpacking a fixed length literal
 			padded := make([]byte, t.Len())
 			copy(padded, data)
 			data = padded
+		} else if len(data) > t.Len() {
+			return nil, fmt.Errorf("%w: fixed[%d] value has %d bytes",
```
tanmayrauth: Worth flagging the failure-mode change this introduces: the manifest evaluators panic(err) on any LiteralFromBytes error (evaluators.go:172, 182, and ~34 other call sites), so an oversized fixed bound now crashes the scan instead of silently mis-pruning. That's a net improvement — a loud panic beats silent wrong results — and the panic(err) is pre-existing and shared by every error branch, so not something to fix here. But since the comment right above concedes non-compliant writers exist (they write short prefixes), a writer emitting oversized fixed bounds would now hard-crash pruning. If graceful degradation is preferred, the real fix is at the call sites: treat a bound-parse error as rowsMightMatch rather than panicking. Fine to leave that out of this PR — just make sure the crash-on-malformed-bound behavior is intended.

tanmayrauth (commented):
Fix looks fine — silently truncating an oversized value into a valid-but-wrong bound is a real pruning-correctness bug, and rejecting matches the spec/Java. One heads-up on the resulting failure mode

literals.go
fallintoplace: I’ll keep this PR scoped to preventing truncation.

literals.go
```diff
@@ -193,13 +193,16 @@ func LiteralFromBytes(typ Type, data []byte) (Literal, error) {
...
 	case FixedType:
-		if len(data) != t.Len() {
+		if len(data) < t.Len() {
 			// looks like some writers will write a prefix of the fixed length
 			// for lower/upper bounds instead of the full length. so let's pad
 			// it out to the full length if unpacking a fixed length literal
 			padded := make([]byte, t.Len())
 			copy(padded, data)
 			data = padded
+		} else if len(data) > t.Len() {
```
laskoviymishka, line 203: the reject itself is right, but I think there's a subtle blast radius here worth settling before merge.

I traced it through: every `LiteralFromBytes` call in `table/evaluators.go` wraps the error in `panic()`, and `VisitExpr` recovers that into a scan-aborting error. so an oversized fixed bound on one manifest entry now takes down the entire scan — where before it truncated and kept going. Java and PyIceberg both tolerate any-length fixed bounds and just keep evaluating, so this ends up being a Go-only read regression.

I'd keep the strict error here but degrade at the callers: when `LiteralFromBytes` returns `ErrInvalidBinSerialization` for a bound, have the evaluator treat that bound as unknown and return `rowsMightMatch` instead of panicking. wdyt?

literals_test.go
```diff
@@ -1201,6 +1202,41 @@ func TestUnmarshalBinary(t *testing.T) {
...
+
+	tests := []struct {
+		name    string
+		data    []byte
+		want    iceberg.FixedLiteral
+		wantErr bool
+	}{
+		{name: "exact", data: []byte{1, 2, 3, 4}, want: iceberg.FixedLiteral{1, 2, 3, 4}},
+		{name: "short", data: []byte{1, 2, 3}, want: iceberg.FixedLiteral{1, 2, 3, 0}},
+		{name: "empty", data: nil, wantErr: true},
```
laskoviymishka: small thing on the table: the `empty` case passes because `data: nil` trips the nil guard at the top of `LiteralFromBytes`, before the `FixedType` branch ever runs — so it's not actually exercising the new path, and the name is a bit misleading.

the interesting gap is a non-nil `[]byte{}`, which hits the pad branch and comes back as `FixedLiteral{0,0,0,0}` today with nothing covering it. I'd rename this one to `nil` and add a zero-length case documenting that pad behavior.

literals_test.go
```diff
@@ -1201,6 +1202,41 @@ func TestUnmarshalBinary(t *testing.T) {
...
+		{name: "one byte oversized", data: []byte{1, 2, 3, 4, 5}, wantErr: true},
+		{name: "large oversized", data: make([]byte, 100), wantErr: true},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			literal, err := iceberg.LiteralFromBytes(iceberg.FixedTypeOf(4), test.data)
+			if test.wantErr {
+				require.ErrorIs(t, err, iceberg.ErrInvalidBinSerialization)
+				if len(test.data) > 4 {
```
laskoviymishka: the `> 4` guard hard-codes the type width — if someone bumps this to `FixedTypeOf(8)` the message assertions silently stop running and the oversized cases go green without checking anything.

I'd move the expected substring onto the table (an `errContains` field on the oversized rows) so each case asserts its own message and the conditional drops out.

laskoviymishka (changes requested):
Nice catch on the underlying bug — padding an oversized value into a valid-but-wrong bound is exactly the kind of silent pruning error that's miserable to track down, and rejecting it is the right instinct. Agreed with tanmayrauth that the direction's sound.

One thing I'd want to settle before merge, and it's the failure mode he flagged from a slightly different angle: I traced the reject path and it flows straight into `panic()` at every `LiteralFromBytes` call site in `table/evaluators.go`, which `VisitExpr` recovers into a scan-aborting error. So a single oversized fixed bound on one manifest entry now makes the whole table unreadable from Go — where Java and PyIceberg tolerate any-length fixed bounds and just keep evaluating. That's a Go-only read regression, and arguably a bit worse than the bug it replaces.

I'd keep the strict error where it is and degrade at the evaluator callers instead — treat an unreadable bound as `rowsMightMatch` rather than panicking. The test table also has a small coverage gap I left inline. Once the panic path is sorted, happy to take another pass.

laskoviymishka (approved):

PR #1586, opened 2026-07-28
fix(types): floor negative nanosecond timestamps

fallintoplace:
## What changed

Use signed floor division when converting nanosecond timestamps to microseconds. Add boundary coverage around zero and verify casts to both timestamp variants.

## Why

Go integer division truncates toward zero. For pre-epoch timestamps with sub-microsecond precision, that could move the converted value forward by one microsecond and even cross the Unix epoch.

## Testing

- `go test .`

### DIFF
--- a/literals_test.go
+++ b/literals_test.go
@@ -201,8 +201,8 @@ func TestTimestampNanoLiteralConversions(t *testing.T) {
 		{
 			name:              "negative timestamp",
 			nanoLit:           iceberg.TimestampNsLiteral(-1234567890123456789),
-			expectedMicro:     iceberg.Timestamp(-1234567890123456),
-			expectedRoundTrip: iceberg.TimestampNano(-1234567890123456000),
+			expectedMicro:     iceberg.Timestamp(-1234567890123457),
+			expectedRoundTrip: iceberg.TimestampNano(-1234567890123457000),
 		},
 		{
 			name:              "maximum precision truncation",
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM, thanks for the fix!

literals_test.go
```diff
@@ -201,8 +201,8 @@ func TestTimestampNanoLiteralConversions(t *testing.T) {
 		{
 			name:              "negative timestamp",
 			nanoLit:           iceberg.TimestampNsLiteral(-1234567890123456789),
-			expectedMicro:     iceberg.Timestamp(-1234567890123456),
-			expectedRoundTrip: iceberg.TimestampNano(-1234567890123456000),
+			expectedMicro:     iceberg.Timestamp(-1234567890123457),
+			expectedRoundTrip: iceberg.TimestampNano(-1234567890123457000),
```
laskoviymishka, line 205: The new value is right — `floor(-1234567890123456.789 µs) = -1234567890123457`, then `*1000`.

I'd add a one-line comment here though. This case was silently wrong for years before the PR, and the asymmetry with the positive case (where we just trim the last 3 digits) is exactly the shape a future reader "corrects" back to the intuitive-but-wrong value. Something like `// floor(-…456.789 µs) = -…457 µs, then *1000` would save that trip. wdyt?

literals_test.go
```diff
@@ -235,6 +235,36 @@ func TestTimestampNanoLiteralConversions(t *testing.T) {
...
+		{nanos: -999, micros: -1},
+		{nanos: -1, micros: -1},
+		{nanos: 0, micros: 0},
+		{nanos: 1, micros: 0},
+		{nanos: 999, micros: 0},
+		{nanos: 1000, micros: 1},
+		{nanos: 1001, micros: 1},
+	}
+
+	for _, tt := range tests {
```
laskoviymishka, line 254: Small thing: this table drops the `name` field and `t.Run` that the other table tests in this file use, so a failing case just points at the assert line with no hint of which `{nanos, micros}` pair broke.

It also mixes `assert` (non-fatal) with the `require.NoError` a few lines down — and since `require` kills the goroutine, an error out of the `To()` path aborts the whole test before the later cases ever run. Wrapping the loop body in `t.Run(tt.name, ...)` fixes both at once (`require` inside a subtest only fails that subtest). Not blocking.

types.go
```diff
@@ -799,7 +799,7 @@ func (t TimestampNano) ToTime() time.Time {
 }
 
 func (t TimestampNano) ToMicros() Timestamp {
-	return Timestamp(int64(t) / 1000)
+	return Timestamp(internal.FloorDiv(int64(t), 1000))
```
laskoviymishka, line 802: This is the right change and it's the last raw-truncation site, so no objection.

Optional: since this exported method now has a specific rounding contract, a one-line godoc noting it floors toward negative infinity (matching Java's `Math.floorDiv`) would keep someone from "simplifying" it back to `/ 1000` later.

laskoviymishka (approved):
The fix is right — flooring toward negative infinity is the correct semantic for a ns→µs downcast, and it lines up with Java's `Math.floorDiv` and PyIceberg's `//`. Nice catch that this was the last raw `/ 1000` site; every other ns→coarser conversion was already going through `FloorDiv`, so this just closes the gap.

Nothing blocking from me, a couple of small things inline, all optional.

:shipit:

PR #1587, opened 2026-07-28
fix(puffin): constrain DV cardinality to int64

fallintoplace:
## What changed

Validate deletion-vector cardinality properties as non-negative signed 64-bit values. Add coverage for the maximum accepted value and the first value above the signed range.

## Why

The writer accepted the full `uint64` range while the deletion-vector reader and manifest record counts use `int64`. This allowed writing Puffin files that the reader could not parse.

## Testing

- `go test ./puffin`

### DIFF
--- a/puffin/puffin_test.go
+++ b/puffin/puffin_test.go
@@ -22,6 +22,7 @@ import (
 	"math"
 	"os"
 	"path"
+	"strconv"
 	"testing"
 
 	"github.com/apache/iceberg-go/puffin"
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM!

puffin/puffin_test.go
```diff
@@ -347,6 +348,28 @@ func TestWriterValidation(t *testing.T) {
 		assert.ErrorContains(t, err, "not a valid non-negative integer")
 	})
 
+	t.Run("deletion vector maximum cardinality", func(t *testing.T) {
```
laskoviymishka, line 351: While we're adding these subtests — the `ParseUint` comments on the two sibling cases just above (~328 and ~344) are now stale. The impl moved to `ParseInt` + an explicit `< 0` guard, so "ParseUint covers both" / "ParseUint catches this via the same invalid-syntax path" is describing a function that's no longer called.

Behavior described is still correct, just the mechanism name drifted. I'd update them to `ParseInt` in the same pass.

puffin/puffin_writer.go
```diff
@@ -158,15 +158,17 @@ func (w *Writer) AddBlob(input BlobMetadataInput, data []byte) (BlobMetadata, er
...
-		if _, err := strconv.ParseUint(properties["cardinality"], 10, 64); err != nil {
+		// Parse the same signed range used by deletion-vector readers and
+		// manifest record counts so every emitted value is readable.
+		cardinality, err := strconv.ParseInt(properties["cardinality"], 10, 64)
+		if err != nil {
 			return BlobMetadata{}, fmt.Errorf("puffin: deletion-vector-v1 cardinality property %q is not a valid non-negative integer: %w",
 				properties["cardinality"], err)
 		}
+		if cardinality < 0 {
+			return BlobMetadata{}, fmt.Errorf("puffin: deletion-vector-v1 cardinality property %q is not a valid non-negative integer",
```
laskoviymishka, line 169: The parse-error path above wraps with `%w` and this one doesn't — which is right, there's no underlying error to wrap here. Only thing is it's easy to skim and read as an accidental `%w` drop. A trailing `// no underlying error to wrap` on the `fmt.Errorf` would make the asymmetry intentional at a glance. Non-blocking.

laskoviymishka (approved):
Nice, clean fix. Moving the writer to `ParseInt` + a `< 0` guard makes it symmetric with the DV reader and with Java's `DeletionVector` field 156 (`LongType`), so we stop emitting cardinalities in `(MaxInt64, MaxUint64]` that nothing downstream — our own reader, Java, or the int64 manifest `record_count` — could ever read back. Worth calling out that this narrows the write-side range rather than breaking anyone: those values never round-tripped through any current client, so it's a bug fix, not an API change.

The boundary-pair test is the right one too — `MaxInt64` accepted, `MaxInt64+1` rejected, and 2^63 being the theoretical max distinct positions anyway is a nice detail.

Left couple thing inline for future, not blocking, happy to merge!

:shipit:

PR #1588, opened 2026-07-28
fix(substrait): preserve nested reference paths

fallintoplace:
## What changed

Build Substrait field references from the positional path retained by the bound Iceberg reference. Add coverage for a colliding top-level leaf name, a missing top-level leaf name, and a deeply nested field.

## Why

The converter rebound only the leaf field name. Nested references could therefore target a same-named top-level column or fail conversion when no top-level field shared that name.

## Testing

- `go test ./table/substrait`

### DIFF
--- a/table/substrait/substrait.go
+++ b/table/substrait/substrait.go
@@ -51,9 +51,10 @@ func NewExtensionSet() exprs.ExtensionIDSet {
 	return exprs.NewExtensionSetDefault(expr.NewEmptyExtensionRegistry(collection))
 }
 
-// ConvertExpr binds the provided expression to the given schema and converts it to a
-// substrait expression so that it can be utilized for computation.
-func ConvertExpr(schema *iceberg.Schema, e iceberg.BooleanExpression, caseSensitive bool) (*expr.ExtensionRegistry, expr.Expression, error) {
+// ConvertExpr converts a bound expression to a Substrait expression so that it can
+// be utilized for computation. Case sensitivity is applied when binding; the third
+// argument is retained for API compatibility.
+func ConvertExpr(schema *iceberg.Schema, e iceberg.BooleanExpression, _ bool) (*expr.ExtensionRegistry, expr.Expression, error) {
 	base, err := ConvertSchema(schema)
 	if err != nil {
 		return nil, nil, err
@@ -62,10 +63,7 @@ func ConvertExpr(schema *iceberg.Schema, e iceberg.BooleanExpression, caseSensit
 	reg := expr.NewEmptyExtensionRegistry(collection)
 
 	bldr := expr.ExprBuilder{Reg: reg, BaseSchema: types.NewRecordTypeFromStruct(base.Struct)}
-	b, err := iceberg.VisitExpr(e, &toSubstraitExpr{
-		bldr: bldr, schema: schema,
-		caseSensitive: caseSensitive,
-	})
+	b, err := iceberg.VisitExpr(e, &toSubstraitExpr{bldr: bldr})
 	if err != nil {
 		return nil, nil, err
 	}
@@ -213,9 +211,7 @@ var (
 )
 
 type toSubstraitExpr struct {
-	schema        *iceberg.Schema
-	bldr          expr.ExprBuilder
-	caseSensitive bool
+	bldr expr.ExprBuilder
 }
 
 func (t *toSubstraitExpr) VisitTrue() expr.Builder {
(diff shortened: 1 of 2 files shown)


table/substrait/substrait.go
```diff
@@ -348,12 +348,7 @@ func toSubstraitLiteralSet(typ iceberg.Type, lits []iceberg.Literal) expr.ListLi
 }
 
 func (t *toSubstraitExpr) getRef(ref iceberg.BoundReference) expr.Reference {
-	updatedRef, err := iceberg.Reference(ref.Field().Name).Bind(t.schema, t.caseSensitive)
-	if err != nil {
-		panic(err)
-	}
-
-	path := updatedRef.Ref().PosPath()
+	path := ref.PosPath()
```
tanmayrauth, line 347: Deleting the rebind also leaves some dead plumbing worth cleaning up in the same PR: getRef was the only reader of t.schema and t.caseSensitive, so both struct fields (lines 216-218) are now write-only, and ConvertExpr's public `caseSensitive` parameter is now entirely unused — its only consumer was this rebind, and callers already bind with case sensitivity upstream (arrow_scanner.go:748 calls iceberg.BindExpr(fileSchema, translatedFilter, as.caseSensitive) right before ConvertExpr). Net effect: a caller who flips caseSensitive on ConvertExpr expecting case-insensitive matching silently gets nothing, and go vet won't flag the write-only fields. Suggest dropping the two struct fields and the constructor args now; for the public param, either remove it (breaking, so worth a maintainer call) or keep it and document that case sensitivity is applied at bind time.

tanmayrauth (commented):
Fix is correct, pulling the path straight off the bound ref fixes both the wrong-column collision and the panic-on-missing-name, and the tests pin both. Two cleanup nits from deleting the rebind

table/substrait/substrait.go
```diff
@@ -51,9 +51,10 @@ func NewExtensionSet() exprs.ExtensionIDSet {
 	return exprs.NewExtensionSetDefault(expr.NewEmptyExtensionRegistry(collection))
 }
 
-// ConvertExpr binds the provided expression to the given schema and converts it to a
-// substrait expression so that it can be utilized for computation.
-func ConvertExpr(schema *iceberg.Schema, e iceberg.BooleanExpression, caseSensitive bool) (*expr.ExtensionRegistry, expr.Expression, error) {
+// ConvertExpr converts a bound expression to a Substrait expression so that it can
+// be utilized for computation. Case sensitivity is applied when binding; the third
+// argument is retained for API compatibility.
+func ConvertExpr(schema *iceberg.Schema, e iceberg.BooleanExpression, _ bool) (*expr.ExtensionRegistry, expr.Expression, error) {
```
laskoviymishka, line 57: The doc still says "case sensitivity is applied when binding," but ConvertExpr doesn't bind anymore — it now assumes `e` is already bound and the third arg is dead. Harmless today since both callers pre-bind, but the comment points future readers the wrong way; I'd reword it to say the expr must be pre-bound and the bool is retained only for compat (probably overlaps one of the cleanup nits tanmayrauth already flagged).

If we want to actually enforce the pre-bound precondition rather than lean on convention, a guard at the top returning an error would beat the `VisitUnbound` panic when someone passes an unbound expr — optional though. wdyt?

table/substrait/substrait.go
```diff
@@ -348,12 +344,7 @@ func toSubstraitLiteralSet(typ iceberg.Type, lits []iceberg.Literal) expr.ListLi
 }
 
 func (t *toSubstraitExpr) getRef(ref iceberg.BoundReference) expr.Reference {
-	updatedRef, err := iceberg.Reference(ref.Field().Name).Bind(t.schema, t.caseSensitive)
-	if err != nil {
-		panic(err)
-	}
-
-	path := updatedRef.Ref().PosPath()
+	path := ref.PosPath()
```
laskoviymishka, line 347: `path[0]` here leans on `PosPath()` always returning at least one element — true by construction today since `acc.pos` is unconditionally first, but it's an implicit invariant. A one-line comment noting that, or a cheap `len` guard, would save a future BoundReference impl from a bare slice-bounds panic. Non-blocking.

table/substrait/substrait_test.go
```diff
@@ -79,6 +79,60 @@ func TestRefTypes(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			bound, err := tt.predicate.Bind(sc, true)
+			require.NoError(t, err)
+
+			_, converted, err := substrait.ConvertExpr(sc, bound, true)
+			require.NoError(t, err)
+			assert.Contains(t, converted.String(), tt.expected)
```
laskoviymishka, line 117: `assert.Contains` means a wrong path could still slip through — e.g. `.field(0).field(1).field(0)` contains `.field(1).field(0)` as a substring and would pass while being wrong. Since pinning the exact path is the whole point of this regression, I'd assert the full converted string (or extract just the field-ref chain) rather than a substring. Same goes for the `.field(0).field(0)` check in `TestNestedReferenceWithoutTopLevelLeafName`.

laskoviymishka (approved):
Nice, this is the right fix. Pulling the path straight off the bound ref instead of rebinding by leaf name fixes both the wrong-column collision on nested leaf-name clashes and the panic when the leaf name doesn't resolve at top level, and the two new tests pin both. Agreed with tanmayrauth on that.

Left couple comments for future, but happy to merge! 

:ship: :+1:

PR #1589, opened 2026-07-28
fix(dv): validate metadata before writing Puffin files

fallintoplace:
## What changed

Construct and validate all deletion-vector `DataFile` metadata before writing the completed Puffin object. Extend the unknown-spec regression test to verify that no file is created.

## Why

`Flush` previously persisted the Puffin file before resolving partition specs and building manifest metadata. A metadata error could therefore return failure while leaving an unreferenced object in storage.

## Testing

- `go test ./table/dv`

### DIFF
--- a/table/dv/dv_writer.go
+++ b/table/dv/dv_writer.go
@@ -223,10 +223,6 @@ func (w *DVWriter) Flush(_ context.Context, location string) ([]iceberg.DataFile
 	}
 
 	fileBytes := buf.Bytes()
-	if err := w.fs.WriteFile(location, fileBytes); err != nil {
-		return nil, fmt.Errorf("write DV puffin file: %w", err)
-	}
-
 	fileSize := int64(len(fileBytes))
 
 	dataFiles := make([]iceberg.DataFile, 0, len(results))
@@ -258,6 +254,10 @@ func (w *DVWriter) Flush(_ context.Context, location string) ([]iceberg.DataFile
 			Build()
 
 		dataFiles = append(dataFiles, df)
+	}
+
+	if err := w.fs.WriteFile(location, fileBytes); err != nil {
+		return nil, fmt.Errorf("write DV puffin file: %w", err)
 	}
 
 	w.entries = make(map[string]*dvEntry)
--- a/table/dv/dv_writer_test.go
+++ b/table/dv/dv_writer_test.go
@@ -19,6 +19,7 @@ package dv
 
 import (
 	"context"
+	stdfs "io/fs"
 	"testing"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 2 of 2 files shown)


tanmayrauth (approved):
Reorder is correct, the metadata loop and the write are independent, so moving the write last cleanly prevents orphaned Puffin objects on metadata errors with no behavior change on the success path. LGTM.

table/dv/dv_writer.go
```diff
@@ -260,6 +256,10 @@ func (w *DVWriter) Flush(_ context.Context, location string) ([]iceberg.DataFile
 		dataFiles = append(dataFiles, df)
 	}
 
+	if err := w.fs.WriteFile(location, fileBytes); err != nil {
```
laskoviymishka, line 259: Not blocking and it's pre-existing, but since this PR is squarely about the error path: if this `WriteFile` fails we return without resetting `w.entries`/`w.order`, so a caller that retries `Flush` re-serializes and writes again — and if the bytes actually landed but the ACK was lost, the retry lands a second copy at a fresh location and orphans the first.

The reorder doesn't make this worse, but it feels like the right moment to either reset state after a successful write or document on `Flush` that it isn't safe to retry after a write failure. wdyt?

table/dv/dv_writer_test.go
```diff
@@ -452,13 +453,17 @@ func TestDVWriterFlushMixedSpecIDs(t *testing.T) {
 func TestDVWriterFlushUnknownSpecID(t *testing.T) {
 	fs := newTestFS()
 	w := NewDVWriter(fs, unpartitionedResolver())
+	location := "mem://test/unknown-spec.puffin"
 
 	// specID 99 is not registered with the resolver.
 	require.NoError(t, w.Add("s3://bucket/file.parquet", []int64{1}, 99, nil))
```
laskoviymishka, line 459: Nice regression test — this does fail on the pre-fix ordering. Since the PR description calls out "mid-loop failure across multiple entries leaves nothing behind," I'd add a second entry with a known spec ahead of the unknown one, so we also prove a partial metadata loop leaves no file behind. Not blocking.

table/dv/dv_writer_test.go
```diff
@@ -452,13 +453,17 @@ func TestDVWriterFlushMixedSpecIDs(t *testing.T) {
...
 
-	_, err := w.Flush(context.Background(), "mem://test/unknown-spec.puffin")
+	_, err := w.Flush(context.Background(), location)
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "unknown partition spec id 99")
+
+	_, openErr := fs.Open(location)
+	assert.ErrorIs(t, openErr, stdfs.ErrNotExist)
```
laskoviymishka, line 466: tiny one: this `ErrorIs` is the actual point of the test — proving no file was written — so I'd make it `require.ErrorIs` to match the `require.Error` above. Not a blocker.

laskoviymishka (approved):
The reorder is the right fix, putting `WriteFile` after the metadata loop means a spec-resolution failure can't leave an orphaned Puffin object behind, and the success path is byte-for-byte unchanged. No objection to landing this.

One thing worth deciding while we're in this error path, though it's pre-existing and not a blocker: after a `WriteFile` failure the writer keeps its `entries`/`order` populated, so a caller that retries `Flush` will re-serialize and write again — and if the bytes actually landed but the ACK was lost, the retry writes a second copy at a fresh location and orphans the first. The reorder doesn't make this worse (the reset was always after a successful write), but since the PR is about error-path correctness it feels like the moment to either reset state after the write or document that `Flush` isn't safe to retry after a write failure.

The rest are small — I left a note on tightening the new test into a multi-entry variant, plus a require-vs-assert nit. Separately, the DV blob's empty `fields` list diverges from Java's `[_pos]`, but that's pre-existing and out of scope here.

None of this blocks, happy to merge!

PR #1590, opened 2026-07-28
fix(substrait): convert nanosecond timestamp literals

fallintoplace:
## Summary

Convert Iceberg nanosecond timestamp literals to Substrait precision-timestamp literals.

## Why

Schema conversion supported timestamp_ns and timestamptz_ns, but expression conversion had no matching literal case and panicked for predicates containing them.

## What changed

Both timestamp variants now preserve their timezone distinction and raw nanosecond value. Equality, pre-epoch, IN, and NOT IN cases are covered.

## Tests

- go test ./table/substrait

### DIFF
--- a/table/substrait/substrait.go
+++ b/table/substrait/substrait.go
@@ -318,6 +318,16 @@ func toSubstraitLiteral(typ iceberg.Type, lit iceberg.Literal) expr.Literal {
 		}
 
 		return toPrimitiveSubstraitLiteral(types.Timestamp(lit))
+	case iceberg.TimestampNsLiteral:
+		if typ.Equals(iceberg.PrimitiveTypes.TimestampTzNs) {
+			return expr.NewPrecisionTimestampTzLiteral(
+				int64(lit), types.PrecisionNanoSeconds, types.NullabilityRequired,
+			)
+		}
+
+		return expr.NewPrecisionTimestampLiteral(
+			int64(lit), types.PrecisionNanoSeconds, types.NullabilityRequired,
+		)
 	case iceberg.DateLiteral:
 		return toPrimitiveSubstraitLiteral(types.Date(lit))
 	case iceberg.TimeLiteral:
--- a/table/substrait/substrait.go
+++ b/table/substrait/substrait.go
@@ -21,6 +21,7 @@ import (
 	_ "embed"
 	"encoding/binary"
 	"fmt"
+	"sort"
 	"strings"
 	"unsafe"
 
@@ -348,6 +349,10 @@ func toSubstraitLiteralSet(typ iceberg.Type, lits []iceberg.Literal) expr.ListLi
 	if len(lits) == 0 {
 		return nil
 	}
+
+	sort.Slice(lits, func(i, j int) bool {
+		return lits[i].String() < lits[j].String()
+	})
 
 	out := make([]expr.Literal, len(lits))
 	for i, l := range lits {
--- a/table/substrait/substrait_test.go
+++ b/table/substrait/substrait_test.go
@@ -142,6 +142,7 @@ func TestNanosecondTimestampLiterals(t *testing.T) {
 		name       string
 		fieldType  iceberg.Type
 		predicate  iceberg.BooleanExpression
+		wantExpr   string
 		wantValues []string
 		wantType   string
 	}{
@@ -149,6 +150,7 @@ func TestNanosecondTimestampLiterals(t *testing.T) {
 			name:       "timestamp equality",
 			fieldType:  iceberg.PrimitiveTypes.TimestampNs,
 			predicate:  iceberg.EqualTo(iceberg.Reference("ts"), iceberg.TimestampNano(123456789)),
+			wantExpr:   "equal(.field(0) => precision_timestamp?<9>, precision_timestamp<9>(1970-01-01 00:00:00.123456789)) => boolean?",
 			wantValues: []string{"1970-01-01 00:00:00.123456789"},
 			wantType:   "precision_timestamp<9>",
 		},
@@ -156,6 +158,7 @@ func TestNanosecondTimestampLiterals(t *testing.T) {
 			name:       "timestamp with timezone equality before epoch",
 			fieldType:  iceberg.PrimitiveTypes.TimestampTzNs,
 			predicate:  iceberg.EqualTo(iceberg.Reference("ts"), iceberg.TimestampNano(-123456789)),
+			wantExpr:   "equal(.field(0) => precision_timestamp_tz?<9>, precision_timestamp_tz<9>(1969-12-31T23:59:59.876543211Z)) => boolean?",
 			wantValues: []string{"1969-12-31T23:59:59.876543211Z"},
 			wantType:   "precision_timestamp_tz<9>",
 		},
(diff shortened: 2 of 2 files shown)


table/substrait/substrait.go
```diff
@@ -318,6 +319,21 @@ func toSubstraitLiteral(typ iceberg.Type, lit iceberg.Literal) expr.Literal {
...
+			result expr.Literal
+			err    error
+		)
+		if typ.Equals(iceberg.PrimitiveTypes.TimestampTzNs) {
+			result, err = literal.NewPrecisionTimestampTz(types.PrecisionNanoSeconds, int64(lit), false)
+		} else {
+			result, err = literal.NewPrecisionTimestamp(types.PrecisionNanoSeconds, int64(lit), false)
+		}
+		if err != nil {
+			panic(err)
```
tanmayrauth: panic(err) here turns a library error into a process-crashing panic, whereas the enclosing ConvertExpr/VisitExpr both return error — so on the one path that could fail, a caller of the public ConvertExpr gets a panic instead of an err. It's effectively unreachable since PrecisionNanoSeconds is a valid constant and NewPrecisionTimestamp won't error on it, and it matches the existing panic style in this file (default case, getRef), so this is just a consistency note rather than a fix I'd block on — surfacing it in case you'd rather thread the error out to match the function's contract.

tanmayrauth (commented):
Looks correct — the ns literal representation lines up with the PrecisionTimestamp type the schema side already emits, and the pre-epoch case renders right. One tiny consistency note inline.

table/substrait/substrait_test.go
```diff
@@ -137,6 +137,60 @@ func TestExprs(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sc := iceberg.NewSchema(1, iceberg.NestedField{ID: 1, Name: "ts", Type: tt.fieldType})
+			bound, err := iceberg.BindExpr(sc, tt.predicate, true)
+			require.NoError(t, err)
+
+			_, converted, err := substrait.ConvertExpr(sc, bound, true)
+			require.NoError(t, err)
+			for _, value := range tt.wantValues {
+				assert.Contains(t, converted.String(), value)
```
laskoviymishka: `assert.Contains` here is doing less than it looks — a wrong literal whose rendering happened to contain this substring would still pass, and the assertion goes quietly vacuous if substrait-go ever changes how it formats `PrecisionTimestampLiteral`. `TestExprs` right above uses `assert.Equal` on the full expression string; I'd match that here so the regression test actually pins the value. wdyt?

table/substrait/substrait.go
```diff
@@ -318,6 +318,16 @@ func toSubstraitLiteral(typ iceberg.Type, lit iceberg.Literal) expr.Literal {
 		}
 
 		return toPrimitiveSubstraitLiteral(types.Timestamp(lit))
+	case iceberg.TimestampNsLiteral:
+		if typ.Equals(iceberg.PrimitiveTypes.TimestampTzNs) {
```
laskoviymishka, line 323: This `err` can't actually be non-nil — `literal.NewPrecisionTimestamp[Tz]` only fails on an unknown precision, and we hand it the `PrecisionNanoSeconds` constant — so the panic is guarding a dead path. `expr.NewPrecisionTimestampLiteral` / `NewPrecisionTimestampTzLiteral` are already in scope and don't return an error, which lets this arm drop the var block and mirror the two-line `TimestampLiteral` case above:

```go
case iceberg.TimestampNsLiteral:
	if typ.Equals(iceberg.PrimitiveTypes.TimestampTzNs) {
		return expr.NewPrecisionTimestampTzLiteral(int64(lit), types.PrecisionNanoSeconds, types.NullabilityRequired)
	}

	return expr.NewPrecisionTimestampLiteral(int64(lit), types.PrecisionNanoSeconds, types.NullabilityRequired)
```

table/substrait/substrait_test.go
```diff
@@ -137,6 +137,60 @@ func TestExprs(t *testing.T) {
 	}
 }
 
+func TestNanosecondTimestampLiterals(t *testing.T) {
```
laskoviymishka, line 140: The original panic hit every predicate carrying a ns literal, including range comparisons — but the cases here only cover equality and in/not-in. A single `GreaterThan`/`LessThan` case per ns type would lock the comparison path that was also broken. Not blocking, just cheap insurance while we're here.

laskoviymishka (approved):
Nice catch! This closes the `toSubstraitLiteral` panic for ns timestamp literals, and the encoding looks right to me: raw ns-since-epoch through at precision 9, tz vs non-tz dispatched on the bound column type. Happy to see it.

Two things I'd tidy before merge, neither a blocker. The one I'd genuinely like fixed is the test: `assert.Contains(converted.String(), ...)` will still pass on a wrong value as long as the substring shows up, and it goes quietly vacuous if substrait-go ever changes its string formatting. `TestExprs` just above uses `assert.Equal` on the full expression string — I'd match that here so the test actually locks the value.

The other is smaller: the `panic(err)` guards an error that can't fire (`NewPrecisionTimestamp[Tz]` only errors on an unknown precision, and we pass a constant), so the var-block-plus-panic can collapse to the two-line shape the µs case uses. Left a suggestion inline.

Once the test's tightened I'm happy to merge.

table/substrait/substrait_test.go
```diff
@@ -137,6 +137,80 @@ func TestExprs(t *testing.T) {
...
+			sc := iceberg.NewSchema(1, iceberg.NestedField{ID: 1, Name: "ts", Type: tt.fieldType})
+			bound, err := iceberg.BindExpr(sc, tt.predicate, true)
+			require.NoError(t, err)
+
+			_, converted, err := substrait.ConvertExpr(sc, bound, true)
+			require.NoError(t, err)
+			if tt.wantExpr != "" {
+				assert.Equal(t, tt.wantExpr, converted.String())
+			}
+			for _, value := range tt.wantValues {
```
zeroshade: Non-blocking: for the IN and NOT IN cases, `Contains` can pass when one literal is wrong and becomes vacuous if rendering changes. Consider giving those cases complete `wantExpr` values and asserting `Equal`, as the tests directly above do; these are the two cases that specifically exercise `toSubstraitLiteralSet`.

table/substrait/substrait_test.go
```diff
@@ -137,6 +137,80 @@ func TestExprs(t *testing.T) {
 	}
 }
 
+func TestNanosecondTimestampLiterals(t *testing.T) {
```
zeroshade, line 140: Worth a follow-up: boundary cases for `MinInt64`/`MaxInt64`, a cross-precision `TimestampNsLiteral.To(...)` case that pins ns-to-us truncation direction, and one execution round-trip would complement the current rendered-string coverage.

zeroshade (commented):
This change is correct: I verified the precision path end to end. The `int64` nanoseconds pass through unmodified with `PrecisionNanoSeconds`; substrait-go v8.1.1 stores the value verbatim with no unit conversion, the timezone dispatch matches the existing microsecond path, and Arrow materializes both precision timestamp forms through `scalar.NewTimestampScalar`.

The main non-blocking test nit is the IN/NOT IN coverage: those are precisely the cases exercising `toSubstraitLiteralSet`, but they do not set `wantExpr` and therefore never use the exact equality assertion used by `TestExprs` directly above. Consider tightening those two cases as noted inline; the implementation itself looks sound.

PR #1591, opened 2026-07-28
fix(table): validate last-updated-ms presence

fallintoplace:
## What changed

Require `last-updated-ms` to be present and non-null while decoding metadata versions 1, 2, and 3. Numeric values remain unrestricted, so Unix epoch zero and existing negative fixture values are preserved.

Add coverage for missing, null, epoch-zero, and negative values across all metadata format versions.

## Why

The decoder previously used a numeric sentinel for presence validation. That made a valid number indistinguishable from an absent field and either accepted missing metadata or rejected explicitly encoded values.

## Testing

- `go test ./table ./table/internal -count=1 -timeout=180s`
- `go test ./... -count=1 -timeout=300s`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -18,6 +18,7 @@
 package table
 
 import (
+	"bytes"
 	"cmp"
 	"encoding/binary"
 	"encoding/json"
@@ -1519,6 +1520,9 @@ func assignMissingPartitionFieldIDs(b []byte) ([]byte, error) {
 	if err := json.Unmarshal(b, &metadata); err != nil {
 		return nil, err
 	}
+	if err := requireLastUpdatedMS(metadata); err != nil {
+		return nil, err
+	}
 
 	type rawPartitionSpec struct {
 		ID     int                          `json:"spec-id"`
@@ -1638,7 +1642,6 @@ type commonMetadata struct {
 
 func initCommonMetadataForDeserialization() commonMetadata {
 	return commonMetadata{
-		LastUpdatedMS:      -1,
 		LastColumnId:       -1,
 		CurrentSchemaID:    -1,
 		DefaultSpecID:      -1,
@@ -1646,6 +1649,15 @@ func initCommonMetadataForDeserialization() commonMetadata {
 		SortOrderList:      nil,
 		Specs:              nil,
 	}
+}
+
+func requireLastUpdatedMS(fields map[string]json.RawMessage) error {
+	value, ok := fields["last-updated-ms"]
+	if !ok || bytes.Equal(bytes.TrimSpace(value), []byte("null")) {
+		return fmt.Errorf("%w: missing last-updated-ms", ErrInvalidMetadata)
+	}
+
+	return nil
 }
 
 func (c *commonMetadata) Ref() SnapshotRef {
@@ -2222,9 +2234,6 @@ func (c *commonMetadata) constructRefs() {
 
 func (c *commonMetadata) validate() error {
 	switch {
-	case c.LastUpdatedMS == 0:
-		// last-updated-ms is required
-		return fmt.Errorf("%w: missing last-updated-ms", ErrInvalidMetadata)
 	case c.LastColumnId < 0:
 		// last-column-id is required
 		return fmt.Errorf("%w: missing last-column-id", ErrInvalidMetadata)
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -18,7 +18,6 @@
 package table
 
 import (
-	"bytes"
 	"cmp"
 	"encoding/binary"
 	"encoding/json"
@@ -1492,7 +1491,7 @@ func ParseMetadataBytes(b []byte) (Metadata, error) {
 		FormatVersion int `json:"format-version"`
 	}{}
 	if err := json.Unmarshal(b, &ver); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w: %w", ErrInvalidMetadata, err)
 	}
 
 	var ret Metadata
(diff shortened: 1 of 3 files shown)


table/metadata.go
```diff
@@ -1648,6 +1648,20 @@ func initCommonMetadataForDeserialization() commonMetadata {
 	}
 }
 
+func requireLastUpdatedMS(data []byte) error {
```
tanmayrauth: This re-parses the whole metadata document into map[string]json.RawMessage, but ParseMetadataBytes already builds exactly that map in assignMissingPartitionFieldIDs (metadata.go:1519) two steps before the struct unmarshal — so on the main load path the document now gets fully parsed into the same map type three times (once here per version). Beyond the redundant parse, this scatters the presence invariant across all three UnmarshalJSON methods. Consider asserting presence once in ParseMetadataBytes, reusing the map that assignMissingPartitionFieldIDs already has, so there's a single check for all format versions and no extra parse.

table/metadata.go
```diff
@@ -2422,6 +2433,10 @@ func (m *metadataV1) preValidate() {
 }
 
 func (m *metadataV1) UnmarshalJSON(b []byte) error {
+	if err := requireLastUpdatedMS(b); err != nil {
```
tanmayrauth: Related to the above: this PR also removes the last-updated-ms case from commonMetadata.validate(), which is the shared invariant checker for both the JSON and builder paths. After this change validate() no longer asserts the field is set at all — the builder path is safe only because buildCommonMetadata forces
  time.Now() when lastUpdatedMS is 0 (metadata.go:1010). So the "must be present" invariant now lives incidentally in two spots (this unmarshal hook + the builder default) rather than in validate(). If the check moves into ParseMetadataBytes per the note above, that's fine; the thing to avoid is leaving validate()
  with no last-updated-ms assertion while the presence guarantee depends on two unrelated code paths.

tanmayrauth (commented):
One design note on where the presence check lives. Details inline.

table/metadata.go
```diff
@@ -1648,6 +1651,15 @@ func initCommonMetadataForDeserialization() commonMetadata {
 	}
 }
 
+func requireLastUpdatedMS(fields map[string]json.RawMessage) error {
```
laskoviymishka, line 1657: this bare `return err` means a malformed-JSON blob comes back unwrapped, so `errors.Is(err, ErrInvalidMetadata)` is false for broken JSON but true for the null/missing case right below. Since this is a fresh call site I'd wrap it with `fmt.Errorf("%w: %w", ErrInvalidMetadata, err)` so callers get a consistent sentinel.

table/metadata.go
```diff
@@ -1648,6 +1651,15 @@ func initCommonMetadataForDeserialization() commonMetadata {
...
+	value, ok := fields["last-updated-ms"]
+	if !ok || bytes.Equal(bytes.TrimSpace(value), []byte("null")) {
+		return fmt.Errorf("%w: missing last-updated-ms", ErrInvalidMetadata)
+	}
```
laskoviymishka, line 1661: the `bytes.Equal(bytes.TrimSpace(value), []byte("null"))` is doing more work than it needs to. A `json.RawMessage` coming out of a `map[string]json.RawMessage` is already whitespace-trimmed by the decoder, so `string(value) == "null"` is equivalent and lets us drop the new `bytes` import entirely.

```suggestion
	if !ok || string(value) == "null" {
```

table/metadata.go
```diff
@@ -1648,6 +1651,15 @@ func initCommonMetadataForDeserialization() commonMetadata {
...
+
```
laskoviymishka, line 1662: small thing: this says "missing" but it also fires for an explicit `"last-updated-ms": null`, and the test asserts that exact substring for the null case, so the wording gets locked in. Something like `"last-updated-ms is absent or null"` (or just `"required"`) would be accurate for both paths.

table/metadata_internal_test.go
```diff
@@ -310,6 +310,69 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+		data string
+	}{
+		{name: "v1", data: ExampleTableMetadataV1},
+		{name: "v2", data: ExampleTableMetadataV2},
+		{name: "v3", data: ExampleTableMetadataV3},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var metadata map[string]any
```
laskoviymishka, line 325: this map is declared once per version and then mutated in place by each inner subtest (missing/null/epoch/negative), so the cases only pass because they run in order. A `t.Parallel()` or `-shuffle` would have them stepping on each other. I'd `maps.Clone(metadata)` into a local at the top of each subtest so they're independent.

table/metadata_internal_test.go
```diff
@@ -310,6 +310,69 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+				assert.Zero(t, parsed.LastUpdatedMillis())
+			})
+
+			t.Run("negative", func(t *testing.T) {
+				metadata["last-updated-ms"] = float64(-1)
+				raw, err := json.Marshal(metadata)
+				require.NoError(t, err)
+				parsed, err := ParseMetadataBytes(raw)
+				require.NoError(t, err)
+				assert.Equal(t, int64(-1), parsed.LastUpdatedMillis())
```
laskoviymishka, line 364: the matrix covers missing/null/epoch/negative but not a present-but-wrong-type value like `"last-updated-ms": "2024-01-01"`. That case slips past `requireLastUpdatedMS` (present and non-null) and instead fails later with a `json.UnmarshalTypeError`, which isn't wrapped as `ErrInvalidMetadata`, so `errors.Is` holds for null but not for a wrong type. Worth a subtest that pins down which error we expect there. wdyt?

laskoviymishka (changes requested):
Nice catch on the underlying bug: the old sentinel/validate mismatch rejected a valid `"last-updated-ms": 0` while letting an absent field slip through, and requiring the field is right per spec (V1/V2/V3 all mark it required, same as Java and Rust). The v1/v2/v3 test matrix is a good way to pin it down.

Almost there, but I'd want a bit of work before it lands, mostly around where the check goes. Agreed with tanmayrauth that the presence assertion belongs once in `ParseMetadataBytes`, reusing the map `assignMissingPartitionFieldIDs` already builds, instead of a third full parse copied across the three `UnmarshalJSON` methods. Doing it there also lets `validate()` keep a real last-updated-ms guard rather than dropping the case, so the invariant stops leaning on the builder's `time.Now()` default everywhere off the JSON path.

One thing worth a line in the PR description: this is a behavioral change. Metadata we used to ingest without last-updated-ms will now fail, which is correct and matches Java/Rust but stricter than PyIceberg (it auto-fills). The rest are smaller and inline.

laskoviymishka (approved):
👍 x

table/metadata.go
haithemalsabr: **الإسمالمستلم هيثم سعيديحيىأحمدالصبري**رقم الجوال 0096 777711 7998

PR #1593, opened 2026-07-28
fix(partitions): validate decoded partition specs

fallintoplace:
## What changed

Apply structural partition-field and partition-spec validation during JSON decoding. Reject invalid spec IDs, missing or non-positive source IDs, empty and duplicate names, duplicate field IDs, redundant definitions, and invalid transforms.

The validation remains independent of the current schema so historical specs may still reference dropped source columns.

## Why

The JSON path directly populated internal fields and bypassed invariants enforced by the constructor. Malformed standalone specs and historical table specs could therefore enter the model and be serialized again.

## Testing

- `go test . -run 'Test(PartitionFieldUnmarshalJSON|PartitionSpecUnmarshalRejectsInvalidStructure|DeserializePartitionSpec)' -count=1`\n- `go test ./table -run 'TestRejectStructurallyInvalidHistoricalPartitionSpec$' -count=1`

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -497,6 +497,10 @@ func validatePartitionFields(fields []PartitionField) error {
 		}
 		names[field.Name] = struct{}{}
 
+		if _, ok := field.Transform.(VoidTransform); ok {
+			continue
+		}
+
 		definition := fmt.Sprintf("%v:%s", field.SourceIDs, field.Transform)
 		if _, ok := definitions[definition]; ok {
 			return fmt.Errorf("%w: redundant partition field for source IDs %v and transform %s",
--- a/partitions.go
+++ b/partitions.go
@@ -121,6 +121,8 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 		if _, ok := raw["source-ids"]; ok {
 			return errors.New("partition field cannot contain both source-id and source-ids")
 		}
+	} else if _, ok := raw["source-ids"]; !ok {
+		return fmt.Errorf("%w: partition field requires source-id or source-ids", ErrInvalidPartitionSpec)
 	}
 
 	aux := struct {
@@ -143,10 +145,21 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 	} else {
 		p.SourceIDs = []int{aux.SourceID}
 	}
+	for _, sourceID := range p.SourceIDs {
+		if sourceID <= 0 {
+			return fmt.Errorf("%w: partition source ID must be positive: %d", ErrInvalidPartitionSpec, sourceID)
+		}
+	}
+	if p.Name == "" {
+		return fmt.Errorf("%w: partition name cannot be empty", ErrInvalidPartitionSpec)
+	}
 
 	var err error
 	if p.Transform, err = ParseTransform(aux.TransformString); err != nil {
-		return err
+		return fmt.Errorf("%w: %w", ErrInvalidPartitionSpec, err)
+	}
+	if err := validateTransform(p.Transform); err != nil {
+		return fmt.Errorf("%w: %w", ErrInvalidPartitionSpec, err)
 	}
 
 	return nil
@@ -439,6 +452,9 @@ func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, &aux); err != nil {
 		return err
 	}
+	if aux.ID < 0 {
+		return fmt.Errorf("%w: spec ID must be non-negative: %d", ErrInvalidPartitionSpec, aux.ID)
+	}
 
 	fields := make([]PartitionField, len(aux.Fields))
 	for i, rawField := range aux.Fields {
@@ -459,12 +475,35 @@ func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
 			return err
 		}
 	}
+	if err := validatePartitionFields(fields); err != nil {
+		return err
+	}
 
 	ps.id, ps.fields = aux.ID, fields
 	if err := ps.assignPartitionFieldIds(nil); err != nil {
 		return fmt.Errorf("%w: %w", ErrInvalidPartitionSpec, err)
 	}
 	ps.initialize()
+
+	return nil
+}
+
+func validatePartitionFields(fields []PartitionField) error {
+	names := make(map[string]struct{}, len(fields))
+	definitions := make(map[string]struct{}, len(fields))
+	for _, field := range fields {
+		if _, ok := names[field.Name]; ok {
+			return fmt.Errorf("%w: duplicate partition name: %s", ErrInvalidPartitionSpec, field.Name)
+		}
+		names[field.Name] = struct{}{}
+
+		definition := fmt.Sprintf("%v:%s", field.SourceIDs, field.Transform)
+		if _, ok := definitions[definition]; ok {
+			return fmt.Errorf("%w: redundant partition field for source IDs %v and transform %s",
+				ErrInvalidPartitionSpec, field.SourceIDs, field.Transform)
+		}
+		definitions[definition] = struct{}{}
+	}
 
 	return nil
 }
--- a/partitions.go
+++ b/partitions.go
@@ -510,10 +510,6 @@ func validatePartitionFields(fields []PartitionField) error {
 			continue
 		}
 
-		if _, ok := field.Transform.(VoidTransform); ok {
-			continue
-		}
-
 		definition := fmt.Sprintf("%v:%s", field.SourceIDs, field.Transform)
 		if _, ok := definitions[definition]; ok {
 			return fmt.Errorf("%w: redundant partition field for source IDs %v and transform %s",
--- a/partitions.go
+++ b/partitions.go
@@ -119,11 +119,11 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 
 	if _, ok := raw["source-id"]; ok {
 		if _, ok := raw["source-ids"]; ok {
-			return errors.New("partition field cannot contain both source-id and source-ids")
+			return fmt.Errorf("%w: partition field cannot contain both source-id and source-ids", ErrInvalidPartitionSpec)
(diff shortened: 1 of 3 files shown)


partitions.go
```diff
@@ -469,6 +488,26 @@ func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
...
+func validatePartitionFields(fields []PartitionField) error {
+	names := make(map[string]struct{}, len(fields))
+	definitions := make(map[string]struct{}, len(fields))
+	for _, field := range fields {
+		if _, ok := names[field.Name]; ok {
+			return fmt.Errorf("%w: duplicate partition name: %s", ErrInvalidPartitionSpec, field.Name)
+		}
+		names[field.Name] = struct{}{}
+
+		definition := fmt.Sprintf("%v:%s", field.SourceIDs, field.Transform)
```
tanmayrauth, line 513: This rejects two fields that share (source-ids, transform), but `void` is the drop marker: dropping a partition field in a v1 spec rewrites it to void while keeping its source-id, so a column that had two
  partition fields both later dropped legally ends up with two `[source]:void` entries (distinct names/field-ids). This check collapses them and fails the whole spec on decode — so a table Java/PyIceberg
  wrote, and older iceberg-go read fine, now fails to open with ErrInvalidPartitionSpec. It's also self-inconsistent: NewPartitionSpecID builds such a spec with no complaint, MarshalJSON emits it, and this
  path then refuses to read it back.
  
  Exempt void from the redundancy check (keep the name-uniqueness check): after `names[field.Name] = struct{}{}`, add `if _, ok := field.Transform.(VoidTransform); ok { continue }` before building
  `definition`. ParseTransform returns value-typed transforms, so the non-pointer VoidTransform assertion is right. That still rejects two identity/bucket/etc. on the same source, which is the case you're
  actually targeting.

partitions_test.go
```diff
@@ -647,10 +647,65 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
...
+			data:    `{"spec-id":0,"fields":[{"source-id":1,"field-id":1000,"name":"part","transform":"identity"},{"source-id":2,"field-id":1001,"name":"part","transform":"identity"}]}`,
+			message: "duplicate partition name",
+		},
+		{
+			name:    "duplicate field IDs",
+			data:    `{"spec-id":0,"fields":[{"source-id":1,"field-id":1000,"name":"first","transform":"identity"},{"source-id":2,"field-id":1000,"name":"second","transform":"identity"}]}`,
+			message: "duplicate field ID provided",
+		},
+		{
+			name:    "redundant fields",
```
tanmayrauth, line 700: This covers two identity fields on one source being rejected, but there's no positive case for two `void` fields on the same source-id — which is exactly the legal historical spec the redundancy check above
  wrongly rejects. Add a case that unmarshals `{spec-id, fields:[{source-id:1,field-id:1000,name:"a",transform:"void"},{source-id:1,field-id:1001,name:"b",transform:"void"}]}` and asserts NoError, so the void exemption is locked in.

tanmayrauth (commented):
The structural checks are good, but the redundancy rule mis-handles void and rejects legitimate historical specs — that one's a blocker. Details inline.

partitions.go
```diff
@@ -121,6 +121,8 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 		if _, ok := raw["source-ids"]; ok {
 			return errors.New("partition field cannot contain both source-id and source-ids")
```
laskoviymishka: This one returns a bare `errors.New`, but the new branch just below wraps `ErrInvalidPartitionSpec`. So `errors.Is(err, ErrInvalidPartitionSpec)` is true for one source-id problem and false for the other, which callers will trip over. I'd wrap this one too while we're here.

partitions.go
```diff
@@ -121,6 +121,8 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
...
 		}
+	} else if _, ok := raw["source-ids"]; !ok {
+		return fmt.Errorf("%w: partition field requires source-id or source-ids", ErrInvalidPartitionSpec)
```
laskoviymishka: This branch rejects any field carrying neither source-id nor source-ids, and the changed test now expects `void` with no source-id to fail. void is the tombstone we write for dropped partition fields, and we used to accept it without a source column — the old test asserted `field.SourceID() == 0` on exactly this input. So this turns a documented contract into a hard parse failure for any metadata that already has a source-id-less void field.

@tanmayrauth already called this out and I agree it's the crux. The spec does cut the other way — Java's `buildFromJsonFields` and PyIceberg both require source-id — so I don't think "accept everything" is obviously right either. I'd make it an explicit decision: either a void carve-out that synthesizes a sentinel source-id when neither key is present, or keep the hard reject but land it with a changelog/migration note. Either is fine, but it shouldn't be a side effect of this branch. wdyt?

partitions.go
```diff
@@ -143,10 +145,21 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 	} else {
 		p.SourceIDs = []int{aux.SourceID}
 	}
+	for _, sourceID := range p.SourceIDs {
```
laskoviymishka, line 165: `{"source-ids":[]}` slips past the else-if above (the key is present), then falls through to `p.SourceIDs = []int{aux.SourceID}` = `[]int{0}`, so this loop reports "must be positive: 0" — which points at the wrong thing; the real problem is the empty array. PyIceberg rejects this explicitly with "Empty source-ids is not allowed."

The "empty source IDs" test asserts on that "must be positive" message, so it's green only because of this side effect — if we ever allowed source-id 0 it'd silently regress. I'd add an explicit empty-array check with its own message and update the test to match.

partitions.go
```diff
@@ -469,6 +488,30 @@ func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
...
+		if _, ok := names[field.Name]; ok {
+			return fmt.Errorf("%w: duplicate partition name: %s", ErrInvalidPartitionSpec, field.Name)
+		}
+		names[field.Name] = struct{}{}
+
+		if _, ok := field.Transform.(VoidTransform); ok {
+			continue
+		}
+
+		definition := fmt.Sprintf("%v:%s", field.SourceIDs, field.Transform)
```
laskoviymishka, line 513: This keys redundancy off `Transform.String()`, which assumes `String()` is injective per transform. The constructor path uses `Transform.Equals()`, so we now have two independent notions of "same field" that can drift — a shared helper both paths call would keep the semantics in one place.

It also carries forward a pre-existing gap rather than adding one: Java's `dedupName()` collapses year/month/day/hour to `"time"` and rejects e.g. `year(ts)` + `month(ts)` on the same column, whereas both our paths accept it. Not this PR's job to fix, just flagging. (Minor: the `fmt.Sprintf` might trip perfsprint depending on the version pinned in CI — worth a quick check.)

partitions_test.go
```diff
@@ -647,10 +647,73 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
 		}`
 		var field iceberg.PartitionField
 		err := json.Unmarshal([]byte(jsonData), &field)
-		require.NoError(t, err)
-		assert.Zero(t, field.SourceID())
-		assert.Equal(t, 1003, field.FieldID)
-		assert.Equal(t, "void", field.Name)
-		assert.Equal(t, iceberg.VoidTransform{}, field.Transform)
+		require.ErrorIs(t, err, iceberg.ErrInvalidPartitionSpec)
+		assert.ErrorContains(t, err, "requires source-id or source-ids")
```
laskoviymishka: The subtest name still says "unmarshal with no source id" (describing the input) but the outcome flipped from pass to reject — I'd rename it to something like "...is rejected" so the intent is obvious.

Also, whichever way the void decision lands above, there's no positive case that a void field *with* a source-id still unmarshals cleanly — worth adding one asserting `SourceID() == 1`.

table/metadata_internal_test.go
```diff
@@ -564,6 +564,25 @@ func TestRejectInvalidSchemaEntries(t *testing.T) {
 	})
 }
 
+func TestRejectStructurallyInvalidHistoricalPartitionSpec(t *testing.T) {
```
laskoviymishka, line 605: This is named for historical specs but only exercises the negative spec-id rejection. The PR's rationale is that historical specs may still reference dropped source columns — that's the case I'd most want pinned down: load a two-spec history where the earlier spec has an identity transform on a column that's gone from the later schema, and assert it still parses. Right now nothing locks that in.

laskoviymishka (changes requested):
The structural checks are a real improvement — validating spec-id, source-id positivity, empty names, and duplicates at the JSON boundary closes a gap the constructors couldn't, and the table-driven negative tests are easy to follow.

I'd hold this before merging though, mostly on the void handling.

Rejecting a void field with no source-id is defensible per spec — Java and PyIceberg both require source-id — but void is the tombstone we write for dropped partition fields, and iceberg-go used to accept it without a source column (the old test asserted exactly that). So this flips a documented contract and makes existing metadata with a source-id-less void field unparseable. @tanmayrauth already flagged this and I agree it's the crux; where I'd land is that this should be a deliberate call — a spec-correct hard break with a changelog note, or a void carve-out that synthesizes a sentinel source-id — rather than a side effect of the new required-source-id branch.

The other one worth calling out before it ships: `{"source-ids":[]}` currently reports "source ID must be positive: 0", which points at the wrong problem, and the test locks that message in — so it's giving us false confidence today.

- Decide the void contract explicitly — carve-out or documented hard break, not an accident of the new branch.
- Give empty `source-ids:[]` its own error and stop asserting on the "positive: 0" side effect.
- Wrap the pre-existing "both source-id and source-ids" error so `errors.Is(ErrInvalidPartitionSpec)` is consistent across the branch.
- Add a positive test for a historical spec that references a dropped source column — the case the PR is really about.

Rest is inline and minor. Once the void decision is settled and those land, happy to take another pass.

laskoviymishka (approved):
:shipit:

PR #1594, opened 2026-07-28
fix(puffin): poison writers after write failures

fallintoplace:
## What changed

Record the first physical write failure on a Puffin writer and reject every subsequent mutation, blob write, or finish attempt while preserving the original cause. Route blob and footer writes through the shared failure-tracking path.

Add fault-injection coverage for partial failures during blob, footer magic, footer payload, and trailer writes, plus short writes that return no underlying error. The tests verify that no additional writes occur afterward.

## Why

An `io.Writer` may consume bytes and return an error or report a short write. Reusing the Puffin writer afterward left its logical offset behind the physical stream and could emit corrupt metadata.

## Testing

- `go test ./puffin -count=1`

### DIFF
--- a/puffin/puffin_test.go
+++ b/puffin/puffin_test.go
@@ -74,10 +74,12 @@ func defaultBlobInput() puffin.BlobMetadataInput {
 
 func TestWriterRejectsOperationsAfterPartialWriteFailure(t *testing.T) {
 	tests := []struct {
-		name     string
-		failCall int
+		name           string
+		failCall       int
+		failsDuringAdd bool
 	}{
-		{name: "blob", failCall: 2},
+		// NewWriter writes first, AddBlob writes second, and Finish writes calls 3-5.
+		{name: "blob", failCall: 2, failsDuringAdd: true},
 		{name: "footer magic", failCall: 3},
 		{name: "footer payload", failCall: 4},
 		{name: "footer trailer", failCall: 5},
@@ -91,7 +93,7 @@ func TestWriterRejectsOperationsAfterPartialWriteFailure(t *testing.T) {
 			require.NoError(t, err)
 
 			_, err = writer.AddBlob(defaultBlobInput(), []byte("first blob payload"))
-			if tt.failCall == 2 {
+			if tt.failsDuringAdd {
 				require.ErrorIs(t, err, writeErr)
 			} else {
 				require.NoError(t, err)
@@ -125,6 +127,9 @@ func TestWriterRejectsOperationsAfterShortWrite(t *testing.T) {
 	_, err = writer.AddBlob(defaultBlobInput(), []byte("second blob payload"))
 	assert.ErrorContains(t, err, "short write")
 	assert.ErrorContains(t, writer.Finish(), "short write")
+	assert.ErrorContains(t, writer.AddProperties(map[string]string{"key": "value"}), "short write")
+	assert.ErrorContains(t, writer.ClearProperties(), "short write")
+	assert.ErrorContains(t, writer.SetCreatedBy("test"), "short write")
 	assert.Equal(t, callsAfterFailure, output.calls)
 }
 
--- a/puffin/puffin_test.go
+++ b/puffin/puffin_test.go
@@ -19,6 +19,7 @@ package puffin_test
 
 import (
 	"bytes"
+	"errors"
 	"math"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM!

puffin/puffin_test.go
```diff
@@ -52,6 +72,62 @@ func defaultBlobInput() puffin.BlobMetadataInput {
 	}
 }
 
+func TestWriterRejectsOperationsAfterPartialWriteFailure(t *testing.T) {
+	tests := []struct {
+		name     string
+		failCall int
+	}{
+		{name: "blob", failCall: 2},
```
laskoviymishka: These `failCall` numbers encode the global ordinal of each write across NewWriter, AddBlob, and Finish — 2 is the blob write, 3–5 the footer writes. Nothing ties them to the site they're meant to hit, though: add a checksum or a codec byte anywhere in AddBlob and every index below it shifts by one, the `if tt.failCall == 2` branch at line 94 silently misclassifies, and the suite stays green while intercepting the wrong call.

I'd swap the raw number for something that names the intent — a `failsDuringAdd bool` on the table (true only for the blob case) driving that branch, plus a short comment enumerating which call each index is. I mostly want the drift to be loud instead of silent. wdyt?

puffin/puffin_writer.go
```diff
@@ -53,6 +53,7 @@ type Writer struct {
 	blobs     []BlobMetadata
 	props     map[string]string
 	done      bool
+	failed    error
```
laskoviymishka, line 59: This `failed` field is a real behavioral contract — once any physical write errors, every subsequent call is rejected — and it isn't written down anywhere. I think poisoning on failure is the right instinct (the stream's already half-written, the file is toast), but it's worth a line on the Writer type or the AddBlob/Finish godoc so callers know a failed writer is dead and not retryable.

One thing to be deliberate about: AddProperties, ClearProperties, and SetCreatedBy get poisoned too, and none of them touch the stream. Blocking them is defensible — the writer's tainted either way — but it'll surprise someone who expects only the byte-writing methods to fail, and it diverges from Java's PuffinWriter, which has no failed state at all. I'd keep the poison broad and just document that it's intentional.

puffin/puffin_test.go
```diff
@@ -52,6 +72,62 @@ func defaultBlobInput() puffin.BlobMetadataInput {
...
+	writer, err := puffin.NewWriter(output)
+	require.NoError(t, err)
+
+	_, err = writer.AddBlob(defaultBlobInput(), []byte("first blob payload"))
+	require.ErrorContains(t, err, "short write")
+
+	callsAfterFailure := output.calls
+	_, err = writer.AddBlob(defaultBlobInput(), []byte("second blob payload"))
+	assert.ErrorContains(t, err, "short write")
+	assert.ErrorContains(t, writer.Finish(), "short write")
```
laskoviymishka, line 129: This test only checks AddBlob and Finish get rejected, but the partial-write test above asserts all five methods including the three property setters. Same code path, so no real risk — just mirror the AddProperties/ClearProperties/SetCreatedBy asserts here so the two stay symmetric.

puffin/puffin_writer.go
```diff
@@ -89,6 +90,9 @@ func NewWriter(w io.Writer) (*Writer, error) {
 // SetProperties merges the provided properties into the file-level properties
 // written to the footer. Can be called multiple times before Finish.
 func (w *Writer) AddProperties(props map[string]string) error {
+	if w.failed != nil {
```
laskoviymishka: Tiny thing that recurs in all five methods: `failed` is checked before `done`. They can't both be set, so there's no bug — but it reads as though `failed` outranks `finalized`, which is backwards. I'd check `done` first. Non-blocking.

laskoviymishka (changes requested):
Nice fix — poisoning the writer after a write failure is the right instinct, and I like that the tests fault every write site and then assert nothing more gets written.

Two things I'd want settled before merge. The test is more fragile than it looks: the `failCall` values are global write ordinals with nothing tying them to the site they're meant to hit, so any new write in AddBlob shifts them and the suite stays green while testing the wrong failure. And the new `failed` field is a behavioral contract — a failed writer is permanently dead, including the no-I/O property setters — that isn't documented and diverges from Java's PuffinWriter. Poisoning everything is the right call; it just needs to be written down.

The rest is inline and minor.

laskoviymishka (approved):

PR #1595, opened 2026-07-28
fix(table): cancel record writes on iterator stop

fallintoplace:
## What changed

Cancel unpartitioned and fanout write contexts when a `WriteRecords` consumer stops early. Fanout writes now cancel record production independently from rolling writers and release retained batches left in the input queue after workers exit.

Add unpartitioned and partitioned regressions showing that breaking after the first output stops the source before all records are consumed. The checked allocator also verifies retained Arrow memory is released.

## Why

The iterator previously drained output without canceling production. Breaking iteration could therefore consume the entire source, continue writing files, or never return for an unbounded source.

## Testing

- `go test ./table -run 'Test(WriteRecords|FanoutWriter|PositionDeletePartitionedFanoutWriter)' -count=1 -timeout=180s`\n- `go test -race ./table -run 'TestWriteRecords/TestEarlyStopCancelsRecordProduction' -count=1 -timeout=120s`

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1866,12 +1866,13 @@ func recordsToDataFiles(ctx context.Context, rootLocation string, meta *Metadata
 func unpartitionedWrite(ctx context.Context, factory *writerFactory, records iter.Seq2[arrow.RecordBatch, error]) iter.Seq2[iceberg.DataFile, error] {
 	outputCh := make(chan iceberg.DataFile, 1)
 	errCh := make(chan error, 1)
+	writerCtx, cancel := context.WithCancel(ctx)
 
 	go func() {
 		defer close(outputCh)
 		defer factory.stopCount()
 
-		writer := factory.newRollingDataWriter(ctx, "", nil, outputCh)
+		writer := factory.newRollingDataWriter(writerCtx, "", nil, outputCh)
 		for rec, err := range records {
 			if err != nil {
 				errCh <- err
@@ -1879,6 +1880,15 @@ func unpartitionedWrite(ctx context.Context, factory *writerFactory, records ite
 				writer.abortAndWait()
 
 				return
+			}
+			select {
+			case <-writerCtx.Done():
+				errCh <- context.Cause(writerCtx)
+				close(errCh)
+				writer.abortAndWait()
+
+				return
+			default:
 			}
 			if err := writer.Add(rec); err != nil {
 				errCh <- err
@@ -1899,6 +1909,7 @@ func unpartitionedWrite(ctx context.Context, factory *writerFactory, records ite
 			for range outputCh {
 			}
 		}()
+		defer cancel()
 		for df := range outputCh {
 			if !yield(df, nil) {
 				return
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1881,15 +1881,7 @@ func unpartitionedWrite(ctx context.Context, factory *writerFactory, records ite
 
 				return
 			}
-			select {
-			case <-writerCtx.Done():
-				errCh <- context.Cause(writerCtx)
-				close(errCh)
-				writer.abortAndWait()
 
-				return
-			default:
-			}
 			if err := writer.Add(rec); err != nil {
 				errCh <- err
 				close(errCh)
--- a/table/partitioned_fanout_writer.go
+++ b/table/partitioned_fanout_writer.go
@@ -180,7 +180,7 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, writerCtx c
 	return nil
 }
 
-func (p *partitionedFanoutWriter) yieldDataFiles(fanoutWorkers *errgroup.Group, inputRecordsCh chan arrow.RecordBatch, outputDataFilesCh chan iceberg.DataFile, cancel context.CancelFunc) iter.Seq2[iceberg.DataFile, error] {
+func (p *partitionedFanoutWriter) yieldDataFiles(fanoutWorkers *errgroup.Group, inputRecordsCh <-chan arrow.RecordBatch, outputDataFilesCh chan iceberg.DataFile, cancel context.CancelFunc) iter.Seq2[iceberg.DataFile, error] {
 	return yieldDataFiles(
 		p.writerFactory,
 		fanoutWorkers,
@@ -195,7 +195,7 @@ func (p *partitionedFanoutWriter) yieldDataFiles(fanoutWorkers *errgroup.Group, 
 func yieldDataFiles(
 	writerFactory *writerFactory,
 	fanoutWorkers *errgroup.Group,
-	inputRecordsCh chan arrow.RecordBatch,
+	inputRecordsCh <-chan arrow.RecordBatch,
 	outputDataFilesCh chan iceberg.DataFile,
 	closeAll func() error,
 	abortAll func(),
@@ -208,6 +208,8 @@ func yieldDataFiles(
 		defer close(outputDataFilesCh)
 		defer cancel()
 		err := fanoutWorkers.Wait()
+		// Wait includes the feeder, which closes inputRecordsCh. Any batches
+		// remaining here were retained by the feeder but never dequeued.
 		for record := range inputRecordsCh {
 			record.Release()
 		}
@@ -221,6 +223,8 @@ func yieldDataFiles(
 	}()
 
 	return func(yield func(iceberg.DataFile, error) bool) {
+		// LIFO defer order matters: cancel signals the producer first,
+		// then draining outputDataFilesCh lets in-flight sends complete.
 		defer func() {
 			for range outputDataFilesCh {
 			}
--- a/table/partitioned_fanout_writer.go
+++ b/table/partitioned_fanout_writer.go
@@ -180,7 +180,7 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, writerCtx c
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


table/partitioned_fanout_writer.go
```diff
@@ -175,10 +180,11 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, writerCtx c
 	return nil
 }
 
-func (p *partitionedFanoutWriter) yieldDataFiles(fanoutWorkers *errgroup.Group, outputDataFilesCh chan iceberg.DataFile, writerCancel context.CancelFunc) iter.Seq2[iceberg.DataFile, error] {
+func (p *partitionedFanoutWriter) yieldDataFiles(fanoutWorkers *errgroup.Group, inputRecordsCh chan arrow.RecordBatch, outputDataFilesCh chan iceberg.DataFile, writerCancel context.CancelFunc) iter.Seq2[iceberg.DataFile, error] {
```
tanmayrauth: The param here is still named `writerCancel`, but it now receives the combined `cancel` that stops both the fanout/feeder context and the writer context. The old name understates that — a maintainer reading
  `writerCancel` won't realize calling it also cancels record production, which is the whole point of this change. Suggest renaming it to `cancel` to match the renamed param on the shared `yieldDataFiles`
  below.

table/write_records_test.go
```diff
@@ -233,6 +234,63 @@ func (s *WriteRecordsTestSuite) TestSmallTargetFileSizeProducesMultipleFiles() {
 	s.Equal(int64(1000), totalRows)
 }
 
+func (s *WriteRecordsTestSuite) TestEarlyStopCancelsRecordProduction() {
```
tanmayrauth, line 237: Nice regression. One gap: it covers unpartitioned + partitioned data writes but not  positionDeletePartitionedFanoutWriter, which took the identical cancel/drain change. The shared
  yieldDataFiles/startRecordFeeder path is exercised by the partitioned case, so this is low priority — but its processBatch path is the one bit that differs and currently has no early-stop coverage. A pos-delete variant of this test would close that.

tanmayrauth (commented):
Traced the cancellation paths end-to-end — the fix is correct and the checked allocator genuinely validates the memory-release claim. Just two non-blocking nits inline.

table/arrow_utils.go
```diff
@@ -1880,6 +1881,15 @@ func unpartitionedWrite(ctx context.Context, factory *writerFactory, records ite
 
 				return
 			}
+			select {
+			case <-writerCtx.Done():
+				errCh <- context.Cause(writerCtx)
```
laskoviymishka: I think there's a subtle wrinkle here. This select fires after we've already pulled `rec` off the iterator but before `writer.Add(rec)`, so on cancellation that batch is dropped between the two lines — harmless today since the `releasing` wrapper frees it, but worth noting.

The bigger thing is what we send. `writerCtx` is a plain `WithCancel`, so `context.Cause(writerCtx)` is always `context.Canceled`, and on the early-stop path the iterator returns before it ever reads `errCh` — so that value is silently discarded. Anyone who later wires up a read of `errCh` after early stop would see a spurious `context.Canceled` reported as the outcome of what was actually a clean stop.

`writer.Add`'s own blocking `<-ctx.Done()` check is already a cancellation point, so one option is to drop this non-blocking select entirely and lean on that. If we want to keep the eager check, I'd switch to `WithCancelCause`, call `cancel(nil)` on clean stop, and only send to `errCh` when `context.Cause` is non-nil. wdyt?

table/arrow_utils.go
```diff
@@ -1880,6 +1881,15 @@ func unpartitionedWrite(ctx context.Context, factory *writerFactory, records ite
 
 				return
 			}
+			select {
```
laskoviymishka: Small one, assuming the select stays: there's no blank line between the closing `}` of the `if err != nil` block and this `select`, which wsl will flag — I'm fairly sure CI fails on it. Worth adding one.

table/partitioned_fanout_writer.go
```diff
@@ -175,32 +180,37 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, writerCtx c
...
 ) iter.Seq2[iceberg.DataFile, error] {
 	// Use a channel to safely communicate the error from the goroutine
 	// to avoid a data race between writing err in the goroutine and reading it in the iterator.
 	errCh := make(chan error, 1)
 	go func() {
 		defer close(outputDataFilesCh)
-		defer writerCancel()
+		defer cancel()
 		err := fanoutWorkers.Wait()
+		for record := range inputRecordsCh {
```
laskoviymishka, line 214: Could we add a one-liner on why this drain is safe? It took me a second: `startRecordFeeder` is part of `fanoutWorkers`, so by the time `Wait()` returns the feeder has already exited and closed `inputRecordsCh`. This only ever releases batches the feeder `Retain()`ed but no worker dequeued — dequeued ones are freed by the worker's own `defer Release()`, so there's no double-release. Mostly the comment guards against someone later pulling the feeder out of the errgroup and quietly deadlocking this loop.

table/partitioned_fanout_writer.go
```diff
@@ -215,6 +225,7 @@ func yieldDataFiles(
 			for range outputDataFilesCh {
 			}
 		}()
+		defer cancel()
```
laskoviymishka, line 235: `cancel()` is deferred here and again in the goroutine at line 209 — both idempotent so it's fine, but the intent isn't obvious (this one fires first via LIFO and is the early-stop signal; the goroutine's is the fallback). `clustered_writer.go:183-186` already has a comment for exactly this — mind mirroring it?

table/partitioned_fanout_writer.go
```diff
@@ -175,32 +180,37 @@ func (p *partitionedFanoutWriter) processRecord(ctx context.Context, writerCtx c
...
 		p.writerFactory.abortAll,
-		writerCancel,
+		cancel,
 	)
 }
 
 func yieldDataFiles(
 	writerFactory *writerFactory,
 	fanoutWorkers *errgroup.Group,
+	inputRecordsCh chan arrow.RecordBatch,
```
laskoviymishka: Since `yieldDataFiles` only ever drains this, could we type it `<-chan arrow.RecordBatch` here (and in the two wrapper methods + the pos-delete one)? Makes the read-only intent explicit and stops anyone accidentally sending into a channel `startRecordFeeder` owns.

table/pos_delete_partitioned_fanout_writer_test.go
```diff
@@ -402,6 +403,66 @@ func TestPositionDeletePartitionedFanoutWriterRoutesPartitionsIndependently(t *t
 	assert.Equal(t, int64(1), byPart[2].Count(), "id=2 delete file must contain only the one row targeting pathB")
 }
 
+func TestPositionDeletePartitionedFanoutWriterEarlyStopCancelsRecordProduction(t *testing.T) {
```
laskoviymishka, line 409: Every other top-level test in this file opens with `t.Parallel()` — this one's missing it. Worth adding for consistency.

table/write_records_test.go
```diff
@@ -233,6 +234,63 @@ func (s *WriteRecordsTestSuite) TestSmallTargetFileSizeProducesMultipleFiles() {
...
+				s.ctx, tbl, schema, records, table.WithTargetFileSize(1), table.WithMaxWriteWorkers(1),
+			) {
+				s.Require().NoError(writeErr)
+				s.Require().NotNil(df)
+
+				break
+			}
+
+			s.Positive(produced.Load())
+			s.Less(produced.Load(), int32(1000))
```
laskoviymishka: `< 1000` is a pretty loose bound — with `WithMaxWriteWorkers(1)` and `WithTargetFileSize(1)` we break after the first file, so real production is only 1-3 batches. As written this still passes even if cancellation were delayed by 990 records, which is most of what we're trying to catch. I'd tighten it to something like `< 10`. Same assertion at `pos_delete_partitioned_fanout_writer_test.go:463`.

table/pos_delete_partitioned_fanout_writer_test.go
```diff
@@ -402,6 +403,66 @@ func TestPositionDeletePartitionedFanoutWriterRoutesPartitionsIndependently(t *t
...
+
+	for dataFile, writeErr := range writer.Write(t.Context(), 1) {
+		require.NoError(t, writeErr)
+		require.NotNil(t, dataFile)
+
+		break
+	}
+
+	assert.Positive(t, produced.Load())
+	assert.Less(t, produced.Load(), int32(1000))
```
laskoviymishka: The PR description calls out that retained Arrow memory gets released, verified by a checked allocator — but this direct-writer test doesn't assert that (only the table-level path does). Could we thread a `memory.CheckedAllocator` through the context and assert `CurrentAllocated() == 0` after the loop? Cheap way to pin the memory claim on this path too.

laskoviymishka (changes requested):
The cancellation wiring looks correct across all three paths, and the checked-allocator test actually verifies that retained memory is released. I agree with Tanmay that the main change is sound.

I’d clean up one thing in the unpartitioned path before merging. The new non-blocking `select` can receive `rec`, notice cancellation, and send `context.Cause(writerCtx)` to `errCh`. But `writerCtx` comes from plain `WithCancel`, so that will always be `context.Canceled`. On early iterator exit, nothing reads the value from `errCh` anyway.

It is harmless, but it turns a normal early stop into an error path without surfacing the error. `writer.Add` already checks the context while blocking, so I’d either remove the eager `select` or switch this path to `WithCancelCause`. I left the details inline.

The rest are small:

* a missing blank line that may trip `wsl` in CI
* short comments explaining why the fanout drain and double cancel are safe
* a receive-only channel type
* a tighter `< 1000` bound in the early-stop test so it would catch a regression

Once the `select` is cleaned up, I’m good to approve.

laskoviymishka (approved):

PR #1596, opened 2026-07-29
feat(metrics):    Instrument commits and emit CommitReport ( #1236)

tanmayrauth:
**What**

  Emits a spec-compliant CommitReport from the commit path on successful completion:

  - doCommit's retry loop now records total-duration and attempts.
  - The committed snapshot's summary is mapped into a CommitReport under Java's commit-report metric names. iceberg-go's summary keys differ from those names, so the mapping translates them (added-files-size →
  added-files-size-bytes, deleted-data-files → removed-data-files, added-position-deletes → added-positional-deletes, …) so the wire format matches Java and dashboards line up across implementations.
  - Metrics with no iceberg-go summary key (DVs, manifest counts) are omitted rather than reported as zero, mirroring Java's counterFrom.

  **Behavior notes**

  - No behavior change for existing users: emission is guarded by metrics.IsNop, so under the default no-op reporter it's a pure no-op (no report assembled). Report failures can never fail a commit.
  - Metadata-only commits emit nothing: a property/schema-only change produces no snapshot, so reporting the unchanged branch head would misattribute a prior snapshot's metrics. Skipped via
  commitAddedSnapshot.
  - Non-default branches resolve the just-committed head via SnapshotByName(branch) rather than the table's current snapshot.

  **Deferred**

  CommitReport.Metadata (Java's engine/commit-context map) is left unset with a TODO — iceberg-go doesn't thread that context into the commit path yet; a follow-up will populate it.

  **Testing**

  - Unit tests for buildCommitReport: name-translated mappings, nil snapshot, absent/unparseable summary values omitted.
  - Integration tests: an append emits a CommitReport with correct operation/attempts/records; a metadata-only commit emits none.
  - go build, go vet, and full table/metrics suites pass.
  
  Realted to #1236

### DIFF
--- a/table/commit_metrics.go
+++ b/table/commit_metrics.go
@@ -18,12 +18,28 @@
 package table
 
 import (
+	"context"
+	"log"
 	"strconv"
 	"time"
 
 	"github.com/apache/iceberg-go"
 	"github.com/apache/iceberg-go/metrics"
 )
+
+// safeReport forwards a report to rep, recovering from a panic in a misbehaving
+// third-party reporter so it can never fail a commit that has already durably
+// succeeded. The Reporter contract requires a reporter never affect the observed
+// operation; compositeReporter isolates each inner reporter this way, but a bare
+// non-composite reporter gets no such protection at the emit site.
+func safeReport(ctx context.Context, rep metrics.Reporter, report metrics.MetricsReport) {
+	defer func() {
+		if v := recover(); v != nil {
+			log.Printf("Warning: metrics reporter %T panicked; recovered: %v", rep, v)
+		}
+	}()
+	rep.Report(ctx, report)
+}
 
 // commitAddedSnapshot reports whether the commit produced a new snapshot.
 // Metadata-only commits (property or schema changes) carry no addSnapshotUpdate
@@ -66,15 +82,15 @@ func summaryCounter(props iceberg.Properties, key string, unit metrics.Unit) *me
 // and every other metric is read from the snapshot summary under its spec key
 // and emitted under Java's commit-report field name so dashboards line up
 // across implementations. Metrics whose summary key iceberg-go does not yet
-// populate (DVs, manifest counts) are absent and therefore omitted — exactly as
-// Java's counterFrom omits summary keys it does not find.
+// populate (DVs) are absent and therefore omitted — exactly as Java's
+// counterFrom omits summary keys it does not find.
 //
 // TODO: the report's Metadata map is left unset. Java's CommitReport carries a
 // metadata map (e.g. engine name/version and other commit context); iceberg-go
 // does not thread that context into the commit path yet, so it is omitted
 // rather than reported empty. Populate it in a follow-up once the context is
 // available.
-func buildCommitReport(tableName string, snap *Snapshot, attempts int, dur time.Duration) metrics.CommitReport {
+func buildCommitReport(tableName string, snap *Snapshot, attempts int64, dur time.Duration) metrics.CommitReport {
 	var (
 		snapshotID int64
 		seqNum     int64
@@ -100,7 +116,7 @@ func buildCommitReport(tableName string, snap *Snapshot, attempts int, dur time.
 		Operation:      operation,
 		Metrics: metrics.CommitMetricsResult{
 			TotalDuration: metrics.NewNanosTimerResult(1, dur.Nanoseconds()),
-			Attempts:      metrics.NewCounterResult(metrics.UnitCount, int64(attempts)),
+			Attempts:      metrics.NewCounterResult(metrics.UnitCount, attempts),
 
 			AddedDataFiles:   count(addedDataFilesKey),
 			RemovedDataFiles: count(deletedDataFilesKey),
@@ -131,6 +147,12 @@ func buildCommitReport(tableName string, snap *Snapshot, attempts int, dur time.
 			AddedEqualityDeletes:   count(addedEqDeletesKey),
 			RemovedEqualityDeletes: count(removedEqDeletesKey),
 			TotalEqualityDeletes:   count(totalEqDeletesKey),
+
+			ManifestsCreated:  count(manifestsCreatedKey),
+			ManifestsReplaced: count(manifestsReplacedKey),
+			ManifestsKept:     count(manifestsKeptKey),
+			// entries-processed in the summary maps to manifest-entries-processed.
+			ManifestEntriesProcessed: count(entriesProcessedKey),
 		},
 	}
 }
--- a/table/commit_metrics.go
+++ b/table/commit_metrics.go
@@ -1,0 +1,130 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/metrics"
+)
+
+// commitAddedSnapshot reports whether the commit produced a new snapshot.
+// Metadata-only commits (property or schema changes) carry no addSnapshotUpdate
+// and must not emit a commit report: they create no snapshot, so the branch head
+// is unchanged and reporting it would attribute a prior snapshot's metrics to
+// this commit. This mirrors Java, whose CommitReport is emitted only from the
+// snapshot-producing path (SnapshotProducer.commit).
+func commitAddedSnapshot(updates []Update) bool {
+	for _, u := range updates {
+		if _, ok := u.(*addSnapshotUpdate); ok {
+			return true
+		}
+	}
+
+	return false
+}
+
+// summaryCounter reads a snapshot-summary property and returns it as a
+// CounterResult, or nil if the key is absent or unparseable. This mirrors
+// Java's CommitMetricsResult.counterFrom exactly: a metric is omitted rather
+// than reported as a zero unless the snapshot summary carries a parseable value
+// for it. As a result iceberg-go and Java emit the same present/absent metric
+// set for an equivalent commit.
+func summaryCounter(props iceberg.Properties, key string, unit metrics.Unit) *metrics.CounterResult {
+	v, ok := props[key]
+	if !ok {
+		return nil
+	}
+	n, err := strconv.ParseInt(v, 10, 64)
+	if err != nil {
+		return nil
+	}
+
+	return metrics.NewCounterResult(unit, n)
+}
+
+// buildCommitReport assembles a CommitReport from the committed snapshot's
+// summary, mirroring Java's CommitMetricsResult.from(commitMetrics,
+// snapshotSummary): attempts and total-duration come from the commit itself,
+// and every other metric is read from the snapshot summary under its spec key
+// and emitted under Java's commit-report field name so dashboards line up
+// across implementations. Metrics whose summary key iceberg-go does not yet
+// populate (DVs, manifest counts) are absent and therefore omitted — exactly as
+// Java's counterFrom omits summary keys it does not find.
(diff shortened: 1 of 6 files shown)


tanmayrauth:
@laskoviymishka  @zeroshade  Can you please take a look when you get chance?

table/table.go
```diff
@@ -464,6 +464,9 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
 		return nil, err
 	}
 
+	commitStart := time.Now()
```
laskoviymishka: `commitStart` is captured here, before `t.fsF(ctx)` and `requireWriteFileIO` resolve the filesystem (both can do I/O), and `time.Since(commitStart)` is sampled at the emit site after `deleteOldMetadata`, which can also hit the network. So `TotalDuration` starts before the commit boundary and ends after it: on a REST catalog with dynamic FS resolution plus old-metadata cleanup it's systematically larger than Java, which times only the `CommitTable` submission loop.

Since the whole point here is to mirror Java's CommitReport, an inflated duration makes cross-implementation comparison misleading. I'd move `commitStart` to just before the retry loop and capture the elapsed time right after the `break` (before orphan cleanup and `deleteOldMetadata`), so the window brackets the commit itself.

table/commit_metrics.go
```diff
@@ -0,0 +1,136 @@
...
+	return metrics.NewCounterResult(unit, n)
+}
+
+// buildCommitReport assembles a CommitReport from the committed snapshot's
+// summary, mirroring Java's CommitMetricsResult.from(commitMetrics,
+// snapshotSummary): attempts and total-duration come from the commit itself,
+// and every other metric is read from the snapshot summary under its spec key
+// and emitted under Java's commit-report field name so dashboards line up
+// across implementations. Metrics whose summary key iceberg-go does not yet
+// populate (DVs, manifest counts) are absent and therefore omitted — exactly as
```
laskoviymishka: This comment says iceberg-go doesn't yet populate manifest counts, but `rewrite_manifests.go` already writes `manifests-created`, `manifests-replaced`, `manifests-kept`, and `entries-processed` into the snapshot summary on every non-trivial RewriteManifests commit. So a CommitReport emitted after a RewriteManifests silently drops all four manifest metrics that Java's `CommitMetricsResult.from` would populate, and RewriteManifests is the one path that always produces them.

The four summary keys are package-level constants in this same package, so we can populate them directly in the `CommitMetricsResult` below:

```go
ManifestsCreated:         count(manifestsCreatedKey),
ManifestsReplaced:        count(manifestsReplacedKey),
ManifestsKept:            count(manifestsKeptKey),
ManifestEntriesProcessed: count(entriesProcessedKey),
```

One thing to watch on the wire mapping: `entries-processed` in the summary emits under `manifest-entries-processed`, so the lookup uses `entriesProcessedKey` but the field is `ManifestEntriesProcessed` (the fixture already has this right). And drop the "does not yet populate ... manifest counts" line from the comment.

table/table.go
```diff
@@ -632,6 +637,28 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
...
+	// attribute a prior snapshot's metrics to this commit.
+	if rep := t.MetricsReporter(); !metrics.IsNop(rep) && commitAddedSnapshot(updates) {
+		committed := newMeta.CurrentSnapshot()
+		if co.branch != "" {
+			if s := newMeta.SnapshotByName(co.branch); s != nil {
+				committed = s
+			}
+		}
+		if committed != nil {
+			rep.Report(ctx,
```
laskoviymishka: This `Report` call isn't guarded, and it runs after `CommitTable` has already durably succeeded. The `Reporter` contract says a reporter must never fail the operation it observes, but nothing structurally stops a buggy third-party impl from panicking here, and if one does, the panic propagates out of `doCommit` and the caller sees an error for a commit that actually landed. That's the worst kind of confusion: a retry or rollback against already-committed state.

`compositeReporter` recovers around each inner report, but a bare non-composite reporter gets no protection. I'd wrap this in a small `safeReport` helper that recovers and logs, mirroring the composite's isolation. The scan path in `scanner.go` has the same gap, though that one's pre-existing and can be a follow-up.

table/table_metrics_commit_test.go
```diff
@@ -0,0 +1,158 @@
...
+	require.NotNil(t, cr.Metrics.AddedRecords)
+	assert.Equal(t, int64(3), cr.Metrics.AddedRecords.Value)
+}
+
+// TestMetadataOnlyCommitEmitsNoCommitReport pins that a commit which produces no
+// snapshot (here a property-only change) emits no CommitReport, even though the
+// table already has a snapshot. Reporting the unchanged branch head would
+// misattribute the earlier append's metrics to this commit.
+func TestMetadataOnlyCommitEmitsNoCommitReport(t *testing.T) {
+	registerCommitReportSink()
```
laskoviymishka, line 112: `commitReportSink` is a package-level singleton shared with `TestCommitEmitsCommitReport`, and this test never resets it at the top: it only resets mid-test after the seeding append. Under sequential order it happens to pass, but with `-shuffle=on`, `-count=2`, or a future committing test running first, stale CommitReports would already be in the sink and this assertion passes for the wrong reason. Worse, a real regression where the metadata-only commit does emit would get masked.

I'd add `commitReportSink.Reset()` as the first statement here. Better still, drop the singleton and inject a fresh `InMemoryReporter` per test via `WithMetricsReporter` so there's no shared state to reason about at all.

table/table.go
```diff
@@ -632,6 +637,28 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
...
+	//
+	// Mirrors the scan path: building the report is skipped for a no-op
+	// reporter (the opt-in default), since a nop discards it and assembling one
+	// would be pure overhead. A metadata-only commit produces no snapshot and
+	// must be skipped too — its branch head is unchanged, so reporting it would
+	// attribute a prior snapshot's metrics to this commit.
+	if rep := t.MetricsReporter(); !metrics.IsNop(rep) && commitAddedSnapshot(updates) {
+		committed := newMeta.CurrentSnapshot()
+		if co.branch != "" {
+			if s := newMeta.SnapshotByName(co.branch); s != nil {
```
laskoviymishka: If `co.branch` is set but `SnapshotByName` returns nil (e.g. the first commit that creates a new non-default branch), `committed` silently stays as `CurrentSnapshot()`, the default-branch head. That's a different snapshot than the one we just committed, so the report would carry the wrong snapshot ID, sequence number, and operation. This is exactly the case the "prefer branch head" comment is meant to handle, so the silent fallback inverts the intent.

I'd either compare the resolved snapshot's ID against what was committed, or skip emission when the branch lookup comes back nil rather than attributing the wrong snapshot. A test that commits to a fresh branch would pin this down.

table/table.go
```diff
@@ -464,6 +464,9 @@ func (t Table) doCommit(ctx context.Context, updates []Update, reqs []Requiremen
 		return nil, err
 	}
 
+	commitStart := time.Now()
+	attemptsUsed := 0
```
laskoviymishka: Two small things on `attemptsUsed`. It's declared `int` and later set with `int(attempt) + 1`, but the `Attempts` counter is `int64` and `attempt` ranges over a `uint` bounded by `maxRetryCount = math.MaxUint32`. On a 32-bit build a large attempt index wraps negative through the `int` cast and emits a negative counter. I'd declare it `int64`, use `int64(attempt) + 1`, and take `attempts int64` in `buildCommitReport` to drop the cast.

Separately, initializing to `0` is a latent trap: today the budget-exhausted path returns an error before the emit block so `0` never escapes, but nothing structurally enforces that. If someone later adds a second success exit, a `0` attempts counter would slip out. Initializing to `1` (or deriving from `attempt` at the emit site) makes the invariant hold by construction.

table/commit_metrics_test.go
```diff
@@ -0,0 +1,118 @@
...
+	require.NotNil(t, m.TotalDuration)
+	assert.Equal(t, (7 * time.Millisecond).Nanoseconds(), m.TotalDuration.TotalDuration)
+	require.NotNil(t, m.Attempts)
+	assert.Equal(t, int64(2), m.Attempts.Value)
+
+	// Direct (same-name) mappings.
+	require.NotNil(t, m.AddedDataFiles)
+	assert.Equal(t, int64(4), m.AddedDataFiles.Value)
+	assert.Equal(t, int64(10), m.TotalDataFiles.Value)
+	assert.Equal(t, int64(12345), m.AddedRecords.Value)
```
laskoviymishka, line 77: These `.Value` dereferences aren't nil-guarded: `m.TotalDataFiles`, `m.AddedRecords`, `m.TotalRecords`, and the ones below (`m.TotalFilesSizeBytes` / `m.AddedPositionalDeletes` / `m.AddedEqualityDeletes` / `m.AddedEqualityDeleteFiles`). The test guards `AddedDataFiles` and `RemovedDataFiles` with `require.NotNil` first but drops the pattern for the rest. If `buildCommitReport` ever returns nil for one of these (key rename, ParseInt failure, refactor), the test panics with a nil deref instead of a clean FAIL. I'd add `require.NotNil` before each.

While we're here: the summary seeds `removedFileSizeKey: "100"` but nothing asserts `RemovedFilesSizeBytes` maps through, so that translation isn't actually covered. Worth an assert, since round-trip coverage on every populated field is the point of this test.

table/table_metrics_commit_test.go
```diff
@@ -0,0 +1,158 @@
...
+		if cr, ok := r.(metrics.CommitReport); ok {
+			commits = append(commits, cr)
+		}
+	}
+	require.NotEmpty(t, commits, "an append commit must emit a CommitReport")
+
+	cr := commits[len(commits)-1]
+	assert.Equal(t, "append", cr.Operation)
+	require.NotNil(t, cr.Metrics.Attempts)
+	assert.GreaterOrEqual(t, cr.Metrics.Attempts.Value, int64(1))
```
laskoviymishka, line 99: This only asserts `Attempts >= 1`, and nothing in the suite ever drives `attemptsUsed` above 1; every test is a clean single-attempt success. The `int(attempt) + 1` assignment is the most novel bit of the emit path, so the coverage here is effectively vacuous for it.

I'd add a test that injects a one-time retryable `ErrCommitFailed` and asserts `Attempts.Value == 2`, so the retry-count logic is actually exercised.

laskoviymishka (changes requested):
Nice work on this. The `buildCommitReport` assembly is clean, and reading each metric from the snapshot summary and omitting absent keys rather than zeroing them mirrors Java's `counterFrom` exactly. Skipping metadata-only commits and no-op reporters on the emit path is the right call too.

I'd hold it before merging though, since the stated goal is to mirror Java's CommitReport and a couple of things break that parity today.

Main concern is metric fidelity. `TotalDuration` doesn't bracket the commit boundary: it starts before filesystem resolution and stops after `deleteOldMetadata`, both of which can do I/O, so it comes out systematically larger than Java, which times only the `CommitTable` loop. And the doc comment on `buildCommitReport` says iceberg-go doesn't yet populate manifest counts, but `rewrite_manifests.go` already writes all four keys, so every RewriteManifests commit emits a report that silently drops the four manifest metrics Java would include, on the one path that always produces them.

- Move the `commitStart`/`time.Since` window to bracket the commit loop itself, not FS setup and metadata cleanup.
- Populate the four manifest fields from the existing summary keys, and fix the comment that claims they're absent.
- Guard the `rep.Report` call so a panicking third-party reporter can't fail a commit that already succeeded.
- Reset the shared sink at the top of the metadata-only test (or inject a per-test reporter), and add a test that actually drives a retry so `Attempts > 1` is exercised.

The rest are smaller notes inline. Once those are addressed I'm happy to take another pass and approve.

tanmayrauth:
Thanks @laskoviymishka  for the thorough review,  all points addressed. 

- TotalDuration now brackets only the commit loop: commitStart moved to just before the retry loop, and the elapsed time is snapped right after the success break, before orphan cleanup and deleteOldMetadata.
- Manifest metrics: populated the four fields (manifests-created/replaced/kept and entries-processed → manifest-entries-processed) from the existing summary keys, and dropped the stale "doesn't yet populate manifest counts" comment.
- Panicking reporter: emission is now wrapped in a safeReport helper that recovers and logs, so a misbehaving third-party reporter can never fail an already-committed commit. Since Scan.Reporter() also hands back the user's reporter unwrapped, I folded the scan path into the same helper rather than deferring it.
- Non-default branch: now skips emission when SnapshotByName returns nil instead of silently falling back to CurrentSnapshot(), so it can't attribute the wrong snapshot.
- attemptsUsed: now int64 (initialized to 1, set via int64(attempt)+1) to avoid the 32-bit wrap and the 0-leak trap, with buildCommitReport taking int64 directly.
- Tests:
    - the metadata-only test resets the sink up front so it no longer depends on ordering;
    - the unit test now require.NotNil-guards every .Value deref and asserts the removed-files-size round-trip plus all four manifest fields;
    - added a test that injects one retryable ErrCommitFailed and asserts Attempts == 2, so the retry-count path is actually exercised.

laskoviymishka (approved):

PR #1597, opened 2026-07-29
feat(expr): bbox predicate evaluation for geo pruning

### LINKED ISSUE #994: feat(expr): bbox predicate evaluation for geo pruning

tanmayrauth:
Summary
  
  Adds geospatial BBoxIntersects / BBoxNotIntersects predicates and wires them through the bound-expression visitor so data files whose geometry bounds cannot overlap a query bounding box are pruned during scan planning. Reuses the single-value geo bound codec from #993.
 

  What changed

  - New predicate (exprs.go) — BoundingBox (planar XY box with Valid/Equals) and the BBoxIntersects(term, bbox) constructor producing an UnboundPredicate. Binding rejects non-geo columns (ErrType) and  transform terms; the constructor panics on a nil term or an invalid box (NaN / inverted min-max) to prevent silent mis-pruning. Adds OpBBoxIntersects / OpBBoxNotIntersects (kept after the boolean ops so existing op-group ranges are undisturbed) with negation and String() support.
  - Visitor interface (visitors.go) — adds VisitBBoxIntersects / VisitBBoxNotIntersects to BoundBooleanExprVisitor, dispatches them in VisitBoundPredicate, and collapses bbox predicates to AlwaysTrue in the  sanitize visitor (they carry query-box coordinates, not user literals, and have no REST JSON form). The row-level exprEvaluator conservatively keeps every row — full ST_Intersects remains a query-engine
  concern.
  - Data-file pruning (table/evaluators.go) — inclusiveMetricsEval.VisitBBoxIntersects prunes files whose decoded geometry bounds are disjoint from the query box (all-null files also prune). Geography columns are never pruned (geodesic bounds may cross the antimeridian). BBoxNotIntersects, strict eval, manifest eval, and the bloom collector all conservatively keep files, since bounds alone can't answer them.
  - Codec helpers (table/internal/geo_codec.go) — GeoBoundsXY decodes lower/upper bounds to planar extents (fails safe on missing/malformed/NaN), and BBoxIntersectsXY does the closed-interval overlap test.
  - Serialization guards — bbox predicates return an error from MarshalJSON (no REST grammar) and panic in the substrait converter, since they exist only for local scan planning.

  Testing

  expr_bbox_test.go, table/evaluators_geo_test.go, table/internal/geo_codec_internal_test.go, and additions to visitors_test.go cover construction/validation, binding, pruning decisions (geometry vs geography, null-only, malformed bounds), and codec edge cases.

Closes: #994

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -35,4 +35,11 @@ var (
 	ErrBadLiteral              = errors.New("invalid literal value")
 	ErrInvalidBinSerialization = errors.New("invalid binary serialization")
 	ErrResolve                 = errors.New("cannot resolve type")
+	// ErrBBoxNotSerializable is returned when marshaling a geospatial bbox
+	// predicate to REST expression JSON: bbox predicates exist only for local
+	// scan planning and have no representation in the REST expression grammar.
+	// It wraps ErrNotImplemented (not ErrInvalidArgument) because this is a
+	// contract/not-implemented condition, not a bad caller argument - matching
+	// how the substrait backstop surfaces the same predicate.
+	ErrBBoxNotSerializable = fmt.Errorf("%w: geospatial bbox predicates cannot be serialized to REST expression JSON", ErrNotImplemented)
 )
--- a/expr_bbox_test.go
+++ b/expr_bbox_test.go
@@ -1,0 +1,229 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package iceberg_test
+
+import (
+	"encoding/json"
+	"math"
+	"testing"
+
+	"github.com/apache/iceberg-go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func geoTestSchema() *iceberg.Schema {
+	return iceberg.NewSchema(1,
+		iceberg.NestedField{ID: 1, Name: "geom", Type: iceberg.GeometryType{}, Required: false},
+		iceberg.NestedField{ID: 2, Name: "geog", Type: iceberg.GeographyType{}, Required: false},
+		iceberg.NestedField{ID: 3, Name: "num", Type: iceberg.PrimitiveTypes.Int32, Required: false},
+	)
+}
+
+func TestBBoxIntersectsConstruction(t *testing.T) {
+	bbox := iceberg.BoundingBox{MinX: 0, MinY: 0, MaxX: 10, MaxY: 10}
+	pred := iceberg.BBoxIntersects(iceberg.Reference("geom"), bbox)
+
+	assert.Equal(t, iceberg.OpBBoxIntersects, pred.Op())
+	assert.Equal(t, iceberg.Reference("geom"), pred.Term())
+	assert.Contains(t, pred.String(), "BBoxIntersects")
+
+	// Negation flips the operation and round-trips back.
+	neg := pred.Negate()
+	assert.Equal(t, iceberg.OpBBoxNotIntersects, neg.Op())
+	assert.True(t, pred.Equals(neg.Negate()))
+	assert.False(t, pred.Equals(neg))
+}
+
+func TestBBoxIntersectsNilTermPanics(t *testing.T) {
+	assert.Panics(t, func() {
+		iceberg.BBoxIntersects(nil, iceberg.BoundingBox{})
+	})
+}
+
+func TestBoundingBoxValid(t *testing.T) {
+	nan := math.NaN()
+	inf := math.Inf(1)
+	tests := []struct {
+		name string
+		bbox iceberg.BoundingBox
+		want bool
+	}{
+		{"zero", iceberg.BoundingBox{}, true},
+		{"normal", iceberg.BoundingBox{MinX: 0, MinY: 0, MaxX: 10, MaxY: 10}, true},
+		{"open half-plane via inf", iceberg.BoundingBox{MinX: -inf, MinY: -inf, MaxX: inf, MaxY: inf}, true},
+		{"inverted x", iceberg.BoundingBox{MinX: 10, MinY: 0, MaxX: 0, MaxY: 10}, false},
+		{"inverted y", iceberg.BoundingBox{MinX: 0, MinY: 10, MaxX: 10, MaxY: 0}, false},
+		{"nan min", iceberg.BoundingBox{MinX: nan, MinY: 0, MaxX: 10, MaxY: 10}, false},
+		{"nan max", iceberg.BoundingBox{MinX: 0, MinY: 0, MaxX: 10, MaxY: nan}, false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, tt.bbox.Valid())
+		})
+	}
+}
+
+func TestBBoxIntersectsInvalidBoxPanics(t *testing.T) {
+	// An inverted or NaN box would silently mis-prune, so construction rejects it.
+	assert.Panics(t, func() {
+		iceberg.BBoxIntersects(iceberg.Reference("geom"),
+			iceberg.BoundingBox{MinX: 10, MinY: 0, MaxX: 0, MaxY: 10})
+	})
+	assert.Panics(t, func() {
+		iceberg.BBoxIntersects(iceberg.Reference("geom"),
+			iceberg.BoundingBox{MinX: math.NaN(), MinY: 0, MaxX: 10, MaxY: 10})
+	})
+}
+
+func TestBBoxIntersectsBind(t *testing.T) {
+	sc := geoTestSchema()
+	bbox := iceberg.BoundingBox{MinX: 1, MinY: 2, MaxX: 3, MaxY: 4}
+
+	for _, name := range []string{"geom", "geog"} {
+		t.Run(name, func(t *testing.T) {
+			bound, err := iceberg.BBoxIntersects(iceberg.Reference(name), bbox).Bind(sc, true)
+			require.NoError(t, err, "binding to %s", name)
+
+			bp, ok := bound.(iceberg.BoundBBoxPredicate)
+			require.True(t, ok)
+			assert.Equal(t, iceberg.OpBBoxIntersects, bp.Op())
+			assert.True(t, bbox.Equals(bp.BBox()))
+			assert.Equal(t, name, bp.Ref().Field().Name)
+		})
+	}
+}
+
+func TestBBoxIntersectsBindNonGeoRejected(t *testing.T) {
+	sc := geoTestSchema()
+	_, err := iceberg.BBoxIntersects(iceberg.Reference("num"), iceberg.BoundingBox{}).Bind(sc, true)
+	require.Error(t, err)
+	assert.ErrorIs(t, err, iceberg.ErrType)
+}
+
+// TestBBoxIntersectsNotAUnaryPredicate locks in the invariant that a bound bbox
+// predicate must NOT satisfy BoundUnaryPredicate. If it did (e.g. by regaining an
+// AsUnbound(Reference) method), every type-switch on BoundUnaryPredicate -
+// column-name translation, transform projection - would silently match it and
+// rebuild it as a generic unary predicate, which reaches substrait (panic) or
+// drops rows on a missing column. Bbox is dispatched via BoundBBoxPredicate only.
+func TestBBoxIntersectsNotAUnaryPredicate(t *testing.T) {
+	sc := geoTestSchema()
+	bound, err := iceberg.BBoxIntersects(iceberg.Reference("geom"),
+		iceberg.BoundingBox{MinX: 0, MinY: 0, MaxX: 10, MaxY: 10}).Bind(sc, true)
+	require.NoError(t, err)
+
+	_, isBBox := bound.(iceberg.BoundBBoxPredicate)
+	require.True(t, isBBox)
+	_, isUnary := bound.(iceberg.BoundUnaryPredicate)
+	assert.False(t, isUnary, "a bound bbox predicate must not satisfy BoundUnaryPredicate")
+}
+
+func TestBBoxIntersectsBindUnknownField(t *testing.T) {
+	sc := geoTestSchema()
+	_, err := iceberg.BBoxIntersects(iceberg.Reference("missing"), iceberg.BoundingBox{}).Bind(sc, true)
+	require.Error(t, err)
+}
+
+// Bound negation preserves the box and can be rewritten through RewriteNotExpr.
+func TestBBoxNotIntersectsRewrite(t *testing.T) {
+	sc := geoTestSchema()
+	bbox := iceberg.BoundingBox{MinX: 1, MinY: 2, MaxX: 3, MaxY: 4}
+
+	bound, err := iceberg.BBoxIntersects(iceberg.Reference("geom"), bbox).Bind(sc, true)
+	require.NoError(t, err)
+
+	rewritten, err := iceberg.RewriteNotExpr(iceberg.NewNot(bound))
+	require.NoError(t, err)
+	assert.Equal(t, iceberg.OpBBoxNotIntersects, rewritten.Op())
+}
+
+// Geospatial predicates have no REST JSON representation and must surface an
+// error rather than silently marshalling to an empty object.
+func TestBBoxIntersectsNotJSONSerializable(t *testing.T) {
+	sc := geoTestSchema()
+	unbound := iceberg.BBoxIntersects(iceberg.Reference("geom"), iceberg.BoundingBox{})
+
+	_, err := json.Marshal(unbound)
+	require.Error(t, err)
+
+	bound, err := unbound.Bind(sc, true)
+	require.NoError(t, err)
+	_, err = json.Marshal(bound)
+	require.Error(t, err)
+}
+
+// TestBBoxTranslateColumnNames guards column-name translation, the single
+// function where a bbox predicate previously went wrong in two directions: with
+// the geo column present it survived as a BoundUnaryPredicate and later panicked
+// in substrait, and with the column absent (a file written before the geo column
+// existed) it collapsed to AlwaysFalse and silently dropped every row. Both must
+// resolve to AlwaysTrue so the record filter conservatively keeps every row -
+// bbox pruning is the metrics evaluator's job, not the record filter's.
+func TestBBoxTranslateColumnNames(t *testing.T) {
+	sc := geoTestSchema()
+	bound, err := iceberg.BBoxIntersects(iceberg.Reference("geom"),
+		iceberg.BoundingBox{MinX: 0, MinY: 0, MaxX: 10, MaxY: 10}).Bind(sc, true)
+	require.NoError(t, err)
+
+	t.Run("column present", func(t *testing.T) {
+		out, err := iceberg.TranslateColumnNames(bound, sc)
+		require.NoError(t, err)
+		assert.Equal(t, iceberg.AlwaysTrue{}, out)
+	})
+
+	t.Run("column absent (schema evolution)", func(t *testing.T) {
+		// A file schema that predates the geometry column.
+		fileSchema := iceberg.NewSchema(1,
+			iceberg.NestedField{ID: 3, Name: "num", Type: iceberg.PrimitiveTypes.Int32})
+		out, err := iceberg.TranslateColumnNames(bound, fileSchema)
+		require.NoError(t, err)
+		assert.Equal(t, iceberg.AlwaysTrue{}, out,
+			"a file missing the geo column must keep every row, not drop them")
+	})
+}
+
+// SanitizeExpression must not fail on a bbox predicate: it has no user literal to
+// mask and no REST JSON form, so it collapses to always-true while the rest of
+// the expression sanitizes normally. Covers both the unbound path (how a scan's
+// row filter is stored) and the bound path.
+func TestBBoxIntersectsSanitize(t *testing.T) {
+	sc := geoTestSchema()
+	bbox := iceberg.BoundingBox{MinX: 0, MinY: 0, MaxX: 10, MaxY: 10}
+	unbound := iceberg.BBoxIntersects(iceberg.Reference("geom"), bbox)
+
+	// Unbound path: scan.rowFilter is stored unbound.
+	sanitized, err := iceberg.SanitizeExpression(unbound)
+	require.NoError(t, err)
+	assert.Equal(t, iceberg.AlwaysTrue{}, sanitized)
+
+	// Bound path.
+	bound, err := unbound.Bind(sc, true)
+	require.NoError(t, err)
+	sanitized, err = iceberg.SanitizeExpression(bound)
+	require.NoError(t, err)
+	assert.Equal(t, iceberg.AlwaysTrue{}, sanitized)
+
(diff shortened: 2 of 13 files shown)


table/substrait/substrait.go
```diff
@@ -443,3 +443,11 @@ func (t *toSubstraitExpr) VisitNotStartsWith(term iceberg.BoundTerm, lit iceberg
 		t.makeLitFunc(startsWithID, term, lit).(expr.FuncArgBuilder),
 	)
 }
+
+func (t *toSubstraitExpr) VisitBBoxIntersects(iceberg.BoundTerm, iceberg.BoundingBox) expr.Builder {
+	panic("geospatial bbox predicates cannot be converted to substrait")
```
laskoviymishka: This panic isn't just defensive — it's reachable on what looks like the primary use of the new API. Run `table.Scan(WithFilter(BBoxIntersects(...)))` and the predicate survives column-name translation (because `boundBBoxPredicate` satisfies `BoundUnaryPredicate`), survives `Bind`, and lands here via `getRecordFilter` → `ConvertExpr`. `VisitExpr`'s `recover()` turns it into an error, so the whole scan fails — which contradicts the "conservatively keeps every row" contract the `exprEvaluator` advertises (that path only runs for partition eval, never the record filter).

The real fix is upstream: drop the bbox predicate to `AlwaysTrue` before it reaches substrait, either in `getRecordFilter` or by special-casing `*boundBBoxPredicate` in `columnNameTranslator.VisitBound` (see my note there — same fix closes the all-rows-dropped case too). Even as a backstop here, I'd return `fmt.Errorf("%w: ...", ErrNotImplemented)` rather than a bare panic string so callers get something typed and wrappable. wdyt?

visitors.go
```diff
@@ -578,6 +604,13 @@ func (sanitizeVisitor) VisitBound(pred BoundPredicate) BooleanExpression {
 	// literal serializes without needing the field's type.
 	ref := Reference(pred.Term().Ref().Field().Name)
 	switch p := pred.(type) {
+	case *boundBBoxPredicate:
```
laskoviymishka: You guard this correctly here in `sanitizeVisitor`, but `columnNameTranslator.VisitBound` has the same shape of switch and doesn't special-case `*boundBBoxPredicate` — and its column-not-found early return fires before the `BoundUnaryPredicate` arm is ever reached.

So a geo-filtered scan against a data file that predates the geometry column hits the not-found path, returns `AlwaysFalse`, and silently drops every row instead of keeping them. That's the one direction pruning must never be wrong.

The same guard you wrote here (match `*boundBBoxPredicate` → `AlwaysTrue`) in `columnNameTranslator.VisitBound` fixes it, and because it also drops the predicate before substrait, it resolves the panic path from my other comment in one shot. More fundamentally, this is the cost of `boundBBoxPredicate` deliberately satisfying `BoundUnaryPredicate` — every type-switch on that interface now has to remember to match the concrete type first. Worth a hard look at whether that trick is paying for itself. wdyt?

table/internal/geo_codec.go
```diff
@@ -315,6 +315,34 @@ func decodeGeoBound(data []byte) (vals [geoNumDims]float64, layout geom.Layout,
...
+	hi, _, okHi := decodeGeoBound(upper)
+	if !okLo || !okHi {
+		return 0, 0, 0, 0, false
+	}
+	if math.IsNaN(lo[geoDimX]) || math.IsNaN(lo[geoDimY]) ||
+		math.IsNaN(hi[geoDimX]) || math.IsNaN(hi[geoDimY]) {
+		return 0, 0, 0, 0, false
+	}
+
+	return lo[geoDimX], lo[geoDimY], hi[geoDimX], hi[geoDimY], true
```
laskoviymishka, line 341: We reject NaN here but not inverted bounds. If a third-party writer emits `lo[X] > hi[X]`, `BBoxIntersectsXY` gets `aMaxX < aMinX`, the intersection test is always false, and we prune a file we should have kept — a dropped-rows false negative, same wrong direction as the schema-evolution case.

iceberg-go's own accumulator never produces inverted bounds, but this is a trust-on-read decode path, so the writer isn't ours to trust. I'd add `if lo[geoDimX] > hi[geoDimX] || lo[geoDimY] > hi[geoDimY] { return 0, 0, 0, 0, false }` right after the NaN check so a malformed bound just makes the file unprunable. wdyt?

table/evaluators_geo_test.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+var geoMetricsSchema = iceberg.NewSchema(0,
+	iceberg.NestedField{ID: 1, Name: "geom", Type: iceberg.GeometryType{}, Required: false},
+	iceberg.NestedField{ID: 2, Name: "geog", Type: iceberg.GeographyType{}, Required: false},
+)
+
+// TestInclusiveMetricsBBoxIntersects verifies that a data file is dropped only
+// when its geometry bounds cannot intersect the query bbox.
+func TestInclusiveMetricsBBoxIntersects(t *testing.T) {
```
laskoviymishka, line 50: These are solid unit tests for the metrics evaluator, but everything here calls `newInclusiveMetricsEvaluator` directly on a hand-built `mockDataFile`. Nothing drives the predicate through a real `table.Scan(WithFilter(BBoxIntersects(...)))` — binding, column-name translation, projection, rewrite — into the manifest→metrics path.

That gap is exactly where the two blockers I flagged live (the substrait panic and the `AlwaysFalse` drop both happen in translation, upstream of this evaluator). A scan-level test with two data files, one overlapping and one disjoint, asserting only the overlapping file lands in the scan tasks, would have caught both and would keep them caught. Could we add one?

visitors.go
```diff
@@ -60,6 +60,8 @@ type BoundBooleanExprVisitor[T any] interface {
 	VisitLess(BoundTerm, Literal) T
 	VisitStartsWith(BoundTerm, Literal) T
 	VisitNotStartsWith(BoundTerm, Literal) T
+	VisitBBoxIntersects(BoundTerm, BoundingBox) T
```
laskoviymishka, line 76: Adding two methods to this exported interface breaks every external implementation — Trino/Dremio adapters and downstream visitors won't compile until they add the geo methods. `iceberg` is v0 so semver doesn't strictly bind, but it's churn we're choosing.

Worth noting Java deliberately keeps geo out of `BoundExpressionVisitor` and dispatches it through standalone `GeospatialPredicateEvaluators` instead. Either we follow that and keep bbox eval outside the visitor via a type-assertion path, or we accept the break here — but I'd make it a deliberate call, and if we keep it in the visitor, either note it in the changelog or ship a `DefaultBoundBooleanExprVisitor` with no-op geo methods embedders can mix in. wdyt?

exprs.go
```diff
@@ -1132,3 +1143,159 @@ func rejectTransformTerm(term BoundTerm) error {
...
 }
+
+// BoundingBox is a planar (XY) query bounding box used by the geospatial
+// BBoxIntersects predicate. MinX/MinY are the lower-left corner and MaxX/MaxY
+// the upper-right corner (X is longitude/easting, Y is latitude/northing).
+//
+// The box is two-dimensional: pruning only compares the X and Y extents of a
+// geometry column's bounds, which is all the Iceberg spec requires for
+// bbox-based data skipping. Z/M extents in a column's bounds are ignored.
+type BoundingBox struct {
```
laskoviymishka, line 1160: This box is XY-only, which is fine for the pruning the spec requires — no correctness issue, inclusive eval just keeps more files than Java would for Z/M-bounded data. My concern is forward-compat: Java's `BoundingBox` carries optional Z and M, and once this struct ships in the public `iceberg` package, its `Valid()`/`Equals()` semantics are effectively frozen.

Since callers build these as literals, adding optional `*float64` `MinZ`/`MaxZ`/`MinM`/`MaxM` later is mostly additive, but the validity/equality contract would shift under existing callers. I'd at least decide now whether to reserve those fields so the door stays open. Not blocking — just cheaper to settle before it's public. wdyt?

exprs.go
```diff
@@ -1132,3 +1143,159 @@ func rejectTransformTerm(term BoundTerm) error {
...
+
+// BBoxIntersects constructs an unbound geospatial predicate that matches rows
+// whose geometry/geography value's bounding box intersects the query box. It is
+// used to prune data files whose stored geo bounds cannot overlap the query
+// region; the spec only requires bbox-based pruning, so full geometric predicate
+// evaluation (ST_Intersects/ST_Within) remains a query-engine concern.
+//
+// Panics if the term is nil or the bbox is not Valid (NaN coordinate or an
+// inverted min/max), either of which would cause silent mis-pruning.
+func BBoxIntersects(t UnboundTerm, bbox BoundingBox) UnboundPredicate {
```
laskoviymishka, line 1212: This is the outermost user-facing entry point, and a caller deriving a box from user input (a map viewport, a request param) can't guard construction without wrapping in `recover()`. I'd lean toward `BBoxIntersects(t, bbox) (UnboundPredicate, error)`.

`NewNot`/`NewAnd` panic too, so there's precedent for the style — if we keep the panic, I'd at least document the panic contract in the godoc so it isn't a surprise. wdyt?

expr_json.go
```diff
@@ -195,6 +195,14 @@ func (bsp *boundSetPredicate[T]) MarshalJSON() ([]byte, error) {
 	return marshalSetPredicate(bsp.op, bsp.term, bsp.lits.Members())
 }
 
+// Geospatial bbox predicates have no representation in the REST expression JSON
+// grammar; they are used only for local scan planning. Report an error rather
+// than silently emitting an empty object.
+func (u *unboundBBoxPredicate) MarshalJSON() ([]byte, error) { return nil, errBBoxNotSerializable }
+func (b *boundBBoxPredicate) MarshalJSON() ([]byte, error)   { return nil, errBBoxNotSerializable }
+
+var errBBoxNotSerializable = fmt.Errorf("%w: geospatial bbox predicates cannot be serialized to REST expression JSON", ErrInvalidArgument)
```
laskoviymishka: Small thing — the sentinel is declared after the methods that use it, which reads backward, and the repo keeps wrapped sentinel errors in `errors.go`. I'd move it there (or at least above these methods) to match convention.

expr_bbox_test.go
```diff
@@ -0,0 +1,179 @@
...
+		iceberg.BBoxIntersects(iceberg.Reference("geom"),
+			iceberg.BoundingBox{MinX: math.NaN(), MinY: 0, MaxX: 10, MaxY: 10})
+	})
+}
+
+func TestBBoxIntersectsBind(t *testing.T) {
+	sc := geoTestSchema()
+	bbox := iceberg.BoundingBox{MinX: 1, MinY: 2, MaxX: 3, MaxY: 4}
+
+	for _, name := range []string{"geom", "geog"} {
```
laskoviymishka, line 98: This loop doesn't wrap the body in `t.Run`, so a failure on `"geom"` aborts before `"geog"` runs — and the rest of the PR uses subtests. I'd wrap it in `t.Run(name, func(t *testing.T){...})` so both geo types are exercised independently.

visitors.go
```diff
@@ -385,6 +391,20 @@ func (e *exprEvaluator) VisitNotStartsWith(term BoundTerm, lit Literal) bool {
...
+// Row-level geometric evaluation (parsing each row's WKB and testing its
+// bounding box) is a query-engine concern - the Iceberg spec only requires
+// bbox-based data skipping, which happens in the metrics/manifest evaluators.
+// Here the predicate conservatively keeps every row so it never drops a row that
+// an engine's own ST_Intersects would retain.
+func (e *exprEvaluator) VisitBBoxIntersects(BoundTerm, BoundingBox) bool {
+	return true
+}
+
+func (e *exprEvaluator) VisitBBoxNotIntersects(BoundTerm, BoundingBox) bool {
```
laskoviymishka, line 436: `VisitBBoxIntersects` just above has a solid comment on why it keeps every row; `VisitBBoxNotIntersects` returns `true` with none. The `true` is correct (bounds can't prove a not-intersects), but it's exactly the kind of bare `return true` a future maintainer might "tidy" to `false` and start dropping rows. A one-line comment mirroring the one above would protect it.

laskoviymishka (changes requested):
The evaluator wiring here is careful: the conservative-keep reasoning across the inclusive/strict/manifest/bloom visitors is sound, the geography/antimeridian guard is the right call, and the sanitize-to-always-true handling of the unserializable predicate is well thought through. Nice job threading the spec's inclusive-pruning contract through the parts that are covered.

I'd hold this before merging though. The core thing is that `columnNameTranslator.VisitBound` doesn't special-case `*boundBBoxPredicate` the way you handled it in `sanitizeVisitor`, and that goes wrong in two directions. When the geo column is present, the predicate flows through as a `BoundUnaryPredicate` into substrait and panics — so `table.Scan(WithFilter(BBoxIntersects(...)))`, which reads like the primary use of this API, fails the whole scan rather than "conservatively keeping every row." When the column is absent (a file that predates the geometry column), the not-found branch returns `AlwaysFalse` and silently drops every row. The second one is the direction a pruning bug really can't go. The good news is that one guard — match `*boundBBoxPredicate` → `AlwaysTrue` in `columnNameTranslator.VisitBound` — closes both.

The reason both slipped through is that there's no end-to-end test: everything exercises `newInclusiveMetricsEvaluator` directly, nothing drives a real `table.Scan` with a bbox filter through translation into the metrics path. A scan-level test with one overlapping and one disjoint file would have caught both and would keep them caught.

A few things I'd want to settle in this PR before merge:
- special-case `*boundBBoxPredicate` in `columnNameTranslator.VisitBound` (fixes both the substrait panic and the all-rows-dropped case)
- reject inverted file-level bounds in `GeoBoundsXY` so a malformed third-party bound makes the file unprunable rather than silently pruned
- add a scan-level integration test proving a disjoint file is pruned and an overlapping one isn't
- make a deliberate call on adding the two methods to `BoundBooleanExprVisitor` (breaking for external implementers) — either follow Java and keep geo out of the visitor, or accept the break and note it

Everything else I left inline is smaller. Once these are addressed I'm happy to take another pass.

tanmayrauth:
@laskoviymishka Thanks for the thorough review — really useful pass. All four blockers plus every inline item are addressed.  
  
  Blockers
  - Column-name translation (the crux): columnNameTranslator.VisitBound now special-cases *boundBBoxPredicate → AlwaysTrue at the top, before both the FindColumnName not-found return and the switch. That  closes both directions in one guard — the substrait panic when the geo column is present, and the AlwaysFalse-drops-every-row case when the file predates the column.
  - Inverted bounds: GeoBoundsXY now rejects lo > hi on X/Y right after the NaN check, so a malformed third-party bound makes the file unprunable rather than silently pruned.
  - End-to-end test: added TestScanPrunesDisjointGeometryFile (real Scan(WithRowFilter(BBoxIntersects(...))) over an overlapping + a disjoint file, asserting only the overlapping one survives PlanFiles). Since the two blockers actually live in TranslateColumnNames/substrait on the record-read path (which PlanFiles doesn't reach), I also added TestBBoxTranslateColumnNames (present + absent column) and a substrait backstop test to pin those two directly — verified they fail if the VisitBound guard is reverted.
  - Visitor interface: made the deliberate call to follow Java — the two geo methods are removed from  BoundBooleanExprVisitor and moved to a new optional BoundGeospatialExprVisitor[T], dispatched via
  type-assertion in VisitBoundPredicate. No external break; internal visitors satisfy it for free, with var _ BoundGeospatialExprVisitor[…] = (*…)(nil) assertions keeping that compile-checked.
   

  Smaller items: substrait backstop now panics with a wrapped ErrNotImplemented (typed/wrappable, though unreachable now that bbox is dropped upstream); errBBoxNotSerializable moved to errors.go as exported ErrBBoxNotSerializable; the geom/geog test loop uses t.Run; VisitBBoxNotIntersects got a comment mirroring the intersects one; BoundingBox stays XY-only with the forward-compat contract documented (Z/M  additive later); and BBoxIntersects keeps the panic for consistency with the other predicate constructors, with the contract documented — happy to switch to an error return if you'd prefer.

errors.go
```diff
@@ -35,4 +35,8 @@ var (
 	ErrBadLiteral              = errors.New("invalid literal value")
 	ErrInvalidBinSerialization = errors.New("invalid binary serialization")
 	ErrResolve                 = errors.New("cannot resolve type")
+	// ErrBBoxNotSerializable is returned when marshaling a geospatial bbox
+	// predicate to REST expression JSON: bbox predicates exist only for local
+	// scan planning and have no representation in the REST expression grammar.
+	ErrBBoxNotSerializable = fmt.Errorf("%w: geospatial bbox predicates cannot be serialized to REST expression JSON", ErrInvalidArgument)
```
laskoviymishka: `ErrBBoxNotSerializable` wraps `ErrInvalidArgument`, which means a caller doing `errors.Is(err, ErrInvalidArgument)` on a marshal result now matches a "this type has no JSON form" case as if it were a bad user argument. It's really a not-implemented/contract condition, not an argument error.

I'd wrap `ErrNotImplemented` instead (or give it its own root sentinel) — consistent with how the substrait backstop already surfaces the same predicate. wdyt?

table/evaluators_geo_test.go
```diff
@@ -0,0 +1,256 @@
...
+	assert.True(t, got, "not-intersects cannot prune from bounds alone")
+}
+
+// TestScanPrunesDisjointGeometryFile drives a bbox filter through a real
+// table.Scan - binding, column-name translation, projection, and the
+// manifest->metrics path - rather than calling the metrics evaluator directly.
+// It proves the end-to-end contract: a data file whose geometry bounds are
+// disjoint from the query box is pruned, while an overlapping one survives. This
+// is the path where a bbox predicate would otherwise either panic in substrait
+// or be dropped to AlwaysFalse during column translation, so it guards both
```
laskoviymishka: Small thing on the comment: `PlanFiles` doesn't run `ReadTasks`/`arrowScan`, so this doesn't actually exercise column-name translation or substrait — the "would otherwise panic in substrait or be dropped to AlwaysFalse during column translation" regressions are pinned by `TestBBoxPredicateConvertsToTypedError` and `TestBBoxTranslateColumnNames`, not here.

What this test genuinely proves is the metrics-pruning path end-to-end, which is worth having on its own. I'd just retarget the comment to say that and point at the two unit tests for the translation/substrait regressions.

visitors.go
```diff
@@ -140,6 +155,23 @@ func VisitBoundPredicate[T any](e BoundPredicate, visitor BoundBooleanExprVisito
 		return visitor.VisitStartsWith(e.Term(), e.(BoundLiteralPredicate).Literal())
 	case OpNotStartsWith:
 		return visitor.VisitNotStartsWith(e.Term(), e.(BoundLiteralPredicate).Literal())
+	case OpBBoxIntersects, OpBBoxNotIntersects:
+		// Geospatial predicates are dispatched through the optional
+		// BoundGeospatialExprVisitor extension so visitors that predate (or don't
+		// care about) geo support are not forced to implement them. A visitor that
+		// doesn't implement it doesn't support geo predicates; panic here is
+		// recovered into an error by VisitExpr.
+		gv, ok := visitor.(BoundGeospatialExprVisitor[T])
```
laskoviymishka, line 170: The dispatch reads well, but this panic-when-the-visitor-doesn't-implement-the-extension is a new part of `VisitBoundPredicate`'s public contract, and the godoc says nothing about it. An external visitor that implements `BoundBooleanExprVisitor` without the geo methods and calls `VisitBoundPredicate` directly (not through `VisitExpr`) gets an unrecovered panic, and even the `VisitExpr` path returns an error the doc doesn't mention.

I'd add a line to the godoc: if the predicate is a bbox op and the visitor doesn't implement `BoundGeospatialExprVisitor[T]`, this panics with a wrapped `ErrNotImplemented`, which `VisitExpr` recovers into a returned error. Cheap to write, saves a downstream adapter a surprise on upgrade.

visitors.go
```diff
@@ -497,6 +552,18 @@ func (columnNameTranslator) VisitUnbound(pred UnboundPredicate) BooleanExpressio
...
 func (c columnNameTranslator) VisitBound(pred BoundPredicate) BooleanExpression {
+	// A bbox predicate has no substrait/record-filter form; it is evaluated only
+	// during metrics-based file pruning, so the record filter must conservatively
+	// keep every row. It must be matched before the column-not-found early return,
+	// which would otherwise return AlwaysFalse and silently drop every row of a
+	// file that predates the geo column, and before the switch below, whose
+	// default panics on an unhandled predicate. Collapse it to always-true
+	// (mirrors exprEvaluator and sanitizeVisitor); this also keeps it out of
+	// substrait, where it would otherwise error.
+	if _, ok := pred.(*boundBBoxPredicate); ok {
```
laskoviymishka: This guards the concrete `*boundBBoxPredicate`, but `BoundBBoxPredicate` is the exported interface — a downstream implementation or a test double satisfying it wouldn't match and would fall through to `FindColumnName` and the `default:` panic. Nothing external can construct one today (Bind always hands back the concrete type), so this is defensive rather than a live bug, but guarding on the interface (`if _, ok := pred.(BoundBBoxPredicate); ok`) is the same cost and future-proofs it.

Same shape in `sanitizeVisitor.VisitBound` — and there the guard sits after `ref := Reference(pred.Term()...)`, so hoisting it above `ref` would match this ordering. wdyt?

visitors_test.go
```diff
@@ -172,6 +172,18 @@ func (e *FooBoundExprVisitor) VisitNotStartsWith(iceberg.BoundTerm, iceberg.Lite
 	return e.visitHistory
 }
 
+func (e *FooBoundExprVisitor) VisitBBoxIntersects(iceberg.BoundTerm, iceberg.BoundingBox) []string {
```
laskoviymishka, line 175: `FooBoundExprVisitor` grew these two methods but nothing in the suite drives a bbox predicate through it, so the new type-assert-and-dispatch path in `VisitBoundPredicate` is uncovered — a refactor that broke the `BoundGeospatialExprVisitor[T]` assertion wouldn't be caught here.

I'd add a case binding a `BBoxIntersects` predicate and asserting the visit history contains `BBOX_INTERSECTS`, plus one negative case: a bare `BoundBooleanExprVisitor` without the geo methods, passed a bbox expr through `VisitExpr`, asserting `errors.Is(err, ErrNotImplemented)`. That second one pins the load-bearing error path external callers actually hit.

laskoviymishka (approved):
Nice, this addressed everything from last round

No blockers this round. What's left all trails from that new extension interface, and none of it changes behavior: `VisitBoundPredicate` now panics for a visitor that doesn't implement the extension, and that contract isn't in the godoc — worth a line, since an external adapter calling it directly (not via `VisitExpr`) gets an unrecovered panic. And the new dispatch has no direct test — I'd add one bbox case through `FooBoundExprVisitor` plus a negative case asserting `errors.Is(err, ErrNotImplemented)` for a visitor without the geo methods.

Smaller, and both judgment calls I left inline: I'd reconsider wrapping `ErrInvalidArgument` for `ErrBBoxNotSerializable` (a serialization caller doing `errors.Is` on it matches a non-argument error), and the `*boundBBoxPredicate` guard could match the exported `BoundBBoxPredicate` interface for robustness.

tanmayrauth:
@laskoviymishka Thanks,  I've addressed the non blocking items
  
  - ErrBBoxNotSerializable now wraps ErrNotImplemented instead of ErrInvalidArgument — it's a contract/not-implemented condition, consistent with the substrait backstop.
  - VisitBoundPredicate godoc now documents the contract: a bbox op with a visitor that doesn't implement BoundGeospatialExprVisitor[T] panics with a wrapped ErrNotImplemented, which VisitExpr recovers into a returned error (and a direct caller must implement the extension or recover).
  - Dispatch is now tested directly: a positive case driving BBOX_INTERSECTS through the extension visitor, plus a negative case with a BoundBooleanExprVisitor lacking the geo methods asserting errors.Is(err,  ErrNotImplemented).
  - Both guards (columnNameTranslator and sanitizeVisitor) now match the exported BoundBBoxPredicate interface rather than the concrete type, and the sanitizeVisitor guard is hoisted above the ref := line to mirror the other.
  - Retargeted the TestScanPrunesDisjointGeometryFile comment: it proves the metrics-pruning path end-to-end; the translation/substrait regressions are pinned by TestBBoxTranslateColumnNames and  TestBBoxPredicateConvertsToTypedError.

PR #1598, opened 2026-07-29
fix(table): accept EWKB-encoded geometry values when computing geo bounds

twuebi:
Iceberg prescribes ISO WKB, but geometry values with EWKB type codes (dimension flags in the high bits, an optional embedded SRID) exist in the wild (Trino 483), and the ISO-only decoder used for bounds failed with "wkb: unknown type: 2147483649", aborting the entire file rewrite.

Dispatch on the flag bits so both encodings decode. Coordinates are identical either way, so bounds are encoding-independent, and the stored bytes pass through unmodified.

This fixes the read/stats path, but it doesn't normalize the stored bytes, so a table compacted by iceberg-go still carries EWKB in its data files. Fully closing the gap is a write-path normalization follow-up

### DIFF
--- a/table/internal/geo_codec.go
+++ b/table/internal/geo_codec.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/apache/iceberg-go"
 	"github.com/twpayne/go-geom"
+	"github.com/twpayne/go-geom/encoding/ewkb"
 	"github.com/twpayne/go-geom/encoding/wkb"
 )
 
@@ -75,9 +76,9 @@ func newGeoBoundsAccumulator(isGeography bool) *geoBoundsAccumulator {
 }
 
 // AddWKB unmarshals a single WKB value and extends the bounding box with its
-// coordinates.
+// coordinates. Both ISO WKB and EWKB values are accepted (see decodeWKB).
 func (a *geoBoundsAccumulator) AddWKB(data []byte) error {
-	g, err := wkb.Unmarshal(data)
+	g, err := decodeWKB(data)
 	if err != nil {
 		return err
 	}
@@ -85,6 +86,57 @@ func (a *geoBoundsAccumulator) AddWKB(data []byte) error {
 	a.extend(g)
 
 	return nil
+}
+
+// EWKB dimension and SRID flags, carried in the high bits of the WKB type word.
+const (
+	ewkbFlagZ    = 0x80000000
+	ewkbFlagM    = 0x40000000
+	ewkbFlagSRID = 0x20000000
+	ewkbFlags    = ewkbFlagZ | ewkbFlagM | ewkbFlagSRID
+)
+
+// WKB byte-order markers, the first byte of every WKB value.
+const (
+	wkbBigEndian    = 0
+	wkbLittleEndian = 1
+)
+
+// decodeWKB decodes a geospatial value for statistics purposes, accepting both
+// encodings found in the wild. Iceberg prescribes ISO WKB, which encodes the
+// dimension in the type value itself (PointZ = 1001); some writers instead emit
+// EWKB, which flags Z/M in the high bits of the type word and may embed an SRID
+// after it. Both yield the same coordinates, so bounds do not depend on the
+// encoding. Stored values are never rewritten - this decode is read-only.
+func decodeWKB(data []byte) (geom.T, error) {
+	if isEWKB(data) {
+		return ewkb.Unmarshal(data)
+	}
+
+	return wkb.Unmarshal(data)
+}
+
+// isEWKB reports whether data's type word carries EWKB flags. The two decoders
+// are not interchangeable: the ISO decoder rejects flagged type words, and the
+// EWKB decoder rejects the ISO dimension offsets (it reads 1001 as an unknown
+// type rather than PointZ), so the flags select which one applies. Values too
+// short to hold a type word are left to the ISO decoder to reject.
+func isEWKB(data []byte) bool {
+	if len(data) < 5 {
+		return false
+	}
+
+	var typeWord uint32
+	switch data[0] {
+	case wkbBigEndian:
+		typeWord = binary.BigEndian.Uint32(data[1:5])
+	case wkbLittleEndian:
+		typeWord = binary.LittleEndian.Uint32(data[1:5])
+	default:
+		return false
+	}
+
+	return typeWord&ewkbFlags != 0
 }
 
 // extend walks every coordinate of g, recursing into geometry collections,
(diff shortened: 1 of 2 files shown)


table/internal/geo_codec_internal_test.go
```diff
@@ -237,6 +237,278 @@ func TestGeoBoundsAccumulatorInvalidWKB(t *testing.T) {
...
+// of the type word and optionally embeds an SRID after it.
+const (
+	wkbPoint       = 1
+	wkbLineString  = 2
+	wkbPointZ      = 1001
+	wkbPointM      = 2001
+	wkbPointZM     = 3001
+	wkbLineStringZ = 1002
+
+	ewkbTestZ    = 0x80000000
```
laskoviymishka: The test declares its own `ewkbTestZ/M/SRID` right next to production's `ewkbFlagZ/M/SRID` with identical values. Since this file is `package internal` it can reference the production constants directly — I'd drop the test copies and use `ewkbFlagZ|ewkbFlagM|ewkbFlagSRID`. As it stands, if someone ever corrects a flag value in geo_codec.go the tests stay green against the stale bits, so they'd be validating the wrong pattern.

Same story with the `0x01`/`0x00` byte-order markers in newWKBBuilder/newXDRWKBBuilder just below — those are `wkbLittleEndian`/`wkbBigEndian` in production. wdyt?

table/internal/geo_codec_internal_test.go
```diff
@@ -237,6 +237,278 @@ func TestGeoBoundsAccumulatorInvalidWKB(t *testing.T) {
...
+			isoLower, isoUpper := isoAcc.Bounds()
+
+			ewkbAcc := newGeoBoundsAccumulator(tt.geograph)
+			require.NoError(t, ewkbAcc.AddWKB(tt.ewkb))
+			ewkbLower, ewkbUpper := ewkbAcc.Bounds()
+
+			assert.Equal(t, isoLower, ewkbLower)
+			assert.Equal(t, isoUpper, ewkbUpper)
+			if tt.geograph {
+				assert.Nil(t, ewkbLower, "geography bounds must be omitted")
```
laskoviymishka, line 520: For the geography case, `assert.Equal(isoLower, ewkbLower)` is comparing nil to nil and this `assert.Nil` just confirms it stayed nil — so this passes even if `AddWKB` decoded nothing at all. The require.NoError above catches a hard decode error, but not a decode that silently succeeds without extending anything.

I'd assert the accumulator actually consumed the value (its internal geom/coordinate count moved past zero) after the add, so the case distinguishes "geography suppresses bounds" from "the value was quietly ignored." Otherwise it stays green if decodeWKB ever starts returning early.

table/internal/geo_codec_internal_test.go
```diff
@@ -237,6 +237,278 @@ func TestGeoBoundsAccumulatorInvalidWKB(t *testing.T) {
...
+		},
+		{
+			name:       "ewkb point z big endian",
+			wkb:        newXDRWKBBuilder(wkbPoint|ewkbTestZ).f64(1, 2, 3).bytes(),
+			wantLower:  []float64{1, 2, 3},
+			wantUpper:  []float64{1, 2, 3},
+			wantLength: 24,
+		},
+		{
+			name:       "ewkb linestring z with srid",
```
laskoviymishka, line 403: The table covers points and a linestring but not a GeometryCollection, which is exactly where the recursive path matters — `ewkb.Unmarshal` decodes each sub-geometry, and Trino/PostGIS emit these regularly. I'd add one case (EWKB collection with a Z flag) so a future go-geom bump can't silently regress nested decoding.

While you're there, worth pinning the mixed-encoding behavior: `isEWKB` only sniffs the outer type word, so an EWKB-flagged outer with ISO-encoded sub-geometries (or the reverse) hits the wrong decoder and errors. That's a fine failure mode — it aborts rather than corrupting bounds — but a case that documents "mixed encoding within a collection is unsupported, and errors" makes the boundary explicit.

table/internal/geo_codec.go
```diff
@@ -87,6 +88,57 @@ func (a *geoBoundsAccumulator) AddWKB(data []byte) error {
...
+// encoding. Stored values are never rewritten - this decode is read-only.
+func decodeWKB(data []byte) (geom.T, error) {
+	if isEWKB(data) {
+		return ewkb.Unmarshal(data)
+	}
+
+	return wkb.Unmarshal(data)
+}
+
+// isEWKB reports whether data's type word carries EWKB flags. The two decoders
```
laskoviymishka, line 119: The comment explains why the decoders aren't interchangeable for flagged and ISO-dimension values, but not the plain-2D case: a 2D EWKB value carries no flags, so `isEWKB` returns false and it goes to the ISO decoder — which is correct only because 2D EWKB is byte-identical to ISO WKB. That invariant is load-bearing and currently unstated. I'd add a sentence so nobody later "tightens" the heuristic and breaks the 2D path.

table/internal/geo_codec.go
```diff
@@ -87,6 +88,57 @@ func (a *geoBoundsAccumulator) AddWKB(data []byte) error {
 	return nil
 }
 
+// EWKB dimension and SRID flags, carried in the high bits of the WKB type word.
+const (
+	ewkbFlagZ    = 0x80000000
```
laskoviymishka: Minor, but I'd type these `uint32` explicitly (`const ewkbFlagZ uint32 = 0x80000000`). `0x80000000` is above MaxInt32, so today it only works because `typeWord` is uint32 and the untyped constant resolves against it — the moment one of these lands in a signed-int context it's a compile error on 32-bit. Typing them makes the intent obvious at the declaration too.

laskoviymishka (approved):
Nice, this is a clean fix — sniffing the EWKB flags and routing to the right decoder is the right call, and keeping the stored bytes untouched (read-only, stats-only) is exactly right given the rewrite path shouldn't be re-encoding geometry on the fly. The coordinate-equivalence tests between the two encodings are a good way to pin it.

The fix itself I'm happy with; my asks are all on the test side, plus one thing worth noting in the PR description.

The tests declare their own copies of the EWKB flag constants right next to the production ones — since the test is in package internal it can just use ewkbFlagZ/M/SRID directly, otherwise a future correction to those values leaves the tests silently green against stale bits. And the geography case passes vacuously today: comparing nil bounds to nil bounds doesn't prove the value was actually decoded, so I'd assert the accumulator really consumed it. I'd also add a GeometryCollection case, since that's the one path that exercises recursive sub-geometry decoding.

Worth a line in the PR description too: this fixes the read/stats path, but it doesn't normalize the stored bytes, so a table compacted by iceberg-go still carries EWKB in its data files. Fully closing the gap is a write-path normalization follow-up — just want it clear this is the read-side half, not the whole thing.

Tighten the tests and I'm happy to merge.

twuebi:
@laskoviymishka addressed the comments & amended the PR description as requested

laskoviymishka (approved):
🌬️  ⛵

PR #1601, opened 2026-07-29
fix(table): preserve branch/tag rentention settings on snapshot ref updates

### LINKED ISSUE #1600: Branch/Tag retention settings silently wiped out on every commit and rollback
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

When a snapshot ref is updated via a normal `Append/Overwrite` commit or via `RollbackToSnapshot`, the _snapshotRef_ is build from scratch and overwrites the existing entry.

It is because, the `commit/rollback` paths pass `0, 0, 0` for the retention params: `maxRefAgeMs`, `maxSnapshotAgeMs`, `minSnapshotsToKeep`, any retention policy previously configured on the `branch/tag` is discarded.

badalprasadsingh:
## Description

Fixes #1600, silent loss of _snapshot-ref_ retention settings (`min-snapshots-to-keep`, `max-snapshot-age-ms`, `max-ref-age-ms`) on every commit and rollback.

Have added the necessary tests for it.

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -908,23 +908,6 @@ func (b *MetadataBuilder) SetSnapshotRef(
 			return fmt.Errorf("%w: invalid snapshot ref option: %w", iceberg.ErrInvalidArgument, err)
 		}
 	}
-
-	// preserve retention properties from an existing ref when not explicitly overridden by caller
-	if existingRef, ok := b.refs[name]; ok && existingRef.SnapshotRefType == refType {
-		if ref.MinSnapshotsToKeep == nil && existingRef.MinSnapshotsToKeep != nil {
-			v := *existingRef.MinSnapshotsToKeep
-			ref.MinSnapshotsToKeep = &v
-		}
-		if ref.MaxSnapshotAgeMs == nil && existingRef.MaxSnapshotAgeMs != nil {
-			v := *existingRef.MaxSnapshotAgeMs
-			ref.MaxSnapshotAgeMs = &v
-		}
-		if ref.MaxRefAgeMs == nil && existingRef.MaxRefAgeMs != nil {
-			v := *existingRef.MaxRefAgeMs
-			ref.MaxRefAgeMs = &v
-		}
-	}
-
 	if err := ref.validate(); err != nil {
 		return fmt.Errorf("%w: invalid snapshot ref: %w", iceberg.ErrInvalidArgument, err)
 	}
@@ -978,6 +961,24 @@ func (b *MetadataBuilder) SetSnapshotRef(
 	b.refs[name] = ref
 
 	return nil
+}
+
+func (b *MetadataBuilder) NewRetainingSnapshotRefUpdate(name string, snapshotID int64, refType RefType) *setSnapshotRefUpdate {
+	var maxRefAgeMs, maxSnapshotAgeMs int64
+	var minSnapshotsToKeep int
+	if existing, ok := b.refs[name]; ok && existing.SnapshotRefType == refType {
+		if existing.MaxRefAgeMs != nil {
+			maxRefAgeMs = *existing.MaxRefAgeMs
+		}
+		if existing.MaxSnapshotAgeMs != nil {
+			maxSnapshotAgeMs = *existing.MaxSnapshotAgeMs
+		}
+		if existing.MinSnapshotsToKeep != nil {
+			minSnapshotsToKeep = *existing.MinSnapshotsToKeep
+		}
+	}
+
+	return NewSetSnapshotRefUpdate(name, snapshotID, refType, maxRefAgeMs, maxSnapshotAgeMs, minSnapshotsToKeep)
 }
 
 func (b *MetadataBuilder) RemoveSnapshotRef(name string) error {
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -908,6 +908,23 @@ func (b *MetadataBuilder) SetSnapshotRef(
 			return fmt.Errorf("%w: invalid snapshot ref option: %w", iceberg.ErrInvalidArgument, err)
 		}
 	}
+
+	// preserve retention properties from an existing ref when not explicitly overridden by caller
+	if existingRef, ok := b.refs[name]; ok && existingRef.SnapshotRefType == refType {
+		if ref.MinSnapshotsToKeep == nil && existingRef.MinSnapshotsToKeep != nil {
+			v := *existingRef.MinSnapshotsToKeep
+			ref.MinSnapshotsToKeep = &v
+		}
+		if ref.MaxSnapshotAgeMs == nil && existingRef.MaxSnapshotAgeMs != nil {
+			v := *existingRef.MaxSnapshotAgeMs
+			ref.MaxSnapshotAgeMs = &v
+		}
+		if ref.MaxRefAgeMs == nil && existingRef.MaxRefAgeMs != nil {
+			v := *existingRef.MaxRefAgeMs
+			ref.MaxRefAgeMs = &v
+		}
+	}
+
 	if err := ref.validate(); err != nil {
 		return fmt.Errorf("%w: invalid snapshot ref: %w", iceberg.ErrInvalidArgument, err)
 	}
--- a/table/metadata_builder_internal_test.go
+++ b/table/metadata_builder_internal_test.go
@@ -482,6 +482,120 @@ func TestSetRef(t *testing.T) {
 	require.Len(t, builder.snapshotLog, 1)
 }
 
+func TestSetSnapshotRefPreservesRetentionOnUpdate(t *testing.T) {
+	builder := builderWithoutChanges(2)
+	schemaID := 0
+	snapshot1 := Snapshot{
+		SnapshotID:       1,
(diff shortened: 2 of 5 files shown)


table/metadata.go
```diff
@@ -908,6 +908,23 @@ func (b *MetadataBuilder) SetSnapshotRef(
 			return fmt.Errorf("%w: invalid snapshot ref option: %w", iceberg.ErrInvalidArgument, err)
 		}
 	}
+
+	// preserve retention properties from an existing ref when not explicitly overridden by caller
+	if existingRef, ok := b.refs[name]; ok && existingRef.SnapshotRefType == refType {
```
laskoviymishka: I'd think about moving this inheritance up to the call sites rather than baking it into `SetSnapshotRef`.

Java keeps `setRef` a pure replace — it stores the `SnapshotRef` as-is — and does the retention carry-over one level up in `setBranchSnapshotInternal` via `SnapshotRef.builderFrom(ref, newSnapshotId)`. Doing it inside `SetSnapshotRef` also changes `setSnapshotRefUpdate.Apply()`: a set-snapshot-ref update that arrives with absent retention (which Java treats as "clear") now silently inherits whatever the current ref holds. For the embedded/Hadoop catalog replay path that's a real semantic divergence, and it's the same root reason there's now no way to clear a retention value once set — no options means inherit, and the `WithXxx` validators reject zero.

No current Go caller passes non-zero retention on an update-in-place, so this is latent design debt rather than a live bug. But if `snapshot_producers.go` and `transaction.go` read the existing ref's retention and pass it explicitly into `NewSetSnapshotRefUpdate`, the emitted update is fully specified, replay stays a pure replace, and clearing becomes representable again. wdyt?

table/metadata.go
```diff
@@ -908,6 +908,23 @@ func (b *MetadataBuilder) SetSnapshotRef(
...
+		if ref.MinSnapshotsToKeep == nil && existingRef.MinSnapshotsToKeep != nil {
```
laskoviymishka: While we're here — `metadata.go` already has `clonePtr[T]` a few hundred lines up that does exactly this nil-check-then-copy. Since `clonePtr(nil)` returns nil, you can drop the outer `!= nil` guards too:

```go
if ref.MinSnapshotsToKeep == nil {
    ref.MinSnapshotsToKeep = clonePtr(existingRef.MinSnapshotsToKeep)
}
```

table/metadata_builder_internal_test.go
```diff
@@ -482,6 +482,120 @@ func TestSetRef(t *testing.T) {
...
+	require.Equal(t, minKeep, *ref.MinSnapshotsToKeep)
+	require.NotNil(t, ref.MaxSnapshotAgeMs)
+	require.Equal(t, maxSnapAgeMs, *ref.MaxSnapshotAgeMs)
+	require.NotNil(t, ref.MaxRefAgeMs)
+	require.Equal(t, maxRefAgeMsIn, *ref.MaxRefAgeMs)
+
+	// Simulate a normal commit/rollback: only the snapshot pointer is updated,
+	// with no retention options. Retention must be preserved.
+	require.NoError(t, builder.AddSnapshot(&snapshot2))
+	require.NoError(t, builder.SetSnapshotRef(MainBranch, 2, BranchRef))
```
laskoviymishka: This test never exercises the path the bug actually lives on.

The regression happens because `snapshot_producers.go` and `transaction.go` build `NewSetSnapshotRefUpdate(branch, snap, BranchRef, 0, 0, 0)` and let `.Apply()` call `SetSnapshotRef` with no options — calling `builder.SetSnapshotRef` directly skips that translation entirely. If someone later changes how `Apply()` maps those zeros, this test stays green while the bug comes back.

I'd drive it through the real path: construct a `setSnapshotRefUpdate{MaxRefAgeMs: 0, ...}` on a builder whose `refs["main"]` already has retention, call `.Apply(builder)`, and assert it's preserved — plus a `RollbackToSnapshot` case for the rollback caller. That's what actually locks the contract.

table/metadata_builder_internal_test.go
```diff
@@ -482,6 +482,120 @@ func TestSetRef(t *testing.T) {
...
+		TimestampMs:      builder.base.LastUpdatedMillis() + 1,
+		ManifestList:     "/snap-1.avro",
+		Summary:          &Summary{Operation: OpAppend},
+		SchemaID:         &schemaID,
+	}
+	require.NoError(t, builder.AddSnapshot(&snapshot))
+	require.NoError(t, builder.SetSnapshotRef(MainBranch, 1, BranchRef, WithMinSnapshotsToKeep(5)))
+
+	// An explicitly provided option overrides the inherited value.
+	require.NoError(t, builder.SetSnapshotRef(MainBranch, 1, BranchRef, WithMinSnapshotsToKeep(10)))
```
laskoviymishka: This only proves that overriding the one field that was set works — it doesn't prove the invariant that matters most.

I'd set all three retention fields in the first call, override just `MinSnapshotsToKeep` in the second, and assert `MaxSnapshotAgeMs` and `MaxRefAgeMs` are still carried over. That's the "only fill when nil" contract; right now nothing fails if a future change accidentally drops the un-overridden fields.

table/metadata_builder_internal_test.go
```diff
@@ -482,6 +482,120 @@ func TestSetRef(t *testing.T) {
...
+	require.NoError(t, builder.AddSnapshot(&snapshot))
+	require.NoError(t, builder.SetSnapshotRef(MainBranch, 1, BranchRef, WithMinSnapshotsToKeep(5)))
+
+	// A tag with the same name must not inherit branch-only retention such as
+	// min-snapshots-to-keep, which tags do not support.
+	require.NoError(t, builder.SetSnapshotRef(MainBranch, 1, TagRef))
+	ref := builder.refs[MainBranch]
+	require.Equal(t, TagRef, ref.SnapshotRefType)
+	require.Nil(t, ref.MinSnapshotsToKeep)
+	require.Nil(t, ref.MaxSnapshotAgeMs)
```
laskoviymishka, line 605: Two gaps on the tag side here.

This asserts `MinSnapshotsToKeep` and `MaxSnapshotAgeMs` are nil after the cross-type switch but not `MaxRefAgeMs` — and `MaxRefAgeMs` is the one retention field valid on both branches and tags, so it's exactly the one a relaxed condition could leak. I'd add `require.Nil(t, ref.MaxRefAgeMs)` here.

There's also no positive tag→tag case: set `MaxRefAgeMs` on a tag, re-point it with `TagRef` and no options, assert it's preserved. Tags support that field, so it should inherit just like branches do.

laskoviymishka (changes requested):
Good catch on #1600! This is a real silent data-loss bug: branch and tag retention gets dropped on every commit and rollback. The fix is in the right area, and copying the fields by value avoids aliasing the existing ref.

I’d still hold the merge for the tests and one design question.

The regression test calls `builder.SetSnapshotRef` directly, but the bug is in the `.Apply()` translation used by commit and rollback producers. As written, the test can pass even if that path breaks again. I’d run one case through `.Apply()` and another through `RollbackToSnapshot`.

I’m also not sure the inheritance belongs inside `SetSnapshotRef`. Java keeps `setRef` as a replace operation and carries retention settings at the `setBranchSnapshotInternal` call site. Putting the inheritance in `SetSnapshotRef` changes replay behavior and makes it impossible to clear a retention value once it has been set.

No current Go caller seems to rely on clearing it, so this is not an immediate bug, but keeping the inheritance at the call sites looks cleaner to me and avoids that issue. Curious if there is a reason to keep it in `SetSnapshotRef`.

Before merge, I’d add:

* coverage through the actual `.Apply()` and rollback paths
* a partial override case showing that untouched retention fields are preserved
* tag coverage: `MaxRefAgeMs` should be nil on a cross-type switch, plus one positive tag-to-tag inheritance case

table/metadata.go
badalprasadsingh: Done.

You are right. Now, `NewRetainingSnapshotRefUpdate` is called both from `commitManifests` and `RollbackToSnapshot`, which reads the current ref's retention and passes it explicitly into `NewSetSnapshotRefUpdate`.

So the emitted update is fully specified, `Apply` stays a pure replace (absent retention = clear, matching Java's `setRef`), and clearing is representable again.

There was no reason to keep it in `SetSnapshotRef` other than fewer call-site touch-points.

table/metadata.go
badalprasadsingh: Well, this becomes no longer relevant with the refactor. Happy to use `clonePtr` if any pointer-copy reappears.

table/metadata_builder_internal_test.go
badalprasadsingh: Done.

table/metadata_builder_internal_test.go
badalprasadsingh: Done, this as well.

table/metadata_builder_internal_test.go
badalprasadsingh: Done. Both of them are covered now.

laskoviymishka (approved):

PR #1603, opened 2026-07-29
fix(table): mark transaction committed only after successful commit

### LINKED ISSUE #1602: Transaction Commit marks the transaction committed before the catalog accepts it, breaking commit retries
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Here, the transaction commit sets the internal `committed` flag to `true` before calling `doCommit`. If `doCommit` returns an error because of any reason : one possible reasons can be exhausted OCC retries, the transaction still retains `committed` as `true`.

This makes the `transaction` permanently unusable after any commit failure.

badalprasadsingh:
## Description

Fixes #1602, issue of setting `t.committed = true` before invoking `doCommit`.

Moved `t.committed = true` so it is set only after `doCommit` returns successfully. The transaction is left usable for retry. Add a test for it.

### DIFF
--- a/table/commit_retry_test.go
+++ b/table/commit_retry_test.go
@@ -944,3 +944,32 @@ func TestDoCommit_RetryProgressesFreshMeta(t *testing.T) {
 	require.NotContains(t, wfs.files, cat.observedManifestLists[1],
 		"attempt-1 rebuild manifest list must be cleaned as orphan after success")
 }
+
+func TestTransactionCommit_MarksCommittedOnlyOnSuccess(t *testing.T) {
+	// first CommitTable : fails (non-ErrCommitFailed, so doCommit returns immediately), the second succeeds
+	cat := &sequentialCatalog{
+		errs: []error{errors.New("catalog unavailable")},
+	}
+	tbl := newRetryTestTable(t, cat, nil)
+	cat.metadata = tbl.Metadata()
+
+	tx := tbl.NewTransaction()
+	require.NoError(t, tx.SetProperties(map[string]string{"key": "value"}))
+
+	_, err := tx.Commit(t.Context())
+	require.Error(t, err, "first commit must surface the catalog failure")
+	assert.False(t, tx.committed, "failed commit must leave committed == false")
+
+	// transaction stays usable: further changes and a retry are both allowed
+	require.NoError(t, tx.SetProperties(map[string]string{"key2": "value2"}),
+		"apply must be allowed after a failed commit")
+
+	committed, err := tx.Commit(t.Context())
+	require.NoError(t, err, "commit retry must be allowed after a transient failure")
+	require.NotNil(t, committed)
+	assert.True(t, tx.committed, "committed must be set only after a successful commit")
+
+	_, err = tx.Commit(t.Context())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "already been committed")
+}
--- a/table/commit_retry_test.go
+++ b/table/commit_retry_test.go
@@ -50,6 +50,7 @@ type sequentialCatalog struct {
 	loadMeta Metadata // optional: returned by LoadTable if non-nil
 	errs     []error
 	attempts atomic.Int32
+	lastReqs []Requirement
 }
 
 func (c *sequentialCatalog) LoadTable(_ context.Context, ident Identifier) (*Table, error) {
@@ -62,7 +63,8 @@ func (c *sequentialCatalog) LoadTable(_ context.Context, ident Identifier) (*Tab
 		func(context.Context) (iceio.IO, error) { return iceio.LocalFS{}, nil }, c), nil
 }
 
-func (c *sequentialCatalog) CommitTable(_ context.Context, _ Identifier, _ []Requirement, updates []Update) (Metadata, string, error) {
+func (c *sequentialCatalog) CommitTable(_ context.Context, _ Identifier, reqs []Requirement, updates []Update) (Metadata, string, error) {
+	c.lastReqs = reqs
 	n := int(c.attempts.Add(1)) - 1 // 0-indexed
 	if n < len(c.errs) && c.errs[n] != nil {
 		return nil, "", c.errs[n]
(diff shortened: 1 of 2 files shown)


table/transaction.go
```diff
@@ -2372,8 +2372,6 @@ func (t *Transaction) Commit(ctx context.Context) (*Table, error) {
 		return nil, errors.New("transaction has already been committed")
 	}
 
-	t.committed = true
-
 	if len(meta.updates) > 0 {
 		t.reqs = append(t.reqs, AssertTableUUID(meta.uuid))
```
laskoviymishka: This append runs on every `Commit()` call, and now that we intentionally allow a retry after a failed `doCommit`, a second `Commit()` tacks on another `AssertTableUUID` for the same UUID. `apply()` dedupes requirements via `requirementSemanticKey`, but this append bypasses `apply()` entirely, so nothing collapses them — the `reqs` slice grows by one on every retry.

Lenient catalogs shrug it off, but a strict REST catalog will reject the duplicate requirement, which defeats the retry this PR is trying to enable. I'd guard the append so it fires at most once (check for an existing `assertTableUUID`, or save `len(t.reqs)` before the append and truncate back on failure).

The new test won't catch this today — `sequentialCatalog.CommitTable` ignores `reqs` — so asserting the `reqs` contents after the retry would pin it. wdyt?

table/transaction.go
```diff
@@ -2384,6 +2382,8 @@ func (t *Transaction) Commit(ctx context.Context) (*Table, error) {
 			return tbl, err
```
laskoviymishka, line 2390: One thing this change surfaces: `doCommit` treats a clean 409 (`ErrCommitFailed`) differently from a 5xx/timeout, where the commit state is genuinely unknown — the catalog may have accepted it silently. Now that any failure here leaves `committed == false`, a caller who retries after an unknown-state error could double-apply the snapshot if the first attempt actually landed.

Java guards this by never retrying `CommitStateUnknownException`. I'd distinguish the failure class before allowing the retry — e.g. treat a non-`ErrCommitFailed` failure as terminal (still surface the error, but don't leave the transaction retriable). Happy to punt it to a follow-up if you'd rather keep this PR narrow, but I'd want it tracked. wdyt?

table/transaction.go
```diff
@@ -2384,6 +2382,8 @@ func (t *Transaction) Commit(ctx context.Context) (*Table, error) {
...
 		}
 
+		t.committed = true
```
laskoviymishka, line 2396: Setting `committed` here — after the catalog accepts but before `PostCommit` — is the right call: a `PostCommit` failure must not leave the transaction retriable, or we'd re-run the catalog commit. Worth a one-line comment saying exactly that, since the obvious "improvement" is to move it below the loop, which would quietly reintroduce the bug.

table/commit_retry_test.go
```diff
@@ -944,3 +944,32 @@ func TestDoCommit_RetryProgressesFreshMeta(t *testing.T) {
...
+
+	_, err := tx.Commit(t.Context())
+	require.Error(t, err, "first commit must surface the catalog failure")
+	assert.False(t, tx.committed, "failed commit must leave committed == false")
+
+	// transaction stays usable: further changes and a retry are both allowed
+	require.NoError(t, tx.SetProperties(map[string]string{"key2": "value2"}),
+		"apply must be allowed after a failed commit")
+
+	committed, err := tx.Commit(t.Context())
```
laskoviymishka, line 984: Could we assert `cat.attempts` went 1→2 here? Right now nothing proves the second `Commit` actually reached the catalog rather than short-circuiting — the other tests in this file pin that with `cat.attempts.Load()`.

While we're here, a sub-case where `doCommit` exhausts its internal retries on `ErrCommitFailed` (and still leaves `committed == false`) would cover the other failure path this fix touches.

table/commit_retry_test.go
```diff
@@ -944,3 +944,32 @@ func TestDoCommit_RetryProgressesFreshMeta(t *testing.T) {
...
+	require.NoError(t, err, "commit retry must be allowed after a transient failure")
+	require.NotNil(t, committed)
+	assert.True(t, tx.committed, "committed must be set only after a successful commit")
+
+	_, err = tx.Commit(t.Context())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "already been committed")
```
laskoviymishka: `assert.ErrorContains(t, err, "already been committed")` reads cleaner here and matches the rest of the file — it nil-checks internally, so we skip reaching through `err.Error()`.

laskoviymishka (changes requested):
Good catch on the root cause. Marking the transaction committed before `doCommit` ran was the bug, and moving that state change into the successful branches fixes #1602.

There is still one retry issue I’d fix before merging. After a failed commit, calling `Commit()` again appends another `AssertTableUUID` to `t.reqs`. That append does not go through `apply()` deduplication, so the requirements list grows with every retry. A strict REST catalog may reject the duplicate requirements.

This should be enough to guard the append so `AssertTableUUID` is added only once.

Two smaller points:

* Retrying after an unknown-state failure, such as a 5xx or timeout, may apply the snapshot twice if the first commit actually succeeded. We should either distinguish that case from a clean `409` here or track it separately.
* The new test should verify that the retry reaches the catalog and assert the contents of `t.reqs`, which would catch the accumulation above.

Fix the duplicate requirement on retry, and I’m good to take another look.

table/transaction.go
badalprasadsingh: Done. 

The `t.reqs` is left untouched — so a retry can't accumulate a second assertion.

table/transaction.go
badalprasadsingh: Done.

Handled in this PR rather than punting.

EDIT: On error, only a clean `ErrCommitFailed` leaves the transaction retriable. Covered by the test `TestTransactionCommit_TerminalOnUnknownState`.

table/transaction.go
badalprasadsingh: Done.

table/commit_retry_test.go
badalprasadsingh: Done.

1. `TestTransactionCommit_RetriableAfterCleanConflict` now asserts `cat.attempts` goes `1->2`, proving the retry reaches the catalog.
2. Added `TestTransactionCommit_RetriableAfterExhaustedInternalRetries` (numRetries=2, all _ErrCommitFailed_) which asserts 3 internal attempts and `committed == false` afterward.

table/commit_retry_test.go
badalprasadsingh: Done.

laskoviymishka (approved):

PR #1604, opened 2026-07-29
fix(io): make MemFS WalkDir emit directory entries

fallintoplace:
## Summary

- synthesize root and intermediate directory entries in `MemFS.WalkDir`
- emit entries in lexical order
- implement `SkipDir` and `SkipAll` behavior
- keep callbacks outside the filesystem lock

## Why

`ListableIO.WalkDir` requires callbacks for every file or directory, including the root. MemFS previously emitted only files in map iteration order, so directory-aware behavior and `SkipDir` differed from other filesystem implementations.

## Testing

- `go test ./io`

### DIFF
--- a/io/mem.go
+++ b/io/mem.go
@@ -131,6 +131,7 @@ func (m *MemFS) WalkDir(root string, fn fs.WalkDirFunc) error {
 
 		if key == root {
 			addEntry(walkEntry{path: key, size: int64(len(data))})
+
 			continue
 		}
 
@@ -260,6 +261,7 @@ func (fi *memFileInfo) Mode() fs.FileMode {
 	if fi.isDir {
 		return fs.ModeDir
 	}
+
 	return 0
 }
 func (fi *memFileInfo) ModTime() time.Time { return time.Time{} }
--- a/io/mem.go
+++ b/io/mem.go
@@ -24,7 +24,6 @@ import (
 	"io/fs"
 	"net/url"
 	"path"
-	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
@@ -77,7 +76,7 @@ func (m *MemFS) Open(name string) (File, error) {
 	cp := make([]byte, len(data))
 	copy(cp, data)
 
-	return &memFile{data: cp, name: filepath.Base(name)}, nil
+	return &memFile{data: cp, name: path.Base(name)}, nil
 }
 
 func (m *MemFS) Remove(name string) error {
@@ -120,7 +119,7 @@ func (m *MemFS) WalkDir(root string, fn fs.WalkDirFunc) error {
 	entriesByPath := make(map[string]walkEntry)
 	addEntry := func(entry walkEntry) {
 		current, exists := entriesByPath[entry.path]
-		if !exists || entry.isDir || !current.isDir {
+		if !exists || (!entry.isDir && current.isDir) {
 			entriesByPath[entry.path] = entry
 		}
 	}
@@ -152,14 +151,39 @@ func (m *MemFS) WalkDir(root string, fn fs.WalkDirFunc) error {
 	}
 	m.mu.RUnlock()
 	if len(entriesByPath) == 0 {
-		addEntry(walkEntry{path: root, isDir: true})
+		err := fn(root, nil, &fs.PathError{Op: "lstat", Path: root, Err: fs.ErrNotExist})
+		if err == fs.SkipDir || err == fs.SkipAll {
+			return nil
+		}
+
+		return err
+	}
+	if rootEntry, ok := entriesByPath[root]; ok && !rootEntry.isDir {
+		entriesByPath = map[string]walkEntry{root: rootEntry}
 	}
 
 	paths := make([]string, 0, len(entriesByPath))
 	for entryPath := range entriesByPath {
 		paths = append(paths, entryPath)
 	}
-	sort.Strings(paths)
+	sort.Slice(paths, func(i, j int) bool {
+		if paths[i] == root {
+			return true
+		}
+		if paths[j] == root {
+			return false
+		}
+
+		left := strings.Split(strings.TrimPrefix(paths[i], root+"/"), "/")
+		right := strings.Split(strings.TrimPrefix(paths[j], root+"/"), "/")
+		for index := 0; index < min(len(left), len(right)); index++ {
+			if left[index] != right[index] {
+				return left[index] < right[index]
+			}
+		}
+
+		return len(left) < len(right)
+	})
 
 	var skipPrefix string
 	for _, entryPath := range paths {
--- a/io/mem.go
+++ b/io/mem.go
@@ -23,7 +23,9 @@ import (
 	"io"
 	"io/fs"
 	"net/url"
+	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


io/mem.go
```diff
@@ -107,28 +109,78 @@ func (m *MemFS) WriteFile(name string, content []byte) error {
...
 	}
 
 	root = strings.TrimRight(root, "/")
 
 	m.mu.RLock()
-	var entries []walkEntry
+	entriesByPath := make(map[string]walkEntry)
+	addEntry := func(entry walkEntry) {
+		current, exists := entriesByPath[entry.path]
+		if !exists || entry.isDir || !current.isDir {
```
laskoviymishka: I think there's a real bug in this condition. `!exists || entry.isDir || !current.isDir` means a synthetic dir will happily overwrite a real file entry — the `entry.isDir` clause fires for any incoming directory regardless of what's already stored.

That bites when a key is both `root` and a prefix of other keys. Say `m.files` holds `mem://bucket/root` (a file) and `mem://bucket/root/a.txt`: the first key stores `{root, isDir:false}`, then walking the second calls `addEntry({root, isDir:true})`, which passes the `entry.isDir` clause and clobbers the file. Root gets reclassified as a directory and the real file entry is lost — `isTableDir`/`scanMetadataFiles` skip it via `d.IsDir()` and never see it.

I'd flip the precedence so real files always win over synthetic dirs — `!exists || (!entry.isDir && current.isDir)`. Same one-liner also fixes the broader-than-intent read that makes this condition hard to reason about. A regression test with root as a file key would lock it down.

io/mem.go
```diff
@@ -107,28 +109,78 @@ func (m *MemFS) WriteFile(name string, content []byte) error {
...
+			entry := walkEntry{path: entryPath, isDir: i < len(parts)-1}
+			if !entry.isDir {
+				entry.size = int64(len(data))
+			}
+			addEntry(entry)
+		}
 	}
 	m.mu.RUnlock()
+	if len(entriesByPath) == 0 {
+		addEntry(walkEntry{path: root, isDir: true})
```
laskoviymishka: This is where MemFS stops being a faithful stand-in for LocalFS. When nothing matches `root` — whether the dir is empty or simply doesn't exist — we synthesize a `{root, isDir:true}` entry and hand the callback a successful dir with a nil error.

`filepath.WalkDir` (what LocalFS delegates to) does the opposite: if the initial stat on root fails it calls `fn(root, nil, fs.ErrNotExist)`. So a test asserting the not-found path against MemFS silently passes where it'd correctly fail on a real FS. `hadoop.isTableDir` only comes out right here because its callback short-circuits on `path == metaDir`.

I'd have this branch return `fn(root, nil, &fs.PathError{Op: "lstat", Path: root, Err: fs.ErrNotExist})` and only synthesize the root dir when at least one key actually lives under it. If instead we want MemFS to not model directory existence at all, that's a defensible call — but then it's worth a doc comment saying so, because right now it reads like a bug. wdyt?

io/mem.go
```diff
@@ -107,28 +109,78 @@ func (m *MemFS) WriteFile(name string, content []byte) error {
...
+	}
 
-	for _, entry := range entries {
-		info := &memFileInfo{name: filepath.Base(entry.path), size: entry.size}
-		if err := fn(entry.path, fs.FileInfoToDirEntry(info), nil); err != nil {
+	paths := make([]string, 0, len(entriesByPath))
+	for entryPath := range entriesByPath {
+		paths = append(paths, entryPath)
+	}
+	sort.Strings(paths)
```
laskoviymishka: Sorting the full URI strings doesn't reproduce `fs.WalkDir`'s traversal order, and I think that's worth deciding on deliberately rather than by accident.

Because `.` (0x2E) sorts before `/` (0x2F), a sibling file can land between a directory and its children. With `root/a/child.txt` and `root/a.txt`, `sort.Strings` gives `root/a`, `root/a.txt`, `root/a/child.txt` — but real pre-order traversal visits `root/a`, `root/a/child.txt`, then `root/a.txt`, descending into `a/` before its sibling. The test name (`...EmitsDirectoriesLexically...`) reads like this is intentional, so I want to name the choice: any caller relying on parent-before-all-children ordering (the stdlib contract) would break, even if none in-tree do today.

If we're fine declaring MemFS emits in lexical order and not pre-order, let's say so in the doc comment. Otherwise a recursive descent over the synthesized tree (or a comparator that treats the separator as the primary key) gets us real pre-order. Which contract do we want here?

io/mem.go
```diff
@@ -107,28 +109,78 @@ func (m *MemFS) WriteFile(name string, content []byte) error {
...
+
+	var skipPrefix string
+	for _, entryPath := range paths {
+		if skipPrefix != "" && strings.HasPrefix(entryPath, skipPrefix) {
+			continue
+		}
+
+		entry := entriesByPath[entryPath]
+		info := &memFileInfo{name: path.Base(entry.path), size: entry.size, isDir: entry.isDir}
```
laskoviymishka, line 195: WalkDir now uses `path.Base` here, but `Open` still uses `filepath.Base` (line 83, just outside this diff). On Windows those disagree for `mem://` keys — `filepath.Base` splits on backslash and would return the whole string, so `Stat().Name()` differs depending on whether you reached the file via Open or WalkDir.

Since these are URI keys, `path.Base` is the right one everywhere. I'd switch line 83 too, and `path/filepath` looks unused after that so it can probably go.

io/mem_test.go
```diff
@@ -257,5 +262,32 @@ func TestMemIO_WalkDirCallbackCanRemoveFiles(t *testing.T) {
 
 		return nil
 	}))
-	assert.Empty(t, remaining)
+	assert.Equal(t, []string{"mem://bucket/root"}, remaining)
```
laskoviymishka: This assertion now leans entirely on the synthetic root entry — `mem://bucket/root` was never written as a file, so it shows up in `remaining` only because WalkDir emits a directory node for it. The test still covers the deadlock fix, but the assertion is now checking synthetic-dir emission rather than file-removal-during-walk, and it'd pass just the same if someone added `mem://bucket/root` as a real file.

If the synthetic-root behavior changes per the ErrNotExist discussion, this expectation moves with it — so it's worth re-grounding the assertion on the real files once that lands.

io/mem_test.go
```diff
@@ -257,5 +262,32 @@ func TestMemIO_WalkDirCallbackCanRemoveFiles(t *testing.T) {
...
+}
+
+func TestMemIO_WalkDirEmitsDirectoriesLexicallyAndHonorsSkipDir(t *testing.T) {
```
laskoviymishka, line 262: Nice to see the SkipDir case covered here. A few sibling cases would round it out: SkipAll from mid-walk (the `fs.SkipAll` arm isn't exercised at all), SkipDir when `path == root`, and — most important given the condition above — a root that's also a file key, which is exactly where the `addEntry` overwrite bites.

The nonexistent-root case is worth one too, since that's the LocalFS divergence I flagged. These are all a couple of lines each.

laskoviymishka (changes requested):
Thanks for tackling this — MemFS silently dropping directory entries was a real gap, and tree-reconstruction is the right approach. The SkipDir boundary handling looks correct.

I'd hold this before merging on one correctness bug and one fidelity issue.

The `addEntry` condition is inverted: `!exists || entry.isDir || !current.isDir` lets a synthetic directory overwrite a real file, so any key that's both the walk root and a prefix of other keys loses its file entry. It's a one-liner to flip (files win over synthetic dirs), but it needs a regression test with root as a file key.

The other one is that MemFS is used as a LocalFS stand-in in the catalog tests, so its behavior on a missing root matters. Today an empty-or-absent root emits a fake-success directory entry; real `filepath.WalkDir` calls back with `fs.ErrNotExist`. That means an error-path test can pass against MemFS and fail against a real FS. I'd return the ErrNotExist and only synthesize the root dir when something actually lives under it — or, if we deliberately don't want MemFS to model directory existence, say so in a doc comment.

One thing to decide rather than fix: `sort.Strings` on full URIs isn't stdlib pre-order (a sibling file can sort between a dir and its children). The test name reads like lexical order is intentional — if so, let's document it; if not, a recursive descent gets us real pre-order.

Rest is inline — a `path.Base`/`filepath.Base` inconsistency and a few test cases I'd add. Once the condition and the ErrNotExist behavior are sorted, happy to take another pass.

laskoviymishka (approved):

PR #1605, opened 2026-07-29
fix(hive): release acquired locks with a cleanup context

fallintoplace:
## Summary

- release acquired Hive locks with a live cleanup context
- bound cleanup with the existing five-second timeout
- return unlock failures alongside the operation result
- apply the cleanup path to drop, rename, and commit

## Why

These operations deferred `Release` with the caller's context. If that context was canceled after lock acquisition, Hive could reject the unlock immediately, and the error was discarded. This could leave other writers blocked until the lock was cleared externally.

## Testing

- `go test ./catalog/hive`

### DIFF
--- a/catalog/hive/hive.go
+++ b/catalog/hive/hive.go
@@ -417,7 +417,7 @@ func sqlFromVersion(v *view.Version) (string, error) {
 	return "", errors.New("view version has no SQL representation")
 }
 
-func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) error {
+func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) (err error) {
 	database, tableName, err := identifierToTableName(identifier)
 	if err != nil {
 		return err
@@ -428,7 +428,7 @@ func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) er
 		return fmt.Errorf("%w: failed to acquire lock for %s.%s: %w", table.ErrCommitFailed, database, tableName, err)
 	}
 	defer func() {
-		_ = lock.Release(ctx)
+		err = errors.Join(err, lock.releaseForCleanup(ctx))
 	}()
 
 	// Re-read after acquiring the lock so drop cannot act on table state that a
@@ -464,7 +464,7 @@ func (c *Catalog) PurgeTable(ctx context.Context, identifier table.Identifier) e
 	return nil
 }
 
-func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*table.Table, error) {
+func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (_ *table.Table, err error) {
 	fromDB, fromTable, err := identifierToTableName(from)
 	if err != nil {
 		return nil, err
@@ -502,7 +502,7 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
 			table.ErrCommitFailed, fromDB, fromTable, toDB, toTable, err)
 	}
 	defer func() {
-		_ = lock.Release(ctx)
+		err = errors.Join(err, lock.releaseForCleanup(ctx))
 	}()
 
 	hiveTbl, err := c.getIcebergTable(ctx, fromDB, fromTable)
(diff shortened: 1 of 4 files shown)


catalog/hive/hive.go
```diff
@@ -428,7 +428,7 @@ func (c *Catalog) DropTable(ctx context.Context, identifier table.Identifier) er
 		return fmt.Errorf("%w: failed to acquire lock for %s.%s: %w", table.ErrCommitFailed, database, tableName, err)
 	}
 	defer func() {
-		_ = lock.Release(ctx)
+		err = errors.Join(err, lock.releaseForCleanup(ctx))
```
laskoviymishka: Same success-path issue as the CommitTable defer below — if the drop lands but `releaseForCleanup` fails, this overwrites the nil `err`, so the caller sees a failure for a table that's already gone and a retry hits `ErrNoSuchTable`. I'd only join the release error when `err` is already non-nil, and log-and-drop it otherwise.

catalog/hive/hive.go
```diff
@@ -502,7 +502,7 @@ func (c *Catalog) RenameTable(ctx context.Context, from, to table.Identifier) (*
 			table.ErrCommitFailed, fromDB, fromTable, toDB, toTable, err)
 	}
 	defer func() {
-		_ = lock.Release(ctx)
+		err = errors.Join(err, lock.releaseForCleanup(ctx))
```
laskoviymishka: Same issue, slightly worse here: the success path ends `return c.LoadTable(ctx, to)`, so a landed rename with a failed unlock hands the caller a valid `*table.Table` and a non-nil error together, which breaks the usual (value, error) contract. Same conditional-join guard fixes it.

catalog/hive/hive.go
```diff
@@ -548,7 +548,7 @@ func (c *Catalog) CommitTable(ctx context.Context, identifier table.Identifier,
 			table.ErrCommitFailed, database, tableName, err)
 	}
 	defer func() {
-		_ = lock.Release(ctx)
+		err = errors.Join(err, lock.releaseForCleanup(ctx))
```
laskoviymishka: This is the one I'd hold the PR on. Joining the release error into `err` unconditionally makes the CommitTable success path worse than the bug it fixes.

When the commit lands we `return staged.Metadata(), staged.MetadataLocation(), nil`, then this defer overwrites `err` with the release error. doCommit in `table/table.go` then sees a non-nil err that doesn't wrap `ErrCommitFailed`, takes the unknown-state branch (`cleanupOrphans = false; return nil, err`), and hands the caller `(nil, unlockErr)` for a commit that's durably in the metastore — so it retries or aborts a write that already succeeded. Wrapping in `ErrCommitFailed` isn't the fix either; that'd retry a landed commit.

I'd surface it out-of-band — only join when the op already failed, otherwise log:

```go
defer func() {
    if releaseErr := lock.releaseForCleanup(ctx); releaseErr != nil {
        if err != nil {
            err = errors.Join(err, releaseErr)
        } else {
            log.Printf("hive: lock release failed after a successful commit: %v", releaseErr)
        }
    }
}()
```

Same guard on the DropTable and RenameTable defers.

catalog/hive/lock_test.go
```diff
@@ -323,6 +323,35 @@ func TestReleaseLock(t *testing.T) {
 	mockClient.AssertExpectations(t)
 }
 
+func TestReleaseLockForCleanupUsesLiveBoundedContext(t *testing.T) {
```
laskoviymishka, line 326: These two tests only cover `releaseForCleanup` in isolation — neither goes through the DropTable/RenameTable/CommitTable defers, so nothing here catches the success-path regression above. A test that drives CommitTable with Unlock failing after a successful AlterTable, then asserts we get the committed metadata back rather than a nil-metadata error, would fail today and lock in the fix.

laskoviymishka (changes requested):
The core of this is right — detaching the cleanup context so unlock always gets a live, bounded window is the correct fix, and I'm glad the silent `_ = lock.Release(ctx)` swallow is gone.

One thing before merge: joining the release error into `err` on the success path makes CommitTable worse than the bug it fixes. When the commit lands but unlock fails, the defer overwrites the nil error, doCommit takes the unknown-state branch (the error doesn't wrap `ErrCommitFailed`), and the caller gets `(nil, unlockErr)` for a commit that's durably in the metastore — so it retries or aborts a write that already succeeded. DropTable and RenameTable have the same shape (RenameTable can even hand back a valid table and a non-nil error together).

I'd only join the release error when the op already failed and log-and-drop it otherwise — wrapping it in `ErrCommitFailed` would be wrong since that retries a landed commit. Left a note inline on a test that would catch this.

Fix the success-path join and this is good to land.

laskoviymishka (approved):

PR #1606, opened 2026-07-29
fix(rest): close internally owned transports

fallintoplace:
## Summary

Close HTTP and OAuth transports allocated internally by the REST catalog.

## Why

Initialization built cleanup callbacks for internal transports but discarded the callback for the final long-lived session. Catalog.Close could therefore leave idle connection pools open.

## What changed

Catalog.Close now closes catalog-owned transports and keeps caller-provided transports caller-owned. Session cleanup is idempotent.

## Tests

- go test ./catalog/rest

### DIFF
--- a/catalog/rest/options.go
+++ b/catalog/rest/options.go
@@ -28,6 +28,11 @@ import (
 
 type Option func(*options)
 
+// TransportFactory creates a transport owned by the REST catalog and returns
+// the cleanup function to run when the catalog is closed. The TLS config is
+// the one configured with WithTLSConfig.
+type TransportFactory func(*tls.Config) (http.RoundTripper, func())
+
 func WithCredential(cred string) Option {
 	return func(o *options) {
 		o.credential = cred
@@ -162,6 +167,15 @@ func WithCustomTransport(transport http.RoundTripper) Option {
 	}
 }
 
+// WithTransportFactory configures a catalog-owned transport factory. The
+// returned cleanup function is called when the catalog is closed. Use
+// WithCustomTransport when the caller owns the transport instead.
+func WithTransportFactory(factory TransportFactory) Option {
+	return func(o *options) {
+		o.transportFactory = factory
+	}
+}
+
 type options struct {
 	awsConfig         aws.Config
 	awsConfigSet      bool
@@ -180,7 +194,7 @@ type options struct {
 	audience          string
 	resource          string
 	transport         http.RoundTripper
-	transportFactory  func(*tls.Config) (http.RoundTripper, func())
+	transportFactory  TransportFactory
 	headers           map[string]string
 
 	oauthTLSConfig *tls.Config
--- a/catalog/rest/options.go
+++ b/catalog/rest/options.go
@@ -180,6 +180,7 @@ type options struct {
 	audience          string
 	resource          string
 	transport         http.RoundTripper
+	transportFactory  func(*tls.Config) (http.RoundTripper, func())
 	headers           map[string]string
 
 	oauthTLSConfig *tls.Config
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -1078,7 +1078,8 @@ func (r *Catalog) fetchConfig(ctx context.Context, opts *options) (*options, err
 func (r *Catalog) Name() string              { return r.name }
 func (r *Catalog) CatalogType() catalog.Type { return catalog.REST }
 
-// Close releases transports created by the catalog and its metrics reporter.
+// Close drains idle connections from transports created by the catalog and
+// closes its metrics reporter.
 // Caller-provided transports remain caller-owned. Close is safe to call more
 // than once. Callers holding a [catalog.Catalog] can reach this via a
 // [catalog.Closer] type assertion.
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -34,6 +34,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/apache/iceberg-go"
@@ -766,6 +767,10 @@ var _ catalog.PurgeableTable = (*Catalog)(nil)
 type Catalog struct {
 	baseURI *url.URL
 	cl      *http.Client
+	// closeSession releases transports owned by the catalog. Caller-provided
+	// transports are excluded when the session is constructed.
+	closeSession     func()
+	closeSessionOnce sync.Once
 
 	name string
 	// Retained catalog properties are reused for table/view IO and may carry
@@ -914,7 +919,7 @@ func (r *Catalog) init(ctx context.Context, ops *options, uri string) error {
 		return err
 	}
 
-	r.cl, _, err = r.createSession(ctx, ops)
+	r.cl, r.closeSession, err = r.createSession(ctx, ops)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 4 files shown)


catalog/rest/rest_internal_test.go
```diff
@@ -1696,6 +1696,15 @@ func (t *closeTrackingTransport) CloseIdleConnections() {
 	t.closed.Store(true)
 }
 
+func TestCatalogCloseReleasesOwnedSessionOnce(t *testing.T) {
```
laskoviymishka, line 1699: This test doesn't actually exercise the fix.

It constructs a `Catalog` literal with `closeSession` already populated, so all it proves is that the `sync.Once` wrapper fires the callback once. The one line that fixes the leak — `r.cl, r.closeSession, err = r.createSession(ctx, ops)` in `init` — is entirely outside its scope. Revert that back to `r.cl, _, err = ...` and this test still passes green, which is exactly the regression we'd want it to catch.

I'd add an integration-level test that builds a real catalog through `NewCatalog`/`newCatalogFromProps` with an internally-owned transport, calls `cat.Close()`, and asserts that transport's `CloseIdleConnections` ran — `closeTrackingTransport` is already right there. And while we're in here, it'd be worth mirroring `TestFetchConfigDoesNotCloseCustomTransport` for `Close()`: build a catalog `WithCustomTransport(base)`, close it, assert `base.closed` stays false, so the caller-ownership boundary is pinned too. Right now that invariant is only documented, not tested.

catalog/rest/rest.go
```diff
@@ -1066,11 +1071,18 @@ func (r *Catalog) fetchConfig(ctx context.Context, opts *options) (*options, err
...
+// Caller-provided transports remain caller-owned. Callers holding a
+// [catalog.Catalog] can reach this via a [catalog.Closer] type assertion.
+func (r *Catalog) Close() error {
+	r.closeSessionOnce.Do(func() {
+		if r.closeSession != nil {
+			r.closeSession()
+		}
+	})
+
+	return r.reporter.Close()
```
laskoviymishka: `reporter.Close()` sits outside the `Once`, so it fires on every `Close()` call while the session cleanup fires once.

It's not a live bug — `CachedReporter.Close()` is idempotent today (it nils itself under a mutex, so the second call is a no-op). But the new test advertises `Close()` as safe to call twice, and that guarantee currently leans on an undocumented property of the concrete reporter; a future non-idempotent or wrapped reporter breaks double-close callers quietly.

I'd pull `reporter.Close()` inside the `Once` so the whole method is idempotent as one unit, and capture the error:

```go
func (r *Catalog) Close() error {
	var err error
	r.closeSessionOnce.Do(func() {
		if r.closeSession != nil {
			r.closeSession()
		}
		err = r.reporter.Close()
	})

	return err
}
```

The godoc could also mention `Close` is safe to call more than once, since that's now the intent (`closeSessionOnce`). wdyt?

catalog/rest/rest_internal_test.go
```diff
@@ -1696,6 +1696,15 @@ func (t *closeTrackingTransport) CloseIdleConnections() {
...
+	var calls atomic.Int32
```
laskoviymishka: Every other top-level test in this file calls `t.Parallel()` and there's no shared state forcing this one to be serial — I'd add it as the first line to stay consistent (and keep `paralleltest` happy if it's enabled).

laskoviymishka (changes requested):
Nice catch on the leak — throwing away `createSession`'s cleanup with `_` in `init` meant the long-lived session's internally-owned transports never got drained, and routing that cleanup through `Close()` behind a `sync.Once` (while leaving caller-provided transports alone) is exactly the right shape.

Almost there. One blocking thing before merge: the new test doesn't actually exercise the fix. It builds a `Catalog` literal with `closeSession` already set, so it only proves the `sync.Once` wrapper fires once — the line that fixes the bug (`r.cl, r.closeSession, err = r.createSession(...)`) is outside its scope, and reverting that line keeps the test green. For a regression test on a resource leak, I'd want it to fail if the wiring regresses, so I'd build a real catalog through the constructor with an internally-owned transport and assert `CloseIdleConnections` actually ran.

A couple of smaller things I'd want alongside that:

- `reporter.Close()` sits outside the `Once`, so it fires on every call. Not a bug today since `CachedReporter` is idempotent, but I'd fold it into the `Once` so the whole method is idempotent as one unit, and note that in the godoc.
- The `catalog.Closer` godoc in `catalog/catalog.go` still calls a stateful HTTP-backed closer a "future" thing — this PR makes it present, so that line is now stale.

Once the test actually pins the fix, happy to take another pass and approve.

laskoviymishka (approved):

catalog/rest/options.go
```diff
@@ -180,6 +180,7 @@ type options struct {
 	audience          string
 	resource          string
 	transport         http.RoundTripper
+	transportFactory  func(*tls.Config) (http.RoundTripper, func())
```
zeroshade: Non-blocking: `transportFactory` has no production writer and no exported option, so this leaves a permanently dead production branch whose only current role is test observation. Consider either exposing it as a supported injection point or keeping the hook test-local through a design that does not add unreachable production state.

catalog/rest/rest_internal_test.go
```diff
@@ -1696,6 +1696,60 @@ func (t *closeTrackingTransport) CloseIdleConnections() {
...
+	mux.HandleFunc("/v1/config", func(w http.ResponseWriter, _ *http.Request) {
+		json.NewEncoder(w).Encode(map[string]any{
+			"defaults":  map[string]any{},
+			"overrides": map[string]any{},
+		})
+	})
+
+	var transports []*closeTrackingTransport
+	withTrackingTransport := func(o *options) {
+		o.transportFactory = func(*tls.Config) (http.RoundTripper, func()) {
```
zeroshade: Non-blocking: this factory sends the test through the new test-only `else if`, so it never asserts cleanup of the real `*http.Transport` created by the final `else` branch. Consider observing that shipping branch directly; the bootstrap/session lifetime checks here are otherwise well targeted.

catalog/rest/rest.go
```diff
@@ -1066,11 +1078,20 @@ func (r *Catalog) fetchConfig(ctx context.Context, opts *options) (*options, err
 func (r *Catalog) Name() string              { return r.name }
 func (r *Catalog) CatalogType() catalog.Type { return catalog.REST }
 
-// Close releases the catalog's metrics reporter. The REST catalog does not own
-// the lifetime of the HTTP client it was configured with, so only the reporter
-// is released. Callers holding a [catalog.Catalog] can reach this via a
+// Close releases transports created by the catalog and its metrics reporter.
```
zeroshade: Consider describing this as draining idle connections rather than releasing transports. `CloseIdleConnections` does not destroy the transport, and the catalog client remains usable after `Close`, so the current wording implies a stronger lifecycle transition than the implementation provides.

zeroshade (commented):
This change is correct: I verified that ownership is structurally enforced. Cleanup is registered only when the catalog creates the transport; `WithCustomTransport` takes the separate caller-owned branch, and there is no option to inject an `http.Client`, so `Close` cannot reach a caller's transport or client. The `sync.Once` handling also makes repeated and concurrent calls safe.

The main non-blocking nit is that `transportFactory` is production state with no production writer or exported option, while the test uses it to bypass the actual `*http.Transport` branch under review. Consider either making the factory an intentional option or observing the real transport path, as noted inline.

Worth follow-up coverage for the OAuth transport cleanup at `catalog/rest/rest.go:883`, which becomes reachable from `Close` through this change. A concurrent `Close` case alongside `catalog/rest/rest_internal_test.go:1699-1730`, run under the race detector, would also pin the synchronization behavior.

Non-blocking coordination note: #1614 also edits the `Close`/`checkValidNamespace` area in `catalog/rest/rest.go` (here around lines 1081-1103), so the two PRs will need a small textual rebase when one lands.

PR #1206, opened 2026-06-15
feat(table): reject Variant in IdentityTransform and bound literal predicates

nssalian:
First part of [#1016](https://github.com/apache/iceberg-go/issues/1016). 

## Changes

- Adds `VariantType` to `IdentityTransform`'s deny-list (matches Java's `Identity.UNSUPPORTED_TYPES`), the other transforms already reject via their allow-list defaults. Adds explicit `case VariantType` in `createBoundLiteralPredicate` returning `ErrInvalidArgument` with a clear message.

- Added a test `TestVariantBoundLiteralRejectionMessage`

This is a precursor to adding the `variant_get` expression and stats-based pruning of #1016. Will address those after the shredding fixes.

PR #1607, opened 2026-07-29
feat(table): variant predicate pushdown and reject variant partition source

### LINKED ISSUE #1016: feat(table): predicate evaluation and partition-transform behavior on variant fields
Parent: #589

Tail of the Variant track. Once non-shredded support (#929 / PR #932) and shredded reader/writer (#986 / #987) land, two pieces remain:

- **Predicate binding.** Predicates over a variant column need a `variant_get(col, '$.path')` projection at the binder so a typed predicate lowers to a column comparison the scanner can evaluate. On a shredded variant the binder should additionally use the typed-value subtree's stats for pruning.
- **Partition-spec posture.** Per spec, variant fields cannot appear directly in partition specs (same posture as `UnknownType`). The schema visitor should reject any partition field whose source column is a variant. Future variant-aware transforms (e.g. hash over a fixed `variant_get` path) are not standardized yet — same parking posture as multi-arg transforms.

Stub issue so the gap has a tracker; concrete design lands once the shredded reader/writer pin the typed-value plumbing.

nssalian:
Adds predicate pushdown for shredded variant fields and rejects Variant as a partition
source. Closes [#1016](https://github.com/apache/iceberg-go/issues/1016).

## Changes

1. **Extract expressions** - `extract(variant, path, type)` referencing a field inside a variant column. RFC-9535 path parse, including bracket notation and `\u` escapes with UTF-16 surrogate pairs (`variant_path.go`); bind against the shredded schema (`variant_extract.go`); cast to the target type, gated on tz-awareness and guarded against int64 overflow (`variant_cast.go`). Follows Java `Expressions.extract` / `PathUtil.parse` / `VariantExpressionUtil.castTo`. Extract terms have no REST expression-JSON form, so they marshal to an explicit `ErrExtractNotSerializable` rather than an empty object.

2. **File pruning** - the inclusive/strict metrics evaluators decode a shredded field's stored variant bounds for an extract term and prune at the file level (`table/evaluators.go`, `table/internal/variant_bounds.go`, bounds split via arrow-go `Metadata.SizeBytes()`); the strict path never prunes on an extract. Extract terms carry no row-group statistics, so they are removed from the row-group stats/bloom filter (`stripExtractPredicates` in `table/arrow_filter_plan.go`); a `NOT` over an extract collapses to `AlwaysTrue` so negation can never fold the pruning filter to a lossy `AlwaysFalse`. Follows Java `InclusiveMetricsEvaluator` / `StrictMetricsEvaluator`.

3. **Residual filtering** - surviving extract terms become synthetic derived columns; the predicate is then evaluated columnar-ly through the existing arrow-go `compute` record-filter path rather than a per-row expression walk (`scanTranslator` in `visitors.go`, `table/variant_residual.go`, `table/arrow_scanner.go`). The derived column itself is materialized with a per-row pass over the variant array (`buildExtractColumn`); a missing variant column or non-`VariantArray` returns an error, and an undecodable row is logged and treated as null.

4. **Reject Variant partition source** - `addSpecFieldInternal` (`partitions.go`) rejects a Variant source for any transform. Builds on [#1206](https://github.com/apache/iceberg-go/pull/1206) (which added Variant to `IdentityTransform`'s deny-list); this path does not consult `CanTransform`, so it closes the gap for sources added via `AddPartitionFieldBySourceID` and for `VoidTransform` (whose `CanTransform` returns true unconditionally). `AddPartitionFieldByName`, `AddPartitionFieldBySourceID`, and the `BindToSchema` replay all route through it.

## Testing

- Unit tests: path parse/normalize incl. escapes, bracket, surrogate pairs, and rejects (`variant_path_test.go`); cast table incl. tz/zoneless, scale, and int64-overflow cases (`variant_cast_test.go`); extract bind incl. the required-column no-fast-path case (`variant_extract_test.go`); residual eval and `appendExtractLiteral` (`table/variant_residual_test.go`); metrics pruning and JSON not-serializable (`table/evaluators_test.go`, `expr_json_test.go`); partition-source rejection (`partitions_test.go`).
- End-to-end through the public `Scan().WithRowFilter(...)` API, incl. `IsNull`/`NotNull`/`Eq` over an extract, a heterogeneous residual column, and a negated-extract case that must not lossily prune (`table/variant_shredded_test.go`); shredded-write + file-pruning matrix and name-mapping field-id fallback (`table/variant_shredded_write_test.go`).


variant_cast.go
```diff
@@ -0,0 +1,269 @@
...
+		if b, ok := raw.([]byte); ok && len(b) == t.Len() {
+			return b, true
+		}
+	case DecimalType:
+		return castVariantDecimal(raw, t)
+	case BooleanType:
+		if b, ok := raw.(bool); ok {
+			return b, true
+		}
+	case TimestampType, TimestampTzType:
```
zeroshade: **Blocking: this conflates tz-aware and zoneless timestamps, and it produces wrong rows rather than just wrong pruning.**

`exactVariantMatch` (`variant_cast.go:62-84`) gets this right — `TimestampMicrosNTZ` -> `timestamp`, `TimestampMicros` -> `timestamptz`. But the fallback dispatch here drops the distinction entirely: both `TimestampType` and `TimestampTzType` route to the same `castVariantToMicros`, and both `TimestampNsType` and `TimestampTzNsType` route to the same `castVariantToNanos`. `castVariantToMicros` (`:195-227`) then accepts both `variant.TimestampNanos` *and* `variant.TimestampNanosNTZ` regardless of which target was requested.

Net effect: `Extract(p, "$.ts", TimestampTz)` silently matches a zoneless nanosecond leaf, treating local time as UTC, and `Extract(p, "$.ts", Timestamp)` silently matches a tz-aware nanosecond leaf. Since this feeds both file pruning and the residual filter, the rows that come back are wrong. The inconsistency with `exactVariantMatch` reads as an oversight rather than a decision.

Suggested fix: pass the target's tz-awareness into the three helpers and require the source physical type to agree.

```go
func castVariantToMicros(pt variant.Type, raw any, tz bool) (any, bool) {
    switch {
    case tz && pt == variant.TimestampNanos, !tz && pt == variant.TimestampNanosNTZ:
        return Timestamp(floorDiv(int64(raw.(arrow.Timestamp)), nanosPerMicro)), true
    ...
```

table/variant_residual.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return nil, arrow.Field{}, err
+	}
+
+	bldr := array.NewBuilder(mem, dt)
+	defer bldr.Release()
+
+	n := int(rec.NumRows())
+	varIdx := fieldIndexByID(rec.Schema(), col.Term.Ref().Field().ID)
+	varr, _ := columnAt(rec, varIdx).(*extensions.VariantArray)
```
zeroshade: **Blocking: the discarded `ok` turns a lookup miss into a silently emptied batch.**

If `fieldIndexByID` misses (returns -1) or the column comes back as something other than a `*extensions.VariantArray`, `varr` is nil, the loop at `:72` appends null for *every* row, and `payload.a = 5` then evaluates false everywhere. That is a wrong-results failure wearing the costume of an empty result — nothing errors and nothing logs.

`varIdx == -1` in particular is reachable when the arrow field lacks `PARQUET:field_id` metadata (name-mapping reads), which is precisely the schema-evolution case where "no rows" is indistinguishable from a legitimate answer.

Suggested fix: distinguish the two cases. Return an explicit error when the column is present but is not a `VariantArray`, and handle "column absent from this file" the way `scanTranslator.VisitBound` already does for a missing column — `AlwaysFalse` for comparison ops, `AlwaysTrue` for `IsNull` — rather than fabricating an all-null column.

variant_extract.go
```diff
@@ -0,0 +1,248 @@
...
+	// ExtractValue navigates v to this term's path and casts the leaf to the target type.
+	ExtractValue(v variant.Value) (Literal, bool)
+}
+
+// Extract creates an unbound variant sub-path term for a dotted JSONPath.
+func Extract(ref Reference, path string, typ PrimitiveType) UnboundTerm {
+	return &unboundExtract{ref: ref, path: path, typ: typ}
+}
+
+type unboundExtract struct {
```
zeroshade, line 41: **Blocking: extract terms serialize to `"term":{}` with no error, corrupting REST scan-planning filters.**

Neither `unboundExtract` nor `boundExtract[T]` has a `MarshalJSON`. `marshalLiteralPredicate` (`expr_json.go:230-239`) calls `json.Marshal(term)`, and a struct with only unexported fields marshals to `{}`. So a row filter containing an extract, sent through server-side scan planning (`catalog/rest/scan_planning.go:389`), is encoded as `{"type":"eq","term":{},"value":5}` and shipped to the catalog with no client-side error. There is also no `decodeTerm` case, so it cannot round-trip back.

The repo already has the right precedent for a non-serializable term — `expr_json.go:204-205` returns `ErrBBoxNotSerializable`.

Suggested fix:

```go
func (u *unboundExtract) MarshalJSON() ([]byte, error) {
    return nil, fmt.Errorf("%w: variant extract terms have no REST expression form", ErrNotImplemented)
}
func (b *boundExtract[T]) MarshalJSON() ([]byte, error) { /* same */ }
```

table/evaluators.go
```diff
@@ -868,25 +880,44 @@ func (m *inclusiveMetricsEval) VisitNotNan(t iceberg.BoundTerm) bool {
...
+	if raw == nil {
+		return nil, false
+	}
 
-	if m.containsNullsOnly(fieldID) || m.containsNansOnly(fieldID) {
-		return rowsCannotMatch
+	if ext, ok := t.(iceberg.BoundExtract); ok {
+		lit, found, err := internal.VariantBoundLiteral(raw, ext.Path(), ext.Type().(iceberg.PrimitiveType))
+		if err != nil {
+			panic(err)
```
zeroshade: Non-blocking: `panic(err)` turns an undecodable bound into a hard scan error — for example a bound written by another engine, or a `TestRowGroup` path where `lowerBounds[-1]` holds arbitrary variant-leaf stats. Metrics pruning should be best-effort; a bound we cannot decode means we cannot prune, not that the scan should fail.

Worth noting the round trip is self-defeating as written: `VariantBoundLiteral` installs a `recover()` (`table/internal/variant_bounds.go:317-321`) specifically to convert arrow-go panics into errors, and `boundFor` immediately re-panics them.

Suggested fix: return `rowsMightMatch` here instead of panicking.

table/evaluators.go
```diff
@@ -868,25 +880,44 @@ func (m *inclusiveMetricsEval) VisitNotNan(t iceberg.BoundTerm) bool {
 	return rowsMightMatch
 }
 
-func (m *inclusiveMetricsEval) VisitLess(t iceberg.BoundTerm, lit iceberg.Literal) bool {
-	field := t.Ref().Field()
-	fieldID := field.ID
+// boundFor decodes the file bound for term t from raw: a scalar for a reference, or the
```
zeroshade, line 958: Non-blocking, just flagging an unremarked behavior change: the non-primitive `panic` guard in `boundFor` now fires only when `raw != nil`, whereas previously it fired unconditionally in `VisitLess`, `VisitIn`, `VisitStartsWith`, and friends. That is more permissive and is very likely fine, but it is not mentioned anywhere in the PR description.

Suggested fix: confirm the loosening is intended and note it in the description (or in a comment here).

variant_path.go
```diff
@@ -0,0 +1,132 @@
...
+				b.WriteRune(r)
+			}
+		}
+	}
+
+	return b.String()
+}
+
+// parseVariantPath parses a dot-shorthand variant path ($.a.b) into its member names.
+func parseVariantPath(path string) ([]string, error) {
```
zeroshade, line 84: Non-blocking round-trip asymmetry: `parseVariantPath` rejects bracket notation, but `BoundExtract.Path()` (`variant_extract.go:146`) *emits* it. So `Extract(ref, term.Path(), typ)` cannot round-trip a path that normalizes to brackets. It also makes field names containing `.`, quotes, or a leading digit unreachable, which Java's `PathUtil.parse` supports.

Suggested fix: accept `$['a']` on input even if array indices stay unsupported for now.

table/variant_residual.go
```diff
@@ -0,0 +1,193 @@
...
+	for i := 0; i < n; i++ {
+		if varr == nil || varr.IsNull(i) {
+			bldr.AppendNull()
+
+			continue
+		}
+
+		v, verr := varr.Value(i)
+		if verr != nil {
+			bldr.AppendNull()
```
zeroshade, line 81: Non-blocking: a variant decode failure (`verr != nil`) is silently converted into "no match" for that row. That is a different thing from the value being absent, and the caller has no way to tell them apart.

Suggested fix: at minimum log it; ideally surface it, since a decode failure usually means the file is malformed rather than that the path is missing.

table/variant_residual.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/compute"
+	"github.com/apache/arrow-go/v18/arrow/extensions"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/iceberg-go"
+	"github.com/google/uuid"
+)
+
+// augmentSchemaWithExtracts returns fileSchema plus one primitive column per variant extract term.
+func augmentSchemaWithExtracts(fileSchema *iceberg.Schema, cols []iceberg.VariantExtractColumn) (*iceberg.Schema, error) {
```
zeroshade: Minor: `augmentSchemaWithExtracts` declares an `error` return it never produces — every caller has to handle an impossible case.

Suggested fix: drop the return value. Separately, `iceberg.NewSchema(fileSchema.ID, fields...)` on `:45` drops the source schema's identifier field IDs; harmless for binding today, but easy to preserve while you are here.

variant_path.go
```diff
@@ -0,0 +1,132 @@
...
+
+package iceberg
+
+import (
+	"fmt"
+	"strings"
+)
+
+// NormalizeVariantPath renders member names as the spec's RFC-9535 normalized JSON path.
+func NormalizeVariantPath(fields []string) string {
```
zeroshade, line 30: Minor API-surface note: `NormalizeVariantPath` and `CastVariantLiteral` (`variant_cast.go:41`) appear to be exported only so `table/internal` can reach them. That permanently widens the root package's public surface to serve an internal need.

Suggested fix: consider moving `VariantBoundLiteral`'s decode into the root package instead, or at minimum document both as internal-use and not covered by compatibility guarantees.

zeroshade (changes requested):
Thanks for this — it is a substantial piece of work and most of the machinery is well built. I read the full diff and traced the surrounding paths (binding, metrics evaluation, the substrait/record filter, partition-spec construction, and REST scan-plan serialization) rather than reviewing it in isolation.

What is genuinely solid: the RFC-9535 path parsing and normalization, the bounds decode in `table/internal/variant_bounds.go` (including the `recover()` that converts arrow-go panics into errors), the columnar residual filter design, the conservatism in strict evaluation, and the partition-source rejection. On that last one — `addSpecFieldInternal` (`partitions.go:266`) is exactly the right chokepoint. I confirmed that `AddPartitionFieldByName`, `AddPartitionFieldBySourceID`, and `table.UpdateSpec` (which routes through `AddPartitionFieldBySourceID` at `table/update_spec.go:199-202`) all funnel through it, so this genuinely closes the gap left by `IdentityTransform.CanTransform`, and it notably also covers `VoidTransform`, whose `CanTransform` returns `true` unconditionally.

Four correctness holes need fixing before this lands. Three are inline; the fourth is below because it lives in a hunk this PR does not touch.

## Blocking: `IsNull`/`NotNull` over an extract on a *required* variant column binds away to a constant

`exprs.go:596-603` (reached via `variant_extract.go:68`).

`unboundUnaryPredicate.Bind` applies its required-field fast path using `bound.Ref().Field()`. For a `BoundExtract`, `Ref()` is the **variant column**, not the sub-path. So for a schema containing `NestedField{ID: 2, Name: "payload", Type: VariantType{}, Required: true}`:

- `IsNull(Extract("payload", "$.a", Int64))` binds to `AlwaysFalse{}` — **zero rows**, even when `$.a` is absent from every row.
- `NotNull(Extract("payload", "$.a", Int64))` binds to `AlwaysTrue{}` — every row survives, even where `$.a` is absent.

A required variant column says nothing about whether a path *inside* it exists. This PR's own guards in `inclusiveMetricsEval.VisitIsNull`/`VisitNotNull` acknowledge exactly that, but binding runs first and collapses the predicate before the evaluator ever sees it. No test uses a required variant column, which is why it is invisible today.

Suggested fix — skip the fast path for extract terms:

```go
_, isExtract := bound.(BoundExtract)
switch up.op {
case OpIsNull:
    if !isExtract && bound.Ref().Field().Required && !schema.FieldHasOptionalParent(bound.Ref().Field().ID) {
        return AlwaysFalse{}, nil
    }
case OpNotNull:
    if !isExtract && bound.Ref().Field().Required && !schema.FieldHasOptionalParent(bound.Ref().Field().ID) {
        return AlwaysTrue{}, nil
    }
```

## Non-blocking, outside the diff

- `visitors.go:670` (`sanitizeVisitor.VisitBound`) — only the *comment* was updated here. An extract predicate still rebuilds as `Reference(pred.Term().Ref().Field().Name)`, so a ScanReport shows `payload = (redacted)`: a filter on the whole variant column, which the user never wrote. Suggested fix: mirror the `BoundBBoxPredicate` handling and collapse extracts to `AlwaysTrue{}`.
- PR description accuracy — the residual section says filtering is done "not row-by-row". The *predicate evaluation* is columnar, but `buildExtractColumn` materializes the derived column with a per-row loop (`table/variant_residual.go:56-79`), including a per-row `varr.Value(i)` variant decode. Worth restating so reviewers do not over-read the performance claim.

## Test coverage

The mechanical layers are covered genuinely well: path parse/normalize including escapes and rejects (`variant_path_test.go`), the cast table with widening, scale-mismatch, and floor-division cases (`variant_cast_test.go`), bounds decode including missing key, type mismatch, and malformed-bytes non-panic (`variant_bounds_test.go:694-728`), extract bind rejects (`variant_extract_test.go:50-61`), an end-to-end shredded-write-to-prune matrix (`variant_shredded_write_test.go:1159-1207`), shredded *and* non-shredded residual scans, AND/OR over multiple extracts, and a `CheckedAllocator` leak assertion. The leak test in particular is the right instinct for this code.

Missing, in rough priority order:

1. **A required (`Required: true`) variant column anywhere.** This single gap is what hides the blocking binding issue above; `TestVariantExtractPruningGuards` and `TestShreddedVariantExtractPruning` both use optional columns.
2. **`IsNull` / `NotNull` through the residual filter.** Only the metrics-guard side is tested. The interesting case — path absent in some rows and present in others, with `IsNull(extract)` selecting exactly the absent ones — is untested end to end.
3. **Anything through the public `table.Scan().Filter(...)` API.** Every residual test constructs `&arrowScan{...}` directly and calls the unexported `getRecordFilter`. Nothing exercises `PlanFiles` -> manifest eval -> `inclusiveProjection` with an extract filter, so the `FieldsBySourceID` -> `AlwaysTrue` behavior (correct as far as I traced it) is unverified, as is `projectedFieldIDs` pulling the variant column into the pruned read set.
4. **`boundExtract.eval` / `evalToLiteral` / `evalIsNull` (`variant_extract.go:208-241`) have zero coverage.** These are the `StructLike` row-evaluation path and they assert `raw.(variant.Value)` from the accessor; nothing confirms an accessor ever yields a `variant.Value`.
5. **Timezone cast cases** (`castVariantToMicros`/`Nanos`/`Date`) have no tests at all. A `TimestampTz`-target test over an NTZ leaf would have surfaced the cast issue immediately.
6. **`TestAppendExtractLiteral`'s comment claims it "covers every extract target builder"** but it tests only int64, string, binary, uuid, and fixed. Bool, int32, float32/64, date, time, both timestamp precisions, and decimal — including the `default:` arm inside the `TimestampBuilder` case — are untested.
7. **Schema evolution** — a file whose variant column is absent, and a file read via name mapping where the arrow field carries no `PARQUET:field_id` (which drives `fieldIndexByID` to -1, the residual issue flagged inline).
8. **JSON round-trip** of an expression containing an extract term.
9. **Nested path (`$.a.b`) through the residual filter.** Covered in `TestExtractBind`, but never materialized against real data.

variant_path.go
nssalian: Left these exported for now - they're used by both the root package and `table/internal`, so they can't move into `internal` or go unexported without root depending on the subpackage. Kept the "not part of the stable public API" note. Can revisit in a follow-up if you want them tucked somewhere shared.

table/variant_residual.go
nssalian: Both done - dropped the error return, and switched to `NewSchemaWithIdentifiers` so the identifier field IDs survive

table/variant_residual.go
nssalian: Added a `slog.Warn` for the undecodable-value case so it's not silent. Kept it non-fatal (didn't want one bad row to kill the whole scan) - happy to make it a hard error if you'd prefer.

variant_path.go
nssalian: Fixed - `parseVariantPath` now takes both dot and bracket forms, so `Path()` round-trips. Went a bit further and taught it to decode `\uXXXX` surrogate pairs (astral chars) instead of mangling them to U+FFFD; lone/unpaired surrogates error now. I had a PR for the java one too but it needed some fallback mechanism that was not trivial - will do that later.

table/evaluators.go
nssalian: Intentional - a nil bound just means "nothing to prune on", so keeping the file is the right best-effort behavior. Same posture as the extract decode path above.

table/evaluators.go
nssalian: Done for extracts - `boundFor` now returns `(nil, false)` on a decode error so pruning stays best-effort, no panic. Left the scalar-reference panic as-is since it's pre-existing and out of scope for this PR

variant_extract.go
nssalian: Now, both `unboundExtract` and `boundExtract` now have `MarshalJSON` returning `ErrExtractNotSerializable`, so it errors instead of shipping `{}`. Checked that the marshal sites (expr_json, REST scan-planning, codec) all propagate it. Asserted in `expr_json_test.go`.

table/variant_residual.go
nssalian: Done. A missing column now returns an error, and the `*VariantArray` assertion uses `, ok` and errors instead of leaving `varr` nil, so it can't null out the whole batch. Testing is covered by `TestBuildExtractColumnNameFallback` and `TestVariantExtractOrAbsentColumnNoPanic`

variant_cast.go
nssalian: This was a great catch. The nanos/date casts now gate on tz-awareness and only accept the matching physical type, so a zoneless leaf can't land in a tz target. `TestCastVariantTimestampTZ` covers each pair. Also guarded the multiply against int64 overflow while I was in there; it was silently wrapping past ~year 2262.

nssalian:
Also fixed two issues found while revisiting the diff: a nil-base panic when an `extract` is OR'd with `IsNull` on a column absent from the file, and an `int64` overflow in the Date/timestamp casts past ~year 2262 (now returns not-castable). In addition to that: `appendExtractLiteral` errors instead of panicking, `parseVariantPath` handles surrogate-pair escapes, and the scan tests are consolidated into `variant_shredded_test.go` with some cleanup.

zeroshade (changes requested):
The Variant extraction and partition-source work addresses seven of my nine earlier findings, but the new pruning rewrite can silently drop matching rows for a negated extract predicate.

### Blocking — negated extract is turned into a lossy pruning filter (`table/arrow_filter_plan.go:383-386`)

`stripExtractPredicates` replaces a `BoundExtract` predicate with `AlwaysTrue` at `:383-386`, while its `VisitNot` method at `:367-369` applies `NewNot(child)`. With `NOT(Equal(Extract("payload", "$.a", Int64), 1))`, the extract child becomes `AlwaysTrue`, and `NewNot(AlwaysTrue{})` folds to `AlwaysFalse{}`. `compileFileFilterPlan` then passes that false pruning expression through `RewriteNotExpr` at `:321`, so the cached Parquet stats evaluator rejects **every** row group before the residual synthetic-column filter ever runs.

A row whose `payload.a` is `2` should satisfy that predicate, but its entire row group is skipped and the scan returns silently incomplete results. This is a correctness issue rather than a performance one, which is why I'm marking it blocking.

The invariant is that a pruning filter may only ever be **more permissive** than the true predicate. An unevaluatable extract must therefore remain `AlwaysTrue` in *every* polarity; negating the placeholder inverts it to `AlwaysFalse` and turns pruning from conservative into lossy.

Suggested fixes, in rough order of preference:

- Replace the entire `NOT` subtree containing an extract with `AlwaysTrue`.
- Or disable row-group/bloom pruning altogether for any expression containing an extract.
- Rewriting `NOT` before stripping may also avoid this particular fold, if the resulting predicate is still handled conservatively.

Please add a regression test using the predicate and data shape above (`NOT(Equal(Extract("payload","$.a",Int64),1))` with a `payload.a = 2` row).

### Previously-raised findings that are now addressed

For completeness, and thank you for working through these:

- Timezone-aware vs zoneless timestamp casts are now gated per target and overflow-checked.
- Missing Variant columns and type assertions now return errors instead of silently null-filling every row.
- Extract JSON marshaling returns `ErrExtractNotSerializable`, with a regression test.
- Extract bound decode no longer panics; pruning stays best-effort.
- Bracket notation plus escaped/Unicode member names now parse.
- Decode failures log before the intentional null fallback.
- Schema augmentation preserves identifier field IDs via `NewSchemaWithIdentifiers`.

### Points I'm conceding

- **Nil-bound pruning** — accepted as intentional and conservative. Absent or undecodable bounds simply disable pruning, and no bound is ever used to reject a file. Your explanation holds.
- **Exported path/cast helpers** — accepted given the package-cycle constraint and the documentation that these are not stable public API. An internal package or a narrow accessor could be a future cleanup; it is not a merge blocker.
- **Decode-failure logging** — the diagnostic I asked for landed, and I am not requiring a hard error for malformed or unsupported Variant encodings.

### Partition-source rejection

`addSpecFieldInternal:405-414` is the right chokepoint: `AddPartitionFieldByName`, `AddPartitionFieldBySourceID`, and `BindToSchema` all route through it, and `NewPartitionSpec` cannot perform this check since it has no schema to inspect. No finding here.

Reviewed at head `aa6929f`; no failing checks, with two integration setup jobs still in progress.

---

>

nssalian:
Pushed the fix for the negated-extract pruning blocker - a NOT over an extract now collapses to AlwaysTrue (and hasExtract propagates through AND/OR, so nested NOTs are covered), so negation can't fold the pruning filter into a lossy AlwaysFalse. Added a regression test as well - NOT(Eq(Extract("payload","$.a",Int64),1)) with a payload.a=2 row, asserting it   
survives - plus coverage for the required-variant-column bind, the tz/zoneless cast rejection, and missing/wrong-type residual columns. Also strips extracts before TranslateColumnNames so the incremental-scan path no longer hits ErrNotImplemented.

table/variant_residual.go
```diff
@@ -0,0 +1,252 @@
...
+
+	bldr := array.NewBuilder(mem, dt)
+	defer bldr.Release()
+
+	n := int(rec.NumRows())
+	varName := col.Term.Ref().Field().Name
+	varIdx := fieldIndexByID(rec.Schema(), col.Term.Ref().Field().ID)
+	if varIdx < 0 {
+		// Arrow field may lack PARQUET:field_id metadata on name-mapping reads; fall back to the column name.
+		if idxs := rec.Schema().FieldIndices(varName); len(idxs) == 1 {
```
zeroshade: This fallback does not work after a name-mapped column rename. For an ID-less file, the Arrow batch retains the legacy physical name (for example, `old_payload`), while `col.Term.Ref().Field().Name` is the current logical name (`payload`), so the scan aborts instead of evaluating the extract. This resolution is also top-level-only, so a Variant nested in a struct cannot be found/materialized. Please carry or derive the translated physical field path and descend through nested containers rather than falling back only to the current top-level name.

visitors.go
```diff
@@ -822,6 +901,11 @@ func (sanitizeVisitor) VisitOr(left, right BooleanExpression) BooleanExpression
 }
 
 func (sanitizeVisitor) VisitUnbound(pred UnboundPredicate) BooleanExpression {
+	// An extract term has no REST expression-JSON form; keeping it would fail json.Marshal. Collapse it like bbox (mirrors VisitBound).
+	if _, ok := pred.Term().(*unboundExtract); ok {
+		return AlwaysTrue{}
```
zeroshade: The unsupported-term collapse needs the same polarity protection as `stripExtractPredicates`. Today `NOT(extract(...) = 1)` visits the extract as `AlwaysTrue` and then becomes `AlwaysFalse`, so the emitted `ScanReport` misstates the filter even though the scan returns rows. Please track whether the child contained an extract/bbox and make a `NOT` over such a subtree conservatively `AlwaysTrue` (or cause the report filter to be omitted).

zeroshade (changes requested):
The previous negated-extract pruning blocker is fixed: extract predicates beneath `NOT` are now removed conservatively, and the incremental-scan path strips extracts before row-group translation. Two correctness issues remain before this is ready.

### Blocking — extract source resolution (`table/variant_residual.go:61-74`)

`buildExtractColumn` resolves only top-level Arrow fields and falls back to the bound term's current logical name. For a valid name-mapped file without embedded field IDs, a renamed Variant column remains under its legacy physical name in the Arrow batch, so a filter on the current name fails with `variant extract column ... not found in file`. Nested Variant fields have the same structural problem because their field ID is not top-level and `rec.Column(varIdx)` cannot descend through the containing struct. Please resolve the source through the translated physical field path/ID and materialize nested fields as needed.

### Scan report can invert unsupported predicates (`visitors.go:891-941`)

Sanitizing `NOT(extract(...) = value)` currently produces `NOT(alwaysTrue)`, i.e. `alwaysFalse`. The scan itself executes the negated extract correctly, but its `ScanReport` claims a materially different filter. Please propagate whether a subtree contained a non-serializable term and conservatively collapse a `NOT` subtree containing one to `alwaysTrue`, or omit the report filter.

Validation: current CI is green; `go test ./ ./table/...` and targeted `go test -race ./table` passed at `8321091`.

---

>

table/variant_residual.go
nssalian: Done. I dropped the logical-name fallback entirely. The synthetic column now carries the translated physical field path (SourcePath, derived from fileSchema.FindColumnName in extractRef), and buildExtractColumn resolves the source by field id (descending nested structs) with the physical path as the id-less fallback. Rename and struct-nested cases both resolve now; an    unresolved source still errors cleanly instead of null-filling. Covered by TestBuildExtractColumnRenamedSource and TestBuildExtractColumnNested.                                            

re the "nested containers" - I checked whether an extract can even bind to a variant nested in a list/map: Reference.Bind rejects it with "invalid schema" (same as any  list/map-element ref, so it's the existing binding behavior, not extract-specific), which means those never reach materialization. So struct descent covers everything that can actually bind. Left it struct-only for that reason - happy to extend the descent to list/map if you'd rather have it explicit.

visitors.go
nssalian: Good catch - gave it the same polarity protection as stripExtractPredicates. sanitizeVisitor now returns a sanitizedResult carrying a hasUnserializable flag (set for extract and bbox), OR-propagated through And/Or, and VisitNot collapses a subtree containing one to AlwaysTrue instead of NewNot(AlwaysTrue) = AlwaysFalse. So the ScanReport no longer misstates the filter. TestSanitizeExpressionExtractCollapses now asserts the NOT(extract) case (unbound + bound) resolves to AlwaysTrue and still marshals.

table/variant_residual.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	return descendByPath(rec.Schema(), rec.Columns(), strings.Split(sourcePath, "."))
+}
+
+func descendByFieldID(f arrow.Field, col arrow.Array, fieldID int) arrow.Array {
+	if v, ok := f.Metadata.GetValue(ArrowParquetFieldIDKey); ok {
+		if id, err := strconv.Atoi(v); err == nil && id == fieldID {
+			return col
+		}
```
zeroshade: **minor** — Nested field-id descent is not pinned by any test and appears redundant

Deleting the entire struct-descent loop in descendByFieldID leaves the full table package green, including TestBuildExtractColumnNested whose assertion message claims 'nested variant resolved by descending the struct' but which actually resolves via the descendByPath fallback. Because fileSchema is derived from the same pruned Arrow schema the record batches carry, the path walk always aligns, so the field-id recursion has no exercised code path. Either add a test that only the field-id descent can satisfy (field ids present, path unresolvable) or drop the recursion and keep descendByPath.

table/variant_residual.go
```diff
@@ -0,0 +1,316 @@
...
+// nested structs; it falls back to the file-schema path when field ids are absent.
+func resolveVariantSource(rec arrow.RecordBatch, fieldID int, sourcePath string) arrow.Array {
+	for i, f := range rec.Schema().Fields() {
+		if a := descendByFieldID(f, rec.Column(i), fieldID); a != nil {
+			return a
+		}
+	}
+	if sourcePath == "" {
+		return nil
+	}
```
zeroshade, line 120: **minor** — strings.Split on '.' misresolves Variant columns whose Iceberg name contains a dot

resolveVariantSource re-splits the flattened column name returned by FindColumnName, so a legal Iceberg column literally named 'a.b' is looked up as top-level 'a' then child 'b'. On a name-mapped (field-id-less) file this aborts the scan. Resolve by walking the Iceberg schema's nested field structure (carrying the path segments through from FindColumnName) instead of re-splitting the flattened name.

visitors.go
```diff
@@ -802,54 +884,79 @@ const sanitizedLiteralMask = "(redacted)"
...
+		return nil, err
+	}
+
+	return res.expr, nil
+}
+
+// sanitizedResult is the masked expression plus whether its subtree held a non-serializable term.
+type sanitizedResult struct {
+	expr              BooleanExpression
+	hasUnserializable bool
```
zeroshade, line 898: **nit** — NOT over a mixed subtree discards the serializable conjuncts from the ScanReport filter

VisitNot collapses the whole subtree to AlwaysTrue whenever any descendant was unserializable, so NOT(AND(extract=5, x<10)) reports AlwaysTrue and the surviving x<10 never reaches ScanReport.Filter. The direction is safe (over-broad, never AlwaysFalse) and matches the bbox convention, but the emitted report can under-report the executed filter. Worth a one-line comment on the tradeoff, or push the collapse down to the offending leaf and negate the rest.

zeroshade (approved):
Both previously-requested changes are genuinely fixed and pinned by tests that go red under mutation; only three narrow non-gating observations remain (an untested redundant field-id descent, dot-in-name path splitting, and lossy NOT sanitization in ScanReport).

Re-review verification: 2 of 2 prior findings confirmed fixed at `b48f52c` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust).

```text
go build ./... (PASS); go vet ./ ./table/ (PASS); go test -race -count=1 ./ ./table/ (PASS: iceberg-go 1.883s, iceberg-go/table 18.534s); mutation run 1 - VisitNot polarity removed -> TestSanitizeExpressionExtractCollapses FAIL (expected AlwaysTrue, actual AlwaysFalse) x2; mutation run 2 - descendByPath fallback stubbed -> TestBuildExtractColumnNameFallback + TestBuildExtractColumnRenamedSource FAIL; mutation run 3 - descendByFieldID nested recursion deleted -> full table package still PASS (no coverage); 4 throwaway probes in table/ and 1 in root, plus a Parquet round-trip probe, all deleted afterwards (git status --porcelain clean).
```

</details>

---

>

table/variant_residual.go
nssalian: Dropped the recursion - resolveVariantSource now does a top-level field-id match then descendByPath for nesting, since the recursion wasn't covering anything the path walk didn't. Added   TestBuildExtractColumnResolvesByFieldID (ids present, path deliberately mismatched) so the id branch is actually pinned now.

table/variant_residual.go
nssalian: Fixed this. This now carries structured path segments from the schema (columnPathSegments walks the nested fields) instead of re-splitting the flattened name, so a column literally named a.b stays one segment. Added TestBuildExtractColumnDottedName.

visitors.go
nssalian: Added the one-line note - the collapse is over-broad (drops the serializable siblings from the report) but never AlwaysFalse, matching the bbox convention.

schema.go
```diff
@@ -524,6 +524,33 @@ func (s *Schema) accessorForField(id int) (accessor, bool) {
 	return acc, ok
 }
 
+// columnPathSegments returns the own-name path segments from the top-level column down to id, so a name containing '.' stays one segment.
+func (s *Schema) columnPathSegments(id int) []string {
+	idToField, err := s.lazyIDToField()
```
zeroshade, line 529: **major** — columnPathSegments — the actual dot-in-name fix — is guarded by no test, and the test that appears to guard it bypasses it

The production fix for the dot-in-name bug lives in Schema.columnPathSegments (schema.go:528), which replaced strings.Split(FindColumnName(id), ".") at visitors.go:721. It has zero coverage. The new TestBuildExtractColumnDottedName (table/variant_shredded_write_test.go:1995) constructs `SourcePath: []string{"a.b"}` by hand, so it never calls the producer; it only exercises descendByPath, which is unchanged in this delta. The fix is correct but free to silently regress, and the test name misrepresents what is covered. Ask: add a root-package test asserting columnPathSegments returns ["a.b"] for a top-level field named "a.b" and ["wrap.per","pay.load"] for the nested case (or assert SourcePath through TranslateColumnNamesForScan, which I confirmed yields ["a.b"]).

```text
Mutation 1: loop body -> 'return nil' unconditionally => `go test . -count=1` ok github.com/apache/iceberg-go 0.393s ; `go test ./table/ -count=1` ok github.com/apache/iceberg-go/table 8.266s. Mutation 2: body -> 'strings.Split(s.FindColumnName(id), ".")' (old buggy behavior) => `go test .` ok 0.390s ; `go test ./table/` ok 6.916s. Both files restored; git status --porcelain empty.
```

</details>

zeroshade (commented):
Delta is clean and all three prior items were addressed, but the dot-in-name fix in schema.go:columnPathSegments is pinned by no test — replacing its body with `return nil` leaves both affected package suites green.

Re-review verification: 1 of 3 prior findings confirmed fixed at `ee35a41` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *partially fixed* — Dot-in-name path splitting in table/variant_residual.go
- *not fixed* — Lossy NOT sanitization in visitors.go

```text
go build ./... OK; go vet . ./table ./catalog/... clean; go test . -count=1 -race ok (1.854s); go test ./table/ ./table/internal/ -count=1 -race ok (19.168s / 2.230s); go test ./catalog/glue/ ok (0.851s, refutes the stale pi-lens 'setup failed' report); targeted -run TestBuildExtractColumn|TestStripExtractPredicates|TestVariant|TestSanitize|TestTranslate ok; 5 mutation probes (2 on schema.go columnPathSegments, 2 on table/variant_residual.go resolveVariantSource branches) plus a throwaway pr1607_probe_test.go, all reverted.
```

</details>

---

>

schema.go
nssalian: Added TestColumnPathSegments in the root package asserting columnPathSegments returns ["a.b"] for a top-level a.b and ["wrap.per","pay.load"] for the nested case.

re the Lossy NOT sanitization in visitors.go: Kept this as the documented tradeoff for now - the collapse is over-broad but never AlwaysFalse, and it only affects the ScanReport filter string (never query results), so I'd rather not re-architect the sanitizer for a report-only nit.

PR #1609, opened 2026-07-30
fix(expr): reject trailing data after boolean expressions

fallintoplace:
## What changed

Fully decode top-level boolean expressions before returning `AlwaysTrue` or `AlwaysFalse`. This rejects malformed input containing a valid boolean followed by additional JSON or other trailing data.

## Why

The expression decoder previously inspected only the first token for booleans. Inputs such as `true false` were therefore accepted even though they are not valid JSON values. Object expressions already reject trailing data through `json.Unmarshal`.

## Testing

- `go test ./...`


laskoviymishka (approved):
⛵

PR #1610, opened 2026-07-30
fix(manifest): make empty writer close terminal

fallintoplace:
## Summary

Make the first ManifestWriter.Close call terminal, including for an empty manifest.

## Why

An empty-manifest error previously left the writer open. Entries could be added afterward, and a later Close or ToManifestFile call could turn the original failure into success.

## What changed

The writer now closes its OCF writer, stores the combined close result, and returns that same result from later close attempts.

## Tests

- Added coverage for repeated close, adding after close, and ToManifestFile after a failed close
- go test ./...

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1365,7 +1365,7 @@ func (w *ManifestWriter) Close() error {
 
 	var emptyErr error
 	if w.addedFiles+w.existingFiles+w.deletedFiles == 0 {
-		emptyErr = errors.New("empty manifest file has been written")
+		emptyErr = ErrEmptyManifest
 	}
 
 	var writerErr error
--- a/manifest.go
+++ b/manifest.go
@@ -1368,7 +1368,19 @@ func (w *ManifestWriter) Close() error {
 		emptyErr = errors.New("empty manifest file has been written")
 	}
 
-	w.closeErr = errors.Join(emptyErr, w.writer.Close())
+	var writerErr error
+	if w.writer != nil {
+		writerErr = w.writer.Close()
+	}
+
+	switch {
+	case emptyErr == nil:
+		w.closeErr = writerErr
+	case writerErr == nil:
+		w.closeErr = emptyErr
+	default:
+		w.closeErr = errors.Join(emptyErr, writerErr)
+	}
 
 	return w.closeErr
 }
@@ -1756,7 +1768,11 @@ func WriteManifest(
 	if err != nil {
 		return nil, err
 	}
-	defer internal.CheckedClose(w, &err)
+	defer func() {
+		if !w.closed {
+			internal.CheckedClose(w, &err)
+		}
+	}()
 
 	for _, entry := range entries {
 		if err := w.addEntry(entry.(*manifestEntry)); err != nil {
@@ -1793,7 +1809,11 @@ func WriteManifestV3(
 	if err != nil {
 		return nil, 0, err
 	}
-	defer internal.CheckedClose(w, &err)
+	defer func() {
+		if !w.closed {
+			internal.CheckedClose(w, &err)
+		}
+	}()
 
 	for _, entry := range entries {
 		if err := w.addEntry(entry.(*manifestEntry)); err != nil {
--- a/manifest.go
+++ b/manifest.go
@@ -1261,9 +1261,10 @@ func constructPartitionSummaries(spec PartitionSpec, schema *Schema, partitions 
 }
 
 type ManifestWriter struct {
-	closed  bool
-	version int
-	impl    writerImpl
+	closed   bool
+	closeErr error
+	version  int
+	impl     writerImpl
 
 	output io.Writer
 	writer *ocf.Writer
@@ -1358,16 +1359,18 @@ func NewManifestWriter(version int, out io.Writer, spec PartitionSpec, schema *S
 
 func (w *ManifestWriter) Close() error {
 	if w.closed {
-		return nil
+		return w.closeErr
 	}
-
-	if w.addedFiles+w.existingFiles+w.deletedFiles == 0 {
-		return errors.New("empty manifest file has been written")
-	}
-
 	w.closed = true
 
-	return w.writer.Close()
+	var emptyErr error
+	if w.addedFiles+w.existingFiles+w.deletedFiles == 0 {
+		emptyErr = errors.New("empty manifest file has been written")
+	}
+
+	w.closeErr = errors.Join(emptyErr, w.writer.Close())
+
+	return w.closeErr
 }
 
 type ManifestFileOption func(mf *manifestFile)
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -531,6 +531,24 @@ func (c manifestCloseErrorCodec) Close() error {
 	return c.err
 }
 
+func (m *ManifestTestSuite) setManifestWriterCloseError(writer *ManifestWriter, closeErr error) {
+	m.T().Helper()
+
+	avroSchema := writer.writer.Schema()
+	m.Require().NoError(writer.writer.Close())
+
+	replacement, err := ocf.NewWriter(
+		io.Discard,
+		avroSchema,
+		ocf.WithCodec(manifestCloseErrorCodec{
+			Codec: ocf.DeflateCodec(flate.DefaultCompression),
+			err:   closeErr,
+		}),
+	)
+	m.Require().NoError(err)
+	writer.writer = replacement
+}
+
 func (m *ManifestTestSuite) writeManifestList() {
 	err := WriteManifestList(1, &m.v1ManifestList, snapshotID, nil, nil, 0, manifestFileRecordsV1)
 	m.Require().NoError(err)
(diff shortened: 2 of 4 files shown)


manifest.go
```diff
@@ -1358,16 +1359,18 @@ func NewManifestWriter(version int, out io.Writer, spec PartitionSpec, schema *S
...
+	w.closed = true
 
+	var emptyErr error
 	if w.addedFiles+w.existingFiles+w.deletedFiles == 0 {
-		return errors.New("empty manifest file has been written")
+		emptyErr = errors.New("empty manifest file has been written")
 	}
 
-	w.closed = true
+	w.closeErr = errors.Join(emptyErr, w.writer.Close())
```
laskoviymishka: This line does two things I'd tighten. `w.writer.Close()` runs unconditionally now, so an empty close where the writer was never built (`NewManifestWriter` can hand back a non-nil `w` alongside a non-nil error) panics where the old early return returned cleanly. And `errors.Join` wraps the empty-manifest error in a `*joinError`, so `errors.As` and sentinel/`==` checks against it stop matching (`errors.Is` still works).

Both go away if we guard the writer and only join when there's a real second error:

```go
var writerErr error
if w.writer != nil {
	writerErr = w.writer.Close()
}
if emptyErr != nil {
	w.closeErr = errors.Join(emptyErr, writerErr)
} else {
	w.closeErr = writerErr
}
```

Exposure is narrow today (callers check the constructor error, no exported sentinel yet), but it's a cheap fix that keeps the contract clean. wdyt?

manifest.go
```diff
@@ -1358,16 +1359,18 @@ func NewManifestWriter(version int, out io.Writer, spec PartitionSpec, schema *S
...
 
-	return w.writer.Close()
+	return w.closeErr
```
laskoviymishka, line 1385: Now that `Close()` is terminal and returns the cached error on every call, I want to check the public wrappers. `WriteManifest` and `WriteManifestV3` both call `Close()` explicitly and also `defer internal.CheckedClose(w, &err)`, so on an empty writer they'll join the returned error with a second `Close()` result — `errors.Join(emptyErr, emptyErr)` — and hand the caller a doubled `empty manifest file has been written\nempty manifest file has been written`.

If those paths are reachable with zero entries (they look like they are — nothing guards the entry count before `WriteManifest`), I'd drop the redundant deferred `CheckedClose` there since `Close()` is already checked explicitly, or have `CheckedClose` skip an already-closed writer. Thoughts?

manifest_test.go
```diff
@@ -2091,6 +2091,20 @@ func (m *ManifestTestSuite) TestManifestWriterMeta() {
 	m.Equal("[]", string(md["partition-spec"]))
 }
 
+func (m *ManifestTestSuite) TestEmptyManifestWriterCloseIsTerminal() {
+	var out bytes.Buffer
+	writer, err := NewManifestWriter(2, &out, *UnpartitionedSpec, testSchema, snapshotID)
+	m.Require().NoError(err)
+
+	firstErr := writer.Close()
```
laskoviymishka: Good that this pins the terminal empty-close behavior. It only covers the empty branch though — I'd add the non-empty happy path (close a writer with entries, call `Close()` again, assert both return nil) so the new `closeErr` field's nil case is locked in, plus a case where the underlying writer errors on close so the two-error `errors.Join` path actually gets exercised.

Minor: `m.Equal(firstErr, writer.Close())` passes on pointer identity (both calls hand back the same cached error), which is correct but implicit — a one-line comment there would help the next reader.

laskoviymishka (changes requested):
Making `Close()` terminal is the right call — caching the result so a repeat `Close()` and `ToManifestFile()` return the same error is what `io.Closer` callers expect, and the test captures the intent well.

I'd hold before merging though. The `errors.Join(emptyErr, w.writer.Close())` line does two things I'd change: it calls `w.writer.Close()` unconditionally (which panics if the writer never got constructed), and it wraps the empty-manifest error in a `*joinError`, so callers doing `errors.As` or a sentinel check silently stop matching. Both go away if we assign the writer-close result directly and only join when there's genuinely a second error.

The other thing I want to confirm before merge: since `Close()` is now terminal and returns the cached error every time, the public `WriteManifest`/`WriteManifestV3` paths that pair an explicit `Close()` with a deferred `CheckedClose` will join the same error with itself on an empty writer and hand callers a doubled `empty manifest file has been written\nempty manifest file has been written` string. If that's reachable I'd want it fixed here.

Not blocking this PR, but worth noting: the empty-manifest-is-an-error policy itself has no spec backing (Java and PyIceberg allow zero-entry manifests), and making it terminal bakes it further into the contract. If we keep it, a package-level sentinel would at least let callers `errors.Is` it.

Details inline. Once those are addressed, happy to take another pass and approve.

manifest.go
```diff
@@ -1358,16 +1359,30 @@ func NewManifestWriter(version int, out io.Writer, spec PartitionSpec, schema *S
...
 	if w.closed {
-		return nil
+		return w.closeErr
 	}
+	w.closed = true
 
+	var emptyErr error
 	if w.addedFiles+w.existingFiles+w.deletedFiles == 0 {
-		return errors.New("empty manifest file has been written")
+		emptyErr = errors.New("empty manifest file has been written")
```
zeroshade: Consider promoting this to an exported sentinel — `var ErrEmptyManifest = errors.New("empty manifest file has been written")` — and wrapping it here. As an inline `errors.New`, callers have no way to `errors.Is` the condition, and the new tests have to match on the message text. Given that this PR is precisely about making the empty-manifest error stable and observable across repeat closes, a sentinel feels like the natural companion. Non-blocking.

manifest.go
```diff
@@ -1753,7 +1768,11 @@ func WriteManifest(
 	if err != nil {
 		return nil, err
 	}
-	defer internal.CheckedClose(w, &err)
+	defer func() {
```
zeroshade, line 1771: Non-blocking, for consistency: this `if !w.closed` guard (and its twin in `WriteManifestV3`) is the right pattern, but it is applied only at these two call sites. The `defer internal.CheckedClose(wr, &retErr)` plus explicit `wr.Close()` pairs in `table/snapshot_producers.go` (`:223`, `:378`, `:837`, `:907`) will now join the cached error with itself and surface a doubled message. Worth a follow-up that either repeats the guard there or moves the once-only behavior into a helper on the writer, so callers don't each have to know about it.

manifest_test.go
```diff
@@ -2091,6 +2108,117 @@ func (m *ManifestTestSuite) TestManifestWriterMeta() {
 	m.Equal("[]", string(md["partition-spec"]))
 }
 
+func (m *ManifestTestSuite) TestEmptyManifestWriterCloseIsTerminal() {
```
zeroshade, line 2129: The central question this change settles — that the file left on the output is now a readable Avro file containing zero entries — isn't asserted anywhere. Worth extending this case to decode `out.Bytes()` back and assert an empty entry list. That readability property is exactly what distinguishes the new behavior from the pre-PR truncated output, so it seems worth pinning. Suggestion only; the terminal-close assertions here are already solid.

manifest_test.go
```diff
@@ -2091,6 +2108,117 @@ func (m *ManifestTestSuite) TestManifestWriterMeta() {
...
+	var out bytes.Buffer
+	writer, err := NewManifestWriter(2, &out, *UnpartitionedSpec, testSchema, snapshotID)
+	m.Require().NoError(err)
+	m.Require().NoError(writer.Add(manifestEntryV2Records[0]))
+
+	m.NoError(writer.Close())
+	m.NoError(writer.Close())
+}
+
+func (m *ManifestTestSuite) TestEmptyManifestWriterCloseAfterConstructionFailure() {
```
zeroshade, line 2176: Consider a companion case: `Close()` after an `addEntry` failure with entries already written. That exercises the branch where `emptyErr` is nil but `writerErr` is not, and confirms the cached result stays stable across repeat closes on that path too. Non-blocking.

manifest_test.go
```diff
@@ -2091,6 +2108,117 @@ func (m *ManifestTestSuite) TestManifestWriterMeta() {
...
+	m.EqualError(
+		firstErr,
+		"empty manifest file has been written\nunderlying close failed",
+	)
+
+	// Close returns the cached combined result after becoming terminal.
+	m.Equal(firstErr, writer.Close())
+}
+
+func (m *ManifestTestSuite) TestWriteManifestEmptyErrorIsNotDuplicated() {
```
zeroshade, line 2250: Non-blocking coverage note: all the new cases construct v2 writers. Since `Close()` is shared across versions and `WriteManifestV3` received the same guard, running these table-driven over v1/v2/v3 would be cheap insurance against a version-specific regression.

zeroshade (commented):
The fix is correct. I verified the terminal-close semantics specifically:

- `Close()` records `closed` and caches its result in `closeErr` before doing any work, so a second call returns the identical error value rather than recomputing it (`manifest.go:1360-1386`). The empty-manifest error can no longer resurrect into a success on a repeat close, which was the point of the change.
- The empty and non-empty paths no longer diverge: both set `closed` first and both run `w.writer.Close()`, so the OCF writer is flushed even when the empty-manifest error is returned. Pre-change that path never closed the writer and left a truncated file; now a complete zero-entry Avro file is written. It is an unreferenced orphan — `WriteManifest`/`WriteManifestV3` still return the error and no `ManifestFile`, so it is never entered into a manifest list — but that is strictly better than the previous truncated output, not corruption.
- The `emptyErr`/`writerErr` switch joins only when both are non-nil, so each single-error case keeps its exact wording and `errors.Is` behavior intact.

Everything below is optional.

One cross-file note that falls outside this diff: the `if !w.closed` guard added at `manifest.go:1771` and `manifest.go:1812` is local to `WriteManifest`/`WriteManifestV3`. Four other sites pair a deferred `internal.CheckedClose(wr, &retErr)` with an explicit `wr.Close()` — `table/snapshot_producers.go:223`, `:378`, `:837`, and `:907` — and will now `errors.Join` the newly cached error with itself, producing a doubled message. Worth applying the same guard there, or factoring it into a small `closeOnce` helper, in a follow-up.

Separately, the new tests assert on error strings (`EqualError`, `ErrorContains`), which pins the wording into the test contract; an exported sentinel would let them assert on identity instead. Noted inline.

laskoviymishka (approved):
Now looks good!

PR #1611, opened 2026-07-30
fix(codec): reject nil data files in scan tasks

fallintoplace:
## Summary

Return a codec error instead of panicking when a FileScanTask contains a nil file.

## Why

The shared spec ID check called SpecID without checking the interface value first.

## What changed

The primary file and all three secondary file lists now reject nil entries before accessing file metadata.

## Tests

- Added coverage for the primary file and all three secondary file lists
- go test ./codec


laskoviymishka (approved):
:shipit:

PR #1612, opened 2026-07-30
fix(manifest): reject unsupported format versions

fallintoplace:
## What changed

Validate manifest and manifest-list format versions before selecting a reader schema. Versions 1, 2, and 3 remain supported, and a missing version still defaults to version 1.

## Why

Manifest lists previously routed every version other than 1 through the v2/v3 reader. Manifest files also accepted arbitrary integers. Unknown versions could therefore be interpreted with a schema whose semantics may not match the file.

## Testing

- Added manifest and manifest-list cases for versions -1, 0, and 4
- Existing version 1, 2, 3, and missing-version coverage remains intact
- `go test ./...`

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -546,10 +546,10 @@ type hasFieldToIDMap interface {
 	setFieldIDToDecimalScaleMap(map[int]int)
 }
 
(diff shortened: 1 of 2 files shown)


laskoviymishka (approved):
👍

PR #1613, opened 2026-07-30
fix(rest): reject blob offsets on non-Puffin delete files

fallintoplace:
## What changed

Reject `content-offset` and `content-size-in-bytes` on position-delete files unless the file format is Puffin.

## Why

These fields locate a deletion-vector blob inside a Puffin file. The decoder required them for Puffin files but accepted them on Parquet, Avro, and ORC position-delete files, creating an inconsistent `DataFile`.

## Testing

- Added cases for offset only, size only, and both fields on a non-Puffin position-delete file
- Existing Puffin and equality-delete validation remains covered
- `go test ./catalog/rest`

### DIFF
--- a/catalog/rest/scan_task_decoder.go
+++ b/catalog/rest/scan_task_decoder.go
@@ -343,6 +343,9 @@ func decodeRESTDeleteFile(
 		if len(wire.EqualityIDs) != 0 {
 			return nil, errors.New("position-deletes file must not carry equality-ids")
 		}
+		if format != iceberg.PuffinFile && (wire.ContentOffset != nil || wire.ContentSizeInBytes != nil) {
+			return nil, errors.New("content-offset and content-size-in-bytes are only valid for Puffin deletion vectors")
+		}
 		if wire.ContentOffset != nil {
 			if *wire.ContentOffset < 0 {
 				return nil, fmt.Errorf("content-offset must be non-negative: %d", *wire.ContentOffset)
--- a/catalog/rest/scan_task_decoder_test.go
+++ b/catalog/rest/scan_task_decoder_test.go
@@ -391,6 +391,28 @@ func TestDecodeScanTasksRejectsMalformedPayloads(t *testing.T) {
 			want: "must not carry equality-ids",
 		},
 		{
+			name: "non-Puffin position delete with content offset",
+			mutate: func(w *ScanTasks) {
+				w.DeleteFiles[0].ContentOffset = int64Ptr(10)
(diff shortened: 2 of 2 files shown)


catalog/rest/scan_task_decoder.go
```diff
@@ -343,6 +343,9 @@ func decodeRESTDeleteFile(
 		if len(wire.EqualityIDs) != 0 {
 			return nil, errors.New("position-deletes file must not carry equality-ids")
 		}
+		if format != iceberg.PuffinFile && (wire.ContentOffset != nil || wire.ContentSizeInBytes != nil) {
```
laskoviymishka, line 346: I think this guard is over-strict and I'd drop it. The REST OpenAPI schema puts `content-offset`/`content-size-in-bytes` on `PositionDeleteFile` with no `file-format: puffin` constraint, so a conformant server can legally send them on a non-Puffin position delete, and we'd now reject that response.

Both reference impls already accept it. Java's `ContentFileParser.fromJson` reads both fields off any delete file with no PUFFIN check, and PyIceberg's `test_scan_planning_models` fixture builds a `parquet` position-delete with both fields set and validates it fine. So we'd be the only client hard-failing on a wire response Java can produce.

The existing code just below already enforces the invariant we actually care about: the `format == PuffinFile` block requires these fields and fills `ReferencedDataFile`, and the non-Puffin path accepts them optionally with the value checks (size required when offset present, non-negative). Dropping the guard restores parity without losing any of that, and the three new test cases go with it.

Am I missing a spec reading that justifies the stricter rule here? wdyt?

laskoviymishka (changes requested):
Keeping blob offsets scoped to Puffin DVs is the right instinct, and I like that the fix comes with a case for each field.

I'd hold this before merge though, I think the guard is over-strict.

`content-offset` and `content-size-in-bytes` live on `PositionDeleteFile` in the REST OpenAPI schema with no `file-format: puffin` constraint, and both Java (`ContentFileParser.fromJson`) and PyIceberg accept a non-Puffin position delete that carries them. PyIceberg's own test fixture builds a parquet one with both fields set and validates it fine. So this guard would make iceberg-go the only client that rejects a response a spec-conformant server can legally send. The existing `format == PuffinFile` block already enforces the invariant we actually want, so I'd just drop the new guard, and the three test cases with it.

Left one inline with the specifics. If there's a spec reading that justifies the stricter rule I'd love to see it, but as it stands I'd want this resolved before we merge.

PR #1614, opened 2026-07-30
fix(rest): reject NUL in namespace identifiers

fallintoplace:
## Summary

Reject NUL characters in namespace identifiers at REST catalog boundaries.

## Why

REST namespace operations previously checked only that an identifier had at least one component. A NUL character is not a valid namespace component character and can lead to inconsistent behavior across catalog implementations.

## What changed

The catalog package now exposes shared namespace validation, and REST namespace creation uses it before performing storage or catalog work.

The validation intentionally remains compatible with existing namespaces: empty components, `.`, `..`, path separators, and newlines are not rejected here. Empty namespace identifiers remain rejected by catalog operations that require a concrete namespace.

## Tests

- Added valid single-level and nested namespace cases
- Added empty, NUL, and compatibility cases
- `go test ./catalog ./catalog/rest`

### DIFF
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -255,10 +255,6 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
 			notFoundErr, strings.Join(ident, "."))
 	}
 
-	return validateIdentifierComponents(ident, notFoundErr)
-}
-
-func validateIdentifierComponents(ident table.Identifier, notFoundErr error) error {
 	for _, part := range ident {
 		if part == "" || part == "." || part == ".." || strings.Contains(part, "/") {
 			return fmt.Errorf("%w: invalid identifier component %q in %v",
@@ -281,7 +277,7 @@ func validateIdentifierComponents(ident table.Identifier, notFoundErr error) err
 // intentionally left to the catalog implementation so existing namespaces
 // remain readable across clients.
 func ValidateNamespaceIdentifier(ident table.Identifier) error {
-	if len(ident) < 1 {
+	if len(ident) == 0 {
 		return fmt.Errorf("%w: empty namespace identifier", ErrNoSuchNamespace)
 	}
 
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -255,6 +255,10 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
 			notFoundErr, strings.Join(ident, "."))
 	}
 
+	return validateIdentifierComponents(ident, notFoundErr)
+}
+
+func validateIdentifierComponents(ident table.Identifier, notFoundErr error) error {
 	for _, part := range ident {
 		if part == "" || part == "." || part == ".." || strings.Contains(part, "/") {
 			return fmt.Errorf("%w: invalid identifier component %q in %v",
@@ -270,6 +274,15 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
 	}
 
 	return nil
+}
+
+// ValidateNamespaceIdentifier checks that an identifier contains at least one valid namespace level.
+func ValidateNamespaceIdentifier(ident table.Identifier) error {
+	if len(ident) < 1 {
+		return fmt.Errorf("%w: empty namespace identifier", ErrNoSuchNamespace)
+	}
+
+	return validateIdentifierComponents(ident, ErrNoSuchNamespace)
 }
 
 // ValidateTableIdentifier checks that an identifier contains at least one valid namespace level and a table name.
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -276,7 +276,10 @@ func validateIdentifierComponents(ident table.Identifier, notFoundErr error) err
 	return nil
 }
 
-// ValidateNamespaceIdentifier checks that an identifier contains at least one valid namespace level.
+// ValidateNamespaceIdentifier checks that an identifier contains at least one
+// namespace level and no null characters. Other namespace component rules are
+// intentionally left to the catalog implementation so existing namespaces
+// remain readable across clients.
 func ValidateNamespaceIdentifier(ident table.Identifier) error {
 	if len(ident) < 1 {
 		return fmt.Errorf("%w: empty namespace identifier", ErrNoSuchNamespace)
--- a/catalog/catalog.go
+++ b/catalog/catalog.go
@@ -282,7 +282,14 @@ func ValidateNamespaceIdentifier(ident table.Identifier) error {
 		return fmt.Errorf("%w: empty namespace identifier", ErrNoSuchNamespace)
 	}
 
-	return validateIdentifierComponents(ident, ErrNoSuchNamespace)
+	for _, part := range ident {
+		if strings.ContainsRune(part, '\x00') {
+			return fmt.Errorf("%w: invalid null character in namespace identifier component %q in %v",
+				ErrNoSuchNamespace, part, strings.Join(ident, "."))
+		}
+	}
+
+	return nil
 }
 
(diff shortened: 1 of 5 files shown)


catalog/catalog.go
```diff
@@ -272,6 +276,15 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
 	return nil
 }
 
+// ValidateNamespaceIdentifier checks that an identifier contains at least one valid namespace level.
+func ValidateNamespaceIdentifier(ident table.Identifier) error {
```
laskoviymishka, line 279: I think there's a real interop question buried in here. These component rules (rejecting `.`, `..`, `/`, and control chars) are stricter than the REST spec, Java's `Namespace.of`, and PyIceberg — Java only rejects the null byte, so `["ns/child"]` and `["ns\tname"]` round-trip fine through those clients.

Because `ValidateNamespaceIdentifier` guards every REST namespace op, including the read paths (Load/List/Drop all go through `checkValidNamespace`), a namespace another client already created with one of these components isn't just un-creatable here — it's un-loadable and un-droppable, and listing silently returns nothing.

What's the intended contract? If this is meant as write-side defensiveness I'd scope it to `CreateNamespace` and loosen it toward Java's rule (null byte only), and have the godoc note that it's stricter than the spec. wdyt?

catalog/catalog.go
```diff
@@ -272,6 +276,15 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
...
+	if len(ident) < 1 {
```
laskoviymishka: tiny thing: `len(ident) == 0` reads as the empty check more directly than `< 1`.

catalog/identifier_test.go
```diff
@@ -45,6 +45,24 @@ func TestValidateTableIdentifier(t *testing.T) {
 	}
 }
 
+func TestValidateNamespaceIdentifier(t *testing.T) {
+	require.NoError(t, catalog.ValidateNamespaceIdentifier(table.Identifier{"namespace"}))
+	require.NoError(t, catalog.ValidateNamespaceIdentifier(table.Identifier{"parent", "namespace"}))
+
+	for _, ident := range []table.Identifier{
```
laskoviymishka: The error table can't actually tell the two failure modes apart. `nil` and `{}` fail on the length check; `{""}`, `{"."}`, `{"namespace/child"}` fail on component validation — but both wrap `ErrNoSuchNamespace`, so if the component check silently stopped running the suite would still pass.

Could we split these into `t.Run` subtests (or at least `require.ErrorIsf` with the input) so a failure names the offending case? While we're here, a multi-component case like `{"parent", "namespace\nchild"}` would cover a control char in a non-first position — `TestValidateTableIdentifier` already does that for tables.

laskoviymishka (changes requested):
Nice cleanup — pulling the shared check into `ValidateNamespaceIdentifier` and unifying the empty-identifier handling reads well, and the new test covers the cases the helper rejects.

I'd hold this before merging though, on one design question. The component rules here (rejecting `.`, `..`, `/`, and control chars) are stricter than the REST spec, Java's `Namespace.of`, and PyIceberg — Java only rejects the null byte. And since `ValidateNamespaceIdentifier` guards every REST namespace op including the read paths, a namespace another client legitimately created with one of those components becomes un-loadable and un-droppable through this client, and listing silently returns nothing. So before we lock it in: what's the intended contract? If it's write-side defensiveness I'd scope it to `CreateNamespace` and loosen toward Java's rule.

Two smaller things: `sql/sql.go` still has its own `checkValidNamespace` with the old empty-only check — if REST-only is intentional here, a follow-up pointer would help. And the error table in the new test can't distinguish its length failures from its component failures since both wrap `ErrNoSuchNamespace`; splitting into subtests would make a failure name the input.

Once the contract question is settled, happy to take another pass and approve.

catalog/catalog.go
```diff
@@ -272,6 +276,25 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
...
+// namespace level and no null characters. Other namespace component rules are
+// intentionally left to the catalog implementation so existing namespaces
+// remain readable across clients.
+func ValidateNamespaceIdentifier(ident table.Identifier) error {
+	if len(ident) < 1 {
+		return fmt.Errorf("%w: empty namespace identifier", ErrNoSuchNamespace)
+	}
+
+	for _, part := range ident {
+		if strings.ContainsRune(part, '\x00') {
```
zeroshade, line 285: Non-blocking: consider rejecting the decoded REST namespace separator as well as NUL. With the default separator, `PathEscape("\x1f") == "%1F"`, so one level containing `0x1F` collides with two separate levels after encoding.

zeroshade (commented):
This change is correct for its stated scope: I verified that it matches Java's `Namespace.of` behavior and reaches all eight REST namespace call sites.

The main non-blocking security-hardening nit is that the dangerous character here is the decoded namespace separator, not NUL. With the default `%1F` separator (`catalog/rest/rest.go:84`), `url.PathEscape("\x1f")` is also `%1F`, so the single-level namespace `a<0x1F>b` and the two-level namespace `a`,`b` both encode to `a%1Fb` in `encodeNamespace` (`catalog/rest/rest.go:1235-1244`). The query-parameter path has the same collision at `catalog/rest/rest.go:1247-1256`. Consider rejecting the decoded configured separator rather than only a hardcoded NUL. Java has the same gap, and this PR does not widen it.

Worth a follow-up at `catalog/rest/rest.go:1219-1224`: identifiers returned by list responses are passed back without validation, so a server can return a namespace that the request path would reject. Response-path coverage would make that contract explicit.

Cross-PR coordination notes, all non-blocking here: #1571 defines the same validator and helper as `catalog/catalog.go:258-295`, so keeping this version's more accurate godoc while rebasing #1571 avoids duplicate definitions; #1571's `catalog/sql/sql.go:1236` applies the namespace rule only on `CreateNamespace`, while this PR applies it throughout REST, so a documented common contract would help; and #1606 also edits the `Close`/`checkValidNamespace` region at `catalog/rest/rest.go:1069-1078`. A separate follow-up around `catalog/catalog.go:283-305` could also document the pre-existing divergence where namespace `{".."}` is valid but no table or view can live inside it.

catalog/catalog.go
```diff
@@ -255,6 +255,10 @@ func validateIdentifier(ident table.Identifier, notFoundErr error) error {
 			notFoundErr, strings.Join(ident, "."))
 	}
 
+	return validateIdentifierComponents(ident, notFoundErr)
```
zeroshade: Worth simplifying in a follow-up: `validateIdentifierComponents` still has only this one caller, while the new namespace validator deliberately does not share these component rules. Keeping the loop in `validateIdentifier` would avoid a dead abstraction left over from the earlier shared-validation design.

catalog/rest/rest_test.go
```diff
@@ -807,6 +807,33 @@ func (r *RestCatalogSuite) TestListNamespaceWithParent200() {
...
+		requestCount++
+	})
+	r.mux.HandleFunc("/v1/namespaces/", func(w http.ResponseWriter, req *http.Request) {
+		requestCount++
+	})
+
+	cat, err := rest.NewCatalog(context.Background(), "rest", r.srv.URL, rest.WithOAuthToken(TestToken))
+	r.Require().NoError(err)
+
+	namespace := table.Identifier{"bad\x00name"}
```
zeroshade: Consider adding `0x1F` coverage alongside the NUL case, including both `[]string{"a\x1fb"}` and `[]string{"a", "b"}`. Optional coverage of `ListTables`, `ListViews`, and `ListFunctions` would also pin the call sites newly reached through `checkValidNamespace`.

laskoviymishka (approved):

PR #1615, opened 2026-07-30
fix(dv): validate deletion-vector blob identity

fallintoplace:
## Summary

Validate that a Puffin deletion-vector blob matches its manifest entry.

## Why

ReadDV selected a blob by offset and length and checked cardinality, but did not verify the blob type or referenced data file. A malformed manifest entry could therefore read a deletion vector for another data file.

## What changed

ReadDV now requires a non-empty referenced data file, checks for type deletion-vector-v1, and verifies the footer's referenced-data-file property before decoding. Existing cardinality checks remain unchanged.

## Tests

- Added mismatched and missing data-file reference cases
- Added a non-DV blob case
- Added nil and empty manifest reference cases
- go test ./table/dv
- go test ./...

### DIFF
--- a/table/dv/deletion_vector.go
+++ b/table/dv/deletion_vector.go
@@ -20,6 +20,7 @@ package dv
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"hash/crc32"
 	"log/slog"
@@ -39,6 +40,8 @@ const (
 	dvCardinalityProperty        = "cardinality"
 	dvReferencedDataFileProperty = "referenced-data-file"
 )
+
+var ErrInvalidDeletionVector = errors.New("invalid deletion vector")
 
 const (
 	// DVMagicNumber is the magic number for deletion vectors.
@@ -151,11 +154,13 @@ func SerializeDV(bitmap *RoaringPositionBitmap) ([]byte, error) {
 //
 // ReadDV also requires the selected blob to be a deletion-vector blob whose
 // referenced-data-file property matches the manifest. This is stricter than
-// clients that only use the manifest offset and size, but prevents a valid
-// Puffin blob for another data file from being applied here. A missing or
-// empty referenced-data-file property is fatal because it cannot establish
-// blob identity; a missing cardinality property is only warned about because
-// the manifest record_count still bounds the decoded bitmap.
+// Java's DVUtil.readDV and PyIceberg, which read by offset and size without
+// validating blob type or referenced-data-file, but prevents a valid Puffin
+// blob for another data file from being applied here. A missing or empty
+// referenced-data-file property is fatal because it cannot establish blob
+// identity; a missing cardinality property is only warned about because the
+// manifest record_count still bounds the decoded bitmap. The compression
+// codec is also rejected because deletion-vector-v1 stores raw bytes here.
 //
 // Blobs missing the spec-required cardinality property are still validated
 // against the manifest record_count and accepted with a slog warning rather
@@ -171,7 +176,7 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
 	}
 	manifestReferencedDataFile := dvFile.ReferencedDataFile()
 	if manifestReferencedDataFile == nil || *manifestReferencedDataFile == "" {
-		return nil, fmt.Errorf("DV file %s missing ReferencedDataFile", dvFile.FilePath())
+		return nil, fmt.Errorf("%w: DV file %s missing or empty %s property", ErrInvalidDeletionVector, dvFile.FilePath(), dvReferencedDataFileProperty)
 	}
 
 	size := *dvFile.ContentSizeInBytes()
@@ -193,20 +198,24 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
 	offset := *dvFile.ContentOffset()
 	blob, err := findBlobMetadataByRange(reader.Blobs(), offset, size)
 	if err != nil {
-		return nil, fmt.Errorf("DV file %s: %w", dvFile.FilePath(), err)
+		return nil, fmt.Errorf("%w: DV file %s: %w", ErrInvalidDeletionVector, dvFile.FilePath(), err)
 	}
 	if blob.Type != puffin.BlobTypeDeletionVector {
-		return nil, fmt.Errorf("DV file %s: blob at offset %d has type %q, expected %q",
+		return nil, fmt.Errorf("%w: DV file %s: blob at offset %d has type %q, expected %q", ErrInvalidDeletionVector,
 			dvFile.FilePath(), offset, blob.Type, puffin.BlobTypeDeletionVector)
+	}
+	if blob.CompressionCodec != nil && *blob.CompressionCodec != "" {
+		return nil, fmt.Errorf("%w: DV file %s: blob at offset %d uses unsupported compression codec %q",
+			ErrInvalidDeletionVector, dvFile.FilePath(), offset, *blob.CompressionCodec)
 	}
 
 	referencedDataFile, ok := blob.Properties[dvReferencedDataFileProperty]
 	if !ok || referencedDataFile == "" {
-		return nil, fmt.Errorf("DV file %s: blob at offset %d missing or empty %s property",
+		return nil, fmt.Errorf("%w: DV file %s: blob at offset %d missing or empty %s property", ErrInvalidDeletionVector,
 			dvFile.FilePath(), offset, dvReferencedDataFileProperty)
 	}
 	if referencedDataFile != *manifestReferencedDataFile {
-		return nil, fmt.Errorf("DV file %s: manifest referenced_data_file %q does not match puffin %s %q",
+		return nil, fmt.Errorf("%w: DV file %s: manifest referenced_data_file %q does not match puffin %s %q", ErrInvalidDeletionVector,
 			dvFile.FilePath(), *manifestReferencedDataFile, dvReferencedDataFileProperty, referencedDataFile)
 	}
 
@@ -252,7 +261,8 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
 }
 
 // findBlobMetadataByRange returns the footer entry identified by the
-// manifest's content offset and size.
+// manifest's content offset and size. On error it returns a zero-value
+// puffin.BlobMetadata; callers must check the error before reading the value.
 func findBlobMetadataByRange(blobs []puffin.BlobMetadata, offset, size int64) (puffin.BlobMetadata, error) {
 	for _, b := range blobs {
 		if b.Offset != offset {
--- a/table/dv/deletion_vector.go
+++ b/table/dv/deletion_vector.go
@@ -35,7 +35,10 @@ import (
 // writer attaches it; the reader uses it to detect truncated bitmaps whose
 // CRC happens to validate (the CRC covers the bytes that ARE present, not
 // the bytes that should have been).
-const dvCardinalityProperty = "cardinality"
+const (
+	dvCardinalityProperty        = "cardinality"
+	dvReferencedDataFileProperty = "referenced-data-file"
+)
 
 const (
 	// DVMagicNumber is the magic number for deletion vectors.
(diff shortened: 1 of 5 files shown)


table/dv/deletion_vector_test.go
```diff
@@ -425,9 +450,55 @@ func TestReadDVInvalidPuffin(t *testing.T) {
...
 }
 
+// Why: offset, size, and cardinality cannot prove that the selected Puffin blob
+// is a deletion vector for the manifest's referenced data file.
+// Condition: the matched blob has conflicting, missing, or non-DV identity metadata.
+// Assertion: ReadDV rejects each case before decoding the blob payload.
+func TestReadDVValidatesBlobMetadata(t *testing.T) {
+	dvBlobBytes := readDVTestData(t, "small-alternating-values-position-index.bin")
+
+	t.Run("mismatched referenced data file", func(t *testing.T) {
```
laskoviymishka, line 483: This doesn't actually exercise the redirect it's guarding against.

Every fixture here is a single-blob Puffin file, so this subtest proves we compare the blob's `referenced-data-file` against the manifest's. It never proves that `findBlobMetadataByRange` picked the right blob to compare in the first place.

I'd add a subtest that writes two DV blobs (data-001 and data-002) into one Puffin file, then calls `ReadDV` with the manifest's `referenced_data_file` set to data-001 but `(offset, size)` pointing at the data-002 blob. That's the actual attack. It's also the only case where a future refactor of `findBlobMetadataByRange` that returns the first blob regardless of offset would go undetected today.

table/dv/deletion_vector_test.go
```diff
@@ -425,9 +450,55 @@ func TestReadDVInvalidPuffin(t *testing.T) {
...
+	})
+
+	t.Run("missing puffin referenced data file", func(t *testing.T) {
+		dir := t.TempDir()
+		path, meta := writeRawPuffinBlob(t, dir, "raw-dv-no-reference.puffin", dvBlobBytes,
+			puffin.BlobTypeDeletionVector, map[string]string{"cardinality": "5"})
+		offset, size := meta.Offset, meta.Length
+
+		_, err := ReadDV(iceio.LocalFS{}, newDVTestFile(path, 5, &offset, &size))
+		assert.ErrorContains(t, err, "missing referenced-data-file property")
```
laskoviymishka: We cover the absent key but not the present-but-empty value. The guard is `!ok || referencedDataFile == ""`, and nothing here would fail if someone trimmed it to `!ok`. Worth a sibling subtest with `props: {"referenced-data-file": "", "cardinality": "5"}` so that half of the condition is pinned too.

table/dv/deletion_vector.go
```diff
@@ -176,9 +183,23 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
...
+		return nil, fmt.Errorf("DV file %s: %w", dvFile.FilePath(), err)
+	}
+	if blob.Type != puffin.BlobTypeDeletionVector {
+		return nil, fmt.Errorf("DV file %s: blob at offset %d has type %q, expected %q",
+			dvFile.FilePath(), offset, blob.Type, puffin.BlobTypeDeletionVector)
+	}
+
+	referencedDataFile, ok := blob.Properties[dvReferencedDataFileProperty]
+	if !ok || referencedDataFile == "" {
+		return nil, fmt.Errorf("DV file %s: blob at offset %d missing %s property",
```
laskoviymishka: When the property is present but empty we report it as "missing", which points an operator diagnosing a corrupt blob at the wrong thing. I'd make it `missing or empty %s property`, or split the two conditions if we want them separately diagnosable. Either is fine, just not "missing" for a key that's there.

table/dv/deletion_vector_test.go
```diff
@@ -139,28 +139,31 @@ func writePuffinWithDVBlobAndProps(t *testing.T, dir string, dvBlobBytes []byte,
 // writer) is to assemble the bytes directly. The reader's validateBlobs does
 // not require the property, so this file loads cleanly.
 func writeRawPuffinWithDVBlobNoCardinality(t *testing.T, dir string, dvBlobBytes []byte) (string, puffin.BlobMetadata) {
+	return writeRawPuffinBlob(t, dir, "raw-dv-no-cardinality.puffin", dvBlobBytes, puffin.BlobTypeDeletionVector, map[string]string{
+		"referenced-data-file": "s3://bucket/data/data-001.parquet",
```
laskoviymishka: `dvReferencedDataFileProperty` is visible here (we're in package `dv`, not `dv_test`), so I'd use the constant rather than the literal. Same for the `"cardinality"` keys in the other fixtures while we're here. As written, renaming the constant compiles clean and the fixtures quietly stop matching what the reader looks up.

table/dv/deletion_vector.go
```diff
@@ -176,9 +183,23 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
 	}
 
 	offset := *dvFile.ContentOffset()
-	blobData := make([]byte, size)
-	if _, err := reader.ReadAt(blobData, offset); err != nil {
-		return nil, fmt.Errorf("read DV blob at offset %d: %w", offset, err)
+	blob, err := findBlobMetadataByRange(reader.Blobs(), offset, size)
```
laskoviymishka, line 199: Worth a couple of sentences in the `ReadDV` doc about how deliberate this strictness is.

Java's `DVUtil.readDV` reads purely by `contentOffset`/`contentSizeInBytes` and validates neither the blob type nor `referenced-data-file`; PyIceberg does the same. So after this we reject spec-non-conformant third-party Puffin files that both of them would happily read. I think that's the right call, since both properties are spec-required and our own writer always sets them, but it's a real divergence and the next person to hit it should find the reasoning in the doc rather than in git history.

The asymmetry deserves a line too: a missing `referenced-data-file` is fatal while a missing `cardinality` is only a warning, and both are spec-required. Identity check versus redundancy check is a sound distinction, it just isn't visible from the code.

table/dv/deletion_vector.go
```diff
@@ -211,23 +232,20 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
...
-//   - (n, true, nil)  — property found and parsed successfully
-//   - (0, false, nil) — matching blob found but no cardinality property
-//   - (_, _, err)     — manifest/footer mismatch or property unparseable
-//
-// Keeping the sentinel out of the int64 return channel avoids leaking
-// DeserializeDV's "-1 means skip" convention up the call chain.
-func blobCardinality(blobs []puffin.BlobMetadata, offset, size int64) (int64, bool, error) {
+// findBlobMetadataByRange returns the footer entry identified by the
+// manifest's content offset and size.
+func findBlobMetadataByRange(blobs []puffin.BlobMetadata, offset, size int64) (puffin.BlobMetadata, error) {
```
laskoviymishka, line 266: Worth noting in the doc that the returned `puffin.BlobMetadata` is a zero value on the error paths and must not be read. The caller is correct today (the `err` check precedes `blob.Type`), but returning `*puffin.BlobMetadata` with nil on error would make that self-documenting instead of conventional. Either works, wdyt?

laskoviymishka (changes requested):
Good catch on the gap. Offset plus size plus cardinality genuinely can't prove the blob we landed on is a DV for the data file the manifest names, and moving the allocation and `ReadAt` below the metadata gates means a bad manifest entry no longer buys a large allocation first. The `findBlobMetadataByRange` / `blobCardinality` split is cleaner than what was there.

I'd hold this before merging though, mostly on the test side.

The scenario this PR exists to prevent, a manifest entry whose `(offset, size)` redirects to a different data file's DV, is never directly constructed. All three subtests in `TestReadDVValidatesBlobMetadata` write single-blob Puffin files, so "mismatched referenced data file" proves we compare the property, but not that we selected the right blob to compare. Those two invariants only compose in a reviewer's head right now. A two-blob fixture (data-001 and data-002 in one file, manifest naming data-001, offset pointing at data-002) makes it real, and it's what would catch a future refactor of `findBlobMetadataByRange` that stops honoring the offset.

A few things I'd ask to fix before merge:

- a two-blob fixture that exercises the actual redirect
- a subtest for `referenced-data-file` present-but-empty, since the `!ok ||` half of that guard is currently unpinned
- `missing %s property` reworded, because it also fires when the property is present and empty
- the test fixtures using `dvReferencedDataFileProperty` instead of the raw string

The rest is doc polish and can ride along or not. I would like a note in the `ReadDV` doc that we're now stricter than Java here (`DVUtil.readDV` validates neither blob type nor `referenced-data-file`) and that the referenced-data-file-fatal versus cardinality-warned asymmetry is intentional.

table/dv/deletion_vector.go
```diff
@@ -146,6 +149,14 @@ func SerializeDV(bitmap *RoaringPositionBitmap) ([]byte, error) {
 // stale manifest record_count against a freshly written blob) is a writer bug
 // and fails fast. The bitmap is then validated against the manifest count.
 //
+// ReadDV also requires the selected blob to be a deletion-vector blob whose
```
zeroshade, line 155: Non-blocking: Consider tightening this documentation by naming Java's `DVUtil.readDV`, which makes the interoperability comparison directly checkable, and collapsing the repeated missing-cardinality rationale now present at lines 157-163.

table/dv/deletion_vector.go
```diff
@@ -158,6 +169,10 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
 	if dvFile.ContentOffset() == nil || dvFile.ContentSizeInBytes() == nil {
 		return nil, fmt.Errorf("DV file %s missing ContentOffset/ContentSizeInBytes", dvFile.FilePath())
 	}
+	manifestReferencedDataFile := dvFile.ReferencedDataFile()
+	if manifestReferencedDataFile == nil || *manifestReferencedDataFile == "" {
+		return nil, fmt.Errorf("DV file %s missing ReferencedDataFile", dvFile.FilePath())
```
zeroshade: Non-blocking: Consider spelling this field `referenced_data_file` to match the specification and the mismatch diagnostic below. Since the new identity failures otherwise expose only strings, a package sentinel wrapped with `%w` would also let operator tooling distinguish metadata corruption with `errors.Is`.

table/dv/deletion_vector.go
```diff
@@ -176,9 +191,23 @@ func ReadDV(fs iceio.IO, dvFile iceberg.DataFile) (*RoaringPositionBitmap, error
...
 
 	offset := *dvFile.ContentOffset()
-	blobData := make([]byte, size)
-	if _, err := reader.ReadAt(blobData, offset); err != nil {
-		return nil, fmt.Errorf("read DV blob at offset %d: %w", offset, err)
+	blob, err := findBlobMetadataByRange(reader.Blobs(), offset, size)
+	if err != nil {
+		return nil, fmt.Errorf("DV file %s: %w", dvFile.FilePath(), err)
+	}
+	if blob.Type != puffin.BlobTypeDeletionVector {
```
zeroshade, line 203: Non-blocking: Consider rejecting a non-empty `blob.CompressionCodec` here. `deletion-vector-v1` omits the compression codec, and this path reads the blob bytes raw; a blob declaring zstd currently falls through to a confusing `invalid deletion vector magic` error instead of reporting the metadata violation.

table/dv/deletion_vector_test.go
```diff
@@ -425,9 +468,95 @@ func TestReadDVInvalidPuffin(t *testing.T) {
 	assert.ErrorContains(t, err, "create puffin reader")
 }
 
+// Why: offset, size, and cardinality cannot prove that the selected Puffin blob
+// is a deletion vector for the manifest's referenced data file.
+// Condition: the matched blob has conflicting, missing, or non-DV identity metadata.
+// Assertion: ReadDV rejects each case before decoding the blob payload.
+func TestReadDVValidatesBlobMetadata(t *testing.T) {
```
zeroshade, line 480: Non-blocking: Consider extending this coverage with a scan-layer propagation case proving an identity mismatch escapes `readAllDeletionVectors`, an explicit exact-match case (`s3a://...` versus `s3://...`), and a compression-codec-bearing blob. Lower-priority additions would use `BlobTypeDataSketchesTheta` for the wrong-type collision and add a Go-written positive multi-blob case.

zeroshade (commented):
This change is correct. I verified the complete identity chain: the Puffin blob property equals the manifest entry's `referenced_data_file`, which is also the key used to look up the bitmap in both scan and transaction paths. Every failure branch propagates an error, and moving allocation below the footer and metadata gates means a bogus entry can no longer trigger an allocation of up to 256 MiB before being rejected.

The exact path comparison is the right identity check; do not add path normalization. The Puffin specification says the property is equal to the data-file location, this repository already uses exact equality for the same field at `table/data_file_meta.go:364-367`, and folding schemes such as `s3://` and `s3a://` would create false accepts.

Non-blocking residual interoperability considerations:

- A table relocated by tooling predating Java's `RewriteTablePathUtil.rewriteDVFile`, or by a rewriter that changed only manifests, may retain a stale blob property. Go will now fail that scan, while Java's `DVUtil.readDV` still reads the byte range without opening the Puffin footer. The error is precise, but this is a behavior change worth acknowledging.
- `table/dv/deletion_vector.go:203-240` treats a missing `referenced-data-file` as fatal but a missing `cardinality` as a warning, although both properties appear in the same specification MUST clause. The identity-versus-redundant-count rationale is defensible; this deliberate deviation would benefit from explicit maintainer sign-off.
- `table/dv/dv_cross_client_test.go:85,176` still uses raw `"referenced-data-file"` and `"cardinality"` literals. Consider finishing the constant conversion for consistency.

laskoviymishka (approved):
🌬️ ⛵

PR #1616, opened 2026-07-30
fix(table): reject ambiguous row-id columns in position deltas

fallintoplace:
## What changed

Require position-delta batches to contain at most one `_row_id` column. The shared lookup is used by `Reinsert`, `Insert`, and the null-column append path.

## Why

Arrow schemas can contain duplicate field names. Selecting the first `_row_id` silently made reserved row-lineage metadata ambiguous and could ignore a conflicting or malformed second column.

## Testing

- Added duplicate `_row_id` coverage for both `Reinsert` and `Insert`
- `go test ./table`


laskoviymishka (approved):
👍

PR #1617, opened 2026-07-30
fix(table): reject ambiguous equality-delete columns

fallintoplace:
## Summary

Resolve equality-delete fields by Iceberg field ID on both delete files and data files.

## Why

Arrow permits duplicate field names, and Iceberg columns can be renamed or stored in legacy files without embedded field IDs. Equality-delete filtering must bind to Iceberg field identity instead of selecting the first matching Arrow name.

## What changed

- Build a per-file field ID to Arrow path index from the ID-resolved physical schema
- Use embedded Parquet field IDs when present and the table name mapping when IDs are absent
- Support equality fields nested in structs, including null parent structs
- Reject duplicate physical field IDs and ambiguous equality-delete columns
- Validate equality field ID and column-name metadata before filtering
- Reuse resolved field paths across record batches while rebuilding array encoders per batch

## Tests

- Duplicate equality columns and duplicate field IDs
- Renamed physical columns with embedded field IDs
- Renamed physical columns resolved through a name mapping
- Literal dotted column names without embedded field IDs
- Nested struct paths and null parent structs
- Mismatched equality field metadata
- `go test ./table`
- Focused equality-delete tests with the race detector

### DIFF
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -217,11 +217,15 @@ func readEqualityDeleteFile(ctx context.Context, fs iceio.IO, tableSchema *icebe
 
 		colNames[i] = name
 		indices := tbl.Schema().FieldIndices(name)
-		if len(indices) == 0 {
+		switch len(indices) {
+		case 0:
 			return nil, nil, fmt.Errorf("equality delete column %q not found in delete file %s", name, dataFile.FilePath())
+		case 1:
+			colIndices[i] = indices[0]
+		default:
+			return nil, nil, fmt.Errorf("equality delete column %q is ambiguous in delete file %s: found %d columns",
+				name, dataFile.FilePath(), len(indices))
 		}
-
-		colIndices[i] = indices[0]
 	}
 
 	// Build the set of encoded delete keys by iterating aligned batches.
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -148,15 +148,19 @@ func resolveArrowField(schema *arrow.Schema, fieldID int, fieldName, filePath st
 }
 
 func arrowArrayAtField(record arrow.RecordBatch, ref arrowFieldRef, fieldID int, fieldName, filePath string) (arrow.Array, error) {
+	location := filePath
+	if location == "" {
+		location = "data record"
+	}
 	if len(ref.path) == 0 || ref.path[0] >= int(record.NumCols()) {
-		return nil, fmt.Errorf("equality field ID %d (%s) not found in %s", fieldID, fieldName, filePath)
+		return nil, fmt.Errorf("equality field ID %d (%s) not found in %s", fieldID, fieldName, location)
 	}
 
 	result := record.Column(ref.path[0])
 	for _, index := range ref.path[1:] {
 		structArray, ok := result.(*array.Struct)
 		if !ok || index >= structArray.NumField() {
-			return nil, fmt.Errorf("equality field ID %d (%s) has unsupported nested path in %s", fieldID, fieldName, filePath)
+			return nil, fmt.Errorf("equality field ID %d (%s) has unsupported nested path in %s", fieldID, fieldName, location)
 		}
 		result = structArray.Field(index)
 	}
@@ -330,7 +334,7 @@ func readEqualityDeleteFile(ctx context.Context, fs iceio.IO, tableSchema *icebe
 	for i, fid := range fieldIDs {
 		name, ok := tableSchema.FindColumnName(fid)
 		if !ok {
-			return nil, nil, fmt.Errorf("equality delete field ID %d not found in table schema", fid)
+			return nil, nil, fmt.Errorf("equality delete field ID %d not found in table schema for %s", fid, dataFile.FilePath())
 		}
 
 		ref, err := resolveArrowField(tbl.Schema(), fid, name, dataFile.FilePath())
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -23,7 +23,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"strings"
 	"unsafe"
 
 	"github.com/apache/arrow-go/v18/arrow"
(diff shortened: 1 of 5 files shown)


table/equality_delete_reader.go
```diff
@@ -217,11 +217,15 @@ func readEqualityDeleteFile(ctx context.Context, fs iceio.IO, tableSchema *icebe
 
 		colNames[i] = name
 		indices := tbl.Schema().FieldIndices(name)
-		if len(indices) == 0 {
+		switch len(indices) {
+		case 0:
 			return nil, nil, fmt.Errorf("equality delete column %q not found in delete file %s", name, dataFile.FilePath())
+		case 1:
+			colIndices[i] = indices[0]
+		default:
```
laskoviymishka: The fail-safe guard here is the right instinct, but we only apply it on the delete-file side. The mirror lookup in `processEqualityDeletesColumnar` (around line 550) does the same `r.Schema().FieldIndices(name)` and then takes `indices[0]` unconditionally against the data record, so a data file with a duplicate physical name for an equality column silently compares the wrong column and drops the wrong rows. That's the same ambiguity we're rejecting here, just on the read side, and it fails silently instead of loudly.

I'd apply the same guard at 550 so both paths fail the same way, and add a test that puts the duplicate column in the data file rather than the delete file, since the current test only exercises this branch. Once both sides are covered, this is good to land.

table/equality_delete_reader_test.go
```diff
@@ -119,6 +123,55 @@ func TestEqualityDeleteReadRoundTrip(t *testing.T) {
...
+	first := builder.NewArray()
+	builder.Append(2)
+	second := builder.NewArray()
+	builder.Release()
+	batch := array.NewRecordBatch(duplicateSchema, []arrow.Array{first, second}, 1)
+	first.Release()
+	second.Release()
+	deleteTable := array.NewTableFromRecords(duplicateSchema, []arrow.RecordBatch{batch})
+	batch.Release()
+	file, err := iceio.LocalFS{}.Create(deletePath)
```
laskoviymishka, line 357: `file` here is a WriteCloser but we never close it, and `pqarrow.WriteTable` writes without closing. That leaks the fd and, depending on buffering, can leave the parquet footer unflushed, so the later `tbl.Scan()` could read a truncated file and flake.

`defer file.Close()` right after the NoError check fixes it, or we could reuse the `writeParquetFile` helper with a custom schema. wdyt?

table/equality_delete_reader_test.go
```diff
@@ -119,6 +123,55 @@ func TestEqualityDeleteReadRoundTrip(t *testing.T) {
...
+	deleteBuilder.EqualityFieldIDs([]int{1})
+	tx = tbl.NewTransaction()
+	rd := tx.NewRowDelta(nil)
+	rd.AddDeletes(deleteBuilder.Build())
+	require.NoError(t, rd.Commit(t.Context()))
+	tbl, err = tx.Commit(t.Context())
+	require.NoError(t, err)
+
+	_, _, err = tbl.Scan().ToArrowRecords(t.Context())
+	require.ErrorContains(t, err, `equality delete column "id" is ambiguous`)
```
laskoviymishka: This asserts the literal error text, so a harmless rewording of the message would break the test with no behavior change. We already have `ErrEmptyEqualityFieldIDs` as a sentinel in this package, so I'd make the ambiguity error a sentinel too (`ErrAmbiguousEqualityColumn`, wrapped with `%w`) and assert with `require.ErrorIs`. Keeps the message free to change and reads the intent more clearly.

laskoviymishka (changes requested):
Nice catch on this one. Rejecting an ambiguous equality-delete column instead of silently grabbing the first match is exactly the right call, and the error message and test read cleanly.

I'd hold it before merging though, for one reason: the fix only covers the delete-file side. `processEqualityDeletesColumnar` does the same name-based `FieldIndices` lookup against the data record and takes `indices[0]` unconditionally, so a data file with a duplicate physical name for an equality column silently compares the wrong column and drops the wrong rows. That's the same ambiguity we're now rejecting, just on the read side, so I'd apply the guard symmetrically and add a test that puts the duplicate column in the data file rather than the delete file.

One thing worth thinking about, though I don't think it needs to block this PR: both lookups resolve columns by name, but equality deletes are matched by field id per the spec. A delete file written after a column rename still carries the right field id in its Parquet metadata but a stale name, so a name lookup drops the delete. Java resolves by field id throughout and avoids both this and the ambiguity. Larger change, probably a follow-up, but flagging the direction.

A couple of smaller things before merge:
- the delete file opened in the test is never closed, so it can leak the fd and leave the footer unflushed (possible flake)
- the test asserts the literal error string, so I'd make the error a sentinel (`ErrAmbiguousEqualityColumn`) and use `require.ErrorIs`, matching `ErrEmptyEqualityFieldIDs`

Once the data-record path is covered, happy to take another pass and approve.

table/equality_delete_reader.go
```diff
@@ -217,11 +220,15 @@ func readEqualityDeleteFile(ctx context.Context, fs iceio.IO, tableSchema *icebe
 
 		colNames[i] = name
 		indices := tbl.Schema().FieldIndices(name)
-		if len(indices) == 0 {
+		switch len(indices) {
```
zeroshade: This name-based ambiguity check false-rejects a valid, ID-unambiguous file. Two physical fields can have the same name but distinct IDs while `equality_ids` identifies exactly one; production Parquet-to-Arrow schemas retain `PARQUET:field_id`, so the authoritative identity is available here. `FieldIndices` is also exact, case-sensitive, and top-level-only, so it cannot resolve nested equality IDs represented as dotted Iceberg paths.

Suggested fix: recursively resolve the physical field ID first and require exactly one ID match, using the metadata visitor support in `table/arrow_utils.go`; use a unique exact-name fallback only when ID metadata is absent. Include the field ID in missing/ambiguous errors.

table/equality_delete_reader.go
```diff
@@ -551,6 +558,10 @@ func processEqualityDeletesColumnar(ctx context.Context, eqDeleteSets []*equalit
 				if len(indices) == 0 {
 					return nil, fmt.Errorf("equality delete column %q not found in data record", name)
 				}
+				if len(indices) > 1 {
```
zeroshade: This still false-accepts the opposite case. If the requested equality field was renamed or differs only by case while another physical field uniquely has the current table name, this code encodes that other field's values as delete keys without confirming its ID. That can silently delete the wrong rows.

Suggested fix: carry `eqDel.fieldIDs` into this path and resolve each data-record column by physical field ID, requiring exactly one match; use unique-name fallback only when ID metadata is absent. Include the equality field ID and data-file path in the error.

zeroshade (changes requested):
Thanks for tightening both equality-delete paths. The new sentinel and `%w` wrapping are appropriate, and the delete-file and data-record paths now fail closed for exact duplicate top-level names.

The remaining blocker is that equality-field identity is still resolved from names rather than the physical field IDs. That creates both failure directions: valid ID-unambiguous files are rejected, while a renamed or case-differing field can still bind to a different physical column and silently delete the wrong rows. The inline comments describe the ID-first resolution needed in each path.

The added tests prove the narrow exact-duplicate case, but coverage should also include duplicate names with distinct IDs, renamed fields with the correct ID, a unique name carrying the wrong ID, case-only collisions, nested equality fields with repeated leaf names, duplicate IDs under different names, and end-to-end duplicate data-file coverage. Including the equality field ID in missing/ambiguous errors would also make failures actionable.

There is no direct code conflict with #1611 or #1618; their ordering dependency is confined to `codec/file_scan_task.go`.

table/equality_delete_reader.go
```diff
@@ -547,12 +677,20 @@ func processEqualityDeletesColumnar(ctx context.Context, eqDeleteSets []*equalit
...
-				indices := r.Schema().FieldIndices(name)
-				if len(indices) == 0 {
-					return nil, fmt.Errorf("equality delete column %q not found in data record", name)
+				fieldID := 0
+				if i < len(eqDel.fieldIDs) {
+					fieldID = eqDel.fieldIDs[i]
 				}
 
-				encoders[i] = makeColEncoder(r.Column(indices[0]))
+				ref, err := resolveArrowField(r.Schema(), fieldID, name, dataFilePath)
```
laskoviymishka: I'd resolve these once before returning the closure rather than per batch. `resolveArrowField` runs on every record batch here, but the schema is constant across all batches of a file, so we're repeating the full recursive walk with identical inputs each time, and the recursive struct descent makes that more expensive than the old flat `FieldIndices` scan it replaced.

`readEqualityDeleteFile` already does this the right way: it resolves the `fieldRefs` once up front and reuses them in the batch loop. I'd mirror that here, computing the refs above the `return func(...)` and closing over them. wdyt?

table/equality_delete_reader.go
```diff
@@ -547,12 +677,20 @@ func processEqualityDeletesColumnar(ctx context.Context, eqDeleteSets []*equalit
 		for _, eqDel := range eqDeleteSets {
 			encoders := make([]colEncoder, len(eqDel.colNames))
 			for i, name := range eqDel.colNames {
-				indices := r.Schema().FieldIndices(name)
-				if len(indices) == 0 {
-					return nil, fmt.Errorf("equality delete column %q not found in data record", name)
+				fieldID := 0
```
laskoviymishka: This defensive guard quietly makes the ID-first resolution optional. When `i >= len(eqDel.fieldIDs)`, `fieldID` stays `0`, so `resolveArrowField` skips the `fieldID > 0` branch entirely and falls back to name-only lookup, which is the behavior this PR is trying to get away from.

Today `fieldIDs` and `colNames` are always co-built at equal length so this never trips, but if that invariant ever slips we'd silently lose ID matching with no error to point at. I'd drop the guard and index `eqDel.fieldIDs[i]` directly, or assert the two are equal-length at construction, so a mismatch fails loudly instead of degrading. wdyt?

table/equality_delete_reader.go
```diff
@@ -46,6 +50,124 @@ type equalityDeleteSet struct {
...
+	location := filePath
+	if location == "" {
+		location = "data record"
+	}
+
+	if fieldID > 0 {
+		switch len(idMatches) {
+		case 1:
+			return idMatches[0].ref, nil
+		case 0:
```
laskoviymishka: The empty `case 0:` reads like a dropped line. It took me a second to see that zero ID matches is meant to fall through to the name-based resolution below rather than being an unfinished branch, and revive's empty-block check will likely flag it too.

I'd add a `// zero ID matches, fall through to name resolution` line inside the case, or drop the switch for explicit `if len(idMatches) == 1 {...}` / `if len(idMatches) > 1 {...}`. Either's fine, just something that signals the fall-through is intentional.

table/equality_delete_reader.go
```diff
@@ -350,7 +476,7 @@ func encodeArrowValue(buf *bytes.Buffer, arr arrow.Array, idx int) {
 // rows whose equality key columns match any entry in the delete sets.
 // Each set is applied independently (they may have different field IDs).
 func processEqualityDeletes(ctx context.Context, eqDeleteSets []*equalityDeleteSet) (recProcessFn, error) {
-	return processEqualityDeletesColumnar(ctx, eqDeleteSets)
+	return processEqualityDeletesColumnarForFile(ctx, eqDeleteSets, "")
```
laskoviymishka: Now that `arrow_scanner.go` calls `processEqualityDeletesColumnarForFile` directly, this wrapper (and `processEqualityDeletesColumnar` just below) has no production callers left: `processEqualityDeletes` is dead, and `processEqualityDeletesColumnar` only survives because the internal test calls it.

I'd collapse both into `processEqualityDeletesColumnarForFile` and point the test at it with a synthetic path. That keeps one entry point and stops the bench in `equality_delete_reader_bench_test.go` from quietly exercising the empty-path variant instead of the real one.

laskoviymishka (changes requested):
Almost there. The direction I flagged last round is now the actual implementation: resolution goes field-ID-first through `resolveArrowField`, and the ambiguity guard is symmetric across both the delete-file and data-record paths. That's more than I asked for, and the ID-first fallback logic itself reads correctly to me.

One thing I'd fix before merge. `resolveArrowField` runs inside the per-batch closure on the data-record path, so the full recursive schema walk repeats for every record batch even though the schema is constant across a file. The delete-file path already resolves once up front, and I'd hoist the data-record resolution out of the closure the same way so it's computed once per file rather than once per batch.

A couple of smaller cleanups while we're here. The `fieldID := 0` guard silently falls back to name-only resolution if `fieldIDs` and `colNames` ever get out of sync, which quietly defeats the ID check we just added, so I'd index directly or assert equal length. And now that `arrow_scanner.go` calls `...ForFile` directly, `processEqualityDeletes` and `processEqualityDeletesColumnar` are dead wrappers worth collapsing.

On tests, the resolver-level matrix in `TestResolveArrowFieldRejectsAmbiguousOrMismatchedIDs` covers most of what I'd have wanted: renamed field with the matching ID, unique name with the wrong ID, duplicate IDs, repeated nested leaf names, and the case collision all read well. The gap left is end to end. Those exercise `resolveArrowField` directly, and the batch-level `TestProcessEqualityDeletesRejectsAmbiguousDataColumns` uses fields with no ID metadata, so nothing drives the full scan through the ID-first path. A scan test with a data file carrying the field ID under the old physical name would close that, but I'll defer to zeroshade's thread on the exact matrix rather than pile on here.

- resolve the data-record columns once per file, not per batch
- drop the `fieldID = 0` fallback so a length mismatch fails loudly
- collapse the dead `processEqualityDeletes` wrappers

Fix the per-batch resolution and I'm happy to approve.

zeroshade (approved):
This is now the right design: equality-field identity resolved by field ID against a structural per-file index (embedded IDs first, name mapping for legacy files), applied symmetrically to delete files and data records, resolved once per file with only encoders rebuilt per batch. The failure modes from the earlier rounds — false-rejecting ID-unambiguous duplicates and silently binding renamed/case-differing columns — are both closed and pinned by tests, and the mismatched-metadata guard failing loudly instead of quietly downgrading to name resolution is exactly right.

Cosmetic only: `arrowArraysAtField` re-inlines the `"data record"` location fallback instead of using the `equalityFieldLocation` helper defined just above. Fine to leave.

---

table/equality_delete_reader.go
```diff
@@ -46,11 +49,140 @@ type equalityDeleteSet struct {
...
+	}
+	if len(matches) > 1 {
+		return arrowFieldRef{}, fmt.Errorf("%w: equality field ID %d (%s) in %s: found %d fields",
+			ErrAmbiguousEqualityColumn, fieldID, fieldName, location, len(matches))
+	}
+
+	return arrowFieldRef{}, fmt.Errorf("equality field ID %d (%s) not found in %s", fieldID, fieldName, location)
+}
+
+func arrowArrayAtField(record arrow.RecordBatch, ref arrowFieldRef, fieldID int, fieldName, filePath string) (arrow.Array, error) {
```
laskoviymishka, line 128: `arrowArrayAtField` isn't called anywhere. `makeArrowFieldEncoder` uses `arrowArraysAtField` directly, so I'd just drop it while we're here.

table/equality_delete_reader.go
```diff
@@ -205,23 +337,50 @@ func readEqualityDeleteFile(ctx context.Context, fs iceio.IO, tableSchema *icebe
...
+		return nil, nil, err
+	}
+
+	var fileSchema *iceberg.Schema
+	if !hasFieldIDs {
+		if nameMapping == nil {
+			nameMapping = tableSchema.NameMapping()
+		}
+
+		fileSchema, err = ArrowSchemaToIcebergWithOptions(tbl.Schema(), ArrowToIcebergOptions{
```
laskoviymishka, line 351: On the `!hasFieldIDs` path we convert the delete file's full Arrow schema through the name mapping, so an extra context column whose physical name isn't in the mapping fails the whole file. The old lookup only resolved the specific equality columns, so this is a narrow regression for legacy files that dropped a column.

Low-probability and not blocking; I'd project down to just the equality columns before the conversion in a follow-up. wdyt?

laskoviymishka (approved):
That's everything from last round. The data-record resolution is hoisted out of the per-batch closure in `processEqualityDeletesColumnarForFile` so it's resolved once per file now, the `fieldID = 0` fallback is gone with a length-mismatch check that fails loudly on `ErrInvalidArgument`, and the dead `processEqualityDeletes` wrappers are collapsed. Reads exactly how I hoped.

Two small things, neither blocking. `arrowArrayAtField` isn't called anywhere anymore, so I'd drop it. And on the no-embedded-IDs path we run the delete file's full Arrow schema through the name mapping, which can fail a legacy file that carries an extra context column not in the mapping. The old lookup only touched the equality columns, so it's a narrow regression, fine as a follow-up. I'd also agree with zeroshade that the `arrowArraysAtField` location fallback re-inlining `equalityFieldLocation` is fine to leave.

Approving. Thanks for staying with this one across the rounds.

PR #1618, opened 2026-07-30
fix(codec): validate scan ranges against file size

fallintoplace:
## Summary

Validate FileScanTask byte ranges produced by the codec without making decode stricter than the interchange format.

## Why

The encoder should not produce negative or out-of-file ranges. Decode still rejects negative values, but Java and other implementations may send a range whose end is beyond the declared file size, so decode should preserve that task rather than add a client-specific EOF rule.

## What changed

- Encode validates that the range fits within the primary data file.
- Encode and decode reject negative starts and lengths.
- The encode-time check uses subtraction so start plus length cannot overflow.
- Decode does not compare the range with file size.

## Tests

- Added full-file, suffix, and zero-length-at-EOF cases
- Added beyond-EOF and integer-overflow encode cases
- Added decode coverage for a range beyond file size
- Added malformed-envelope decode coverage
- go test ./codec

### DIFF
--- a/codec/file_scan_task.go
+++ b/codec/file_scan_task.go
@@ -135,9 +135,6 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 	if err != nil {
 		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: file: %w", err)
 	}
-	if err := validateScanRange(envelope.Start, envelope.Length, file.FileSizeBytes()); err != nil {
-		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: %w", err)
-	}
 	del, err := decodeDataFileSlice(envelope.DeleteFiles, spec, schema, version)
 	if err != nil {
 		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: delete files: %w", err)
--- a/codec/file_scan_task.go
+++ b/codec/file_scan_task.go
@@ -58,6 +58,9 @@ func EncodeFileScanTask(task table.FileScanTask, spec iceberg.PartitionSpec, sch
 	if err := checkDataFileSpecID(task.File, spec); err != nil {
 		return nil, fmt.Errorf("codec: EncodeFileScanTask: %w", err)
 	}
+	if err := validateScanRange(task.Start, task.Length, task.File.FileSizeBytes()); err != nil {
+		return nil, fmt.Errorf("codec: EncodeFileScanTask: %w", err)
+	}
 	fileBytes, err := EncodeDataFile(task.File, spec, schema, version)
 	if err != nil {
 		return nil, fmt.Errorf("codec: EncodeFileScanTask: file: %w", err)
@@ -132,6 +135,9 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 	if err != nil {
 		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: file: %w", err)
 	}
+	if err := validateScanRange(envelope.Start, envelope.Length, file.FileSizeBytes()); err != nil {
+		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: %w", err)
+	}
 	del, err := decodeDataFileSlice(envelope.DeleteFiles, spec, schema, version)
 	if err != nil {
 		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: delete files: %w", err)
@@ -160,6 +166,14 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 		FirstRowID:          envelope.FirstRowID,
 		DataSequenceNumber:  envelope.DataSequenceNumber,
 	}, nil
+}
+
+func validateScanRange(start, length, fileSize int64) error {
+	if start > fileSize || length > fileSize-start {
+		return fmt.Errorf("scan range start=%d length=%d exceeds file size %d", start, length, fileSize)
+	}
+
+	return nil
 }
 
 // fileScanTaskShape is a compile-time drift guard for FileScanTask.
--- a/codec/file_scan_task.go
+++ b/codec/file_scan_task.go
@@ -169,6 +169,8 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 }
 
 func validateScanRange(start, length, fileSize int64) error {
+	// Check the end as a subtraction after confirming start <= fileSize so
+	// start+length cannot overflow int64.
 	if start > fileSize || length > fileSize-start {
 		return fmt.Errorf("scan range start=%d length=%d exceeds file size %d", start, length, fileSize)
 	}
--- a/codec/file_scan_task.go
+++ b/codec/file_scan_task.go
@@ -43,6 +43,8 @@ import (
 // partition spec than the data file; the caller is responsible for
 // partitioning the FileScanTask by per-file specID and calling
 // EncodeFileScanTask once per group.
+// The scan range is checked against the DataFile's recorded file size, which
+// is manifest metadata rather than a filesystem stat.
 func EncodeFileScanTask(task table.FileScanTask, spec iceberg.PartitionSpec, schema *iceberg.Schema, version int) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


codec/file_scan_task_internal_test.go
```diff
@@ -19,12 +19,37 @@ package codec
...
+		name              string
+		start, length     int64
+		fileSize          int64
+		shouldReturnError bool
+	}{
+		{name: "full file", length: 100, fileSize: 100},
+		{name: "empty range at EOF", start: 100, fileSize: 100},
+		{name: "start after EOF", start: 101, fileSize: 100, shouldReturnError: true},
+		{name: "end after EOF", start: 99, length: 2, fileSize: 100, shouldReturnError: true},
+		{name: "overflowing end", start: math.MaxInt64, length: 1, fileSize: 100, shouldReturnError: true},
```
laskoviymishka: this case doesn't actually exercise the overflow-safe branch. `start=math.MaxInt64` trips `start > fileSize` first, so the `length > fileSize-start` subtraction never runs - which is the whole reason we wrote it as a subtraction instead of `start+length > fileSize`.

If someone "simplified" the helper to `start+length > fileSize`, this test would still pass (`MaxInt64+1` wraps negative, and `start > fileSize` still fires on the first clause). The case that actually pins the subtraction form is one where the sum overflows but `start <= fileSize` holds:

```go
{name: "start+length overflows int64", start: math.MaxInt64 - 1, length: 2, fileSize: math.MaxInt64, shouldReturnError: true},
```

There, naive `start+length` wraps to a negative and passes, while the subtraction form correctly rejects. Same gap in the `overflowing range` case over in `file_scan_task_test.go`. wdyt?

codec/file_scan_task.go
```diff
@@ -132,6 +135,9 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 	if err != nil {
 		return table.FileScanTask{}, fmt.Errorf("codec: DecodeFileScanTask: file: %w", err)
 	}
+	if err := validateScanRange(envelope.Start, envelope.Length, file.FileSizeBytes()); err != nil {
```
laskoviymishka: I'm a little torn on the decode-time guard. For any task we encoded ourselves it's tautological - `EncodeFileScanTask` already validated against the same `FileSizeBytes()` that gets baked into the DataFile blob, so this can only ever fire on a foreign encoder.

That's the part I'd think about for #1178: once REST scan planning lands, a remote Java/PyIceberg planner could hand us a task where `length` and the manifest `file_size_in_bytes` came from different sources, and we'd reject something the caller legitimately wants. No other client enforces this at decode, and the spec doesn't bound start/length either.

Not blocking, and I'd keep the encode-time check regardless as an assertion on our own output. But I'd either drop a comment noting decode is a defensive coherence guard (not spec-mandated) or consider demoting it to a warning. wdyt?

codec/file_scan_task.go
```diff
@@ -162,6 +168,14 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 	}, nil
 }
 
+func validateScanRange(start, length, fileSize int64) error {
+	if start > fileSize || length > fileSize-start {
```
laskoviymishka, line 178: Could we add a one-line note on why this is written as a subtraction rather than `start + length > fileSize`? The `||` short-circuit is load-bearing here - `fileSize - start` only runs once `start <= fileSize` (leaning on the caller's `start >= 0` guard), which dodges both the underflow and the `start + length` overflow. Without a comment it reads as trivially simplifiable, and nothing in the tests catches that rewrite today (see the overflow case above).

codec/file_scan_task_internal_test.go
```diff
@@ -110,3 +135,19 @@ func TestDecodeFileScanTaskRejectsNegativeScanRanges(t *testing.T) {
...
+	require.NoError(t, err)
+	file, err := EncodeDataFile(builder.Build(), spec, nil, 2)
+	require.NoError(t, err)
+	envelope, err := fileScanTaskSchema.Encode(&fileScanTaskEnvelope{
+		File: file, Start: 90, Length: 11,
+	})
+	require.NoError(t, err)
+
+	_, err = DecodeFileScanTask(envelope, spec, nil, 2)
+	require.ErrorContains(t, err, "scan range start=90 length=11 exceeds file size 100")
```
laskoviymishka: minor: this pins the full formatted message including the numbers, while everything else in the suite (and the encode-side test) matches the looser `"exceeds file size"`. If we ever tweak the format string this fails for no real reason - I'd loosen it to `require.ErrorContains(t, err, "exceeds file size")`.

laskoviymishka (changes requested):
The validation logic is correct, and placing it after the negative-range guards on both paths is the right call.

Almost there - one thing I'd want fixed before merge: the overflow test doesn't actually test overflow. With `start=math.MaxInt64` the `start > fileSize` clause fires first, so the `length > fileSize-start` subtraction never runs, and a naive rewrite to `start+length > fileSize` would still pass. A case like `start=math.MaxInt64-1, length=2, fileSize=math.MaxInt64` pins the subtraction form (same for the external test's `overflowing range` case).

Everything else is minor and non-blocking - loosening the verbatim error assertion, a one-line comment on why the short-circuit is safe, and a question worth chewing on: whether the decode-time guard should stay once #1178 starts feeding us tasks from foreign planners. I left those inline.

Fix the overflow test and I'm happy to approve.

codec/file_scan_task.go
```diff
@@ -58,6 +58,9 @@ func EncodeFileScanTask(task table.FileScanTask, spec iceberg.PartitionSpec, sch
 	if err := checkDataFileSpecID(task.File, spec); err != nil {
 		return nil, fmt.Errorf("codec: EncodeFileScanTask: %w", err)
 	}
+	if err := validateScanRange(task.Start, task.Length, task.File.FileSizeBytes()); err != nil {
```
zeroshade, line 63: Non-blocking: `FileSizeBytes()` is the manifest's recorded `file_size_in_bytes` (`manifest.go:2042,2208`), not a filesystem stat. Iceberg's immutable-file model normally makes it authoritative, but stale or incorrectly small metadata can reject a physically readable file during encode. Consider documenting that tradeoff.

codec/file_scan_task.go
```diff
@@ -162,6 +165,16 @@ func DecodeFileScanTask(data []byte, spec iceberg.PartitionSpec, schema *iceberg
 	}, nil
 }
 
+func validateScanRange(start, length, fileSize int64) error {
```
zeroshade, line 177: Non-blocking: This helper relies on both callers checking negative start and length first. Consider documenting that precondition or consolidating those checks here so a future caller cannot accidentally make `fileSize-start` the only guard.

codec/file_scan_task_internal_test.go
```diff
@@ -19,12 +19,37 @@ package codec
...
 import (
 	"encoding/binary"
+	"math"
 	"testing"
 
 	"github.com/apache/iceberg-go"
 	"github.com/stretchr/testify/require"
 )
 
+func TestValidateScanRange(t *testing.T) {
```
zeroshade, line 29: Non-blocking: Consider rounding out the boundary matrix with `(fileSize, 1)`, `(MaxInt64, 0, MaxInt64)`, `(MaxInt64-1, 1, MaxInt64)`, and `(0, 0)`, plus invalid or non-positive sizes from a custom `DataFile`.

zeroshade (commented):
This change is correct. I verified that encode-side range validation is off-by-one correct and overflow-safe: exact EOF endpoints and the empty range at EOF are accepted, while starts or ends beyond EOF are rejected using subtraction rather than `start + length`. Decode intentionally remains permissive for foreign encodings.

Non-blocking integration and scope notes:

- When this is combined with #1611, keep #1611's nil/typed-nil file guard before `codec/file_scan_task.go:61` calls `FileSizeBytes()`; that ordering preserves the panic fix.
- `codec/file_scan_task.go:128-165` rejects negative decoded values but intentionally accepts a range beyond the declared file size. Consider documenting that decode/re-encode can therefore fail and that consumers honoring ranges need their own validation.
- The local Arrow path still reads the whole file: `table/arrow_scanner.go:982` calls `GetRecords` without the task range, and `table/arrow_scanner.go:1054` prepares the reader from only the file. This is pre-existing and outside this PR, but split tasks can consequently duplicate full-file results and deserve a separate correctness issue.

laskoviymishka (approved):
LGTM now

PR #1619, opened 2026-07-30
fix(table): keep generated snapshot IDs non-negative

fallintoplace:
## What changed

Clear the sign bit before converting the UUID-derived snapshot ID to `int64`. The UUID folding logic is extracted into a small deterministic helper for regression coverage.

## Why

Negating a negative value does not make `math.MinInt64` positive because signed negation overflows back to the same value. Clearing the sign bit guarantees that generated snapshot IDs are nonnegative for every input.

## Testing

- Added deterministic coverage for an XOR result with only the sign bit set
- Added coverage showing the remaining value bits are preserved
- `go test ./table`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -45,22 +45,18 @@ const (
 )
 
 func generateSnapshotID() int64 {
-	var (
-		rndUUID = uuid.New()
-		out     [8]byte
-	)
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):

laskoviymishka (approved):
:shipit:

PR #1621, opened 2026-07-31
fix(cli): report correct entity type for missing property

jx2lee:
## Summary
- `properties get` always reported "on namespace" in its error message even when the target was a table, since the message was hardcoded instead of using the actual entity type.
- Switched to `fmt.Errorf` with `get.Type` so the message reflects "namespace" or "table" correctly.
- Add unit test verifying `properties get` reports correct entity type on missing property

## Test plan
`go test ./cmd/iceberg/`

Ran against a local hadoop catalog, requesting a non-existent property:

**Before**
```
$ iceberg properties get table ns1.t1 nonexistent_prop
could not find property nonexistent_prop on namespace ns1.t1   # wrong: says "namespace" for a table
```

**After**
```
$ iceberg properties get table ns1.t1 nonexistent_prop
could not find property nonexistent_prop on table ns1.t1       # correct

$ iceberg properties get namespace ns1 nonexistent_prop
could not find property nonexistent_prop on namespace ns1      # unchanged, still correctn

### DIFF
--- a/cmd/iceberg/main.go
+++ b/cmd/iceberg/main.go
@@ -652,8 +652,8 @@ func runProperties(ctx context.Context, output Output, cat catalog.Catalog, cmd 
 		if val, ok := props[get.PropName]; ok {
 			output.Text(val)
 		} else {
-			output.Error(errors.New("could not find property " + get.PropName + " on namespace " + get.Identifier))
-			os.Exit(1)
+			output.Error(fmt.Errorf("could not find property %s on %s %s", get.PropName, get.Type, get.Identifier))
+			osExit(1)
 		}
 	case cmd.Set != nil:
 		set := cmd.Set
--- a/cmd/iceberg/properties_test.go
+++ b/cmd/iceberg/properties_test.go
@@ -1,0 +1,156 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: cmd/iceberg/properties_test.go [hunk 1]
```diff
@@ -0,0 +1,156 @@
...
+
+func (m *mockCatalogForProperties) CheckNamespaceExists(context.Context, table.Identifier) (bool, error) {
+	panic("CheckNamespaceExists must not be called")
+}
+
+func (m *mockCatalogForProperties) UpdateNamespaceProperties(context.Context, table.Identifier, []string, iceberg.Properties) (catalog.PropertiesUpdateSummary, error) {
+	panic("UpdateNamespaceProperties must not be called")
+}
+
+func TestRunPropertiesGetNamespaceNotFound로(t *testing.T) {
```
tanmayrauth: The test name `TestRunPropertiesGetNamespaceNotFound로` has a stray Korean character `로` on the end — looks like an accidental keystroke. It compiles and runs (Go allows Unicode identifiers), but it shows up as garbage in CI logs and when grepping test names, and it's inconsistent with the sibling `TestRunPropertiesGetTableNotFound`. Rename to `TestRunPropertiesGetNamespaceNotFound`.

cmd/iceberg/main.go
```diff
@@ -652,8 +652,10 @@ func runProperties(ctx context.Context, output Output, cat catalog.Catalog, cmd
 		if val, ok := props[get.PropName]; ok {
 			output.Text(val)
 		} else {
-			output.Error(errors.New("could not find property " + get.PropName + " on namespace " + get.Identifier))
-			os.Exit(1)
+			output.Error(fmt.Errorf("could not find property %s on %s %s", get.PropName, get.Type, get.Identifier))
+			osExit(1)
+
+			return
```
tanmayrauth: The `return` here is unreachable — `osExit` is `os.Exit` in prod (never returns) and panics in tests, and this is already the end of the case. Harmless, but you can drop it.

tanmayrauth (commented):
Fix is correct — validateEntityType guarantees get.Type is only ever "namespace" or "table", so the interpolated message is safe. Couple of nits inline.

tanmayrauth (approved):

laskoviymishka (approved):
:shipit:

PR #1624, opened 2026-08-01
fix(table): guard DV filtering against stale row counts

fallintoplace:
## Summary

Prevent deletion-vector filtering from panicking when a data file emits more rows than its manifest `record_count`.

## Problem

The contiguous fast path builds a keep mask sized from `DataFile.Count()` and slices it using the rows emitted by each Arrow batch. If the manifest count is stale or otherwise smaller than the physical file, the batch end falls outside the mask and `array.NewSlice` panics.

The row-group-pruned path already handles this mismatch by preserving positions beyond the mask. The contiguous path should behave the same way.

## Fix

Keep the existing zero-copy mask slice while a batch is within the declared row count. If a batch crosses that boundary, build a bounded mask for that batch and keep rows whose positions are beyond the declared count. Once a batch is entirely beyond the mask, pass it through without allocating an all-true filter. In-range deletion-vector positions are still applied normally.

## Tests

Added coverage across three batches: one within the manifest count, one crossing it, and one entirely beyond it. The test verifies in-range deletes while preserving rows beyond the declared count.

```
go test ./table -count=1
```

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -576,6 +576,11 @@ func filterByDeletionVector(ctx context.Context, bitmap *dv.RoaringPositionBitma
 
 				return compute.FilterRecordBatch(ctx, r, sliced, compute.DefaultFilterOptions())
 			}
+			if currentIdx >= rowCount {
+				r.Retain()
+
+				return r, nil
+			}
 
 			// A stale manifest count can be smaller than the rows emitted by
 			// the file. Preserve rows beyond the mask rather than slicing past
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -560,16 +560,36 @@ func filterByDeletionVector(ctx context.Context, bitmap *dv.RoaringPositionBitma
 			currentIdx := nextIdx
 			nextIdx += nrows
 
-			// Wrap (and slice) the shared keep-mask buffer for this batch.
-			// array.NewSlice on a Boolean array tracks the bit-level offset,
-			// so we don't need byte-aligned slicing — currentIdx can land
-			// anywhere within a byte.
-			full := array.NewBoolean(int(rowCount), buf, nil, 0)
-			defer full.Release()
-			sliced := array.NewSlice(full, currentIdx, nextIdx).(*array.Boolean)
-			defer sliced.Release()
+			if nextIdx <= rowCount {
+				// Wrap (and slice) the shared keep-mask buffer for this batch.
+				// array.NewSlice on a Boolean array tracks the bit-level offset,
+				// so we don't need byte-aligned slicing — currentIdx can land
+				// anywhere within a byte.
+				full := array.NewBoolean(int(rowCount), buf, nil, 0)
+				defer full.Release()
+				sliced := array.NewSlice(full, currentIdx, nextIdx).(*array.Boolean)
+				defer sliced.Release()
 
-			return compute.FilterRecordBatch(ctx, r, sliced, compute.DefaultFilterOptions())
+				return compute.FilterRecordBatch(ctx, r, sliced, compute.DefaultFilterOptions())
+			}
+
+			// A stale manifest count can be smaller than the rows emitted by
+			// the file. Preserve rows beyond the mask rather than slicing past
+			// its bounds and panicking.
+			bldr := array.NewBooleanBuilder(mem)
+			defer bldr.Release()
+			bldr.Reserve(int(nrows))
+			for pos := currentIdx; pos < nextIdx; pos++ {
+				if pos >= rowCount {
+					bldr.Append(true)
+				} else {
+					bldr.Append(keepBits[pos>>3]&(1<<(uint(pos)&7)) != 0)
+				}
+			}
+			mask := bldr.NewBooleanArray()
+			defer mask.Release()
+
+			return compute.FilterRecordBatch(ctx, r, mask, compute.DefaultFilterOptions())
 		}
 
 		bldr := array.NewBooleanBuilder(mem)
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -530,6 +530,10 @@ func processPositionalDeletes(ctx context.Context, deletes set[int64], cursor *r
 	}
 }
 
+func deletionVectorKeepsRow(keepBits []byte, rowCount, pos int64) bool {
+	return pos >= rowCount || keepBits[pos>>3]&(1<<(uint(pos)&7)) != 0
+}
+
 // filterByDeletionVector returns a pipeline step that drops rows present in
 // the bitmap by precomputing a bit-packed keep-mask covering the whole file
 // once and slicing the relevant range per batch into compute.FilterRecordBatch.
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):

table/arrow_scanner.go
```diff
@@ -530,6 +530,10 @@ func processPositionalDeletes(ctx context.Context, deletes set[int64], cursor *r
 	}
 }
```
laskoviymishka, line 532: When we pulled this out of the slow-path loop it lost the bit-layout note that used to sit above it. Could we keep a one-liner here — byte `pos>>3`, bit `pos&7`, LSB-first (the layout `array.NewBoolean` reads in the fast path) — so the next reader doesn't wonder whether the `uint(pos)` cast is guarding negative positions?

table/arrow_scanner.go
```diff
@@ -560,16 +564,37 @@ func filterByDeletionVector(ctx context.Context, bitmap *dv.RoaringPositionBitma
...
+				// anywhere within a byte.
+				full := array.NewBoolean(int(rowCount), buf, nil, 0)
+				defer full.Release()
+				sliced := array.NewSlice(full, currentIdx, nextIdx).(*array.Boolean)
+				defer sliced.Release()
+
+				return compute.FilterRecordBatch(ctx, r, sliced, compute.DefaultFilterOptions())
+			}
+			if currentIdx >= rowCount {
+				r.Retain()
```
laskoviymishka, line 580: This `Retain` is the one genuinely subtle line in the change — it's counteracting the unconditional `defer r.Release()` at the top of the closure so the caller ends up with the batch at its original refcount.

Without a comment I think a future reader could easily read it as a leak and "fix" it by deleting the `Retain`, which turns the pending `Release` into a use-after-free that won't surface until GC or the sanitizer. I'd drop a one-line note right here to pin that down.

table/arrow_scanner.go
```diff
@@ -560,16 +564,37 @@ func filterByDeletionVector(ctx context.Context, bitmap *dv.RoaringPositionBitma
...
+
+				return r, nil
+			}
+
+			// A stale manifest count can be smaller than the rows emitted by
```
laskoviymishka, line 585: A stale `record_count` almost always means the writer emitted more physical rows than the manifest declares — a real defect in whatever wrote the file. We handle it correctly, but silently, so an operator has no way to find the offending file.

Would it be worth a one-time `slog.Warn` (data file path, `rowCount`, the batch range) the first time we cross this boundary, guarded so it fires at most once per file rather than per batch? `dv/deletion_vector.go` already warns on the missing-cardinality case, so there's precedent for treating this as a writer defect worth surfacing. wdyt?

table/dv_scanner_read_test.go
```diff
@@ -347,3 +347,39 @@ func TestFilterByDeletionVectorOutOfBoundsPosition(t *testing.T) {
...
 }
+
+func TestFilterByDeletionVectorStaleRowCount(t *testing.T) {
+	ctx := context.Background()
+	mem := memory.NewGoAllocator()
+
+	bitmap := dv.NewRoaringPositionBitmap()
+	bitmap.Set(1)
+	bitmap.Set(3)
+	filter := filterByDeletionVector(ctx, bitmap, 4, (&rowPositionSource{}).cursor())
```
laskoviymishka: Two edges of the three-way split aren't pinned by these batches. The test steps from `nextIdx=3` (case 1) straight to `nextIdx=6` (straddle), so neither exact boundary — `nextIdx==rowCount` nor `currentIdx==rowCount` — ever gets hit. Those conditions are `<=` / `>=`, and that's load-bearing: the `<=` is what keeps `array.NewSlice(full, currentIdx, rowCount)` valid when a batch ends exactly on `rowCount`. If a refactor tightened either to a strict inequality, a batch landing on the boundary would silently take the wrong branch and still produce the right rows here. A two-batch variant where batch 1 ends exactly at `rowCount` and batch 2 starts there would catch that.

Separately, `rowCount=0` (a fully stale manifest) is a valid value that's untested — every batch should fall to the pass-through branch. Worth one more case.

table/dv_scanner_read_test.go
```diff
@@ -347,3 +347,39 @@ func TestFilterByDeletionVectorOutOfBoundsPosition(t *testing.T) {
...
+
+		return array.NewRecordBatch(schema, []arrow.Array{col}, int64(len(values)))
+	}
+
+	withinCount, err := filter(mkBatch(0, 1, 2))
+	require.NoError(t, err)
+	defer withinCount.Release()
+	assert.Equal(t, []int64{0, 2}, withinCount.Column(0).(*array.Int64).Int64Values())
+
+	beyondCount, err := filter(mkBatch(3, 4, 5))
```
laskoviymishka: I think this straddle case passes even against a subtly-wrong implementation. In `mkBatch(3, 4, 5)` the only in-range position is 3, and 3 is deleted — so 4 and 5 (beyond `rowCount`) are the only kept rows, and the assertion is `[]int64{4, 5}`. Nothing here exercises an in-range position that should be *kept*: a straddle impl that just dropped every in-range row would produce the same `[4, 5]` and pass all three assertions.

I'd reshape so the straddle batch carries an in-range keep, e.g. batch1 = `(0,1)` [case 1], batch2 = `(2,3,4,5)` [straddle]: pos 2 in-range and kept, 3 in-range and deleted, 4/5 beyond and kept → expect `[2, 4, 5]`. That pins the branch that actually matters.

table/dv_scanner_read_test.go
```diff
@@ -347,3 +347,39 @@ func TestFilterByDeletionVectorOutOfBoundsPosition(t *testing.T) {
...
+	require.NoError(t, err)
+	defer beyondCount.Release()
+	assert.Equal(t, []int64{4, 5}, beyondCount.Column(0).(*array.Int64).Int64Values())
+
+	fullyBeyondCount, err := filter(mkBatch(6, 7))
```
laskoviymishka: Small thing, but these two names are inverted: `beyondCount` above is actually the straddle result (`currentIdx 3 < rowCount 4 < nextIdx 6`), while this `mkBatch(6, 7)` case is the fully-beyond one. I'd rename to something like `straddleResult` / `beyondResult` so the intent reads straight — the value/position coincidence (values happening to equal absolute positions) already makes this test a little hard to follow, so anything that names the branch helps.

laskoviymishka (changes requested):
Nice fix — the guard is the right instinct, and breaking the closure into fast-path / straddle / fully-beyond reads cleanly. Spec- and parity-wise it checks out: rows past a stale `record_count` have no position in the bitmap by construction, so keeping them matches V3 and what Java and PyIceberg do.

I'd tighten the test before we merge, though. The straddle case passes even against a subtly-wrong impl — the only in-range position in `mkBatch(3,4,5)` is deleted, so nothing exercises the in-range-kept path that branch exists for. Reshaping so the straddle batch carries an in-range keep (e.g. batch2 = `(2,3,4,5)` → expect `[2,4,5]`) fixes that.

While we're in there, the exact edges of the split — a batch landing right on `rowCount`, and `rowCount=0` — aren't pinned by any case, and those `<=`/`>=` conditions are load-bearing. A couple of small additions would turn this into a test that'd actually catch a regression.

Left a few smaller notes inline — mainly a comment on the `r.Retain()` line, which is the one genuinely subtle spot. Once the straddle test exercises the in-range keep, happy to take another pass and approve.

laskoviymishka (approved):

PR #1626, opened 2026-08-01
fix(udf): replace state during JSON decoding

fallintoplace:
## Summary

Make UDF JSON decoding replace receiver state instead of merging into it.

## Problem

Several UDF `UnmarshalJSON` implementations decode through aliases of their existing receivers. Go leaves struct fields unchanged when they are absent from a new JSON document, so reusing a receiver can retain values from an earlier decode.

At the metadata level, a malformed document without `definitions` can pass validation using retained definitions, and the lazy definition index can remain stale. Exported nested values such as `Parameter`, `DefinitionVersion`, `Definition`, and `DefinitionLogEntry` also retain omitted optional fields when decoded directly.

## Fix

Decode each value into fresh state and assign it to the receiver only after decoding succeeds. Top-level metadata is also validated before assignment, then initializes its definition index from the replacement state.

## Tests

Added receiver-reuse coverage for top-level metadata and exported nested values. The tests verify omitted optional fields are cleared, cached indexes are replaced, and failed metadata decoding does not partially overwrite the last valid state.

```
go test ./udf -count=1
```

### DIFF
--- a/udf/metadata.go
+++ b/udf/metadata.go
@@ -870,14 +870,13 @@ func (m *metadata) init() {
 
 func (m *metadata) UnmarshalJSON(b []byte) error {
 	type Alias metadata
+	next := metadata{FormatVersionValue: -1}
 	// definition-log is shadowed to reject the required field being absent
 	// or null, which the default decoding cannot distinguish from empty.
 	aux := struct {
 		DefinitionLog json.RawMessage `json:"definition-log"`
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM!

laskoviymishka (approved):
:shipit:

PR #1627, opened 2026-08-01
fix(manifest): reject row ID allocation overflow

fallintoplace:
## What changed

Validate every row-ID range assigned by the v3 manifest writers before advancing the cursor:

- reject negative initial row IDs
- reject negative added or existing row counts
- detect `int64` overflow while allocating a manifest range
- advance `NextRowID` only after manifest validation and encoding succeed

The same checked arithmetic is used by `WriteManifestV3` when it returns the next available row ID.

## Why

V3 row lineage requires non-negative row IDs. The manifest-list writer previously added `existing_rows_count` and `added_rows_count` directly to its cursor. Near `math.MaxInt64`, that addition could wrap and assign a negative `first_row_id` to the next manifest. Invalid counts or a later validation failure could also move the in-memory cursor even though the manifest was rejected.

This could produce overlapping or invalid row-ID ranges in metadata written near the boundary.

## Testing

- negative initial row ID
- negative manifest row count
- manifest-list range overflow
- `WriteManifestV3` range overflow
- cursor remains unchanged after overflow or later validation failure
- `go test .`
- `go vet .`

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1599,6 +1599,10 @@ func NewManifestListWriterV3(out io.Writer, snapshotId, sequenceNumber, firstRow
 }
 
 func advanceRowID(firstRowID, existingRows, addedRows int64) (int64, error) {
+	if existingRows == -1 || addedRows == -1 {
+		return 0, fmt.Errorf("%w: cannot assign row-lineage IDs with unknown row counts: existing=%d added=%d",
+			ErrInvalidArgument, existingRows, addedRows)
+	}
 	if existingRows < 0 || addedRows < 0 {
 		return 0, fmt.Errorf("%w: row counts must be non-negative: existing=%d added=%d",
 			ErrInvalidArgument, existingRows, addedRows)
--- a/manifest.go
+++ b/manifest.go
@@ -1599,9 +1599,14 @@ func NewManifestListWriterV3(out io.Writer, snapshotId, sequenceNumber, firstRow
 }
 
 func advanceRowID(firstRowID, existingRows, addedRows int64) (int64, error) {
-	if existingRows == -1 || addedRows == -1 {
-		return 0, fmt.Errorf("%w: cannot assign row-lineage IDs with unknown row counts: existing=%d added=%d",
-			ErrInvalidArgument, existingRows, addedRows)
+	if firstRowID < 0 {
+		return 0, fmt.Errorf("%w: first row ID must be non-negative: %d", ErrInvalidArgument, firstRowID)
+	}
+	if existingRows == -1 {
+		existingRows = 0
+	}
+	if addedRows == -1 {
+		addedRows = 0
 	}
 	if existingRows < 0 || addedRows < 0 {
 		return 0, fmt.Errorf("%w: row counts must be non-negative: existing=%d added=%d",
@@ -1646,9 +1651,17 @@ func (m *ManifestListWriter) NextRowID() *int64 {
 	return m.nextRowID
 }
 
-func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
+func (m *ManifestListWriter) AddManifests(files []ManifestFile) (err error) {
 	if len(files) == 0 {
 		return nil
+	}
+	if m.version == 3 && m.nextRowID != nil {
+		batchStartRowID := *m.nextRowID
+		defer func() {
+			if err != nil {
+				*m.nextRowID = batchStartRowID
+			}
+		}()
 	}
 
 	switch m.version {
@@ -1846,7 +1859,7 @@ func WriteManifestV3(
 	raw.FirstRowIDValue = &v
 	nextFirstRowID, err = advanceRowID(firstRowID, raw.ExistingRowsCount, raw.AddedRowsCount)
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, fmt.Errorf("manifest %q: %w", filename, err)
 	}
 
 	return raw, nextFirstRowID, nil
--- a/manifest.go
+++ b/manifest.go
@@ -1666,6 +1666,8 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 
 	case 2, 3:
 		for _, file := range files {
+			var assignedNextRowID *int64
+
 			// Per the Iceberg spec a v2 manifest list may reference v1 manifest
 			// files (and a v3 list may reference v1 or v2 manifests) so that a
 			// table can be upgraded without rewriting historical manifests. The
@@ -1693,7 +1695,7 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 								return fmt.Errorf("manifest %q: %w", wrapped.Path, err)
 							}
 							wrapped.FirstRowIDValue = &firstRowID
-							*m.nextRowID = nextRowID
+							assignedNextRowID = &nextRowID
 						}
 					}
 				}
@@ -1721,6 +1723,9 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 			}
 			if err := m.writer.Encode(wrapped); err != nil {
 				return err
+			}
+			if assignedNextRowID != nil {
+				*m.nextRowID = *assignedNextRowID
 			}
 		}
 	default:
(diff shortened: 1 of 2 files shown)


manifest.go
```diff
@@ -1594,6 +1598,19 @@ func NewManifestListWriterV3(out io.Writer, snapshotId, sequenceNumber, firstRow
 	})
 }
 
+func advanceRowID(firstRowID, existingRows, addedRows int64) (int64, error) {
+	if existingRows < 0 || addedRows < 0 {
```
tanmayrauth, line 1611: -1 isn't invalid here — it's the sentinel for a v1 manifest with unset row counts (toFile at 263-273; see the comment at snapshot_producers.go:170). Since append defaults to fastAppend (manifest-merge is off by default) and passes  inherited manifests through untouched into AddManifests, an inherited v1 data manifest with unset counts hits this branch and fails the whole append with "row counts must be non-negative: existing=-1 added=-1" — misleading, since  they're missing, not negative. Rejecting is fine (old code did += -2 and corrupted the cursor), but handle the -1 case on purpose: recompute the count, or bail with a message that names the real cause like "manifest %q has unknown row counts, can't assign row-lineage IDs".

tanmayrauth (commented):
Overflow guard and cursor reorder look good. One issue with the negative-count check, inline.

manifest.go
fallintoplace: Thank you for the review, I handled the case here [1014f6e](https://github.com/apache/iceberg-go/pull/1627/commits/1014f6ebf29215999fea059ecbc7ad258616f9e7).

manifest.go
```diff
@@ -1671,8 +1694,12 @@ func (m *ManifestListWriter) AddManifests(files []ManifestFile) error {
 					if wrapped.FirstRowIDValue == nil {
 						if m.nextRowID != nil {
 							firstRowID := *m.nextRowID
+							nextRowID, err := advanceRowID(firstRowID, wrapped.ExistingRowsCount, wrapped.AddedRowsCount)
```
laskoviymishka, line 1710: The overflow rejection is right, but I think this guard hard-aborts a spec-legal path. `toFile()` sets `AddedRowsCount` and `ExistingRowsCount` to -1 when the Avro count was null, which is how pre-1.4 v1/v2 manifests decode. Carrying one of those into a v3 list — the upgrade-without-rewrite path the comment right above even calls out — now trips `existingRows < 0 || addedRows < 0` and returns `ErrInvalidArgument`, so the whole write aborts on a count the caller never set.

The old code was also wrong here (it advanced the cursor by -2), so this surfaces the bug rather than introducing it, but a hard error on old tables is a worse failure mode than the silent one. I'd normalize the -1 sentinel to 0 for the cursor math and keep the hard error for genuinely-negative counts.

Can you confirm the `toFile` -1 path actually reaches here? That's the one link I couldn't fully trace.

manifest.go
```diff
@@ -1594,6 +1598,23 @@ func NewManifestListWriterV3(out io.Writer, snapshotId, sequenceNumber, firstRow
 	})
 }
 
+func advanceRowID(firstRowID, existingRows, addedRows int64) (int64, error) {
```
laskoviymishka, line 1601: Both callers guard `firstRowID < 0` before this runs, but `advanceRowID` itself assumes it and the arithmetic quietly breaks if that assumption ever slips — `math.MaxInt64 - firstRowID` wraps for a negative `firstRowID`, so the overflow check would misfire. I'd move the `firstRowID < 0` check in here as the first clause.

That also lets us drop the duplicated `first row ID must be non-negative` string from both call sites — right now two copies feed two `ErrorContains` assertions, and if one drifts a test passes against the wrong text. One guard, one message, self-contained. wdyt?

manifest.go
```diff
@@ -1810,7 +1844,10 @@ func WriteManifestV3(
 	raw := mf.(*manifestFile)
 	v := firstRowID
 	raw.FirstRowIDValue = &v
-	nextFirstRowID = firstRowID + raw.AddedRowsCount + raw.ExistingRowsCount
+	nextFirstRowID, err = advanceRowID(firstRowID, raw.ExistingRowsCount, raw.AddedRowsCount)
```
laskoviymishka, line 1860: `AddManifests` wraps this as `manifest %q: %w` so the caller knows which file blew up, but here the bare `err` comes back with no filename. Since `WriteManifestV3` already takes the path, I'd wrap it the same way for parity.

manifest_test.go
```diff
@@ -2430,6 +2446,47 @@ func (m *ManifestTestSuite) TestV3ManifestListWriterRowIDTracking() {
...
+		var buf bytes.Buffer
+		writer, err := NewManifestListWriterV3(&buf, snapshotID, 1, 0, nil)
+		m.Require().NoError(err)
+		manifest := NewManifestFile(3, "negative.avro", 100, 1, snapshotID).AddedRows(-2).Build()
+		err = writer.AddManifests([]ManifestFile{manifest})
+		m.Require().ErrorIs(err, ErrInvalidArgument)
+		m.Require().ErrorContains(err, "row counts must be non-negative")
+	})
+
+	m.Run("overflow", func() {
```
laskoviymishka, line 2478: This subtest and the negative-count one above only drive `AddedRows`, so the existing-rows branch of the guard (`existingRows > MaxInt64-firstRowID`) never actually runs. A swap or drop of that clause would sail through green.

I'd add `ExistingRows` parallels for both — one with `ExistingRows(-1)`, one with `ExistingRows(1)` at `firstRowID = math.MaxInt64` — so both overflow clauses are pinned.

manifest_test.go
```diff
@@ -2430,6 +2446,47 @@ func (m *ManifestTestSuite) TestV3ManifestListWriterRowIDTracking() {
...
+		writer, err := NewManifestListWriterV3(&buf, snapshotID, 1, math.MaxInt64, nil)
+		m.Require().NoError(err)
+		manifest := NewManifestFile(3, "overflow.avro", 100, 1, snapshotID).AddedRows(1).Build()
+		err = writer.AddManifests([]ManifestFile{manifest})
+		m.Require().ErrorIs(err, ErrInvalidArgument)
+		m.Require().ErrorContains(err, "overflows int64")
+		m.EqualValues(math.MaxInt64, *writer.NextRowID())
+	})
+
+	m.Run("later validation failure leaves cursor unchanged", func() {
```
laskoviymishka, line 2500: This nicely covers the cursor staying put when the first manifest fails, but the deferral you built is really there for the batch case: early manifests succeed and advance, a later one overflows, and the cursor has to stay at where the batch started. That's the path a shared-state or advance-outside-the-loop bug would slip through, and it isn't exercised yet.

I'd add a `[valid, overflow]` case — writer at 10, first manifest adds 5 and succeeds, second overflows — then assert the error comes back and `*writer.NextRowID()` is still 10. That's the assertion that actually proves the deferral.

laskoviymishka (changes requested):
Nice catch on this one. Rejecting the overflow rather than letting the row-ID cursor silently wrap is the right instinct, and it's stricter than Java here (`ManifestListWriter.V3Writer.prepare` just does `nextRowId += ...` with no guard).

I'd hold it before merging, though. The main thing: `toFile()` assigns -1 to `AddedRowsCount`/`ExistingRowsCount` when the Avro count was null, which is how pre-1.4 v1/v2 manifests decode. The new negativity guard will hard-abort the moment one of those is carried into a v3 manifest list — the upgrade-without-rewrite path the surrounding comment explicitly calls out. The old code was also wrong there (it advanced the cursor by -2), so this surfaces the bug rather than introducing it, but a hard error on historical tables is a worse failure mode than the silent one. I'd normalize the -1 sentinel to 0 for the cursor math and keep the hard error for genuinely-negative counts.

- Confirm and handle the -1 null-count path so upgrading an old table doesn't abort.
- Push the `firstRowID < 0` check into `advanceRowID` itself — the arithmetic wraps for a negative `firstRowID`, and it lets the duplicated non-negative message collapse to one copy.
- Add the batch cursor test: early manifests succeed and advance, a later one overflows, cursor stays at where the batch started. That's the case the deferral actually protects, and it's not covered yet.

laskoviymishka (approved):

PR #1628, opened 2026-08-01
fix(schema): replace state during JSON decoding

fallintoplace:
## What changed

Make `Schema.UnmarshalJSON` decode and validate into fresh temporary state, then replace the receiver only after the complete document is valid. On success, clear every cached lookup index and rebuild the lazy parent and name-mapping functions for the new schema.

## Why

The decoder previously unmarshaled through the existing receiver. Reusing a `Schema` could therefore retain values omitted from the next document, including `schema-id` and `identifier-field-ids`. Cached lookups could continue returning fields from the previous schema.

Duplicate-field validation also ran after the receiver had already been modified, so invalid input could leave a partially updated schema behind. Decoding transactionally gives successful reuse replacement semantics while preserving the original value on error.

## Testing

- reused receiver clears omitted schema metadata
- ID, name, case-insensitive name, and name-mapping caches reflect the replacement schema
- invalid duplicate IDs preserve the original schema
- `go test .`
- `go vet .`

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -277,10 +277,11 @@ func (s *Schema) FieldIDs() []int {
 
 func (s *Schema) UnmarshalJSON(b []byte) error {
 	type Alias Schema
+	var decoded Schema
 	aux := struct {
 		Fields []NestedField `json:"fields"`
 		*Alias
-	}{Alias: (*Alias)(s)}
+	}{Alias: (*Alias)(&decoded)}
 
 	if err := json.Unmarshal(b, &aux); err != nil {
 		return err
@@ -290,12 +291,13 @@ func (s *Schema) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	s.init()
-
-	s.fields = aux.Fields
-	if s.IdentifierFieldIDs == nil {
-		s.IdentifierFieldIDs = []int{}
+	decoded.fields = aux.Fields
+	if decoded.IdentifierFieldIDs == nil {
+		decoded.IdentifierFieldIDs = []int{}
 	}
+
+	*s = decoded
+	s.init()
 
 	return nil
 }
--- a/schema.go
+++ b/schema.go
@@ -296,7 +296,14 @@ func (s *Schema) UnmarshalJSON(b []byte) error {
 		decoded.IdentifierFieldIDs = []int{}
 	}
 
-	*s = decoded
+	s.ID = decoded.ID
+	s.IdentifierFieldIDs = decoded.IdentifierFieldIDs
+	s.fields = decoded.fields
+	s.idToName.Store(nil)
+	s.idToField.Store(nil)
+	s.nameToID.Store(nil)
+	s.nameToIDLower.Store(nil)
+	s.idToAccessor.Store(nil)
 	s.init()
 
 	return nil
--- a/schema_test.go
+++ b/schema_test.go
@@ -570,6 +570,50 @@ func TestUnmarshalSchema(t *testing.T) {
 	assert.True(t, tableSchemaSimple.Equals(&schema))
 }
 
+func TestUnmarshalSchemaReplacesExistingState(t *testing.T) {
+	schema := iceberg.NewSchemaWithIdentifiers(7, []int{1},
+		iceberg.NestedField{ID: 1, Name: "old", Type: iceberg.PrimitiveTypes.String},
+	)
+	_, ok := schema.FindFieldByID(1)
+	require.True(t, ok)
+
+	require.NoError(t, json.Unmarshal([]byte(`{
+		"type": "struct",
+		"fields": [{"id": 2, "name": "new", "type": "long", "required": true}]
+	}`), schema))
+
+	assert.Zero(t, schema.ID)
+	assert.Empty(t, schema.IdentifierFieldIDs)
+	assert.Equal(t, 1, schema.NumFields())
+	_, ok = schema.FindFieldByID(1)
+	assert.False(t, ok)
+	field, ok := schema.FindFieldByID(2)
+	require.True(t, ok)
+	assert.Equal(t, "new", field.Name)
+}
+
+func TestUnmarshalSchemaPreservesExistingStateOnError(t *testing.T) {
+	schema := iceberg.NewSchemaWithIdentifiers(7, []int{1},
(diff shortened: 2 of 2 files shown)


tanmayrauth (approved):
LGTM!

schema.go
```diff
@@ -290,13 +291,21 @@ func (s *Schema) UnmarshalJSON(b []byte) error {
...
-
-	s.fields = aux.Fields
-	if s.IdentifierFieldIDs == nil {
-		s.IdentifierFieldIDs = []int{}
+	decoded.fields = aux.Fields
+	if decoded.IdentifierFieldIDs == nil {
+		decoded.IdentifierFieldIDs = []int{}
 	}
 
+	s.ID = decoded.ID
```
laskoviymishka, line 299: This manual three-field copy is a bit of a trap. If someone later adds a new JSON-tagged field to `Schema`, they'll wire up the tag and `MarshalJSON` but won't know to update this block, so the value decodes into `decoded` and then gets silently dropped on the re-unmarshal path, with nothing at compile time to catch it.

I'd add a comment right here spelling out the contract: decode-into-temp-then-copy is deliberate (validate before we touch the receiver, the same reason Java's `SchemaParser.fromJson` always returns a fresh `Schema`), and any new JSON-decodable field has to be copied here too. wdyt?

schema.go
```diff
@@ -290,13 +291,21 @@ func (s *Schema) UnmarshalJSON(b []byte) error {
...
+	s.IdentifierFieldIDs = decoded.IdentifierFieldIDs
+	s.fields = decoded.fields
+	s.idToName.Store(nil)
+	s.idToField.Store(nil)
+	s.nameToID.Store(nil)
+	s.nameToIDLower.Store(nil)
+	s.idToAccessor.Store(nil)
+	s.init()
```
laskoviymishka, line 307: Small ordering thing: we write `s.fields`/`s.ID`/`s.IdentifierFieldIDs` first and then clear the five atomics. In the window between, a concurrent reader on a lazy-rebuild path can recompute an index from the new fields and store it, only for the `Store(nil)` to immediately throw it away. Harmless today, and mutating a `*Schema` concurrently is racy regardless, but the idiomatic order is clear the derived caches first, then overwrite the authoritative fields, then `init()`.

While we're here: this clears 5 of the 7 cached values explicitly but leaves `lazyIDToParent` and `lazyNameMapping` to `init()`. That's correct, but non-obvious. A one-line comment noting those two are reset by `init()` because they're func values rather than atomics would save the next person from wondering why they're not in the `Store(nil)` list.

schema_test.go
```diff
@@ -570,6 +570,65 @@ func TestUnmarshalSchema(t *testing.T) {
...
+	)
+	_, ok := schema.FindFieldByID(1)
+	require.True(t, ok)
+	_, ok = schema.FindColumnName(1)
+	require.True(t, ok)
+	_, ok = schema.FindFieldByName("old")
+	require.True(t, ok)
+	_, ok = schema.FindFieldByNameCaseInsensitive("OLD")
+	require.True(t, ok)
+	assert.Contains(t, schema.NameMapping().String(), "old")
```
laskoviymishka, line 585: The pre-warm seeds `lazyNameMapping` and the four atomic caches, but skips `lazyIDToParent`, and that's the one cache reset by `init()` rather than an explicit `Store(nil)`, so it's the most likely to regress if `init()` ever changes.

Could we seed it with `schema.FieldHasOptionalParent(1)` before the unmarshal, then assert `FieldHasOptionalParent(2)` is `false` after the reuse? wdyt?

schema_test.go
```diff
@@ -570,6 +570,65 @@ func TestUnmarshalSchema(t *testing.T) {
...
+			{"id": 2, "name": "first", "type": "long", "required": true},
+			{"id": 2, "name": "duplicate", "type": "string", "required": false}
+		]
+	}`), schema)
+	require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
+
+	assert.Equal(t, 7, schema.ID)
+	assert.Equal(t, []int{1}, schema.IdentifierFieldIDs)
+	assert.Equal(t, 1, schema.NumFields())
+	field, ok := schema.FindFieldByID(1)
```
laskoviymishka, line 637: This doesn't actually prove the caches survive the error.

`FindFieldByID` goes through `lazyIDToField`, which rebuilds from `s.fields` on a nil atomic. So even if the error path had wrongly zeroed the caches, this would still pass by recomputing from the unchanged `fields` slice, and the test can't tell "caches left intact" from "caches cleared but recomputed". Since stale cached lookups are exactly the bug we're guarding against, I'd warm every lookup (`FindFieldByID`, `FindFieldByName`, `FindColumnName`, `FindFieldByNameCaseInsensitive`, `NameMapping()`) before the failing unmarshal, then assert they still return the old field afterward.

laskoviymishka (changes requested):
Nice fix. Decoding into a temp `Schema` and only copying into the receiver once validation passes is exactly right. It closes the stale-ID/stale-cache reuse bug and the half-updated-on-error case in one move, and it's about as close to Java's always-fresh-`Schema` model as we can get given `UnmarshalJSON` has to mutate the receiver.

I'd hold before merging on one thing mainly.

`TestUnmarshalSchemaPreservesExistingStateOnError` doesn't actually prove what it's guarding against. `FindFieldByID` rebuilds from `s.fields` on a nil cache, so the test would still pass even if the error path had zeroed the caches. It can't distinguish "caches intact" from "caches cleared but recomputed". Since stale cached lookups are the whole point of the fix, I'd warm every lookup before the failing unmarshal and re-assert after.

The other thing worth doing before merge: the manual three-field copy is a quiet trap. Add a new JSON-tagged field to `Schema` later and it'll decode into the temp and get silently dropped on the re-unmarshal path, with nothing at compile time to catch it. A short comment at the copy block stating the contract (and that any new decodable field must be copied here) would cover it.

A couple of smaller ones I left inline: seeding `lazyIDToParent` in the reuse test since it's the one cache not cleared by `Store(nil)`, and the clear-then-write ordering plus a note on the two lazy funcs.

Once the error test actually exercises the cache path, happy to take another pass and approve.

laskoviymishka (approved):

Issue #989, opened 2026-05-05
feat: geometry and geography type support (Iceberg v3)


laskoviymishka:
Parent: #589 (v3 support tracking)

Iceberg v3 adds `geometry` and `geography` as first-class primitive types. WKB ([Well-Known Binary](https://libgeos.org/specifications/wkb/)) is the on-disk encoding; the GeoArrow extension provides the canonical Arrow representation. Spec: [Iceberg primitive types](https://iceberg.apache.org/spec/#primitive-types), [Parquet Geospatial types](https://github.com/apache/parquet-format/blob/master/Geospatial.md).

Two stale WIP PRs exist (#628, #663) but no parent issue tied them together. This umbrella tracks the full surface and breaks it into six independently mergeable pieces. Looking for an owner — happy to mentor the contributor who picks it up.

Sub-issues:

- #990 — type + schema plumbing (foundational)
- #991 — WKB encoding + GeoArrow conversion (depends on #990)
- #992 — Parquet GeoStatistics emission (depends on #991)
- #993 — manifest-level geo bounds aggregation (depends on #992)
- #994 — bbox predicate evaluation + pruning (depends on #993)
- #995 — geo defaults + write path (depends on #991 + #994)

Each child is sized to ship in isolation. The chain is mostly linear: types → Arrow → parquet stats → manifest bounds → predicates → write. Cross-client coverage matters at every step — Java/pyiceberg fixtures should pin the wire format.

happydave1:
I'm happy to pick up the work for adding Geography and Geometry types! Thanks for breaking the issue up into sub issues!

C-Loftus:
I would be happy to help implementing geo support. Is there a specific issue that others are not working on and would be straightforward to start with as a new contributor? 

Also, should I wait for https://github.com/apache/iceberg-go/pull/984 to be merged before proceeding?

Thanks!

laskoviymishka:
@C-Loftus, thanks for jumping in!

Honest answer: the sub-issues are mostly a linear chain (#990 → #991 → #992 → #993 → #994 → #995), and the first two are already with @happydave1. So nothing downstream is truly unblocked yet.

That said, there are two good ways to start contributing on geo now:

[...]

PR #1632, opened 2026-08-02
test(table): geometry write path populates manifest bounds end-to-end

### LINKED ISSUE #995: feat(table): geometry/geography defaults + write path
Parent: #989

Depends on #991 + #994. Parses the WKT-encoded `initial-default` / `write-default` into WKB at metadata-load time (using the helper added in #991) and applies them via the existing default-value enforcement (#820). The write path routes records through the partition writer by transform — only `identity` is defined for geo today, but plumb so future bbox-bucket transforms drop in.

Tests write a table with a geometry column and verify Parquet GeoStatistics + manifest bounds are populated end-to-end. Cross-client read by Java pins the wire format.

tanmayrauth:
Write a geometry and geography column through the data file writer and assert the geometry column's manifest bounds decode to the expected planar XY box while geography stays unbounded. Complements the internal TestWriteDataFileGeoBounds by driving the full public writer path.

Geo defaults remain null-only per the Iceberg spec, already enforced and tested elsewhere.

Closes: #995

### DIFF
--- a/table/geo_write_test.go
+++ b/table/geo_write_test.go
@@ -1,0 +1,134 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/io"
+	tblutils "github.com/apache/iceberg-go/table/internal"
+	"github.com/geoarrow/geoarrow-go"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/twpayne/go-geom/encoding/wkb"
+	"github.com/twpayne/go-geom/encoding/wkt"
+)
+
+// wktToWKB converts Well-Known Text into little-endian (NDR) Well-Known Binary,
+// the on-disk encoding iceberg-go writes for geometry/geography columns.
+func wktToWKB(t *testing.T, s string) geoarrow.WKBBytes {
+	t.Helper()
+
+	g, err := wkt.Unmarshal(s)
+	require.NoError(t, err)
+	b, err := wkb.Marshal(g, wkb.NDR)
+	require.NoError(t, err)
+
+	return geoarrow.WKBBytes(b)
+}
+
+// TestWriteGeometryColumnPopulatesBounds writes a geometry and a geography
+// column end-to-end through the data file writer and checks the manifest-level
+// geo bounds that iceberg-go computes from the WKB values. arrow-go's Parquet
+// writer does not emit native GeoStatistics, so iceberg-go derives the column
+// bounds itself (see geoBoundsAccumulator) and threads them into the DataFile
+// exactly as the manifest carries them. Geometry gets a planar XY bounding box;
+// geography stays unbounded because a planar box over geodesic edges is unsafe.
+//
+// This is the public-path counterpart to internal.TestWriteDataFileGeoBounds:
+// that test feeds hand-built StatsCols straight to WriteDataFile, whereas this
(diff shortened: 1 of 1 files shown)


tanmayrauth:
@laskoviymishka  @zeroshade  Can you please take a look when you get chance?

table/geo_write_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	arrowSchema, err := SchemaToArrowSchema(schema, nil, true, false)
+	require.NoError(t, err)
+
+	// Two geometry points spanning X in [0, 30] and Y in [-5, 10].
+	geomLo := wktToWKB(t, "POINT (0 -5)")
+	geomHi := wktToWKB(t, "POINT (30 10)")
+	geog := wktToWKB(t, "POINT (12 4)")
+
+	rec, _, err := array.RecordFromJSON(memory.DefaultAllocator, arrowSchema, strings.NewReader(`[
```
laskoviymishka, line 98: Follow-up, not for this PR: at some point I'd run this on a checked allocator rather than `memory.DefaultAllocator`. The suite tests (`WriteRecordsTestSuite`, `ClusteredWriterTestSuite`) use `memory.NewCheckedAllocator` and assert zero residual on teardown, and for a path wiring the geo-bounds accumulator a silent allocation escape is the kind of thing we'd want a test to catch. `memory.NewCheckedAllocator(memory.DefaultAllocator)` plus a `mem.AssertSize(t, 0)` once the writer's done matches `arrow_utils_internal_test.go`.

table/geo_write_test.go
```diff
@@ -0,0 +1,134 @@
...
+		{"id": 1, "geom": "`+geomLo.String()+`", "geog": "`+geog.String()+`"},
```
laskoviymishka, line 99: Not blocking, just a note for whoever reads this next: the test quietly depends on geoarrow's JSON parser round-tripping `geomLo.String()` (lowercase hex) back into WKB. If that ever stopped, the accumulator would see no values and the `require.Contains` below would fail for a fairly confusing reason. It works today and `parquet_files_test.go` leans on the same `.String()` hex pattern, so a one-line comment pointing at that precedent would save someone the archaeology later.

table/geo_write_test.go
```diff
@@ -0,0 +1,134 @@
...
+	upper := df.UpperBoundValues()
+	require.Contains(t, lower, geomFieldID, "geometry column must record a lower bound")
+	require.Contains(t, upper, geomFieldID, "geometry column must record an upper bound")
+
+	minX, minY, maxX, maxY, ok := tblutils.GeoBoundsXY(lower[geomFieldID], upper[geomFieldID])
+	require.True(t, ok, "geometry bounds must decode to a planar XY box")
+	assert.Equal(t, 0.0, minX)
+	assert.Equal(t, -5.0, minY)
+	assert.Equal(t, 30.0, maxX)
+	assert.Equal(t, 10.0, maxY)
```
laskoviymishka, line 129: Follow-up idea: alongside the geo bounds, asserting that the `id` column still gets ordinary min/max would be cheap insurance. The geo-type guard in `DataFileStatsFromMeta` suppresses generic Parquet stats for geo columns, and a regression that over-suppressed an adjacent non-geo column would sail straight past this test. A single `require.Contains(t, lower, 1)` on the `id` field ID pins it down.

table/geo_write_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Geography stays unbounded: a planar box over geodesic edges is unsafe.
+	assert.NotContains(t, lower, geogFieldID, "geography column must not record bounds")
```
laskoviymishka, line 132: Wording nit for a follow-up: the assertion itself is right, but "must not record bounds" reads as a spec mandate and it isn't one. The V3 spec permits geography bounds (including the xmin > xmax wrapping convention) — iceberg-go omits them as a deliberate conservative choice, which the accumulator's own comment spells out ("geography is left unbounded until geodesic/antimeridian-aware computation is added"). Phrasing it as "geography does not record bounds in the current implementation" keeps us from encoding a temporary limitation as an invariant, and makes it obvious this assertion should flip when that computation lands.

laskoviymishka (approved):
This is good coverage and I'm happy to see it land. Driving geometry bounds through `newDataFileWriter`/`writeFile` over a `MetadataBuilder` exercises `computeStatsPlan` and the geo stats-collector setup that `TestWriteDataFileGeoBounds` skips by feeding `WriteDataFile` pre-built `StatsCols`, so a `computeStatsPlan` misconfig on a geo column would slip past the internal test but not this one. The XY box math and the WKB encoding path both check out.

Everything below is follow-up material. None of it needs to change before merge.

The one I'd genuinely like tracked is the issue reference. `Closes #995` claims more than this delivers: #995 is geo defaults *plus* the write path, which means parsing the WKT `initial-default`/`write-default` into WKB at metadata-load time, applying them through the existing default-value enforcement, and plumbing partition-writer routing by transform. End-to-end tests are one clause of that, and `feature-status.md` still lists geo transforms/statistics/pruning as in progress under #989. Since GitHub acts on `Closes:` at merge time, #995 will auto-close on its own. Either narrow it to "partially addresses #995" now (a description-only edit, no re-push or CI re-run) or reopen the issue after merging, whichever you prefer.

Coverage follow-ups, in rough order of what I think is worth doing:

- an all-null geometry column, asserting the field drops out of the bounds map through the full path
- a metrics-mode sub-test (`write.metadata.metrics.column.geom=none`) so we know `computeStatsPlan` actually suppresses geo bounds when asked; that's the dispatch the internal test can't reach
- the `id` column still getting ordinary min/max, so an over-eager geo-suppression regression can't hide

A few smaller notes inline, all optional.

PR #1633, opened 2026-08-03
fix(cli): use standard property keys when creating a namespace

jx2lee:
## Summary
`iceberg create namespace` set the `Description`/`Location` properties with capitalized keys. Glue's `constructDatabaseInput` only recognizes the lowercase `comment`/`location` keys (Hive's uppercase handling is a legacy fallback, not the standard), so `--location-uri` was silently dropped into the namespace's generic parameter map instead of `DatabaseInput.LocationUri` for Glue catalogs, while the CLI reported success.

Use `comment`/`location`, matching the keys catalogs actually read
(`catalog/glue/glue.go`, `catalog/hive/hive.go`, `catalog/internal/utils.go`).

## Test Plan
`go test ./cmd/iceberg`

### DIFF
--- a/cmd/iceberg/create_test.go
+++ b/cmd/iceberg/create_test.go
@@ -1,0 +1,78 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM! Thanks for the fix.

laskoviymishka (approved):
LGTM!

PR #1634, opened 2026-08-03
fix(table): default absent geoarrow CRS metadata to OGC:CRS84

twuebi:
The Parquet geospatial spec defines OGC:CRS84 as the default when a GEOMETRY/GEOGRAPHY logical type carries no CRS; the arrow-to-Iceberg conversion previously fabricated srid:0 for such fields, producing a type mismatch that failed projection with "cannot promote geometry(srid:0) to geometry".

Map an absent CRS to Iceberg's default geometry/geography types instead; explicit CRS values, including an explicit srid, are unchanged. An absent Parquet geography algorithm already resolves to spherical before reaching this conversion, so the default algorithm needs no change.

Symmetrically, stop serializing srid:0 as an omitted GeoArrow CRS on the write side: an unset Parquet CRS means OGC:CRS84, so an unknown CRS must be written explicitly to round-trip.

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -2183,16 +2183,12 @@ func isWKT2CRSString(crs string) bool {
 	return false
 }
 
-// defaultGeoCRS is the CRS assumed when a Parquet GEOMETRY or GEOGRAPHY logical
-// type carries no CRS; it is also the default CRS of the Iceberg geo types.
-const defaultGeoCRS = "OGC:CRS84"
-
 // geoArrowCRSToIcebergCRS maps GeoArrow CRS metadata to an Iceberg CRS string.
 // Absent CRS metadata means the default CRS OGC:CRS84, matching the Parquet
 // geospatial spec and Iceberg's default geometry/geography types.
 func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 	if len(meta.CRS) == 0 {
-		return defaultGeoCRS, nil
+		return iceberg.DefaultGeoCRS, nil
 	}
 
 	switch {
@@ -2206,8 +2202,8 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 			return "", errors.New("unsupported CRS: empty string CRS")
 		}
 
-		if strings.EqualFold(crs, defaultGeoCRS) || strings.EqualFold(crs, "EPSG:4326") {
-			return defaultGeoCRS, nil
+		if strings.EqualFold(crs, iceberg.DefaultGeoCRS) || strings.EqualFold(crs, "EPSG:4326") {
+			return iceberg.DefaultGeoCRS, nil
 		}
 
 		switch meta.CRSType {
@@ -2224,6 +2220,11 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 			return "", errWKT2CRSNotSupported
 		}
 		if meta.CRSType == geoarrow.CRSTypeSRID {
+			// Some writers store the identifier already prefixed; don't prefix twice.
+			if strings.HasPrefix(strings.ToLower(crs), "srid:") {
+				return crs, nil
+			}
+
 			return "srid:" + crs, nil
 		}
 
@@ -2281,8 +2282,8 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 		}
 
 		authorityCode := authority + ":" + code
-		if strings.EqualFold(authorityCode, defaultGeoCRS) || strings.EqualFold(authorityCode, "EPSG:4326") {
-			return defaultGeoCRS, nil
+		if strings.EqualFold(authorityCode, iceberg.DefaultGeoCRS) || strings.EqualFold(authorityCode, "EPSG:4326") {
+			return iceberg.DefaultGeoCRS, nil
 		}
 
 		return authorityCode, nil
@@ -2385,7 +2386,7 @@ func icebergCRSToGeoArrowMetadata(crs string, props iceberg.Properties) (geoarro
 
 	if lowerCRS == "epsg:4326" {
 		// collapse EPSG:4326 to OGC:CRS84
-		raw, _ = json.Marshal(defaultGeoCRS) //nolint:errcheck // Marshalling a string can't fail
+		raw, _ = json.Marshal(iceberg.DefaultGeoCRS) //nolint:errcheck // Marshalling a string can't fail
 	} else {
 		raw, _ = json.Marshal(crs) //nolint:errcheck // Marshalling a string can't fail
 	}
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -2183,9 +2183,16 @@ func isWKT2CRSString(crs string) bool {
 	return false
 }
 
+// defaultGeoCRS is the CRS assumed when a Parquet GEOMETRY or GEOGRAPHY logical
+// type carries no CRS; it is also the default CRS of the Iceberg geo types.
+const defaultGeoCRS = "OGC:CRS84"
+
+// geoArrowCRSToIcebergCRS maps GeoArrow CRS metadata to an Iceberg CRS string.
+// Absent CRS metadata means the default CRS OGC:CRS84, matching the Parquet
+// geospatial spec and Iceberg's default geometry/geography types.
 func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 	if len(meta.CRS) == 0 {
-		return "srid:0", nil
+		return defaultGeoCRS, nil
 	}
 
 	switch {
@@ -2199,8 +2206,8 @@ func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 			return "", errors.New("unsupported CRS: empty string CRS")
 		}
 
-		if strings.EqualFold(crs, "OGC:CRS84") || strings.EqualFold(crs, "EPSG:4326") {
-			return "OGC:CRS84", nil
+		if strings.EqualFold(crs, defaultGeoCRS) || strings.EqualFold(crs, "EPSG:4326") {
+			return defaultGeoCRS, nil
 		}
 
 		switch meta.CRSType {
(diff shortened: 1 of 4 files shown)


twuebi:
@laskoviymishka should be the last geo fix aside https://github.com/apache/arrow-go/pull/1051 / https://github.com/apache/arrow-go/pull/1052 to enable reading trino-written geo columns through iceberg-go

laskoviymishka:
That's quite a milestone!!!

table/arrow_utils.go
```diff
@@ -2183,9 +2183,16 @@ func isWKT2CRSString(crs string) bool {
...
+// type carries no CRS; it is also the default CRS of the Iceberg geo types.
+const defaultGeoCRS = "OGC:CRS84"
+
+// geoArrowCRSToIcebergCRS maps GeoArrow CRS metadata to an Iceberg CRS string.
+// Absent CRS metadata means the default CRS OGC:CRS84, matching the Parquet
+// geospatial spec and Iceberg's default geometry/geography types.
 func geoArrowCRSToIcebergCRS(meta geoarrow.Metadata) (string, error) {
 	if len(meta.CRS) == 0 {
-		return "srid:0", nil
+		return defaultGeoCRS, nil
```
laskoviymishka: I think this changes the read result for files old iceberg-go already wrote. Before, a `geometry(srid:0)` column serialized to absent GeoArrow CRS (via the `srid:0` to omitted special case this PR also removes), and absent CRS read back as `srid:0`, internally consistent. After this change, absent CRS reads as plain `geometry`, so projecting one of those files against a `geometry(srid:0)` table schema hits PromoteType's default branch and fails with `cannot promote geometry to geometry(srid:0)`.

Blast radius is bounded, since that write path only shipped a few weeks ago, but it's a silent read regression for anyone who wrote `srid:0` geometry in that window. I'd allow `geometry(srid:0)` → `geometry` (and `geography(srid:0, spherical)` → `geography`) as permitted promotions, since old iceberg-go used `srid:0` as the stand-in for absent CRS anyway; at minimum I'd document the migration. wdyt?

table/arrow_utils.go
```diff
@@ -2340,10 +2347,6 @@ func icebergCRSToGeoArrowMetadata(crs string, props iceberg.Properties) (geoarro
 	if strings.HasPrefix(lowerCRS, "srid:") {
 		id := crs[len("srid:"):]
 
-		if id == "0" {
-			return geoarrow.NewMetadata(), nil // srid:0 maps to omitted GeoArrow CRS
-		}
-
 		raw, _ := json.Marshal(id) //nolint:errcheck // Marshalling a string can't fail
```
laskoviymishka, line 2351: I think there's a subtle issue with what actually lands in the Parquet CRS field here. Dropping the `srid:0` special case means `icebergCRSToGeoArrowMetadata("srid:0")` now emits GeoArrow crs `"0"` with crs_type srid, and when that serializes out, the Parquet GEOMETRY logical type carries the bare string `"0"`.

The Parquet geospatial spec wants the full `"srid:0"` form with the prefix, and arrow-rs keys its unknown-CRS sentinel off exactly `"srid:0"`: it won't read `"0"` as SRID 0, it'll treat it as an opaque authority code. So we'd be writing `srid:0` geometry other engines can't round-trip. I'd store the full `"srid:<id>"` string in the GeoArrow CRS field so `ParquetCRS()` emits `"srid:0"`. Can we confirm what `ParquetCRS()` produces here and pin it with a Parquet-level round-trip?

table/arrow_utils_internal_test.go
```diff
@@ -495,11 +496,16 @@ func TestIcebergCRSToGeoArrowMetadata(t *testing.T) {
...
 		meta, err := icebergCRSToGeoArrowMetadata("srid:0", nil)
 		require.NoError(t, err)
-		assert.Empty(t, meta.CRS)
-		assert.Empty(t, meta.CRSType)
+		assert.JSONEq(t, `"0"`, string(meta.CRS))
+		assert.Equal(t, geoarrow.CRSTypeSRID, meta.CRSType)
+
+		crs, err := geoArrowCRSToIcebergCRS(meta)
+		require.NoError(t, err)
+		assert.Equal(t, "srid:0", crs)
```
laskoviymishka, line 508: This round-trips through the GeoArrow metadata but stops there: it never serializes to a Parquet CRS string and reads it back, which is exactly where the `"0"` vs `"srid:0"` question bites. It'd still pass even if what we write to Parquet is `"0"` rather than the spec's `"srid:0"`. A round-trip that goes iceberg type → Parquet CRS → iceberg type would pin the wire format we actually care about.

table/arrow_utils_test.go
```diff
@@ -2756,3 +2774,244 @@ func TestToRequestedSchemaMismatchedExtensionNameNotRewrapped(t *testing.T) {
...
+	wkb, ok := written.(*geoarrow.WKBType)
+	require.True(t, ok, "expected geoarrow.wkb, got %T", written)
+
+	dt, err := geoarrow.NewWKBType().Deserialize(wkb.StorageType(), wkb.Serialize())
+	require.NoError(t, err)
+
+	return dt
+}
+
+func TestToRequestedSchemaGeoSRID0Projection(t *testing.T) {
```
laskoviymishka, line 2919: This covers `srid:0` written by the new code and read straight back, which is the easy direction. The case that regresses is the other one: a file with absent CRS (what old iceberg-go wrote for a `srid:0` column) projected against a `geometry(srid:0)` schema. I'd add a test pinning that path, either asserting the promotion now succeeds (if we allow it) or documenting the failure, so the back-compat behavior is deliberate rather than incidental.

table/arrow_utils.go
```diff
@@ -2183,9 +2183,16 @@ func isWKT2CRSString(crs string) bool {
 	return false
 }
 
+// defaultGeoCRS is the CRS assumed when a Parquet GEOMETRY or GEOGRAPHY logical
+// type carries no CRS; it is also the default CRS of the Iceberg geo types.
+const defaultGeoCRS = "OGC:CRS84"
```
laskoviymishka: iceberg already has an unexported `defaultGeoCRS = "OGC:CRS84"` over in types.go, so this is a second copy of the same magic string with no compiler link between them. I'd export `iceberg.DefaultGeoCRS` and use it here; the new internal-test assertions hard-code `"OGC:CRS84"` too, same drift risk. Minor, but it's the kind of thing that silently diverges.

laskoviymishka (changes requested):
Nice fix, almost there! 

Defaulting absent CRS to OGC:CRS84 is the right call per the Parquet geospatial spec, and I like that you did it symmetrically on both the read and write sides rather than patching just the read path.

I'd hold this before merging though, mostly around `srid:0`.

The one I'd most want to nail down: dropping the `srid:0` to omitted special case means we now write the bare string `"0"` into the Parquet CRS field, where the spec (and arrow-rs, which you reference) expect the full `"srid:0"`. So `srid:0` geometry we write may not round-trip through other engines, and the round-trip test we have doesn't catch it because it stops at the GeoArrow metadata and never touches the Parquet wire format.

The other is back-compat: files old iceberg-go wrote for `srid:0` columns went out with absent CRS, and those now read back as plain `geometry`, which fails to promote against a `geometry(srid:0)` table schema. Blast radius is bounded since that write path only shipped a few weeks ago, but it's a silent read regression for anyone who used it. I'd lean toward allowing `geometry(srid:0)` to `geometry` as a promotion rather than just documenting it.

- write `srid:0` as `"srid:0"` (not `"0"`) to Parquet, and pin it with a type → Parquet CRS → type round-trip
- decide back-compat: allow `geometry(srid:0)` → `geometry`, or document the migration, and add a test for the absent-CRS-vs-`srid:0`-schema path
- small one: reuse a single OGC:CRS84 constant instead of a second copy

Once those are sorted, happy to take another pass. The read direction also has a mirror bug, arrow-rs `"srid:0"` decodes to `"srid:srid:0"` on the way in; not blocking for this PR, but worth a follow-up while it's fresh.

twuebi:
Thanks, dug into all three.

**`"0"` vs `"srid:0"`:** we don't write a Parquet CRS field at all right now. arrow-go v18.7.0 has no GEOMETRY/GEOGRAPHY logical type in its schema layer (only the thrift structs in `parquet/internal/gen-go/parquet/parquet.go`), so a geo column goes out as plain `BYTE_ARRAY` with logical type `None`, and the CRS travels in the GeoArrow metadata of the stored `ARROW:schema`. I verified that against the real write path (`table/internal/parquet_files.go`, pqarrow with `WithStoreSchema`). Storing `"srid:0"` in the GeoArrow `crs` field would also double-prefix on our own read path — `geoArrowCRSToIcebergCRS` prepends `srid:` for `crs_type: srid`, giving `geometry(srid:srid:0)`, which is the arrow-rs bug you mention at the end — and it would contradict `srid:4326` → `{"crs":"4326","crs_type":"srid"}`, which predates this PR, plus GeoArrow's definition of `crs_type: srid` as an opaque identifier. So `{"crs":"0","crs_type":"srid"}` is the right GeoArrow form; the prefixed `srid:<id>` spelling belongs to whoever emits the Parquet logical type.

I added the round-trip you asked for, at the level that exists today: `TestGeoTypeParquetRoundTrip` writes a real Parquet file per CRS shape, pins the GeoArrow metadata, pins that the column logical type is still `None` (so we get a failing test to revisit the CRS spelling the moment arrow-go starts emitting GEOMETRY), then reads back through `ArrowSchemaToIceberg` and pins the iceberg type. `geometry(srid:0)` and `geography(srid:0)` survive.

**Back-compat:** narrower than it looks — the old `srid:0` → omitted encoding landed in 38d0d12 and is in no release: `git tag --contains 38d0d12` is empty and the newest tag, v0.6.0, predates it. Only builds from main in the last few weeks can have written such files. Also, the case you describe needs `geometry` → `geometry(srid:0)`, not `geometry(srid:0)` → `geometry`: the file now reads as the default CRS and the table declares `srid:0`. Either direction is a CRS promotion, which isn't in the spec's promotion list that `PromoteType` implements, so I'd rather not add one unilaterally — it would diverge from other engines and mask genuinely mismatched CRS. I added `TestToRequestedSchemaGeoAbsentCRSAgainstSRID0SchemaFails` so the rejection is deliberate rather than incidental.

**Constant:** exported `iceberg.DefaultGeoCRS` and switched both `types.go` and `table/arrow_utils.go` to it, so the second copy is gone. I kept the literal `"OGC:CRS84"` in the test assertions on purpose — an assertion written against the constant under test can't catch a change to that constant.

**Read-side `srid:srid:0`:** agreed, and it turned out to be a two-line prefix check in the same spot, so I folded it in rather than deferring: with `crs_type: srid`, a `crs` value that already starts with `srid:` is taken as-is. Pinned both at the CRS-mapping level (`srid:0`, `srid:3857`, `SRID:3857`) and end to end, where `{"crs":"srid:0","crs_type":"srid"}` now reads as `geometry(srid:0)` instead of `geometry(srid:srid:0)`.

laskoviymishka (approved):
🌬️ ⛵

PR #1635, opened 2026-08-04
fix(catalog/sql): return ErrNamespaceAlreadyExists on a concurrent create

iremcaginyurtturk:
## Problem

`CreateNamespace` in the sql catalog resolves the namespace key, returns `ErrNamespaceAlreadyExists` if it is present, and otherwise inserts the properties. Two callers creating the same namespace both pass that check, and the loser's insert violates the primary key on `iceberg_namespace_properties`. The driver's error is returned unwrapped:

```
error inserting namespace properties for namespace '[raw]': constraint failed:
UNIQUE constraint failed: iceberg_namespace_properties.catalog_name,
iceberg_namespace_properties.namespace, iceberg_namespace_properties.property_key (1555)
```

Callers that write `errors.Is(err, catalog.ErrNamespaceAlreadyExists)` — the contract the other implementations honour, and the natural way to write create-if-absent — see a failure for a namespace that does exist. It surfaces as an intermittent first-run failure in any tool that loads two tables into a new namespace concurrently; that is how I ran into it.

## Fix

Re-resolve the namespace key when the insert fails, and return `ErrNamespaceAlreadyExists` if it is present now. An insert that failed for any other reason is returned exactly as before.

I kept the pre-check rather than replacing it with an upsert: it keeps the common path a single read, and the behaviour on the raced path is then identical to the non-raced one.

### DIFF
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1284,8 +1284,8 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, insertErr)
 		}
 
-		// Only Postgres aborts the whole tx on a failed insert, so only it needs a
-		// savepoint to keep the re-check runnable; Oracle has no RELEASE SAVEPOINT.
+		// Postgres aborts the whole tx on a failed insert, so only it needs the
+		// savepoint for the re-check; SQLite/MySQL don't, and Oracle can't RELEASE it.
 		if tx.Dialect().Name() == dialect.PG {
 			sp, err := tx.BeginTx(ctx, nil)
 			if err != nil {
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1269,15 +1269,33 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 			})
 		}
 
-		_, err := tx.NewInsert().Model(&toInsert).Exec(ctx)
+		// Run the insert in a savepoint so a failure can be rolled back without
+		// aborting the whole transaction (Postgres); the re-check then still runs.
+		sp, err := tx.BeginTx(ctx, nil)
 		if err != nil {
-			// A concurrent writer may have inserted since the check above; if the
-			// re-check itself fails, fall through to the insert error.
-			if _, exists, checkErr := c.resolveNamespaceKeyInTx(ctx, tx, namespace); checkErr == nil && exists {
+			return fmt.Errorf("error creating savepoint for namespace '%s': %w", namespace, err)
+		}
+
+		if _, err = sp.NewInsert().Model(&toInsert).Exec(ctx); err != nil {
+			if rbErr := sp.Rollback(); rbErr != nil {
+				// A failed rollback poisons the tx (Postgres); keep both causes.
+				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(err, rbErr))
+			}
+			// A concurrent writer may have won the race; return the sentinel in the
+			// same form as the pre-check above so both paths are identical.
+			_, exists, checkErr := c.resolveNamespaceKeyInTx(ctx, tx, namespace)
+			if checkErr == nil && exists {
 				return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(namespace, "."))
+			}
+			if checkErr != nil {
+				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(err, checkErr))
 			}
 
 			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, err)
+		}
+
+		if err = sp.Commit(); err != nil {
+			return fmt.Errorf("error releasing savepoint for namespace '%s': %w", namespace, err)
 		}
 
 		return nil
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1271,9 +1271,9 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 
 		_, err := tx.NewInsert().Model(&toInsert).Exec(ctx)
 		if err != nil {
-			// A concurrent writer can insert between the check above and here, and
-			// the driver's duplicate-key error is not what callers match on.
-			if _, exists, checkErr := c.resolveNamespaceKey(ctx, namespace); checkErr == nil && exists {
+			// A concurrent writer may have inserted since the check above; if the
+			// re-check itself fails, fall through to the insert error.
+			if _, exists, checkErr := c.resolveNamespaceKeyInTx(ctx, tx, namespace); checkErr == nil && exists {
 				return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(namespace, "."))
 			}
 
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1269,33 +1269,46 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 			})
 		}
 
-		// Run the insert in a savepoint so a failure can be rolled back without
-		// aborting the whole transaction (Postgres); the re-check then still runs.
-		sp, err := tx.BeginTx(ctx, nil)
-		if err != nil {
-			return fmt.Errorf("error creating savepoint for namespace '%s': %w", namespace, err)
-		}
-
-		if _, err = sp.NewInsert().Model(&toInsert).Exec(ctx); err != nil {
-			if rbErr := sp.Rollback(); rbErr != nil {
-				// A failed rollback poisons the tx (Postgres); keep both causes.
-				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(err, rbErr))
-			}
-			// A concurrent writer may have won the race; return the sentinel in the
-			// same form as the pre-check above so both paths are identical.
+		// A concurrent writer may have won the race; return the sentinel in the same
+		// form as the pre-check above. This must stay the first read on the outer tx
+		// so MySQL's REPEATABLE READ snapshot lands after the winner committed.
+		recheck := func(insertErr error) error {
 			_, exists, checkErr := c.resolveNamespaceKeyInTx(ctx, tx, namespace)
 			if checkErr == nil && exists {
 				return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(namespace, "."))
 			}
 			if checkErr != nil {
-				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(err, checkErr))
+				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(insertErr, checkErr))
 			}
 
-			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, err)
+			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, insertErr)
 		}
 
-		if err = sp.Commit(); err != nil {
-			return fmt.Errorf("error releasing savepoint for namespace '%s': %w", namespace, err)
+		// Only Postgres aborts the whole tx on a failed insert, so only it needs a
+		// savepoint to keep the re-check runnable; Oracle has no RELEASE SAVEPOINT.
+		if tx.Dialect().Name() == dialect.PG {
+			sp, err := tx.BeginTx(ctx, nil)
+			if err != nil {
+				return fmt.Errorf("error creating savepoint for namespace '%s': %w", namespace, err)
+			}
+
+			if _, err = sp.NewInsert().Model(&toInsert).Exec(ctx); err != nil {
+				if rbErr := sp.Rollback(); rbErr != nil {
+					return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(err, rbErr))
+				}
+
+				return recheck(err)
+			}
+
+			if err = sp.Commit(); err != nil {
+				return fmt.Errorf("error releasing savepoint for namespace '%s': %w", namespace, err)
+			}
+
+			return nil
+		}
+
+		if _, err := tx.NewInsert().Model(&toInsert).Exec(ctx); err != nil {
+			return recheck(err)
 		}
 
 		return nil
--- a/catalog/sql/sql.go
+++ b/catalog/sql/sql.go
@@ -1271,6 +1271,12 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 
 		_, err := tx.NewInsert().Model(&toInsert).Exec(ctx)
 		if err != nil {
+			// A concurrent writer can insert between the check above and here, and
+			// the driver's duplicate-key error is not what callers match on.
+			if _, exists, checkErr := c.resolveNamespaceKey(ctx, namespace); checkErr == nil && exists {
+				return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(namespace, "."))
+			}
+
 			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, err)
 		}
 
--- a/catalog/sql/sql_test.go
+++ b/catalog/sql/sql_test.go
@@ -20,6 +20,7 @@ package sql_test
 import (
 	"context"
 	"database/sql"
+	"database/sql/driver"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -31,6 +32,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/apache/arrow-go/v18/arrow"
@@ -1951,6 +1953,7 @@ func (s *SqliteCatalogTestSuite) TestCreateNamespaceConcurrent() {
 	close(errs)
 
 	created, alreadyExists := 0, 0
+	var unexpected []error
 	for err := range errs {
 		switch {
 		case err == nil:
@@ -1958,12 +1961,101 @@ func (s *SqliteCatalogTestSuite) TestCreateNamespaceConcurrent() {
 		case errors.Is(err, catalog.ErrNamespaceAlreadyExists):
 			alreadyExists++
 		default:
-			s.Failf("unexpected error", "want nil or ErrNamespaceAlreadyExists, got %v", err)
+			unexpected = append(unexpected, err)
 		}
 	}
 
+	s.Empty(unexpected, "want nil or ErrNamespaceAlreadyExists")
 	s.Equal(1, created)
 	s.Equal(writers-1, alreadyExists)
+}
+
+// plantingDriver wraps the sqlite driver and, once, inserts the namespace row
+// on the same connection just before the catalog's own insert.
+type plantingDriver struct {
+	base      driver.Driver
+	namespace string
+	planted   atomic.Bool
+}
+
+func (d *plantingDriver) Open(dsn string) (driver.Conn, error) {
+	conn, err := d.base.Open(dsn)
+	if err != nil {
+		return nil, err
+	}
+
+	return &plantingConn{Conn: conn, drv: d}, nil
+}
+
+type plantingConn struct {
+	driver.Conn
+	drv *plantingDriver
+}
+
+func (c *plantingConn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	execer, ok := c.Conn.(driver.ExecerContext)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+	// Same statement would be rolled back with the failing insert, so the plant
+	// goes in as its own statement first.
+	if strings.Contains(query, "iceberg_namespace_properties") && strings.HasPrefix(strings.ToUpper(strings.TrimSpace(query)), "INSERT") && c.drv.planted.CompareAndSwap(false, true) {
+		if _, err := execer.ExecContext(ctx, "INSERT INTO iceberg_namespace_properties (catalog_name, namespace, property_key, property_value) VALUES (?, ?, ?, ?)",
+			[]driver.NamedValue{
+				{Ordinal: 1, Value: "default"},
+				{Ordinal: 2, Value: c.drv.namespace},
+				{Ordinal: 3, Value: "exists"},
+				{Ordinal: 4, Value: "true"},
+			}); err != nil {
+			return nil, err
+		}
+	}
+
+	return execer.ExecContext(ctx, query, args)
+}
+
+func (c *plantingConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	queryer, ok := c.Conn.(driver.QueryerContext)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+
+	return queryer.QueryContext(ctx, query, args)
+}
+
+func (c *plantingConn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
+	return c.Conn.(driver.ConnBeginTx).BeginTx(ctx, opts)
+}
+
+func (c *plantingConn) PrepareContext(ctx context.Context, query string) (driver.Stmt, error) {
+	return c.Conn.(driver.ConnPrepareContext).PrepareContext(ctx, query)
+}
+
+// The concurrent test above cannot tell whether a loser returned from the check
+// before the insert or from the recovery after it. This one only passes if the
+// insert ran and failed.
+func (s *SqliteCatalogTestSuite) TestCreateNamespaceLosesInsertRace() {
+	ctx := context.Background()
+	namespace := table.Identifier{databaseName()}
+
+	base, err := sql.Open(sqliteshim.ShimName, ":memory:")
+	s.Require().NoError(err)
+	s.Require().NoError(base.Close())
+
+	drvName := "sqlite-planting-" + namespace[0]
+	sql.Register(drvName, &plantingDriver{base: base.Driver(), namespace: namespace[0]})
+
+	sqldb, err := sql.Open(drvName, s.catalogUri())
+	s.Require().NoError(err)
+	defer sqldb.Close()
+	sqldb.SetMaxOpenConns(1)
+
+	cat, err := sqlcat.NewCatalog("default", sqldb, sqlcat.SQLite, iceberg.Properties{"warehouse": "file://" + s.warehouse})
+	s.Require().NoError(err)
+
+	err = cat.CreateNamespace(ctx, namespace, nil)
+	s.ErrorIs(err, catalog.ErrNamespaceAlreadyExists)
+	s.Contains(err.Error(), namespace[0])
 }
 
 func (s *SqliteCatalogTestSuite) TestCreateNamespaceRejectsReservedProperty() {
--- a/catalog/sql/sql_test.go
+++ b/catalog/sql/sql_test.go
@@ -2067,6 +2067,12 @@ func (c *pgAbortConn) QueryContext(ctx context.Context, query string, args []dri
 		return nil, errEmulatedAborted
 	}
 
+	// Postgres probes information_schema for schema-version detection; sqlite has
+	// no such view, so answer as the V1 table this suite creates (column present).
+	if strings.Contains(query, "information_schema") {
+		return &boolRows{val: true}, nil
+	}
+
 	// The namespace exists on the re-check only in the unique-violation case: a
 	// concurrent winner committed the row. An unrelated failure leaves it absent.
 	if isNamespaceExistsProbe(query) {
@@ -2118,7 +2124,7 @@ func (r *boolRows) Next(dest []driver.Value) error {
 	return nil
 }
 
-func (s *SqliteCatalogTestSuite) newAbortCatalog(mode insertFailure) *sqlcat.Catalog {
+func (s *SqliteCatalogTestSuite) newAbortCatalog(mode insertFailure) (*sqlcat.Catalog, *pgAbortDriver) {
 	base, err := sql.Open(sqliteshim.ShimName, ":memory:")
 	s.Require().NoError(err)
 	s.Require().NoError(base.Close())
@@ -2126,7 +2132,8 @@ func (s *SqliteCatalogTestSuite) newAbortCatalog(mode insertFailure) *sqlcat.Cat
 	// drvName derives from databaseName() (unique per call); sql.Register panics
 	// on a duplicate name, so this must not become a fixed string.
 	drvName := "sqlite-pgabort-" + databaseName()
-	sql.Register(drvName, &pgAbortDriver{base: base.Driver(), mode: mode})
+	drv := &pgAbortDriver{base: base.Driver(), mode: mode}
+	sql.Register(drvName, drv)
 
 	sqldb, err := sql.Open(drvName, s.catalogUri())
 	s.Require().NoError(err)
@@ -2135,33 +2142,37 @@ func (s *SqliteCatalogTestSuite) newAbortCatalog(mode insertFailure) *sqlcat.Cat
 	// transaction that follows it.
 	sqldb.SetMaxOpenConns(1)
 
-	cat, err := sqlcat.NewCatalog("default", sqldb, sqlcat.SQLite, iceberg.Properties{"warehouse": "file://" + s.warehouse})
+	// Postgres dialect so the savepoint-backed recovery path runs; the driver
+	// above emulates Postgres aborting the tx over an sqlite base.
+	cat, err := sqlcat.NewCatalog("default", sqldb, sqlcat.Postgres, iceberg.Properties{"warehouse": "file://" + s.warehouse})
 	s.Require().NoError(err)
 
-	return cat
+	return cat, drv
 }
 
 // Emulates Postgres aborting the transaction: only passes if the savepoint let
 // the re-check recover the sentinel, which SQLite alone cannot show.
 func (s *SqliteCatalogTestSuite) TestCreateNamespaceLosesInsertRace() {
 	namespace := table.Identifier{databaseName()}
-	cat := s.newAbortCatalog(failUnique)
+	cat, drv := s.newAbortCatalog(failUnique)
 
 	err := cat.CreateNamespace(context.Background(), namespace, nil)
 	s.ErrorIs(err, catalog.ErrNamespaceAlreadyExists)
 	s.Contains(err.Error(), namespace[0])
+	s.True(drv.insertAttempted.Load(), "the emulated insert must have run")
 }
 
 // An unrelated insert failure on an absent namespace must surface its own
 // error, not a misleading ErrNamespaceAlreadyExists.
 func (s *SqliteCatalogTestSuite) TestCreateNamespaceInsertFailureSurfacesOriginalError() {
 	namespace := table.Identifier{databaseName()}
-	cat := s.newAbortCatalog(failUnrelated)
+	cat, drv := s.newAbortCatalog(failUnrelated)
 
 	err := cat.CreateNamespace(context.Background(), namespace, nil)
 	s.Error(err)
 	s.NotErrorIs(err, catalog.ErrNamespaceAlreadyExists)
 	s.Contains(err.Error(), "disk I/O error")
+	s.True(drv.insertAttempted.Load(), "the emulated insert must have run")
 }
 
 func (s *SqliteCatalogTestSuite) TestCreateNamespaceRejectsReservedProperty() {
--- a/catalog/sql/sql_test.go
+++ b/catalog/sql/sql_test.go
@@ -21,13 +21,16 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"maps"
 	"math/rand/v2"
+	"net/url"
 	"os"
 	"path/filepath"
 	"slices"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/apache/arrow-go/v18/arrow"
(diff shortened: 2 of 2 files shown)


catalog/sql/sql_test.go
```diff
@@ -1911,6 +1914,58 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+			created++
+		case errors.Is(err, catalog.ErrNamespaceAlreadyExists):
+			alreadyExists++
+		default:
+			s.Failf("unexpected error", "want nil or ErrNamespaceAlreadyExists, got %v", err)
+		}
+	}
+
+	s.Equal(1, created)
+	s.Equal(writers-1, alreadyExists)
```
laskoviymishka, line 1971: This assertion can't tell whether the new branch ran.

The pre-check at line 1246 sits outside `withWriteTx`, which doesn't open until 1261. So a loser has two ways to get here: it sees `exists == true` at 1251 and returns before touching the insert, or it passes the pre-check, loses the write-lock race, and reaches the new code at 1276. Both return `fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(namespace, "."))`, byte for byte, so `alreadyExists == 7` holds either way and nothing in the drain loop can distinguish them. Which path a given goroutine takes is down to scheduling.

I'd add a deterministic test next to this one that forces the insert to fail: create the namespace first, then drive the insert path with the row already present, so 1276-1278 are actually covered. Keeping this one as a concurrency smoke test is fine, I'd just not lean on it as the regression test for the fix.

catalog/sql/sql_test.go
```diff
@@ -1911,6 +1914,58 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+
+	created, alreadyExists := 0, 0
+	for err := range errs {
+		switch {
+		case err == nil:
+			created++
+		case errors.Is(err, catalog.ErrNamespaceAlreadyExists):
+			alreadyExists++
+		default:
+			s.Failf("unexpected error", "want nil or ErrNamespaceAlreadyExists, got %v", err)
```
laskoviymishka: `s.Failf` marks the test failed but doesn't stop the loop, so one unexpected error leaves `created` and `alreadyExists` short and you get three failures instead of one, with the message that actually matters buried under two count mismatches.

I'd collect and assert after the drain:

```go
var unexpected []error
// ... default: unexpected = append(unexpected, err)
s.Empty(unexpected, "want nil or ErrNamespaceAlreadyExists")
s.Equal(1, created)
s.Equal(writers-1, alreadyExists)
```

catalog/sql/sql.go
```diff
@@ -1271,6 +1271,12 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 
 		_, err := tx.NewInsert().Model(&toInsert).Exec(ctx)
 		if err != nil {
+			// A concurrent writer can insert between the check above and here, and
+			// the driver's duplicate-key error is not what callers match on.
+			if _, exists, checkErr := c.resolveNamespaceKey(ctx, namespace); checkErr == nil && exists {
```
laskoviymishka: non-blocking: the comment covers the race but not the `checkErr == nil` half of the guard. I'd add a clause saying that if the re-check itself fails we fall through and return the original insert error, since that's the non-obvious part. wdyt?

laskoviymishka (changes requested):
Good catch, and the fix is in the right place. Worth noting this is actually better than both reference implementations: Java's `JdbcCatalog.createNamespace` calls `insertProperties` with no handling around the insert, so a race loser there gets an `IllegalStateException`, and PyIceberg's `SqlCatalog.create_namespace` surfaces a raw `IntegrityError`. Returning `ErrNamespaceAlreadyExists` also matches what the REST catalog already does with a 409, so the two catalogs now agree on the observable contract. Thanks for the repro in the description too, made this easy to follow.

One blocking thing before merge: `TestCreateNamespaceConcurrent` doesn't pin the new branch. The pre-check at line 1246 runs outside `withWriteTx`, and it returns `ErrNamespaceAlreadyExists` with formatting byte-identical to the new path at 1277, so `alreadyExists == 7` holds whether a goroutine was gated at the pre-check or actually collided on the insert. Which path each one takes comes down to scheduling. I left a note inline with what I'd add.

Separately, and not for this PR: `CreateNamespace` is the only create path in this file still on `withWriteTx` with its existence check outside the transaction. `CreateTable` and `CreateView` both do the check inside `withSerializableWriteTx`, which closes the window structurally instead of recovering from it. Probably worth a follow-up issue.

Once the test covers that branch, happy to take another pass.

iremcaginyurtturk:
You were right that the concurrent test proved nothing about the new branch. Fixed in `358e302`.

`TestCreateNamespaceLosesInsertRace` only passes if the insert ran and failed — a driver wrapper plants the row on the same connection just before the catalog's insert. It has to be a separate statement: sqlite rolls back the failing statement's own changes, trigger effects included, which is why the trigger version I tried first does not work. With the recovery removed it fails, with it in place it passes.

The re-check also moved into the transaction. On a separate connection it cannot see an uncommitted winner, and with a constrained pool it deadlocks — the new test sets `SetMaxOpenConns(1)` and the old code hung there.

Error collection and the comment are done as suggested.

On the follow-up: agreed, moving the check inside `withSerializableWriteTx` closes the window rather than recovering from it. Happy to open a PR for that — as a separate one, or folded in here if you prefer.

catalog/sql/sql.go
```diff
@@ -1271,6 +1271,12 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
 
 		_, err := tx.NewInsert().Model(&toInsert).Exec(ctx)
 		if err != nil {
+			// A concurrent writer may have inserted since the check above; if the
+			// re-check itself fails, fall through to the insert error.
+			if _, exists, checkErr := c.resolveNamespaceKeyInTx(ctx, tx, namespace); checkErr == nil && exists {
```
laskoviymishka: This guard can't fire on Postgres. A statement error there marks the whole transaction aborted and every command after it gets `25P02` until a rollback, so once the insert fails with a `23505` unique violation the `SELECT` that `resolveNamespaceKeyInTx` runs on this same `tx` comes back `checkErr != nil` rather than `exists`. We fall through and return the raw insert error, which is exactly what we returned before this PR. `postgres` is one of the five dialects this file declares, so it's a supported target.

SQLite rolls back only the offending statement, which is why the new test is green. MySQL and Oracle behave like SQLite here; Postgres is the outlier, and it's the one most deployments run against.

Two shapes hold across dialects: a `SAVEPOINT` around the insert so the re-check has a live transaction to run in, or matching the driver's unique-violation code directly and dropping the re-query entirely. Moving the existence check inside a `withSerializableWriteTx`, the way `CreateTable` and `DropNamespace` do, is worth doing regardless, but on Postgres unique-index enforcement isn't covered by SSI predicate locking, so a loser can still come back with `23505` instead of a serialization failure. It narrows the window rather than replacing this block. wdyt?

catalog/sql/sql_test.go
```diff
@@ -1911,6 +1916,148 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+	driver.Conn
+	drv *plantingDriver
+}
+
+func (c *plantingConn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	execer, ok := c.Conn.(driver.ExecerContext)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+	// Same statement would be rolled back with the failing insert, so the plant
```
laskoviymishka: This comment has the mechanism slightly off, and the wrong version is what hides the dialect gap. The plant isn't "its own statement" in any transactional sense; it's in the same transaction as the catalog's insert and never separately committed. What actually saves it is that SQLite rolls back only the failing statement, so the plant row survives to be seen by the re-check.

I'd say that part out loud, because on Postgres the same sequence loses both and the re-check can't run at all (see `sql.go:1276`). Something like "SQLite rolls back only the failing statement, so the plant stays visible to the re-check; other dialects differ" keeps the next reader from taking this green test as multi-dialect coverage.

catalog/sql/sql_test.go
```diff
@@ -1911,6 +1916,148 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+	queryer, ok := c.Conn.(driver.QueryerContext)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+
+	return queryer.QueryContext(ctx, query, args)
+}
+
+func (c *plantingConn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
+	return c.Conn.(driver.ConnBeginTx).BeginTx(ctx, opts)
```
laskoviymishka: non-blocking: these two panic rather than fail if the shim ever stops implementing the interface, and a panic in a driver callback takes the whole test binary down with a trace that won't point back here. `ExecContext` and `QueryContext` just above already do the ok-check, so it's mostly consistency:

```go
beginTx, ok := c.Conn.(driver.ConnBeginTx)
if !ok {
	return nil, driver.ErrBadConn
}

return beginTx.BeginTx(ctx, opts)
```

Same for `PrepareContext` below it.

laskoviymishka (changes requested):
`plantingDriver`/`plantingConn` is exactly what I was asking for. Planting the row on the same connection just before the catalog's own insert forces the violation deterministically instead of leaving it to scheduling, and reverting the six lines in `sql.go` does make `TestCreateNamespaceLosesInsertRace` fail, so the branch is pinned now. The `unexpected []error` drain and the `checkErr == nil` clause in the comment both landed too.

One new blocker this round: the recovery never fires on Postgres. A statement error there puts the whole transaction in aborted state, so the `SELECT` inside `resolveNamespaceKeyInTx` on the same `tx` comes back `25P02` instead of a row, `checkErr != nil`, and we return the raw insert error, which is what we returned before this PR. `postgres` is one of the five dialects this file declares, so it isn't hypothetical. SQLite rolls back only the offending statement, which is the only reason the new test is green.

Which means walking back something I said last round. I called `CreateNamespace` keeping its existence check outside the transaction a follow-up, "not for this PR." It turns out to be load-bearing for the fix, so I'd rather deal with it here. One caveat on the obvious version of that: moving the check inside a `withSerializableWriteTx` like `CreateTable` and `DropNamespace` do narrows the window a lot, but Postgres doesn't cover unique-index enforcement under SSI, so a loser can still surface `23505` rather than a serialization failure. The two shapes that hold across dialects are a `SAVEPOINT` around the insert so the re-check has a live transaction to run in, or mapping the driver's unique-violation code straight to the sentinel and dropping the re-query.

If you'd rather keep the scope tight I won't fight it, but then the comment and the test should say SQLite-only out loud, so a green test doesn't get read as multi-dialect coverage. Either way, happy to take another pass once it's settled.

catalog/sql/sql.go
zeroshade: **Blocking — this branch cannot fire on PostgreSQL, so the bug is unfixed for that dialect.**

The re-check runs on `tx`, the same transaction whose `INSERT` just failed on line 1272. PostgreSQL aborts a transaction on *any* statement error (SQLSTATE 25P02); every subsequent command in that transaction fails with `current transaction is aborted, commands ignored until end of transaction block`.

So on Postgres `checkErr != nil` always, the `&& exists` branch is never taken, and control falls through to line 1280 — returning the raw duplicate-key error, which is precisely the behavior this PR sets out to fix. Postgres is a first-class dialect here (`sql.go:58`, `sql.go:243`), not an exotic one.

MySQL and SQLite don't abort the transaction on a duplicate-key error, so the fix works as intended there. Net effect: correct for two of three dialects, quietly inert on the third, and nothing in the test suite would catch it.

**Suggested fix, either approach:**

- Wrap the insert in a savepoint — `tx.Exec("SAVEPOINT ns_insert")` before, `ROLLBACK TO SAVEPOINT ns_insert` on failure — which restores the transaction to a usable state on Postgres and is a harmless no-op on the others.
- Or do the re-check outside the failed transaction, on `c.db` via `resolveNamespaceKey` rather than `resolveNamespaceKeyInTx`. A fresh connection isn't poisoned by the aborted transaction. Slightly weaker isolation for the read, but for an "does it exist now" check after a failure that is fine.

The savepoint version is the more faithful of the two, since it keeps the read inside the same transactional context.

catalog/sql/sql.go
```diff
@@ -1271,6 +1271,12 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
+				return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(namespace, "."))
+			}
+
 			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, err)
```
zeroshade: Minor: when the re-check succeeds and reports the namespace exists, the original insert error is discarded. That's defensible — the observable state *is* "already exists," and the caller gets the sentinel they need — but it does mean an insert that failed for some unrelated reason, on a namespace that happens to exist, reports a cause that isn't the real one.

**Suggested fix:** consider `errors.Join`-ing the driver error into the returned error on the recovery branch above, so `errors.Is(err, catalog.ErrNamespaceAlreadyExists)` still works for callers while operators keep the underlying cause in logs. Not blocking either way.

catalog/sql/sql_test.go
```diff
@@ -1911,6 +1916,148 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+func (s *SqliteCatalogTestSuite) TestCreateNamespaceLosesInsertRace() {
+	ctx := context.Background()
+	namespace := table.Identifier{databaseName()}
+
+	base, err := sql.Open(sqliteshim.ShimName, ":memory:")
+	s.Require().NoError(err)
+	s.Require().NoError(base.Close())
+
+	drvName := "sqlite-planting-" + namespace[0]
+	sql.Register(drvName, &plantingDriver{base: base.Driver(), namespace: namespace[0]})
```
zeroshade: `sql.Register` installs a driver name into a process-global registry and panics with `Register called twice for driver ...` on a duplicate. This is safe here only because `drvName` derives from `databaseName()` (`sql_test.go:152`), which is unique per call.

That's a load-bearing property of a helper defined ~1900 lines away, and the failure mode is a panic that takes the whole test binary down rather than a single failing test.

**Suggested fix:** a one-line comment noting the dependency on `databaseName()` uniqueness would keep someone from later switching to a fixed driver name for readability.

zeroshade (changes requested):
Thanks for this — the shape of the fix is the right one. Re-checking existence after a failed insert avoids driver-specific error-code sniffing entirely, which is the trap most implementations of this fall into: no `pq.Error` code table, no `sqlite3.ErrConstraintUnique` type switch, nothing that breaks when someone swaps drivers. And the sentinel wrapping is exactly right — `fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, …)` matches the pre-check path at `sql.go:1252` byte for byte, so `errors.Is` behaves identically whether the caller lost the race or not.

`TestCreateNamespaceLosesInsertRace` is also a genuinely good test. The planting driver proves the recovery branch actually *executed*, which a plain concurrency test can never do — it can only show the right error came out, not which code path produced it.

One blocking issue: as written, the fix is silently inert on PostgreSQL. Details inline on `sql.go:1276`.

## Test coverage

Beyond the planting-driver test, two gaps:

1. **No Postgres or MySQL coverage.** The entire portability question — which is where the blocking issue lives — is untested, and there's no pg driver in `go.mod` for CI to use. Even a comment on the test naming which dialects are actually exercised would help the next person, but given that the bug is dialect-specific, this is where a build-tagged integration test would earn its keep.
2. **No negative case.** Nothing proves that an insert failure for an *unrelated* reason (constraint violation on some other column, connection error, serialization failure) still surfaces the original error unchanged rather than being converted into a misleading `ErrNamespaceAlreadyExists`. The planting driver you already built makes this cheap to add — plant a different failure and assert the original error survives.

catalog/sql/sql.go
```diff
@@ -1269,11 +1269,30 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
+		}
+
+		if _, err = sp.NewInsert().Model(&toInsert).Exec(ctx); err != nil {
+			if rbErr := sp.Rollback(); rbErr != nil {
+				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, err)
+			}
+			// A concurrent writer may have won the race; if the re-check confirms
+			// it, return the sentinel joined with the cause, else the insert error.
+			if _, exists, checkErr := c.resolveNamespaceKeyInTx(ctx, tx, namespace); checkErr == nil && exists {
+				return errors.Join(fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(namespace, ".")), err)
```
laskoviymishka: The race-loser here returns `errors.Join(sentinel, err)`, but the pre-check path a few lines up returns a plain `fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, …)`. `errors.Is` holds either way so the 409 mapping is safe, but `.Error()` now carries the raw insert error on a second line, and `errors.As` will unwrap to the driver's constraint error only on the race path — so two paths that mean the same thing produce structurally different errors depending on scheduling.

I'd drop `err` from the join and return the same form as the pre-check so they're identical. If we want the cause for logs, `%v` keeps it in the string without putting it in the unwrap chain.

catalog/sql/sql.go
```diff
@@ -1269,11 +1269,30 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
-		_, err := tx.NewInsert().Model(&toInsert).Exec(ctx)
+		// Run the insert in a savepoint so a failure can be rolled back without
+		// aborting the whole transaction (Postgres); the re-check then still runs.
+		sp, err := tx.BeginTx(ctx, nil)
 		if err != nil {
+			return fmt.Errorf("error creating savepoint for namespace '%s': %w", namespace, err)
+		}
+
+		if _, err = sp.NewInsert().Model(&toInsert).Exec(ctx); err != nil {
+			if rbErr := sp.Rollback(); rbErr != nil {
```
laskoviymishka: Two failure branches in this block drop their cause. On rollback failure we return the insert `err` and throw away `rbErr` — but a failed `ROLLBACK TO SAVEPOINT` is a different and worse condition than a failed insert, and on Postgres it's exactly what leaves the tx poisoned, so that's the diagnostic I'd most want to keep. Same shape at the re-check just below: when `checkErr != nil` we fall straight through to the insert-error return and lose that the re-check itself failed.

I'd `errors.Join(err, rbErr)` on the rollback path and thread `checkErr` into the fall-through return so neither cause vanishes. wdyt?

catalog/sql/sql_test.go
```diff
@@ -1911,6 +1917,249 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+
+// boolRows is a single-row, single-column result carrying a SQL EXISTS answer.
+type boolRows struct {
+	val  bool
+	done bool
+}
+
+func (r *boolRows) Columns() []string { return []string{"exists"} }
+func (r *boolRows) Close() error      { return nil }
+func (r *boolRows) Next(dest []driver.Value) error {
```
laskoviymishka, line 2120: `Next` returns a row (with `int64(0)`) when `val` is false rather than signaling EOF, so this only reads as "not exists" if bun's existence check scans the column value rather than row presence. If it ever keys on presence, the `failUnrelated` case would see a row, conclude the namespace exists, and hand back `ErrNamespaceAlreadyExists` — the exact thing that test asserts against — so it'd fail opaquely instead of catching a regression.

Simplest way to make it robust either way is to return `io.EOF` when `!val` so "not exists" is genuinely an empty result:

```go
func (r *boolRows) Next(dest []driver.Value) error {
	if r.done || !r.val {
		return io.EOF
	}
	r.done = true
	dest[0] = int64(1)

	return nil
}
```

Worth a quick check that reverting the recovery in `sql.go` still fails `TestCreateNamespaceInsertFailureSurfacesOriginalError` — if it does, the negative case is genuinely pinned.

catalog/sql/sql_test.go
```diff
@@ -1911,6 +1917,249 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+	return strings.HasPrefix(strings.ToUpper(strings.TrimSpace(query)), "INSERT") &&
+		strings.Contains(query, "iceberg_namespace_properties")
+}
+
+func isNamespaceExistsProbe(query string) bool {
+	return strings.Contains(strings.ToUpper(query), "EXISTS") &&
+		strings.Contains(query, "iceberg_namespace_properties")
+}
+
+func isRollbackToSavepoint(query string) bool {
```
laskoviymishka, line 2030: Small robustness thing on the emulation: both `isRollbackToSavepoint` and `isNamespaceExistsProbe` match bun's exact SQL text (the `ROLLBACK TO SAVEPOINT` prefix, the literal `EXISTS`). Postgres and SQLite also accept `ROLLBACK TO <name>` without the keyword, and if a bun version ever changes either shape the `aborted` flag never clears or the probe stops firing, and the test falls through to an opaque `ErrorIs` failure instead of exercising the recovery.

Not blocking, but a one-line comment naming the bun format these depend on — or asserting the savepoint query was actually seen — would save the next person the debugging. The driver-scoped `aborted`/`insertAttempted` state is fine given `SetMaxOpenConns(1)` and you've already documented that, so I'd leave it.

laskoviymishka (changes requested):
The SAVEPOINT route is exactly one of the two shapes I floated last round, and it does resolve the Postgres blocker — the insert rolls back to the savepoint and the re-check runs on a live transaction instead of coming back `25P02`, so the recovery actually fires on pg and not just SQLite. The `failUnrelated` negative case landed too, and the `pgAbortDriver` emulation is a nice way to prove the aborted-transaction path without a real pg in CI. This is close.

One thing I'd sort before merge: the race-loser return isn't quite the same shape as the serial pre-check. The pre-check returns a plain `fmt.Errorf("%w: %s", …)`, and this path wraps that in an `errors.Join` with the raw insert error, so `.Error()` picks up an extra line and `errors.As` unwraps to the driver error only when you lost the race. `errors.Is` holds either way, but I don't think it's byte-identical to the pre-check the way the thread reads it — dropping the insert error from the join gets them back in sync. Details inline.

The rest is small:

- the rollback-failure and re-check-failure branches both drop their diagnostic (`rbErr`, `checkErr`) — I'd join them in so a poisoned tx doesn't get reported as a plain insert error.
- worth confirming the `failUnrelated` emulation actually fails without the recovery (the `boolRows` EOF thing inline), so the negative case is genuinely pinned.
- one sanity check that bun emits standard savepoint SQL for the other declared dialects — mssql uses `SAVE TRANSACTION`, so it's worth a look before we call this multi-dialect.

The broader "pre-check lives outside the tx" restructure I went back and forth on last round — I think the savepoint settles the load-bearing part of it, so I'm happy to leave the full move to a follow-up issue rather than hold this on it.

Fix the error shape and I'm happy to approve.

iremcaginyurtturk:
Thanks for the detailed pass. Pushed the fixes:

- **Error shape:** the race-loser now returns the same `fmt.Errorf("%w: %s", …)` as the serial pre-check, so the two paths are byte-identical and `errors.As` no longer unwraps to the driver error only on the race path.
- **Dropped causes:** the rollback-failure branch now `errors.Join(err, rbErr)` and the re-check-failure branch `errors.Join(err, checkErr)`, so a poisoned tx / failed re-check isn't reported as a plain insert error.
- **mssql:** confirmed bun emits `SAVE TRANSACTION` for the MSSavepoint feature, so the savepoint path holds across the declared dialects.
- **Emulation robustness:** added a comment naming the bun SQL shape the matchers depend on.

On the `boolRows` EOF suggestion — I went a different way, and wanted to flag why: bun's `Exists()` runs `SELECT EXISTS(...)` and scans the value via `QueryRowContext(...).Scan(&bool)`, so an empty result comes back as `sql.ErrNoRows` rather than "false" and the probe errors out (I hit this — the tests fail). So I kept `boolRows` emitting an explicit `0`/`1` row and documented that it must, rather than signalling EOF. I did confirm the pinning holds: reverting the savepoint makes `TestCreateNamespaceLosesInsertRace` fail, and the negative case would fail if the re-check ever reported "exists" for the unrelated-failure path — so both are genuinely pinned. Happy to revisit if you'd prefer a different shape there.

zeroshade (approved):
The savepoint shape resolves the Postgres aborted-transaction blocker cleanly — the insert failure rolls back to the savepoint and the re-check runs on a live transaction, with the `pgAbortDriver` emulation pinning exactly that dialect behavior without needing pg in CI. The race-loser now returns the sentinel byte-for-byte the same as the pre-check path, and the previously-dropped rollback/re-check causes are preserved via `errors.Join`. Driver-independent recovery (no error-code sniffing) remains the strongest property of this fix. Nice work.

Withdrawing my earlier suggestion to join the insert error onto the exists-path sentinel — path consistency argues the other way and the comment documents the choice.

---

catalog/sql/sql.go
```diff
@@ -1269,11 +1269,35 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
+				// A failed rollback poisons the tx (Postgres); keep both causes.
+				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(err, rbErr))
+			}
+			// A concurrent writer may have won the race; return the sentinel in the
+			// same form as the pre-check above so both paths are identical.
+			_, exists, checkErr := c.resolveNamespaceKeyInTx(ctx, tx, namespace)
```
laskoviymishka, line 1276: While we're here: this re-check being correct on MySQL leans on it being the first consistent read in the outer tx. Under REPEATABLE READ the snapshot is taken at the first non-locking SELECT, and since the insert is the only statement before it, the snapshot lands after the race winner committed and `exists` comes back true. Safe today.

It's a quiet invariant though. If anyone later adds an existence SELECT before the insert inside `withWriteTx`, the snapshot would establish too early and the re-check could miss the winner. A one-line comment that this must stay the first read on the outer tx would keep it from breaking silently. Not a blocker.

catalog/sql/sql.go
```diff
@@ -1269,11 +1269,35 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
+				return fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, strings.Join(namespace, "."))
+			}
+			if checkErr != nil {
+				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(err, checkErr))
+			}
+
 			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, err)
 		}
 
+		if err = sp.Commit(); err != nil {
```
laskoviymishka: This is the other leg of the multi-dialect savepoint worry from last round. I checked mssql's `SAVE TRANSACTION` on the create side, but the commit side has its own gap: `sp.Commit()` on a savepoint-backed tx emits `RELEASE SAVEPOINT`, and Oracle has no such statement. It supports `SAVEPOINT` and `ROLLBACK TO SAVEPOINT` but frees savepoints implicitly. bun only skips the RELEASE for the mssql feature flag as far as I can tell.

If that's right, this regresses Oracle on the happy path: `CreateNamespace` was a plain `Exec` before this PR with no savepoint, so the commit would now fail and the `RunInTx` defer rolls back the insert that just succeeded. Oracle's a declared dialect, so it's not hypothetical.

Worth confirming against bun's oracle dialect before merge. If it doesn't emit a no-op RELEASE, I'd gate the savepoint to Postgres, the only dialect that actually needs the aborted-transaction recovery, rather than take it on every path. wdyt?

catalog/sql/sql_test.go
```diff
@@ -1911,6 +1917,253 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+	return strings.HasPrefix(strings.ToUpper(strings.TrimSpace(query)), "ROLLBACK TO SAVEPOINT")
+}
+
+func (c *pgAbortConn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
```
laskoviymishka, line 2034: One robustness gap in the emulation: if the wrapped conn ever doesn't implement `ExecerContext` (same for `QueryerContext` just below), this returns `driver.ErrSkip`, and database/sql falls back to prepare-then-exec, which skips the interceptor entirely. The injected failure never fires, the test sees a clean `CreateNamespace`, and it fails on a confusing nil-vs-sentinel mismatch instead of "the simulation didn't run." sqliteshim implements both today so it's not live, but nothing pins it.

Asserting `drv.insertAttempted.Load()` after the call in both race tests would catch that, and doubles as proof the savepoint path actually ran rather than getting skipped.

laskoviymishka (changes requested):
Everything from round 3 landed. The race-loser return is a plain `fmt.Errorf("%w: %s", catalog.ErrNamespaceAlreadyExists, …)` now, matching the pre-check byte-for-byte, so the two paths finally agree and `errors.As` unwraps the same on both. I see zeroshade withdrew the join-the-insert-error suggestion, so we're settled there. The rollback and re-check branches `errors.Join` their diagnostics now, `failUnrelated` pins the negative case, and the mssql `SAVE TRANSACTION` question got looked at. zeroshade's already approved and deferred my threads back to me, so consider those cleared.

One new thing before merge, and it's the other half of the multi-dialect savepoint worry from last round. I checked mssql's `SAVE TRANSACTION` on the create side, but the commit side has its own dialect gap: `sp.Commit()` on a savepoint-backed tx emits `RELEASE SAVEPOINT`, and Oracle has no such statement. It supports `SAVEPOINT` and `ROLLBACK TO SAVEPOINT` but releases savepoints implicitly. bun only skips the RELEASE for the mssql feature flag as far as I can tell, so on Oracle the commit should fail and roll the insert back.

Oracle's a declared dialect and it worked before this PR (plain `Exec`, no savepoint), so this would be a regression on the happy path, not a pre-existing gap. I've left the details inline. Worth confirming against bun's oracle dialect before we merge, and if it holds up, I'd gate the savepoint to Postgres, which is the only dialect that actually needs it, rather than take it on every path.

The rest is small and inline: an `insertAttempted` assertion so a missed intercept fails loudly instead of as a confusing error-shape mismatch, and a comment pinning the MySQL re-check as the first consistent read. Neither blocks.

Sort the Oracle path and I'm happy to approve.

zeroshade (commented):
All prior blockers (Postgres 25P02, Oracle RELEASE SAVEPOINT, error shape, dropped diagnostics, negative case) are genuinely fixed and mutation-verified, but the Postgres gate left the non-Postgres recovery branch and the Postgres happy path at coverage count 0.

Re-review verification: 11 of 13 prior findings confirmed fixed at `25fd793` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *partially fixed* — `laskoviymishka` R1: TestCreateNamespaceConcurrent cannot tell whether the new branch ran; needs a deterministic pin

```text
go build ./... => OK; go vet ./catalog/sql/ ./catalog/hadoop/ => OK; go test ./catalog/sql/ -count=1 => ok (3 consecutive runs); go test ./catalog/sql/ -count=1 -race => ok (3.612s); TestCreateNamespaceConcurrent / LosesInsertRace / InsertFailureSurfacesOriginalError all PASS; unmutated concurrency test 0/25 spurious failures; coverage run 79.5% stmts; 4 mutation probes (savepoint removal, non-PG recovery removal, '&& exists' removal, exists-probe forced true) each produced the expected pass/fail; 1 throwaway probe file written and deleted, worktree ends with zero modifications.
```

</details>

---

>

catalog/sql/sql.go
```diff
@@ -1269,9 +1269,46 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
+
+			if err = sp.Commit(); err != nil {
+				return fmt.Errorf("error releasing savepoint for namespace '%s': %w", namespace, err)
+			}
+
+			return nil
+		}
+
+		if _, err := tx.NewInsert().Model(&toInsert).Exec(ctx); err != nil {
+			return recheck(err)
```
zeroshade, line 1311: **major** — Non-Postgres recovery branch has no deterministic test; the PR's own repro dialect is unpinned

The Postgres gate added in 25fd793 moved the deterministic driver test entirely onto the PG branch, so the SQLite/MySQL/Oracle/MSSQL dispatch at 1310-1312 is covered only by the scheduling-dependent TestCreateNamespaceConcurrent. This re-opens `laskoviymishka`'s round-1 blocker for exactly the dialect in the PR description's repro (the quoted error is SQLite's 1555). Add a SQLite-dialect deterministic test using a NON-aborting driver. Note a dialect parameter on newAbortCatalog will NOT work: pgAbortDriver unconditionally sets `aborted` on the insert and only ROLLBACK TO SAVEPOINT clears it, so a SQLite-dialect run never clears it and the re-check fails with errEmulatedAborted. It needs a separate ~40-line driver reusing isNamespaceInsert/isNamespaceExistsProbe/boolRows.

```text
go tool cover: 'catalog/sql/sql.go:1311.4,1312.1 1 0' (count 0). Reverting 1310-1312 to the pre-PR raw error leaves 'go test ./catalog/sql/ -count=1' => ok. Detection rate measured over 15 independent single runs: 13/15. Throwaway probe with a non-aborting dupDriver at SQLite dialect: unmodified head => 'race-loser err = namespace already exists: my-iceberg-db-mpifixpfynaunvszosha' PASS; reverted mutant => 'error inserting namespace properties ...: UNIQUE constraint failed' FAIL.
```

</details>

catalog/sql/sql.go
```diff
@@ -1269,9 +1269,46 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
+			if _, err = sp.NewInsert().Model(&toInsert).Exec(ctx); err != nil {
+				if rbErr := sp.Rollback(); rbErr != nil {
+					return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(err, rbErr))
+				}
+
+				return recheck(err)
+			}
+
+			if err = sp.Commit(); err != nil {
+				return fmt.Errorf("error releasing savepoint for namespace '%s': %w", namespace, err)
```
zeroshade, line 1304: **minor** — Postgres happy path (SAVEPOINT/RELEASE SAVEPOINT) is never executed by any test

Both pgAbortDriver tests force the insert to fail, so sp.Commit() at 1303 and the 'return nil' at 1307 never run. This is newly added code on the hot path of every successful CreateNamespace against Postgres, and there is no real Postgres anywhere in CI (grep over .github/workflows and dev/ finds none). A success-path assertion through the existing pg-dialect-over-sqlite harness would cover it cheaply.

catalog/sql/sql.go
```diff
@@ -1269,9 +1269,46 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
+			}
+			if checkErr != nil {
+				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(insertErr, checkErr))
+			}
+
+			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, insertErr)
+		}
+
+		// Only Postgres aborts the whole tx on a failed insert, so only it needs a
+		// savepoint to keep the re-check runnable; Oracle has no RELEASE SAVEPOINT.
```
zeroshade: **nit** — Gate comment conflates two independent justifications

'Only Postgres aborts the whole tx on a failed insert, so only it needs a savepoint to keep the re-check runnable; Oracle has no RELEASE SAVEPOINT.' These are two separate facts: the first justifies why PG needs the savepoint, the second why it must not be applied universally. As written, Oracle reads like the reason for a Postgres-only gate. Also, the 'only Postgres' claim is asserted for MSSQL without evidence (SQL Server's behavior depends on XACT_ABORT); scoping the sentence to the dialects actually reasoned about would be more honest.

catalog/sql/sql_test.go
```diff
@@ -1911,6 +1917,384 @@ func (s *SqliteCatalogTestSuite) TestLoadEmptyNamespaceProperties() {
...
+}
+
+// Exercises the Postgres savepoint happy path: SAVEPOINT -> insert -> RELEASE.
+func (s *SqliteCatalogTestSuite) TestCreateNamespaceSucceedsThroughSavepoint() {
+	namespace := table.Identifier{databaseName()}
+	cat, drv := s.newAbortCatalog(insertSucceeds)
+
+	err := cat.CreateNamespace(context.Background(), namespace, nil)
+	s.Require().NoError(err)
+	s.True(drv.insertAttempted.Load(), "the emulated insert must have run")
```
zeroshade, line 2270: **major** — Savepoint happy-path test cannot distinguish RELEASE SAVEPOINT from ROLLBACK TO SAVEPOINT — a silent data-loss mutant survives

TestCreateNamespaceSucceedsThroughSavepoint asserts only s.Require().NoError(err) and drv.insertAttempted, neither of which depends on whether the savepoint was released or rolled back. The whole point of the new Postgres branch is that sp.Commit() emits RELEASE SAVEPOINT (bun v1.2.18 db.go:670) so the insert survives into the outer tx; if it instead emitted ROLLBACK TO SAVEPOINT (db.go:691) CreateNamespace would return nil having created nothing — silent data loss on the hot path of every successful Postgres CreateNamespace, with no real Postgres anywhere in CI. Fix is one assertion: the pgAbortDriver only intercepts EXISTS probes (isNamespaceExistsProbe), so a plain COUNT(*) on iceberg_namespace_properties, or CheckNamespaceExists via a non-intercepting handle, passes through to the real sqlite base and pins it.

```text
Mutation `sp.Commit()` -> `sp.Rollback()` at sql.go:1303. PR's own test: 'ok github.com/apache/iceberg-go/catalog/sql 0.753s'; full suite also 'ok ... 0.980s'. Probe asserting persistence: unmutated 'PROBE: persisted namespace property rows = 1' PASS; mutated 'PROBE: persisted namespace property rows = 0' / '"0" is not positive' / 'CreateNamespace returned nil but persisted no namespace rows' FAIL. Confirmed bun Tx.BeginTx sets a non-empty savepoint name (db.go:753-776), so Commit/Rollback really are RELEASE/ROLLBACK TO, not whole-tx operations.
```

</details>

zeroshade (commented):
The non-Postgres recovery branch is now deterministically pinned (prior major fixed), but the new Postgres savepoint happy-path test is vacuous: swapping RELEASE SAVEPOINT for ROLLBACK TO SAVEPOINT leaves the whole suite green while CreateNamespace returns nil and persists zero rows.

Re-review verification: 1 of 2 prior findings confirmed fixed at `dbb54f9` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *partially fixed* — minor — Postgres happy path (sp.Commit()/RELEASE SAVEPOINT at 1303, return nil at 1307) is never executed by any test

```text
go build ./catalog/... ; go vet ./catalog/sql/... (both clean). go test ./catalog/sql/ -run TestSqlCatalog -race -count=1 => ok 3.793s. Coverage: -covermode=atomic -coverprofile; sql.go:1311 count 1, 1303 count 1, 1307 count 1 (all 0 at 25fd793); 1281 (recheck checkErr!=nil) still 0. go test -run 'TestSqlCatalog/TestCreateNamespace' -race -count=5 => ok 2.506s (no flakiness). Mutation A (drop non-PG recheck dispatch) => TestCreateNamespaceLosesInsertRaceSQLite RED. Mutation B (sp.Commit->sp.Rollback) => suite GREEN (gap). Mutation C (drop `&& exists` from recheck) => TestCreateNamespaceInsertFailureSurfacesOriginalError RED. Mutation D (force PG gate false) => TestCreateNamespaceLosesInsertRace RED. All mutations reverted via git checkout --; final git status --porcelain empty at dbb54f9.
```

</details>

---

>

catalog/sql/sql.go
```diff
@@ -1269,9 +1269,46 @@ func (c *Catalog) CreateNamespace(ctx context.Context, namespace table.Identifie
...
+				return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, errors.Join(insertErr, checkErr))
+			}
+
+			return fmt.Errorf("error inserting namespace properties for namespace '%s': %w", namespace, insertErr)
+		}
+
+		// Postgres aborts the whole tx on a failed insert, so only it needs the
+		// savepoint for the re-check; SQLite/MySQL don't, and Oracle can't RELEASE it.
+		if tx.Dialect().Name() == dialect.PG {
+			sp, err := tx.BeginTx(ctx, nil)
```
zeroshade, line 1290: **nit** — Postgres-only gate on the savepoint branch is unpinned by any test

The dialect gate exists so Oracle (no RELEASE SAVEPOINT) and SQLite/MySQL (statement-level rollback) skip the savepoint. Removing it is behaviourally invisible to the suite because sqlite supports savepoints, so the emulation cannot distinguish gated from ungated. Cheaply pinnable: have dupInsertConn.ExecContext (sql_test.go:2188) record whether any statement matching a SAVEPOINT prefix was seen, and assert it was NOT in TestCreateNamespaceLosesInsertRaceSQLite. Follow-up, not a merge blocker.

zeroshade (approved):
The savepoint happy-path test is no longer vacuous — mutating sp.Commit() to sp.Rollback() now drives TestCreateNamespaceSucceedsThroughSavepoint RED at sql_test.go:2281, closing the last open finding.

Re-review verification: 3 of 3 prior findings confirmed fixed at `874dbb9` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust).

```text
go build ./catalog/... OK; go vet ./catalog/sql/ OK. go test ./catalog/sql/ -count=1 -race -timeout=300s => ok 3.924s. go test ./catalog/sql/ -run 'TestSqlCatalog/TestCreateNamespace' -count=3 -race => ok 2.239s (no flakiness). Mutation 1 (sp.Commit->sp.Rollback, sql.go:1303) => TestCreateNamespaceSucceedsThroughSavepoint RED at sql_test.go:2281. Mutation 2 (PG branch disabled, sql.go:1289) => TestCreateNamespaceLosesInsertRace RED. Mutation 3 (PG gate forced true, sql.go:1289) => GREEN (nit above). All mutations reverted; final git status --porcelain empty; catalog/... builds.
```

</details>

---

>

laskoviymishka (approved):
This is good to merge!

PR #1640, opened 2026-08-04
fix(hive): jitter lock-retry waits to decorrelate contending clients

1991santhu:
## The problem

`calculateBackoff` is a pure function of the attempt number and the configured bounds, so every client computes the same sequence of delays — 100ms, 200ms, 400ms, 800ms on the defaults.

That matters more here than in a typical retry loop, because of *why* a client is in the loop. `acquireLocks` only retries when the lock was not granted immediately, which means another client holds it. Multiple waiters are the precondition, not an unlucky coincidence. They therefore sleep for identical durations and issue their `CheckLock` calls at the same instants, so each polling round reaches the metastore as a burst — the load pattern backoff exists to prevent.

## The change

```go
waitTime := applyJitter(
    calculateBackoff(attempt, opts.LockMinWaitTime, opts.LockMaxWaitTime),
    opts.LockMinWaitTime,
    opts.LockMaxWaitTime,
)
```

**The jitter is added on top of the interval, not centred on it.** Equal jitter (`d/2 + rand(d/2)`) is the more familiar formula, but it halves the wait, which would let a caller poll sooner than the `lock-check-min-wait-time` they configured. The result is drawn from `[d, min(2d, maxWait)]`.

**Once the sequence saturates at `maxWait` there is no headroom left to add into.** That is not an edge case — it is the state a persistently contended client stays in, so leaving it unjittered would put clients back in lockstep exactly when it matters most. The wait is spread *downwards* there instead. A cap is an upper bound on the polling interval rather than a target, so drawing below it breaks no contract.

**How far down it may draw is the subtle part.** Flooring at half the interval would let the first saturated attempt wait *less* than the attempt before it, because the last unsaturated interval lands in `[maxWait/2, maxWait)`. Concretely, with `lock-check-min-wait-time=100ms` and `lock-check-max-wait-time=1s`: attempt 3 draws from `[800ms, 1s]`, but a `maxWait/2` floor would give attempt 4 `[500ms, 1s]`. Both bounds still hold, but a later check polling sooner than an earlier one inverts the property the retry loop exists to provide.

The floor is therefore the last interval the doubling sequence produced before it hit the cap, recovered by replaying the sequence. That value is a bound the schedule has already cleared, and it cannot fall below `minWait`, because that is where the sequence starts.

**Jitter is applied at the call site, not inside `calculateBackoff`.** `TestCalculateBackoff` asserts exact durations per attempt, and determinism is a reasonable property for that function to keep. Leaving the calculation pure keeps the existing contract and its tests untouched, and makes the new behaviour separately testable.

## Reachability, stated plainly

The saturated path is config-gated. The defaults — `lock-check-min-wait-time=100ms`, `lock-check-max-wait-time=1 minute`, `lock-check-retries=4` — top out at 800ms and **never** reach the cap. Hitting it needs a lowered `lock-check-max-wait-time` or roughly `lock-check-retries >= 11`. The below-cap additive jitter is what applies on the defaults.

## Testing

`go build ./...`, `go vet ./catalog/hive/`, and `gofmt` are clean. The full `catalog/hive` package passes, and the jitter tests were run with `-count=5` to check for flakiness. `TestCalculateBackoff` is unmodified and still passes.

New tests:

| Test | Asserts |
|---|---|
| `TestApplyJitterBelowCapNeverShorterThanInput` | result in `[d, min(2d, maxWait)]` across three magnitudes |
| `TestApplyJitterMinimumIsMonotonicAcrossTheCap` | no attempt may draw below an earlier attempt's guaranteed floor |
| `TestApplyJitterAtCapFloorsAtTheLastUncappedInterval` | the boundary above, pinned to the exact numbers |
| `TestApplyJitterAtCapStaysWithinBoundsAndVaries` | the saturated wait still varies, so clients do not re-lockstep |
| `TestApplyJitterAtCapNeverPollsSoonerThanMinWait` | the downward spread respects a configured minimum |
| `TestApplyJitterVaries` | fails if the implementation regresses to deterministic |
| `TestApplyJitterNonPositive` | non-positive inputs pass through unchanged |

The two monotonicity tests were checked against the defect they describe: reverting only the floor logic fails both, with draws around 500–750ms against an 800ms floor.

The monotonicity test compares against the exact interval rather than a sampled minimum. Sampling both sides compares two noisy values sitting a hair above the same true floor, and which one lands lower is chance — that version was flaky and was replaced.

## Note

I have not reproduced multi-client metastore contention; the reasoning is from the code. If maintainers would prefer a different floor at the cap, or the jitter bound made configurable rather than fixed at one interval, both are small changes and I am happy to make them.

### DIFF
--- a/catalog/hive/lock.go
+++ b/catalog/hive/lock.go
@@ -208,10 +208,7 @@ func applyJitter(d, minWait, maxWait time.Duration) time.Duration {
 	// is itself >= maxWait the loop cannot run at all, and options.go accepts that
 	// configuration, so leaving the floor at d/2 there would allow a wait of a third
 	// of the configured minimum.
-	floor := d / 2
-	if minWait > floor {
-		floor = minWait
-	}
+	floor := max(minWait, d/2)
 	for scheduled := minWait; scheduled > 0 && scheduled < maxWait; scheduled <<= 1 {
 		if scheduled > floor {
 			floor = scheduled
--- a/catalog/hive/lock.go
+++ b/catalog/hive/lock.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand/v2"
 	"slices"
 	"strings"
 	"time"
@@ -100,7 +101,11 @@ func acquireLocks(ctx context.Context, client HiveClient, identifiers []tableLoc
 	// If not acquired immediately, wait and retry
 	for attempt := 0; attempt < opts.LockRetries; attempt++ {
 		// Wait before checking again
-		waitTime := calculateBackoff(attempt, opts.LockMinWaitTime, opts.LockMaxWaitTime)
+		waitTime := applyJitter(
+			calculateBackoff(attempt, opts.LockMinWaitTime, opts.LockMaxWaitTime),
+			opts.LockMinWaitTime,
+			opts.LockMaxWaitTime,
+		)
 
 		select {
 		case <-ctx.Done():
@@ -158,6 +163,75 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
 	}
 
 	return minWait << attempt
+}
+
+// applyJitter spreads a backoff interval by adding a random amount on top of it,
+// bounded so the result never exceeds maxWait.
+//
+// calculateBackoff is a pure function of the attempt number and the configured
+// bounds, so every client waiting on the same table lock computes an identical
+// sequence of delays. Contention is the precondition for entering the retry loop
+// at all, which means those clients are waiting simultaneously by construction:
+// they re-check the lock in lockstep, and each round of CheckLock calls arrives
+// at the metastore as a burst. Spreading the wait decorrelates them.
+//
+// The jitter is added rather than subtracted so that the result is never shorter
+// than the interval calculateBackoff produced. That keeps the guarantee implied
+// by the lock-check-min-wait-time property: a caller who configures a minimum
+// wait never polls sooner than it.
+//
+// Once the backoff saturates at maxWait there is no headroom left to add into.
+// calculateBackoff reaches that point deliberately, so leaving it unjittered
+// would put contending clients back in lockstep for every retry after the
+// sequence tops out. The wait is therefore spread downwards instead. Drawing
+// below maxWait breaks no contract, because it is an upper bound on the polling
+// interval rather than a target.
+//
+// How far down it may draw is the subtle part. Flooring at half the interval is
+// wrong: the last attempt that did not saturate waited for its own full interval,
+// which is somewhere in [maxWait/2, maxWait), so a floor of maxWait/2 lets the
+// first saturated attempt wait less than the attempt before it did. That inverts
+// the one property callers rely on, which is that the wait between lock checks
+// only ever grows. The floor is therefore the last interval the doubling sequence
+// produced before it hit the cap, recovered by replaying the sequence. It is a
+// bound the schedule has already cleared, so honouring it keeps the guaranteed
+// minimum monotonic across the saturation boundary, and it can never fall below
+// minWait because minWait is where the sequence starts.
+func applyJitter(d, minWait, maxWait time.Duration) time.Duration {
+	if d <= 0 {
+		return d
+	}
+
+	// A caller that hands in an interval already past the cap is outside the
+	// contract; leave it exactly as given rather than silently reshaping it.
+	headroom := maxWait - d
+	if headroom < 0 {
+		return d
+	}
+
+	// Add up to another full interval, without exceeding the configured maximum.
+	extra := d
+	if headroom < extra {
+		extra = headroom
+	}
+	if extra > 0 {
+		return d + time.Duration(rand.Int64N(int64(extra)+1))
+	}
+
+	// Replay the doubling sequence and keep the largest interval that still fitted
+	// under the cap. The guard on scheduled keeps a non-positive or overflowing
+	// minWait from spinning here; in that case the half-interval floor stands.
+	floor := d / 2
+	for scheduled := minWait; scheduled > 0 && scheduled < maxWait; scheduled <<= 1 {
+		if scheduled > floor {
+			floor = scheduled
+		}
+	}
+	if floor >= d {
+		return d
+	}
+
+	return d - time.Duration(rand.Int64N(int64(d-floor)+1))
 }
 
 func (l *HiveLock) Release(ctx context.Context) error {
--- a/catalog/hive/lock.go
+++ b/catalog/hive/lock.go
@@ -165,38 +165,20 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
 	return minWait << attempt
 }
 
-// applyJitter spreads a backoff interval by adding a random amount on top of it,
-// bounded so the result never exceeds maxWait.
+// applyJitter spreads a backoff interval so clients contending for the same lock
+// stop re-polling in lockstep. calculateBackoff is a pure function of the attempt
+// and the configured bounds, and contention is the precondition for retrying at
+// all, so without this every waiter issues its CheckLock calls at the same instants
+// and each round reaches the metastore as a burst.
 //
-// calculateBackoff is a pure function of the attempt number and the configured
-// bounds, so every client waiting on the same table lock computes an identical
-// sequence of delays. Contention is the precondition for entering the retry loop
-// at all, which means those clients are waiting simultaneously by construction:
-// they re-check the lock in lockstep, and each round of CheckLock calls arrives
-// at the metastore as a burst. Spreading the wait decorrelates them.
-//
-// The jitter is added rather than subtracted so that the result is never shorter
-// than the interval calculateBackoff produced. That keeps the guarantee implied
-// by the lock-check-min-wait-time property: a caller who configures a minimum
-// wait never polls sooner than it.
-//
-// Once the backoff saturates at maxWait there is no headroom left to add into.
-// calculateBackoff reaches that point deliberately, so leaving it unjittered
-// would put contending clients back in lockstep for every retry after the
-// sequence tops out. The wait is therefore spread downwards instead. Drawing
-// below maxWait breaks no contract, because it is an upper bound on the polling
-// interval rather than a target.
-//
-// How far down it may draw is the subtle part. Flooring at half the interval is
-// wrong: the last attempt that did not saturate waited for its own full interval,
-// which is somewhere in [maxWait/2, maxWait), so a floor of maxWait/2 lets the
-// first saturated attempt wait less than the attempt before it did. That inverts
-// the one property callers rely on, which is that the wait between lock checks
-// only ever grows. The floor is therefore the last interval the doubling sequence
-// produced before it hit the cap, recovered by replaying the sequence. It is a
-// bound the schedule has already cleared, so honouring it keeps the guaranteed
-// minimum monotonic across the saturation boundary, and it can never fall below
-// minWait because minWait is where the sequence starts.
+// The invariants, in the order the code establishes them:
+//   - below the cap the jitter is added rather than centred, so the wait is never
+//     shorter than the interval calculateBackoff produced;
+//   - at the cap there is no headroom left to add into, so the wait is spread
+//     downward instead, floored at the last interval the sequence produced before
+//     it saturated — a bound the schedule has already cleared, which stops a later
+//     attempt from being allowed to wait less than an earlier one;
+//   - the result never exceeds maxWait and never falls below minWait.
 func applyJitter(d, minWait, maxWait time.Duration) time.Duration {
 	if d <= 0 {
 		return d
@@ -220,8 +202,16 @@ func applyJitter(d, minWait, maxWait time.Duration) time.Duration {
 
 	// Replay the doubling sequence and keep the largest interval that still fitted
 	// under the cap. The guard on scheduled keeps a non-positive or overflowing
-	// minWait from spinning here; in that case the half-interval floor stands.
+	// minWait from spinning here.
+	//
+	// minWait is applied before the replay rather than relying on it. When minWait
+	// is itself >= maxWait the loop cannot run at all, and options.go accepts that
+	// configuration, so leaving the floor at d/2 there would allow a wait of a third
+	// of the configured minimum.
 	floor := d / 2
+	if minWait > floor {
+		floor = minWait
+	}
 	for scheduled := minWait; scheduled > 0 && scheduled < maxWait; scheduled <<= 1 {
 		if scheduled > floor {
 			floor = scheduled
--- a/catalog/hive/lock_test.go
+++ b/catalog/hive/lock_test.go
@@ -508,6 +508,68 @@ func TestApplyJitterAtCapFloorsAtTheLastUncappedInterval(t *testing.T) {
 	}
 }
 
+// A caller may configure minWait above maxWait; options.go accepts it, and
+// calculateBackoff resolves it by returning maxWait. The downward spread must not
+// then draw below the configured minimum. Flooring at half the interval would
+// permit a third of it, because the replay loop cannot run when minWait is already
+// past the cap.
+func TestApplyJitterHonoursMinWaitAboveMaxWait(t *testing.T) {
+	minWait := 90 * time.Second
+	maxWait := 60 * time.Second
+
+	d := calculateBackoff(0, minWait, maxWait)
+	require.Equal(t, maxWait, d, "calculateBackoff resolves this configuration to the cap")
+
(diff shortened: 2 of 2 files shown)


1991santhu:
Corrected two things in my own patch since opening this.

First, the jitter was centred, so it could poll sooner than `lock-check-min-wait-time` allows. It adds now instead.

Second, `calculateBackoff` returns `maxWait` outright in three branches. My first version clamped the extra to the headroom, which is zero there, so those retries came back with no jitter at all. It now spreads downward in that case, floored at `minWait`. The floor matters when `minWait > maxWait/2`, where the old code would have polled at 30s against a configured 40s.

I also ran the numbers rather than just asserting the problem. 50 clients on one lock, defaults of 100ms/60s/4 retries: without jitter all 50 re-check inside the same 10ms window on every attempt. With it, about 10 on the first retry and 3 by the fourth. That's a simulation of arrival times, not something I've seen in production.

`go build ./...` and `go test -count=2 ./catalog/hive/` pass.

catalog/hive/lock.go
```diff
@@ -160,6 +165,60 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
...
+//
+// The jitter is added rather than subtracted so that the result is never shorter
+// than the interval calculateBackoff produced. That keeps the guarantee implied
+// by the lock-check-min-wait-time property: a caller who configures a minimum
+// wait never polls sooner than it.
+//
+// Once the backoff saturates at maxWait there is no headroom left to add into.
+// calculateBackoff reaches that point deliberately, so leaving it unjittered
+// would put contending clients back in lockstep for every retry after the
+// sequence tops out. The wait is therefore spread downwards instead, floored at
```
tanmayrauth: Small doc nit: this says the downward spread is "floored at minWait", but the code at line 211 floors at max(d/2, minWait) — so the real floor is d/2, and minWait only takes over when it's larger than half the interval. Worth  rewording to "floored at half the interval, or at minWait when minWait exceeds half" so a future reader doesn't assume the at-cap wait can fall all the way to minWait when it can't.

tanmayrauth (approved):
Mechanics are correct — bounds hold in both branches, no rand.Int64N(0) path, rand/v2 is concurrency-safe for the parallel acquireLocks callers. One minor comment-accuracy nit inline

1991santhu:
Good catch, you're right. The code floors at `max(d/2, minWait)`, so half the interval is the real floor and `minWait` only takes over when it's bigger than that. As written the comment would have someone expecting the at-cap wait to reach `minWait`, which it can't.

Reworded to "floored at half the interval, or at minWait when minWait exceeds half", and said explicitly that the `minWait` half only matters when `minWait` is more than half of `maxWait`.

Thanks for checking the bounds in both branches and the rand/v2 concurrency point. Pushed in 4ab50a5.

One note for whoever picks this up: the `integration-test (4.0, spark4)` failure is not from this change. The job dies in setup while pulling images, before any test runs:

```
rest Error unauthorized: authentication required
Error response from daemon: unauthorized: authentication required
make: *** [Makefile:46: integration-setup-spark4] Error 1
```

The commit it ran against only edits a doc comment, so it can't have broken an image pull. Looks like registry auth or rate limiting. A re-run should clear it.

catalog/hive/lock.go
```diff
@@ -160,6 +165,75 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
...
+		extra = headroom
+	}
+	if extra > 0 {
+		return d + time.Duration(rand.Int64N(int64(extra)+1))
+	}
+
+	// Replay the doubling sequence and keep the largest interval that still fitted
+	// under the cap. The guard on scheduled keeps a non-positive or overflowing
+	// minWait from spinning here; in that case the half-interval floor stands.
+	floor := d / 2
```
zeroshade: Non-blocking: consider clamping this initial floor to at least `minWait`, or rejecting `minWait >= maxWait` while parsing the options. With `minWait=90s` and `maxWait=60s`, `calculateBackoff` returns 60s, the replay guard is false on entry, and the floor remains 30s. That permits waits down to one third of the configured minimum, contrary to the documented promise. `catalog/hive/options.go:91-105` currently makes this configuration reachable. A focused 90s/60s regression case would also pin the behavior.

catalog/hive/lock.go
```diff
@@ -160,6 +165,75 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
 	return minWait << attempt
 }
 
+// applyJitter spreads a backoff interval by adding a random amount on top of it,
```
zeroshade: Non-blocking: consider condensing this comment. The detailed floor derivation is useful design context, but a roughly 31-line comment is disproportionate to this small helper; the full rationale could live in the commit message with a shorter invariant-focused note here.

catalog/hive/lock.go
```diff
@@ -100,7 +101,11 @@ func acquireLocks(ctx context.Context, client HiveClient, identifiers []tableLoc
 	// If not acquired immediately, wait and retry
 	for attempt := 0; attempt < opts.LockRetries; attempt++ {
 		// Wait before checking again
-		waitTime := calculateBackoff(attempt, opts.LockMinWaitTime, opts.LockMaxWaitTime)
+		waitTime := applyJitter(
```
zeroshade, line 104: Non-blocking: consider adding one `acquireLocks`-level test that exercises this wiring and cancellation path, plus an assertion that the aggregate retry delay stays within the intended bound. The helper tests cover the distribution thoroughly, but do not execute the new call site.

zeroshade (commented):
This change is correct: I verified that below the cap it draws uniformly from `[d, min(2d, maxWait)]`, while capped attempts spread downward to keep independently polling clients decorrelated. This is not AWS decorrelated jitter, but the independent draws achieve the intended decorrelation. `math/rand/v2.Int64N` uses the lock-free per-P runtime source, so caller seeding is unnecessary, and cancellation remains honored by the existing `ctx.Done()` select at `catalog/hive/lock.go:110-113`.

Optional follow-ups:

- `catalog/hive/lock.go:104`: the worst-case total wait for the default four retries roughly doubles from 1.5s to 3s; consider mentioning that behavioral change in the release notes.
- `catalog/hive/options.go:91-105`: the parser also accepts negative durations because `time.ParseDuration` permits them. This is pre-existing and the `d <= 0` guard handles it, but validating the configured range here would make the contract clearer.

1991santhu:
Thanks — the 90s/60s case is a real bug and it was a regression in my own previous commit, not something pre-existing. The replay loop only raises the floor while walking intervals below `maxWait`, so when `minWait >= maxWait` it never runs at all and the floor stayed at `d/2`. I reproduced it before changing anything: lowest observed wait was 30.0014s against a configured 90s minimum, exactly as you described.

Fixed in a48bc64 by applying `minWait` as a floor before the replay rather than relying on the replay to reach it. Where `minWait` is at or past the cap the floor now exceeds the interval and the wait comes back unspread, which is the only honest answer when the configured minimum can't be respected any other way. Regression test uses your 90s/60s numbers, and I checked it fails without the fix.

Shortened the comment to the invariants. You were right that 31 lines was disproportionate, and I'll note the removed text ended by claiming the wait could never fall below `minWait` — which is exactly what it did. The derivation is in the commit message now.

Added the `acquireLocks`-level test. It runs the retry loop through the real call site and bounds the aggregate delay. One caveat worth stating: the lower bound is asserted tightly but the ceiling is loose, because a wall-clock upper bound is the kind of assertion that flakes on a busy runner, and a flaky timing test is worse than none.

On the release note: agreed, and I can confirm the number. Defaults are 100ms/1min/4 retries, so the unjittered schedule is 100+200+400+800ms = 1.5s and each draw can add up to one further interval, giving a worst case of 3s.

On `options.go` accepting negative durations and `minWait >= maxWait`: agreed that validating there would make the contract clearer, but both predate this PR and fixing them changes behaviour for configurations that currently parse, so I'd rather not widen this one. Happy to open a follow-up if you want it.

catalog/hive/lock.go
```diff
@@ -160,6 +165,65 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
...
+	// minWait from spinning here.
+	//
+	// minWait is applied before the replay rather than relying on it. When minWait
+	// is itself >= maxWait the loop cannot run at all, and options.go accepts that
+	// configuration, so leaving the floor at d/2 there would allow a wait of a third
+	// of the configured minimum.
+	floor := d / 2
+	if minWait > floor {
+		floor = minWait
+	}
```
zeroshade:
```suggestion
	floor := max(minWait, d / 2)
```

zeroshade (approved):
just one nit, otherwise looks great!

1991santhu:
Applied, thanks. The builtin is already how `catalog/rest/scan_planning.go` spells its retry delay, so this matches the local style better than what I had.

catalog/hive/lock.go
```diff
@@ -160,6 +165,62 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
...
+// and each round reaches the metastore as a burst.
+//
+// The invariants, in the order the code establishes them:
+//   - below the cap the jitter is added rather than centred, so the wait is never
+//     shorter than the interval calculateBackoff produced;
+//   - at the cap there is no headroom left to add into, so the wait is spread
+//     downward instead, floored at the last interval the sequence produced before
+//     it saturated — a bound the schedule has already cleared, which stops a later
+//     attempt from being allowed to wait less than an earlier one;
+//   - the result never exceeds maxWait and never falls below minWait.
```
laskoviymishka, line 181: This last bullet isn't quite true when minWait > maxWait. calculateBackoff resolves that config to maxWait, and applyJitter returns maxWait too, so the result sits below minWait. TestApplyJitterHonoursMinWaitAboveMaxWait actually pins exactly that (got == maxWait with minWait=90s, maxWait=60s).

I'd qualify it: never below minWait when minWait <= maxWait, otherwise exactly maxWait.

catalog/hive/lock.go
```diff
@@ -160,6 +165,62 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
...
+		return d
+	}
+
+	// Add up to another full interval, without exceeding the configured maximum.
+	extra := d
+	if headroom < extra {
+		extra = headroom
+	}
+	if extra > 0 {
+		return d + time.Duration(rand.Int64N(int64(extra)+1))
```
laskoviymishka, line 200: Adding up to a full interval here gives us [d, 2d] below the cap. Java's Tasks.exponentialBackoff only jitters by ~10% of the current delay, so a mixed Java/Go fleet hitting the same HMS will spread quite differently, with Go clients sitting in a window twice as wide.

I'm fine with the wider spread if it's deliberate, I'd just call the divergence out in the doc comment so the next person doesn't assume we're matching Java. wdyt?

catalog/hive/lock.go
```diff
@@ -160,6 +165,62 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
...
+	//
+	// minWait is applied before the replay rather than relying on it. When minWait
+	// is itself >= maxWait the loop cannot run at all, and options.go accepts that
+	// configuration, so leaving the floor at d/2 there would allow a wait of a third
+	// of the configured minimum.
+	floor := max(minWait, d/2)
+	for scheduled := minWait; scheduled > 0 && scheduled < maxWait; scheduled <<= 1 {
+		if scheduled > floor {
+			floor = scheduled
+		}
```
laskoviymishka, line 215: This downward-spread branch only runs once the sequence saturates at maxWait, and under the defaults (100ms/60s/4 retries) we top out at 800ms and never get there, so the replay loop is effectively dead code in a default deployment. It's also the most intricate part of the helper, and I'm a little uneasy about the most fragile code being the least exercised in practice.

Is the strict monotonicity across the cap worth it? A flat max(minWait, d/2) floor drops the loop entirely, at the cost of letting the first capped attempt dip slightly below the last uncapped one. If we do keep the loop, I'd at least note plainly that it's only reachable under non-default config. Thoughts?

catalog/hive/lock_test.go
```diff
@@ -404,3 +404,195 @@ func TestLockConfigurationDefaults(t *testing.T) {
...
+// is spread downwards. It must still vary, or every retry after the sequence
+// saturates puts contending clients back in lockstep.
+func TestApplyJitterAtCapStaysWithinBoundsAndVaries(t *testing.T) {
+	minWait := 100 * time.Millisecond
+	maxWait := time.Second
+
+	seen := make(map[time.Duration]struct{})
+	for i := 0; i < 500; i++ {
+		got := applyJitter(maxWait, minWait, maxWait)
+		assert.GreaterOrEqual(t, got, maxWait/2,
```
laskoviymishka, line 437: This lower bound is looser than the code actually guarantees. The floor at the cap is 800ms (the last uncapped interval), but this only asserts >= 500ms, so a regression letting the spread dip to 600-700ms would pass here. TestApplyJitterAtCapFloorsAtTheLastUncappedInterval does pin the real 800ms, so we're covered, but this assertion reads stronger than it is. I'd tighten it to 800ms so the two agree.

Separately, the message ("must not exceed one half interval") is upper-bound wording on a GreaterOrEqual check. I'd flip it to something like "must be at least half of maxWait" so a failure dump isn't confusing.

catalog/hive/lock_test.go
```diff
@@ -404,3 +404,195 @@ func TestLockConfigurationDefaults(t *testing.T) {
...
+	minWait := 100 * time.Millisecond
+	maxWait := time.Second
+
+	assert.Equal(t, 800*time.Millisecond, calculateBackoff(3, minWait, maxWait),
+		"the last interval before the cap")
+	assert.Equal(t, maxWait, calculateBackoff(4, minWait, maxWait),
+		"the first interval at the cap")
+
+	for i := 0; i < 2000; i++ {
+		assert.GreaterOrEqual(t, applyJitter(maxWait, minWait, maxWait), 800*time.Millisecond,
```
laskoviymishka, line 506: The pinned cases all use clean doubling ratios (100ms to 1s, floor 800ms). The floor's hard case is a non-power-of-2 ratio where d/2 and the real last-uncapped interval diverge, e.g. 300ms/1000ms, where the floor must be 600ms, not 500ms. I'd add one such case here; that's where the replay loop earns its keep and the only spot the arithmetic could quietly go wrong.

catalog/hive/lock_test.go
```diff
@@ -404,3 +404,195 @@ func TestLockConfigurationDefaults(t *testing.T) {
...
+	start := time.Now()
+	lock, err := acquireLocks(ctx, mockClient,
+		[]tableLockIdentifier{{database: "testdb", table: "testtable"}}, opts)
+	elapsed := time.Since(start)
+
+	require.Error(t, err)
+	require.Nil(t, lock)
+	assert.ErrorIs(t, err, ErrLockAcquisitionFailed)
+
+	assert.GreaterOrEqual(t, elapsed, 150*time.Millisecond,
```
laskoviymishka, line 565: This asserts real elapsed wall-clock time, so on a busy CI box the >= 150ms lower bound can flake, and the test sleeps for real every run. The additive never-shorter property is already covered by TestApplyJitterBelowCapNeverShorterThanInput without touching the clock.

I'd drop the timing assertions and keep this as a wiring check that's count-based (assert CheckLock was called exactly 4 times) so we exercise the call site without the clock dependency.

laskoviymishka (approved):
This is good to go from my side, approving. The core mechanics are already well vetted (bounds in both branches, rand/v2 concurrency-safety, cancellation through the existing ctx.Done() select), so I won't re-tread those; agreed on all of it. Everything below is follow-up, none of it blocking the approval.

The one I'd genuinely like a call on is the downward-spread branch at the cap. Under the default config (100ms/60s/4 retries) the sequence tops out at 800ms and never saturates, so that replay loop is dead code in a normal deployment, and it's simultaneously the most intricate part of the helper. I'm a little uneasy about the most fragile code being the least exercised. Either flatten it to a plainer max(minWait, d/2) floor (a hair shorter at one boundary, much less to reason about), or keep it and note plainly that the branch is only reachable under non-default config. Happy either way, I just don't want it sitting there unmarked.

Smaller stuff, all inline: the at-cap test asserts >= maxWait/2 when the real floor is 800ms (the pinned test covers the real bound, so it's not a safety gap, just weaker than it looks), the doc bullet claiming the result never falls below minWait isn't true when minWait > maxWait, and the wall-clock timing assertion in the acquireLocks test is worth swapping for a count-based check so it can't flake on slow CI.

There are also a couple of pre-existing Java-compat gaps this sits on top of (the backoff scale is 2.0 vs Java's 1.5, and the lock-check config keys don't match Java's names/defaults). Both out of scope here, but worth a tracking issue. None of this gates the approval; treating it all as follow-up.

laskoviymishka:
I think this is overall good to merge.

PR #1643, opened 2026-08-04
fix(view): snapshot property update inputs

fallintoplace:
## Summary

Copy map and slice inputs when creating view property updates.

## Why

The constructors retained caller-owned maps and slices. A later mutation could make the metadata builder and the emitted catalog update disagree.

## What changed

NewSetPropertiesUpdate and NewRemovePropertiesUpdate now copy their inputs, so a staged update cannot change after construction.

## Tests

- Added constructor ownership tests for maps and slices
- Added builder coverage for metadata and emitted changes
- go test ./view

### DIFF
--- a/view/updates_test.go
+++ b/view/updates_test.go
@@ -145,12 +145,25 @@ func TestPropertyUpdateConstructorsCopyInputs(t *testing.T) {
 	assert.Equal(t, []string{"old-property"}, remove.Removals)
 }
 
+func TestPropertyUpdateConstructorsPreserveNilness(t *testing.T) {
+	var nilProps iceberg.Properties
+	emptyProps := iceberg.Properties{}
+	assert.Nil(t, NewSetPropertiesUpdate(nilProps).Updates)
+	assert.NotNil(t, NewSetPropertiesUpdate(emptyProps).Updates)
+
+	var nilRemovals []string
+	emptyRemovals := []string{}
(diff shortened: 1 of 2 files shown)


view/updates_test.go
```diff
@@ -131,3 +131,41 @@ func TestUpdatesUnmarshalJSONReplacesExistingSlice(t *testing.T) {
...
+		SetLoc("location").
+		AddSchema(newTestSchema(0)).
+		AddVersion(newTestVersion(1, 0)).
+		SetCurrentVersionID(1).
+		Build()
+	require.NoError(t, err)
+
+	builder, err := MetadataBuilderFromBase(base)
+	require.NoError(t, err)
+	result, err := builder.SetProperties(props).RemoveProperties(removals).Build()
```
zeroshade: Non-blocking: consider separating `SetProperties`/`RemoveProperties` from `Build`, mutating `props` and `removals` in between, and only then building. That directly exercises the staging interval this fix protects. Explicit nil and non-nil-empty cases would also pin the preservation semantics of `maps.Clone` and `slices.Clone`.

zeroshade (commented):
This change is correct: I verified that `maps.Clone` and `slices.Clone` snapshot caller-owned inputs at constructor time while preserving nil versus non-nil-empty values. The constructor test at `view/updates_test.go:135-145` detects the former aliasing behavior for both maps and slices.

zeroshade (approved):

PR #1645, opened 2026-08-04
fix(table): advance string upper bounds across surrogate gap

fallintoplace:
## Summary

Return a tighter upper bound when a truncated string ends at U+D7FF.

## Why

The old logic incremented the final rune and stopped when the result was invalid. The surrogate range U+D800 through U+DFFF cannot appear in UTF-8, so U+D7FF incorrectly produced no upper bound.

## What changed

The next valid scalar value, U+E000, is now used as the upper bound. This improves pruning without changing query correctness.

## Tests

- Added a regression test for the U+D7FF boundary
- go test ./table/internal
- go test ./table

### DIFF
--- a/table/internal/utils.go
+++ b/table/internal/utils.go
@@ -541,7 +541,7 @@ func TruncateUpperBoundText(s string, trunc int) string {
 		if next, ok := nextValidRune(result[i]); ok {
 			result[i] = next
 
-			return string(result)
+			return string(result[:i+1])
 		}
 	}
 
--- a/table/internal/utils.go
+++ b/table/internal/utils.go
@@ -538,8 +538,7 @@ func TruncateUpperBoundText(s string, trunc int) string {
 
 	result := []rune(s)[:trunc]
 	for i := len(result) - 1; i >= 0; i-- {
-		next := result[i] + 1
-		if utf8.ValidRune(next) {
+		if next, ok := nextValidRune(result[i]); ok {
 			result[i] = next
 
 			return string(result)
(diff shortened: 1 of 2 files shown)


table/internal/utils.go
```diff
@@ -538,8 +538,7 @@ func TruncateUpperBoundText(s string, trunc int) string {
 
 	result := []rune(s)[:trunc]
 	for i := len(result) - 1; i >= 0; i-- {
-		next := result[i] + 1
-		if utf8.ValidRune(next) {
+		if next, ok := nextValidRune(result[i]); ok {
 			result[i] = next
 
 			return string(result)
```
zeroshade: Non-blocking: after carrying past one or more trailing U+10FFFF runes, consider returning `string(result[:i+1])` here. Retaining the trailing maximum runes produces a safe but looser upper bound; returning through the rune that was actually incremented gives the tighter prefix bound.

table/internal/utils_test.go
```diff
@@ -67,6 +67,7 @@ func TestMetricsModePairs(t *testing.T) {
 
 func TestTruncateUpperBoundString(t *testing.T) {
 	assert.Equal(t, "ab", internal.TruncateUpperBoundText("aaaa", 2))
+	assert.Equal(t, "\uE000", internal.TruncateUpperBoundText("\uD7FFx", 1))
```
zeroshade: Non-blocking: consider asserting `utf8.ValidString` and bytewise ordering for the returned bound, since those are the two correctness properties behind this regression. Neighboring scalar values and a preceding-rune carry case would round out the boundary coverage.

zeroshade (commented):
This change is correct: I verified that U+D7FF advances to valid U+E000 while preserving bytewise UTF-8 ordering, and that U+10FFFF carries to an earlier rune or yields no upper bound when every retained rune is maximal. The safety direction is right: a too-low upper bound could silently discard matching rows, whereas this implementation can only produce an occasionally too-high, less selective bound.

zeroshade (approved):

PR #1646, opened 2026-08-04
fix(view): replace metadata state during JSON decoding

fallintoplace:
## Summary

Make view metadata JSON decoding replace the receiver state only after validation succeeds.

## Why

Reusing a metadata value could retain omitted properties or version-log entries. A failed decode could also leave the receiver partially updated.

## What changed

The decoder now uses fresh state, initializes lookup indexes, validates the complete document, and replaces the receiver only after success. The final receiver's indexes are rebuilt as well.

## Tests

- Added coverage for omitted fields being cleared on reuse
- Added lookup-cache replacement coverage
- Added failed-decode state preservation coverage
- go test ./view

### DIFF
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -671,6 +671,7 @@ func (m *metadata) UnmarshalJSON(b []byte) error {
 	}
 
 	*m = next
+	m.init()
 
 	return nil
 }
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -671,6 +671,7 @@ func (m *metadata) UnmarshalJSON(b []byte) error {
 	}
 
 	*m = next
+	// init() rebinds the lazy-index closures to m after the value copy.
 	m.init()
 
 	return nil
--- a/view/metadata.go
+++ b/view/metadata.go
@@ -654,18 +654,25 @@ func (m *metadata) init() {
 
 func (m *metadata) UnmarshalJSON(b []byte) error {
 	type Alias metadata
-	aux := (*Alias)(m)
-
-	aux.FormatVersionValue = -1
-	aux.CurrentVersionIDValue = -1
+	next := metadata{
+		FormatVersionValue:    -1,
+		CurrentVersionIDValue: -1,
+	}
+	aux := (*Alias)(&next)
 
 	if err := json.Unmarshal(b, aux); err != nil {
 		return err
 	}
 
-	m.init()
+	next.init()
 
-	return m.validate()
+	if err := next.validate(); err != nil {
+		return err
+	}
+
+	*m = next
+
+	return nil
 }
 
 // NewMetadata creates a new view metadata object using the provided version, schema, location, and props,
--- a/view/metadata_test.go
+++ b/view/metadata_test.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 	"testing"
 
 	"github.com/apache/iceberg-go"
@@ -141,6 +142,28 @@ func TestUnmarshalViewMetadata(t *testing.T) {
 	assert.Equal(t, []VersionLogEntry{{TimestampMS: 1000, VersionID: 1}}, md.VersionLog())
 	assert.Equal(t, []Representation{NewRepresentation("select * from ns.tbl", "trino")}, md.CurrentVersion().Representations)
 	assert.Equal(t, iceberg.Properties{"prop": "value"}, md.Properties())
+}
+
+func TestMetadataUnmarshalReplacesReceiverState(t *testing.T) {
+	var metadata metadata
+	require.NoError(t, json.Unmarshal([]byte(exampleViewJSON), &metadata))
+
+	var reduced map[string]json.RawMessage
+	require.NoError(t, json.Unmarshal([]byte(exampleViewJSON), &reduced))
+	delete(reduced, "properties")
+	delete(reduced, "version-log")
+	reducedData, err := json.Marshal(reduced)
+	require.NoError(t, err)
+
+	require.NoError(t, json.Unmarshal(reducedData, &metadata))
+	assert.Empty(t, metadata.Props)
+	assert.Empty(t, metadata.VersionLogList)
+
+	invalid := strings.Replace(exampleViewJSON, `"current-version-id": 1`, `"current-version-id": 99`, 1)
+	require.Error(t, json.Unmarshal([]byte(invalid), &metadata))
+	assert.Equal(t, int64(1), metadata.CurrentVersionIDValue)
+	assert.Empty(t, metadata.Props)
+	assert.Empty(t, metadata.VersionLogList)
 }
 
 func TestValidMetadataDeserialization(t *testing.T) {
--- a/view/metadata_test.go
+++ b/view/metadata_test.go
@@ -166,6 +166,42 @@ func TestMetadataUnmarshalReplacesReceiverState(t *testing.T) {
 	assert.Empty(t, metadata.VersionLogList)
 }
 
+func TestMetadataUnmarshalReplacesLookupCaches(t *testing.T) {
+	var metadata metadata
+	require.NoError(t, json.Unmarshal([]byte(exampleViewJSON), &metadata))
+
+	// Prime every lookup cache before replacing the metadata.
+	assert.Equal(t, int64(1), metadata.CurrentVersion().VersionID)
+	assert.Equal(t, 0, metadata.CurrentSchema().ID)
+	assert.Contains(t, metadata.SchemasByID(), 0)
+
+	var replacement map[string]any
+	require.NoError(t, json.Unmarshal([]byte(exampleViewJSON), &replacement))
+	replacement["current-version-id"] = int64(2)
+	replacement["versions"].([]any)[0].(map[string]any)["version-id"] = int64(2)
+	replacement["versions"].([]any)[0].(map[string]any)["schema-id"] = 2
+	replacement["schemas"].([]any)[0].(map[string]any)["schema-id"] = 2
+	replacement["version-log"].([]any)[0].(map[string]any)["version-id"] = int64(2)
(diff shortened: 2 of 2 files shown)


tanmayrauth (approved):

zeroshade (commented):
The change is correct. Verified against `5955033`:

- `next` is a fresh composite literal, so every map and slice on the decoded value (`Props`, `VersionList`, `SchemaList`, `VersionLogList`) is allocated by the decoder rather than merged into whatever the receiver already held. No field is left merged.
- `next` is local and unreachable after `*m = next`, so the assigned value aliases neither a previous metadata nor the input `b`.
- The failure path is fully atomic: both error returns (`view/metadata.go:663-665` for the decode, `:669-671` for `validate()`) precede the assignment, so a rejected document leaves the receiver untouched.
- Round-trip output is unchanged — `init()`'s nil-to-empty normalization of `VersionLogList` already ran before this PR.

The second `init()` at `view/metadata.go:674` is required rather than redundant, and both new tests are meaningful and non-vacuous. Notes inline.

**Non-blocking**

Beyond the inline points, two smaller test gaps: nothing asserts that decoding does not alias or mutate the input `b`, and there is no decode → encode byte-equivalence check after a receiver has been reused.

view/metadata.go
```diff
@@ -654,18 +654,26 @@ func (m *metadata) init() {
...
 	}
 
+	next.init()
+
+	if err := next.validate(); err != nil {
+		return err
+	}
+
+	*m = next
 	m.init()
```
zeroshade, line 675: This call is required, not redundant — which is worth a one-line comment saying so. `init()` installs `sync.OnceValue` closures that capture the *receiver pointer* (`view/metadata.go:647-652`), so the closures copied in by `*m = next` on the line above are bound to the dead `next` local rather than to `m`.

Without a note the call reads as dead code, and deleting it produces a latent bug rather than an immediate one: the lazy indexes still return correct answers, they just keep `next` alive through the closure. That is the kind of defect that survives review indefinitely.

Consider something like:

```go
	// init() binds the lazy-index closures to the receiver; rebind after the copy.
	m.init()
```

`TestMetadataUnmarshalReplacesLookupCaches` already covers the behavior well; this is only about making the reason legible at the call site.

view/metadata.go
```diff
@@ -654,18 +654,26 @@ func (m *metadata) init() {
...
+		FormatVersionValue:    -1,
+		CurrentVersionIDValue: -1,
+	}
+	aux := (*Alias)(&next)
 
 	if err := json.Unmarshal(b, aux); err != nil {
 		return err
 	}
 
+	next.init()
```
zeroshade, line 667: Non-blocking: the schema index is now built twice per successful parse — once here, because `validate()` forces `next.lazySchemasByID()` via `checkVersionSchemasExist`, and again lazily after `m.init()` on line 674 resets the `OnceValue`. Harmless at these sizes.

If you would rather avoid it, one option is to validate against a locally computed index and install the closures once, after the copy.

view/metadata_test.go
```diff
@@ -143,6 +144,64 @@ func TestUnmarshalViewMetadata(t *testing.T) {
...
+	reducedData, err := json.Marshal(reduced)
+	require.NoError(t, err)
+
+	require.NoError(t, json.Unmarshal(reducedData, &metadata))
+	assert.Empty(t, metadata.Props)
+	assert.Empty(t, metadata.VersionLogList)
+
+	invalid := strings.Replace(exampleViewJSON, `"current-version-id": 1`, `"current-version-id": 99`, 1)
+	require.Error(t, json.Unmarshal([]byte(invalid), &metadata))
+	assert.Equal(t, int64(1), metadata.CurrentVersionIDValue)
```
zeroshade: Non-blocking: these three field assertions cannot catch a *partial* regression — a future change that reset `Props` and `VersionLogList` but forgot `SchemaList` would still pass here. `assert.Equal` on the whole struct is not an option, since `reflect.DeepEqual` is always false for a struct holding non-nil func fields (the `sync.OnceValue` closures).

A workable substitute: snapshot `json.Marshal(&metadata)` immediately before the failing decode and assert the bytes are identical afterwards. That covers every serialized field at once, at roughly the cost of the three assertions it replaces.

zeroshade (approved):
Re-reviewed at `f258696`. The three notes from my review of `5955033` are addressed; approving.

**Gates at `f258696`:** `go build ./...`, `go vet ./view/...`, `gofmt -l view/` (clean), `go test ./view/...`, and `go test ./... -count=1` all pass.

**Core fix re-confirmed at the new head.** `next` is still a fresh composite literal, `init()` and `validate()` both complete before the receiver assignment, both error paths return before `*m = next`, and `next` is unreachable afterwards. Serialized output is unchanged from pre-PR.

**On the previous notes:**

- The `init()` comment is accurate — both closures at `view/metadata.go:647` capture `m`, so after `*m = next` the copied func values are still bound to `&next`. The rebind is load-bearing, and now says so at the call site.
- The marshal snapshot is a genuine strengthening over the three field assertions it replaces. There is no custom `metadata.MarshalJSON`; `CurrentVersionIDValue`, `VersionList`, `SchemaList` and `VersionLogList` are always serialized, and `Props` is `omitempty` but non-empty in the invalid document, so an empty-to-populated mutation stays visible. No coverage lost.
- Verified the protection is real by reverting the fix in a scratch tree: `TestMetadataUnmarshalReplacesReceiverState` fails on stale properties, stale version log, and the new before/after snapshot; `TestMetadataUnmarshalReplacesLookupCaches` fails after the failed decode.

**Non-blocking — `TestMetadataUnmarshalDoesNotMutateInputAndRoundTripsAfterReuse` does not currently protect anything.** It passes with the pre-PR decoder restored (checked in an isolated tree). Both of its assertions are weaker than they look:

- `assert.Equal(originalInput, input)` catches direct writes to `b`, but `encoding/json` never writes to its input, so there is no realistic bug in range. It is not an aliasing check — that would mutate `input` *after* decoding and assert `md` is unaffected.
- `assert.JSONEq(encoded, reencoded)` compares a value against a decode-then-re-encode of itself. That tests idempotence, not whether the reused decode preserved every field from `replacementData`.

One line fixes the second one: compare `replacementData` against `encoded` directly, which would actually catch a field dropped during a reused decode.

Not a correctness concern either way — the decoded object graph retains no input-backed `[]byte` or `json.RawMessage`, so the aliasing gap is theoretical. Happy to see this land as-is and tightened separately.

**Minor, take or leave:** the unchecked `.([]any)` / `.(map[string]any)` fixture assertions panic rather than fail cleanly if `exampleViewJSON` changes. And the schema index is still built twice per successful parse — validation forces `lazySchemasByID()` via `checkVersionSchemasExist`, then `m.init()` resets the `OnceValue` — which remains cosmetic at these sizes.

zeroshade:
Follow-up on the one non-blocking point from my re-review, now with a verified snippet — approval above still stands, this is the only thing I'd like changed before merge.

`TestMetadataUnmarshalDoesNotMutateInputAndRoundTripsAfterReuse` passes with the production fix in `UnmarshalJSON` reverted, so it currently protects nothing. The cause is the final assertion:

```go
assert.JSONEq(t, string(encoded), string(reencoded))
```

That compares `encoded` against a decode-then-re-encode of itself, which tests encoder idempotence rather than whether the reused decode preserved every field from `replacementData`.

I initially wanted to suggest simply comparing `replacementData` against `encoded`, but I verified that and it does **not** work, for two reasons worth recording so this isn't re-attempted:

1. A literal substitution doesn't compile — `reencoded` becomes unused (`declared and not used: reencoded`), so the round-trip block has to go with it.
2. Even after removing it, the compare fails at `f258696` on exactly one key: `NestedField.Doc` is `json:"doc,omitempty"`, so the fixture's `"doc": ""` is omitted on encode. `version-log` was **not** a problem here — the fixture has it present, so `init()`'s nil-to-empty normalization never triggers.
3. And once `doc` is normalized away, the compare passes *both* at the head and with the fix reverted — still vacuous. Fixing the tautology alone isn't enough to give the test teeth.

What makes it non-vacuous is dropping `properties` from the replacement document: the fixed decoder clears the omitted state, the old one retains stale `{"prop":"value"}`. That's the assertion doing real work.

Verified minimal form:

```go
var replacement map[string]any
require.NoError(t, json.Unmarshal([]byte(exampleViewJSON), &replacement))
replacement["current-version-id"] = int64(2)
replacement["versions"].([]any)[0].(map[string]any)["version-id"] = int64(2)
replacement["versions"].([]any)[0].(map[string]any)["schema-id"] = 2
replacement["schemas"].([]any)[0].(map[string]any)["schema-id"] = 2
delete(replacement["schemas"].([]any)[0].(map[string]any)["fields"].([]any)[0].(map[string]any), "doc")
delete(replacement, "properties")
replacementData, err := json.Marshal(replacement)
require.NoError(t, err)
require.NoError(t, json.Unmarshal(replacementData, &md))

encoded, err := json.Marshal(&md)
require.NoError(t, err)
assert.JSONEq(t, string(replacementData), string(encoded))
```

The trailing `roundTripped` / `reencoded` block is then deleted. Keep the existing input-mutation assertion at the top as-is.

Confirmed empirically:

- At `f258696`, targeted test: **passes**.
- With pre-PR `UnmarshalJSON` restored: **fails**, and on the right thing — the only diff is a surviving `"properties": {"prop": "value"}` in the actual output.
- `go test ./view/ -count=1` with the snippet in place: **green**.

Two notes on the `delete` for `doc`: it's papering over an encode/decode asymmetry rather than testing anything, so if you'd rather avoid it, giving that fixture field a non-empty `doc` would remove the need for the line. Either is fine by me.

Not worth blocking a correct production fix over — if you'd prefer to land this as-is and tighten the test in a follow-up, I'm happy with that too. The `.([]any)` / `.(map[string]any)` assertions stay unchecked either way, which remains ordinary test brittleness.

laskoviymishka (approved):
:shipit:

PR #1647, opened 2026-08-04
fix(table): preserve field state on failed JSON decoding

fallintoplace:
## Summary

Make partition and sort field JSON decoding safe to retry on a reused receiver.

## Why

Malformed transforms or invalid sort options could leave a field containing a mixture of new and old state because assignments happened before validation finished.

## What changed

Both decoders now validate a local value completely and assign it only after success. Valid JSON behavior and existing validation rules are unchanged.

## Tests

- Added failed-decode coverage for invalid transforms, source IDs, names, sort directions, and null orders
- go test . ./table

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -142,14 +142,16 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	p.FieldID = aux.FieldID
-	p.Name = aux.Name
+	next := PartitionField{
+		FieldID: aux.FieldID,
+		Name:    aux.Name,
+	}
 
 	var err error
-	if p.Transform, err = ParseTransform(aux.TransformString); err != nil {
+	if next.Transform, err = ParseTransform(aux.TransformString); err != nil {
 		return fmt.Errorf("%w: %w", ErrInvalidPartitionSpec, err)
 	}
-	if err := validateTransform(p.Transform); err != nil {
+	if err := validateTransform(next.Transform); err != nil {
 		return fmt.Errorf("%w: %w", ErrInvalidPartitionSpec, err)
 	}
 
@@ -157,25 +159,27 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 		return fmt.Errorf("%w: partition source-ids cannot be empty", ErrInvalidPartitionSpec)
 	}
 	if !hasSourceID && !hasSourceIDs {
-		if _, isVoid := p.Transform.(VoidTransform); !isVoid {
+		if _, isVoid := next.Transform.(VoidTransform); !isVoid {
 			return fmt.Errorf("%w: partition field requires source-id or source-ids", ErrInvalidPartitionSpec)
 		}
 		// Preserve compatibility with historical source-less void tombstones.
-		p.SourceIDs = []int{0}
+		next.SourceIDs = []int{0}
 	} else if len(aux.SourceIDs) > 0 {
-		p.SourceIDs = aux.SourceIDs
+		next.SourceIDs = aux.SourceIDs
 	} else {
-		p.SourceIDs = []int{aux.SourceID}
+		next.SourceIDs = []int{aux.SourceID}
 	}
-	for _, sourceID := range p.SourceIDs {
-		_, isVoid := p.Transform.(VoidTransform)
+	for _, sourceID := range next.SourceIDs {
+		_, isVoid := next.Transform.(VoidTransform)
 		if sourceID <= 0 && (!isVoid || hasSourceID || hasSourceIDs) {
 			return fmt.Errorf("%w: partition source ID must be positive: %d", ErrInvalidPartitionSpec, sourceID)
 		}
 	}
-	if p.Name == "" {
+	if next.Name == "" {
 		return fmt.Errorf("%w: partition name cannot be empty", ErrInvalidPartitionSpec)
 	}
+
+	*p = next
 
 	return nil
 }
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -778,6 +778,27 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
 	})
 }
 
+func TestPartitionFieldUnmarshalPreservesStateOnError(t *testing.T) {
+	field := iceberg.PartitionField{
+		SourceIDs: []int{7},
+		FieldID:   1007,
+		Name:      "old",
+		Transform: iceberg.IdentityTransform{},
+	}
+
+	err := json.Unmarshal([]byte(`{
+		"source-id": 1,
+		"field-id": 1000,
+		"transform": "not-a-transform",
+		"name": "new"
+	}`), &field)
+	require.Error(t, err)
+	assert.Equal(t, []int{7}, field.SourceIDs)
+	assert.Equal(t, 1007, field.FieldID)
+	assert.Equal(t, "old", field.Name)
+	assert.Equal(t, iceberg.IdentityTransform{}, field.Transform)
+}
+
 func TestPartitionSpecUnmarshalRejectsInvalidStructure(t *testing.T) {
 	tests := []struct {
 		name    string
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -803,7 +803,7 @@ func TestPartitionFieldUnmarshalPreservesStateOnError(t *testing.T) {
 	}{
 		{
 			name: "invalid transform",
-			data: `{"source-id":1,"field-id":1000,"transform":"not-a-transform","name":"new"}`,
+			data: `{"source-id":1,"field-id":1000,"transform":"bucket[0]","name":"new"}`,
 		},
 		{
 			name: "non-positive source ID",
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -779,24 +779,40 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
 }
 
 func TestPartitionFieldUnmarshalPreservesStateOnError(t *testing.T) {
-	field := iceberg.PartitionField{
+	initial := iceberg.PartitionField{
 		SourceIDs: []int{7},
 		FieldID:   1007,
 		Name:      "old",
 		Transform: iceberg.IdentityTransform{},
 	}
 
-	err := json.Unmarshal([]byte(`{
-		"source-id": 1,
-		"field-id": 1000,
-		"transform": "not-a-transform",
-		"name": "new"
-	}`), &field)
-	require.Error(t, err)
-	assert.Equal(t, []int{7}, field.SourceIDs)
-	assert.Equal(t, 1007, field.FieldID)
-	assert.Equal(t, "old", field.Name)
-	assert.Equal(t, iceberg.IdentityTransform{}, field.Transform)
+	for _, test := range []struct {
+		name string
+		data string
+	}{
+		{
+			name: "invalid transform",
+			data: `{"source-id":1,"field-id":1000,"transform":"not-a-transform","name":"new"}`,
+		},
+		{
+			name: "non-positive source ID",
+			data: `{"source-id":0,"field-id":1000,"transform":"identity","name":"new"}`,
+		},
+		{
+			name: "missing source ID",
+			data: `{"field-id":1000,"transform":"identity","name":"new"}`,
+		},
+		{
+			name: "empty name",
+			data: `{"source-id":1,"field-id":1000,"transform":"identity","name":""}`,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			field := initial
+			require.Error(t, json.Unmarshal([]byte(test.data), &field))
+			assert.Equal(t, initial, field)
+		})
+	}
 }
 
 func TestPartitionSpecUnmarshalRejectsInvalidStructure(t *testing.T) {
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -852,6 +852,7 @@ func TestPartitionFieldUnmarshalReplacesStateOnSuccess(t *testing.T) {
 	assert.Equal(t, []int{2}, field.SourceIDs)
 	assert.Equal(t, 1001, field.FieldID)
 	assert.Equal(t, "new field", field.Name)
+	assert.Equal(t, iceberg.IdentityTransform{}, field.Transform)
 	assert.Equal(t, "new+field", field.EscapedName())
 }
 
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -757,6 +757,23 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
 		assert.ErrorContains(t, err, "partition field cannot contain both source-id and source-ids")
 	})
 
+	t.Run("unmarshal rejects empty source-ids", func(t *testing.T) {
+		var field iceberg.PartitionField
+		err := json.Unmarshal([]byte(`{
+			"source-ids": [],
+			"field-id": 1002,
+			"transform": "identity",
+			"name": "identity"
+		}`), &field)
+		require.ErrorIs(t, err, iceberg.ErrInvalidPartitionSpec)
+		assert.ErrorContains(t, err, "source-ids cannot be empty")
+	})
+
+	t.Run("unmarshal rejects malformed JSON", func(t *testing.T) {
+		var field iceberg.PartitionField
+		require.Error(t, json.Unmarshal([]byte(`{"source-id":`), &field))
+	})
+
 	t.Run("unmarshal source-less void tombstone", func(t *testing.T) {
 		jsonData := `
 		{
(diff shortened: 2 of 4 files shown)


tanmayrauth (approved):

partitions.go
```diff
@@ -137,41 +137,45 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
...
 		if sourceID <= 0 && (!isVoid || hasSourceID || hasSourceIDs) {
 			return fmt.Errorf("%w: partition source ID must be positive: %d", ErrInvalidPartitionSpec, sourceID)
 		}
 	}
-	if p.Name == "" {
+	if next.Name == "" {
 		return fmt.Errorf("%w: partition name cannot be empty", ErrInvalidPartitionSpec)
 	}
 
+	*p = next
```
zeroshade, line 182: This line silently fixes a second bug that neither the code nor the PR description claims, and it is worth claiming.

`escapedName` (`partitions.go:55-57`) is a cached `url.QueryEscape(Name)`. The old code assigned `p.Name` but never reset `escapedName`, so re-decoding into an already-populated `PartitionField` left `EscapedName()` returning the **previous** field's escaped name. `*p = next` zeroes it. That is a real correctness fix, not just a hygiene improvement.

Consider mentioning it in the description and pinning it with a test. As written, nothing in the code or the test suite records the dependency, so a future refactor back to in-place assignment would silently reintroduce it.

partitions_test.go
```diff
@@ -660,6 +660,43 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
...
+			name: "missing source ID",
+			data: `{"field-id":1000,"transform":"identity","name":"new"}`,
+		},
+		{
+			name: "empty name",
+			data: `{"source-id":1,"field-id":1000,"transform":"identity","name":""}`,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			field := initial
```
zeroshade, line 828: `field := initial` copies the struct but shares `SourceIDs`' backing array with `initial`, so `assert.Equal(t, initial, field)` on line 695 cannot detect in-place mutation of that slice. An implementation that decoded into `p.SourceIDs` would overwrite element 0 of the array **both** structs point at, and the assertion would still pass — the test would go green on the class of bug it exists to catch.

Consider constructing `initial` inside the subtest, or comparing against an independently constructed expected value that owns its own slice.

table/sorting_test.go
```diff
@@ -319,6 +319,43 @@ func TestUnmarshalInvalidSortTransform(t *testing.T) {
...
+			name: "invalid direction",
+			data: `{"source-id":1,"transform":"identity","direction":"not-a-direction","null-order":"nulls-first"}`,
+		},
+		{
+			name: "invalid null order",
+			data: `{"source-id":1,"transform":"identity","direction":"asc","null-order":"not-a-null-order"}`,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			field := initial
```
zeroshade, line 389: Same as on `partitions_test.go:693`: `field := initial` shares `SourceIDs`' backing array with `initial`, so `assert.Equal(t, initial, field)` on line 354 cannot detect in-place mutation of that slice — a buggy implementation writing through `s.SourceIDs` would mutate the array both structs point at and the assertion would still pass.

Consider building `initial` inside the subtest, or comparing against an independently constructed value with its own slice.

table/sorting.go
```diff
@@ -158,36 +158,40 @@ func (s *SortField) UnmarshalJSON(b []byte) error {
...
 	if hasSourceIDs {
-		s.SourceIDs = aux.SourceIDs
+		next.SourceIDs = aux.SourceIDs
 	} else {
-		s.SourceIDs = []int{aux.SourceID}
+		next.SourceIDs = []int{aux.SourceID}
 	}
 
-	if err := validateSortSourceIDs(s.SourceIDs); err != nil {
+	if err := validateSortSourceIDs(next.SourceIDs); err != nil {
```
zeroshade, line 176: Non-blocking, pre-existing, and noted only because the block was rewritten here: `validateSortSourceIDs` still runs before `iceberg.ParseTransform` on line 177, so a document with both a bad source ID and a bad transform reports only the former. Fine as-is — flagging in case the ordering was meant to change along with the restructure.

zeroshade (commented):
The change is correct. Verified against `46c4f92` by auditing every field on both receivers for anything still merged from a previous value:

- `PartitionField` has four exported fields plus one unexported cache. `next` sets `FieldID`/`Name` (`partitions.go:140-143`), `Transform` (`:146`), and `SourceIDs` on all three branches (`:161`, `:163`, `:165`). Nothing merged.
- `SortField` has four fields. `next` sets `Direction`/`NullOrder` (`table/sorting.go:161-164`), `SourceIDs` on both branches (`:167`, `:169`), and `Transform` (`:177`). Nothing merged.
- Every error return precedes `*p = next` / `*s = next`, including the early returns that fire before `aux` is populated, so the failure path is genuinely atomic.
- No aliasing: `next.SourceIDs = aux.SourceIDs` shares its backing array with `aux`, a decoder-allocated local — not with the previous receiver and not with `b`.

Two items are worth acting on and are noted inline: an unclaimed correctness fix that deserves a mention in the description plus a pinning test, and a flaw in the new failure-path tests that makes them weaker than they look.

**Non-blocking**

- `partitions.go:461-464` — out of scope, but the same file and the same bug class: `aux := struct{...}{ID: ps.id}` seeds the spec ID from the receiver, so `PartitionSpec` still merges `spec-id` from a previous value when the document omits it. Worth folding in here while the context is loaded. (Flagged independently on #1662.)
- The highest-value missing test is a success-path dirty-receiver case, since `*p = next` is the whole change: populate a `PartitionField` through `PartitionSpec.initialize()` (`partitions.go:528-533`, so `escapedName` is set), decode a valid *different* field into it, then assert both that `SourceIDs` is exactly the new value and that `EscapedName()` reflects the new name. That is the one behavior this PR fixes with zero coverage today.
- Also missing, all trivially safe because they return before mutation: "both source-id and source-ids", `"source-ids": []`, malformed JSON, and for `SortField` "neither source-id nor source-ids".

partitions_test.go
```diff
@@ -639,6 +639,23 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
...
+			"transform": "identity",
+			"name": "identity"
+		}`), &field)
+		require.ErrorIs(t, err, iceberg.ErrInvalidPartitionSpec)
+		assert.ErrorContains(t, err, "source-ids cannot be empty")
+	})
+
+	t.Run("unmarshal rejects malformed JSON", func(t *testing.T) {
+		var field iceberg.PartitionField
+		require.Error(t, json.Unmarshal([]byte(`{"source-id":`), &field))
```
zeroshade: This case never executes any line of `PartitionField.UnmarshalJSON`. `encoding/json` scans the entire buffer for well-formedness *before* dispatching to a custom unmarshaler, so `{"source-id":` fails in the stdlib scanner:

```
err=unexpected end of JSON input  type=*json.SyntaxError
```

Coverage over just this subtest confirms it — every block in the method is at hit count 0:

```
partitions.go:114.56,116.48 2 0
partitions.go:116.48,118.3  1 0    <- the raw-unmarshal error return
partitions.go:120.2,120.35  1 0
partitions.go:125.2,136.48  4 0
```

If the intent was to cover the `json.Unmarshal(b, &raw)` guard at `partitions.go:116-118`, that needs input which is **well-formed JSON but structurally wrong** for the target — e.g. `[1,2]` or `"a string"`, which parse fine but fail the `map[string]json.RawMessage` decode inside the method. That guard is currently unreached by any test. Otherwise this subtest is asserting that the standard library rejects truncated JSON, and is better dropped.

partitions_test.go
```diff
@@ -660,6 +677,63 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
...
+			data: `{"field-id":1000,"transform":"identity","name":"new"}`,
+		},
+		{
+			name: "empty name",
+			data: `{"source-id":1,"field-id":1000,"transform":"identity","name":""}`,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			initial := iceberg.PartitionField{
+				SourceIDs: []int{7},
```
zeroshade, line 823: Moving `initial` into the subtest fixed real cross-subtest contamination — previously all four subtests shared one `[]int{7}` backing array, so a mutation in the first would leak into the rest.

Within a single subtest, though, `field := initial` on line 709 is a shallow copy: `field.SourceIDs` and `initial.SourceIDs` still point at the same array. So if the code under test ever mutated the slice *elements* in place, both sides would change together and `assert.Equal(t, initial, field)` would still pass:

```
PROBE aliasing: initial.SourceIDs=[999] field.SourceIDs=[999]  assert.Equal_would_pass=true
```

Not reachable in the current implementation — `next.SourceIDs` is only ever assigned a fresh slice (`partitions.go:159/161/163`), never indexed into. So this is a regression-detection gap rather than a live bug. `SourceIDs: slices.Clone([]int{7})`, or comparing `field` against a separately-constructed literal, would close it.

partitions_test.go
```diff
@@ -660,6 +677,63 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
...
+	require.NoError(t, json.Unmarshal([]byte(`{
+		"source-id": 2,
+		"field-id": 1001,
+		"transform": "identity",
+		"name": "new field"
+	}`), &field))
+
+	assert.Equal(t, []int{2}, field.SourceIDs)
+	assert.Equal(t, "new field", field.Name)
+	assert.Equal(t, "new+field", field.EscapedName())
```
zeroshade, line 856: This is the assertion that makes the whole test load-bearing — on a build with the fix reverted, it's the *only* one that fails (`expected: "new+field", actual: "old+field"`). Lines 732 and 733 pass either way, because the old incremental code overwrote `SourceIDs` and `Name` too; the stale unexported `escapedName` cache was the only state that actually survived.

That's a lot of weight on one line, and on the incidental fact that `escapedName` rides through `clonePartitionField` (`partitions.go:571-573`) as part of a by-value copy. Worth making the test harder to accidentally defeat:

- `assert.Equal(t, 1001, field.FieldID)` for completeness.
- A multi-arg → single-arg shrink case: seed `SourceIDs: []int{1, 2}`, decode a document with a single `source-id`, and assert the result is exactly `[]int{2}`. That directly exercises full replacement over merge, independent of the cache.

table/sorting_test.go
```diff
@@ -337,6 +373,11 @@ func TestSortFieldMultiArgSourceIDs(t *testing.T) {
 		assert.Contains(t, err.Error(), "cannot contain both source-id and source-ids")
 	})
 
+	t.Run("unmarshal rejects malformed JSON", func(t *testing.T) {
+		var field table.SortField
+		require.Error(t, json.Unmarshal([]byte(`{"source-id":`), &field))
```
zeroshade: Same issue as the `partitions_test.go` malformed-JSON case: this never reaches `SortField.UnmarshalJSON`. All 24 statement blocks in `table/sorting.go:134-196` report hit count 0 when only this subtest runs — `encoding/json` rejects the truncated buffer before the method is called.

Use well-formed-but-structurally-wrong input (`[1,2]`) if the goal is to cover the `raw` decode guard at `table/sorting.go:136-138`, which no test currently reaches.

Minor, while here: this subtest lives inside `TestSortFieldMultiArgSourceIDs`, which is about multi-arg source IDs. The partition-side equivalent went into `TestPartitionFieldUnmarshalJSON`. Worth aligning.

table/sorting_test.go
```diff
@@ -319,6 +319,42 @@ func TestUnmarshalInvalidSortTransform(t *testing.T) {
...
+			data: `{"source-id":1,"transform":"identity","direction":"not-a-direction","null-order":"nulls-first"}`,
+		},
+		{
+			name: "invalid null order",
+			data: `{"source-id":1,"transform":"identity","direction":"asc","null-order":"not-a-null-order"}`,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			initial := table.SortField{
+				SourceIDs: []int{7},
```
zeroshade, line 384: Same latent aliasing as the partition-side preservation test: `field := initial` on line 351 shares the `SourceIDs` backing array with `initial`, so an in-place element mutation would be invisible to `assert.Equal` on line 353.

Not triggerable today — `next.SourceIDs` only ever receives a fresh slice (`table/sorting.go:167/169`). `slices.Clone` here keeps it that way if the implementation later changes.

zeroshade (changes requested):
Re-reviewed at `d306952`. The production code is unchanged from `46c4f92` and my prior audit of it stands: the `next` + `*p = next` / `*s = next` shape is correct, atomic on every error path, and complete across both receivers. All gates are green (`go build ./...`, `go vet ./ ./table/`, `go test ./ ./table/`, `gofmt`, plus `-race` on the touched tests).

The new tests fix the gap I flagged last round, and `TestPartitionFieldUnmarshalReplacesStateOnSuccess` is genuinely load-bearing — verified by reverting only the `partitions.go` hunks and re-running:

```
=== RUN   TestPartitionFieldUnmarshalReplacesStateOnSuccess
    partitions_test.go:734:
        	Error:      	Not equal:
        	            	expected: "new+field"
        	            	actual  : "old+field"
--- FAIL: TestPartitionFieldUnmarshalReplacesStateOnSuccess (0.00s)
```

Requesting changes on the **tests only** — nothing here suggests the fix is wrong. Two of the four added cases don't test this PR's code at all, and I'd rather not land tests that look like coverage but aren't:

1. **The two malformed-JSON cases are vacuous.** `encoding/json` validates the whole input buffer before dispatching to a custom `UnmarshalJSON`, so `{"source-id":` never reaches either method. Confirmed by coverage: every statement block in `PartitionField.UnmarshalJSON` and all 24 in `SortField.UnmarshalJSON` report hit count **0** when only those subtests run. They would pass with both functions deleted.
2. **The success-path test rests on a single assertion.** On the reverted build, the `SourceIDs` and `Name` asserts both still pass — only `EscapedName()` discriminates.

Also worth tightening, though these are optional:

- **Latent aliasing in the preservation tests.** Moving `initial` inside the subtest closed cross-subtest contamination (all four previously shared one `[]int{7}` backing array) — good catch — but `field := initial` is still a shallow copy, so an in-place element mutation of `SourceIDs` would change both and slip past `assert.Equal`. Not reachable today, since both methods only ever assign fresh slices; worth closing as a regression guard.
- `partitions_test.go` "unmarshal source-less void tombstone" (pre-existing, so no inline): `assert.Equal(t, 0, field.SourceID())` can't distinguish the intended `[]int{0}` compatibility value from a nil/empty slice. `assert.Equal(t, []int{0}, field.SourceIDs)` pins the actual contract.

No success-path test for `SortField` is the right call, for the record — it has no unexported cache, so such a test could only assert fields the old incremental code already overwrote.

For context on scope: `partitions.go:461-464` (`PartitionSpec` seeding `aux.ID` from `ps.id`) is the same bug class in this file but out of scope here — tracked on #1662.

Methodology: audited independently by two different models; both returned zero blocking correctness findings on the production diff, and the test-efficacy gaps above were established empirically via revert + coverage rather than by inspection.

partitions_test.go
```diff
@@ -778,6 +677,63 @@ func TestPartitionFieldUnmarshalJSON(t *testing.T) {
...
```
laskoviymishka, line 856: building on zeroshade's point that this test really rests on `EscapedName()` — I'd also pin `FieldID` and `Transform` here. On the reverted build `SourceIDs`/`Name` both still pass, so a regression that cleared the cache but copied the wrong `FieldID` or left a stale `Transform` would slip through, and this is the one test covering `*p = next`.

```go
assert.Equal(t, 1001, field.FieldID)
assert.Equal(t, iceberg.IdentityTransform{}, field.Transform)
```

wdyt?

laskoviymishka (changes requested):
Late to this one — zeroshade's already done a thorough audit and has an open changes-request on the tests, and I agree with holding on exactly those grounds. Nothing to add on the production side: I ran the diff down the same way and the `next` + `*p = next` / `*s = next` shape is correct and atomic on every error path.

Agreed with zeroshade that the two `unmarshal rejects malformed JSON` subtests are vacuous — `encoding/json` rejects the truncated buffer before either custom unmarshaler runs, so they'd pass with both methods deleted. Same read on the `field := initial` shallow copy in the preservation tests: it shares the `SourceIDs` backing array, so an in-place element mutation would update both and slip past `assert.Equal`. Both worth closing before merge.

One thing to add on top: `TestPartitionFieldUnmarshalReplacesStateOnSuccess` only really discriminates on `EscapedName()` — I'd also pin `FieldID` and `Transform` so a regression that clears the cache but copies the wrong field can't sail through. Left it inline.

I'll defer to zeroshade on skipping the `SortField` success-path test — there's no unexported cache there, so I think that's the right call. And `partitions.go:461-464` (the `aux.ID: ps.id` seeding) is the same merge-from-previous bug in this file but out of scope here; good that it's tracked on #1662.

I don't think this needs a separate hold from me on top of zeroshade's — happy to approve once those test gaps close.

zeroshade (approved):
All of my remaining items are addressed: the preservation tests clone the `SourceIDs` backing array so in-place mutation is detectable, the success-path test pins every field including `EscapedName()` (with `FieldID`/`Transform` per laskoviymishka's note), and the failure cases now genuinely execute the unmarshaler. The production shape was already verified sound two rounds ago.

One last non-blocking ask: please add a line to the PR description claiming the `escapedName` cache fix — this PR repairs it, the test proves it, and future archaeology should be able to find it from the description.

Deferring thread closure to `laskoviymishka` since his change request predates the fix push.

---

laskoviymishka (approved):

PR #1648, opened 2026-08-04
fix(table): replace metadata state during JSON decoding

fallintoplace:
## Summary

Make table metadata JSON decoding replace receiver state for format versions 1, 2, and 3.

## Why

Decoding through aliases of the existing receiver allowed omitted properties, snapshots, refs, logs, and current-snapshot state to survive from an earlier document. Validation errors could also leave partial changes behind.

## What changed

Each version decodes into fresh state, applies version-specific defaults and sentinels before validation, and replaces the receiver only after success. The deserialization sentinels are applied on every path, so direct v2 and v3 decoding rejects omitted normalized state such as `default-spec-id` and `next-row-id`, matching `ParseMetadataBytes`. Reused receivers also no longer leak stale snapshot refs or v1 schema state into the new document.

## Tests

- Added receiver-reuse coverage for v1, v2, and v3
- Added failed-decode state preservation coverage for validation and malformed JSON errors
- Added coverage for omitted required state
- go test ./table

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2762,10 +2762,15 @@ type SequenceNumberValidator interface {
 
 // checkLastSequenceNumber validates that all snapshots have sequence numbers <= the last sequence number
 func checkLastSequenceNumber(validator SequenceNumberValidator, snapshotList []Snapshot) error {
+	lastSequenceNumber := validator.LastSequenceNumber()
+	if lastSequenceNumber == -1 {
+		return fmt.Errorf("%w: last-sequence-number is required for format versions greater than 1", ErrInvalidMetadata)
+	}
+
 	for _, snap := range snapshotList {
-		if snap.SequenceNumber > validator.LastSequenceNumber() {
+		if snap.SequenceNumber > lastSequenceNumber {
 			return fmt.Errorf("%w: snapshot %d has sequence number %d which is greater than last-sequence-number %d",
-				ErrInvalidMetadata, snap.SnapshotID, snap.SequenceNumber, validator.LastSequenceNumber())
+				ErrInvalidMetadata, snap.SnapshotID, snap.SequenceNumber, lastSequenceNumber)
 		}
 	}
 
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1692,7 +1692,8 @@ type commonMetadata struct {
 
 func initCommonMetadataForDeserialization() commonMetadata {
 	return commonMetadata{
-		// Negative sentinels distinguish omitted required fields from explicit zero values.
+		// These fields use negative sentinels so validation can distinguish omitted
+		// values from explicit zero values.
 		LastColumnId:       -1,
 		CurrentSchemaID:    -1,
 		DefaultSpecID:      -1,
@@ -2666,6 +2667,10 @@ func (m *metadataV3) UnmarshalJSON(b []byte) error {
 	aux := (*Alias)(next)
 
 	if err := json.Unmarshal(b, aux); err != nil {
+		return err
+	}
+
+	if err := rejectFieldsBeyondVersion(aux.FormatVersion); err != nil {
 		return err
 	}
 
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1512,11 +1512,11 @@ func ParseMetadataBytes(b []byte) (Metadata, error) {
 	var ret Metadata
 	switch ver.FormatVersion {
 	case 1:
-		ret = initMetadataV1Deser()
+		ret = &metadataV1{}
 	case 2:
-		ret = initMetadataV2Deser()
+		ret = &metadataV2{}
 	case 3:
-		ret = initMetadataV3Deser()
+		ret = &metadataV3{}
 	default:
 		return nil, ErrInvalidMetadataFormatVersion
 	}
@@ -1692,6 +1692,7 @@ type commonMetadata struct {
 
 func initCommonMetadataForDeserialization() commonMetadata {
 	return commonMetadata{
+		// Negative sentinels distinguish omitted required fields from explicit zero values.
 		LastColumnId:       -1,
 		CurrentSchemaID:    -1,
 		DefaultSpecID:      -1,
@@ -2505,9 +2506,6 @@ func (m *metadataV1) UnmarshalJSON(b []byte) error {
 	next := initMetadataV1Deser()
 	aux := (*Alias)(next)
 
-	// Set LastColumnId to -1 to indicate that it is not set as LastColumnId = 0 is a valid value for when no schema is present
-	aux.LastColumnId = -1
-
 	if err := json.Unmarshal(b, aux); err != nil {
 		return err
 	}
@@ -2586,9 +2584,6 @@ func (m *metadataV2) UnmarshalJSON(b []byte) error {
 	type Alias metadataV2
 	next := initMetadataV2Deser()
 	aux := (*Alias)(next)
-
-	// Set LastColumnId to -1 to indicate that it is not set as LastColumnId = 0 is a valid value for when no schema is present
-	aux.LastColumnId = -1
 
 	if err := json.Unmarshal(b, aux); err != nil {
 		return err
@@ -2669,9 +2664,6 @@ func (m *metadataV3) UnmarshalJSON(b []byte) error {
 	type Alias metadataV3
 	next := initMetadataV3Deser()
 	aux := (*Alias)(next)
-
-	// Set LastColumnId to -1 to indicate that it is not set as LastColumnId = 0 is a valid value for when no schema is present
-	aux.LastColumnId = -1
 
 	if err := json.Unmarshal(b, aux); err != nil {
 		return err
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2502,7 +2502,8 @@ func (m *metadataV1) preValidate() {
 
 func (m *metadataV1) UnmarshalJSON(b []byte) error {
 	type Alias metadataV1
-	aux := (*Alias)(m)
+	next := initMetadataV1Deser()
+	aux := (*Alias)(next)
 
 	// Set LastColumnId to -1 to indicate that it is not set as LastColumnId = 0 is a valid value for when no schema is present
 	aux.LastColumnId = -1
@@ -2530,9 +2531,15 @@ func (m *metadataV1) UnmarshalJSON(b []byte) error {
 		}
 	}
 
-	m.preValidate()
+	next.preValidate()
 
-	return m.validate()
+	if err := next.validate(); err != nil {
+		return err
+	}
+
+	*m = *next
+
+	return nil
 }
 
 func (m *metadataV1) ToV2() metadataV2 {
@@ -2577,7 +2584,8 @@ func (m *metadataV2) Equals(other Metadata) bool {
 
 func (m *metadataV2) UnmarshalJSON(b []byte) error {
 	type Alias metadataV2
-	aux := (*Alias)(m)
+	next := initMetadataV2Deser()
+	aux := (*Alias)(next)
 
 	// Set LastColumnId to -1 to indicate that it is not set as LastColumnId = 0 is a valid value for when no schema is present
 	aux.LastColumnId = -1
@@ -2594,9 +2602,15 @@ func (m *metadataV2) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	m.preValidate()
+	next.preValidate()
 
-	return m.validate()
+	if err := next.validate(); err != nil {
+		return err
+	}
+
+	*m = *next
+
+	return nil
 }
 
 func (m *metadataV2) validate() error {
@@ -2653,7 +2667,8 @@ func (m *metadataV3) Equals(other Metadata) bool {
 
 func (m *metadataV3) UnmarshalJSON(b []byte) error {
 	type Alias metadataV3
-	aux := (*Alias)(m)
+	next := initMetadataV3Deser()
+	aux := (*Alias)(next)
 
 	// Set LastColumnId to -1 to indicate that it is not set as LastColumnId = 0 is a valid value for when no schema is present
 	aux.LastColumnId = -1
@@ -2662,9 +2677,15 @@ func (m *metadataV3) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	m.preValidate()
+	next.preValidate()
 
-	return m.validate()
+	if err := next.validate(); err != nil {
+		return err
+	}
+
+	*m = *next
+
+	return nil
 }
 
 // MarshalJSON serializes v3 metadata with the spec-mandated emission of
--- a/table/metadata_internal_test.go
+++ b/table/metadata_internal_test.go
@@ -421,6 +421,9 @@ func TestMetadataUnmarshalReplacesReceiverState(t *testing.T) {
 			assert.Empty(t, common.StatisticsList)
 			assert.Empty(t, common.PartitionStatsList)
 			assert.Empty(t, common.EncryptionKeyList)
+			// With no snapshots, an omitted last-sequence-number retains the
+			// legacy -1 sentinel. This documents current behavior, not the
+			// desired metadata normalization.
 			switch metadata := tt.target.(type) {
 			case *metadataV2:
 				assert.Equal(t, int64(-1), metadata.LastSeqNum)
@@ -454,7 +457,7 @@ func TestMetadataUnmarshalPreservesStateOnError(t *testing.T) {
 		{
 			name:    "v1",
 			data:    ExampleTableMetadataV1,
-			invalid: strings.Replace(ExampleTableMetadataV1, `"id": 2, "name": "y"`, `"id": 1, "name": "y"`, 1),
+			invalid: strings.Replace(ExampleTableMetadataV1, `"current-snapshot-id": -1`, `"current-snapshot-id": 999`, 1),
 			target:  &metadataV1{},
 		},
 		{
--- a/table/metadata_internal_test.go
+++ b/table/metadata_internal_test.go
@@ -382,6 +382,22 @@ func TestMetadataV2UnmarshalPreservesStateOnError(t *testing.T) {
 	assert.Len(t, metadata.SnapshotList, 2)
 }
 
+func TestMetadataV2UnmarshalRejectsMissingRequiredStateOnReuse(t *testing.T) {
+	var metadata metadataV2
+	require.NoError(t, json.Unmarshal([]byte(ExampleTableMetadataV2), &metadata))
+
+	var reduced map[string]json.RawMessage
+	require.NoError(t, json.Unmarshal([]byte(ExampleTableMetadataV2), &reduced))
+	delete(reduced, "default-spec-id")
+	reducedData, err := json.Marshal(reduced)
+	require.NoError(t, err)
+
+	require.Error(t, json.Unmarshal(reducedData, &metadata))
+	assert.Equal(t, 0, metadata.DefaultSpecID)
+	assert.Equal(t, 1, metadata.CurrentSchemaID)
+	assert.Len(t, metadata.SnapshotList, 2)
+}
+
 func TestLastUpdatedMSPresence(t *testing.T) {
 	tests := []struct {
 		name string
--- a/table/metadata_internal_test.go
+++ b/table/metadata_internal_test.go
@@ -337,7 +337,29 @@ func TestMetadataUnmarshalReplacesReceiverState(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			require.NoError(t, json.Unmarshal([]byte(tt.data), tt.target))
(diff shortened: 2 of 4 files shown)


table/metadata_internal_test.go
```diff
@@ -1402,6 +1478,7 @@ func TestMetadataV2Validation(t *testing.T) {
 		"current-schema-id": 0,
 		"last-partition-id": 1000,
 		"schemas": [{"type":"struct","schema-id":0,"fields":[]}],
+		"default-spec-id": 0,
```
zeroshade, line 1667: This added line — and the matching one at `:1501` — is the proof that the description is inaccurate, and it is the main thing worth changing here.

"Existing version-specific defaults and sentinels are preserved" reads as a no-op, but they are now applied unconditionally, so direct `UnmarshalJSON` is strictly more rejecting than before. `initMetadataV2Deser()` yields `DefaultSpecID = -1`, and `commonMetadata.preValidate` has no `-1 → max(specs)` fallback — only `metadataV1.preValidate` does, at `table/metadata.go:2444-2446`. That is exactly why these two fixtures had to gain `"default-spec-id": 0`. The same applies to v3 `next-row-id`: a zero-value receiver gave `0` and passed, whereas `-1` now reaches `checkNextRowID`'s "next-row-id is required for v3 tables" (`table/metadata.go:2705-2709`).

This is the right behavior — it matches what `ParseMetadataBytes` already enforced, and `metadataV{1,2,3}` are unexported so no external caller can observe the change. Consider rewording the commit message to say so, e.g. "the deserialization sentinels are now applied on every decode path, so `default-spec-id` / `next-row-id` are required for v2/v3 documents decoded directly, matching `ParseMetadataBytes`."

table/metadata.go
```diff
@@ -2465,7 +2465,8 @@ func (m *metadataV1) preValidate() {
 
 func (m *metadataV1) UnmarshalJSON(b []byte) error {
 	type Alias metadataV1
-	aux := (*Alias)(m)
+	next := initMetadataV1Deser()
```
zeroshade, line 2507: Worth adding to the description: this fixes two things beyond the stated one.

First, `commonMetadata.preValidate()` injects a `main` entry into `SnapshotRefs` when `CurrentSnapshotID != nil` (`table/metadata.go:2161-2168`). On a reused receiver whose new document omitted `refs`, that injection mutated the **previous** document's map. `TestMetadataUnmarshalReplacesReceiverState`'s `assert.Nil(t, common.CurrentSnapshotID)` plus `assert.Empty(t, common.SnapshotRefs)` is exactly the right assertion for it.

Second, v1's `if aux.CurrentSchemaID == -1 && aux.Schema != nil` branch (`table/metadata.go:2488-2495`) previously never fired on a zero-value receiver, because `CurrentSchemaID` was `0` rather than the `-1` sentinel. Direct `UnmarshalJSON` therefore rejected v1 documents that `ParseMetadataBytes` accepted. The two paths are now consistent.

table/metadata.go
```diff
@@ -2465,7 +2465,8 @@ func (m *metadataV1) preValidate() {
...
+	aux := (*Alias)(next)
 
 	// Set LastColumnId to -1 to indicate that it is not set as LastColumnId = 0 is a valid value for when no schema is present
 	aux.LastColumnId = -1
```
zeroshade: Non-blocking: this assignment and its comment are now dead code. `next` comes from `initMetadataV1Deser()`, and `initCommonMetadataForDeserialization()` (`table/metadata.go:1660`) already sets `LastColumnId` to `-1`. Same at `:2554` and `:2637`.

Leaving both makes it ambiguous which one is authoritative. Consider dropping the redundant assignments and keeping the explanatory comment at the `init...Deser` helper, where the value actually originates.

table/metadata_internal_test.go
```diff
@@ -324,6 +324,80 @@ func TestMetadataV3Parsing(t *testing.T) {
 	assert.Equal(t, int64(2000), *secondSnapshot.FirstRowID)
 }
 
+func TestMetadataUnmarshalReplacesReceiverState(t *testing.T) {
+	tests := []struct {
+		name   string
+		data   string
+		target any
+	}{
+		{name: "v1", data: ExampleTableMetadataV1, target: &metadataV1{}},
```
zeroshade, line 333: Non-blocking: the v1 subtest is close to vacuous. `ExampleTableMetadataV1` (`table/metadata_internal_test.go:181-199`) has `"properties": {}`, no `refs`, no `snapshot-log`, no `metadata-log`, and `"current-snapshot-id": -1` (normalized to nil). Five of the six `assert.Empty`/`assert.Nil` calls below therefore already hold *before* the second decode, and would pass with the fix reverted. Only `SnapshotList` is genuinely exercised for v1.

Either extend the v1 fixture, or — cheaper and more durable — add guard assertions right after the first decode (`require.NotEmpty(t, common.Props)`, and so on) so no version's subtest can silently go vacuous as fixtures change.

table/metadata_internal_test.go
```diff
@@ -324,6 +324,80 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require.NoError(t, json.Unmarshal([]byte(tt.data), tt.target))
+
+			var reduced map[string]json.RawMessage
+			require.NoError(t, json.Unmarshal([]byte(tt.data), &reduced))
+			for _, key := range []string{
+				"properties", "current-snapshot-id", "snapshots", "snapshot-log", "metadata-log", "refs",
```
zeroshade, line 406: Non-blocking: this delete-list omits several `omitempty` `commonMetadata` fields in the same bug class — `statistics`, `partition-statistics`, `encryption-keys`, `last-partition-id`, and `last-sequence-number` would all have survived a reuse identically. `last-sequence-number` is free to add; the rest would need fixture extension first.

table/metadata_internal_test.go
```diff
@@ -324,6 +324,80 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+			assert.Empty(t, common.SnapshotList)
+			assert.Empty(t, common.SnapshotLog)
+			assert.Empty(t, common.MetadataLog)
+			assert.Empty(t, common.SnapshotRefs)
+			assert.Nil(t, common.CurrentSnapshotID)
+		})
+	}
+}
+
+func TestMetadataV2UnmarshalPreservesStateOnError(t *testing.T) {
```
zeroshade: Non-blocking: failure-path coverage exists only for v2, yet all three `UnmarshalJSON`s changed.

v1 needs it most — `metadataV1.preValidate` (`table/metadata.go:2432-2464`) mutates `SchemaList`, `Specs`, `DefaultSpecID`, `LastPartitionID`, and `SortOrderList` before validation runs, by far the largest partial-mutation surface of the three. v3 likewise fails inside `checkNextRowID` after `preValidate` has already run.

zeroshade (commented):
The change is correct. Verified against `7abcb95`:

- `next` comes from `initMetadataV{1,2,3}Deser()`, so every slice and map in `commonMetadata` starts nil and is allocated by the decoder rather than merged into what the receiver already held.
- `next` is unreachable after `*m = *next`, so the assigned value aliases nothing.
- Every error return — `json.Unmarshal`, `rejectFieldsBeyondVersion`, `validate` — precedes the assignment on all three versions, so failure is fully atomic.
- `*m = *next` is vet-clean: `commonMetadata` contains no locks, and `metadataV1.ToV2()` (`table/metadata.go:2508-2516`) already performs the equivalent value copy.

The one substantive ask is that the commit message describe what the change actually does; details inline at `table/metadata_internal_test.go:1481`. In short, "existing version-specific defaults and sentinels are preserved" reads as a no-op, but they are now applied **unconditionally**, which makes direct `UnmarshalJSON` strictly more rejecting. That is the correct behavior — it matches what `ParseMetadataBytes` already enforced, and `metadataV{1,2,3}` are unexported so nothing external can observe the tightening — it just should not be described as a preservation.

The PR also fixes two things without saying so; noted inline at `table/metadata.go:2468`.

**Non-blocking**

- `table/metadata.go:1514-1518` — `ParseMetadataBytes` still builds `ret` via `initMetadataVNDeser()`, and `json.Unmarshal` now immediately discards that state in favor of its own fresh copy. Correct but wasteful and slightly misleading; `ret = &metadataV1{}` (and so on) now suffices.
- Pre-existing, but this PR extends it to a new path: for a v2 document omitting `last-sequence-number`, `LastSeqNum` ends as the `-1` sentinel with no normalization, and `checkLastSequenceNumber` (`table/metadata.go:2730`) does not catch it when there are no snapshots — so `-1` round-trips back out to JSON.

table/metadata_internal_test.go
```diff
@@ -324,6 +324,80 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+	var metadata metadataV2
+	require.NoError(t, json.Unmarshal([]byte(ExampleTableMetadataV2), &metadata))
+
+	invalid := strings.Replace(ExampleTableMetadataV2, `"current-schema-id": 1`, `"current-schema-id": 99`, 1)
+	require.Error(t, json.Unmarshal([]byte(invalid), &metadata))
+
+	assert.Equal(t, 1, metadata.CurrentSchemaID)
```
zeroshade: Non-blocking, and the same suggestion made on #1646: rather than three hand-picked fields, snapshot `json.Marshal(&metadata)` before the failing decode and assert the bytes are unchanged afterwards. That covers every serialized field at once and would catch a partial regression that these three assertions would miss.

zeroshade:
Re-reviewed after `95a0d56` and `1efe8e2`. These pick up six of the seven notes from my last pass, and several of them go further than I asked — seeding the fixture with the previously-absent optional fields is a better fix for the vacuous-v1 problem than the guard assertions I suggested, and overriding `last-partition-id` to 1001 rather than deleting it tests replacement instead of just absence. Thanks for working through all of them.

For the record, what landed:

- `ParseMetadataBytes` now builds `&metadataV1{}` / `&metadataV2{}` / `&metadataV3{}` (`table/metadata.go:1514-1518`).
- The three dead `aux.LastColumnId = -1` assignments are gone, with the explanation moved to `initCommonMetadataForDeserialization` (`table/metadata.go:1662`) where the value originates.
- The delete-list gained `statistics`, `partition-statistics`, `encryption-keys`, and `last-sequence-number`.
- Failure-path coverage is table-driven across all three versions.
- The failure-path assertion is now a `json.Marshal` before/after byte comparison.

Two things I found this round, both cheap, and then one wording item that is still outstanding.

**1. `TestMetadataUnmarshalPreservesStateOnError/v1` does not currently guard anything.**

I checked this by reverting all three `UnmarshalJSON`s to decode into the receiver via `aux := (*Alias)(m)`. The v2 and v3 subtests fail, as they should. The v1 subtest passes.

The reason is that the invalid v1 document differs from the valid one only inside `"schema"`, and the error it raises — `invalid schema: multiple fields for id 1: x and y` — comes out of `json.Unmarshal` itself. Every field the decoder writes before reaching that point is byte-identical to what the receiver already holds, so `before == after` whether or not the fix is present.

This matters because v1 is the version with the largest partial-mutation surface: `metadataV1.preValidate` (`table/metadata.go:2432-2464`) mutates `SchemaList`, `Specs`, `DefaultSpecID`, `LastPartitionID`, and `SortOrderList` before validation runs.

The fix is to perturb a top-level scalar so a partial write is observable. I verified two candidates — both fail with the fix reverted and pass with it in place:

- `"current-snapshot-id": -1` → `999`, which fails with `invalid metadata: snapshot ref main with ID 999 does not exist in snapshot list`. This is the direct parallel to what the v2 and v3 subtests already do.
- `"last-column-id": 3` → `-5`, which fails with `invalid metadata: missing last-column-id`.

Either works; the first keeps the three subtests symmetric.

**2. The new `assert.Equal(t, int64(-1), metadata.LastSeqNum)` pins a pre-existing bug as expected behavior.**

I raised this leak last time as non-blocking, and it is genuinely pre-existing rather than anything this PR introduces — I confirmed that on `main`. But it is now asserted, which changes its status: a future fix to the sentinel normalization will break this line and read as a regression.

To be concrete about what is being pinned: for a v2 document that omits `last-sequence-number` and has no snapshots, `checkLastSequenceNumber` (`table/metadata.go:2709`) only iterates the snapshot list, so with zero snapshots nothing checks it. On `main` today, `ParseMetadataBytes` on such a document returns successfully, `LastSequenceNumber()` reports `-1`, and re-serializing emits `"last-sequence-number": -1` — an invalid v2 document, through the public API.

A short comment on the assertion noting that `-1` is the un-normalized sentinel and not the desired end state would be enough. Happy to file the underlying issue separately.

**3. Still outstanding from last time: the description.**

`ccc5e4e` has no commit body, so the wording lives only in the PR description, which still reads "Existing version-specific defaults and sentinels are preserved". They are now applied *unconditionally*, which makes direct `UnmarshalJSON` strictly more rejecting — `default-spec-id` becomes required for v2 and `next-row-id` for v3, matching what `ParseMetadataBytes` already enforced. That is the right behavior and unobservable externally since `metadataV{1,2,3}` are unexported, but it is a tightening rather than a preservation, and the description is what carries into the record on squash.

Nothing here touches the production change, which I verified last pass and which is unchanged. On my side: `go test ./table/... ./ -count=1` green, `gofmt -l table/` and `go vet ./table/...` clean, all 15 CI checks passing.

Happy to approve once the v1 fixture is swapped and the description is updated.

table/metadata_internal_test.go
```diff
@@ -324,6 +324,187 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+			require.NotEmpty(t, common.Props)
+			require.NotEmpty(t, common.SnapshotList)
+			require.NotEmpty(t, common.SnapshotLog)
+			require.NotEmpty(t, common.MetadataLog)
+			require.NotEmpty(t, common.SnapshotRefs)
+			require.NotNil(t, common.CurrentSnapshotID)
+			require.NotNil(t, common.LastPartitionID)
+			require.NotEmpty(t, common.StatisticsList)
+			require.NotEmpty(t, common.PartitionStatsList)
+			switch metadata := tt.target.(type) {
```
laskoviymishka, line 391: I think both of these type switches will trip the exhaustive linter we run in CI. They cover `*metadataV2` and `*metadataV3` but there's no `case *metadataV1` and no `default`, and the second switch further down (line 427) has the same gap.

Adding `case *metadataV1:` to both keeps CI green, and it means the test won't silently skip v1 if that type ever grows a `LastSeqNum`-like field.

table/metadata.go
```diff
@@ -1692,6 +1692,7 @@ type commonMetadata struct {
 
 func initCommonMetadataForDeserialization() commonMetadata {
 	return commonMetadata{
+		// Negative sentinels distinguish omitted required fields from explicit zero values.
 		LastColumnId:       -1,
```
laskoviymishka, line 1697: This comment promises more than the validators actually deliver.

`LastUpdatedMS` is set to -1 in the same initializer, but `validate()` checks `== 0`, so a document that omits `last-updated-ms` keeps the -1 sentinel and passes. The "distinguish omitted from explicit zero" invariant doesn't hold for it.

Same shape on `LastSeqNum`, which is the -1 round-trip zeroshade flagged and which you now document in the reuse test. It's spec-required for v2/v3, so a `validate()` guard would close it properly, but I'd be fine scoping this comment to the fields that are really enforced now and leaving the guard as a follow-up. wdyt?

table/metadata_internal_test.go
```diff
@@ -324,6 +324,187 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+	require.NoError(t, json.Unmarshal([]byte(ExampleTableMetadataV2), &metadata))
+
+	var reduced map[string]json.RawMessage
+	require.NoError(t, json.Unmarshal([]byte(ExampleTableMetadataV2), &reduced))
+	delete(reduced, "default-spec-id")
+	reducedData, err := json.Marshal(reduced)
+	require.NoError(t, err)
+
+	require.Error(t, json.Unmarshal(reducedData, &metadata))
+	assert.Equal(t, 0, metadata.DefaultSpecID)
```
laskoviymishka: This assertion doesn't discriminate what we want it to. The preserved `DefaultSpecID` is 0, but 0 is also the Go zero value, so a hypothetical "zero-fill the receiver on failure" regression would pass this too.

I'd seed the first decode with a non-zero `default-spec-id` and assert that value here. The `SnapshotList` length check below is the strong one, so keep that. (This test also overlaps `TestMetadataUnmarshalPreservesStateOnError` on the state-preservation angle, so you could narrow it to just the missing-field error if you'd rather not duplicate.)

table/metadata_internal_test.go
```diff
@@ -324,6 +324,187 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+		{
+			name:    "v1",
+			data:    ExampleTableMetadataV1,
+			invalid: strings.Replace(ExampleTableMetadataV1, `"current-snapshot-id": -1`, `"current-snapshot-id": 999`, 1),
+			target:  &metadataV1{},
+		},
+		{
+			name:    "v2",
+			data:    ExampleTableMetadataV2,
+			invalid: strings.Replace(ExampleTableMetadataV2, `"current-schema-id": 1`, `"current-schema-id": 99`, 1),
```
laskoviymishka: All three invalid cases here are semantically invalid but syntactically valid, so they only exercise the `validate()` error path, which runs after `json.Unmarshal` succeeds. The earlier `json.Unmarshal` error path, the one that returns before `*m = *next`, isn't covered by any case.

A fourth case per version with something like ``invalid: `not json {` `` would exercise it in a line each.

table/metadata.go
```diff
@@ -2653,18 +2662,22 @@ func (m *metadataV3) Equals(other Metadata) bool {
...
-	aux.LastColumnId = -1
+	next := initMetadataV3Deser()
+	aux := (*Alias)(next)
 
 	if err := json.Unmarshal(b, aux); err != nil {
 		return err
 	}
 
-	m.preValidate()
+	next.preValidate()
```
laskoviymishka, line 2677: v1 and v2 `UnmarshalJSON` both call `rejectFieldsBeyondVersion`, v3 doesn't (pre-existing, but this is the natural spot since the PR reworks this body). It's harmless today, but once v4 lands, a v3 object read via `json.Unmarshal` would silently accept v4-only fields where Java's parser rejects them.

Adding the call now with an empty field list documents the pattern and makes the v4 addition mechanical. Non-blocking, up to you whether it's in scope here.

table/metadata_internal_test.go
```diff
@@ -324,6 +324,187 @@ func TestMetadataV3Parsing(t *testing.T) {
...
+			var initial map[string]json.RawMessage
+			require.NoError(t, json.Unmarshal(initialData, &initial))
+			setJSON := func(key string, value any) {
+				encoded, err := json.Marshal(value)
+				require.NoError(t, err)
+				initial[key] = encoded
+			}
+			snapshotID := int64(1925)
+			if tt.name != "v1" {
+				snapshotID = 3055729675574597004
```
laskoviymishka: This is the `current-snapshot-id` from `ExampleTableMetadataV2/V3` hardcoded as a literal. If the fixture's snapshot id ever changes, the test keeps using this value and fails as a confusing "snapshot not found" rather than pointing at the fixture.

I'd pull it from the parsed fixture at setup or give it a named constant.

laskoviymishka (changes requested):
Nice, the fresh-struct-then-assign shape reads cleanly and the atomicity is exactly right (zeroshade already walked through why, so I won't repeat it).

One thing that actually gates merge for me: the two type switches in `TestMetadataUnmarshalReplacesReceiverState` only handle `*metadataV2` and `*metadataV3`, no `*metadataV1` and no `default`. I'm fairly sure that trips the exhaustive linter we run, so it'd fail CI as-is. Adding a `case *metadataV1:` to both fixes it and keeps the test honest.

Agreed with zeroshade on the commit message: the defaults and sentinels aren't "preserved", they now apply unconditionally, which is the actual (and correct) behavior change, and it's worth saying it that way.

The rest is polish. I left an inline on the sentinel comment (it overpromises for `last-updated-ms`, and ties into the `last-sequence-number` round-trip zeroshade flagged — I see you now document that behavior in the test, so scoping the comment would close the loop), plus a couple of test-quality notes so the new tests prove what they mean to.

Fix the switch and reword the commit and I'm happy to approve.

zeroshade (approved):
The two gating items are resolved: the type switches are exhaustive across all three metadata versions, and the description now tells the truth about the behavior change — sentinels applied unconditionally with omitted normalized state rejected, matching `ParseMetadataBytes`, plus the stale-ref-injection and v1-schema-leakage fixes this also delivers. Atomicity shape (fresh struct, assign only on success) was already right.

The remaining open notes are polish; of those I'd only flag v3's missing `rejectFieldsBeyondVersion` as worth an actual follow-up issue so it doesn't get forgotten before a v4 exists.

Deferring thread closure to `laskoviymishka` since his change request predates the fix push.

---

laskoviymishka (approved):

PR #1650, opened 2026-08-05
fix(cli): accept case-insentive catalog types in cli

jx2lee:
## Summary
- `--catalog` values were matched lowercase constants without normarlization, so inputs like `--catalog Rest(or REST)` fell throught to un recognized catalog type.
- Normalize catalog type before dispatching in `initCatalog`
- Add CLI regression coverage that runs `list --catalog REST` againts minimal REST catalog. I'll remove this test func if not needed

## Test
`go test ./cmd/iceberg`

### DIFF
--- a/cmd/iceberg/args_test.go
+++ b/cmd/iceberg/args_test.go
@@ -18,6 +18,8 @@
 package main
 
 import (
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


tanmayrauth (approved):

cmd/iceberg/main.go
```diff
@@ -358,7 +358,7 @@ func initCatalog(ctx context.Context, args Args) catalog.Catalog {
 		err error
 	)
 
-	switch catalog.Type(args.Catalog) {
+	switch catalog.Type(strings.ToLower(args.Catalog)) {
```
zeroshade, line 361: Non-blocking: consider normalizing the catalog type in `catalog.Load` as well. `catalog/registry.go:150-161` currently passes the configured type to an exact-match registry lookup, so the library API remains case-sensitive while the CLI is now lenient. Applying `strings.ToLower` there would make the two entry points consistent.

cmd/iceberg/args_test.go
```diff
@@ -382,6 +384,33 @@ func TestCLIExplicitMissingConfigFailsBeforeCatalogInit(t *testing.T) {
 	assert.Contains(t, string(out), path)
```
zeroshade, line 385: Non-blocking: consider making this table-driven across mixed-case `REST`, `Glue`, `HIVE`, and `Hadoop`, with a negative `RESTX` case. That would cover every existing switch branch and show that normalization does not make unknown types acceptable.

zeroshade (commented):
This change is correct: I verified that `cmd/iceberg/main.go:361` is the CLI's only comparison of `args.Catalog`, and the value is assigned from configuration at `cmd/iceberg/main.go:884`, so config-file and environment-derived catalog types benefit as well. The normalization does not touch case-sensitive URI, warehouse, or credential values.

Optional follow-ups:

- **PR title:** consider correcting the two typos before squash-merge so the commit lands as `fix(cli): accept case-insensitive catalog types`.
- `cmd/iceberg/main.go:361-427`: the nearby `initCatalog` switch has no `catalog.SQL` or `catalog.DynamoDB` branch even though both are declared catalog types. This is pre-existing, but it may be worth tracking separately.

jx2lee:
> This change is correct: I verified that `cmd/iceberg/main.go:361` is the CLI's only comparison of `args.Catalog`, and the value is assigned from configuration at `cmd/iceberg/main.go:884`, so config-file and environment-derived catalog types benefit as well. The normalization does not touch case-sensitive URI, warehouse, or credential values.
> 
> Optional follow-ups:
> 
> - **PR title:** consider correcting the two typos before squash-merge so the commit lands as `fix(cli): accept case-insensitive catalog types`.
> - `cmd/iceberg/main.go:361-427`: the nearby `initCatalog` switch has no `catalog.SQL` or `catalog.DynamoDB` branch even though both are declared catalog types. This is pre-existing, but it may be worth tracking separately.

thanks for the review and comment 🎶

I'll follow up seperately for missing catalog type in CLI! (dynamo and sql type) #1692

cmd/iceberg/main.go
jx2lee: @zeroshade good point. fixed to lower in registry, and added test case!

cmd/iceberg/args_test.go
jx2lee: @zeroshade added mixed-case for existed test. Please review that. 👍

zeroshade (approved):

PR #945, opened 2026-04-27
feat(table): refresh-and-replay between commit retries

laskoviymishka:
Closes #830 (concurrent-writer conflict detection) — fifth and final PR in the sequence. The retry loop in Table.doCommit now actually does something useful for contended commits.

Before this change, doCommit's retry loop re-issued the same updates and requirements between attempts. After a peer advanced the branch with a non-conflicting commit, AssertRefSnapshotID deterministically failed on every retry and the budget was wasted. The default num-retries was set to 0 in #912 precisely to avoid this slow-fail.

What changes:

  - On ErrCommitFailed mid-loop, reload the catalog's current metadata via cat.LoadTable.
  - Build a fresh conflictContext with (base = writer's view, current = freshly loaded). Validators now run against real concurrent-snapshot state — the framework wired in #934 finally fires.
  - Rewrite AssertRefSnapshotID requirements to point at the new branch head before re-submitting. Mirrors Java's SnapshotProducer between-retry behavior.
  - The pre-flight pass moved inside the loop so attempt 0 still gets a no-op validation (base == current → empty concurrent list) without a separate code path.

Behavior matrix after this PR (with num-retries > 0):

  - Peer advance with no semantic conflict: validators pass on refresh, requirements rewritten, retry succeeds. Was: spin to exhaustion.
  - Peer advance with semantic conflict (e.g. same partition overwrite): validators reject on refresh, retry exits terminally. Was: spin to exhaustion.
  - Diverged base (branch deleted): newConflictContext returns ErrCommitDiverged, terminal exit. Same as before.

Also tightens flakyCatalog's LoadTable to return a real Table backed by its current metadata, so refresh paths in retry tests no longer dereference nil.

E2E tests cover the headline success path (peer advance, retry succeeds with rewritten assertion), the terminal-on-conflict path (validator rejects on refresh), and the rewriteRefSnapshot- Requirements helper directly.

Issue #1015, opened 2026-05-05
discussion: align CommitNumRetriesDefault with Java/PyIceberg (0 → 4)?


laskoviymishka:
Parent: #830

Open question, not a blocker. iceberg-go's `CommitNumRetriesDefault` was set to `0` in PR #912 because the retry loop without refresh-and-replay was harmful (re-issued the same stale conflict). PR #945 added refresh-and-replay, so the original reason for the `0` default is gone — but the constant has not been flipped.

Java and PyIceberg both default to `4` retries. A user porting from either of those clients sees fail-fast on the first conflict today and has to dig into properties to figure out why. The fix is a one-line constant change plus a sweep of tests that assume the no-retry path.

Reasons to flip:
- Java parity / out-of-box experience.
- The retry mechanism is fully wired and exercised in CI.
- Existing users who *want* fail-fast can set `commit.retry.num-retries=0` explicitly.

Reasons to hold off:
- Existing users (since #912 in April) have been running with default `0`. Flipping silently changes their commit-latency profile under contention — a 4-attempt exponential-backoff loop can add seconds before a terminal error surfaces.
- Some retry tests calibrated for `0` would need adjustment.

Filing as a question, not a feature, because the answer is a defaults-tuning judgment call rather than a missing capability. Looking for input from maintainers on whether the parity argument outweighs the silent-behavior-change concern, and ideally a release note if we do flip.

PR #1652, opened 2026-08-05
Fix stale commit-retry doc comment

alessandro-nori:
The comment says the 0 default exists "until refresh-and-replay lands," but refresh-and-replay was added in #945.

### DIFF
--- a/table/properties.go
+++ b/table/properties.go
@@ -111,12 +111,13 @@ const (
 	// CommitNumRetriesKey is the number of commit retry attempts before
 	// giving up on ErrCommitFailed from the catalog.
 	//
-	// The default is 0 (no retries) until refresh-and-replay lands; a
-	// retry loop that reuses the original updates/requirements will
-	// fail deterministically on genuine OCC conflicts and only slow
-	// down the final error. Callers that observe transient catalog
-	// flakiness (dropped connections, brief 409 during leader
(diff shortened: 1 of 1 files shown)


twuebi:
We could also default to some sane amount of retries, wdyt?

alessandro-nori:
> #945

@twuebi there is an [issue open](https://github.com/apache/iceberg-go/issues/1015) for that and I don't know if we want to make a breaking change now

zeroshade (approved):
I'm fine with adding that to this PR or doing it as a follow-up. Just let me know which you plan on, if you plan on doing it as a follow-up I'll merge this

laskoviymishka (approved):

laskoviymishka:
>  We could also default to some sane amount of retries, wdyt?

commit.retry.num-retries still defaults to 0 because it was parked there as a safety stop while blind retries could corrupt state, and the planned flip to 4 never happened after refresh-and-replay landed, so it's now just a stale default rather than a real blocker (open question tracked in #1015).

But changing default has some risks:
- It's a silent behavior change: commits that used to fail instantly now go through a few backoff cycles first, so hard conflicts surface a bit slower (small with the default 100ms/60s backoff, but non-zero).
- The retry path exercises more machinery by default

I would do this as separate PR.

PR #1655, opened 2026-08-05
feat(metrics): add REST metrics reporter (#1236)

tanmayrauth:
6th PR of the metrics reporting framework. Implements the previously stubbed catalog metrics endpoint:
  
  restMetricsReporter POSTs a ReportMetricsRequest to a table's POST .../tables/{table}/metrics endpoint. The send runs on a background goroutine detached from the caller's cancellation, and any error is logged and swallowed, so
  reporting never blocks or fails the observed scan/commit.
  The REST catalog combines it with the configured reporter when loading a table, gated on the opt-in rest.metrics-reporting-enabled property and on the server advertising the endpoint. It is off by default, so existing REST users
  see no new /metrics traffic.
  Tests use a custom RoundTripper (no network listener) to assert the async POST, the flattened report-type wire format, nil-report no-op, and that Report never blocks on a slow server.

  Behavior change (all REST catalog users): OAuth token refresh now runs on an *http.Client with a 60s timeout (defaultOAuthTimeout). Because oauth2's TokenSource.Token() takes no context, this timeout is what bounds a stalled or  black-holing token endpoint. It applies to every REST catalog user, not just metrics opt-ins — a token endpoint that previously hung indefinitely will now fail refresh after 60s. Worth calling out in release notes.

  Related to #1236

### DIFF
--- a/catalog/rest/metrics_reporter.go
+++ b/catalog/rest/metrics_reporter.go
@@ -118,8 +118,11 @@ type metricsJob struct {
 // rather than accumulating goroutines and connections. Close cancels in-flight
 // reports and drains the workers.
 type metricsDispatcher struct {
-	jobs      chan metricsJob
-	timeout   time.Duration
+	jobs    chan metricsJob
+	timeout time.Duration
+	// ctx/cancel are the dispatcher's own lifecycle context, deliberately stored
+	// on the struct (the documented exception to "don't store a context in a
+	// struct"): Close cancels it to abort in-flight reports and stop the workers.
 	ctx       context.Context
 	cancel    context.CancelFunc
 	wg        sync.WaitGroup
@@ -278,9 +281,20 @@ func (d *metricsDispatcher) close() {
 			close(done)
 		}()
 
+		// Explicit timer with a deferred Stop rather than time.After, which would
+		// leak a live timer (up to the full timeout) on the common path where the
+		// workers finish first — it adds up in a process that cycles through many
+		// catalogs.
+		t := time.NewTimer(d.timeout)
+		defer t.Stop()
 		select {
 		case <-done:
-		case <-time.After(d.timeout):
+		case <-t.C:
+			// Shutdown budget expired before the workers returned (a report wedged in
+			// auth, say). d.ctx is already cancelled, so each worker self-exits once
+			// its attempt unblocks — nothing leaks — but Close can return here before
+			// the workers have fully stopped; do not tear down a shared transport on
+			// the assumption that they have.
 		}
 
 		close(d.closeDone)
--- a/catalog/rest/metrics_reporter.go
+++ b/catalog/rest/metrics_reporter.go
@@ -1,0 +1,76 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package rest
+
+import (
+	"context"
+	"log/slog"
+	"net/http"
+	"net/url"
+
+	"github.com/apache/iceberg-go/metrics"
+)
+
+// keyReportMetricsEnabled opts a REST catalog into POSTing scan/commit reports
+// to the catalog's metrics endpoint. It is disabled by default so existing
+// users see no new network traffic unless they turn it on.
+const keyReportMetricsEnabled = "rest.metrics-reporting-enabled"
+
+// restMetricsReporter POSTs metrics reports to a table's REST metrics endpoint
+// (POST .../tables/{table}/metrics). It is bound to a single table's path and
+// satisfies metrics.Reporter.
+type restMetricsReporter struct {
+	baseURI *url.URL
+	cl      *http.Client
+	path    []string // table metrics path, relative to baseURI
+}
+
+var _ metrics.Reporter = (*restMetricsReporter)(nil)
+
+// Report wraps the report in a ReportMetricsRequest and POSTs it on a
+// background goroutine. Per the Reporter contract it never blocks or fails the
+// observed scan/commit: the send is detached from the caller's cancellation,
+// and any error is logged and swallowed.
+func (rep *restMetricsReporter) Report(ctx context.Context, report metrics.MetricsReport) {
+	if report == nil {
+		return
+	}
+
+	req := metrics.NewReportMetricsRequest(report)
+	// Detach from the caller's cancellation (the scan/commit is already done)
+	// while preserving any context values used by the HTTP client.
+	sendCtx := context.WithoutCancel(ctx)
+
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				slog.Default().Warn("iceberg: panic while reporting metrics to REST catalog", "recovered", r)
+			}
+		}()
+
+		if _, err := doPost[metrics.ReportMetricsRequest, struct{}](
+			sendCtx, rep.baseURI, rep.path, req, rep.cl, nil, allowNoContent()); err != nil {
+			slog.Default().Warn("iceberg: failed to report metrics to REST catalog", "error", err)
+		}
+	}()
+}
+
+// Close satisfies [metrics.Reporter]. The reporter is stateless — it borrows
+// the catalog's shared HTTP client rather than owning one — so there is nothing
+// to release.
+func (rep *restMetricsReporter) Close() error { return nil }
--- a/catalog/rest/metrics_reporter.go
+++ b/catalog/rest/metrics_reporter.go
@@ -22,7 +22,9 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	iceberg "github.com/apache/iceberg-go"
@@ -39,23 +41,28 @@ const (
 	// keyReportMetricsEnabledLegacy is the historical dotted spelling, accepted
 	// as an alias so existing configs keep working.
 	keyReportMetricsEnabledLegacy = "rest.metrics-reporting-enabled"
-	// keyReportMetricsTimeoutMs bounds a single report's total auth + request +
-	// response cycle, in milliseconds. Read from the client-supplied properties
-	// only.
+	// keyReportMetricsTimeoutMs bounds a single report's request and response
+	// cycle, in milliseconds. Read from the client-supplied properties only. Auth
+	// (token refresh) is bounded separately by the OAuth refresh client Timeout.
 	keyReportMetricsTimeoutMs = "rest-metrics-reporting-timeout-ms"
 
-	// defaultReportMetricsTimeout bounds a single report's total auth + request
-	// + response cycle when keyReportMetricsTimeoutMs is unset. Telemetry is safe
-	// to drop, so the bound is deliberately short.
+	// defaultReportMetricsTimeout bounds a single report's request and response
+	// cycle when keyReportMetricsTimeoutMs is unset. Telemetry is safe to drop, so
+	// the bound is deliberately short.
 	defaultReportMetricsTimeout = 10 * time.Second
 	// metricsDispatchWorkers is the fixed number of goroutines draining the
 	// dispatch queue, capping the reporter's concurrent connection use.
 	metricsDispatchWorkers = 4
 	// metricsDispatchQueueSize bounds how many reports may await dispatch.
-	// Reports offered while the queue is full are dropped (and logged) rather
-	// than queued without limit, so a stalled endpoint cannot make reporting grow
-	// without bound.
+	// Reports offered while the queue is full are dropped (and counted, then
+	// logged in aggregate) rather than queued without limit, so a stalled
+	// endpoint cannot make reporting grow without bound.
 	metricsDispatchQueueSize = 128
+	// metricsStatsInterval is how often the dispatcher emits an aggregated
+	// warning summarizing dropped and failed reports. Aggregating keeps a stalled
+	// or unavailable endpoint from amplifying into one log line per report while
+	// still surfacing back-pressure.
+	metricsStatsInterval = 30 * time.Second
 )
 
 // reportMetricsEnabled reports whether the client opted into REST metrics
@@ -63,9 +70,19 @@ const (
 // given the client-supplied properties only (never the server-merged config) so
 // a server cannot flip the default and turn on outbound telemetry the client
 // never asked for.
+//
+// The canonical key wins whenever it is present, so a client that has migrated
+// to it and explicitly set it to false is honored even if a stale legacy dotted
+// key lingers in the config. The legacy key is consulted only when the canonical
+// key is absent.
 func reportMetricsEnabled(props iceberg.Properties) bool {
-	return props.GetBool(keyReportMetricsEnabled, false) ||
-		props.GetBool(keyReportMetricsEnabledLegacy, false)
+	if v, ok := props[keyReportMetricsEnabled]; ok {
+		enabled, err := strconv.ParseBool(v)
+
+		return err == nil && enabled
+	}
+
+	return props.GetBool(keyReportMetricsEnabledLegacy, false)
 }
 
 // reportMetricsTimeout resolves the per-report deadline from the client-supplied
@@ -82,6 +99,11 @@ func reportMetricsTimeout(props iceberg.Properties) time.Duration {
 
 // metricsJob is a single report awaiting dispatch to a table's metrics endpoint.
 type metricsJob struct {
+	// ctx carries the observed scan/commit's context values (trace spans,
+	// request-scoped attributes) with its cancellation detached — the scan/commit
+	// is already done, so its cancellation must not abort the report, but its
+	// values should still propagate to the outbound request.
+	ctx     context.Context
 	baseURI *url.URL
 	cl      *http.Client
 	path    []string
@@ -92,32 +114,38 @@ type metricsJob struct {
 // pool of workers draining a bounded queue. It is owned by the catalog and
 // shared across that catalog's table reporters, so concurrent report volume
 // stays bounded no matter how many tables are loaded or how often they are
-// scanned. A stalled endpoint sheds load — reports are dropped and logged —
+// scanned. A stalled endpoint sheds load — reports are dropped and counted —
 // rather than accumulating goroutines and connections. Close cancels in-flight
 // reports and drains the workers.
 type metricsDispatcher struct {
-	jobs    chan metricsJob
-	timeout time.Duration
-	ctx     context.Context
-	cancel  context.CancelFunc
-	wg      sync.WaitGroup
-	logger  *slog.Logger // nil means resolve slog.Default at call time
+	jobs      chan metricsJob
+	timeout   time.Duration
+	ctx       context.Context
+	cancel    context.CancelFunc
+	wg        sync.WaitGroup
+	logger    *slog.Logger // nil means resolve slog.Default at call time
+	dropped   atomic.Uint64
+	failed    atomic.Uint64
+	closeOnce sync.Once
+	closeDone chan struct{}
 }
 
 func newMetricsDispatcher(workers, queueSize int, timeout time.Duration, logger *slog.Logger) *metricsDispatcher {
 	ctx, cancel := context.WithCancel(context.Background())
 	d := &metricsDispatcher{
-		jobs:    make(chan metricsJob, queueSize),
-		timeout: timeout,
-		ctx:     ctx,
-		cancel:  cancel,
-		logger:  logger,
+		jobs:      make(chan metricsJob, queueSize),
+		timeout:   timeout,
+		ctx:       ctx,
+		cancel:    cancel,
+		logger:    logger,
+		closeDone: make(chan struct{}),
 	}
 
-	d.wg.Add(workers)
+	d.wg.Add(workers + 1)
 	for range workers {
 		go d.worker()
 	}
+	go d.reportStats()
 
 	return d
 }
@@ -155,28 +183,38 @@ func (d *metricsDispatcher) send(job metricsJob) {
 		}
 	}()
 
-	// Derive from the dispatcher context so Close cancels in-flight requests,
-	// and add the per-report deadline so a stalled endpoint cannot pin a worker
-	// (and its connection) indefinitely. The deadline covers auth plus the full
-	// request and response cycle.
-	ctx, cancel := context.WithTimeout(d.ctx, d.timeout)
+	// Bound the report by the per-report deadline, starting from the job's
+	// context so the caller's values (trace spans, request-scoped attributes)
+	// still reach the transport, and tie the derived context to dispatcher
+	// shutdown so Close cancels in-flight requests. The deadline covers the
+	// request and response cycle; a token refresh triggered by auth is bounded
+	// separately by the OAuth refresh client's Timeout.
+	ctx, cancel := context.WithTimeout(job.ctx, d.timeout)
 	defer cancel()
+	stop := context.AfterFunc(d.ctx, cancel)
+	defer stop()
 
 	if _, err := doPost[metrics.ReportMetricsRequest, struct{}](
 		ctx, job.baseURI, job.path, job.req, job.cl, nil, allowNoContent()); err != nil {
 		// A report interrupted by Close (dispatcher context cancelled) is expected
-		// shutdown behavior, not a failure worth logging. A per-report timeout
-		// leaves the dispatcher context live, so genuine timeouts still surface.
+		// shutdown behavior, not a failure worth counting. A per-report timeout
+		// leaves the dispatcher context live, so genuine failures still surface.
 		if d.ctx.Err() != nil {
 			return
 		}
-		d.log().Warn("iceberg: failed to report metrics to REST catalog", "error", err)
+		d.failed.Add(1)
+		// The transport error embeds the request URL, which can carry a sensitive
+		// host, prefix, namespace or table name, so keep the detail at Debug and
+		// let reportStats surface the aggregate count at Warn.
+		d.log().Debug("iceberg: failed to report metrics to REST catalog", "error", err)
 	}
 }
 
 // submit offers a job to the queue without blocking. It drops the report (and
-// logs the drop, so back-pressure is visible rather than silent) when the queue
-// is full, and ignores reports once the dispatcher is closed.
+// counts the drop, aggregated and logged by reportStats so back-pressure is
+// visible without amplifying into one log line per drop) when the queue is
+// full, and ignores reports once the dispatcher is closed. Every path here is
+// non-blocking so Report never blocks the observed scan/commit.
 func (d *metricsDispatcher) submit(job metricsJob) {
 	select {
 	case <-d.ctx.Done():
@@ -188,25 +226,67 @@ func (d *metricsDispatcher) submit(job metricsJob) {
 	case d.jobs <- job:
 	case <-d.ctx.Done():
 	default:
-		d.log().Warn("iceberg: metrics report dropped; dispatch queue full")
+		d.dropped.Add(1)
+	}
+}
+
+// reportStats periodically emits an aggregated warning summarizing reports
+// dropped (queue full) and failed (endpoint errors) since the last tick, so a
+// stalled or unavailable endpoint surfaces as bounded, rate-limited log volume
+// rather than one line per report. A final summary is emitted on shutdown.
+func (d *metricsDispatcher) reportStats() {
+	defer d.wg.Done()
+
+	ticker := time.NewTicker(metricsStatsInterval)
+	defer ticker.Stop()
+
+	var lastDropped, lastFailed uint64
+	flush := func() {
+		dropped, failed := d.dropped.Load(), d.failed.Load()
+		if dropped == lastDropped && failed == lastFailed {
+			return
+		}
+		d.log().Warn("iceberg: REST metrics reports dropped or failed",
+			"dropped", dropped-lastDropped, "dropped_total", dropped,
+			"failed", failed-lastFailed, "failed_total", failed)
+		lastDropped, lastFailed = dropped, failed
+	}
+
+	for {
+		select {
+		case <-d.ctx.Done():
+			flush()
+
+			return
+		case <-ticker.C:
+			flush()
+		}
 	}
 }
 
 // close cancels in-flight reports and waits for the workers to return, bounded
-// by the report timeout so shutdown cannot hang on a stalled endpoint.
+// by the report timeout so shutdown cannot hang on a stalled endpoint. It is
+// one-shot: repeated calls block until the first completes, then return, so no
+// extra waiter goroutine is spawned per call.
 func (d *metricsDispatcher) close() {
-	d.cancel()
+	d.closeOnce.Do(func() {
+		d.cancel()
 
-	done := make(chan struct{})
-	go func() {
-		d.wg.Wait()
-		close(done)
-	}()
+		done := make(chan struct{})
+		go func() {
+			d.wg.Wait()
+			close(done)
+		}()
 
-	select {
-	case <-done:
-	case <-time.After(d.timeout):
-	}
+		select {
+		case <-done:
+		case <-time.After(d.timeout):
+		}
+
+		close(d.closeDone)
+	})
+
+	<-d.closeDone
 }
 
 // restMetricsReporter POSTs metrics reports for a single table to its REST
@@ -224,14 +304,21 @@ var _ metrics.Reporter = (*restMetricsReporter)(nil)
 // Report wraps the report in a ReportMetricsRequest and hands it to the
 // catalog's dispatcher, returning immediately. Per the Reporter contract it
 // never blocks or fails the observed scan/commit: dispatch is asynchronous and
-// bounded, the send is detached from the caller's cancellation (the scan/commit
-// is already done), and any error is logged and swallowed by the dispatcher.
-func (rep *restMetricsReporter) Report(_ context.Context, report metrics.MetricsReport) {
+// bounded, and any error is counted and swallowed by the dispatcher. The
+// caller's context is detached from cancellation (the scan/commit is already
+// done) but its values are preserved so trace spans and request-scoped
+// attributes still propagate to the outbound report.
+func (rep *restMetricsReporter) Report(ctx context.Context, report metrics.MetricsReport) {
 	if report == nil {
 		return
 	}
 
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
 	rep.dispatcher.submit(metricsJob{
+		ctx:     context.WithoutCancel(ctx),
 		baseURI: rep.baseURI,
 		cl:      rep.cl,
 		path:    rep.path,
--- a/catalog/rest/metrics_reporter.go
+++ b/catalog/rest/metrics_reporter.go
@@ -22,55 +22,225 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"sync"
+	"time"
 
+	iceberg "github.com/apache/iceberg-go"
 	"github.com/apache/iceberg-go/metrics"
 )
 
-// keyReportMetricsEnabled opts a REST catalog into POSTing scan/commit reports
-// to the catalog's metrics endpoint. It is disabled by default so existing
-// users see no new network traffic unless they turn it on.
-const keyReportMetricsEnabled = "rest.metrics-reporting-enabled"
+const (
+	// keyReportMetricsEnabled opts a REST catalog into POSTing scan/commit
+	// reports to the catalog's metrics endpoint. It is disabled by default so
+	// existing users see no new network traffic unless they turn it on. This is
+	// the canonical, cross-implementation spelling used by Iceberg Java and the
+	// Iceberg docs; keyReportMetricsEnabledLegacy is accepted as an alias.
+	keyReportMetricsEnabled = "rest-metrics-reporting-enabled"
+	// keyReportMetricsEnabledLegacy is the historical dotted spelling, accepted
+	// as an alias so existing configs keep working.
+	keyReportMetricsEnabledLegacy = "rest.metrics-reporting-enabled"
+	// keyReportMetricsTimeoutMs bounds a single report's total auth + request +
+	// response cycle, in milliseconds. Read from the client-supplied properties
+	// only.
+	keyReportMetricsTimeoutMs = "rest-metrics-reporting-timeout-ms"
 
-// restMetricsReporter POSTs metrics reports to a table's REST metrics endpoint
-// (POST .../tables/{table}/metrics). It is bound to a single table's path and
-// satisfies metrics.Reporter.
-type restMetricsReporter struct {
+	// defaultReportMetricsTimeout bounds a single report's total auth + request
+	// + response cycle when keyReportMetricsTimeoutMs is unset. Telemetry is safe
+	// to drop, so the bound is deliberately short.
+	defaultReportMetricsTimeout = 10 * time.Second
+	// metricsDispatchWorkers is the fixed number of goroutines draining the
+	// dispatch queue, capping the reporter's concurrent connection use.
+	metricsDispatchWorkers = 4
+	// metricsDispatchQueueSize bounds how many reports may await dispatch.
+	// Reports offered while the queue is full are dropped (and logged) rather
+	// than queued without limit, so a stalled endpoint cannot make reporting grow
+	// without bound.
+	metricsDispatchQueueSize = 128
+)
+
+// reportMetricsEnabled reports whether the client opted into REST metrics
+// reporting, accepting both the canonical and the legacy dotted key. It must be
+// given the client-supplied properties only (never the server-merged config) so
+// a server cannot flip the default and turn on outbound telemetry the client
+// never asked for.
+func reportMetricsEnabled(props iceberg.Properties) bool {
+	return props.GetBool(keyReportMetricsEnabled, false) ||
+		props.GetBool(keyReportMetricsEnabledLegacy, false)
+}
+
+// reportMetricsTimeout resolves the per-report deadline from the client-supplied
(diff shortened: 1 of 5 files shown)


catalog/rest/rest.go
```diff
@@ -1109,6 +1109,15 @@ func (r *Catalog) tableFromResponse(_ context.Context, identifier []string, meta
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
+	// Opt-in: POST scan/commit reports to the catalog's metrics endpoint, but
+	// only when enabled and the server advertises the endpoint.
+	if config.GetBool(keyReportMetricsEnabled, false) && r.endpoints.check(endpointReportMetrics) == nil {
```
zeroshade: `config` here is the **merged** table and catalog configuration: the server's `defaults`, the server's `overrides`, and the properties on the table response all feed into it. So this condition is satisfied whenever a *server* sets `rest.metrics-reporting-enabled=true`, and this client will start POSTing scan and commit telemetry to that server without the user having opted in.

That is the inverse of what the PR description and issue #1236 describe. The constant's own doc comment says reporting is "disabled by default so existing users see no new network traffic unless they turn it on" — but "they" here can be the server. A default that a remote party can flip is not a default.

The comment immediately above this block already articulates the principle: server-vended reporter selection "should be an explicit decision rather than a side effect of merge order." Turning on network telemetry is a strictly larger decision than choosing between reporter implementations, so it should be held to at least the same standard.

Suggested fix: read enablement only from client-originated configuration — `r.reporterProps`, as line 1108 already does for reporter selection, or a dedicated explicit catalog option — and never from the merged `config`. Then pin the precedence with a test: server `defaults`, server `overrides`, and table-response properties each setting the key to `true` must leave reporting off when the client has not enabled it.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,76 @@
...
+// and any error is logged and swallowed.
+func (rep *restMetricsReporter) Report(ctx context.Context, report metrics.MetricsReport) {
+	if report == nil {
+		return
+	}
+
+	req := metrics.NewReportMetricsRequest(report)
+	// Detach from the caller's cancellation (the scan/commit is already done)
+	// while preserving any context values used by the HTTP client.
+	sendCtx := context.WithoutCancel(ctx)
```
zeroshade: Detaching from the caller is the right instinct — the scan or commit has already finished and its cancellation should not abort the report. The problem is that `context.WithoutCancel` removes the caller's deadline without installing one of its own, and the catalog's shared `http.Client` has no `Timeout` set either. This request therefore has no bound at any layer.

Concretely, that means an unreachable or black-holing endpoint, a token refresh that never returns, a wait on response headers, or a response body that never finishes draining will block for the lifetime of the process.

Suggested fix: keep the detach, and add a deadline of its own:

```go
sendCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), reportTimeout)
```

with `defer cancel()` inside the goroutine so it covers auth plus the complete request and response cycle. A conservative default of a few seconds, ideally configurable, seems right for telemetry that is safe to drop.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,76 @@
...
+
+	go func() {
```
zeroshade: Every call to `Report` spawns a goroutine that nothing tracks and nothing bounds. Paired with the missing timeout above, a metrics endpoint that accepts connections but never responds accumulates one permanently blocked goroutine — and one held connection — per scan and per commit, for as long as the process runs.

That is where the isolation guarantee stops holding. Any individual report is harmless, but the unbounded accumulation can exhaust connections and eventually degrade the very scans and commits this reporter is meant to stay out of the way of.

Suggested fix: move dispatch to a catalog-owned bounded queue served by a small fixed worker pool, rather than spawning per report. When the queue is full, drop the report and log the drop — shedding telemetry is the correct behavior here, and logging it makes back-pressure visible instead of silent. Tracking the workers also gives `Close` something to drain (see below).

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,76 @@
...
+			sendCtx, rep.baseURI, rep.path, req, rep.cl, nil, allowNoContent()); err != nil {
+			slog.Default().Warn("iceberg: failed to report metrics to REST catalog", "error", err)
+		}
+	}()
+}
+
+// Close satisfies [metrics.Reporter]. The reporter is stateless — it borrows
+// the catalog's shared HTTP client rather than owning one — so there is nothing
+// to release.
+func (rep *restMetricsReporter) Close() error { return nil }
```
zeroshade, line 347: The comment is accurate that the reporter borrows the catalog's HTTP client rather than owning one — but as of this PR it does own goroutines, and a no-op `Close` cannot reclaim them. In-flight reports outlive the catalog with nothing to cancel them and nothing to wait on, so a caller that closes the catalog and expects its outbound network activity to stop does not get that.

Suggested fix: once reporting is backed by a catalog-owned worker pool, have `Close` cancel the pool's context and wait for the workers to return, bounded by the reporting timeout so shutdown cannot hang either. That makes shutdown observable and gives the drain path something to test.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,76 @@
...
+	"net/http"
+	"net/url"
+
+	"github.com/apache/iceberg-go/metrics"
+)
+
+// keyReportMetricsEnabled opts a REST catalog into POSTing scan/commit reports
+// to the catalog's metrics endpoint. It is disabled by default so existing
+// users see no new network traffic unless they turn it on.
+const keyReportMetricsEnabled = "rest.metrics-reporting-enabled"
```
zeroshade: The property name here is `rest.metrics-reporting-enabled`, with a dot after `rest`. Iceberg Java and the Iceberg documentation both use `rest-metrics-reporting-enabled`, with a hyphen.

That one character makes this feature unreachable for anyone configuring it the documented, cross-implementation way: they set the canonical key, this code never sees it, and reporting silently stays off with no error and no warning to explain why. Silent no-ops on a configuration key are particularly expensive to debug, since the natural conclusion is that the feature is broken rather than misnamed.

Suggested fix: make `rest-metrics-reporting-enabled` the canonical key. If the dotted spelling is worth keeping, accept it as a deliberate alias rather than as the only accepted name. It would also help to export the constant, or expose a typed catalog option, so callers are not hardcoding the string at all.

catalog/rest/metrics_reporter_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/apache/iceberg-go/metrics"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type capturedRequest struct {
```
zeroshade, line 43: Non-blocking. `capturedRequest` records only the method, path, and body, so no test can currently assert anything about headers. Reusing the catalog's authenticated client is one of the more valuable properties of this design, and right now nothing verifies it.

Suggested fix: capture `r.Header` as well, then assert that the metrics POST carries the `Authorization` header — including after a token refresh — along with `Content-Type: application/json` and any headers configured on the catalog.

catalog/rest/metrics_reporter_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	return &http.Response{
+		StatusCode: http.StatusNoContent,
+		Body:       io.NopCloser(bytes.NewReader(nil)),
+		Header:     make(http.Header),
+	}, nil
+}
+
+func newTestReporter(t *testing.T, tr *captureTransport) *restMetricsReporter {
```
zeroshade: Non-blocking. This helper hands the reporter a pre-built relative path, so the assertion at line 89 only ever checks `/namespaces/db/tables/t/metrics`. The parts most likely to break in practice are exactly the parts skipped: `reqPath` composing the full `/v1/{prefix}/...` URL, multi-level namespaces, and identifier escaping.

Suggested fix: add a test that goes through `tableFromResponse` — or at minimum `endpointReportMetrics.reqPath` — with a configured prefix and a nested namespace containing a character that requires escaping, and assert the complete request path.

zeroshade (changes requested):
Thanks for taking this on — the wire-level half of this is right, and that is the part that is easiest to get wrong. The endpoint (`POST .../namespaces/{ns}/tables/{table}/metrics`), the `ReportMetricsRequest` payload with its `report-type` discriminator and flattened report fields, the prefix handling through `endpointReportMetrics.reqPath`, and the reuse of the catalog's already-authenticated `http.Client` all conform to the Iceberg REST spec. Gating on `r.endpoints.check(endpointReportMetrics)` and composing through `metrics.Combine` so a user-configured reporter still receives its reports are both the right calls.

The failure-isolation shape is also mostly right: `Report` dispatches and returns immediately, HTTP errors are logged and swallowed rather than propagated, and scans and commits only report after their work has completed. A single failed POST therefore does not synchronously fail or block the operation that triggered it. That property is worth preserving through the changes below.

My concerns are about who gets to turn this on, and about what happens when the metrics endpoint is not merely failing but hanging.

## Enablement is server-controlled, not client-controlled

This is the one I would like settled before the rest. At `catalog/rest/rest.go:1114` the enablement flag is read from `config` — the **merged** table and catalog configuration. That map includes the server's `defaults`, the server's `overrides`, and the properties returned on the table response itself. A server can therefore set `rest.metrics-reporting-enabled=true` and this client will begin POSTing scan and commit telemetry to it, having never opted in to anything.

That contradicts both the PR description and issue #1236, which frame this as opt-in and off by default. As written, the default holds only until a server decides otherwise. The comment three lines above already states the principle: server-vended reporter selection "should be an explicit decision rather than a side effect of merge order." Enablement of network reporting deserves at least the same protection. Details and a suggested fix are inline.

## Isolation holds against a failing endpoint, not a hanging one

The three remaining blocking items compound into one failure mode. `context.WithoutCancel` at `metrics_reporter.go:57` removes the caller's deadline without installing a replacement, and the shared HTTP client carries no `Timeout` — so a report can hang indefinitely. Each report at `:59` occupies its own untracked, unbounded goroutine, and `Close` at `:76` is a no-op, so those goroutines and their connections accumulate and outlive the catalog.

Individually each is small; together they mean a metrics endpoint that black-holes traffic leaks one permanently blocked goroutine and one held connection per scan and per commit, indefinitely, with no way to stop them. That is the point at which telemetry stops being isolated from the operations it observes. Each is commented inline with a suggested fix.

## Naming

`metrics_reporter.go:32` uses `rest.metrics-reporting-enabled`, where Iceberg Java and the Iceberg docs use `rest-metrics-reporting-enabled`. Anyone configuring this from the documented cross-implementation property name gets silence — no reporting, no error, no warning. Inline.

## Test coverage

What is there is genuinely useful: scan and commit POST dispatch, method and relative path, the report-type discriminators, the flattened payload shape, nil-report handling, and the proof that `Report` returns promptly while the transport is blocked.

The gaps that matter most, roughly in order:

1. **A request that blocks until `Context.Done()`**, proving a finite reporting timeout actually exists.
2. **Burst reporting against a stalled endpoint**, proving concurrency stays bounded and excess reports are dropped rather than queued forever.
3. **Catalog `Close` cancelling in-flight reports.**
4. **Catalog-level enablement precedence** — default-disabled; explicitly client-enabled; and, most importantly, server `defaults`/`overrides`/table-response properties setting the key to `true` *not* enabling reporting.
5. **Advertised vs. unadvertised `report-metrics` endpoint**, and confirmation that a user-configured custom reporter still receives reports alongside the REST one.
6. **Full path construction** — `/v1/{prefix}/...` with a nested namespace and an identifier needing escaping.
7. **Auth and token refresh** on the metrics request.
8. **Transport errors and non-2xx responses** being logged and swallowed.
9. **End-to-end scan/commit isolation** against a permanently unavailable endpoint.

tanmayrauth:
Thanks for the review @zeroshade. I've pushed a commit that addresses every item. Summary:

  - Enablement is now client-only. Reporting is gated on reportMetricsEnabled(r.reporterProps), which reads the properties captured before /v1/config merges in, exactly as reporter selection already does. Server defaults, overrides,
  and table-response properties can no longer turn it on.
  - Reports have a finite timeout. Each POST runs under context.WithTimeout(dispatcherCtx, timeout) (default 10s, configurable via rest-metrics-reporting-timeout-ms), covering auth + full request/response.
  - Dispatch is bounded. Per-report goroutines are gone; reports now go through a catalog-owned worker pool (4 workers, 128-slot queue). A full queue drops the report and logs it.
  - Close now does its job. Catalog.Close cancels in-flight reports and drains the workers, bounded by the timeout so shutdown can't hang.
  - Canonical key. Renamed to rest-metrics-reporting-enabled; the dotted spelling is kept as a backward-compatible alias.
  - Tests. Added coverage for finite timeout, bounded concurrency + drop, Close cancelling in-flight and dropping queued reports, full /v1/{prefix}/... path construction with escaping, and the enablement precedence matrix (server
  defaults/overrides can't enable; endpoint must be advertised).

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,246 @@
...
+		if r := recover(); r != nil {
+			d.log().Warn("iceberg: panic while reporting metrics to REST catalog", "recovered", r)
+		}
+	}()
+
+	// Derive from the dispatcher context so Close cancels in-flight requests,
+	// and add the per-report deadline so a stalled endpoint cannot pin a worker
+	// (and its connection) indefinitely. The deadline covers auth plus the full
+	// request and response cycle.
+	ctx, cancel := context.WithTimeout(d.ctx, d.timeout)
```
zeroshade: The deadline is real and it does bound the request and response cycle — but it does not bound authentication, which runs before the transport ever sees this context.

`sessionTransport.RoundTrip` calls `AuthHeader()` with no context (`rest.go:264-269`), `oauthAuthManager` calls `TokenSource.Token()` with no context (`auth.go:40-43`), and the built-in token source is constructed against `context.Background()` (`rest.go:880-903`) on a client with no `Timeout`. Nothing in that chain observes `ctx`.

With one worker, `timeout` at 30ms, and an `AuthManager` blocked inside `AuthHeader`, the worker was still pinned at 80ms; `close` returned only when its own timer expired, and the worker plus the `wg.Wait` goroutine survived until auth was released by hand. That is the same unbounded-hang shape as the previous round, just relocated from the metrics endpoint to the token endpoint — and a stalled token endpoint is at least as likely as a stalled metrics endpoint.

Worth noting that the comment on lines 160-161 asserts the deadline covers auth. It is the one thing it does not cover, so the comment will mislead the next reader.

Suggested fix: thread a context through auth retrieval so `AuthHeader`/`Token` honor the caller's deadline, including built-in OAuth refresh. Wrapping the context-free call in another goroutine only moves the leak somewhere harder to see. Failing that, a `Timeout` on the client used for token refresh would at least bound it.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,246 @@
...
+	case <-d.ctx.Done():
+		return
+	default:
+	}
+
+	select {
+	case d.jobs <- job:
+	case <-d.ctx.Done():
+	default:
+		d.log().Warn("iceberg: metrics report dropped; dispatch queue full")
```
zeroshade: This log call is synchronous, and it runs on the caller's goroutine — the scan or commit being observed. A slow or blocking `slog.Handler` therefore blocks `Report`, which contradicts the contract at `metrics/reporter.go:66-70` that reporting returns promptly. A full-queue probe with a deliberately blocking handler held `Report` until the handler was released.

This is my fault as much as anything: logging the drops was my suggestion last round, to keep back-pressure visible rather than silent. Putting that log on the synchronous submit path is what turned the safety valve into a blocking path. The goal stands, the placement needs to move.

There is a secondary concern even with a fast handler: one warning per drop means log amplification precisely during overload, when the queue is full and drops are arriving at scan/commit rate.

Suggested fix: keep `submit` to non-blocking operations only — bump an atomic drop counter, or do a non-blocking send on a small notification channel — and have a dispatcher-owned goroutine emit an aggregated, rate-limited warning ("N reports dropped in the last interval"). That preserves visibility, keeps the caller unblocked, and removes the amplification.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+}
+
+// close cancels in-flight reports and waits for the workers to return, bounded
+// by the report timeout so shutdown cannot hang on a stalled endpoint.
+func (d *metricsDispatcher) close() {
+	d.cancel()
+
+	done := make(chan struct{})
+	go func() {
```
zeroshade: `close` is bounded now, which was the main thing, and because `d.jobs` is never closed a concurrent `Report` cannot panic on a send to a closed channel. Cooperative concurrent `Report`/`close` and double-close stress came back race-clean.

Two residual issues:

Each call to `close` spawns another waiter goroutine here. If a worker is stuck in the uncancellable auth path described above, every repeated `close` leaves one more `wg.Wait` goroutine behind permanently. Making shutdown one-shot — a `sync.Once` plus a shared completion channel that later calls select on — makes repeated `Close` cheap and leak-free.

More importantly, the bound is on how long `close` *waits*, not on the work itself. When a worker is pinned in auth, `close` returns after `d.timeout` and reports success while the worker and its connection are still live. A caller that closes the catalog and expects outbound network activity to have stopped does not get that guarantee. Fixing the auth cancellation above is what makes this bound meaningful.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+var _ metrics.Reporter = (*restMetricsReporter)(nil)
+
+// Report wraps the report in a ReportMetricsRequest and hands it to the
+// catalog's dispatcher, returning immediately. Per the Reporter contract it
+// never blocks or fails the observed scan/commit: dispatch is asynchronous and
+// bounded, the send is detached from the caller's cancellation (the scan/commit
+// is already done), and any error is logged and swallowed by the dispatcher.
+func (rep *restMetricsReporter) Report(_ context.Context, report metrics.MetricsReport) {
```
zeroshade: Non-blocking, but a small regression against the previous revision. Discarding the caller's context drops its *values* as well as its cancellation; the earlier `context.WithoutCancel(ctx)` detached cancellation while preserving values, and the dispatcher now starts from `context.Background()` (`:108`, `:162`). A probe confirmed a value set by the caller no longer reaches the transport.

That matters for anything propagated contextually rather than through the request — tracing spans and request-scoped logging attributes being the usual cases. Detaching from cancellation is right; detaching from values is an unrelated side effect.

Suggested fix: keep `context.WithoutCancel(ctx)` on the queued job, and in the worker combine it with dispatcher cancellation and the per-report deadline.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,246 @@
...
+	metricsDispatchQueueSize = 128
+)
+
+// reportMetricsEnabled reports whether the client opted into REST metrics
+// reporting, accepting both the canonical and the legacy dotted key. It must be
+// given the client-supplied properties only (never the server-merged config) so
+// a server cannot flip the default and turn on outbound telemetry the client
+// never asked for.
+func reportMetricsEnabled(props iceberg.Properties) bool {
+	return props.GetBool(keyReportMetricsEnabled, false) ||
```
zeroshade: Non-blocking. Accepting both spellings with the canonical one first, and the alias documented as legacy, is exactly right.

One edge worth pinning down: because this is an `||`, a canonical `rest-metrics-reporting-enabled=false` combined with a legacy `rest.metrics-reporting-enabled=true` enables reporting. Someone migrating to the canonical key and explicitly setting it to `false` while an old dotted key lingers in their config would get telemetry they just tried to turn off.

Suggested fix: let the canonical key win when it is present at all, and fall back to the legacy key only when the canonical one is absent. A short comment plus a test for the conflicting case would settle it.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if _, err := doPost[metrics.ReportMetricsRequest, struct{}](
+		ctx, job.baseURI, job.path, job.req, job.cl, nil, allowNoContent()); err != nil {
+		// A report interrupted by Close (dispatcher context cancelled) is expected
+		// shutdown behavior, not a failure worth logging. A per-report timeout
+		// leaves the dispatcher context live, so genuine timeouts still surface.
+		if d.ctx.Err() != nil {
+			return
+		}
+		d.log().Warn("iceberg: failed to report metrics to REST catalog", "error", err)
```
zeroshade: Non-blocking. Distinguishing shutdown cancellation from genuine failures via `d.ctx.Err()` is a nice touch.

One caution: the raw transport error typically embeds the full request URL, which here carries the catalog host, any configured prefix, the namespace and the table name, plus any query parameters on the base URI. For deployments where table names or URI parameters are sensitive, that lands in operator logs on every failed report. Authorization headers are not logged, so this is limited to URL content.

Also, since a persistently unavailable endpoint fails one report per scan and per commit, this warning can become steady-state noise. Rate limiting or aggregating it — alongside the drop counter suggested on the `submit` path — would keep it useful.

catalog/rest/metrics_reporter_test.go
```diff
@@ -0,0 +1,487 @@
...
+	"testing"
+	"time"
+
+	iceberg "github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/metrics"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type capturedRequest struct {
```
zeroshade, line 43: Still open from last round: `capturedRequest` records method, decoded path, escaped path and body, but not headers, so nothing asserts that the metrics POST actually goes out authenticated.

That reuse of the catalog's already-authenticated client is one of the better properties of this design, and it is currently load-bearing but untested — a refactor that sent metrics through a bare `http.Client` would keep every existing test green.

Suggested fix: add `header http.Header` to `capturedRequest`, populate it in `captureTransport.RoundTrip` (clone it — the SDK may reuse the request), and assert `Authorization` is present, `Content-Type: application/json`, and any catalog-configured headers are forwarded. A case where the token refreshes between two reports, asserting the second carries the new credential, would cover the auth path this PR depends on and currently exercises nowhere.

catalog/rest/metrics_reporter_test.go
```diff
@@ -0,0 +1,487 @@
...
+// /v1/{prefix}/... URL and that namespace and table segments needing escaping
+// are percent-encoded.
+func TestRESTMetricsReporterEscapesFullPath(t *testing.T) {
+	received := make(chan capturedRequest, 1)
+	base, err := url.Parse("http://catalog.invalid/v1/my-prefix")
+	require.NoError(t, err)
+	rep := &restMetricsReporter{
+		baseURI:    base,
+		cl:         &http.Client{Transport: &captureTransport{ch: received}},
+		path:       []string{"namespaces", "a b", "tables", "t x", "metrics"},
```
zeroshade, line 254: This is better than the previous version — the prefix is now real and the escaping assertions on lines 256-257 are meaningful. Two things still keep it from guarding the fragile part.

`path` is supplied pre-built here, so the test never runs the production composition: `tableFromResponse`, `splitIdentForPath`, `encodeNamespace`, and `endpointReportMetrics.reqPath` (`rest.go:1140-1146`) are all bypassed. If that composition broke, this test would still pass. I checked the real path separately and it does produce the correct nested result, so this is about protecting a working behavior rather than a live bug.

The namespace is also single-level (`"a b"`). Multi-level namespaces are where separator handling and escaping interact, and that is the case most likely to regress.

Suggested fix: build the reporter through `tableFromResponse` (or at minimum call `endpointReportMetrics.reqPath`) with a configured prefix and a nested namespace such as `{"a b", "c/d"}`, then assert the complete escaped path. Worth coordinating with `9accf20` (#1614), which touched `encodeNamespace` and is part of the rebase.

catalog/rest/metrics_reporter_test.go
```diff
@@ -0,0 +1,487 @@
...
+		"a non-positive timeout falls back to the default")
+	assert.Equal(t, defaultReportMetricsTimeout, reportMetricsTimeout(iceberg.Properties{keyReportMetricsTimeoutMs: "bogus"}))
+}
+
+// TestMetricsReportingEnablementPrecedence pins that reporting is enabled only
+// by client-supplied configuration: server-vended defaults and overrides setting
+// the key must not turn it on, and the server must advertise the endpoint.
+// Table-response properties likewise cannot enable it, since enablement is
+// resolved once at init from the client properties alone.
+func TestMetricsReportingEnablementPrecedence(t *testing.T) {
```
zeroshade, line 748: This test covers the important cases well: default-off, client opt-in, server `defaults`, and server `overrides`. Given that server-controlled enablement was the security-relevant defect, having it pinned here is the most valuable test in the file.

The gap is that the doc comment on lines 407-408 asserts table-response properties cannot enable reporting, but no case exercises that. It is currently true by construction — enablement resolves once at init from client properties — and that structural argument is exactly what a future refactor could invalidate, silently, while every test here still passes.

Suggested fix: add a case that loads a table whose *table response* carries `rest-metrics-reporting-enabled=true` with the client not opting in, and assert no metrics traffic is produced. That converts the comment into an enforced invariant and completes the precedence matrix from the last round.

zeroshade (changes requested):
Thanks for the thorough revision — this is a substantial improvement, and the hardest item is genuinely fixed.

**What is now settled.** Client-only enablement is correct: the client properties are captured into `r.reporterProps` before server config is fetched and merged (`rest.go:914-919`), dispatcher construction consults only `reportMetricsEnabled(r.reporterProps)` (`rest.go:932-940`), server `defaults`/`overrides` merge on a separate path (`rest.go:1043-1058`), and table-response config reaches only the table-local `config` (`rest.go:1632-1638`) where it cannot create a dispatcher. A server can no longer flip this on. The dispatch bound is real too — four workers behind a 128-entry queue, non-blocking submission, drops when full — and a burst of 2,000 reports against a stalled transport held in-flight work at exactly four and logged every drop. `rest-metrics-reporting-enabled` is now canonical with the dotted spelling as a commented legacy alias, matching Iceberg Java's `RESTCatalogProperties` and the published catalog-properties docs. The REST wire format is unchanged and still conforms to the OpenAPI definition. `-race -count=3` is clean, `golangci-lint` reports 0 issues, and the timeout and concurrency tests are genuine — they fail when their fixes are mutated away, with no flakiness across ten runs.

Two blocking items remain, both in the same failure-isolation area, and I want to be clear that neither is visible from the test suite as written.

## The report deadline still does not cover authentication

`context.WithTimeout` at `metrics_reporter.go:162` bounds the request and response cycle, but not the auth step that runs in front of it. `sessionTransport.RoundTrip` calls the context-free `AuthHeader()` (`rest.go:264-269`), the OAuth manager calls `TokenSource.Token()` with no context (`auth.go:40-43`), and the built-in token source is rooted in `context.Background()` (`rest.go:880-903`) on a client with no `Timeout`. With one worker, a 30ms report timeout and an `AuthManager` blocked inside `AuthHeader`, the worker was still pinned after 80ms, `close` returned only because its own shutdown timer expired, and both the worker and the `wg.Wait` goroutine stayed alive until auth was released manually.

So the original process-lifetime hang survives for a stalled or black-holing *token* endpoint, and `Close` cannot reclaim catalog-owned work. The comment at `metrics_reporter.go:160-161` states the deadline covers auth, which is the one part it does not cover.

## Drop logging can block the caller it is meant to protect

`submit` calls `d.log().Warn(...)` synchronously on the queue-full path (`metrics_reporter.go:191`). That executes on the scan/commit goroutine, so a slow or blocking `slog.Handler` blocks `Report` — contradicting the prompt-return requirement at `metrics/reporter.go:66-70`. A full-queue probe with a blocking handler kept `Report` blocked until the handler was released. This one is new in this revision: making back-pressure visible was my suggestion, and putting the log on the synchronous path is what turned it into a blocking path. Sorry for the loop.

## Tests

The suite is much stronger, and the two tests doing the most work are honest ones. The systematic gap is that every timeout and shutdown transport *cooperates* with request cancellation, which is exactly why the auth hang above passes CI. The most valuable additions now would be a report blocked in auth rather than in transport, an actual `Catalog.Close` with work in flight, and headers on `capturedRequest` so the reuse of the authenticated client is pinned rather than assumed.

One housekeeping item: the PR currently conflicts with `main`. Please rebase — and note the conflict is not purely textual. `7efed97 fix(rest): close internally owned transports (#1606)` reworked catalog close and transport ownership, which is the same lifecycle this PR extends with dispatcher shutdown, so the two need reconciling deliberately. `9accf20 fix(rest): reject NUL in namespace identifiers (#1614)` also touched `encodeNamespace`, which is relevant to the path-construction comment below.

Details and suggested fixes are inline. The wire-level work and the enablement fix are solid; what is left is about what happens when the metrics or token endpoint hangs rather than fails.

tanmayrauth:
Thanks @zeroshade  for the review.

Auth deadline. Since TokenSource.Token() takes no context, I bounded auth two ways rather than threading the per-report deadline through it (which would need either a cache-losing reconstruction or the goroutine wrapper you   flagged): an optional ContextAuthManager that sessionTransport prefers, and a Timeout on the built-in OAuth refresh client. A stalled token endpoint now fails within that timeout instead of hanging. One residual, documented   in-code: a refresh in flight is bounded by the OAuth timeout, not the shorter per-report deadline, so Close can return before that worker's auth attempt finishes. It's bounded, not infinite. Happy to make the timeout configurable  if you'd prefer.

 Drop logging. Off the caller's goroutine now — submit bumps an atomic counter, and a dispatcher-owned goroutine emits a single aggregated, rate-limited warning for drops and failures. Also Close is one-shot (sync.Once + shared channel); Report queues context.WithoutCancel(ctx) to preserve values; the canonical enablement key wins when present; and failed-report detail (which carries the URL) is now at Debug,  with only aggregate counts at Warn. 
 
 Tests added for auth forwarding, per-report credential re-fetch, the ContextAuthManager preference, context-value propagation, non-blocking drops, the aggregated warning, idempotent Close, and the real path composition with a   multi-level namespace. -race -count=3 clean, lint 0.
  
The "table-response cannot enable" case is currently pinned by a unit-level invariant plus the existing precedence test; let me know if you'd like the full LoadTable version.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,333 @@
...
+// reporting, accepting both the canonical and the legacy dotted key. It must be
+// given the client-supplied properties only (never the server-merged config) so
+// a server cannot flip the default and turn on outbound telemetry the client
+// never asked for.
+//
+// The canonical key wins whenever it is present, so a client that has migrated
+// to it and explicitly set it to false is honored even if a stale legacy dotted
+// key lingers in the config. The legacy key is consulted only when the canonical
+// key is absent.
+func reportMetricsEnabled(props iceberg.Properties) bool {
```
laskoviymishka, line 78: I'm on board with defaulting this off — the no-surprise-outbound-traffic instinct is the right one, and the comment above makes the intent clear.

The thing I'd flag is that Java defaults `METRICS_REPORTING_ENABLED_DEFAULT` to true, so anyone porting a Java client that never set the flag explicitly will silently stop getting metrics here. I'd call that out in CHANGES.md or the catalog property docs so the divergence reads as a documented choice rather than a surprise. wdyt?

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,333 @@
...
+// pool of workers draining a bounded queue. It is owned by the catalog and
+// shared across that catalog's table reporters, so concurrent report volume
+// stays bounded no matter how many tables are loaded or how often they are
+// scanned. A stalled endpoint sheds load — reports are dropped and counted —
+// rather than accumulating goroutines and connections. Close cancels in-flight
+// reports and drains the workers.
+type metricsDispatcher struct {
+	jobs      chan metricsJob
+	timeout   time.Duration
+	ctx       context.Context
```
laskoviymishka, line 126: Storing `ctx`/`cancel` on the struct will draw a lint eye since the Go docs steer away from contexts in struct fields. I think it's the right call here — the dispatcher genuinely owns a lifecycle context that Close cancels — but I'd add a one-line comment saying exactly that, so the next reader doesn't try to "fix" it.

catalog/rest/metrics_reporter.go
```diff
@@ -0,0 +1,333 @@
...
+
+		done := make(chan struct{})
+		go func() {
+			d.wg.Wait()
+			close(done)
+		}()
+
+		select {
+		case <-done:
+		case <-time.After(d.timeout):
```
laskoviymishka: `time.After(d.timeout)` leaves a live timer around until it fires whenever the `<-done` arm wins, which is the common path — with the 10s default that's a 10s-lived timer per catalog Close, and it adds up in a process that cycles through many catalogs. I'd switch to `time.NewTimer` with a `defer Stop`:

```go
t := time.NewTimer(d.timeout)
defer t.Stop()
select {
case <-done:
case <-t.C:
}
```

While we're here: on the timeout arm this returns while the `wg.Wait()` goroutine (and maybe a worker) is still alive. It self-exits once `d.cancel()` takes effect so it's not a leak, but it does mean Close can return before workers have fully stopped — worth a one-line comment so nobody later tears down a shared transport assuming the workers are done.

catalog/rest/metrics_reporter_test.go
```diff
@@ -0,0 +1,844 @@
...
+
+	assert.LessOrEqual(t, tr.maxSeen.Load(), int32(workers), "concurrency exceeded the worker pool")
+}
+
+// TestMetricsDispatcherCloseCancelsInFlight proves Close cancels an in-flight
+// report and returns promptly rather than waiting on the stalled endpoint.
+func TestMetricsDispatcherCloseCancelsInFlight(t *testing.T) {
+	tr := &ctxBlockTransport{started: make(chan struct{}, 1), ctxErr: make(chan error, 1)}
+	// A long per-report timeout so it is Close, not the deadline, that unblocks.
+	d := newMetricsDispatcher(1, 1, time.Minute, slog.New(slog.NewTextHandler(io.Discard, nil)))
```
laskoviymishka, line 651: This test and `TestMetricsDispatcherCloseDropsQueuedReports` construct the dispatcher directly without registering `t.Cleanup(d.close)` the way `newTestDispatcher` does. Both use a 1-minute timeout, so if an early `t.Fatal` fires before the close runs — the "report never reached the transport" path, say — the workers keep running and under `-race`/goleak that shows up as a flake in some later test.

I'd add `t.Cleanup(d.close)` right after each `newMetricsDispatcher` call here and keep the explicit close too.

catalog/rest/metrics_reporter_test.go
```diff
@@ -0,0 +1,844 @@
...
+			wantEnabled: true,
+		},
+		{
+			name:        "client legacy key enables",
+			serverCfg:   cfg(nil, nil, nil),
+			clientProps: iceberg.Properties{keyReportMetricsEnabledLegacy: "true"},
+			wantEnabled: true,
+		},
+		{
+			name:        "client enables but endpoint not advertised",
```
laskoviymishka, line 792: This proves the negative — endpoints advertised but not the metrics one — but there's no case advertising a non-empty list that *does* include the metrics endpoint. The "client enables" cases above only hit the fallback-to-`defaultEndpoints` path (server returns no endpoints key), so the production path where `resolveEndpoints` parses an explicitly-advertised metrics endpoint template never runs.

I'd add a sibling entry with `endpoints: []string{"POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/metrics"}` and `wantEnabled: true`. That's the case that catches `endpointFromString` silently failing to parse the template and disabling reporting for servers that actually advertise it.

catalog/rest/rest.go
```diff
@@ -929,6 +955,16 @@ func (r *Catalog) init(ctx context.Context, ops *options, uri string) error {
 	}
 	r.props = toProps(ops)
 
+	// Stand up the metrics dispatcher only when the client opted in and the
+	// server advertises the endpoint. Enablement is read from the
+	// client-supplied properties alone (see reporterProps): a server-vended
+	// default, override, or table-response property must never turn on outbound
+	// telemetry the client never asked for.
+	if reportMetricsEnabled(r.reporterProps) && r.endpoints.check(endpointReportMetrics) == nil {
```
laskoviymishka: Reading enablement from `r.reporterProps` is exactly right for the direction that matters — a server must not be able to turn *on* telemetry the client never asked for. No argument there.

The asymmetry I'd think about is the other direction: because this only ever looks at client props, a server that sets `rest-metrics-reporting-enabled=false` in its overrides can't turn reporting *off* for a Go client the way it can for a Java one (Java resolves this from the merged config). Server-side disable is always safe — it never creates surprise outbound traffic — so honoring an explicit `=false` from overrides seems like the friendlier behavior for an operator trying to suppress all clients at once.

Either honor a server override that flips it off, or leave it client-only and document that server-side disable is intentionally ignored. I'd lean toward the former, but either's defensible as long as it's a deliberate call. wdyt?

catalog/rest/rest.go
```diff
@@ -1143,6 +1183,21 @@ func (r *Catalog) tableFromResponse(_ context.Context, identifier []string, meta
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize metrics reporter: %w", err)
 	}
+	// Opt-in: POST scan/commit reports to the catalog's metrics endpoint via the
+	// catalog-owned dispatcher. It is non-nil only when the client enabled
+	// reporting and the server advertises the endpoint (see init).
+	if r.metricsDispatcher != nil {
+		if ns, tbl, idErr := r.splitIdentForPath(identifier); idErr == nil {
```
laskoviymishka: Both `splitIdentForPath` and `reqPath` errors get swallowed here — if either fails, the REST reporter just isn't composed in, with no log. It's unreachable for a valid table today, but if it ever regresses, metrics silently vanishing with no signal is a rough thing to debug.

I'd drop a debug-level `slog` on the two error branches, matching the warn-on-drop pattern the dispatcher already uses. Even a comment on why these can't fail would help the next reader.

laskoviymishka (changes requested):
The wire-level side of this is genuinely well done — the flat report payload with `report-type` as a sibling of the flattened fields, the prefix-aware path building and escaping, reusing the authenticated client, gating on the endpoint capability check, and composing through `metrics.Combine` are all the things that are easy to get subtly wrong, and they're right here. The bounded dispatcher with drop-on-full, per-report deadlines, and a Close that cancels in-flight work is a solid shape.

I'd hold this before merging though, mostly to settle one design call while it's still cheap to change.

Enablement is read from client props only, which is the right guard for the direction that matters — a server can't switch on telemetry the client never asked for. But it also means the default diverges from Java (opt-in here vs opt-out there), and a server that sets the flag `=false` in its overrides can't suppress a Go client the way it can a Java one. I don't think either is wrong, but they're the kind of cross-client surprise I'd want to be a deliberate, documented decision rather than something an operator hits in production. My lean is to honor a server-side `=false` (disabling is always safe) and to document the opt-in default, but I'm open.

- Make the opt-in-vs-Java's-opt-out default a documented choice (CHANGES.md or the catalog property docs), and decide whether a server override setting the flag off should be honored.
- Add the positive test case where the server advertises a non-empty endpoint list that includes the metrics endpoint — right now only the fallback-to-defaults path is covered, so a parse regression on the endpoint template would silently disable reporting with no test catching it.
- Register `t.Cleanup(d.close)` on the two dispatcher tests that construct it directly, and tidy the `time.After` timer in `close()`.

One more small thing: worth a line in the Go-specific docs that `rest-metrics-reporting-timeout-ms` is a Go extension with no Java/PyIceberg/Rust equivalent, so multi-client deployment notes don't assume parity on it.

None of this is a correctness problem — the behavior looks right and the tests are thorough. Once the parity call is made and those couple of test/lifecycle items land, I'm happy to take another pass and approve.

tanmayrauth:
Thanks @laskoviymishka  for the review.
  
I went with your lean and now honor a server-side disable. Enablement requires both a client opt-in and that the merged config doesn't resolve the key to false, so a server override setting rest-metrics-reporting-enabled=false suppresses a client that opted in — same way a Java client resolves it, so an operator can switch reporting off fleet-wide. A server default still can't flip off a client (client props win over defaults in the merge), and the direction that actually matters is unchanged: a server can never turn reporting on.

 For the parity divergence, I added a Metrics reporting section to website/src/configuration.md — there's no  CHANGES.md in the repo. It documents the three properties, calls out that we default off where Java defaults on
  (METRICS_REPORTING_ENABLED_DEFAULT=true), and notes that rest-metrics-reporting-timeout-ms is a Go-only extension with no Java/PyIceberg/Rust equivalent.

 On the smaller items: added the positive test case that advertises the metrics endpoint explicitly, so the  resolveEndpoints/endpointFromString template parse gets exercised instead of just the fallback path — plus server-override-false and server-default-false cases for the new behavior. close() now uses time.NewTimer with a deferred Stop instead of time.After, with a note that the timeout arm can return before the workers have fully stopped so nobody tears down a shared transport assuming otherwise. Registered t.Cleanup(d.close) on the two directly-constructed dispatcher tests, added the one-liner on the dispatcher's ctx/cancel fields explaining the deliberate lifecycle-context exception, and put debug logging on the two swallowed splitIdentForPath/reqPath branches so a regression surfaces instead of metrics quietly disappearing.

laskoviymishka (approved):
Thanks! This looking good to me now!

tanmayrauth:
@zeroshade  what are you thoughts?

catalog/rest/rest.go
```diff
@@ -880,11 +905,12 @@ func setupOAuthManager(r *Catalog, cl *http.Client, opts *options) (AuthManager,
...
-	// dedicated HTTP client for token requests instead of reusing the catalog
-	// client. This is needed when the OAuth2 server is a different host with
-	// different TLS requirements.
-	oauthClient := cl
+	// Token refresh uses a client with a Timeout so a stalled token endpoint
+	// cannot block auth forever (oauth2's Token() takes no context). By default
+	// this reuses the catalog client's transport — preserving its TLS, proxy and
+	// header behavior — but as a distinct *http.Client so the Timeout applies to
+	// refresh alone and not to ordinary catalog requests.
+	oauthClient := &http.Client{Transport: cl.Transport, Timeout: defaultOAuthTimeout}
```
zeroshade, line 913: Non-blocking: this 60s refresh timeout now applies to every REST catalog user, not just metrics opt-ins. It's the right fix (oauth2's `Token()` takes no context), but since it's a behavior change landing inside a metrics PR, please add a line to the PR description calling it out so it surfaces in release notes.

zeroshade (approved):
All of my previous concerns are addressed in this revision — the bounded dispatcher with catalog-owned lifecycle, the per-report deadline with auth bounded via the refresh-client timeout, canonical-key precedence, server-side disable (but not enable), and the auth-header and production-path test coverage that had stayed open across rounds. The enablement-precedence and table-response-property tests pin the security-relevant behavior nicely.

One non-blocking note inline about the new 60s OAuth refresh timeout being a repo-wide behavior change worth a mention in the description. Nice work seeing this through — this is a solid piece of infrastructure.

---

catalog/rest/rest.go
tanmayrauth: Thanks @zeroshade! Added a Behavior change note to the PR description so it lands in release notes.

PR #1657, opened 2026-08-05
fix(table): reject tags as transaction targets

fallintoplace:
## Summary

Reject existing tags as transaction targets.

## Why

NewTransactionOnBranch accepted a tag name as a write target. A later commit could then treat the tag like a branch and replace it.

## What changed

NewTransactionOnBranchWithError checks the existing ref type before creating the transaction. Existing tags return ErrInvalidArgument, while branches and missing refs keep their current behavior.

## Tests

- go test ./table

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -183,6 +183,12 @@ func (t Table) newBrokenTransaction(branch string, err error) *Transaction {
 // callers to receive the precise initialization error instead of hitting
 // panic/undefined behavior later.
 func (t Table) NewTransactionOnBranchWithError(branch string) (*Transaction, error) {
+	for name, ref := range t.metadata.Refs() {
+		if name == branch && ref.SnapshotRefType != BranchRef {
+			return nil, fmt.Errorf("%w: %s is not a branch", iceberg.ErrInvalidArgument, branch)
+		}
+	}
+
 	meta, err := MetadataBuilderFromBase(t.metadata, t.metadataLocation)
 	if err != nil {
 		return nil, err
--- a/table/table.go
+++ b/table/table.go
@@ -184,9 +184,15 @@ func (t Table) newBrokenTransaction(branch string, err error) *Transaction {
 // panic/undefined behavior later.
 func (t Table) NewTransactionOnBranchWithError(branch string) (*Transaction, error) {
 	for name, ref := range t.metadata.Refs() {
-		if name == branch && ref.SnapshotRefType != BranchRef {
-			return nil, fmt.Errorf("%w: %s is not a branch", iceberg.ErrInvalidArgument, branch)
+		if name != branch {
+			continue
 		}
+		if ref.SnapshotRefType != BranchRef {
+			return nil, fmt.Errorf("%w: ref %q is a %s; tags cannot be transaction targets",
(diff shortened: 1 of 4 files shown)


table/table.go
```diff
@@ -183,6 +183,12 @@ func (t Table) newBrokenTransaction(branch string, err error) *Transaction {
 // callers to receive the precise initialization error instead of hitting
 // panic/undefined behavior later.
 func (t Table) NewTransactionOnBranchWithError(branch string) (*Transaction, error) {
+	for name, ref := range t.metadata.Refs() {
+		if name == branch && ref.SnapshotRefType != BranchRef {
```
zeroshade: This type check runs only when the transaction is created. If the ref changes from a branch to a tag at the same snapshot ID before commit, `AssertRefSnapshotID` still succeeds because it checks only the ID, and the set-ref update replaces the tag with a branch. #1659's UUID check does not help because the table UUID is unchanged.

Suggested fix: validate that the target is still branch-typed on the atomic commit/update path and after retry refreshes, then add a same-ID branch→tag race test.

table/table.go
```diff
@@ -183,6 +183,12 @@ func (t Table) newBrokenTransaction(branch string, err error) *Transaction {
 // callers to receive the precise initialization error instead of hitting
 // panic/undefined behavior later.
 func (t Table) NewTransactionOnBranchWithError(branch string) (*Transaction, error) {
+	for name, ref := range t.metadata.Refs() {
```
zeroshade, line 186: This loop continues scanning all refs even after it finds a matching branch.

Suggested fix: stop once the matching ref is found, or centralize this in a typed ref lookup helper.

table/table.go
```diff
@@ -183,6 +183,12 @@ func (t Table) newBrokenTransaction(branch string, err error) *Transaction {
...
+		if name == branch && ref.SnapshotRefType != BranchRef {
+			return nil, fmt.Errorf("%w: %s is not a branch", iceberg.ErrInvalidArgument, branch)
```
zeroshade: The error identifies the name but does not quote it or explain that tags cannot be transaction targets.

Suggested fix: quote the ref name and state explicitly that a tag cannot be used as a transaction target while continuing to wrap `iceberg.ErrInvalidArgument`.

zeroshade (changes requested):
Thanks for tightening transaction construction — all public constructor paths now reject a tag that already exists, valid branches and missing refs remain accepted, and the error preserves `iceberg.ErrInvalidArgument`. That initial validation is sound.

The remaining blocker is the same-snapshot branch→tag race described inline. A transaction can be constructed against a branch, the ref can become a tag at the same snapshot ID, and the commit can still replace it with a branch because `AssertRefSnapshotID` validates only the ID.

The UUID guard proposed in #1659 does **not** protect this path: changing a ref from branch to tag within the same table leaves the table UUID unchanged. Ref-type validation must therefore run against refreshed/commit-time metadata, not only at transaction construction.

Constructor coverage is otherwise complete. Please add a race test that changes the target from branch to tag without changing its snapshot ID before commit. The legacy-constructor test should also exercise a public operation on the broken transaction rather than checking only the private `initErr`.

zeroshade (approved):

PR #1658, opened 2026-08-05
fix(table): keep scan ref selectors exclusive

fallintoplace:
## Summary

Keep scan reference selectors mutually exclusive.

## Why

UseRef could be applied after an as-of timestamp had been selected. The scan then contained both selectors, and the ref snapshot took precedence over the historical timestamp.

## What changed

UseRef treats main as a no-op and returns ErrInvalidArgument when a non-main ref is selected after a snapshot ID or as-of timestamp. Snapshot ID and as-of timestamp options also record a selector conflict instead of silently replacing one another.

## Tests

- go test ./table

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -239,6 +239,7 @@ type Scan struct {
 	caseSensitive  bool
 	snapshotID     *int64
 	asOfTimestamp  *int64
+	selectorErr    error
 	options        iceberg.Properties
 	limit          int64
 
@@ -266,11 +267,18 @@ func (scan *Scan) Reporter() metrics.Reporter {
 	return scan.reporter
 }
 
+// UseRef selects a named snapshot reference. UseRef(MainBranch) is an
+// intentional no-op that returns a clone, so it remains valid even when the
+// scan already has an explicit snapshot or as-of selector. Any conflicting
+// selectors recorded by scan options are still surfaced by scan execution.
 func (scan *Scan) UseRef(name string) (*Scan, error) {
 	if name == MainBranch {
 		out := *scan
 
 		return &out, nil
+	}
+	if scan.selectorErr != nil {
+		return nil, scan.selectorErr
 	}
 
 	if scan.snapshotID != nil {
@@ -296,6 +304,10 @@ func (scan *Scan) UseRef(name string) (*Scan, error) {
 // the table's current snapshot; explicit snapshot IDs and as-of timestamps
 // must resolve to an existing snapshot.
 func (scan *Scan) ResolveSnapshot() (*Snapshot, error) {
+	if scan.selectorErr != nil {
+		return nil, scan.selectorErr
+	}
+
 	if scan.snapshotID != nil {
 		snap := scan.metadata.SnapshotByID(*scan.snapshotID)
 		if snap == nil {
@@ -332,6 +344,10 @@ func (scan *Scan) Snapshot() *Snapshot {
 }
 
 func (scan *Scan) Projection() (*iceberg.Schema, error) {
+	if scan.selectorErr != nil {
+		return nil, scan.selectorErr
+	}
+
 	curSchema, err := scan.effectiveSchema()
 	if err != nil {
 		return nil, err
@@ -386,6 +402,10 @@ func (scan *Scan) Projection() (*iceberg.Schema, error) {
 }
 
 func (scan *Scan) effectiveSchema() (*iceberg.Schema, error) {
+	if scan.selectorErr != nil {
+		return nil, scan.selectorErr
+	}
+
 	curSchema := scan.metadata.CurrentSchema()
 	if scan.snapshotID == nil && scan.asOfTimestamp == nil {
 		// Live scans intentionally use the table's current schema. A schema-only
@@ -817,6 +837,10 @@ func (scan *Scan) collectManifestEntriesWithSchema(
 // scan's reporter on success; remote (server-side) planning reports its own
 // metrics and does not emit here.
 func (scan *Scan) PlanFiles(ctx context.Context) ([]FileScanTask, error) {
+	if scan.selectorErr != nil {
+		return nil, scan.selectorErr
+	}
+
 	if scan.asOfTimestamp != nil {
 		snapshot, err := scan.ResolveSnapshot()
 		if err != nil {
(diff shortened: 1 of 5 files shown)


table/scanner.go
```diff
@@ -267,10 +267,20 @@ func (scan *Scan) Reporter() metrics.Reporter {
 }
 
 func (scan *Scan) UseRef(name string) (*Scan, error) {
+	if name == MainBranch {
```
zeroshade, line 275: This deliberately treats `UseRef("main")` as a clone/no-op even after another selector. That matches Java, but it is an exception to the exclusivity rule and is not obvious from the API.

Suggested fix: document the `main` exception on `UseRef` so callers do not infer that all refs conflict with an existing snapshot selector.

zeroshade (changes requested):
Thanks for closing the as-of→non-main-ref hole — `UseRef` now rejects that conflict immediately, preserves the immutable original scan, and still allows live branch/tag selection. That part is sound.

## Snapshot ID and as-of remain last-wins

`table/table.go:967-976` still makes `WithSnapshotID` clear `asOfTimestamp` and `WithSnapshotAsOf` clear `snapshotID`. Option order therefore silently changes which snapshot is read, so the full selector set is not mutually exclusive. The test at `table/time_travel_test.go:250-259` explicitly preserves this last-wins behavior.

Suggested fix: record and surface a selector-conflict error (or add fallible selector methods) and reject snapshot-ID/as-of conflicts in both orders rather than discarding the earlier selector.

The current pair behavior is inconsistent:

- Snapshot ID → non-main ref: immediate error.
- As-of timestamp → non-main ref: immediate error.
- Non-main ref → another non-main ref: immediate error because the first ref resolves to a snapshot ID.
- Snapshot ID ↔ as-of timestamp: silent last-wins in both orders.
- Any selector → `main`: no-op clone preserving the earlier selector.
- Live scan → existing branch/tag: accepted.
- Live scan → unknown non-main ref: immediate `ErrInvalidArgument`.

Please add both snapshot-ID/as-of orderings, snapshot-ID→ref, second-ref branch/tag combinations, exact error context, preservation of the original scan after failure, and unknown-ref behavior when another selector is active.

zeroshade:
Re-reviewed after `a46d30d`. This addresses the main ask from my last pass: `WithSnapshotID` and `WithSnapshotAsOf` no longer discard each other, and the conflict is reported in both orderings. Given that `ScanOption` is `func(*Scan)` with no error return, recording a deferred `selectorErr` and surfacing it at the entry points is the right shape — it keeps `Table.Scan`'s signature intact while making the conflict unignorable. The `main` exception on `UseRef` is documented now too.

I checked the guard coverage rather than assuming it. Every exported `Scan` method that can produce data routes through a guarded call: `ToArrowTable` → `ToArrowRecords` → `PlanFiles` and `ReadTasks`, and `Snapshot()` → `ResolveSnapshot`. There is no unguarded reader, and nothing outside `table/table.go` reads `snapshotID` or `asOfTimestamp` directly.

**One thing I would like a test for before this merges.**

Four of the six `selectorErr` guards can be deleted with `go test ./table` still passing: the ones in `PlanFiles`, `ReadTasks`, `effectiveSchema`, and `UseRef`. Only the `ResolveSnapshot` and `Projection` guards are exercised by the new tests.

For local planning that is harmless — `ResolveSnapshot` is the chokepoint, and with only its guard present every local path still errors and the suite still passes. `effectiveSchema` reaches it at `table/scanner.go:412`, and `PlanFiles` reaches it through `effectiveSchema`.

But `planFilesRemote` (`table/scanner.go:974-995`) never calls `ResolveSnapshot`. It passes `SnapshotID: scan.snapshotID` straight to the planner. `PlanFiles` does resolve as-of into a snapshot ID before the planning-mode switch, so the as-of-first ordering is covered by that block — but the snapshot-ID-first ordering skips it entirely and goes directly to the remote planner.

I confirmed this with a `fakeScanPlanner` in `ScanPlanningRemote` mode, using `WithSnapshotID(1000)` followed by `WithSnapshotAsOf(ts)`:

```
PlanFiles guard removed:  PlanFiles err = <nil>
                          -> dispatched to planner with SnapshotID=1000, conflict silently ignored

PR as written:            PlanFiles err = invalid argument: cannot select as-of timestamp ...
                          when snapshot ID 1000 is already selected
```

So that guard is the only thing standing between a conflicting scan and a remote plan against the first selector — which is precisely the silent-last-wins behavior this PR exists to remove, just relocated to the remote path. It deserves a test. The harness already exists: `fakeScanPlanner` and `fakePlanIO` in `table/scan_planning_test.go:144-163`, used the same way as `TestScanPlanningRemoteStoresPlanIO`.

**Smaller item.** `Transaction.Scan` (`table/transaction.go:2304-2334`) applies the same `ScanOption`s and already returns `(*Scan, error)`, but does not surface `selectorErr`. `Table.Scan` cannot, which is the whole reason for the deferred design — but the transaction variant can, and reporting the conflict at construction there would be strictly better than deferring it:

```go
for _, opt := range opts {
    opt(s)
}
if s.selectorErr != nil {
    return nil, s.selectorErr
}
```

**Still open from my previous list**, all lower priority than the above: non-main ref → another non-main ref, unknown-ref behavior while another selector is active (today that reports the selector conflict rather than the unknown ref, which I think is the better answer but is untested), and any assertion on error content — the current tests only check `errors.Is` against the sentinel.

**On the merge conflict.** It is mechanical. `table/time_travel_test.go` is the only conflicted file; both sides appended test functions to the same region and share a trailing brace, so the resolution is to keep both and close `TestScanUseRefKeepsSnapshotSelectorsExclusive` before `TestSnapshotAsOfWithOutOfOrderSnapshotLog`. I did that locally and the merged tree builds, is gofmt-clean, and passes `go test ./table`. `table/scanner.go` and `table/table.go` auto-merge, and I checked that main's change to `scanner.go` lands inside `ResolveSnapshot`'s as-of branch, orthogonal to the guard added at the top of that function.

CI is green on the current head (15/15).

Happy to approve once the remote-planning case has a test and `Transaction.Scan` surfaces the error — you will need to push for the conflict anyway.

zeroshade (approved):
This resolves everything from my last round: snapshot-ID and as-of conflicts are now rejected in both orders via the recorded `selectorErr` — surfaced at every fallible consumption point, including before remote planning is invoked — as-of→ref is symmetric with snapshot-id→ref, and the `UseRef(main)` exception is documented and pinned (including that a conflicted scan's main-clone keeps its error). The selector set is genuinely mutually exclusive now, with the one documented exception. Nice work.

---

PR #1659, opened 2026-08-05
fix(table): reject refreshes with changed table UUID

fallintoplace:
## Summary

Prevent a table handle from switching to a replacement table during refresh.

## Why

If a table was dropped and recreated under the same identifier, refresh could silently replace the metadata for the original table.

## What changed

Refresh compares the current and freshly loaded table UUIDs before changing state. A mismatch returns ErrInvalidMetadata and leaves metadata, location, planner, filesystem, and reporter state unchanged.

## Tests

- go test ./table

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -202,6 +202,12 @@ func (t *Table) Refresh(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	if t.metadata != nil && fresh.metadata != nil {
+		if expected, actual := t.metadata.TableUUID(), fresh.metadata.TableUUID(); expected != actual {
+			return fmt.Errorf("%w: table UUID changed during refresh: expected %s, got %s",
+				ErrInvalidMetadata, expected, actual)
+		}
+	}
 
 	t.metadata = fresh.metadata
 	t.fsF = fresh.fsF
--- a/table/table.go
+++ b/table/table.go
@@ -41,6 +41,7 @@ import (
 	icebergio "github.com/apache/iceberg-go/io"
 	"github.com/apache/iceberg-go/metrics"
 	tblutils "github.com/apache/iceberg-go/table/internal"
+	"github.com/google/uuid"
 	"github.com/klauspost/compress/zstd"
 	"golang.org/x/sync/errgroup"
 )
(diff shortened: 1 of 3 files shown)


table/table.go
```diff
@@ -202,6 +202,12 @@ func (t *Table) Refresh(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	if t.metadata != nil && fresh.metadata != nil {
+		if expected, actual := t.metadata.TableUUID(), fresh.metadata.TableUUID(); expected != actual {
```
zeroshade: An absent v1 `table-uuid` decodes as `uuid.Nil`, but this comparison treats that sentinel as a real identity. Valid transitions between UUID-less v1 metadata and metadata with a UUID are therefore rejected. The repository's valid UUID-less fixture begins at `table/metadata_internal_test.go:864`, and the value field that produces `uuid.Nil` is `table/metadata.go:1634`.

Suggested fix: compare only when both UUIDs are known: `expected != uuid.Nil && actual != uuid.Nil && expected != actual`, and test absent→present, present→absent, and both-absent refreshes.

table/table.go
```diff
@@ -202,6 +202,12 @@ func (t *Table) Refresh(ctx context.Context) error {
...
+			return fmt.Errorf("%w: table UUID changed during refresh: expected %s, got %s",
```
zeroshade: This error includes both UUIDs and wraps `ErrInvalidMetadata`, but it does not tell the caller how to recover.

Suggested fix: say that the table was likely replaced and instruct the caller to load a new table handle; including the table identifier would make the diagnosis more actionable.

zeroshade (changes requested):
Thanks for adding table-identity validation — equal known UUIDs continue to refresh, unequal known UUIDs are rejected, and the tests verify that table state is preserved after rejection. That core guard is useful.

## Commit retry bypasses the guard

`table/table.go:558-570` has a separate refresh-and-replay path: it calls `LoadTable`, assigns `fresh.metadata` to `current`, rewrites requirements, and rebuilds manifests without applying the UUID check. A retry can therefore run validators and rebuild manifests against a dropped-and-recreated table before the catalog eventually rejects the commit.

Suggested fix: factor UUID validation into a shared helper and call it immediately after `LoadTable` here, before consuming refreshed metadata for requirements, validation, or manifest rebuilding. Add a retry test that changes UUID and proves the path stops before validation/rebuild work.

## UUID validation does not protect ref type

The UUID guard also does not close #1657's same-table branch→tag race. At this head, transaction construction does not reject tags (`table/table.go:185-197`), `AssertRefSnapshotID` checks only the snapshot ID (`table/requirements.go:173-199`), and the commit emits/applies a `BranchRef` replacement (`table/snapshot_producers.go:1374-1376`, `table/updates.go:390-407`). A branch can become a tag at the same snapshot ID while the table UUID remains unchanged, and the transaction can convert it back into a branch.

Suggested fix: validate ref type against refreshed/commit-time metadata independently of table UUID, including the same-ID branch→tag race.

Also cover all UUID-nullability combinations (absent→present, present→absent, both absent), `t.metadata == nil`, and error text that tells callers to load a new table handle.

zeroshade (approved):

PR #1661, opened 2026-08-05
fix(schema): validate schema JSON shape

fallintoplace:
## Summary

Validate the top-level shape of schema JSON before decoding it.

## Why

encoding/json ignores unknown top-level properties, so a list-type object or a schema without fields could be accepted as an empty table schema.

## What changed

Schema decoding now requires type=struct and a non-null fields array. The envelope and fields are decoded into temporary state, so the receiver changes only after validation succeeds. Existing duplicate-field validation is unchanged.

## Tests

- go test .


schema_test.go
```diff
@@ -570,6 +570,27 @@ func TestUnmarshalSchema(t *testing.T) {
 	assert.True(t, tableSchemaSimple.Equals(&schema))
 }
 
+func TestUnmarshalSchemaRejectsInvalidTopLevelShape(t *testing.T) {
```
zeroshade, line 573: Non-blocking: Consider adding a positive legacy-v1 case omitting both `schema-id` and `identifier-field-ids`, plus `type: null`, non-string type, scalar/malformed `fields`, an assertion that nil fields marshal as `[]`, and receiver preservation after a field-decoding failure. Nested struct/list/map missing-property cases would also pin the follow-up behavior described in the review body.

zeroshade (commented):
This change is correct. I verified that schema decoding now requires top-level `type: "struct"` and a present, non-null `fields` array, preserves the receiver until validation succeeds, and serializes nil fields as `[]`. It also correctly leaves `schema-id` optional for v1 and does not require `identifier-field-ids`.

Non-blocking follow-up: nested shape validation remains inconsistent with the new top-level rule. `types.go:375-401` accepts nested structs whose `fields` member is missing or null; `types.go:471-489` accepts lists with a missing/null `element` or omitted `element-required`; and `types.go:555-580` accepts maps with a missing/null `key` or `value` and defaults an omitted `value-required`. This is pre-existing and outside the stated top-level scope, but equivalent validation in a follow-up would keep validity from depending on nesting position.

zeroshade (approved):

PR #1662, opened 2026-08-05
fix(partitions): require fields in partition spec JSON

fallintoplace:
## Summary

Require spec-id and fields when decoding partition specs.

## Why

Missing fields were treated as an empty list, and a reused PartitionSpec could retain its old ID when spec-id was omitted.

## What changed

Both fields are now required and validated in temporary state. The receiver is replaced only after successful decoding. Compatibility for omitted partition field-id values is unchanged.

## Tests

- go test .

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -18,6 +18,7 @@
 package iceberg
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -454,20 +455,34 @@ func (ps PartitionSpec) MarshalJSON() ([]byte, error) {
 }
 
 func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
-	aux := struct {
-		ID     int               `json:"spec-id"`
-		Fields []json.RawMessage `json:"fields"`
-	}{ID: ps.id}
-
-	if err := json.Unmarshal(b, &aux); err != nil {
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(b, &raw); err != nil {
 		return err
 	}
-	if aux.ID < 0 {
-		return fmt.Errorf("%w: spec ID must be non-negative: %d", ErrInvalidPartitionSpec, aux.ID)
+
+	rawID, ok := raw["spec-id"]
+	if !ok || bytes.Equal(bytes.TrimSpace(rawID), []byte("null")) {
+		return fmt.Errorf("%w: partition spec is missing required spec-id", ErrInvalidPartitionSpec)
+	}
+	var id int
+	if err := json.Unmarshal(rawID, &id); err != nil {
+		return fmt.Errorf("%w: invalid partition spec ID: %w", ErrInvalidPartitionSpec, err)
+	}
+	if id < 0 {
+		return fmt.Errorf("%w: spec ID must be non-negative: %d", ErrInvalidPartitionSpec, id)
 	}
 
-	fields := make([]PartitionField, len(aux.Fields))
-	for i, rawField := range aux.Fields {
+	rawFields, ok := raw["fields"]
+	if !ok || bytes.Equal(bytes.TrimSpace(rawFields), []byte("null")) {
+		return fmt.Errorf("%w: partition spec is missing required fields", ErrInvalidPartitionSpec)
+	}
+	var rawFieldList []json.RawMessage
+	if err := json.Unmarshal(rawFields, &rawFieldList); err != nil {
+		return fmt.Errorf("%w: invalid partition spec fields: %w", ErrInvalidPartitionSpec, err)
+	}
+
+	fields := make([]PartitionField, len(rawFieldList))
+	for i, rawField := range rawFieldList {
 		var keys map[string]json.RawMessage
 		if err := json.Unmarshal(rawField, &keys); err != nil {
 			return err
@@ -489,11 +504,12 @@ func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	ps.id, ps.fields = aux.ID, fields
-	if err := ps.assignPartitionFieldIds(nil); err != nil {
+	decoded := PartitionSpec{id: id, fields: fields}
+	if err := decoded.assignPartitionFieldIds(nil); err != nil {
 		return fmt.Errorf("%w: %w", ErrInvalidPartitionSpec, err)
 	}
-	ps.initialize()
+	decoded.initialize()
+	*ps = decoded
 
 	return nil
 }
--- a/partitions.go
+++ b/partitions.go
@@ -457,16 +457,17 @@ func (ps PartitionSpec) MarshalJSON() ([]byte, error) {
 func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
 	var raw map[string]json.RawMessage
 	if err := json.Unmarshal(b, &raw); err != nil {
-		return err
+		return fmt.Errorf("%w: invalid partition spec JSON: %w", ErrInvalidPartitionSpec, err)
 	}
 
-	rawID, ok := raw["spec-id"]
-	if !ok || bytes.Equal(bytes.TrimSpace(rawID), []byte("null")) {
-		return fmt.Errorf("%w: partition spec is missing required spec-id", ErrInvalidPartitionSpec)
-	}
-	var id int
-	if err := json.Unmarshal(rawID, &id); err != nil {
-		return fmt.Errorf("%w: invalid partition spec ID: %w", ErrInvalidPartitionSpec, err)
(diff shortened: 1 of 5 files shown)


partitions.go
```diff
@@ -454,20 +455,34 @@ func (ps PartitionSpec) MarshalJSON() ([]byte, error) {
...
-	if err := json.Unmarshal(b, &aux); err != nil {
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(b, &raw); err != nil {
 		return err
 	}
-	if aux.ID < 0 {
-		return fmt.Errorf("%w: spec ID must be non-negative: %d", ErrInvalidPartitionSpec, aux.ID)
+
+	rawID, ok := raw["spec-id"]
+	if !ok || bytes.Equal(bytes.TrimSpace(rawID), []byte("null")) {
```
zeroshade: This is the blocking issue. Requiring `spec-id` is correct for stored table metadata, but this decoder is not only used for stored table metadata.

The REST OpenAPI `PartitionSpec` schema requires only `fields`; `spec-id` is optional and marked read-only, deliberately, so that a client can send a spec and let the server assign the ID. `PartitionSpec.UnmarshalJSON` is the single entry point for both shapes, so applying the stored-metadata requirement here applies it to requests too.

The concrete regression is at `table/updates.go:242` — `addPartitionSpecUpdate.Spec` is an `*iceberg.PartitionSpec`, so it decodes through this function. A spec-valid update that worked before this change:

```json
{"action": "add-spec", "spec": {"fields": []}}
```

now fails with `partition spec is missing required spec-id`. Create-table requests whose spec omits the ID are exposed identically. A shared unmarshaller cannot impose the stored-metadata requirement globally without breaking the request path.

Suggested fix, in two parts:

1. **Here:** keep `fields` required — that part is right and matches both schemas — but allow `spec-id` to be absent, defaulting to `InitialPartitionSpecID`. Note that the default must be the constant, *not* the receiver's current `ps.id`: the old code seeded `aux` from `ps.id`, and reintroducing that would make decoding depend on whatever the receiver happened to hold, which is the inheritance problem the `decoded` local at `:507` otherwise fixes. Explicit `null` should still be rejected, as should a negative or non-integer value.

2. **In the metadata decoder:** enforce the requirement where it actually belongs, in the code that parses full table metadata and therefore knows both the format version and the context. That is the only place with enough information to distinguish "stored metadata missing a required field" from "a request legitimately omitting a read-only one."

partitions_test.go
```diff
@@ -317,6 +317,27 @@ func TestSerializePartitionSpec(t *testing.T) {
 	assert.True(t, spec.Equals(outspec))
 }
 
+func TestDeserializePartitionSpecRequiresTopLevelFields(t *testing.T) {
+	for _, tt := range []struct {
+		name string
+		data string
+	}{
+		{name: "missing spec id", data: `{"fields": []}`},
```
zeroshade: This case encodes the behavior that needs to change. `{"fields": []}` is a **valid** REST `PartitionSpec` — the OpenAPI schema requires `fields` and nothing else — so asserting that it returns `ErrInvalidPartitionSpec` locks in the request-path regression described at `partitions.go:464`.

Suggested fix: move this case out of the error table and into a positive test asserting that it decodes successfully with `ID() == InitialPartitionSpecID` and zero fields. The other four rows here are all correct and should stay: `null` `spec-id` is genuinely invalid even where the field is optional, and the three `fields` cases match the schema. Their assertions on receiver preservation are a nice touch and worth keeping in the positive test too — starting from `NewPartitionSpecID(7)` and confirming the decoded ID is `0`, not `7`, is exactly the check that catches the receiver-inheritance trap.

partitions.go
```diff
@@ -454,20 +455,34 @@ func (ps PartitionSpec) MarshalJSON() ([]byte, error) {
...
 func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
-	aux := struct {
-		ID     int               `json:"spec-id"`
-		Fields []json.RawMessage `json:"fields"`
-	}{ID: ps.id}
-
-	if err := json.Unmarshal(b, &aux); err != nil {
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(b, &raw); err != nil {
 		return err
```
zeroshade: Non-blocking, for consistency. Every other failure path added in this PR wraps its error with `ErrInvalidPartitionSpec`, so callers can use `errors.Is` uniformly. This one returns the bare `json.Unmarshal` error, which means a malformed top-level value — a string or an array where an object was expected — is the one shape error that does not match the sentinel.

Suggested fix: wrap it the same way as the others, e.g. `fmt.Errorf("%w: %w", ErrInvalidPartitionSpec, err)`. The same applies to the two bare `return err` paths inside the field loop, at `:488` and `:500`.

zeroshade (changes requested):
Thanks for this — the change is careful, and two of the judgement calls in it are exactly right.

First, it does **not** make `field-id` mandatory on partition fields. That matters more than it might look: v1 partition specs were written without field IDs and depend on sequential assignment from 1000, so requiring the ID would have broken reading them. `partitions.go:490` correctly treats the key as optional while still rejecting an explicit `null`, which is the right line to draw.

Second, the decode is now properly all-or-nothing. Building into a local `decoded` at `partitions.go:507` and only assigning `*ps = decoded` at `:512` after `assignPartitionFieldIds` succeeds means a failed unmarshal leaves the receiver untouched — a real improvement over the previous version, which set `ps.id` and `ps.fields` before validating them. Using raw presence checks rather than typed zero values to separate "absent" from `null` from a legitimate `0` is the right technique for this, and `:475` applies it correctly to `fields`.

The blocking problem is one of scope: this decoder serves two different schemas, and only one of them requires `spec-id`.

## The blocking issue

`partitions.go:463-465` unconditionally rejects a partition spec without `spec-id`. That is correct for stored table metadata, where the table spec does require it on the full partition-spec object. It is not correct for REST requests, where the OpenAPI `PartitionSpec` schema requires only `fields` and declares `spec-id` optional and read-only — precisely so clients can omit an ID the server will assign.

`PartitionSpec.UnmarshalJSON` is the single decoder for both. The concrete regression is at **`table/updates.go:242`**, where `addPartitionSpecUpdate.Spec` is an `*iceberg.PartitionSpec` and so decodes through this same path. A spec-valid update:

```json
{"action": "add-spec", "spec": {"fields": []}}
```

now fails to decode. Create-table requests carrying a spec without an ID are exposed the same way. Full reasoning and the suggested fix are inline at `partitions.go:464`.

## Where this sits relative to #1660, #1661 and #1667

Read as a group, these four are mostly coherent: #1660 accepts values Java genuinely emitted, and #1661 and #1667 reject only shapes the spec and the OpenAPI schemas already declare invalid. #1662 is the exception. It is the only one of the four that applies the strictness of the **stored table metadata** schema to a type that is also used in **REST requests**, where the same field is optional and read-only.

That is why this one produces an interop regression and the others do not: the validation is context-insensitive, applied at a decoder that cannot see which of the two contexts it is running in. Worth keeping in view if more shape validation lands on shared types — the question to ask each time is whether the type is request-facing as well as metadata-facing.

## Non-blocking

- **`partitions.go:490`** — a pre-existing ambiguity this PR inherits rather than introduces: an explicit `"field-id": 0` is indistinguishable from omission and gets reassigned by `assignPartitionFieldIds`. Relatedly, the decoder accepts a *mix* of fields with and without IDs, where Java expects all-or-none. Worth rejecting partially specified field IDs in a follow-up, so a spec that supplies some IDs and not others fails loudly rather than having the gaps silently filled.
- **`partitions.go:460`** — a top-level type error surfaces as a bare `json.Unmarshal` error, unlike every other failure path here. Noted inline.

## Test coverage

The new table at `partitions_test.go:320` is a good shape — it covers the null/missing/non-array cases and asserts receiver preservation on each. Required before merge:

- A REST partition spec with `fields` but no `spec-id`, asserting it **decodes** and defaults to `InitialPartitionSpecID`.
- `{"action": "add-spec", "spec": {"fields": []}}` through `table/updates.go`, asserting it decodes.
- A create-table request whose spec omits `spec-id`.

Also missing, and worth adding while this area is open:

- A v1 spec whose fields omit `field-id`, asserting sequential assignment from 1000 — this PR deliberately preserves that behavior, and nothing currently pins it.
- An explicit valid `"spec-id": 0`, confirming it is not mistaken for absence.
- A malformed `spec-id` — non-integer, and negative — hitting the two distinct error paths at `:469` and `:472`.
- A spec mixing fields with and without `field-id`.
- Receiver preservation after a field-parsing error, a duplicate-ID error, and a duplicate-name error, not only after the top-level shape errors.

zeroshade (approved):

PR #1667, opened 2026-08-05
fix(table): require fields in update decoders

fallintoplace:
## Summary

Require required wire fields when decoding metadata updates.

## Why

Missing payload fields could become Go zero values, nil objects, or empty no-op updates.

## What changed

Every update action now validates its required fields before decoding into the existing update types. Explicit empty collections remain valid where the format allows them.

## Tests

- go test ./table -count=1

### DIFF
--- a/table/updates.go
+++ b/table/updates.go
@@ -92,7 +92,7 @@ var requiredUpdateFields = map[string][]string{
 	UpdateRemoveProperties:          {"removals"},
 	UpdateRemoveSpec:                {"spec-ids"},
 	UpdateRemoveSchemas:             {"schema-ids"},
-	UpdateSetStatistics:             {"snapshot-id", "statistics"},
+	UpdateSetStatistics:             {"statistics"},
 	UpdateRemoveStatistics:          {"snapshot-id"},
 	UpdateSetPartitionStatistics:    {"partition-statistics"},
 	UpdateRemovePartitionStatistics: {"snapshot-id"},
--- a/table/updates.go
+++ b/table/updates.go
@@ -132,10 +132,17 @@ func (u *Updates) UnmarshalJSON(data []byte) error {
 		updates = make(Updates, 0, len(rawUpdates))
 	}
 	for _, raw := range rawUpdates {
-		var base baseUpdate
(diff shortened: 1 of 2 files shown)


table/updates.go
```diff
@@ -140,6 +188,9 @@ func (u *Updates) UnmarshalJSON(data []byte) error {
 		default:
```
zeroshade, line 195: Non-blocking: Consider validating a missing or null `action` explicitly before this switch. Both currently become `unknown update action`, while a field-specific diagnostic would match the clarity introduced for the action payloads.

table/updates_test.go
```diff
@@ -564,6 +564,53 @@ func TestUnmarshalUpdatesReplacesExistingSlice(t *testing.T) {
 	assert.Empty(t, updates)
 }
 
+func TestUnmarshalUpdatesRejectsMissingRequiredFields(t *testing.T) {
```
zeroshade, line 567: Non-blocking: Consider independently omitting `type` and `snapshot-id` from `set-snapshot-ref`, covering null scalar/map/array fields, and confirming explicitly empty collections remain valid. Positive cases for `add-schema` without `last-column-id` and `set-snapshot-ref` without retention fields, plus receiver preservation after required-field failure, would complete the contract.

zeroshade (commented):
This change is correct. I verified that the required-field map covers all 23 REST update actions and rejects both absent and explicit-null required values before mutating the receiver. It correctly leaves deprecated `last-column-id`, the deprecated top-level `snapshot-id` on `set-statistics`, and the `set-snapshot-ref` retention fields optional.

Additional non-blocking follow-ups:

- `table/updates.go:410-450` stores optional snapshot-reference retention settings as values, so omission and explicit zero are indistinguishable and `Apply` ignores non-positive values. Consider pointer-backed fields in a separate change if that distinction becomes meaningful.
- `table/updates.go:478-498` recognizes only the canonical `updates` and `removals` properties. If compatibility with legacy Java `updated`/`removed` aliases is desired, consider supporting those deliberately rather than allowing them to decode as empty updates.

zeroshade (approved):

PR #1669, opened 2026-08-05
fix(table): require fields in requirement decoders

fallintoplace:
## Summary

Require required JSON fields when decoding table and view requirements.

## Why

The decoders used ordinary Go value fields, so omitted fields became zero values. A malformed requirement could pass when zero was a valid current ID.

## What changed

Required fields are now checked while explicit zero values remain valid. A nullable snapshot ID is still accepted for assert-ref-snapshot-id. Both JSON entry points use the same strict parsing path.

## Tests

- go test ./table ./view -count=1

### DIFF
--- a/table/requirement_test.go
+++ b/table/requirement_test.go
@@ -93,7 +93,11 @@ func TestParseRequirementBytes(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			actual, err := table.ParseRequirementBytes(tc.data)
 			assert.Equal(t, tc.expected, actual)
-			assert.Equal(t, tc.expectedErr, err)
+			if tc.expectedErr != nil {
+				require.ErrorIs(t, err, tc.expectedErr)
+			} else {
+				require.NoError(t, err)
+			}
 		})
 	}
 }
@@ -173,42 +177,76 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
 
 func TestParseRequirementRejectsMissingRequiredFields(t *testing.T) {
 	tests := []struct {
-		name string
-		data string
+		name          string
+		data          string
+		expectedField string
 	}{
-		{name: "table uuid", data: `{"type":"assert-table-uuid"}`},
-		{name: "missing ref", data: `{"type":"assert-ref-snapshot-id"}`},
-		{name: "missing snapshot id", data: `{"type":"assert-ref-snapshot-id","ref":"main"}`},
-		{name: "default spec id", data: `{"type":"assert-default-spec-id"}`},
-		{name: "current schema id", data: `{"type":"assert-current-schema-id"}`},
-		{name: "default sort order id", data: `{"type":"assert-default-sort-order-id"}`},
-		{name: "last assigned field id", data: `{"type":"assert-last-assigned-field-id"}`},
-		{name: "last assigned partition id", data: `{"type":"assert-last-assigned-partition-id"}`},
-		{name: "null current schema id", data: `{"type":"assert-current-schema-id","current-schema-id":null}`},
+		{name: "missing type", data: `{}`, expectedField: "type"},
+		{name: "null type", data: `{"type":null}`, expectedField: "type"},
+		{name: "table uuid", data: `{"type":"assert-table-uuid"}`, expectedField: "uuid"},
+		{name: "null table uuid", data: `{"type":"assert-table-uuid","uuid":null}`, expectedField: "uuid"},
+		{name: "missing ref", data: `{"type":"assert-ref-snapshot-id"}`, expectedField: "ref"},
+		{name: "missing snapshot id", data: `{"type":"assert-ref-snapshot-id","ref":"main"}`, expectedField: "snapshot-id"},
+		{name: "default spec id", data: `{"type":"assert-default-spec-id"}`, expectedField: "default-spec-id"},
+		{name: "null default spec id", data: `{"type":"assert-default-spec-id","default-spec-id":null}`, expectedField: "default-spec-id"},
+		{name: "current schema id", data: `{"type":"assert-current-schema-id"}`, expectedField: "current-schema-id"},
+		{name: "null current schema id", data: `{"type":"assert-current-schema-id","current-schema-id":null}`, expectedField: "current-schema-id"},
+		{name: "default sort order id", data: `{"type":"assert-default-sort-order-id"}`, expectedField: "default-sort-order-id"},
+		{name: "null default sort order id", data: `{"type":"assert-default-sort-order-id","default-sort-order-id":null}`, expectedField: "default-sort-order-id"},
+		{name: "last assigned field id", data: `{"type":"assert-last-assigned-field-id"}`, expectedField: "last-assigned-field-id"},
+		{name: "null last assigned field id", data: `{"type":"assert-last-assigned-field-id","last-assigned-field-id":null}`, expectedField: "last-assigned-field-id"},
+		{name: "last assigned partition id", data: `{"type":"assert-last-assigned-partition-id"}`, expectedField: "last-assigned-partition-id"},
+		{name: "null last assigned partition id", data: `{"type":"assert-last-assigned-partition-id","last-assigned-partition-id":null}`, expectedField: "last-assigned-partition-id"},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			_, err := table.ParseRequirementBytes([]byte(tt.data))
-			require.Error(t, err)
+			require.ErrorIs(t, err, table.ErrInvalidRequirement)
+			require.ErrorContains(t, err, tt.expectedField)
 
 			var requirements table.Requirements
 			err = json.Unmarshal([]byte("["+tt.data+"]"), &requirements)
-			require.Error(t, err)
+			require.ErrorIs(t, err, table.ErrInvalidRequirement)
+			require.ErrorContains(t, err, tt.expectedField)
 		})
 	}
 }
 
 func TestParseRequirementAcceptsExplicitZero(t *testing.T) {
-	actual, err := table.ParseRequirementBytes([]byte(`{"type":"assert-current-schema-id","current-schema-id":0}`))
-	require.NoError(t, err)
-	assert.Equal(t, table.AssertCurrentSchemaID(0), actual)
+	tests := []struct {
+		name     string
+		data     string
+		expected table.Requirement
+	}{
+		{name: "default spec id", data: `{"type":"assert-default-spec-id","default-spec-id":0}`, expected: table.AssertDefaultSpecID(0)},
+		{name: "current schema id", data: `{"type":"assert-current-schema-id","current-schema-id":0}`, expected: table.AssertCurrentSchemaID(0)},
+		{name: "default sort order id", data: `{"type":"assert-default-sort-order-id","default-sort-order-id":0}`, expected: table.AssertDefaultSortOrderID(0)},
+		{name: "last assigned field id", data: `{"type":"assert-last-assigned-field-id","last-assigned-field-id":0}`, expected: table.AssertLastAssignedFieldID(0)},
+		{name: "last assigned partition id", data: `{"type":"assert-last-assigned-partition-id","last-assigned-partition-id":0}`, expected: table.AssertLastAssignedPartitionID(0)},
+		{name: "snapshot id", data: `{"type":"assert-ref-snapshot-id","ref":"main","snapshot-id":0}`, expected: table.AssertRefSnapshotID("main", ptr(int64(0)))},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			actual, err := table.ParseRequirementBytes([]byte(tt.data))
+			require.NoError(t, err)
+			assert.Equal(t, tt.expected, actual)
+		})
+	}
 }
 
 func TestParseRequirementRefRequiresRefButAllowsNullSnapshotID(t *testing.T) {
-	actual, err := table.ParseRequirementBytes([]byte(`{"type":"assert-ref-snapshot-id","ref":"main","snapshot-id":null}`))
+	data := []byte(`{"type":"assert-ref-snapshot-id","ref":"main","snapshot-id":null}`)
+
+	actual, err := table.ParseRequirementBytes(data)
 	require.NoError(t, err)
 	assert.Equal(t, table.AssertRefSnapshotID("main", nil), actual)
+
+	var requirements table.Requirements
+	require.NoError(t, json.Unmarshal([]byte("["+string(data)+"]"), &requirements))
+	require.Len(t, requirements, 1)
+	assert.Equal(t, table.AssertRefSnapshotID("main", nil), requirements[0])
 }
 
 func TestParseRequirementRefAcceptsNumericSnapshotID(t *testing.T) {
--- a/table/requirement_test.go
+++ b/table/requirement_test.go
@@ -171,6 +171,45 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
 	assert.Empty(t, requirements)
 }
 
+func TestParseRequirementRejectsMissingRequiredFields(t *testing.T) {
+	tests := []struct {
+		name string
+		data string
+	}{
+		{name: "table uuid", data: `{"type":"assert-table-uuid"}`},
+		{name: "ref", data: `{"type":"assert-ref-snapshot-id"}`},
+		{name: "default spec id", data: `{"type":"assert-default-spec-id"}`},
+		{name: "current schema id", data: `{"type":"assert-current-schema-id"}`},
+		{name: "default sort order id", data: `{"type":"assert-default-sort-order-id"}`},
+		{name: "last assigned field id", data: `{"type":"assert-last-assigned-field-id"}`},
+		{name: "last assigned partition id", data: `{"type":"assert-last-assigned-partition-id"}`},
+		{name: "null current schema id", data: `{"type":"assert-current-schema-id","current-schema-id":null}`},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := table.ParseRequirementBytes([]byte(tt.data))
+			require.Error(t, err)
+
+			var requirements table.Requirements
+			err = json.Unmarshal([]byte("["+tt.data+"]"), &requirements)
+			require.Error(t, err)
+		})
+	}
+}
+
+func TestParseRequirementAcceptsExplicitZero(t *testing.T) {
+	actual, err := table.ParseRequirementBytes([]byte(`{"type":"assert-current-schema-id","current-schema-id":0}`))
+	require.NoError(t, err)
+	assert.Equal(t, table.AssertCurrentSchemaID(0), actual)
+}
+
+func TestParseRequirementRefRequiresRefButAllowsNullSnapshotID(t *testing.T) {
+	actual, err := table.ParseRequirementBytes([]byte(`{"type":"assert-ref-snapshot-id","ref":"main","snapshot-id":null}`))
+	require.NoError(t, err)
+	assert.Equal(t, table.AssertRefSnapshotID("main", nil), actual)
+}
+
 func TestAssertRefSnapshotIDValidate(t *testing.T) {
 	meta, err := table.ParseMetadataBytes([]byte(table.ExampleTableMetadataV2))
 	require.NoError(t, err)
--- a/table/requirement_test.go
+++ b/table/requirement_test.go
@@ -236,6 +236,11 @@ func TestParseRequirementAcceptsExplicitZero(t *testing.T) {
 			actual, err := table.ParseRequirementBytes([]byte(tt.data))
 			require.NoError(t, err)
 			assert.Equal(t, tt.expected, actual)
+
+			var requirements table.Requirements
+			require.NoError(t, json.Unmarshal([]byte("["+tt.data+"]"), &requirements))
+			require.Len(t, requirements, 1)
+			assert.Equal(t, tt.expected, requirements[0])
 		})
 	}
 }
--- a/table/requirement_test.go
+++ b/table/requirement_test.go
@@ -48,7 +48,7 @@ func TestParseRequirementBytes(t *testing.T) {
 		},
 		{
 			name:        "Should parse an assert ref snapshot id",
-			data:        []byte(`{"type": "assert-ref-snapshot-id", "ref": "branch"}`),
+			data:        []byte(`{"type": "assert-ref-snapshot-id", "ref": "branch", "snapshot-id": null}`),
 			expected:    table.AssertRefSnapshotID("branch", nil),
 			expectedErr: nil,
 		},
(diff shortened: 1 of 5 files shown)


table/requirements.go
```diff
@@ -105,6 +82,151 @@ type baseRequirement struct {
...
+
+	case reqAssertRefSnapshotID:
+		var req assertRefSnapshotIDWire
+		if err := json.Unmarshal(b, &req); err != nil {
+			return nil, err
+		}
+		if req.Ref == nil {
+			return nil, requiredRequirementField("ref")
+		}
+		if !req.SnapshotID.Set {
```
zeroshade: Worth calling out as the thing this PR gets right. `snapshot-id` is nullable-not-optional in the spec — a required member with `nullable: true`, where an explicit `null` asserts that the ref must **not** exist. Treating it as merely optional (the natural `*int64` reading) would silently accept a document that omits it and quietly turn a "ref must not exist" assertion into a no-op.

The `nullableInt64` presence flag (`table/requirements.go:93-116`) plus this rejection is the correct implementation, and `TestParseRequirementRefRequiresRefButAllowsNullSnapshotID` pins both halves of it.

table/requirement_test.go
```diff
@@ -171,6 +171,52 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
 	assert.Empty(t, requirements)
 }
 
+func TestParseRequirementRejectsMissingRequiredFields(t *testing.T) {
```
zeroshade, line 179: Non-blocking: consider asserting `errors.Is(err, table.ErrInvalidRequirement)` rather than bare `require.Error`, and that the message names the missing field. As written, an unrelated decode failure — a malformed body, a renamed type constant — would satisfy these subtests just as well as the rejection they are meant to pin.

Two cases are also worth adding to the table: a missing `type`, and `{"type":null}`. The common discriminator is checked at `table/requirements.go:128-130` but nothing exercises it. `{"type":"assert-table-uuid","uuid":null}` and `{"type":"assert-ref-snapshot-id","ref":null}` would round out the explicit-null coverage the same way the `null current-schema-id` case does.

table/requirement_test.go
```diff
@@ -171,6 +171,52 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
...
+			require.Error(t, err)
+
+			var requirements table.Requirements
+			err = json.Unmarshal([]byte("["+tt.data+"]"), &requirements)
+			require.Error(t, err)
+		})
+	}
+}
+
+func TestParseRequirementAcceptsExplicitZero(t *testing.T) {
```
zeroshade, line 220: Non-blocking: this covers `current-schema-id: 0` only, but the explicit-zero distinction applies to every integer requirement — `default-spec-id`, `default-sort-order-id`, `last-assigned-field-id`, `last-assigned-partition-id`, and `snapshot-id: 0`.

Consider turning it into a table so each scalar decoder is exercised. `snapshot-id: 0` is the most valuable addition, since it is the one value that also has to be distinguished from `null`.

table/requirement_test.go
```diff
@@ -171,6 +171,52 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
...
+	actual, err := table.ParseRequirementBytes([]byte(`{"type":"assert-current-schema-id","current-schema-id":0}`))
+	require.NoError(t, err)
+	assert.Equal(t, table.AssertCurrentSchemaID(0), actual)
+}
+
+func TestParseRequirementRefRequiresRefButAllowsNullSnapshotID(t *testing.T) {
```
zeroshade, line 248: Non-blocking: this exercises the valid `snapshot-id: null` case through `ParseRequirementBytes` only. `TestParseRequirementRejectsMissingRequiredFields` checks both entry points; consider doing the same here, so `Requirements.UnmarshalJSON` is confirmed to *accept* the valid-null document rather than only to reject the invalid ones.

view/requirements_test.go
```diff
@@ -95,3 +95,16 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
 	require.NoError(t, json.Unmarshal([]byte(`[]`), &requirements))
 	assert.Empty(t, requirements)
 }
+
+func TestParseRequirementRejectsMissingUUID(t *testing.T) {
```
zeroshade, line 101: Non-blocking, mirroring the table-side note: consider asserting the sentinel (`errors.Is(err, table.ErrInvalidRequirement)`, which `view.requiredRequirementField` wraps at `view/requirements.go:84-86`) rather than bare `require.Error`, and adding a missing-`type` and `{"type":null}` case so the common discriminator check at `view/requirements.go:93-94` is exercised on this decoder too.

zeroshade (commented):
The change is correct, and there are no blocking issues. Verified against `18e55f5`:

- No field is wrongly made required. Every mapping matches the OpenAPI schema, and `assert-ref-snapshot-id.snapshot-id` is handled as required-but-nullable rather than optional — the easy thing to get wrong here, and this gets it right. Note inline at `table/requirements.go:155`.
- Explicit zero values survive: `current-schema-id: 0` and its siblings are no longer indistinguishable from an omitted member.
- Interop: requirements flow client → catalog, and both Java and PyIceberg emit an explicit `null` for the ref case, so no break was identified against Polaris, Nessie, Unity, Gravitino, Tabular, or S3Tables.
- Missing-field rejection is covered through both decoder entry points (`ParseRequirementBytes` and `Requirements.UnmarshalJSON`). CI green.

**Non-blocking**

Everything remaining is test coverage; no implementation defects found. The gaps cluster in three places, noted inline: no assertions on a missing or null `type` discriminator, no `errors.Is` sentinel checks on the new rejections, and explicit-zero coverage that exercises only `current-schema-id: 0`.

table/requirement_test.go
```diff
@@ -171,6 +175,87 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
...
+		{name: "null last assigned field id", data: `{"type":"assert-last-assigned-field-id","last-assigned-field-id":null}`, expectedField: "last-assigned-field-id"},
+		{name: "last assigned partition id", data: `{"type":"assert-last-assigned-partition-id"}`, expectedField: "last-assigned-partition-id"},
+		{name: "null last assigned partition id", data: `{"type":"assert-last-assigned-partition-id","last-assigned-partition-id":null}`, expectedField: "last-assigned-partition-id"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := table.ParseRequirementBytes([]byte(tt.data))
+			require.ErrorIs(t, err, table.ErrInvalidRequirement)
+			require.ErrorContains(t, err, tt.expectedField)
```
zeroshade: Non-blocking, and the fix is one line — but this assertion is weaker than it looks for the two `type` cases at lines 184-185.

For the 15 field cases it is solidly discriminating. I built a cross-satisfaction matrix over all 17 actual error strings and found **zero** instances of one case's error satisfying another case's `expectedField`, and mutation-tested it: changing `current schema id`'s expectation to `"default-spec-id"`, or `last assigned field id`'s to `"last-assigned-partition-id"`, both correctly fail. So field identity is genuinely pinned.

The `"type"` token is the problem — it is short and appears in unrelated error text. Three experiments, production otherwise at this commit:

1. A wrong-reason rejection that happens to mention type — `fmt.Errorf("%w: unknown requirement type", ErrInvalidRequirement)` → both `type` cases **pass**.
2. A leaked `json.UnmarshalTypeError` wrapped in the sentinel → both **pass**. That error's text always contains "type" (`json: cannot unmarshal object into Go value of type string`).
3. The realistic one: full pre-PR production revert **plus** one plausible independent improvement (sentinel-wrapping the unknown-type errors) → `missing type` and `null type` both **pass** on a build containing none of this PR's fixes, while the other 15 correctly fail.

Asserting the whole message closes all three at once and also removes a latent substring hazard (`"spec-id"` is a substring of `"default-spec-id"`, so a future author writing the short form would get a silently loose test — confirmed it still passes today):

```go
require.ErrorContains(t, err, fmt.Sprintf("missing required field %q", tt.expectedField))
```

Same change applies to line 212 for the list path.

table/requirement_test.go
```diff
@@ -93,7 +93,11 @@ func TestParseRequirementBytes(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			actual, err := table.ParseRequirementBytes(tc.data)
 			assert.Equal(t, tc.expected, actual)
-			assert.Equal(t, tc.expectedErr, err)
+			if tc.expectedErr != nil {
+				require.ErrorIs(t, err, tc.expectedErr)
```
zeroshade: Non-blocking: this loosening was not required by the PR, and it gives up a real bit of coverage.

Only one case in this table has a non-nil `expectedErr` — `"invalid requirement"` at line 89 — and its value is the **bare unwrapped sentinel**, since `ParseRequirementBytes` returns `ErrInvalidRequirement` verbatim (`table/requirements.go:476`). So no case here was ever pinning a wrapped or formatted message that `ErrorIs` would now miss. I confirmed the change was unnecessary: restoring `assert.Equal(t, tc.expectedErr, err)` against the production code at this commit passes 9/9.

What now slips through: mutate `ParseRequirementBytes`'s unknown-type handler to `requiredRequirementField("type")`, so an unknown requirement type is misreported as a *missing-`type`-field* error. The new assertion passes 9/9; the old one catches it:

```
expected: *errors.errorString{s:"invalid requirement"}
actual  : *fmt.wrapError{msg:"invalid requirement: missing required field \"type\"", ...}
```

Note this compounds with the `ErrorContains` note above — that exact misclassification is also what would let the `"type"` cases pass for the wrong reason, and after this change nothing in the suite pins that the unknown-type error is bare.

The `else { require.NoError(...) }` branch is a genuine improvement over `assert.Equal` on nil, so keep that. Suggest either keeping `assert.Equal` for the error case, or adding an `expectedErrMsg string` column and asserting both the sentinel and the exact message.

table/requirement_test.go
```diff
@@ -171,6 +175,87 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
...
+		name     string
+		data     string
+		expected table.Requirement
+	}{
+		{name: "default spec id", data: `{"type":"assert-default-spec-id","default-spec-id":0}`, expected: table.AssertDefaultSpecID(0)},
+		{name: "current schema id", data: `{"type":"assert-current-schema-id","current-schema-id":0}`, expected: table.AssertCurrentSchemaID(0)},
+		{name: "default sort order id", data: `{"type":"assert-default-sort-order-id","default-sort-order-id":0}`, expected: table.AssertDefaultSortOrderID(0)},
+		{name: "last assigned field id", data: `{"type":"assert-last-assigned-field-id","last-assigned-field-id":0}`, expected: table.AssertLastAssignedFieldID(0)},
+		{name: "last assigned partition id", data: `{"type":"assert-last-assigned-partition-id","last-assigned-partition-id":0}`, expected: table.AssertLastAssignedPartitionID(0)},
+		{name: "snapshot id", data: `{"type":"assert-ref-snapshot-id","ref":"main","snapshot-id":0}`, expected: table.AssertRefSnapshotID("main", ptr(int64(0)))},
```
zeroshade, line 231: This is the case I most wanted from the last round, and it is the right one to have added — `snapshot-id: 0` is the only value that must be distinguished from both an omitted member *and* an explicit `null`, and it is now pinned alongside the `null` case at line 241 and the rejection case at line 190. Those three together fully cover the nullable-not-optional semantics.

Worth knowing (not a defect): none of the six subtests here fail if the production fix is reverted, because the old decoder accepted explicit zero too. Their value is in the other direction — they fail against a naive `!Set || Value == nil` presence check, which is the most likely way someone would reimplement this and accidentally reject legitimate zeros. Confirmed by mutation. Good regression guards to keep.

`ptr` is the pre-existing generic helper in this package (`table/snapshots_test.go`), not a new duplicate — no action needed.

view/requirements_test.go
```diff
@@ -95,3 +96,34 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
...
+		require.ErrorContains(t, err, "uuid")
+	}
+}
+
+func TestParseRequirementRejectsMissingOrNullType(t *testing.T) {
+	for _, data := range []string{`{}`, `{"type":null}`} {
+		t.Run(data, func(t *testing.T) {
+			_, err := view.ParseRequirementBytes([]byte(data))
+			require.ErrorIs(t, err, table.ErrInvalidRequirement)
+			require.ErrorContains(t, err, "type")
```
zeroshade: Non-blocking, same as the table-side note: `"type"` as a bare substring is incidentally satisfiable — a wrapped `json.UnmarshalTypeError` or any message mentioning "unknown requirement type" satisfies it without the missing-field rejection this test is meant to pin. `require.ErrorContains(t, err, fmt.Sprintf("missing required field %q", "type"))` closes it. Applies to line 126 too, and the `"uuid"` assertions at lines 107/112 would benefit from the same full-message form (note `json: cannot unmarshal number into Go value of type *uuid.UUID` contains both "type" and "uuid").

The new coverage itself is correct and was a real gap — both cases fail on a production revert, and the common discriminator check now gets exercised on this decoder rather than only the table one. Confirmed the `errors.Is` chain works here: `view.requiredRequirementField` wraps `table.ErrInvalidRequirement`, which was already the view parser's public sentinel before this PR, so importing `table` for it in the test is correct rather than an odd coupling — there is no separate view sentinel to prefer.

zeroshade (approved):
Re-reviewed at `d34d1a1`. The delta since `18e55f5` is test-only, the production code is unchanged from my prior audit of it, and all five notes from that round are addressed. Approving — everything below is non-blocking.

Gates green: `go build ./...`, `go vet ./table/ ./view/`, `go test ./table/ ./view/ -count=1`, `gofmt`, and `go test -race -run TestParseRequirement`.

**The new rejection table is genuinely load-bearing.** Reverting only the production hunks and re-running the tests at this commit fails **17 of 17** subtests, on both entry points, with zero passing. Nothing in that table is dead weight.

What makes that worth stating plainly is what the old decoders actually *did* with those 17 documents: **15 of 17 were silently accepted** as zero-valued requirements. `{"type":"assert-ref-snapshot-id","ref":null}` previously produced a requirement asserting that the `""` ref must not exist. The remaining two (`type` cases) were rejected, but on the list path with an unwrapped error, so `errors.Is` was false. That is the bug class, concretely, and this PR closes it.

The six `TestParseRequirementAcceptsExplicitZero` cases and the null-snapshot test do *not* fail on revert — correctly so. They are over-correction guards, and they do fail against a naive `!Set || Value == nil` implementation. Right design; worth keeping.

Spec fidelity was audited independently in both directions across all nine requirement types against the REST OpenAPI schemas: no field wrongly made required (the dangerous direction here — it would reject documents real catalogs legitimately send), and no spec-required field left optional. The three-state nullable handling for `snapshot-id` is correct and the presence flag is not spoofable; `AssertRefSnapshotID("main", nil)` emits explicit `null` and round-trips through both strict entry points; no partially-decoded value escapes on error.

Two test-quality items are worth a follow-up, noted inline: the bare `"type"` substring assertions are incidentally satisfiable, and the `assert.Equal` → `require.ErrorIs` change in `TestParseRequirementBytes` gives up a bit of coverage it did not need to. Neither weakens what this PR fixes.

**Separate pre-existing issue, explicitly NOT a gate on this PR.** Not every invalid requirement wraps `ErrInvalidRequirement`: `Requirements.UnmarshalJSON` returns an unwrapped `unknown requirement type: X` (`table/requirements.go:66`, `view/requirements.go:56`) while `ParseRequirementBytes` returns the sentinel, and wrong-typed fields, invalid UUIDs, and malformed `snapshot-id` all surface raw errors with `errors.Is(err, ErrInvalidRequirement) == false`. I ran the same probe against the base commit `d785b27` and the results are **identical on every case**, so this PR neither causes nor worsens it — the unwrapped `fmt.Errorf` exists verbatim pre-PR. Happy to open a separate issue for the error-taxonomy cleanup.

Methodology: audited by two independent models on separate axes (test efficacy via revert + mutation experiments; correctness via spec/marshal/caller audit). Every claim above was verified empirically rather than by inspection, and one reviewer's blocking finding was downgraded only after I reproduced it against the base commit and confirmed it predates the change.

table/requirements.go
laskoviymishka: This is the one spot I'd want a second look at before it lands. Making an absent `snapshot-id` an error is spec-correct — the OpenAPI `AssertRefSnapshotId` lists it in `required` — but it's stricter than the Java and rust reference implementations, which both treat an absent key as equivalent to `null` (`getLongOrNull`/`hasNonNull` on the Java side, `Option<i64>` on rust).

It's also a behavior change for us specifically: `{"type":"assert-ref-snapshot-id","ref":"branch"}` with no `snapshot-id` used to parse to `AssertRefSnapshotID("branch", nil)`, and this PR removes that test case and now rejects the input. Java and PyIceberg always send explicit `null`, so I don't think this breaks live catalog traffic — but any caller that hand-builds requirement docs and omits the key will start getting an error.

I'm fine landing it strict since that matches the spec, I'd just want us to pick "strict" knowingly. wdyt — keep it as-is, or fall back to treating absent as null like the other clients?

table/requirement_test.go
```diff
@@ -171,6 +176,89 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
...
+		{name: "last assigned field id", data: `{"type":"assert-last-assigned-field-id"}`, expectedField: "last-assigned-field-id"},
+		{name: "null last assigned field id", data: `{"type":"assert-last-assigned-field-id","last-assigned-field-id":null}`, expectedField: "last-assigned-field-id"},
+		{name: "last assigned partition id", data: `{"type":"assert-last-assigned-partition-id"}`, expectedField: "last-assigned-partition-id"},
+		{name: "null last assigned partition id", data: `{"type":"assert-last-assigned-partition-id","last-assigned-partition-id":null}`, expectedField: "last-assigned-partition-id"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			expectedError := fmt.Sprintf("missing required field %q", tt.expectedField)
+
```
laskoviymishka, line 207: Seconding the substring point from the earlier round — `requiredRequirementField` formats the name with `%q`, so the error reads `missing required field "type"`. Asserting the bare `tt.expectedField` (`type`, `uuid`, ...) passes on any error text that happens to contain those letters. Matching the quoted form (assert on `"`+tt.expectedField+`"`) pins it to the actual field name.

table/requirement_test.go
```diff
@@ -171,6 +176,89 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
...
+		expected table.Requirement
+	}{
+		{name: "default spec id", data: `{"type":"assert-default-spec-id","default-spec-id":0}`, expected: table.AssertDefaultSpecID(0)},
+		{name: "current schema id", data: `{"type":"assert-current-schema-id","current-schema-id":0}`, expected: table.AssertCurrentSchemaID(0)},
+		{name: "default sort order id", data: `{"type":"assert-default-sort-order-id","default-sort-order-id":0}`, expected: table.AssertDefaultSortOrderID(0)},
+		{name: "last assigned field id", data: `{"type":"assert-last-assigned-field-id","last-assigned-field-id":0}`, expected: table.AssertLastAssignedFieldID(0)},
+		{name: "last assigned partition id", data: `{"type":"assert-last-assigned-partition-id","last-assigned-partition-id":0}`, expected: table.AssertLastAssignedPartitionID(0)},
+		{name: "snapshot id", data: `{"type":"assert-ref-snapshot-id","ref":"main","snapshot-id":0}`, expected: table.AssertRefSnapshotID("main", ptr(int64(0)))},
+	}
+
```
laskoviymishka, line 233: Every other new test here exercises both `ParseRequirementBytes` and `Requirements.UnmarshalJSON`, but this one only hits `ParseRequirementBytes`. They converge on the same path so it's not a correctness gap, just an asymmetry — a parallel `json.Unmarshal([]byte("["+tt.data+"]"), &requirements)` assertion would keep it consistent with the rest.

table/requirements.go
```diff
@@ -105,6 +82,151 @@ type baseRequirement struct {
...
+type requirementWire struct {
+	Type *string `json:"type"`
+}
+
+type assertTableUUIDWire struct {
+	UUID *uuid.UUID `json:"uuid"`
+}
+
+type nullableInt64 struct {
+	Set   bool
```
laskoviymishka: `Set` and `Value` are exported on an unexported type, so nothing outside the package can reach them anyway — I'd lowercase them to `set`/`value` to match the usual convention. Small thing.

table/requirements.go
```diff
@@ -105,6 +82,151 @@ type baseRequirement struct {
...
+	Value *int64
+}
+
+func (n *nullableInt64) UnmarshalJSON(data []byte) error {
+	n.Set = true
+	n.Value = nil
+	if bytes.Equal(bytes.TrimSpace(data), []byte("null")) {
```
laskoviymishka: `encoding/json` never hands `UnmarshalJSON` whitespace-padded input — the token's already trimmed by the time it reaches here, so the `bytes.TrimSpace` is dead work, and it's the only reason this file needs the `bytes` import. `bytes.Equal(data, []byte("null"))` or just `string(data) == "null"` does the same thing.

view/requirements.go
```diff
@@ -93,6 +73,43 @@ type baseRequirement struct {
 	Type string `json:"type"`
 }
 
+type requirementWire struct {
```
laskoviymishka: `requirementWire` and `requiredRequirementField` here are byte-for-byte identical to the ones in `table` (which `view` already imports). Not worth blocking over, but it's the kind of thing that quietly drifts out of sync — worth a shared internal helper if we're touching this area again.

view/requirements_test.go
```diff
@@ -95,3 +97,36 @@ func TestParseRequirementListReplacesExistingSlice(t *testing.T) {
 	require.NoError(t, json.Unmarshal([]byte(`[]`), &requirements))
 	assert.Empty(t, requirements)
 }
+
+func TestParseRequirementRejectsMissingUUID(t *testing.T) {
```
laskoviymishka, line 101: This loop isn't wrapped in `t.Run`, so if the first input fails the `require` call stops the whole test and the second case (`"uuid":null`) never runs. `TestParseRequirementRejectsMissingOrNullType` right below it uses `t.Run(data, ...)` — I'd match that here so both cases always execute.

laskoviymishka (approved):
Took a second maintainer pass over this on top of the approval, and it holds up — the three-state snapshot-id decode and the revert-proof that the rejection table actually bites are the right call.

One thing I don't think came up earlier: making an absent `snapshot-id` key an error is spec-correct, but it's stricter than both Java and rust (which treat absent as null), and it drops the old `{"ref":"branch"}`-with-no-snapshot-id case that used to parse. I don't think it breaks live catalog traffic since Java and PyIceberg always send explicit `null` — I'd just want us to pick "strict" on purpose rather than as a side effect. Left it inline.

The rest are nits: a `bytes.TrimSpace` that never does anything, a view test loop that should be `t.Run` so both cases run, and light seconds on the test-tightening points from the earlier round. None of them block.

Nothing here changes the approval — happy to see it land once the snapshot-id call is settled one way or the other.

PR #1670, opened 2026-08-05
fix(table): normalize EWKB geometry values on write

fallintoplace:
## Summary

Normalize EWKB geometry values to ISO WKB before writing Parquet data.

## Why

EWKB values could be read and have their bounds computed correctly, but rewrites stored the original EWKB bytes.

## What changed

- EWKB values are converted before the write, while valid ISO WKB values and nulls are preserved.
- The original byte order is retained.
- Nested WKB values are normalized without mutating the input batch.
- Normalization only visits schema columns that can contain GeoArrow WKB, so non-geo writes skip the nested traversal.
- Values hidden behind null ancestors are copied without decoding. They remain unreachable and do not make the write fail.
- ISO WKB empty points are accepted during bounds accumulation and skipped as NaN coordinates.

## Tests

- go test ./...
- go test -race ./table/internal -count=1
- go vet ./...

### DIFF
--- a/table/internal/geo_codec.go
+++ b/table/internal/geo_codec.go
@@ -26,6 +26,7 @@ import (
 	"github.com/twpayne/go-geom"
 	"github.com/twpayne/go-geom/encoding/ewkb"
 	"github.com/twpayne/go-geom/encoding/wkb"
+	"github.com/twpayne/go-geom/encoding/wkbcommon"
 )
 
 // Geo bounding box dimension indices. X is longitude/easting, Y is
@@ -38,6 +39,8 @@ const (
 	geoDimM
 	geoNumDims
 )
+
+var wkbEmptyPointOption = wkbcommon.WKBOptionEmptyPointHandling(wkbcommon.EmptyPointHandlingNaN)
 
 // geoBoundsAccumulator computes a geospatial bounding box from a stream of WKB
 // values. Iceberg stores geometry/geography column bounds using the single-value
@@ -113,7 +116,7 @@ func decodeWKB(data []byte) (geom.T, error) {
 		return ewkb.Unmarshal(data)
 	}
 
-	return wkb.Unmarshal(data)
+	return wkb.Unmarshal(data, wkbEmptyPointOption)
 }
 
 func normalizeWKB(data []byte) ([]byte, error) {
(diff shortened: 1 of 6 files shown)


zeroshade (commented):
The core conversion logic in `normalizeWKB` (table/internal/geo_codec.go) is correct — byte order preserved, ISO passthrough zero-copy, SRID stripped — and the test coverage (sliced containers, null masking at every nesting level, end-to-end Parquet round-trip) is genuinely thorough. Tests build and pass on the PR head. A few things I'd like addressed/answered before this leaves draft:

1. **Per-batch overhead for non-geo schemas.** `normalizeGeoBatch` recurses into every column of every batch via `normalizeNestedArrayReachable` (table/internal/parquet_files.go). For any nested column — geo or not — `normalizeListChild` allocates a `childActive []bool` sized to the full child value range, slices the values array, and walks all offsets, on every `Write` call. For a writer streaming large batches of nested non-geo data this is measurable pure waste. A cheap datatype-level pre-check (does the column's `arrow.DataType` contain a `*geoarrow.WKBType` anywhere — computable once per writer, or at least once per column type) would make the common no-geo case zero-cost. Since `parquetFormat.NewFileWriter` already inspects the schema for WKB types around line 404, the writer could cache which column indices even need visiting.

2. **Undocumented behavior change on the read/bounds path.** `decodeWKB` now passes `EmptyPointHandlingNaN` to `wkb.Unmarshal`, so ISO WKB empty points (NaN coords) that previously failed bounds accumulation — and therefore failed the write — are now accepted. That looks like a deliberate and reasonable fix (the accumulator already documents NaN-skipping), but it's a distinct behavior change not mentioned in the PR description. Please call it out there, or split it if it's separable. Also note the asymmetry: `normalizeWKB`'s `ewkb.Unmarshal(data)` call doesn't pass the empty-point option while the marshal side does — the empty-point tests pass, so it works today, but a one-line comment on why the option is only needed on the ISO side would prevent someone "fixing" the asymmetry later.

3. **Confirming intent on masked values:** values hidden behind null ancestors are copied verbatim (still EWKB, and malformed bytes don't error), while the same bytes in a reachable slot would either be normalized or fail the write. I see from the tests this is deliberate — it avoids failing writes on garbage in masked slots — but it does mean written files can still contain EWKB bytes in unreachable child positions. Fine by me, just want it confirmed as the intended contract since `accumulateGeoBounds` makes the same reachability distinction.

Minor: dictionary- or REE-encoded columns wrapping a WKB extension type would silently pass through unnormalized; probably impossible via the current write path, but worth a defensive comment if so.

Since this is a draft I'm leaving a comment rather than a formal review outcome.

zeroshade:
Since this is a draft, I'll hold off of further review until this is marked ready

zeroshade (approved):

PR #1671, opened 2026-08-05
fix(table): use data file metrics in conflict checks

fallintoplace:
## Summary

Use data file metrics to refine concurrent overwrite conflict checks.

## Why

Conflict validation already pruned manifest and partition data, but it did not use file-level bounds.

## What changed

For an unpartitioned table, a concurrent file whose bounds cannot match the overwrite filter is ignored. Missing or uncertain metrics remain conservative and still report a possible conflict.

## Tests

- go test ./table -count=1

### DIFF
--- a/table/conflict_validation.go
+++ b/table/conflict_validation.go
@@ -367,7 +367,7 @@ func validateDataFilesExist(ctx *conflictContext, referencedPaths []string) erro
 //     straddles the filter only triggers a conflict when at least
 //     one actual added file's partition value satisfies the filter.
 //  3. The surviving files are evaluated against their column metrics so
-//     unpartitioned files whose bounds cannot match the filter are ignored.
+//     files whose bounds cannot match the filter are ignored.
 func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.BooleanExpression) error {
 	if len(ctx.concurrent) == 0 {
 		return nil
@@ -392,6 +392,10 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
 	partitionEvals := newKeyDefaultMapWrapErr(func(specID int) (func(iceberg.DataFile) (bool, error), error) {
 		return buildPartitionEvaluator(specID, ctx.current, ctx.current.CurrentSchema(), partitionFilters, ctx.caseSensitive)
 	})
+	// Eval copies each file's metrics into fresh evaluator state, so this
+	// evaluator can be reused across concurrent manifests.
+	// includeEmptyFiles=false is intentional: an empty file has no rows that
+	// can match the filter and must not create a conflict.
 	metricsEval, err := newInclusiveMetricsEvaluator(ctx.current.CurrentSchema(), filter, ctx.caseSensitive, false)
 	if err != nil {
 		return fmt.Errorf("failed to build metrics evaluator: %w", err)
--- a/table/conflict_validation.go
+++ b/table/conflict_validation.go
@@ -358,7 +358,7 @@ func validateDataFilesExist(ctx *conflictContext, referencedPaths []string) erro
 // predicate call this so that a concurrent append into the same
 // partition is rejected before the commit overwrites it.
 //
-// The check runs in two layers:
+// The check runs in three layers:
 //  1. A manifest-level partition-summary evaluator prunes manifests
 //     whose summaries cannot overlap the filter.
 //  2. Inside surviving manifests, every ADDED entry attributed to the
@@ -366,10 +366,8 @@ func validateDataFilesExist(ctx *conflictContext, referencedPaths []string) erro
 //     evaluator on its partition tuple, so a manifest whose summary
 //     straddles the filter only triggers a conflict when at least
 //     one actual added file's partition value satisfies the filter.
-//
-// Per-file metric evaluation (a third pass in Java that refines
-// beyond partition for columns not in the spec) is TODO and tracked
-// under issue #830 follow-ups.
+//  3. The surviving files are evaluated against their column metrics so
+//     unpartitioned files whose bounds cannot match the filter are ignored.
 func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.BooleanExpression) error {
 	if len(ctx.concurrent) == 0 {
 		return nil
@@ -394,6 +392,10 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
 	partitionEvals := newKeyDefaultMapWrapErr(func(specID int) (func(iceberg.DataFile) (bool, error), error) {
 		return buildPartitionEvaluator(specID, ctx.current, ctx.current.CurrentSchema(), partitionFilters, ctx.caseSensitive)
 	})
+	metricsEval, err := newInclusiveMetricsEvaluator(ctx.current.CurrentSchema(), filter, ctx.caseSensitive, false)
+	if err != nil {
+		return fmt.Errorf("failed to build metrics evaluator: %w", err)
+	}
 
 	for _, snap := range ctx.concurrent {
 		manifests, err := snap.Manifests(ctx.fs)
@@ -431,6 +433,13 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
 				matches, err := pEval(e.DataFile())
 				if err != nil {
 					return err
+				}
+				if !matches {
+					continue
+				}
+				matches, err = metricsEval(e.DataFile())
+				if err != nil {
+					return fmt.Errorf("evaluating metrics for data file %s: %w", e.DataFile().FilePath(), err)
 				}
 				if !matches {
 					continue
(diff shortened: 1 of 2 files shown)


table/conflict_validation.go
```diff
@@ -358,18 +358,16 @@ func validateDataFilesExist(ctx *conflictContext, referencedPaths []string) erro
...
 //     concurrent snapshot is evaluated against a per-spec partition
 //     evaluator on its partition tuple, so a manifest whose summary
 //     straddles the filter only triggers a conflict when at least
 //     one actual added file's partition value satisfies the filter.
-//
-// Per-file metric evaluation (a third pass in Java that refines
-// beyond partition for columns not in the spec) is TODO and tracked
-// under issue #830 follow-ups.
+//  3. The surviving files are evaluated against their column metrics so
+//     unpartitioned files whose bounds cannot match the filter are ignored.
```
zeroshade: Small wording nit. This bullet — and the PR description — says "unpartitioned files", but the code applies `metricsEval` to every file that survives the partition check, under every spec. The behavior is the right one; the wording just understates its scope. Consider dropping "unpartitioned" so the doc matches what the loop actually does.

table/conflict_validation.go
```diff
@@ -394,6 +392,10 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
 	partitionEvals := newKeyDefaultMapWrapErr(func(specID int) (func(iceberg.DataFile) (bool, error), error) {
 		return buildPartitionEvaluator(specID, ctx.current, ctx.current.CurrentSchema(), partitionFilters, ctx.caseSensitive)
 	})
+	metricsEval, err := newInclusiveMetricsEvaluator(ctx.current.CurrentSchema(), filter, ctx.caseSensitive, false)
```
zeroshade, line 399: Non-blocking: the trailing `false` is `includeEmptyFiles`, which means a concurrently added data file with `record_count == 0` hits the `rowsCannotMatch` early return (`table/evaluators.go:803-804`) and is silently skipped. That is defensible — there are no rows to lose — but it is currently neither documented nor tested, so the next reader has to derive it. Consider a short comment justifying the choice, or passing `true` if you'd rather keep this check maximally conservative.

table/conflict_validation.go
```diff
@@ -394,6 +392,10 @@ func validateAddedDataFilesMatchingFilter(ctx *conflictContext, filter iceberg.B
...
+	if err != nil {
+		return fmt.Errorf("failed to build metrics evaluator: %w", err)
+	}
 
 	for _, snap := range ctx.concurrent {
```
zeroshade, line 404: Worth a one-line comment noting that hoisting `metricsEval` out of this loop is safe because `Eval` copies its per-file state into a fresh struct on each call (`table/evaluators.go:807-811`). The hoist is correct, but the justification lives in another file.

table/conflict_validation_test.go
```diff
@@ -734,6 +734,73 @@ func TestValidateAddedDataFilesMatchingFilter_NoConcurrent(t *testing.T) {
 	require.NoError(t, validateAddedDataFilesMatchingFilter(ctx, nil))
 }
 
+func TestValidateAddedDataFilesMatchingFilterUsesFileMetrics(t *testing.T) {
+	tests := []struct {
+		name         string
+		lower, upper int64
+		wantConflict bool
+	}{
+		{name: "bounds cannot match", lower: 100, upper: 200, wantConflict: false},
```
zeroshade, line 743: The case I'd most like to see added here: a data file with **no** bounds at all, which must still report a conflict. That is the entire safety argument for this change — absent metrics fall through to `rowsMightMatch` — and it's the one property these three cases don't exercise. Non-blocking, but it's the highest-value addition to this table.

table/conflict_validation_test.go
```diff
@@ -734,6 +734,73 @@ func TestValidateAddedDataFilesMatchingFilter_NoConcurrent(t *testing.T) {
 	require.NoError(t, validateAddedDataFilesMatchingFilter(ctx, nil))
 }
 
+func TestValidateAddedDataFilesMatchingFilterUsesFileMetrics(t *testing.T) {
```
zeroshade, line 737: Further coverage worth picking up over time, roughly in priority order after the no-bounds case: `record_count == 0`; a file carrying bounds for other columns but not the filter column; NaN or float bounds; a null-count-only column paired with an `IsNull` filter; a truncated string upper bound; and a genuinely partitioned spec with a filter on a non-partition column — the helper below writes against `UnpartitionedSpec`, so the partitioned scenario the PR description highlights isn't reached yet. All suggestions; the three cases here do cover the core int64 path well.

zeroshade (commented):
The change is correct, and the property that matters most here holds. I checked the conservative handling of absent metrics directly:

- Every bound-based visitor guards on the bound actually being present — `table/evaluators.go:900`, `:932`, `:964`, `:995`, and `:1027` each test `m.lowerBounds[fieldID] != nil` / `m.upperBounds[fieldID] != nil` and fall through to `rowsMightMatch` when it is missing.
- `containsNullsOnly` (`table/evaluators.go:675-687`) returns `false` whenever `valueCounts` or `nullCounts` is absent, so a missing count never becomes a "cannot match".
- Together those mean a concurrently added file written with `write.metadata.metrics.default=none` still reports a conflict. That is the correct, conservative outcome and the thing I most wanted to confirm before this lands on the commit path.
- NaN bounds are treated as unreliable and fall through to `rowsMightMatch` (`table/evaluators.go:906`, `:1033`).
- Truncated bounds are only ever compared inclusively, and `VisitStartsWith` (`table/evaluators.go:1127`) truncates both sides to the prefix length before comparing — there is no exact comparison against a possibly-truncated bound.

One item worth flagging that sits outside this diff, so it goes here rather than inline: the inclusive metrics evaluator `panic`s on a bound it cannot decode (`table/evaluators.go:903`, `:967`, `:998`, `:1030`) instead of returning an error. A realistic trigger is an `int` → `long` promotion that leaves four-byte bounds on older files while `CurrentSchema()` reports `long`. The hazard predates this PR, but this PR is what newly exposes the *commit* path to it. Worth considering in a follow-up: either recover around the `metricsEval` call, or treat a decode failure as "might conflict", which would match the conservatism applied everywhere else here.

The inline notes are all optional.

zeroshade (approved):
Re-reviewed after `e4e2c77`. It picks up four of the five notes from my last pass: the doc no longer says "unpartitioned", both the hoist-safety and `includeEmptyFiles=false` justifications are now comments at the call site, and the missing-bounds case is covered. That last one was the item I cared about, so this is good to go.

I confirmed the new comments are accurate rather than taking them on faith — `Eval` does rebuild its per-file state into a fresh struct (`table/evaluators.go:802-822`), and `Count()` is `RecordCount` (`manifest.go:2271`), so the empty-file short-circuit is keyed on the right field.

**The new test is not vacuous.** I mutated `VisitEqual` to return `rowsCannotMatch` on a nil lower bound; `TestValidateAddedDataFilesMatchingFilterWithMissingBounds` fails with exactly its intended message while the three bounds-present subtests keep passing. It really does pin the conservative fall-through.

A few things I checked this round that weren't settled last time:

- **Retracting my `record_count == 0` test suggestion.** `NewDataFileBuilder` rejects `recordCount <= 0` (`manifest.go:2518`), so that branch can't be reached through the public builder at all — it only fires on manifests written by other engines. Constructing the test would mean hand-rolling a manifest entry, which isn't worth it. The comment is the right substitute here; please don't chase this one.
- **Truncated upper bounds are safe on the writer side.** `TruncateUpperBoundText` / `TruncateUpperBoundBinary` (`table/internal/utils.go:534,552`) increment the trailing rune/byte, and return empty/nil when they can't carry — in which case `MaxAsBytes` emits a nil bound and we fall back to conservative. So this repo cannot produce a truncated upper bound that under-reports. That was the one failure mode that would have turned layer 3 into silent data loss, and it's closed.
- **`IsNull` stays conservative.** `VisitIsNull` (`table/evaluators.go:843`) only returns cannot-match when the null count is explicitly present and zero. This matters because it's genuinely reachable: `eqDeletePartitionsToFilter` emits `IsNull(col)` for null partition values (`table/conflict_validation.go:578`).
- **Hoisting the bind can't introduce a new error class.** `eqDeletePartitionsToFilter` resolves names via `meta.CurrentSchema().FindFieldByID` (`table/conflict_validation.go:571`), so the filter is always bindable against the schema the evaluator binds to. One nuance worth naming: the bind now happens before the manifest loop, so a bind failure surfaces even when every concurrent snapshot carries only delete manifests. That's fail-fast and arguably an improvement.

Still open, and explicitly not a blocker for this PR: the inclusive metrics evaluator `panic`s on a bound it cannot decode (`table/evaluators.go:903, 967, 998, 1030`). An `int` → `long` promotion leaving four-byte bounds on older files while `CurrentSchema()` reports `long` still panics, and this PR is what puts that on the commit path. It predates the change and I flagged it as follow-up last time, so I'm not moving the goalpost — but I searched and there's no tracking issue for it (#830 is closed). I'll open one.

The remaining coverage wishlist from my last pass — bounds on other columns, NaN/float bounds, a null-count-only column with `IsNull`, a truncated string bound, a genuinely partitioned spec — is still worth picking up over time, but none of it blocks.

Verification on my side: `go test ./table/... ./ -count=1` all green, `gofmt -l table/` clean, `go vet ./table/...` clean.

LGTM.

PR #1672, opened 2026-08-05
fix(table): validate numeric Parquet writer properties

fallintoplace:
## Summary

Validate numeric Parquet writer properties before creating an output file.

## Why

Malformed, zero, negative, or out-of-range values were accepted by permissive parsing and then passed to Arrow.

## What changed

Row-group, page, dictionary, and bloom-filter sizes are checked before the writer is created. Existing defaults are unchanged.

## Tests

- go test ./table/internal -count=1

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -388,6 +388,13 @@ func ValidateParquetWriteProperties(props iceberg.Properties) error {
 		}
 	}
 
+	if value, ok := props[ParquetCompressionLevelKey]; ok {
+		if _, err := strconv.Atoi(value); err != nil {
+			return fmt.Errorf("%w: invalid %s value %q: %v", iceberg.ErrInvalidArgument,
+				ParquetCompressionLevelKey, value, err)
+		}
+	}
+
 	return nil
 }
 
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -73,6 +73,15 @@ const (
 	parquetBloomFilterMaxBytesMax            = 128 * 1024 * 1024
 	ParquetBloomFilterColumnEnabledKeyPrefix = "write.parquet.bloom-filter-enabled.column"
 
+	// These bounds mirror the compression libraries used by Arrow's Parquet
+	// writers. -1 is the shared default level and is included in each range.
+	parquetGzipCompressionLevelMin   = -3
+	parquetGzipCompressionLevelMax   = 9
+	parquetZstdCompressionLevelMin   = -5
+	parquetZstdCompressionLevelMax   = 22
+	parquetBrotliCompressionLevelMin = 0
+	parquetBrotliCompressionLevelMax = 11
+
 	// Deliberately not namespaced under write.parquet: this is the parquet-mr key
 	// Iceberg Java reads straight from the table properties, so a table carrying
 	// it gets the same dictionary behavior from either implementation.
@@ -358,6 +367,9 @@ func javaBool(val string) bool {
 }
 
 func ValidateParquetWriteProperties(props iceberg.Properties) error {
+	// A missing property uses the writer default, but an explicitly configured
+	// empty value is invalid. This keeps validation from silently accepting a
+	// value that GetInt would otherwise replace with its default.
 	for _, key := range []string{
 		ParquetRowGroupSizeBytesKey,
 		ParquetRowGroupLimitKey,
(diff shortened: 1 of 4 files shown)


table/internal/parquet_files.go
```diff
@@ -354,6 +357,47 @@ func javaBool(val string) bool {
...
+			(parsed < parquetBloomFilterMaxBytesMin || parsed > parquetBloomFilterMaxBytesMax) {
+			return fmt.Errorf("%w: %s must be between %d and %d bytes, got %d",
+				iceberg.ErrInvalidArgument, key, parquetBloomFilterMaxBytesMin, parquetBloomFilterMaxBytesMax, parsed)
+		}
+		if strconv.IntSize == 32 && parsed > int64(^uint(0)>>1) {
+			return fmt.Errorf("%w: %s value %d exceeds int range", iceberg.ErrInvalidArgument, key, parsed)
+		}
+	}
+
+	if value, ok := props[ParquetCompressionLevelKey]; ok {
```
zeroshade, line 406: `strconv.Atoi` only proves that the level is an integer; it still accepts invalid codec levels such as gzip 10 and zstd 23. Gzip can panic later, while zstd silently maps arbitrary values, so invalid configuration is not being rejected safely here.

Suggested fix: validate the level against the selected codec's supported range, preserve the default `-1`, and report the property name, codec, and valid range in the error.

table/internal/parquet_files.go
```diff
@@ -354,6 +357,47 @@ func javaBool(val string) bool {
...
+		}
+		if parsed <= 0 {
+			return fmt.Errorf("%w: %s must be greater than 0, got %d", iceberg.ErrInvalidArgument, key, parsed)
+		}
+		if key == ParquetBloomFilterMaxBytesKey &&
+			(parsed < parquetBloomFilterMaxBytesMin || parsed > parquetBloomFilterMaxBytesMax) {
+			return fmt.Errorf("%w: %s must be between %d and %d bytes, got %d",
+				iceberg.ErrInvalidArgument, key, parquetBloomFilterMaxBytesMin, parquetBloomFilterMaxBytesMax, parsed)
+		}
+		if strconv.IntSize == 32 && parsed > int64(^uint(0)>>1) {
```
zeroshade: This blanket 32-bit `int` ceiling also applies to `write.parquet.row-group-size-bytes`, even though that value remains `int64` through `ParquetRowGroupTargetSizeBytes`, `ParquetFileWriter.rowGroupBytes`, and the row-group comparison. On 32-bit builds this rejects legitimate `int64` row-group targets.

Suggested fix: exempt the row-group-size property from the `int` limit, or migrate all of its downstream consumers consistently if an `int` bound is actually required.

table/internal/parquet_files.go
```diff
@@ -354,6 +357,47 @@ func javaBool(val string) bool {
 	return strings.EqualFold(val, "true")
 }
 
+func ValidateParquetWriteProperties(props iceberg.Properties) error {
```
zeroshade, line 369: One constraint should deliberately remain absent: under Arrow/Parquet, page size does not need to be less than or equal to the row-group target.

Suggested fix: do not add that cross-property relationship as a hard validity rule; validate each property's actual consumer constraints instead.

table/internal/parquet_files.go
```diff
@@ -354,6 +357,47 @@ func javaBool(val string) bool {
...
+		ParquetPageRowLimitKey,
+		ParquetDictSizeBytesKey,
+		ParquetBloomFilterMaxBytesKey,
+	} {
+		value, ok := props[key]
+		if !ok {
+			continue
+		}
+
+		parsed, err := strconv.ParseInt(value, 10, 64)
```
zeroshade, line 386: An explicitly configured empty string now returns a parse error, whereas it previously selected the default. That may be the desired stricter behavior, but it is a compatibility change.

Suggested fix: document the change and add explicit tests distinguishing absent/unset properties, empty strings, and default-valued properties.

zeroshade (changes requested):
Thanks for centralizing numeric writer-property validation. The positive checks are useful, the bloom-filter bounds correctly include 32 bytes through 128 MiB, and there is no unsupported FPP property to validate.

Two blocking validation gaps remain inline: compression levels need codec-specific bounds, and the 32-bit `int` ceiling must not reject the row-group byte target that remains `int64` downstream. I also called out the empty-string compatibility change and the page-size relationship separately.

Please add codec boundary matrices with exact range errors, nil/unset/default cases, a 32-bit large-row-group case, and proof that invalid properties fail before file creation.

zeroshade (approved):

PR #1673, opened 2026-08-05
fix(table): close manifest lists before staging snapshots

### LINKED ISSUE #1666: snapshot producer corrupts a table by producing commits with missing manifest list file when getting blob storage error on deferred close
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

We have a streaming ingestion into iceberg that uses iceberg-go to write to iceberg tables and we saw repeated cases of tables getting corrupted due to snapshots referring to missing manifest list files (and potentially missing manifest files as well). 

At a high-level, what we're seeing in those cases of corrupted snapshots is:
* We create a transaction that does [AddFiles](https://pkg.go.dev/github.com/apache/iceberg-go@v0.6.0/table#Transaction.AddFiles)
* We call [transaction.Commit](https://pkg.go.dev/github.com/apache/iceberg-go@v0.6.0/table#Transaction.Commit) and get an error back. For example, we might get transient 503 errors on GCS that show up like this: 

```
blob (key "<redacted>/metadata/snap-1200212384152206267-0-deb76786-3779-4042-b0d2-8e321fcb1929.avro") (code=Unknown): googleapi: got HTTP response code 503 with body: Service Unavailable
```
* Our application retries by refreshing the table and trying the commit again and it succeeds. However, the failed commit still produced a snapshot pointing to a missing manifest list file and this causes queries / reads on that table to fail. The only path to recovery is restoring from a previous snapshot and replaying write operations.

## Investigation
I reviewed the implementation of [commitManifests](https://github.com/apache/iceberg-go/blob/main/table/snapshot_producers.go#L1160) and it looks like the issue has to do with how the errors on Close are all evaluated as deferred calls (example [here](https://github.com/apache/iceberg-go/blob/main/table/snapshot_producers.go#L1187) for the snapshot list writer:

```go
	out, err := sp.io.Create(manifestListFilePath)
	if err != nil {
		return nil, nil, err
	}
	defer internal.CheckedClose(out, &err)
...
                 err = iceberg.WriteManifestList(sp.txn.meta.formatVersion, out,
			sp.snapshotID, parentSnapshot, &nextSequence, firstRowID, newManifests)
		if err != nil {
			return nil, nil, err
		}
```

The implementation assumes that errors writing files will be returned on the write operations *but* the gocloud.dev blob writer [states](https://github.com/google/go-cloud/blob/master/blob/blob.go#L432-L436) that writes can't be guaranteed to be successful until Close returns without errors:
```go
// Close closes the blob writer. The write operation is not guaranteed to have succeeded until
// Close returns with no error.
// Close may return an error if the context provided to create the Writer is
// canceled or reaches its deadline.
func (w *Writer) Close() (err error)
```

This lines up with the behavior we're seeing where we get an error back from Commit because it happens on the deferred `Close` calls but the actual implementation had already gone through the full commit steps only checking the errors on `Write`.

fallintoplace:
## Summary

Close manifest-list writers before staging snapshot updates.

## Why

Manifest-list uploads can report their real failure from Close. The commit path built snapshot updates before that close ran, so a failed upload could leave a snapshot pointing at a missing manifest list.

## What changed

The manifest-list writer and underlying output are closed before snapshot updates are created. Close failures remove the incomplete object and return no updates or requirements. The retry rebuild path is covered too.

## Tests

- go test ./table -count=1

Fixes #1666

### DIFF
--- a/table/snapshot_producers.go
+++ b/table/snapshot_producers.go
@@ -1153,23 +1153,55 @@ func (sp *snapshotProducer) commit(ctx context.Context) (_ []Update, _ []Require
 	return sp.commitManifests(newManifests, addedContent)
 }
 
-func closeManifestListOutput(
+func writeManifestListFile(
 	fs iceio.WriteFileIO,
 	path string,
-	out iceio.FileWriter,
-	writer *iceberg.ManifestListWriter,
-	removeOnError bool,
-) error {
-	var err error
-	if writer != nil {
-		err = errors.Join(err, writer.Close())
-	}
-	err = errors.Join(err, out.Close())
-	if removeOnError || err != nil {
-		err = errors.Join(err, fs.Remove(path))
+	version int,
+	snapshotID int64,
+	sequenceNumber int64,
+	firstRowID int64,
+	parentSnapshotID *int64,
+	manifests []iceberg.ManifestFile,
+) (addedRows int64, err error) {
+	out, err := fs.Create(path)
+	if err != nil {
+		return 0, err
 	}
 
-	return err
+	defer func() {
+		if err != nil {
+			err = errors.Join(err, fs.Remove(path))
+		}
+	}()
+	defer internal.CheckedClose(out, &err)
+
+	var writer *iceberg.ManifestListWriter
+	switch version {
+	case 1:
+		writer, err = iceberg.NewManifestListWriterV1(out, snapshotID, parentSnapshotID)
+	case 2:
+		writer, err = iceberg.NewManifestListWriterV2(out, snapshotID, sequenceNumber, parentSnapshotID)
+	case 3:
+		writer, err = iceberg.NewManifestListWriterV3(out, snapshotID, sequenceNumber, firstRowID, parentSnapshotID)
+	default:
+		return 0, fmt.Errorf("unsupported manifest version: %d", version)
+	}
+
+	if writer != nil {
+		defer internal.CheckedClose(writer, &err)
+	}
+	if err != nil {
+		return 0, err
+	}
+
+	if err = writer.AddManifests(manifests); err != nil {
+		return 0, err
+	}
+	if version == 3 && writer.NextRowID() != nil {
+		addedRows = *writer.NextRowID() - firstRowID
+	}
+
+	return addedRows, nil
 }
 
 // commitManifests stages the snapshot for an already-built manifest set.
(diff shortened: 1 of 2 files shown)


table/snapshot_producers.go
```diff
@@ -1328,6 +1364,12 @@ func (sp *snapshotProducer) commitManifests(newManifests, addedContent []iceberg
 				return nil, fmt.Errorf("rebuild manifest list: write: %w", wErr)
 			}
 		}
+		if closeErr := closeManifestListOutput(fio, manifestListPath, out, writer, false); closeErr != nil {
+			closed = true
+
+			return nil, fmt.Errorf("rebuild manifest list: close: %w", closeErr)
+		}
+		closed = true
```
alexandre-normand: Maybe it's just me missing nuances and being naive but this seems like a lot of code. Why can't we just have the writer's closing be done after the `write` operation and return the error if there's one? After all, the contract with blob storage is essentially that a write is only guaranteed to be successful when `Close` has returned without an error. 

If _that_ approach worked, it seems like we could extend the same pattern to other paths that are prone to the same flaw/bug. I _think_ that some of those other paths would *not* have the same impact/severity nor would they all lead to corrupted snapshots so this might be more just about using the blob storage API like it's been intended to but it would help to promote better hygiene?

fallintoplace: Thanks, you were right that this could be simpler. I checked the current Java SnapshotProducer implementation and the relevant writer and IO contracts. Java closes the manifest-list writer before constructing the snapshot. In Go, the manifest-list writer closes the Avro writer, while the separate FileWriter still needs to be closed to finalize the blob.

alexandre-normand: Thanks, I think the updated version does look a lot better ❤️ .

zeroshade (commented):
The change is correct. I verified the two properties the fix turns on:

- **LIFO defer ordering.** The defers in `writeManifestListFile` (`table/snapshot_producers.go:1171-1192`) unwind in exactly the required order: the manifest-list/OCF writer closes first (registered last, `:1191`), the underlying object-store writer second (`:1176`), and the failure-cleanup defer runs last (`:1171-1175`). So the OCF close-time flush lands in the output before that output is closed — which is the bug being fixed.
- **Close errors are joined, not discarded.** Both closes go through `internal.CheckedClose` (`internal/utils.go:204-205`), which does `*err = errors.Join(*err, c.Close())` against the named return `err`. A close failure therefore propagates rather than being swallowed.

Following from those: nothing is staged on failure. The snapshot is not constructed until the helper returns successfully, on both the initial path (`:1241-1253`, with `snapshot := Snapshot{...}` at `:1255`) and the retry path (`:1358-1370`), so a close failure yields no snapshot and no update. The partial manifest list is removed and the removal error is preserved in the join. There is no double close either — `ManifestListWriter.Close` (`manifest.go:1656-1662`) closes only its internal OCF writer, and the `if writer != nil` guard keeps the nil case out of the defer.

The suggestions below are all about pinning this behavior harder in the tests. None of them indicate a defect.

table/snapshot_producers_test.go
```diff
@@ -719,6 +732,165 @@ func (t *trackingIO) GetWriterCount() int {
...
+				var output []byte
+				for _, writer := range fs.writers {
+					output = append([]byte(nil), writer.buf.Bytes()...)
+				}
+				fs.writersMu.Unlock()
+
+				require.Equal(t, 1, writerCount)
+				writtenManifests, readErr := iceberg.ReadManifestList(bytes.NewReader(output))
+				require.NoError(t, readErr)
+				require.Len(t, writtenManifests, 1, "manifest-list writer must flush before the output is closed")
```
zeroshade, line 846: This assertion proves the flush happened, which is the right thing to check — but it can't distinguish "flushed before the underlying close" from "flushed at all", since a buffer that keeps accepting writes after `Close()` satisfies it either way. To pin the ordering strictly, consider having `trackingWriteCloser.Write` return an error once `closed` is set, or recording an ordered event log across both layers and asserting inner-before-outer. Non-blocking — the defers are in the right order today; this would just keep them that way.

table/snapshot_producers_test.go
```diff
@@ -719,6 +732,165 @@ func (t *trackingIO) GetWriterCount() int {
...
+		trackingIO: newTrackingIO(),
+		closeErr:   closeErr,
+		removeErr:  removeErr,
+	}
+}
+
+func (c *closeErrorIO) Create(name string) (iceio.FileWriter, error) {
+	writer, err := c.trackingIO.Create(name)
+	if err == nil {
+		writer.(*trackingWriteCloser).closeErr = c.closeErr
```
zeroshade: Non-blocking observation about what this injects: the failure lands on the *underlying* output writer only. The scenario this fix specifically exists for — a failure surfacing during the OCF close-time flush while the outer `Close()` succeeds — isn't reachable through this helper, so it currently goes untested. Worth a follow-up variant that fails inside the manifest-list writer's own close, which would exercise the join from the inner layer.

table/snapshot_producers_test.go
```diff
@@ -719,6 +732,165 @@ func (t *trackingIO) GetWriterCount() int {
...
+func (c *closeErrorIO) Remove(name string) error {
+	c.removes = append(c.removes, name)
+	if c.removeErr != nil {
+		return c.removeErr
+	}
+
+	return c.trackingIO.Remove(name)
+}
+
+func TestCommitManifestsCloseFailureReturnsNoUpdates(t *testing.T) {
```
zeroshade, line 803: Worth adding while you're in here: an assertion that each writer layer is closed exactly once. `trackingWriteCloser` already tracks `closed` as a bool; making it a counter would let this test catch a double-close regression for free. Optional.

table/snapshot_producers_test.go
```diff
@@ -719,6 +732,165 @@ func (t *trackingIO) GetWriterCount() int {
...
+			manifest := writeTestManifestFileWithVersion(
+				t, fs, spec, simpleSchema(), sp.snapshotID, 1, manifestVersion,
+			)
+
+			updates, requirements, err := sp.commitManifests([]iceberg.ManifestFile{manifest}, nil)
+			require.ErrorIs(t, err, closeErr)
+			require.ErrorIs(t, err, removeErr)
+			require.Nil(t, updates)
+			require.Nil(t, requirements)
+			require.Len(t, fs.removes, 1)
```
zeroshade, line 829: This confirms a removal was attempted. The complementary case — that a *successful* cleanup leaves no partial manifest-list object behind — isn't covered anywhere. Consider a variant with `removeErr == nil` that asserts the path is actually gone from the backing store. Suggestion only.

zeroshade (approved):

PR #1675, opened 2026-08-06
feat(table): build transformed dynamic partition predicates

fallintoplace:
Summary

Build match predicates for dynamic overwrite partitions using each partition field transform.

Why

Partition values in data files are post-transform values. Direct source comparisons only work for identity transforms.

What changed

- Keep identity fields as raw references.
- Build transform(source) predicates for bucket, truncate, year, month, day, and hour.
- Preserve full nested source paths and recognize pointer-form identity transforms.
- Validate transform and source compatibility and fail closed for missing or impossible partition values.
- Handle null and void partitions.

Limitation

This is phase 1 of #1216. Transform-term binding landed in #1679. The current partial-file rewrite path still cannot execute non-identity predicates because source-column metrics are conservative and Substrait rejects transformed terms. The partition-level strict matching follow-up is tracked in #1215.

Tests

- go test ./table/internal -run TestBuildPartitionMatchPredicate -count=1
- go test ./table/internal ./table -count=1
- go test -race ./table/internal ./table

### DIFF
--- a/exprs.go
+++ b/exprs.go
@@ -1250,6 +1250,7 @@ func isUnknownTransform(transform Transform) bool {
 		return false
 	}
 }
+
 // rejectTransformTerm guards predicates that still do not support transform
 // terms, such as bounding-box predicates.
 func rejectTransformTerm(term BoundTerm) error {
--- a/exprs.go
+++ b/exprs.go
@@ -932,6 +932,7 @@ func (usp *unboundSetPredicate) Bind(schema *Schema, caseSensitive bool) (Boolea
 	if err != nil {
 		return nil, err
 	}
+
 	return createBoundSetPredicate(usp.op, bound, usp.lits)
 }
 
--- a/exprs.go
+++ b/exprs.go
@@ -932,7 +932,6 @@ func (usp *unboundSetPredicate) Bind(schema *Schema, caseSensitive bool) (Boolea
 	if err != nil {
 		return nil, err
 	}
-
 	return createBoundSetPredicate(usp.op, bound, usp.lits)
 }
 
@@ -1229,7 +1228,6 @@ func (u *UnboundTransform) Bind(schema *Schema, caseSensitive bool) (BoundTerm, 
 
 	return &BoundTransform{transform: u.transform, term: bound}, nil
 }
-
 func isNilTransform(transform Transform) bool {
 	if transform == nil {
 		return true
@@ -1252,7 +1250,6 @@ func isUnknownTransform(transform Transform) bool {
 		return false
 	}
 }
-
 // rejectTransformTerm guards predicates that still do not support transform
 // terms, such as bounding-box predicates.
 func rejectTransformTerm(term BoundTerm) error {
--- a/exprs.go
+++ b/exprs.go
@@ -1229,6 +1229,7 @@ func (u *UnboundTransform) Bind(schema *Schema, caseSensitive bool) (BoundTerm, 
 
 	return &BoundTransform{transform: u.transform, term: bound}, nil
 }
+
 func isNilTransform(transform Transform) bool {
 	if transform == nil {
 		return true
--- a/table/evaluators_test.go
+++ b/table/evaluators_test.go
@@ -3101,6 +3101,21 @@ func TestManifestEvaluatorKeepsTransformedTerms(t *testing.T) {
 	assert.True(t, keep)
 }
 
+func TestPrepareBatchFilterRejectsTransformedTerms(t *testing.T) {
+	schema := iceberg.NewSchema(1,
+		iceberg.NestedField{ID: 1, Name: "category", Type: iceberg.PrimitiveTypes.String},
+	)
+	term := iceberg.NewUnboundTransform(
+		iceberg.TruncateTransform{Width: 3},
+		iceberg.Reference("category"),
+	)
+
+	_, err := prepareBatchFilter(iceberg.EqualTo(term, "boo"), schema, true)
+	require.Error(t, err)
+	assert.ErrorIs(t, err, iceberg.ErrNotImplemented)
+	assert.Contains(t, err.Error(), "transformed terms")
+}
+
 func TestGetCmpLiteralRejectsGeo(t *testing.T) {
 	// Little-endian float64 X,Y - a valid geospatial single-value bound that
 	// must never be routed into an ordering comparison.
--- a/table/evaluators_test.go
+++ b/table/evaluators_test.go
@@ -3101,21 +3101,6 @@ func TestManifestEvaluatorKeepsTransformedTerms(t *testing.T) {
 	assert.True(t, keep)
 }
 
-func TestPrepareBatchFilterRejectsTransformedTerms(t *testing.T) {
-	schema := iceberg.NewSchema(1,
-		iceberg.NestedField{ID: 1, Name: "category", Type: iceberg.PrimitiveTypes.String},
-	)
-	term := iceberg.NewUnboundTransform(
-		iceberg.TruncateTransform{Width: 3},
-		iceberg.Reference("category"),
-	)
-
-	_, err := prepareBatchFilter(iceberg.EqualTo(term, "boo"), schema, true)
-	require.Error(t, err)
-	assert.ErrorIs(t, err, iceberg.ErrNotImplemented)
-	assert.Contains(t, err.Error(), "transformed terms")
-}
-
 func TestGetCmpLiteralRejectsGeo(t *testing.T) {
 	// Little-endian float64 X,Y - a valid geospatial single-value bound that
 	// must never be routed into an ordering comparison.
--- a/table/internal/partition_predicate.go
+++ b/table/internal/partition_predicate.go
@@ -49,12 +49,17 @@ import (
 //
 // The transform is kept in the row predicate so the match is evaluated against
 // the partition value rather than comparing the post-transform value directly
-// with the source column.
+// with the source column. This is phase 1 of issue #1216. The current overwrite
+// path cannot execute non-identity predicates for partial-file rewrites because
+// source-column metrics are conservative and the Substrait row-filter converter
+// rejects transformed terms; partition-level strict matching is still needed
+// before this helper can drive those rewrites (tracked in issue #1215).
 func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Schema, partitions []map[int]any) (iceberg.BooleanExpression, error) {
 	type fieldRef struct {
-		id        int
-		name      string
-		transform iceberg.Transform
+		id         int
+		name       string
+		transform  iceberg.Transform
+		resultType iceberg.Type
 	}
 
 	var fields []fieldRef
@@ -70,11 +75,17 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 			return nil, fmt.Errorf("%w: partition field %q references unknown source id %d",
 				iceberg.ErrInvalidArgument, f.Name, f.SourceIDs[0])
 		}
-		if _, err := iceberg.NewUnboundTransform(f.Transform, iceberg.Reference(sourceName)).Bind(schema, true); err != nil {
+		bound, err := iceberg.NewUnboundTransform(f.Transform, iceberg.Reference(sourceName)).Bind(schema, true)
+		if err != nil {
+			return nil, fmt.Errorf("partition field %q: %w", f.Name, err)
+		}
+		if _, err := f.Transform.MarshalText(); err != nil {
 			return nil, fmt.Errorf("partition field %q: %w", f.Name, err)
 		}
 
-		fields = append(fields, fieldRef{id: f.FieldID, name: sourceName, transform: f.Transform})
+		fields = append(fields, fieldRef{
+			id: f.FieldID, name: sourceName, transform: f.Transform, resultType: bound.Type(),
+		})
 	}
 
 	// NewAnd/NewOr fold away the AlwaysTrue/AlwaysFalse seeds, so a single-field,
@@ -94,18 +105,31 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 		for _, fr := range fields {
 			term := partitionTerm(fr.transform, fr.name)
 
-			// A missing key (!ok) is treated like an explicit nil. In practice
 			// DataFile.Partition() stores a null partition value as a nil-valued
-			// entry rather than an absent key; !ok is a defensive guard.
+			// entry. A missing field ID is malformed and must fail closed rather
+			// than silently under-deleting.
 			val, ok := part[fr.id]
-			if !ok || val == nil {
+			if !ok {
+				return nil, fmt.Errorf("%w: partition field %q (field id %d) is missing from partition tuple",
+					iceberg.ErrInvalidArgument, fr.name, fr.id)
+			}
+			if val == nil {
 				clause = iceberg.NewAnd(clause, iceberg.IsNull(term))
 				sigParts = append(sigParts, strconv.Itoa(fr.id)+":null")
 
 				continue
 			}
 
-			if isNaN(val) {
+			lit, err := LiteralForPartitionValue(val)
+			if err != nil {
+				return nil, fmt.Errorf("partition field %q: %w", fr.name, err)
+			}
+			lit, err = validatePartitionValue(fr.transform, fr.resultType, lit)
+			if err != nil {
+				return nil, fmt.Errorf("partition field %q: %w", fr.name, err)
+			}
+
+			if isNaN(lit.Any()) {
 				// x == NaN is never true (IEEE 754), so a NaN partition value must
 				// match via IsNaN. This intentionally diverges from PyIceberg's
 				// _build_partition_predicate, which emits EqualTo for every value:
@@ -119,11 +143,6 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 				sigParts = append(sigParts, strconv.Itoa(fr.id)+":nan")
 
 				continue
-			}
-
-			lit, err := LiteralForPartitionValue(val)
-			if err != nil {
-				return nil, fmt.Errorf("partition field %q: %w", fr.name, err)
 			}
 
 			clause = iceberg.NewAnd(clause, iceberg.LiteralPredicate(iceberg.OpEQ, term, lit))
@@ -152,6 +171,11 @@ func partitionTerm(transform iceberg.Transform, name string) iceberg.UnboundTerm
 	if isIdentityTransform(transform) {
 		return ref
 	}
+	if isVoidTransform(transform) {
+		// Use the value form so binding the null predicate can fold it to
+		// AlwaysTrue. Pointer forms are accepted by the Transform interface too.
+		return iceberg.NewUnboundTransform(iceberg.VoidTransform{}, ref)
+	}
 
 	return iceberg.NewUnboundTransform(transform, ref)
 }
@@ -165,6 +189,85 @@ func isIdentityTransform(transform iceberg.Transform) bool {
 	default:
 		return false
 	}
+}
+
+func isVoidTransform(transform iceberg.Transform) bool {
+	switch t := transform.(type) {
+	case iceberg.VoidTransform:
+		return true
+	case *iceberg.VoidTransform:
+		return t != nil
+	default:
+		return false
+	}
+}
+
+func isTruncateTransform(transform iceberg.Transform) bool {
+	switch t := transform.(type) {
+	case iceberg.TruncateTransform:
+		return true
+	case *iceberg.TruncateTransform:
+		return t != nil
+	default:
+		return false
+	}
+}
+
+func validatePartitionValue(transform iceberg.Transform, resultType iceberg.Type, lit iceberg.Literal) (iceberg.Literal, error) {
+	normalized, err := lit.To(resultType)
+	if err != nil {
+		return nil, fmt.Errorf("%w: partition value type %s cannot be converted to transform result type %s: %v",
+			iceberg.ErrInvalidArgument, lit.Type(), resultType, err)
+	}
+
+	switch normalized.(type) {
+	case iceberg.AboveMaxLiteral, iceberg.BelowMinLiteral:
+		return nil, fmt.Errorf("%w: partition value %s is outside transform result type %s",
+			iceberg.ErrInvalidArgument, normalized, resultType)
+	}
+
+	switch t := transform.(type) {
+	case iceberg.BucketTransform:
+		if err := validateBucketPartitionValue(t.NumBuckets, normalized); err != nil {
+			return nil, err
+		}
+	case *iceberg.BucketTransform:
+		if t == nil {
+			return nil, fmt.Errorf("%w: bucket transform cannot be nil", iceberg.ErrInvalidArgument)
+		}
+		if err := validateBucketPartitionValue(t.NumBuckets, normalized); err != nil {
+			return nil, err
+		}
+	case iceberg.VoidTransform, *iceberg.VoidTransform:
+		return nil, fmt.Errorf("%w: void transform only accepts a nil partition value", iceberg.ErrInvalidArgument)
+	}
+
+	if (isIdentityTransform(transform) || isTruncateTransform(transform)) && !isNaN(normalized.Any()) {
+		applied := transform.Apply(iceberg.Optional[iceberg.Literal]{Valid: true, Val: normalized})
+		if !applied.Valid || !applied.Val.Equals(normalized) {
+			return nil, fmt.Errorf("%w: partition value %s is not in the range of transform %s",
+				iceberg.ErrInvalidArgument, normalized, transform)
+		}
+	}
+
+	return normalized, nil
+}
+
+func validateBucketPartitionValue(numBuckets int, lit iceberg.Literal) error {
+	value, ok := lit.(iceberg.Int32Literal)
+	if !ok {
+		return fmt.Errorf("%w: bucket partition value must be int32, got %s", iceberg.ErrInvalidArgument, lit.Type())
+	}
+	if numBuckets <= 0 || numBuckets > math.MaxInt32 {
+		return fmt.Errorf("%w: bucket transform requires numBuckets in [1, %d], got %d",
+			iceberg.ErrInvalidArgument, math.MaxInt32, numBuckets)
+	}
+	if value.Value() < 0 || int64(value.Value()) >= int64(numBuckets) {
+		return fmt.Errorf("%w: bucket partition value %d is outside [0, %d)",
+			iceberg.ErrInvalidArgument, value.Value(), numBuckets)
+	}
+
+	return nil
 }
 
 // isNaN reports whether v is a floating-point NaN.
--- a/table/internal/partition_predicate.go
+++ b/table/internal/partition_predicate.go
@@ -65,19 +65,13 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 				iceberg.ErrInvalidArgument, f.Name, len(f.SourceIDs))
 		}
 
-		src, ok := schema.FindFieldByID(f.SourceIDs[0])
-		if !ok {
-			return nil, fmt.Errorf("%w: partition field %q references unknown source id %d",
-				iceberg.ErrInvalidArgument, f.Name, f.SourceIDs[0])
-		}
 		sourceName, ok := schema.FindColumnName(f.SourceIDs[0])
 		if !ok {
 			return nil, fmt.Errorf("%w: partition field %q references unknown source id %d",
 				iceberg.ErrInvalidArgument, f.Name, f.SourceIDs[0])
 		}
-		if !f.Transform.CanTransform(src.Type) {
-			return nil, fmt.Errorf("%w: transform %s cannot be applied to source field %q of type %s",
-				iceberg.ErrInvalidArgument, f.Transform, src.Name, src.Type)
+		if _, err := iceberg.NewUnboundTransform(f.Transform, iceberg.Reference(sourceName)).Bind(schema, true); err != nil {
+			return nil, fmt.Errorf("partition field %q: %w", f.Name, err)
 		}
 
 		fields = append(fields, fieldRef{id: f.FieldID, name: sourceName, transform: f.Transform})
--- a/table/internal/partition_predicate.go
+++ b/table/internal/partition_predicate.go
@@ -46,6 +46,9 @@ import (
 // Duplicate tuples collapse to a single clause, and an empty input yields
 // AlwaysFalse (matching nothing). Callers are expected to pass a partitioned
 // spec; dynamic partition overwrite rejects unpartitioned tables upstream.
+// Void fields are also accepted when their source column has been dropped (the
+// spec represents those tombstones with source ID 0), because void always
+// produces a null partition value and does not need a source column.
 //
 // The transform is kept in the row predicate so the match is evaluated against
 // the partition value rather than comparing the post-transform value directly
@@ -60,6 +63,7 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 		name       string
 		transform  iceberg.Transform
 		resultType iceberg.Type
+		isVoid     bool
 	}
 
 	var fields []fieldRef
@@ -68,6 +72,18 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 		if len(f.SourceIDs) != 1 {
 			return nil, fmt.Errorf("%w: partition field %q must have exactly one source id, got %d",
 				iceberg.ErrInvalidArgument, f.Name, len(f.SourceIDs))
+		}
+		if isVoidTransform(f.Transform) {
+			// A void field can survive source-column removal as a source-less
+			// tombstone. Its output is always null, so resolving or binding the
+			// source would add no information and would reject source ID 0.
+			if _, err := f.Transform.MarshalText(); err != nil {
+				return nil, fmt.Errorf("partition field %q: %w", f.Name, err)
+			}
+
+			fields = append(fields, fieldRef{id: f.FieldID, name: f.Name, transform: f.Transform, isVoid: true})
+
+			continue
 		}
 
 		sourceName, ok := schema.FindColumnName(f.SourceIDs[0])
@@ -103,8 +119,6 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 		sigParts := make([]string, 0, len(fields))
 
 		for _, fr := range fields {
-			term := partitionTerm(fr.transform, fr.name)
-
 			// DataFile.Partition() stores a null partition value as a nil-valued
 			// entry. A missing field ID is malformed and must fail closed rather
 			// than silently under-deleting.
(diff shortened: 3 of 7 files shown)


zeroshade:

exprs.go
```diff
@@ -790,7 +788,7 @@ type BoundLiteralPredicate interface {
 
 func newBoundLiteralPredicate[T LiteralType](op Operation, term BoundTerm, lit Literal) BoundPredicate {
 	return &boundLiteralPredicate[T]{
-		op: op, term: term.(bound[T]),
+		op: op, term: term,
```
zeroshade, line 791: **Blocking:** accepting a `BoundTransform` here exposes it to the metrics and bloom evaluators, which call `term.Ref()` and compare the transformed literal against raw source-column bounds/bloom values. For example, `truncate[3](category) == "boo"` with a file containing `"books"` is incorrectly pruned because the inclusive metrics evaluator compares raw `"books"` with `"boo"`, even though the transformed predicate matches. This can skip matching files during scans and dynamic overwrite. Please make all source-column pruning evaluators conservatively handle transformed terms (or correctly transform metrics where safe); strict metrics and bloom pruning need the same audit.

exprs.go
```diff
@@ -838,7 +836,7 @@ func createBoundLiteralPredicate(op Operation, term BoundTerm, lit Literal) (Bou
 
 type boundLiteralPredicate[T LiteralType] struct {
 	op   Operation
-	term bound[T]
+	term BoundTerm
```
zeroshade, line 839: **Blocking:** the Substrait row-filter converter is not transform-aware. It converts predicates through `Term().Ref()` and therefore emits `equal(.field(0), "boo")` for `truncate[3](category) == "boo"`, silently dropping the transform. Partial-file rewrites can consequently retain rows that should be replaced, and bucket/time transforms may also produce type-invalid filters. Please encode the transform in Substrait or reject transformed predicates on this path rather than changing their semantics.

zeroshade (changes requested):
Thanks for adding transformed-term binding support. I found two downstream consumers that still interpret the predicate as if its term were a raw field reference. Both can change filtered overwrite semantics, so I think they need to be addressed before merge. The full module suite and all CI checks otherwise pass, and I did not find coding-style issues.

exprs.go
```diff
@@ -1121,7 +1124,17 @@ func (u *UnboundTransform) Equals(other UnboundTerm) bool {
 }
 
 func (u *UnboundTransform) Bind(schema *Schema, caseSensitive bool) (BoundTerm, error) {
-	bound, err := u.term.Bind(schema, caseSensitive)
+	if isUnknownTransform(u.transform) {
```
zeroshade, line 1209: `NewUnboundTransform` is exported and accepts a `Transform` interface, so callers can pass `nil`. In that case this check falls through and `u.transform.CanTransform(...)` panics instead of returning an error. I reproduced this by binding `EqualTo(NewUnboundTransform(nil, Reference("x")), int32(1))`. Please reject a nil transform here with an appropriate error (for example `ErrInvalidArgument`) and add regression coverage. This is the same issue currently present in the stacked dependency #1679.

zeroshade (changes requested):
Thanks—the two previously reported correctness issues are addressed. Source-column evaluators now remain conservative for transformed predicates, and Substrait conversion rejects unsupported transformed row filters rather than changing their semantics.

One blocker remains in the stacked transform-term implementation: `NewUnboundTransform(nil, ref)` panics during binding instead of returning an error. This is also tracked on dependency PR #1679; please incorporate that fix here when updating the stack.

Please also add the required DCO sign-offs to all commits and reference the associated dynamic-overwrite issue.

---

table/internal/partition_predicate.go
```diff
@@ -77,8 +70,12 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 			return nil, fmt.Errorf("%w: partition field %q references unknown source id %d",
 				iceberg.ErrInvalidArgument, f.Name, f.SourceIDs[0])
 		}
+		if !f.Transform.CanTransform(src.Type) {
+			return nil, fmt.Errorf("%w: transform %s cannot be applied to source field %q of type %s",
+				iceberg.ErrInvalidArgument, f.Transform, src.Name, src.Type)
+		}
 
-		fields = append(fields, fieldRef{id: f.FieldID, name: src.Name})
+		fields = append(fields, fieldRef{id: f.FieldID, name: src.Name, transform: f.Transform})
```
zeroshade: `FindFieldByID` returns the field's leaf name, so a partition source such as `location.category` is stored as `category`. The resulting `truncate(category)` predicate cannot bind to the schema; I reproduced the `invalid schema: could not bind reference 'category'` failure 10/10 times. If a top-level field shares that leaf name, this can instead bind the wrong source and select the wrong rows for overwrite.

Please obtain the full path with `schema.FindColumnName(f.SourceIDs[0])` while retaining `src` for type validation, and add nested transformed-source coverage.

table/internal/partition_predicate.go
```diff
@@ -151,6 +148,15 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 	return result, nil
 }
 
+func partitionTerm(transform iceberg.Transform, name string) iceberg.UnboundTerm {
+	ref := iceberg.Reference(name)
+	if _, ok := transform.(iceberg.IdentityTransform); ok {
```
zeroshade: Pointer forms of built-in transforms satisfy `Transform` and are accepted elsewhere, but this assertion recognizes only the value form. With `&IdentityTransform{}`, the builder emits `identity(id) == 5` instead of the promised unchanged `id == 5`; reproduced 10/10 times. That unnecessarily enters the transformed-filter path, which `prepareBatchFilter` rejects.

Please recognize both pointer and value identity transforms and add a regression. This is adjacent to, but distinct from, the pointer-transform equality issue on dependency PR #1679.

zeroshade (changes requested):
The previously reported pruning, Substrait, nil-transform, DCO, and issue-reference findings are addressed, and the updated branch passes the affected validation suites and latest-main integration.

Two dynamic-partition predicate issues remain inline: nested source columns lose their full schema path, and pointer-form identity transforms are incorrectly emitted as transformed terms. Both can prevent the generated overwrite predicate from following the intended path; the nested case can also bind the wrong field when leaf names collide.

---

>

zeroshade (changes requested):
The two prior findings that still applied to this PR's files are fixed — nested source paths now resolve the full schema path via `FindColumnName`, and pointer-identity transforms emit a raw reference. My other earlier findings on this work (metrics/bloom transformed-term handling, Substrait rejection, nil/typed-nil transforms) landed through separate merged work and are no longer in scope for this diff. What remains is test coverage.

### Major — transformed predicate behaviour lacks execution-level coverage (`table/internal/partition_predicate_test.go:279`)

The new test only compares the generated expression tree for `bucket` and `truncate`; it never binds and evaluates the predicate against rows, and there are no cases at all for `year`, `month`, `day`, or `hour`.

That means regressions in transform application or result literal types — `day`'s `DateLiteral` vs `int32` output, negative-timestamp floor behaviour, timezone handling for `hour` — or in operand coercion can pass CI while dynamic overwrite deletes the wrong rows. Given that the consequence is destructive and silent, AST-shape assertions alone aren't enough.

Please add bind + `ExpressionEvaluator` tests per transform, with matching and non-matching source values:

- `bucket` — including a bucket collision (two distinct sources hashing to the same bucket).
- `truncate` — same-prefix and different-prefix values.
- `year` / `month` / `day` / `hour` — boundary values on each side of the truncation unit.
- Null handling for each transform.

An end-to-end overwrite-deletion assertion would be ideal, though I understand that belongs with the caller.

### Context, not a finding

The helper currently has no production caller — this is phase-1 groundwork for the `DynamicPartitionOverwrite` caller planned under #1215. That's the reason I'm not asking for the end-to-end overwrite test in this PR, and it's also what keeps the coverage gap low-risk today. The execution-level helper tests are still needed before anything relies on this.

The transform/value construction itself reads as spec-correct: identity uses a direct reference, the others compare `transform(source)` against the post-transform partition value, nulls go through `IsNull(term)`, and NaN is handled only for identity. CI is green.

---

>

zeroshade (changes requested):
The transform/value construction is spec-correct and the test suite is genuinely strong — 14 of 18 mutations killed, and every code fix from my earlier rounds is pinned by a mutation-verified test. Two things hold it: the null-path coverage I asked for last round provably cannot fail, and the predicate this emits can't actually be executed by the repo's overwrite path.

**First, the failing check is not yours.** `ubuntu-latest go1.25.9` fails at `catalog/sql/sql_test.go:2378` (`TestConcurrentTableViewCollisionReturnsCatalogSentinel`) — the pre-existing tracked flake in #1793 ("hits SQLITE_BUSY on ubuntu runners"). The same commit passes `ubuntu go1.26.1`, both macOS jobs and both Spark integration suites; locally 8/8 focused and 3/3 full runs of `catalog/sql` pass under `-race`. Nothing this PR touches is reachable from `catalog/sql`.

## Blocking

**1. `table/internal/partition_predicate_test.go:533` — the per-transform null test is inert, and `void` is entirely uncovered.**

Last round I asked for "null handling for each transform." `TestBuildPartitionMatchPredicate_EvaluatesTransformedNull` was added with bucket/truncate/year/month/day/hour cases — and it cannot fail when the thing it protects is broken.

Mutation: change `partition_predicate.go:102` from `iceberg.IsNull(term)` to `iceberg.IsNull(iceberg.Reference(fr.name))` — drop the transform from exactly the path this test covers. The entire `table/internal` suite stays green. It survives because for every transform in the table `t(null)=null` and `t(non-null)=non-null`, so `IsNull(t(x)) ≡ IsNull(x)`.

The one transform where they differ is **`void`**, which appears nowhere in the file. That matters: `void` puts every row in one partition whose stored value is always null, so `IsNull(void(src))` must match **all** rows. It correctly binds to `AlwaysTrue` (`exprs.go:592-596`) — a void-partitioned dynamic overwrite replaces everything. With the transform dropped it becomes `IsNull(category)`: only null rows get deleted, every non-null row survives the delete and is then re-appended → **duplicate rows, silently**.

A one-line addition closes it. Adding to the case table:

```go
{name: "void", sourceType: iceberg.PrimitiveTypes.String,
 transform: iceberg.VoidTransform{}, nonNil: "books", nonNilMatches: true},
```

(and asserting `tc.nonNilMatches` rather than a hard-coded `assert.False`) turns the mutation from surviving into:

```
--- FAIL: TestBuildPartitionMatchPredicate_EvaluatesTransformedNull/void
    expected: true
    actual  : false
```

Please add that, plus an assertion that a void-only spec's predicate binds to `AlwaysTrue`. That's a delete-the-whole-table expression coming out of a destructive-path builder — it deserves an explicit named test rather than being an emergent property.

**2. `table/internal/partition_predicate.go:95` + the PR title — the emitted predicate isn't executable by this repo's overwrite path for any non-identity transform.**

Probed against the real classification path, `truncate[3](category) == "boo"` versus a file wholly inside that partition (bounds `books..boots`):

```
generated: Equal(term=UnboundTransform(transform=truncate[3], term=Reference(name='category')), literal=boo)
inclusive (file might match)                 = true
strict (all rows match -> whole-file delete) = false
prepareBatchFilter(complement) err           = prepareBatchFilter: convert expression:
                                               not implemented: transformed terms cannot be converted to substrait
```

Identity control, same shape: `strict = true`, `prepareBatchFilter err = <nil>`.

The chain: `strictMetricsEval.VisitBound` returns `rowsMightNotMatch` for any `*BoundTransform` (`evaluators.go:1400-1402`) → `classifyFilesForFilteredDeletions` never reaches `filesToDelete` (`transaction.go:2701-2704`) → every candidate goes to `filesWithPartialDeletes` → the rewrite calls `prepareBatchFilter` (`transaction.go:2756`) → `substrait.go:245-247` rejects it. So the first caller wired to this helper gets a hard `ErrNotImplemented` for **100% of non-identity specs**.

It fails closed — no data loss — and that conservatism is itself one of my earlier fixes. But "match transformed partitions in dynamic overwrite" describes a capability the tree can't execute, and the doc says nothing about it. My last review noted "no production caller yet"; what it didn't establish is that the downstream is *incapable*, which changes the plan for #1215.

The missing piece is a **strict partition-level** evaluator. The inclusive partition projection already handles transformed terms fine today — `newInclusiveProjection(sch, spec, true)(expr)` → `Equal(term=Reference(name='category_trunc'), literal=boo)`. There's no `strictProjection` anywhere in the tree (grep: 0 hits). Whole-file deletion for a transformed predicate should be decided by projecting onto partition values, not by strict metrics on source columns.

Ask: document the limitation on `BuildPartitionMatchPredicate` and reference it from #1215. Cheap to satisfy, expensive to rediscover at runtime.

## Major

**3. `partition_predicate.go:100-106` — a missing partition key silently degrades to `IsNull`, i.e. under-deletion with no error.**

```go
val, ok := part[fr.id]
if !ok || val == nil {
    clause = iceberg.NewAnd(clause, iceberg.IsNull(term))
```

The comment calls `!ok` "a defensive guard," but the guard is *permissive*. Probe with a tuple keyed by a field ID absent from the spec:

```
STALE-FIELD-ID tuple -> predicate=IsNull(term=UnboundTransform(truncate[3], Reference('category')))  err=<nil>
  -> matches the real 'books' row? false
```

No error; the data that must be replaced isn't selected; the append then duplicates it. That's exactly the "too few and you get duplicate rows" failure mode, and it's most reachable under partition evolution — a caller harvesting tuples from *existing* files (which carry old-spec field IDs) rather than newly written ones degrades **every** field to `IsNull` in silence.

Pre-existing (base has the same condition), but this is the partition-matching-correctness PR and #1215's caller will be written against this contract. Fix: return `ErrInvalidArgument` when the key is absent; reserve `IsNull` for an explicit `nil`. Two lines, converting a silent wrong answer into a fail-closed error. Note the current behaviour *is* pinned by `TestBuildPartitionMatchPredicate_NullValue`, so this needs a deliberate change rather than a drive-by.

**4. `partition_predicate.go:124-135` — no post-condition check that the partition value lies in the transform's range.** Values this spec could never have produced are accepted and yield a predicate matching zero rows:

```
bucket[4]    partition value int32(7):  err=<nil>  -> matched   0/200 rows
truncate[10] partition value int32(7):  err=<nil>  -> matched   0/200 rows
truncate[3]  partition value "books":   err=<nil>  -> matched 0 of {books, boo, boots}
```

The third is the realistic partition-evolution case: `"books"` is what a prior `truncate[5]` spec stored, and `truncate[3](x)` can never equal it. Silently deletes nothing. Cheap guards in increasing generality: assert the literal's type equals `transform.ResultType(sourceType)`; for `truncate`/`identity` assert the value is a fixed point (`transform.Apply(lit).Val.Equals(lit)`); for `bucket` assert `0 <= v < NumBuckets`. Given this is the core of a destructive operation with no caller yet to enforce the contract, validating its own inputs is worth the few lines.

## What checks out

Per-transform end-to-end probes, all exact: identity; bucket[4] including a genuine collision (wrong-typed value fails closed at bind); truncate[3] (`boo`/`books`/`booster` match, `bar`/`science` don't); year/month/day/hour on both sides of every boundary including pre-epoch hour `−1`; `day` literal typing on **both** decodes (real `DataFile.Partition()` → `iceberg.Date` → `DateLiteral`, and raw `int32(18322)` coerced to `DateLiteral 2020-03-01`); multi-field spec (identity+bucket+day) across 2 partitions, **7/7 with no false cross-product matches**; row-level partition evolution 5/5 (a new `truncate[3]` spec correctly matches old identity-spec rows `books`/`boots`/`bookkeeper` and rejects `bar`).

Fail-closed matrix: unknown transform → `ErrNotImplemented`; nil and typed-nil → `ErrInvalidArgument`; day-on-string, year-on-string, hour-on-int, truncate-on-timestamp, bucket-on-double → `ErrInvalidArgument`. All fail closed.

Build clean; `./table/internal` and `./table` pass with `-race`; `gofmt` clean; `golangci-lint` 0 issues. True diff vs base `87789102` is 3 files, +401/−33 — the four `exprs.go` commits net to zero and that file is byte-identical to base.

## Minor

- **`0383df14` is missing its DCO sign-off.** The other 8 commits have one; `CONTRIBUTING.md:40` requires it on all. My round-2 review asked for exactly this, so it's a regression on an addressed item, and no DCO status check is configured so CI won't catch it.
- `partition_predicate_test.go:47` — `identitySpec(...)` is now used to build bucket/truncate/year/month/day/hour specs throughout. The name actively misleads; `specWithFields` would be honest.
- `partition_predicate.go:118` — the `IsNaN(term)` → `IsNaN(raw ref)` mutation survives, *correctly*: the NaN branch is only reachable when the result type is float (identity, void), and void always stores null. Flagging so the surviving mutation isn't later mistaken for a gap. I did confirm the comment's justification is true rather than folklore: `IsNaN(score)` gives `strict = true` on a NaN-only file while `EqualTo(score, NaN)` gives false.
- `partition_predicate.go:89-92`,`:135` — the comment says the `(field id, value bytes)` signature is what prevents tuple collisions, but the field-id component is provably redundant (`sigParts` is built in fixed `fields` order with hex-encoded values; replacing `fr.id` with a constant survives the full suite). Harmless, but the comment overstates what's load-bearing.
- `partition_predicate.go:63-66` — the `len(f.SourceIDs) != 1` guard is untested; deleting it survives. Pre-existing; without it an empty `SourceIDs` panics.
- `partition_predicate.go:50-52` — doc wording is muddled. Suggest: "the transform is retained in the row predicate so the post-transform partition value is compared against `transform(source)`, not against the raw source column."
- `partition_predicate.go:73` — the new validation binds `NewUnboundTransform(IdentityTransform{}, ref)` for identity fields too, and `IdentityTransform.CanTransform` rejects `Geometry`/`Geography`/`Variant`. Identity partitions on those types now error where they previously produced `ref == value`. Right direction, but an unmentioned behaviour change.
- Range-over-int respected; no new files.

## Prior items

**Round 1 (08-21):** metrics/bloom evaluators conservative for transformed terms → **Fixed** via #1679; the `isTransformedTerm` guards are in place at `evaluators.go:601`, `:895`, `:1400`, `:1826`, and the false-prune from my original comment no longer occurs. Substrait converter silently dropping the transform → **Fixed**, and this PR adds the regression; the test is live (deleting the rejection fails it).

**Round 2 (08-26):** `NewUnboundTransform(nil, ref)` panicking on Bind → **Fixed**, mutation-verified. DCO on all commits → **Partially fixed, regressed** (above). Reference the issue → **Fixed.**

**Round 3 (08-26):** nested source losing its full schema path → **Fixed**, and the test is properly adversarial — it uses a colliding top-level `category` so the wrong-field variant is caught; two separate mutations are killed by it. Pointer identity emitted as a transformed term → **Fixed**, mutation-verified.

**Round 4 (09-01):** bucket incl. collision → **Fixed** (the collision test discovers two distinct sources in one bucket at runtime). truncate same/different prefix → **Fixed.** year/month/day/hour boundaries → **Fixed**, including the pre-epoch negative-floor case I was worried about; I independently checked the arithmetic (year 50, month 601, hour 439714 = 18321·24+10). `day`'s `DateLiteral` vs `int32` → **Fixed, verified on both decodes.** Null handling per transform → **Partially fixed, inert** (Blocking 1). "An end-to-end overwrite-deletion assertion would be ideal" → **Still open**, and Blocking 2 explains why it currently can't be written. "The transform/value construction reads as spec-correct" → **confirmed independently.**

## Description

- **Stale dependency claim.** "intended to be bound by the transform-term predicate support in #1679" — #1679 merged 2026-08-26, and `exprs.go` here is byte-identical to base. Should read "landed in #1679".
- **Overstated title.** "match transformed partitions in dynamic overwrite" is true of the *predicate builder*, false of dynamic overwrite as a whole. This is phase 1 of #1216, not #1215.
- **Understated changes.** "add bucket and truncate coverage" omits year/month/day/hour, bucket collision, transformed null, nested source path, pointer identity, and the invalid-transform matrix.
- **Incomplete test command.** The listed `go test ./table/internal -run TestBuildPartitionMatchPredicate` doesn't run `TestPrepareBatchFilterRejectsTransformedTerms`, which this PR adds.

---

table/internal/partition_predicate.go
```diff
@@ -151,6 +196,110 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 	return result, nil
 }
 
+func partitionTerm(transform iceberg.Transform, name string) iceberg.UnboundTerm {
+	ref := iceberg.Reference(name)
+	if isIdentityTransform(transform) {
+		return ref
+	}
+	if isVoidTransform(transform) {
+		// Use the value form so binding the null predicate can fold it to
```
zeroshade, line 205: **minor** — partitionTerm's void branch is unreachable dead code with a misleading comment

partitionTerm is called from only two sites (:141 and :155), both reached only when fr.isVoid is false, and fr.isVoid is set exactly when isVoidTransform(f.Transform) is true. The void branch at :204-208 can therefore never execute. Its comment ('Use the value form so binding the null predicate can fold it to AlwaysTrue') describes an outcome that never occurs, because void fields never produce a predicate at all — they are skipped at :136-142. Delete the branch and the comment.

table/internal/partition_predicate.go
```diff
@@ -151,6 +196,110 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
...
+		}
+	case *iceberg.BucketTransform:
+		if t == nil {
+			return nil, fmt.Errorf("%w: bucket transform cannot be nil", iceberg.ErrInvalidArgument)
+		}
+		if err := validateBucketPartitionValue(t.NumBuckets, normalized); err != nil {
+			return nil, err
+		}
+	case iceberg.VoidTransform, *iceberg.VoidTransform:
+		return nil, fmt.Errorf("%w: void transform only accepts a nil partition value", iceberg.ErrInvalidArgument)
```
zeroshade, line 272: **minor** — Three more unreachable defensive branches in validatePartitionValue / validateBucketPartitionValue

(a) :271-272 'void transform only accepts a nil partition value' cannot fire — the caller at :150-153 already returns an error for every void field with a non-nil value. (b) :265-267 '*BucketTransform cannot be nil' cannot fire — UnboundTransform.Bind at :95 rejects typed-nil pointers via isNilTransform first. (c) :291-294 'numBuckets in [1, MaxInt32]' cannot fire — f.Transform.MarshalText() at :98 already calls BucketTransform.validateNumBuckets, which rejects the same two conditions. Drop these and rely on the single checks that already exist, so the reader is not led to believe there are two independent validation layers.

table/internal/partition_predicate_test.go
```diff
@@ -276,38 +290,406 @@ func TestBuildPartitionMatchPredicate_NoSeparatorCollision(t *testing.T) {
...
+
+			matched, err = eval(partitionPredicateRow{tc.nonNil})
+			require.NoError(t, err)
+			assert.Equal(t, tc.nonNilMatches, matched, "unexpected match for non-null source value")
+		})
+	}
+}
+
+func TestBuildPartitionMatchPredicate_VoidBindsAlwaysTrue(t *testing.T) {
+	spec := specWithFields(iceberg.PartitionField{
```
zeroshade, line 572: **minor** — TestBuildPartitionMatchPredicate_VoidBindsAlwaysTrue is vacuous

The test builds a void-only spec, calls BindExpr, and asserts the result equals AlwaysTrue{} — but BuildPartitionMatchPredicate already returns the concrete value iceberg.AlwaysTrue{} before BindExpr runs, so BindExpr is a no-op identity and the assertion cannot fail for any reason related to void handling. The same applies to the {name: "void", ..., nonNilMatches: true} case at :542, which only asserts that AlwaysTrue matches every row. Either rename to reflect what is actually covered (a source-less void field contributes no clause) or drop the BindExpr step.

table/internal/partition_predicate.go
```diff
@@ -39,46 +39,69 @@ import (
 // result is an OR across distinct partitions, each clause an AND across the
 // spec's fields:
 //
-//	source == value   when the partition value is present
-//	IsNaN(source)      when the value is a floating-point NaN (x == NaN is never true)
-//	IsNull(source)     when the partition value is absent or nil
+//	transform(source) == value when the partition value is present
+//	IsNaN(transform(source)) when the value is a floating-point NaN (x == NaN is never true)
+//	IsNull(transform(source)) when the partition value is absent or nil
 //
```
zeroshade, line 45: **minor** — Doc comment contradicts the new fail-closed behaviour for absent partition keys

Line 44 still documents 'IsNull(transform(source)) when the partition value is absent or nil', but :127-130 now returns ErrInvalidArgument when the field id is absent from the tuple — that is the headline behaviour change of commit 28628f7 and the doc was not updated with it. Fix the comment to say absent keys are rejected.

zeroshade (commented):
All five prior findings are genuinely fixed and the transformed-predicate semantics verify correct under probe testing, but the new validation code carries ~25 lines of provably unreachable branches and one vacuous void test.

Re-review verification: 5 of 5 prior findings confirmed fixed at `9077e16` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust).

```text
go build ./... (PASS); go vet ./table/... (clean); go test ./table/internal/ -count=1 (PASS); go test ./table/ -count=1 (PASS, 6.8s); go test ./table/substrait/ -count=1 (PASS); go test ./table/internal -run TestBuildPartitionMatchPredicate -coverprofile (PASS, used to prove unreachable branches); 14 throwaway probe tests in table/internal/pr1675_probe_test.go covering identity over 16 primitive-type/value shapes, truncate on int32/int64/binary/decimal/unicode, bucket on string/uuid/decimal/date, year/month/day on a Date source, overflow guards, dedup partial-failure, duplicate field ids, and pointer-void handling — all PASS, no defect found; probe file deleted, worktree clean.
```

</details>

---

>

table/internal/partition_predicate.go
```diff
@@ -151,6 +196,110 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
...
+		return true
+	case *iceberg.IdentityTransform:
+		return t != nil
+	default:
+		return false
+	}
+}
+
+func isVoidTransform(transform iceberg.Transform) bool {
+	switch t := transform.(type) {
```
zeroshade, line 225: **minor** — Pointer-form void is accepted here but not by the upstream IsUnpartitioned guard the doc relies on

The doc comment at :48 leans on 'dynamic partition overwrite rejects unpartitioned tables upstream', but PartitionSpec.IsUnpartitioned (partitions.go:756) matches only the value form VoidTransform, while isVoidTransform here deliberately accepts *VoidTransform too. A spec whose only field is &VoidTransform{} therefore passes the upstream 'is partitioned' guard and yields AlwaysTrue from this builder — a whole-table match. Not reachable from parsed metadata (ParseTransform returns value forms only), so this is hand-constructed-input territory; either narrow isVoidTransform to the value form or widen IsUnpartitioned so the two agree.

table/internal/partition_predicate.go
```diff
@@ -151,6 +196,110 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
...
+		}
+		if err := validateBucketPartitionValue(t.NumBuckets, normalized); err != nil {
+			return nil, err
+		}
+	case iceberg.VoidTransform, *iceberg.VoidTransform:
+		return nil, fmt.Errorf("%w: void transform only accepts a nil partition value", iceberg.ErrInvalidArgument)
+	}
+
+	if (isIdentityTransform(transform) || isTruncateTransform(transform)) && !isNaN(normalized.Any()) {
+		applied := transform.Apply(iceberg.Optional[iceberg.Literal]{Valid: true, Val: normalized})
```
zeroshade, line 276: **nit** — PR body overstates the impossible-value validation

The description says the change will 'fail closed for missing or impossible partition values', but the range/fixed-point validation at :259-281 covers only bucket, truncate, identity and void. Year/month/day/hour partition values get no plausibility check. This is harmless in practice (an impossible value yields a predicate that matches nothing, i.e. under-delete rather than over-delete), but the claim should be scoped to the transforms actually validated.

table/evaluators_test.go
```diff
@@ -3100,6 +3100,21 @@ func TestManifestEvaluatorKeepsTransformedTerms(t *testing.T) {
 	assert.True(t, keep)
 }
 
+func TestPrepareBatchFilterRejectsTransformedTerms(t *testing.T) {
+	schema := iceberg.NewSchema(1,
```
zeroshade, line 3104: **nit** — TestPrepareBatchFilterRejectsTransformedTerms sits in the wrong test file

prepareBatchFilter is defined in table/transaction.go:3356; its test was added to table/evaluators_test.go. The test itself is good and correctly pins the limitation the PR description depends on — it just belongs next to the code it covers.

table/internal/partition_predicate.go
```diff
@@ -39,46 +39,69 @@ import (
...
 		if len(f.SourceIDs) != 1 {
-			return nil, fmt.Errorf("%w: identity partition field %q must have exactly one source id, got %d",
+			return nil, fmt.Errorf("%w: partition field %q must have exactly one source id, got %d",
 				iceberg.ErrInvalidArgument, f.Name, len(f.SourceIDs))
 		}
+		if isVoidTransform(f.Transform) {
+			// A void field can survive source-column removal as a source-less
+			// tombstone. Its output is always null, so resolving or binding the
+			// source would add no information and would reject source ID 0.
+			if _, err := f.Transform.MarshalText(); err != nil {
```
laskoviymishka: This MarshalText call can't fail. `VoidTransform.MarshalText` always returns `("void", nil)`, so the error branch is unreachable.

The twin call in the non-void path just below at line 98 has the same problem: anything that made it past `Bind` is already representable. I'd drop both error checks.

table/internal/partition_predicate.go
```diff
@@ -151,6 +196,110 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
 	return result, nil
 }
 
+func partitionTerm(transform iceberg.Transform, name string) iceberg.UnboundTerm {
+	ref := iceberg.Reference(name)
+	if isIdentityTransform(transform) {
+		return ref
+	}
+	if isVoidTransform(transform) {
```
laskoviymishka: This branch is dead with the current callers. `partitionTerm` is only reached for fields where `fr.isVoid` is false, and that's set precisely when `isVoidTransform` is true up in the field loop, so a void transform never flows down here.

I'd drop the `isVoidTransform` arm and let it fall through to the identity check plus the `NewUnboundTransform` default. wdyt?

table/internal/partition_predicate.go
```diff
@@ -151,6 +196,110 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
...
+		return t != nil
+	default:
+		return false
+	}
+}
+
+func validatePartitionValue(transform iceberg.Transform, resultType iceberg.Type, lit iceberg.Literal) (iceberg.Literal, error) {
+	normalized, err := lit.To(resultType)
+	if err != nil {
+		return nil, fmt.Errorf("%w: partition value type %s cannot be converted to transform result type %s: %v",
```
laskoviymishka: The inner conversion error goes in with `%v`, so it lands in the message text but not the error chain, and `errors.Is/As` can't reach the cause.

I'd switch it to `%w`, or drop the fragment if only the sentinel is meant to be unwrappable.

table/internal/partition_predicate.go
```diff
@@ -151,6 +196,110 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
...
+			iceberg.ErrInvalidArgument, normalized, resultType)
+	}
+
+	switch t := transform.(type) {
+	case iceberg.BucketTransform:
+		if err := validateBucketPartitionValue(t.NumBuckets, normalized); err != nil {
+			return nil, err
+		}
+	case *iceberg.BucketTransform:
+		if t == nil {
```
laskoviymishka: Same story as `partitionTerm` here: both of these are unreachable. `validatePartitionValue` only runs after the `isVoid` guard, so the `VoidTransform` case a few lines down never fires, and a typed-nil `*BucketTransform` is already rejected by `Bind` before we get here (the typed-nil-bucket test confirms it), so the `t == nil` guard can't trip either.

I'd drop both and keep just the value-form bucket case.

laskoviymishka (approved):
I came at this fresh and walked the spec/Java parity and the fail-closed paths independently, and they hold up. Year/month/hour land as Int32, day as a date, the bucket range guard is [0, N), and erroring on a missing partition key instead of degrading to IsNull is the right call under partition evolution. The last review was already non-blocking, and I don't see anything new that changes that, so this is a light concurring pass rather than a hold.

I did double-check the void null test that the earlier rounds flagged as inert. On its own it can't tell "skip the clause" from `IsNull(void(x))`, but the dedicated `VoidBindsAlwaysTrue` and `SourceLessVoidDoesNotDropOtherFields` tests backstop the structural behavior, so I'm not worried about it for phase 1.

The only real residue is the unreachable validation code from the last review. It's smaller than the ~25-line estimate: the void branch in `partitionTerm`, the void case and nil-pointer bucket guard in `validatePartitionValue`, and both `MarshalText` guards that can't fail once a field is past `Bind`. All of it is dead because the void and typed-nil cases are caught upstream, and none of it is a correctness issue, so I'd treat it as cleanup rather than a blocker.

Trim those branches (and maybe switch the secondary conversion error to `%w` so `errors.As` can reach the cause) and this is good to land from my side.

table/internal/partition_predicate.go
```diff
@@ -39,46 +39,62 @@ import (
...
 
-		src, ok := schema.FindFieldByID(f.SourceIDs[0])
+		sourceName, ok := schema.FindColumnName(f.SourceIDs[0])
 		if !ok {
 			return nil, fmt.Errorf("%w: partition field %q references unknown source id %d",
 				iceberg.ErrInvalidArgument, f.Name, f.SourceIDs[0])
 		}
-
-		fields = append(fields, fieldRef{id: f.FieldID, name: src.Name})
+		bound, err := iceberg.NewUnboundTransform(f.Transform, iceberg.Reference(sourceName)).Bind(schema, true)
```
zeroshade, line 91: `Bind` validates nil/unknown transforms and source compatibility, but not transform *parameters*: `BucketTransform.NumBuckets` and `TruncateTransform.Width` are never range-checked, and `NewPartitionSpec` deliberately accepts unvalidated fields.

So `bucket[0]` or `truncate[0]` paired with an explicit nil partition value builds successfully as `IsNull(transform(value))`. The invalid transform's `Apply` returns an invalid optional for every non-null source, which reads as null, so the predicate evaluates true against non-null input — probing `int32(1)` matched for both. Once the overwrite path is wired, that is an all-row match where the caller intended a single partition, i.e. deleting rows outside the target.

Restoring the parameter check the 09-03 cleanup removed (the former `MarshalText` validation, or an equivalent validator) would close it. Non-blocking today only because nothing reaches it yet.

table/internal/partition_predicate.go
```diff
@@ -151,6 +189,92 @@ func BuildPartitionMatchPredicate(spec iceberg.PartitionSpec, schema *iceberg.Sc
...
+
+	switch normalized.(type) {
+	case iceberg.AboveMaxLiteral, iceberg.BelowMinLiteral:
+		return nil, fmt.Errorf("%w: partition value %s is outside transform result type %s",
+			iceberg.ErrInvalidArgument, normalized, resultType)
+	}
+
+	switch t := transform.(type) {
+	case iceberg.BucketTransform:
+		if err := validateBucketPartitionValue(t.NumBuckets, normalized); err != nil {
```
zeroshade, line 249: This case covers only the value form:

```go
switch t := transform.(type) {
case iceberg.BucketTransform:
```

The non-nil pointer case went away with the 09-03 cleanup, so `&BucketTransform{NumBuckets: 4}` with the impossible partition value `4` passes validation and yields a predicate that can never match. Under-pruning is perf-only rather than a correctness hazard, which is the other reason I'm not blocking — but the asymmetry between `BucketTransform` and `*BucketTransform` will surprise someone eventually.

If you pick this up, three regressions would cover it: nil-valued `bucket[0]`, nil-valued `truncate[0]`, and `&BucketTransform{NumBuckets: 4}` with partition value `4`.

zeroshade (commented):
Not blocking. The valid-input behavior is correct and well covered, and the gap below isn't reachable from a production path today — flagging it so it's on record before the overwrite caller lands.

For valid inputs the predicate is exact: identity stays a raw reference, and bucket collision, truncate prefixes, year/month/day/hour boundaries including pre-epoch hour, null/void, nested source paths, and pointer identity are all exercised and pass. Metrics and bloom handling stay conservative, and Substrait rejects unsupported transformed terms rather than silently changing semantics. The earlier rounds' items check out: nested binding, pointer/void handling, missing-key rejection, pruning conservatism, and per-transform execution coverage.

The 09-03 dead-code cleanup was mostly right — dropping the genuinely dead void and typed-nil branches was sound. Two of the removals were load-bearing; details inline. Since the overwrite caller isn't wired up yet these are latent rather than live, so a follow-up issue is fine by me if you'd rather not spin another round on this PR.

zeroshade (approved):
APPROVE, which clears my `CHANGES_REQUESTED` from 09-02 — that predates the current head, and I've since assessed what remains as non-blocking.

The two findings from my comment review earlier today stand as written: the missing transform-parameter validation (`table/internal/partition_predicate.go:91`) and the dropped non-nil `*BucketTransform` range case (`:249`). Neither is reachable from a production path until the overwrite caller is wired up, so neither should gate this. I'm happy to carry them as a follow-up issue rather than another round here — say the word and I'll open one.

PR #1676, opened 2026-08-06
fix(table): apply per-task residual filters in ReadTasks

fallintoplace:
## Summary

Apply the residual filter attached to each scan task when reading tasks.

## Why

Manifest planning can produce a residual that is narrower than the scan filter. Ignoring it can return rows that the task was not supposed to read.

## What changed

Bind task residuals against the effective schema, use them for projection and row-group pruning, and keep the global scan filter as the fallback. The caller task slice is not modified.

## Tests

- `go test ./table -run TestReadTasksUsesTaskResidual -count=1`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -710,7 +710,7 @@ func collectLeafIDs(typ iceberg.Type, fieldID int, idset set[int]) {
 	}
 }
 
-func (as *arrowScan) projectedFieldIDs() (set[int], error) {
+func (as *arrowScan) projectedFieldIDs(rowFilter iceberg.BooleanExpression) (set[int], error) {
 	idset := set[int]{}
 	// Collect leaf field IDs for column pruning.
 	// For nested types (map, list, struct), we recursively descend to find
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
The core contract is right: `ReadTasks` now honors per-task residuals, binding them against the scan's effective schema — not the table's current one — so historical-snapshot scans filter correctly, with the caller's task slice cloned rather than mutated and the `FileScanTask.Residual` doc updated to match reality. The mixed bound/unbound rejection in `bindTaskFilter` fails the right way, and the row-group/bloom pruning now keyed on the per-task filter is a nice compounding win.

Two notes, neither blocking since #1702 (stacked on this) already carries the fix: the `bindTaskFilter(as.metadata.CurrentSchema(), …)` fallback inside `recordsFromTask` is unreachable today (residuals arrive pre-bound or nil) but should become `as.scanSchema` when #1702 lands so the latent wrong-schema bind never activates; and the `rowFilter := as.boundRowFilter` dead store plus the post-hoc nil-residual override in the same block reads awkwardly — worth folding into a single conditional whenever this area is next touched.

---

PR #1677, opened 2026-08-06
feat(table): allow RewriteFiles to add delete files

fallintoplace:
## Summary

Allow rewrite operations to add position and equality delete files atomically.

## Why

A rewrite can produce delete files together with replacement data files, but RewriteFiles currently rejects delete files as additions.

## What changed

Add delete-file routing to RewriteFiles and a transaction replacement path that validates and appends delete files in the same snapshot. Validate file types, paths, references, format version, and equality field IDs.

## Tests

- `go test ./table -run TestRewriteFiles_AddsDeleteFile -count=1`
- `go test ./table -run TestRewriteFiles -count=1`

### DIFF
--- a/table/snapshot_producers.go
+++ b/table/snapshot_producers.go
@@ -576,12 +576,22 @@ type snapshotProducer struct {
 	snapshotID         int64
 	parentSnapshotID   int64
 	addedFiles         []iceberg.DataFile
-	addedDeleteFiles   []iceberg.DataFile
+	addedDeleteFiles   []deleteFileAddition
 	manifestCount      atomic.Int32
 	deletedFiles       map[string]iceberg.DataFile
 	deletedDeleteFiles map[string]iceberg.DataFile
 	deletedDVsByRef    map[string]iceberg.DataFile
 	snapshotProps      iceberg.Properties
+}
+
+// deleteFileAddition carries the data sequence number of a rewritten delete
+// file. A nil sequence means the file is a normal delete added by the current
+// snapshot and should inherit that snapshot's sequence number. Rewrite paths
+// use an explicit sequence so replacing delete files cannot change which data
+// files the delete applies to.
+type deleteFileAddition struct {
+	file               iceberg.DataFile
+	dataSequenceNumber *int64
 }
 
 func createSnapshotProducer(op Operation, txn *Transaction, fs iceio.WriteFileIO, commitUUID *uuid.UUID, snapshotProps iceberg.Properties) *snapshotProducer {
@@ -630,7 +640,16 @@ func (sp *snapshotProducer) appendDataFile(df iceberg.DataFile) *snapshotProduce
 }
 
 func (sp *snapshotProducer) appendDeleteFile(df iceberg.DataFile) *snapshotProducer {
-	sp.addedDeleteFiles = append(sp.addedDeleteFiles, df)
+	sp.addedDeleteFiles = append(sp.addedDeleteFiles, deleteFileAddition{file: df})
+
+	return sp
+}
+
+func (sp *snapshotProducer) appendDeleteFileWithDataSequenceNumber(df iceberg.DataFile, seq int64) *snapshotProducer {
+	sp.addedDeleteFiles = append(sp.addedDeleteFiles, deleteFileAddition{
+		file:               df,
+		dataSequenceNumber: &seq,
+	})
 
 	return sp
 }
@@ -771,7 +790,7 @@ func (sp *snapshotProducer) addedContentManifests() ([]iceberg.ManifestFile, err
 	}
 
 	if len(sp.addedDeleteFiles) > 0 {
-		g.Go(sp.manifestProducer(iceberg.ManifestContentDeletes, sp.addedDeleteFiles, &positionDeleteManifests))
+		g.Go(sp.deleteManifestProducer(&positionDeleteManifests))
 	}
 
 	if err := g.Wait(); err != nil {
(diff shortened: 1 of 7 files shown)


zeroshade (commented):
The spec-level validation here is thorough (DV uniqueness per data file, applicability by sequence number, surviving pos-delete suppression, v2/v3 gating, dropped-equality-field tolerance), and the test coverage is genuinely strong. I built the branch and ran the full `./table` suite — all green. A few concerns before this leaves draft:

1. **No commit-time conflict validation for the new delete-file invariants.** All of the new checks in `Transaction.replaceFiles` (transaction.go: "deletion vector for data file %s already exists and must be replaced", "would survive", referenced-data-file liveness in `validateAddedDeletionVectorTargets`) run against `meta.currentSnapshot()` at staging time only. `RewriteFiles.Commit` (rewrite_files.go:376-378) still registers `rewriteValidator(r.dataFilesToDelete)` — which only guards the rewritten *data* files — and registers nothing at all for a pure delete-file rewrite (`dataFilesToDelete` empty), while `skipDefaultValidator` suppresses the overwrite isolation validator. So two concurrent transactions that each rewrite pos-deletes into a DV for the same referenced data file can both commit on retry, producing two DVs for one data file (a v3 spec violation); similarly a concurrent `RowDelta` can land a new delete on the DV's target that the staged "must be replaced" check would have rejected. I think the rewrite validator needs to also cover the data files referenced by added DVs (and be registered even when no data files are rewritten), re-checking the single-DV/no-surviving-delete invariants against concurrent snapshots the way `validateNoNewDeletesForRewrittenFiles` does.

2. **`RewriteFiles.DataSequenceNumber` has no upper bound.** It's validated `>= 0` (rewrite_files.go:180) but nothing prevents passing a value greater than the new snapshot's sequence number, which would stamp ADDED data entries with a sequence number above the snapshot's own — breaking the spec invariant and making future deletes mis-apply. I'd validate `seq <= meta.nextSequenceNumber()` in `replaceFiles` where you already compute `addedDataSequenceNumber`.

3. **Membership-only sequence validation for added pos-deletes is a documented footgun.** For pos-delete rewrites, an added file's explicit sequence number only has to be *one of* the removed files' sequence numbers (transaction.go, "expected one of the replaced delete sequence numbers"). If a caller merges content from sources with seqs {3, 7} into one output and assigns 3, deletes silently stop applying to data files with seq in (3, 7] — data resurrection with no error. The `DeleteFileAddition` doc acknowledges grouping isn't representable in the API, but given the failure mode is silent, it may be worth either representing the group explicitly or at least calling this out more loudly. Question rather than a demand — what's your take?

Minor: `writeAddedDeleteManifest` (snapshot_producers.go:996) skips the `writerClosed` guard that `writeAddedManifest` uses; it works because `ManifestWriter.Close` is idempotent, but matching the surrounding pattern would be more consistent. Also, the `for path := range setToDelete { if _, replacement := setToAdd[path]; ... }` branch in `replaceFiles` is dead code — the entries loop already rejects any `setToAdd` path referenced by the table, so a path can never be in both sets.

Since this is a draft I'm leaving comments only; happy to re-review once the conflict-validation story for delete additions is settled.

zeroshade:

table/snapshot_producers.go
```diff
@@ -953,6 +1001,29 @@ func (sp *snapshotProducer) writeAddedManifest(content iceberg.ManifestContent,
 	return wr.ToManifestFile(path, counter.Count, iceberg.WithManifestFileContent(content))
 }
 
+func (sp *snapshotProducer) writeAddedDeleteManifest(specID int, additions []deleteFileAddition) (_ iceberg.ManifestFile, retErr error) {
+	wr, path, counter, out, err := sp.newManifestWriter(sp.spec(specID), iceberg.WithManifestWriterContent(iceberg.ManifestContentDeletes))
+	if err != nil {
+		return nil, err
+	}
+	defer internal.CheckedClose(out, &retErr)
+	defer internal.CheckedClose(wr, &retErr)
```
zeroshade: Could we mirror `writeAddedManifest`'s `writerClosed` guard here? As written, the defer closes `wr` again after the explicit `wr.Close()` below, relying on `ManifestWriter.Close` remaining idempotent. Matching the adjacent lifecycle pattern avoids that assumption and keeps both writer paths consistent.

table/transaction.go
```diff
@@ -1469,42 +1817,202 @@ func (t *Transaction) ReplaceFiles(ctx context.Context, dataFilesToDelete, dataF
...
+	if cfg.dataSequenceNumber != nil {
+		addedDataSequenceNumber = *cfg.dataSequenceNumber
+	}
+	for _, df := range dataFilesToAdd {
+		liveDataFiles[df.FilePath()] = rewriteFileState{
+			file: df, dataSequenceNumber: addedDataSequenceNumber,
+		}
+	}
+	for path := range setToDelete {
+		if _, replacement := setToAdd[path]; !replacement {
```
zeroshade: This replacement branch appears unreachable: the earlier entry scan rejects any `setToAdd` path already referenced by the table, so a path cannot successfully be present in both `setToDelete` and `setToAdd`. Could this be simplified to unconditionally delete every `setToDelete` path from `liveDataFiles`?

table/transaction.go
```diff
@@ -1541,7 +2062,14 @@ func (t *Transaction) ReplaceFiles(ctx context.Context, dataFilesToDelete, dataF
 		return err
 	}
 
-	return t.apply(updates, reqs)
+	if err := t.apply(updates, reqs); err != nil {
+		return err
+	}
+	if len(rewriteValidationFiles) > 0 {
+		t.addValidator(rewriteValidatorWithReferencedDataFiles(rewriteValidationFiles, referencedDataFilePaths))
```
zeroshade, line 2067: For rewrites that add delete files, this extended validator already includes a clone of `dataFilesToDelete`, but `RewriteFiles.Commit` subsequently registers the basic `rewriteValidator(r.dataFilesToDelete)` as well. That validates the same rewritten data files twice and can duplicate manifest conflict scanning during retries. Could registration be centralized so exactly one validator covers the rewritten files plus any referenced DV targets?

zeroshade (changes requested):
The correctness concerns from my prior review are resolved, and the updated tests and CI are green. Requesting cleanup of the three inline items before merge: align manifest-writer close handling, remove unreachable same-path replacement logic, and avoid duplicate conflict-validator registration.

zeroshade (approved):

PR #1678, opened 2026-08-06
fix(puffin): read LZ4-compressed footer payloads

fallintoplace:
## Summary

Read Puffin files whose footer payload uses the LZ4 compression flag.

## Why

The reader rejected compressed footer payloads even though LZ4 is the codec defined for Puffin footer compression.

This follows the [Puffin spec](https://iceberg.apache.org/puffin-spec/) directly. It defines a compressed footer as a single LZ4 compression frame with content size present.

For context, Apache Iceberg Java main recognizes the compressed footer flag in [PuffinReader](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/puffin/PuffinReader.java) and routes it through PuffinFormat.decompress(...), but the [LZ4 branch in PuffinFormat](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/puffin/PuffinFormat.java) is still TODO and currently falls through to unsupported. This PR implements that narrow reader path in Go; it is not a comparison of overall Puffin support in either project.

## What changed

Decompress the footer payload as an LZ4 frame before decoding its JSON.

- Validate the fixed LZ4 header first and reject missing or oversized advertised content before scanning the rest of the payload.
- Carry the BD-declared maximum block size into envelope validation and reject oversized blocks before reading their bodies.
- Validate the standard single-frame envelope, including reserved descriptor bits, block checksums, and trailing compressed data.
- Buffer both envelope inspection and LZ4 decoding through independent 4 MiB section readers to avoid one tiny object-store range read per frame block.
- Keep the decompressed-size limit and LimitedReader backstop for compressed footers only, so uncompressed footer behavior stays unchanged.
- Decode Puffin FooterPayloadSize as a signed 32-bit little-endian value.
- Reject unknown footer flags and verify the decoded size matches the LZ4 content size.
- Preserve late LZ4 checksum and read errors instead of reporting them as trailing JSON.

## Tests

- go test ./...
- go test -race ./puffin
- go vet ./...
- golangci-lint v2.8.0
- Regression coverage for concatenated frames, reserved descriptor bits, buffered frame inspection, oversized BD-declared blocks failing before body reads, positive and corrupted block checksums, checksum error propagation, signed footer sizes, size limits, and content-size mismatches.
- Independent compressed-footer fixture generated with LZ4 CLI 1.10.0.


puffin/puffin_reader.go
```diff
@@ -224,11 +238,32 @@ func (r *Reader) readFooter() error {
...
-	decoder := json.NewDecoder(payloadReader)
+	var footerReader io.Reader = payloadReader
+	if flags&FooterFlagCompressed != 0 {
+		lz4Reader := lz4.NewReader(payloadReader)
+		var firstByte [1]byte
+		n, err := lz4Reader.Read(firstByte[:])
+		if err != nil && !errors.Is(err, io.EOF) {
+			return fmt.Errorf("puffin: initialize compressed footer: %w", err)
+		}
+		if lz4Reader.Size() == 0 {
```
laskoviymishka: This leans on the 1-byte `Read()` having populated `Size()` as a side effect. pierrec parses the frame descriptor on the first Read today, but that ordering isn't part of the documented contract. A lazy-parsing version, or a different lz4 library, would return `Size() == 0` for a perfectly valid frame and we'd reject every Trino-written file.

`Size() == 0` also can't tell "no content-size field" (the case the spec actually prohibits) from "field present, declared as 0", so the "missing content size" error can misfire on the degenerate case.

I'd read the frame descriptor explicitly via the lz4 header API instead of the peek-and-reconstruct dance, or just decompress into a bounded buffer and let the `LimitedReader` be the only size gate. wdyt?

puffin/puffin_reader.go
```diff
@@ -224,11 +238,32 @@ func (r *Reader) readFooter() error {
...
+			return errors.New("puffin: compressed footer LZ4 frame is missing content size")
+		}
+		if int64(lz4Reader.Size()) > r.maxFooterSize {
```
laskoviymishka: `lz4Reader.Size()` is a `uint64`; casting to `int64` first means a frame advertising a content size above `math.MaxInt64` wraps negative and sails straight past this guard. The `LimitedReader` still catches it downstream, but this early check is exactly the fail-fast-before-allocation gate, and a one-bit flip in the frame header disables it.

Comparing in uint64 space closes it, since `r.maxFooterSize` is already guaranteed positive above:

```go
if lz4Reader.Size() > uint64(r.maxFooterSize) {
```

puffin/puffin_reader.go
```diff
@@ -224,11 +238,32 @@ func (r *Reader) readFooter() error {
...
+			return fmt.Errorf("puffin: footer exceeds maximum size %d", r.maxFooterSize)
+		}
+		footerReader = io.MultiReader(bytes.NewReader(firstByte[:n]), lz4Reader)
+	}
+
+	limitedFooter := &io.LimitedReader{R: footerReader, N: r.maxFooterSize + 1}
```
laskoviymishka: Small edge: if a caller passes `WithMaxFooterSize(math.MaxInt64)`, `N: r.maxFooterSize + 1` overflows to `MinInt64`, `LimitedReader` treats `N <= 0` as an empty stream, and the decoder fails with "unexpected end of JSON input", which points nowhere near the actual misconfig. Worth capping `N` at the sentinel when `maxFooterSize` is near `MaxInt64`.

puffin/puffin_reader.go
```diff
@@ -224,11 +238,32 @@ func (r *Reader) readFooter() error {
...
+	decoder := json.NewDecoder(limitedFooter)
 	var footer Footer
 	if err := decoder.Decode(&footer); err != nil {
 		return fmt.Errorf("puffin: decode footer JSON: %w", err)
 	}
+	if limitedFooter.N == 0 {
```
laskoviymishka: `json.Decoder` reads ahead into its own buffer, so by the time `Decode` returns `N` may already be 0 because the decoder pulled in trailing bytes past the JSON value, not because the footer itself overflowed. In that case this reports "footer exceeds maximum size" when the real problem is trailing content, and the deliberate trailing-content check just below never gets a chance to fire.

I'd move the `N == 0` guard to wrap the trailing-content read rather than sit between `Decode` and it, or check `decoder.Buffered()` to tell the two apart.

puffin/puffin_test.go
```diff
@@ -619,6 +638,60 @@ func TestReaderRejectsTrailingFooterData(t *testing.T) {
...
+	data := fileWithCompressedFooterPayload(t, []byte(`{"blobs":[],"properties":{"source":"lz4"}}`))
+
+	r, err := puffin.NewReader(bytes.NewReader(data))
+	require.NoError(t, err)
+	assert.Equal(t, "lz4", r.Properties()["source"])
+	assert.Empty(t, r.Blobs())
+}
+
+func TestReaderRejectsLZ4CompressedFooterWithoutContentSize(t *testing.T) {
+	payload := []byte(`{"blobs":[]}`)
```
laskoviymishka, line 787: This inlines the same frame-building as `fileWithCompressedFooterPayload` minus the `SizeOption` call, so if the helper ever regressed in a way that dropped `SizeOption` this test would still pass green. I'd factor out a `fileWithCompressedFooterPayloadNoSize(t, payload)` so the two stay in sync.

puffin/puffin_test.go
```diff
@@ -619,6 +638,60 @@ func TestReaderRejectsTrailingFooterData(t *testing.T) {
...
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "missing content size")
+}
+
+func TestReaderEnforcesFooterSizeLimit(t *testing.T) {
+	t.Run("json payload", func(t *testing.T) {
+		data := fileWithFooterPayload([]byte(`{"blobs":[],"properties":{"large":"0123456789"}}`))
+		_, err := puffin.NewReader(bytes.NewReader(data), puffin.WithMaxFooterSize(16))
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "footer")
```
laskoviymishka: `assert.Contains(t, err.Error(), "footer")` would also pass on unrelated errors like "invalid footer start magic". The two sibling subtests assert the specific "footer exceeds maximum size", so I'd match that here too.

puffin/puffin_test.go
```diff
@@ -619,6 +638,60 @@ func TestReaderRejectsTrailingFooterData(t *testing.T) {
...
+
+	t.Run("trailing content beyond limit", func(t *testing.T) {
+		payload := append([]byte(`{"blobs":[]}`), bytes.Repeat([]byte{' '}, 32)...)
+		data := fileWithFooterPayload(payload)
+		_, err := puffin.NewReader(bytes.NewReader(data), puffin.WithMaxFooterSize(16))
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "footer exceeds maximum size")
+	})
+
+	t.Run("compressed advertised size", func(t *testing.T) {
```
laskoviymishka, line 899: All three size-limit subtests hit the early advertised-size rejection; none exercise the `LimitedReader` as the actual backstop, i.e. a frame that advertises a size within the limit but decompresses to more than it. That's the one path that matters if the early check is ever bypassed (see the uint64 issue above), and lz4 won't always validate actual output against the advertised size.

I'd add a subtest that builds a frame with `SizeOption(<= maxFooterSize)` but writes a payload that decompresses past the limit, and assert it still errors with "footer exceeds maximum size".

laskoviymishka (changes requested):
Nice work getting compressed-footer reads working. Dropping the hard "not supported" error and mirroring the `WithMaxBlobSize` shape with `WithMaxFooterSize` is the right structure, and I like that you kept this reader-only rather than reaching for writer support. Java and PyIceberg both still reject compressed footers, so writing them would produce files neither can read; worth a note whenever the writer side lands, but nothing to do here.

I'd hold this before merging, though. The main thing is the early size guard: `int64(lz4Reader.Size())` narrows a `uint64`, so a frame advertising a size above `math.MaxInt64` wraps negative and skips the check entirely. The `LimitedReader` still backstops it, but that's exactly the case there's no test for: every size-limit subtest exercises the early rejection, none exercise the LimitedReader catching a frame that lies about its size. Since the early check is bypassable, that backstop is the one that has to hold, so I'd like it covered.

- compare the advertised size in uint64 space so the guard can't be flipped off
- add a subtest where the frame advertises a small size but decompresses past the limit, asserting the LimitedReader still errors
- the "missing content size" detection leans on undocumented pierrec Read/Size() ordering; I'd read the frame descriptor explicitly or let the LimitedReader be the only gate
- the post-Decode `N == 0` check can misreport trailing content as a size overflow

The rest are small: a couple of test-tightening nits and an overflow edge on `maxFooterSize + 1`. Once the size guard and its test are sorted, happy to take another pass and approve.

zeroshade (approved):
This landed in a very good place. The explicit frame-descriptor parse (rather than leaning on the lz4 reader's side effects), uint64-space size guards, the decoder read-ahead ordering fix, and the decoded-size cross-check close every hole from the previous rounds, and the envelope validator's handling of checksums, reserved bits, concatenated frames, and truncated end marks is thorough — with the CLI-written testdata file anchoring interoperability. The negative-payload-size fix via sign extension is a nice catch beyond the original scope.

Two non-blocking thoughts: the two 4 MB `bufio` buffers could size down to `min(payloadSize, 4MB)` for typical small footers, and the envelope walk is a second full pass over the payload — both fine here, just worth keeping in mind if this pattern ever extends to blob reads.

---

puffin/puffin_reader.go
```diff
@@ -74,6 +79,17 @@ func WithMaxBlobSize(size int64) ReaderOption {
 	}
 }
 
+// WithMaxFooterSize sets the maximum decompressed size allowed for compressed
```
laskoviymishka: The limit here only binds compressed footers. An uncompressed footer ignores `WithMaxFooterSize` entirely, and a non-positive value that errors for compressed reads is silently a no-op for uncompressed ones. That's a little surprising from the name: someone setting `WithMaxFooterSize(N)` on untrusted input gets no protection when the footer happens to be uncompressed.

I'd either enforce it on the uncompressed path too, or rename to `WithMaxCompressedFooterSize` / `DefaultMaxCompressedFooterSize` so the scope is obvious from the call site. Either's fine, I just don't want the asymmetry to be a surprise. wdyt?

puffin/puffin_test.go
```diff
@@ -671,6 +730,204 @@ func TestReaderRejectsTrailingFooterData(t *testing.T) {
...
+func TestReaderPreservesLZ4BlockChecksumError(t *testing.T) {
+	payload := []byte(`{"blobs":[]}`)
+	frame := compressedLZ4Frame(
+		t,
+		payload,
+		lz4.SizeOption(uint64(len(payload))),
+		lz4.BlockChecksumOption(true),
+		lz4.ChecksumOption(false),
+	)
+	const lz4FrameHeaderSizeWithContentExternal = 15
```
laskoviymishka: These two locals shadow the real package consts, and `blockChecksumOffset` is computed straight from them. If `lz4FrameHeaderSizeWithContent` or the block-size mask ever changes, this test keeps using the stale copies, mutates the wrong byte, and quietly stops testing the block-checksum path (or panics). Since `puffin_reader_internal_test.go` is already `package puffin`, I'd move this case there and reference the real consts.

Same theme, not worth its own thread: the `data[12]` and `data[len(data)-13]` offsets in the flag and checksum tests are hand-derived from the header layout and would drift the same way. A named const or a short comment would anchor them.

puffin/puffin_reader.go
```diff
@@ -230,11 +441,65 @@ func (r *Reader) readFooter() error {
 		payloadReader = io.NewSectionReader(r.r, footerStart+MagicSize, payloadSize)
 	}
 
-	decoder := json.NewDecoder(payloadReader)
+	var footerReader io.Reader = payloadReader
+	var compressedFooter *countingReader
+	var compressedContentSize uint64
+	var limitedFooter *io.LimitedReader
+	if flags&FooterFlagCompressed != 0 {
```
laskoviymishka, line 454: In round 1 I said the ecosystem note could wait for the writer side. I'll walk that back a touch. Java's `PuffinFormat.decompress()` still throws on the LZ4 branch, and PyIceberg and rust reject compressed footers too, so a file we read here won't reopen in any of them today.

A one-line comment right at this branch noting that would be a nice breadcrumb for whoever adds writer support, so they know not to flip `FooterFlagCompressed` on until the others catch up. Still not a blocker, just cheap insurance.

laskoviymishka (approved):
This landed in a really good spot. The revision closed every thread from my last pass, and cleanly. Quick confirm on each:

- the size guard compares in uint64 space now (`frameHeader.contentSize > uint64(r.maxFooterSize)`), so it can't be flipped off by an oversized advertised size
- there's a subtest that advertises a small size and then decompresses past the limit, exercising the `LimitedReader` backstop directly. That was the gap I cared about most.
- the frame descriptor is parsed explicitly in `readLZ4FrameHeader` rather than leaning on pierrec's Read/Size() ordering
- the post-Decode `N == 0` path no longer misreports trailing content as a size overflow (nice touch reporting the buffered-trailing case separately)
- the `maxFooterSize+1` overflow edge is covered, including the `math.MaxInt64` case

One thing I'd still like us to at least name: `WithMaxFooterSize` only binds compressed footers. Uncompressed ones ignore it entirely, and a non-positive value errors for compressed reads but is a silent no-op for uncompressed. Not a blocker, but I left an inline; either enforce it on both paths or rename so the scope is obvious.

The rest are small and non-blocking. There's a test that reshadows package consts and could drift silently (inline), and, walking back my round-1 "nothing to do here" a touch, a one-line breadcrumb near the `FooterFlagCompressed` branch noting Java/PyIceberg/rust still reject these would help a future writer know not to enable the flag yet. If you want to tidy while you're in there, comparing `compressedFooter.count` in the int64 domain instead of casting to uint64 reads a hair cleaner, but it's optional. Agreed with zeroshade that the double 4 MB buffers and the second envelope pass are fine as-is.

With the round-1 threads closed and CI green, I'm happy to approve.

PR #1679, opened 2026-08-06
feat(expr): bind predicates over transform terms

fallintoplace:
## Summary

Bind and evaluate predicates whose term is a partition transform.

## Why

Transform terms can be parsed from REST expression JSON, but binding them was rejected before the typed predicate code ran. This left valid residual filters unusable.

## What changed

Generalize predicate evaluation to work with any bound term, including transformed values. Preserve transform terms when predicates are rewritten or serialized, and cover unary, literal, and set predicates.

## Tests

- `go test . -run TestExpressionTransformTermRoundTrip -count=1`
- `go test . -run TestEvaluatorTransformTerms -count=1`

Refs #1216


zeroshade:

exprs.go
```diff
@@ -790,7 +788,7 @@ type BoundLiteralPredicate interface {
 
 func newBoundLiteralPredicate[T LiteralType](op Operation, term BoundTerm, lit Literal) BoundPredicate {
 	return &boundLiteralPredicate[T]{
-		op: op, term: term.(bound[T]),
+		op: op, term: term,
```
zeroshade, line 791: **Blocking:** allowing a `BoundTransform` here exposes it to metrics and bloom evaluators that still compare the transform-result literal with raw source-column values from `term.Ref()`. I reproduced `truncate[3](category) == "boo"` incorrectly pruning a file containing `"books"`. More seriously, strict metrics reports that every row matches `truncate[3](category) != "boo"` for a required file containing only `"books"`, even though no row matches; filtered deletion can therefore drop the entire file. Bloom pruning has the same source-domain/transformed-domain mismatch. Please make all source-column pruning consumers conservatively handle transformed terms or correctly transform their statistics.

exprs.go
```diff
@@ -838,7 +836,7 @@ func createBoundLiteralPredicate(op Operation, term BoundTerm, lit Literal) (Bou
 
 type boundLiteralPredicate[T LiteralType] struct {
 	op   Operation
-	term bound[T]
+	term BoundTerm
```
zeroshade, line 839: **Blocking:** the Substrait converter is not transform-aware. Its reference, literal, and set helpers all emit `term.Ref()` directly, so `truncate[3](category) == "boo"` becomes `category == "boo"`, silently changing the predicate. Bucket and temporal transforms can instead fail because the source and result types differ. This path is used by scan filtering and partial-file rewrites; please encode the transform in Substrait or reject transformed predicates there rather than dropping it.

visitors.go
```diff
@@ -406,8 +406,23 @@ func nullsFirstCmp[T LiteralType](cmp Comparator[T], v1, v2 Optional[T]) int {
 	return cmp(v1.Val, v2.Val)
 }
 
+func typedTermEval[T LiteralType](st StructLike, term BoundTerm) Optional[T] {
+	v := term.evalToLiteral(st)
+	if !v.Valid {
+		return Optional[T]{}
+	}
+
+	lit, ok := v.Val.(TypedLiteral[T])
```
zeroshade, line 415: **Major:** this assertion makes `day` predicates unevaluable. `DayTransform.ResultType` is `Date`, so binding selects `TypedLiteral[Date]`, but `DayTransform.Apply` returns an `Int32Literal`. A direct `day(ts) == Date(0)` evaluation returns `type error: ... evaluated to literal type int, expected date`. Please align the transform's evaluated literal with its declared result type (and add coverage for `day` alongside bucket/truncate).

zeroshade (changes requested):
Found three correctness issues in downstream consumers of transformed bound terms, including a whole-file deletion path. The existing and CI tests pass, but they do not exercise metrics pruning, Substrait conversion, or `day` transform evaluation. Please also update the `ParseExpr` documentation, which still says binding transform predicates is unsupported.

---

>

exprs.go
```diff
@@ -1121,7 +1124,17 @@ func (u *UnboundTransform) Equals(other UnboundTerm) bool {
 }
 
 func (u *UnboundTransform) Bind(schema *Schema, caseSensitive bool) (BoundTerm, error) {
-	bound, err := u.term.Bind(schema, caseSensitive)
+	if isUnknownTransform(u.transform) {
```
zeroshade, line 1158: `NewUnboundTransform` is exported and accepts a `Transform` interface, so callers can pass `nil`. In that case this check falls through and `u.transform.CanTransform(...)` panics instead of returning an error. I reproduced this with `NewUnboundTransform(nil, Reference("x"))` followed by predicate binding. Since `Bind` already reports unknown and incompatible transforms as errors, please reject a nil transform here (for example with `ErrInvalidArgument`) and add a regression test.

zeroshade (changes requested):
Thanks—the three previously reported correctness issues are addressed. Transformed terms are now conservative in pruning consumers, Substrait rejects them rather than changing predicate semantics, and the day transform returns the declared literal type.

One new public-API robustness issue remains: `NewUnboundTransform(nil, ref)` panics during `Bind` when `CanTransform` is called on the nil transform. Please reject a nil transform with an error and add a regression test.

Please also add the required DCO sign-offs to all commits and reference the associated issue.

---

exprs.go
```diff
@@ -1116,12 +1128,40 @@ func (u *UnboundTransform) Equals(other UnboundTerm) bool {
...
+			return false
+		}
+		if u.term == nil || rhs.term == nil {
+			return u.term == nil && rhs.term == nil
+		}
+
+		return u.term.Equals(rhs.term)
+	}
 
 	return u.transform.Equals(rhs.transform) && u.term.Equals(rhs.term)
```
zeroshade, line 1151: Non-nil pointers to the built-in transforms are valid `Transform` values—the existing APIs and tests already use forms such as `&BucketTransform{...}`. However, the built-in `Equals` implementations only accept value forms, so delegating here breaks equality reflexivity for pointer transforms.

I reproduced this 10/10 times with `NewUnboundTransform(&BucketTransform{NumBuckets: 16}, ref)`: the unbound term does not equal itself, its bound predicate does not equal itself, and projecting that predicate through the identical pointer transform returns `nil` rather than removing the matching transform.

The typed-nil regressions cover the nil pointer case but miss the adjacent non-nil pointer case. Please make built-in transform equality pointer/value aware or normalize built-in pointers, and add regressions for unbound equality, bound-predicate equality, and same-transform projection.

zeroshade (changes requested):
The previously reported pruning, Substrait, temporal typing, nil-transform, DCO, and issue-reference findings are addressed. The updated implementation and latest-main integration pass the full validation set.

One public API issue remains inline: non-nil pointers to built-in transforms violate equality reflexivity and prevent matching transform projection. Please make pointer and value forms compare consistently and add coverage for the affected expression paths.

---

zeroshade (approved):

PR #1680, opened 2026-08-06
feat(metrics): include environment context in commit reports

fallintoplace:
## Summary

Include process-wide environment context in commit and scan report metadata.

## Why

Commit reports did not include metadata, so applications could not attach process or deployment context. Scan reports also need the same standard context for parity with Java.

## What changed

- Add a concurrency-safe environment context with copy-on-read behavior.
- Seed `iceberg-version` from the Iceberg Go module build information, both when it is the main module and when it is a dependency.
- Include independent environment context snapshots in commit report metadata.
- Merge scan options and environment context into scan report metadata, with environment context taking precedence to match Java.
- Keep environment properties mutable to match Java behavior.
- Add coverage for module version discovery, replacements, concurrent access, metadata precedence, and snapshot independence.

## Tests

- `go test -race . ./table`
- `go test ./...`


environment_context.go
```diff
@@ -0,0 +1,82 @@
...
+package iceberg
+
+import (
+	"maps"
+	"sync"
+)
+
+const (
+	// EnvironmentContextEngineNameKey identifies the engine name property.
+	EnvironmentContextEngineNameKey = "engine-name"
```
laskoviymishka: Since this is exported API and awkward to change once it ships, worth getting the name right now: `EnvironmentContextEngineNameKey` is a mouthful and the `Context` in the middle is structural noise. Java just uses `ENGINE_NAME`. I'd go with `EngineNameKey`/`EngineVersionKey`, or at least drop the `Context` — `EnvironmentEngineNameKey`. wdyt?

environment_context.go
```diff
@@ -0,0 +1,82 @@
...
+	// EnvironmentContextEngineVersionKey identifies the engine version property.
+	EnvironmentContextEngineVersionKey = "engine-version"
+
+	environmentContextIcebergVersionKey = "iceberg-version"
+)
+
+var environmentContext = struct {
+	sync.RWMutex
```
laskoviymishka: Minor, but the rest of the repo uses a named mutex field (`registryMu`, `regMutex`, `mu`) rather than embedding. Embedding promotes `Lock`/`RLock`/… onto the package var, so package code can bypass the public API with `environmentContext.Lock()`, and it trips `copylocks` if the struct ever gets copied. I'd make it `mu sync.RWMutex` and use `environmentContext.mu.Lock()`.

environment_context.go
```diff
@@ -0,0 +1,82 @@
...
+	properties map[string]string
+}{
+	properties: make(map[string]string),
+}
+
+var environmentContextInit sync.Once
+
+func initializeEnvironmentContext() {
```
laskoviymishka: Could we replace the `sync.Once` + `initializeEnvironmentContext()` with a plain `func init()` that seeds the key? `Version()` is set by `init()` in `utils.go` in this same package, so ordering is fine, and today every `EnvironmentContext`/`Set`/`Remove` call pays a `Once.Do` check for a seed that's really just a package-load constant.

It also cleans up a subtlety: once `Remove` deletes `iceberg-version`, the `Once` has already fired, so nothing re-seeds it. A plain `init()` makes "seeded at startup" the honest story. wdyt?

environment_context.go
```diff
@@ -0,0 +1,82 @@
...
+	environmentContextInit.Do(func() {
+		environmentContext.Lock()
+		environmentContext.properties[environmentContextIcebergVersionKey] = Version()
```
laskoviymishka: `Version()` returns the bare Go module semver, or the literal `(unknown version)` when built from source — which is basically every dev/test/CI run. So in those environments the report advertises `(unknown version)` as the writer.

Java seeds this key from `IcebergBuild.fullVersion()` (`Apache Iceberg X.Y.Z (commit …)`) and PyIceberg emits `PyIceberg X.Y.Z`. I'd format this as something like `Apache Iceberg Go <version>` so dashboards that key off `iceberg-version` to identify the client get a real, greppable value. A small helper in `utils.go` would do it.

environment_context.go
```diff
@@ -0,0 +1,82 @@
...
+
+	environmentContext.Lock()
+	defer environmentContext.Unlock()
+
+	environmentContext.properties[key] = value
+}
+
+// RemoveEnvironmentProperty removes one process-wide environment context
+// property.
+func RemoveEnvironmentProperty(key string) {
```
laskoviymishka, line 64: One thing I'd like us to settle: `iceberg-version` is fully mutable and removable through these setters, and `TestEnvironmentContextAllowsIcebergVersionMutation` locks that in. A single `RemoveEnvironmentProperty("iceberg-version")` from app code or an adapter silently drops the version from every subsequent report, and there's no re-seed path.

Java's `ConcurrentHashMap` is unguarded too, so freely-mutable does match Java. But I lean toward protecting this one key — either no-op the setters for it, or keep it as a separate field that's always merged in at read time in `EnvironmentContext()` — so a stray delete can't blank it out everywhere.

If we keep it mutable, I'd at least add a line to this doc noting that removing `iceberg-version` drops it permanently. wdyt?

environment_context_test.go
```diff
@@ -0,0 +1,94 @@
...
+	preserveEnvironmentProperties(t, keys...)
+
+	var wg sync.WaitGroup
+	for worker, key := range keys {
+		wg.Add(1)
+		go func(worker int, key string) {
+			defer wg.Done()
+			for iteration := range iterations {
+				SetEnvironmentProperty(key, strconv.Itoa(iteration))
+				_ = EnvironmentContext()[key]
```
laskoviymishka, line 85: This test proves the lock holds up under contention, which is worth having. But each worker uses its own key, and the `EnvironmentContext()` result here is read and discarded — so the thing that would actually be a bug, `EnvironmentContext()` handing back a live reference to the stored map instead of a clone, isn't exercised.

I'd add a case where one goroutine calls `EnvironmentContext()` and then mutates the returned map while another goroutine `SetEnvironmentProperty`s the same key, run under `-race`. If the clone ever regressed to a live reference, that's what would catch it.

table/commit_metrics_test.go
```diff
@@ -115,6 +115,36 @@ func TestBuildCommitReport(t *testing.T) {
 	assert.Nil(t, m.AddedDVs)
 }
 
+func TestBuildCommitReportIncludesEnvironmentContext(t *testing.T) {
+	keys := []string{
+		iceberg.EnvironmentContextEngineNameKey,
+		iceberg.EnvironmentContextEngineVersionKey,
+	}
+	previous := iceberg.EnvironmentContext()
```
laskoviymishka: This save/restore block is now copy-pasted here and in `scan_metrics_test.go`, and it's a near-duplicate of `preserveEnvironmentProperties` in the root package (which the table tests can't import since it's in package `iceberg`). Worth exporting a small test helper — `iceberg.PreserveEnvironmentProperties(t, keys...)` — or re-adding a bulk `SetEnvironmentContext(map)` so cleanup is one call. Otherwise the three copies drift when the reset logic changes.

table/scan_metrics.go
```diff
@@ -106,6 +107,9 @@ func (acc *scanMetricsAccumulator) applyResultDeleteMetrics(tasks []FileScanTask
 // are left unset and omitted.
 func (scan *Scan) buildScanReport(acc *scanMetricsAccumulator, schema, projected *iceberg.Schema, planning time.Duration) metrics.ScanReport {
 	ids, names := projectedFields(projected)
+	metadata := make(map[string]string, len(scan.options)+1)
+	maps.Copy(metadata, scan.options)
+	maps.Copy(metadata, iceberg.EnvironmentContext())
```
laskoviymishka: Copying `EnvironmentContext()` second means env context wins on key collisions, so a per-scan `engine-name` passed in `scan.options` gets silently overwritten by the process-wide value. That matches Java's ordering (`options()` first, then context), so I think it's the right call — but it's non-obvious enough that I'd drop a one-line comment saying env context intentionally takes precedence.

While you're here: `EnvironmentContext()` takes an RLock and clones on each call, and the `len(scan.options)+1` hint under-counts when there's more than one env key. I'd hoist it into a local — `envCtx := iceberg.EnvironmentContext()` — and size the map with `len(scan.options)+len(envCtx)`.

laskoviymishka (changes requested):
Nice work, this is clean and the test coverage is solid. Mirroring Java's `EnvironmentContext` as a process-wide map, wiring it into both the commit and scan reports, and getting copy-on-read right (the returned map is a clone) all look good.

I'd hold before merging on a couple of things, neither of them big.

The one I'd actually want fixed: `iceberg-version` is seeded from `Version()`, which returns the bare module semver, or the literal `(unknown version)` when built from source — so nearly every dev/CI report ends up advertising `(unknown version)` as the writer. Java seeds this from `IcebergBuild.fullVersion()` and PyIceberg emits `PyIceberg X.Y.Z`; I'd emit something like `Apache Iceberg Go <version>` so tooling that parses `iceberg-version` to identify the writer sees a real value.

The one I'd like us to decide on: right now `iceberg-version` is freely mutable and removable, and a test bakes that in. Java's map is unguarded too, so this matches Java — but a stray `RemoveEnvironmentProperty("iceberg-version")` from an adapter silently drops the version from every subsequent report with no re-seed path. I left a comment on whether we want to protect it (or re-merge it at read time); I lean protected, but it's a real trade-off against Java parity. wdyt?

Separately, worth tracking as a follow-up (not for this PR): Java also injects `EnvironmentContext` into the snapshot summary via `SnapshotProducer.updateSummary`, so every snapshot carries engine/version durably. This PR covers the ephemeral reports but not `rebaseSummary`, so summary-reading tooling will still see blanks for iceberg-go snapshots. Fine to scope out given the PR title.

Rest of my notes are inline and minor. Once the version string and the protection question are settled, happy to take another pass.

zeroshade (approved):
Verified the morning-review items at head: `iceberg-version` now seeds from `fullVersion()` (`Apache Iceberg Go <version>`), so reports identify the writer implementation even on source builds; the naming, mutex-field, and init-shape nits are all addressed; and `TestEnvironmentContextSnapshotConcurrentMutation` pins the clone-not-live-reference contract under contention, which was the meaningful gap in the concurrency coverage.

On the open question of `iceberg-version` mutability: I'd keep it as-is. Java's `EnvironmentContext` map is equally unguarded, engines legitimately override these keys, and guarding one key creates a surprising special case — the mutation test documenting the behavior is the right amount of protection. If we ever see real-world reports missing the version we can revisit with a re-seed-on-report approach.

---

laskoviymishka (approved):
:shipit:

PR #1681, opened 2026-08-06
perf(table): prune row groups during row-lineage rewrites

fallintoplace:
## Summary

Restore Parquet row-group pruning during copy-on-write rewrites that preserve row lineage.

## Why

The rewrite scanner had to keep the row filter out of the main scan while it synthesized `_row_id`. That also disabled Parquet statistics and bloom-filter pruning.

## What changed

- Keep the post-lineage survivor filter separate from the Parquet pruning filter.
- Translate and bind pruning filters to each physical file schema.
- Preserve original row positions when row groups are skipped.
- Skip early pruning when a missing field has a non-null `initial-default`.
- Validate bound task residuals against the scan schema.
- Normalize negated filters before collecting bloom predicates.

## Benchmark

The benchmark is in `table/row_lineage_rewrite_bench_test.go`.

It covers 1M rows split into 100 Parquet row groups and measures the end-to-end copy-on-write delete path for 1/100, 10/100, and 100/100 retained groups.

## Tests

- `go test -vet=off ./... -count=1`
- `go test -vet=off -race ./table -run "Test(CoWRewritePrunesRowGroupsBeforeLineageSynthesis|CoWRewriteDoesNotPruneMissingInitialDefault|ProcessRecordsUsesRowGroupFilterForPruning|ProcessRecordsRebindsRowGroupFilterToPromotedFileSchema|ProcessRecordsDoesNotPruneMissingInitialDefault)$" -count=1`
- `go vet ./...`
- `go test -vet=off ./table -run "^$" -bench "^BenchmarkRowLineageRewriteRowGroupPruning$" -benchtime=1x -count=1`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"iter"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -685,6 +686,7 @@ type arrowScan struct {
 	// pruning. It lets callers keep boundRowFilter as AlwaysTrue while they
 	// must evaluate the real row filter after position-dependent enrichment.
 	rowGroupFilter iceberg.BooleanExpression
+	filterSchema   *iceberg.Schema
 	caseSensitive  bool
 	rowLimit       int64
 	options        iceberg.Properties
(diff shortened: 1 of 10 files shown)


zeroshade:

zeroshade (approved):

PR #1682, opened 2026-08-06
feat(table): commit partial rewrites per group

fallintoplace:
## Summary

Make partial compaction progress durable one group at a time.

## Why

PartialProgress previously staged separate updates in memory but sent them in one catalog commit. A process failure before that commit lost all completed groups.

## What changed

Commit each non-empty group through a fresh child transaction and keep the latest committed table on the parent transaction. The existing final Commit call returns that table, while additional work on the completed parent is rejected.

## Tests

- `go test ./table -run TestRewriteDataFiles_PartialProgress -count=1`
- `go test ./table -run TestRewriteDataFiles -count=1`

### DIFF
--- a/cmd/iceberg/compact.go
+++ b/cmd/iceberg/compact.go
@@ -164,9 +164,11 @@ func compactRun(ctx context.Context, output Output, tbl *table.Table, plan compa
 		os.Exit(1)
 	}
 
-	if _, err := tx.Commit(ctx); err != nil {
-		output.Error(fmt.Errorf("commit failed: %w", err))
-		os.Exit(1)
+	if !cfg.partialProgress {
+		if _, err := tx.Commit(ctx); err != nil {
+			output.Error(fmt.Errorf("commit failed: %w", err))
+			os.Exit(1)
+		}
 	}
 
 	totalRemovedDeletes := result.RemovedPositionDeleteFiles + result.RemovedEqualityDeleteFiles
--- a/cmd/iceberg/compact.go
+++ b/cmd/iceberg/compact.go
@@ -165,8 +165,16 @@ func compactRun(ctx context.Context, output Output, tbl *table.Table, plan compa
 		os.Exit(1)
 	}
 
-	if _, err := tx.Commit(ctx); err != nil {
-		output.Error(fmt.Errorf("commit failed: %w", err))
+	if !cfg.partialProgress {
+		if _, err := tx.Commit(ctx); err != nil {
+			output.Error(fmt.Errorf("commit failed: %w", err))
+			os.Exit(1)
+		}
+	}
+	if len(result.FailedGroups) > 0 {
+		failure := result.FailedGroups[0]
+		output.Error(fmt.Errorf("compaction failed for %d group(s); first group %q: %w",
+			len(result.FailedGroups), failure.PartitionKey, failure.Err))
 		os.Exit(1)
 	}
 
--- a/table/rewrite_data_files.go
+++ b/table/rewrite_data_files.go
@@ -446,7 +446,11 @@ func ExecuteCompactionGroup(ctx context.Context, tbl *Table, group CompactionTas
 	)
 	for df, err := range WriteRecords(ctx, tbl, arrowSchema, records, writeOpts...) {
 		if err != nil {
-			return CompactionGroupResult{}, fmt.Errorf("write compacted files for group %q: %w", group.PartitionKey, err)
+			return CompactionGroupResult{
+				PartitionKey: group.PartitionKey,
+				NewDataFiles: newFiles,
+				BytesAfter:   bytesAfter,
+			}, fmt.Errorf("write compacted files for group %q: %w", group.PartitionKey, err)
 		}
 		newFiles = append(newFiles, df)
 		bytesAfter += df.FileSizeBytes()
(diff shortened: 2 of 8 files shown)


table/rewrite_data_files.go
```diff
@@ -433,64 +488,204 @@ func allTasksHaveRowLineage(tasks []FileScanTask) bool {
...
 			continue
 		}
 
-		gr, err := ExecuteCompactionGroup(ctx, t.tbl, group, opts.GroupOptions...)
+		fs, err := current.fsF(ctx)
 		if err != nil {
-			return result, err
+			return result, fmt.Errorf("open table IO for partial rewrite batch: %w", err)
+		}
+		deadPositionDeletes, err := CollectDeadPositionDeletes(
```
zeroshade, line 600: **Blocking:** This always resolves `main`, even when `t.branch` targets an existing divergent branch. `deletesToRemove` can therefore be calculated from main’s survivors and remove a shared partition-scoped position delete that is still needed by a branch-only file, resurrecting rows on that branch. Please resolve the snapshot with the same branch-aware semantics as the child transaction (including the main fallback only when creating a new branch), and add a divergent-main/branch regression test.

table/rewrite_data_files.go
```diff
@@ -433,64 +488,204 @@ func allTasksHaveRowLineage(tasks []FileScanTask) bool {
...
 func (t *Transaction) rewriteDataFilesPartial(ctx context.Context, groups []CompactionTaskGroup, opts RewriteDataFilesOptions) (*RewriteResult, error) {
-	result := &RewriteResult{}
-	props := maps.Clone(opts.SnapshotProps)
-	var allRewritten []iceberg.DataFile
+	result := &RewriteResult{Table: t.tbl}
+	meta, err := t.txnMeta()
+	if err != nil {
+		return nil, err
+	}
+	if len(meta.updates) > 0 || len(t.reqs) > 0 || len(t.validators) > 0 {
```
zeroshade, line 511: This freshness check does not include `t.committed`. After the first successful partial batch, `recordCommittedRewriteBatch` marks the parent terminal, but a second call to `RewriteDataFiles(...PartialProgress: true)` proceeds and commits more child transactions. I reproduced this directly: the second call returned `nil`. Please reject an already committed/unknown-state parent before performing any rewrite I/O and cover the reuse case.

cmd/iceberg/compact.go
```diff
@@ -165,9 +165,11 @@ func compactRun(ctx context.Context, output Output, tbl *table.Table, plan compa
 		os.Exit(1)
 	}
 
-	if _, err := tx.Commit(ctx); err != nil {
-		output.Error(fmt.Errorf("commit failed: %w", err))
-		os.Exit(1)
+	if !cfg.partialProgress {
```
zeroshade, line 168: `RewriteDataFiles` can return `nil` with `result.FailedGroups` populated while failures remain within the configured limit. This path then skips the parent commit, ignores those failures, and later prints `Done`; with the current default (`MaxFailedCommits == 0`, unlimited), even every batch can fail while the command exits successfully. Please report/check failed groups and return a nonzero status, or expose an explicit failure-tolerance contract in the CLI instead of silently succeeding.

table/rewrite_data_files.go
```diff
@@ -433,64 +488,204 @@ func allTasksHaveRowLineage(tasks []FileScanTask) bool {
...
+			// ErrCommitFailed is the only error that proves the catalog did not
+			// commit. Every other error leaves commit state unknown, so continuing
+			// with the old table could apply later batches on stale state.
+			if !errors.Is(err, ErrCommitFailed) {
+				t.committed = true
+
+				return result, err
+			}
+
+			failedCommits++
```
zeroshade, line 662: At this point every `NewDataFiles` output for the batch has already been written. `ErrCommitFailed` is the known-not-committed case, but this path records the failure and continues without deleting those outputs; `FailedGroups` also does not expose their paths to the caller. Repeated tolerated failures therefore leak full unreferenced compacted files. Please remove the batch’s generated files before continuing and add a cleanup regression test.

zeroshade (changes requested):
Thanks for implementing partial-progress rewrites. I’m requesting changes for the four correctness issues called out inline. The branch-snapshot issue is blocking because it can change row-deletion semantics on a divergent branch.

Additionally, all nine commits currently lack the `Signed-off-by:` trailer required by `CONTRIBUTING.md`. Please rebase or squash with sign-off before this is merged.

I verified the exact PR head with the rewrite/table/command tests, targeted race tests, `go vet`, and `git diff --check`; those checks pass. I also reproduced the completed-parent reuse issue with a focused regression test: a second partial rewrite on the same parent returned `nil` instead of being rejected.

table/rewrite_data_files.go
```diff
@@ -433,64 +484,237 @@ func allTasksHaveRowLineage(tasks []FileScanTask) bool {
...
+		rewrittenFiles := make([]iceberg.DataFile, 0)
+
+		for _, group := range batchGroups {
+			if err := ctx.Err(); err != nil {
+				return result, err
+			}
+
+			gr, err := ExecuteCompactionGroup(ctx, current, group, opts.GroupOptions...)
+			if err != nil {
+				return result, err
```
zeroshade: These pre-commit exits can orphan files already written by earlier groups in the same batch. For example, with two groups and `MaxCommits: 1`, group 1 can write its replacement successfully and group 2 can fail here while reading its source; `batchResults` then owns group 1's `NewDataFiles`, but this return leaves them on storage. I reproduced that case and observed the extra generated Parquet file remain. Please clean generated batch outputs on every known pre-commit failure path (context cancellation, group execution, delete collection, transaction creation, and staging), while continuing not to clean on unknown catalog commit outcomes. `ExecuteCompactionGroup` also needs to clean files already yielded before a later `WriteRecords` error, since its current error return discards those paths.

zeroshade (changes requested):
The previous findings are addressed, and the latest style-only follow-up is clean. One remaining pre-commit output-ownership issue still needs correction; details are inline.

zeroshade (approved):

PR #1683, opened 2026-08-06
fix(table): close stale scan plan IO

fallintoplace:
## Summary

Close plan-scoped IO when scan planning finishes or a plan is replaced.

## Why

Remote replanning and plan-loading errors could leave the previous plan IO open, which leaks resources during retries and failed reads.

## What changed

Close local plan IO after planning, close the previous remote plan before replacing it, and close plan IO when ReadTasks fails to load or initialize a plan.

## Tests

- `go test ./table -run TestScanPlanningRemote -count=1`
- `go test ./table -run TestReadTasksClosesPlanIOWhenLoadFails -count=1`

### DIFF
--- a/table/scan_planning.go
+++ b/table/scan_planning.go
@@ -110,16 +110,19 @@ type ScanPlanningRequest struct {
 }
 
 // PlanIO lazily loads the FileIO used to read a planned scan and closes any
-// resources it holds (e.g. plan-scoped credentials) once reading is done. Nil
-// means the scan keeps using the table's normal FileIO. Remote planners may
-// return a PlanIO backed by plan-scoped storage credentials.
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
This turns the plan-IO lifecycle from a single-owner close-after-first-iterator into a proper refcounted lease, and every rule has a test pinning it: derived scans retain ownership (fixing the latent bug where `UseRowLimit`/`UseRef` copies shared a plan the first finisher would close under the other), replanning replaces only on success so a failed replan can't destroy a usable plan, retired plans refuse new leases while in-flight iterators keep theirs, the same-`PlanIO` dedup avoids pointless churn, and the comparable-type validation turns a would-be panic into a clear error. The honest doc updates — including the accepted unread-iterator edge — are appreciated. Clean, careful work.

---

PR #1684, opened 2026-08-06
fix(catalog/sql): preserve close errors

fallintoplace:
## Summary

Preserve both reporter and database errors when closing a SQL catalog.

## Why

The close path returned only the first error, hiding a database close failure when the metrics reporter also failed.

## What changed

Join the database close error with the reporter close error and add a regression test using failures from both resources.

## Tests

- `go test ./catalog/sql -run TestCloseReturnsReporterAndDatabaseErrors -count=2`

### DIFF
--- a/catalog/sql/close_test.go
+++ b/catalog/sql/close_test.go
@@ -1,0 +1,89 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package sql
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"errors"
+	"testing"
+
+	"github.com/apache/iceberg-go/metrics"
+	"github.com/stretchr/testify/require"
+	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/dialect/sqlitedialect"
+)
+
+const (
+	closeErrorDriverName   = "iceberg_go_sql_close_error"
+	closeErrorReporterName = "iceberg_go_sql_close_reporter"
+)
+
+var closeErrorDBErr error
+
+func init() {
+	sql.Register(closeErrorDriverName, closeErrorDriver{})
+}
+
+type closeErrorDriver struct{}
+
+func (closeErrorDriver) Open(string) (driver.Conn, error) {
+	return closeErrorConn{err: closeErrorDBErr}, nil
+}
+
+type closeErrorConn struct{ err error }
+
+func (c closeErrorConn) Prepare(string) (driver.Stmt, error) { return nil, driver.ErrSkip }
+func (c closeErrorConn) Close() error                        { return c.err }
+func (closeErrorConn) Begin() (driver.Tx, error)             { return nil, driver.ErrSkip }
+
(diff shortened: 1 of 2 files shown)


catalog/sql/close_test.go
```diff
@@ -0,0 +1,89 @@
...
+type closeErrorReporter struct{ err error }
+
+func (closeErrorReporter) Report(context.Context, metrics.MetricsReport) {}
+func (r closeErrorReporter) Close() error                                { return r.err }
+
+func TestCloseReturnsReporterAndDatabaseErrors(t *testing.T) {
+	dbErr := errors.New("database close")
+	reporterErr := errors.New("reporter close")
+
+	closeErrorDBErr = dbErr
```
laskoviymishka: The reporter side resets itself with `t.Cleanup` (line 70), but `closeErrorDBErr` never does, so it leaks across runs: under `go test -count=2` the second run starts with the stale error still set. I'd add `t.Cleanup(func() { closeErrorDBErr = nil })` right here to match.

While we're here, `Open` reads this global (line 47) on whatever goroutine database/sql calls it from, and we write it here with no synchronization, so `go test -race` on this package would flag it. Threading the error through the DSN instead of a package global would kill both problems at once, since the driver would read it per-connection and there'd be nothing shared to reset. wdyt?

catalog/sql/close_test.go
```diff
@@ -0,0 +1,89 @@
...
+	metrics.Register(closeErrorReporterName, func(map[string]string) (metrics.Reporter, error) {
+		return closeErrorReporter{err: reporterErr}, nil
+	})
+	t.Cleanup(func() {
+		metrics.Deregister(closeErrorReporterName)
+	})
+
+	db, err := sql.Open(closeErrorDriverName, "")
+	require.NoError(t, err)
+	require.NoError(t, db.Ping())
```
laskoviymishka, line 73: `db.Ping()` is doing more than it looks here: it's the only thing that opens a connection into the pool, which is what gives `bun.DB.Close()` something to drain so `closeErrorConn.Close()` actually fires. Worth a one-line comment saying so, otherwise someone trimming setup could drop it and quietly lose the db-error-path coverage without any test going red.

catalog/sql/sql.go
```diff
@@ -455,8 +455,8 @@ func (c *Catalog) CatalogType() catalog.Type {
 func (c *Catalog) Close() error {
 	err := c.reporter.Close()
 	if c.ownsDB {
-		if dbErr := c.db.Close(); dbErr != nil && err == nil {
-			err = dbErr
+		if dbErr := c.db.Close(); dbErr != nil {
+			err = errors.Join(err, dbErr)
```
laskoviymishka: This is the right fix for the both-fail case, but it also changes the common single-failure path: when the reporter closes cleanly and only the DB errors, `errors.Join(nil, dbErr)` wraps `dbErr` in a `*joinError` rather than returning it directly like the old code did. `errors.Is`/`As` still traverse fine, but direct equality and `errors.Unwrap` on the result no longer see `dbErr`.

If we want to keep the old identity for the single-error case:

```go
if dbErr := c.db.Close(); dbErr != nil {
    if err != nil {
        err = errors.Join(err, dbErr)
    } else {
        err = dbErr
    }
}
```

Not blocking since the test uses `ErrorIs`, but it's a cheap way to avoid surprising any caller doing value comparisons.

laskoviymishka (approved):
Nice, this is a clean fix. `errors.Join` is the right call here and it lines up with the cleanup pattern already used elsewhere in the file. Good to see the close error stop getting dropped.

Couple of small things, none blocking a merge. The test threads the db-close error through a mutable package global that `Open` reads and that never gets reset. The reporter global right beside it already resets via `t.Cleanup`, so it's a little inconsistent, and as-is it's stale under `go test -count=2` and a `-race` hazard. Threading it through the DSN instead would tidy both. Separately, `errors.Join(nil, dbErr)` does change the returned error identity on the single-failure path (reporter fine, DB errors); `errors.Is` still works so it's not a real problem, but the guard-only-when-both-non-nil shape is a cheap way to keep the old value.

None of that blocks merge for me, the fix itself looks good. Details inline.

zeroshade (approved):
Correct and minimal — both close failures now survive, and keeping the single-failure case unwrapped (rather than unconditionally joining) preserves error identity for existing callers. The dual-failure regression test through a real driver/reporter pair is exactly the right shape.

---

PR #1686, opened 2026-08-06
fix(table): track parents of added columns so nested moves apply

### LINKED ISSUE #1685: UpdateSchema: moving a nested column added in the same transaction is silently ignored
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

When a nested column is added via `AddColumn` / `UnionByNameWith` and then reordered with `MoveFirst`/`MoveBefore`/`MoveAfter` in the same `UpdateSchema`, the move is silently discarded (or fails to resolve a sibling), and the applied schema keeps the field at its default appended position.

Reason for this: `parentID` map, which is seeded once in `initParentID` is never updated. `moveColumn` to `findParentID` therefore returns `TableRootID (-1)`, thus filing under the root struct and dropped by `moveFields` during apply.

Top-level added columns are unaffected.

### Contribution

Will submit a PR for the fix.

badalprasadsingh:
## Description

Fixes #1685 

Record the parent id of freshly-added fields, so a nested column added and moved in the same transaction resolves to the correct parent struct instead of the table root.

## Testing

Added the regression test covering it : `TestMoveAddedNestedColumn`

### DIFF
--- a/table/update_schema.go
+++ b/table/update_schema.go
@@ -357,6 +357,10 @@ func (u *UpdateSchema) addColumn(path []string, fieldType iceberg.Type, doc stri
 	}
 	u.adds[parentID] = append(u.adds[parentID], sch.Field(0))
 	u.addedNameToID[fullName] = sch.Field(0).ID
+	// Skipping TableRootID as findParentID already defaults unknown ids to it.
+	if parentID != TableRootID {
+		u.parentID[sch.Field(0).ID] = parentID
+	}
 
 	return nil
 }
@@ -785,6 +789,10 @@ func (u *UpdateSchema) unionAddColumn(path []string, newField iceberg.NestedFiel
 	}
 	u.adds[parentID] = append(u.adds[parentID], sch.Field(0))
 	u.addedNameToID[fullName] = sch.Field(0).ID
+	// Skipping TableRootID as findParentID already defaults unknown ids to it.
+	if parentID != TableRootID {
+		u.parentID[sch.Field(0).ID] = parentID
+	}
 
 	return nil
 }
--- a/table/update_schema_test.go
+++ b/table/update_schema_test.go
@@ -742,6 +742,119 @@ func TestMoveColumn(t *testing.T) {
 	})
 }
 
+func addressFieldNames(t *testing.T, schema *iceberg.Schema) []string {
+	t.Helper()
+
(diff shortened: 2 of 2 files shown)


badalprasadsingh:
While working on this fix, I ended up finding a few more bugs in `UpdateSchema`. Will raise PRs for those as well.

Looking forward to contribute more over the weekend :)

table/update_schema.go
```diff
@@ -357,6 +357,10 @@ func (u *UpdateSchema) addColumn(path []string, fieldType iceberg.Type, doc stri
 	}
 	u.adds[parentID] = append(u.adds[parentID], sch.Field(0))
 	u.addedNameToID[fullName] = sch.Field(0).ID
```
zeroshade, line 359: Only the added root's name and ID are indexed here. If `fieldType` is a struct/list/map, none of its freshly assigned descendants are entered in `addedNameToID` or `parentID`, so `findFieldForMove` cannot resolve a move within that newly added complex field. The analogous union path has the same one-level indexing. **Suggested fix:** recursively index every added descendant's full name and parent relationship, then apply moves inside the staged complex field.

zeroshade (changes requested):
Thanks for preserving per-parent move order and for keeping the cross-parent rejection in place; those parts are sound.

The parent tracking is still only one level deep, so moves within a newly added complex field remain unresolved, as noted inline. A second blocking path is at `table/update_schema.go:1363-1380`: `moveFields` removes the source field before confirming that a `MoveBefore`/`MoveAfter` anchor still exists. If the relative anchor was staged for deletion, the `continue` silently discards the user's moved field. The silence is the critical problem here: a requested schema change is lost without an error. **Suggested fix:** resolve and validate both the source and relative field before mutating the slice, and reject moves relative to staged deletions.

This needs to be coordinated with #1688. Both changes require ancestor-aware tracking in `table/update_schema.go`; merging either one alone leaves silent-loss paths. **Suggested fix:** use a shared ancestor-aware model for staged additions, deletions, and move resolution. Please add coverage for newly added complex roots, list/map descendants, multiple moves under one parent, and `MoveBefore`/`MoveAfter` against deleted anchors.

badalprasadsingh:
Will merge these PR changes into #1688! As they both have a lot of common logical fixes.

Thanks for the review.

PR #1688, opened 2026-08-06
fix(table): reject AddColumn/UnionByNameWith targeting a column staged for deletion

### LINKED ISSUE #1687: UpdateSchema: AddColumn silently discards nested columns added under a parent staged for deletion
### Apache Iceberg version

None

### Please describe the bug 🐞

When a column is deleted and a new nested column is added under that same (now-deleted) parent within the same `UpdateSchema` or `Transaction` chain, `AddColumn` succeeds without error, but `Apply()` / `Commit()` silently drops the newly added column because the parent struct field is removed from the schema before the visitor considers pending additions under it.

What is expected? An error, since the parent is already deleted

### Status

I'm working on this and will be submitting a PR.

badalprasadsingh:
### Description

Fixes #1687 

Right now, `UpdateSchema.addColumn` and `unionAddColumn` resolved the parent field against the original schema without checking whether that parent was already staged for deletion in the same builder chain. This let `Apply()`/`Commit()` succeed while silently dropping the newly added nested column, because `applyChanges.Field` drops deleted fields (and everything staged under them) before considering pending additions.

Added an explicit `u.isDeleted(parentID)` check right after resolving the parent field in both `addColumn` and `unionAddColumn`, returning an error instead of silently staging an addition that will be discarded.

### Testing

Added: the required **regression** tests for it along with, a **symmetry** test and an, **invariant** test

### DIFF
--- a/table/update_schema.go
+++ b/table/update_schema.go
@@ -320,6 +320,10 @@ func (u *UpdateSchema) addColumn(path []string, fieldType iceberg.Type, doc stri
 		}
 
 		parentID = parentField.ID
+
+		if u.isDeleted(parentID) {
+			return fmt.Errorf("cannot add to a column that will be deleted: %s", parentFullPath)
+		}
 	}
 
 	name := path[len(path)-1]
@@ -746,6 +750,10 @@ func (u *UpdateSchema) unionAddColumn(path []string, newField iceberg.NestedFiel
 		}
 
 		parentID = parentField.ID
+
+		if u.isDeleted(parentID) {
+			return fmt.Errorf("cannot add to a column that will be deleted: %s", parentFullPath)
+		}
 	}
 
 	name := path[len(path)-1]
--- a/table/update_schema.go
+++ b/table/update_schema.go
@@ -265,6 +265,38 @@ func (u *UpdateSchema) findParentID(fieldID int) int {
 	return parentID
 }
 
+// findDeletedAncestor walks fieldID and all of its ancestors (following the parent chain
+// up to the table root) and returns the id of the first field staged for deletion, if any.
+//
+// This prevents adding a column under a deleted ancestor, even when the
+// immediate parent is still present.
+func (u *UpdateSchema) findDeletedAncestor(fieldID int) (int, bool) {
+	for id := fieldID; id != TableRootID; id = u.findParentID(id) {
+		if u.isDeleted(id) {
+			return id, true
+		}
+	}
+
+	return TableRootID, false
+}
+
+// hasStagedAddUnder reports whether any column addition is staged
+// beneath fieldID or any of its descendants.
+func (u *UpdateSchema) hasStagedAddUnder(fieldID int) bool {
+	for addParentID, added := range u.adds {
+		if len(added) == 0 {
+			continue
+		}
+		for id := addParentID; id != TableRootID; id = u.findParentID(id) {
+			if id == fieldID {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
 func (u *UpdateSchema) AddColumn(path []string, fieldType iceberg.Type, doc string, required bool, defaultValue iceberg.Literal) *UpdateSchema {
 	u.ops = append(u.ops, func() error {
 		return u.addColumn(path, fieldType, doc, required, defaultValue)
@@ -321,8 +353,13 @@ func (u *UpdateSchema) addColumn(path []string, fieldType iceberg.Type, doc stri
 
 		parentID = parentField.ID
 
-		if u.isDeleted(parentID) {
-			return fmt.Errorf("cannot add to a column that will be deleted: %s", parentFullPath)
+		if ancestorID, ok := u.findDeletedAncestor(parentID); ok {
+			deletedName, found := u.schema.FindColumnName(ancestorID)
+			if !found {
+				deletedName = parentFullPath
+			}
+
+			return fmt.Errorf("cannot add to a column that will be deleted: %s", deletedName)
 		}
 	}
 
@@ -380,7 +417,10 @@ func (u *UpdateSchema) deleteColumn(path []string) error {
 		return fmt.Errorf("field not found: %s", fullName)
 	}
 
-	if _, ok := u.adds[field.ID]; ok {
+	// Reject deletion when an addition is staged beneath this field or any of its descendants;
+	// applyChanges drops the whole subtree, so any such add keyed under a descendant
+	// would otherwise be silently discarded.
+	if u.hasStagedAddUnder(field.ID) {
 		return fmt.Errorf("field that has additions cannot be deleted: %s", fullName)
 	}
 
@@ -751,8 +791,13 @@ func (u *UpdateSchema) unionAddColumn(path []string, newField iceberg.NestedFiel
 
 		parentID = parentField.ID
 
-		if u.isDeleted(parentID) {
-			return fmt.Errorf("cannot add to a column that will be deleted: %s", parentFullPath)
+		if ancestorID, ok := u.findDeletedAncestor(parentID); ok {
+			deletedName, found := u.schema.FindColumnName(ancestorID)
+			if !found {
+				deletedName = parentFullPath
+			}
+
+			return fmt.Errorf("cannot add to a column that will be deleted: %s", deletedName)
 		}
 	}
 
--- a/table/update_schema_test.go
+++ b/table/update_schema_test.go
@@ -490,6 +490,116 @@ func TestAddColumn(t *testing.T) {
 	})
 }
 
+// threeLevelStructSchema builds: id, outer{ inner{ leaf } }
+func threeLevelStructSchema() *iceberg.Schema {
+	return iceberg.NewSchema(1,
+		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int32, Required: true},
+		iceberg.NestedField{ID: 2, Name: "outer", Required: false, Type: &iceberg.StructType{
+			FieldList: []iceberg.NestedField{
+				{ID: 3, Name: "inner", Required: false, Type: &iceberg.StructType{
+					FieldList: []iceberg.NestedField{
+						{ID: 4, Name: "leaf", Type: iceberg.PrimitiveTypes.String, Required: false},
+					},
+				}},
+			},
+		}},
+	)
+}
+
+// listOfStructSchema builds: id, data list<struct{ a }>
+func listOfStructSchema() *iceberg.Schema {
+	return iceberg.NewSchema(1,
+		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int32, Required: true},
+		iceberg.NestedField{ID: 2, Name: "data", Required: false, Type: &iceberg.ListType{
+			ElementID:       3,
+			ElementRequired: false,
+			Element: &iceberg.StructType{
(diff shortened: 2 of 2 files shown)


table/update_schema.go
```diff
@@ -320,6 +320,10 @@ func (u *UpdateSchema) addColumn(path []string, fieldType iceberg.Type, doc stri
 		}
 
 		parentID = parentField.ID
+
+		if u.isDeleted(parentID) {
```
zeroshade: This checks only the immediate parent. If `outer` is staged for deletion and the add targets `outer.inner.new_leaf`, `parentID` is `inner`, so this passes and the later schema application silently discards the requested addition with `outer`. **Suggested fix:** walk all ancestors of the resolved parent, reject if any is staged for deletion, and report the actual deleted ancestor.

table/update_schema.go
```diff
@@ -746,6 +750,10 @@ func (u *UpdateSchema) unionAddColumn(path []string, newField iceberg.NestedFiel
...
```
zeroshade: `UnionByNameWith` has the same immediate-parent gap: an addition below a descendant of a deleted ancestor is accepted and then silently lost. **Suggested fix:** apply the same full ancestor-deletion check here and report the deleted ancestor rather than only the immediate parent path.

zeroshade (changes requested):
Thanks for adding symmetric checks for direct parent deletion and preserving delete-then-re-add with a different type; that legitimate pattern remains supported.

The new checks only cover immediate parents, so ancestor deletion still silently discards user-requested additions in either operation order. For the reverse order, `table/update_schema.go:383-385` checks only `u.adds[field.ID]`. If an add is staged beneath `outer.inner` and `outer` is then deleted, the addition is keyed by the descendant parent and deletion of `outer` succeeds. The requested addition is silently lost rather than rejected. **Suggested fix:** reject deletion when the field contains any staged-add parent, walking the full descendant/ancestor relationship.

This must be coordinated with #1686: both changes need ancestor-aware tracking in `table/update_schema.go`, and merging either one alone leaves silent-loss paths. **Suggested fix:** share ancestor traversal for direct adds, `UnionByNameWith`, deletions, and moves. Errors should identify the actual deleted ancestor. Please test both operation orders for multi-level ancestors, both `UnionByNameWith` orders, list-element structs, map-value structs, and assertions naming the conflicting ancestor.

table/update_schema.go
badalprasadsingh: Done.

Now, `findDeletedAncestor` walks the full ancestor chain and reports the deleted ancestor's name.

The case: `delete outer`, add `outer.inner.new_leaf` : now errors `cannot add to a column that will be deleted: outer`. This is now covered by `TestAddColumnUnderDeletedAncestor` (delete-then-add, all nesting kinds).

table/update_schema.go
badalprasadsingh: Done.

Now, `unionAddColumn` uses `findDeletedAncestor` and names the deleted ancestor. Covered by `TestUnionByNameRejectsAddUnderDeletedAncestor` in both orders (delete-then-union and union-then-delete).

badalprasadsingh:
Addressed. Added shared ancestor traversal, which now prevents the silent-loss in both the operation orders.

Thanks for the review @zeroshade 🤩

badalprasadsingh:
Will merge the PR #1686 changes into this, and create a single PR. 

Drafting it now. Please review it again once, I re-open this.

table/update_schema.go
```diff
@@ -320,6 +352,15 @@ func (u *UpdateSchema) addColumn(path []string, fieldType iceberg.Type, doc stri
 		}
 
 		parentID = parentField.ID
+
+		if ancestorID, ok := u.findDeletedAncestor(parentID); ok {
```
laskoviymishka, line 356: This walks the full ancestor chain, but Java's `SchemaUpdate` only checks the immediate resolved parent (`!deletes.contains(parentId)`), and PyIceberg does the same. So `DeleteColumn(["outer"]).AddColumn(["outer","inner","new_leaf"])` is rejected here but succeeds against Java/Py — a real cross-client divergence where the same script passes on one engine and fails at `Apply()` on ours.

The reason Java can afford the narrow check is that its apply-time tree-drop makes "delete wins" a defined, non-lossy outcome. Ours drops silently, which is exactly the hole @zeroshade asked to close — so narrowing back to `isDeleted(parentID)` would reopen it unless we also made the drop explicit.

So I don't think "just match Java" is right here given the thread above. But I'd want the divergence chosen deliberately: either keep the walk and document that we're intentionally stricter than Java/Py, or narrow the check and fix the silent drop at apply time. wdyt — and worth pulling @zeroshade in since it's his ancestor-awareness ask?

table/update_schema.go
```diff
@@ -320,6 +352,15 @@ func (u *UpdateSchema) addColumn(path []string, fieldType iceberg.Type, doc stri
...
+			deletedName, found := u.schema.FindColumnName(ancestorID)
+			if !found {
+				deletedName = parentFullPath
```
laskoviymishka, line 359: When `FindColumnName(ancestorID)` misses, the fallback uses `parentFullPath`, which is the immediate parent's path, not the deleted ancestor's. For the deleted-`outer` / add-`outer.inner.new_leaf` case the error would then read `...: outer.inner` and point at the wrong field.

In practice an id that's in `u.deletes` should always resolve, so this is a dead branch today — but it's a wrong-field branch, and the same fallback is duplicated in `unionAddColumn`. I'd resolve the fallback name from `ancestorID` itself (or just drop the name) so the two paths can't drift into lying.

table/update_schema.go
```diff
@@ -376,7 +417,10 @@ func (u *UpdateSchema) deleteColumn(path []string) error {
 		return fmt.Errorf("field not found: %s", fullName)
 	}
 
-	if _, ok := u.adds[field.ID]; ok {
+	// Reject deletion when an addition is staged beneath this field or any of its descendants;
+	// applyChanges drops the whole subtree, so any such add keyed under a descendant
+	// would otherwise be silently discarded.
+	if u.hasStagedAddUnder(field.ID) {
```
laskoviymishka, line 423: `hasStagedAddUnder` closes the add-under-deleted-subtree hole, but the same mechanism still exists for staged updates. `UpdateColumn(["a","b","c"])` then `DeleteColumn(["a"])` stores the update keyed at `c`'s id, this guard doesn't see it, and `applyChanges` drops the subtree before the update runs — lost exactly the way adds were.

I'd add a `hasStagedUpdateUnder` twin over `u.updates` and call it alongside this one, plus a test that registers an update on a grandchild then deletes the ancestor. Otherwise a reader sees the add case fixed and reasonably concludes the whole class is closed.

table/update_schema.go
```diff
@@ -265,6 +265,38 @@ func (u *UpdateSchema) findParentID(fieldID int) int {
...
+}
+
+// hasStagedAddUnder reports whether any column addition is staged
+// beneath fieldID or any of its descendants.
+func (u *UpdateSchema) hasStagedAddUnder(fieldID int) bool {
+	for addParentID, added := range u.adds {
+		if len(added) == 0 {
+			continue
+		}
+		for id := addParentID; id != TableRootID; id = u.findParentID(id) {
```
laskoviymishka, line 290: The termination here is correct but silent and load-bearing in two ways. The `id != TableRootID` bound means a top-level pending add (parent == root) never matches, so `AddColumn(["new_top"])` then `DeleteColumn(["unrelated_top"])` correctly succeeds — and separately, `findParentID` returns `TableRootID` for ids it doesn't know while `u.parentID` is only seeded from the committed schema, so a freshly-assigned add id exits the walk immediately rather than looping forever.

Both are right today, but they're resting on an undocumented coupling between the sentinel and this bound. I'd add a one-line comment naming the invariant and a test for the root-level-add-then-unrelated-delete case, so this can't quietly regress into either a false block or a spin if the bound ever changes.

table/update_schema_test.go
```diff
@@ -407,6 +407,197 @@ func TestAddColumn(t *testing.T) {
...
+		_, err := NewUpdateSchema(txn, true, true).
+			DeleteColumn([]string{"address"}).
+			AddColumn([]string{"address", "code"}, iceberg.PrimitiveTypes.String, "", false, nil).
+			Apply()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "cannot add to a column that will be deleted: address")
+	})
+
+	t.Run("test add column under a deleted nested parent is rejected", func(t *testing.T) {
+		schema := iceberg.NewSchema(1,
```
laskoviymishka, line 424: This inline schema is the same three-level shape as `threeLevelStructSchema()` you add lower in the file. Could this subtest use `txnForSchema(t, threeLevelStructSchema())` like the table-driven cases do? Keeps the one fixture in one place.

table/update_schema_test.go
```diff
@@ -407,6 +407,197 @@ func TestAddColumn(t *testing.T) {
...
+
+		fields := structType.Fields()
+		require.Len(t, fields, 2)
+
+		cityField, ok := newSchema.FindFieldByName("address.city")
+		require.True(t, ok)
+		assert.Equal(t, iceberg.PrimitiveTypes.Int64, cityField.Type)
+		assert.Equal(t, "recreated", cityField.Doc)
+		// The re-added field must receive a fresh id, not reuse the deleted one.
+		assert.NotEqual(t, 5, cityField.ID)
```
laskoviymishka, line 489: `NotEqual(t, 5, cityField.ID)` proves the id isn't reused, which is the important part. While we're here, `assert.Greater(t, cityField.ID, testMetadata.LastColumnID())` would also pin the monotonicity invariant — otherwise a future regression that hands out some other stale-but-different id still passes this.

laskoviymishka (changes requested):
Nice follow-up — this reads as a real response to your thread, @zeroshade, not a box-tick. The ancestor-aware `findDeletedAncestor`/`hasStagedAddUnder` block the add-under-deleted-subtree race in both operation orders now, the tests cover struct/list/map ancestors and both `UnionByNameWith` orders, and delete-then-re-add-with-a-different-type is preserved. That last one is easy to break and it's still working.

I'd hold this before merging though, on two points.

Still open from your review, @zeroshade: the silent-loss class isn't fully closed. Adds are handled, but the exact same mechanism still drops staged *updates* under a deleted subtree (`UpdateColumn(["a","b","c"])` then `DeleteColumn(["a"])`), and `moveColumn` is still unguarded for deleted ancestors. Both fail the same way — apply-time tree-drop wins, the staged op vanishes. Given your #1686 coordination note, I think this wants the shared ancestor traversal you asked for rather than a third bespoke walk.

New from my pass: the ancestor walk makes us stricter than Java and PyIceberg, which only check the immediate resolved parent, not grandparents. `DeleteColumn(["outer"]).AddColumn(["outer","inner","new_leaf"])` is rejected here and accepted there. I don't think we should just match Java — its narrow check is safe only because its apply-time drop is a defined "delete wins", and ours drops silently. But I'd want this divergence chosen deliberately: keep the walk and document that we're intentionally stricter, or narrow the check and make the apply-time drop explicit. Flagged inline for both of you to weigh in.

- fallback error message names the wrong field when the ancestor name lookup misses (dead branch today, but wrong, and duplicated in `unionAddColumn`)
- `hasStagedAddUnder`'s root-level skip is correct but undocumented and untested
- a couple of test cleanups (reuse the three-level fixture, assert id monotonicity not just non-reuse)

Once the update/move gap and the Java-divergence call are settled, happy to take another pass and approve.

PR #1695, opened 2026-08-07
feat(table): add refs metadata table

fallintoplace:
## Summary

Adds the `refs` metadata table to `Table.Inspect`.

- Exposes branch and tag names, types, snapshot IDs, and retention settings.
- Uses the Iceberg metadata-table field IDs and nullable retention fields.
- Sorts refs by name so the result is deterministic.

## Tests

- `go test ./table`
- Covers schema IDs and types, branch and tag rows, optional retention values, empty refs, and out-of-range `min_snapshots_to_keep`.

## Scope

This adds metadata inspection only. It does not change snapshot-reference creation or retention behavior.


table/inspect.go
```diff
@@ -247,6 +251,87 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+func (i InspectTable) Refs(ctx context.Context) (array.RecordReader, error) {
+	arrowSchema, err := SchemaToArrowSchema(RefsSchema(), nil, true, false)
+	if err != nil {
+		return nil, fmt.Errorf("inspect refs: build arrow schema: %w", err)
+	}
+
+	type refRow struct {
+		name string
+		ref  SnapshotRef
+	}
```
laskoviymishka, line 479: tiny thing while we're here: the sibling methods accumulate into a plain `var refs []refRow` rather than `make([]refRow, 0)`, or size the make with `len(...)`. Matching them keeps this consistent and skips the eager empty alloc.

table/inspect.go
```diff
@@ -247,6 +251,87 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+		name.Append(row.name)
+		refType.Append(strings.ToUpper(string(row.ref.SnapshotRefType)))
+		snapshotID.Append(row.ref.SnapshotID)
+
+		if row.ref.MaxRefAgeMs != nil {
+			maxReferenceAge.Append(*row.ref.MaxRefAgeMs)
+		} else {
+			maxReferenceAge.AppendNull()
+		}
+		if row.ref.MinSnapshotsToKeep != nil {
```
laskoviymishka, line 512: `MinSnapshotsToKeep` is a `*int`, so on 64-bit platforms this cast silently truncates anything above `math.MaxInt32`. The concrete way it bites: `MinSnapshotsToKeepDefault` is `math.MaxInt`, so if that default ever lands in a stored ref this appends `-1` rather than erroring.

I'd guard it before the append, either returning an error when the value is out of int32 range, or switching the field to `*int32` if we want the wider change. wdyt?

table/inspect_internal_test.go
```diff
@@ -245,6 +247,152 @@ func TestInspectHistoryNoCurrentSnapshot(t *testing.T) {
...
+	require.Equal(t, iceberg.PrimitiveTypes.Int64, fields[3].Type)
+	require.Equal(t, iceberg.PrimitiveTypes.Int32, fields[4].Type)
+	require.Equal(t, iceberg.PrimitiveTypes.Int64, fields[5].Type)
+	require.True(t, fields[0].Required)
+	require.True(t, fields[1].Required)
+	require.True(t, fields[2].Required)
+	require.False(t, fields[3].Required)
+	require.False(t, fields[4].Required)
+	require.False(t, fields[5].Required)
+}
```
laskoviymishka, line 351: one gap vs. the sibling coverage: `TestInspectAllocatorOption` runs `Snapshots` through a `memory.CheckedAllocator` so a builder leak surfaces at test exit, but there's no equivalent for `Refs`. Cheap to add, and it'd catch a leak in this new builder path. I'd add a `TestInspectRefsAllocator` following that pattern.

table/inspect_internal_test.go
```diff
@@ -245,6 +247,152 @@ func TestInspectHistoryNoCurrentSnapshot(t *testing.T) {
...
+
+func TestInspectRefs(t *testing.T) {
+	tbl := historyTestTable()
+	minSnapshotsToKeep := 2
+	maxSnapshotAge := int64(3000)
```
laskoviymishka, line 355: the other inspect tests build a complete `*metadataV2` in a helper (`historyTestTable`, `snapshotsTestTable`) and hand it to `New()`. Reaching in with `tbl.metadata.(*metadataV2)` to mutate `SnapshotRefs` is more fragile: it panics if `metadata` is ever wrapped.

I'd pull a `refsTestTable()` helper that sets the branch and tag refs up front and share it between `TestInspectRefs` and `TestInspectRefsEmpty`, matching the sibling pattern. wdyt?

laskoviymishka (changes requested):
Really clean clone of the existing inspect pattern: schema IDs, names, and flags match Java's `RefsTable`, the `BRANCH`/`TAG` casing matches Java's enum names, and sorting by ref name plus the empty-refs test are exactly right.

Almost there. The one thing I'd want before merge is a guard on the `int32(*row.ref.MinSnapshotsToKeep)` cast. `MinSnapshotsToKeep` is a `*int`, and with `MinSnapshotsToKeepDefault` being `math.MaxInt`, that default landing in a ref would silently append `-1`. Either bounds-check and return an error before the append, or move the field to `*int32`. Narrow trigger, but it's silent when it's wrong.

The rest is just matching the sibling tests: a checked-allocator test for `Refs`, and pulling a `refsTestTable()` helper instead of reaching into `tbl.metadata.(*metadataV2)`. Both are inline.

Once the overflow guard's in, happy to take another pass and approve.

zeroshade (approved):
The overflow guard landed exactly as asked — explicit range check with a descriptive error before the `int32` append, closing the silent `math.MaxInt` → `-1` truncation — and the sibling-parity items (checked-allocator test for `Refs`, the `refsTestTable()` helper) are in too. Schema IDs/names/casing match Java's `RefsTable` and the sort-by-name plus empty-refs coverage was already right. `laskoviymishka` said he'd approve once the guard was in, so leaving his threads for him to close.

---

laskoviymishka (approved):
LGTM

laskoviymishka:
@fallintoplace once conflicts resolved ready for merge

PR #1696, opened 2026-08-07
feat(table): add metadata log entries table

fallintoplace:
## Summary

Adds the `metadata_log_entries` metadata table to `Table.Inspect`.

- Includes previous metadata log entries and the current metadata file.
- Resolves the latest available snapshot, schema, and sequence information as of each metadata timestamp.
- Handles missing snapshots, out-of-order entries, and equal timestamps.

## Tests

- `go test ./table`
- Covers normal history, missing snapshots, out-of-order log entries, and equal timestamps.

## Scope

This adds metadata inspection only. It does not change metadata loading or snapshot selection behavior.

### DIFF
--- a/table/inspect.go
+++ b/table/inspect.go
@@ -247,6 +247,96 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
 	return rr, nil
 }
 
+// MetadataLogEntries returns one row for every metadata file in the table's
+// metadata log, including the current metadata file. Snapshot information is
+// resolved from the snapshot log at each metadata file's timestamp.
+//
+// Columns:
+//   - timestamp (timestamptz, required): when the metadata file was written
+//   - file (string, required): metadata file location
+//   - latest_snapshot_id (long, optional): latest snapshot visible then
+//   - latest_schema_id (int, optional): schema used by that snapshot
+//   - latest_sequence_number (long, optional): sequence number of that snapshot
+//
+// The returned reader holds a single record batch. The caller must Release it.
+func (i InspectTable) MetadataLogEntries(ctx context.Context) (array.RecordReader, error) {
+	arrowSchema, err := SchemaToArrowSchema(MetadataLogEntriesSchema(), nil, true, false)
+	if err != nil {
+		return nil, fmt.Errorf("inspect metadata log entries: build arrow schema: %w", err)
+	}
+
+	entries := make([]MetadataLogEntry, 0)
+	for entry := range i.tbl.metadata.PreviousFiles() {
+		entries = append(entries, entry)
+	}
+	entries = append(entries, MetadataLogEntry{
+		MetadataFile: i.tbl.metadataLocation,
+		TimestampMs:  i.tbl.metadata.LastUpdatedMillis(),
+	})
+
+	bldr := array.NewRecordBuilder(i.alloc, arrowSchema)
+	defer bldr.Release()
+
+	timestamp := bldr.Field(0).(*array.TimestampBuilder)
+	file := bldr.Field(1).(*array.StringBuilder)
+	latestSnapshotID := bldr.Field(2).(*array.Int64Builder)
+	latestSchemaID := bldr.Field(3).(*array.Int32Builder)
+	latestSequenceNumber := bldr.Field(4).(*array.Int64Builder)
+
+	for _, entry := range entries {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+
+		timestamp.Append(arrow.Timestamp(entry.TimestampMs * 1000))
+		file.Append(entry.MetadataFile)
+
+		snapshot := latestSnapshotAt(i.tbl.metadata, entry.TimestampMs)
+		if snapshot == nil {
+			latestSnapshotID.AppendNull()
+			latestSchemaID.AppendNull()
+			latestSequenceNumber.AppendNull()
+			continue
+		}
+
+		latestSnapshotID.Append(snapshot.SnapshotID)
+		if snapshot.SchemaID != nil {
+			latestSchemaID.Append(int32(*snapshot.SchemaID))
+		} else {
+			latestSchemaID.AppendNull()
+		}
+		latestSequenceNumber.Append(snapshot.SequenceNumber)
+	}
+
+	rr, err := singleBatchReader(arrowSchema, bldr)
+	if err != nil {
+		return nil, fmt.Errorf("inspect metadata log entries: %w", err)
+	}
+
+	return rr, nil
+}
+
+// latestSnapshotAt follows the metadata-table behavior used by Java's
+// MetadataLogEntriesTable: find the snapshot-log entry at or before the
+// metadata timestamp, then expose its details only if that snapshot is still
+// present in the current metadata.
+func latestSnapshotAt(metadata Metadata, timestampMs int64) *Snapshot {
+	var snapshotID *int64
+	for entry := range metadata.SnapshotLogs() {
+		if entry.TimestampMs > timestampMs {
+			break
+		}
+
+		id := entry.SnapshotID
+		snapshotID = &id
+	}
+	if snapshotID == nil {
+		return nil
+	}
+
+	return metadata.SnapshotByID(*snapshotID)
+}
+
 // HistorySchema returns the Iceberg schema of the history metadata table. The
 // field IDs are fixed by the Iceberg metadata-tables spec and match the Java,
 // PyIceberg, and Rust clients for cross-client parity. A fresh schema value is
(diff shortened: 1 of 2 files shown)


table/inspect.go
```diff
@@ -247,6 +247,104 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+// latestSnapshotAt follows the metadata-table behavior used by Java's
+// MetadataLogEntriesTable: find the snapshot-log entry at or before the
+// metadata timestamp, then resolve its details independently. The snapshot
+// ID remains available when the snapshot itself has expired from metadata.
+func latestSnapshotAt(metadata Metadata, timestampMs int64) (int64, *Snapshot, bool) {
+	var snapshotID int64
+	var latestTimestamp int64
+	found := false
+	for entry := range metadata.SnapshotLogs() {
+		if entry.TimestampMs <= timestampMs && (!found || entry.TimestampMs > latestTimestamp) {
```
laskoviymishka: The behavior here is right and matches Java's `SnapshotUtil.snapshotIdAsOfTime`, but I'd tighten the doc comment above so nobody unwinds it later.

Two things it should call out. First, this resolves against the *current* snapshot log, so if an older snapshot was trimmed by ExpireSnapshots the result can shift to a later entry (or none). The doc's "the snapshot-log entry at or before the metadata timestamp" reads like point-in-time semantics, which it isn't under trimming — it's the correct, Java-compatible choice, it just needs to say so.

Second, the `> latestTimestamp` tie-break keeps the *first* entry on equal-ms timestamps. That matches Java, but PyIceberg's `snapshot_as_of_timestamp` iterates in reverse and returns the *last*. One line on the `>` noting the divergence is intentional would keep someone from "correcting" it. wdyt?

table/inspect.go
```diff
@@ -247,6 +247,104 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+
+		latestSnapshotID.Append(snapshotID)
+		if snapshot == nil {
+			latestSchemaID.AppendNull()
+			latestSequenceNumber.AppendNull()
+
+			continue
+		}
+		if snapshot.SchemaID != nil {
+			latestSchemaID.Append(int32(*snapshot.SchemaID))
```
laskoviymishka, line 314: `SchemaID` is spec-bounded to int32 so this never actually overflows, but the bare `int32(*snapshot.SchemaID)` is a narrowing cast that'll trip gosec G115 if we ever enable it, and the sibling tables don't cast (snapshots.go uses the value as a plain int). I'd add a short `//nolint:gosec` noting it's bounded by spec, or a guard — either's fine, just so the intent is explicit. wdyt?

table/inspect.go
```diff
@@ -247,6 +247,104 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+	if err != nil {
+		return nil, fmt.Errorf("inspect metadata log entries: build arrow schema: %w", err)
+	}
+
+	entries := make([]MetadataLogEntry, 0)
+	for entry := range i.tbl.metadata.PreviousFiles() {
+		entries = append(entries, entry)
+	}
+	entries = append(entries, MetadataLogEntry{
+		MetadataFile: i.tbl.metadataLocation,
```
laskoviymishka: If a table is ever built with an empty `metadataLocation`, this synthetic current entry writes `""` into `file`, which is a required (non-null) column — non-null but semantically empty. TestInspectMetadataLogEntriesEmpty passes a real path so it doesn't cover this. Java skips the synthetic entry when the location isn't set; I'd either match that or document the assumption that `metadataLocation` is always populated here.

table/inspect.go
```diff
@@ -247,6 +247,104 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+			latestSequenceNumber.AppendNull()
+
+			continue
+		}
+		if snapshot.SchemaID != nil {
+			latestSchemaID.Append(int32(*snapshot.SchemaID))
+		} else {
+			latestSchemaID.AppendNull()
+		}
+		latestSequenceNumber.Append(snapshot.SequenceNumber)
```
laskoviymishka, line 318: For a V1 snapshot `SequenceNumber` is the Go zero value 0, and it's appended unconditionally to this optional column, so there's no way to tell "V1, no sequence numbers" from "V2, sequence 0". Java and PyIceberg both emit 0 here too, so this matches every client and I wouldn't block on it. If we wanted to be stricter than Java we could emit null for V1 (`Version() < 2`), but that's genuinely optional.

table/inspect.go
```diff
@@ -247,6 +247,104 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+	entries := make([]MetadataLogEntry, 0)
+	for entry := range i.tbl.metadata.PreviousFiles() {
+		entries = append(entries, entry)
+	}
+	entries = append(entries, MetadataLogEntry{
+		MetadataFile: i.tbl.metadataLocation,
+		TimestampMs:  i.tbl.metadata.LastUpdatedMillis(),
+	})
+
+	bldr := array.NewRecordBuilder(i.alloc, arrowSchema)
```
laskoviymishka, line 278: This builds five column builders and can early-return on `ctx.Err()` mid-build, so I'd like a checked-allocator test asserting everything's released. TestInspectAllocatorOption only exercises Snapshots today. Could we add a TestInspectMetadataLogEntriesAllocator mirroring it with `memory.NewCheckedAllocator` and asserting zero bytes after release? wdyt?

table/inspect_internal_test.go
```diff
@@ -404,6 +404,216 @@ func TestInspectSnapshotsEmpty(t *testing.T) {
...
+	meta := &metadataV2{commonMetadata: commonMetadata{
+		SnapshotList: []Snapshot{
+			{SnapshotID: firstSnapshot},
+			{SnapshotID: secondSnapshot},
+			{SnapshotID: lateSnapshot},
+		},
+		SnapshotLog: []SnapshotLogEntry{
+			{SnapshotID: firstSnapshot, TimestampMs: 2000},
+			{SnapshotID: secondSnapshot, TimestampMs: 3000},
+			{SnapshotID: lateSnapshot, TimestampMs: 2500},
```
laskoviymishka: This fixture is out of chronological order by 500ms (2000, 3000, 2500), which passes today only because `validateChronologicalSnapshotLogs` allows a 1-minute skew tolerance. If that tolerance is ever tightened this becomes invalid metadata and the test fails for an unrelated reason. A one-line comment on whether this is meant to be genuinely out-of-order or just within-tolerance clock skew would make the intent clear.

table/inspect.go
```diff
@@ -247,6 +247,104 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+//   - latest_sequence_number (long, optional): sequence number of that snapshot
+//
+// The returned reader holds a single record batch. The caller must Release it.
+func (i InspectTable) MetadataLogEntries(ctx context.Context) (array.RecordReader, error) {
+	arrowSchema, err := SchemaToArrowSchema(MetadataLogEntriesSchema(), nil, true, false)
+	if err != nil {
+		return nil, fmt.Errorf("inspect metadata log entries: build arrow schema: %w", err)
+	}
+
+	entries := make([]MetadataLogEntry, 0)
```
laskoviymishka: Small one: `PreviousFiles()` has a known length so this could be pre-sized, or the loop collapsed with `slices.Collect` (already used in metadata.go). Not load-bearing.

laskoviymishka (approved):
Nice addition — this lines up with Java's `MetadataLogEntriesTable` really well: the equal-timestamp tie-break, keeping the snapshot ID when the snapshot itself has expired, and the microsecond timestamp all match, and the tests hit the cases that matter.

The one thing I'd like to settle before merge is the documentation around `latestSnapshotAt`. The behavior is right, but the doc comment reads like point-in-time semantics when it actually resolves against the current snapshot log (so trimming can move the answer), and the equal-timestamp tie-break quietly diverges from PyIceberg. Both are the correct, Java-compatible calls — they just need a line or two of comment so nobody unwinds them later.

A few smaller things I left inline:
- the `int32(*snapshot.SchemaID)` cast is unguarded (fine by spec, but it'll trip G115 and the sibling tables don't cast)
- empty `metadataLocation` would write `""` into the required `file` column
- worth a checked-allocator test for the new method, mirroring the Snapshots one

None of these block — happy to approve once the `latestSnapshotAt` semantics are documented. Nice work.

zeroshade (approved):
Matches Java's `MetadataLogEntriesTable` on all the subtle points — first-entry tie-break, snapshot-ID retention past expiry, µs timestamps — and the `latestSnapshotAt` doc comment now records exactly why the current-log resolution and the PyIceberg divergence are deliberate, which was the one thing left from the earlier review. The clock-skew scan-all test with the intentional timestamp inversion is a particularly good pin.

---

PR #1697, opened 2026-08-07
feat(table): add manifests metadata table

fallintoplace:
## Summary

Adds the current-snapshot `manifests` metadata table to `Table.Inspect`.

- Exposes manifest content, paths, lengths, counts, and partition specification IDs.
- Converts partition summaries to the human-readable values used by Iceberg clients.
- Represents missing partition summaries as a non-null empty list and preserves unknown legacy V1 file counts as null.
- Keeps `added_snapshot_id` required and rejects malformed negative values.
- Rejects missing partition specs when partition summaries require decoding, excess partition summaries, and unknown manifest content.

## Tests

- `go test -race ./table`
- `go test ./...`
- Covers data and delete manifests, V1 unknown and invalid negative counts, invalid snapshot IDs, invalid partition metadata, field IDs, and partition-summary conversion.

## Scope

This adds current-snapshot metadata inspection only. It does not change manifest reading or writing.

### DIFF
--- a/table/inspect.go
+++ b/table/inspect.go
@@ -290,39 +290,51 @@ func (i InspectTable) Manifests(ctx context.Context) (array.RecordReader, error)
 			return nil, err
 		}
 
-		content.Append(int32(manifest.ManifestContent()))
+		manifestContent := manifest.ManifestContent()
+		switch manifestContent {
+		case iceberg.ManifestContentData, iceberg.ManifestContentDeletes:
+		default:
+			return nil, fmt.Errorf("manifest %s has unknown content %d", manifest.FilePath(), manifestContent)
+		}
+
+		content.Append(int32(manifestContent))
 		path.Append(manifest.FilePath())
 		length.Append(manifest.Length())
 		partitionSpecID.Append(manifest.PartitionSpecID())
 		addedSnapshotID.Append(manifest.SnapshotID())
 
-		if manifest.ManifestContent() == iceberg.ManifestContentData {
-			addedDataFiles.Append(manifest.AddedDataFiles())
-			existingDataFiles.Append(manifest.ExistingDataFiles())
-			deletedDataFiles.Append(manifest.DeletedDataFiles())
+		switch manifestContent {
+		case iceberg.ManifestContentData:
+			appendManifestCount(addedDataFiles, manifest.AddedDataFiles())
+			appendManifestCount(existingDataFiles, manifest.ExistingDataFiles())
+			appendManifestCount(deletedDataFiles, manifest.DeletedDataFiles())
 			addedDeleteFiles.Append(0)
 			existingDeleteFiles.Append(0)
 			deletedDeleteFiles.Append(0)
-		} else {
+		case iceberg.ManifestContentDeletes:
 			addedDataFiles.Append(0)
 			existingDataFiles.Append(0)
 			deletedDataFiles.Append(0)
-			addedDeleteFiles.Append(manifest.AddedDataFiles())
-			existingDeleteFiles.Append(manifest.ExistingDataFiles())
-			deletedDeleteFiles.Append(manifest.DeletedDataFiles())
+			appendManifestCount(addedDeleteFiles, manifest.AddedDataFiles())
+			appendManifestCount(existingDeleteFiles, manifest.ExistingDataFiles())
+			appendManifestCount(deletedDeleteFiles, manifest.DeletedDataFiles())
 		}
 
+		spec := i.tbl.metadata.PartitionSpecByID(int(manifest.PartitionSpecID()))
+		if spec == nil {
+			return nil, fmt.Errorf("manifest %s references missing partition spec %d",
+				manifest.FilePath(), manifest.PartitionSpecID())
+		}
+		partType := spec.PartitionType(i.tbl.metadata.CurrentSchema())
 		partitions := manifest.Partitions()
 		if partitions == nil {
 			partitionSummaries.AppendNull()
 
 			continue
 		}
-
-		spec := i.tbl.metadata.PartitionSpecByID(int(manifest.PartitionSpecID()))
-		var partType *iceberg.StructType
-		if spec != nil {
-			partType = spec.PartitionType(i.tbl.metadata.CurrentSchema())
+		if len(partitions) > spec.NumFields() {
+			return nil, fmt.Errorf("manifest %s has %d partition summaries for partition spec %d with %d fields",
+				manifest.FilePath(), len(partitions), manifest.PartitionSpecID(), spec.NumFields())
 		}
 
 		partitionSummaries.Append(true)
@@ -335,19 +347,6 @@ func (i InspectTable) Manifests(ctx context.Context) (array.RecordReader, error)
 				summaryContainsNaN.Append(*summary.ContainsNaN)
 			}
 
-			if spec == nil || idx >= spec.NumFields() {
-				summaryLower.AppendNull()
-				summaryUpper.AppendNull()
-
-				continue
-			}
-
-			if idx >= len(partType.FieldList) {
-				summaryLower.AppendNull()
-				summaryUpper.AppendNull()
-
-				continue
-			}
 			fieldType := partType.FieldList[idx].Type
 			transform := spec.Field(idx).Transform
 			if err := appendManifestBound(summaryLower, fieldType, transform, summary.LowerBound); err != nil {
@@ -365,6 +364,16 @@ func (i InspectTable) Manifests(ctx context.Context) (array.RecordReader, error)
 	}
 
 	return rr, nil
+}
+
+func appendManifestCount(builder *array.Int32Builder, count int32) {
+	if count < 0 {
+		builder.AppendNull()
+
+		return
+	}
+
+	builder.Append(count)
 }
 
 func (i InspectTable) currentSnapshotManifests(ctx context.Context) ([]iceberg.ManifestFile, error) {
--- a/table/inspect.go
+++ b/table/inspect.go
@@ -345,8 +345,12 @@ func (i InspectTable) Manifests(ctx context.Context) (array.RecordReader, error)
 			}
 			fieldType := partType.FieldList[idx].Type
 			transform := spec.Field(idx).Transform
-			appendManifestBound(summaryLower, fieldType, transform, summary.LowerBound)
-			appendManifestBound(summaryUpper, fieldType, transform, summary.UpperBound)
+			if err := appendManifestBound(summaryLower, fieldType, transform, summary.LowerBound); err != nil {
+				return nil, fmt.Errorf("manifest %s partition field %d lower bound: %w", manifest.FilePath(), idx, err)
+			}
+			if err := appendManifestBound(summaryUpper, fieldType, transform, summary.UpperBound); err != nil {
+				return nil, fmt.Errorf("manifest %s partition field %d upper bound: %w", manifest.FilePath(), idx, err)
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


table/inspect.go
```diff
@@ -247,6 +248,224 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+	for _, manifest := range manifests {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+
+		manifestContent := manifest.ManifestContent()
+		switch manifestContent {
+		case iceberg.ManifestContentData, iceberg.ManifestContentDeletes:
+		default:
+			return nil, fmt.Errorf("manifest %s has unknown content %d", manifest.FilePath(), manifestContent)
```
laskoviymishka, line 297: For a read-only inspect API I'm a little wary of hard-erroring the whole call when we hit a single manifest with a content value we don't recognize — one forward-version manifest makes the entire table uninspectable through this method. Java/PyIceberg lean toward graceful degradation here.

Not blocking, and I can see the case for failing loud. But I'd at least consider skip-with-warning or a wrapped sentinel the caller can pick out. Same thought applies to the excess-summaries reject a few lines down. wdyt?

table/inspect.go
```diff
@@ -247,6 +248,224 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+		}
+
+		content.Append(int32(manifestContent))
+		path.Append(manifest.FilePath())
+		length.Append(manifest.Length())
+		partitionSpecID.Append(manifest.PartitionSpecID())
+		addedSnapshotID.Append(manifest.SnapshotID())
```
laskoviymishka: `added_snapshot_id` is declared `Required:true`, but `SnapshotID()` returns `-1` for a V1 manifest whose `added_snapshot_id` field was absent, and we append it straight through. Java returns null there, so a query joining `manifests.added_snapshot_id` against the snapshots table would silently miss instead of seeing null, and `-1` reads like a real id.

I don't think any of the current constructors can actually produce the `-1` (`NewManifestFile` always injects a concrete id), so this may not be reachable today — but if the Avro read path can surface it, I'd declare the field `Required:false` and `AppendNull()` when `SnapshotID()==-1`. Worth confirming which way that goes.

table/inspect.go
```diff
@@ -247,6 +248,224 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+				return nil, err
+			}
+			addedDeleteFiles.Append(0)
+			existingDeleteFiles.Append(0)
+			deletedDeleteFiles.Append(0)
+		case iceberg.ManifestContentDeletes:
+			addedDataFiles.Append(0)
+			existingDataFiles.Append(0)
+			deletedDataFiles.Append(0)
+			if err := appendCount(addedDeleteFiles, "added_delete_files", manifest.AddedDataFiles()); err != nil {
```
laskoviymishka, line 337: This routing is going to make the next reader do a double-take: for delete manifests we're pulling the delete-file counts out of `AddedDataFiles()`/`ExistingDataFiles()`/`DeletedDataFiles()`. It's correct today because those accessors return the generic `added_files_count` field, which holds delete-file counts for a delete manifest — but the names say the opposite.

A one-line comment noting that would save the confusion, and flag it if iceberg-go ever grows distinct `AddedDeleteFiles()` accessors.

table/inspect.go
```diff
@@ -247,6 +248,224 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+			if err := appendCount(existingDeleteFiles, "existing_delete_files", manifest.ExistingDataFiles()); err != nil {
+				return nil, err
+			}
+			if err := appendCount(deletedDeleteFiles, "deleted_delete_files", manifest.DeletedDataFiles()); err != nil {
+				return nil, err
+			}
+		}
+
+		spec := i.tbl.metadata.PartitionSpecByID(int(manifest.PartitionSpecID()))
+		if spec == nil {
```
laskoviymishka, line 356: I think there's a subtle ordering bug here. We look up the spec and error out if it's nil before we check `partitions == nil`, so a manifest that carries no partition summaries but was written under a spec that's since been removed (spec evolution + metadata cleanup) fails the whole `Manifests()` call.

Java and PyIceberg return early on the no-summaries case and never touch the spec, so they succeed where we'd error. I'd move the `partitions == nil` check above the spec lookup — that also gives a clean home for the empty-list change from the other thread. wdyt?

table/inspect.go
```diff
@@ -247,6 +248,224 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+			return nil, fmt.Errorf("manifest %s references missing partition spec %d",
+				manifest.FilePath(), manifest.PartitionSpecID())
+		}
+		partType := spec.PartitionType(i.tbl.metadata.CurrentSchema())
+		partitions := manifest.Partitions()
+		if partitions == nil {
+			partitionSummaries.AppendNull()
```
laskoviymishka: Same contract problem as the count columns, one column over. `partition_summaries` is field 8, `Required:true` → `Nullable:false`, but when `partitions == nil` we `AppendNull()` into it. PyIceberg and Rust both emit an empty list here, never null.

I'd emit `partitionSummaries.Append(true)` with no children (a zero-length list), then `continue`. One gotcha: `Append(false)` would still be a null slot, so it has to be `Append(true)` to get a non-null empty list. That matches the reference clients and keeps the column honestly non-nullable.

table/inspect.go
```diff
@@ -247,6 +248,224 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+			if summary.ContainsNaN == nil {
+				summaryContainsNaN.AppendNull()
+			} else {
+				summaryContainsNaN.Append(*summary.ContainsNaN)
+			}
+
+			fieldType := partType.FieldList[idx].Type
+			transform := spec.Field(idx).Transform
+			if err := appendManifestBound(summaryLower, fieldType, transform, summary.LowerBound); err != nil {
+				return nil, fmt.Errorf("manifest %s partition field %d lower bound: %w", manifest.FilePath(), idx, err)
```
laskoviymishka: nit: this reports the partition field by positional index ("partition field 3"), which makes the user count schema fields to find it. We've got `partType.FieldList[idx].Name` right here — worth using the name instead.

table/inspect.go
```diff
@@ -247,6 +248,224 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+		return nil, fmt.Errorf("inspect manifests: %w", err)
+	}
+
+	return rr, nil
+}
+
+func appendManifestCount(builder *array.Int32Builder, version int, name string, count int32) error {
+	if count < 0 {
+		if version == 1 {
+			builder.AppendNull()
```
laskoviymishka, line 402: This is the one I'd block on. The schema declares the six count fields `Required:true`, which `SchemaToArrowSchema` turns into Arrow `Nullable:false` — but here we `AppendNull()` for V1 manifests, so we emit a batch whose schema says non-nullable while the data carries nulls. Arrow-Go won't panic on that, but any consumer that trusts the schema (a Parquet writer, a scan engine) reads it as corrupt.

Every reference client is internally consistent and none emit null here: PyIceberg and Java declare the fields and write 0 for V1 unknowns; Rust declares them nullable. I'd match PyIceberg/Java and `Append(int32(0))` instead of `AppendNull()` — that keeps `Required:true` honest. If we'd rather preserve the null, the fields (5, 6, 7, 15, 16, 17) have to flip to `Required:false`. Either is fine, but the schema and the data have to agree.

table/inspect.go
```diff
@@ -247,6 +248,224 @@ func (i InspectTable) Snapshots(ctx context.Context) (array.RecordReader, error)
...
+	if snapshot == nil {
+		return nil, nil
+	}
+	if i.tbl.fsF == nil {
+		return nil, errors.New("table file IO is not configured")
+	}
+
+	fs, err := i.tbl.fsF(ctx)
+	if err != nil {
+		return nil, err
```
laskoviymishka: Small thing — this `err` comes back unwrapped while everything else in the chain gets an `fmt.Errorf`. The outer wrap is just `"inspect manifests: %w"`, so a failure in the IO factory loses the context that it happened here. I'd do `return nil, fmt.Errorf("get file IO: %w", err)`.

table/inspect_internal_test.go
```diff
@@ -404,6 +457,308 @@ func TestInspectSnapshotsEmpty(t *testing.T) {
...
+	require.EqualValues(t, 0, record.Column(8).(*array.Int32).Value(0))
+	require.Equal(t, manifestPath, record.Column(1).(*array.String).Value(0))
+
+	summaries := record.Column(11).(*array.List)
+	require.False(t, summaries.IsNull(0))
+	start, end := summaries.ValueOffsets(0)
+	require.EqualValues(t, 1, end-start)
+	summary := summaries.ListValues().(*array.Struct)
+	require.False(t, summary.Field(0).(*array.Boolean).Value(0))
+	require.Equal(t, "7", summary.Field(2).(*array.String).Value(0))
```
laskoviymishka, line 811: The non-null `contains_nan` path never gets exercised. Every test leaves `ContainsNaN` nil, so we only ever hit the `AppendNull` branch — the `Append(*summary.ContainsNaN)` side would pass just as well if it were deleted. This block also skips asserting `Field(1)` entirely.

I'd add a case with `ContainsNaN` set to a bool and assert `Field(1)` is non-null with the right value.

table/inspect_internal_test.go
```diff
@@ -404,6 +457,308 @@ func TestInspectSnapshotsEmpty(t *testing.T) {
...
+
+func TestInspectManifestsRejectsNegativeCountsForV2AndV3(t *testing.T) {
+	tests := []struct {
+		name                     string
+		version                  int
+		added, existing, deleted int32
+		invalidCountName         string
+	}{
+		{name: "v2 added data files", version: 2, added: -123, existing: 1, deleted: 1, invalidCountName: "added_data_files"},
+		{name: "v2 existing data files", version: 2, added: 1, existing: -123, deleted: 1, invalidCountName: "existing_data_files"},
```
laskoviymishka, line 1048: This table only covers data manifests — every row is the default `ManifestContentData`. The delete-file counts go through the same `appendManifestCount`, so a negative delete-count on a V2/V3 delete manifest hits the identical error path but isn't verified. I'd add a few `ManifestContentDeletes` rows with negative counts to cover the symmetry.

laskoviymishka (changes requested):
Nice work, this slots cleanly into the existing History/Snapshots methods, the field IDs match the spec, and reaching for the reject-path tests (missing spec, excess summaries, unknown content, negative counts) is the right instinct.

I'd hold it before merging, though, mostly around one issue that shows up in two places: the schema and the data disagree on nullability. `ManifestsSchema()` declares the six count columns and `partition_summaries` as `Required:true`, which `SchemaToArrowSchema` turns into Arrow `Nullable:false` — but the code `AppendNull()`s into all of them (count columns for V1 unknowns, `partition_summaries` when a manifest has no summaries). Arrow-Go doesn't enforce this at append time, so it passes tests, but we're emitting batches whose schema claims non-nullable while the validity bits say otherwise, and any consumer that trusts the schema (a Parquet writer, a scan engine) reads that as corrupt. Worth noting no reference client does this: PyIceberg and Java write 0 for the counts and an empty list for summaries; Rust declares the fields nullable. So either match them (0 / empty list, keep `Required:true`) or flip the fields to `Required:false` — but the two have to agree.

The other design one I'd want resolved: the missing-spec guard fires before the no-summaries check, so a manifest with no partition summaries written under a since-removed spec errors the whole call, where Java/PyIceberg succeed. Reordering the two guards fixes it and gives a clean home for the empty-list change above.

A few smaller things I've left inline:
- `added_snapshot_id` emits the `-1` sentinel instead of null (may not be reachable today, worth confirming)
- the delete-count routing reads backwards without a comment
- the IO-factory error comes back unwrapped
- the `contains_nan` non-null branch is never exercised (would pass if deleted)
- the negative-count reject only covers data manifests, not delete manifests
- there's no no-current-snapshot (zero-row) case to match the History/Snapshots empty tests

Once the nullability contract and the guard ordering are sorted, happy to take another pass and approve.

zeroshade (approved):
Verified at head that the blocking items from the morning review are resolved: the schema/data nullability contract now holds (counts and `partition_summaries` follow the reference implementations instead of appending null into non-nullable columns), and the spec-lookup ordering is fixed so a manifest with no summaries under a removed spec emits an empty list rather than failing the whole `Manifests()` call. The `contains_nan` and delete-manifest count-validation coverage rounds it out. The remaining open notes (graceful degradation for forward-version content values, the delete-counts-via-data-accessors readability point) read as follow-up material to me.

Deferring final confirmation of `laskoviymishka`'s threads to him since his change request predates these commits.

---

laskoviymishka (approved):
LGTM!

PR #1698, opened 2026-08-07
feat(table): add data files metadata table

fallintoplace:
## Summary

Adds the current live `data_files` metadata table to `Table.Inspect`.

- Exposes content-file metrics, partition values, bounds, delete metadata, and row-lineage fields.
- Uses the Iceberg content-file field IDs and a table-wide partition type for evolved partition specs.
- Provides shared content-file schema and conversion helpers for the related metadata tables.

## Tests

- `go test ./table`
- Covers the content-file schema, decimal partition values, and active partition specs.

## Scope

This reports live data-file metadata for the current snapshot. It does not change data-file reads, writes, or scan behavior.

## Related

The shared content-file foundation is used by #1699, #1700, and #1701.

### DIFF
--- a/table/inspect_files.go
+++ b/table/inspect_files.go
@@ -334,6 +334,15 @@ func inspectValueScalar(value any, typ iceberg.Type, arrowType arrow.DataType) (
 		if value, ok := value.(uuid.UUID); ok {
 			return scalar.MakeScalarParam(value[:], arrowType)
 		}
+	case iceberg.DecimalType:
+		switch value := value.(type) {
+		case iceberg.DecimalLiteral:
+			return scalar.NewDecimal128Scalar(value.Val, arrowType), nil
+		case iceberg.Decimal:
+			return scalar.NewDecimal128Scalar(value.Val, arrowType), nil
+		default:
+			return nil, fmt.Errorf("unsupported decimal partition value %T", value)
+		}
 	}
 
 	return scalar.MakeScalarParam(value, arrowType)
--- a/table/inspect_files.go
+++ b/table/inspect_files.go
@@ -21,10 +21,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"iter"
 	"sort"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/arrow-go/v18/arrow/scalar"
 	"github.com/apache/iceberg-go"
 	"github.com/google/uuid"
@@ -40,23 +42,13 @@ func (i InspectTable) DataFiles(ctx context.Context) (array.RecordReader, error)
 		return nil, fmt.Errorf("inspect data files: build arrow schema: %w", err)
 	}
 
-	files, err := i.currentContentFiles(ctx, iceberg.ManifestContentData)
-	if err != nil {
-		return nil, fmt.Errorf("inspect data files: %w", err)
-	}
-
-	bldr := array.NewRecordBuilder(i.alloc, arrowSchema)
-	defer bldr.Release()
-	for _, file := range files {
-		if err := ctx.Err(); err != nil {
-			return nil, err
-		}
-		if err := appendContentFileRecord(bldr, partitionType, file); err != nil {
-			return nil, fmt.Errorf("inspect data files: append %s: %w", file.FilePath(), err)
-		}
-	}
-
-	rr, err := singleBatchReader(arrowSchema, bldr)
+	rr, err := i.manifestEntryReader(ctx, arrowSchema, true,
+		func(manifest iceberg.ManifestFile) bool {
+			return manifest.ManifestContent() == iceberg.ManifestContentData
+		},
+		func(bldr *array.RecordBuilder, entry iceberg.ManifestEntry) error {
+			return appendContentFileRecord(bldr, partitionType, entry.DataFile())
+		})
 	if err != nil {
 		return nil, fmt.Errorf("inspect data files: %w", err)
 	}
@@ -64,10 +56,21 @@ func (i InspectTable) DataFiles(ctx context.Context) (array.RecordReader, error)
 	return rr, nil
 }
 
-func (i InspectTable) currentContentFiles(ctx context.Context, content iceberg.ManifestContent) ([]iceberg.DataFile, error) {
+const inspectRecordBatchSize = 4096
+
+// manifestEntryReader streams manifest entries into bounded Arrow record
+// batches. It keeps only the current batch and manifest decoder state in
+// memory instead of materializing every entry before returning a reader.
+func (i InspectTable) manifestEntryReader(
+	ctx context.Context,
+	arrowSchema *arrow.Schema,
+	discardDeleted bool,
+	includeManifest func(iceberg.ManifestFile) bool,
+	appendEntry func(*array.RecordBuilder, iceberg.ManifestEntry) error,
+) (array.RecordReader, error) {
 	snapshot := i.tbl.metadata.CurrentSnapshot()
 	if snapshot == nil {
-		return nil, nil
+		return array.ReaderFromIter(arrowSchema, emptyInspectRecordBatch(i.alloc, arrowSchema)), nil
 	}
 	if i.tbl.fsF == nil {
 		return nil, errors.New("table file IO is not configured")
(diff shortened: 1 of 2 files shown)


table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+}
+
+func appendContentFile(builder *array.StructBuilder, partitionType *iceberg.StructType, file iceberg.DataFile) error {
+	builder.Append(true)
+
+	return appendContentFileFields(builder.FieldBuilder, partitionType, file)
+}
+
+func appendContentFileFields(fieldBuilder func(int) array.Builder, partitionType *iceberg.StructType, file iceberg.DataFile) error {
+	idx := 0
```
laskoviymishka: The schema in `inspectContentFileFields` and this builder loop have to stay in perfect lock-step by hand, every `idx++` here mirroring one field there, including the conditional bump for the optional partition column. There's no guard, so inserting or reordering a single field silently shifts everything downstream: wrong order writes mismatched values, wrong type panics at the assertion, and nothing catches it until runtime. `inspect.go` sidesteps this by naming each builder explicitly.

Since you've built this as shared scaffolding for delete_files/entries too, the drift surface only grows. I'd either name the builders like `inspect.go` or look them up by field ID, and failing that at least assert `idx == bldr.NumField()` at the end so a mismatch fails loudly. wdyt?

table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+		case iceberg.DecimalLiteral:
+			return scalar.NewDecimal128Scalar(value.Val, arrowType), nil
+		case iceberg.Decimal:
+			return scalar.NewDecimal128Scalar(value.Val, arrowType), nil
+		default:
+			return nil, fmt.Errorf("unsupported decimal partition value %T", value)
+		}
+	}
+
+	return scalar.MakeScalarParam(value, arrowType)
```
laskoviymishka, line 606: The temporal and UUID cases only handle the happy-path assertion. If the type-switch matches but the value isn't the expected concrete type, we fall through to `scalar.MakeScalarParam(value, arrowType)` here instead of erroring. For a `TimestampType` fed an `int64` that quietly builds an `Int64Scalar` rather than a timestamp (silently wrong data), and for `UUIDType` fed a `uuid.UUID` it hits `MakeScalarParam` with a `[16]byte` and panics.

`DecimalType` already does the right thing with an explicit error return. I'd give the temporal and UUID cases the same treatment so a bad value is a clear error, not corrupt output or a panic.

table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+}
+
+func appendInspectBytes(builder array.Builder, value []byte) {
+	if value == nil {
+		builder.AppendNull()
+
+		return
+	}
+	builder.(*array.BinaryBuilder).Append(value)
```
laskoviymishka: This `array.Builder` gets asserted to `*array.BinaryBuilder` here with no comma-ok, so if field order ever drifts from the schema this panics with nothing to recover from. Every other append helper takes a concrete builder type, which moves the assertion to the `fieldBuilder(idx)` call site. I'd have this one take `*array.BinaryBuilder` directly to match, same failure but at least consistent with the rest.

table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+	if partitionType != nil && len(partitionType.FieldList) > 0 {
+		fields = append(fields, iceberg.NestedField{ID: 102, Name: "partition", Type: partitionType, Required: true})
+	}
+	fields = append(fields,
+		iceberg.NestedField{ID: 103, Name: "record_count", Type: iceberg.PrimitiveTypes.Int64, Required: true},
+		iceberg.NestedField{ID: 104, Name: "file_size_in_bytes", Type: iceberg.PrimitiveTypes.Int64, Required: true},
+		iceberg.NestedField{ID: 108, Name: "column_sizes", Type: inspectInt64MapType(117, 118), Required: false},
+		iceberg.NestedField{ID: 109, Name: "value_counts", Type: inspectInt64MapType(119, 120), Required: false},
+		iceberg.NestedField{ID: 110, Name: "null_value_counts", Type: inspectInt64MapType(121, 122), Required: false},
```
laskoviymishka: `distinct_value_counts` (field 111, `map<int, long>` with key 112 / value 113) is missing from the schema and the append loop. `DataFile` already exposes `DistinctValueCounts()` and it's populated, so the data's right there, a query for it against an iceberg-go table just won't find the column, and that gap propagates when this helper backs delete_files/entries.

Spec order puts it between `null_value_counts` (110) and `nan_value_counts` (137), so I'd add the field here and the matching `appendInspectInt64Map(...file.DistinctValueCounts())` at the same position.

table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+		iceberg.NestedField{ID: 137, Name: "nan_value_counts", Type: inspectInt64MapType(138, 139), Required: false},
+		iceberg.NestedField{ID: 125, Name: "lower_bounds", Type: inspectBinaryMapType(126, 127), Required: false},
```
laskoviymishka: Both Java (`BaseFilesTable` joining `MetricsUtil.readableMetricsSchema`) and PyIceberg (`_get_files_schema`) append a `readable_metrics` struct here, per-column bounds decoded from binary into native types. We only emit the raw `lower_bounds`/`upper_bounds` as `map<int, binary>`, so `SELECT readable_metrics FROM ..._files` against an iceberg-go table comes back column-not-found.

Building it means threading the table schema in and decoding each primitive column's bounds, which is real work. Is that intended for a follow-up, or in scope here? Fine either way, I'd just want it called out explicitly rather than silently diverging from the other engines.

table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+}
+
+// DataFilesSchema returns the common content-file schema used by the data_files
+// and delete_files metadata tables. The partition field is omitted for an
+// unpartitioned table, as required by the Iceberg metadata-table spec.
+func DataFilesSchema(partitionType *iceberg.StructType) *iceberg.Schema {
+	return iceberg.NewSchema(0, inspectContentFileFields(partitionType)...)
+}
+
+func inspectContentFileType(partitionType *iceberg.StructType) *iceberg.StructType {
```
laskoviymishka: `inspectContentFileType` (and `appendContentFile` just below) have no callers or tests, they're scaffolding for the future entries table. The deadcode linter will flag `inspectContentFileType`, and `appendContentFile`'s `Append(true)` + `FieldBuilder` convention differs from `appendContentFileRecord`'s `bldr.Field`, so the first real caller could easily wire it up wrong. I'd drop both and reintroduce them with the entries table where they'll have a test, rather than let them bitrot. wdyt?

table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+		fields = append(fields, field)
+	}
+	sort.Slice(fields, func(left, right int) bool { return fields[left].ID < fields[right].ID })
+
+	return &iceberg.StructType{FieldList: fields}, nil
+}
+
+func isInspectVoidTransform(transform iceberg.Transform) bool {
+	switch transform.(type) {
+	case iceberg.VoidTransform, *iceberg.VoidTransform:
```
laskoviymishka, line 253: `VoidTransform` is always returned by value (`transforms.go`, and every test constructs it that way), so the `*iceberg.VoidTransform` arm never matches. Not harmful today, but if a future parser ever returns the pointer form it'd slip past void-detection and trip the "incompatible transforms" error instead. I'd drop the pointer case unless there's a path that actually produces it.

table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+	builder.Append(true)
+	keys := builder.KeyBuilder().(*array.Int32Builder)
+	items := builder.ItemBuilder().(*array.Int64Builder)
+	ids := make([]int, 0, len(values))
+	for id := range values {
+		ids = append(ids, id)
+	}
+	sort.Ints(ids)
+	for _, id := range ids {
+		keys.Append(int32(id))
```
laskoviymishka: These `int32(id)` conversions (and the ones in `appendInspectInt32List` and `appendInspectOptionalInt32` a bit further down) truncate silently, `int` is 64-bit here, so a column or equality-field ID above `MaxInt32` wraps into a corrupted key with no error. Probably can't happen with real field IDs, but a one-line range guard or a comment noting the spec bound would make that assumption explicit.

table/inspect_internal_test.go
```diff
@@ -404,6 +409,197 @@ func TestInspectSnapshotsEmpty(t *testing.T) {
...
+	require.Equal(t, 102, fields[4].ID)
+	require.Equal(t, 145, fields[len(fields)-1].ID)
+
+	unpartitioned := DataFilesSchema(&iceberg.StructType{})
+	require.NotContains(t, testFieldNames(unpartitioned), "partition")
+}
+
+func TestInspectDataFilesStreamsBatchesAndSkipsDeleted(t *testing.T) {
+	const snapshotID = int64(1)
+	spec := *iceberg.UnpartitionedSpec
```
laskoviymishka, line 1269: This test (and the whole suite) only ever uses `UnpartitionedSpec`, so the partition column branch in `appendContentFileFields` / `appendInspectPartition` / `inspectValueScalar` never runs end-to-end. `TestInspectValueScalarDecimal` and `TestInspectPartitionTypeUsesAllActiveSpecs` exercise those pieces in isolation, but nothing writes a real partitioned data file and reads it back through `DataFiles`, which is exactly the path most likely to panic (see the temporal/UUID fallthrough). I'd add an integration test with at least an identity-partitioned int32 column that verifies the partition values come back correct.

table/inspect_internal_test.go
```diff
@@ -404,6 +409,197 @@ func TestInspectSnapshotsEmpty(t *testing.T) {
...
+
+	tbl := New(Identifier{"db", "tbl"}, built, "metadata.json",
+		func(context.Context) (iceio.IO, error) { return memIO, nil }, nil)
+	rr, err := tbl.Inspect().DataFiles(context.Background())
+	require.NoError(t, err)
+	defer rr.Release()
+
+	var batchRows []int
+	var paths []string
+	for rr.Next() {
```
laskoviymishka, line 1318: This streaming path has the trickiest ownership in the file, the `emit` closure releasing batches, `NewRecordBatch`, the yield-false early exit, and none of it runs under a checked allocator. `TestInspectAllocatorOption` does `checked.AssertSize(t, 0)` for the Snapshots path; I'd add the equivalent here, with one variant that drains fully and one that abandons mid-stream (`rr.Release()` after a single batch) so the early-exit release path is actually covered.

table/inspect_files.go
```diff
@@ -0,0 +1,533 @@
...
+			emitted = true
+			if yield(batch, nil) {
+				return true
+			}
+
+			batch.Release()
+
+			return false
+		}
+		emitEmpty := func() {
```
laskoviymishka, line 111: Worth noting `DataFiles` always yields at least one batch, a snapshot with only delete manifests skips every entry, `emitted` stays false, and `emitEmpty` fires a zero-row batch, same as the empty-snapshot case but through a different code path. That seems intentional and matches the existing Snapshots/History behavior, so mostly I'd just document the "always >=1 batch" invariant and add a test case for the all-deleted-manifest path so the two routes stay pinned.

laskoviymishka (changes requested):
The streaming design here is the right call. `manifestEntryReader` keeps just the current batch and decoder state in memory instead of materializing every entry before handing back a reader, the field ordering matches Java's `DataFile.getType()` exactly, and the partition-type union across all specs (with the void-transform reconciliation) is a genuinely thoughtful way to handle files written before partition evolution.

I'd hold this before merging though. My main concern is structural: `appendContentFileFields` keeps the schema and the builder loop in lock-step through a hand-maintained positional `idx` counter, with no guard that the two stay aligned. Combined with the unchecked builder type-assertions, any future field insertion or reorder silently shifts everything downstream and panics at runtime rather than failing at compile time. Naming the builders (like `inspect.go` does) or looking them up by field ID would take that whole class of bug off the table.

The other thing I'd want settled is the partition-value path. `inspectValueScalar` falls through to a generic `MakeScalarParam` for temporal and UUID types when the value assertion fails, which means wrong-but-silent data for timestamps and a panic for UUIDs, and there's no partitioned-table test anywhere in the suite that would catch it since everything runs on `UnpartitionedSpec`.

- Make `appendContentFileFields` robust to field drift (named builders / ID lookup, or at minimum assert the field count).
- Give the temporal and UUID cases in `inspectValueScalar` explicit errors like `DecimalType` already has.
- Add `distinct_value_counts` (field 111), it's exposed by `DataFile` and just missing from the schema.
- Add an integration test with a partitioned table that reads partition values back through `DataFiles`, plus a checked-allocator test that abandons the stream mid-way.
- Decide whether `readable_metrics` is in scope or a follow-up, and call it out either way.
- Drop the dead scaffolding (`inspectContentFileType`, `appendContentFile`, the `*VoidTransform` arm) or land it with the table that uses it.

zeroshade (commented):
Re-checked this at head: the scalar type-mismatch fallthrough is properly closed (explicit errors on every matched-type arm), and `TestInspectDataFilesReturnsPartitionValues` gives the partition branch real end-to-end coverage — nice.

+1 to `laskoviymishka`'s remaining ask before this merges: the streaming path (the `emit` closure's batch release, `NewRecordBatch` ownership, and the yield-false early exit) still has no checked-allocator coverage. Given that's the code most likely to leak or double-release under a consumer that stops early, the two variants he described — full drain and early exit, both under `checked.AssertSize(t, 0)` — are worth adding. Everything else looks ready to me.

---

zeroshade (approved):
The streaming-path allocator coverage closes the last gap: both the full-drain and abandon-after-first-batch variants run under a checked allocator with the entry count sized to force multiple batches, so the `emit` closure's release discipline and the early-exit path are both leak-checked now. With the scalar type-mismatch errors and the partitioned end-to-end coverage from the previous push, this is ready from my side.

---

laskoviymishka (approved):
Good to merge now

laskoviymishka:
@fallintoplace  let's fix conflict and then ready for merge

PR #1701, opened 2026-08-07
feat(table): add partitions metadata table

fallintoplace:
## Summary

Adds the current-snapshot `partitions` metadata table to `Table.Inspect`.

- Aggregates data-file, position-delete, and equality-delete counts by partition.
- Preserves the latest update timestamp, snapshot ID, and partition specification ID.
- Coerces compatible evolved partition specs into a table-wide partition value.

## Tests

- `go test ./table`
- Covers the partition schema, unified partition fields, and evolved partition specs.

## Scope

This reports partition-level metadata for the current snapshot. It does not change file writing, delete application, or scan behavior.

## Related

Builds on the shared content-file and partition representation in #1698.


zeroshade (approved):
just needs merge conflicts resolved

zeroshade (approved):

PR #1702, opened 2026-08-07
feat(table): add incremental append scan

fallintoplace:
## Summary

Adds an incremental append scan through `Table.NewIncrementalAppendScan`.

- Plans data files added by append snapshots between a starting snapshot and an end snapshot.
- Supports inclusive and exclusive starting boundaries plus snapshot ID and as-of selectors.
- Validates snapshot ancestry while preserving the existing file-scan task shape.
- Binds task residuals during planning so historical schema and case-sensitivity semantics survive a separately configured `Scan.ReadTasks` call.
- Keeps explicit historical end snapshots consistent with iceberg-go explicit-snapshot schema behavior.

## Tests

- `go test -race ./table`
- `go test ./...`
- Covers default starts, snapshot ID and as-of selectors, inclusive and exclusive boundaries, expired and multiple roots, ancestry validation, mixed snapshot operations, manifest pruning, historical schema renames, case-insensitive planning, and separate task consumption.

## Scope

This is an append-only incremental scan. Broader changelog scanning and row-level change tracking are out of scope. Branch and tag selectors can be added separately.


table/arrow_scanner.go
```diff
@@ -1091,14 +1146,23 @@ func (as *arrowScan) recordsFromTask(ctx context.Context, task tblutils.Enumerat
...
+		rowFilter  iceberg.BooleanExpression
 		rdr        tblutils.FileReader
 		iceSchema  *iceberg.Schema
 		colIndices []int
 		filterFunc recProcessFn
 		dropFile   bool
 	)
 
-	iceSchema, colIndices, rdr, err = as.prepareToRead(ctx, task.Value.File)
+	rowFilter, err = bindTaskFilter(as.metadata.CurrentSchema(), task.Value.Residual, as.caseSensitive)
```
laskoviymishka: I think there's a subtle trap here. `recordsFromTask` binds the residual against `as.metadata.CurrentSchema()`, but `ReadTasks` already pre-binds every residual against `effectiveSchema` (which can be an older snapshot's schema) before it ever reaches here. Today that's fine — `bindTaskFilter` short-circuits already-bound expressions, so the current-schema bind is a no-op on the production path.

But `arrowScan` is internal and the tests instantiate it directly. The moment a caller hands `recordsFromTask` a task with an unbound residual referencing a column that lives in a historical schema but not the current one, it binds against the wrong schema silently — wrong field ID, or a spurious error. I'd thread `effectiveSchema` into `arrowScan` and bind against that here, so the invariant is enforced in one place rather than assumed.

While we're here, the assign-then-override-then-check order reads a little backwards — the `err` from `bindTaskFilter` is checked after the `Residual == nil` branch that throws its result away. I'd flip it to `if task.Value.Residual != nil { rowFilter, err = ...; if err != nil { return err } } else { rowFilter = as.boundRowFilter }`. wdyt?

table/incremental_append_scan.go
```diff
@@ -0,0 +1,266 @@
...
+
+	if s.fromSnapshotID == nil {
+		slices.Reverse(ancestors)
+
+		return appendOnlySnapshots(ancestors), nil
+	}
+
+	fromID := *s.fromSnapshotID
+	if !s.fromInclusive {
+		between, found := AncestorsBetween(toSnapshotID, fromID, s.scan.metadata.SnapshotByID)
```
laskoviymishka, line 267: When `fromSnapshotID == toSnapshotID` on the exclusive path, `AncestorsBetween` short-circuits and returns `(nil, true)`, so `found` is true, we get an empty snapshot list, and `PlanFiles` returns zero tasks with no error. Java's `BaseIncrementalScan` rejects this — `from == to` exclusive can't have `from` as a parent-ancestor of `to`, so it throws.

Silently returning empty for what's really an invalid range feels risky: a caller polling `from=X exclusive, to=X` sees "no new data" rather than a clear error. I'd guard `from == to` explicitly and return `ErrInvalidArgument`.

There's also no test pinning either from==to case. I'd add one for exclusive (expect the error) and one for inclusive `FromSnapshotInclusive(X).ToSnapshot(X)` (expect just X's files) — the inclusive path leans on `IsAncestorOf(X, X)` returning true, and that's currently an untested assumption. wdyt?

table/incremental_append_scan.go
```diff
@@ -0,0 +1,266 @@
...
+	out := *s
+	out.fromSnapshotID = &snapshotID
+	out.fromInclusive = true
+
+	return &out, nil
+}
+
+// FromSnapshotExclusive starts after the given snapshot. The starting
+// snapshot must be an ancestor of the ending snapshot when planning.
+func (s *IncrementalAppendScan) FromSnapshotExclusive(snapshotID int64) *IncrementalAppendScan {
```
laskoviymishka, line 68: The three builder methods return different shapes — `FromSnapshotInclusive` and `ToSnapshot` return `(*IncrementalAppendScan, error)`, but `FromSnapshotExclusive` returns just `*IncrementalAppendScan`. That breaks uniform chaining, and since this is a brand-new public API, I'd settle it before it ships.

The asymmetry also means a bogus exclusive from-ID gets no feedback until `PlanFiles`, and the error there ("starting snapshot X is not an ancestor of ending snapshot Y") is misleading when the snapshot simply doesn't exist. I get that exclusive intentionally tolerates an expired/pruned parent (the expired-exclusive test), so eager existence-checking isn't quite right either.

Either way is fine, but it should be deliberate: move all validation into `PlanFiles` and have every builder return `*IncrementalAppendScan` only (matches how `Scan` defers errors to `PlanFiles`, most idiomatic here), or make all three return `(*..., error)` and document why exclusive still allows a missing/expired from-snapshot. wdyt?

table/incremental_append_scan.go
```diff
@@ -0,0 +1,266 @@
...
+	snapshots, err := s.snapshotsBetween(toSnapshot.SnapshotID)
+	if err != nil {
+		return nil, err
+	}
+	if len(snapshots) == 0 {
+		return nil, nil
+	}
+	appendSnapshots := make(map[int64]struct{}, len(snapshots))
+	for _, snapshot := range snapshots {
+		if snapshot.Summary != nil && snapshot.Summary.Operation == OpAppend {
```
laskoviymishka: This append-only filter is dead code — `snapshotsBetween` already runs everything through `appendOnlySnapshots`, so every snapshot reaching here is guaranteed `OpAppend` and this `if` is always true (and the `len(appendSnapshots) == 0` check just below is unreachable, given the `len(snapshots) == 0` guard above it).

Harmless today, but it makes it look like there are two independent append-only fences when there's really one. The actual invariant is the `appendSnapshots` membership checks at the manifest level and entry level below — if someone later widens `snapshotsBetween` to return all snapshots, `appendSnapshots` silently fills with overwrite/delete IDs and this filter won't save us. I'd pick a single authority: keep `snapshotsBetween` append-only and drop this guard with a comment, or have it return all snapshots and let `PlanFiles` do the filtering.

Related — I don't see an end-to-end test that drives a range containing a non-append snapshot (append->overwrite->append) through `PlanFiles` and asserts only the two appends' files show up. `TestIncrementalAppendScanSnapshotBoundaries` checks `snapshotsBetween` directly, but the manifest/entry-level fence, the thing that actually protects correctness, isn't exercised in the full planning flow. I'd add that fixture.

table/incremental_append_scan.go
```diff
@@ -0,0 +1,266 @@
...
+	for path := range manifestsByPath {
+		paths = append(paths, path)
+	}
+	sort.Strings(paths)
+	manifestList := make([]iceberg.ManifestFile, 0, len(paths))
+	for _, path := range paths {
+		manifestList = append(manifestList, manifestsByPath[path])
+	}
+
+	planningScan := *s.scan
```
laskoviymishka: Not a blocker, and it's consistent with the existing `Scan` builder shallow-copy pattern, but `planningScan := *s.scan` aliases the reference-typed fields — `options` (map), `selectedFields` and `identifier` (slices) still point at the caller's underlying data. Nothing mutates through `planningScan` today so it's safe, but it's the kind of thing that bites later: a future change to `filterManifestsWithSchema` that appends to `selectedFields` would corrupt the caller's scan.

If it's cheap, I'd `maps.Clone` the options and `slices.Clone` the slices on copy (or add a `Scan.clone()` helper) to make the isolation explicit.

table/incremental_append_scan.go
```diff
@@ -0,0 +1,266 @@
...
+			}
+			manifestsByPath[manifest.FilePath()] = manifest
+		}
+	}
+
+	paths := make([]string, 0, len(manifestsByPath))
+	for path := range manifestsByPath {
+		paths = append(paths, path)
+	}
+	sort.Strings(paths)
```
laskoviymishka: The surrounding code uses `slices` consistently, so I'd swap `sort.Strings(paths)` for `slices.Sort(paths)` here and the `sort.Slice(tasks, ...)` below for `slices.SortFunc(tasks, func(a, b FileScanTask) int { return cmp.Compare(a.File.FilePath(), b.File.FilePath()) })`, then drop the `sort` import.

table/incremental_append_scan_test.go
```diff
@@ -0,0 +1,437 @@
...
+	exclusive, err = exclusive.ToSnapshot(102)
+	require.NoError(t, err)
+	snapshots, err = exclusive.snapshotsBetween(102)
+	require.NoError(t, err)
+	require.Empty(t, snapshots, "the only snapshot after 101 is not an append")
+}
+
+func TestIncrementalAppendScanRejectsUnknownStart(t *testing.T) {
+	_, err := snapshotsTestTable().NewIncrementalAppendScan().FromSnapshotInclusive(999)
+	require.Error(t, err)
```
laskoviymishka: Every other error assertion in this file uses `require.ErrorIs(t, err, iceberg.ErrInvalidArgument)` — I'd match that here instead of bare `require.Error`, plus a `require.ErrorContains(t, err, "starting snapshot not found")` so we're actually pinning the not-found path. Small thing while here: `999` could be a named `const nonExistentSnapshotID = int64(999)` to mirror the `expiredSnapshotID` const the neighboring test uses.

table/incremental_append_scan_test.go
```diff
@@ -0,0 +1,437 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
```
laskoviymishka, line 9: This header stops at the LICENSE-2.0 URL and is missing the closing boilerplate ("Unless required by applicable law..." through "...limitations under the License.") that every other file in the tree carries — compare `task_residual_test.go` in this same PR. I'd paste the full 16-line header for consistency.

laskoviymishka (changes requested):
Nice work on this, it tracks Java's incremental append scan closely, and the parts that matter for correctness are thought through: deduping inherited manifests into a set before opening them, following a single ancestry chain, and pinning the historical-vs-live schema choice on whether the end snapshot is explicit. The residual-binding plumbing in `ReadTasks` with the bound/unbound visitor is careful, and the boundary and schema-evolution coverage is genuinely good.

I'd hold this before merging though, mostly around one correctness invariant and the shape of the new public API.

The residual binding in `recordsFromTask` binds against `CurrentSchema()` while `ReadTasks` pre-binds against `effectiveSchema`. It's a no-op on the production path today because already-bound expressions short-circuit, but `arrowScan` is internal and the tests construct it directly — so the moment something feeds an unbound residual referencing a historical-only column, it binds against the wrong schema silently. I'd thread `effectiveSchema` through so the invariant lives in one place instead of being assumed.

The other thing is the public API surface. The three builder methods return inconsistent shapes (`FromSnapshotExclusive` has no error, the other two do), which breaks uniform chaining, and `from == to` on the exclusive path silently returns zero tasks where Java throws. Since this is a fresh public API, I'd rather settle the return-shape convention and the empty-range semantics now than change them after people build against it.

- thread `effectiveSchema` into `recordsFromTask` so residuals never bind against the wrong schema off the `ReadTasks` path
- decide the builder return convention (all `(*scan, error)`, or all bare and defer validation to `PlanFiles`), and make `from == to` exclusive an explicit error
- collapse the dead append-only filter in `PlanFiles` to a single authority, and add an end-to-end test that runs a mixed append/overwrite range through `PlanFiles`
- the smaller stuff inline: `sort` -> `slices`, `require.ErrorIs`, and the truncated license header

Branch/tag selectors being out of scope is totally fine for this PR — the only thing I'd note for a follow-up is that an end snapshot resolved from a different branch isn't guarded, but that can wait.

zeroshade (approved):
All of the morning-review items check out at head: the residual now binds against the scan's effective schema on both the planning and read paths, exclusive `from == to` fails fast with the Java-matching error, the builder API is uniform before it ships, and `appendOnlySnapshots` has real validation semantics — appends planned, replace/overwrite/delete skipped per Java, missing/unknown operations rejected. The boundary, ancestry, schema-evolution, and separate-task-consumption coverage is genuinely thorough.

One note for the record: after #1726 lands, unknown snapshot operations will surface here as errors rather than skips — that's the right conservative call for incremental scans, just flagging that the interaction is intentional.

Deferring final confirmation of `laskoviymishka`'s threads to him since his change request predates these commits.

---

laskoviymishka (approved):
LGTM

laskoviymishka:
@fallintoplace - can you solve conflicts, so i can merge this?

PR #1705, opened 2026-08-07
Add more conformance tests

rambleraptor:
Since we've gotten the conformance test suite merged, I've built out some more tests. These tests are:

- create a table that already exists
- load a missing table
- load a table with a non-existent namespace

All of them pass, so our catalogs match for these use cases. Great to verify!


tanmayrauth (approved):

laskoviymishka (approved):
👍

PR #1706, opened 2026-08-07
test(table): expand geometry write-path coverage

tanmayrauth:
Add end-to-end sub-tests around the geometry write path that exercise dispatch the internal TestWriteDataFileGeoBounds cannot reach (it feeds hand-built StatsCols straight to WriteDataFile):

- non-geo id column keeps ordinary min/max bounds, so an over-eager geo stats suppression on an adjacent column cannot pass unnoticed
- all-null geometry column drops out of the manifest bounds map
- write.metadata.metrics.column.geom=none suppresses geo bounds via the stats-plan dispatch, scoped to the geom column
- checked allocator asserts zero residual to catch allocation escapes in the geo-bounds accumulator wiring

Also reword the geography assertion: leaving bounds unbounded is the current conservative implementation, not a spec mandate, and note the .String() hex round-trip the JSON parser depends on.

### DIFF
--- a/table/geo_write_test.go
+++ b/table/geo_write_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/compute"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/iceberg-go"
 	"github.com/apache/iceberg-go/io"
@@ -33,6 +34,12 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/twpayne/go-geom/encoding/wkb"
 	"github.com/twpayne/go-geom/encoding/wkt"
+)
+
+const (
+	geoTestIDFieldID   = 1
+	geoTestGeomFieldID = 2
+	geoTestGeogFieldID = 3
 )
 
 // wktToWKB converts Well-Known Text into little-endian (NDR) Well-Known Binary,
@@ -46,6 +53,45 @@ func wktToWKB(t *testing.T, s string) geoarrow.WKBBytes {
 	require.NoError(t, err)
 
 	return geoarrow.WKBBytes(b)
+}
+
+// newGeoTestWriter builds an unpartitioned v3 table with an id/geom/geog schema
+// and returns a data file writer over it. props feed the metrics-mode
+// configuration (e.g. write.metadata.metrics.column.geom=none) so tests can
+// exercise the stats-plan dispatch that decides whether geo bounds are recorded.
+func newGeoTestWriter(t *testing.T, dir string, props iceberg.Properties) (*defaultDataFileWriter, *iceberg.Schema, *arrow.Schema) {
+	t.Helper()
+
+	geogType, err := iceberg.GeographyTypeOf("OGC:CRS84", "spherical")
+	require.NoError(t, err)
+
+	schema := iceberg.NewSchema(0,
+		iceberg.NestedField{ID: geoTestIDFieldID, Name: "id", Type: iceberg.PrimitiveTypes.Int32, Required: false},
+		iceberg.NestedField{ID: geoTestGeomFieldID, Name: "geom", Type: iceberg.GeometryType{}, Required: false},
+		iceberg.NestedField{ID: geoTestGeogFieldID, Name: "geog", Type: geogType, Required: false},
+	)
+
+	// Geometry/geography are v3 types.
+	mb, err := NewMetadataBuilder(3)
+	require.NoError(t, err)
+	require.NoError(t, mb.AddSchema(schema))
+	require.NoError(t, mb.SetCurrentSchemaID(0))
+	unpartitioned := *iceberg.UnpartitionedSpec
+	require.NoError(t, mb.AddPartitionSpec(&unpartitioned, true))
+	require.NoError(t, mb.SetDefaultSpecID(0))
+	// The stats plan is computed from the table properties on the builder (see
+	// defaultDataFileWriter.writeFile), so metrics-mode config must live here.
+	if len(props) > 0 {
+		require.NoError(t, mb.SetProperties(props))
+	}
+
+	arrowSchema, err := SchemaToArrowSchema(schema, nil, true, false)
+	require.NoError(t, err)
+
+	writer, err := newDataFileWriter(dir, &io.LocalFS{}, mb, props)
+	require.NoError(t, err)
+
+	return writer, schema, arrowSchema
 }
 
 // TestWriteGeometryColumnPopulatesBounds writes a geometry and a geography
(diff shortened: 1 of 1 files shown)


table/geo_write_test.go
```diff
@@ -48,34 +55,20 @@ func wktToWKB(t *testing.T, s string) geoarrow.WKBBytes {
...
-	t.Parallel()
-
-	const (
-		geomFieldID = 2
-		geogFieldID = 3
-	)
+// newGeoTestWriter builds an unpartitioned v3 table with an id/geom/geog schema
+// and returns a data file writer over it. props feed the metrics-mode
+// configuration (e.g. write.metadata.metrics.column.geom=none) so tests can
+// exercise the stats-plan dispatch that decides whether geo bounds are recorded.
```
laskoviymishka, line 61: `newGeoTestWriter` only builds top-level geo columns, and nested geo columns are still unhandled in the writer (the `TODO(#992)` over in `parquet_files.go`). A one-line note in this docstring that nested geo isn't covered here would stop someone reading only the test file from assuming it is.

table/geo_write_test.go
```diff
@@ -118,17 +142,145 @@ func TestWriteGeometryColumnPopulatesBounds(t *testing.T) {
...
-	assert.NotContains(t, upper, geogFieldID, "geography column must not record bounds")
+	// The non-geo id column still gets ordinary min/max bounds. The geo-type guard
+	// in DataFileStatsFromMeta suppresses generic Parquet stats for geo columns;
+	// pinning id here catches a regression that over-suppresses an adjacent
+	// non-geo column.
+	require.Contains(t, lower, geoTestIDFieldID, "non-geo id column must still record a lower bound")
+	require.Contains(t, upper, geoTestIDFieldID, "non-geo id column must still record an upper bound")
+
+	// Geography does not record bounds in the current implementation: the V3 spec
+	// permits geography bounds (with the xmin > xmax antimeridian-wrapping
```
laskoviymishka, line 166: nice that the comment already calls out the `xmin > xmax` antimeridian convention.

One thing I'd add while we're here: the geometry decode path (`GeoBoundsXY`) rejects `xmin > xmax` as inverted, which is correct for geometry but would silently drop valid antimeridian-crossing geography bounds from Java/PyIceberg if it ever got reused for the geography pruning path. A short note that geography will need its own decode would save someone a subtle bug later. Not blocking, wdyt?

table/geo_write_test.go
```diff
@@ -118,17 +142,145 @@ func TestWriteGeometryColumnPopulatesBounds(t *testing.T) {
...
+		FileCount: 1,
+		Schema:    schema,
+		Batches:   []arrow.RecordBatch{rec},
+	})
+	require.NoError(t, err)
+	require.EqualValues(t, 2, df.Count())
+
+	lower := df.LowerBoundValues()
+	upper := df.UpperBoundValues()
+	assert.NotContains(t, lower, geoTestGeomFieldID, "all-null geometry column must not record a lower bound")
```
laskoviymishka, line 211: this passes both when the accumulator ran and found nothing, and when the accumulator was never wired up at all (say `collectGeoColumns` stopped registering geo columns), which is the more interesting regression.

Fine to leave since `PopulatesBounds` is the positive guard against the bypass case, but I'd add a one-line comment here pointing at that so the intent is clear. wdyt?

table/geo_write_test.go
```diff
@@ -118,17 +142,145 @@ func TestWriteGeometryColumnPopulatesBounds(t *testing.T) {
...
+	assert.NotContains(t, lower, geoTestGeomFieldID, "geometry column bounds must be suppressed under metrics mode none")
+	assert.NotContains(t, upper, geoTestGeomFieldID, "geometry column bounds must be suppressed under metrics mode none")
+
+	// Suppression is scoped to the geom column: id still gets ordinary bounds.
+	require.Contains(t, lower, geoTestIDFieldID, "non-geo id column must keep bounds when only geom is set to none")
+	require.Contains(t, upper, geoTestIDFieldID, "non-geo id column must keep bounds when only geom is set to none")
+}
+
+// TestWriteGeometryColumnCheckedAllocator runs the geometry write path on a
+// checked allocator and asserts zero residual once the writer is done, catching
```
laskoviymishka: the comment says this catches "a silent allocation escape in the geo-bounds accumulator wiring," but I don't think the checked allocator can actually see that. `geoBoundsAccumulator` does its work on plain Go struct fields and hands back `[]byte` from `encodeGeoBound`, none of which goes through an Arrow allocator, so a leak there is a GC concern that `AssertSize` never observes.

What this test genuinely validates is that the Arrow write pipeline (`ToRequestedSchema`, the pqarrow writer) releases its buffers, so I'd reframe the comment to that. If we do want a real guard on the accumulator holding no references, that's a separate small unit test on `geoBoundsAccumulator` after `Bounds()`/`StatsAgg()`. wdyt?

table/geo_write_test.go
```diff
@@ -118,17 +142,145 @@ func TestWriteGeometryColumnPopulatesBounds(t *testing.T) {
...
+	geomLo := wktToWKB(t, "POINT (0 -5)")
+	geomHi := wktToWKB(t, "POINT (30 10)")
+	geog := wktToWKB(t, "POINT (12 4)")
+
+	rec, _, err := array.RecordFromJSON(mem, arrowSchema, strings.NewReader(`[
+		{"id": 1, "geom": "`+geomLo.String()+`", "geog": "`+geog.String()+`"},
+		{"id": 2, "geom": "`+geomHi.String()+`", "geog": null}
+	]`))
+	require.NoError(t, err)
+	defer rec.Release()
```
laskoviymishka: `writeFile` owns everything in `task.Batches` and releases each batch itself (that deferred `b.Release()` loop in `writer.go`), so this defer double-releases `rec`: `writeFile` drops it to zero before returning, then this fires and takes it below zero.

Under `DefaultAllocator` that's silent, which is how the other tests get away with it, but here it corrupts the checked allocator's accounting before the deferred `AssertSize(t, 0)` runs, so the zero-residual check isn't measuring what we think it is.

I'd drop this defer, and the same one in the other three tests plus the original `PopulatesBounds`, since `writeFile` owns the batch in all of them.

table/geo_write_test.go
```diff
@@ -118,17 +142,145 @@ func TestWriteGeometryColumnPopulatesBounds(t *testing.T) {
...
+	df, err := writer.writeFile(ctx, nil, WriteTask{
+		Uuid:      uuid.New(),
+		ID:        0,
+		FileCount: 1,
+		Schema:    schema,
+		Batches:   []arrow.RecordBatch{rec},
+	})
+	require.NoError(t, err)
+	require.EqualValues(t, 2, df.Count())
+	require.Contains(t, df.LowerBoundValues(), geoTestGeomFieldID, "geometry column must record a lower bound")
```
laskoviymishka, line 296: the other three tests check both lower and upper bounds; here we only assert the lower. I'd add the matching `df.UpperBoundValues()` contains check so this one lines up and pins both ends.

laskoviymishka (changes requested):
Nice expansion of the geo write-path coverage, and I appreciate that the comments actually explain the why (geography staying unbounded, what each id-column assertion guards). That's the kind of test doc that ages well.

I'd hold this before merging though. Every one of these tests defers `rec.Release()` right after `RecordFromJSON`, but `writeFile` already owns and releases each batch in `task.Batches`, so it's a double-release. Silent under `DefaultAllocator`, which is how the original test got away with it, but in the new checked-allocator test it corrupts the accounting before the deferred `AssertSize(t, 0)` fires, so the zero-residual check isn't really measuring anything.

Related: that same checked-allocator test says it catches an allocation escape in the geo-bounds accumulator, but the accumulator works on plain Go fields and returns `[]byte`, none of it through an Arrow allocator, so the checked allocator can't see a leak there. I'd reframe the comment to what the test genuinely covers (the Arrow write pipeline releasing its buffers).

A few smaller things I've left inline:
- the checked-allocator test only asserts the lower bound; the others check both ends
- `TestWriteGeometryColumnAllNull` passes whether the accumulator ran or was never invoked, worth a one-liner pointing at `PopulatesBounds` as the positive guard
- an optional breadcrumb that reusing `GeoBoundsXY` for a future geography decode would drop antimeridian bounds

Once the double-release is out and the checked-allocator comment matches what it validates, happy to take another pass and approve.

tanmayrauth:
Thanks for the thorough look, this was really helpful. Pushed fixes for all of it.
  
  The double-release one was a good catch. I hadn't clocked that writeFile already releases everything in task.Batches, so my defer rec.Release() was just doing it a second time. Doesn't show up under DefaultAllocator, but yeah, in the checked-allocator test it was quietly going negative before AssertSize ran, so that assertion wasn't really testing anything. Dropped it everywhere.

  And you're right that the checked allocator can't actually see the accumulator, that comment was overclaiming. It's plain Go fields and []byte, nothing goes through Arrow there. Rewrote the comment to say what it really covers (the write pipeline releasing its buffers) and added the upper-bound check so it matches the others.

  Rest of the notes are in too: pointed the all-null test at PopulatesBounds as the real positive guard, added the nested-geo caveat to newGeoTestWriter, and left the antimeridian breadcrumb, since GeoBoundsXY throwing out xmin > xmax is fine for geometry but would eat valid geography bounds if someone reused it.

laskoviymishka (approved):
Nice! That looks ready to merge now <3

PR #1707, opened 2026-08-08
fix(cli): support SQL catalog and validate DynamoDB type

### LINKED ISSUE #1692: CLI initCatalog does not handle all catalog.Type values (SQL/DynamoDB)
### Apache Iceberg version

None

### Please describe the bug 🐞

Addressed by [this](https://github.com/apache/iceberg-go/pull/1650#pullrequestreview-4878472108)

## Description

CLI `initCatalog` switch currently handles only REST, Glue, Hive, and Hadoop catalog types, while `catalog.Type` also defines `SQL` and `DynamoDB`.

This means values such as `--catalog sql` currently fall through to `unrecognized catalog type`, even though the SQL catalog is implemented and registered through the catalog registry.

## Current behavior

  ```sh
  iceberg --catalog sql list

  fails with:

  unrecognized catalog type

  ## Expected behavior

The CLI should either support all implemented catalog types exposed by catalog.Type, or clearly document/validate which catalog types are supported by the CLI.

dgvj-work:
## Summary
- Wire `--catalog sql` through `catalog.Load` so the CLI uses the registered SQL catalog instead of failing with `unrecognized catalog type`.
- Add `--sql-driver` / `--sql-dialect` flags and matching `sql-driver` / `sql-dialect` config keys.
- Recognize `--catalog dynamodb` with a clear `not implemented` error (type exists; catalog package does not yet).
- Document the new flags/config and cover mixed-case SQL plus DynamoDB validation in CLI tests.

Fixes #1692

## Testing
- `go test ./config/ ./cmd/iceberg/ -count=1`
- `gofmt` clean on touched Go files

### DIFF
--- a/cmd/iceberg/args_test.go
+++ b/cmd/iceberg/args_test.go
@@ -414,6 +414,11 @@ func TestCLIAcceptsMixedCaseCatalogType(t *testing.T) {
 
 	hadoopWarehouse := t.TempDir()
 
+	sqlWarehouse := t.TempDir()
+	sqlDB := filepath.Join(t.TempDir(), "catalog.db")
+	sqlURI := "file:" + sqlDB
+	sqlWarehouseURI := "file://" + sqlWarehouse
+
 	// A raw TCP listener stands in for a Hive metastore: the CLI only needs to
 	// get far enough to open a connection to prove "HIVE"/"Hive" were recognized
 	// as the Hive catalog type. It has no Thrift handshake, so the command still
@@ -444,6 +449,9 @@ func TestCLIAcceptsMixedCaseCatalogType(t *testing.T) {
 		// rejected outright, to keep that failure mode distinguishable from
 		// wantErr above.
 		wantUnrecognized bool
+		// wantContains, when set with wantErr, asserts a substring of stderr/stdout
+		// so "not implemented" failures stay distinguishable from other errors.
+		wantContains string
 	}{
 		{
 			name: "rest lowercase",
@@ -496,6 +504,48 @@ func TestCLIAcceptsMixedCaseCatalogType(t *testing.T) {
 			wantErr: true,
 		},
 		{
+			name: "sql lowercase",
+			args: []string{
+				"list", "--catalog", "sql",
+				"--uri", sqlURI,
+				"--sql-driver", "sqliteshim",
+				"--sql-dialect", "sqlite",
+				"--warehouse", sqlWarehouseURI,
+			},
+		},
+		{
+			name: "sql uppercase",
+			args: []string{
+				"list", "--catalog", "SQL",
+				"--uri", sqlURI,
+				"--sql-driver", "sqliteshim",
+				"--sql-dialect", "sqlite",
+				"--warehouse", sqlWarehouseURI,
+			},
+		},
+		{
+			name: "sql mixed case",
+			args: []string{
+				"list", "--catalog", "Sql",
+				"--uri", sqlURI,
+				"--sql-driver", "sqliteshim",
+				"--sql-dialect", "sqlite",
+				"--warehouse", sqlWarehouseURI,
+			},
+		},
+		{
+			name:         "dynamodb is recognized but not implemented",
+			args:         []string{"list", "--catalog", "dynamodb"},
+			wantErr:      true,
+			wantContains: "dynamodb catalog is not implemented",
+		},
+		{
+			name:         "dynamodb uppercase is recognized but not implemented",
+			args:         []string{"list", "--catalog", "DYNAMODB"},
+			wantErr:      true,
+			wantContains: "dynamodb catalog is not implemented",
+		},
+		{
 			name:             "unknown catalog type is still rejected",
 			args:             []string{"list", "--catalog", "RESTX", "--uri", restSrv.URL},
 			wantUnrecognized: true,
@@ -519,6 +569,9 @@ func TestCLIAcceptsMixedCaseCatalogType(t *testing.T) {
 			if tt.wantErr {
 				require.Error(t, err, "output: %s", out)
 				require.NotContains(t, string(out), "unrecognized catalog type")
+				if tt.wantContains != "" {
+					require.Contains(t, string(out), tt.wantContains)
+				}
 
 				return
 			}
@@ -526,6 +579,27 @@ func TestCLIAcceptsMixedCaseCatalogType(t *testing.T) {
 			require.NoError(t, err, "output: %s", out)
 		})
 	}
+}
+
+func TestMergeConfSQLOptions(t *testing.T) {
+	fileCfg := &config.CatalogConfig{
+		SQLDriver:  "sqliteshim",
+		SQLDialect: "sqlite",
+	}
+
+	t.Run("file values applied when flags absent", func(t *testing.T) {
+		var a Args
+		mergeConf(fileCfg, &a, map[string]bool{})
+		assert.Equal(t, "sqliteshim", a.SQLDriver)
+		assert.Equal(t, "sqlite", a.SQLDialect)
+	})
+
+	t.Run("explicit flags win over file", func(t *testing.T) {
+		a := Args{SQLDriver: "mysql", SQLDialect: "mysql"}
+		mergeConf(fileCfg, &a, map[string]bool{"sql-driver": true, "sql-dialect": true})
+		assert.Equal(t, "mysql", a.SQLDriver)
+		assert.Equal(t, "mysql", a.SQLDialect)
+	})
 }
 
 func TestMergeConfAwsProfile(t *testing.T) {
--- a/cmd/iceberg/args_test.go
+++ b/cmd/iceberg/args_test.go
@@ -534,6 +534,19 @@ func TestCLIAcceptsMixedCaseCatalogType(t *testing.T) {
 			},
 		},
 		{
+			// Missing --sql-driver must reach the SQL registrar (catalog.Load),
+			// not fail as an unrecognized catalog type.
+			name: "sql without sql-driver reaches registrar",
+			args: []string{
+				"list", "--catalog", "sql",
+				"--uri", sqlURI,
+				"--sql-dialect", "sqlite",
+				"--warehouse", sqlWarehouseURI,
+			},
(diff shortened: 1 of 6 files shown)


cmd/iceberg/main.go
```diff
@@ -36,11 +36,13 @@ import (
 	"github.com/apache/iceberg-go/catalog/hadoop"
 	"github.com/apache/iceberg-go/catalog/hive"
 	"github.com/apache/iceberg-go/catalog/rest"
+	sqlcat "github.com/apache/iceberg-go/catalog/sql"
 	"github.com/apache/iceberg-go/config"
 	_ "github.com/apache/iceberg-go/io/gocloud"
 	"github.com/apache/iceberg-go/table"
 
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	_ "github.com/uptrace/bun/driver/sqliteshim"
```
laskoviymishka, line 47: This blank-imports sqliteshim unconditionally but no other driver, so a prebuilt binary run with `--sql-driver pgx --sql-dialect postgres` dies with `sql: unknown driver` and no hint. The `--sql-dialect` help plus cli.md both advertise postgres/mysql/mssql/oracle as if they work.

Since #1692 is really about the production SQL catalogs, I'd either blank-import the common drivers (lib/pq or pgx, go-sql-driver/mysql) so the CLI works out of the box, or narrow the help text and docs to say only sqliteshim is compiled in and anything else needs a custom build. Either is fine, but the advertised-vs-available gap needs to close one way or the other. wdyt?

cmd/iceberg/main.go
```diff
@@ -422,6 +426,26 @@ func initCatalog(ctx context.Context, args Args) catalog.Catalog {
 		}); err != nil {
 			log.Fatal(err)
 		}
+	case catalog.SQL:
+		// Always set uri/warehouse keys (even when empty) so catalog.Load does not
+		// fill them from EnvConfig for a differently typed catalog of the same name.
+		props := iceberg.Properties{
+			"type":      string(catalog.SQL),
+			"uri":       args.URI,
```
laskoviymishka: Setting `uri`/`warehouse` unconditionally fixes the bleed for the flag-driven path, but it also breaks the config-file-only workflow: if someone puts `uri` under this catalog name in `~/.iceberg-go.yaml` and omits `--uri` to inherit it, they now get an empty DSN instead.

The cleaner fix is probably to have mergeConf populate args.URI/args.Warehouse from the config file before initCatalog runs, so the value reaching here is never empty when the file has it, then this guard isn't load-bearing. Same reasoning applies to `credential`, which isn't in the map at all and can still get pulled from EnvConfig. wdyt?

cmd/iceberg/main.go
```diff
@@ -422,6 +426,26 @@ func initCatalog(ctx context.Context, args Args) catalog.Catalog {
...
+		}
+		if len(args.SQLDialect) > 0 {
+			props[sqlcat.DialectKey] = args.SQLDialect
+		}
+
+		if cat, err = catalog.Load(ctx, args.CatalogName, props); err != nil {
+			log.Fatal(err)
+		}
+	case catalog.DynamoDB:
+		log.Fatal("dynamodb catalog is not implemented")
```
laskoviymishka: Every other user-facing error here goes through output.Error + os.Exit(1) so it respects `--output json`; this one log.Fatals straight to stderr. initCatalog doesn't take the output writer today so it may be more than a one-liner, but I'd at least route it through the same path as the other graceful errors, or leave a note that this bypasses the JSON layer.

cmd/iceberg/main.go
```diff
@@ -422,6 +426,26 @@ func initCatalog(ctx context.Context, args Args) catalog.Catalog {
 		}); err != nil {
 			log.Fatal(err)
 		}
+	case catalog.SQL:
```
laskoviymishka, line 438: There's no pre-flight check that `--sql-driver`/`--sql-dialect` are set when `--catalog sql`; a missing driver surfaces as a log.Fatal from deep inside the registrar. We validate other flag combos up front (the rollback selector), so I'd add an early check here that catches both missing flags at once and keeps the message consistent. Not a blocker.

cmd/iceberg/main.go
```diff
@@ -200,7 +202,7 @@ type Args struct {
 	Upgrade          *UpgradeCmd          `arg:"subcommand:upgrade" help:"upgrade table format version"`
 	Rollback         *RollbackCmd         `arg:"subcommand:rollback" help:"roll back to a previous snapshot"`
 
-	Catalog     string `arg:"--catalog" default:"rest" help:"catalog type"`
+	Catalog     string `arg:"--catalog" default:"rest" help:"catalog type (rest, glue, hive, hadoop, sql)"`
```
laskoviymishka: Small one: this lists rest, glue, hive, hadoop, sql but not dynamodb, even though it's recognized now and cli.md mentions it. I'd add it here with the not-implemented caveat so `--help` and the docs agree.

cmd/iceberg/args_test.go
```diff
@@ -495,6 +503,48 @@ func TestCLIAcceptsMixedCaseCatalogType(t *testing.T) {
 			args:    []string{"list", "--catalog", "Hive", "--uri", hiveURI},
 			wantErr: true,
 		},
+		{
+			name: "sql lowercase",
```
laskoviymishka, line 507: These cases prove the happy path (list against an empty sqlite exits 0), nice. Could we also add a `wantErr` case for `--catalog sql` with no `--sql-driver`, asserting it reaches the registrar's `must provide driver` rather than being rejected as an unrecognized type? That's the bit that proves the new branch is actually wired through catalog.Load.

laskoviymishka (changes requested):
Thanks for picking this up, the flag-and-config wiring is clean and symmetric, and returning a clear `dynamodb catalog is not implemented` instead of the generic unrecognized-type error is exactly right. I'd hold it before merging for one thing though.

Right now the binary only compiles in sqliteshim, but the `--sql-dialect` help and both docs pages list postgres, mysql, sqlite, mssql, and oracle as if they all work. A prebuilt CLI pointed at Postgres will die with `sql: unknown driver "pgx"` and no hint about why. Since #1692 is really motivated by the production SQL catalogs, that advertised-vs-actually-available gap is the thing I'd want closed before this lands: either blank-import the common drivers so the CLI works out of the box, or narrow the help text and docs to say only sqliteshim ships by default and anything else needs a custom build. Either is fine, it just needs to be one or the other.

A few smaller things I'd like to settle in this PR before merge:

- the second commit's empty-uri guard fixes the flag path but breaks the config-file-only workflow (omit `--uri`, inherit it from the file, get an empty DSN); I think mergeConf should populate args.URI from the file before initCatalog instead. Same gap exists for `credential`.
- the dynamodb branch log.Fatals straight to stderr, bypassing the output.Error/`--output json` path every other user error uses.
- a `--catalog sql` with no `--sql-driver` negative test, to prove the new branch actually delegates to catalog.Load.

None of these are big. Sort out the driver/docs story and I'm happy to take another pass and approve.

cmd/iceberg/main.go
dgvj-work: Went with the narrower docs path: the default binary still only blank-imports `sqliteshim`, and the `--sql-driver` / `--sql-dialect` help plus `cli.md` / `configuration.md` now say that explicitly. Other dialects remain supported by the SQL catalog library, but need a custom build that blank-imports their drivers.

cmd/iceberg/main.go
dgvj-work: `mergeConf` already copied `uri` / `warehouse` / `credential` from the file when those flags were omitted; I tightened the tests around that path and also always pass `credential` into the SQL props (same empty-key guard as `uri` / `warehouse`) so a differently typed EnvConfig entry can't bleed through.

cmd/iceberg/main.go
dgvj-work: Moved the DynamoDB rejection into `main` before `initCatalog`, so it goes through `output.Error` + `os.Exit(1)` like the other user-facing checks.

cmd/iceberg/main.go
dgvj-work: Left the missing-driver check to the registrar on purpose so the failure stays `must provide driver` from `catalog.Load` — that also lets the new negative test prove the SQL branch is actually wired through. Happy to add a CLI-level preflight later if you'd rather fail earlier with a flag-oriented message.

cmd/iceberg/main.go
dgvj-work: Added `dynamodb` to the `--catalog` help with the not-implemented caveat so it matches `cli.md`.

cmd/iceberg/args_test.go
dgvj-work: Added a `wantErr` case for `--catalog sql` without `--sql-driver` that asserts `must provide driver` and still rejects `unrecognized catalog type`.

dgvj-work:
Thanks for the review - addressed in bf931dd:

- Docs/help now state that the default CLI binary only ships `sqliteshim` / `sqlite`; other dialects need a custom build.
- Kept `mergeConf` populating `uri` / `warehouse` / `credential` from the config file, expanded tests for that, and always pass `credential` into the SQL props so EnvConfig can't bleed through.
- DynamoDB now fails via `output.Error` before `initCatalog`.
- Added the `--catalog sql` without `--sql-driver` negative test asserting `must provide driver`.

Ready for another pass when you have a moment.

zeroshade (approved):
All six review items are addressed well: the DynamoDB rejection now routes through `output.Error` (case-insensitively, with both spellings tested), the docs honestly state that the default binary ships only `sqliteshim` while the flags accept any compiled-in driver, the config-file `uri`/`warehouse`/`credential` path is covered by tightened `mergeConf` tests, and the deliberate choice to let the registrar own the missing-driver error — proving the branch is wired through `catalog.Load` via the new negative test — is well reasoned. Nice, focused CLI improvement.

Leaving `laskoviymishka`'s formal re-review to him since his change request predates the fix commit, though all his threads are resolved.

---

laskoviymishka (approved):
LGTM

PR #1708, opened 2026-08-08
fix(types): reject decimal scales above precision

fallintoplace:
## Description

validateDecimalPrecisionScale checked positive precision, maximum precision, and non-negative scale, but it did not check that scale is less than or equal to precision. As a result, decimal(5,6) was accepted even though the type is invalid.

Add the missing check. decimal(5,5) remains valid, while decimal(5,6) is rejected by both DecimalTypeOf and JSON decoding.

## Testing

- go test ./... -count=1


zeroshade (approved):
Spec-correct tightening — `scale ≤ precision` enforced at the shared validation point, with the constructor-panic, JSON-parse, and Arrow-conversion surfaces all covered.

---

PR #1709, opened 2026-08-08
fix(types): reject malformed fixed type names

fallintoplace:
## Description

The fixed type parser used strings.HasPrefix(typename, "fixed") together with a generic bracket regex. This allowed fixedfoo[16] to be decoded as FixedType{16}.

Use a fixed-specific regex so only the fixed[N] spelling is accepted. Add regression coverage for fixedfoo[16] and fixedx[5].

## Testing

- go test ./... -count=1

### DIFF
--- a/types.go
+++ b/types.go
@@ -33,10 +33,10 @@ import (
 )
 
 var (
-	regexFromBrackets = regexp.MustCompile(`^\w+\[(\d+)\]$`)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Good catch — the `HasPrefix("fixed")` guard plus the generic `\w+` bracket regex quietly accepted names like `fixedfoo[16]` as valid fixed types. Anchoring on the exact keyword closes it, and the two malformed-prefix cases pin it.

---

PR #1710, opened 2026-08-08
fix(types): require list element type during JSON decoding

fallintoplace:
## Description

ListType.UnmarshalJSON checked element-id but did not require the element type. Missing element data therefore left Element nil and allowed malformed list metadata to be constructed.

Return ErrInvalidSchema at the JSON boundary when the required element field is missing.

## Testing

- go test ./... -count=1


zeroshade (approved):
CI is green after the macOS runner flake cleared on re-run. As noted in my earlier comment, the guard itself mirrors the `element-id` check and the map key/value siblings with the right sentinel and test coverage. (Same heads-up as the siblings: this touches the shared decoding area, so later merges in the family may need trivial rebases.)

---

PR #1711, opened 2026-08-08
fix(types): require map key type during JSON decoding

fallintoplace:
## Description

MapType.UnmarshalJSON checked key-id but did not require the key type. Missing key data therefore left KeyType nil and allowed malformed map metadata to be constructed.

Return ErrInvalidSchema at the JSON boundary when the required key field is missing.

## Testing

- go test ./... -count=1


zeroshade (approved):
Same shape as the sibling decoding guards — right sentinel, message matches the existing convention, test pins identity and message. (Heads-up: this and its sibling PR touch adjacent lines of `MapType.UnmarshalJSON`, so the second to merge will need a trivial rebase.)

---

PR #1712, opened 2026-08-08
fix(types): require map value type during JSON decoding

fallintoplace:
## Description

MapType.UnmarshalJSON checked value-id but did not require the value type. Missing value data therefore left ValueType nil and allowed malformed map metadata to be constructed.

Return ErrInvalidSchema at the JSON boundary when the required value field is missing.

## Testing

- go test ./... -count=1

### DIFF
--- a/types.go
+++ b/types.go
@@ -577,6 +577,9 @@ func (m *MapType) UnmarshalJSON(b []byte) error {
 	if aux.ValueID == nil {
 		return fmt.Errorf("%w: field is missing required 'value-id' key in JSON", ErrInvalidSchema)
 	}
+	if aux.Value.Type == nil {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Same shape as the sibling decoding guards — right sentinel, message matches the existing convention, test pins identity and message. (Heads-up: this and its sibling PR touch adjacent lines of `MapType.UnmarshalJSON`, so the second to merge will need a trivial rebase.)

---

zeroshade:
@fallintoplace Approved and ready — its sibling #1711 merged first and their guards sit on adjacent lines of `MapType.UnmarshalJSON`, so this now needs a trivial rebase. Once CI re-runs green it can merge right away.

PR #1719, opened 2026-08-08
fix(catalog): reject invalid current metadata version before staging

fallintoplace:
## What changed

- Reject invalid current metadata locations before staging updates.
- Keep legacy vN.metadata.json and vN.gz.metadata.json locations working.
- Use version 0 for creates and reject empty locations for existing tables.

## Why

A failed metadata filename parse used to become version 0 through -1 + 1. That could hide malformed catalog state and stage the next update at the wrong version.

## Tests

- `go test ./catalog/internal -count=1`
- `go test ./... -count=1`


zeroshade (approved):
just needs the merge conflicts resolved

PR #1722, opened 2026-08-08
fix(table): compare sort order IDs by value

fallintoplace:
## What

`MetadataBuilder.AddSortOrder` compared `lastAddedSortOrderID` with the address of a fresh local variable. Re-adding an equivalent sort order could therefore append duplicate add-sort-order updates. This compares the stored ID by value instead.

## Testing

- `go test ./table -count=1`


zeroshade (approved):
Good catch — comparing against the address of a per-call local made the guard always-true, so equivalent re-adds stacked duplicate `add-sort-order` updates. Value comparison with the nil guard is the correct fix, and assigning the resolved ID to the caller's order on the dedup path too is a nice consistency improvement pinned by the test.

---

PR #1723, opened 2026-08-08
fix(literals): return independent bool marshal bytes

fallintoplace:
## Summary

BoolLiteral.MarshalBinary currently returns a slice backed by a package-level buffer. Since the returned []byte is mutable, changing it also changes the result of later calls.

Return a new one-byte slice for each call instead. The regression test covers both true and false, mutates the first result, and checks that later calls still return the correct value.

## Testing

- go test . -run TestBoolLiteralMarshalBinaryReturnsIndependentBytes -count=1
- go list ./... | rg -v io/gocloud | xargs go test -count=1
- go test ./io/gocloud -run ^$ -count=1

### DIFF
--- a/literals.go
+++ b/literals.go
@@ -424,15 +424,13 @@ func (b BoolLiteral) Equals(l Literal) bool {
 	return literalEq(b, l)
 }
 
-var falseBin, trueBin = [1]byte{0x0}, [1]byte{0x1}
-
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Genuine aliasing bug — the returned slice was backed by shared package-level state, so one caller's mutation poisoned every subsequent marshal. Per-call allocation is the right fix at this size, and the regression test's mutate-then-recheck shape pins it.

---

PR #1725, opened 2026-08-08
fix(table): read legacy property update field names

fallintoplace:
## What

Accept the legacy `updated` and `removed` fields when decoding `set-properties` and `remove-properties` updates.

The current `updates` and `removals` names remain canonical when both forms are present and when the update is marshaled.

## Tests

- `go test ./table -count=1`
- `go test ./...`


tanmayrauth (approved):
LGTM!

laskoviymishka (approved):
LGTM

PR #1726, opened 2026-08-08
fix(table): preserve unknown snapshot operation values

fallintoplace:
## What

Preserve unknown non-empty `summary.operation` values when reading snapshots. This keeps readers forward-compatible with newer operation values.

Explicitly empty operation values are still rejected, and known operation validation for write/update paths is unchanged.

## Tests

- `go test ./table -count=1`
- `go test ./...`

### DIFF
--- a/table/snapshots.go
+++ b/table/snapshots.go
@@ -233,9 +233,10 @@ func (s *Summary) UnmarshalJSON(b []byte) (err error) {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Right call for forward compatibility — Java's `SnapshotParser` likewise stores the operation string without closed-set validation, so this brings the read boundary in line while keeping empty/null rejection and leaving write-path validation strict. The round-trip test pinning preservation is a good touch.

---

PR #1727, opened 2026-08-08
fix(table): preserve embedded v1 snapshot manifests

fallintoplace:
## Summary

Preserve the V1 `manifests` array when reading and writing snapshot metadata.

- Keep embedded manifest locations during JSON decoding.
- Decode snapshots into a fresh value so stale `manifest-list` and `manifests` state cannot survive receiver reuse.
- Normalize dual-field input to the `manifest-list` representation.
- Preserve an explicitly empty `manifests` array.
- Resolve each embedded manifest length through the supplied IO before creating a manifest-list entry.
- Omit V1 snapshot sequence numbers when serializing the embedded representation.
- Prefer `manifest-list` when both representations are present.

Previously, the `manifests` array was ignored and synthetic entries used `-1` for the required manifest length. That value could be carried into a later manifest list when a snapshot inherited the V1 manifests.

## Testing

- `go test ./table -count=1`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2064,7 +2064,10 @@ func cloneSnapshotRef(ref SnapshotRef) SnapshotRef {
 
 func cloneSnapshot(snapshot Snapshot) Snapshot {
 	clone := snapshot
-	clone.ManifestLocations = slices.Clone(snapshot.ManifestLocations)
+	if snapshot.ManifestLocations != nil {
+		clone.ManifestLocations = make([]string, len(snapshot.ManifestLocations))
+		copy(clone.ManifestLocations, snapshot.ManifestLocations)
+	}
 	if snapshot.ParentSnapshotID != nil {
 		value := *snapshot.ParentSnapshotID
 		clone.ParentSnapshotID = &value
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2064,6 +2064,7 @@ func cloneSnapshotRef(ref SnapshotRef) SnapshotRef {
 
 func cloneSnapshot(snapshot Snapshot) Snapshot {
 	clone := snapshot
+	clone.ManifestLocations = slices.Clone(snapshot.ManifestLocations)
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
Builds clean and the full `./table` suite passes on the PR head. The diff matches the description, and the semantics line up with Java's SnapshotParser: prefer `manifest-list` when both fields are present on read and write, spec-id 0 for synthesized entries, and v1 embedded manifests always carry explicit entry snapshot IDs so hard-coding `AddedSnapshotID` to the containing snapshot in `Snapshot.Manifests` (table/snapshots.go:397-411) is safe. Nice side effect: orphan cleanup (table/orphan_cleanup.go:461) previously saw zero manifests for embedded-v1 snapshots and could have treated them as unreferenced; now they're accounted for, and a missing manifest fails loudly instead of silently vanishing.

A few non-blocking notes:

- table/snapshots.go:315-317: the `sequence-number` omission only applies to the embedded-manifests form. Java omits it whenever it equals the initial sequence number regardless of representation, so v1 snapshots with a `manifest-list` still serialize `"sequence-number": 0` here. That's pre-existing behavior and readers tolerate it, but worth knowing the divergence isn't fully closed.
- table/snapshots.go:419-444: `embeddedManifestLength` does a stat/open per manifest on every `Manifests()` call with no caching, and `Manifests` sits on hot paths (`dataFiles`, `entries`, orphan cleanup). Only legacy v1 embedded tables pay this, so I think it's fine, but if it shows up in practice a cached length map would be the fix.
- The `-1` added/existing/deleted counts get persisted into a rebuilt manifest list on inheritance (as your v1 inheritance test exercises). Java uses null for unknown counts; `-1` is the established unknown sentinel in this codebase's builders, so consistent — just noting it's a written artifact, not only in-memory.

Test coverage is thorough (roundtrip, empty-array preservation, dual-field precedence, receiver-reuse, and the end-to-end fast-append inheritance case). LGTM.

PR #1731, opened 2026-08-09
fix(table): align float partition equality

fallintoplace:
## Summary

Use one Iceberg-compatible equality path for partition values in equality-delete matching, partitioned fanout grouping, and clustered partition tracking.

## Problem

The old writer path used raw float values as Go map keys, while scanner matching used `reflect.DeepEqual`. Those paths did not have the same floating-point behavior:

- NaN values did not match themselves in equality-delete partition matching.
- Different NaN payloads were not treated as the same value by scanner matching.
- `-0.0` and `+0.0` could be grouped together by Go map equality even though their float bit patterns are different.
- The same partition could therefore be handled differently by delete matching and by the writers.

## What changed

- Convert float32 and float64 partition values to width-specific bit-pattern keys.
- Canonicalize NaNs separately for float32 and float64.
- Preserve signed zero.
- Keep float32 and float64 values as distinct partition types.
- Keep binary partition values compared by byte content.
- Reuse the normalized comparator for scanner partition matching and writer key generation.

This keeps equality-delete matching, fanout grouping, and clustered partition tracking consistent. The behavior is aligned with Java Iceberg type-aware partition comparators in [`PartitionMap`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/PartitionMap.java), [`StructLikeMap`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/StructLikeMap.java), and [`Comparators`](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Comparators.java).

## Tests

- Added coverage for float32 and float64 NaNs and NaN payloads, signed zero, binary values, float widths, equality-delete matching, fanout grouping, and clustered tracking.
- `go test ./table -count=1`
- `go test -race ./table -run "Test(ClusteredWriter|FanoutWriter|PartitionsMatchUsesFloatSemantics|MatchEqualityDeletesToDataHandlesFloatPartitions)" -count=1`

### DIFF
--- a/table/clustered_writer_test.go
+++ b/table/clustered_writer_test.go
@@ -42,40 +42,56 @@ import (
 func TestClusteredPartitionTrackingUsesComparableKeys(t *testing.T) {
 	t.Parallel()
 
-	records := []partitionRecord{
-		{[]byte{1, 2, 3}},
-		{math.NaN()},
-		{math.Float64frombits(0x7ff8000000000001)},
-	}
-	completed := make(closedPartitionSet)
-	for _, record := range records {
-		completed.add(record)
-		require.True(t, completed.contains(record))
-		require.True(t, partitionRecordsEqual(record, record))
+	tests := []struct {
+		name       string
+		record     partitionRecord
+		equivalent partitionRecord
+	}{
+		{
+			name:       "binary values",
+			record:     partitionRecord{[]byte{1, 2, 3}},
+			equivalent: partitionRecord{[]byte{1, 2, 3}},
+		},
+		{
+			name:       "float64 NaN payloads",
+			record:     partitionRecord{math.Float64frombits(0x7ff8000000000001)},
+			equivalent: partitionRecord{math.Float64frombits(0x7ff8000000000002)},
+		},
+		{
+			name:       "float32 NaN payloads",
+			record:     partitionRecord{math.Float32frombits(0xffc00001)},
+			equivalent: partitionRecord{math.Float32frombits(0x7fc00002)},
+		},
 	}
 
-	require.True(t, completed.contains(partitionRecord{[]byte{1, 2, 3}}))
-	require.True(t, completed.contains(partitionRecord{math.NaN()}))
-	require.True(t, completed.contains(partitionRecord{math.Float64frombits(0x7ff8000000000002)}))
-	signedZeros := make(closedPartitionSet)
-	signedZeros.add(partitionRecord{math.Copysign(0, -1)})
-	require.True(t, signedZeros.contains(partitionRecord{math.Copysign(0, -1)}))
-	require.False(t, signedZeros.contains(partitionRecord{float64(0)}))
-	require.True(t, partitionRecordsEqual(partitionRecord{[]byte{1, 2, 3}}, partitionRecord{[]byte{1, 2, 3}}))
-	require.True(t, partitionRecordsEqual(partitionRecord{math.NaN()}, partitionRecord{math.NaN()}))
-	require.True(t, partitionRecordsEqual(
-		partitionRecord{math.Float64frombits(0x7ff8000000000001)},
-		partitionRecord{math.Float64frombits(0x7ff8000000000002)},
-	))
-	require.True(t, partitionRecordsEqual(
-		partitionRecord{math.Float32frombits(0xffc00001)},
-		partitionRecord{math.Float32frombits(0x7fc00002)},
-	))
-	require.False(t, partitionRecordsEqual(
-		partitionRecord{math.Float32frombits(0x7fc00001)},
-		partitionRecord{math.Float64frombits(0x7ff8000000000001)},
-	))
-	require.False(t, partitionRecordsEqual(partitionRecord{math.Copysign(0, -1)}, partitionRecord{float64(0)}))
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			completed := make(closedPartitionSet)
+			completed.add(tt.record)
+			require.True(t, completed.contains(tt.record))
+			require.True(t, completed.contains(tt.equivalent))
+			require.True(t, partitionRecordsEqual(tt.record, tt.record))
+			require.True(t, partitionRecordsEqual(tt.record, tt.equivalent))
+		})
+	}
+
+	t.Run("float widths stay distinct", func(t *testing.T) {
+		require.False(t, partitionRecordsEqual(
+			partitionRecord{math.Float32frombits(0x7fc00001)},
+			partitionRecord{math.Float64frombits(0x7ff8000000000001)},
+		))
+	})
+
+	t.Run("signed zero stays distinct", func(t *testing.T) {
+		signedZeros := make(closedPartitionSet)
+		signedZeros.add(partitionRecord{math.Copysign(0, -1)})
+		require.True(t, signedZeros.contains(partitionRecord{math.Copysign(0, -1)}))
+		require.False(t, signedZeros.contains(partitionRecord{float64(0)}))
+		require.False(t, partitionRecordsEqual(
+			partitionRecord{math.Copysign(0, -1)},
+			partitionRecord{float64(0)},
+		))
+	})
 }
 
 type ClusteredWriterTestSuite struct {
--- a/table/clustered_writer_test.go
+++ b/table/clustered_writer_test.go
@@ -45,6 +45,7 @@ func TestClusteredPartitionTrackingUsesComparableKeys(t *testing.T) {
 	records := []partitionRecord{
 		{[]byte{1, 2, 3}},
 		{math.NaN()},
+		{math.Float64frombits(0x7ff8000000000001)},
 	}
 	completed := make(closedPartitionSet)
 	for _, record := range records {
@@ -55,8 +56,18 @@ func TestClusteredPartitionTrackingUsesComparableKeys(t *testing.T) {
 
 	require.True(t, completed.contains(partitionRecord{[]byte{1, 2, 3}}))
 	require.True(t, completed.contains(partitionRecord{math.NaN()}))
+	require.True(t, completed.contains(partitionRecord{math.Float64frombits(0x7ff8000000000002)}))
+	signedZeros := make(closedPartitionSet)
+	signedZeros.add(partitionRecord{math.Copysign(0, -1)})
+	require.True(t, signedZeros.contains(partitionRecord{math.Copysign(0, -1)}))
+	require.False(t, signedZeros.contains(partitionRecord{float64(0)}))
 	require.True(t, partitionRecordsEqual(partitionRecord{[]byte{1, 2, 3}}, partitionRecord{[]byte{1, 2, 3}}))
 	require.True(t, partitionRecordsEqual(partitionRecord{math.NaN()}, partitionRecord{math.NaN()}))
+	require.True(t, partitionRecordsEqual(
+		partitionRecord{math.Float64frombits(0x7ff8000000000001)},
+		partitionRecord{math.Float64frombits(0x7ff8000000000002)},
+	))
+	require.False(t, partitionRecordsEqual(partitionRecord{math.Copysign(0, -1)}, partitionRecord{float64(0)}))
 }
 
 type ClusteredWriterTestSuite struct {
--- a/table/clustered_writer_test.go
+++ b/table/clustered_writer_test.go
@@ -67,6 +67,14 @@ func TestClusteredPartitionTrackingUsesComparableKeys(t *testing.T) {
 		partitionRecord{math.Float64frombits(0x7ff8000000000001)},
 		partitionRecord{math.Float64frombits(0x7ff8000000000002)},
 	))
+	require.True(t, partitionRecordsEqual(
+		partitionRecord{math.Float32frombits(0xffc00001)},
+		partitionRecord{math.Float32frombits(0x7fc00002)},
+	))
+	require.False(t, partitionRecordsEqual(
+		partitionRecord{math.Float32frombits(0x7fc00001)},
+		partitionRecord{math.Float64frombits(0x7ff8000000000001)},
+	))
 	require.False(t, partitionRecordsEqual(partitionRecord{math.Copysign(0, -1)}, partitionRecord{float64(0)}))
 }
 
--- a/table/equality_delete_index.go
+++ b/table/equality_delete_index.go
@@ -20,7 +20,6 @@ package table
 import (
 	"cmp"
 	"fmt"
-	"math"
 	"slices"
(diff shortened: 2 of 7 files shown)


zeroshade (approved):
LGTM just needs to resolve the merge conflict

catalog/rest/load_table_bench_test.go
```diff
@@ -81,7 +81,7 @@ func BenchmarkDecodeTableMetadata(b *testing.B) {
 			b.ResetTimer()
 			b.ReportAllocs()
 
-			for i := 0; i < b.N; i++ {
+			for range b.N {
```
laskoviymishka, line 84: This range-over-int cleanup is unrelated to the float-equality fix. I'd split it into its own commit so a bisect on this change stays clean. Not blocking.

table/clustered_writer_test.go
```diff
@@ -55,8 +56,26 @@ func TestClusteredPartitionTrackingUsesComparableKeys(t *testing.T) {
 
 	require.True(t, completed.contains(partitionRecord{[]byte{1, 2, 3}}))
 	require.True(t, completed.contains(partitionRecord{math.NaN()}))
+	require.True(t, completed.contains(partitionRecord{math.Float64frombits(0x7ff8000000000002)}))
```
laskoviymishka: This block stacks four distinct scenarios — cross-payload NaN equality, float32 NaN, cross-width inequality, signed-zero distinctness — as flat `require` calls, so a failure won't tell you which one broke. I'd wrap each in a `t.Run` so failures point at the scenario. Non-blocking.

table/partitioned_fanout_writer.go
```diff
@@ -75,35 +76,50 @@ type partitionExtractionPlan struct {
...
+	binaryPartitionKey  string
+	float32PartitionKey uint32
+	float64PartitionKey uint64
+)
 
-type nanPartitionKey struct {
-	bits int
-}
+const (
+	canonicalFloat32NaNBits uint32 = 0x7fc00000
```
laskoviymishka, line 90: Worth a short comment on these constants spelling out the two deliberate choices, since both are behavioral.

Canonicalizing every NaN payload to `0x7fc00000` matches Java's `floatToIntBits`, but it collapses `-NaN` and `+NaN`, which the spec's sort-order table actually keeps distinct — following Java over the spec total order is the right call here, just worth saying so.

And keeping signed zeros separate (`-0.0 ≠ +0.0`) flips the old `==` path where they compared equal. It's spec-correct per `Float.compare`, but a behavioral shift that I'd also call out in the PR description.

table/partitioned_fanout_writer.go
```diff
@@ -75,35 +76,50 @@ type partitionExtractionPlan struct {
...
+	canonicalFloat64NaNBits uint64 = 0x7ff8000000000000
+)
 
 func comparablePartitionKey(value any) any {
 	switch value := value.(type) {
 	case []byte:
 		return binaryPartitionKey(value)
 	case float32:
+		bits := math.Float32bits(value)
```
laskoviymishka: `bits` here (and in the float64 case below) shadows the `math/bits` import, which this file uses in `initialPartitionRowCapacity` via `bits.Len64`. It's inert since each case returns before touching the package, but `go vet -shadow` and gocritic will flag it and it's a double-take right next to the import. I'd rename both to `rawBits`.

table/partitioned_fanout_writer.go
```diff
@@ -75,35 +76,50 @@ type partitionExtractionPlan struct {
...
 		}
+
+		return float64PartitionKey(bits)
 	}
 
 	return value
 }
 
+func partitionValuesEqual(left, right any) bool {
+	return reflect.DeepEqual(comparablePartitionKey(left), comparablePartitionKey(right))
```
laskoviymishka: I'd drop `reflect.DeepEqual` here for plain `==`.

Everything `comparablePartitionKey` returns is already a comparable type — the named uint32/uint64 keys, `binaryPartitionKey` (string), or a pass-through primitive — and the map-key paths (`getOrCreate`, `closedPartitionSet`) already compare these with `==`. `reflect.DeepEqual` just adds reflection overhead and makes the intent murkier than the `!=` it replaced.

The one thing `==` gives up is that it'll panic rather than quietly mishandle a future non-comparable pass-through type. If that's a worry I'd guard it in `comparablePartitionKey`'s default branch (or leave a one-line note on the invariant) rather than paying for reflect on every call.

table/scanner.go
```diff
@@ -609,7 +609,7 @@ func matchEqualityDeletesToData(dataEntry iceberg.ManifestEntry, eqDeleteEntries
 }
 
 func partitionsMatch(a, b map[int]any) bool {
-	return maps.EqualFunc(a, b, reflect.DeepEqual)
+	return maps.EqualFunc(a, b, partitionValuesEqual)
```
laskoviymishka, line 645: This swaps `partitionsMatch` to the new comparator, but `matchEqualityDeletesToData` — its only caller — is dead code on current main. The live equality-delete path is `buildEqualityDeleteIndex` + `eqDeleteIndex.forDataFile` over in `equality_delete_index.go`. So the fanout-writer half of this PR is genuinely active and fixed, but this half corrects a dormant function and won't change any production scan.

The active path also normalizes floats differently: `comparableEqualityDeletePartitionValue` promotes float32 to float64 bits and uses one canonical NaN type across widths, so `float32(1.5)` matches `float64(1.5)` there — while `comparablePartitionKey` keeps the widths distinct. Under a float→double partition-spec evolution those two conventions disagree.

Given the description says this "fixes equality-delete matching," I'd either reword to say the scanner change future-proofs a dormant path, or apply the same normalization to `comparableEqualityDeletePartitionValue` so the live index path actually benefits — and settle which float-width convention is the intended one. The added `TestMatchEqualityDeletesToDataHandlesFloatPartitions` exercises the dormant function directly, so it passes without touching the active path. wdyt?

table/scanner_internal_test.go
```diff
@@ -285,6 +284,108 @@ func TestEqualityDeletePartitionKeyDistinguishesSignedZero(t *testing.T) {
...
+		right any
+		match bool
+	}{
+		{name: "float32 NaN values", left: float32(math.NaN()), right: float32(math.NaN()), match: true},
+		{name: "float64 NaN values", left: math.NaN(), right: math.NaN(), match: true},
+		{name: "float32 NaN payloads", left: math.Float32frombits(0x7fc00001), right: math.Float32frombits(0x7fc00002), match: true},
+		{name: "float64 NaN payloads", left: math.Float64frombits(0x7ff8000000000001), right: math.Float64frombits(0x7ff8000000000002), match: true},
+		{name: "float32 signed zero", left: float32(math.Copysign(0, -1)), right: float32(0), match: false},
+		{name: "float64 signed zero", left: math.Copysign(0, -1), right: float64(0), match: false},
+		{name: "float widths differ", left: float32(1), right: float64(1), match: false},
```
laskoviymishka, line 486: This table only exercises float and binary values. A pass-through case (`int64(42)` matches, `string("a") != string("b")`) plus a `nil`-vs-`nil` case would pin down the two branches that aren't float — especially relevant if we switch to `==` above, since that's exactly where a non-comparable type would surface. Cheap to add here.

table/scanner_internal_test.go
```diff
@@ -285,6 +284,108 @@ func TestEqualityDeletePartitionKeyDistinguishesSignedZero(t *testing.T) {
...
+func TestMatchEqualityDeletesToDataHandlesBinaryPartitions(t *testing.T) {
+	dataSeqNum := int64(1)
+	deleteSeqNum := int64(2)
+	dataFile := &mockDataFile{
+		path:      "data.parquet",
+		partition: map[int]any{1000: []byte{0xde, 0xad}},
+	}
+	matchingDelete := &mockDataFile{
+		path:      "matching-delete.parquet",
+		partition: map[int]any{1000: append([]byte(nil), 0xde, 0xad)},
```
laskoviymishka: The `append([]byte(nil), ...)` reads like a copy of something when it's really just constructing a fresh slice. A plain `[]byte{0xde, 0xad}` literal makes the same point — it's already a different backing array from the data file's literal. Minor.

laskoviymishka (changes requested):
Nice fix — unifying the three equality call-sites behind one comparator is the right instinct, and canonicalizing NaN, keeping signed zeros distinct, and keeping the float widths separate is exactly the set of properties to nail here. The fanout-writer coverage is thorough.

I see @zeroshade already approved modulo the merge conflict; I'd still hold for one thing before it goes in.

The scanner.go change swaps `partitionsMatch` over to the new comparator, but that function's only live caller is gone on current main — the active equality-delete path is the index in `equality_delete_index.go`, and it normalizes floats differently (float32 promoted to float64 bits, one canonical NaN across widths, so `float32(1.5)` and `float64(1.5)` match there). So "fixes equality-delete matching" only half-holds: the writer side is genuinely fixed, but the scanner side corrects a dormant function and leaves two disagreeing float conventions in the same package. I'd either reword the claim to say it future-proofs a dormant path, or push the same normalization into the active index path and settle which width convention is intended.

- drop `reflect.DeepEqual` in `partitionValuesEqual` for plain `==` — every normalized type is comparable and the map-key paths already rely on that; guard the default branch if a future non-comparable type is a concern
- rename the `bits` locals in `comparablePartitionKey` — they shadow the `math/bits` import the file uses elsewhere
- add a comment on the NaN/signed-zero semantics (and a line in the PR description), since `-0.0 ≠ +0.0` is a behavioral change from the old `==` path
- a pass-through and a `nil` case in `TestPartitionsMatchUsesFloatSemantics` would cover the non-float branches

Also worth splitting the unrelated range-over-int cleanup in `load_table_bench_test.go` into its own commit.

Once the equality-delete path is settled and those are in, happy to take another pass and approve.

table/partitioned_fanout_writer.go
```diff
@@ -75,35 +75,54 @@ type partitionExtractionPlan struct {
...
+	canonicalFloat32NaNBits uint32 = 0x7fc00000
+	canonicalFloat64NaNBits uint64 = 0x7ff8000000000000
+)
 
 func comparablePartitionKey(value any) any {
 	switch value := value.(type) {
 	case []byte:
 		return binaryPartitionKey(value)
 	case float32:
+		bits := math.Float32bits(value)
```
laskoviymishka: This is the rename I flagged last round, and it's live: `math/bits` is imported (line 26) and used as `bits.Len64` in `initialPartitionRowCapacity`, so these `bits` locals shadow the package inside the float cases.

It compiles today because the usage is in a different function, but the shadow linter will trip and anyone adding a `math/bits` call to these cases gets a confusing error. I'd rename both here and in the float64 case at line 103 to something like `rawBits`.

table/scanner.go
laskoviymishka, line 645: Good — routing this through `partitionValuesEqual` unifies the float convention with the writer, which is what I was after last round.

One thing that's still uneven: `partitionValuesEqual` compares integer partition values with a raw `==`, so `int32(7)` and `int64(7)` come out unequal here, whereas the active index path (`comparableEqualityDeletePartitionValue`) normalizes them to a common width and treats them as equal. Since `matchEqualityDeletesToData` is the only caller of `partitionsMatch` and it's still dead code, this doesn't bite production — but the two paths still disagree, just on ints now instead of floats.

I'd either normalize integers in `partitionValuesEqual` too, or drop a short comment on `matchEqualityDeletesToData` noting it's a dormant reference impl and the index path is the live one. The PR description still reads as though this is the production path. wdyt?

table/scanner_internal_test.go
```diff
@@ -232,12 +232,39 @@ func TestEqualityDeletePartitionKeyNormalizesValues(t *testing.T) {
 		name  string
 		left  any
 		right any
+		equal bool
 	}{
-		{name: "integer widths", left: int32(7), right: int64(7)},
-		{name: "date and integer", left: iceberg.Date(7), right: int32(7)},
+		{name: "integer widths", left: int32(7), right: int64(7), equal: true},
+		{name: "date and integer", left: iceberg.Date(7), right: int32(7), equal: true},
 		{name: "float widths", left: float32(1.5), right: float64(1.5)},
```
laskoviymishka, line 239: This case has no `equal` field now, so it's silently asserting `float32(1.5)` and `float64(1.5)` are distinct — which is correct, but the name gives no hint that's the intent.

Your new `TestEqualityDeleteIndexUsesFloatSemantics` already names the equivalent case `float widths stay distinct`; I'd match that here.

table/equality_delete_index.go
```diff
@@ -106,19 +103,10 @@ func comparableEqualityDeletePartitionValue(value any) (any, error) {
...
-	case float64:
-		if math.IsNaN(value) {
-			return equalityDeleteNaNPartitionValue{}, nil
-		}
-
-		return equalityDeleteFloatPartitionValue(math.Float64bits(value)), nil
+	case float32, float64:
+		// Keep equality-delete indexing aligned with writer partition keys:
+		// canonicalize NaNs within each width and preserve signed zero.
+		return comparablePartitionKey(value), nil
```
laskoviymishka, line 109: Nice — delegating to `comparablePartitionKey` is exactly the unification I wanted.

Not for this PR, but `comparablePartitionKey` and `partitionValuesEqual` are now shared partition-equality primitives that live in `partitioned_fanout_writer.go`, and this reader-side file reaches across into the writer file to use them. A follow-up that pulls them into something like `partition_equality.go` would read more naturally. Non-blocking.

laskoviymishka (approved):
Almost there — the float-convention concern from last round is genuinely resolved. Both the scanner path and the delete index now go through `comparablePartitionKey`, the canonical-NaN and signed-zero handling matches Java's `Comparators`, and Float vs Double stay distinct by construction. Nice work.

Two bits of cleanup I'd still like before this goes in, neither a hard blocker. First, the `bits` rename I flagged last round didn't make it — `math/bits` is imported and used as `bits.Len64` in the same file, so the `bits` locals in the float cases really do shadow it, and the linter will catch it. Second, `matchEqualityDeletesToData` is still framed as the live equality-delete path, but production runs through the index; I'd either add a line noting it's a dormant reference impl or unify its integer handling with the index path (details inline).

Everything else I asked for last round is in:
- `partitionsMatch` uses `partitionValuesEqual` instead of `reflect.DeepEqual`
- the canonical-NaN / signed-zero rationale is captured in the const block
- the pass-through and nil cases are covered, plus new signed-zero and NaN-payload cases
- the delete index delegates its float cases to `comparablePartitionKey` rather than carrying its own normalization

Fix the shadow and settle the dormant-path framing and I'm happy to approve.

PR #1735, opened 2026-08-09
fix(manifest): reject unknown legacy row counts

fallintoplace:
## Summary

V1 manifest lists can omit added_rows_count and existing_rows_count. The [First Row ID Assignment](https://iceberg.apache.org/spec/#first-row-id-assignment) section of the Iceberg specification says missing row counts are assumed to be non-zero. During v3 first_row_id assignment, converting the legacy -1 sentinel to zero can leave no space for rows that readers later assign from record_count.

For upgraded tables, the specification also requires post-upgrade snapshots to assign row IDs to existing and added files. See [Row Lineage for Upgraded Tables](https://iceberg.apache.org/spec/#row-lineage-for-upgraded-tables).

This is the follow-up to [#1627](https://github.com/apache/iceberg-go/pull/1627).

## Changes

- reject unknown legacy row counts during v3 first_row_id assignment
- keep the row ID cursor unchanged when validation fails
- cover both unknown counts and each asymmetric case in the regression test

## Testing

- go test . -run TestManifests/TestV3ManifestList(RejectsV1ManifestWithUnknownRowCounts|WriterRejectsInvalidRowIDRanges) -count=1 -v
- go test ./... -count=1

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1648,6 +1648,7 @@ type ManifestListWriter struct {
 	sequenceNumber   int64
 	writer           *ocf.Writer
 	nextRowID        *int64
+	failure          error
 }
 
 func NewManifestListWriterV1(out io.Writer, snapshotID int64, parentSnapshot *int64) (*ManifestListWriter, error) {
@@ -1722,7 +1723,7 @@ func advanceRowID(firstRowID, existingRows, addedRows int64) (int64, error) {
 		return 0, fmt.Errorf("%w: first row ID must be non-negative: %d", ErrInvalidArgument, firstRowID)
 	}
 	if existingRows == -1 || addedRows == -1 {
-		return 0, fmt.Errorf("%w: cannot assign row-lineage IDs with unknown row counts: existing=%d added=%d",
+		return 0, fmt.Errorf("%w: cannot assign row-lineage IDs because at least one row count is unknown (existing=%d added=%d); rewrite or compact the legacy manifest before upgrading to v3",
 			ErrInvalidArgument, existingRows, addedRows)
 	}
 	if existingRows < 0 || addedRows < 0 {
@@ -1768,18 +1769,29 @@ func (m *ManifestListWriter) NextRowID() *int64 {
 	return m.nextRowID
 }
 
+// AddManifests appends manifest files to the list. If it returns an error, the
+// writer is poisoned: callers must close and discard it, and subsequent calls
+// return the original error.
 func (m *ManifestListWriter) AddManifests(files []ManifestFile) (err error) {
+	if m.failure != nil {
+		return m.failure
+	}
 	if len(files) == 0 {
 		return nil
 	}
-	if m.version == 3 && m.nextRowID != nil {
-		batchStartRowID := *m.nextRowID
-		defer func() {
-			if err != nil {
+	hasRowID := m.version == 3 && m.nextRowID != nil
+	var batchStartRowID int64
+	if hasRowID {
+		batchStartRowID = *m.nextRowID
+	}
+	defer func() {
+		if err != nil {
+			if hasRowID {
 				*m.nextRowID = batchStartRowID
 			}
-		}()
-	}
+			m.failure = err
+		}
+	}()
 
 	switch m.version {
 	case 1:
--- a/manifest.go
+++ b/manifest.go
@@ -1721,11 +1721,9 @@ func advanceRowID(firstRowID, existingRows, addedRows int64) (int64, error) {
 	if firstRowID < 0 {
 		return 0, fmt.Errorf("%w: first row ID must be non-negative: %d", ErrInvalidArgument, firstRowID)
 	}
-	if existingRows == -1 {
-		existingRows = 0
-	}
-	if addedRows == -1 {
-		addedRows = 0
+	if existingRows == -1 || addedRows == -1 {
+		return 0, fmt.Errorf("%w: cannot assign row-lineage IDs with unknown row counts: existing=%d added=%d",
+			ErrInvalidArgument, existingRows, addedRows)
 	}
 	if existingRows < 0 || addedRows < 0 {
 		return 0, fmt.Errorf("%w: row counts must be non-negative: existing=%d added=%d",
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -22,6 +22,7 @@ import (
 	"compress/flate"
 	"encoding/json"
 	"errors"
+	"fmt"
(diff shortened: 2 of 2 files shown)


manifest.go
```diff
@@ -1631,11 +1631,9 @@ func advanceRowID(firstRowID, existingRows, addedRows int64) (int64, error) {
 	if firstRowID < 0 {
 		return 0, fmt.Errorf("%w: first row ID must be non-negative: %d", ErrInvalidArgument, firstRowID)
 	}
-	if existingRows == -1 {
-		existingRows = 0
-	}
-	if addedRows == -1 {
-		addedRows = 0
+	if existingRows == -1 || addedRows == -1 {
```
laskoviymishka, line 1725: Rejecting here is the right call and matches Java's behavior (it NPEs on the same input). The trade-off worth naming: the spec says a null count "is assumed to be non-zero", so a V1 table whose manifests legally omit these counts now can't produce a V3 snapshot after upgrade, with no recovery short of rewriting those manifests. The spec does describe a fallback (open the manifest and sum `record_count` for files whose `first_row_id` is null), but that's expensive and probably out of scope here.

I'd keep the rejection and just make the dead-end recoverable in the operator's eyes, either in the error text or a doc comment: the way out is to rewrite or compact those manifests before upgrading. wdyt, reject-and-document, or is the fallback worth implementing?

manifest.go
```diff
@@ -1631,11 +1631,9 @@ func advanceRowID(firstRowID, existingRows, addedRows int64) (int64, error) {
...
+		return 0, fmt.Errorf("%w: cannot assign row-lineage IDs with unknown row counts: existing=%d added=%d",
```
laskoviymishka: In the partial cases (one count is -1, the other a real value) the message reads "unknown row counts" plural while printing a legitimate count right next to it, so an operator looking at `existing=-1 added=1` has to work out which one is the sentinel. I'd either say "at least one row count unknown" or split into two guards, one message per field. The split also lets the test's `ErrorContains` assert on which field specifically it's rejecting.

manifest_test.go
```diff
@@ -3416,23 +3416,41 @@ func (m *ManifestTestSuite) TestV3ManifestListAcceptsV1AndV2Manifests() {
...
-	legacy := *(manifestFileRecordsV1[0].(*manifestFile))
-	legacy.AddedRowsCount = -1
-	legacy.ExistingRowsCount = -1
+func (m *ManifestTestSuite) TestV3ManifestListRejectsV1ManifestWithUnknownRowCounts() {
+	tests := []struct {
+		name         string
+		addedRows    int64
+		existingRows int64
+	}{
+		{name: "both row counts unknown", addedRows: -1, existingRows: -1},
```
laskoviymishka, line 3856: Small naming thing: "existing row count unknown" and "added row count unknown" read as if only that field is set, but the point of these two cases is that the sibling is a known positive count and the guard still fires. "only existing row count unknown" / "only added row count unknown" captures that intent. While we're here, the loop var is `test` where the rest of this file uses `tt`.

manifest_test.go
```diff
@@ -3416,23 +3416,41 @@ func (m *ManifestTestSuite) TestV3ManifestListAcceptsV1AndV2Manifests() {
...
+			m.Require().NoError(
+				WriteManifestList(1, &v1Buf, snapshotID, nil, nil, 0, []ManifestFile{&legacy}))
+			manifests, err := ReadManifestList(&v1Buf)
+			m.Require().NoError(err)
+			m.Require().Len(manifests, 1)
+
+			var v3Buf bytes.Buffer
+			writer, err := NewManifestListWriterV3(&v3Buf, snapshotID, 1, 1000, nil)
+			m.Require().NoError(err)
+			err = writer.AddManifests(manifests)
```
laskoviymishka, line 3877: All three cases send a single-manifest batch, so `NextRowID() == 1000` and `Close()` succeeding hold trivially here: nothing was encoded before the error fired. The case that actually exercises the risk is a mixed batch, `AddManifests([validV1, unknownCountV1])`, where the first manifest is already Avro-encoded when the second is rejected.

I'd add that case, assert the error, then read the output back and confirm we didn't leave a manifest with a colliding `first_row_id` range. That's also the case that tells us whether the writer is safe to reuse after a partial failure, so I'd pin the contract down here too: either poison the writer after a partial-batch error, or document that it isn't reusable.

manifest_test.go
```diff
@@ -3416,23 +3416,41 @@ func (m *ManifestTestSuite) TestV3ManifestListAcceptsV1AndV2Manifests() {
...
+			m.Require().ErrorIs(err, ErrInvalidArgument)
+			m.Require().ErrorContains(err, "cannot assign row-lineage IDs with unknown row counts")
+			m.Require().ErrorContains(err, legacy.Path)
+			m.EqualValues(1000, *writer.NextRowID())
```
laskoviymishka: This assertion is bare `m.EqualValues` while the rest of the block uses `Require()`, so if the cursor-rollback check fails, execution falls through to `Close()` and you get an ambiguous second failure. I'd make it `m.Require().EqualValues` to stop right there. Old test had the same bare pattern, so no worse than before, just tidier while we're in here.

laskoviymishka (approved):
Good catch, and the fix is right. Coercing the -1 sentinels to 0 was silently handing two manifests the same `first_row_id`, and rejecting is both safer and closer to what Java does (it NPEs on the same input, so an explicit `ErrInvalidArgument` is nicer parity). The table-driven rejection cases read well too.

I'd hold this before merging on one thing.

The new rejection is now reachable from ordinary data. A V1 table whose manifests legally omit row counts is a common artifact, and `AddManifests([validV1, unknownCountV1])` will hit it mid-batch. When it does, the first manifest is already Avro-encoded but the deferred cursor reset rolls `nextRowID` back, so a reused writer will re-assign `first_row_id` over records it already wrote. The overflow path had this same gap, but it was never realistically reachable; this change makes it reachable from data people actually have. I'd at minimum pin the contract with a mixed-batch test, and either poison the writer after a partial-batch failure or document that it isn't reusable after an error.

A couple of things I'd want before merge:

- a `[valid, unknown-count]` two-manifest test that asserts the error and then verifies writer state (Close plus readback, no colliding `first_row_id` ranges), plus a decision on whether the writer is reusable after a batch error
- a note on the recovery path: since an upgraded V1 table with legally-null counts now can't produce a V3 snapshot, it's worth telling the caller (in the error or a doc comment) that rewriting or compacting those manifests first is the way out

PR #1736, opened 2026-08-09
fix(table): align snapshot expiration retention semantics

fallintoplace:
## Summary

`ExpireSnapshots` was using non-standard table-property names and defaults, and it expired young snapshots that were no longer referenced by a branch or tag.

This change:
- uses the standard `history.expire.*` property names
- matches the standard defaults: 1 snapshot, 5 days, and no ref-age limit
- retains young unreferenced snapshots according to the expiration age
- accepts the old unprefixed names as compatibility fallbacks
- gives the standard names precedence when both forms are set
- updates the retention documentation and adds regression coverage

## Test plan

- `go test ./... -count=1`
- CI checks on the pull request

### DIFF
--- a/table/properties.go
+++ b/table/properties.go
@@ -99,14 +99,18 @@ const (
 	ParquetVariantBufferSizeKey     = internal.ParquetVariantBufferSizeKey
 	ParquetVariantBufferSizeDefault = internal.ParquetVariantBufferSizeDefault
 
-	MinSnapshotsToKeepKey     = "min-snapshots-to-keep"
-	MinSnapshotsToKeepDefault = math.MaxInt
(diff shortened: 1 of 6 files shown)


zeroshade (approved):
Retention semantics now line up with Java on all three axes — standard `history.expire.*` names with well-tested legacy fallback and precedence (including the malformed-standard-key case), spec defaults, and young-unreferenced snapshot retention with the `WithOlderThan` override applying to the unreferenced cutoff. Main-branch protection and tag semantics (head-only, no ancestry) check out too.

One request for the record rather than the code: the default-behavior shift (previously "keep everything" absent properties, now 5-day/keep-1 spec defaults) is exactly right but is a real behavior change for existing callers of `ExpireSnapshots()` — please make sure it's flagged prominently in the release notes.

---

PR #1737, opened 2026-08-09
fix(table): respect gc.enabled during snapshot expiration

fallintoplace:
## Summary

`ExpireSnapshots` could remove snapshots and later garbage-collect their files even when the table property `gc.enabled=false`.

This change:
- rejects snapshot expiration when garbage collection is disabled
- applies the guard before staging metadata requirements
- keeps the behavior independent of `WithPostCommit(false)`
- exposes the shared `gc.enabled` property constant and default
- updates the cleanup documentation

The diagnostic follows the current Iceberg Java implementation, using Go's lowercase error-string convention.

## Test plan

- `go test ./... -count=1`
- `go vet ./...`
- `go test -race ./codec/... ./table/...`

### DIFF
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -3274,7 +3274,7 @@ func (t *Transaction) Commit(ctx context.Context) (*Table, error) {
 }
 
 func validateGCEnabledForSnapshotExpiration(props iceberg.Properties) error {
-	if !props.GetBool(GCEnabledKey, GCEnabledDefault) {
+	if !isGCEnabled(props) {
 		return errors.New("cannot expire snapshots: GC is disabled (deleting files may corrupt other tables)")
 	}
 
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -705,7 +705,7 @@ func (t *Transaction) ExpireSnapshots(opts ...ExpireSnapshotsOpt) error {
 		return err
 	}
 	if !meta.props.GetBool(GCEnabledKey, GCEnabledDefault) {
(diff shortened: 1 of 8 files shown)


table/transaction.go
```diff
@@ -3266,6 +3269,14 @@ func (t *Transaction) Commit(ctx context.Context) (*Table, error) {
 	return t.tbl, nil
 }
 
+func validateGCEnabledForSnapshotExpiration(props iceberg.Properties) error {
+	if !props.GetBool(GCEnabledKey, GCEnabledDefault) {
```
zeroshade: `GetBool(..., true)` fails open here: Go's parser accepts values such as `"1"` as true, while parse errors such as `"garbage"` or `"false "` return the true default. That means malformed/non-Java values can still enable snapshot expiration and physical file deletion. Iceberg Java only enables this property for case-insensitive `"true"`. Please make this safety gate fail closed or exactly match Java parsing, use the same helper for post-commit cleanup and `PurgeFiles`, and add tests for these values.

zeroshade (changes requested):
Thanks for tightening the `gc.enabled` enforcement. One safety-critical parsing issue still leaves destructive cleanup enabled for malformed/non-Java boolean values; I left an inline comment with details.

Also, all four commits currently lack the `Signed-off-by:` trailer required by `CONTRIBUTING.md`. Please add DCO sign-offs when updating the branch.

zeroshade (approved):

PR #1743, opened 2026-08-09
fix(rest): reject trailing JSON responses

fallintoplace:
## What changed

Use one response decoder for successful GET and POST requests, and require EOF after the first JSON value.

## Why

The REST client currently decodes one value and ignores the rest of the response body. A response containing a valid object followed by another JSON value is accepted as successful even though it is not a single JSON document.

Whitespace after the response remains valid.

## Testing

- added a trailing JSON test for GET responses
- added a trailing JSON test for POST responses
- `go test ./...`
- `go vet ./...`


zeroshade (approved):
LGTM

Issue #1703, opened 2026-08-07
Table.Inspect: complete metadata table support


fallintoplace:
## Background

`Table.Inspect` was introduced in [#1465](https://github.com/apache/iceberg-go/pull/1465) with the `history` metadata table and extended in [#1476](https://github.com/apache/iceberg-go/pull/1476) with `snapshots`.

The original PRs describe this subsystem as mirroring the Java, PyIceberg, and Rust clients. This issue tracks the metadata tables and keeps the related work together.

Current status: all 16 metadata tables are implemented. `Table.Inspect` now has complete metadata table support.

## Implemented

- [x] `history` ([#1465](https://github.com/apache/iceberg-go/pull/1465))
- [x] `snapshots` ([#1476](https://github.com/apache/iceberg-go/pull/1476))
- [x] `refs` ([#1695](https://github.com/apache/iceberg-go/pull/1695))
- [x] `metadata_log_entries` ([#1696](https://github.com/apache/iceberg-go/pull/1696))
- [x] `manifests` ([#1697](https://github.com/apache/iceberg-go/pull/1697))
- [x] `data_files` ([#1698](https://github.com/apache/iceberg-go/pull/1698))
- [x] `delete_files` ([#1699](https://github.com/apache/iceberg-go/pull/1699))
- [x] `entries` ([#1700](https://github.com/apache/iceberg-go/pull/1700))
- [x] `partitions` ([#1701](https://github.com/apache/iceberg-go/pull/1701))

### Current and all-snapshot file tables

- [x] `files`: combines the current snapshot data and delete files
- [x] `all_files`: returns data and delete files across all tracked snapshots
- [x] `all_data_files`: returns data files across all tracked snapshots
- [x] `all_delete_files`: returns delete files across all tracked snapshots

Implemented in [#1744](https://github.com/apache/iceberg-go/pull/1744), which is merged.

### All-snapshot manifests

- [x] `all_manifests`: returns all manifests across all tracked snapshots ([#1745](https://github.com/apache/iceberg-go/pull/1745), merged)

### All-snapshot entries

- [x] `all_entries`: returns all manifest entries across all tracked snapshots ([#1746](https://github.com/apache/iceberg-go/pull/1746), merged)

### Positional deletes

- [x] `position_deletes`: exposes the individual positional delete records referenced by the current snapshot ([#1747](https://github.com/apache/iceberg-go/pull/1747), merged)

## Implementation notes

[...]

PR #1744, opened 2026-08-09
feat(table): add files metadata tables

fallintoplace:
## Summary

- add files, all_files, all_data_files, and all_delete_files to Table.Inspect
- reuse the content-file Arrow mapping and stream manifest entries in bounded batches
- scan each reachable manifest once for the all_* tables

## Dependency

Built on #1698. I will rebase this after the parent PR lands.

## Testing

- go test ./...

Refs #1703


zeroshade:

table/inspect_files.go
```diff
@@ -106,12 +212,20 @@ func (i InspectTable) manifestEntryReader(
 			rows = 0
 			emitted = true
 
-			return yield(batch, nil)
+			if yield(batch, nil) {
+				return true
+			}
+
+			batch.Release()
```
zeroshade: Once `batch` has been passed to `yield`, ownership has transferred to `ReaderFromIter`, even when `yield` eventually returns false because the reader was released early. In that path the producer releases here, then `iterReader.Release()` releases its `ir.cur` again. An assert-enabled build panics with `too many releases`; normal builds underflow the reference count. Please return the result of `yield` without releasing `batch`. The same issue is present in `emitEmpty` below.

table/inspect_files.go
```diff
@@ -164,7 +278,9 @@ func emptyInspectRecordBatch(alloc memory.Allocator, schema *arrow.Schema) iter.
 		defer bldr.Release()
 
 		batch := bldr.NewRecordBatch()
-		_ = yield(batch, nil)
+		if !yield(batch, nil) {
+			batch.Release()
```
zeroshade: This empty-reader path has the same double-release. If a caller does `Next()` and then `Release()`, `stop()` makes this `yield` return false, this release runs, and `iterReader.Release()` releases the current batch a second time. The producer should not release a record after yielding it; please restore `_ = yield(batch, nil)` and add an early-release regression test for an empty result.

zeroshade (changes requested):
The metadata-table behavior and manifest traversal look good, but the new iterator cleanup has a blocking Arrow ownership bug. Once a record has been passed to `yield`, `ReaderFromIter` owns it. Releasing a partially consumed reader resumes the producer with `yield == false`, then `iterReader.Release` releases the same current record again. I reproduced this under Arrow's `assert` build tag as `panic: too many releases`. Please remove the producer-side releases after `yield` (including the empty-batch paths) and add early-release coverage for populated and empty readers.

table/inspect_internal_test.go
```diff
@@ -1686,6 +1686,263 @@ func inspectDataFileEntries(t *testing.T, spec iceberg.PartitionSpec, count int)
...
+		{name: "all data files", read: tbl.Inspect(WithInspectAllocator(checked)).AllDataFiles},
+		{name: "all delete files", read: tbl.Inspect(WithInspectAllocator(checked)).AllDeleteFiles},
+	}
+
+	for _, tt := range reads {
+		t.Run(tt.name, func(t *testing.T) {
+			rr, err := tt.read(context.Background())
+			require.NoError(t, err)
+			require.True(t, rr.Next())
+			rr.Release()
```
zeroshade, line 1914: **Blocking: this regression is not enforced by CI.** This exercises the right `Next()` → `Release()` path, but `memory.CheckedAllocator` catches leaks, not an Arrow record refcount dropping below zero. Arrow's over-release panic is compiled in only with `-tags assert`, while CI currently runs untagged and race tests. I restored the old producer-side `batch.Release()` while retaining this test: the normal test passed, but the assert-tagged run failed with `panic: too many releases`. Please add assert-tagged CI coverage for these ownership tests (or another assertion mechanism that fails untagged), so the original bug cannot return unnoticed.

zeroshade (changes requested):
The production ownership bug is fixed, and the simplified manifest iterator correctly transfers yielded record ownership to `ReaderFromIter`. One test-enforcement blocker remains: the new early-release tests do not fail under the repository's normal or race CI configurations when the old producer-side release is restored. Arrow detects that underflow only with the `assert` build tag. I mutation-tested this exact path: untagged tests passed, while `go test -tags assert` failed with `panic: too many releases`. Please add assert-tagged CI coverage for these ownership tests, or another mechanism that detects over-release under the standard suite.

---

>

zeroshade (approved):
The ownership fix is now protected by assert-tagged CI coverage. The shared `make test-assert` target exercises populated, filtered-empty, and empty-table early-release paths, and mutation testing confirms that restoring the previous producer-side over-release fails with Arrow's `too many releases` panic. The targeted assert, normal, and race table suites all pass. No remaining findings from me.

---

>

PR #1745, opened 2026-08-09
feat(table): add all manifests metadata table

fallintoplace:
## Summary

- add all_manifests to Table.Inspect
- include reference_snapshot_id and key_metadata
- preserve one row for each snapshot-to-manifest reference

## Dependency

Built on #1697. I will rebase this after the parent PR lands.

## Testing

- go test ./...

Refs #1703


zeroshade:

table/inspect_all_manifests.go
```diff
@@ -0,0 +1,87 @@
...
+// AllManifests returns every manifest referenced by every snapshot currently
+// tracked by the table. A manifest referenced by multiple snapshots produces
+// one row per reference, identified by reference_snapshot_id.
+func (i InspectTable) AllManifests(ctx context.Context) (array.RecordReader, error) {
+	arrowSchema, err := SchemaToArrowSchema(AllManifestsSchema(), nil, true, false)
+	if err != nil {
+		return nil, fmt.Errorf("inspect all manifests: build arrow schema: %w", err)
+	}
+
+	bldr := array.NewRecordBuilder(i.alloc, arrowSchema)
```
zeroshade: This builder remains live across every retained snapshot, so memory grows with the complete set of manifest references (including repeated shared manifests, partition summaries, and key metadata). Because `AllManifests` returns a `RecordReader`, this should use an iterator-backed reader with a bounded builder rather than accumulating the entire history before returning.

table/inspect_all_manifests.go
```diff
@@ -0,0 +1,87 @@
...
+					snapshot.SnapshotID, err)
+			}
+			referenceSnapshotID := snapshot.SnapshotID
+			if err := i.appendManifestRows(ctx, bldr, manifests, &referenceSnapshotID); err != nil {
+				return nil, fmt.Errorf("inspect all manifests: snapshot %d: %w", snapshot.SnapshotID, err)
+			}
+		}
+	}
+
+	rr, err := singleBatchReader(arrowSchema, bldr)
```
zeroshade: Returning a single batch here means callers receive no rows until every snapshot manifest list has been read and every row has been materialized, and they cannot stop early. Please emit fixed-size batches from `ReaderFromIter` while traversing snapshots, retaining at most the current manifest list plus one batch. A test with more rows than the batch limit should assert that multiple batches are produced.

zeroshade (changes requested):
The all-manifests row semantics, schema extensions, and shared row-building refactor look good, but the new all-history path has an unbounded-memory scalability issue. It accumulates every manifest reference from every retained snapshot before returning one record batch. On valid tables with long histories this can create extreme time-to-first-row or OOM before the caller can consume or stop the reader. Please stream bounded Arrow batches while traversing snapshots and add coverage that crosses the batch boundary.

zeroshade (approved):

PR #1747, opened 2026-08-09
feat(table): add position deletes metadata table

fallintoplace:
## Summary

- add position_deletes to Table.Inspect
- expose records from Parquet position delete files and v3 Puffin deletion vectors
- include partition/spec metadata, physical delete file paths, and v3 DV offsets
- omit the empty partition struct for unpartitioned tables and deconflict partition field IDs

## Dependency

Built on #1699, which currently depends on #1698. I will rebase as the parent PRs land.

## Testing

- go test ./...

Refs #1703

### DIFF
--- a/table/inspect_internal_test.go
+++ b/table/inspect_internal_test.go
@@ -20,6 +20,7 @@ package table
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
@@ -3006,7 +3007,7 @@ func TestInspectPositionDeletesParquet(t *testing.T) {
 	require.Equal(t, deletePath, deleteFilePaths.Value(1))
 }
 
-func TestInspectPositionDeletesV3ParquetUsesFileSize(t *testing.T) {
+func TestInspectPositionDeletesV3ParquetLeavesDVMetadataNull(t *testing.T) {
 	ctx := context.Background()
 	memFS := iceio.NewMemFS()
 	deletePath := "mem://position-deletes/table/data/delete-v3.parquet"
@@ -3035,8 +3036,7 @@ func TestInspectPositionDeletesV3ParquetUsesFileSize(t *testing.T) {
 	require.EqualValues(t, 2, record.NumRows())
 	require.EqualValues(t, 7, record.NumCols())
 	require.EqualValues(t, 2, record.Column(5).NullN())
-	require.Equal(t, []int64{fileSize, fileSize},
-		record.Column(6).(*array.Int64).Int64Values())
+	require.EqualValues(t, 2, record.Column(6).NullN())
 }
 
 func TestAppendParquetPositionDeleteRowsRejectsNegativePosition(t *testing.T) {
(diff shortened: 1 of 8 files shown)


zeroshade:

table/inspect_position_deletes.go
```diff
@@ -0,0 +1,694 @@
...
+
+	return nil
+}
+
+func appendPositionDeleteValue(builder array.Builder, value scalar.Scalar) error {
+	if arrow.TypeEqual(builder.Type(), value.DataType()) {
+		return scalar.Append(builder, value)
+	}
+	if !canPromotePositionDeleteValue(value.DataType(), builder.Type()) {
+		return scalar.Append(builder, value)
```
zeroshade: This fallback makes valid Iceberg schema evolution fail whenever the promoted value is nested in a list/map, because `scalar.Append` requires exact Arrow type equality. For example, projecting a deleted row written as `list<int>` into a current `list<long>` field reaches this branch and returns `cannot append scalar of type list<int32> to builder for type list<int64>`. The same issue affects promoted map values and evolution inside container structs; decimal precision widening is also valid but is not handled by `canPromotePositionDeleteValue`. Could this projection recurse through list/map children and support all Iceberg-compatible primitive promotions? Please add coverage for at least nested `list<int>` → `list<long>` and decimal precision widening.

zeroshade (changes requested):
Thanks for adding the position-deletes metadata table. I found one blocking schema-evolution issue in deleted-row projection. Valid Iceberg promotions inside nested collections, and decimal precision widening, currently make the scan fail. Please make projection recursive for container values and cover these evolution cases end to end.

table/inspect_position_deletes.go
fallintoplace: Thanks. This PR is now ready to be reviewed again.

table/inspect_position_deletes.go
```diff
@@ -0,0 +1,865 @@
...
+
+	for row := range int(record.NumRows()) {
+		if err := ctx.Err(); err != nil {
+			return false, err
+		}
+		var deletedRow scalar.Scalar
+		projected := false
+		if rowIndex >= 0 {
+			if projectedRows != nil {
+				deletedRow, err = scalar.GetScalar(projectedRows, row)
```
zeroshade, line 708: Blocking: the nested projected-row path still leaks Arrow allocations. I reran `TestInspectPositionDeletesParquetProjectsEvolvedNestedRow` with only `Inspect` wired to a checked allocator; after releasing the reader and returned record, the one-row scan retained 320 bytes (three builder allocations and two compute-kernel allocations, each 64 bytes). Please balance the ownership in this projection/scalar lifecycle so the checked allocator returns to zero.

table/inspect_internal_test.go
```diff
@@ -2868,6 +2873,1196 @@ func TestAllEntriesSchemaMatchesEntries(t *testing.T) {
...
+			return true, nil
+		},
+	)
+	require.False(t, keepGoing)
+	require.ErrorIs(t, err, iceberg.ErrInvalidSchema)
+	require.ErrorContains(t, err, "null row")
+	require.Zero(t, rows)
+}
+
+func TestInspectPositionDeletesParquetProjectsEvolvedNestedRow(t *testing.T) {
```
zeroshade, line 3167: Please make this nested evolution test use `WithInspectAllocator(memory.NewCheckedAllocator(...))` and assert zero bytes remain. The existing value assertions pass while the scan leaks 320 bytes, so the checked allocator is needed to prevent this ownership regression.

zeroshade (changes requested):
Thanks for the updates. The previous nested list/map promotion blocker is resolved, but one blocking Arrow ownership issue remains; details are inline.

Also, 17 of the 19 commits in this PR still lack the `Signed-off-by:` trailer required by `CONTRIBUTING.md`. Only `b9e9b7b4b1e08b465a8c1c1ec6a9474f171e8c65` and `c450ad7212f8eba4c0c411b9d45e355fe33b6b7e` are signed. Please add sign-offs to the remaining commits before merge.

zeroshade (approved):

Issue #1748, opened 2026-08-10
REST: Add support for ?snapshots query parameter in loadTable


technicolorbeat:
### Proposed Change

The `loadTable` endpoint (`GET /v1/{prefix}/namespaces/{namespace}/tables/{table}`)
accepts an optional `snapshots` query parameter that controls which snapshots are
included in the response:

| Value       | Behaviour                                                        |
|-------------|------------------------------------------------------------------|
| _(omitted)_ | Server default — equivalent to `all`                             |
| `all`       | Return all currently valid snapshots                             |
| `refs`      | Return only snapshots referenced by a named branch or tag        |

The `refs` mode lets clients request a significantly smaller metadata payload for
tables with long snapshot histories, since the server omits unreferenced snapshots.

The iceberg-go REST client's `LoadTable` always omits the parameter (defaulting to
`all`). There is currently no way for callers to opt into the lighter `refs` response.

**Backward compatibility:**

- `LoadTable` behaviour is unchanged; default behavior returns all snapshots, all existing callers are unaffected. `refs` behavior returns only the snapshot referenced to the named branch.

**Spec reference:**
REST Catalog Open API – loadTable:
https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/apache/iceberg/apache-iceberg-1.10.1/open-api/rest-catalog-open-api.yaml#tag/Catalog-API/operation/loadTable

**Willingness to contribute**

- [x] I can contribute this improvement/feature independently
- [ ] I would be willing to contribute this improvement/feature with guidance from the Iceberg community
- [ ] I cannot contribute this improvement/feature at this time

### Specifications

- [x] Table
- [ ] View
- [x] REST
- [ ] Puffin
- [ ] Encryption
- [ ] Other

PR #1749, opened 2026-08-10
feat(rest): support snapshot-loading-mode catalog property in loadTable

technicolorbeat:
## Summary

- add `SnapshotMode` type with `SnapshotModeAll` and `SnapshotModeRefs` constants
- add `snapshot-loading-mode` catalog property (values: `refs` | `all`), read once at
  initialization and applied to every `LoadTable` call — matching the Java `RESTSessionCatalog` behaviour
- add `withQueryParams(url.Values) reqOption` and `queryParams` field to `reqConfig` so
  the `do()` helper merges query parameters onto the request URL without discarding
  existing params on the base URI
- `LoadTable` sends `?snapshots=<mode>` when `snapshot-loading-mode` is set; omits the
  parameter entirely when unset (server default applies, existing behaviour unchanged)

## Testing

- go test ./...

Refs #1748


catalog/rest/rest.go
```diff
@@ -1607,7 +1632,17 @@ func (r *Catalog) RegisterTable(ctx context.Context, identifier table.Identifier
 	return r.tableFromResponse(ctx, identifier, ret.Metadata, ret.MetadataLoc, config, credsVended)
 }
 
+// Loads a table from the catalog. It implements [catalog.Catalog].
 func (r *Catalog) LoadTable(ctx context.Context, identifier table.Identifier) (*table.Table, error) {
+	return r.loadTableWithMode(ctx, identifier, "")
```
alessandro-nori: in the [Java implementation](https://github.com/apache/iceberg/blob/d7d3bd5c3788b882f52fa79c44cf09a11459deab/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L266) `snapshotMode` is configured once at catalog initialization and not per request.
I wonder if we should follow the same approach and, like we did for `pageSize`, extract `snapshotMode` from the context.

alessandro-nori (commented):
When a table is loaded with snapshotMode=refs and the clients tries to load previous snapshots, there should be a way to lazily load the rest of the snapshots and let the client knows that not all snapshots were loaded with the first loadTable request.
See `snapshotsSupplier` in the Java implementation

tanmayrauth (commented):
+1 to @alessandro-nori, this should read snapshot mode from the context rather than add a new exported method.

  Every per-request knob in this catalog already lives on the context: SetPageSize / getPageSize (rest.go:2111-2120), read inside the list paths. Snapshot mode is the same kind of knob, so the context approach keeps it consistent and, more importantly, keeps it usable through the catalog.Catalog interface. As written,  LoadTableWithSnapshotMode exists only on *rest.Catalog, so any code holding a catalog.Catalog has to type-assert to *rest.Catalog to reach it — which generic callers won't do, so the knob effectively  doesn't exist for anything programming against the interface. That's the concrete cost of a bespoke method here.

  Concretely:

    - add SetSnapshotMode(ctx, SnapshotMode) context.Context and getSnapshotMode(ctx),
      mirroring SetPageSize/getPageSize
    - have the existing LoadTable read the mode from ctx; default "" → no ?snapshots
      param, so current behavior is byte-for-byte unchanged
    - drop LoadTableWithSnapshotMode

  LoadTable stays the single entry point, we don't commit a new exported method to maintain/deprecate, and it lines up with the spirit of Java configuring the mode once instead of per-call. The withQueryParams / queryParams plumbing is the right shape and stays exactly as-is under either approach.

  One small thing:
    - refs mode returns a partial snapshot set with no signal to the caller, so a later time-travel to a non-referenced snapshot surfaces as a plain "snapshot not found" with no way to distinguish "absent" from "not loaded." Fine to leave the lazy snapshotsSupplier equivalent for a follow-up, but document the limitation wherever we expose the mode.

catalog/rest/rest.go
```diff
@@ -421,7 +442,11 @@ func do[T any](ctx context.Context, method string, baseURI *url.URL, path []stri
 		rsp *http.Response
 	)
 
-	uri := baseURI.JoinPath(path...).String()
+	u := baseURI.JoinPath(path...)
+	if len(cfg.queryParams) > 0 {
+		u.RawQuery = cfg.queryParams.Encode()
```
zeroshade: `JoinPath` preserves any query string already present on the base URI, and this assignment silently discards it. Safer to merge:

```go
if len(cfg.queryParams) > 0 {
	q := u.Query()
	for k, vs := range cfg.queryParams {
		for _, v := range vs {
			q.Add(k, v)
		}
	}
	u.RawQuery = q.Encode()
}
```

catalog/rest/rest.go
```diff
@@ -116,6 +116,16 @@ var (
 	ErrOAuthError         = fmt.Errorf("%w: oauth error", ErrRESTError)
 )
 
+// Controls which snapshots are included in a loadTable response.
```
zeroshade: Go doc comments on exported identifiers should start with the identifier name, e.g. `// SnapshotMode controls which snapshots are included in a loadTable response.` — same for `SnapshotModeAll`, `SnapshotModeRefs`, and `LoadTableWithSnapshotMode` below. (Noting the `style(rest): remove identifier names from comments` commit went the opposite direction of the convention.)

catalog/rest/rest_test.go
```diff
@@ -1571,6 +1571,73 @@ func (r *RestCatalogSuite) TestLoadTable200() {
 	}))
 }
 
+func (r *RestCatalogSuite) TestLoadTableWithSnapshotModeRefs() {
```
zeroshade, line 1601: Two cheap additions worth making: (1) assert `snapshots` is *absent* from the query in `TestLoadTable200` — that guards the "LoadTable is unchanged" claim in the PR description; (2) one case for `SnapshotModeAll`.

zeroshade (commented):
Thanks — implementation is clean and the test is solid. Two things before this can move forward:

1. +1 to the open question from `alessandro-nori` on per-request vs. catalog-level configuration. Java configures `snapshot-loading-mode` once at catalog initialization, and this repo followed a similar context/config pattern for `pageSize`. I'd like to hear your take before we commit to the `LoadTableWithSnapshotMode` public API shape — a catalog property (possibly alongside this method) may serve the common case better.
2. A few smaller points inline: query-string merging in `do()`, Go doc-comment convention on the new exported identifiers, and two cheap test additions.

CI is green; happy to re-review once the API-shape question settles.

---

technicolorbeat:
> Thanks — implementation is clean and the test is solid. Two things before this can move forward:
> 
> 1. +1 to the open question from `alessandro-nori` on per-request vs. catalog-level configuration. Java configures `snapshot-loading-mode` once at catalog initialization, and this repo followed a similar context/config pattern for `pageSize`. I'd like to hear your take before we commit to the `LoadTableWithSnapshotMode` public API shape — a catalog property (possibly alongside this method) may serve the common case better.
> 2. A few smaller points inline: query-string merging in `do()`, Go doc-comment convention on the new exported identifiers, and two cheap test additions.
> 
> CI is green; happy to re-review once the API-shape question settles.
> 
> _This review was drafted with an AI-assisted tool and may contain mistakes; an Apache Iceberg Go maintainer has reviewed and confirmed the submission. See the [contributing docs](https://github.com/apache/iceberg-go/blob/main/CONTRIBUTING.md) for what the project considers a maintainer review._


thanks for taking a look! thinking about this more, I agree the catalog-level implementation is a better than a per-request implementation since most of our users who want refs want it for all loads on that catalog and this is more ergonomic in that sense. I'll update this and address docu / test comments 🙏

technicolorbeat:
> When a table is loaded with snapshotMode=refs and the clients tries to load previous snapshots, there should be a way to lazily load the rest of the snapshots and let the client knows that not all snapshots were loaded with the first loadTable request. See `snapshotsSupplier` in the Java implementation

@zeroshade I think it would be valuable to implement lazy loading for rest of the snapshot as a followup, wdyt?

thanks for the reviews @alessandro-nori  and @tanmayrauth!

alexandre-normand:
> Thanks — implementation is clean and the test is solid. Two things before this can move forward:
> 
> 1. +1 to the open question from `alessandro-nori` on per-request vs. catalog-level configuration. Java configures `snapshot-loading-mode` once at catalog initialization, and this repo followed a similar context/config pattern for `pageSize`. I'd like to hear your take before we commit to the `LoadTableWithSnapshotMode` public API shape — a catalog property (possibly alongside this method) may serve the common case better.

It's probably too late to chime in but here's another alternative to having a new public API or passing things through the context (which is not ideal for discovery of the feature). 

The idea is just functional options (more go idiomatic and easier for discovery) and this is kind of a high-level of what this could look like. Note that existing callers would just have the same existing/default behavior and users that need to set a different snapshot mode would just need 

```go
type LoadTableOption func(c *loadTableCall)

type loadTableCall struct {
    snapshotMode SnapshotMode
}

func WithLoadTableSnapshotMode(m SnapshotMode) LoadTableOption {
    return func(c *loadTableCall) {
        c.snapshotMode = m
    }
}

func (r *Catalog) LoadTable(ctx context.Context, identifier table.Identifier, opts ...LoadTableOption) (*table.Table, error) {
...
}  
```

zeroshade (approved):
This landed exactly where the design discussion pointed: `snapshot-loading-mode` as a catalog property read once at initialization, matching Java's `RESTSessionCatalog`, with the per-request method dropped so the public surface is just the `SnapshotMode` type and constants. The query-merge in `do()` no longer discards base-URI params, the doc comments follow convention (the time-travel caveat on `SnapshotModeRefs` is a nice touch), and the default-absence assertion in `TestLoadTable200` guards the no-behavior-change claim. Thanks for the thorough rework — `alessandro-nori`, I think this resolves your thread.

Optional polish, not blocking: an unrecognized property value passes through to the server as-is (Java parses to an enum and fails client-side); a one-line validation at `fetchConfig` would fail faster, but the server 400 is acceptable.

---

PR #1751, opened 2026-08-10
Add Transaction.RemoveProperties

alessandro-nori:
## Key changes
- Adds `Transaction.RemoveProperties(keys []string) error`, mirroring the existing `SetProperties`.

## Motivation
`Transaction` currently supports setting table properties but has no way to remove them.

### DIFF
--- a/table/table_test.go
+++ b/table/table_test.go
@@ -3218,6 +3218,35 @@ func (t *TableTestSuite) TestRefresh() {
 	t.Equal(originalSpec, tbl.Spec())
 }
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Clean addition — mirrors `SetProperties` exactly (including the empty-input no-op and `txnMeta` guard), reuses the existing `removePropertiesUpdate`, and the `keys []string` signature matches `MetadataBuilder.RemoveProperties` and the catalog-level removals APIs. The commit round-trip test covering both the removal and the ignored-absent-key case is just right for the scope. Thanks!

---

PR #1752, opened 2026-08-10
perf(table): index equality deletes by partition and sequence

fallintoplace:
## Summary

Equality delete matching currently scans every equality delete entry for every data file. This makes scan planning grow with the product of data and delete file counts.

This change builds a small index before creating scan tasks:

* unpartitioned equality deletes are kept in a global sequence-sorted list
* partitioned equality deletes are grouped by partition spec and normalized partition value
* each data file uses binary search to skip delete files with non-applicable sequence numbers

Partitioned matching now follows Java's DeleteFileIndex: deletes apply within the same spec and partition, while unpartitioned and void-only deletes remain global. The common paths with no equality deletes or only global deletes avoid partition-key construction.

The existing exported compaction helpers keep their original signatures as conservative compatibility wrappers. New WithSpecs variants provide exact spec-aware cleanup, and the CLI and rewrite flow use those variants.

## Benchmarks

The main benchmark includes index construction and matching 10,000 data files. Each data file matches at most the newest delete so output materialization stays bounded.

| Delete files | Partitions | Before | After | Speedup |
|---:|---:|---:|---:|---:|
| 100 | 1 | 3.17 ms | 1.47 ms | 2.2x |
| 100 | 10 | 2.89 ms | 1.09 ms | 2.6x |
| 100 | 100 | 2.86 ms | 0.95 ms | 3.0x |
| 1,000 | 1 | 27.16 ms | 1.51 ms | 18.0x |
| 1,000 | 10 | 27.14 ms | 1.21 ms | 22.4x |
| 1,000 | 100 | 27.29 ms | 1.12 ms | 24.3x |
| 10,000 | 1 | 269.21 ms | 2.46 ms | 109.3x |
| 10,000 | 10 | 265.48 ms | 2.42 ms | 109.8x |
| 10,000 | 100 | 265.76 ms | 2.34 ms | 113.6x |

There are also focused benchmark cases for old data files that materialize many matching deletes and for building the index when there are no data files.

Benchmark command:

    go test ./table -run '^$' -bench '^BenchmarkEqualityDeleteIndex' -benchtime=200ms -count=3

## Testing

* full repository test suite
* race tests for table and table/compaction
* golangci-lint
* end-to-end partition evolution test that writes data and a partitioned delete under different specs with the same tuple, then verifies only the global delete is attached and applied
* compatibility coverage for the original conservative compaction APIs, including void-only delete safety


zeroshade (approved):
This is both the right optimization and a quiet correctness fix: grouping by (spec ID, normalized partition tuple) with sequence-sorted lists and binary search matches Java's `DeleteFileIndex` semantics — including cross-spec isolation the old `DeepEqual` tuple comparison didn't provide, void-only specs correctly treated as global, and the strict `seq >` contract preserved and documented. The key normalization (NaN folding, float-bit widening, single-field no-allocation fast path, encoded-tuple fallback) is careful, and the 100×-at-10k-deletes benchmark table with bounded-materialization cases is exactly how to justify it. The conservative compatibility wrappers with spec-aware `WithSpecs` variants for the exact paths are a good API story.

One mechanical note before merge: today's #1683 landed a `reflect` use in `scanner.go`, and this branch removes that import — please rebase onto current `main` (keeping the import) and let CI re-run; the semantic surfaces don't overlap otherwise.

---

PR #1754, opened 2026-08-10
perf(table): reuse equality-delete key sets across scan tasks

fallintoplace:
## Summary

Equality delete files are already deduplicated and read once, but their decoded key maps are copied into a fresh union for every scan task. When many tasks share the same delete files, setup time and memory therefore grow with the number of tasks rather than the number of distinct delete-file combinations.

This change reuses the decoded key sets during per-task assembly:

* a group backed by one delete file points directly at that file's key set
* multi-file groups are canonicalized by delete-file ID and identical combinations share one merged set
* duplicate references to the same delete file are removed before building a union
* delete files with different equality field IDs remain separate, as before

The branch is based on the current equality-delete reader. Name mapping, structural field-ID resolution, nested fields, ambiguous-column validation, and per-data-file field resolution all happen unchanged. Sharing begins only after each delete file has been decoded.

The shared sets are immutable during record filtering, so the scan hot path still performs one lookup per equality-field group. It does not add a chain of underlying maps or change row filtering behavior.

This is independent of #1752. That PR speeds up deciding which equality delete files apply during planning; this change avoids repeatedly materializing their decoded key unions before reading rows.

## Benchmarks

Representative medians from five runs with three iterations each on an Apple M1 Pro:

| Workload | Copy per task | Shared | Speedup | Memory before | Memory after |
|---|---:|---:|---:|---:|---:|
| 1 file, 1 combination | 68.27 ms | 0.31 ms | 219x | 109.3 MB | 0.22 MB |
| 4 files, 1 combination | 71.98 ms | 0.87 ms | 83x | 109.4 MB | 0.49 MB |
| 4 files, 100 combinations | 71.24 ms | 8.57 ms | 8.3x | 109.4 MB | 11.30 MB |
| 2 files, 1,000 unique combinations | 10.97 ms | 10.78 ms | 1.0x | 13.8 MB | 13.97 MB |

Each workload has 1,000 tasks. The shared cases model global or partition-level deletes used by multiple data files. The fully unique case is included as a control and remains effectively neutral.

Command:

    go test ./table -run '^$' -bench '^BenchmarkEqualityDeleteSetAssembly$' -benchmem -benchtime=3x -count=5

## Testing

* full repository test suite
* focused equality-delete tests under the race detector
* golangci-lint
* scan-level regression with two data tasks sharing the same two-file delete union and four reader workers
* helper coverage for singleton sharing, order-independent combinations, duplicate references, distinct combinations, and separate field-ID groups

### DIFF
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -19,10 +19,13 @@ package table
 
 import (
 	"bytes"
+	"cmp"
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"math"
+	"slices"
 	"unsafe"
 
 	"github.com/apache/arrow-go/v18/arrow"
@@ -42,7 +45,8 @@ var ErrAmbiguousEqualityColumn = errors.New("equality delete column is ambiguous
 // equalityDeleteSet holds the set of delete keys and the column names
 // used to look them up in data records. Each set corresponds to one
 // group of equality field IDs — delete files with different field IDs
-// produce separate sets.
+// produce separate sets. The set is immutable after construction so it
+// can be shared by tasks with the same delete files.
 type equalityDeleteSet struct {
 	keys     set[string]
 	fieldIDs []int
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
Verified the sharing is safe end-to-end: `readAllEqualityDeleteFiles` fully materializes all sets before `recordBatchesFromTasksAndDeletes` spawns workers, and the only downstream access is the read-only lookup in `processEqualityDeletesColumnarForFile` (table/arrow_scanner.go / equality_delete_reader.go:793), so concurrent map reads on shared sets are fine — confirmed by the new scan-level test passing under `-race` locally. A couple of things I checked that hold up:

- Taking `fieldIDs`/`colNames` from `files[0]` after canonical sort is equivalent to the old first-seen-in-task behavior, since `colNames` come from `tableSchema.FindColumnName(fid)` and are deterministic per field ID regardless of which file resolved them.
- The empty-set skip (`len(deleteSet.keys) > 0`) matches prior behavior for both the singleton and merged paths.
- The `sharedSets` key of sorted unique file IDs is sound: IDs are unique per call (assigned from `len(uniqueDeletes)`), a file belongs to exactly one field group, and the cache doesn't outlive the call.

One minor note, non-blocking: `buildEqualityDeleteSetsPerTaskWithCopies` in the bench file is a ~50-line copy of the pre-PR algorithm kept only as a baseline comparator. That's useful now but will rot; I'd be fine dropping the copy-per-task sub-benchmarks in a follow-up once the numbers are recorded here. Nice test coverage on order-independence and duplicate file references.

PR #1755, opened 2026-08-10
perf(table): batch concurrent manifest entry collection

fallintoplace:
## Summary

- collect each manifest batch under one mutex acquisition
- reduce synchronization from once per accepted entry to once per manifest
- preserve the existing data, delete, equality delete, and deletion vector classification

## Benchmarks

`go test ./table -run "^$" -bench "^BenchmarkManifestEntryCollection$" -benchmem -benchtime=2s -count=6 -cpu=4`

The benchmark now models manifest count and scanner concurrency separately. `-cpu=4` sets GOMAXPROCS, while `concurrency=4` applies the same active manifest limit as the scanner.

With 1,000 entries per manifest on an Apple M1 Pro:

| Workload | Manifest concurrency | Before | After | Speedup |
| 64 data manifests | 4 | 7.484 ms | 2.761 ms | 2.7x |
| 64 delete manifests | 4 | 7.616 ms | 1.803 ms | 4.2x |

For example, the 64-delete-manifest workload is about 4x faster with the production-style concurrency limit. `B/op` and `allocs/op` are unchanged in both cases.

The benchmark also covers larger manifests:

- 10,000 entries per manifest at 8, 16, and 64 manifests, with concurrency 1, 4, and 16
- 100,000 entries per manifest at 8 and 16 manifests, with concurrency 4 and 16
- 38 total subbenchmarks across data and delete workloads

A one-iteration run of all 38 subbenchmarks completed successfully on an Apple M1 Pro.

## Validation

- `go test ./...`
- `go test ./table -count=1`
- `go test -race ./table -run "TestManifestEntries_(DVClassification|ConcurrentMerge)$" -count=1`
- `go test ./table -run "^$" -bench "^BenchmarkManifestEntryCollection$" -benchmem -benchtime=1x -count=1`

### DIFF
--- a/table/dv_scan_planning_test.go
+++ b/table/dv_scan_planning_test.go
@@ -152,6 +152,20 @@ func TestManifestEntries_ConcurrentMerge(t *testing.T) {
 			path:        "s3://bucket/data/pos-del-001.parquet",
 			contentType: iceberg.EntryContentPosDeletes,
 		}),
+		iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+			path:        "s3://bucket/data/eq-del-001.parquet",
+			contentType: iceberg.EntryContentEqDeletes,
+		}),
+		iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &dvMockDataFile{
+			mockDataFile: mockDataFile{
+				path:        "s3://bucket/data/dv-001.puffin",
+				contentType: iceberg.EntryContentPosDeletes,
+				format:      iceberg.PuffinFile,
+			},
+			referencedDataFile: strPtr("s3://bucket/data/data-001.parquet"),
+			contentOffset:      int64Ptr(0),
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM. The consolidation of the four `add*` methods into a single `merge` in table/scanner.go is behavior-preserving: the classification switch (data / pos-delete / DV / eq-delete) and the `ErrInvalidMetadata` error are byte-for-byte the same as the old inline loop in `collectManifestEntriesWithSchema`, and the errgroup still aborts the scan on error, so the partial-append-before-error behavior is unobservable just as before. Verified locally on the PR head: `go build ./...`, `go test ./table -count=1`, `go test -race` on the DVClassification/ConcurrentMerge tests, and a one-shot run of the benchmark all pass. The new concurrent-merge race test is a nice addition.

One minor observation, not blocking: `merge` now runs `ContentType()`/`IsDeletionVector()` under the mutex, serializing classification across manifests. Those are cheap field accesses so it doesn't matter in practice (your benchmarks confirm the net win), but if classification ever gets heavier it would be worth pre-bucketing outside the lock.

PR #1756, opened 2026-08-11
fix(hive): use 1.5 lock-check backoff scale to match Java

### LINKED ISSUE #1715: Hive lock retry backoff scale diverges from the Java implementation (2.0 vs 1.5)
Raised by @laskoviymishka during review of #1640 and deferred as out of scope there. Checked against the Java source before filing this update.

`catalog/hive/lock.go`'s `calculateBackoff` doubles the wait on each attempt:

```go
return minWait << attempt
```

The Java equivalent for **this specific loop** is [`MetastoreLock.java:206`](https://github.com/apache/iceberg/blob/main/hive-metastore/src/main/java/org/apache/iceberg/hive/MetastoreLock.java#L206), which passes a scale factor of **1.5**:

```java
.exponentialBackoff(lockCheckMinWaitTime, lockCheckMaxWaitTime, lockAcquireTimeout, 1.5)
```

Worth being precise, because the number depends on the call site: `Tasks.Builder`'s default `scaleFactor` is 2.0 ([`Tasks.java:89`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/Tasks.java#L89)), and the lock **creation** loop at [`MetastoreLock.java:295`](https://github.com/apache/iceberg/blob/main/hive-metastore/src/main/java/org/apache/iceberg/hive/MetastoreLock.java#L295) does use 2.0. It is only the lock-**check** path — the one `acquireLocks` mirrors — that uses 1.5.

So a Go and a Java client waiting on the same contended table walk different schedules from identical configuration.

**A second, larger divergence in the same call site.** Java retries `Integer.MAX_VALUE - 100` times and is bounded by `lockAcquireTimeout`, i.e. it keeps polling until a deadline. Go bounds by *attempt count* instead — `LockRetries`, default 4 — so with the default 100ms minimum it gives up after roughly 1.5s of waiting regardless of how long the lock holder needs. These are different failure semantics, not just different tuning, and arguably matter more than the scale factor.

Filing rather than patching, because each is a behaviour change for existing deployments:

- Matching 1.5 makes Go clients retry more often and reach the cap later.
- Moving to a deadline bound changes when `ErrLockAcquisitionFailed` surfaces.
- Keeping either as-is is defensible, but the divergence should be intentional and documented rather than incidental.

Related: the `[d, 2d]` jitter window in `applyJitter` is wider than Java's, which adds up to 10% of the current delay ([`Tasks.java:456`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/Tasks.java#L456)). Both are additive, so the direction matches; only the width differs. That one is now called out in the doc comment in #1714.

dgvj-work:
## Summary
- Align Hive lock-check `calculateBackoff` with the Java `MetastoreLock` checkLock path, which uses `Tasks.exponentialBackoff(..., 1.5)` rather than the previous Go doubling schedule (`minWait << attempt`).
- Keep the attempt-count bound unchanged; this PR does not move to Java's deadline-based retry semantics.
- Update the jitter floor replay to walk the same 1.5× sequence so the last uncapped interval stays correct after the scale change.

Fixes #1715

## Testing
- `go test ./catalog/hive/ -count=1`
- `gofmt` clean on touched Go files

### DIFF
--- a/catalog/hive/lock.go
+++ b/catalog/hive/lock.go
@@ -158,11 +158,21 @@ func calculateBackoff(attempt int, minWait, maxWait time.Duration) time.Duration
 	if attempt <= 0 || minWait <= 0 {
 		return minWait
 	}
-	if attempt > 62 || minWait > maxWait>>attempt {
-		return maxWait
+
+	// Match Java's lock-check schedule: minWait * 1.5^attempt, capped at maxWait.
+	// (MetastoreLock uses scaleFactor 1.5 for checkLock retries; the lock-create
+	// path still uses 2.0, and is not mirrored here.)
+	wait := float64(minWait)
+	for i := 0; i < attempt; i++ {
+		next := wait * lockCheckBackoffScale
+		// next <= wait catches +Inf / non-advancing overflow on extreme inputs.
+		if next >= float64(maxWait) || next <= wait {
+			return maxWait
+		}
+		wait = next
 	}
 
-	return minWait << attempt
+	return time.Duration(wait)
 }
 
 // applyJitter spreads a backoff interval so clients contending for the same lock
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Careful alignment work: the 1.5× schedule matches Java's `MetastoreLock` check-lock path (leaving the 2.0 create path alone, as noted in the comment), the non-advancing/overflow guard is a tidy replacement for the shift-count check, and — most importantly — the jitter-floor replay was updated *with* the scale rather than left to drift: the recomputed worked examples and the monotonicity-across-the-cap test at the exact 759.375ms boundary keep the least-exercised branch honest. Exactly how a subtle-constants change should be done.

---

PR #1758, opened 2026-08-11
feat(dv): add SetRange and RunLengthEncode to RoaringPositionBitmap

### LINKED ISSUE #1757: RoaringPositionBitmap: range mutator and run-length encoding at DV serialization (Java parity)

`table/dv.RoaringPositionBitmap`'s only mutator is `Set(pos uint64)`, and `SerializeDV` writes containers as-is without run-length encoding. For writers whose deletion sets are range-shaped — deletes at block or row-group granularity, where a deletion vector is a handful of contiguous runs — this has two costs:

1. **Construction** is one `Set` call per deleted position: a multi-million-row dead range costs millions of calls where one range-add would do.
2. **Size**: incremental `Set` produces array/dense bitmap containers and nothing ever re-encodes them as runs, so a run-shaped DV serializes dramatically larger than the portable roaring format allows. Measured: 100,000 contiguous positions produce an 8,235-byte DV envelope; run-encoded, the same set is 49 bytes.

Java Iceberg has both affordances and applies them by default: `RoaringPositionBitmap` declares `setRange(posStartInclusive, posEndExclusive)` and `runLengthEncode()`, and `BitmapPositionDeleteIndex.serialize` calls `runLengthEncode()` before serializing. The Go port has neither, so Go-written DVs are byte-inflated relative to Java's for the same positions (they remain spec-valid — run containers are flagged in the portable format's cookie header — just larger).

Proposal, mirroring Java's names and layering:

- `SetRange(startInclusive, endExclusive uint64)` — bucket-split across the high-32-bit keys, one roaring `AddRange` per bucket.
- `RunLengthEncode()` — re-encode each bucket's containers as runs where smaller.
- `SerializeDV` calls `RunLengthEncode()` before emitting, matching `BitmapPositionDeleteIndex.serialize`.

The existing Java golden fixtures (including `all-container-types-position-index.bin`, which Java builds from range deletes and serializes run-encoded) continue to match byte-for-byte with this change — the all-container-types fixture only matches *because* of it.

I'd like to work on this — implementation and tests are ready to submit.

dimakuz:
Fixes #1757.

`Set` was the only mutator on `RoaringPositionBitmap`, so range-shaped deletes (dead blocks or row groups) cost one call per position and serialized as array or dense bitmap containers. This adds, mirroring Java's `RoaringPositionBitmap`:

- **`SetRange(startInclusive, endExclusive uint64)`** — marks `[start, end)` with one roaring `AddRange` per 2^32-position bucket (Java: `setRange`).
- **`RunLengthEncode()`** — re-encodes each bucket's containers as runs where smaller (Java: `runLengthEncode`).
- **`SerializeDV`** now run-length encodes before emitting bytes, matching `BitmapPositionDeleteIndex.serialize`. For 100,000 contiguous positions the DV envelope drops from 8,235 to 49 bytes; output remains spec-valid (run containers are flagged in the portable format's cookie header).

Two deliberate deviations from Java, documented in the code: an inverted range is a no-op rather than an exception (no mutator on this type panics or errors, and uint64 already excludes Java's negative-position case), and `RunLengthEncode` returns nothing (roaring's `RunOptimize` doesn't report whether it changed anything).

## Testing

- New unit tests: range within/spanning/ending-on bucket boundaries, empty and inverted ranges, `SetRange`≡`Set`-loop equivalence, serialized-size reduction with round-trip membership/cardinality equality.
- New cross-client test rebuilds Java's `all-container-types-position-index.bin` fixture (132,561 positions across array/run/dense containers in two buckets) via `SetRange`/`Set` and asserts the Go envelope is byte-identical to Java's 94 bytes — it passes only with serialize-time run encoding, pinning the parity decision.
- Existing Java golden fixtures still match byte-for-byte; the comment on `TestCrossClientGoSerializeMatchesJavaSingleArrayContainer` that described the old serialize-time divergence from Java is corrected to match.
- `go test ./...`, `gofmt`, `golangci-lint run` all clean locally.

### DIFF
--- a/table/dv/deletion_vector.go
+++ b/table/dv/deletion_vector.go
@@ -114,7 +114,12 @@ func DeserializeDV(data []byte, expectedCardinality int64) (*RoaringPositionBitm
 //   - Magic  (4 bytes, little-endian): DVMagicNumber
 //   - Bitmap (variable): roaring bitmap in Iceberg portable format
 //   - CRC-32 (4 bytes, big-endian): checksum over magic + bitmap
+//
+// The bitmap is run-length encoded in place before it is written, matching
+// Java's BitmapPositionDeleteIndex.serialize.
 func SerializeDV(bitmap *RoaringPositionBitmap) ([]byte, error) {
+	bitmap.RunLengthEncode()
+
 	var bitmapBuf bytes.Buffer
 	if err := bitmap.Serialize(&bitmapBuf); err != nil {
 		return nil, fmt.Errorf("serialize roaring bitmap: %w", err)
--- a/table/dv/deletion_vector_test.go
+++ b/table/dv/deletion_vector_test.go
@@ -220,6 +220,69 @@ type failingOpenFS struct {
 func (f failingOpenFS) Open(string) (iceio.File, error) { return nil, f.err }
 func (f failingOpenFS) Remove(string) error             { return nil }
 
+// Why: SerializeDV run-length encodes before emitting bytes, so a caller that
+// records contiguous deletes one position at a time still gets a compact blob
+// rather than dense containers. Without that step the envelope carries a full
+// 8 KiB word block per container.
+// Condition: 100,000 contiguous positions recorded with individual Set calls,
+// measured against what the bare bitmap encoder would have produced.
+// Assertion: the envelope is two orders of magnitude smaller and still decodes
+// to the same positions and cardinality.
+func TestSerializeDVRunLengthEncodesContiguousPositions(t *testing.T) {
+	const positions = 100_000
+
+	bm := NewRoaringPositionBitmap()
+	for pos := uint64(0); pos < positions; pos++ {
+		bm.Set(pos)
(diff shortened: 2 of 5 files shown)


tanmayrauth (approved):
LGTM!

zeroshade (approved):
Excellent first contribution. The bucket arithmetic is exactly right — keying off the last *included* position so boundary-ending ranges don't allocate an empty bucket above, with partial/full/partial `AddRange` splits across buckets — and wiring `RunLengthEncode` into `SerializeDV` mirrors Java's `BitmapPositionDeleteIndex.serialize` for a dramatic size win on contiguous deletes. Both deviations from Java are documented with sound Go rationale, and the cross-client test that rebuilds the Java fixture through the new API and asserts byte-identical output is the gold standard for this kind of change. The `Set`≡`SetRange` equivalence and boundary matrices round it out. Welcome, and thanks!

---

Issue #1691, opened 2026-08-06
Catalog Conformance Testing Suite


rambleraptor:
### Feature Request / Improvement

Right now, it appears that every catalog has its own test suite. A shared conformance test suite allows all catalogs to run against a shared suite to ensure identical functionality.

iceberg does this through the CatalogTests and pyiceberg has a similar suite.

PR #1761, opened 2026-08-11
test(catalog): add namespace lifecycle conformance tests

laskoviymishka:
Follows up on #1694, which stood up the shared catalog conformance suite as the
first part of #1691. That one landed the harness plus a few table tests and left
room for more, so this adds the namespace side: create, list, and drop, along
with the error cases (create-already-exists, drop-missing, drop-not-empty).

Adding the not-empty-drop test surfaced a real bug in the REST catalog: its
DropNamespace mapped only 404, so a 409 on a non-empty namespace came back as a
generic REST error instead of catalog.ErrNamespaceNotEmpty. Fixed that here (with
a unit test), which is exactly the kind of divergence the shared suite is meant
to catch.

The suite runs against the SQL and REST catalogs today. Each test takes a fresh
catalog and unique identifiers so it stays isolated on shared-server backends too.

I kept this focused on the structural lifecycle. Namespace properties
(LoadNamespaceProperties / UpdateNamespaceProperties) and the Config capability
flags to gate them per backend are the obvious next step, and I'll send those in
a separate PR so this one stays easy to review.

### DIFF
--- a/catalog/catalogtest/catalogtest.go
+++ b/catalog/catalogtest/catalogtest.go
@@ -94,6 +94,13 @@ func RunCatalogTests(t *testing.T, cfg Config) {
 	t.Run("BasicCreateTableThatAlreadyExists", func(t *testing.T) { testBasicCreateTableThatAlreadyExists(t, cfg) })
 	t.Run("LoadMissingTable", func(t *testing.T) { testLoadMissingTable(t, cfg) })
 	t.Run("LoadTableWithNonExistingNamespace", func(t *testing.T) { testLoadTableWithNonExistingNamespace(t, cfg) })
+
+	t.Run("CreateNamespace", func(t *testing.T) { testCreateNamespace(t, cfg) })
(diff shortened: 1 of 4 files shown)


rambleraptor:
I was excited for a second that the test suite found an actual bug. Looks like we just need a better test regex.

rambleraptor:
Wait, it did find a bug! That's really exciting. Great work!

zeroshade (approved):
The namespace lifecycle slice is exactly what the shared suite is for — and it immediately paid for itself by catching the REST 409 mapping gap, now fixed with the right sentinel and pinned by both the conformance test and a REST unit test. The suite hygiene is model quality: fresh catalog and unique identifiers per test for shared-backend isolation, state-intact assertions after every rejected operation, and the documented Go-vs-Java divergence on drop-missing so nobody "fixes" it later. Deferring properties + capability flags to the next slice keeps this cleanly reviewable.

---

PR #1762, opened 2026-08-11
fix(manifest): infer manifest list version from the embedded writer schema

### LINKED ISSUE #1760: `ReadManifestList` silently misreads v2/v3 manifest lists as v1 when the optional `format-version` metadata key is absent
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

`ReadManifestList` picks its reader schema from the `format-version` key in the Avro file metadata, and assumes v1 when the key is missing. As the doc comment on the function itself notes, manifest lists aren't required to carry that key — but the v1 fallback means a perfectly valid v2/v3 list from a writer that skips it gets decoded through the v1 schema, and the fields v1 doesn't have are silently zeroed:

- `content` reads as `data`, so delete manifests look like data manifests
- `sequence-number` / `min-sequence-number` read as 0
- v3 `first-row-id` is lost

No error, no warning — Avro schema resolution just drops the extra writer fields.

I hit this with real files: DuckDB's iceberg extension (v1.5.x) writes manifest lists without the key, so reading any DuckDB-written v3 table with iceberg-go makes its deletion vectors invisible and zeroes every sequence number. Java and PyIceberg read the same files fine since they project from the embedded writer schema.

Observed on v0.6.0.

## Reproduction

Round-trip through iceberg-go itself: write a v2 list, drop the optional key from the header (data blocks untouched), read it back.

```go
package main

import (
	"bytes"
	"fmt"

)

func main() {
	// A v2 manifest list holding one delete manifest with sequence number 5.
	mf := iceberg.NewManifestFile(2, "s3://bucket/metadata/deletes-m0.avro", 100, 0, 42).
		Content(iceberg.ManifestContentDeletes).
		SequenceNum(5, 3).
		Build()

	var buf bytes.Buffer
	seq := int64(5)
	if err := iceberg.WriteManifestList(2, &buf, 42, nil, &seq, 0, []iceberg.ManifestFile{mf}); err != nil {
		panic(err)
	}

	stripped := stripMetadataKey(buf.Bytes(), "format-version")

	got, err := iceberg.ReadManifestList(bytes.NewReader(stripped))
	if err != nil {
		panic(err)
	}
	fmt.Printf("content             = %v  (want deletes)\n", got[0].ManifestContent())
	fmt.Printf("sequence-number     = %d  (want 5)\n", got[0].SequenceNum())
	fmt.Printf("min-sequence-number = %d  (want 3)\n", got[0].MinSequenceNum())
}

// stripMetadataKey re-encodes an Avro OCF header without one metadata key;
// the sync marker and data blocks are copied through untouched.
func stripMetadataKey(data []byte, key string) []byte {
	off := 4 // past "Obj\x01"
	meta := map[string][]byte{}
	for {
		n, next := readLong(data, off)
		if n == 0 {
			off++ // consume terminator

			break
		}
		if n < 0 {
			n = -n
			_, next = readLong(data, next) // block byte size
		}
		off = next
		for i := int64(0); i < n; i++ {
			var k, v []byte
			k, off = readBytes(data, off)
			v, off = readBytes(data, off)
			meta[string(k)] = v
		}
	}
	delete(meta, key)

	out := append([]byte(nil), data[:4]...)
	out = appendLong(out, int64(len(meta)))
	for k, v := range meta {
		out = appendLong(out, int64(len(k)))
		out = append(out, k...)
		out = appendLong(out, int64(len(v)))
		out = append(out, v...)
	}
	out = append(out, 0)

	return append(out, data[off:]...)
}

func readLong(data []byte, off int) (int64, int) {
	var u uint64
	var shift uint
	for {
		b := data[off]
		off++
		u |= uint64(b&0x7f) << shift
		if b&0x80 == 0 {
			break
		}
		shift += 7
	}

	return int64(u>>1) ^ -int64(u&1), off
}

func readBytes(data []byte, off int) ([]byte, int) {
	n, off := readLong(data, off)

	return data[off : off+int(n)], off + int(n)
}

func appendLong(dst []byte, v int64) []byte {
	u := uint64(v<<1) ^ uint64(v>>63)
	for u >= 0x80 {
		dst = append(dst, byte(u)|0x80)
		u >>= 7
	}

	return append(dst, byte(u))
}
```

Output on v0.6.0:

```
content             = data  (want deletes)
sequence-number     = 0  (want 5)
min-sequence-number = 0  (want 3)
```

Or with real files: create a v3 table with DuckDB's iceberg extension and read `metadata/snap-*.avro` with `ReadManifestList` — same result.

## Suggested fix

The version is recoverable from the embedded writer schema when the key is absent: `sequence_number`/`content` fields (ids 515/517) imply v2+, `first_row_id` (id 520) implies v3. Even an error would beat silently returning wrong values, since anything that rewrites metadata based on them (compaction, snapshot management) writes the corruption back out.

Happy to send a PR if the schema-inference approach sounds right to you.

ryanworl:
Fixes #1760

## What

`ReadManifestList` selected its Avro reader schema solely from the `format-version` file-metadata key, assuming v1 when the key is absent. Per the spec that key is required only for [manifest files](https://iceberg.apache.org/spec/#manifests); for [manifest lists](https://iceberg.apache.org/spec/#manifest-lists) it is optional, and real writers (DuckDB's iceberg extension ≤ 1.5.4) omit it. Reading a keyless v2/v3 list through the v1 reader schema made Avro schema resolution silently drop every writer field the v1 schema doesn't declare: `content` (field-id 517) read as 0 so delete manifests became invisible as deletes, `sequence_number` (515) / `min_sequence_number` (516) read as 0, and v3 `first_row_id` (520) was lost — with no error. Any read-modify-rewrite path (compaction, snapshot expiry, manifest rewriting) would persist that corruption.

## How

Field-ID inference from the embedded writer schema, feeding the existing v1/v2+ decode paths:

- Inside the reader-schema callback, the writer schema's top-level fields are inspected: `content` (517) or `sequence_number` (515) imply v2+, `first_row_id` (520) implies v3, neither means v1. Fields are discriminated primarily by their `field-id` prop — the stable contract per the spec (DuckDB embeds them) — with the spec field names as a fallback for writers that don't annotate IDs.
- Key absent: the inferred version is used. A genuine keyless v1 list still reads as v1 (existing behavior, now pinned by a test).
- Key present and consistent: unchanged behavior.
- Key present but claiming v1 for a schema that carries v2+ fields: an error is returned instead of silently dropping writer fields.

This matches how Java's `ManifestLists.read` and PyIceberg handle these files: they never consult the key and let field-ID resolution against the embedded writer schema decide what is present. The alternative design discussed in the issue — a Java-style single superset reader schema (v2+/v3 fields optional, defaulted in code) with no version branch at all — would also work and is arguably closer to Java; I went with inference as the smaller diff that feeds the existing v1/v2+ code paths. Happy to rework toward the superset approach if maintainers prefer it.

## Tests

- Key-stripped v2 and v3 round-trips: written with `WriteManifestList` containing a delete-content manifest with non-zero sequence numbers (and an assigned `first_row_id` for v3), `format-version` stripped from the OCF header (data blocks byte-identical), read back asserting `content`, `sequence_number`, `min_sequence_number`, and `first_row_id` survive.
- Keyless v1 list still decodes as v1, identical to the same list with the key.
- Key claiming v1 over a v2/v3 schema returns an error.
- A real DuckDB-written fixture (`testdata/duckdb_v3_manifest_list.avro`, 1.7 KB): produced by DuckDB v1.5.4's iceberg extension against a REST catalog — `format-version = '3'` table, one INSERT, one DELETE — and checked in. Under the previous fallback its delete manifest read as a data manifest with zeroed sequence numbers.

`go test ./...` and `golangci-lint run` pass; the fixture is listed in `dev/release/rat_exclude_files.txt` alongside the existing binary test files.


manifest.go
```diff
@@ -999,17 +999,66 @@ func ReadManifest(m ManifestFile, f io.Reader, discardDeleted bool) ([]ManifestE
...
+const (
+	fieldIDManifestSequenceNumber = 515
+	fieldIDManifestContent        = 517
+	fieldIDManifestFirstRowID     = 520
+)
+
+// schemaFieldID returns the Iceberg "field-id" property of an Avro schema
+// field, if present. Props hold decoded JSON, so numbers arrive as float64.
+func schemaFieldID(f avro.SchemaField) (int, bool) {
+	if id, ok := f.Props["field-id"].(float64); ok {
```
tanmayrauth: The name fallback here answers the open question from the issue (writers that don't annotate ids), and it's the right call to keep since DuckDB does embed  them, so it's just belt-and-suspenders. One robustness gap though: this only accepts `float64`, while the existing field-id reader in this same file  (`manifest.go:476`, `switch v := field.Props["field-id"].(type)`) handles both `int` and `float64`. For the read path here the writer schema is JSON-parsed from the OCF header, so numbers come through as `float64` and the DuckDB fixture proves it works — no live bug. But the two field-id readers now disagree on accepted types, and that sibling `int` case exists because an in-memory-built schema yields `int` props. If inference is ever handed such a schema, `schemaFieldID` returns false and silently degrades to name-matching, and a writer that annotates only field-ids (non-spec names) then misfires back to the v1 fallback this PR removes.  Worth mirroring the sibling type switch to accept `int` too — ideally one shared helper both sites call so they can't drift again.

manifest.go
```diff
@@ -1023,6 +1072,10 @@ func ReadManifestList(in io.Reader) ([]ManifestFile, error) {
 		}
 
 		if version == 1 {
```
tanmayrauth: `if version == 1 { if inferred > 1 { ... } }` - This only half-satisfies the "key present must agree with the schema, error beats zeroed values" rule from the issue. It catches a key claiming v1 over a v2+ schema, but not a key claiming a *lower v2+* version than the schema carries: a file with key="2" but a v3 schema (first_row_id present, inferred 3) reads with version=2 and no error. first_row_id is read into the struct, but Version() reports 2, so a read-modify-rewrite through WriteManifestList(2, …) uses the v2 writer schema and silently drops v3 row lineage — the same corruption this PR fixes, one version up. No real writer emits a key that undershoots its schema today (DuckDB omits the key; Java/Py write consistent keys), so it's adversarial rather than a field bug — but it's cheap to close the whole class: generalize the check to `if version < inferred { return error }`, and add a key=2/schema=v3 case to  TestReadManifestListKeyContradictsSchema.

tanmayrauth (commented):
The inference approach is fine as-is — no need to rework toward the Java superset-schema model; feeding the existing v1/v2+ paths is the smaller diff and field-id-primary discrimination is the right anchor. Two non-blocking notes inline, both about type/version edges rather than the main path.

manifest.go
ryanworl: Done in 88d1401 — schemaFieldID now mirrors the sibling switch (int and float64), and the partition-field reader at the other site calls the same helper so the two can't drift.

manifest.go
ryanworl: Done in 88d1401 — generalized to `version < inferred` with the error message naming both versions, and added the key=2/v3-schema case to TestReadManifestListKeyContradictsSchema.

zeroshade (approved):
Textbook fix for a nasty silent-corruption path: field-ID-based version inference honors the spec's actual contract (the metadata key is only required for manifests, not lists), the ID-first/name-fallback discrimination handles both annotating and non-annotating writers, and turning a key-vs-schema downgrade mismatch into a loud error is exactly the right fail-closed choice given rewrite paths would otherwise persist the loss. The committed DuckDB fixture makes the interop claim concrete. Thanks — this is a great first PR in the series.

---

PR #1763, opened 2026-08-11
fix(hive): align lock-check config keys and defaults with Java

### LINKED ISSUE #1716: Hive lock-check config keys do not match the Java implementation's names or defaults
Raised by @laskoviymishka during review of #1640 and deferred as out of scope there.

`catalog/hive/options.go` defines the lock-check tuning as:

| Key | Default |
|---|---|
| `lock-check-min-wait-time` | 100ms |
| `lock-check-max-wait-time` | 1 minute |
| `lock-check-retries` | 4 |

These do not line up with the Java implementation's property names or its default values, so the same table property set cannot be carried across the two clients, and an operator tuning one has to re-derive the settings for the other.

Two separate questions here, worth deciding together:

1. **Names** — should the Go client accept Java's property keys, either as the primary names or as aliases?
2. **Defaults** — should the values match Java's, or is divergence acceptable as long as it is documented?

Both are user-visible changes for anyone already setting these, which is why this is a tracking issue rather than a patch.

One adjacent robustness note from the same review: `options.go` parses these with `time.ParseDuration`, which accepts negative durations, and nothing rejects `min-wait-time >= max-wait-time`. Neither is currently dangerous — `applyJitter` guards both — but validating at parse time would make the contract clearer than relying on downstream defensiveness.

dgvj-work:
## Summary
- Accept Java's `iceberg.hive.lock-check-min-wait-ms` / `iceberg.hive.lock-check-max-wait-ms` as the primary Hive lock-check property keys (integer milliseconds).
- Keep the existing Go duration-string keys as aliases; Java keys win when both are set.
- Match Java defaults (`50ms` / `5s`) and ignore invalid values (non-positive waits/retries, or min ≥ max) at parse time.
- Document the remaining Go-specific `lock-check-retries` vs Java `iceberg.hive.lock-timeout-ms` gap as follow-up.

Fixes #1716

## Test plan
- [x] `go test ./catalog/hive/ -count=1`
- [x] Cover Java keys, legacy aliases, precedence, defaults, and invalid-value handling
- [ ] CI green

### DIFF
--- a/catalog/hive/lock.go
+++ b/catalog/hive/lock.go
@@ -222,11 +222,11 @@ func applyJitter(d, minWait, maxWait time.Duration) time.Duration {
 	}
 
 	// Everything below this point is unreachable under the default configuration.
-	// lock-check-min-wait-time=100ms, lock-check-max-wait-time=1m and
-	// lock-check-retries=4 top the sequence out at 337.5ms, so it never saturates and
-	// the branch above always wins. Getting here needs a lowered maximum or a raised
-	// retry count. Stated plainly because this is the most intricate part of the
-	// helper and the least exercised in practice.
+	// iceberg.hive.lock-check-min-wait-ms=50, iceberg.hive.lock-check-max-wait-ms=5000
+	// and lock-check-retries=4 top the sequence out at 168.75ms (50×1.5³), so it never
+	// saturates and the branch above always wins. Getting here needs a lowered
+	// maximum or a raised retry count. Stated plainly because this is the most
+	// intricate part of the helper and the least exercised in practice.
 	//
 	// Replay the 1.5× backoff sequence and keep the largest interval that still
 	// fitted under the cap. The guard on scheduled keeps a non-positive or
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
The alignment is exactly right: Java's millisecond keys as primary (with Java's 50ms/5s defaults verified against `HiveCatalog`), the Go duration keys as documented aliases with Java-wins precedence, and — the subtle part done well — the min/max window applied atomically only when consistent, so a bad value can't half-apply into an inverted window. Documenting `lock-check-retries` as Go-specific with the `iceberg.hive.lock-timeout-ms` deadline model as follow-up is the honest scoping.

One thing on rebase (your #1756 merged first): the `applyJitter` comment this PR updates still assumes the old doubling schedule — under the merged 1.5× scale the new 50ms default tops out at 168.75ms (50×1.5³), not 400ms, and the surrounding saturation narrative in the tests references the old defaults too. Please recompute those comment values when resolving the `lock.go`/`lock_test.go` conflicts, then CI re-run and this is good to merge.

---

dgvj-work:
Rebased on main after #1756 and recomputed the applyJitter / saturation comments for the 1.5× schedule with the new 50ms default (tops out at 168.75ms).

dgvj-work:
Rebased onto main after #1756 and recomputed the applyJitter/test narratives for the 1.5× schedule with the Java 50ms default (tops out at 168.75ms).

PR #1765, opened 2026-08-11
perf(table): reuse partition extraction plans

fallintoplace:
## What changed

- Build partition extraction descriptors once from the first record batch.
- Share the immutable plan across fanout workers.
- Reuse the same plan in clustered writes.
- Rebuild for a batch only if its Arrow schema actually changes.
- Keep the existing literal conversion and transform behavior unchanged.

## Why

Partition routing rebuilt the partition type, field lookup map, source names, source types, and Arrow column indexes for every batch. This fixed setup cost is noticeable for streams with many small batches.

The plan is built from the first real batch because partitioned equality-delete input may include routing columns that are omitted from the output file schema.

## Benchmark

Apple M1 Pro, median of 3 runs:

    go test ./table -run ^$ -bench ^BenchmarkPartitionExtraction$ -benchmem -benchtime=1s -count=3 -cpu=1

| Rows per batch | main | this PR | main allocations | this PR allocations |
| ---: | ---: | ---: | ---: | ---: |
| 0 | 1695 ns | 197 ns | 1432 B, 10 allocs | 128 B, 2 allocs |
| 1 | 3926 ns | 2412 ns | 3928 B, 28 allocs | 2624 B, 20 allocs |
| 16 | 20558 ns | 18070 ns | 21456 B, 150 allocs | 20152 B, 142 allocs |
| 1024 | 184282 ns | 156782 ns | 21456 B, 150 allocs | 20152 B, 142 allocs |

## Tests

- go test ./...
- go test -race ./table with the fanout, clustered, equality-delete, and shredded-variant writer tests
- go vet ./...
- git diff --check

### DIFF
--- a/table/clustered_writer.go
+++ b/table/clustered_writer.go
@@ -65,6 +65,7 @@ func clusteredPartitionedWrite(
 			currentRec          partitionRecord
 			currentWriter       *RollingDataWriter
 			completedPartitions = make(closedPartitionSet)
+			extractionPlan      *partitionExtractionPlan
 		)
 
 		closeCurrentWriter := func() error {
@@ -144,7 +145,16 @@ func clusteredPartitionedWrite(
 				return
 			}
 
-			partitions, err := getRecordPartitions(spec, schema, rec)
+			if extractionPlan == nil {
+				extractionPlan, err = newPartitionExtractionPlan(spec, schema, rec.Schema())
+				if err != nil {
+					fail(err)
+
+					return
+				}
+			}
+
+			partitions, err := extractionPlan.getRecordPartitions(rec)
 			if err != nil {
 				fail(err)
 
--- a/table/partition_extraction_bench_test.go
+++ b/table/partition_extraction_bench_test.go
@@ -1,0 +1,80 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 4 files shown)


zeroshade (approved):
I verified this locally: build, `go vet`, and `go test -race ./table` over the fanout/clustered/partition tests all pass on the PR head, and the benchmark reproduces the claimed direction.

The concurrency story checks out: `sync.Once` gives the necessary happens-before for `plan`/`planErr` across fanout workers, the plan is immutable after construction, and all per-batch state (`partitionMap`, `partitionRec`, `partitionColumns`) stays call-local in `partitionExtractionPlan.getRecordPartitions` (table/partitioned_fanout_writer.go:375). The schema-mismatch fallback building a throwaway local plan (rather than mutating the shared one) is the right call for worker safety, and the dropped-source-column path (`columnIndex = -1` → nil partition value) preserves the existing equality-delete behavior — confirmed by `TestGetRecordPartitionsWithDroppedLeadingSourceColumn` still passing. Nice test coverage on the reordered/changed-schema cases.

Two non-blocking observations:

1. `planErr` is sticky: if the first batch fails plan construction, every subsequent `getPartitions` call returns that error even for batches whose schema would succeed. In practice any error aborts the whole write, so this doesn't matter today, but a short comment on the field might save a future reader some head-scratching.
2. When batch schemas alternate, each mismatched batch rebuilds a plan from scratch (no caching of the alternate). That's the same cost as pre-PR, so fine — just noting the reuse only benefits the stable-schema case, which is the common one.

Also, `matchesSchema` using `Schema.Equal` (ignoring metadata differences) is safe here since extraction only depends on field names, types, and order.

PR #1766, opened 2026-08-11
perf(table): avoid copying contiguous partition batches

fallintoplace:
## What changed

- Reuse the input record when a partition contains the full batch.
- Use Arrow slices when the selected rows are one contiguous range.
- Check the complete row range and bounds before taking the fast path.
- Keep the existing `compute.Take` path for scattered rows.

## Why

`partitionBatchByKey` always built an Int64 index array and ran `Take`. This copied every selected column even when the result was the full batch or one contiguous range.

## Benchmark

Apple M1 Pro, 65,536 rows, median of 3 runs:

```
go test ./table -run '^$' -bench '^BenchmarkPartitionBatchByKey$' -benchmem -benchtime=1s -count=3 -cpu=1
```

| Case | main | this PR |
| --- | ---: | ---: |
| 1 column, full | 649,907 ns/op, 1,610,469 B/op, 47 allocs/op | 49,908 ns/op, 0 B/op, 0 allocs/op |
| 1 column, contiguous half | 319,600 ns/op, 815,066 B/op, 47 allocs/op | 18,402 ns/op, 288 B/op, 4 allocs/op |
| 8 columns, full | 2,360,602 ns/op, 5,351,519 B/op, 236 allocs/op | 35,814 ns/op, 0 B/op, 0 allocs/op |
| 8 columns, contiguous half | 1,217,218 ns/op, 2,721,096 B/op, 236 allocs/op | 18,282 ns/op, 1,760 B/op, 19 allocs/op |

Scattered selections still use `Take`. Their allocation counts are unchanged and runtime stayed within normal benchmark variance.

## Tests

- `go test ./...`
- `go test -race ./table -run 'TestFanoutWriter|TestClusteredWriter|TestWriteEqualityDeleteFilesPartitioned|TestShreddedVariantPartitioned' -count=1`
- `go vet ./...`
- `git diff --check`

### DIFF
--- a/table/partitioned_fanout_writer.go
+++ b/table/partitioned_fanout_writer.go
@@ -557,7 +557,7 @@ func partitionBatchByKey(ctx context.Context) partitionBatchFn {
 				return record, nil
 			}
 
-			if recordHasRowBoundedStorage(record) && contiguousSliceHasBoundedRetention(start, end, record.NumRows()) {
+			if contiguousSliceHasBoundedRetention(start, end, record.NumRows()) && recordHasRowBoundedStorage(record) {
 				return record.NewSlice(start, end), nil
 			}
 		}
--- a/table/partitioned_fanout_writer.go
+++ b/table/partitioned_fanout_writer.go
@@ -569,6 +569,11 @@ func partitionBatchByKey(ctx context.Context) partitionBatchFn {
 		rowIndicesArr := bldr.NewInt64Array()
 		defer rowIndicesArr.Release()
 
+		recordMetadata := arrow.Metadata{}
+		if recordWithMetadata, ok := record.(arrow.RecordBatchWithMetadata); ok {
+			recordMetadata = recordWithMetadata.Metadata()
+		}
+
 		partitionedRecord, err := compute.Take(
 			ctx,
 			*compute.DefaultTakeOptions(),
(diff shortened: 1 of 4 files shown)


table/partitioned_fanout_writer.go
```diff
@@ -520,6 +520,22 @@ func partitionBatchByKey(ctx context.Context) partitionBatchFn {
...
+		}
+
+		if start, end, ok := contiguousRowRange(rowIndices, record.NumRows()); ok {
+			if start == 0 && end == record.NumRows() {
+				record.Retain()
+
+				return record, nil
+			}
+
+			return record.NewSlice(start, end), nil
```
zeroshade: `RecordBatch.NewSlice` retains the source arrays' complete backing buffers, not just `[start:end]`. `RollingDataWriter.Add` then retains this record and can queue 64 records asynchronously, so a slow partition containing only a tiny contiguous range can pin up to 64 complete input batches. A checked-allocator probe selecting one row from a 100,000-row string batch retained 34,095,168 bytes after releasing the source record; the previous `compute.Take` path retained 384 bytes. The synchronous benchmark does not measure this queued peak-memory behavior. Could we keep the full-record fast path but use `Take` (or another bounded-copy policy) for disproportionately small partial ranges, and add a queued-writer memory regression test?

zeroshade (changes requested):
The full-record reuse looks safe, but the partial contiguous-slice path can retain an entire input batch for a very small partition. Because rolling writers retain and asynchronously queue up to 64 records, this can cause substantial memory amplification and undermines the existing per-worker input-memory bound. Please bound retained backing storage for partial ranges and add a queued-writer peak-memory regression test.

table/partitioned_fanout_writer.go
```diff
@@ -541,6 +559,35 @@ func partitionBatchByKey(ctx context.Context) partitionBatchFn {
 	}
 }
 
+// contiguousSliceHasBoundedRetention reports whether a partial zero-copy slice
+// retains no more than twice as many input rows as it returns. The range must
+// already have been validated by contiguousRowRange.
+func contiguousSliceHasBoundedRetention(start, end, numRows int64) bool {
+	selectedRows := end - start
+
+	return selectedRows >= numRows-selectedRows
```
zeroshade, line 800: This bounds retained rows, not retained bytes. For a 4,096-row string batch where the omitted first half contains 16 KiB strings and the selected contiguous second half contains one-byte strings, this condition accepts `NewSlice`. After releasing the input, the result retained 67,125,824 bytes; `compute.Take` for the same rows retained 10,560 bytes. A 64-record queue can therefore still pin several gigabytes. Please use a byte-aware policy, or keep partial zero-copy slices only for storage types where row count actually bounds backing-buffer size.

table/partitioned_fanout_writer_test.go
```diff
@@ -683,6 +683,144 @@ func (s *FanoutWriterTestSuite) TestGetRecordPartitionsWithDroppedLeadingSourceC
...
+		s.Require().NoError(err)
+
+		s.Require().NoError(writer.Add(partitioned))
+		partitioned.Release()
+		record.Release()
+
+		peakBytes = max(peakBytes, s.mem.CurrentAlloc())
+	}
+
+	s.Less(peakBytes, fullBatchBytes*4, "queued partial batches should not retain complete input batches")
```
zeroshade, line 1129: This regression uses uniform 128-byte payloads, so selected-row ratio happens to approximate retained-byte ratio. Please add a variable-width skew case where omitted rows contain large payloads and selected rows are small; the current implementation retains about 67 MB instead of roughly 10 KB for that shape.

zeroshade (changes requested):
The previous memory-retention blocker is not fully resolved. The new threshold bounds retained row count, but variable-width columns can concentrate almost all bytes in the omitted rows, so a qualifying zero-copy slice can still pin an arbitrarily large backing buffer. See the inline comments for the reproduced case and missing regression coverage.

All four commits also lack the `Signed-off-by:` trailer required by `CONTRIBUTING.md`: `4f064bf`, `3f8ac83`, `d3b39be`, and `b4699a5`. The branch also currently conflicts with `main` in `table/partitioned_fanout_writer_test.go`; please rebase when updating it.

---

>

zeroshade (approved):

PR #1767, opened 2026-08-11
feat(metrics): add OpenTelemetry reporter (#1236)

tanmayrauth:
Seventh and final PR of the metrics reporting stack. Adds an in-module metrics/otel package with an OpenTelemetry-backed metrics.Reporter that maps every field of a ScanReport's ScanMetricsResult and a CommitReport's CommitMetricsResult to an OpenTelemetry instrument, so scan/commit metrics can flow to any OTLP
  backend (Prometheus, Datadog, CloudWatch, Grafana Cloud, …).

  Design
  - Host owns the SDK. The reporter obtains a Meter from the global MeterProvider, or one supplied via WithMeter. With no SDK registered, the global provider returns no-op instruments and metric calls are silently dropped — the standard OpenTelemetry contract.
  - Instruments. Durations (iceberg.scan.planning.duration, iceberg.commit.duration) are histograms in ms; every other field is a monotonic counter (bytes carry the By unit, counts are unit-less). Unset fields are skipped, so no empty time series is created. Names, units, and descriptions mirror Java's
  ScanInstruments/CommitInstruments (17 scan + 30 commit instruments).
  - Bounded cardinality. By default scan metrics carry iceberg.table.name and commit metrics carry iceberg.table.name + iceberg.operation. iceberg.schema.id is opt-in via WithAttributes; the snapshot id is never a metric attribute (unique per commit → unbounded cardinality). Because iceberg.table.name is on by
  default, series grow with table count; the OTel SDK caps a stream at 2000 attribute combinations and folds the overflow into a single otel.metric.overflow=true point, so drop iceberg.table.name in very-large-table deployments.
  - Kept as an in-module package (not a separate module) because go.opentelemetry.io is already a dependency; go.mod only promotes the three directly-imported OTel modules from indirect to direct.

  Testing
  metrics/otel unit tests cover full scan/commit instrument coverage, unit correctness, null-metrics handling, counter accumulation across reports, the default and configured attribute sets (including schema-id opt-in and empty allowlist), pointer/typed-nil report dispatch, Close, and the default global-meter path.
  
  
  Important Note: 

  ▎ Names, units, and the attribute set mirror Java's OtelMetricsReporter (apache/iceberg#16250) so metrics stay consistent across implementations; treat them as provisional until that PR merges.
  
  Related #1236

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,6 +58,9 @@ require (
 	github.com/uptrace/bun/dialect/sqlitedialect v1.2.18
 	github.com/uptrace/bun/driver/sqliteshim v1.2.18
 	github.com/uptrace/bun/extra/bundebug v1.2.18
+	go.opentelemetry.io/otel v1.44.0
+	go.opentelemetry.io/otel/metric v1.44.0
+	go.opentelemetry.io/otel/sdk/metric v1.44.0
 	gocloud.dev v0.46.0
 	golang.org/x/exp v0.0.0-20260218203240-3dfff04db8fa
 	golang.org/x/oauth2 v0.36.0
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
A fitting capstone to the metrics stack. The construction-time instrument creation, allowlist-bounded cardinality with the snapshot-ID exclusion rationale spelled out, nil-field skipping, and host-owns-the-SDK contract are all the right OpenTelemetry idioms — and `Report` staying synchronous is correct here since counter adds and histogram records are SDK-buffered, unlike the REST reporter's network sends. Mirroring Java's instrument names with the explicit provisional caveat against apache/iceberg#16250 sets the right expectations. Nice work across all seven PRs of this series.

---

PR #1769, opened 2026-08-11
perf(table): size partition row indices adaptively

fallintoplace:
## What changed

- Keep the existing maximum starting capacity of 128 row indexes.
- Reduce the starting capacity as more partitions are discovered in a batch.
- Use `batch rows / discovered partitions` as the estimate, with a minimum of one.
- Keep grouping and row ordering unchanged.

## Why

Every new partition currently reserves space for 128 row indexes. This works well for low-cardinality batches, but a batch with one partition per row reserves about 1 KiB for every single row.

The adaptive estimate keeps the current behavior for a few large partitions. It gradually moves toward one row of capacity for high-cardinality batches.

## Benchmark

Apple M1 Pro, 32,768 rows, median of 5 runs:

```
go test ./table -run '^$' -bench '^BenchmarkPartitionRowCapacity$' -benchmem -benchtime=1s -count=5 -cpu=1
```

| Distribution | main | this PR | Change |
| --- | ---: | ---: | ---: |
| 1 partition | 1,160,120 B/op | 1,160,120 B/op | unchanged |
| 16 partitions | 951,784 B/op | 951,784 B/op | unchanged |
| 1,024 partitions | 1,759,064 B/op | 1,348,728 B/op | -23% |
| 1 partition per row | 50,331,068 B/op | 18,241,259 B/op | -64% |
| 90% in one partition | 5,994,673 B/op | 3,580,624 B/op | -40% |

Allocation counts are unchanged in every case. The unique-partition case was also about 14% faster in the repeated comparison.

## Tests

- `go test ./...`
- `go test -race ./table -run '^TestFanoutWriter$' -count=1`
- `go vet ./...`
- `git diff --check`


table/partitioned_fanout_writer.go
```diff
@@ -486,16 +487,35 @@ func (n *partitionMapNode) getOrCreate(partitionRec partitionRecord, fieldInfo [
 	}
 
 	partVal = &partitionInfo{
-		rows:            make([]int64, 0, 128), // modest starting capacity
+		rows:            make([]int64, 0, initialPartitionRowCapacity(numRows, n.partitionCount)),
```
zeroshade, line 491: This estimate depends on partition discovery order and can land just below Go's slice-growth thresholds. In a 32,768-row probe where 300 partitions are discovered first and partitions 256–299 later grow to roughly 110–128 rows, this branch increased usage from 1,369,248 to 1,455,136 B/op (+6.3%) and from 11,950 to 11,994 allocs/op versus the merge base. The included skew benchmark only exercises the favorable hot-partition-first ordering. Could we quantize these capacities to growth-safe boundaries, or otherwise prevent sub-128 slices from growing beyond the previous 128 allocation? Please also add a late-discovered/skewed benchmark covering this case.

zeroshade (changes requested):
The adaptive sizing improves the benchmarked high-cardinality cases, but it can increase memory and allocations for valid row orderings when late-discovered partitions grow past their estimate. I reproduced a roughly 6.3% B/op regression with 44 additional allocations versus the merge base. Please make the sizing growth-safe and add a late-discovered/skewed benchmark before merging.

zeroshade (approved):

PR #1771, opened 2026-08-12
perf(table): skip metrics after partition rejection

fallintoplace:
## What changed

- Short-circuit manifest entry filtering when the partition evaluator rejects an entry.
- Keep partition evaluation first.
- Run the existing inclusive metrics evaluator for entries accepted by the partition filter.
- Add regression coverage for the evaluation order and all three outcomes.
- Add a rejection-heavy benchmark with unique data-file paths.

## Why

Partition filtering can reject files before their column metrics are needed. Skipping metrics evaluation for those files avoids unnecessary work during scan planning. The candidate file results remain the same.

There are no API changes.

## Validation

- `go test ./... -count=1`
- `go test -race ./table -run TestOpenManifestShortCircuitsMetricsEvaluation -count=1`
- `go vet ./...`

## Benchmark

Local run on an Apple M1 Pro. The benchmark scans 1,000 entries, rejects 95% with the partition filter, and runs five samples.

| branch | time/op | bytes/op | allocs/op |
| --- | ---: | ---: | ---: |
| `origin/main` | 2.527 ms | 2.666 MB | 30,521 |
| `this branch` | 2.117 ms | 1.914 MB | 22,479 |
| change | **-16.2%** | **-28.2%** | **-26.3%** |

Command:

```text
go test ./table -run ^$ -bench ^BenchmarkOpenManifestPartitionRejectsMostEntries$ -benchtime=2s -count=5
```

Results will vary by machine, but this shows the expected benefit when partition rejection is common.

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -215,7 +215,7 @@ func openManifest(io io.IO, manifest iceberg.ManifestFile,
 			return nil, err
 		}
 
-		if p && m {
+		if m {
 			out = append(out, entry)
 		}
 	}
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -206,6 +206,9 @@ func openManifest(io io.IO, manifest iceberg.ManifestFile,
 		if err != nil {
 			return nil, err
 		}
+		if !p {
+			continue
+		}
 
 		m, err := metricsEval(entry.DataFile())
 		if err != nil {
--- a/table/scanner_bench_test.go
+++ b/table/scanner_bench_test.go
@@ -1,0 +1,115 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"testing"
+
+	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
+)
+
+func BenchmarkOpenManifestPartitionRejectsMostEntries(b *testing.B) {
+	const (
+		entryCount    = 1_000
+		rejectThrough = 950
+	)
+
+	spec := partitionedSpec()
+	schema := simpleSchema()
+	snapshotID := int64(1)
+	entries := make([]iceberg.ManifestEntry, entryCount)
+	for i := range entries {
+		value := int32(i)
+		builder, err := iceberg.NewDataFileBuilder(
+			spec,
+			iceberg.EntryContentData,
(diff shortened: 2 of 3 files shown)


table/scanner.go
```diff
@@ -206,6 +206,9 @@ func openManifest(io io.IO, manifest iceberg.ManifestFile,
 		if err != nil {
 			return nil, err
 		}
+		if !p {
```
laskoviymishka, line 209: Now that we `continue` on `!p` here, the `p &&` in the `if p && m {` a few lines down is dead, since `p` is always true by the time we reach it.

I'd collapse it to `if m {` so the control-flow intent reads cleanly; otherwise the next reader has to stop and convince themselves `p` can't be false at that gate.

table/scanner_internal_test.go
```diff
@@ -285,6 +285,99 @@ func TestEqualityDeletePartitionKeyDistinguishesSignedZero(t *testing.T) {
...
+		partitionMatches bool
+		metricsMatches   bool
+		wantEvaluations  []string
+		wantEntries      int
+	}{
+		{
+			name:             "partition rejection skips metrics",
+			partitionMatches: false,
+			metricsMatches:   true,
+			wantEvaluations:  []string{"partition"},
```
laskoviymishka, line 341: This case proves metricsEval isn't *called* when the partition filter rejects, which is exactly right. The more interesting behavior change, though, is that a metricsEval *error* is now swallowed on that path, and nothing covers it.

Before this PR, if the partition filter returned `(false, nil)` and metricsEval would have errored, that error propagated; now it's skipped entirely. That's almost certainly what we want since we're not going to read those files anyway, but it's an untested contract shift. I'd add a case where `partitionMatches` is false and the metrics closure returns an error, asserting `openManifest` still returns a nil error. wdyt?

table/scanner_bench_test.go
```diff
@@ -0,0 +1,115 @@
...
+		b.Fatal(err)
+	}
+
+	fs := iceio.NewMemFS()
+	if err := fs.WriteFile(manifestPath, manifestBytes.Bytes()); err != nil {
+		b.Fatal(err)
+	}
+
+	partitionFilter := func(df iceberg.DataFile) (bool, error) {
+		return df.Partition()[1000].(int32) >= rejectThrough, nil
```
laskoviymishka, line 94: One caveat on the reported percentages: this partition filter is a raw map-lookup-plus-compare, but in production the partition evaluator runs through `buildPartitionEvaluator` → `ExpressionEvaluator` → `GetPartitionRecord`, which is a good bit heavier. So the absolute count of skipped metricsEval calls is real, but the savings *fractions* will read larger here than in a real scan where the denominator is bigger.

I'd either build the filter via `buildPartitionEvaluator` for an end-to-end number, or drop a one-line comment noting the filter is deliberately simplified and the percentages reflect metricsEval savings only. Either's fine, just worth making explicit.

laskoviymishka (approved):
Really clean change, and thanks for the bench and unit coverage. That's exactly the kind of evidence I want to see on a perf tweak. The short-circuit itself is correct and matches how Java's `ManifestReader` and PyIceberg both order the partition and metrics evaluators, so no objections to the approach.

A couple of small things and I'm happy to approve. Now that we `continue` on `!p`, the `p &&` in the `if p && m` just below is dead, so I'd collapse it to `if m` so the control flow reads cleanly. And the change quietly stops propagating metricsEval errors for partition-rejected entries; that's the behavior we want, but I'd add a test case for it so the contract is pinned down.

One non-blocking note on the benchmark: the partition filter there is a simplified closure rather than the real `buildPartitionEvaluator` path, so the savings percentages read a little high relative to a production scan. Worth a one-line comment noting that, or wiring the real evaluator if it's easy.

(Also, the red CI is an unrelated `intrange` lint failure in a file this PR doesn't touch, not anything here.)

Fix the first two and this is good to land.

zeroshade (approved):
LGTM, i'll leave for @laskoviymishka to do the merge

PR #1772, opened 2026-08-12
perf(table): skip clean positional delete batches

fallintoplace:
## What changed

When a data file has positional deletes, the scanner currently builds a survivor index and calls `compute.Take` for every record batch. Most batches may not contain a deleted row, so this does extra work and rebuilds batches that could be passed through unchanged.

This PR:

- returns the original Arrow record batch when no positional delete applies;
- creates the survivor index only after the first deleted row in a batch;
- keeps the existing batch-local index behavior for later batches;
- keeps row-position tracking correct across batches and pruned row groups.

The no-op path retains the input batch before returning it, so ownership stays compatible with the scanner pipeline.

## Tests

- `go test ./table -count=1`
- `go test -race ./table -run '^TestProcessPositionalDeletes' -count=1`
- Coverage for zero deletes, first/middle/last row deletes, batch boundaries, all rows deleted, pruned row groups, and allocator cleanup.

## Benchmarks

Apple M1 Pro, 64K-row batch:

- Clean batch: `189,498 ns/op`, `112 B/op`, `3 allocs/op`
- Partially deleted batch: `859,138 ns/op`, `1,610,925 B/op`, `51 allocs/op`

The partial-delete path keeps the existing reconstruction behavior.

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -564,7 +564,7 @@ func combinePositionalDeletes(mem memory.Allocator, deletes set[int64], cursor *
 		if _, deleted := deletes[cursor.next()]; deleted {
 			if bldr == nil {
 				bldr = array.NewInt64Builder(mem)
-				bldr.Reserve(int(nrows))
+				bldr.Reserve(int(i))
 				for j := range i {
 					bldr.Append(j)
 				}
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -553,15 +553,35 @@ type set[T comparable] map[T]struct{}
 // batch handed to compute.Take, not into the whole file. Without batch-local
 // indices, the second and later batches of a file would pass indices >= the batch
 // length and compute.Take would fail with "index error: N out of bounds".
+//
+// A nil result means that no row in the batch was deleted. The index builder is
+// allocated lazily after the first deleted row so clean batches can pass through
+// without reconstructing their columns.
 func combinePositionalDeletes(mem memory.Allocator, deletes set[int64], cursor *rowPositionCursor, nrows int64) arrow.Array {
-	bldr := array.NewInt64Builder(mem)
-	defer bldr.Release()
+	var bldr *array.Int64Builder
 
 	for i := range nrows {
-		if _, ok := deletes[cursor.next()]; !ok {
+		if _, deleted := deletes[cursor.next()]; deleted {
+			if bldr == nil {
+				bldr = array.NewInt64Builder(mem)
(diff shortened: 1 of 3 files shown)


zeroshade:
Will review once this is marked ready and the conflicts are resolved

table/arrow_scanner.go
```diff
@@ -522,16 +522,36 @@ type set[T comparable] map[T]struct{}
...
-	bldr := array.NewInt64Builder(mem)
-	defer bldr.Release()
+	var bldr *array.Int64Builder
 
 	for i := range nrows {
-		if _, ok := deletes[cursor.next()]; !ok {
+		if _, deleted := deletes[cursor.next()]; deleted {
+			if bldr == nil {
+				bldr = array.NewInt64Builder(mem)
+				bldr.Reserve(int(nrows))
```
zeroshade: Reserving `nrows` regresses delete-heavy batches because the output index only contains survivors. In a 64K-row all-deleted probe, `main` used ~2.75 KB/op at ~294–299 µs/op, while this branch used ~1.078 MB/op at ~357–364 µs/op. Reserving only the prefix that must be backfilled (`int(i)`) restored ~2.9 KB/op and ~291–303 µs/op while preserving the lazy-builder optimization. Please avoid reserving the full batch here.

table/arrow_scanner_posdelete_bench_test.go
```diff
@@ -209,3 +210,52 @@ func benchmarkChunkBoundaries(length, numChunks int) []int {
...
+
+func BenchmarkProcessPositionalDeletes(b *testing.B) {
+	const numRows = 64 * 1024
+
+	for _, tc := range []struct {
+		name    string
+		deletes set[int64]
+	}{
+		{name: "clean", deletes: set[int64]{numRows: {}}},
+		{name: "partial", deletes: set[int64]{numRows / 2: {}}},
```
zeroshade, line 224: This `partial` case deletes only one row, so it does not expose the full-batch allocation above. Please add a delete-heavy/all-deleted case so the benchmark protects both the clean-batch improvement and performance when few survivor indices are needed.

zeroshade (changes requested):
The no-op ownership and cursor handling look correct, but the full-batch reservation introduces a substantial delete-heavy performance regression. I left inline comments on the implementation and benchmark coverage.

Also, both commits currently lack the `Signed-off-by:` trailer required by `CONTRIBUTING.md`. Please add DCO sign-offs when updating the branch.

zeroshade (approved):

PR #1773, opened 2026-08-12
perf(table): group positional deletes by file path

fallintoplace:
## What changed

- Build positional-delete indexes by file path in one pass over the delete rows.
- Keep the existing row order, including repeated positions.
- Keep support for string, large-string, and dictionary-encoded file paths.
- Add cleanup and ownership checks for Arrow data and partial errors.
- Add a benchmark that varies the number of referenced files.

## Why

The old code found the unique file paths first, then scanned the full delete columns once for every path. This repeated work when one delete file referenced many data files.

The new code walks `file_path` and `pos` together once and appends each position to the builder for its path.

## Tests

- `go test ./table -count=1`
- `go test ./... -count=1`
- `go vet ./table`
- `git diff --check`

## Benchmark

The benchmark uses 1,000,000 delete rows and covers 1, 10, 100, 1,000, and 10,000 paths.

- 1 path: 21.8 ms/op
- 10,000 paths: 87.9 ms/op


zeroshade:
since this is a draft i'll hold off on reviewing until it is marked ready and the conflicts are resolved

zeroshade (approved):

PR #1774, opened 2026-08-12
perf(table): stream projected equality deletes

fallintoplace:
## Summary

- Read equality delete files with only the fields used by the delete key.
- Build encoded delete keys while record batches are being read.
- Keep the existing field ID and name mapping behavior.
- Preserve null handling, type encoding, cancellation, and the current worker limit.
- Add an equivalence test against the old materialized path.

## Why

`ReadTable` loaded the full equality delete file before the keys were built. This also loaded columns that were not part of the equality key. Streaming the projected columns lowers peak memory use and avoids that extra work.

## Checks

- `go test ./table -count=1`
- `go test -race ./table -run "EqualityDelete|ReadEqualityDeleteFile" -count=1`
- `go vet ./table`

## Benchmark

For a 100k-row equality delete file on an Apple M1 Pro:

- Streamed projection: 30.6 MB/op, 11.6 ms/op
- Materialized table: 161.2 MB/op, 25.0 ms/op

### DIFF
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -368,7 +368,7 @@ func equalityDeleteSetForFiles(
 
 // readEqualityDeleteFile reads a single equality delete file and returns
 // the set of encoded delete keys and the column names used.
-func readEqualityDeleteFile(ctx context.Context, fs iceio.IO, tableSchema *iceberg.Schema, nameMapping iceberg.NameMapping, dataFile iceberg.DataFile, fieldIDs []int) (set[string], []string, error) {
+func readEqualityDeleteFile(ctx context.Context, fs iceio.IO, tableSchema *iceberg.Schema, nameMapping iceberg.NameMapping, dataFile iceberg.DataFile, fieldIDs []int) (keys set[string], colNames []string, err error) {
 	src, err := internal.GetFile(ctx, fs, dataFile, true)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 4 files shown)


table/equality_delete_reader.go
```diff
@@ -380,24 +380,28 @@ func readEqualityDeleteFile(ctx context.Context, fs iceio.IO, tableSchema *icebe
...
+	if nameMapping == nil {
+		nameMapping = tableSchema.NameMapping()
+	}
+
+	projectedIDs := make(map[int]struct{}, len(fieldIDs))
+	for _, fieldID := range fieldIDs {
+		projectedIDs[fieldID] = struct{}{}
+	}
+
+	projectedSchema, colIndices, err := rdr.PrunedSchema(projectedIDs, nameMapping)
```
zeroshade, line 392: `PrunedSchema` gives this name mapping precedence over embedded Parquet field IDs. Because the fallback mapping above is built from the *current* table names, a legal rename followed by name reuse can prune away the equality field: for example, a delete file stores ID 1 as `old_name`, while the current schema has ID 1 renamed to `new_name` and reuses `old_name` for ID 2. I reproduced that case: the previous materialized reader succeeds using embedded ID 1, while this path maps the file column to ID 2 and fails with `equality field ID 1 ... not found`. Please preserve embedded IDs as authoritative and use name mapping only for files/fields without IDs, with a regression test covering rename plus name reuse.

zeroshade (changes requested):
The streaming approach and benchmark results look good, but there is one schema-evolution correctness regression that needs to be addressed; details are inline. Please also add the required `Signed-off-by:` trailer to both commits (`f448563d48e3d7c4881378e032d907200f66349d` and `d97da3a572b0c3ef13563d264945785079368e07`) per [CONTRIBUTING.md](https://github.com/apache/iceberg-go/blob/main/CONTRIBUTING.md#submitting-a-pull-request).

---

>

zeroshade (approved):
The previous schema-evolution blocker is fixed. Embedded Parquet field IDs are now authoritative during pruning, with name mapping retained only as a fallback for fields without IDs. The added rename-plus-name-reuse regression matches the old materialized path.

I also verified the relevant tests under the race detector, the full table package, internal Parquet pruning tests, vet, and the benchmark. The streaming path preserves behavior while substantially reducing memory use. All DCO trailers and CI checks are complete.

---

PR #1775, opened 2026-08-12
perf(table): pre-resolve Parquet stats columns

fallintoplace:
## What does this change do?

`DataFileStatsFromMeta` currently resolves each Parquet column path while processing every row group. The same path lookup and variant-parent walk is repeated for the whole file.

This change resolves the physical columns once from the file schema and stores small immutable descriptors for them. The row-group loop can then reuse the resolved path, Iceberg field ID, statistics collector, and variant-child status.

The same descriptors are also reused while collecting shredded variant bounds.

## Behavior

- Keeps existing counts, sizes, null counts, min/max bounds, and metrics modes.
- Keeps nested column paths and variant child handling.
- Keeps the existing behavior for missing field mappings.
- Keeps metrics-plan validation for missing collectors, including reserved row-lineage columns.
- Keeps historical schema handling by using the resolved Parquet file schema and the caller-provided path mapping.
- Does not add a package-global cache.

## Tests

- Added semantic-equivalence coverage for direct fields, nested fields, variant descendants, and missing mappings.
- Added a paired before/after benchmark for the repeated column-resolution work.
- Added a multi-row-group end-to-end benchmark with nested and shredded variant columns.
- `go test -race ./table/internal -run 'TestResolveParquetStatsColumns|TestMetricsPrimitiveTypes|TestShreddedVariantStatsDoesNotPanic|TestShreddedVariantReadRoundTrip|TestCollectVariantBounds|TestMetricsSkipColumnOutsidePlan' -count=1`
- `go vet ./table/internal`

## Benchmark

The benchmark uses 32 row groups, 10,000 rows per group, and 9 physical columns. It includes nested ordinary columns plus shredded variant residual and typed-value columns.

I ran the same end-to-end benchmark on `origin/main` at `4390d2af2` and this branch at `5f74dd7c1`.

Command:

`go test ./table/internal -run '^$' -bench '^BenchmarkDataFileStatsFromMetaNestedVariant$' -benchmem -benchtime=2s -count=5`

Apple M1 Pro, Go 1.26.3:

| | Before (`origin/main`) | After |
| --- | ---: | ---: |
| Time | 376-401 us/op | 323-329 us/op |
| Allocated bytes | 898,454-898,458 B/op | 878,677-878,681 B/op |
| Allocations | 5,746 allocs/op | 4,987 allocs/op |
| Rows | 320,000 rows/op | 320,000 rows/op |
| Physical columns | 288 columns/op | 288 columns/op |

Median time improved from about 386 us/op to 328 us/op, or about 15%. Allocations dropped by about 13%.

The paired lookup benchmark is also included in the branch:

`go test ./table/internal -run '^$' -bench '^BenchmarkParquetStatsColumnResolution$' -benchmem -benchtime=2s -count=5`

It measured about 70-78 us/op before and 42-47 us/op after, with 1,376 versus 873 allocations per operation.

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -1378,8 +1378,72 @@ func (w wrappedDecByteArrayStats) Max() iceberg.Decimal {
 	return iceberg.Decimal{Val: dec, Scale: w.scale}
 }
 
+// parquetStatsColumn contains the immutable per-file resolution needed to
+// interpret a physical Parquet column's statistics. The row-group metadata
+// still carries the values that are accumulated, but the path-to-field lookup
+// is independent of the row group and is done once from the resolved schema.
+type parquetStatsColumn struct {
+	path         string
+	fieldID      int
+	statsCol     StatisticsCollector
+	variantChild bool
+	skipStats    bool
+	resolveErr   error
+}
+
+func resolveParquetStatsColumns(meta *metadata.FileMetaData, statsCols map[int]StatisticsCollector, colMapping map[string]int, variantFieldIDs map[int]struct{}) []parquetStatsColumn {
+	columns := make([]parquetStatsColumn, meta.Schema.NumColumns())
+	for pos := range meta.Schema.NumColumns() {
+		column := meta.Schema.Column(pos)
+		columnPath := column.ColumnPath()
+		descriptor := parquetStatsColumn{path: column.Path()}
+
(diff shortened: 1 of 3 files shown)


zeroshade:
since this is a draft i'll hold off reviewing until it is marked ready and the conflicts are resolved

zeroshade (approved):

PR #1779, opened 2026-08-12
fix(table): delete column should check the column's own pending update

### LINKED ISSUE #1778: UpdateSchema: delete column validation checks children's updates instead of the column's own updates
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

In Update Schema, `deleteColumn` function guards against deleting a column that has a pending update with 
```go
if _, ok := u.updates[field.ID]; ok
```

However, `u.updates` is a `map[int]map[int]iceberg.NestedField` keyed by `[parentID][fieldID]`, so `u.updates[field.ID]` inspects updates staged for the field's children, not the field itself.

Because of this, 
- Deleting a col that has its own pending update (e.g., a rename, type or doc change) is wrongly **allowed**, silently discarding the staged update - _ideally, we should reject it_
- Deleting a struct whose child was updated is wrongly **rejected** with a misleading "field that has updates cannot be deleted" error, even though the struct itself has no updates

### Contribution

We raise a PR for this fix

badalprasadsingh:
### Description

Fixes #1778 

Fixed `UpdateSchema.deleteColumn` to correctly check for pending updates using the field's own ID under its parent's update map.

### Testing

Added the required regression tests for it.

### DIFF
--- a/table/update_schema.go
+++ b/table/update_schema.go
@@ -380,8 +380,13 @@ func (u *UpdateSchema) deleteColumn(path []string) error {
 		return fmt.Errorf("field that has additions cannot be deleted: %s", fullName)
 	}
 
-	if _, ok := u.updates[field.ID]; ok {
-		return fmt.Errorf("field that has updates cannot be deleted: %s", fullName)
+	// u.updates is keyed by parent ID, then by the field's own ID, so a
+	// field's own pending update lives under its parent's map
+	parentID := u.findParentID(field.ID)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Correct re-index of the two-level updates map, and the semantics land exactly on Java's behavior: a field's own pending update blocks its deletion while deleting a parent legitimately discards staged child updates. The test matrix covering both false-negative and false-positive directions of the old check — including the survivor assertions — is just right.

---

PR #1781, opened 2026-08-12
perf(table): reuse FileIO during local scan planning

fallintoplace:
## Summary

- Reuse one `FileIO` across each concurrent manifest-worker batch.
- Load through the factory again at every batch boundary so credential-renewing factories keep their checkpoints.
- Stop loading later manifest batches after a worker error.
- Keep empty-table planning lazy.
- Leave remote planning and its plan-scoped IO lifecycle unchanged.

## Benchmark

Run with:

`go test ./table -run ^$ -bench BenchmarkPlanFilesFileIOFactory -benchmem -benchtime=1s -count=3`

Results below are from a macOS arm64 machine with max concurrency 8. The factory-call count is the important metric; wall time includes in-memory Avro planning and varies by machine.

| Manifests | Before factory calls/op | After factory calls/op | After ns/op |
| ---: | ---: | ---: | ---: |
| 1 | 2 | 2 | 393,944 |
| 10 | 11 | 3 | 1,130,422 |
| 100 | 101 | 14 | 9,499,501 |

This keeps the renewal checkpoints while reducing backend-factory calls by 73% for 10 manifests and 86% for 100 manifests.

## Tests

- `go test ./table/...`
- `go test -race ./table -run ^TestPlanFiles(ReusesFileIOWithinManifestBatches|RefreshesFileIODuringManifestPlanning|StopsLoadingManifestIOAfterWorkerError)$ -count=1`
- `go vet ./table/...`
- `golangci-lint run --timeout=10m ./table/...`

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -1001,7 +1001,7 @@ func (scan *Scan) collectManifestEntriesWithSchema(
 	manifestIO := newManifestIOBatch(scan.ioF, concurrencyLimit)
 
 	entries := newManifestEntries()
-	g, _ := errgroup.WithContext(ctx)
+	g, gctx := errgroup.WithContext(ctx)
 	g.SetLimit(concurrencyLimit)
 
 	partitionFilters := scan.partitionFiltersForSchema(schema)
@@ -1015,7 +1015,7 @@ func (scan *Scan) collectManifestEntriesWithSchema(
 		}
 
 		g.Go(func() error {
-			fs, err := manifestIO.acquire(ctx)
+			fs, err := manifestIO.acquire(gctx)
 			if err != nil {
 				return err
 			}
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -941,6 +941,7 @@ func (scan *Scan) collectManifestEntries(
 	if err != nil {
 		return nil, err
 	}
+
 	return scan.collectManifestEntriesWithSchema(ctx, manifestList, schema)
 }
 
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -941,20 +941,11 @@ func (scan *Scan) collectManifestEntries(
 	if err != nil {
 		return nil, err
 	}
-	var fs io.IO
-	if len(manifestList) > 0 {
-		fs, err = scan.ioF(ctx)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	return scan.collectManifestEntriesWithSchema(ctx, fs, manifestList, schema)
+	return scan.collectManifestEntriesWithSchema(ctx, manifestList, schema)
 }
 
 func (scan *Scan) collectManifestEntriesWithSchema(
 	ctx context.Context,
-	fs io.IO,
 	manifestList []iceberg.ManifestFile,
 	schema *iceberg.Schema,
 ) (*manifestEntries, error) {
@@ -986,6 +977,13 @@ func (scan *Scan) collectManifestEntriesWithSchema(
 		}
 
 		g.Go(func() error {
+			// FileIO factories may renew credentials between manifest reads. Keep
+			// this load at the worker boundary instead of retaining the IO used for
+			// the manifest list across the whole plan.
+			fs, err := scan.ioF(ctx)
+			if err != nil {
+				return err
+			}
 			partEval, err := partitionEvaluators.Get(int(mf.PartitionSpecID()))
 			if err != nil {
 				return fmt.Errorf("failed to build partition evaluator for spec %d: %w", mf.PartitionSpecID(), err)
@@ -1124,7 +1122,7 @@ func (scan *Scan) planFilesLocal(ctx context.Context, acc *scanMetricsAccumulato
 	}
 
 	// Step 2: Read manifest entries concurrently, accumulating data and positional deletes.
-	entries, err := scan.collectManifestEntriesWithSchema(ctx, fs, manifestList, schema)
+	entries, err := scan.collectManifestEntriesWithSchema(ctx, manifestList, schema)
 	if err != nil {
 		return nil, err
 	}
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -952,6 +952,10 @@ func (b *manifestIOBatch) acquire(ctx context.Context) (io.IO, error) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
 	if b.remaining == 0 {
 		fs, err := b.factory(ctx)
 		if err != nil {
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -931,6 +931,42 @@ func (scan *Scan) filterManifestsWithSchema(
 	return filtered, nil
 }
 
+// manifestIOBatch shares a FileIO within a concurrency-sized batch. A new
+// batch loads through the factory again so factories that renew credentials
+// still get regular checkpoints without rebuilding the backend for every
+// manifest.
+type manifestIOBatch struct {
+	factory FSysF
+	limit   int
+
+	mu        sync.Mutex
+	fs        io.IO
+	remaining int
+}
+
+func newManifestIOBatch(factory FSysF, limit int) *manifestIOBatch {
+	return &manifestIOBatch{factory: factory, limit: max(limit, 1)}
+}
+
+func (b *manifestIOBatch) acquire(ctx context.Context) (io.IO, error) {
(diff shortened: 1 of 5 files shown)


table/scanner.go
```diff
@@ -900,14 +924,23 @@ func (scan *Scan) planFilesLocal(ctx context.Context, acc *scanMetricsAccumulato
 		}
 	}()
 
+	snap, err := scan.ResolveSnapshot()
+	if err != nil || snap == nil {
+		return nil, err
+	}
+	fs, err := scan.ioF(ctx)
```
zeroshade, line 1142: Retaining this concrete FileIO for the entire plan bypasses later factory calls that may refresh credentials. REST tables wire `vendedCredentialRefresher.loadFS` as `ioF`, and each invocation checks `needsRenewal()` before returning an IO. A deterministic expiring-credential factory probe passes on `main`—later manifest workers reacquire FileIO—but fails on this branch with `credentials expired before open`. Long-running plans that cross the renewal window can therefore fail with storage authorization errors. Please preserve credential-renewal checkpoints while avoiding repeated backend construction, and add coverage for expiry during planning.

zeroshade (changes requested):
The FileIO load-count improvement is measurable and the planning tests are green, but two issues need to be addressed before merge. First, retaining one concrete FileIO for the whole plan bypasses the renewal checks in REST's vended-credential factory; details and a differential repro are inline. Second, commit `e60669eed065563c175dea59b029e45488652ed2` has no `Signed-off-by:` trailer, while `CONTRIBUTING.md` requires one on every commit—the signed merge commit does not sign off the earlier commit.

Smaller observation: incremental local planning still loads FileIO at `table/incremental_append_scan.go:162` and then again through `collectManifestEntriesWithSchema` at line 207, so that path performs two loads rather than one per planning operation.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Iceberg Go maintainer. After you've addressed the points above and pushed an update, an Apache Iceberg Go maintainer — a real person — will take the next look at the PR. If you think a finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

table/scanner.go
```diff
@@ -789,6 +785,9 @@ func (scan *Scan) collectManifestEntriesWithSchema(
 		}
 
 		g.Go(func() error {
+			// FileIO factories may renew credentials between manifest reads. Keep
+			// this load at the worker boundary instead of retaining the IO used for
+			// the manifest list across the whole plan.
 			fs, err := scan.ioF(ctx)
```
zeroshade: Each manifest worker now calls `scan.ioF(ctx)` again, so current head measures 2, 11, and 101 factory calls/op for 1, 10, and 100 manifests—the same values documented as “Before,” not the claimed one call/op. This fixes credential renewal by restoring the original repeated factory behavior, but removes the performance benefit that is the purpose of this PR. Please preserve renewal checkpoints while retaining a measurable reuse benefit, or revise the PR scope if this is intentionally only a refactor. The comment below saying one FileIO is shared with all workers should also be updated.

zeroshade (changes requested):
The credential-renewal issue from the previous review is fixed, but the fix restores one `ioF` call per manifest worker. The current benchmark reports 2, 11, and 101 factory calls for 1, 10, and 100 manifests—exactly the description's “Before” values—so the PR no longer delivers its stated FileIO reuse optimization. Please retain the renewal checkpoints while preserving a measurable reuse benefit, or revise the scope and benchmark claims accordingly.

Also, commit `6b623741` needs a `Signed-off-by` trailer, and the PR should reference its associated issue. The description also references the removed `BenchmarkPlanFilesFileIOReuse` and `TestPlanFilesReusesFileIO`.

---

table/scanner.go
```diff
@@ -742,6 +738,42 @@ func (scan *Scan) filterManifestsWithSchema(
...
+	mu        sync.Mutex
+	fs        io.IO
+	remaining int
+}
+
+func newManifestIOBatch(factory FSysF, limit int) *manifestIOBatch {
+	return &manifestIOBatch{factory: factory, limit: max(limit, 1)}
+}
+
+func (b *manifestIOBatch) acquire(ctx context.Context) (io.IO, error) {
```
zeroshade, line 951: `acquire` relies on the factory itself to reject a canceled context. `FSysF` receives a context but is not required to inspect it, so after one worker fails, a queued worker can still invoke a context-agnostic factory and open another manifest.

I reproduced this 10/10 times with concurrency 1: after manifest 0 fails, factory calls increase from the expected 2 to 3 and manifest 1 is opened once. The new cancellation test masks this because its fake factory explicitly returns `ctx.Err()`.

Please have `manifestIOBatch.acquire` check cancellation itself after acquiring the mutex and before returning cached IO or invoking the factory. The regression should use a factory that ignores context so the collector—not the test double—enforces the invariant.

zeroshade (changes requested):
The earlier credential-renewal, reuse-performance, DCO, and stale-description concerns are addressed. Batch reuse now provides the claimed reduction while preserving renewal checkpoints.

One cancellation issue remains inline: a later batch can still invoke a context-agnostic factory and open another manifest after an earlier worker fails. Please enforce cancellation inside `manifestIOBatch.acquire` and strengthen the regression accordingly.

Please also add the associated issue reference to the PR description.

---

zeroshade (approved):
The batch-scoped FileIO reuse now retains renewal checkpoints *and* cancellation, so this is good to go. Approving.

All three of my earlier reviews target force-pushed commits (`bf3e470c`, `6b623741`, `4755cc21`) that are absent from the current eight-commit history, so they were void regardless — but each of their concerns is genuinely addressed in the current head:

| Prior concern | Status |
|---|---|
| Retaining one concrete FileIO for the whole plan bypassed renewal checks in REST's vended-credential factory | addressed — renewal now happens at batch checkpoints, refreshing every `limit` acquisitions |
| The renewal fix regressed reuse to no benefit (2/11/101 factory calls, identical to "Before") | addressed — batch reuse restores the reduction while keeping checkpoints |
| A later batch could invoke a context-agnostic factory and open another manifest after an earlier worker failed | addressed — `manifestIOBatch.acquire` (`table/scanner.go:951-969`) checks `ctx.Err()` while holding the mutex, before either cached reuse or a factory call |
| Missing `Signed-off-by` on individual commits | addressed — all 8 current commits carry the trailer |
| Stale description referencing removed benchmarks/tests | addressed |

The cancellation fix is pinned by `TestManifestIOBatchRejectsCanceledContext` and `TestPlanFilesStopsLoadingManifestIOAfterWorkerError`, both using a context-agnostic factory — which is the right way to test it, since a context-aware factory would mask the bug.

Beyond the prior findings, I checked the things a shared-handle change can break:

- **Close ownership.** `io.IO` has no `Close`, so local planning owns no close obligation and there's no close-while-in-use path. Per-manifest file handles are closed by the `Entries` iterator. Remote plan-scoped `PlanIO` remains separate with its owner/reader leases, and `planFilesLocal`'s deferred `closePlanIO` releases a prior remote plan only after successful local planning, with `planIOState` waiting on active readers. Given this codebase's earlier PlanIO lease bug, I wanted that boundary explicitly clear — it is.
- **Concurrency.** The built-ins are safe for concurrent reads: `LocalFS` is stateless over `os`, `MemFS` uses an RWMutex, and gocloud `blobfs` is immutable over a `Bucket` that guards its operations and `Close` with an RWMutex. The arrow scanner already shares one IO across concurrent delete/DV workers, so this isn't a new assumption.
- **Reuse correctness.** `FSysF` takes only a context; the table closure fixes location and config, and REST `prefixScopedIO` resolves credentials by actual path and caches per credential+authority. The batched IO stays within the same table/path scope.
- **Context.** Workers run under the errgroup context, `acquire` gates queued workers after cancellation, and a fresh `LoadFS` receives that context. The REST vended refresher can return an IO created under an earlier outer context, but that cross-scan caching predates this PR and isn't introduced here.

Benchmark: 2/3/14 factory calls for 1/10/100 manifests versus 2/11/101 before — 73%/86% reduction, backed by the committed `BenchmarkPlanFilesFileIOFactory`.

### Minor — please add the issue reference

The PR description still lacks the associated issue number that `CONTRIBUTING.md` asks for. I raised this on an earlier pass and it's the one item still outstanding; not worth holding the merge over, but please add it.

Also still visible, and I'm satisfied it's intentional: the incremental append path performs one manifest-list factory load plus its worker-batch checkpoint, so two loads remain there by design for credential renewal.

CI green, DCO satisfied on all 8 commits.

---

>

PR #1782, opened 2026-08-12
perf(table): avoid writer-factory lock on cache hits

fallintoplace:
## What changed

- Return cached partition writers before taking the factory mutex.
- Keep the locked recheck so concurrent cache misses still create one writer.
- Add a concurrent creation regression test and a focused cache-hit benchmark.

## Benchmark

Apple M1 Pro, Go 1.26.3, 8 benchmark workers:

```text
                                 before       after
1 partition                     139.5 ns/op   2.6 ns/op
8 partitions                    146.7 ns/op   3.1 ns/op
128 partitions                  152.3 ns/op   3.8 ns/op
```

All cases remain at 0 allocs/op.

## Testing

- `go test ./table`
- `go test -race ./table -run TestRollingDataWriter/TestConcurrentGetOrCreateCreatesOneWriter -count=20`
- `go vet ./table`
- `golangci-lint run --timeout=10m ./table/...`


zeroshade (approved):
LGTM

PR #1783, opened 2026-08-12
fix(table): skip DELETED manifest entries in Snapshot.dataFiles

ryanworl:
## What

`Snapshot.dataFiles` read manifest entries with `discardDeleted=false` and applied no status filter, so DELETED entries — tombstones recording that a file was removed — were yielded as if the files were still reachable. This PR flips that one flag to `true` (the manifest reader's existing `discardDeleted` support) and adds regression tests.

## Why

Every caller of `dataFiles` wants live files only, so the leak made each one subtly wrong:

- `AddFiles` / `AddDataFiles` duplicate walks reported false "already referenced" errors for a path whose previous incarnation was deleted (e.g. by a compaction), blocking a legitimate re-add.
- The `ReplaceFiles`-family belong-to-table checks matched already-removed files, silently staging a redundant re-delete of a file the table no longer references.
- `classifyFilesForDeletions`' full-table branch collected tombstoned files as deletion candidates.

I audited every call site to make sure no caller wants tombstones (all are in `transaction.go`):

| Call site | Wants |
|---|---|
| `ReplaceDataFiles` belong-to-table / already-referenced | live only |
| `AddDataFiles` duplicate check | live only |
| `ReplaceDataFilesWithDataFiles` belong-to-table | live only |
| `ReplaceFiles` belong-to-table | live only |
| `AddFiles` duplicate walk | live only |
| `classifyFilesForDeletions` full-table branch | live only |

Code that genuinely needs tombstones (conflict validation, `collectExistingDVs`) already uses the separate `Snapshot.entries` iterator and filters by status itself, so it's untouched. Scan planning already discards DELETED entries — this brings the transaction-side walks in line.

This also matches Java, where the transaction/validation walks go through `ManifestReader.liveEntries()` / `ManifestGroup`, which filter DELETED entries by default.

## Behavior changes

Both are bug-fix-shaped, but worth stating plainly:

- Re-adding a path whose only head entry is a tombstone now succeeds instead of failing with a false "already referenced" error.
- Asking the `ReplaceFiles` family to delete such a path now correctly fails with "cannot delete files that do not belong to the table" instead of silently staging a redundant re-delete. This matches Java's `BaseReplaceFiles` validation.

## Tests

- `TestSnapshotDataFilesSkipsDeletedEntries` (new internal test): a snapshot whose data and delete manifests each carry ADDED, EXISTING, and DELETED entries yields only the live ones.
- `TestAddFiles_ReAddsPathDeletedByRewrite`: a rewrite removes a file leaving a tombstone; re-adding the same path succeeds and the file is scannable again.

Both fail on the pre-fix code. `go test ./...` and `golangci-lint run` are clean.

### DIFF
--- a/table/rewrite_files_test.go
+++ b/table/rewrite_files_test.go
@@ -343,6 +343,47 @@ func TestRewriteFiles_DeleteFile_RoutesByContentType(t *testing.T) {
 	assert.NotContains(t, err.Error(), dataSliceMiss)
 }
 
+// TestAddFiles_ReAddsPathDeletedByRewrite pins the duplicate check's
+// tombstone handling: a rewrite that removes a data file leaves a
+// DELETED manifest entry for its path on the branch head. That entry
+// records a removal — the path is no longer referenced by the table —
+// so a subsequent AddFiles for the same path must not report a false
+// "already referenced" conflict.
+func TestAddFiles_ReAddsPathDeletedByRewrite(t *testing.T) {
+	tbl := newRewriteTestTable(t)
+
+	arrowSc, err := table.SchemaToArrowSchema(tbl.Schema(), nil, false, false)
+	require.NoError(t, err)
+	dataPath := tbl.Location() + "/data/seed.parquet"
+	writeParquetFile(t, dataPath, arrowSc, `[{"id": 1, "data": "seed"}]`)
+	seedTx := tbl.NewTransaction()
+	require.NoError(t, seedTx.AddFiles(t.Context(), []string{dataPath}, nil, false))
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Minimal fix, maximal diligence — the call-site audit table showing every `dataFiles` consumer wants live files only is exactly what makes a one-flag change trustworthy, and the rewrite-then-re-add regression captures the real-world failure (compaction tombstones blocking legitimate re-adds) end to end. The hand-built manifest tests distinguishing ADDED/EXISTING from DELETED close the loop at the unit level.

---

PR #1785, opened 2026-08-12
feat(table): RowDelta.RemoveDeletes for deletion-vector supersession

ryanworl:
> [!NOTE]
> Builds on #1783 and #1784 (both merged); rebased onto `main` so the diff is just this PR's four commits.

## What

Adds `RowDelta.RemoveDeletes` (Java's `RowDelta#removeDeletes`): a v3 writer replacing a data file's deletion vector can now remove the old DV entry in the same snapshot that adds the merged replacement. Also adds a `Positions()` iterator to `RoaringPositionBitmap` (Java's `forEach`) so writers can read the old DV's positions to build that replacement. Happy to split `Positions()` into its own PR if preferred.

## Why

The v3 spec allows "at most one deletion vector for a given data file in a snapshot" and requires writers to merge new deletes into the existing vector. Today the public API can only supersede a DV in two commits — add the replacement, then remove the old entry — and the snapshot in between has two live DVs for one data file. A reader can apply the stale DV and resurrect deleted rows, and if the writer crashes between the commits, the table stays that way.

## How it works

- Removals are resolved against the current snapshot's delete manifests, so validation and the DELETED entries use the manifest's own metadata rather than the caller's copy.
- Validation requires each removed file to be a live DV with an explicit `referenced_data_file`, paired with a replacement in the same delta — and the converse: a replacement for a data file whose live DV isn't removed is an error. Duplicate replacements and pre-existing duplicate live DVs are also rejected.
- Deltas with removals route through the overwrite producer (which handles dropping manifest entries); deltas without removals keep the fast-append path unchanged.
- Removal identity is snapshot-relative, so these commits use the no-replay flag from #1784: on a CAS conflict they fail with `ErrCommitFailed` rather than replaying, and the caller rebuilds against a fresh table.

## Out of scope

- Plain `AddDeletes` without removals still doesn't check for an existing live DV. This matches Java, where `validateAddedDVs` only checks *concurrently added* DVs (entries from snapshots newer than the writer's starting snapshot), never the base snapshot's live DVs — keeping the at-most-one-DV invariant is the writer's job in both implementations. Checking the base would also cost a delete-manifest walk on every DV-adding commit.
- The replacement isn't verified to contain the removed DV's positions — writer responsibility, matching Java.

## Tests

End-to-end supersession happy path, a CAS-conflict race asserting no-replay behavior, the full validation matrix (missing replacement, surviving live DV, duplicates, non-DV files, stale references, shared Puffin files, unknown files, v2 tables), and `Positions()` ordering/boundary cases. `go test ./table/...` and `golangci-lint run` are clean.

### DIFF
--- a/table/dv/roaring_bitmap.go
+++ b/table/dv/roaring_bitmap.go
@@ -143,6 +143,11 @@ func (b *RoaringPositionBitmap) Contains(pos uint64) bool {
 // Positions returns an iterator over every set position in ascending
 // order. The positions yielded are the same 64-bit values passed to Set,
 // mirroring Java's RoaringPositionBitmap#forEach.
+//
+// The bitmap must not be modified while iterating: the key set is
+// captured up front while each bucket is consulted lazily, so a
+// concurrent or re-entrant Set/Or produces an inconsistent view (as
+// with the rest of the type, which is not safe for concurrent use).
 func (b *RoaringPositionBitmap) Positions() iter.Seq[uint64] {
 	return func(yield func(uint64) bool) {
 		for _, key := range slices.Sorted(maps.Keys(b.bitmaps)) {
--- a/table/dv/roaring_bitmap.go
+++ b/table/dv/roaring_bitmap.go
@@ -22,6 +22,9 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"iter"
+	"maps"
+	"slices"
 	"sort"
 
 	"github.com/RoaringBitmap/roaring/v2"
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
This completes the v3 DV-supersession contract the right way on every axis: removals resolved against the snapshot's own manifest entries so caller-side copies can neither stale nor forge their way into a wrong pairing, the (path, referenced-data-file) identity that keeps multi-blob Puffin siblings alive with ambiguity rejected rather than guessed, bidirectional replacement validation so neither half of a supersession can commit alone, and the overwrite-producer routing with the default validator suppressed in favor of RowDelta's own — plus #1784's no-replay so a conflicted removal fails loudly instead of stranding double DVs on replay. The doc comments on `RemoveDeletes` and the resolution/validation helpers are reference-quality. Together with #1783/#1784 this is an outstanding series.

---

zeroshade:
@ryanworl Approved — and #1783/#1784 are now squash-merged, so this branch's copies of their commits conflict with `main`. A rebase dropping the merged parent commits should leave just your four; once CI is green it can merge right away.

ryanworl:
Rebased onto `main` — the branch is now just the four commits (the merged parents from #1783/#1784 are dropped). The only conflict was an import collision in `table/dv/roaring_bitmap_test.go` with the new `SetRange`/`RunLengthEncode` tests from #1758. `go test ./table/...` and `golangci-lint run` are clean locally.

ryanworl:
CI note: `ubuntu-latest go1.26.1` failed on `TestSqlCatalog/TestConcurrentTableViewCollisionReturnsCatalogSentinel` in `catalog/sql` — the concurrent create raced into `database is locked (5) (SQLITE_BUSY)` instead of the `table already exists` sentinel. Unrelated to this PR (the diff only touches `table/` and `table/dv/`), and the same test passes on the other matrix jobs. I don't have permission to re-run the job — could a maintainer retrigger it? Happy to file the flake as a separate issue.

PR #1788, opened 2026-08-12
fix(table): validate write default type in update column

### LINKED ISSUE #1787: UpdateSchema: update column does not validate new write-default value against the column type
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Right now, update column accepts a write-default and writes it directly into the field without validating it against the column type. 

- `UnionByName`: does validate the incoming defaults via `validateDefaultValue`, but,
- `UpdateColumn`: doesn't

Thus, the two entry points are inconsistent as well.

What's expected? It should reject a write-default that doesn't match the column type, and clearing a write default works without panicking

### Contribution

Working on a PR for the fix

badalprasadsingh:
### Description

Fixes #1787 

1. Validate the new write-default in update column against the column's type (possibly, promoted) via `validateDefaultValue)`, matching the `UnionByName` path
2. Fix the `nil` pointer panic when clearing the write-default (val: `nil`)

### Testing

Added the regression tests for the same: `valid`/`invalid`/`promoted`/`non-primitive` defaults and default clearing.


zeroshade (approved):
Consistent and complete — routing update-column write-defaults through the same validator as `UnionByName` closes the gap, validating against the promoted type when the type changes in the same update is the subtle case handled and tested in both directions, and the nil-clear panic fix keeps the required-column guard semantics intact. Thanks for the quick pair of fixes today!

---

PR #1794, opened 2026-08-12
fix(partitions): preserve source-less void tombstones

### LINKED ISSUE #1791: BindToSchema rejects accepted source-less void partition tombstones
### Apache Iceberg version

main (development), reproduced at `a56ca68bdf56bbf3181506cbb9a7cb2de482e0e2`; also reproducible in v0.6.0.

### Please describe the bug 🐞

`PartitionField.UnmarshalJSON` intentionally accepts a historical source-less `void` partition tombstone and represents its omitted source as `SourceIDs: []int{0}`. However, `PartitionSpec.BindToSchema` rebuilds every field through `AddPartitionFieldBySourceID`. That attempts to resolve synthetic source ID 0 in the current schema and rejects the same spec that the metadata parser accepted.

This breaks operations that rebind or rebuild an accepted partition spec, including `MetadataBuilder.AddPartitionSpec` and partition evolution paths. `UpdateSpec.Apply` has a second manifestation: retaining a source-less tombstone whose historical partition name matches a live schema column is rejected as a name collision because sentinel source ID 0 differs from the column's real ID.

Minimal reproduction:

```go
package main

import (
	"encoding/json"
	"fmt"

	iceberg "github.com/apache/iceberg-go"
)

func main() {
	schema := iceberg.NewSchema(1, iceberg.NestedField{
		ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64, Required: true,
	})

	for _, raw := range []string{
		`{"spec-id":0,"fields":[{"field-id":1000,"name":"id","transform":"void"}]}`,
		`{"spec-id":0,"fields":[{"source-id":1,"field-id":1000,"name":"id","transform":"void"}]}`,
	} {
		var spec iceberg.PartitionSpec
		parseErr := json.Unmarshal([]byte(raw), &spec)
		_, bindErr := spec.BindToSchema(schema, nil, nil)
		fmt.Printf("parse=%v bind=%v\n", parseErr, bindErr)
	}
}
```

Actual output:

```text
parse=<nil> bind=invalid partition spec: cannot find source column with id: 0 in schema
parse=<nil> bind=<nil>
```

Expected: both accepted specs bind successfully. The source-less `void` field should retain its field ID, name, and transform without trying to resolve synthetic source ID 0 against the schema. Partition evolution should retain the same tombstone, including when its historical name matches a live schema column. Explicit `source-id: 0` and source-less non-`void` fields must remain invalid.

Regression coverage should include:

- source-less `void` parse → `BindToSchema`, with a normal sourced `void` control;
- partition evolution that retains a source-less tombstone while changing an unrelated field;
- the common historical case where the tombstone name matches a live schema column;
- preservation of spec ID, partition field ID, name, and transform;
- continued rejection of explicit zero source IDs and source-less non-`void` transforms.

Related history: #1674 noted this as a separate pre-existing bug but intentionally did not fix it. #1664/#1665 concern explicit ordinal source ID 0 in unbound create-table requests, which is a different path.

### Contribution

Implementation is open in #1794.

mattfaltyn:
## Summary

- preserve historical source-less `void` partition tombstones when binding a partition spec to a schema
- keep those tombstones source-less when marshaling JSON, so accepted metadata round-trips without inventing `source-id: 0`
- allow partition evolution to retain source-less tombstones even when their historical partition name matches a live schema column
- route partition-spec updates through the shared binding path and cover both direct binding and table evolution with regressions

## Problem

`PartitionField.UnmarshalJSON` intentionally accepts historical `void` fields that omit both `source-id` and `source-ids`, representing them internally with the compatibility sentinel `[]int{0}`. `PartitionSpec.BindToSchema` and `UpdateSpec.Apply` then treated that sentinel as a real schema field ID and failed with `field id 0 not found`, preventing otherwise valid metadata from being rebound or evolved. `UpdateSpec.Apply` could also reject a retained tombstone before binding when its historical partition name matched a live schema column.

The fix recognizes only the exact source-less `void` sentinel, preserves it without a schema lookup, and exempts that same compatibility case from the schema-name collision check. Every sourced field, and every non-`void` field, continues through the existing validation. JSON marshaling omits the synthetic sentinel so the accepted historical representation remains round-trippable.

Fixes #1791

## Testing

- `go test . ./table -run 'TestBindToSchemaPreservesSourceLessVoidTombstone|TestUpdateSpecPreservesSourceLessVoidTombstone' -count=2 -v`
- `go test ./... -count=1`
- `go test -race -v ./codec/... ./table/... -count=1`
- `golangci-lint run --timeout=10m` (v2.11.4, 0 issues)
- `go vet ./...`
- `GOOS=linux GOARCH=s390x CGO_ENABLED=0 go build ./...`
- `dev/check-license`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -80,6 +80,15 @@ func (p *PartitionField) EscapedName() string {
 }
 
 func (p PartitionField) MarshalJSON() ([]byte, error) {
+	if len(p.SourceIDs) == 1 && p.SourceIDs[0] == 0 {
+		if _, isVoid := p.Transform.(VoidTransform); isVoid {
+			return json.Marshal(struct {
+				FieldID   int       `json:"field-id"`
+				Name      string    `json:"name"`
+				Transform Transform `json:"transform"`
+			}{p.FieldID, p.Name, p.Transform})
+		}
+	}
 	if len(p.SourceIDs) > 1 {
 		return json.Marshal(struct {
 			SourceIDs []int     `json:"source-ids"`
@@ -208,6 +217,17 @@ func (p *PartitionSpec) BindToSchema(schema *Schema, lastPartitionID *int, newSp
 	}
 
 	for _, field := range p.Fields() {
+		if len(field.SourceIDs) == 1 && field.SourceIDs[0] == 0 {
+			if _, isVoid := field.Transform.(VoidTransform); isVoid {
+				opts = append(opts, func(spec *PartitionSpec) error {
+					spec.fields = append(spec.fields, clonePartitionField(field))
+
+					return nil
+				})
+
+				continue
+			}
+		}
 		opts = append(opts, AddPartitionFieldBySourceID(field.SourceID(), field.Name, field.Transform, schema, &field.FieldID))
 	}
 
--- a/partitions_test.go
+++ b/partitions_test.go
@@ -943,6 +943,34 @@ func TestBindToSchemaAllowsRepeatedVoidTransforms(t *testing.T) {
 	assert.Equal(t, 2, bound.NumFields())
 }
 
+func TestBindToSchemaPreservesSourceLessVoidTombstone(t *testing.T) {
+	schema := iceberg.NewSchema(1,
+		iceberg.NestedField{ID: 1, Name: "s", Type: iceberg.PrimitiveTypes.String, Required: true},
+	)
+
+	data := `{"spec-id":0,"fields":[{"field-id":1000,"name":"old_partition","transform":"void"}]}`
+	var spec iceberg.PartitionSpec
+	require.NoError(t, json.Unmarshal([]byte(data), &spec))
+
+	newSpecID := 2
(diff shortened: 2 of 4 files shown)


zeroshade (approved):
This closes the loop #1665 opened: once explicit `source-id: 0` void fields are rejected at parse, the internal sentinel must marshal back source-less or accepted metadata can't round-trip through our own reader — and the fix threads that needle exactly, preserving tombstones verbatim through `BindToSchema` while keeping them subject to set-level validation, exempting only the sentinel from the schema-name collision, and consolidating `UpdateSpec.Apply` onto the shared binding path with field IDs preserved. The strip-the-source-id metadata test is a particularly faithful reproduction of the historical form.

Coordination note: #1651 (approved, pending its final confirmation) also rewrites `UpdateSpec.Apply` to bind against the staged transaction schema — whichever merges second should rebase carefully, and the staged schema should then feed the `BindToSchema` call added here.

---

PR #1795, opened 2026-08-12
Namespace properties conformance tests

rambleraptor:
Continuing the work on the catalog conformance tests.

This adds namespace properties tests...if the catalog supports those methods. We've got the configuration built-in to the suite, so we just need to add an option.

### DIFF
--- a/catalog/catalogtest/catalogtest.go
+++ b/catalog/catalogtest/catalogtest.go
@@ -82,6 +82,13 @@ type Config struct {
 	// once per test; implementations should register any teardown with
 	// t.Cleanup rather than relying on the suite to release resources.
 	NewCatalog func(t *testing.T) catalog.Catalog
+
+	// SupportsNamespaceProperties reports whether the catalog can store
+	// properties against a namespace. Tests that write properties are skipped
+	// when it is false, as they are for the Hadoop catalog, whose namespaces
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
Ran the suite against the sqlite-backed SQL catalog on the PR head — all new subtests pass — and vet is clean across catalogtest/rest/sql/hive. The tests are careful in the right places: using `assert.Subset`/checking only the created key (catalogs may inject their own properties), asserting `ErrNoSuchNamespace` for the missing-namespace paths (which all three wired catalogs map correctly), and the summary assertions match the shared `getUpdatedPropsAndUpdateSummary` semantics in catalog/catalog.go:358 (only value-changing keys land in `Updated`, only actually-present keys in `Removed`). Gating write tests behind `SupportsNamespaceProperties` while always running `LoadNamespaceProperties` is the right split — the Hadoop catalog's synthetic-location load (catalog/hadoop/hadoop.go:1076) would satisfy the ungated test whenever it gets wired up.

One minor consistency nit, take it or leave it: `testUpdateNamespaceProperties` and `testRemoveNamespaceProperties` use `assert.Equal` on single-element summary slices where the other tests use `ElementsMatch`; harmless with one element, but `ElementsMatch` everywhere would keep the suite robust if these grow.

PR #1797, opened 2026-08-12
fix(partitions): reject redundant time transforms on the same source column

### LINKED ISSUE #1796: Partitions: Spec permits conflicing time transforms on the same source column
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

The validate partition fields function rejects a redundant partition field only when two fields on the same source column carry exactly equal transforms. Thus, `hour(ts)` and `day(ts)` can end up co-existing in one place.

The `addedTimeFields` function cannot catch it either: it is seeded only from fields added in the current update, never from the table's current spec.

### Status

Am already working on the fix, will raise a PR for it.

badalprasadsingh:
### Description

Fixes #1796 

Extends the function `validatePartitionFields` with `partitionFieldsRedundant` folding _year_, _month_, _day_ and _hour_ into one slot per source column.

Other families keep equality-only conflict and stay legal.

### Testing

Added the necessary regression tests for it.

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -551,26 +551,26 @@ func (ps *PartitionSpec) UnmarshalJSON(b []byte) error {
 }
 
 // validatePartitionFields rejects duplicate partition names and redundant
-// fields (the same transform applied to the same source columns). It is the
-// single definition of a well-formed field set, shared by NewPartitionSpecOpts
-// and UnmarshalJSON so the builder cannot produce a spec the parser rejects.
+// fields (transforms that overlap when applied to the same source columns). It
+// is the single definition of a well-formed field set, shared by
+// NewPartitionSpecOpts and UnmarshalJSON so the builder cannot produce a spec
+// the parser rejects.
 //
 // Repeated void fields are exempt: void is the tombstone written for a dropped
 // partition field, so a spec that has dropped several fields legitimately
 // carries more than one.
 //
-// TODO: this does not implement Java's dedupName rule, which collapses year,
-// month, day and hour to a single name per source column and so rejects
-// year(ts) alongside month(ts). We accept that pairing, which means a spec
-// built here can be refused by Java's PartitionSpec.Builder. UpdateSpec.addField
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
This lands the dedup rule cleanly and the authored-vs-replayed split is the right call — strict validation in `NewPartitionSpecOpts`, lenient equality-only in `UnmarshalJSON`/`BindToSchema` so metadata written by other clients (or older iceberg-go) stays readable. That matches Java, where `PartitionSpec.Builder` enforces one time transform per source but existing metadata is loaded via unchecked builds. Test coverage is thorough, including the order-independence of `AddField`/`RemoveField` in `UpdateSpec` and the v1 void-tombstone case.

A few notes, none blocking:

- partitions.go:578-579 — the diff dropped the first half of the nil-transform comment, so it now starts mid-thought at "validateTransform's default branch passes nil through...". Worth restoring the leading sentence ("Reject a nil transform before the redundancy comparison...") or trimming the fragment.
- table/update_spec.go:208 — I want to flag explicitly (as your test `TestUpdateSpecOnInheritedTimeOverlap` documents) that a table whose current spec already carries e.g. day(ts)+hour(ts) now has *all* spec evolution blocked until the redundant field is removed, even unrelated adds/renames. I think that's the correct behavior since it mirrors Java rebuilding the spec through the checked builder, and the remove-first escape hatch is tested, but it is a behavior change for anyone with such tables.
- `TestTimeTransformMembership` in partitions_test.go asserts a Go type assertion rather than exercising any code in this PR; it's a fine canary against someone adding `Transformer` to another transform, just noting it doesn't test the validator itself.
- The moved added-vs-added check means the time-conflict error now surfaces at `BuildUpdates`/`Apply` instead of being recorded during `AddField`; since `us.err` was only ever surfaced at build time anyway, callers see no observable difference — confirmed by the updated `TestUpdateSpecAddField` case.

Built and ran both packages' full test suites on the PR head; all green.

PR #1800, opened 2026-08-13
perf(expr): short-circuit boolean evaluator traversal

fallintoplace:
## Summary

- Add a short-circuit traversal for boolean evaluator visitors.
- Skip the right side of AND when the left side is false.
- Skip the right side of OR when the left side is true.
- Use it for row, manifest, inclusive metrics, and strict metrics evaluation.
- Keep VisitExpr unchanged for visitors that need to see every expression node.

## Benchmark

On an Apple M1 Pro, with a 32-field short-circuit-heavy expression:

- Full traversal: about 1,189 ns/op
- Short-circuit traversal: about 35 ns/op
- About 34x faster
- 0 B/op and 0 allocs/op in both paths

The benchmark is included in visitors_bench_test.go.

## Tests

- go test . ./table
- go test -race . ./table

### DIFF
--- a/table/evaluators.go
+++ b/table/evaluators.go
@@ -76,7 +76,7 @@ func (m *manifestEvalVisitor) Eval(manifest iceberg.ManifestFile) (bool, error) 
 		partitionFilter: m.partitionFilter,
 	}
 
-	result, err := iceberg.VisitExpr(ev.partitionFilter, &ev)
+	result, err := iceberg.VisitExprEvaluator(ev.partitionFilter, &ev)
 	if errors.Is(err, iceberg.ErrInvalidFixedLength) {
 		return rowsMightMatch, nil
 	}
@@ -792,7 +792,7 @@ func (m *inclusiveMetricsEval) TestRowGroup(rgmeta *metadata.RowGroupMetaData, c
 		}
 	}
 
-	result, err := iceberg.VisitExpr(m.expr, m)
+	result, err := iceberg.VisitExprEvaluator(m.expr, m)
 	if errors.Is(err, iceberg.ErrInvalidFixedLength) {
 		return rowsMightMatch, nil
 	}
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
Verified the short-circuit semantics against each converted visitor and this is sound. All four call sites (`exprEvaluator`, `manifestEvalVisitor`, `inclusiveMetricsEval`, `strictMetricsEval`) encode their tri-state results as bools where false = cannot-match / might-not-match and true = might-match / must-match, with pure `&&`/`||` combiners, so skipping the right side of AND-with-false-left and OR-with-true-left is correct for inclusive, strict, and row evaluation alike. Full `go build` and `go test . ./table` pass on the PR head, and the new visited-order tests in visitors_test.go cover both short-circuit and non-short-circuit branches.

Two non-blocking notes:

1. visitors.go (`visitEvaluator`, AND/OR cases): on short-circuit you call `visitor.VisitFalse()`/`VisitTrue()` rather than returning `left` directly. That's only equivalent for visitors where `VisitFalse() == false` and `VisitTrue() == true` — true for all current call sites, and the doc comment does state the contract ("boolean results represent expression truth"), but it might be worth strengthening the comment to say explicitly that `VisitTrue`/`VisitFalse` must return `true`/`false`, since any future `BooleanExprVisitor[bool]` that doesn't satisfy that (e.g. an aggregating visitor) would silently misbehave here.
2. Subtle behavior change worth being aware of: previously, an `ErrInvalidFixedLength` panic from the right subtree caused `Eval` to conservatively return might-match even when the left subtree of an AND was already cannot-match. With short-circuiting, that file is now pruned instead. That's strictly more precise (AND with a cannot-match left is cannot-match regardless of the right side), so it's an improvement, not a regression — just noting it since it can change which files a scan opens for expressions that previously hit that error path.

Nice that the benchmark exercises both short-circuit and no-short-circuit shapes across multiple widths.

PR #1801, opened 2026-08-13
perf(table): avoid quadratic snapshot-log pruning

fallintoplace:
## What changed

- Build one snapshot-ID set in RemoveSnapshots.
- Reuse it for snapshot, snapshot-log, statistics, and partition-statistics pruning.
- Add a benchmark covering larger snapshot logs.

## Benchmark

Same benchmark, 5 runs, 500 ms each, on an Apple M1 Pro:

- 1,000 entries: ~167 µs -> ~15.7 µs, about 10x faster.
- 10,000 entries: ~13.4 ms -> ~171 µs, about 78x faster.

The speedup comes from replacing repeated linear membership scans with map lookups.

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -609,16 +609,18 @@ func (b *MetadataBuilder) RemoveSnapshots(snapshotIds []int64, postCommit bool) 
 
 	b.snapshotList = slices.DeleteFunc(b.snapshotList, func(e Snapshot) bool {
 		_, ok := removedIDs[e.SnapshotID]
-		return ok
-	})
-	b.snapshotLog = slices.DeleteFunc(b.snapshotLog, func(e SnapshotLogEntry) bool {
-		_, ok := removedIDs[e.SnapshotID]
+
 		return ok
 	})
 
-	newRefs := make(map[string]SnapshotRef)
+	validSnapshotIDs := make(map[int64]struct{}, len(b.snapshotList))
+	for _, snapshot := range b.snapshotList {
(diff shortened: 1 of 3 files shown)


table/metadata.go
```diff
@@ -596,20 +596,31 @@ func (b *MetadataBuilder) NextRowID() int64 {
...
 
 	b.snapshotList = slices.DeleteFunc(b.snapshotList, func(e Snapshot) bool {
-		return slices.Contains(snapshotIds, e.SnapshotID)
-	})
-	b.snapshotLog = slices.DeleteFunc(b.snapshotLog, func(e SnapshotLogEntry) bool {
-		return slices.Contains(snapshotIds, e.SnapshotID)
+		_, ok := removedIDs[e.SnapshotID]
+
+		return ok
 	})
```
zeroshade, line 614: looks like you lost the snapshotLog update unless i'm missing something. I don't see `b.snapshotLog` getting updated anywhere anymore.

zeroshade (changes requested):

table/metadata.go
fallintoplace: That delay is intentional: the pruning logic needs to see the removed entry so it can preserve the correct snapshot-history gap behavior. 

The new regression test explicitly verifies that the builder is unchanged before Build(), then the final metadata log is pruned.

fallintoplace: I can add a short comment here to make that less confusing.

zeroshade (approved):

PR #1803, opened 2026-08-13
perf(rest): cache partition decode plans

fallintoplace:
## Summary

- Cache the resolved partition decode plan by spec ID while decoding REST scan tasks.
- Reuse the plan across inline and fetched task envelopes in one remote scan.
- Keep decoded partition values and maps independent for every file.
- Add focused reuse tests and a decoder benchmark.

## Performance

Compared with the parent commit on an Apple M1 Pro, using `GOMAXPROCS=1` and 5 benchmark samples:

- 64 tasks: 365 us -> 241 us, about 1.5x faster
- 1024 tasks: 4.23 ms -> 3.82 ms, about 1.1x faster
- Allocations: 2,049 -> 1,862 per operation, about 9% fewer

The cache is shared across inline and fetched task envelopes, so the setup cost is paid once per spec for a remote scan.

## Testing

- go test ./... -count=1
- go test -race ./catalog/rest -count=1
- go vet ./catalog/rest
- go test ./catalog/rest -run "^$" -bench "^BenchmarkDecodeScanTasks$" -benchmem -benchtime=300ms -count=3

### DIFF
--- a/catalog/rest/scan_planning.go
+++ b/catalog/rest/scan_planning.go
@@ -327,6 +327,7 @@ func (r *Catalog) collectScanTasks(ctx context.Context, ident table.Identifier, 
 // references are indexes into the envelope-local delete-files array.
 func remoteScanTasks(envelopes []ScanTasks, req table.ScanPlanningRequest) ([]table.FileScanTask, error) {
 	var result []table.FileScanTask
+	plans := newPartitionDecodePlanCache()
 	for i, envelope := range envelopes {
 		// Keep empty completed plans cheap and allow the low-level planner method
 		// to return no tasks even when a caller did not provide metadata.
@@ -334,7 +335,7 @@ func remoteScanTasks(envelopes []ScanTasks, req table.ScanPlanningRequest) ([]ta
 			continue
 		}
 
-		tasks, err := DecodeScanTasks(envelope, req.Metadata, req.Schema, req.RowFilter)
+		tasks, err := decodeScanTasks(envelope, req.Metadata, req.Schema, req.RowFilter, plans)
 		if err != nil {
 			return nil, fmt.Errorf("decoding remote scan task envelope %d: %w", i, err)
 		}
--- a/catalog/rest/scan_task_decoder.go
+++ b/catalog/rest/scan_task_decoder.go
@@ -118,6 +118,16 @@ func DecodeScanTasks(
 	schema *iceberg.Schema,
 	fallbackResidual iceberg.BooleanExpression,
 ) ([]table.FileScanTask, error) {
+	return decodeScanTasks(wire, metadata, schema, fallbackResidual, newPartitionDecodePlanCache())
+}
+
+func decodeScanTasks(
+	wire ScanTasks,
+	metadata table.ScanPlanningMetadata,
+	schema *iceberg.Schema,
+	fallbackResidual iceberg.BooleanExpression,
+	plans *partitionDecodePlanCache,
+) ([]table.FileScanTask, error) {
 	if metadata == nil {
 		return nil, fmt.Errorf("%w: decoding scan tasks: table metadata is required", ErrRESTError)
 	}
@@ -139,7 +149,7 @@ func DecodeScanTasks(
 	// entries, including an unreferenced entry that will be rejected below.
 	deletes := make([]iceberg.DataFile, len(wire.DeleteFiles))
 	for i := range wire.DeleteFiles {
-		decoded, err := decodeRESTDeleteFile(wire.DeleteFiles[i], metadata, "")
+		decoded, err := decodeRESTDeleteFile(wire.DeleteFiles[i], metadata, "", plans)
 		if err != nil {
 			return nil, fmt.Errorf("%w: decoding scan tasks: delete-files[%d]: %w", ErrRESTError, i, err)
 		}
@@ -155,7 +165,7 @@ func DecodeScanTasks(
 			return nil, fmt.Errorf("%w: decoding scan tasks: file-scan-tasks[%d] is missing data-file", ErrRESTError, i)
 		}
 
-		dataFile, err := decodeRESTDataFile(*wireTask.DataFile, metadata)
+		dataFile, err := decodeRESTDataFile(*wireTask.DataFile, metadata, plans)
 		if err != nil {
 			return nil, fmt.Errorf("%w: decoding scan tasks: file-scan-tasks[%d].data-file: %w", ErrRESTError, i, err)
 		}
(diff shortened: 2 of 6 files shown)


zeroshade:

catalog/rest/scan_planning.go
```diff
@@ -145,60 +145,47 @@ const headerIdempotencyKey = "Idempotency-Key"
...
-//
-// TODO(#1178): return SupportsFullRemoteScanPlanning() once PlanFiles is wired
-// end-to-end. Until then, callers probing endpoint capability should use
-// SupportsFullRemoteScanPlanning / SupportsPlanTableScan directly.
+// SupportsRemoteScanPlanning reports whether this catalog can submit a remote
+// plan. Any continuation capability is validated against the response returned
+// by the server.
 func (r *Catalog) SupportsRemoteScanPlanning() bool {
-	return false
+	return r.SupportsPlanTableScan()
```
zeroshade, line 183: Blocking: activating REST remote planning here makes the existing `PlanIO` owner leak reachable in normal use. `planFilesRemote` installs an owning `planIOState`, but exhausting the `ReadTasks` iterator releases only its reader lease; the owner is released only by replacement/local replanning through the private `closePlanIO`. `Scan` has no public `Close` and no finalizer, so a one-shot remote scan that reads successfully and is then discarded never invokes `PlanIO.Close`, leaving the cached prefix-scoped cloud filesystems/credential resources alive. Cloning an already planned scan adds more owners that public callers likewise cannot release. The package tests can clean up only by calling the private helper. Please provide a deterministic public lifetime mechanism or redesign ownership so completed/abandoned scans release their plan IO.

catalog/rest/scan_planning.go
```diff
@@ -280,49 +299,53 @@ func (r *Catalog) collectScanTasks(ctx context.Context, ident table.Identifier,
 
 		resp, err := r.FetchScanTasks(ctx, ident, FetchScanTasksRequest{PlanTask: handle})
 		if err != nil {
-			return nil, nil, err
+			return nil, err
 		}
-		files = append(files, resp.FileScanTasks...)
-		deletes = append(deletes, resp.DeleteFiles...)
+		envelopes = append(envelopes, resp.ScanTasks)
```
zeroshade, line 318: Blocking: this accepts a zero response produced by a JSON `null` body and silently drops the plan-task's work. `FetchScanTasks` uses `requireBody`, but that guard rejects only `Content-Length: 0`; `json.Decoder` successfully unmarshals `null` into a zero `FetchScanTasksResponse`. We then append an empty envelope, discover no child handles, and return a successful scan with the original handle's files omitted. A focused HTTP probe returning `200` + `null` confirmed `FetchScanTasks` returns nil error. Please validate the decoded response (or give it strict `UnmarshalJSON` handling) so `null` is `ErrRESTError`, with a regression next to the empty-body test.

zeroshade (changes requested):
Thanks for completing the end-to-end remote scan-planning path. I found two blocking lifecycle/response-validation issues in the newly activated path; details are inline. Separately, all 10 commits in `apache/main..HEAD` are missing the `Signed-off-by` trailer required by `CONTRIBUTING.md`, so the series needs DCO sign-off before merge. Please also update the PR title and description: they currently describe only partition decode-plan caching, while this patch activates and implements remote scan planning across 15 files and roughly 1,400 added lines.

table/scanner.go
```diff
@@ -1231,14 +1344,29 @@ func (scan *Scan) ReadTasks(ctx context.Context, tasks []FileScanTask) (*arrow.S
...
 
 	planIO := scan.planIO
 	scan.planIO = nil
-	planIO.releaseOwner()
+
+	return planIO.releaseOwner()
+}
+
+// Close releases the plan-scoped resources owned by this scan. It is safe to
+// call more than once. Active ReadTasks iterators retain their reader lease and
```
zeroshade, line 1576: The active-iterator guarantee leaves an unreleasable edge: `ReadTasks` increments the reader count before returning its lazy `iter.Seq2`, while `releasePlanIOAfter` runs only if that sequence is invoked. If the caller receives the iterator, never ranges over it, and calls `Scan.Close`, the owner is released but the reader count remains permanently nonzero; `PlanIO.Close` is never called, and `iter.Seq2` has no separate close operation. I reproduced this with a focused test. Please defer acquisition until iteration starts or otherwise provide deterministic abandonment cleanup.

zeroshade (changes requested):
The `null` fetch-response blocker is fixed, and all ten commits now carry the required `Signed-off-by` trailer. The focused REST and table lifecycle suites pass under the race detector, package tests and vet are clean, and CI is green.

The PlanIO lifecycle issue is still not fully resolved. `ReadTasks` acquires a reader lease before returning its lazy iterator, but that lease is released only if the iterator is invoked. If the caller abandons the returned iterator without ranging over it and then calls `Scan.Close`, the owner is released while the reader count remains permanently nonzero, so `PlanIO.Close` is never called. The inline comment includes the focused reproduction.

Please also update the PR title and description to represent the end-to-end REST remote scan-planning implementation rather than only partition decode-plan caching, and reference issue #1178 as required by `CONTRIBUTING.md`.

---

zeroshade (approved):
This PR was reduced to the cache-only change, so I want to be explicit about what happened to my earlier review: my PlanIO lifetime, `null` fetch-response, and end-to-end title/description findings were raised against the much larger earlier version of this PR. That implementation has since landed via merged and split-out work, and none of those files or lines are in this six-file diff. Those findings are therefore no longer in scope rather than outstanding, and I'm clearing my `CHANGES_REQUESTED` accordingly.

On the change that is actually here, the partition decode-plan cache looks correct:

- **Key correctness** — all envelopes in one remote scan share the same immutable table metadata, and Iceberg partition-spec IDs uniquely identify specs within that metadata, so `specID` is a collision-free key for this scope.
- **Lifetime and memory** — `remoteScanTasks` creates one map for the sequential envelope walk and `DecodeScanTasks` creates one per call. It is not process-global: the map is discarded when the decode operation ends and holds only the distinct spec IDs actually encountered, bounded by the finite spec set. No eviction or invalidation is needed for that lifetime.
- **Concurrency** — the cache never escapes the decode call and all access is sequential; concurrent scans get independent caches. No mutex or `sync.Map` required.
- **Isolation** — entries retain only immutable spec/type/logical metadata, while decoded partition values and output maps stay per-file allocations.

Benchmark evidence is present (Apple M1 Pro, `GOMAXPROCS=1`, five samples: 64 tasks 365 µs → 241 µs; 1024 tasks 4.23 ms → 3.82 ms; allocations 2,049 → 1,862) and the patch adds `BenchmarkDecodeScanTasks`. Checked-in raw output or a `benchstat` comparison would be a nice-to-have for reproducibility — explicitly not a condition of this approval.

CI is green across all 15 checks. Thanks for splitting this out; the narrowed scope made it much easier to review.

---

>

PR #1806, opened 2026-08-13
perf(table): index snapshots by ID

fallintoplace:
## Summary

- Index snapshot IDs in table metadata and metadata builders.
- Keep the index in sync when snapshots are added or removed.
- Keep a safe fallback for package-level fixtures that replace the snapshot slice.
- Add regression coverage and a lookup benchmark.

## Performance

Snapshot lookup is roughly 150x faster in the 10,000-snapshot benchmark on an Apple M1 Pro.

| Approach | Result | Relative speed |
| --- | ---: | ---: |
| Indexed lookup | 49 ns/op | ~150x faster |
| Linear scan | 7.8 us/op | Baseline |

This keeps the existing defensive copy behavior for returned snapshots.

## Validation

- go test ./table
- go test -race ./table/...
- go vet ./table
- repository-wide tests for the remaining packages

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2062,12 +2062,12 @@ func (c *commonMetadata) Snapshots() []Snapshot {
 }
 
 func (c *commonMetadata) SnapshotByID(id int64) *Snapshot {
-	if snapshotIndexNeedsRebuild(c.snapshotIndex, c.SnapshotList) {
-		c.snapshotIndex = buildSnapshotIndex(c.SnapshotList)
+	index := c.snapshotIndex
+	if snapshotIndexNeedsRebuild(index, c.SnapshotList) {
+		index = buildSnapshotIndex(c.SnapshotList)
 	}
 
-	i, ok, index := snapshotIndexPosition(c.snapshotIndex, c.SnapshotList, id)
-	c.snapshotIndex = index
+	i, ok, _ := snapshotIndexPosition(index, c.SnapshotList, id)
 	if ok {
 		return cloneSnapshotPtr(&c.SnapshotList[i])
 	}
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -63,9 +63,21 @@ func snapshotIDFromUUID(id uuid.UUID) int64 {
 
 type snapshotIndexData struct {
 	positions map[int64]int
+	// firstSnapshot identifies the snapshot slice used to build positions.
+	// It lets read-only lookups distinguish a complete index from an index
+	// left behind by an in-package fixture that replaced the slice.
+	firstSnapshot *Snapshot
 	// shared means positions is owned by more than one builder or metadata
 	// value and must be copied before a builder mutates it.
 	shared bool
+}
+
+func snapshotListFirst(snapshots []Snapshot) *Snapshot {
+	if len(snapshots) == 0 {
+		return nil
+	}
+
+	return &snapshots[0]
 }
 
 func buildSnapshotIndex(snapshots []Snapshot) *snapshotIndexData {
@@ -76,7 +88,7 @@ func buildSnapshotIndex(snapshots []Snapshot) *snapshotIndexData {
 		}
 	}
 
-	return &snapshotIndexData{positions: positions}
+	return &snapshotIndexData{positions: positions, firstSnapshot: snapshotListFirst(snapshots)}
 }
 
 func cloneSnapshotIndex(index *snapshotIndexData) *snapshotIndexData {
@@ -84,11 +96,18 @@ func cloneSnapshotIndex(index *snapshotIndexData) *snapshotIndexData {
 		return nil
 	}
 
-	return &snapshotIndexData{positions: maps.Clone(index.positions)}
+	return &snapshotIndexData{
+		positions:     maps.Clone(index.positions),
+		firstSnapshot: index.firstSnapshot,
+	}
 }
 
 func snapshotIndexNeedsRebuild(index *snapshotIndexData, snapshots []Snapshot) bool {
-	return index == nil || len(index.positions) != len(snapshots)
+	if index == nil || len(index.positions) != len(snapshots) {
+		return true
+	}
+
+	return len(snapshots) > 0 && index.firstSnapshot != &snapshots[0]
 }
 
 // snapshotIndexPosition returns the position for id. Metadata loaded or built
@@ -106,6 +125,10 @@ func snapshotIndexPosition(index *snapshotIndexData, snapshots []Snapshot, id in
 				return i, true
 			}
 
+			return 0, false
+		}
+
+		if !snapshotIndexNeedsRebuild(index, snapshots) {
 			return 0, false
 		}
 	}
@@ -380,7 +403,6 @@ func (b *MetadataBuilder) clone() *MetadataBuilder {
 		lastPartitionID:      clonePtr(b.lastPartitionID),
 		props:                maps.Clone(b.props),
 		snapshotList:         slices.Clone(b.snapshotList),
-		snapshotIndex:        b.snapshotIndex,
 		currentSnapshotID:    clonePtr(b.currentSnapshotID),
 		snapshotLog:          slices.Clone(b.snapshotLog),
 		metadataLog:          slices.Clone(b.metadataLog),
@@ -396,6 +418,13 @@ func (b *MetadataBuilder) clone() *MetadataBuilder {
 		lastAddedSchemaID:    clonePtr(b.lastAddedSchemaID),
 		lastAddedPartitionID: clonePtr(b.lastAddedPartitionID),
 		lastAddedSortOrderID: clonePtr(b.lastAddedSortOrderID),
+	}
+	if b.snapshotIndex != nil {
+		cloned.snapshotIndex = &snapshotIndexData{
+			positions:     b.snapshotIndex.positions,
+			firstSnapshot: snapshotListFirst(cloned.snapshotList),
+			shared:        true,
+		}
 	}
 	if b.snapshotIndex != nil {
 		b.snapshotIndex.shared = true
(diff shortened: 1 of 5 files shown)


zeroshade:

table/metadata.go
```diff
@@ -319,6 +363,7 @@ func (b *MetadataBuilder) clone() *MetadataBuilder {
 		lastPartitionID:      clonePtr(b.lastPartitionID),
 		props:                maps.Clone(b.props),
 		snapshotList:         slices.Clone(b.snapshotList),
+		snapshotIndex:        maps.Clone(b.snapshotIndex),
```
zeroshade: This adds an O(snapshot-count) map clone to every `MetadataBuilder.clone()`. `Transaction.apply` calls `clone()` for each staged operation, so large snapshot histories pay this cost even when no snapshot lookup occurs. Please make the index lazy/immutable-shareable, or otherwise avoid copying it until the snapshot list is actually mutated.

table/metadata.go
```diff
@@ -1121,6 +1176,7 @@ func (b *MetadataBuilder) buildCommonMetadata() (*commonMetadata, error) {
 		LastPartitionID:    b.lastPartitionID,
 		Props:              b.props,
 		SnapshotList:       b.snapshotList,
+		snapshotIndex:      buildSnapshotIndex(b.snapshotList),
```
zeroshade: This rebuilds the complete index during `Build()` immediately after the transaction staging path cloned the existing index. At 10,000 snapshots, the combined `clone()` + `Build()` path increased locally from 228 µs / 1.34 MB to 494 µs / 1.93 MB. Please avoid this second O(n) construction for unchanged snapshots and add a benchmark that includes index construction rather than only warmed lookups.

zeroshade (changes requested):
Requesting changes for the transaction-path performance regression and the missing DCO trailer.

The warmed lookup benchmark shows a strong improvement, but the eager index work substantially regresses metadata construction and staging with large histories. With 10,000 snapshots, my local `clone()` + `Build()` benchmark changed from 228 µs / 1.34 MB on the merge base to 494 µs / 1.93 MB on this branch. `Transaction.apply` performs that pair for every staged operation. Please avoid cloning and rebuilding the complete map when the snapshot list is unchanged, and add a construction or transaction-path benchmark so this cost remains visible.

Please also add the required `Signed-off-by:` trailer to the commit.

table/metadata.go
```diff
@@ -1964,10 +2062,14 @@ func (c *commonMetadata) Snapshots() []Snapshot {
...
-	for i := range c.SnapshotList {
-		if c.SnapshotList[i].SnapshotID == id {
-			return cloneSnapshotPtr(&c.SnapshotList[i])
-		}
+	if snapshotIndexNeedsRebuild(c.snapshotIndex, c.SnapshotList) {
+		c.snapshotIndex = buildSnapshotIndex(c.SnapshotList)
+	}
+
+	i, ok, index := snapshotIndexPosition(c.snapshotIndex, c.SnapshotList, id)
+	c.snapshotIndex = index
```
zeroshade: **Major:** This assignment makes every steady-state `SnapshotByID` call write to published table metadata, even when `index` is the unchanged existing pointer. Metadata is shared by tables and scans, so parallel calls through `SnapshotByID`, `SnapshotByName`, or `CurrentSnapshot` now race. A focused eight-goroutine test under `go test -race` reported read/write and write/write races at lines 2065, 2069, and 2070. Please keep the normal lookup path read-only—by completing index initialization before publication or using concurrency-safe initialization/repair—and add a concurrent lookup regression test.

zeroshade (changes requested):
Thanks for the update—the previous eager `clone()`/`Build()` performance regression is addressed by the copy-on-write index, the requested construction benchmarks are present, and every commit now has the required `Signed-off-by:` trailer.

One merge-blocking concurrency regression remains, described inline: the hot metadata lookup path now mutates shared metadata and races under concurrent reads.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Iceberg Go maintainer. After you've addressed the point above and pushed an update, an Apache Iceberg Go maintainer — a real person — will take the next look at the PR. If you think the finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>
> *More on how Apache Iceberg Go handles contributions:* [CONTRIBUTING.md](https://github.com/apache/iceberg-go/blob/main/CONTRIBUTING.md).

table/metadata.go
```diff
@@ -1211,10 +1304,13 @@ func (b *MetadataBuilder) GetSortOrderByID(id int) (*SortOrder, error) {
 }
 
 func (b *MetadataBuilder) SnapshotByID(id int64) (*Snapshot, error) {
-	for _, s := range b.snapshotList {
-		if s.SnapshotID == id {
-			return &s, nil
-		}
+	b.ensureSnapshotIndex()
+	i, ok, index := snapshotIndexPosition(b.snapshotIndex, b.snapshotList, id)
+	b.snapshotIndex = index
```
zeroshade: This unconditional assignment makes concurrent read-only calls to the exported `MetadataBuilder.SnapshotByID` race, even when `index` is the unchanged existing pointer. A focused eight-goroutine probe under `go test -race` reports read/write and write/write races between `ensureSnapshotIndex` and this line; the base implementation's linear lookup was read-only. Please keep this lookup path read-only—mirroring `commonMetadata.SnapshotByID` where appropriate—and add a concurrent builder-lookup regression test.

zeroshade (changes requested):
Thanks for the update—the previous published-metadata race and eager clone/build index costs are addressed, the requested benchmarks and regression test are present, and the DCO trailers are complete. One analogous race remains in the exported `MetadataBuilder.SnapshotByID` path, described inline.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Iceberg Go maintainer. After you've addressed the point above and pushed an update, an Apache Iceberg Go maintainer — a real person — will take the next look at the PR. If you think the finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

table/metadata.go
```diff
@@ -335,6 +419,16 @@ func (b *MetadataBuilder) clone() *MetadataBuilder {
...
+	if b.snapshotIndex != nil {
+		cloned.snapshotIndex = &snapshotIndexData{
+			positions:     b.snapshotIndex.positions,
+			firstSnapshot: snapshotListFirst(cloned.snapshotList),
+			shared:        true,
+		}
+	}
+	if b.snapshotIndex != nil {
+		b.snapshotIndex.shared = true
+	}
```
laskoviymishka, line 431: These two `if b.snapshotIndex != nil` blocks always fire together, so I'd merge them into one: build the child struct, then set `b.snapshotIndex.shared = true` right after. Keeping them split means a future CoW field could land in only one block and quietly break the shared invariant, with no compile-time signal to catch it. wdyt?

table/metadata.go
```diff
@@ -1964,10 +2094,14 @@ func (c *commonMetadata) Snapshots() []Snapshot {
 }
 
 func (c *commonMetadata) SnapshotByID(id int64) *Snapshot {
-	for i := range c.SnapshotList {
-		if c.SnapshotList[i].SnapshotID == id {
-			return cloneSnapshotPtr(&c.SnapshotList[i])
-		}
+	index := c.snapshotIndex
```
laskoviymishka, line 2097: The rebuild here goes into a local `index` and never writes back to `c.snapshotIndex`, which is exactly what keeps this path safe under the concurrent published-metadata reads we fixed earlier; writing back would be an unsynchronised mutation again. Could we drop a one-line comment noting the local rebuild is deliberate and must not be cached back? The same shape lives in `MetadataBuilder.SnapshotByID`. Without a note, a future "let's cache the fresh index" edit would quietly reintroduce the race.

table/snapshot_index_test.go
```diff
@@ -0,0 +1,243 @@
...
+func TestCommonMetadataSnapshotLookupsConcurrent(t *testing.T) {
+	currentSnapshotID := int64(2)
+	metadata := &commonMetadata{
+		SnapshotList:      []Snapshot{{SnapshotID: 1}, {SnapshotID: 2}},
+		CurrentSnapshotID: &currentSnapshotID,
+		SnapshotRefs: map[string]SnapshotRef{
+			MainBranch: {SnapshotID: currentSnapshotID, SnapshotRefType: BranchRef},
+		},
+	}
+	metadata.snapshotIndex = buildSnapshotIndex(metadata.SnapshotList)
```
laskoviymishka, line 177: Both concurrent tests call `buildSnapshotIndex` before launching the goroutines, so `snapshotIndexNeedsRebuild` returns false on every goroutine call and the rebuild branch never runs under contention. That's the exact branch the round-2 and round-3 races lived in, so the `-race` guard is weaker than it looks. Could we add a sub-case that leaves `snapshotIndex` nil with a non-empty `SnapshotList` before starting the goroutines, to force the rebuild path concurrently? wdyt?

laskoviymishka (approved):
Thanks for sticking with this through the rounds. The analogous race I flagged last round in `MetadataBuilder.SnapshotByID` is closed now: the read path rebuilds into a local and never writes the fresh index back, same as the `commonMetadata` path, so concurrent reads no longer mutate shared state. That was the last correctness blocker for me, and everything from the earlier rounds has landed:

- clone()+Build() perf regression: resolved via the copy-on-write index, with construction/transaction benchmarks in place so the cost stays visible.
- DCO Signed-off-by trailers: complete on every commit.
- commonMetadata.SnapshotByID published-read race (round 2): resolved by the read-only fallback.
- MetadataBuilder.SnapshotByID race (round 3): resolved the same way.

What's left is polish rather than anything blocking. The two read-only rebuild sites are now load-bearing for the concurrency contract, so I'd add a one-line comment at each saying the local rebuild is deliberate and must not be cached back, otherwise a well-meaning "cache the fresh index" edit walks the race right back in. I'd also merge the two back-to-back `if b.snapshotIndex != nil` blocks in clone() into one, so a future CoW field can't land in only half of them. And the concurrent tests warm the index before spawning goroutines, so they never actually exercise the rebuild branch the earlier races lived in; a nil-index sub-case would tighten that guard. All inline.

Unrelated to your change: the unit-test CI jobs are red on a pre-existing `intrange` lint error in catalog/rest/load_table_bench_test.go, a file this PR doesn't touch. A rebase on main should clear it.

None of that blocks merge from my side. Fix the comments and the test sub-case and I'm happy to approve.

This review was drafted by an AI-assisted tool and confirmed by an Apache Iceberg Go maintainer.

zeroshade (approved):
LGTM now, thanks!

PR #1809, opened 2026-08-13
perf(puffin): reuse prefetched footer payload

fallintoplace:
## What

- Reuse the footer payload from the initial 8 KiB tail read when the whole footer fits.
- Keep the existing `io.SectionReader` fallback for larger footers.
- Add read-count coverage for both paths.

## Why

`readFooter` already reads the tail to avoid extra object-store round trips. Small footers were still decoded from a section reader backed by the underlying `ReaderAt`, so the payload was read again.

## Performance

With a small footer and a simulated 5 ms delay for every range read:

- Before: 17.8 ms/op
- After: 12.2 ms/op
- About 1.5x faster

This removes one redundant range read from the small-footer path.

The benchmark used 5 runs of 20 iterations on an Apple M1 Pro. The exact numbers will vary with object-store latency.

## Tests

- `go test ./puffin`
- `go test -race ./puffin`
- `go test ./...`

### DIFF
--- a/puffin/puffin_reader.go
+++ b/puffin/puffin_reader.go
@@ -205,6 +205,8 @@ func (r *Reader) readFooter() error {
 	// Total footer size: magic(4) + payload + trailer(12)
 	totalFooterSize := MagicSize + payloadSize + footerTrailerSize
 
+	var payloadReader io.Reader
+
 	// Validate footer start magic
 	if totalFooterSize <= readSize {
 		// We already have the footer magic in buf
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. The slice bounds in the prefetch branch check out: since `totalFooterSize <= readSize`, `payloadOffset+int(payloadSize)` lands exactly at `len(buf)-footerTrailerSize`, and the `int(payloadSize)` conversion is safe in that branch because the payload is bounded by the 8 KiB read. The decoded bytes are identical to what the section reader would have produced, so no behavioral change, and the large-footer fallback path is untouched. The read-count tests are a nice touch — I verified the expected counts (read 1 = header magic in NewReader, read 2 = the tail read), and both new tests pass with `-count=1`. One tiny nit, non-blocking: `TestReaderReadsLargeFooterPayloadFromUnderlyingReader` hard-codes read indices 2/3, which would need updating if NewReader's read pattern ever changes, but that's acceptable for what it's asserting.

PR #1810, opened 2026-08-13
perf(table): flatten orphan URI equivalences

### LINKED ISSUE #1808: Table: flatten orphan URI equivalences
The orphan cleanup URI equivalence configuration is currently searched in its raw, comma-separated form for every URI normalization.

For example, with `map[string]string{"s3,s3a,s3n": "s3"}`, `applySchemeEquivalence` scans the configured groups and splits the key each time. `applyAuthorityEquivalence` has the same pattern. On a large table, the same equivalence rules are applied repeatedly to both referenced paths and listed candidates.

This is a small cost per lookup, but orphan cleanup processes many paths. The impact is speed x amount: any speed-up in equivalence lookup is multiplied by the amount of paths compared, and by the number of configured equivalence groups.

### Proposal

Flatten the configured equivalences once when building the cleanup config:

- turn each comma-separated group into direct `scheme -> canonical` and `authority -> canonical` lookups
- preserve the current `WithEqualSchemes` and `WithEqualAuthorities` API
- define behavior for overlapping groups or conflicting canonical values
- use the flattened maps in per-path normalization and prefix-mismatch checks

### Acceptance criteria

- Equivalent schemes and authorities match exactly as they do today.
- Per-path lookup does not repeatedly split or scan the configured groups.
- Tests cover comma-separated groups, single entries, overlapping or conflicting groups, and ADLS `container@host` authorities.
- Add a benchmark showing the effect as the number of paths and equivalence groups grows.

fallintoplace:
Closes #1808

## What changed

- Expand comma-separated scheme and authority groups once when options are applied.
- Use direct map lookups during orphan path normalization and prefix checks.
- Exact single-key mappings take precedence over comma-separated groups.
- Overlapping groups are deterministic: the lexicographically last group wins. This intentionally differs from Java’s input-iteration-order behavior because Go map iteration is unordered.
- Comma-joined group keys are configuration syntax and are not retained as lookup keys.
- Add tests and a benchmark across path and group counts.

## Benchmark

I ran the same benchmark before and after this change on an Apple M1 Pro with GOMAXPROCS=1. These are the medians from 3 runs. ns/op is for one benchmark operation, which processes all paths in that case.

| Paths | Groups | Before | After | Speedup |
| ---: | ---: | ---: | ---: | ---: |
| 100 | 1 | 15,965 ns/op | 908 ns/op | 17.6x |
| 100 | 100 | 480,338 ns/op | 1,258 ns/op | 381.8x |
| 10,000 | 1 | 1,102,166 ns/op | 72,268 ns/op | 15.3x |
| 10,000 | 100 | 51,545,935 ns/op | 103,446 ns/op | 498.3x |

The old implementation also allocated 100, 5,050, 10,000, and 505,611 objects per operation for those four cases. The new implementation reported 0 allocations per operation in every case.

## Tests run

- go test ./... passed
- go vet ./table passed
- GOMAXPROCS=1 go test ./table -run BenchmarkApplyURIEquivalence -benchmem -count=3 -v passed

### DIFF
--- a/table/orphan_cleanup.go
+++ b/table/orphan_cleanup.go
@@ -842,14 +842,44 @@ func normalizeURLPath(path string, cfg *orphanCleanupConfig) string {
 	}
 
 	normalizedScheme := applySchemeEquivalence(parsedURL.Scheme, equalSchemes)
-	normalizedAuthority := applyAuthorityEquivalence(parsedURL.Host, equalAuthorities)
+	normalizedUser, normalizedHost := normalizeURLAuthority(parsedURL, equalAuthorities)
 	normalizedURL := &url.URL{
 		Scheme: normalizedScheme,
-		Host:   normalizedAuthority,
+		User:   normalizedUser,
+		Host:   normalizedHost,
 		Path:   filepath.Clean(parsedURL.Path),
 	}
 
 	return normalizedURL.String()
+}
+
+// completeURLAuthority returns the URI authority, including user-info when it
+// is present. url.URL stores user-info separately from Host, but URI authority
+// equivalence applies to the complete authority (for example,
+// container@account.dfs.core.windows.net).
+func completeURLAuthority(parsedURL *url.URL) string {
+	if parsedURL.User == nil {
+		return parsedURL.Host
+	}
+
+	return parsedURL.User.String() + "@" + parsedURL.Host
+}
+
+func normalizeURLAuthority(parsedURL *url.URL, equalAuthorities map[string]string) (*url.Userinfo, string) {
+	authority := completeURLAuthority(parsedURL)
+	normalizedAuthority := applyAuthorityEquivalence(authority, equalAuthorities)
+	if normalizedAuthority == authority {
+		return parsedURL.User, parsedURL.Host
+	}
+
+	// Parse the canonical authority back into User and Host because url.URL
+	// serializes user-info separately from the host component.
+	canonicalURL, err := url.Parse("//" + normalizedAuthority)
+	if err != nil {
+		return parsedURL.User, parsedURL.Host
+	}
+
+	return canonicalURL.User, canonicalURL.Host
 }
 
 // normalizeNonURLPath provides basic path normalization for non-URL paths.
@@ -943,8 +973,8 @@ func checkPrefixMismatch(referencedPath, filesystemPath string, cfg *orphanClean
 
 	refScheme := applySchemeEquivalence(refURL.Scheme, cfg.equalSchemes)
 	fsScheme := applySchemeEquivalence(fsURL.Scheme, cfg.equalSchemes)
-	refAuth := applyAuthorityEquivalence(refURL.Host, cfg.equalAuthorities)
-	fsAuth := applyAuthorityEquivalence(fsURL.Host, cfg.equalAuthorities)
+	refAuth := applyAuthorityEquivalence(completeURLAuthority(refURL), cfg.equalAuthorities)
+	fsAuth := applyAuthorityEquivalence(completeURLAuthority(fsURL), cfg.equalAuthorities)
 
 	// Check for mismatches
 	schemeMismatch := refScheme != fsScheme
--- a/table/orphan_cleanup.go
+++ b/table/orphan_cleanup.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	stdfs "io/fs"
 	"log/slog"
-	"maps"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


zeroshade (commented):
The flattening approach is correct and the precedence tests are thorough. I verified all production paths go through `newOrphanCleanupConfigWithMode`, so the flatten step is applied consistently, and the raw-option test (`TestOrphanCleanupOptions`) still exercises the pre-flatten maps as intended. Tests pass on the PR head.

A few points I'd like your take on before this leaves draft:

1. `flattenURIEquivalences` (table/orphan_cleanup.go:176) introduces a defined precedence contract — exact keys beat groups, and among overlapping groups the lexicographically last wins. The old code was nondeterministic (map iteration order), so determinism is strictly better, but note this also diverges from Java's `flattenMap`, which is last-put-wins over input iteration order with no exact-key precedence. I think your choice is the more sensible one; just confirm it's intentional and consider stating the divergence in the function comment alongside the existing Java references, since the other helpers in this file cite Java line numbers as the behavioral source of truth.

2. The flattened map retains the original comma-joined group keys (`"s3,s3a,s3n"` stays as a key via the overlay loop). Harmless since a URI scheme/authority can't contain a comma, but it's slightly surprising and shows up in `TestFlattenURIEquivalences`'s expected map. If it's only there to make the overlay loop simpler, a short comment (or skipping comma keys in the overlay) would help future readers.

3. Nit: the license header in table/orphan_cleanup_bench_test.go has a blank line after the license URL, splitting it into two comment blocks — the other files in the package use one contiguous block. Worth fixing so RAT/header tooling doesn't trip on it.

4. Benchmark nit: the `result == ""` guard at the end can never trigger (`applySchemeEquivalence` always returns a non-empty string for non-empty input), so it's dead code — fine to keep as a sink to prevent dead-code elimination, but the `b.Fatal` implies it's a real assertion.

None of these are blockers; happy to re-review once it's marked ready.

zeroshade:
Since this is a draft, i'll hold off on further review until it is marked ready

table/orphan_cleanup.go
```diff
@@ -888,24 +917,10 @@ func applySchemeEquivalence(scheme string, equalSchemes map[string]string) strin
 // Based on Apache Iceberg Java's equalAuthorities logic (lines 546, 161-165, 392-403).
 // https://github.com/apache/iceberg/blob/07c088fce9c54369864dcb6da16006e78206048b/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/DeleteOrphanFilesSparkAction.java#L1
 func applyAuthorityEquivalence(authority string, equalAuthorities map[string]string) string {
-	if equalAuthorities == nil {
-		return authority
-	}
-
 	if canonical, exists := equalAuthorities[authority]; exists {
```
zeroshade, line 950: **Blocking:** `normalizeURLPath` and `checkPrefixMismatch` pass `parsedURL.Host` here, but Go's `net/url` separates `abfs://container@host/...` into `User=container` and `Host=host`. As a result, flattened keys such as `container@account-a.dfs.core.windows.net` never match, and `normalizeURLPath` also reconstructs the URL without `User`, dropping the container. A direct probe with two configured `container@host` authorities normalized them to different URLs (`abfs://account-a...` vs `abfs://account-b...`). Please perform authority lookup on the complete authority (`User@Host`) and preserve/reconstruct user-info when normalizing; the same helper should be used by the prefix-mismatch path.

table/orphan_cleanup_test.go
```diff
@@ -86,6 +86,78 @@ func TestOrphanCleanupOptions(t *testing.T) {
...
+	assert.Equal(t, "canonical", applyAuthorityEquivalence("host2", cfg.equalAuthorities))
+}
+
+func TestNewOrphanCleanupConfigFlattensURIEquivalences(t *testing.T) {
+	cfg := newOrphanCleanupConfig(
+		WithEqualSchemes(map[string]string{
+			"s3,s3a": "s3",
+		}),
+		WithEqualAuthorities(map[string]string{
+			"host1,host2": "canonical",
```
zeroshade, line 151: Could you make this exercise the required ADLS behavior through `normalizeURLPath` (and ideally `checkPrefixMismatch`) using `abfs://container@account-*.dfs.core.windows.net/...`, rather than only calling `applyAuthorityEquivalence` directly? That regression test currently fails because `net/url` removes `container@` from `URL.Host`, which is the gap hidden by these host-only values.

zeroshade (changes requested):
Requesting changes for the ADLS authority correctness issue described inline. The flattening implementation and existing tests otherwise look good, but the linked issue explicitly requires coverage for ADLS `container@host` authorities, and that production path currently fails.

zeroshade (approved):

PR #1811, opened 2026-08-13
perf(io): delete GoCloud files concurrently

fallintoplace:
## What changed

- Delete files with up to 16 workers in `BlobFileIO.DeleteFiles`.
- Keep best-effort deletes, missing-file handling, and joined errors.
- Keep deleted paths and errors in input order.
- Add a test for the concurrency cap and result ordering.

## Why

`DeleteFiles` was deleting one object at a time. This makes cleanup slow when a cloud store has many files.

## Speed

Local synthetic benchmark on an Apple M1 Pro:

- 64 files
- 5 ms simulated delete latency per file
- serial: 386.6 ms
- concurrent: 24.0 ms
- speedup: 16.1x

This measures request latency hiding. It is not a provider-specific benchmark.

## Tests

- `go test -p 2 ./...`
- `go test -race ./io/gocloud -run TestBlobFileIODeleteFiles -count=1`
- `go vet ./io/gocloud`

### DIFF
--- a/io/gocloud/blob.go
+++ b/io/gocloud/blob.go
@@ -26,6 +26,7 @@ import (
 	pathpkg "path"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	icebergio "github.com/apache/iceberg-go/io"
@@ -187,6 +188,10 @@ type BlobFileIO struct {
 }
 
 var _ icebergio.ListableIO = (*BlobFileIO)(nil)
+
+// deleteFilesMaxConcurrency bounds the number of in-flight object-store
+// deletes without creating one goroutine per path for large cleanup jobs.
+const deleteFilesMaxConcurrency = 16
 
 type blobFileInfo struct {
 	name    string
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. The worker-pool implementation is correct: results are written to distinct indices of a pre-sized slice so there's no data race, deleted paths and joined errors preserve input order, and the NotFound-is-success and empty-slice semantics from the `BulkRemovableIO` contract in io/io.go:114-126 are preserved. I verified the new `TestBlobFileIODeleteFilesIsConcurrentAndBounded` is deterministic (workers block in `Delete` until released, so the max-in-flight assertion can't flake) and it passes repeatedly under `-race`, along with the existing DeleteFiles tests.

- io/gocloud/blob.go:456-459 could use the `min()` builtin instead of the manual clamp.
- Elsewhere the codebase reaches for `errgroup` with `SetLimit` (e.g. table/scanner.go:751, table/arrow_utils.go:1722); the manual pool is justified here since best-effort semantics conflict with errgroup's cancel-on-first-error, but a one-line comment saying so would save the next reader from "simplifying" it back to errgroup.

The hard-coded cap of 16 seems a reasonable default; if it ever needs tuning per store we can make it a property later.

PR #1813, opened 2026-08-13
perf(table): precompute data writer invariants

fallintoplace:
## What changed

- Precompute the stats plan, current partition spec, Parquet row-group size, write properties, schema options, and file extension when the writer is created.
- Reuse those values for every write task.
- Add a focused writer benchmark.

## Benchmark

Apple M1 Pro, `GOMAXPROCS=1`, median of 5 runs with `-benchtime=1s`:

- `origin/main`: 8.53 us/op, 97 allocs/op, 6.49 KB/op
- this branch: 6.38 us/op, 72 allocs/op, 5.01 KB/op
- about 1.34x faster and 26% fewer allocations

The benchmark uses a no-op file format so it measures the per-task writer setup and schema conversion without filesystem or Parquet I/O.

## Tests

- `go test ./...`
- `go test -race ./table/...`
- `go vet ./table/...`

### DIFF
--- a/table/writer.go
+++ b/table/writer.go
@@ -64,8 +64,19 @@ type defaultDataFileWriter struct {
 	format           tblutils.FileFormat
 	props            iceberg.Properties
 	content          iceberg.ManifestEntryContent
-	meta             *MetadataBuilder
 	equalityFieldIDs []int
+	invariants       dataWriterInvariants
+}
+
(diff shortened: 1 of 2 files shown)


zeroshade:

zeroshade (approved):

PR #1816, opened 2026-08-13
fix(table): look up removed-file partition specs by ID

### LINKED ISSUE #1815: Snapshot summary: Delete/Overwrite panics when a removed file's spec ID isn't its slice position
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

`snapshotProducer.accumulateSummaryDelta` resolves partition specs two different ways depending on the branch. Added files go through `sp.spec(int(df.SpecID()))`, which is a by-ID lookup on top of `MetadataBuilder.GetSpecByID`. Removed data files, delete files and deletion vectors instead index the spec slice directly, `specs[df.SpecID()]`, in three places.

Spec ID is an identifier, not a slice position, and the two come apart on ordinary tables:

- `MetadataBuilder.RemovePartitionSpecs` filters the slice and never renumbers the survivors, so removing a non-default spec that sits before the default one leaves `specs[i].ID() != i`.
- Specs loaded from existing metadata keep the order of the `partition-specs` JSON array (`slices.Clone`, no sort). The spec format identifies each entry by `spec-id` and does not promise ascending order, so metadata written by another engine can also break the `index == id` assumption. Java is unaffected because it keys specs by ID in a `Map<Integer, PartitionSpec>`.

`specs[df.SpecID()]` is evaluated as a call argument, so it fails before `removeFile` ever runs. On such a table, any `Delete` / `Overwrite` / `RewriteFiles` commit that removes files panics with `index out of range`, or — when the ID happens to stay within bounds but points at the wrong entry — silently records the changed-partition summary under a partition path derived from the wrong spec.

**Reproduction**, using only public metadata updates: evolve to spec ids 0/1/2, make 2 the default, then `RemovePartitionSpecs([]int{1})` so `specs == [id0, id2]` (len 2). Now feed the *same* spec-2 data file to both branches of the function:

| path | result |
| --- | --- |
| add | `partitions.id_trunc=0`, no error |
| remove | `panic: runtime error: index out of range [2] with length 2` |

Same file, same spec, same function — the only variable is which branch it goes down.

```
panic: runtime error: index out of range [2] with length 2
github.com/apache/iceberg-go/table.(*snapshotProducer).accumulateSummaryDelta
	table/snapshot_producers.go:1023
```

Expected: removed files resolve their spec by ID, exactly as added files already do.

### Contribution

Working on a PR for the fix (regression test included).

codeAnqiang-ma:
Fixes #1815

`accumulateSummaryDelta` resolves partition specs two different ways. Added files use `sp.spec(int(df.SpecID()))`, a by-ID lookup over `GetSpecByID`; removed data files, delete files and deletion vectors index the slice directly, `specs[df.SpecID()]`.

Spec ID is an identifier, not a slice position. `RemovePartitionSpecs` filters the slice without renumbering the survivors, and specs loaded from metadata keep the `partition-specs` array order (no sort), so positions and IDs come apart on ordinary tables. Because `specs[df.SpecID()]` is evaluated as a call argument, it fails before `removeFile` ever runs: any Delete/Overwrite/RewriteFiles commit that removes files panics with `index out of range`, or — when the ID stays in bounds but points at the wrong entry — silently attributes the change to the wrong partition.

`RemovePartitionSpecs([]int{1})` over specs 0/1/2 leaves `specs == [id0, id2]`. Sending the *same* spec-2 data file down both branches:

| path | result |
| --- | --- |
| add | `partitions.id_trunc=0` |
| remove | `panic: runtime error: index out of range [2] with length 2` |

Same file, same spec, same function; only the branch differs.

The fix calls `sp.spec(...)` in the three removal loops and drops the now-unused local. Unknown IDs then yield an empty unpartitioned spec instead of panicking, matching the add path.

<details>
<summary>Test evidence</summary>

Regression test `TestAccumulateSummaryDeltaResolvesRemovedFileSpecByID` in `table/snapshot_producers_test.go`. It asserts the add path first, so the failure output itself shows the asymmetry rather than just "remove panics".

Before the fix:

```
=== RUN   TestAccumulateSummaryDeltaResolvesRemovedFileSpecByID
--- FAIL: TestAccumulateSummaryDeltaResolvesRemovedFileSpecByID (0.00s)
panic: runtime error: index out of range [2] with length 2 [recovered, repanicked]
...
github.com/apache/iceberg-go/table.(*snapshotProducer).accumulateSummaryDelta(...)
	table/snapshot_producers.go:1023 +0x5e8
github.com/apache/iceberg-go/table.TestAccumulateSummaryDeltaResolvesRemovedFileSpecByID(...)
	table/snapshot_producers_test.go:1115 +0x950
FAIL	github.com/apache/iceberg-go/table	0.933s
```

After the fix:

```
=== RUN   TestAccumulateSummaryDeltaResolvesRemovedFileSpecByID
--- PASS: TestAccumulateSummaryDeltaResolvesRemovedFileSpecByID (0.00s)
ok  	github.com/apache/iceberg-go/table	0.922s
```

| check | result |
| --- | --- |
| `go test ./...` | all 26 packages with tests ok |
| `go test -race ./codec/... ./table/...` | ok |
| `golangci-lint run` (v2.11.4, whole repo) | 0 issues |
| `gofmt -l` | clean |

Not run: the Docker-gated integration suite (`go test -tags integration ./...`) and the s390x cross-compile job. The change is confined to in-memory summary accounting, with no catalog, IO or platform-specific code involved.

</details>

Assisted by Cursor; I reproduced the panic locally and reviewed every line of the change.

### DIFF
--- a/table/snapshot_producers.go
+++ b/table/snapshot_producers.go
@@ -1018,9 +1018,8 @@ func (sp *snapshotProducer) accumulateSummaryDelta(countDeleteRemoval func(icebe
 		}
 	}
 
-	specs := sp.txn.meta.specs
 	for _, df := range sp.deletedFiles {
-		if err = ssc.removeFile(df, currentSchema, specs[df.SpecID()]); err != nil {
+		if err = ssc.removeFile(df, currentSchema, sp.spec(int(df.SpecID()))); err != nil {
 			return nil, err
 		}
 	}
@@ -1028,7 +1027,7 @@ func (sp *snapshotProducer) accumulateSummaryDelta(countDeleteRemoval func(icebe
 		if countDeleteRemoval != nil && !countDeleteRemoval(df) {
 			continue
 		}
-		if err = ssc.removeFile(df, currentSchema, specs[df.SpecID()]); err != nil {
+		if err = ssc.removeFile(df, currentSchema, sp.spec(int(df.SpecID()))); err != nil {
 			return nil, err
 		}
 	}
@@ -1036,7 +1035,7 @@ func (sp *snapshotProducer) accumulateSummaryDelta(countDeleteRemoval func(icebe
 		if countDeleteRemoval != nil && !countDeleteRemoval(df) {
 			continue
 		}
-		if err = ssc.removeFile(df, currentSchema, specs[df.SpecID()]); err != nil {
+		if err = ssc.removeFile(df, currentSchema, sp.spec(int(df.SpecID()))); err != nil {
 			return nil, err
 		}
 	}
--- a/table/snapshot_producers_test.go
+++ b/table/snapshot_producers_test.go
@@ -26,6 +26,7 @@ import (
 	"io/fs"
 	"runtime"
 	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
Verified locally: the three removal loops in `accumulateSummaryDelta` (table/snapshot_producers.go:1021-1041 on main) index `sp.txn.meta.specs` positionally while the add loops resolve by ID via `sp.spec()`, so any commit removing files after `RemovePartitionSpecs` (or on metadata whose `partition-specs` order doesn't match IDs) panics or misattributes the partition. Switching the removal paths to `sp.spec(int(df.SpecID()))` makes both directions symmetric, including the fall-back to an empty unpartitioned spec for unknown IDs, which is the established behavior of the add path. The regression test is well constructed — asserting the add path first so the add/remove asymmetry is visible, and covering the exact spec-removal scenario that decouples positions from IDs. I ran the test on the PR head and it passes. LGTM.

PR #1817, opened 2026-08-13
perf(transform): hash decimal buckets without allocating

fallintoplace:
## What

- Hash decimal bucket values directly from decimal128 signed two-complement bytes.
- Keep DecimalLiteral.MarshalBinary unchanged.
- Add boundary and random compatibility tests.
- Add focused benchmarks.

## Why

Decimal bucketing used to build a big.Int and an allocated byte slice for every value.

## Benchmark

Apple M1 Pro, arm64. Inputs are pre-boxed:

| Path | Before | After | Result |
|---|---:|---:|---|
| Bucket Apply | 124.9 ns, 97 B, 4 allocs | 17.3 ns, 0 B, 0 allocs | 7.2x faster, zero allocations |
| Bound Transformer | 126.3 ns, 97 B, 4 allocs | 13.5 ns, 0 B, 0 allocs | 9.4x faster, zero allocations |

## Tests

- go test ./...
- go vet .

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -375,6 +375,28 @@ func hashTimestampNano(v any) uint32 {
 	return hashHelperInt[int64](micros)
 }
 
+func hashDecimal(v Decimal) uint32 {
+	// Decimal bucket transforms hash the minimum two's-complement
+	// big-endian representation of the unscaled value. Keep the full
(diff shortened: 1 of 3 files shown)


zeroshade:

zeroshade (approved):

PR #1820, opened 2026-08-13
perf(compaction): compute survivor minimum once

fallintoplace:
## Summary

- Compute the survivor minimum sequence once per decision.
- Avoid repeating the partition-minimum scan for every equality-delete candidate.
- Work changes from O(P x D) to O(P + D).
- Add a scaling benchmark.

## Benchmark

Apple M1 Pro, arm64. 1,000 partitions and 1,000 equality-delete candidates:

| | Before | After |
|---|---:|---:|
| Time | 14.999 ms/op | 65.7 us/op |
| Bytes | 25,223,048 B/op | 96,144 B/op |
| Allocs | 9,006 allocs/op | 15 allocs/op |

That is about 228x faster, with 99.6% fewer bytes and 99.8% fewer allocations.

At 10,000 partitions and 1,000 candidates, time drops from 149.7 ms/op to 254 us/op.

## Tests

- go test ./...
- go vet .

### DIFF
--- a/table/compaction/eq_delete_decision.go
+++ b/table/compaction/eq_delete_decision.go
@@ -138,6 +138,7 @@ func DecideDeadEqualityDeletes(survey *SurvivorSurvey, candidates []iceberg.Mani
 
 	dead := make([]iceberg.DataFile, 0, len(candidates))
 	seen := make(map[string]struct{}, len(candidates))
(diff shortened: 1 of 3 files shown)


zeroshade:

zeroshade (approved):

PR #1822, opened 2026-08-13
perf(table): decode metadata updates in one pass

fallintoplace:
## Summary

- Reuse one raw field map for update action dispatch and required-field validation.
- Keep legacy property field normalization working.
- Avoid repeated JSON object decoding for every update.
- Add canonical and legacy update benchmarks.

## Benchmark

Apple M1 Pro, arm64. 1,000 updates:

| Payload | Before | After | Result |
|---|---:|---:|---|
| Canonical | 4.04 ms, 1.71 MB, 36,149 allocs | 2.96 ms, 1.39 MB, 26,897 allocs | 27% faster, 18% fewer bytes, 26% fewer allocs |
| Legacy properties | 4.19 ms, 1.71 MB, 36,151 allocs | 3.36 ms, 1.45 MB, 28,400 allocs | 20% faster, 16% fewer bytes, 21% fewer allocs |

## Tests

- go test ./...
- go vet .

### DIFF
--- a/table/updates.go
+++ b/table/updates.go
@@ -100,7 +100,7 @@ var requiredUpdateFields = map[string][]string{
 	UpdateRemoveEncryptionKey:       {"key-id"},
 }
 
-func normalizeLegacyPropertyFields(action string, raw json.RawMessage) (json.RawMessage, error) {
+func normalizeLegacyPropertyFields(action string, object map[string]json.RawMessage) bool {
 	var modern, legacy string
 	switch action {
 	case UpdateSetProperties:
@@ -108,33 +108,28 @@ func normalizeLegacyPropertyFields(action string, raw json.RawMessage) (json.Raw
 	case UpdateRemoveProperties:
 		modern, legacy = "removals", "removed"
 	default:
-		return raw, nil
+		return false
 	}
 
-	var object map[string]json.RawMessage
-	if err := json.Unmarshal(raw, &object); err != nil {
-		return nil, err
-	}
-
+	normalized := false
 	if _, ok := object[modern]; !ok {
 		if value, ok := object[legacy]; ok {
 			object[modern] = value
+			normalized = true
 		}
 	}
-	delete(object, legacy)
+	if _, ok := object[legacy]; ok {
+		delete(object, legacy)
+		normalized = true
+	}
 
-	return json.Marshal(object)
+	return normalized
 }
 
-func validateRequiredUpdateFields(action string, raw json.RawMessage) error {
+func validateRequiredUpdateFields(action string, object map[string]json.RawMessage) error {
 	fields := requiredUpdateFields[action]
 	if len(fields) == 0 {
 		return nil
-	}
-
-	var object map[string]json.RawMessage
-	if err := json.Unmarshal(raw, &object); err != nil {
-		return err
 	}
 
 	for _, field := range fields {
@@ -158,17 +153,20 @@ func (u *Updates) UnmarshalJSON(data []byte) error {
 		updates = make(Updates, 0, len(rawUpdates))
 	}
 	for _, raw := range rawUpdates {
-		var baseWire struct {
-			Action *string `json:"action"`
-		}
-		if err := json.Unmarshal(raw, &baseWire); err != nil {
+		var object map[string]json.RawMessage
+		if err := json.Unmarshal(raw, &object); err != nil {
 			return err
 		}
-		if baseWire.Action == nil {
+		actionValue, ok := object["action"]
+		if !ok || bytes.Equal(bytes.TrimSpace(actionValue), []byte("null")) {
 			return fmt.Errorf("%w: update requires field %q", iceberg.ErrInvalidArgument, "action")
 		}
+		var action string
+		if err := json.Unmarshal(actionValue, &action); err != nil {
+			return err
+		}
 
-		base := baseUpdate{ActionName: *baseWire.Action}
+		base := baseUpdate{ActionName: action}
 
 		var upd Update
 		switch base.ActionName {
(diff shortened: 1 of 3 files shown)


zeroshade:

table/updates.go
```diff
@@ -158,17 +153,20 @@ func (u *Updates) UnmarshalJSON(data []byte) error {
...
-		var baseWire struct {
-			Action *string `json:"action"`
-		}
-		if err := json.Unmarshal(raw, &baseWire); err != nil {
+		var object map[string]json.RawMessage
+		if err := json.Unmarshal(raw, &object); err != nil {
 			return err
 		}
-		if baseWire.Action == nil {
+		actionValue, ok := object["action"]
```
zeroshade: Could we keep action dispatch aligned with the final tagged-struct decode? This exact map lookup no longer accepts case-insensitive action keys that `encoding/json` previously accepted. More seriously, for `{"action":"set-location","ACTION":"unknown","location":"s3://bucket/table"}`, this lookup dispatches to `setLocationUpdate`, but the later `json.Unmarshal(raw, upd)` applies case-insensitive struct matching and overwrites the embedded `ActionName` with `"unknown"`. The result is a concrete set-location update whose `Action()` reports `unknown`; the merge base rejects that payload as an unknown action. Please use identical action-key semantics for dispatch and final decode, with regression tests for mixed-case and duplicate case-variant keys.

zeroshade (changes requested):
The one-pass refactor currently changes `encoding/json` action-field semantics and can produce an update whose concrete type disagrees with `Update.Action()`. Please make dispatch and final decoding use the same action interpretation and add regression coverage for mixed-case and duplicate case-variant action keys. Both commits also need the `Signed-off-by:` trailer required by `CONTRIBUTING.md`.

The existing update tests, race checks, full table suite, lint, formatting, and CI are green, and the benchmarked performance improvement is confirmed locally.

table/updates.go
```diff
@@ -158,17 +154,45 @@ func (u *Updates) UnmarshalJSON(data []byte) error {
...
+				actionFieldCount++
+				if field != "action" {
+					caseVariantAction = true
+				}
+			}
 		}
 
-		base := baseUpdate{ActionName: *baseWire.Action}
+		var action string
+		if exactAction && !caseVariantAction {
```
zeroshade, line 183: **Major: exact duplicate `action` keys can still erase an `encoding/json` type error.** By this fast path, decoding into `object` has already collapsed duplicate exact keys. For `[{"action":123,"action":"set-properties","updated":{"k":"v"}}]`, the merge base returns `*json.UnmarshalTypeError`, but the current code selects the final string, legacy-normalizes and re-marshals the map, then succeeds because the earlier invalid number is gone. Please preserve the discriminator's original decode error before any map round-trip and add this payload as a regression test.

zeroshade (changes requested):
Thanks for addressing the prior action-dispatch inconsistency and adding the missing DCO trailers. Mixed-case and distinct case-variant action fields now dispatch and decode consistently, but one `encoding/json` compatibility regression remains.

### Exact duplicate `action` keys can erase a type error (`table/updates.go:175`)

For this payload:

```json
[{"action":123,"action":"set-properties","updated":{"k":"v"}}]
```

the merge base returns `*json.UnmarshalTypeError`, while the current implementation accepts it and returns a valid `set-properties` update. The initial map decode collapses the duplicate exact keys to the final string, and legacy-property normalization re-marshals that map before the concrete decode can observe the earlier invalid value. Please preserve the discriminator decode error before that map round-trip and add this case as a regression test.

All 15 CI checks pass. I also verified the targeted update tests, full table suite, race tests, vet, clean merge with current `main`, and that the performance improvement remains.

---

>
> [contributing-docs/05_pull_requests.rst](https://github.com/apache/iceberg-go/blob/main/contributing-docs/05_pull_requests.rst).

table/updates.go
```diff
@@ -221,12 +245,32 @@ func (u *Updates) UnmarshalJSON(data []byte) error {
 		default:
 			return fmt.Errorf("%w: unknown update action: %s", iceberg.ErrInvalidArgument, base.ActionName)
 		}
-		normalized, err := normalizeLegacyPropertyFields(base.ActionName, raw)
-		if err != nil {
-			return err
+		if normalizeLegacyPropertyFields(base.ActionName, object) {
+			var actionWire struct {
```
zeroshade: **Major:** This raw discriminator check is reached only after successful action dispatch and only for legacy property normalization, so exact duplicate-key type errors still disappear on earlier exits. For both `{"action":123,"action":"unknown"}` and `{"action":123,"action":null}`, the merge base returns `*json.UnmarshalTypeError`; this head instead returns “unknown update action” or the missing/null-action `ErrInvalidArgument` because the map decode collapsed the earlier number before execution reaches here. Please validate the original discriminator semantics before the null and unknown-action exits, and extend the regression test across valid, unknown, and null final values.

zeroshade (changes requested):
Thanks for adding the duplicate-action regression test. The reported valid-final-action payload now preserves its `encoding/json` type error, but the same compatibility issue remains when the collapsed final action takes an earlier exit as unknown or `null`, as described inline.

All CI checks and the regular update/table/race suites pass, but a focused comparison still differs from the merge base for those payloads.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Iceberg Go maintainer. After you've addressed the point above and pushed an update, an Apache Iceberg Go maintainer — a real person — will take the next look at the PR. If you think the finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>
> *More on how Apache Iceberg Go handles contributions:* [CONTRIBUTING.md](https://github.com/apache/iceberg-go/blob/main/CONTRIBUTING.md).

zeroshade (approved):

zeroshade:
just need to fix the linting issue

PR #1823, opened 2026-08-13
perf(table): decode partition and sort fields once

fallintoplace:
## Summary

- Decode partition and sort field objects into one auxiliary value.
- Preserve presence checks for source-id, source-ids, and transform.
- Avoid the temporary raw map and second full-object decode.
- Add single-source and multi-source benchmarks.

## Benchmark

Apple M1 Pro, arm64:

| Decoder | Case | Before | After | Result |
|---|---|---:|---:|---|
| Partition | Single source | 4,254 ns, 1,344 B, 30 allocs | 2,168 ns, 976 B, 19 allocs | 49% faster |
| Partition | Multi source | 4,467 ns, 1,408 B, 34 allocs | 2,735 ns, 1,032 B, 23 allocs | 39% faster |
| Sort | Single source | 3,508 ns, 1,360 B, 31 allocs | 2,234 ns, 1,000 B, 20 allocs | 36% faster |
| Sort | Multi source | 3,977 ns, 1,424 B, 35 allocs | 2,627 ns, 1,048 B, 24 allocs | 34% faster |

Bytes and allocations drop by about 26% and 31 to 37%.

## Tests

- go test ./...
- go vet .

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -135,47 +135,55 @@ func (p *PartitionField) UnmarshalJSON(b []byte) error {
 }
 
 func (p *PartitionField) unmarshal(b []byte, binding specBinding) error {
-	var aux struct {
-		SourceID  json.RawMessage `json:"source-id"`
-		SourceIDs json.RawMessage `json:"source-ids"`
-		FieldID   int             `json:"field-id"`
-		Name      string          `json:"name"`
-		Transform json.RawMessage `json:"transform"`
-	}
-	if err := json.Unmarshal(b, &aux); err != nil {
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(b, &raw); err != nil {
 		return fmt.Errorf("%w: failed to unmarshal partition field", err)
 	}
 
-	hasSourceID := aux.SourceID != nil
-	hasSourceIDs := aux.SourceIDs != nil
+	_, hasSourceID := raw["source-id"]
+	_, hasSourceIDs := raw["source-ids"]
 	if hasSourceID && hasSourceIDs {
 		return fmt.Errorf("%w: partition field cannot contain both source-id and source-ids", ErrInvalidPartitionSpec)
 	}
 
-	if aux.Transform == nil || string(aux.Transform) == "null" {
+	transformJSON, hasTransform := raw["transform"]
+	if !hasTransform || string(transformJSON) == "null" {
 		return fmt.Errorf("%w: partition field requires a transform", ErrInvalidTransform)
 	}
 
 	var sourceID int
 	if hasSourceID {
-		if err := json.Unmarshal(aux.SourceID, &sourceID); err != nil {
+		if err := unmarshalJSONField(raw["source-id"], "source-id", &sourceID); err != nil {
 			return err
 		}
 	}
 	var sourceIDs []int
 	if hasSourceIDs {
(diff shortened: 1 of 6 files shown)


zeroshade:

partitions.go
```diff
@@ -122,32 +122,41 @@ func (p *PartitionField) String() string {
 }
 
 func (p *PartitionField) UnmarshalJSON(b []byte) error {
-	var raw map[string]json.RawMessage
-	if err := json.Unmarshal(b, &raw); err != nil {
+	var aux struct {
+		SourceID  json.RawMessage `json:"source-id"`
```
zeroshade: `encoding/json` matches struct tags case-insensitively, whereas the previous `map[string]json.RawMessage` presence checks used exact keys. As a result, a payload containing `"Source-Id"` (or uppercase `"TRANSFORM"`) was rejected by the base decoder but is accepted here. Iceberg JSON member names are case-sensitive, so please preserve exact-key behavior and add mixed-case regression cases. The same issue exists in `table/sorting.go`.

partitions.go
```diff
@@ -122,32 +122,41 @@ func (p *PartitionField) String() string {
...
+		SourceIDs json.RawMessage `json:"source-ids"`
+		FieldID   int             `json:"field-id"`
+		Name      string          `json:"name"`
+		Transform json.RawMessage `json:"transform"`
+	}
+	if err := json.Unmarshal(b, &aux); err != nil {
```
zeroshade: Decoding `FieldID` and `Name` in this first pass changes validation precedence. For example, with both source fields plus `"field-id":"x"`, the base decoder returns `ErrInvalidPartitionSpec`, but this now exits here with a generic JSON type error, so `errors.Is(err, ErrInvalidPartitionSpec)` changes from true to false. A null transform plus an invalid `field-id` similarly stops matching `ErrInvalidTransform`. These are exported sentinels downstream callers can inspect; please retain the previous semantic-check ordering. `Direction`/`NullOrder` cause the equivalent change in `table/sorting.go`.

partitions.go
```diff
@@ -122,32 +122,41 @@ func (p *PartitionField) String() string {
...
-		SourceIDs       []int  `json:"source-ids,omitempty"`
-		FieldID         int    `json:"field-id"`
-		Name            string `json:"name"`
-		TransformString string `json:"transform"`
-	}{}
-
-	if err := json.Unmarshal(b, &aux); err != nil {
+	var sourceID int
+	if hasSourceID {
+		if err := json.Unmarshal(aux.SourceID, &sourceID); err != nil {
```
zeroshade: Unmarshalling the `RawMessage` as a standalone value drops the field name from type errors: malformed `source-id` used to identify `.source-id`, but now only says it cannot unmarshal into a Go value of type `int`. Please wrap each of these returns with the relevant JSON field name so malformed metadata remains diagnosable; the same applies to the corresponding returns in `table/sorting.go`.

zeroshade (changes requested):
Thanks for the focused optimization—the performance improvement is real, and the affected tests and CI are green. I’m requesting changes because the one-pass struct decode changes two externally observable behaviors: exact Iceberg JSON member-name handling and the precedence of exported validation sentinels. The inline comments include concrete differential repros. There is also a smaller diagnostics regression from decoding each RawMessage without field context.

zeroshade (approved):
LGTM

PR #1824, opened 2026-08-13
perf(dv): avoid dense expansion for sparse masks

fallintoplace:
## Summary

- Clear sparse deletion positions directly in the keep mask.
- Keep the dense word path for dense and run-heavy vectors.
- Skip out-of-range buckets without expanding them.
- Add sparse and dense benchmarks plus a high-position correctness case.

## Benchmark

Apple M1 Pro, arm64. The mask covers 1 MiB of rows:

| Case | Before | After | Result |
|---|---:|---:|---|
| 10 sparse deletes | 95.2 us, 254 KB, 3 allocs | 25.0 us, 131 KB, 2 allocs | 3.8x faster, 48% fewer bytes |
| 1,000 sparse deletes | 108.1 us, 262 KB, 3 allocs | 27.5 us, 131 KB, 2 allocs | 3.9x faster, 50% fewer bytes |
| Dense range | 115.0 us, 262 KB, 3 allocs | 112.4 us, 262 KB, 3 allocs | essentially unchanged |
| Out-of-range delete | 208.3 us, 2.1 MB, 3 allocs | 126.6 ns, 120 B, 2 allocs | no dense temporary |

## Tests

- go test ./...
- go vet .

### DIFF
--- a/table/dv/roaring_bitmap.go
+++ b/table/dv/roaring_bitmap.go
@@ -299,6 +299,7 @@ func (b *RoaringPositionBitmap) KeepMaskBytes(length int64) []byte {
 				globalPos := bucketBitBase + pos
 				out[globalPos>>3] &^= byte(1 << (globalPos & 7))
 			}
+
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


zeroshade:
Since this is a draft, i'll hold off on further review until it's marked ready and the tests are fixed

zeroshade (approved):
LGTM

PR #1826, opened 2026-08-13
fix(table): release owned nested element refs in list / map projection

### LINKED ISSUE #1819: Memory Leak in ToRequestedSchema for list/map with nested struct / list / map elements
### Apache Iceberg version

None

### Please describe the bug 🐞

Right now, `arrowProjectionVisitor.List` and `.Map` do not release the original element / key / value array before `castIfNeeded` returns its retained copy, when the element / key / value type is itself nested ( struct / list / map ).

- `Struct()` : handles this correctly with two defers
- `List` / `Map` : only have one

Thus, one arrow buffer reference leaked per nested container field per batch via `ToRequestedSchema` - affecting any scan over schemas.

### Status

Will be working on the fix ahead.

badalprasadsingh:
### Description

Fixes #1819 

There was a reference counting leak in schema projection - `ToRequestedSchema` for list / map columns whose element / key / value is struct, list or map. 

`List()` / `Map()` overwrote the caller-owned original array with `castIfNeeded`'s retained copy without releasing the original first — `Struct()` already does this correctly via a two-defer pattern; ports the same pattern.

### Testing
Added three regression tests that fail without the fix.

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1304,6 +1304,10 @@ func (a *arrowProjectionVisitor) List(listType iceberg.ListType, listArr arrow.A
 		return nil
 	}
 
+	if _, ok := valArr.DataType().(arrow.NestedType); ok {
+		defer valArr.Release()
+	}
+
 	valArr = a.castIfNeeded(listType.ElementField(), valArr)
 	defer valArr.Release()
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified the fix locally: all four new tests fail on main with checked-allocator leaks (e.g. `invalid memory size exp=0, got=256` in the map-struct-key case) and pass with the change, and the rest of the `ToRequestedSchema`/Arrow projection suite stays green.

The ownership analysis holds up. `castIfNeeded` (table/arrow_utils.go:1067) always returns a caller-owned ref (Retain or fresh cast), so `List()` overwriting `valArr` and `Map()` shadowing `keyResult`/`valResult` without first releasing the incoming owned nested results was a genuine leak — exactly the invariant `Struct()` already maintained. Extracting that predicate into `ownsChildResult` and reusing it in all three visitors keeps the invariant in one place. Two details I checked specifically:

- `defer valArr.Release()` at the top of `List()` (arrow_utils.go:1312) evaluates the receiver at defer time, so the later reassignment to the cast result doesn't change which array gets released — the original and the cast copy each get exactly one release.
- The early-return paths (`!ok || valArr == nil` in `List`, nil key/value in `Map`) now also release owned children they previously leaked, and there's no double-free risk since nested-typed child results only ever originate from `Struct`/`List`/`Map`, which return fresh caller-owned arrays (the updated `Variant` comment at arrow_utils.go:1412 correctly documents why variant stays exempt).

Good targeted fix with solid regression coverage.

PR #1828, opened 2026-08-13
fix(table): scope RollbackToSnapshot to the transaction's target branch

### LINKED ISSUE #1827: Transaction: RollbackToSnapshot ignores the transaction's target branch and always operates on main
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

The function `RollbackToSnapshot` hardcodes the main branch for both the ancestor-check snapshot and the ref it writes to, ignoring the `t.branch`. On a transaction opened via the new transaction on branch, lets say, "feature", calling `RollbackToSnapshot` either:
- silently rolls back to **main** instead of the targeted branch, or,
- silently succeeds as a no-op that never touches the targeted branch

Eventually, the targeted branch is never modified in each case.

### Status

Will quickly raise the PR for this fix. Intentionally keeping it out of PRs for #1636 or #1638.

badalprasadsingh:
### Description

Fixes #1827 

Resolves the branch the same way `AssertRefSnapshotID` already does (`t.branch`, default `main`), looks up that branch's own head from its own ref, and writes the ancestor check, the ref update, and the CAS assertion to it.

Also made sure:

- **Unknown branch fails, never falls back to main:** `currentSnapshotForRef`, from #1637, intentionally falls back to main for a new branch — reusing it here would let rolling back a nonexistent branch silently rewind main, reintroducing this exact bug.
- **Ref type is re-checked at the operation:** not only at transaction construction. Before this fix `RollbackToSnapshot` always wrote to `MainBranch`, a branch by construction, so no type check was needed. Resolving an arbitrary named ref makes rewriting a tag as a branch (and dropping its tag retention) newly reachable.

### Testing

Added tests covering the relevant edge cases.

### DIFF
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -474,9 +474,34 @@ func (t *Transaction) RollbackToSnapshot(snapshotID int64) error {
 	if err != nil {
 		return err
 	}
-	cs := meta.currentSnapshot()
-	if cs == nil {
-		return errors.New("cannot rollback: table has no current snapshot")
+
+	branch := t.branch
+	if branch == "" {
+		branch = MainBranch
+	}
+
+	// Resolve the head from the target branch's OWN ref. This deliberately does
+	// not reuse a resolver that falls back to main's head for an unknown ref:
+	// rolling back a branch that does not exist would then rewind
+	// main instead. An absent ref must fail here, never fall back.
+	ref, ok := meta.refs[branch]
+	if !ok {
+		return fmt.Errorf("cannot rollback: branch %q does not exist", branch)
+	}
+
+	if ref.SnapshotRefType != BranchRef {
+		return fmt.Errorf("%w: cannot rollback: ref %q is a %s, not a branch",
+			iceberg.ErrInvalidArgument, branch, ref.SnapshotRefType)
+	}
+
+	// Present but invalid is not the same as absent: metadata validation
+	// (checkRefsExist) rejects dangling refs on load, so reaching this means
+	// in-flight builder state is inconsistent. Fail closed and name the ref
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified the fix end to end: branch resolution matches how `Transaction.AssertRefSnapshotID` resolves the default (`t.branch`, falling back to `MainBranch`), the CAS requirement is correctly built from the base metadata via `baseRefSnapshotID` rather than the staged builder, and the deliberate refusal to fall back to main for an unknown ref is exactly right — reusing `currentSnapshotForRef` here would have reintroduced #1827 in a worse form. The test coverage is genuinely thorough (divergent-lineage ancestry, tag rejection at both the operation and commit, requirement dedup with a producer, no-retry on type conflict).

Two things I explicitly checked and found fine:

- Replacing `meta.currentSnapshot()` with a `meta.refs[MainBranch]` lookup for the default path is not a regression for legacy metadata that has `current-snapshot-id` but no explicit `main` ref: `constructRefs`/UnmarshalJSON synthesize the main ref from `current-snapshot-id`, so the ref is always present when a current snapshot exists (table/metadata.go:2202, 2282).
- The skipped requirement when `baseRefSnapshotID` returns nil (branch staged within this transaction) matches the existing pattern — the update that created the ref carries its own base-state requirement.

One message nit, non-blocking: the "branch %q does not exist" error at table/transaction.go:490 isn't wrapped in `iceberg.ErrInvalidArgument` while the tag-type error just below is; callers matching with `errors.Is` will only catch one of the two rejection paths. Fine to leave or fix in a follow-up.

LGTM.

PR #1831, opened 2026-08-13
fix(table): do not emit add-sort-order for an existing sort order id

ryanworl:
## What

REST catalogs reject commits that add a sort order under an id that already exists. `MetadataBuilder.AddSortOrder` still emitted `add-sort-order` when the order being "added" was already present in the **base** metadata: the existing guard (`*lastAddedSortOrderID != newOrderID`) only suppresses a second add within the same builder, and a fresh builder starts with `lastAddedSortOrderID == nil`, so the first reuse emitted the update.

`SetDefaultSortOrderID` compounded this by encoding `-1` ("last added") even when no `add-sort-order` preceded it in the update list, so replaying the emitted updates against a fresh builder could not resolve the default.

- The reuse path now records the id as `lastAddedSortOrderID` and returns without appending `add-sort-order`.
- `SetDefaultSortOrderID` encodes `-1` only when the update list actually carries an `add-sort-order` for that id; otherwise it writes the concrete id.

Net effect: setting the default sort order back to an existing order (e.g. unsorted) produces a commit payload with `set-default-sort-order` carrying the concrete id and no `add-sort-order`.

## Tests

- `TestAddSortOrderDoesNotAddDuplicateUpdate` updated: reuse of a base order emits nothing; adding a *new* order twice in one builder still emits exactly one add.
- New `TestAddSortOrderReuseDoesNotDuplicateUpdate`: unsorted→A→unsorted and A→B→A; replaying the emitted updates against a fresh builder reproduces the default.

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -637,12 +637,14 @@ func (b *MetadataBuilder) AddSortOrder(sortOrder *SortOrder) error {
 
 	newOrderID := b.reuseOrCreateNewSortOrderID(sortOrder)
 	if _, err := b.GetSortOrderByID(newOrderID); err == nil {
+		// The order already exists in this metadata: reuse its ID and
+		// do NOT emit an add-sort-order update (REST catalogs reject
+		// adding a sort order under an existing ID). Record it as
+		// last-added so a following SetDefaultSortOrderID(-1) resolves
+		// to the reused ID; SetDefaultSortOrderID encodes the concrete
+		// ID in that case because no add update precedes it.
 		sortOrder.orderID = newOrderID
-
-		if b.lastAddedSortOrderID == nil || *b.lastAddedSortOrderID != newOrderID {
-			b.lastAddedSortOrderID = &newOrderID
-			b.updates = append(b.updates, NewAddSortOrderUpdate(sortOrder))
-		}
+		b.lastAddedSortOrderID = &newOrderID
 
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Verified the two behavioral claims against the diff: the reuse path in `AddSortOrder` (table/metadata.go:639-648) now records `lastAddedSortOrderID` without appending `add-sort-order`, and `SetDefaultSortOrderID` (table/metadata.go:731) only encodes `-1` when an `add-sort-order` for that id is actually in the update list — this matches the Java builder's behavior where reusing an existing order produces no metadata change, and fixes the REST commit rejection.

I walked the edge cases: a new order added twice in one builder still emits exactly one add and `-1` for the default (covered in the updated `TestAddSortOrderDoesNotAddDuplicateUpdate`); reusing a base order after adding a different new order correctly emits the concrete id since `hasAddSortOrderUpdate` keys on the specific order id; and update replay through `addSortOrderUpdate.Apply`/`setDefaultSortOrderUpdate.Apply` round-trips, which the new replay assertions prove directly. Nice that the test verifies self-containedness by replaying the emitted updates against a fresh builder rather than just counting updates.

`go build ./...` and `go test ./table/... -count=1` pass on the PR head. One trivial nit, no action needed: `hasAddSortOrderUpdate` is a linear scan per `SetDefaultSortOrderID` call, which is fine at these sizes.

PR #1834, opened 2026-08-13
fix(table): assert default spec id when the default partition spec changes

ryanworl:
## What

`UpdateSpec.BuildUpdates` fenced spec evolution with `assert-last-assigned-partition-id` only. Two racing **remove-only** spec evolutions assign no new field ids, so both pass that assertion and both commit — last writer wins and the other spec change is silently lost.

This adds the missing requirement: `AssertDefaultSpecID`, pinned to the base default spec id, whenever the default partition spec changes.

## Why

Java parity: [`UpdateRequirements.forUpdateTable`](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/UpdateRequirements.java) registers `AssertDefaultSpecID` whenever a `SetDefaultPartitionSpec` update is present, alongside the last-assigned-partition-id assertion. With this in place, the second of two concurrent remove-only evolutions fails `assert-default-spec-id` instead of clobbering the first.

## Tests

- `TestUpdateSpecBuildChanges` now expects both requirements on add / remove / rename, with `assert-default-spec-id` pinning the base spec id.

### DIFF
--- a/table/update_spec.go
+++ b/table/update_spec.go
@@ -190,6 +190,12 @@ func (us *UpdateSpec) BuildUpdates() ([]Update, []Requirement, error) {
 			requiredLastAssignedPartitionID = &base
 		}
 		requirements = append(requirements, AssertLastAssignedPartitionID(*requiredLastAssignedPartitionID))
+		// Java UpdateRequirements.forUpdateTable registers
+		// AssertDefaultSpecID on SetDefaultPartitionSpec: two racing
+		// remove-only spec evolutions (which assign no new field ids)
+		// can both pass last-assigned-partition-id, so the default
+		// spec id itself must be pinned.
+		requirements = append(requirements, AssertDefaultSpecID(us.txn.tbl.Metadata().DefaultPartitionSpec()))
 	}
 
 	return updates, requirements, nil
--- a/table/update_spec_test.go
+++ b/table/update_spec_test.go
@@ -802,11 +802,12 @@ func TestUpdateSpecBuildChanges(t *testing.T) {
 		assert.NotNil(t, reqs)
 
 		assert.Equal(t, 2, len(updates))
-		assert.Equal(t, 1, len(reqs))
+		assert.Equal(t, 2, len(reqs))
 
 		assert.Equal(t, table.UpdateAddSpec, updates[0].Action())
 		assert.Equal(t, table.UpdateSetDefaultSpec, updates[1].Action())
 		assert.Equal(t, "assert-last-assigned-partition-id", reqs[0].GetType())
+		assert.Equal(t, table.AssertDefaultSpecID(testNonPartitionedTable.Metadata().DefaultPartitionSpec()), reqs[1])
 	})
 
 	t.Run("build changes on removed partition field", func(t *testing.T) {
@@ -821,11 +822,12 @@ func TestUpdateSpecBuildChanges(t *testing.T) {
 		assert.NotNil(t, reqs)
 
 		assert.Equal(t, 2, len(updates))
-		assert.Equal(t, 1, len(reqs))
+		assert.Equal(t, 2, len(reqs))
 
 		assert.Equal(t, table.UpdateAddSpec, updates[0].Action())
 		assert.Equal(t, table.UpdateSetDefaultSpec, updates[1].Action())
 		assert.Equal(t, "assert-last-assigned-partition-id", reqs[0].GetType())
+		assert.Equal(t, table.AssertDefaultSpecID(testPartitionedTable.Metadata().DefaultPartitionSpec()), reqs[1])
 	})
 
 	t.Run("build changes on renamed partition field", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
Verified against Java's `UpdateRequirements.forUpdateTable`: it does register `AssertDefaultSpecID` whenever a `SetDefaultPartitionSpec` update is present, and this is exactly the gap that lets two racing remove-only spec evolutions (no new field ids assigned) both slip past `assert-last-assigned-partition-id`. The fix in table/update_spec.go:170-174 is correctly placed inside the `DefaultPartitionSpec() != newSpec.ID()` branch, so both remove-only paths (`SetDefaultSpecUpdate` to an existing spec id) and add paths are covered, and it pins to `us.txn.tbl.Metadata().DefaultPartitionSpec()` — the base table metadata — consistent with the neighboring `AssertLastAssignedPartitionID` call. Multi-operation transactions are fine too: `requirementSemanticKey` in transaction.go dedupes identical requirements by their JSON encoding. Full `./table/...` suite passes on the PR head. One optional nit: a test that actually exercises the race (validate the new requirement against metadata whose default spec id has moved, expecting failure) would be nice, but `assertDefaultSpecId.Validate` is already covered in requirement_test.go, so I don't consider it blocking.

zeroshade:
This is approved, but it picked up a conflict when #1833 landed just ahead of it in today's merge queue. Could you rebase on latest main? Happy to merge once it's green again.

table/update_spec.go
```diff
@@ -167,6 +167,11 @@ func (us *UpdateSpec) BuildUpdates() ([]Update, []Requirement, error) {
 		}
 		requiredLastAssignedPartitionId := us.txn.tbl.Metadata().LastPartitionSpecID()
 		requirements = append(requirements, AssertLastAssignedPartitionID(*requiredLastAssignedPartitionId))
+		// Java UpdateRequirements.forUpdateTable registers
+		// AssertDefaultSpecID on SetDefaultPartitionSpec so two racing
+		// remove-only spec evolutions (which assign no new field ids)
+		// cannot both pass last-assigned-partition-id.
```
laskoviymishka: The "cannot both pass" clause reads backwards. The whole reason for this assertion is that two racing remove-only evolutions *can* both pass `assert-last-assigned-partition-id` (they allocate no new field ids, so that counter never moves), and the default-spec-id check is what makes the second commit fail.

I'd flip it, something like:

```go
// Java UpdateRequirements.forUpdateTable registers
// AssertDefaultSpecID on SetDefaultPartitionSpec because two racing
// remove-only spec evolutions (which assign no new field ids) can
// both pass assert-last-assigned-partition-id; the default spec id
// check makes the second commit fail.
```

table/update_spec_test.go
```diff
@@ -551,11 +552,12 @@ func TestUpdateSpecBuildChanges(t *testing.T) {
 		assert.NotNil(t, reqs)
 
 		assert.Equal(t, 2, len(updates))
-		assert.Equal(t, 1, len(reqs))
+		assert.Equal(t, 2, len(reqs))
 
 		assert.Equal(t, table.UpdateAddSpec, updates[0].Action())
 		assert.Equal(t, table.UpdateSetDefaultSpec, updates[1].Action())
 		assert.Equal(t, "assert-last-assigned-partition-id", reqs[0].GetType())
+		assert.Equal(t, table.AssertDefaultSpecID(testPartitionedTable.Metadata().DefaultPartitionSpec()), reqs[1])
```
laskoviymishka, line 830: These new assertions confirm the requirement is emitted with the right value, but nothing exercises the actual fence: building the requirement from base metadata, then calling `Validate` against metadata whose `DefaultPartitionSpec()` has already advanced and expecting an error. That `Validate` path is the property this PR exists to protect.

zeroshade flagged this as an optional nit; I'd actually want it before merge, since a future edit that drops the append (or regresses `assertDefaultSpecId.Validate`) wouldn't be caught otherwise. A subtest mirroring `TestAssertLastAssignedPartitionIDValidate` would cover it. wdyt?

laskoviymishka (approved):
Nice catch on this one. The fix is correct and lands in exactly the right branch: appending `AssertDefaultSpecID` inside the `DefaultPartitionSpec() != newSpec.ID()` block fences the remove-only race that `assert-last-assigned-partition-id` can't see, and it mirrors Java's `UpdateRequirements.forUpdateTable` faithfully. zeroshade already verified the Java parity and the dedup, so I'm not re-litigating any of that.

Two small things before I'd merge, neither a correctness objection. The one I care about: there's no test that actually exercises the race. Every new assertion checks the requirement is emitted with the right value, but none calls `Validate` against metadata whose default spec has already moved, which is the exact safety property this PR exists to protect. zeroshade called this an optional nit; I'd actually want it in before merge for that reason, since a future edit that drops the append wouldn't be caught otherwise.

The other is a one-word comment fix: the "cannot both pass" clause inverts the bug it's describing (the point is that the two racing evolutions *can* both pass the last-assigned check).

Fix those two and this is good to land.

ryanworl:
Rebased on latest main and addressed both of @laskoviymishka's points:

- New `TestUpdateSpecRemoveOnlyRaceIsFenced` exercises the race directly: writer A's requirements from a remove-only evolution are validated against metadata where the same evolution already committed. `assert-last-assigned-partition-id` passes (no new field ids were assigned), and exactly `assert-default-spec-id` fails.
- Fixed the inverted comment — it now says the two racing evolutions *can* both pass the last-assigned check.

`go test ./table/...` and golangci-lint (v2.12.2) are clean on the new head.

PR #1835, opened 2026-08-14
feat(table): add CommitEqualityDeletes convenience API

### LINKED ISSUE #1207: Add convenience API for committing equality delete records

iceberg-go already has low-level support for equality deletes via `Transaction.WriteEqualityDeletes`, and `RowDelta` can commit those delete files. For CDC/upsert workloads, callers currently need to manually compose the two steps:

```go
deleteFiles, err := tx.WriteEqualityDeletes(ctx, equalityFieldIDs, records)
if err != nil {
    return err
}

rd := tx.NewRowDelta(snapshotProps)
rd.AddDeletes(deleteFiles...)
err = rd.Commit(ctx)
```

Would maintainers be open to a small convenience API that wraps this flow, for example:

```go
func (t *Transaction) DeleteEquality(
    ctx context.Context,
    equalityFieldIDs []int,
    records iter.Seq2[arrow.RecordBatch, error],
    snapshotProps iceberg.Properties,
) error
```

This would not change `Delete(filter)` semantics. it would only expose an explicit delete-by-equality-records path for CDC-style deletes where the caller already has the equality key values.

I have a downstream CDC/streaming engine using the same Iceberg equality-delete pattern, and I have validated the generated delete files with Spark and Trino readers. My goal is to eventually migrate that downstream implementation to use iceberg-go’s native equality-delete path once the API shape is aligned with the library.

If this direction sounds reasonable, I'd be happy to contribute it with focused tests for:
- format v2 requirement
- empty equality field IDs
- upartitioned equality deletes
- partitioned equality deletes
- scan applying the committed equality deletes
- snapshot summary/delete counts

This would cover delete-only CDC events. For update CDC events that need delete + append in a single snapshot, callers can still use the lower-level RowDelta flow directly. I’d prefer to keep this proposal scoped to the delete-only convenience API first, and discuss a higher-level row-delta records API separately if maintainers think it would be useful.

tanmayrauth:
Adds `Transaction.CommitEqualityDeletes`, a thin convenience wrapper over `WriteEqualityDeletes` + `NewRowDelta().AddDeletes().Commit()` for delete-only CDC/upsert commits where the caller already holds the equality-key values.
  
  Closes #1207 
  
  ## Scope
  
Adds no new behavior beyond composing the two existing steps:
  - format-version >= 2 gate — inherited from `WriteEqualityDeletes`
  - equality-field-ID validation — inherited from `WriteEqualityDeletes` / `RowDelta`
  - concurrent-writer conflict detection — inherited from `RowDelta`, which validates against `write.delete.isolation-level` (serializable by default) at `Transaction.Commit`
  
  The one intentional behavior: an empty record stream is a no-op (returns nil,  stages nothing) rather than surfacing RowDelta's "at least one file" error.
  
  Update CDC events needing delete + append in one snapshot still use the lower-level `RowDelta` flow directly — out of scope here, as discussed.

### DIFF
--- a/table/equality_delete_commit_test.go
+++ b/table/equality_delete_commit_test.go
@@ -1,0 +1,263 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table_test
+
+import (
+	"testing"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/iceberg-go"
+	"github.com/apache/iceberg-go/table"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// idKeyArrowSchema builds the Arrow schema for an equality delete file
+// keyed on the "id" column alone.
+func idKeyArrowSchema(t *testing.T) *arrow.Schema {
+	t.Helper()
+
+	sc, err := table.SchemaToArrowSchema(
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
This is a faithful, thin composition of the two existing steps, and the test coverage is genuinely thorough. I verified the wrapper in table/equality_delete_writer.go adds no new validation paths: format-version and field-ID gates come from `WriteEqualityDeletes`, and conflict detection from `RowDelta` — all matching the PR description. The full suite (`go test ./table/ -run TestCommitEqualityDeletes`) passes on the PR head, including the serializable-conflict and snapshot-isolation companion tests, which are exactly the safety properties I'd want pinned for a CDC entry point.

Two non-blocking notes:

1. The name `CommitEqualityDeletes` slightly overstates what it does — it stages a row delta in the transaction rather than committing to the catalog. The doc comment covers this clearly and it mirrors `RowDelta.Commit` naming, so I'm fine keeping it, but flagging in case anyone prefers something like `ApplyEqualityDeletes` before this becomes public API.
2. If `Transaction.Commit` later fails, the delete files written by `WriteEqualityDeletes` are orphaned — that's inherent to the existing write path, not something this PR introduces, so no action needed here.

LGTM.

PR #1842, opened 2026-08-17
fix(table): retain ref head snapshots in expire snapshots (#1829)

### LINKED ISSUE #1829: Transaction: ExpireSnapshots silently deletes non-main branch when min-snapshots-to-keep is zero
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Well, when `history.expire.min-snapshots-to-keep` is `0` and a non-main branch's head snapshot is older than `max-snapshot-age-ms`, `ExpireSnapshots` expires the head snapshot itself. The branch ref then loses its target snapshot and `RemoveSnapshots` silently prunes the ref entirely — the branch disappears with no error.

IMO, root cause is: only main's head snapshot is unconditionally protected before the age / count trim loop runs; other branches rely on the loop adding their head, which can break on iteration 0 when `numSnapshots(0) >=
minSnapshotsToKeep(0)`.

### Status

Will be working on the fix and raise the PR soon.

badalprasadsingh:
### Description

Closes #1829 

Earlier, `ExpireSnapshots` only protected _main_'s head. With `history.expire.min-snapshots-to-keep=0`, a _non-main_ branch or _tag_ head older than max-snapshot-age-ms was expired and its ref silently dropped, deleting data files. 

From now on, every retained _ref_'s head is kept. Both ancestor walks reuse `AncestorsOf`, which also terminates on cyclic metadata. 

### Testing
Added regression tests, and few guards as well.

### DIFF
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -656,8 +656,9 @@ func getRetentionInt64(props iceberg.Properties, standardKey, legacyKey string, 
 // with reference age falling back through the ref's max-ref-age-ms, the table's
 // MaxRefAgeMsKey property, and the specification default. Snapshot age falls
 // back through the ref's own settings, the options passed here, and finally the
-// table's MinSnapshotsToKeepKey and MaxSnapshotAgeMsKey properties. The current
-// snapshot of the main branch is always kept.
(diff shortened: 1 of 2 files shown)


tanmayrauth (approved):
LGTM 👍🏻

zeroshade (approved):

PR #1843, opened 2026-08-17
fix(deps): bump twmb/avro to v1.8.0 and adapt to its API changes

twmb:
Bumps `github.com/twmb/avro` from v1.7.2 to v1.8.0. Two changes in that release affect this repo — one is a compile error, the other is silent.

### `Schema.Root()` returns `*SchemaNode`

Field access auto-derefs, so only the four value-context uses need a change: passing the node to `NullableNode`/`getField`, and assigning it to `SchemaField.Type`.

### `Root()` numeric metadata is now `int64`, not `float64`

v1.8.0 preserves numeric precision when decoding schema JSON metadata, because integers above 2^53 previously rounded through `float64`. A field's `field-id` property therefore comes back as `int64` where it used to be `float64`.

`schemaFieldID` matched only `int` and `float64`, so on v1.8.0 it returns `false` for **every** field. Nothing errors — `getFieldIDMap` just ends up empty, `avroEncodePartitionData` iterates an empty map, and **partition values are written to the manifest as null**. The header schema is still written correctly, and the read path drops the same keys for the same reason, so the loss is silent in both directions.

Accepting `int64` alongside the existing cases keeps this correct on both avro versions.

### Verification

Two existing tests catch this and fail on the bump without the `schemaFieldID` change; both pass with it:

- `table/internal` — `TestDataFileStatisticsDecimalPartitionManifestRoundTrip` (decimal partition value)
- `table/compaction` — `TestCollectDeadPositionDeletesPartitioned` (string partition values, and this package doesn't import avro at all)

The second one confirms the problem is partition values generally, not anything decimal-specific.

`gofmt` and `go vet` are clean on the changed packages. I ran the two tests above rather than the full suite locally; relying on CI for the rest.

Disclosure: I maintain `twmb/avro`. The `int64` change there was deliberate and matches Avro Java, whose `JsonProperties.getObjectProp` yields an integer for integral metadata — v1.7.2's `float64` was the divergence. It should have been called out in that release's notes, and I'm fixing that separately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/substrait-io/substrait-go/v8 v8.1.1
 	github.com/testcontainers/testcontainers-go/modules/compose v0.44.0
-	github.com/twmb/avro v1.7.2
+	github.com/twmb/avro v1.8.0
 	github.com/twmb/murmur3 v1.1.8
 	github.com/twpayne/go-geom v1.6.1
 	github.com/uptrace/bun v1.2.18
--- a/go.sum
+++ b/go.sum
@@ -631,8 +631,8 @@ github.com/transparency-dev/formats v0.1.1 h1:4bVHJc+KdBgpA1OJD1yjI+g0i5Z1graCpp
 github.com/transparency-dev/formats v0.1.1/go.mod h1:qtZ8goRuJ8FTBG9c9+Bj0rn2rUG7eG/AUTkr+Aw3jFw=
 github.com/transparency-dev/merkle v0.0.2 h1:Q9nBoQcZcgPamMkGn7ghV8XiTZ/kRxn1yCG81+twTK4=
 github.com/transparency-dev/merkle v0.0.2/go.mod h1:pqSy+OXefQ1EDUVmAJ8MUhHB9TXGuzVAT58PqBoHz1A=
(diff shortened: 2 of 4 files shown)


zeroshade (approved):

twmb:
@zeroshade apologies — I rebased this onto `28b989e` and force-pushed, which wiped your merge of main. That wasn't deliberate; I'd fetched upstream but not this branch, so I didn't see your commit until after the push. CONTRIBUTING is clear about not force-pushing and I should have looked first.

Nothing of yours is lost in content — the branch is the same four-file change on top of current main. The reason I rebased rather than merging is that your merge predates `28b989e` ("hotfix: remove redundant loop variable copy in test"), so lint was still failing on `copyloopvar` in `table/positional_delete_index_test.go`, a file this PR doesn't touch. Rebasing onto that hotfix clears it.

Happy to restore your merge commit and stack a merge of main on top instead if you'd prefer the history that way — just say so. I won't force-push this branch again either way.

zeroshade (approved):
That's fine. my goal was just resolving the merge conflicts anyways. This LGTM

twmb:
Sorry for the LLM speak above. The agent went ...a bit... above and beyond. My session was about auditing repo's I know use this package and then fixing them up; the PR opening and noticing your merge were a _bit_ past the point, but also it's all good so, thank you for merging!

PR #1846, opened 2026-08-18
feat(table): type uniformity for variant shredding inference

nssalian:
## Rationale for the change

Ports the type-uniformity rule from Java's [apache/iceberg#17424](https://github.com/apache/iceberg/pull/17424) into the Go shredding-inference analyzer.

## Changes

`AnalyzeVariantShredding` used to shred a mixed-type field by majority vote, leaving a `typed_value` column that only covered the majority rows. Now a field shreds only when all its non-null values fall in one type family: `fieldInfo.admittedType`
(renamed from `mostCommonType`) collapses the integer and decimal families to their widest member and returns "not shreddable" if more than one family remains. Integer and decimal widening still shred; cross-family mixes (int+string, int+decimal, float+double, timestamp micros+nanos) no longer do. The dead majority-vote tie-break is removed.

Write-side inference only - read paths are unaffected.

## Are these changes tested?

- New/updated analyzer cases cover mixed-type, cross-family, empty-container, widening, and nulls-plus-one-type; `TestShreddedVariantWriteMixedTypeField` proves end to end that a mixed field is not shredded and every row still round-trips.
- Cross-engine Spark 3.5 and 4.0 integration suites pass locally


table/internal/variant_shredding.go
```diff
@@ -47,17 +47,6 @@ var decimalPriority = map[variant.Type]int{
...
-	variant.Bool: 0, variant.Int8: 1, variant.Int16: 2, variant.Int32: 3,
-	variant.Int64: 4, variant.Float: 5, variant.Double: 6, variant.Decimal4: 7,
-	variant.Decimal8: 8, variant.Decimal16: 9, variant.Date: 10, variant.Time: 11,
-	variant.TimestampMicros: 12, variant.TimestampMicrosNTZ: 13, variant.Binary: 14,
-	variant.String: 15, variant.TimestampNanos: 16, variant.TimestampNanosNTZ: 17,
-	variant.UUID: 18,
-}
-
 type fieldInfo struct {
 	typeCounts          map[variant.Type]int
```
zeroshade: Non-blocking simplification: now that majority selection is gone, nothing reads these per-type counts; only the map keys remain relevant. Consider storing admission state directly (`admitted`, `hasAdmitted`, `mixed`) and merging each observation through a small helper that widens integer/decimal pairs and rejects cross-family pairs. That would remove this per-node map allocation and most of the `widestInt`/`widestDec`/`otherFamilies` bookkeeping below. If retaining post-hoc calculation, at least model this as `observedTypes map[variant.Type]struct{}` so the representation matches the new semantics.

table/internal/variant_shredding_test.go
```diff
@@ -115,22 +127,96 @@ func TestAnalyzeIntegerWidening(t *testing.T) {
...
+// TestAnalyzeMixedSiblingKeepsUniform: a mixed-type field is dropped from the
+// shredded type while its uniform sibling still shreds.
+func TestAnalyzeMixedSiblingKeepsUniform(t *testing.T) {
+	var sample []variant.Value
+	for range 8 {
+		sample = append(sample, mkVar(t, map[string]any{"good": bigI64, "bad": bigI64}))
+	}
+	for range 2 {
+		sample = append(sample, mkVar(t, map[string]any{"good": bigI64, "bad": "s"}))
+	}
```
zeroshade: Non-blocking simplification: the preceding test already uses a 7/3 split to prove that a majority no longer wins, so this sibling-focused test only needs two rows: one with `bad` as an integer and one with it as a string. You could then compare `dt` directly with `arrow.StructOf` containing only the expected `good` field; that removes the loops, unchecked struct assertion, and field lookups while making the no-extra-fields assertion stronger.

table/internal/variant_shredding_test.go
```diff
@@ -268,22 +354,14 @@ func TestInferredTypedScalarsShred(t *testing.T) {
...
-			v := mk()
+			v := mkBuilt(t, c.build)
 			inner, ok := AnalyzeVariantShredding([]variant.Value{v})
 			require.Truef(t, ok, "%s should infer a shredded type", c.name)
 
 			st := extensions.NewShreddedVariantType(inner)
 			bldr := extensions.NewVariantBuilder(mem, st)
 			defer bldr.Release()
-			bldr.Append(mk())
+			bldr.Append(mkBuilt(t, c.build))
```
zeroshade: Non-blocking nit: `v` was already built above and `Append` does not require a fresh value, so this can simply be `bldr.Append(v)`. That avoids invoking the builder callback twice and makes it clearer that the value used for inference is the one being round-tripped.

zeroshade (approved):
LGTM — this cleanly replaces majority-vote shredding with Java-aligned type-family uniformity, with focused analyzer and end-to-end regression coverage. The inline notes are non-blocking simplification suggestions.

---

>

table/internal/variant_shredding_test.go
nssalian: Makes sense. Simplified.

table/internal/variant_shredding.go
nssalian: Also done

PR #1850, opened 2026-08-19
fix(table): retain equality fields for delete scans

### LINKED ISSUE #1849: Equality-delete scans fail when projection omits equality fields
## Version

`main` at `8f6e791182d4268c02bd0261b135f8134047f124` (also affects the equality-delete read path released in v0.6.0).

## Problem

A scan fails when an applicable equality delete references a field that is not part of the selected output columns.

The Arrow scanner builds the physical data-file projection from the requested output fields and row-filter fields. Equality-delete fields are needed to evaluate row liveness, but are not added to that physical projection. The equality-delete processor then cannot resolve the omitted field and returns an error.

Output projection should not change which rows are considered deleted.

## Reproduction

1. Create a v2 table with schema `id: long, data: string`.
2. Append rows `(1, "one")`, `(2, "two")`, and `(3, "three")`.
3. Add an equality delete for `id = 2` with equality field ID `1`.
4. Read using `tbl.Scan(table.WithSelectedFields("data")).ToArrowRecords(ctx)`.
5. Consume the returned record iterator.

This fails deterministically with:

```text
equality field ID 1 (id) not found in <table>/data/data.parquet
```

The existing equality-delete round-trip succeeds when selecting `id`, which is why current tests do not expose the projection dependency.

## Expected behavior

The scan returns the projected `data` values `"one"` and `"three"`. The scanner may read `id` internally to apply the delete, but it should not expose `id` in the result schema.

## Actual behavior

The record iterator returns an error and no result because `id` was pruned before equality deletes were applied.

## Proposed fix

Include equality-delete field IDs in the physical data-file projection for each scan task, apply equality deletes, and retain the existing final projection to remove execution-only columns from the output. Add a regression test whose selected output excludes the equality key.

mattfaltyn:
Fixes #1849.

## Summary

- include applicable equality-delete field IDs in the physical data-file projection
- keep the requested result projection unchanged, so execution-only equality fields are not exposed
- cover scans that select a non-key column while deleting by an omitted equality key

## Why

Equality deletes are row semantics and must be evaluated independently of the selected output columns. Previously, the scanner pruned an omitted equality key before applying deletes, causing the record iterator to fail with `equality field ID ... not found`.

## Testing

- `go test ./table -run '^TestEqualityDeleteReadRoundTrip$' -count=1 -v`
- `go test ./...`
- `go test -race ./codec/... ./metrics/... ./table/...`

### DIFF
--- a/table/equality_delete_reader_test.go
+++ b/table/equality_delete_reader_test.go
@@ -142,6 +142,22 @@ func TestEqualityDeleteReadRoundTrip(t *testing.T) {
 	}
 
 	assert.Equal(t, []int64{1, 3, 5}, ids, "expected rows with id=2 and id=4 deleted")
+
+	_, itr, err = tbl.Scan(table.WithSelectedFields("data")).ToArrowRecords(t.Context())
+	require.NoError(t, err)
+
+	var data []string
+	for rec, err := range itr {
+		require.NoError(t, err)
+		col := rec.Column(0).(*array.String)
+		for i := range col.Len() {
+			data = append(data, col.Value(i))
+		}
+		rec.Release()
+	}
+
+	assert.Equal(t, []string{"alpha", "gamma", "epsilon"}, data,
+		"expected equality deletes to apply when the equality field is not projected")
 }
 
 func TestEqualityDeleteReadSharesMultiFileUnionAcrossConcurrentTasks(t *testing.T) {
--- a/table/equality_delete_reader_test.go
+++ b/table/equality_delete_reader_test.go
@@ -143,12 +143,15 @@ func TestEqualityDeleteReadRoundTrip(t *testing.T) {
 
 	assert.Equal(t, []int64{1, 3, 5}, ids, "expected rows with id=2 and id=4 deleted")
 
-	_, itr, err = tbl.Scan(table.WithSelectedFields("data")).ToArrowRecords(t.Context())
+	resultSchema, itr, err := tbl.Scan(table.WithSelectedFields("data")).ToArrowRecords(t.Context())
 	require.NoError(t, err)
+	require.Len(t, resultSchema.Fields(), 1)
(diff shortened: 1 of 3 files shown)


table/equality_delete_reader_test.go
```diff
@@ -142,6 +142,22 @@ func TestEqualityDeleteReadRoundTrip(t *testing.T) {
 	}
 
 	assert.Equal(t, []int64{1, 3, 5}, ids, "expected rows with id=2 and id=4 deleted")
+
+	_, itr, err = tbl.Scan(table.WithSelectedFields("data")).ToArrowRecords(t.Context())
```
zeroshade: Non-blocking: since `id` is now read as an execution-only field, could this retain the returned schema and assert that it contains only `data` (and that each record has exactly one column)? That would make the no-hidden-field-leak contract explicit rather than relying on `Column(0)` alone.

zeroshade (approved):
Looks good. The task-specific equality-delete fields are included only for execution, deletes are applied before the final requested-schema projection, and the positional-delete path remains unchanged. The targeted, table, and race tests all pass.

table/equality_delete_reader_test.go
mattfaltyn: > Non-blocking: since `id` is now read as an execution-only field, could this retain the returned schema and assert that it contains only `data` (and that each record has exactly one column)? That would make the no-hidden-field-leak contract explicit rather than relying on `Column(0)` alone.

Implemented in `67a402e`. The regression now retains the returned schema, asserts it has exactly one field named `data`, and requires every record batch to have exactly one column.

Validated with:

- `go test ./table -run '^TestEqualityDeleteReadRoundTrip$' -count=1 -v`
- `go test ./table/...`
- `golangci-lint run --timeout=10m` (`0 issues`)

PR #1851, opened 2026-08-19
fix(table): drain RollingDataWriter's recordCh on mid-stream write errors

### LINKED ISSUE #1825: RollingDataWriter: batches queued in recordCh are stranded when stream exits on a mid-stream write error
RollingDataWriter.stream() deregisters itself from writerFactory.writers via a deferred CompareAndDelete as soon as it returns , including on an early return triggered by a write error partway through processing. If other batches were already queued in recordCh (sent via Add before the error), abortAll() can no longer find this writer once it eventually runs, since it's already gone from the registry — so nothing ever drains or releases those batches.

Confirmed with a CheckedAllocator, and reproduces without any Windows dependency: any write error partway through a rolling writer's stream strands whatever else is already buffered in recordCh.

**Repro**: queue several valid records via Add, then force the next file-open/write to fail (e.g. a FileFormat stub whose NewFileWriter returns an error). stream() exits on the first record, the rest stay in recordCh forever, CheckedAllocator reports the leak.

deepak-2605:
## What

When a `RollingDataWriter`'s `stream()` goroutine exits early on a write
error, batches already queued in its `recordCh` (sent via `Add` before the
error surfaced) were never released, leaking their Arrow buffers.

Closes #1825.

## Root cause

`stream()` deregisters itself from `writerFactory.writers` via a deferred
`CompareAndDelete` the moment it returns. On an early error return, that
runs while `recordCh` may still hold queued batches. The cleanup path that
is supposed to drain those batches, `abortAll()`, walks the writer registry
— but the writer has already removed itself, so `abortAll()` never sees it
and the buffered batches are stranded.

This is independent of any partitioning race: any write error partway
through a rolling writer's stream strands whatever is still buffered in
`recordCh`. (An earlier revision of #1825 described a Windows-specific
trigger for the same symptom; that turned out to be a separate path-parsing
bug, tracked in #1262.)

## Fix

- `stream()`'s cleanup now cancels its context and drains `recordCh`
  (non-blocking) before the `CompareAndDelete` defer deregisters the writer,
  so cleanup no longer depends on `abortAll()` finding it in the registry.
- `Add()` checks `ctx.Done()` up front (non-blocking) so that once a writer
  has died, in-flight sends from other fanout workers back off instead of
  queuing into a channel nobody will drain.
- `recordCh` is never closed from inside `stream()` — that stays with the
  external `closeAll()`/`abortAll()` paths, which only run after all `Add`
  callers have returned, avoiding a send-on-closed-channel panic.

## Test plan

- Added `TestStreamErrorDrainsBufferedRecords`: queues several batches, forces
  the file open to fail, and asserts the checked allocator returns to zero.
  Fails on `main`, passes with this change.
- `go test ./table/ -run TestRollingDataWriter -race` passes.
- 50x `-race` stress run of the fanout/rolling-writer tests: no panics, no
  data races.

### DIFF
--- a/table/rolling_data_writer.go
+++ b/table/rolling_data_writer.go
@@ -389,14 +389,9 @@ func (w *writerFactory) getOrCreateRollingDataWriter(ctx context.Context, partit
 
 // Add appends a record to the writer's buffer.
 func (r *RollingDataWriter) Add(record arrow.RecordBatch) error {
-	// Hold the read lock across the whole check-and-enqueue so it is atomic
-	// with stream's shutdown drain (which takes the write lock). Once the
-	// drain has set noMoreSends, this returns before retaining, so a record
-	// can never be enqueued into a channel that will not be drained again.
-	// The enqueue below can still block for backpressure while the writer is
-	// live, but stream's cleanup cancels ctx before it takes the write lock,
-	// so a parked send unblocks via ctx.Done() and releases the read lock
-	// rather than deadlocking the drain.
+	// Read lock held across the whole check-and-enqueue so it is atomic with
+	// stream's shutdown (which takes the write lock): once noMoreSends is set
+	// this returns before retaining, so nothing is enqueued into a dead channel.
 	r.sendMu.RLock()
 	defer r.sendMu.RUnlock()
 
@@ -432,6 +427,13 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
 	var currentWriter tblutils.FileWriter
 	var fileArrowSchema *arrow.Schema // per-file shredded schema; nil => factory default
 
+	// Set only when stream reaches the end of its input and finalizes without
+	// error. The cleanup defer uses it to skip cancelling ctx on the success
+	// path — cancellation is an error-exit signal that unblocks parked senders,
+	// and on the clean path there are none (input is closed after all senders
+	// finish), so the barrier's write lock is uncontended anyway.
+	exitedCleanly := false
+
 	// bootstrap buffer of converted batches, held until inference runs.
 	var buf []arrow.RecordBatch
 	var bufRows int64
@@ -466,22 +468,19 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
 		if currentWriter != nil {
 			_ = currentWriter.Abort()
 		}
-		// stream is exiting; release any records producers left buffered in
-		// recordCh. This is a hard barrier against in-flight Add calls, not
-		// just a drain: an Add that already observed noMoreSends as false
-		// could otherwise enqueue after a lockless drain saw the channel
-		// empty, leaving a retained record with no consumer. Cancel first so
-		// any send currently parked on recordCh unblocks via ctx.Done() and
-		// releases its read lock; then take the write lock, which cannot be
-		// acquired until every active Add has returned. Setting noMoreSends
-		// under that lock makes the set of records ever placed in recordCh
-		// final, so the drain below releases all of them and nothing can be
-		// enqueued afterward. recordCh is deliberately not closed here: an Add
-		// may still hold the read lock about to observe noMoreSends, and
-		// closing under a live sender could panic. This runs before the
-		// CompareAndDelete defer above, so cleanup does not depend on abortAll
-		// finding this writer in the registry.
-		r.cancel()
+		// Release anything producers left buffered in recordCh, behind a hard
+		// barrier with in-flight Add calls (see sendMu). On the error exit,
+		// cancel first so a send parked on a full channel unblocks via
+		// ctx.Done() and drops its read lock; the clean exit has no parked
+		// senders (closeInput ran after all senders finished), so skip the
+		// cancel. Then the write lock — which waits out every active Add — plus
+		// noMoreSends makes the buffered set final, and the drain releases it.
+		// recordCh is not closed here: another goroutine may close it via
+		// closeInput (the drain's !ok case ends the loop then), and closing
+		// under a live sender could panic.
+		if !exitedCleanly {
+			r.cancel()
+		}
 		r.sendMu.Lock()
 		defer r.sendMu.Unlock()
 		r.noMoreSends = true
@@ -697,7 +696,10 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
 	}
 	if err := closeWriter(); err != nil {
 		r.sendError(err)
+
+		return
 	}
+	exitedCleanly = true
 }
 
 // inferShreddingFromBatches infers the inner type per top-level variant column, keyed
--- a/table/rolling_data_writer.go
+++ b/table/rolling_data_writer.go
@@ -328,6 +328,15 @@ type RollingDataWriter struct {
 	cancel          context.CancelFunc
 	wg              sync.WaitGroup
 	closeRecordCh   sync.Once
+
+	// sendMu forms a barrier between Add (read lock) and stream's shutdown
+	// drain (write lock). While an Add holds the read lock its check of
+	// noMoreSends and its enqueue are atomic with respect to the drain: once
+	// the drain holds the write lock and sets noMoreSends, no Add can be
+	// mid-enqueue and none can start one, so the drain releases every record
+	// that will ever reach recordCh.
+	sendMu      sync.RWMutex
+	noMoreSends bool
 }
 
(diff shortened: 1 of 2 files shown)


table/rolling_data_writer.go
```diff
@@ -430,6 +442,31 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
...
+		// any Add call still in flight from another fanout worker prefers
+		// bailing out via ctx.Done() (see Add's priority check) over queuing
+		// into a channel nobody will drain again, then release whatever is
+		// already buffered right now. recordCh is never closed here: other
+		// callers may still be sending to it, and closing while a send can
+		// still land would risk a send-on-closed-channel panic. That leaves
+		// a narrow window between this drain and Add's ctx.Done() check
+		// taking effect elsewhere, but it is bounded to that one window
+		// rather than persisting for the life of the writer.
+		r.cancel()
```
zeroshade: **Blocking: an in-flight `Add` can enqueue after this drain finishes.** The first `ctx.Done()` check does not synchronize with cleanup. If `Add` passes that check and pauses in `record.Retain()`, this defer can cancel, observe an empty `recordCh`, and return. When `Add` resumes, both the buffered send and `ctx.Done()` are ready, so Go may choose the send; `Add` returns `nil` and the retained record remains queued with no consumer. I reproduced this controlled interleaving 200 times and observed 108 post-drain sends. The acknowledged window therefore still preserves the leak this PR is intended to fix. Please add a barrier with active `Add` calls (or otherwise make send-vs-shutdown atomic) before the final drain completes.

table/rolling_data_writer_test.go
```diff
@@ -905,3 +905,60 @@ func (s *RollingDataWriterTestSuite) TestStreamRecoversWriterClosePanic() {
...
+		factory:  factory,
+		ctx:      ctx,
+		cancel:   cancel,
+	}
+
+	const numRecords = 5
+	for range numRecords {
+		record := s.buildRecord(arrSchema, 3)
+		record.Retain()
+		writer.recordCh <- record
```
zeroshade, line 1029: This setup proves that cleanup releases records already buffered before the stream exits, but bypassing `Add` also bypasses the failing interleaving. Please exercise an `Add` that is in flight between its initial context check and its send while the stream shuts down, and assert that it cannot return success or leave a record in `recordCh` after cleanup.

zeroshade (changes requested):
The current cleanup releases records that are already buffered when the stream exits, but it does not establish a barrier with in-flight `Add` calls. An `Add` that has passed the first cancellation check can enqueue after the drain observes an empty channel, return `nil`, and leave its retained record with no consumer. I reproduced that exact interleaving with an instrumented `RecordBatch.Retain`: 108 of 200 calls sent after cleanup's final drain. Please synchronize cleanup with active senders and extend the regression to exercise `Add` rather than only preloading the channel.

---

>

table/rolling_data_writer.go
laskoviymishka: This defer runs on every `stream()` exit, including the clean channel-closed path, so `r.cancel()` now fires from the stream goroutine on success too, before `wg.Done()` unblocks `closeAndWait`. It's idempotent today (`closeAndWait` cancels again), but it's a quiet semantic shift: the per-writer context is now cancelled "by stream" on the happy path, not "by closeAndWait", so anything that later checks `r.ctx.Err()` after `wg.Wait()` returns would see a cancelled context on a successful write. If we keep this shape I'd guard it behind an exited-cleanly flag so the cancel only fires on the error exit.

table/rolling_data_writer.go
```diff
@@ -430,6 +442,31 @@ func (r *RollingDataWriter) stream(outputDataFilesCh chan<- iceberg.DataFile) {
...
 			_ = currentWriter.Abort()
 		}
+		// stream is exiting; this defer runs before the CompareAndDelete
+		// defer above deregisters the writer, so cleanup does not depend on
+		// abortAll finding this writer in the registry: by the time (if
+		// ever) abortAll looks, this writer is already gone. Cancel first so
+		// any Add call still in flight from another fanout worker prefers
+		// bailing out via ctx.Done() (see Add's priority check) over queuing
+		// into a channel nobody will drain again, then release whatever is
+		// already buffered right now. recordCh is never closed here: other
```
laskoviymishka: Small thing, but "recordCh is never closed here" reads like the `!ok` case a few lines down is dead code. It isn't: `closeInput()` closes the channel and, since `wg.Done()` is the last defer here (LIFO), it can run while this drain is still looping. So `!ok` is genuinely reachable, and someone could prune it as dead code and turn the drain into a spin on a closed channel. I'd reword to say another goroutine may close `recordCh` via `closeInput`/abort and that `!ok` is what ends the drain.

While here: this comment block is much denser than the rest of the file, where the method bodies are mostly one line or none. I'd trim it to the intent plus the one real constraint (don't close recordCh here) and let the code carry the rest.

table/rolling_data_writer_test.go
```diff
@@ -905,3 +905,60 @@ func (s *RollingDataWriterTestSuite) TestStreamRecoversWriterClosePanic() {
...
+// stream ever runs — otherwise Add would race stream's error exit for the
+// later records once errorCh closes.
+func (s *RollingDataWriterTestSuite) TestStreamErrorDrainsBufferedRecords() {
+	arrSchema := arrow.NewSchema([]arrow.Field{
+		{Name: "id", Type: arrow.PrimitiveTypes.Int32, Nullable: true},
+		{Name: "name", Type: arrow.BinaryTypes.String, Nullable: true},
+	}, nil)
+
+	loc := filepath.ToSlash(s.T().TempDir())
+	factory, _ := s.createWriterFactory(loc, arrSchema, 1024*1024)
```
laskoviymishka, line 1010: `createWriterFactory` calls `iter.Pull`, which starts a goroutine that only stops via `stopCount()` inside `closeAll()`/`abortAll()`. Every other test in this file that builds a factory pairs it with `defer factory.closeAll()`; this one doesn't, so the pull goroutine leaks for the rest of the run (a goleak check would flag it). I'd add `defer factory.closeAll()` right here. No writer is registered on this factory, so it just calls `stopCount()`.

table/rolling_data_writer_test.go
```diff
@@ -905,3 +905,60 @@ func (s *RollingDataWriterTestSuite) TestStreamRecoversWriterClosePanic() {
...
+	factory.format = failingOpenFormat{FileFormat: factory.format}
+
+	outputCh := make(chan iceberg.DataFile, 10)
+	ctx, cancel := context.WithCancel(s.ctx)
```
laskoviymishka, line 1015: Missing the `defer cancel()` that both other `WithCancel` tests in this file pair with. `stream`'s cleanup does call `r.cancel()`, but only once the goroutine is running, so if the test bails between here and `go writer.stream(...)` the context leaks (and `go vet`/lint flags the missing cancel). One line right below.

laskoviymishka (changes requested):
Left a handful of smaller things inline. I'm deliberately not re-opening the concurrent-Add barrier or the test-coverage gap, since zeroshade already flagged both on the thread and I agree with where he landed, so these are the separate items I noticed on top of that.

The most substantive of them is the `r.cancel()` added to the cleanup defer: it runs on every `stream()` exit, including the clean channel-closed path, so on success the context now gets cancelled by the stream goroutine before `closeAndWait` would have done it. It's idempotent so nothing breaks today, but it's a quiet semantic shift, and anyone who later reads `ctx.Err()` after the wait would see a cancelled context on the happy path. I'd guard it so it only fires on the error exit.

The rest are cleanup: the new test leaks the `iter.Pull` goroutine (missing `defer factory.closeAll()`) and skips `defer cancel()`, both breaking the pattern every other test in the file follows, and the "recordCh is never closed here" comment reads as if the `!ok` branch below it is dead code when it's actually reachable via `closeInput`. All noted inline.

deepak-2605:
The 4 failing lint jobs aren't from this PR , they're a pre-existing intrange violation in catalog/rest/load_table_bench_test.go (lines 84, 101), which this PR doesn't touch. build/CodeQL pass here and the change is good under go test -race.

zeroshade (approved):
LGTM just needs to rebase onto main to fix the linting issue

laskoviymishka (approved):
LGTM!

@deepak-2605 can you plz resolve conflicts?

deepak-2605:
Both reviews are approved and 13/14 checks pass. The only red is ubuldn't re-run it from the fork. tu-latest go1.25.9, which failed at "Run tests" with a bare exit code 1 (2m16s, not a timeout) while the other 3 matrix jobs passed on this same commit  looks like a flaky test in go test ./.... Could someone re-run that job? Once it's green this should be ready to merge. (I can't re-run from the fork.)

PR #1854, opened 2026-08-20
fix(manifest): inherit writer content when creating manifest files (#1853)

### LINKED ISSUE #1853: Manifest: ToManifestFile hardcodes content to "data", ignoring the writer's content
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Right now, `ManifestWriter` records its content type twice. `ManifestWriter.meta()` writes the writer's content (`w.content`) into the manifest file's own Avro metadata, but `ToManifestFile` hardcodes the `Content` field of the returned `ManifestFile` to `ManifestContentData` instead of using `w.content`.

A writer created with `WithManifestWriterContent(ManifestContentDeletes)` therefore produces a `ManifestFile` that claims to be a data manifest. Committing it into a manifest list yields metadata where the list entry and the manifest file disagree.

Reading such a table back fails in `NewManifestReader`:
```
    manifest file's 'content' metadata indicates "deletes", but entry from manifest
    list indicates "data"
```
Callers can only avoid this by passing the redundant option `ToManifestFile(path, n, WithManifestFileContent(ManifestContentDeletes))`. Every call site in the codebase, right now, does exactly that.

### Status

Will raise a PR for this fix.

badalprasadsingh:
### Description

Fixes #1853 

Earlier, `ToManifestFile` hardcoded the manifest list entry's content to `data`, so a deletes writer produced an entry contradicting the manifest's own metadata, and reading it back failed. 

It now inherits the writer's content, and a conflicting `WithManifestFileContent` is rejected instead of silently written.

### Testing

Added the regression test for it.

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -1476,8 +1476,10 @@ func (w *ManifestWriter) Close() error {
 
 type ManifestFileOption func(mf *manifestFile)
 
-// WithManifestFileContent overrides the ManifestContent of a new manifest file with the provided value
-// Default: ManifestContentData
+// WithManifestFileContent sets the ManifestContent of a new manifest file.
+// The default is the content the ManifestWriter was created with, so this is
+// only needed to state that content explicitly. Passing a value that disagrees
+// with the writer makes ToManifestFile fail.
 func WithManifestFileContent(content ManifestContent) ManifestFileOption {
 	return func(mf *manifestFile) {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):

badalprasadsingh:
Thanks :)

PR #1856, opened 2026-08-20
fix(table): fail closed when a partition spec id is not registered (#1855)

### LINKED ISSUE #1855: SnapshotProducer: silently substitutes an empty partition spec when the spec id is unknown
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

The `snapshotProducer.spec(id)` helper returns `NewPartitionSpec()` when a partition spec id is not registered in table metadata.

Three manifest-writing paths use it: added delete manifests, mainfest merging, and deleted-entry tombstone manifests. A delete file with an unregistered spec id **commits without error** into a manifest declaring the unpartitioned spec 0, while its entry carries a different partition tuple.

This **corrupts** metadata, no warning. The helper should fail closed.

### Status

Will raise a PR for it ASAP.

badalprasadsingh:
### Description

Fixes #1855, a long running issue, unnoticed till #1818.

Previous to it, `snapshotProducer.spec(id)` returned an empty partition spec when a spec id was not registered in table metadata. Four callers relied on it. So, a delete-file carrying an unknown spec id committed happily into a manifest declaring the unpartitioned spec 0, while its entry still held a different partition tuple. 

That is corrupt metadata written with no error at all. As I said earlier, #1818 did spot this but it fixed the added-data-file path only.

This moves the check into the helper itself so every path behaves the same way. 

### Testing

Added regression tests wherever required.

### DIFF
--- a/table/row_delta_test.go
+++ b/table/row_delta_test.go
@@ -1534,3 +1534,57 @@ func TestRowDeltaWritesRegisteredNonDefaultSpec(t *testing.T) {
 		assert.Equal(t, 1, entries)
 	}
 }
+
+func TestRowDeltaMixedSpecCommitLeavesNoManifests(t *testing.T) {
+	validData := func() iceberg.DataFile { return buildDataFile(t, "s3://bucket/data/insert.parquet") }
+	validDelete := func() iceberg.DataFile { return buildPosDeleteFile(t, "s3://bucket/data/pos-del.parquet") }
+	orphanData := func() iceberg.DataFile {
+		return buildDataFileForSpec(t, unregisteredSpec(), "s3://bucket/data/orphan.parquet", map[int]any{1000: int64(7)})
+	}
+	orphanDelete := func() iceberg.DataFile {
+		return buildPosDeleteFileForSpec(t, unregisteredSpec(), "s3://bucket/data/orphan-del.parquet", map[int]any{1000: int64(7)})
+	}
+
+	tests := []struct {
+		name    string
+		rows    []iceberg.DataFile
+		deletes []iceberg.DataFile
+	}{
+		{"valid data with unregistered delete", []iceberg.DataFile{validData()}, []iceberg.DataFile{orphanDelete()}},
+		{"valid delete with unregistered data", []iceberg.DataFile{orphanData()}, []iceberg.DataFile{validDelete()}},
+		{"valid data group before unregistered one", []iceberg.DataFile{validData(), orphanData()}, nil},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tbl := newRowDeltaCommitTestTable(t)
+
+			tx := tbl.NewTransaction()
+			rd := tx.NewRowDelta(nil)
+			rd.AddRows(tt.rows...)
+			rd.AddDeletes(tt.deletes...)
+
+			err := rd.Commit(t.Context())
+			require.Error(t, err)
+			assert.ErrorIs(t, err, table.ErrPartitionSpecNotFound)
+			assert.Empty(t, writtenManifests(t, tbl.Location()),
+				"a rejected row delta must not leave manifests behind")
+
+			validTx := tbl.NewTransaction()
+			validRd := validTx.NewRowDelta(nil)
+			validRd.AddRows(validData())
+			require.NoError(t, validRd.Commit(t.Context()))
+			assert.NotEmpty(t, writtenManifests(t, tbl.Location()),
+				"the same probe must observe the manifests a successful commit writes")
+		})
+	}
+}
+
+func writtenManifests(t *testing.T, location string) []string {
+	t.Helper()
+
+	paths, err := filepath.Glob(filepath.Join(location, "metadata", "*.avro"))
+	require.NoError(t, err)
+
+	return paths
+}
--- a/table/row_delta_test.go
+++ b/table/row_delta_test.go
@@ -1427,3 +1427,110 @@ func assertRowCount(t *testing.T, tbl *table.Table, expected int64) {
 
 	assert.Equal(t, expected, total, "unexpected row count")
 }
+
+func unregisteredSpec() iceberg.PartitionSpec {
+	return iceberg.NewPartitionSpecID(99, iceberg.PartitionField{
+		SourceIDs: []int{1}, FieldID: 1000, Name: "id_identity", Transform: iceberg.IdentityTransform{},
+	})
+}
+
+func buildPosDeleteFileForSpec(t *testing.T, spec iceberg.PartitionSpec, path string, partition map[int]any) iceberg.DataFile {
+	t.Helper()
+
+	b, err := iceberg.NewDataFileBuilder(
+		spec, iceberg.EntryContentPosDeletes,
+		path, iceberg.ParquetFile, partition, nil, nil, 5, 512)
+	require.NoError(t, err)
+
+	return b.Build()
+}
+
+func buildDataFileForSpec(t *testing.T, spec iceberg.PartitionSpec, path string, partition map[int]any) iceberg.DataFile {
+	t.Helper()
+
+	b, err := iceberg.NewDataFileBuilder(
+		spec, iceberg.EntryContentData,
+		path, iceberg.ParquetFile, partition, nil, nil, 10, 1024)
+	require.NoError(t, err)
+
+	return b.Build()
+}
+
+func TestRowDeltaRejectsDeleteFileWithUnregisteredSpec(t *testing.T) {
+	tbl := newRowDeltaCommitTestTable(t)
+
+	tx := tbl.NewTransaction()
+	rd := tx.NewRowDelta(nil)
+	rd.AddDeletes(buildPosDeleteFileForSpec(t, unregisteredSpec(),
+		"s3://bucket/data/pos-del.parquet", map[int]any{1000: int64(7)}))
+
+	err := rd.Commit(t.Context())
+	require.Error(t, err)
+	assert.ErrorIs(t, err, table.ErrPartitionSpecNotFound)
+	assert.ErrorContains(t, err, "99")
+
+	result, err := tx.Commit(t.Context())
+	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


table/snapshot_producers.go
```diff
@@ -1002,7 +1013,12 @@ func (sp *snapshotProducer) writeAddedManifest(content iceberg.ManifestContent,
 }
 
 func (sp *snapshotProducer) writeAddedDeleteManifest(specID int, additions []deleteFileAddition) (_ iceberg.ManifestFile, retErr error) {
-	wr, path, counter, out, err := sp.newManifestWriter(sp.spec(specID), iceberg.WithManifestWriterContent(iceberg.ManifestContentDeletes))
+	spec, err := sp.spec(specID)
```
zeroshade, line 1064: Blocking: this new error path can be reached after another manifest has already been persisted. `addedContentManifests` runs the data and delete producers in parallel, so a RowDelta containing valid data plus a delete file with an unregistered spec writes the data manifest, returns `ErrPartitionSpecNotFound` here, and then `errgroup.Wait` discards the successful manifest path. Since no snapshot is staged, that file is never cleaned up. A focused probe left `metadata/<uuid>-m1.avro`; an earlier valid group followed by an invalid group has the same problem. Please preflight all referenced spec IDs before creating any output, or remove all completed manifests on failure, and add a mixed valid/invalid test asserting that no manifest remains.

zeroshade (changes requested):
Thanks for making partition-spec resolution fail closed and propagating the errors through the snapshot producer. I found one failure-atomicity issue in the new error paths: a rejected mixed-spec operation can leave persisted manifests that no snapshot references. Please validate the referenced specs before starting any writers, or clean up every successfully written manifest when a sibling/group fails, and cover the mixed valid/invalid case with a no-orphans regression test.

table/snapshot_producers.go
badalprasadsingh: Done.

Spec ids, now, preflighted before any writer is created. So, nothing is written. Added no-orphan tests. Also, preflighted the tombstone path and made its group _order deterministic_, so the orphan test is not flaky.

badalprasadsingh:
The CIs will stop failing once #1863 is merged

table/snapshot_producers.go
```diff
@@ -788,6 +798,10 @@ func (sp *snapshotProducer) buildManifests(ctx context.Context, parent *Snapshot
 // written once and reused verbatim across OCC retries (rewriting them would
 // churn manifest paths and orphan object-store files on every attempt).
 func (sp *snapshotProducer) addedContentManifests() ([]iceberg.ManifestFile, error) {
+	if err := sp.validateAddedSpecs(); err != nil {
```
laskoviymishka: Looking at this validation path on its own, I think it closes the sibling-goroutine race but not the cross-phase one. `buildManifests` calls `parentDependentManifests` first, and that can write tombstone manifests for deleted entries before we ever reach `addedContentManifests`, where `validateAddedSpecs` actually runs.

So an overwrite with valid parent-dependent deletions plus an added file on an unregistered spec writes the tombstones, then rejects, and those manifests are orphaned. Same shape on the summary side: the new `removeFile` closure in `accumulateSummaryDelta` errors on an unregistered deleted-file spec, but that runs after `buildManifests` has already written everything.

I'd lift a single validation to the very top of `buildManifests`, before either phase, and widen it to cover the to-be-deleted spec ids too (`deletedFiles` / `deletedDeleteFiles` / `deletedDVsByRef`), not just `addedFiles` — so nothing touches object storage until every spec id in the commit is known-good. I left a note on `TestRowDeltaMixedSpecCommitLeavesNoManifests` about the test that would guard this.

table/snapshot_producers.go
```diff
@@ -626,12 +632,16 @@ func createSnapshotProducer(op Operation, txn *Transaction, fs iceio.WriteFileIO
...
 
-func (sp *snapshotProducer) spec(id int) iceberg.PartitionSpec {
-	if spec, _ := sp.txn.meta.GetSpecByID(id); spec != nil {
-		return *spec
+func (sp *snapshotProducer) spec(id int) (iceberg.PartitionSpec, error) {
+	spec, err := sp.txn.meta.GetSpecByID(id)
+	if err == nil && spec == nil {
+		err = ErrPartitionSpecNotFound
+	}
+	if err != nil {
```
laskoviymishka, line 637: I'd reserve the "unregistered" wording for the actual not-found case. As written, any error out of `GetSpecByID` (a future decode/IO failure, say) comes back to the caller labeled `"unregistered partition spec id N"`, which wouldn't be true.

The `if err == nil && spec == nil` guard also looks dead — `GetSpecByID` returns either `(*spec, nil)` or `(nil, err)`, never `(nil, nil)` — so the not-found signal is already the error it hands back. Something like:
```go
spec, err := sp.txn.meta.GetSpecByID(id)
if errors.Is(err, ErrPartitionSpecNotFound) {
    return iceberg.PartitionSpec{}, fmt.Errorf("unregistered partition spec id %d: %w", id, err)
}
if err != nil {
    return iceberg.PartitionSpec{}, err
}
return *spec, nil
```
Minor while we're here: the id lands in the message twice (`"unregistered partition spec id 99: partition spec not found: id 99"`) — worth trimming one.

table/row_delta_test.go
```diff
@@ -1427,3 +1427,164 @@ func assertRowCount(t *testing.T, tbl *table.Table, expected int64) {
...
+		deletes []iceberg.DataFile
+	}{
+		{"valid data with unregistered delete", []iceberg.DataFile{validData()}, []iceberg.DataFile{orphanDelete()}},
+		{"valid delete with unregistered data", []iceberg.DataFile{orphanData()}, []iceberg.DataFile{validDelete()}},
+		{"valid data group before unregistered one", []iceberg.DataFile{validData(), orphanData()}, nil},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tbl := newRowDeltaCommitTestTable(t)
```
laskoviymishka, line 1560: This is the test that would catch the ordering gap I flagged in `snapshot_producers.go`, but it can't as written — `newRowDeltaCommitTestTable` gives a fresh table, so `deletedEntries()` is empty and no tombstones get written before `validateAddedSpecs` fires. Every subtest here rejects during the added-file phase with nothing on disk yet.

`TestParentDependentManifestsMixedSpecWritesNoManifests` has the same limitation — it drives `parentDependentManifests` in isolation, never the full `buildManifests` where the parent-dependent phase succeeds and the added-spec phase then fails.

Once the validation moves up, I'd add a case that commits a file first, then runs an overwrite with a valid deletion plus an unregistered added-file spec, and asserts no new manifests. That's the path that's currently unguarded.

table/row_delta_test.go
```diff
@@ -1427,3 +1427,164 @@ func assertRowCount(t *testing.T, tbl *table.Table, expected int64) {
 
 	assert.Equal(t, expected, total, "unexpected row count")
 }
+
+func unregisteredSpec() iceberg.PartitionSpec {
+	return iceberg.NewPartitionSpecID(99, iceberg.PartitionField{
+		SourceIDs: []int{1}, FieldID: 1000, Name: "id_identity", Transform: iceberg.IdentityTransform{},
```
laskoviymishka, line 1433: Small one: this and `unregisteredSpec()` in `snapshot_producers_test.go` (around line 2033) are near-duplicates both hard-coding id 99, but with different `Name` values — `"id_identity"` here, `"id"` there. Different packages so they can't be shared, but the mismatch reads as accidental, and anyone who later registers spec 99 or evolves past it breaks both files confusingly. I'd at least align the `Name`.

laskoviymishka (changes requested):
Nice fix — failing closed on an unregistered spec id is the right call, and it's cleaner than Java here, which resolves the id to null and then NPEs at manifest-writer construction. The `spec()` signature change and the caller propagation all read correctly.

I'd hold it on one thing before merge. Reading the new `validateAddedSpecs` path on its own, it closes the within-`addedContentManifests` race but not the cross-phase one: `buildManifests` runs `parentDependentManifests` first, which can write tombstone manifests for deleted entries, and only then reaches `addedContentManifests` where the validation runs. So an overwrite with valid parent-dependent deletions plus an added file on an unregistered spec writes the tombstones, then rejects, and leaves them orphaned. The `removeFile` closure in `accumulateSummaryDelta` has the same shape on the summary side. I'd lift one validation to the top of `buildManifests` before either phase, widened to cover the to-be-deleted spec ids as well as the added ones.

The mixed-spec tests won't catch this as written — they all run on fresh tables, so `deletedEntries()` is empty and nothing gets written before the reject fires. Once the check moves up, a case that commits a file first, then runs an overwrite with a valid deletion plus an unregistered added-file spec and asserts no new manifests, would guard the path that's actually unprotected today.

Left a couple of smaller inline notes too (the "unregistered" wrapper mislabeling non-not-found errors, a dead nil guard, and a duplicated test helper). Once the ordering and the regression test are in, happy to take another pass.

table/snapshot_producers.go
badalprasadsingh: Done.

Now we have single validateSpecs in buildManifests and assembleManifests, covering deletedFiles, deletedDeleteFiles and deletedDVsByRef.

table/snapshot_producers.go
badalprasadsingh: Done.

Dropped the wrapper entirely : GetSpecByID's error already used to name the id. Dead nil guard removed, no relabelling.

table/row_delta_test.go
badalprasadsingh: Done.

Replaced with a buildManifestss test whose parent hold a live Deletion Vector, so tombstones would be written first.

table/row_delta_test.go
badalprasadsingh: Done.

zeroshade (approved):
LGTM, i'll leave it to @laskoviymishka to double check given the existing request-changes review

laskoviymishka (approved):
LGTM

PR #1858, opened 2026-08-21
fix(table): fence multi-table commits with the target branch requirement (#1844)

### LINKED ISSUE #1844: Transaction: TableCommit() omits the target branch snapshot requirement on multi-table commits
### Apache Iceberg version

None

### Please describe the bug 🐞

Right now, `Commit()` in `Transaction` adds a requirement that the target branch still points at the snapshot the transaction read. `TableCommit` in `Transaction`, which builds the payload for multi-table commits, does not.

Snapshot-productions txns add that requirement themselves, so they are fine. Metadata-only ones (`SetProperties`, `UpdateSchema`) do not, so their multi-table commit succeeds even if another writer moved the branch - the writer's read-modify-write silently breaks.

Both paths were matching until #1657, which routed only `Commit` through the shared helper.

EDIT: 
### Status

Will raise a PR for the fix soon.

badalprasadsingh:
### Description

Fixes #1844.

From now on, `TableCommit()` derives requirements with the same helper `Commit()` uses, so multi-tables payloads fence the _target_ branch.

Previously, metadata-only txns shipped no ref assertion and could clobber a concurrent writer.

### Testing

Added the required tests.

### DIFF
--- a/catalog/multi_table_transaction.go
+++ b/catalog/multi_table_transaction.go
@@ -102,8 +102,10 @@ func (m *MultiTableTransaction) AddTransaction(tx *table.Transaction) error {
 
 // Commit extracts pending changes from all added transactions and
 // commits them atomically. On success, all transactions are marked
-// as committed. On failure, no transactions are marked committed
-// and the caller may retry.
+// as committed. On failure, no transactions are marked committed.
+//
+// A retry must be rebuilt from freshly loaded tables:
+// each payload asserts the branch head its transaction read.
 //
 // PostCommit hooks are not executed. Because the multi-table commit
 // endpoint returns 204 No Content, callers must LoadTable individually
--- a/catalog/multi_table_transaction.go
+++ b/catalog/multi_table_transaction.go
@@ -72,6 +72,9 @@ func NewMultiTableTransaction(cat Catalog) (*MultiTableTransaction, error) {
 // with all other transactions in this multi-table transaction.
 // Returns an error if the transaction is nil, already committed, or
 // targets a table that was already added.
+//
+// A transaction may still be empty here and gain updates before Commit,
+// so emptiness is only decided when the payloads are built.
 func (m *MultiTableTransaction) AddTransaction(tx *table.Transaction) error {
 	if tx == nil {
 		return errors.New("transaction must not be nil")
@@ -107,6 +110,10 @@ func (m *MultiTableTransaction) AddTransaction(tx *table.Transaction) error {
 // A retry must be rebuilt from freshly loaded tables:
 // each payload asserts the branch head its transaction read.
 //
+// Transactions that stage no updates are left out of the request, since
+// a catalog may reject a table change carrying none. They are still
+// marked committed. A request with nothing left to send is not made.
+//
 // PostCommit hooks are not executed. Because the multi-table commit
 // endpoint returns 204 No Content, callers must LoadTable individually
 // to obtain updated metadata.
@@ -120,18 +127,26 @@ func (m *MultiTableTransaction) Commit(ctx context.Context) error {
 	}
 
 	commits := make([]table.TableCommit, 0, len(m.txns))
-	for _, tx := range m.txns {
+	names := make([]string, 0, len(m.txns))
+	for i, tx := range m.txns {
 		tc, err := tx.TableCommit()
 		if err != nil {
 			return err
 		}
 
+		if len(tc.Updates) == 0 {
+			continue
+		}
+
 		commits = append(commits, tc)
+		names = append(names, m.tableNames[i])
 	}
 
-	if err := m.cat.CommitTransaction(ctx, commits); err != nil {
-		return fmt.Errorf("commit transaction for tables [%s]: %w",
-			strings.Join(m.tableNames, ", "), err)
+	if len(commits) > 0 {
+		if err := m.cat.CommitTransaction(ctx, commits); err != nil {
+			return fmt.Errorf("commit transaction for tables [%s]: %w",
+				strings.Join(names, ", "), err)
+		}
 	}
 
 	m.committed = true
--- a/catalog/multi_table_transaction_test.go
+++ b/catalog/multi_table_transaction_test.go
@@ -391,7 +391,7 @@ func TestMultiTableTransactionFailsAfterConcurrentAdvance(t *testing.T) {
 	assert.Equal(t, 1, cat.calls)
 }
 
-// Why: a catalog may reject a table change with no updates, which would fail the whole batch 
+// Why: a catalog may reject a table change with no updates, which would fail the whole batch
 // over an entry that changes nothing.
 func TestMultiTableTransactionSkipsTransactionsWithoutUpdates(t *testing.T) {
 	t.Run("empty transactions are left out of the request", func(t *testing.T) {
--- a/catalog/multi_table_transaction_test.go
+++ b/catalog/multi_table_transaction_test.go
@@ -277,3 +277,78 @@ func TestCommitAndReloadPartialFailure(t *testing.T) {
 	// First table was loaded successfully before the second failed.
 	assert.Len(t, tables, 1)
 }
+
+func mtxTableWithHead(t *testing.T, name string, headID, childID int64) (*table.Table, table.Metadata) {
+	t.Helper()
+
+	base := mtxTestTable(t, "db", name).Metadata()
+	withHead := mtxGraftSnapshot(t, base, headID, nil)
+	advanced := mtxGraftSnapshot(t, withHead, childID, &headID)
+
+	return table.New(table.Identifier{"db", name}, withHead, "", nil, nil), advanced
+}
+
+func mtxGraftSnapshot(t *testing.T, base table.Metadata, id int64, parent *int64) table.Metadata {
+	t.Helper()
+
+	builder, err := table.MetadataBuilderFromBase(base, "")
+	require.NoError(t, err)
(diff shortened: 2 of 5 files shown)


tanmayrauth (approved):
LGTM,  routing TableCommit through the same transactionRequirements path as Commit correctly fences metadata-only payloads, and the tests cover the boundary well.

catalog/multi_table_transaction.go
```diff
@@ -102,8 +102,10 @@ func (m *MultiTableTransaction) AddTransaction(tx *table.Transaction) error {
 
 // Commit extracts pending changes from all added transactions and
 // commits them atomically. On success, all transactions are marked
-// as committed. On failure, no transactions are marked committed
-// and the caller may retry.
+// as committed. On failure, no transactions are marked committed.
+//
+// A retry must be rebuilt from freshly loaded tables:
```
laskoviymishka, line 110: A metadata-only transaction with no staged updates returns `{Requirements: [], Updates: []}` from `TableCommit`, and `Commit` here forwards every collected transaction unconditionally. A conformant REST catalog can reject a `TableChange` with zero updates (400/422), which would fail the whole atomic batch over one no-op entry.

I'd either filter zero-update transactions out in `Commit`, or document on `AddTransaction`/`Commit` that a no-update transaction shouldn't be added. wdyt?

catalog/multi_table_transaction_test.go
```diff
@@ -277,3 +277,78 @@ func TestCommitAndReloadPartialFailure(t *testing.T) {
...
+	stub := &stubCatalog{}
+	mtx := &MultiTableTransaction{cat: stub}
+
+	for _, tbl := range []*table.Table{tbl1, tbl2} {
+		tx := tbl.NewTransaction()
+		require.NoError(t, tx.SetProperties(map[string]string{"offsets": "42"}))
+		require.NoError(t, mtx.AddTransaction(tx))
+	}
+
+	require.NoError(t, mtx.Commit(context.Background()))
```
laskoviymishka: Small consistency thing: the new tests in `transaction_internal_test.go` all use `t.Context()`, but this one reaches for `context.Background()`. I'd match `t.Context()` here.

table/commit.go
```diff
@@ -77,9 +78,9 @@ func (t *Transaction) TableCommit() (TableCommit, error) {
 		}, nil
 	}
 
-	reqs := make([]Requirement, len(t.reqs), len(t.reqs)+1)
-	copy(reqs, t.reqs)
-	reqs = append(reqs, AssertTableUUID(meta.uuid))
+	// Same derivation as Commit. Copying t.reqs alone leaves metadata-only transactions unfenced:
+	// their updates stage no ref requirement.
+	reqs := append(transactionRequirements(t.reqs, t.branch, t.tbl.metadata), AssertTableUUID(meta.uuid))
```
laskoviymishka, line 84: This is safe today only because `transactionRequirements` always starts with `slices.Clone(reqs)`, so the returned slice never aliases `t.reqs`. That's a non-local invariant now, and we're appending into the result while holding `t.mx`; the old `make + copy` spelled the ownership out.

I'd add a one-liner here noting the returned slice is a fresh allocation, or state that guarantee in `transactionRequirements`'s doc, so a future fast-path return can't silently reintroduce aliasing.

table/transaction_internal_test.go
```diff
@@ -1785,3 +1786,258 @@ func TestRollbackToSnapshotCommitRejectsBranchTurnedTag(t *testing.T) {
...
+	t.Run("target name created as a tag is rejected", func(t *testing.T) {
+		base := newConflictTestMetadata(t, &head)
+		tx := multiTableTestTable(t, base).NewTransactionOnBranch("audit")
+		require.NoError(t, tx.SetProperties(iceberg.Properties{"offsets": "42"}))
+
+		tc, err := tx.TableCommit()
+		require.NoError(t, err)
+
+		req := soleRefRequirement(t, tc.Requirements, "audit")
+		err = req.Validate(metadataWithRef(t, base, "audit", TagRef))
```
laskoviymishka, line 1942: This sub-case proves the tag guard fires through a local `Validate` call, but that isn't what runs on the multi-table path this PR is fencing. `MultiTableTransaction.Commit` hands the payload to `CommitTransaction` with no local retry or `Validate` loop, and `requireBranch` has no wire representation, so the server only sees `{"type":"assert-ref-snapshot-id","ref":"audit","snapshot-id":...}` and enforces snapshot-id equality, not ref type.

So the "created as a tag is rejected" protection is real for single-table `Commit` but silently absent once the payload goes out over REST. I'd scope this with a comment to client-side validation, so it isn't read as multi-table tag-race coverage. wdyt?

table/transaction_internal_test.go
```diff
@@ -1785,3 +1786,258 @@ func TestRollbackToSnapshotCommitRejectsBranchTurnedTag(t *testing.T) {
...
+			committed := New(Identifier{"db", "multi-table"}, base, "metadata.json",
+				func(context.Context) (iceio.IO, error) { return iceio.LocalFS{}, nil }, cat)
+
+			txCommit := committed.NewTransactionOnBranch(branch)
+			require.NoError(t, txCommit.SetProperties(iceberg.Properties{"offsets": "42"}))
+			_, err := txCommit.Commit(t.Context())
+			require.NoError(t, err)
+			require.Len(t, cat.reqs, 1)
+
+			txPayload := multiTableTestTable(t, base).NewTransactionOnBranch(branch)
```
laskoviymishka: This passes only because `txPayload` is rebuilt from `base`, the same starting metadata the `Commit` side used. `reqCapturingCatalog.CommitTable` advances `c.metadata` in place, so if someone later seeds this from the post-commit metadata, the two sides would start from different heads and the equality could pass for the wrong reason.

I'd assert the two transactions start from identical metadata, or at least leave a comment pinning that assumption.

table/transaction_internal_test.go
```diff
@@ -1785,3 +1786,258 @@ func TestRollbackToSnapshotCommitRejectsBranchTurnedTag(t *testing.T) {
...
+	head := int64(100)
+	base := newConflictTestMetadata(t, &head)
+
+	tx := multiTableTestTable(t, base).NewTransaction()
+	require.NoError(t, tx.AssertRefSnapshotID(MainBranch))
+
+	tc, err := tx.TableCommit()
+	require.NoError(t, err)
+
+	assert.Empty(t, tc.Requirements)
```
laskoviymishka: The `NotNil` pair already catches a nil return, but the intent, a non-nil zero-length slice that serializes as `[]` and not `null`, reads more directly as `assert.Equal(t, []Requirement{}, tc.Requirements)` (same for `Updates`). One assertion, exact intent.

laskoviymishka (changes requested):
Nice fix. Routing `TableCommit` through `transactionRequirements` is exactly right, metadata-only multi-table payloads were shipping with no branch assertion at all, and now they carry the same base-head fence `Commit` does. I know this already has a green light from tanmayrauth, and I agree the routing is correct; the whole-list equality test guarding the two paths against drift is a nice touch.

I'd hold this before merging though, on a couple of things that sit just outside what that approval covered.

The one I actually care about: a transaction with no staged updates yields an empty `{[], []}` payload, and `MultiTableTransaction.Commit` collects and forwards every transaction unconditionally. A conformant REST catalog can reject a `TableChange` with zero updates, which would fail the whole atomic batch over a single no-op entry. I'd either filter zero-update transactions in `Commit` or document on `AddTransaction`/`Commit` that they mustn't be added.

The other is honesty of coverage: the "created as a tag is rejected" test exercises a local `Validate` call, but that path never runs on the multi-table commit. `requireBranch` has no wire form, so the server only enforces snapshot-id equality, not ref type. The guard is real for single-table `Commit`; it just doesn't extend to the REST path this PR is fencing, and the test currently reads as if it does.

- filter or document empty-payload transactions in the multi-table `Commit` path
- scope the tag-rejection test to client-side validation so it isn't read as multi-table coverage
- a one-liner on the `append` in `commit.go` noting `transactionRequirements` returns a fresh slice (the alias-safety is a non-local invariant now)
- lock `TestTableCommitRequirementsMatchCommit` to identical starting metadata on both sides

One thing I didn't turn into a request: the doc says "a retry must be rebuilt from freshly loaded tables," but nothing tests that a second `Commit` after a concurrent advance actually fails, since the stub catalogs don't validate requirements. Worth a follow-up if you'd rather not grow this PR.

catalog/multi_table_transaction.go
badalprasadsingh: Done.

I filtered this in `Commit`. Documenting at `AddTransaction` wouldn't work - a txn can gain updates after being added. Tests added.

catalog/multi_table_transaction_test.go
badalprasadsingh: Done.

table/commit.go
badalprasadsingh: Done.

table/transaction_internal_test.go
badalprasadsingh: Done.

`requireBranch` had no wire form. Renamed to "...locally" and scoped with a comment.

table/transaction_internal_test.go
badalprasadsingh: Done.

Both tables now built from `base` up front, with an assertion pinning identical starting metadata.

table/transaction_internal_test.go
badalprasadsingh: Done.

laskoviymishka (approved):
👍

PR #1859, opened 2026-08-21
docs(readme): correct inaccuracies

badalprasadsingh:
Took a quick look at the `README.md` that we have for the project, and found that it is not up to date with the current developments in the project.

A quick PR to update it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 
 ### Prerequisites
 
-* Go 1.25 or later
+* Go 1.25.9 or later
 
 ### Build
 
@@ -73,7 +73,7 @@ make lint-install
    export AWS_S3_ENDPOINT=http://$(docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' minio):9000
    export AWS_REGION=us-east-1
    export SPARK_CONTAINER_ID=$(docker ps -qf 'name=spark-iceberg')
-   export DOCKER_API_VER=$(docker version -f '{{.Server.APIVersion}}')
+   export DOCKER_API_VERSION=$(docker version -f '{{.Server.APIVersion}}')
    ```
 
 3. Run the integration tests:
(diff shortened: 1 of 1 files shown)


tanmayrauth (approved):

README.md
```diff
@@ -25,7 +25,7 @@
 
 ### Prerequisites
 
-* Go 1.25 or later
+* Go 1.25.9 or later
```
laskoviymishka: `go.mod` declares `go 1.25.8` (and CI runs `1.25.8`), so this now advertises a patch version one ahead of what the module actually requires. I'd either match go.mod exactly (`Go 1.25.8 or later`) or just keep the minor form (`Go 1.25 or later`), which stays forward-safe. wdyt?

laskoviymishka (changes requested):
Nice cleanup, and mostly accurate. The `DOCKER_API_VERSION` rename checks out against the Makefile and `local_spark.go`, the Load View and Rewrite manifests rows both match code that's actually wired up, and the `integration-hive` / `integration-hadoop` suites are real targets.

The one thing I'd fix before merge is the Go version. Bumping `Go 1.25 or later` to `1.25.9` reads as a tightening, but `go.mod` says `go 1.25.8`, so the README now advertises a patch that's ahead of what the module actually requires, which is the opposite of what this PR is going for. Matching go.mod (`Go 1.25.8 or later`), or just reverting to the minor `1.25` form, closes it.

Fix that one line and this is good to land.

README.md
badalprasadsingh: I clarified it, currently:

1. `go.mod` is at `1.25.9`
2. The CI matrix is `['1.25.9', '1.26.1']`

That said, I’m happy to go with the minor version as you suggested. Since the patch version gets bumped roughly every two months, pinning a specific patch version here probably wouldn’t be ideal in the long run. :)

badalprasadsingh:
A re-run of the failed CI should solve the issue.

laskoviymishka (approved):

PR #1861, opened 2026-08-22
fix(manifest): handle foreign time/timestamp partition logical types

tanmayrauth:
Partition decoding asserted v.(int64) unchecked on every time and timestamp arm. A manifest that declares a time/timestamp logicalType on an underlying Avro type the spec forbids (e.g. timestamp-millis on an int) decodes to a raw int32, panicking Partition() with no way to recover. time-millis and timestamp-millis were also returned as milliseconds though iceberg.Time and iceberg.Timestamp count microseconds, 1000x too small.

Accept the raw int32/int64 value and scale milliseconds to microseconds.

Closes #1487


zeroshade (approved):
This looks good to me, just needs to have the linting issues fixed.

PR #1836, opened 2026-08-14
chore: use modernize to improve code

zeroshade:
Used `golang.org/x/tools/go/analysis/passes/modernize/cmd/modernize@latest -fix` to modernize the code with proper structures and code paths.

PR #1863, opened 2026-08-22
hotfix(rest): golangci-lint

badalprasadsingh:
Resolves `golangci-lint` errors.

### DIFF
--- a/catalog/rest/load_table_bench_test.go
+++ b/catalog/rest/load_table_bench_test.go
@@ -81,7 +81,7 @@ func BenchmarkDecodeTableMetadata(b *testing.B) {
 			b.ResetTimer()
 			b.ReportAllocs()
 
-			for i := 0; i < b.N; i++ {
+			for range b.N {
 				var resp loadTableResponse
 				err := tc.decode(body, &resp)
 				require.NoError(b, err)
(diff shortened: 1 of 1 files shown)


badalprasadsingh:
cc: @zeroshade, @laskoviymishka  :)

fallintoplace:
Thanks for flagging this.

laskoviymishka (approved):
Nice!

ferhatelmas:
how was the failing code merged in the first place? what can be done to prevent it from happening again?

badalprasadsingh:
@ferhatelmas, the issue was a classic stale-base merge. #1654 introduced the issue with all checks green, but its last CI run was older and main was never merged into it after #1836 landed, (#1836 enabled `intrange` in `.golangci.yml`). It merged against a 7-day-old green result and turned main red.

This is the second time in two weeks that such an issue like this has occurred.

Possible solution: we can list a required status check in `.asf.yaml` — `strict: true` is already set but has no `contexts`, so it does nothing and stale PRs merge without re-running CI.

cc: @laskoviymishka, @zeroshade should we raise the PR for this fix?

ferhatelmas:
Thanks for the context, I think it's worth of requiring PRs to be up to date.

PR #1867, opened 2026-08-23
fix(table): validate required metadata identity fields

### LINKED ISSUE #1864: Table metadata parser accepts missing required location and v2/v3 table UUID
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

`table.ParseMetadata` accepts table metadata with required identity/location fields missing:

- `location` is accepted as missing or null for format versions 1–3.
- `table-uuid` is accepted as missing or null for format versions 2–3.

The decoded metadata silently exposes default Go values:

- `Metadata.Location()` returns `""`.
- `Metadata.TableUUID()` returns `uuid.Nil`.

Neither field is validated on the parse path. The same repo already enforces both for *view* metadata (`view/metadata.go:503-510`), with the same error wording table metadata is the outlier.

Iceberg requires `location` for v1–v3 metadata. `table-uuid` became required in v2.

Impact is silent rather than loud: `LoadLocationProvider("")` succeeds because `url.Parse("")` succeeds, so writes land at bogus relative paths; and `AssertTableUUID` compares equal across two distinct UUID-less tables.

Java enforces the same rules with two deltas worth naming up front: `Preconditions.checkArgument(formatVersion == 1 || uuid != null, ...)` matches, but Java's location check is `location != null`, so it accepts `""` rejecting empty is deliberately stricter here. Java also makes location optional at v4 (`MIN_FORMAT_VERSION_OPTIONAL_LOCATION = 4`), irrelevant while this repo caps at v3 but relevant to whoever adds v4. PyIceberg diverges: it silently generates a UUID via `default_factory=uuid.uuid4`. Erroring is preferable a fabricated UUID would corrupt `assert-table-uuid` commit requirements.

### Expected behavior

Parsing should return an error wrapping `table.ErrInvalidMetadata` when:

- `location` is missing, null, or empty in v1–v3 metadata.
- `table-uuid` is missing or null in v2/v3 metadata.

Missing `table-uuid` should remain accepted for v1 compatibility.

Revanth14:
### Description

This PR validates required identity fields when parsing table metadata:

- Rejects missing, null, or empty `location` for format versions 1–3.
- Rejects missing, null, or zero-valued `table-uuid` for format versions 2–3.
- Preserves v1 compatibility by allowing an absent `table-uuid`.
- Fixes v1-to-v2 conversion so a UUID is generated when the v1 metadata has no UUID.
- Updates affected test fixtures to contain valid required fields.

Validation is performed at the JSON parsing boundary rather than in the shared metadata validator. This keeps the existing create-via-commit staging flow valid, where metadata may temporarily have an empty location before it is persisted.

### Why are these changes needed?

Previously, invalid table metadata was silently accepted:

- A missing location became `""`.
- A missing UUID became `uuid.Nil`.

This could cause location providers to produce bogus relative paths and could make `assert-table-uuid` requirements incorrectly match unrelated UUID-less tables.

The behavior now matches the Iceberg requirements for the currently supported format versions. Two deliberate deltas from Java: `"location": ""` and an explicitly all-zeros `table-uuid` are rejected here, where Java only checks for null. Both are covered by tests. If format v4 support is added, location validation will need to account for location becoming optional.

**Note: this is a behavior change.** Metadata that previously parsed successfully may now return an error.

Real-world compatibility risk is limited:

- Normal table creation generates a UUID through `NewMetadata`.
- Builder-based v1-to-v2/v3 upgrades assign a UUID through `SetFormatVersion`.
- Create-via-commit temporarily constructs metadata with an empty location, but supplies the actual location through an update before persistence.

Malformed externally produced metadata and the direct v1-to-v2 conversion path are the primary cases affected. This behavior change may warrant a release note.

### Tests

Added coverage for:

- Missing, null, and empty locations across v1–v3.
- Missing, null, and zero UUIDs across v2–v3.
- v1 compatibility for absent UUIDs.
- UUID generation during v1-to-v2 conversion.
- Successful reparsing of converted metadata.

Validated with:

- `go test ./... -count=1`
- `go vet -tags=integration ./...`

Closes #1864

### DIFF
--- a/table/arrow_utils_internal_test.go
+++ b/table/arrow_utils_internal_test.go
@@ -42,6 +42,7 @@ import (
 func constructTestTable(t *testing.T, writeStats []string) (*metadata.FileMetaData, Metadata) {
 	tableMeta, err := ParseMetadataString(`{
 		"format-version": 2,
+		"table-uuid": "9c12d441-03fe-4693-9a96-a0705ddf69c1",
 		"last-sequence-number": 0,
         "location": "s3://bucket/test/location",
         "last-column-id": 7,
--- a/table/internal/parquet_files_test.go
+++ b/table/internal/parquet_files_test.go
@@ -155,6 +155,7 @@ func newAbortTestWriter(t *testing.T, fs iceio.WriteFileIO, fileName string) int
 func constructTestTablePrimitiveTypes(t *testing.T) (*metadata.FileMetaData, table.Metadata) {
 	tableMeta, err := table.ParseMetadataString(`{
         "format-version": 2,
+        "table-uuid": "9c12d441-03fe-4693-9a96-a0705ddf69c1",
         "last-sequence-number": 0,
         "location": "s3://bucket/test/location",
         "last-column-id": 7,
@@ -700,6 +701,7 @@ func TestNanosecondTimestampMetrics(t *testing.T) {
 
 	tableMeta, err := table.ParseMetadataString(`{
 		"format-version": 3,
+		"table-uuid": "9c12d441-03fe-4693-9a96-a0705ddf69c1",
 		"last-sequence-number": 0,
 		"location": "s3://bucket/test/location",
 		"last-column-id": 2,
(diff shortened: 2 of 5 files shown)


zeroshade (approved):
LGTM, just need to rebase onto main to pick up the linting fixes

Revanth14:
> LGTM, just need to rebase onto main to pick up the linting fixes

@zeroshade 
Rebase done and CI is green!

PR #1871, opened 2026-08-23
fix(table): validate required snapshot fields

### LINKED ISSUE #1869: Snapshot parser accepts missing required snapshot-id and timestamp-ms
### Apache Iceberg version

None

### Apache Iceberg version

main (development)

### Please describe the bug 

`table.Snapshot.UnmarshalJSON` accepts snapshot JSON when the required `snapshot-id` or `timestamp-ms` field is missing or null.

Because both fields are decoded directly into non-pointer `int64` values, missing and null values silently become zero:

```go
var snapshot table.Snapshot

err := json.Unmarshal([]byte(`{
  "snapshot-id": null,
  "timestamp-ms": 1234,
  "manifests": []
}`), &snapshot)

// err == nil
// snapshot.SnapshotID == 0
```

The same happens for `timestamp-ms`. A JSON `null` snapshot is also accepted and becomes a zero-value `Snapshot`, replacing any value already held by the receiver.

This affects the table metadata parse path as well. Removing either field from a snapshot in `table/testdata/TableMetadataV2Valid.json` still allows `table.ParseMetadata` to succeed. A null entry in the `snapshots` list is also accepted when it does not trigger another metadata consistency check.

The Iceberg specification requires both fields for format versions 1–3:

https://iceberg.apache.org/spec/#snapshots

Java rejects the same inputs: `SnapshotParser.fromJson` requires an object and reads both fields using `JsonUtil.getLong`, which rejects missing and null values:

https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SnapshotParser.java

### Impact

Malformed snapshot metadata is silently assigned default identity and timestamp values:

- A missing `snapshot-id` becomes `0` and can be treated as an actual snapshot identity.
- A missing `timestamp-ms` becomes the Unix epoch, potentially affecting time-travel resolution, ordering, inspection, and expiration behavior.
- A null snapshot entry becomes a zero-value snapshot rather than producing an error.

The existing metadata validation checks duplicate IDs and reference consistency, but cannot determine whether these fields were present in the original JSON.

### Expected behavior

Parsing should return an error wrapping `table.ErrInvalidMetadata` when:

- `snapshot-id` is missing or null.
- `timestamp-ms` is missing or null.
- A snapshot document or entry is null.

Explicit numeric zero values should remain distinguishable from missing fields and continue to parse. Failed unmarshalling should not partially replace an existing `Snapshot` value.

### Willingness to contribute

I can contribute this fix independently.

Revanth14:
### Description

Validate required snapshot fields during JSON decoding:

- Reject missing or null `snapshot-id`.
- Reject missing or null `timestamp-ms`.
- Reject null snapshot entries.
- Return errors wrapping `table.ErrInvalidMetadata`.
- Preserve explicit zero values and leave the receiver unchanged on failure.

The validation also propagates through table metadata and `add-snapshot` decoding.

### Why?

Previously, malformed snapshots silently decoded missing IDs and timestamps as `0`. This could create invalid snapshot identities and Unix-epoch timestamps.

Both fields are required by the Iceberg specification, and Java rejects them when missing or null.

### Tests

Added coverage for missing, null, and explicit-zero values, null snapshots, metadata parsing, and receiver preservation.

Validated with:

- `go test ./... -count=1`
- `go vet -tags=integration ./...`
- `golangci-lint run ./table/...`

Closes #1869

### DIFF
--- a/table/snapshots.go
+++ b/table/snapshots.go
@@ -272,9 +272,27 @@ type Snapshot struct {
 func (s *Snapshot) UnmarshalJSON(data []byte) error {
 	type Alias Snapshot
 	var next Alias
-	if err := json.Unmarshal(data, &next); err != nil {
+	// snapshot-id and timestamp-ms are required by the spec for every format
+	// version. Decode them through pointers so an absent or null field stays
+	// distinguishable from an explicit zero, which would otherwise be accepted
+	// as a valid identity or as the Unix epoch.
+	aux := struct {
+		SnapshotID  *int64 `json:"snapshot-id"`
+		TimestampMs *int64 `json:"timestamp-ms"`
+		*Alias
+	}{Alias: &next}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
 		return err
 	}
+	if aux.SnapshotID == nil {
+		return fmt.Errorf("%w: snapshot-id is absent or null", ErrInvalidMetadata)
+	}
+	if aux.TimestampMs == nil {
+		return fmt.Errorf("%w: timestamp-ms is absent or null", ErrInvalidMetadata)
+	}
+	next.SnapshotID, next.TimestampMs = *aux.SnapshotID, *aux.TimestampMs
+
 	if next.ManifestList != "" {
 		next.ManifestLocations = nil
 	}
--- a/table/snapshots.go
+++ b/table/snapshots.go
@@ -291,6 +291,7 @@ func (s *Snapshot) UnmarshalJSON(data []byte) error {
 	if aux.TimestampMs == nil {
 		return fmt.Errorf("%w: timestamp-ms is absent or null", ErrInvalidMetadata)
 	}
+	// The pointer fields shadow Alias, so json.Unmarshal does not populate next.
 	next.SnapshotID, next.TimestampMs = *aux.SnapshotID, *aux.TimestampMs
 
 	if next.ManifestList != "" {
--- a/table/snapshots_test.go
+++ b/table/snapshots_test.go
@@ -449,6 +449,11 @@ func TestSnapshotUnmarshalRequiresSnapshotIDAndTimestamp(t *testing.T) {
 			data:    `{"snapshot-id": 25, "timestamp-ms": null, "manifests": []}`,
 			wantErr: "timestamp-ms is absent or null",
 		},
+		{
+			name:    "both fields missing",
+			data:    `{"manifests": []}`,
+			wantErr: "snapshot-id is absent or null",
+		},
 	}
 
 	for _, tt := range tests {
@@ -476,21 +481,55 @@ func TestSnapshotUnmarshalAcceptsExplicitZeroValues(t *testing.T) {
 	assert.NotNil(t, snapshot.ManifestLocations)
 }
 
-func TestSnapshotUnmarshalFailureLeavesSnapshotUnchanged(t *testing.T) {
-	snapshot := Snapshot()
-	original := Snapshot()
+func TestSnapshotUnmarshalPreservesRequiredFields(t *testing.T) {
+	var snapshot table.Snapshot
+	require.NoError(t, json.Unmarshal([]byte(`{
+		"snapshot-id": 1234,
+		"timestamp-ms": 5678,
+		"manifests": []
+	}`), &snapshot))
 
-	err := json.Unmarshal([]byte(`{
-		"timestamp-ms": 1602638573590,
-		"manifest-list": "s3:/a/b/new.avro"
-	}`), &snapshot)
-	require.ErrorIs(t, err, table.ErrInvalidMetadata)
-	assert.True(t, snapshot.Equals(original), "expected snapshot to be untouched, got %s", snapshot)
+	assert.Equal(t, int64(1234), snapshot.SnapshotID)
+	assert.Equal(t, int64(5678), snapshot.TimestampMs)
+}
+
+func TestSnapshotUnmarshalFailureLeavesSnapshotUnchanged(t *testing.T) {
+	tests := []struct {
+		name string
+		data string
+	}{
+		{
+			name: "missing snapshot-id",
+			data: `{
+				"timestamp-ms": 1602638573590,
+				"manifest-list": "s3:/a/b/new.avro"
+			}`,
+		},
+		{
+			name: "missing timestamp-ms",
+			data: `{
+				"snapshot-id": 26,
+				"manifest-list": "s3:/a/b/new.avro"
+			}`,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			snapshot := Snapshot()
+			original := Snapshot()
+
+			err := json.Unmarshal([]byte(tt.data), &snapshot)
+			require.ErrorIs(t, err, table.ErrInvalidMetadata)
+			assert.True(t, snapshot.Equals(original), "expected snapshot to be untouched, got %s", snapshot)
+			assert.Equal(t, original.ManifestList, snapshot.ManifestList)
+		})
+	}
 }
 
 // A null snapshot carries neither identity nor timestamp, so it is rejected
-// rather than decoded into a zero-value snapshot. Java's SnapshotParser
-// likewise refuses a null node.
+// rather than decoded into a zero-value snapshot. Java's SnapshotParser also
+// rejects null, but at its earlier object-node precondition.
 func TestSnapshotUnmarshalRejectsNullDocument(t *testing.T) {
 	var snapshot table.Snapshot
 	err := json.Unmarshal([]byte(`null`), &snapshot)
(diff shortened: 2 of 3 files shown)


table/snapshots.go
```diff
@@ -272,9 +272,27 @@ type Snapshot struct {
...
+	if err := json.Unmarshal(data, &aux); err != nil {
 		return err
 	}
+	if aux.SnapshotID == nil {
+		return fmt.Errorf("%w: snapshot-id is absent or null", ErrInvalidMetadata)
+	}
+	if aux.TimestampMs == nil {
+		return fmt.Errorf("%w: timestamp-ms is absent or null", ErrInvalidMetadata)
+	}
+	next.SnapshotID, next.TimestampMs = *aux.SnapshotID, *aux.TimestampMs
```
laskoviymishka, line 295: This write-back is load-bearing in a way that's easy to miss. The `aux` `*int64` fields shadow the promoted `Alias` fields, so `json.Unmarshal` writes `snapshot-id`/`timestamp-ms` into `aux` and never into `next`. Drop this line and every valid snapshot silently decodes with `SnapshotID==0` and `TimestampMs==0`, with no compile error and nothing in the current tests catching it.

I'd pin the invariant with a round-trip assertion on a known non-zero id (decode `{"snapshot-id": 1234, "timestamp-ms": 5678, ...}` and assert both come back), and add a one-line comment here noting that `aux` shadows the promoted fields so the decoder never touches `next` directly. That way a future refactor can't quietly bring the zero-value bug back.

table/snapshots_test.go
```diff
@@ -421,3 +422,126 @@ func TestValidateRowLineage(t *testing.T) {
...
+func TestSnapshotUnmarshalFailureLeavesSnapshotUnchanged(t *testing.T) {
+	snapshot := Snapshot()
+	original := Snapshot()
+
+	err := json.Unmarshal([]byte(`{
+		"timestamp-ms": 1602638573590,
+		"manifest-list": "s3:/a/b/new.avro"
+	}`), &snapshot)
+	require.ErrorIs(t, err, table.ErrInvalidMetadata)
+	assert.True(t, snapshot.Equals(original), "expected snapshot to be untouched, got %s", snapshot)
```
laskoviymishka: The `Equals` check here passes a bit coincidentally. The failing input sets `manifest-list`, which the decoder does write into the intermediate `next` before the `snapshot-id` nil-check fires, so what we actually want to prove is that none of that leaked into the receiver. `Equals` only gets us there because the receiver happens to still match `original` on the fields it compares.

I'd assert directly on a field the partial decode would have touched, e.g. `assert.Equal(t, "", snapshot.ManifestList)` alongside the `Equals`, so the intent is unambiguous. And since `timestamp-ms` goes through a different branch than `snapshot-id`, a second variant here (snapshot-id present, timestamp-ms absent) would cover the symmetric case. wdyt?

table/snapshots_test.go
```diff
@@ -421,3 +422,126 @@ func TestValidateRowLineage(t *testing.T) {
...
+		{
+			name:    "missing timestamp-ms",
+			data:    `{"snapshot-id": 25, "manifests": []}`,
+			wantErr: "timestamp-ms is absent or null",
+		},
+		{
+			name:    "null timestamp-ms",
+			data:    `{"snapshot-id": 25, "timestamp-ms": null, "manifests": []}`,
+			wantErr: "timestamp-ms is absent or null",
+		},
```
laskoviymishka, line 451: One more case worth adding: both fields absent, e.g. `{"manifests": []}`, asserting the `snapshot-id` error wins. The ordering is deterministic today, but a case here documents that contract so a reorder of the two nil-checks shows up as a test change rather than a silent behavior shift.

table/snapshots_test.go
```diff
@@ -421,3 +422,126 @@ func TestValidateRowLineage(t *testing.T) {
...
+		"timestamp-ms": 1602638573590,
+		"manifest-list": "s3:/a/b/new.avro"
+	}`), &snapshot)
+	require.ErrorIs(t, err, table.ErrInvalidMetadata)
+	assert.True(t, snapshot.Equals(original), "expected snapshot to be untouched, got %s", snapshot)
+}
+
+// A null snapshot carries neither identity nor timestamp, so it is rejected
+// rather than decoded into a zero-value snapshot. Java's SnapshotParser
+// likewise refuses a null node.
```
laskoviymishka: Small accuracy note on the Java comparison. Java's `SnapshotParser` does reject a null node, but earlier and for a different reason: `fromJson` fails the `node.isObject()` precondition ("Cannot parse ... non-object"), not a missing `snapshot-id`. Go reaches the same outcome via `aux.SnapshotID` being nil after the no-op unmarshal. I'd reword the comment so it doesn't imply the two hit the same check.

tanmayrauth (approved):
LGTM!

table/snapshots.go
```diff
@@ -272,9 +272,27 @@ type Snapshot struct {
...
+
+	if err := json.Unmarshal(data, &aux); err != nil {
 		return err
 	}
+	if aux.SnapshotID == nil {
+		return fmt.Errorf("%w: snapshot-id is absent or null", ErrInvalidMetadata)
+	}
+	if aux.TimestampMs == nil {
+		return fmt.Errorf("%w: timestamp-ms is absent or null", ErrInvalidMetadata)
+	}
```
ferhatelmas, line 293: this will be doing double prefix

```
invalid metadata: invalid metadata: snapshot-id is absent or null
```

tests can't catch since only checking suffix

table/snapshots.go
```diff
@@ -272,9 +272,27 @@ type Snapshot struct {
...
 	type Alias Snapshot
 	var next Alias
-	if err := json.Unmarshal(data, &next); err != nil {
+	// snapshot-id and timestamp-ms are required by the spec for every format
+	// version. Decode them through pointers so an absent or null field stays
+	// distinguishable from an explicit zero, which would otherwise be accepted
+	// as a valid identity or as the Unix epoch.
+	aux := struct {
+		SnapshotID  *int64 `json:"snapshot-id"`
+		TimestampMs *int64 `json:"timestamp-ms"`
```
ferhatelmas, line 281: similar problem exists in other types (`SnapshotLogEntry`, `MetadataLogEntry`, etc.) as well

ferhatelmas: specifically for snapshot, how about manifests or manifest-list required check here?

table/snapshots.go
```diff
@@ -272,9 +272,27 @@ type Snapshot struct {
...
+		*Alias
+	}{Alias: &next}
```
ferhatelmas, line 283: nit: perf

I would suggest letting it parse the type itself first but if there is a default value then check existence (via `map[string]json.RawMessage`) so that price is only paid for wrong payload and regular branch isn't affected

laskoviymishka (approved):
This is clean. Using pointer fields to tell an absent or null `snapshot-id`/`timestamp-ms` apart from an explicit zero is exactly right, matches the spec's required-field rule, and the table already covers the missing, null, and zero cases well.

The one thing I'd add before merge is a guard on the write-back. The line that copies the validated pointers back into `next` is load-bearing but non-obvious: the `aux` `*int64` fields shadow the promoted `Alias` fields, so the decoder writes into `aux` and never into `next` directly. A round-trip assertion on a non-zero id would pin that invariant if anyone ever refactors it, plus a one-line comment on why the write-back is there. The rest are small test-strengthening notes inline.

Separately, a heads-up on CI: the Go matrix is red, but it's the pre-existing `intrange` lint in `catalog/rest/load_table_bench_test.go` (from #1654), not anything you touched. It'll need a rebase once main is green before this can land.

None of that is blocking on your end. Add the write-back guard and I'm happy to approve.

table/snapshots.go
```diff
@@ -272,9 +272,27 @@ type Snapshot struct {
...
+
+	if err := json.Unmarshal(data, &aux); err != nil {
 		return err
 	}
+	if aux.SnapshotID == nil {
+		return fmt.Errorf("%w: snapshot-id is absent or null", ErrInvalidMetadata)
```
zeroshade, line 289: `ParseMetadataBytes` wraps every `json.Unmarshal` failure with `ErrInvalidMetadata`, so wrapping the same sentinel here produces duplicated public text: `invalid metadata: invalid metadata: snapshot-id is absent or null`. Please avoid the outer wrap when the nested error already satisfies `errors.Is(err, ErrInvalidMetadata)` (and add an exact error-text assertion so this does not regress).

zeroshade (changes requested):
The snapshot required-field validation is correct, and the targeted regressions plus the full table package pass locally.

Before merge:

- Commit `4cc62fa3` needs the `Signed-off-by` trailer required by `CONTRIBUTING.md`.
- Please address the duplicate `ErrInvalidMetadata` wrapping noted inline.
- Please rebase onto current `main` to inherit the unrelated `intrange` lint fix and restart CI. This is a branch-state issue rather than a changed-line issue, so it is recorded here instead of on an unrelated inline anchor.

---

*This review was drafted by an AI-assisted tool and confirmed by an Apache Iceberg Go maintainer. After you've addressed the points above and pushed an update, an Apache Iceberg Go maintainer — a real person — will take the next look at the PR.*

table/snapshots.go
Revanth14: Fixed. `ParseMetadataBytes` now returns errors that already match `ErrInvalidMetadata` without wrapping them again. I also updated the tests to assert the exact error text, so a duplicate prefix will be caught. Thanks!

table/snapshots.go
Revanth14: Thanks, you’re right that `SnapshotLogEntry` and `MetadataLogEntry` have analogous required-field validation gaps. Since #1869 is specifically scoped to `Snapshot.snapshot-id` and `timestamp-ms`, I’d prefer to track those separately in a follow-up issue.

For manifest locations, v1 snapshots may use `manifests`, while newer snapshots normally use `manifest-list`. Although `Snapshot.UnmarshalJSON` does not know the table format version, it can still require at least one of those representations to be present. Format-version-specific enforcement would need to remain at the metadata level to preserve compatibility with legacy snapshots in upgraded tables.

table/snapshots.go
Revanth14: Thanks, that makes sense. I’ll decode into the alias first and only inspect the raw fields when either `snapshot-id` or `timestamp-ms` has its zero value.

The fallback will check both field presence and JSON `null`, so missing or null values remain invalid while explicit numeric zero remains accepted. This keeps the normal nonzero path on the existing direct decode.

zeroshade (approved):
LGTM thanks for addressing everything!

PR #1872, opened 2026-08-23
fix(expr): drop out-of-range set literals

### LINKED ISSUE #1870: Mixed-range IN predicates can fail during scan planning
## Apache Iceberg version

`main` at `0f08e66e59cc1ec9b40898823d9eb991409bd080` (also present in v0.6.0)

## Please describe the bug 🐞

Binding an `IN` or `NOT IN` predicate to a narrower numeric field retains `AboveMaxLiteral` and `BelowMinLiteral` sentinels in the bound set. Metrics evaluators assume every set member is a regular `TypedLiteral[T]`, so a valid scan can fail during planning when its filter mixes representable and out-of-range values.

For example, an application may build an `int64` filter for an Iceberg `int32` column:

```go
iceberg.IsIn(iceberg.Reference("id"), int64(35), int64(math.MaxInt32)+1)
```

The second value can never equal an `int32` column value and should be discarded during binding. Instead, it becomes `aboveMaxLiteral[int32]`. Inclusive metrics evaluation reaches `removeBoundCmp`, asserts that every member implements `TypedLiteral[int32]`, and returns this error:

```text
interface conversion: iceberg.aboveMaxLiteral[int32] is not iceberg.TypedLiteral[int32]: missing method Comparator
```

This affects the normal scan-planning path because `Scan.collectManifestEntriesWithSchema` creates the inclusive metrics evaluator for the row filter.

### Expected behavior

- Impossible set members are removed when binding to the field type.
- The mixed predicate above behaves like `id IN (35)`.
- A set containing only impossible members simplifies to `AlwaysFalse` for `IN` and `AlwaysTrue` for `NOT IN`.
- Metrics evaluation completes without an error.

### Reproduction

The following regression test can be added to `table/evaluators_test.go`:

```go
func TestInclusiveMetricsMixedRangeIn(t *testing.T) {
	lower, err := iceberg.Int32Literal(30).MarshalBinary()
	require.NoError(t, err)
	upper, err := iceberg.Int32Literal(79).MarshalBinary()
	require.NoError(t, err)

	schema := iceberg.NewSchema(0,
		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int32},
	)
	file := &mockDataFile{
		count:       1,
		valueCounts: map[int]int64{1: 1},
		nullCounts:  map[int]int64{1: 0},
		lowerBounds: map[int][]byte{1: lower},
		upperBounds: map[int][]byte{1: upper},
	}
	filter := iceberg.IsIn(
		iceberg.Reference("id"),
		int64(35),
		int64(math.MaxInt32)+1,
	)

	eval, err := newInclusiveMetricsEvaluator(schema, filter, true, true)
	require.NoError(t, err)
	shouldRead, err := eval(file)
	require.NoError(t, err)
	assert.True(t, shouldRead)
}
```

On current `main`, `eval(file)` returns the interface-conversion error above. An in-range control using `int64(35), int64(36)` succeeds.

### Root cause

`createBoundSetPredicate` calls `Literal.To(boundType)` and unconditionally inserts the result. Scalar literal binding already handles `AboveMaxLiteral` and `BelowMinLiteral` specially, but set binding does not.

Dropping those impossible members before the existing zero/one-member simplification should preserve `IN` and `NOT IN` semantics while preventing non-comparable sentinel values from reaching expression, manifest, metrics, and bloom-filter consumers.

mattfaltyn:
## Summary

- discard above-max and below-min sentinels while binding `IN` and `NOT IN` predicates
- reuse the existing empty- and singleton-set simplifications after impossible values are removed
- cover integer and floating-point narrowing plus the inclusive metrics scan-planning regression

## Why

Applications can construct a filter with a wider Go literal type than the Iceberg field type. When a set mixes representable and out-of-range values, binding currently retains range sentinels that cannot equal any field value. Metrics pruning later treats those sentinels as regular typed literals and returns an interface-conversion error instead of planning the scan.

Removing impossible set members during binding preserves membership semantics and keeps sentinels out of downstream predicate consumers.

Closes #1870.

## Testing

- `go test . -run '^TestBindAboveBelowIntMax$' -count=1`
- `go test ./table -run '^TestEvaluators/TestInMetrics$' -count=1 -v`
- `go test ./...`
- `make test-assert`
- `make test-race`
- `golangci-lint v2.12.2 run --timeout=10m --new-from-rev origin/main` (`0 issues`)

The full repository lint invocation reports two existing `intrange` findings in unchanged `catalog/rest/load_table_bench_test.go` lines 84 and 101. Current upstream `main` fails on the same findings ([run](https://github.com/apache/iceberg-go/actions/runs/32528308563)), so the PR's Go matrix currently stops in lint before its test steps; the diff-scoped lint run is clean.

### DIFF
--- a/exprs_test.go
+++ b/exprs_test.go
@@ -1006,9 +1006,12 @@ func TestBindAboveBelowIntMax(t *testing.T) {
 		}{
 			{"int in mixed range", iceberg.IsIn(ref, int64(34), above, below), iceberg.EqualTo(ref, int32(34))},
 			{"int not in mixed range", iceberg.NotIn(ref, int64(34), above, below), iceberg.NotEqualTo(ref, int32(34))},
+			{"int in mixed range set", iceberg.IsIn(ref, int64(34), int64(35), above, below), iceberg.IsIn(ref, int32(34), int32(35))},
 			{"int in outside range", iceberg.IsIn(ref, above, below), iceberg.AlwaysFalse{}},
 			{"int not in outside range", iceberg.NotIn(ref, above, below), iceberg.AlwaysTrue{}},
 			{"float in mixed range", iceberg.IsIn(ref2, float64(34), above2, below2), iceberg.EqualTo(ref2, float32(34))},
+			{"float not in mixed range", iceberg.NotIn(ref2, float64(34), above2, below2), iceberg.NotEqualTo(ref2, float32(34))},
+			{"float in outside range", iceberg.IsIn(ref2, above2, below2), iceberg.AlwaysFalse{}},
 			{"float not in outside range", iceberg.NotIn(ref2, above2, below2), iceberg.AlwaysTrue{}},
 		}
 
@@ -1016,9 +1019,9 @@ func TestBindAboveBelowIntMax(t *testing.T) {
 			t.Run(tt.name, func(t *testing.T) {
 				bound, err := iceberg.BindExpr(sc, tt.pred, true)
 				require.NoError(t, err)
-				expected, err := iceberg.BindExpr(sc, tt.expected, true)
+				wantBound, err := iceberg.BindExpr(sc, tt.expected, true)
 				require.NoError(t, err)
-				assert.True(t, expected.Equals(bound), "expected %s, got %s", expected, bound)
+				assert.True(t, wantBound.Equals(bound), "expected %s, got %s", wantBound, bound)
 			})
 		}
 	})
--- a/exprs_test.go
+++ b/exprs_test.go
@@ -997,6 +997,31 @@ func TestBindAboveBelowIntMax(t *testing.T) {
 			assert.Equal(t, tt.exp, b)
 		})
 	}
+
+	t.Run("set predicates", func(t *testing.T) {
+		tests := []struct {
+			name     string
+			pred     iceberg.BooleanExpression
+			expected iceberg.BooleanExpression
+		}{
+			{"int in mixed range", iceberg.IsIn(ref, int64(34), above, below), iceberg.EqualTo(ref, int32(34))},
+			{"int not in mixed range", iceberg.NotIn(ref, int64(34), above, below), iceberg.NotEqualTo(ref, int32(34))},
+			{"int in outside range", iceberg.IsIn(ref, above, below), iceberg.AlwaysFalse{}},
+			{"int not in outside range", iceberg.NotIn(ref, above, below), iceberg.AlwaysTrue{}},
+			{"float in mixed range", iceberg.IsIn(ref2, float64(34), above2, below2), iceberg.EqualTo(ref2, float32(34))},
+			{"float not in outside range", iceberg.NotIn(ref2, above2, below2), iceberg.AlwaysTrue{}},
+		}
+
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				bound, err := iceberg.BindExpr(sc, tt.pred, true)
+				require.NoError(t, err)
+				expected, err := iceberg.BindExpr(sc, tt.expected, true)
+				require.NoError(t, err)
+				assert.True(t, expected.Equals(bound), "expected %s, got %s", expected, bound)
+			})
+		}
+	})
 }
 
 func TestVariantBoundLiteralRejectionMessage(t *testing.T) {
--- a/table/evaluators_test.go
+++ b/table/evaluators_test.go
@@ -1929,6 +1929,7 @@ func (suite *InclusiveMetricsTestSuite) TestInMetrics() {
 		{iceberg.IsIn(ref, IntMinValue-1, IntMinValue), true, "should read: id equal to lower bound"},
 		{iceberg.IsIn(ref, IntMaxValue-4, IntMaxValue-3), true, "should read: id between upper and lower bounds"},
 		{iceberg.IsIn(ref, IntMaxValue, IntMaxValue+1), true, "should read: id equal to upper bound"},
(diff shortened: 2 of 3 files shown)


tanmayrauth (approved):
LGTM. Matches Java's bindInOperation, drops out-of-range sentinels, then the existing empty→alwaysFalse/alwaysTrue and single→EQ/NEQ collapse handles the rest, identically for IN and NOT IN. No blockers.

exprs_test.go
```diff
@@ -997,6 +997,31 @@ func TestBindAboveBelowIntMax(t *testing.T) {
...
+			name     string
+			pred     iceberg.BooleanExpression
+			expected iceberg.BooleanExpression
+		}{
+			{"int in mixed range", iceberg.IsIn(ref, int64(34), above, below), iceberg.EqualTo(ref, int32(34))},
+			{"int not in mixed range", iceberg.NotIn(ref, int64(34), above, below), iceberg.NotEqualTo(ref, int32(34))},
+			{"int in outside range", iceberg.IsIn(ref, above, below), iceberg.AlwaysFalse{}},
+			{"int not in outside range", iceberg.NotIn(ref, above, below), iceberg.AlwaysTrue{}},
+			{"float in mixed range", iceberg.IsIn(ref2, float64(34), above2, below2), iceberg.EqualTo(ref2, float32(34))},
+			{"float not in outside range", iceberg.NotIn(ref2, above2, below2), iceberg.AlwaysTrue{}},
```
laskoviymishka, line 1015: The int side has four cases but the float side only two, with no structural reason for the gap. The float path goes through `Float32AboveMaxLiteral` vs int's `Int32AboveMaxLiteral`, so the missing cases aren't redundant. I'd fill in the two symmetric ones:

```go
{"float in outside range", iceberg.IsIn(ref2, above2, below2), iceberg.AlwaysFalse{}},
{"float not in mixed range", iceberg.NotIn(ref2, float64(34), above2, below2), iceberg.NotEqualTo(ref2, float32(34))},
```

While we're here, every case in this table leaves 0 or 1 survivor, so the 2+ survivor branch never runs with sentinels present. One case like `IsIn(ref, int64(34), int64(35), above, below)` -> `IsIn(ref, int32(34), int32(35))` would exercise it.

exprs_test.go
```diff
@@ -997,6 +997,31 @@ func TestBindAboveBelowIntMax(t *testing.T) {
...
+		}
+
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				bound, err := iceberg.BindExpr(sc, tt.pred, true)
+				require.NoError(t, err)
+				expected, err := iceberg.BindExpr(sc, tt.expected, true)
```
laskoviymishka: `expected` here is the bound form while `tt.expected` is the unbound input, so the name's doing double duty. I'd rename the local to `wantBound` to pair with `bound` above and make the `Equals` read cleaner. Non-blocking.

table/evaluators_test.go
```diff
@@ -1929,6 +1929,7 @@ func (suite *InclusiveMetricsTestSuite) TestInMetrics() {
 		{iceberg.IsIn(ref, IntMinValue-1, IntMinValue), true, "should read: id equal to lower bound"},
 		{iceberg.IsIn(ref, IntMaxValue-4, IntMaxValue-3), true, "should read: id between upper and lower bounds"},
 		{iceberg.IsIn(ref, IntMaxValue, IntMaxValue+1), true, "should read: id equal to upper bound"},
+		{iceberg.IsIn(ref, int64(IntMaxValue), int64(math.MaxInt32)+1), true, "should read: ignore value outside int32 range"},
```
laskoviymishka, line 1932: This regression only exercises `IsIn`, but the original panic hit `NotIn` too on the same path (an `AboveMaxLiteral` reaching `cloneBoundLiteral`). The fix is in the shared `createBoundSetPredicate` so both are covered, but this is the metrics evaluator where it actually blew up, so I'd add the NOT IN sibling here to pin the end-to-end path:

```go
{iceberg.NotIn(ref, int64(IntMaxValue), int64(math.MaxInt32)+1), true, "should read: ignore value outside int32 range in not-in"},
```

laskoviymishka (approved):
Nice catch on the root cause. Dropping the out-of-range set members in `createBoundSetPredicate` is the right layer, and it matches Java's `bindInOperation` reduction to AlwaysFalse/EQ. Agreed with tanmayrauth's approval.

Nothing blocking from me, just some test-coverage polish. The main one: the metrics regression in `evaluators_test.go` only covers `IsIn`, but `NotIn` hit the same panic on the same path, so I'd add the NOT IN sibling there to lock the end-to-end path. The float cases in `exprs_test.go` are also a bit asymmetric with the int ones, worth filling in while we're here.

None of that blocks merge. Happy to approve once the NOT IN regression is in.

exprs_test.go
mattfaltyn: Thanks — added the two symmetric float cases and a multi-survivor IN case in 1b28a8d, so the filtered bound-set branch is covered too.

exprs_test.go
mattfaltyn: Thanks — renamed the bound expectation to wantBound in 1b28a8d.

table/evaluators_test.go
mattfaltyn: Thanks — added the NOT IN regression in 1b28a8d under strict metrics, where NOT IN inspects the set and reproduces the pre-fix sentinel failure. It keeps two in-range survivors so the fixed path remains a bound NOT IN set.

laskoviymishka:
I think this need rebase

ferhatelmas (approved):
This looks good but similar class issue exists in int64 -> date. Addressed in #1877 and builds on top of this. When this PR is merged, it should be green as well.

PR #1873, opened 2026-08-23
fix(updates): decode add-spec and add-sort-order unbound

twuebi:
Follow-up to #1665 

A commit request's add-spec and add-sort-order payloads are bound to a schema by Apply, not by the decoder: AddPartitionSpec calls BindToSchema and AddSortOrder calls CheckCompatibility, both against the table's current schema. Decoding them into the bound types therefore enforced an invariant the wire form does not carry.

That rejected a spec a client is entitled to send. A dropped partition field is a void transform over source ID 0, which BindToSchema preserves as a tombstone, so re-sending a spec after dropping a partition column failed with "partition source ID must be positive: 0" before binding ran.

Decode into UnboundPartitionSpec and UnboundSortOrder, and hand the embedded bound values to the builder. Source IDs that do not resolve are still rejected, now by binding, which reports them against the schema that decides. The constructors keep taking bound values, so callers are unaffected, and Apply reports a nil payload rather than dereferencing it.

### DIFF
--- a/table/updates.go
+++ b/table/updates.go
@@ -389,7 +389,11 @@ func (u *setCurrentSchemaUpdate) Apply(builder *MetadataBuilder) error {
 
 type addPartitionSpecUpdate struct {
 	baseUpdate
-	Spec    *iceberg.PartitionSpec `json:"spec"`
+	// The spec arrives unbound: Apply binds it to the table's current schema,
+	// so its source IDs are only required to resolve there. A dropped
+	// partition field in particular is a void transform over source ID 0,
+	// which BindToSchema carries across but a bound decode rejects.
+	Spec    *iceberg.UnboundPartitionSpec `json:"spec"`
 	initial bool
 }
 
@@ -397,15 +401,23 @@ type addPartitionSpecUpdate struct {
 // metadata. If the initial flag is set to true, the spec is considered the initial spec of the table,
 // and all other previously added specs in the metadata builder are removed.
 func NewAddPartitionSpecUpdate(spec *iceberg.PartitionSpec, initial bool) *addPartitionSpecUpdate {
-	return &addPartitionSpecUpdate{
+	upd := &addPartitionSpecUpdate{
 		baseUpdate: baseUpdate{ActionName: UpdateAddSpec},
-		Spec:       spec,
 		initial:    initial,
 	}
+	if spec != nil {
+		upd.Spec = &iceberg.UnboundPartitionSpec{PartitionSpec: *spec}
+	}
+
+	return upd
 }
 
 func (u *addPartitionSpecUpdate) Apply(builder *MetadataBuilder) error {
-	return builder.AddPartitionSpec(u.Spec, u.initial)
+	if u.Spec == nil {
+		return fmt.Errorf("%w: %s requires field %q", iceberg.ErrInvalidArgument, UpdateAddSpec, "spec")
+	}
+
+	return builder.AddPartitionSpec(&u.Spec.PartitionSpec, u.initial)
 }
 
 type setDefaultSpecUpdate struct {
@@ -428,7 +440,10 @@ func (u *setDefaultSpecUpdate) Apply(builder *MetadataBuilder) error {
 
 type addSortOrderUpdate struct {
 	baseUpdate
-	SortOrder *SortOrder `json:"sort-order"`
+	// Unbound for the same reason as addPartitionSpecUpdate.Spec: Apply checks
+	// the order against the table's current schema, which is what decides
+	// whether its source IDs resolve.
+	SortOrder *UnboundSortOrder `json:"sort-order"`
 	initial   bool
 }
 
@@ -436,14 +451,22 @@ type addSortOrderUpdate struct {
 // If the initial flag is set to true, the sort order is considered the initial sort order of the table,
 // and all previously added sort orders in the metadata builder are removed.
 func NewAddSortOrderUpdate(sortOrder *SortOrder) *addSortOrderUpdate {
-	return &addSortOrderUpdate{
+	upd := &addSortOrderUpdate{
 		baseUpdate: baseUpdate{ActionName: UpdateAddSortOrder},
-		SortOrder:  sortOrder,
 	}
+	if sortOrder != nil {
+		upd.SortOrder = &UnboundSortOrder{SortOrder: *sortOrder}
+	}
+
+	return upd
 }
 
 func (u *addSortOrderUpdate) Apply(builder *MetadataBuilder) error {
-	return builder.AddSortOrder(u.SortOrder)
+	if u.SortOrder == nil {
+		return fmt.Errorf("%w: %s requires field %q", iceberg.ErrInvalidArgument, UpdateAddSortOrder, "sort-order")
+	}
+
+	return builder.AddSortOrder(&u.SortOrder.SortOrder)
 }
 
 type setDefaultSortOrderUpdate struct {
--- a/table/updates.go
+++ b/table/updates.go
@@ -466,6 +466,8 @@ func (u *addSortOrderUpdate) Apply(builder *MetadataBuilder) error {
 		return fmt.Errorf("%w: update %q requires field %q", iceberg.ErrInvalidArgument, UpdateAddSortOrder, "sort-order")
 	}
 
+	// AddSortOrder writes the fresh order ID into the order it is given and
+	// keeps a pointer to that field, so the builder aliases this update.
 	return builder.AddSortOrder(&u.SortOrder.SortOrder)
 }
 
--- a/table/updates.go
+++ b/table/updates.go
@@ -414,7 +414,7 @@ func NewAddPartitionSpecUpdate(spec *iceberg.PartitionSpec, initial bool) *addPa
 
 func (u *addPartitionSpecUpdate) Apply(builder *MetadataBuilder) error {
 	if u.Spec == nil {
-		return fmt.Errorf("%w: %s requires field %q", iceberg.ErrInvalidArgument, UpdateAddSpec, "spec")
+		return fmt.Errorf("%w: update %q requires field %q", iceberg.ErrInvalidArgument, UpdateAddSpec, "spec")
 	}
 
 	return builder.AddPartitionSpec(&u.Spec.PartitionSpec, u.initial)
@@ -440,9 +440,9 @@ func (u *setDefaultSpecUpdate) Apply(builder *MetadataBuilder) error {
 
 type addSortOrderUpdate struct {
 	baseUpdate
-	// Unbound for the same reason as addPartitionSpecUpdate.Spec: Apply checks
-	// the order against the table's current schema, which is what decides
-	// whether its source IDs resolve.
+	// Unbound so decoding accepts placeholder source IDs. Unlike the spec
+	// above, none are exempt at Apply: CheckCompatibility still requires every
+	// source ID to resolve in the current schema, so this only defers the error.
 	SortOrder *UnboundSortOrder `json:"sort-order"`
 	initial   bool
 }
@@ -463,7 +463,7 @@ func NewAddSortOrderUpdate(sortOrder *SortOrder) *addSortOrderUpdate {
 
 func (u *addSortOrderUpdate) Apply(builder *MetadataBuilder) error {
 	if u.SortOrder == nil {
-		return fmt.Errorf("%w: %s requires field %q", iceberg.ErrInvalidArgument, UpdateAddSortOrder, "sort-order")
+		return fmt.Errorf("%w: update %q requires field %q", iceberg.ErrInvalidArgument, UpdateAddSortOrder, "sort-order")
 	}
 
 	return builder.AddSortOrder(&u.SortOrder.SortOrder)
--- a/table/updates_test.go
+++ b/table/updates_test.go
@@ -524,12 +524,12 @@ func TestUnmarshalUpdates(t *testing.T) {
 					case "add-partition-spec":
 						expectedAddPartitionSpec := u.(*addPartitionSpecUpdate)
 						actualAddPartitionSpec := actual[idx].(*addPartitionSpecUpdate)
-						assert.True(t, expectedAddPartitionSpec.Spec.Equals(*actualAddPartitionSpec.Spec))
+						assert.True(t, expectedAddPartitionSpec.Spec.Equals(actualAddPartitionSpec.Spec.PartitionSpec))
 						assert.Equal(t, actualAddPartitionSpec.initial, expectedAddPartitionSpec.initial)
 					case "add-sort-order":
 						expectedAddSortOrder := u.(*addSortOrderUpdate)
 						actualAddSortOrder := actual[idx].(*addSortOrderUpdate)
-						assert.True(t, expectedAddSortOrder.SortOrder.Equals(*actualAddSortOrder.SortOrder))
+						assert.True(t, expectedAddSortOrder.SortOrder.Equals(actualAddSortOrder.SortOrder.SortOrder))
 						assert.Equal(t, actualAddSortOrder.initial, expectedAddSortOrder.initial)
 					default:
 						assert.Equal(t, u, actual[idx])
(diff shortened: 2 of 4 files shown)


twuebi:
@laskoviymishka

catalog/rest/load_table_bench_test.go
```diff
@@ -98,7 +98,7 @@ func makeTableResponseWithSnapshots(snapshotCount int64) []byte {
 	schemaID := 0
 	var snapshotTimestamp int64
 	var snapshotID int64
-	for i := int64(0); i < snapshotCount; i++ {
+	for i := range snapshotCount {
```
laskoviymishka, line 101: This `for range` change is unrelated to the decode fix, so I'd pull it into its own commit to keep the bug fix easy to bisect.

It's also not purely cosmetic: in the old loop `i` was shared across iterations, so `parentID = &i` captured one address and every snapshot ended up pointing at the final count. Range-over-int gives each iteration its own `i`, which quietly corrects that. Fine to keep, but worth calling out as a behavior change rather than folding in as style.

table/updates.go
```diff
@@ -389,23 +389,35 @@ func (u *setCurrentSchemaUpdate) Apply(builder *MetadataBuilder) error {
...
+		upd.Spec = &iceberg.UnboundPartitionSpec{PartitionSpec: *spec}
+	}
+
+	return upd
 }
 
 func (u *addPartitionSpecUpdate) Apply(builder *MetadataBuilder) error {
-	return builder.AddPartitionSpec(u.Spec, u.initial)
+	if u.Spec == nil {
+		return fmt.Errorf("%w: %s requires field %q", iceberg.ErrInvalidArgument, UpdateAddSpec, "spec")
```
laskoviymishka: These nil guards format the action name bare, but every other "requires field" error in this file uses the `validateRequiredUpdateFields` template `"%w: update %q requires field %q"`. So this path emits `add-spec requires field "spec"` while the decode path emits `update "add-spec" requires field "spec"`. Same divergence at line 466 for sort-order. I'd match the existing template so the two paths read identically.

table/updates.go
```diff
@@ -428,22 +440,33 @@ func (u *setDefaultSpecUpdate) Apply(builder *MetadataBuilder) error {
 
 type addSortOrderUpdate struct {
 	baseUpdate
-	SortOrder *SortOrder `json:"sort-order"`
+	// Unbound for the same reason as addPartitionSpecUpdate.Spec: Apply checks
```
laskoviymishka: The symmetry with the partition-spec fix only holds at decode time. For specs the fix is end-to-end, but `SortOrder.CheckCompatibility` still calls `validateSortSourceIDs` unconditionally and rejects a void sort tombstone with "source ID must be positive: 0" at Apply.

So a void sort tombstone now decodes fine but still fails during Apply, which isn't obvious from a comment that implies parity. I'd either carry the void exception through `CheckCompatibility` too, or soften the comment to say we're only deferring the error here, not accepting the tombstone. wdyt?

table/updates_test.go
```diff
@@ -524,12 +524,12 @@ func TestUnmarshalUpdates(t *testing.T) {
 					case "add-partition-spec":
 						expectedAddPartitionSpec := u.(*addPartitionSpecUpdate)
 						actualAddPartitionSpec := actual[idx].(*addPartitionSpecUpdate)
-						assert.True(t, expectedAddPartitionSpec.Spec.Equals(*actualAddPartitionSpec.Spec))
+						assert.True(t, expectedAddPartitionSpec.Spec.Equals(actualAddPartitionSpec.Spec.PartitionSpec))
```
laskoviymishka, line 527: This is the assertion the PR just updated, but it never runs. `Action()` returns `add-spec` (`UpdateAddSpec`), and the case label above is `"add-partition-spec"`, so the whole branch falls through to `default` and the spec is never compared.

Even if the label matched, `expectedAddPartitionSpec` and `actualAddPartitionSpec` are both cast from `actual[idx]` (`u` *is* `actual[idx]` in this range loop), so it compares the decoded spec against itself and passes vacuously. The `add-sort-order` branch below has the same self-comparison shape.

I'd switch to `case UpdateAddSpec:` and compare `tc.expected[idx]` against `actual[idx]` so the round-trip is genuinely exercised. This is the one thing I'd want fixed before merge, since it's the coverage this change is really adding.

table/updates_test.go
```diff
@@ -1299,3 +1299,62 @@ func TestRemoveEncryptionKeyUpdate_Apply_NoOp(t *testing.T) {
 	b := buildFromBase(t)
 	require.NoError(t, NewRemoveEncryptionKeyUpdate("nonexistent").Apply(b))
 }
+
+func TestAddPartitionSpecUpdate_UnmarshalVoidTombstone(t *testing.T) {
```
laskoviymishka, line 1303: Not this PR's job to solve, but worth flagging while we're here: this now accepts and persists a partition field with source-id 0, and `PartitionField.MarshalJSON` omits `source-id` when it's 0. Java's `PartitionSpecParser` calls `JsonUtil.getInt("source-id", ...)` unconditionally, so metadata written with such a tombstone won't parse back in Java or PyIceberg.

The create-table path already had this exposure, so this PR only widens the set of inputs that reach it. I wouldn't block on it, but a tracking issue for canonicalizing the void tombstone (Java preserves the original positive source-id rather than storing 0) would be good so we don't lose it.

table/updates_test.go
```diff
@@ -1299,3 +1299,62 @@ func TestRemoveEncryptionKeyUpdate_Apply_NoOp(t *testing.T) {
...
+	// rejected, but by binding rather than by decoding.
+	data := []byte(`[{"action":"add-spec","spec":{"spec-id":1,"fields":[{"source-id":0,"field-id":1000,"transform":"identity","name":"x"}]}}]`)
+
+	var updates Updates
+	require.NoError(t, json.Unmarshal(data, &updates))
+	require.Len(t, updates, 1)
+
+	err := updates[0].Apply(buildFromBaseV3(t))
+	require.Error(t, err)
+	assert.NotContains(t, err.Error(), "must be positive")
```
laskoviymishka, line 1339: This pins the absence of the old decode error but nothing about the new one, so a future regression that fails for a different reason would slip through. The sort-order test right below asserts the positive message ("not compatible with current schema"), so I'd add the analogous `assert.Contains` on the expected binding error here too.

laskoviymishka (changes requested):
This is the right shape for the fix: decoding into the unbound variants and deferring the real binding to Apply, while keeping the constructors bound so existing callers don't change, lines up with how Java stores `AddPartitionSpec`/`AddSortOrder`. Nice and tight.

One blocking thing before merge: the `TestUnmarshalUpdates` assertion this PR updates never actually runs. `Action()` returns `add-spec`, but the switch case is labeled `"add-partition-spec"`, so it falls through to `default`. And even where a case does match, both sides are cast from `actual[idx]`, so the comparison is against itself and passes vacuously. That's exactly the round-trip coverage this change is adding, so I'd want it exercised: switch to `case UpdateAddSpec:` and compare `tc.expected[idx]` against `actual[idx]`.

Everything else is minor: the nil-guard error strings drift from the `validateRequiredUpdateFields` template, the sort-order "unbound for the same reason" comment overstates the symmetry (void sort tombstones still fail at Apply, not just decode), and the `for range` change in the bench test is unrelated to the fix, so it's worth its own commit, especially since it quietly corrects the `parentID = &i` capture.

Fix the dead test so the new assertion runs and this is good to land.

table/updates.go
```diff
@@ -389,23 +389,35 @@ func (u *setCurrentSchemaUpdate) Apply(builder *MetadataBuilder) error {
 
 type addPartitionSpecUpdate struct {
 	baseUpdate
-	Spec    *iceberg.PartitionSpec `json:"spec"`
+	// The spec arrives unbound: Apply binds it to the table's current schema,
+	// so its source IDs are only required to resolve there. A dropped
+	// partition field in particular is a void transform over source ID 0,
+	// which BindToSchema carries across but a bound decode rejects.
+	Spec    *iceberg.UnboundPartitionSpec `json:"spec"`
```
laskoviymishka, line 396: The behavior here is right, and it matches how Java stores this (`AddPartitionSpec` holds an `UnboundPartitionSpec` too), so I'm not worried about correctness.

What gives me pause is that `UnboundPartitionSpec`'s own doc says to use `PartitionSpec` for specs read from table metadata, where source IDs are bound and must be positive — and an add-spec commit payload is exactly that. So we're using the type against its documented contract, and the next person who reads that doc hits a trap: the bound positivity check that would catch a genuinely bad source ID is now bypassed at decode time.

Two ways out, and either is fine as long as we're explicit: keep `*UnboundPartitionSpec` here and loosen that doc comment to acknowledge the commit path, or add a void-tombstone exemption to the bound decoder (parallel to `BindToSchema`) and keep `*iceberg.PartitionSpec` — the new tests would drive the same behavior either way. I lean slightly toward the second since it keeps the type contract honest, but I don't feel strongly. wdyt?

table/updates.go
```diff
@@ -428,22 +440,33 @@ func (u *setDefaultSpecUpdate) Apply(builder *MetadataBuilder) error {
...
+	return upd
 }
 
 func (u *addSortOrderUpdate) Apply(builder *MetadataBuilder) error {
-	return builder.AddSortOrder(u.SortOrder)
+	if u.SortOrder == nil {
+		return fmt.Errorf("%w: update %q requires field %q", iceberg.ErrInvalidArgument, UpdateAddSortOrder, "sort-order")
+	}
+
+	return builder.AddSortOrder(&u.SortOrder.SortOrder)
```
laskoviymishka, line 471: Minor lifetime thing: `AddSortOrder` stashes `&sortOrder.orderID` in `b.lastAddedSortOrderID` and holds it for the builder's lifetime, so after this call the builder is pointing into `u.SortOrder.SortOrder` — a field of this update struct. It's fine as long as the `Updates` slice outlives the builder, which is the normal pattern, but nothing enforces or documents it.

A one-line comment on this call noting the builder aliases the embedded order would save the next person the trace. Not blocking.

table/updates_test.go
```diff
@@ -1299,3 +1299,63 @@ func TestRemoveEncryptionKeyUpdate_Apply_NoOp(t *testing.T) {
...
+	// rejected, but by binding rather than by decoding.
+	data := []byte(`[{"action":"add-spec","spec":{"spec-id":1,"fields":[{"source-id":0,"field-id":1000,"transform":"identity","name":"x"}]}}]`)
+
+	var updates Updates
+	require.NoError(t, json.Unmarshal(data, &updates))
+	require.Len(t, updates, 1)
+
+	err := updates[0].Apply(buildFromBaseV3(t))
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "cannot find source column with id: 0 in schema")
```
laskoviymishka: The `NotContains(..., "must be positive")` below is the load-bearing assertion here — it's what proves the rejection moved from decode-time to binding — but the `Contains` on `"cannot find source column with id: 0 in schema"` pins us to a phrase from deep inside `BindToSchema`. If that wording ever changes, the `require.Error` above still passes and the failure shows up as a bare `false`.

I'd switch it to `require.ErrorContains` so a regression stops with a useful message, and drop a one-line comment that we're intentionally pinning the error source to binding. Small thing.

table/updates_test.go
```diff
@@ -1299,3 +1299,63 @@ func TestRemoveEncryptionKeyUpdate_Apply_NoOp(t *testing.T) {
...
+	// known at Apply, which is where an unresolvable source ID is reported.
+	data := []byte(`[{"action":"add-sort-order","sort-order":{"order-id":1,"fields":[{"source-id":0,"transform":"identity","direction":"asc","null-order":"nulls-first"}]}}]`)
+
+	var updates Updates
+	require.NoError(t, json.Unmarshal(data, &updates))
+	require.Len(t, updates, 1)
+
+	err := updates[0].Apply(buildFromBaseV3(t))
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "not compatible with current schema")
```
laskoviymishka: This one only ever exercises the failure path — decode succeeds, Apply rejects. We've got no test that a resolvable add-sort-order actually round-trips: decodes as `*UnboundSortOrder`, applies, builds, and comes out with the right direction/null-order/transform.

That's the half that actually changed here — the decode now goes through `newSortOrder(..., false)`, which skips source-ID validation — and right now a parsing regression in it would only be caught by `TestUnmarshalUpdates`, which never reaches Apply+Build.

Mirror the void-tombstone test with a source ID that resolves in `baseMetaJSON` (field 1, `x`), apply it, build, and assert the field comes back correct. Then both sides of the new construction are covered.

laskoviymishka (approved):
Thanks for the quick turnaround — the dead `TestUnmarshalUpdates` assertion is properly live now. The switch runs on `UpdateAddSpec`/`UpdateAddSortOrder` and compares `tc.expected[idx]` against `actual[idx]` instead of against itself, so the round-trip coverage this PR adds actually runs (and it surfaced the `NullsLast` fix on field 2 along the way). That was the blocker, and it's resolved.

The one thing I'd like your read on is the type choice. Decoding into `UnboundPartitionSpec` is behaviorally right and matches how Java stores `AddPartitionSpec`, so I'm not worried about correctness — but it runs against that type's own doc comment ("use `PartitionSpec` for specs read from table metadata"), which leaves a trap for the next reader. I left an inline with the two ways I'd resolve it; I don't feel strongly which.

Past that it's small stuff, all inline: the sort-order test only covers the reject path, so the happy-path round-trip for the new `UnboundSortOrder` decode is still untested; one error-string assertion is pinned to `BindToSchema`'s internal wording; and the sort-order Apply now hands the builder a pointer into the update struct.

None of this blocks merge for me. I'd genuinely like your read on the type-contract question, but however you land it, I think this is good to go.

table/updates_test.go
```diff
@@ -1299,3 +1299,94 @@ func TestRemoveEncryptionKeyUpdate_Apply_NoOp(t *testing.T) {
...
+	data := []byte(`[{"action":"add-spec","spec":{"spec-id":1,"fields":[{"source-id":0,"field-id":1000,"transform":"void","name":"x_bucket"}]}}]`)
+
+	var updates Updates
+	require.NoError(t, json.Unmarshal(data, &updates))
+	require.Len(t, updates, 1)
+
+	b := buildFromBaseV3(t)
+	require.NoError(t, updates[0].Apply(b))
+
+	meta, err := b.Build()
```
laskoviymishka, line 1316: This test is the first place a void tombstone rides all the way through decode, bind, and Build, which is exactly the path we want covered, but it stops at Build and never marshals `meta` back out, and that's where I think there's a real interop hazard.

`PartitionField.MarshalJSON` special-cases void tombstones and omits `source-id`, so the spec we persist here serializes with no `source-id` key at all. Java's `PartitionSpecParser.fromJson` and PyIceberg both require `source-id`, so once iceberg-go writes this file neither can read it back. Before this PR the void path was rejected at decode, so that marshal branch never actually ran, which is why it's only surfacing now.

I think the fix is to drop the void special-case in `MarshalJSON`, since the normal path already writes `source-id: 0` and matches Java, and a `json.Marshal` + `assert.JSONEq` round-trip on this fixture would both pin the behavior and have caught this. If you'd rather do the marshal fix as a follow-up with an issue that's fine, but I'd want it decided before merge. wdyt?

table/updates_test.go
```diff
@@ -1299,3 +1299,94 @@ func TestRemoveEncryptionKeyUpdate_Apply_NoOp(t *testing.T) {
...
+	// Pinned to BindToSchema's wording on purpose: the assertion below only
+	// shows the decoder let the field through, not who rejected it.
+	require.ErrorContains(t, err, "cannot find source column with id: 0 in schema")
+	assert.NotContains(t, err.Error(), "must be positive")
+}
+
+func TestAddSortOrderUpdate_UnmarshalDefersBindingToApply(t *testing.T) {
+	// Decoding accepts the order; the schema it must resolve against is only
+	// known at Apply, which is where an unresolvable source ID is reported.
+	data := []byte(`[{"action":"add-sort-order","sort-order":{"order-id":1,"fields":[{"source-id":0,"transform":"identity","direction":"asc","null-order":"nulls-first"}]}}]`)
```
laskoviymishka: source-id 0 here gets rejected by `validateSortSourceIDs` ("source ID must be positive: 0") before any schema lookup runs, so this proves positive-ID enforcement moved to Apply rather than the schema-resolution deferral the name and comment describe. The `FindFieldByID` branch never executes.

If we use a positive source-id that isn't in the V3 base schema (say 999), Apply fails with "sort field with source id 999 not found in schema" and the test actually exercises the deferred schema lookup. wdyt?

table/updates_test.go
```diff
@@ -1299,3 +1299,94 @@ func TestRemoveEncryptionKeyUpdate_Apply_NoOp(t *testing.T) {
...
+	require.NoError(t, updates[0].Apply(b))
+
+	meta, err := b.Build()
+	require.NoError(t, err)
+
+	var order SortOrder
+	var found bool
+	for _, o := range meta.SortOrders() {
+		if o.OrderID() == 1 {
+			order, found = o, true
```
laskoviymishka, line 1391: Small one: this keeps scanning after it finds order 1, so if a future fixture ever produced two matching orders the last would silently win. A `break` after the assignment makes the intent clear.

table/updates_test.go
```diff
@@ -1299,3 +1299,94 @@ func TestRemoveEncryptionKeyUpdate_Apply_NoOp(t *testing.T) {
...
+	require.Equal(t, 1, order.Len())
+
+	field := order.Field(0)
+	assert.Equal(t, []int{1}, field.SourceIDs)
+	assert.Equal(t, SortDESC, field.Direction)
+	assert.Equal(t, NullsLast, field.NullOrder)
+	assert.IsType(t, iceberg.IdentityTransform{}, field.Transform)
+}
+
+func TestAddSpecAndSortOrderUpdates_ApplyRejectNilPayload(t *testing.T) {
```
laskoviymishka, line 1406: These two nil-payload checks run sequentially with `require`, so if the spec assertion fails the sort-order one never runs. I'd split them into two `t.Run` subtests so both always execute and you can see which one broke.

laskoviymishka (changes requested):
Thanks, the type-contract question I asked you to weigh in on last round is properly resolved. The `UnboundPartitionSpec`/`UnboundSortOrder` doc comments now describe both wire forms (the create-table request and the add-spec/add-sort-order commit payload) instead of pointing the next reader away from exactly the use this PR adds. That was the one open thing I cared about, so thanks for taking it head on.

Everything else I flagged is in too:
- `TestAddSortOrderUpdate_UnmarshalRoundTrip` covers the resolvable `UnboundSortOrder` decode end to end
- the `addSortOrderUpdate` comment now says plainly that none are exempt at Apply, which fixes the round-1 symmetry overstatement
- nil-payload guards on both Applys, with a test
- the `BindToSchema`-pinned error assertion is a deliberate pin now and the comment says so, which is a fine answer to the round-2 minor

One genuinely new thing surfaced this round, and it's why I'm not just re-approving. This PR is the first change that lets a void tombstone (a dropped partition field, void over source-id 0) decode, bind, and persist end to end. Before this it was rejected at decode, so the marshal side of that path never ran. `PartitionField.MarshalJSON` special-cases void tombstones and omits `source-id`, so the metadata we write for such a spec comes out with no `source-id` key at all, and Java's `PartitionSpecParser.fromJson` and PyIceberg both require it and fail to read the file back. I left an inline: I think dropping the special-case branch is the fix, since the normal path already writes `source-id: 0` and matches Java, but if you'd rather scope it to a follow-up with an issue I'm open to that. I'd just want it decided before this lands rather than shipping the path that persists a metadata file Java can't read.

A few tiny test things are inline as well (the defers-to-Apply sort test proves something narrower than its name, a missing `break`, and splitting the two nil-payload asserts so both always run). None of those block.

Sort out the void-tombstone marshal one way or the other and I'm happy to approve.

twuebi:
@laskoviymishka I don't think `source-id: 0` is correct either, you're right that a spec serialized with no `source-id` key won't read back in Java or PyIceberg. `PartitionSpecParser.fromJson` reads it with `JsonUtil.getInt(SOURCE_ID, element)`, which throws on a missing key. 

But dropping the special-case so we write `source-id: 0` doesn't match Java either, maybe it's even worse:

- **Java never writes `0` for a void field.** `toJson` writes `field.sourceId()` unconditionally, with no void branch, and `PartitionField` holds the original source id, a voided field serializes with its original positive source id. This is what you noted yourself in the `updates_test.go:1303` thread ("Java preserves the original positive source-id rather than storing 0"). Rust does the same, void fields carry a positive `source_id`. So `source-id: 0` isn't Java's output. It's our go-sentinel value leaking into the file.
- **`0` isn't a valid schema field id.** With `source-id: 0` present, Java/PyIceberg get past the `getInt` read but then reference a column that doesn't exist, so it fails at bind instead of at read. That's not "readable" in any useful sense.
- **Our own bound decoder rejects it.** `validatePartitionSourceID` returns `partition source ID must be positive: 0` for a bound spec, and `PartitionField.UnmarshalJSON` decodes as bound. So the file we'd write couldn't be read back by iceberg-go either.

The root of it: a source-less void tombstone has already discarded the original source id (that's what the `[]int{0}` sentinel is), so there's no positive id to recover at marshal time. `MarshalJSON` is the wrong layer to fix this, the current omit-in/omit-out round-trip at least preserves the historical file exactly, whereas `source-id: 0` turns it into something no implementation, ours included, can bind.

For real Java parity, the fix belongs upstream of serialization, preserve the original positive source id through decode/bind rather than collapsing it to `0`, or reject a source-less void field in the add-spec commit path. I'm happy to work on a follow-up PR to fix this, wdyt?

laskoviymishka (approved):
LGTM

PR #1877, opened 2026-08-25
fix(literals): reject out-of-range int64-date conversion

ferhatelmas:
`Int64Literal.To(DateType)` narrowed with a plain `DateLiteral(i)` conversion, silently wrapping out-of-range values so `Eq`/`In` predicates could match rows the user never asked for. 

Return `Int32AboveMaxLiteral()`/`Int32BelowMinLiteral()` instead, similar to int64 -> int32 narrowing. Binding already handles the sentinels: unary predicates simplify to `AlwaysFalse`/`AlwaysTrue`, and set predicates drop the impossible members (#1870).

related to #1870 and #1872

### DIFF
--- a/exprs_test.go
+++ b/exprs_test.go
@@ -1027,6 +1027,38 @@ func TestBindAboveBelowIntMax(t *testing.T) {
 	})
 }
 
+func TestBindOutOfRangeDate(t *testing.T) {
+	sc := iceberg.NewSchema(1,
+		iceberg.NestedField{ID: 1, Name: "d", Type: iceberg.PrimitiveTypes.Date},
+	)
+
+	ref := iceberg.Reference("d")
+	above, below := int64(math.MaxInt32)+1, int64(math.MinInt32)-1
+
+	tests := []struct {
+		name     string
+		pred     iceberg.BooleanExpression
+		expected iceberg.BooleanExpression
+	}{
+		{"eq above max", iceberg.EqualTo(ref, above), iceberg.AlwaysFalse{}},
+		{"eq below min", iceberg.EqualTo(ref, below), iceberg.AlwaysFalse{}},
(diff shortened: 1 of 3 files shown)


exprs_test.go
```diff
@@ -999,6 +999,38 @@ func TestBindAboveBelowIntMax(t *testing.T) {
...
+
+	tests := []struct {
+		name     string
+		pred     iceberg.BooleanExpression
+		expected iceberg.BooleanExpression
+	}{
+		{"eq above max", iceberg.EqualTo(ref, above), iceberg.AlwaysFalse{}},
+		{"eq below min", iceberg.EqualTo(ref, below), iceberg.AlwaysFalse{}},
+		{"neq above max", iceberg.NotEqualTo(ref, above), iceberg.AlwaysTrue{}},
+		{"in mixed range", iceberg.IsIn(ref, int64(34), above, below), iceberg.EqualTo(ref, iceberg.Date(34))},
```
zeroshade, line 1046: This case (along with the two set cases below it) currently fails because the branch does not contain PR #1872's sentinel filtering for `IN`/`NOT IN`, and #1872 is still open rather than present on `main`. The sentinels remain in the bound set instead of simplifying to the expected predicate. Please rebase after #1872 merges, stack this PR on it, or include the required set-binding change so the PR is independently green.

zeroshade (changes requested):
The int64-to-date range check itself looks correct, but this branch is not independently green. Its `IN`/`NOT IN` regression cases require the sentinel-filtering change from PR #1872, which remains open. On current `main`, those three cases fail and cause the root suite and multiple CI jobs to fail; temporarily combining #1872 makes the complete regression pass.

Please merge and rebase onto #1872, stack this PR on #1872, or otherwise include the required set-binding change.

---

zeroshade (approved):

PR #1881, opened 2026-08-25
chore: align makefile & ci for race test

ferhatelmas:
+ use makefile in ci
+ enable race detector for all unit tests in `test-race` make target

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -89,10 +89,10 @@ jobs:
         # collide on one entry; we manage the cache explicitly above instead.
         skip-cache: true
     - name: Run tests
-      run: go test -v ./...
+      run: make test
     # Race detector is opt-in per package/test.
     - name: Run race detector
-      run: go test -race -v ./codec/... ./table/...
+      run: make test-race
     - name: Run Arrow ownership assertions
       if: matrix.os == 'ubuntu-latest' && matrix.go == '1.26.1'
       run: make test-assert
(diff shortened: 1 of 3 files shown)


.github/workflows/go-ci.yml
```diff
@@ -89,10 +89,10 @@ jobs:
         # collide on one entry; we manage the cache explicitly above instead.
         skip-cache: true
     - name: Run tests
-      run: go test -v ./...
+      run: make test
     # Race detector is opt-in per package/test.
     - name: Run race detector
-      run: go test -race -v ./codec/... ./table/...
+      run: make test-race
```
ferhatelmas: Actually, I would suggest running race for all packages (filtering according to OS).

badalprasadsingh:
1. Can we add `make test-race` in contributing.md? That would help future contributors.
2. Plus, since you would be touching that area for 1, can we update the go commands to `make test` and `make lint`.

LGTM, otherwise.

.github/workflows/go-ci.yml
ferhatelmas: if speed is a concern, we can parallelize it. 

To me, missing a package where it should have run is worse. Speed is clear, improved if painful

zeroshade: Agreed, we should run race for all packages

zeroshade (changes requested):
Just needs to update to use test -race for all packages and not a subset of them, then we're good

.github/workflows/go-ci.yml
ferhatelmas: Updated, dropped regular test since race one covers it now

zeroshade (approved):

PR #1885, opened 2026-08-26
perf(compaction): reuse survivor partition tuple

fallintoplace:
## Summary

- Reuse the canonical partition tuple in `AddSurvivorWithSpec`.
- Avoid sorting and formatting the same partition twice.
- Keep empty-partition and negative-sequence behavior unchanged.

## Benchmark

The benchmark is included in this PR at `table/compaction/eq_delete_decision_bench_test.go`.

Command: `go test ./table/compaction -run=^$ -bench=^BenchmarkAddSurvivorWithSpec$ -benchmem -count=5`

Measured on an Apple M1 Pro. Parent commit -> this PR:

- 1 field: 72 -> 32 B/op, 7 -> 4 allocs/op
- 8 fields: 624 -> 456 B/op, 28 -> 16 allocs/op
- 32 fields: 2544 -> 1816 B/op, 80 -> 42 allocs/op

The 32-field case was about 40% faster in the local comparison.

## Checks

- `go test ./... -count=1`
- `go test -race ./table/compaction -count=1`
- `go vet ./table/compaction`

### DIFF
--- a/table/compaction/eq_delete_decision.go
+++ b/table/compaction/eq_delete_decision.go
@@ -78,28 +78,30 @@ func (s *SurvivorSurvey) AddSurvivor(partition map[int]any, seq int64) {
 		return
 	}
 
-	key := partitionMatchKey(partition)
-	if cur, ok := s.PartMinSeq[key]; ok {
-		seq = min(seq, cur)
-	}
-	s.PartMinSeq[key] = seq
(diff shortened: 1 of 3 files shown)


tanmayrauth (approved):
LGTM!

zeroshade (approved):
LGTM

PR #1886, opened 2026-08-26
docs: expand security threat model

zeroshade:
## Summary

- adapt the expanded Iceberg Python threat-model structure to Iceberg Go's actual catalog, REST, configuration, IO, and CLI boundaries
- preserve Iceberg Go-owned credential isolation across separately constructed catalog/client instances while distinguishing caller-shared state
- clarify private vulnerability reporting, delegated credentials, endpoint routing, resource limits, table provenance, and destructive external-path behavior
- make scanner calibration conditional and require human review instead of blanket rejection

## Review notes

The Go model intentionally does not copy Python's blanket same-process isolation disclaimer or assume universal table-root containment. Availability-only findings remain hardening by default unless they violate a documented Iceberg Go-owned boundary or demonstrate another security impact.

## Verification

- `go test ./...` — 28 packages passed; 6 packages had no tests
- `git diff --check upstream/main..HEAD`
- final task review and whole-branch security-policy review completed with no remaining Critical or Important findings

### DIFF
--- a/SECURITY-THREAT-MODEL.md
+++ b/SECURITY-THREAT-MODEL.md
@@ -59,14 +59,29 @@ rules. A finding that crosses an Iceberg Go-owned boundary or demonstrates a
 security impact must be reviewed even when it also has a normally trusted
 precondition.
 
-In this document, a **new audience** is any log, error, CLI text or JSON output,
-serialized metadata, host, catalog or client, or principal that was not already
-authorized to receive the secret, credential, or credential-bearing request.
+## Definitions
+
+A **new audience** is any log, error, CLI text or JSON output, serialized
+metadata, host, catalog or client, or principal that was not already authorized
+to receive the secret, credential, or credential-bearing request.
 
 A **documented Iceberg Go-owned availability boundary** is an explicit
-project-owned resource limit that Iceberg Go documents and enforces, such as
-the Puffin reader's configured maximum blob size (256 MB by default). It is not
+project-owned resource limit that Iceberg Go documents and enforces. It is not
 a general availability guarantee for an embedding service.
+
+Current examples include:
+
+- the Puffin reader's configured
+  [`puffin.WithMaxBlobSize`](puffin/puffin_reader.go) limit (256 MB by default);
+  and
+- the REST metrics reporter's documented per-report deadline, configured with
+  [`rest-metrics-reporting-timeout-ms`](website/src/configuration.md#metrics-reporting)
+  (10 seconds by default).
+
+This list is non-exhaustive. An integer-width check, slice bound, worker count,
+batch size, or other internal implementation constraint is not a documented
+availability boundary unless a public API or project document presents it as a
+resource limit and Iceberg Go enforces it.
 
 ## Scope
 
@@ -93,8 +108,9 @@ Iceberg Go should:
 - prevent tokens, client secrets, storage credentials, signed requests, and
   credential-bearing configuration from reaching a new audience;
 - preserve per-catalog and per-client authentication and credential isolation
-  for state Iceberg Go creates, including internally managed auth, transport,
-  metrics, and delegated-credential state;
+  for state Iceberg Go creates, including internally managed authentication and
+  transport state, REST metrics reporters and dispatchers, and
+  delegated-credential state;
 - avoid creating network, signing, storage, or destructive capabilities that
   the configured principal did not authorize;
 - preserve documented Iceberg Go-owned availability boundaries by enforcing
@@ -140,7 +156,8 @@ capability.
 
 The embedding application decides which users may invoke Iceberg Go and owns
 its user and tenant boundaries. It may intentionally share an `AuthManager`,
-custom transport, database handle or implementation, registry implementation,
+custom transport, `metrics.Reporter`, a reporter factory registered with
+`metrics.Register`, database handle or implementation, registry implementation,
 or other mutable dependency.
 Such deliberate sharing is caller-owned; Iceberg Go-created state that crosses
 otherwise separate catalog or client instances is not.
@@ -219,24 +236,25 @@ deletes or mutates objects outside the actor's proven storage capability.
 
 ### Boundary 5: Iceberg Go-owned client state
 
-Iceberg Go internally creates REST sessions and per-catalog authentication,
-transport, metrics, and delegated-credential state. Separately constructed
-catalogs and clients must not receive one another's internally managed state.
+Iceberg Go internally creates REST sessions, per-catalog authentication and
+transport state, REST metrics reporters and dispatchers, and
+delegated-credential state. Separately constructed catalogs and clients must
+not receive one another's internally managed state.
 
-Callers may intentionally share an `AuthManager`, custom transport, database
-handle or implementation, registry implementation, or other mutable dependency.
-Effects inherent in an explicitly shared object are caller-owned by default.
-This exception does not
-disclaim credential crossover or state reuse caused by Iceberg Go-owned
-per-catalog or per-client state.
+Callers may intentionally share an `AuthManager`, custom transport,
+`metrics.Reporter`, a reporter factory registered with `metrics.Register`,
+database handle or implementation, registry implementation, or other mutable
+dependency. Effects inherent in an explicitly shared object are caller-owned by
+default. This exception does not disclaim credential crossover or state reuse
+caused by Iceberg Go-owned per-catalog or per-client state.
 
 ### Boundary 6: configuration discovery and process-global registries
 
 Iceberg Go loads operator-controlled configuration from
 `~/.iceberg-go.yaml`, from the directory selected by `GOICEBERG_HOME`, or from
 an explicit CLI `--config` path. The library also initializes process-global
-configuration state, and its catalog and IO registries are process-global
-plugin mechanisms that callers can extend.
+configuration state, and its catalog, IO, and metrics reporter-factory
+registries are process-global plugin mechanisms that callers can extend.
 
 Those documented configuration sources and registries are not tenant-isolation
 boundaries. Control of them normally demonstrates operator or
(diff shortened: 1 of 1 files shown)


SECURITY-THREAT-MODEL.md
```diff
@@ -19,205 +19,461 @@
...
+  destructive impact.
 
-## Scope
+The categories below are conditional triage defaults, not blanket rejection
+rules. A finding that crosses an Iceberg Go-owned boundary or demonstrates a
+security impact must be reviewed even when it also has a normally trusted
+precondition.
 
-This model is scoped to the Apache Iceberg Go repository itself:
+In this document, a **new audience** is any log, error, CLI text or JSON output,
```
laskoviymishka: These two terms, 'new audience' here and 'documented Iceberg Go-owned availability boundary' just below, end up governing almost every Boundary, In-Scope, and Out-of-Scope entry, but they're defined inside the Purpose prose. Someone landing directly on Boundary 3 or Out-of-Scope 7 (which the doc itself says is a real access pattern for maintainers and scanners) hits them with no definition in view. Not blocking, but I think a short Definitions section right before Trust Boundaries would make the governing terms much easier to find. wdyt?

SECURITY-THREAT-MODEL.md
```diff
@@ -19,205 +19,461 @@
...
+serialized metadata, host, catalog or client, or principal that was not already
+authorized to receive the secret, credential, or credential-bearing request.
 
-- table format and metadata handling
-- catalog and REST catalog clients
-- transport, credential, and configuration handling implemented in this repo
-- command-line or helper tooling shipped in this repo
+A **documented Iceberg Go-owned availability boundary** is an explicit
```
laskoviymishka, line 68: Every Out-of-Scope override leans on this term ('reassess if a finding violates a documented Iceberg Go-owned availability boundary'), but the only example given is the Puffin 256 MB limit. A maintainer triaging, say, an Avro decode cap or a manifest slice limit can't tell whether it's a documented and owned boundary without auditing the code first. Not a blocker, but a short non-exhaustive list of the enforced limits, or even just a pointer like 'see the puffin and table package docs', would let the rule be applied without a codebase dig.

SECURITY-THREAT-MODEL.md
```diff
@@ -19,205 +19,461 @@
...
 
-Iceberg Go often accepts metadata locations, table properties, namespace
-properties, and related control-plane information from a catalog. By default,
-Iceberg Go treats those sources as trusted.
+### Boundary 5: Iceberg Go-owned client state
 
-This means a malicious catalog supplying incorrect or malicious metadata is
-usually not an Iceberg Go vulnerability by itself.
+Iceberg Go internally creates REST sessions and per-catalog authentication,
+transport, metrics, and delegated-credential state. Separately constructed
```
laskoviymishka: Small inconsistency worth resolving: 'metrics' is listed here (and in Security Goals) as Iceberg Go-owned state that separately constructed catalogs must not share, but Out-of-Scope 6 says sharing a registry implementation is caller-owned and not a security issue. An app that deliberately shares a `metrics.Registry` across two catalogs lands in both buckets at once, so a maintainer could reach opposite conclusions depending on which section they read first. Metrics also aren't usually credential-bearing, so either dropping 'metrics' from the isolation set here, or a sentence on why cross-catalog metric aggregation is security-relevant, would settle it. Non-blocking.

laskoviymishka (commented):
I left a few non-blocking comments on the parts I'd want us to be deliberate about before this merges.

SECURITY-THREAT-MODEL.md
zeroshade: Good catch—the bare term “metrics” obscures two different ownership cases. The intended in-scope state is a catalog-created REST reporter/dispatcher, which uses that catalog’s authenticated client and therefore must remain isolated. A caller deliberately sharing a `metrics.Reporter`, or registering a factory that returns shared state, is caller-owned. I’ll replace the broad wording with those concrete terms in Security Goals, Boundary 5, In-Scope 2, and Out-of-Scope 6.

SECURITY-THREAT-MODEL.md
zeroshade: Good point. I’ll add a short non-exhaustive list containing the Puffin blob limit and the documented REST metrics-report deadline. I’ll also state that incidental integer, slice, batching, or worker limits do not qualify unless a public API or project document establishes them as an enforced resource boundary. That should make the rule usable without requiring a codebase dig while avoiding a stale inventory of internal constants.

SECURITY-THREAT-MODEL.md
zeroshade: Agreed. These terms carry enough weight that they should be independently linkable rather than embedded in the Purpose prose. I’ll move them into a short Definitions section before the substantive scope and boundary rules.

SECURITY-THREAT-MODEL.md
```diff
@@ -19,205 +19,479 @@
...
+an explicit CLI `--config` path. The library also initializes process-global
+configuration state, and its catalog, IO, and metrics reporter-factory
+registries are process-global plugin mechanisms that callers can extend.
+
+Those documented configuration sources and registries are not tenant-isolation
+boundaries. Control of them normally demonstrates operator or
+embedding-application authority and lowers scanner confidence. Reassess when
+Iceberg Go sends a secret to a new audience, crosses internally managed
+catalog/client state, creates an unauthorized integrity capability, exposes or
+corrupts memory, or causes an unauthorized destructive effect.
```
laskoviymishka, line 264: Small consistency thing while we're in here: Boundary 1 lists "bypasses an operator-selected restriction" among the conditions that defeat the downgrade, and this reassess list covers overlapping config ground (`~/.iceberg-go.yaml`, `GOICEBERG_HOME`) but leaves it out.

A config-load bug that silently ignores an operator's TLS-min or allowed-endpoint restriction wouldn't trip this checklist, though Boundary 1's would. I'd add it here to match Boundary 1. Not blocking.

SECURITY-THREAT-MODEL.md
```diff
@@ -19,205 +19,479 @@
...
+
+A report that requires the selected catalog, OAuth service, storage service, or
+other operator-trusted control plane to be malicious normally has a trusted
+precondition and lower confidence. Trust in that service is not proof that all
+resulting client behavior is authorized.
+
+A new credential audience, unexpected outbound request, cross-catalog or
+cross-client state reuse, bypass of an operator-selected restriction,
+unauthorized integrity or memory impact, or unauthorized destructive effect
+overrides this default downgrade.
```
laskoviymishka, line 405: This override list drops the availability-boundary trigger that the out-of-scope section header says applies to every default: "a demonstrated violation of a documented Iceberg Go-owned availability boundary." Every other out-of-scope category carries it, and the in-scope preamble explicitly puts demonstrated availability-boundary violations in scope.

The gap is load-bearing here specifically: a malicious control plane that pushes Iceberg Go past a documented limit (say the Puffin blob-size cap) reads as lower-confidence under this category, but reviewable under in-scope. Someone reading only this section reaches the opposite call.

I'd add the trigger to both this list and the equivalent-harm one just below (category 5 has the same omission, though it's more contrived there). Once those two match the header, happy to approve.

SECURITY-THREAT-MODEL.md
```diff
@@ -19,205 +19,479 @@
...
+  boundary crossing.
+
+Scanners should report the actor, controlled input, affected boundary, existing
+and new capability, intended and actual credential audience, and demonstrated
+impact. They should not classify the mere presence of a configurable endpoint,
+`unsafe` conversion, global registry, malformed input, or external file
+reference as a vulnerability.
+
+These signals route findings for human review; they do not automatically accept
+or reject a report. Endpoint routing, credential scope, shared state, `unsafe`
```
laskoviymishka, line 495: "shared state" here is unqualified, but section 6 and Boundary 5 carefully split Iceberg Go-owned state crossing separately-constructed clients (in scope) from a caller deliberately sharing an `AuthManager`, transport, or `metrics.Reporter` (caller-owned by default).

A scanner reading only this closing line would flag every shared-state observation for review, including the caller-owned cases the rest of the doc downgrades. I'd tighten it to something like "Iceberg Go-owned shared state crossing separately constructed catalog or client instances" to preserve that carve-out. Also non-blocking. wdyt?

laskoviymishka (changes requested):
This reads a lot cleaner than prev version. The Definitions section now pins down "new audience" and "documented Iceberg Go-owned availability boundary" up front, the availability-boundary examples (the Puffin blob cap and the REST metrics timeout) are concrete and correct, and the metrics-ownership split reads consistently now that internally-managed reporters and dispatchers are separated from a caller-supplied `metrics.Reporter`. That last one was my main worry and I think it's fully resolved.

One thing I'd want us to close before merge. The out-of-scope section header says every default reassesses on a demonstrated availability-boundary violation, and the in-scope preamble puts those squarely in scope, but categories 4 (malicious catalog) and 5 (equivalent-harm) drop that trigger from their own reassess lists. Every other category carries it. The effect is a real false-negative: a malicious control plane pushing Iceberg Go past the documented Puffin limit reads as lower-confidence under category 4 while in-scope says it's reviewable. Adding the trigger to both lists closes it.

Two smaller consistency nits I left inline (Boundary 6 missing the "bypasses an operator-selected restriction" condition that Boundary 1 has, and "shared state" in the Scanner Calibration closer being unqualified against the section-6 carve-out) are both non-blocking.

Fix the category 4/5 trigger and I'm happy to approve.

SECURITY-THREAT-MODEL.md
zeroshade: Fixed in 11334ef4.

Both category 4 and category 5 now carry "a demonstrated violation of a documented Iceberg Go-owned availability boundary" in their override lists, so all eight out-of-scope categories match the section header and the in-scope preamble.

The case you called out — a malicious control plane pushing Iceberg Go past the documented Puffin blob-size cap — now reads the same way whether someone lands on category 4 or on the in-scope section, which was the false negative.

SECURITY-THREAT-MODEL.md
zeroshade: Done in 11334ef4.

Boundary 6's reassess list now includes "bypasses an operator-selected restriction", matching Boundary 1. A config-load bug that silently drops an operator's TLS-min or allowed-endpoint restriction now trips both checklists rather than only Boundary 1's.

SECURITY-THREAT-MODEL.md
zeroshade: Done in 11334ef4.

Changed to "Iceberg Go-owned shared state crossing separately constructed catalog or client instances". That keeps the Boundary 5 and Out-of-Scope 6 carve-out intact, so a scanner reading only the closing line no longer flags deliberately shared caller-owned objects (an `AuthManager`, custom transport, or `metrics.Reporter`) for review.

PR #1891, opened 2026-08-26
feat(table): add explicit default spec/sort-order commit fence (Transaction.AssertDefaultShape)

ryanworl:
## What

Adds `Transaction.AssertDefaultShape`: an explicit fence pinning the table's default partition spec id and default sort order id at their base values — the layout analogue of `Transaction.AssertRefSnapshotID` (#1786), which pins a ref. It applies the existing `AssertDefaultSpecID` + `AssertDefaultSortOrderID` requirements from the base table metadata, and deduplicates against the producer-built assertions `UpdateSpec` and `ReplaceSortOrder` already register (same JSON, same pinned ids).

## Why

Java's [`UpdateRequirements.forUpdateTable`](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/UpdateRequirements.java) registers `AssertDefaultSpecID` / `AssertDefaultSortOrderID` when a commit *changes* those defaults, but a commit that stages no spec or sort-order change of its own (e.g. properties-only) carries no shape assertion, so two racers evolving a table's declaration can interleave. Java has no user-callable equivalent; this is an explicit API for a fence Java only applies implicitly.

Register the fence before staging spec or sort-order changes: requirements are validated against staged metadata at registration time. The fence is never rewritten by refresh-and-replay. A transaction with no updates never contacts the catalog, so the fence alone does not force a commit.

Naming: "shape" is not spec terminology — happy to rename (e.g. `AssertDefaultSpecAndSortOrder`) if preferred.

## Tests

- Properties-only commit succeeds when the shape is unmoved; fails with `ErrCommitFailed` (properties not applied) when the default spec or default sort order moved in between.
- Double-register plus `ReplaceSortOrder` collapses to one spec assertion and one sort-order assertion.

### DIFF
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -453,6 +453,38 @@ func (t *Transaction) AssertRefSnapshotID(branch string) error {
 	return nil
 }
 
+// AssertDefaultShape registers requirements pinning the table's default
+// partition spec ID and default sort order ID at their values in the
+// transaction's base metadata, so a concurrent change to either fails
+// this commit instead of interleaving silently — the shape analogue of
+// AssertRefSnapshotID's explicit branch fence. It gives commits that
+// stage no shape change of their own (for example a properties-only
+// revision marker) the same serialization against concurrent spec or
+// sort-order evolutions that Java UpdateRequirements.forUpdateTable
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM

PR #1894, opened 2026-08-26
perf(table): borrow partition records during scan planning

fallintoplace:
## Summary

- **Borrow** partition data during scan planning through a read-only `StructLike` view.
- Avoid rebuilding a positional `[]any` record for every candidate data file.
- Avoid cloning binary partition values on the internal hot path.
- Keep public `GetPartitionRecord` behavior defensive and unchanged.
- No public API changes.

## Why

Scan planning reuses a partition predicate for every candidate data file. Before this change, each evaluation called `GetPartitionRecord`, which allocated a positional record, walked every partition field, and copied binary values.

The new internal view looks up partition values by field ID in partition-field order and is used only for the current evaluator call. Its `Set` method is read-only, so the borrowed data cannot be changed through the view.

## Performance

Focused `BenchmarkPartitionEvaluator` result for 4,096 files on an Apple M1 Pro:

- 1 partition field: about 0.51 ms/op and 295 KB/op -> 0.43 ms/op and 197 KB/op
- 8 partition fields: about 0.76 ms/op and 754 KB/op -> 0.42 ms/op and 197 KB/op
- 32 partition fields: about 2.84 ms/op and 2.33 MB/op -> 0.52 ms/op and 197 KB/op

Each case also drops from 12,288 to 8,192 allocations per benchmark operation.

## Tests

- `go test ./...`
- `go test -race ./table`
- `go vet ./table`
- `go test -tags=assert -run='^(TestBuildPartitionEvaluator|TestBorrowedPartitionRecord|TestGetPartitionRecord)' ./table`
- `git diff --check`

### DIFF
--- a/data_file_refs.go
+++ b/data_file_refs.go
@@ -31,8 +31,7 @@ import "github.com/apache/iceberg-go/internal"
 // evaluators. ColumnSizes and DistinctValueCounts are excluded because the
 // evaluator does not read them, and cloning those maps would add work to the
 // hot path without changing its result. Partition has a separate borrowed
-// accessor because partition records are built once per file during scan
-// planning.
+// accessor because partition data is read during scan planning.
 func (d *dataFile) DataFileStatsRef(_ internal.DataFileRef) (
 	valueCounts map[int]int64,
 	nullCounts map[int]int64,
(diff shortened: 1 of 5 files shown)


laskoviymishka (approved):
👍

PR #1896, opened 2026-08-26
perf(table): borrow DataFile stats in metadata tables

fallintoplace:
## Summary

- **Borrow** partition data while appending files metadata rows.
- **Reuse** one borrowed stats view for value counts, null counts, NaN counts, and bounds.
- Keep the public `DataFile` getters and fallback behavior unchanged.
- No public API changes.

## Why

The files metadata tables append each `DataFile` into Arrow. The public getters return defensive copies for mutable metadata, but those copies were immediately iterated and discarded.

Built-in manifest `DataFile` values now use the existing read-only internal helpers. Other `DataFile` implementations still use their public getters.

## Performance

`BenchmarkInspectContentFileAppenderDataFileStats` on an Apple M1 Pro with 4,096 files and 32 stat columns:

- **Borrowed:** about 42.5 to 43.9 ms/op, 49.9 MB/op, and 54.2k allocs/op
- **Public copies:** about 58.1 to 59.3 ms/op, 87.8 MB/op, and 406.5k allocs/op

The benchmark also covers 0, 8, 32, and 128 stat columns with 4,096 and 16,384 files.

## Tests

- `go test ./...`
- `go test -race ./table`
- `go vet ./table`
- `go test -tags=assert -run='^TestInspectContentFileBuilder(UsesBorrowedDataFileMetadata|FallsBackToPublicDataFileMetadata)$' ./table`
- `git diff --check`

### DIFF
--- a/table/inspect_files.go
+++ b/table/inspect_files.go
@@ -655,7 +655,7 @@ func (b inspectContentFileBuilder) append(file iceberg.DataFile) error {
 	b.specID.Append(file.SpecID())
 
 	if b.partition != nil {
-		if err := b.partition.append(file.Partition()); err != nil {
+		if err := b.partition.append(dataFilePartition(file)); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


laskoviymishka (approved):
:shipit:

Issue #1884, opened 2026-08-26
Support delete-file-backed row changes in incremental changelog scans


fallintoplace:
## Background

Incremental changelog scanning was added in [#1883](https://github.com/apache/iceberg-go/pull/1883) to mirror Java's `newIncrementalChangelogScan()` API.

The first implementation handles file-level changes from data manifests. This issue tracks delete-file-backed row changes and keeps the related work together.

Current status: file-level changelog planning is implemented in #1883. Snapshots with delete manifests are still rejected.

## Implemented

- [x] `Table.NewIncrementalChangelogScan`
- [x] `INSERT` tasks for added data files
- [x] `DELETE` tasks for removed data files
- [x] Snapshot boundaries, change ordinals, and commit snapshot IDs
- [x] Filters, residuals, row-lineage task fields, and local planning

## Planned

- [ ] Define changelog task and reader types for added rows, deleted data files, and deleted rows
- [ ] Plan position and equality delete files across snapshot ranges
- [ ] Preserve added versus existing delete-file semantics
- [ ] Support deletion vectors and v3 row lineage
- [ ] Add end-to-end tests for same-snapshot changes, prior deletes, file removal, schema evolution, filters, and boundaries
- [ ] Add documentation and decide remote planning support

## Implementation notes

- The normal scan path removes deleted rows while reading. A changelog scan needs to emit rows removed by the current snapshot.
- A deleted data file must be read with delete files that existed before the file was removed.
- A row-level delete needs to distinguish delete files added by the current snapshot from delete files that were already present.
- The task shape should follow Java's `AddedRowsScanTask`, `DeletedDataFileScanTask`, and `DeletedRowsScanTask`, with Go naming.
- Position-delete, equality-delete, and deletion-vector readers already exist in the table scan and inspection paths. This work should reuse them where possible.
- Remote planning can remain unsupported initially.

## Related work

- [PR #1883](https://github.com/apache/iceberg-go/pull/1883)
- [Table.Inspect metadata table support #1703](https://github.com/apache/iceberg-go/issues/1703) (related groundwork only)
- [Delete-file metadata support #1699](https://github.com/apache/iceberg-go/pull/1699)
- [Position deletes metadata support #1747](https://github.com/apache/iceberg-go/pull/1747)

## References

- [Java changelog planner](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/BaseIncrementalChangelogScan.java)
[...]

PR #1897, opened 2026-08-27
feat(table): add delete-file-backed changelog task types

dgvj-work:
## Summary

First slice for delete-file-backed incremental changelog scans.

This adds the Java-shaped task types that later planning and readers need:

- `AddedRowsScanTask` for inserts from added data files, including same-snapshot deletes
- `DeletedDataFileScanTask` for removed data files, with existing deletes applied first
- `DeletedRowsScanTask` for row-level deletes, keeping added vs existing delete files separate

Planning, readers, deletion vectors, and end-to-end scans are left for follow-up. File-level changelog scanning remains in #1883.

Related to #1884

## Test plan

- [x] `go test ./table -count=1 -run 'TestAddedRowsScanTask|TestDeletedDataFileScanTask|TestDeletedRowsScanTask'`


table/changelog_scan_task.go
```diff
@@ -0,0 +1,152 @@
...
+
+package table
+
+import "github.com/apache/iceberg-go"
+
+// AddedRowsScanTask is a changelog insert produced by adding a data file.
+// Matching delete files committed in the same snapshot, or from squashed
+// snapshots, are applied while reading so deleted rows are not emitted as
+// inserts.
+type AddedRowsScanTask struct {
```
laskoviymishka, line 68: Design-direction question for this first slice: Java has these three implement a common `ChangelogScanTask` interface with `operation()`, `changeOrdinal()`, `commitSnapshotId()`. Here each type carries `ChangeOrdinal`/`CommitSnapshotID` but there's no shared interface and no `Operation()`.

Without it the planning follow-up can't return a uniform `[]ChangelogScanTask` and every consumer needs a type switch to tell inserts from deletes. I'd lean toward defining the interface plus a `ChangelogOperation` enum now so the follow-ups have something to build against, but if you'd rather defer until planning lands that's reasonable too. wdyt?

table/changelog_scan_task.go
```diff
@@ -0,0 +1,152 @@
...
+	return allDeleteFiles(t.FileScanTask)
+}
+
+// DeletedRowsScanTask is a changelog delete produced by adding delete files
+// against a data file that remains in the table. AddedDeletes remove rows
+// that should appear in the changelog. ExistingDeletes already applied and
+// those rows must not be emitted again.
+type DeletedRowsScanTask struct {
+	FileScanTask
+	addedDeletes     FileScanTask
```
laskoviymishka: I'd hold off on modeling `addedDeletes` as a `FileScanTask`. Java keeps it as a plain `List<DeleteFile>`, and here we get a second `.File` pointing at the same data file that no method ever reads, plus a whole `FileScanTask` whose `Start`/`Length`/`FirstRowID`/`DataSequenceNumber` are all zero/nil.

That last part is the real trap: once the reader slice lands and passes this into the read path, the zeroed range and nil `FirstRowID`/`DataSequenceNumber` will read as "intentionally absent" and silently suppress row-lineage synthesis (arrow_scanner gates on those being non-nil), so the output looks valid but is wrong.

I'd store the classified lists directly, a small internal struct like `classifiedDeletes{pos, eq, dv []iceberg.DataFile}`, and have `AddedDeletes()` flatten it. That makes the incomplete-metadata state impossible to express. wdyt?

table/changelog_scan_task.go
```diff
@@ -0,0 +1,152 @@
...
+		File:                dataFile,
+		DeleteFiles:         pos,
+		EqualityDeleteFiles: eq,
+		DeletionVectorFiles: dv,
+	}
+}
+
+func classifyDeleteFiles(files []iceberg.DataFile) (pos, eq, dv []iceberg.DataFile) {
+	for _, f := range files {
+		if f == nil {
```
laskoviymishka: Minor, but this nil guard is a little misleading: `f` is an interface, so a typed-nil (a `(*dataFile)(nil)` appended to the slice) passes `f == nil` and then panics on `IsDeletionVector`'s `FileFormat()` call right below. Real callers get `DataFile`s from `Build()` or manifest entries, neither of which produces a typed-nil, so I'd just drop the guard and note that nil elements aren't a supported input rather than half-guarding against a case that can't happen.

table/changelog_scan_task.go
```diff
@@ -0,0 +1,152 @@
...
+			continue
+		}
+		switch {
```
laskoviymishka: This switch has no default, so any file that isn't a DV, eq-delete, or pos-delete (a plain data file, `ContentType` 0) is silently dropped. `scanner.go` classifies into these same three buckets but returns a wrapped `ErrInvalidMetadata` on unknown content (scanner.go:776-791). I'd match that here rather than swallow it, since once the planner is feeding this, a misrouted data file would produce a silently-wrong changelog instead of a loud failure.

Given `scanner.go` already does this exact three-way split with the error branch, it's worth extracting one shared helper so the two don't drift. Returning an error means threading it through `fileScanTaskWithDeletes` and the constructors, but I think that's the right trade.

table/changelog_scan_task.go
```diff
@@ -0,0 +1,152 @@
...
+		case IsDeletionVector(f):
+			dv = append(dv, f)
+		case f.ContentType() == iceberg.EntryContentEqDeletes:
+			eq = append(eq, f)
+		case f.ContentType() == iceberg.EntryContentPosDeletes:
+			pos = append(pos, f)
+		}
+	}
+	return pos, eq, dv
```
laskoviymishka: `nlreturn` wants a blank line before this return, and the same before `return out` in `allDeleteFiles` just below. CI will fail on both until they're added. Quick fix, but it's the thing currently keeping the build red.

table/changelog_scan_task_test.go
```diff
@@ -0,0 +1,80 @@
...
+	t.Helper()
+
+	b, err := iceberg.NewDataFileBuilder(*iceberg.UnpartitionedSpec,
+		content, path, format, nil, nil, nil, 10, 1024)
+	require.NoError(t, err)
+
+	return b.Build()
+}
+
+func TestAddedRowsScanTaskAppliesSameSnapshotDeletes(t *testing.T) {
```
laskoviymishka, line 37: These three tests only exercise the happy path: every input is a cleanly-typed delete file. There's no case for the branch that matters most, a plain data file (or nil) in the deletes slice, which today is silently dropped. If that path becomes an error (per the `classifyDeleteFiles` comment), I'd want a test asserting the error; if it stays a skip, a test that documents the intent.

table/changelog_scan_task_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	task := NewDeletedRowsScanTask(data, []iceberg.DataFile{added}, []iceberg.DataFile{existing}, 2, 44)
+
+	require.Equal(t, 2, task.ChangeOrdinal())
+	require.Equal(t, int64(44), task.CommitSnapshotID())
+	require.Equal(t, []iceberg.DataFile{added}, task.AddedDeletes())
+	require.Equal(t, []iceberg.DataFile{existing}, task.ExistingDeletes())
+	require.Equal(t, existing.FilePath(), task.DeleteFiles[0].FilePath())
+	require.Empty(t, task.EqualityDeleteFiles)
+	require.Equal(t, []iceberg.DataFile{added}, task.addedDeletes.EqualityDeleteFiles)
```
laskoviymishka: The assertion a few lines up already checks `AddedDeletes()` returns `[]{added}`; this one reaches into the unexported `addedDeletes.EqualityDeleteFiles` to assert the same fact through internal layout. I'd drop it. If the goal is to prove eq-vs-pos classification, a dedicated `classifyDeleteFiles` unit test reads better and doesn't couple to the struct shape.

laskoviymishka (changes requested):
Nice start: the three-task taxonomy mirrors Java's changelog scan tasks cleanly, and landing the types before planning/readers is the right way to slice this. The doc comments carry a lot of the intent, which I appreciated.

I'd hold this before merging though. CI won't go green as-is (two `nlreturn` violations), and there's one modeling call I'd really like to settle before it calcifies into the follow-ups.

The main one is `addedDeletes` being a full `FileScanTask`. Java keeps it as a plain list of delete files, and here it buys us a second `.File` pointing at the same data file that nothing reads, plus a `FileScanTask` whose `Start`/`Length`/`FirstRowID`/`DataSequenceNumber` are all zero/nil. That's the part that worries me: when the reader slice lands and consumes this, the zeroed range and nil lineage fields will read as "intentionally absent" and silently suppress row-lineage synthesis, so the changelog output looks fine but is wrong. A small internal struct holding the classified lists makes that failure impossible to express, and it's much cheaper to change now than after readers depend on the shape.

- the two `nlreturn` fixes (CI)
- `classifyDeleteFiles` silently dropping anything that isn't a DV/eq/pos file; I'd return an error like `scanner.go` does rather than swallow it
- the `addedDeletes` shape above
- whether we define the `ChangelogScanTask` interface + `Operation()` now (Java has it) or defer to the planning PR (happy either way, just want it to be a decision)

The rest (the typed-nil guard, the redundant test assertion, thin classify coverage) are inline and minor. None of the deferred scope (planning, readers, DVs, e2e) is a concern for this PR; the taxonomy is the right foundation. Fix the CI and the `addedDeletes` shape and I'm happy to take another pass.

table/changelog_scan_task.go
dgvj-work: Good call — I added `ChangelogScanTask` with `Operation()`, `ChangeOrdinal()`, and `CommitSnapshotID()`, plus a `ChangelogOperation` enum matching Java (`INSERT` / `DELETE` / `UPDATE_BEFORE` / `UPDATE_AFTER`). The three task types implement it so the planning follow-up can return `[]ChangelogScanTask` without a type switch just to tell inserts from deletes.

table/changelog_scan_task.go
dgvj-work: Agreed, thanks for catching that. `addedDeletes` is now a small internal `classifiedDeletes` struct holding the pos/eq/dv lists, and `AddedDeletes()` just flattens it. That way we never carry a second `FileScanTask` whose range and lineage fields would read as intentionally absent.

table/changelog_scan_task.go
dgvj-work: Dropped the nil guard. Real callers get files from `Build()` or manifest entries, so a half-working interface-nil check wasn't worth keeping.

table/changelog_scan_task.go
dgvj-work: Makes sense. I extracted a shared `classifyDataFile` helper and wired it through both `manifestEntries.merge` and the changelog constructors. Unknown content — including a plain data file in the deletes slice — now returns `ErrInvalidMetadata` instead of being dropped.

table/changelog_scan_task.go
dgvj-work: Fixed — the returns now have the blank line `nlreturn` wants. Thanks for flagging it.

table/changelog_scan_task_test.go
dgvj-work: Added `TestClassifyDeleteFiles` for the pos/eq/dv split and for a data file in the deletes slice, which now errors with `ErrInvalidMetadata`.

table/changelog_scan_task_test.go
dgvj-work: Removed it. `AddedDeletes()` already covers the public result, and the classify test now owns the eq-vs-pos split without reaching into the struct layout.

laskoviymishka (approved):

dgvj-work:
@laskoviymishka conflicts with main are resolved. I don't have merge rights on the repo — could you merge this when CI is green?

laskoviymishka:
@dgvj-work all good! i'll merge this

PR #1898, opened 2026-08-27
perf(table): stream projected position-delete reads

fallintoplace:
## Summary

- Read position deletes with `GetRecords` using only `file_path` and `pos`.
- Process record batches incrementally with one grouping accumulator.
- Avoid full-table materialization and `UnifyTableDicts`.
- Keep the existing string, dictionary, validation, and ordering behavior.

## Performance

Local benchmark on an Apple M1 Pro with Go 1.26.3. Each case used a 100 ms benchmark run and included an unused string column in the Parquet file.

| Rows | Before | After |
| ---: | ---: | ---: |
| 100K | 11.1 ms/op, 77.7 MB/op, 1,009 allocs/op | 4.1 ms/op, 12.3 MB/op, 612 allocs/op |
| 1M | 93.8 ms/op, 613.0 MB/op, 4,851 allocs/op | 26.6 ms/op, 75.6 MB/op, 1,960 allocs/op |

The exact numbers will vary by machine and storage backend, but the projected reader avoids decoding and retaining unused columns.

## Tests

- `go vet ./table`
- `go test ./table -run ^$ -bench ^BenchmarkReadDeletesProjected/(rows=100000|rows=1000000)/(before|after) -benchtime=100ms -count=1`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -43,8 +43,10 @@ import (
 )
 
 const (
-	ScanOptionArrowUseLargeTypes              = "arrow.use_large_types"
-	ScanOptionRowLineageEnabled               = "row_lineage.enabled"
+	ScanOptionArrowUseLargeTypes = "arrow.use_large_types"
+	ScanOptionRowLineageEnabled  = "row_lineage.enabled"
+	// This must stay a power of two because appendFilePathChunk uses a bit mask
+	// for periodic cancellation checks.
 	positionalDeleteCancellationCheckInterval = 16 * 1024
 )
 
@@ -381,6 +383,10 @@ func (a *posDeleteAccumulator) release() {
 }
 
 func (a *posDeleteAccumulator) finish() map[string]*arrow.Chunked {
+	if a.builders == nil {
+		panic("position delete accumulator is already finished or released")
+	}
+
 	results := make(map[string]*arrow.Chunked, len(a.builders))
 	for path, builder := range a.builders {
 		positions := builder.NewInt64Array()
@@ -394,8 +400,8 @@ func (a *posDeleteAccumulator) finish() map[string]*arrow.Chunked {
 	return results
 }
 
-func validatePosDeleteColumns(filePathType arrow.DataType, filePathNulls int, filePathLen int,
-	posType arrow.DataType, posNulls int, posLen int,
+func validatePosDeleteColumns(filePathType arrow.DataType, filePathNulls int,
+	posType arrow.DataType, posNulls int,
 ) error {
 	if filePathNulls > 0 {
 		return fmt.Errorf("%w: null file_path in position delete file", iceberg.ErrInvalidSchema)
@@ -411,6 +417,11 @@ func validatePosDeleteColumns(filePathType arrow.DataType, filePathNulls int, fi
 		return fmt.Errorf("%w: unsupported pos column type %s in position delete file",
 			iceberg.ErrInvalidSchema, posType)
 	}
+
+	return nil
+}
+
+func validatePosDeleteColumnLengths(filePathLen, posLen int) error {
 	if filePathLen != posLen {
 		return fmt.Errorf("%w: file_path and pos columns have different lengths: %d and %d",
 			iceberg.ErrInvalidSchema, filePathLen, posLen)
@@ -476,8 +487,11 @@ func (a *posDeleteAccumulator) appendChunked(ctx context.Context, filePathCol, p
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-	if err := validatePosDeleteColumns(filePathCol.DataType(), filePathCol.NullN(), filePathCol.Len(),
-		posCol.DataType(), posCol.NullN(), posCol.Len()); err != nil {
+	if err := validatePosDeleteColumns(filePathCol.DataType(), filePathCol.NullN(),
+		posCol.DataType(), posCol.NullN()); err != nil {
+		return err
+	}
+	if err := validatePosDeleteColumnLengths(filePathCol.Len(), posCol.Len()); err != nil {
 		return err
 	}
 
@@ -503,16 +517,12 @@ func (a *posDeleteAccumulator) appendRecord(ctx context.Context, record arrow.Re
 
 	filePathCol := record.Column(0)
 	posCol := record.Column(1)
-	if err := validatePosDeleteColumns(filePathCol.DataType(), filePathCol.NullN(), filePathCol.Len(),
-		posCol.DataType(), posCol.NullN(), posCol.Len()); err != nil {
+	if err := validatePosDeleteColumns(filePathCol.DataType(), filePathCol.NullN(),
+		posCol.DataType(), posCol.NullN()); err != nil {
 		return err
 	}
 
-	posArr, ok := posCol.(*array.Int64)
-	if !ok {
-		return fmt.Errorf("%w: unsupported pos chunk array type %T in position delete file",
-			iceberg.ErrInvalidSchema, posCol)
-	}
+	posArr := posCol.(*array.Int64)
 	posCursor := posDeleteCursor{chunks: []*array.Int64{posArr}}
 	if err := a.appendFilePathChunk(ctx, filePathCol, &posCursor); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


table/arrow_scanner.go
```diff
@@ -361,106 +361,179 @@ func (c *posDeleteCursor) next() (int64, bool) {
...
+	for _, builder := range a.builders {
+		builder.Release()
 	}
-	if filePathValueType(filePathCol.DataType()).ID() == arrow.STRING_VIEW {
-		return nil, fmt.Errorf("%w: unsupported file_path column type %s in position delete file",
-			iceberg.ErrInvalidSchema, filePathCol.DataType())
+	a.builders = nil
+}
+
+func (a *posDeleteAccumulator) finish() map[string]*arrow.Chunked {
```
laskoviymishka, line 386: `finish()` and `release()` both set `a.builders = nil`, so if `finish()` is ever called after `release()` it ranges a nil map and returns an empty result with no error, and the accumulated deletes just vanish. Today's callers don't do this (success calls `finish()`, the error defer calls `release()`), but nothing on the type enforces it.

Since it's unexported, I'd rather a misuse fail loudly than return a silent empty map. A `panic` in `finish()` when `builders == nil`, or at least a contract comment, would cover it.

table/arrow_scanner.go
```diff
@@ -361,106 +361,179 @@ func (c *posDeleteCursor) next() (int64, bool) {
...
-				builder.Release()
+	var dictionary arrow.Array
+	var indices *array.Dictionary
+	if dict, ok := filePathChunk.(*array.Dictionary); ok && dict.Dictionary().NullN() > 0 {
+		dictionary = dict.Dictionary()
+		indices = dict
+	}
+
+	for i := range filePathChunk.Len() {
+		if i&(positionalDeleteCancellationCheckInterval-1) == 0 {
```
laskoviymishka, line 454: This `i & (positionalDeleteCancellationCheckInterval - 1) == 0` only works while the interval is a power of two. It is (16384), but now that it lives in the hot loop here I'd drop a one-line comment at the const so nobody swaps in a non-power-of-2 and quietly breaks the cancellation cadence.

table/arrow_scanner.go
```diff
@@ -361,106 +361,179 @@ func (c *posDeleteCursor) next() (int64, bool) {
...
+	if record.NumCols() != 2 {
+		return fmt.Errorf("%w: projected position delete record has %d columns, expected 2",
+			iceberg.ErrInvalidSchema, record.NumCols())
 	}
 
-	results = make(map[string]*arrow.Chunked, len(builders))
-	for path, builder := range builders {
-		positions := builder.NewInt64Array()
-		builder.Release()
+	filePathCol := record.Column(0)
```
laskoviymishka, line 519: I chased down whether this hardcoded ordering is safe, since `readDeletes` projects with `GetRecords(ctx, []int{filePathIndex, posIndex}, nil)` and here we assume col 0 is `file_path` and col 1 is `pos`. It's correct: arrow-go's `SchemaManifest.GetFieldIndices` builds the projected field list by iterating the request slice in order, and `GetFieldReaders` fills the output schema positionally from that, so the projection always comes back as `[file_path, pos]` regardless of the file's physical layout, even a reversed `[pos, file_path]` delete file.

What worries me is that nothing in the new path pins that. `TestReadDeletesProjectsColumnsAndAccumulatesBatches` only writes a standard-order file, so if arrow-go ever changed its emission order this would silently mis-assign the columns and surface as a confusing `unsupported pos column type string`. I'd add a `readDeletes` case that writes a reversed-schema file (pos physically first) and asserts the positions still group under the right paths. wdyt?

table/arrow_scanner.go
```diff
@@ -361,106 +361,179 @@ func (c *posDeleteCursor) next() (int64, bool) {
...
+	posCol := record.Column(1)
+	if err := validatePosDeleteColumns(filePathCol.DataType(), filePathCol.NullN(), filePathCol.Len(),
```
laskoviymishka: A couple of the checks in this path can't actually fire. `validatePosDeleteColumns` compares `filePathLen != posLen`, but these are two columns of the same `RecordBatch`, so their lengths are always equal (that cross-column check only had teeth in the old whole-file path). And it rejects `posType.ID() != arrow.INT64` right before the `posCol.(*array.Int64)` assertion just below, which can't independently fail for real pqarrow output.

Not a correctness problem, just dead branches that read as if they're guarding something. I'd either trim the length and INT64 checks from this call or drop the assertion and lean on `validatePosDeleteColumns`, so the two paths don't diverge on which one reports the error.

table/arrow_scanner.go
```diff
@@ -517,29 +590,42 @@ func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_
...
 
-	tbl, err = array.UnifyTableDicts(compute.GetAllocator(ctx), tbl)
+	filePathIndex, posIndex, err := positionDeleteColumnIndices(schema)
 	if err != nil {
 		return nil, err
 	}
-	defer tbl.Release()
 
-	filePathIndex, posIndex, err := positionDeleteColumnIndices(tbl.Schema())
+	records, err := rdr.GetRecords(ctx, []int{filePathIndex, posIndex}, nil)
```
laskoviymishka: Worth a one-line comment that we intentionally don't `UnifyTableDicts` on this path anymore. It's safe because `filePathValues` decodes each dictionary batch down to string values and we key the map on the string, so cross-batch dictionary independence doesn't matter. Without a note, I could see someone re-adding the unify call thinking the streaming path needs it.

table/arrow_scanner.go
```diff
@@ -517,29 +590,42 @@ func readDeletes(ctx context.Context, fs iceio.IO, dataFile iceberg.DataFile) (_
...
 		return nil, err
 	}
-	filePathCol := tbl.Column(filePathIndex).Data()
-	posCol := tbl.Column(posIndex).Data()
-	if posCol.NullN() > 0 {
-		return nil, fmt.Errorf("%w: null pos in position delete file", iceberg.ErrInvalidSchema)
+	defer records.Release()
+
+	acc := newPosDeleteAccumulator(ctx)
+	defer func() {
```
laskoviymishka, line 623: This deferred `release()` only fires on error because it reads the named return `err`, but every early return in here uses `err :=` in an inner scope. It's correct per spec (the `return nil, err` assigns the named return before defers run), but it's subtle, and a refactor to `err = ...` inside an `if` could quietly change what gets released. This defer is the only thing preventing an Arrow leak on the error path, so I'd add a short comment noting the dependency.

table/arrow_scanner_posdelete_regression_test.go
```diff
@@ -97,6 +97,68 @@ func TestReadDeletesRejectsMissingFilePath(t *testing.T) {
 	assert.Contains(t, err.Error(), `exactly one "file_path" column, found 0`)
 }
 
+func TestReadDeletesProjectsColumnsAndAccumulatesBatches(t *testing.T) {
```
laskoviymishka, line 100: This is the only end-to-end test of the new streaming path, and it never exercises dictionary-encoded `file_path`. That's the case that actually flows through `appendFilePathChunk`'s dictionary branch and the `dictionary.IsNull(...)` guard, and it's the primary memory win this PR is after. All the existing dictionary coverage goes through `groupPosDeletesByFilePath` into `appendChunked`, which `readDeletes` no longer calls.

Could we add a case that writes a delete file with repeated `file_path` values (so it comes back dictionary-encoded under `SetReadDict`) and runs it through `readDeletes`, asserting the grouping plus `mem.AssertSize(t, 0)`? A regression in the dict path wouldn't be caught today.

laskoviymishka (changes requested):
Nice work here. Streaming the projected `file_path`/`pos` columns instead of materializing the whole delete table is exactly the right instinct, and splitting the accumulator out of `groupPosDeletesByFilePath` reads cleanly.

My main concern is test coverage of the new path rather than the code itself. `readDeletes` now runs everything through `appendRecord`, but the bulk of the regression suite (dictionary layouts, string variants, negative pos, and so on) still exercises the old `appendChunked` path via `groupPosDeletesByFilePath`, which production no longer calls. The two cases I'd most want pinned on the new path are dictionary-encoded `file_path` (the primary memory win, and the branch most likely to regress) and a reversed-schema file with pos physically first. On the reversed case: I dug into the ordering and the projection is actually correct (arrow-go emits projected columns in request order, so col 0 is always `file_path`), but nothing currently locks that guarantee down and it rides on an arrow-go internal detail.

- a `readDeletes` test that exercises dictionary-encoded `file_path` end to end
- a `readDeletes` test on a reversed-schema delete file (pos before file_path)

The rest (the dead validation branches, the `finish()`-after-`release()` guard, and a couple of clarifying comments) are inline and non-blocking. Once those two tests are in, happy to take another pass and approve.

laskoviymishka (approved):

PR #1900, opened 2026-08-27
perf(table): skip equality-delete filtering for untouched batches

fallintoplace:
## What changed

- Allocate the equality-delete mask only after the first matching row.
- Return the input batch directly when no row matches.
- Keep the existing `compute.Filter` path for batches with matches.
- Add a no-match correctness test and integer/string benchmarks.

## Why

An equality delete set can be non-empty even when a record batch has no matching rows. The old path allocated and filled a full bitmap and called `compute.Filter` for every batch.

## Benchmark

Run on an Apple M1 Pro with:

`go test ./table -run '^$' -bench 'BenchmarkProcessEqualityDeletesNoMatch(Int|String)/rows=(100000|1000000)/deletes=10$' -benchmem -count=5`

Representative medians:

| Case | Before | After |
| --- | --- | --- |
| Int, 100K rows / 10 deletes | 2.76 ms/op, 2.03 MB/op, 86 allocs/op | 2.22 ms/op, 176 B/op, 4 allocs/op |
| Int, 1M rows / 10 deletes | 26.46 ms/op, 20.15 MB/op, 87 allocs/op | 22.12 ms/op, 176 B/op, 4 allocs/op |
| String, 100K rows / 10 deletes | 4.58 ms/op, 2.94 MB/op, 95 allocs/op | 2.63 ms/op, 208 B/op, 4 allocs/op |
| String, 1M rows / 10 deletes | 40.57 ms/op, 29.28 MB/op, 96 allocs/op | 26.18 ms/op, 208 B/op, 4 allocs/op |

The allocation reduction is the main result. Wall time had normal run-to-run noise.

## Testing

- `go test ./table -count=1`

### DIFF
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -781,14 +781,8 @@ func processEqualityDeletesColumnarForFile(ctx context.Context, eqDeleteSets []*
 		mem := compute.GetAllocator(ctx)
 		numRows := int(r.NumRows())
 
-		maskBuf := memory.NewResizableBuffer(mem)
-		defer maskBuf.Release()
-		maskBuf.Resize(int(bitutil.BytesForBits(int64(numRows))))
-		maskBytes := maskBuf.Bytes()
-
-		for i := range maskBytes {
-			maskBytes[i] = 0xFF
-		}
+		var maskBuf *memory.Buffer
+		var maskBytes []byte
 
 		var keyBuf bytes.Buffer
 
@@ -798,12 +792,16 @@ func processEqualityDeletesColumnarForFile(ctx context.Context, eqDeleteSets []*
 				var err error
 				encoders[i], err = makeArrowFieldEncoder(r, fieldRefs[setIdx][i], eqDel.fieldIDs[i], name, dataFilePath)
 				if err != nil {
+					if maskBuf != nil {
+						maskBuf.Release()
+					}
+
 					return nil, err
 				}
 			}
 
 			for row := range numRows {
-				if !bitutil.BitIsSet(maskBytes, row) {
+				if maskBytes != nil && !bitutil.BitIsSet(maskBytes, row) {
 					continue
 				}
 
@@ -813,21 +811,40 @@ func processEqualityDeletesColumnarForFile(ctx context.Context, eqDeleteSets []*
 					enc(&keyBuf, row)
 				}
 
-				if _, deleted := eqDel.keys[bufString(&keyBuf)]; deleted {
-					bitutil.ClearBit(maskBytes, row)
+				if _, deleted := eqDel.keys[bufString(&keyBuf)]; !deleted {
+					continue
 				}
+
+				if maskBuf == nil {
+					maskBuf = memory.NewResizableBuffer(mem)
+					maskBuf.Resize(int(bitutil.BytesForBits(int64(numRows))))
+					maskBytes = maskBuf.Bytes()
+
+					for i := range maskBytes {
+						maskBytes[i] = 0xFF
+					}
+				}
+
+				bitutil.ClearBit(maskBytes, row)
 			}
+		}
+
+		if maskBuf == nil {
+			r.Retain()
+
+			return r, nil
 		}
 
 		mask := array.NewBooleanData(array.NewData(
 			arrow.FixedWidthTypes.Boolean, numRows,
 			[]*memory.Buffer{nil, maskBuf}, nil, 0, 0))
-		defer mask.Release()
 
 		filtered, err := compute.Filter(ctx,
 			compute.NewDatumWithoutOwning(r),
 			compute.NewDatumWithoutOwning(mask),
 			*compute.DefaultFilterOptions())
+		mask.Release()
+		maskBuf.Release()
 		if err != nil {
 			return nil, err
 		}
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -792,10 +792,6 @@ func processEqualityDeletesColumnarForFile(ctx context.Context, eqDeleteSets []*
 				var err error
 				encoders[i], err = makeArrowFieldEncoder(r, fieldRefs[setIdx][i], eqDel.fieldIDs[i], name, dataFilePath)
 				if err != nil {
-					if maskBuf != nil {
-						maskBuf.Release()
-					}
-
 					return nil, err
 				}
 			}
@@ -817,6 +813,7 @@ func processEqualityDeletesColumnarForFile(ctx context.Context, eqDeleteSets []*
 
 				if maskBuf == nil {
 					maskBuf = memory.NewResizableBuffer(mem)
+					defer maskBuf.Release()
 					maskBuf.Resize(int(bitutil.BytesForBits(int64(numRows))))
 					maskBytes = maskBuf.Bytes()
 
@@ -835,16 +832,17 @@ func processEqualityDeletesColumnarForFile(ctx context.Context, eqDeleteSets []*
 			return r, nil
 		}
 
-		mask := array.NewBooleanData(array.NewData(
+		maskData := array.NewData(
 			arrow.FixedWidthTypes.Boolean, numRows,
-			[]*memory.Buffer{nil, maskBuf}, nil, 0, 0))
+			[]*memory.Buffer{nil, maskBuf}, nil, 0, 0)
+		mask := array.NewBooleanData(maskData)
+		maskData.Release()
+		defer mask.Release()
 
 		filtered, err := compute.Filter(ctx,
 			compute.NewDatumWithoutOwning(r),
 			compute.NewDatumWithoutOwning(mask),
 			*compute.DefaultFilterOptions())
-		mask.Release()
-		maskBuf.Release()
 		if err != nil {
 			return nil, err
 		}
--- a/table/equality_delete_reader_bench_test.go
+++ b/table/equality_delete_reader_bench_test.go
@@ -242,10 +242,60 @@ func buildBenchDeleteSetString(numDeletes int) *equalityDeleteSet {
 	}
 }
 
+func buildBenchDeleteSetIntNoMatch(numDeletes int) *equalityDeleteSet {
+	keys := make(set[string])
+	var buf bytes.Buffer
+
+	for i := range numDeletes {
+		buf.Reset()
+		buf.WriteByte(1)
+		_ = binary.Write(&buf, binary.BigEndian, int64(i*3))
+		buf.WriteByte(1)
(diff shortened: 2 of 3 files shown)


table/equality_delete_reader.go
```diff
@@ -813,21 +811,40 @@ func processEqualityDeletesColumnarForFile(ctx context.Context, eqDeleteSets []*
...
 		mask := array.NewBooleanData(array.NewData(
 			arrow.FixedWidthTypes.Boolean, numRows,
 			[]*memory.Buffer{nil, maskBuf}, nil, 0, 0))
-		defer mask.Release()
 
 		filtered, err := compute.Filter(ctx,
 			compute.NewDatumWithoutOwning(r),
 			compute.NewDatumWithoutOwning(mask),
 			*compute.DefaultFilterOptions())
+		mask.Release()
```
laskoviymishka: Small thing, and not a live bug: the old code had `defer mask.Release()` / `defer maskBuf.Release()`, so both got freed even if `compute.Filter` panicked. Now they're explicit calls right before the err check.

Arrow compute is synchronous and won't panic in practice, so this is fine today, but the defers gave that safety for free, and they'd also protect against a future failable call sneaking in between the release and the return. I'd lean toward keeping them as defers (register right after `maskBuf` is allocated and after `mask` is built). wdyt?

table/equality_delete_reader_bench_test.go
```diff
@@ -242,10 +242,60 @@ func buildBenchDeleteSetString(numDeletes int) *equalityDeleteSet {
...
 func BenchmarkProcessEqualityDeletesInt(b *testing.B) {
 	benchEqDeletes(b, buildBenchRecordInt, buildBenchDeleteSetInt)
 }
 
 func BenchmarkProcessEqualityDeletesString(b *testing.B) {
 	benchEqDeletes(b, buildBenchRecordString, buildBenchDeleteSetString)
 }
+
+func BenchmarkProcessEqualityDeletesNoMatchInt(b *testing.B) {
+	benchEqDeletes(b, buildBenchRecordInt, buildBenchDeleteSetIntNoMatch)
```
laskoviymishka: I think these two benchmarks are measuring the wrong function. `benchEqDeletes` routes through `processEqualityDeletes` → `processEqualityDeletesColumnar`, but the lazy-alloc optimization this PR adds lives only in `processEqualityDeletesColumnarForFile`, which this path never calls. So both no-match benchmarks are timing the unchanged eager-allocation code, and the numbers in the PR body don't actually demonstrate the improvement.

I'd add a `benchEqDeletesForFile` helper that calls `processEqualityDeletesColumnarForFile(ctx, delSets, fileSchema, "bench.parquet")` and point the two no-match benchmarks at it. Then the before/after should actually show the skip. wdyt?

table/equality_delete_reader_internal_test.go
```diff
@@ -481,6 +481,43 @@ func TestProcessEqualityDeletesUsesStructuralFieldPaths(t *testing.T) {
 	result.Release()
 }
 
+func TestProcessEqualityDeletesReturnsOriginalBatchWhenNoRowsMatch(t *testing.T) {
+	builder := array.NewInt64Builder(memory.DefaultAllocator)
```
laskoviymishka: This test validates the no-match return path, which does `r.Retain()` balanced by `defer r.Release()`, but with `memory.DefaultAllocator` a refcount imbalance there goes completely undetected. The other tests in this file use `memory.NewCheckedAllocator(memory.DefaultAllocator)` with `defer mem.AssertSize(t, 0)` for exactly this reason. Since leak-safety on the retain/release dance is the whole point of this test, I'd switch it to the checked allocator.

table/equality_delete_reader_internal_test.go
```diff
@@ -481,6 +481,43 @@ func TestProcessEqualityDeletesUsesStructuralFieldPaths(t *testing.T) {
...
+		Name: "id", Type: arrow.PrimitiveTypes.Int64,
+	}}, nil)
+	record := array.NewRecordBatch(schema, []arrow.Array{values}, 2)
+	values.Release()
+
+	fileSchema := iceberg.NewSchema(0,
+		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64},
+	)
+	process, err := processEqualityDeletesColumnarForFile(context.Background(), []*equalityDeleteSet{{
+		keys:     set[string]{unmatchedKey.String(): {}},
```
laskoviymishka: This only exercises a single delete set. The interesting new behavior is across sets: the `maskBytes != nil` guard in the row loop and the mid-loop lazy allocation only fire once there are two or more sets. Two cases worth covering: set 0 matches (allocates the mask) and set 1 then skips an already-cleared row, and the inverse where set 0 matches nothing (mask stays nil) and set 1 matches and allocates mid-loop. I'd add a two-set test asserting both rows drop out. wdyt?

table/equality_delete_reader_internal_test.go
```diff
@@ -481,6 +481,43 @@ func TestProcessEqualityDeletesUsesStructuralFieldPaths(t *testing.T) {
...
+	schema := arrow.NewSchema([]arrow.Field{{
+		Name: "id", Type: arrow.PrimitiveTypes.Int64,
+	}}, nil)
+	record := array.NewRecordBatch(schema, []arrow.Array{values}, 2)
+	values.Release()
+
+	fileSchema := iceberg.NewSchema(0,
+		iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64},
+	)
+	process, err := processEqualityDeletesColumnarForFile(context.Background(), []*equalityDeleteSet{{
```
laskoviymishka: Nit: the rest of the tests in this file pass `t.Context()`. I'd use it here too for consistency; it also gets cancelled when the test finishes, unlike `context.Background()`.

table/equality_delete_reader_internal_test.go
```diff
@@ -481,6 +481,43 @@ func TestProcessEqualityDeletesUsesStructuralFieldPaths(t *testing.T) {
 	result.Release()
 }
 
+func TestProcessEqualityDeletesReturnsOriginalBatchWhenNoRowsMatch(t *testing.T) {
+	builder := array.NewInt64Builder(memory.DefaultAllocator)
+	builder.AppendValues([]int64{1, 2}, nil)
+	values := builder.NewArray()
+	builder.Release()
+
+	unmatchedBuilder := array.NewInt64Builder(memory.DefaultAllocator)
```
laskoviymishka: Minor: this builds a whole Int64 array just to encode the key for `int64(3)`, then releases it. `buildBenchDeleteSetInt` writes the key bytes directly. That said, going through `encodeArrowValue` keeps the key guaranteed-consistent with what the reader produces, so I'm honestly fine either way, just flagging the extra ceremony.

table/equality_delete_reader_bench_test.go
```diff
@@ -242,10 +242,60 @@ func buildBenchDeleteSetString(numDeletes int) *equalityDeleteSet {
...
+		_ = binary.Write(&buf, binary.BigEndian, int64(i*3))
+		buf.WriteByte(1)
+		_ = binary.Write(&buf, binary.BigEndian, int64((i*3+1)%100))
+		keys[buf.String()] = struct{}{}
+	}
+
+	return &equalityDeleteSet{
+		keys:     keys,
+		fieldIDs: []int{1, 2},
+		colNames: []string{"id", "category"},
```
laskoviymishka, line 286: These `colNames` (`"id"`/`"category"`) don't line up with the field names `benchEqDeletes` puts in `fileSchema` (`"first"`/`"second"`). Harmless since resolution is by field ID, and the existing benchmarks already do this, but a mismatched name would show up in any field-resolution error message, so while we're here I'd align them.

table/equality_delete_reader_bench_test.go
```diff
@@ -242,10 +242,60 @@ func buildBenchDeleteSetString(numDeletes int) *equalityDeleteSet {
...
 }
 
+func buildBenchDeleteSetIntNoMatch(numDeletes int) *equalityDeleteSet {
+	keys := make(set[string])
+	var buf bytes.Buffer
+
+	for i := range numDeletes {
+		buf.Reset()
+		buf.WriteByte(1)
+		_ = binary.Write(&buf, binary.BigEndian, int64(i*3))
```
laskoviymishka, line 277: Tiny style thing: the existing `buildBenchDeleteSet*` helpers call `binary.Write` bare, and these new ones assign to `_`. Writing to a `bytes.Buffer` never errors so both are safe, it's just now split within one file. I'd match one way or the other (errcheck prefers the `_ =`, so probably add it to the existing ones).

laskoviymishka (changes requested):
Nice change: skipping `compute.Filter` entirely when a batch has no matching rows is a real saving on the hot path, and I like that it comes with benchmarks and a targeted test rather than just an assertion that it's faster.

I'd hold it before merging though, mostly because I don't think the evidence backs the claim yet. The two new no-match benchmarks go through `benchEqDeletes`, which as far as I can tell dispatches to `processEqualityDeletes` → `processEqualityDeletesColumnar`. But the lazy-allocation optimization only lives in `processEqualityDeletesColumnarForFile`, which those benchmarks never call, so they're timing the unchanged, eager-allocation path, and the numbers in the PR body are measuring the wrong function. I'd add a `benchEqDeletesForFile` helper that calls the `...ForFile` variant and point the no-match benchmarks at it.

The other thing is test confidence on the path this actually changes. The new test uses `memory.DefaultAllocator`, so the retain/release dance in the no-match return, the exact thing the test exists to check, has no leak detection; a checked allocator with `AssertSize(t, 0)` closes that. And it only covers the single-set case, while the `maskBytes != nil` guard and the mid-loop lazy alloc only fire with two or more delete sets, so I'd add a two-set test as well.

A few smaller things are inline. Once the benchmark measures the right function and the new path is properly covered, happy to take another pass.

zeroshade (approved):
Thanks for tightening this up. The benchmarks now exercise the actual `processEqualityDeletesColumnarForFile` path, the checked-allocator tests pin the no-match retain/release behavior, both multi-delete-set lazy-allocation transitions are covered, and the mask resources unwind safely. I also re-ran the focused race tests repeatedly, the full table suite, vet, the benchmark matrix, and a synthetic merge with current main; all passed.

The prior Audit and Verify failure was unrelated to these changes: `TestSqlCatalog/TestConcurrentTableViewCollisionReturnsCatalogSentinel` hit `SQLITE_BUSY`. That test passed locally 10/10 both on this head and after merging current main, so this is consistent with a CI flake; the rerun can gate merge.

Looks good to merge once the required check is green.

laskoviymishka (approved):

PR #1905, opened 2026-08-27
fix(table): reject deleted schema move targets

### LINKED ISSUE #1903: UpdateSchema: moving relative to a deleted column silently drops the moved column
### Apache Iceberg version

main (development) at `c9813324fb8617ba5c889d939b0d883e85f7eb1e`

### Please describe the bug 🐞

`UpdateSchema` silently drops an additional column when a move is staged relative to a column that is deleted in the same update.

Given a schema with `id`, `name`, and `age`, both operation orders succeed:

```go
update.DeleteColumn([]string{"name"}).
    MoveBefore([]string{"age"}, []string{"name"})
```

```go
update.MoveBefore([]string{"age"}, []string{"name"}).
    DeleteColumn([]string{"name"})
```

`Apply()` returns no error, but the resulting schema contains only `id`. The requested deletion of `name` also removes `age` unintentionally. `Commit()` builds its persisted schema update from this result.

Minimal reproduction:

```go
schema := iceberg.NewSchema(1,
    iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int32, Required: true},
    iceberg.NestedField{ID: 2, Name: "name", Type: iceberg.PrimitiveTypes.String},
    iceberg.NestedField{ID: 3, Name: "age", Type: iceberg.PrimitiveTypes.Int32},
)
metadata, err := table.NewMetadata(schema, nil, table.UnsortedSortOrder, "", nil)
require.NoError(t, err)
tbl := table.New(table.Identifier{"db", "people"}, metadata, "", nil, nil)

updated, err := table.NewUpdateSchema(tbl.NewTransaction(), true, true).
    DeleteColumn([]string{"name"}).
    MoveBefore([]string{"age"}, []string{"name"}).
    Apply()

require.NoError(t, err)
require.Equal(t, 1, updated.NumFields())
require.Equal(t, "id", updated.Field(0).Name) // age is silently lost
```

Expected behavior: reject the conflicting operations with an actionable error. A move cannot be applied relative to a field that will not exist in the resulting schema.

Root cause: `moveColumn` validates that the moved field is not deleted, but does not validate its relative target. Later, `moveFields` removes the moved field before searching for the target; when the target is absent, it continues without reinserting the moved field.

This is distinct from #1685/#1686 (moving newly added nested fields), #1687/#1688 (adding beneath deleted ancestors), and #1778/#1779 (deleting a field with its own pending update).

Proposed fix: #1905, with regression coverage for both operation orders and both `MoveBefore` and `MoveAfter`.

mattfaltyn:
### Description

Fixes #1903.

Reject schema updates that either move a field relative to a column already staged for deletion or delete a column already referenced as a move target. This prevents `moveFields` from silently omitting the moved field when the target is absent from the resulting schema.

The regression test covers both operation orders and both `MoveBefore` and `MoveAfter`.

### Testing

- `go test ./table -run 'TestMoveRelativeToDeletedColumnRejected|TestMoveColumn|TestDeleteColumn|TestChainedOperations' -count=1`
- `make test`
- `make test-assert`
- `make test-race`
- `make lint`
- `go build ./...`

### DIFF
--- a/table/update_schema.go
+++ b/table/update_schema.go
@@ -388,6 +388,11 @@ func (u *UpdateSchema) deleteColumn(path []string) error {
 			return fmt.Errorf("field that has updates cannot be deleted: %s", fullName)
 		}
 	}
+	for _, move := range u.moves[parentID] {
+		if move.RelativeTo == field.ID {
+			return fmt.Errorf("field that is used as a move target cannot be deleted: %s", fullName)
+		}
+	}
 
 	delete(u.identifierFieldNames, fullName)
 
@@ -625,6 +630,9 @@ func (u *UpdateSchema) moveColumn(op MoveOp, path []string, relativeTo []string)
 
 		if relativeToFieldID == fieldID {
 			return fmt.Errorf("cannot move a field to itself: %s", fullName)
+		}
+		if u.isDeleted(relativeToFieldID) {
+			return fmt.Errorf("field that has been deleted cannot be used as a move target: %s", relativeToFullName)
 		}
 
 		if u.findParentID(relativeToFieldID) != parentID {
--- a/table/update_schema.go
+++ b/table/update_schema.go
@@ -1388,7 +1388,7 @@ func moveFields(fields []iceberg.NestedField, moves []move) []iceberg.NestedFiel
 			}
 		}
 		if !found {
-			continue
+			panic(fmt.Errorf("cannot move field %d: field not found", move.FieldID))
 		}
 
 		reordered = append(reordered[:fieldIndex], reordered[fieldIndex+1:]...)
@@ -1408,7 +1408,7 @@ func moveFields(fields []iceberg.NestedField, moves []move) []iceberg.NestedFiel
 				}
 			}
 			if !found {
-				continue
+				panic(fmt.Errorf("cannot move field %d: target field %d not found", move.FieldID, move.RelativeTo))
 			}
 
 			if move.Op == MoveOpBefore {
--- a/table/update_schema_test.go
+++ b/table/update_schema_test.go
@@ -490,6 +490,16 @@ func TestApplyChanges(t *testing.T) {
 			{ID: 4, Name: "name", Type: iceberg.PrimitiveTypes.UUID, Required: false, Doc: ""},
 		}, st.(*iceberg.StructType).Fields())
 	})
+
+	t.Run("test move with missing target fails", func(t *testing.T) {
+		_, err := iceberg.Visit(originalSchema, &applyChanges{
+			deletes: map[int]struct{}{2: {}},
+			moves: map[int][]move{
+				TableRootID: {{FieldID: 3, RelativeTo: 2, Op: MoveOpBefore}},
+			},
+		})
+		require.ErrorContains(t, err, "cannot move field 3: target field 2 not found")
+	})
 }
 
 func TestDeleteColumn(t *testing.T) {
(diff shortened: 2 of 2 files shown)


table/update_schema.go
```diff
@@ -388,6 +388,11 @@ func (u *UpdateSchema) deleteColumn(path []string) error {
 			return fmt.Errorf("field that has updates cannot be deleted: %s", fullName)
 		}
 	}
+	for _, move := range u.moves[parentID] {
```
laskoviymishka, line 391: the two guards close both orderings in #1903, but they're patching entry points — the real silent drop is in `moveFields`, where the `!found { continue }` branch (~line 1370) removes the field from the working slice and then bails with no error. this PR makes that branch unreachable for the tested orderings, but it's still a landmine for any future path.

and there's already one it doesn't cover: `MoveBefore(["age"], ["name"])` then `DeleteColumn(["age"])`. this guard only inspects `move.RelativeTo`, not `move.FieldID`, so neither guard fires and age's staged move is silently swallowed at apply (output happens to be correct, but there's no diagnostic).

I'd rather fix `moveFields` to not drop on `!found` — restore the field and continue, or return an error — so the root is closed in one place instead of enumerating call-site orderings. wdyt?

table/update_schema_test.go
```diff
@@ -887,6 +887,47 @@ func TestMoveColumn(t *testing.T) {
...
+				return update.MoveAfter([]string{"age"}, []string{"name"}).DeleteColumn([]string{"name"})
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tbl := New([]string{"id"}, testMetadata, "", nil, nil)
+			_, err := tt.build(NewUpdateSchema(tbl.NewTransaction(), true, true)).Apply()
+			require.ErrorContains(t, err, "move target")
```
laskoviymishka: these assertions can't tell the two guards apart — both error messages contain "move target", so cases 1/3 exercise the `moveColumn` guard and 2/4 the `deleteColumn` guard, but all four pass on the same substring. if one guard broke and the other picked up the slack, this would stay green.

I'd split it by the distinguishing text: assert `"cannot be deleted"` for the move-then-delete cases and `"has been deleted"` for the delete-then-move cases.

table/update_schema_test.go
```diff
@@ -887,6 +887,47 @@ func TestMoveColumn(t *testing.T) {
 	})
 }
 
+func TestMoveRelativeToDeletedColumnRejected(t *testing.T) {
```
laskoviymishka, line 900: all four cases expect an error, so this proves the guards fire but not that they fire only when they should. I'd add a positive case — `DeleteColumn(["name"])` then `MoveBefore(["age"], ["id"])` should still succeed and land in the expected order — so an over-eager guard can't slip through unnoticed.

table/update_schema_test.go
```diff
@@ -887,6 +887,47 @@ func TestMoveColumn(t *testing.T) {
...
+	tests := []struct {
+		name  string
+		build func(*UpdateSchema) *UpdateSchema
+	}{
+		{
+			name: "delete then move before",
+			build: func(update *UpdateSchema) *UpdateSchema {
+				return update.DeleteColumn([]string{"name"}).MoveBefore([]string{"age"}, []string{"name"})
```
laskoviymishka, line 909: these are all flat top-level fields, so the `deleteColumn` guard only ever scans `u.moves[-1]` (parentID resolves to the root). a nested target like `address.city` keys its move under `address`'s ID — a different bucket that's currently untested. one nested case would pin that path down.

laskoviymishka (changes requested):
Nice, targeted fix: the two symmetric guards are the right instinct, and they close the exact #1903 repro cleanly in both orderings.

I'd hold for now. The guards patch the two entry points, but the actual silent drop lives in `moveFields`: the `!found { continue }` branch (~line 1370) pulls the field out of the working slice and then bails with no error. This PR makes that branch unreachable for the two tested orderings, but it's still a landmine for any future path — and there's already one it doesn't cover: `MoveBefore(["age"], ["name"])` then `DeleteColumn(["age"])` slips past both guards and silently swallows the staged move. I'd rather fix `moveFields` to not drop on `!found` so the root is closed in one place, instead of enumerating call-site orderings.

A couple of things I'd like to settle before merge:
- fix (or at least guard) the silent drop at its root in `moveFields`, so the move-source-delete ordering surfaces an error too
- tighten the test to distinguish the two guards (split on the distinct messages) and add a positive case where a delete plus an unrelated move still succeeds

Once those are settled, happy to take another pass.

table/update_schema.go
mattfaltyn: > “fix `moveFields` to not drop on `!found`”

Great catch — `moveFields` now fails closed for missing sources or targets, with source-delete and direct target-loss coverage.

table/update_schema_test.go
mattfaltyn: > “split it by the distinguishing text”

Good suggestion — the two orderings now assert `has been deleted` and `cannot be deleted` separately.

table/update_schema_test.go
mattfaltyn: > “add a positive case”

Agreed — unrelated deletion plus move now succeeds and asserts the final field order.

table/update_schema_test.go
mattfaltyn: > “one nested case would pin that path down”

Nice call — added a nested `address.city` target case to exercise the parent-ID bucket.

laskoviymishka (approved):
🌬️ ⛵

PR #1907, opened 2026-08-27
perf(table): borrow equality-delete partitions

fallintoplace:
## What

- Use the existing borrowed partition view while building the equality-delete index.
- Use it again while matching data files against partitioned equality deletes.
- Keep the public `DataFile.Partition()` defensive-copy behavior unchanged.
- Add a focused benchmark using the built-in `DataFile` implementation.

## Why

`DataFile.Partition()` creates a defensive copy. The equality-delete index only reads the partition while it immediately builds a lookup key, so the copy is unnecessary on this trusted internal path.

The existing `dataFilePartition` helper already falls back to `Partition()` for external `DataFile` implementations.

## Benchmark

Command:

```text
go test ./table -run '^TestEqualityDeleteIndex' -bench '^BenchmarkEqualityDeleteIndexBuiltInPartition$' -benchmem -benchtime=200ms -count=5
```

Median of 5 runs on an Apple M1 Pro with Go 1.26.3. Each value is `ns/op, B/op, allocs/op`.

| Case | Before | After |
| --- | --- | --- |
| lookup, 1 field | 236.5, 256, 2 | 83.1, 0, 0 |
| lookup, 8 fields | 675.2, 648, 9 | 486.0, 392, 7 |
| lookup, 32 fields | 2542, 3472, 13 | 1788, 1592, 9 |
| build, 1 field | 639.3, 1112, 8 | 527.0, 856, 6 |
| build, 8 fields | 1086, 1512, 14 | 937.7, 1256, 12 |
| build, 32 fields | 3095, 4352, 18 | 2182, 2472, 14 |

## Testing

- `go test ./table -race -run '^TestEqualityDeleteIndex' -count=1 -timeout=5m`

### DIFF
--- a/table/equality_delete_index.go
+++ b/table/equality_delete_index.go
@@ -159,7 +159,7 @@ func buildEqualityDeleteIndex(
 			continue
 		}
 
-		partition := df.Partition()
+		partition := dataFilePartition(df)
 		key, err := newEqualityDeletePartitionKey(df.SpecID(), partition)
 		if err != nil {
 			return nil, fmt.Errorf("indexing equality delete file %s: %w", df.FilePath(), err)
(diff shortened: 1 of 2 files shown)


laskoviymishka (approved):
👍

PR #1911, opened 2026-08-27
perf(table): precompute IN predicate extrema

fallintoplace:
**What changed**

- **Precompute min/max for `IN`** when the bound expression is created.
- **Use one comparison per manifest** instead of scanning every literal for both partition bounds.
- **Keep existing visitor behavior** for public visitors and literals without a total ordering.
- Added focused tests and a benchmark for int32, string, decimal, binary, and nanosecond timestamp values.

**Benchmark**

- **Machine:** Apple M1 Pro
- **200 literals, lower-bound prune, median:**
  - int32: **3.14 µs -> 0.18 µs**
  - string: **3.77 µs -> 0.20 µs**
  - decimal: **3.52 µs -> 0.24 µs**
- **Allocations:** int32 **192 B / 7 -> 144 B / 6**; string **208 B / 8 -> 160 B / 7**; decimal **240 B / 10 -> 192 B / 9**
- Command: `go test ./table -run '^$' -bench '^BenchmarkManifestEvaluatorInPredicate$' -benchmem -benchtime=200ms -count=5`

**Checks**

- `go test . ./table ./table/dv ./table/substrait -count=1`
- `go test -race . ./table -run '^(TestLiteralSetExtrema|TestVisitBoundPredicateRefPassesInExtrema|TestVisitBoundPredicateRefDoesNotAllocate|TestManifestEvaluatorInPredicateExtrema|TestManifestEvaluator|TestManifestEvalVisitorEvalRace)$' -count=1`
- `go vet ./...`

### DIFF
--- a/exprs.go
+++ b/exprs.go
@@ -1008,14 +1008,62 @@ func createBoundSetPredicate(op Operation, term BoundTerm, lits Set[Literal]) (B
 		ErrType, term.Type())
 }
 
-func newBoundSetPredicate[T LiteralType](op Operation, term BoundTerm, lits Set[Literal]) *boundSetPredicate[T] {
-	return &boundSetPredicate[T]{op: op, term: term, lits: lits}
+func newBoundSetPredicate[T LiteralType](op Operation, term BoundTerm, lits literalSet) *boundSetPredicate[T] {
+	predicate := &boundSetPredicate[T]{op: op, term: term, lits: lits}
+	if op == OpIn {
+		predicate.minLiteral, predicate.maxLiteral, predicate.hasExtrema = literalSetExtrema[T](lits)
+	}
+
+	return predicate
 }
 
 type boundSetPredicate[T LiteralType] struct {
-	op   Operation
-	term BoundTerm
-	lits Set[Literal]
+	op         Operation
+	term       BoundTerm
+	lits       Set[Literal]
+	minLiteral Literal
+	maxLiteral Literal
+	hasExtrema bool
+}
+
+func literalSetExtrema[T LiteralType](lits literalSet) (min, max Literal, ok bool) {
+	var (
+		cmp                Comparator[T]
+		minValue, maxValue T
+		valid              = true
+	)
+
+	lits.All(func(lit Literal) bool {
+		typed, typeOK := lit.(TypedLiteral[T])
+		if !typeOK {
+			valid = false
+			return false
+		}
+
+		value := typed.Value()
+		if !ok {
+			cmp = typed.Comparator()
+			min, max = lit, lit
+			minValue, maxValue = value, value
+			ok = true
+			return true
+		}
+
+		if cmp(value, minValue) < 0 {
+			min, minValue = lit, value
+		}
+		if cmp(value, maxValue) > 0 {
+			max, maxValue = lit, value
+		}
+
+		return true
+	})
+
+	if !valid {
+		return nil, nil, false
+	}
+
+	return min, max, ok
 }
 
 func (bsp *boundSetPredicate[T]) Equals(other BooleanExpression) bool {
@@ -1031,8 +1079,12 @@ func (bsp *boundSetPredicate[T]) Equals(other BooleanExpression) bool {
 func (bsp *boundSetPredicate[T]) Op() Operation { return bsp.op }
 func (bsp *boundSetPredicate[T]) Negate() BooleanExpression {
 	return &boundSetPredicate[T]{
-		op: bsp.op.Negate(), term: bsp.term,
-		lits: bsp.lits,
+		op:         bsp.op.Negate(),
+		term:       bsp.term,
+		lits:       bsp.lits,
+		minLiteral: bsp.minLiteral,
+		maxLiteral: bsp.maxLiteral,
+		hasExtrema: bsp.hasExtrema,
 	}
 }
 func (bsp *boundSetPredicate[T]) Term() BoundTerm     { return bsp.term }
--- a/exprs.go
+++ b/exprs.go
@@ -1026,7 +1026,7 @@ type boundSetPredicate[T LiteralType] struct {
 	hasExtrema bool
 }
 
-func literalSetExtrema[T LiteralType](lits literalSet) (min, max Literal, ok bool) {
+func literalSetExtrema[T LiteralType](lits literalSet) (minLit, maxLit Literal, ok bool) {
 	var (
 		cmp                Comparator[T]
 		minValue, maxValue T
@@ -1044,7 +1044,7 @@ func literalSetExtrema[T LiteralType](lits literalSet) (min, max Literal, ok boo
 		value := typed.Value()
 		if !ok {
 			cmp = typed.Comparator()
-			min, max = lit, lit
+			minLit, maxLit = lit, lit
 			minValue, maxValue = value, value
 			ok = true
 
@@ -1052,10 +1052,10 @@ func literalSetExtrema[T LiteralType](lits literalSet) (min, max Literal, ok boo
 		}
 
 		if cmp(value, minValue) < 0 {
-			min, minValue = lit, value
+			minLit, minValue = lit, value
 		}
 		if cmp(value, maxValue) > 0 {
-			max, maxValue = lit, value
+			maxLit, maxValue = lit, value
 		}
 
 		return true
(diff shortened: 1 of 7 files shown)


exprs.go
```diff
@@ -1008,14 +1008,64 @@ func createBoundSetPredicate(op Operation, term BoundTerm, lits Set[Literal]) (B
...
-	lits Set[Literal]
+	op         Operation
+	term       BoundTerm
+	lits       Set[Literal]
+	minLiteral Literal
+	maxLiteral Literal
+	hasExtrema bool
+}
+
+func literalSetExtrema[T LiteralType](lits literalSet) (min, max Literal, ok bool) {
```
laskoviymishka: `min` and `max` have been builtins since Go 1.21 and the module's on 1.25, so these named returns shadow them — `golangci-lint`'s `predeclared` will flag it (same for the `boundSetExtremaRef` signature in `bound_set_literal_view.go` and the destructuring in `visitors.go`). I'd rename to `minLit`/`maxLit` to match the `minLiteral`/`maxLiteral` struct fields, consistently across all three sites.

exprs.go
```diff
@@ -1031,8 +1081,12 @@ func (bsp *boundSetPredicate[T]) Equals(other BooleanExpression) bool {
 func (bsp *boundSetPredicate[T]) Op() Operation { return bsp.op }
 func (bsp *boundSetPredicate[T]) Negate() BooleanExpression {
 	return &boundSetPredicate[T]{
-		op: bsp.op.Negate(), term: bsp.term,
-		lits: bsp.lits,
+		op:         bsp.op.Negate(),
+		term:       bsp.term,
```
laskoviymishka: These extrema get carried onto the `NOT IN` result, but the dispatch in `visitBoundPredicate` only consults them on `OpIn`, so for `OpNotIn` they're stored and never read — 24 bytes per negated predicate for nothing. I'd set `hasExtrema: false` here. (Minor gotcha the other way: a double-negate back to `OpIn` loses them, since `Negate` doesn't recompute.)

table/evaluators.go
```diff
@@ -162,7 +164,53 @@ func allBoundCheck(bound iceberg.Literal, set iceberg.Set[iceberg.Literal], want
 	panic(iceberg.ErrType)
 }
 
+func compareBound[T iceberg.LiteralType](bound, value iceberg.Literal) int {
+	val := bound.(iceberg.TypedLiteral[T])
+
+	return val.Comparator()(val.Value(), value.(iceberg.TypedLiteral[T]).Value())
+}
+
+func boundCompare(bound, value iceberg.Literal) int {
```
laskoviymishka: `boundCompare` is a second copy of the type dispatch `getCmpLiteral` already does — `boundCompare(a, b)` is just `getCmpLiteral(a)(a, b)`. The only reason it exists is that `getCmpLiteral` is missing the `TimestampNano` case, so rather than adding it there the PR routes around it with a parallel switch.

The catch is that leaves the root gap in place: `getCmpLiteral` still panics on nanosecond-timestamp columns, so `VisitEqual`/`VisitGreater`/etc. still blow up for a TimestampNs partition column even though the IN path now handles it. `removeBoundCheck` has the same hole for the inclusive/strict metrics IN paths. So we've added TimestampNs coverage to three dispatchers but not the two that share the same shape.

I'd add the `TimestampNano` case to `getCmpLiteral` (and `removeBoundCheck`), then collapse `boundCompare` to `getCmpLiteral(bound)(bound, value)` and drop `compareBound` — that finishes the TimestampNano fix everywhere and removes ~28 lines of parallel dispatch. wdyt?

table/evaluators.go
```diff
@@ -179,7 +227,11 @@ func (m *manifestEvalVisitor) VisitIn(term iceberg.BoundTerm, literals iceberg.S
 		panic(err)
 	}
 
-	if allBoundCheck(lower, literals, 1) {
+	if max != nil {
```
laskoviymishka: I think the extrema check never runs for the sets this is meant to speed up. The `literals.Len() > inPredicateLimit` guard above still fires first and returns `rowsMightMatch`, so any IN set past 200 values short-circuits before we reach `if max != nil` — and extrema get computed at bind time for every set size regardless. The O(1) win is exactly inert for the large predicates that motivated the change.

That guard only makes sense for the O(n) `allBoundCheck` fallback. I'd move it into the `else` arm so the extrema comparison runs unconditionally, and we only fall through to the length guard + `allBoundCheck` when extrema aren't available (`max == nil`).

The benchmark tops out at exactly 200 literals, right at the cutoff, so this regime never gets exercised — a 201+ case with a disjoint partition range would return `rowsMightMatch` where we'd want `rowsCannotMatch`. wdyt?

table/evaluators.go
```diff
@@ -179,7 +227,11 @@ func (m *manifestEvalVisitor) VisitIn(term iceberg.BoundTerm, literals iceberg.S
...
+		if boundCompare(lower, max) == 1 {
```
laskoviymishka: The `Comparator` contract in `literals.go` is defined as `< 0` / `> 0`, not `±1`, so checking `== 1` here (and `== -1` on the upper bound) is a little fragile — a conforming comparator that returns 2 would silently skip the prune and we'd read the manifest anyway. I'd use `boundCompare(lower, max) > 0` and `boundCompare(upper, min) < 0`.

There are pre-existing `== 1`/`== -1` sites elsewhere in this file so this isn't new, but since we're adding two more I'd rather not extend the pattern.

table/in_predicate_extrema_test.go
```diff
@@ -0,0 +1,114 @@
...
+			expectRead: false,
+		},
+		{
+			name:       "timestamp nanos overlaps bound",
+			typ:        iceberg.PrimitiveTypes.TimestampNs,
+			expr:       iceberg.IsIn(iceberg.Reference("value"), iceberg.TimestampNano(100), iceberg.TimestampNano(200), iceberg.TimestampNano(300)),
+			lower:      iceberg.NewLiteral(iceberg.TimestampNano(200)),
+			upper:      iceberg.NewLiteral(iceberg.TimestampNano(250)),
+			expectRead: true,
+		},
```
laskoviymishka, line 91: Could we add a case with 201+ literals and a clearly disjoint partition range here? That's the one that would catch the `inPredicateLimit` ordering issue — today it returns `rowsMightMatch` (read) where the extrema say `rowsCannotMatch` (prune). A boundary case where `lower == max` (expecting a read) would lock the `== 1` edge too.

visitors.go
```diff
@@ -82,6 +82,10 @@ type BoundGeospatialExprVisitor[T any] interface {
 	VisitBBoxNotIntersects(BoundTerm, BoundingBox) T
 }
 
+type boundSetExtremaExprVisitor[T any] interface {
```
laskoviymishka, line 88: Worth a doc comment here: `VisitInWithExtrema` only gets called on the `borrowed` path in `visitBoundPredicate`, so a visitor that implements it but is driven through the public `VisitBoundPredicate` will silently never see it. The geospatial extension interface right above documents its dispatch contract explicitly; I'd mirror that.

Related, only `manifestEvalVisitor` implements it today — the other three evaluators now go through `VisitBoundPredicateRef` and fall back to `VisitIn`, so inclusive/strict metrics and row-group filtering still scan every literal. Fine to leave for a follow-up, but a note here keeps it from reading as if all four are optimized.

laskoviymishka (changes requested):
Nice change — precomputing the IN extrema at bind time and collapsing the manifest check to a single comparison is the right idea, and the benchmark harness makes the intent easy to follow.

I'd hold, though, mostly because the optimization doesn't actually fire where it's supposed to. The `literals.Len() > inPredicateLimit` guard still runs ahead of the new extrema branch, so any IN set past 200 values short-circuits to `rowsMightMatch` before we ever look at min/max — which is exactly the large-predicate case the PR is meant to speed up. Moving the guard into the fallback arm fixes it, and a 201-literal test with a disjoint range would keep it fixed.

The other thing I'd want settled here is the `boundCompare` dispatch. It's a second copy of what `getCmpLiteral` already does, added only because `getCmpLiteral` is missing the `TimestampNano` case — but routing around that leaves `VisitEqual`/`VisitGreater`/etc. still panicking on nanosecond-timestamp columns, and `removeBoundCheck` has the same hole. Adding the case to `getCmpLiteral` and reusing it finishes the TimestampNano fix and drops ~28 lines.

A few smaller things I'd fix while we're here:
- the extrema comparisons use `== 1`/`== -1`, but the `Comparator` contract is `> 0`/`< 0` — a comparator returning ±2 would silently skip the prune
- `min`/`max` named returns shadow the Go builtins and will trip `golangci-lint`'s `predeclared`
- `Negate()` copies extrema onto the `NOT IN` result, where they're never read
- a doc note that `VisitInWithExtrema` only fires on the borrowed path and only `manifestEvalVisitor` implements it today

Once the ordering and the dispatch are sorted, happy to take another pass and approve.

laskoviymishka (approved):

PR #1912, opened 2026-08-27
perf(table): bypass no-op writer schema projection

fallintoplace:
## What

- **Skip the writer schema projection** when the incoming Arrow batch already matches the target Arrow schema.
- Applies to both the default writer and rolling writer.
- Compares field order, physical types, nullability, field metadata, and top-level metadata.
- Keeps the existing projection for reordered columns, missing/default fields, timestamp downcasts, list offset changes, and metadata mismatches.
- Adds checked-allocator identity tests and a focused benchmark.

## Benchmark

`BenchmarkDefaultDataFileWriter` with 128 rows, 5 runs on an Apple M1 Pro:

- **Before:** 4.6 to 5.3 us/op, about 5.0 KB/op, 73 allocs/op
- **After:** 1.7 to 3.1 us/op, about 0.9 KB/op, 15 allocs/op

The focused benchmark measured the exact-schema path at about **35 to 51 ns/op with 0 allocs/op**. Projection stayed around **2.6 to 3.0 us/op** with 60 allocs/op.

## Checks

- `go test ./table`
- `go test -race ./table -run 'Test(ToRequestedSchemaWriteFastPath|DefaultDataFileWriterReusesExactBatch|RollingDataWriterReusesExactBatch)$'`\n- `go vet ./table`\n- `go test ./... -run '^$'`

### DIFF
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1457,6 +1457,19 @@ type SchemaOptions struct {
 // ToRequestedSchema will construct a new record batch matching the requested iceberg schema
 // casting columns if necessary as appropriate.
 func ToRequestedSchema(ctx context.Context, requested, fileSchema *iceberg.Schema, batch arrow.RecordBatch, opts SchemaOptions) (arrow.RecordBatch, error) {
+	return toRequestedSchema(ctx, requested, fileSchema, batch, opts, nil)
+}
+
+// toRequestedSchema is the internal write-path variant of ToRequestedSchema.
+// When targetArrowSchema is provided, an exact Arrow schema match can reuse the
+// input batch without walking the Iceberg schema or rebuilding any arrays.
+func toRequestedSchema(ctx context.Context, requested, fileSchema *iceberg.Schema, batch arrow.RecordBatch, opts SchemaOptions, targetArrowSchema *arrow.Schema) (arrow.RecordBatch, error) {
+	if targetArrowSchema != nil && arrowSchemaEqual(batch.Schema(), targetArrowSchema) {
+		batch.Retain()
+
+		return batch, nil
+	}
+
 	st := array.RecordToStructArray(batch)
 	defer st.Release()
 
@@ -1479,6 +1492,12 @@ func ToRequestedSchema(ctx context.Context, requested, fileSchema *iceberg.Schem
 	result.Release()
 
 	return out, nil
+}
+
+func arrowSchemaEqual(left, right *arrow.Schema) bool {
+	// Schema.Equal intentionally ignores top-level metadata, but projection
+	// creates a schema without it, so include it in the no-op check.
+	return left.Equal(right) && left.Metadata().Equal(right.Metadata())
 }
 
 type schemaCompatVisitor struct {
--- a/table/arrow_utils.go
+++ b/table/arrow_utils.go
@@ -1495,8 +1495,12 @@ func toRequestedSchema(ctx context.Context, requested, fileSchema *iceberg.Schem
 }
 
 func arrowSchemaEqual(left, right *arrow.Schema) bool {
-	// Schema.Equal intentionally ignores top-level metadata, but projection
-	// creates a schema without it, so include it in the no-op check.
+	if left == nil || right == nil {
+		return left == right
+	}
+
+	// Schema.Equal ignores top-level metadata. Full projection strips it, so
+	// guard it here; otherwise the fast path would preserve metadata projection drops.
 	return left.Equal(right) && left.Metadata().Equal(right.Metadata())
 }
 
--- a/table/writer_invariants_bench_test.go
+++ b/table/writer_invariants_bench_test.go
@@ -19,6 +19,7 @@ package table
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/apache/arrow-go/v18/arrow"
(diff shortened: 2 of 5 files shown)


table/writer_invariants_bench_test.go
```diff
@@ -90,3 +91,93 @@ func BenchmarkDefaultDataFileWriter(b *testing.B) {
...
+				for b.Loop() {
+					out, err := toRequestedSchema(b.Context(), requested, provided, conversion, opts, requestedArrowSchema)
+					if err != nil {
+						b.Fatal(err)
+					}
+					out.Release()
+				}
+				b.ReportMetric(float64(rows), "rows/op")
+			})
+
```
laskoviymishka: gofumpt is failing on this one blank line; it doesn't allow a trailing blank at the end of a closure body. Deleting it turns CI green.

table/arrow_utils.go
```diff
@@ -1457,6 +1457,19 @@ type SchemaOptions struct {
...
 // casting columns if necessary as appropriate.
 func ToRequestedSchema(ctx context.Context, requested, fileSchema *iceberg.Schema, batch arrow.RecordBatch, opts SchemaOptions) (arrow.RecordBatch, error) {
+	return toRequestedSchema(ctx, requested, fileSchema, batch, opts, nil)
+}
+
+// toRequestedSchema is the internal write-path variant of ToRequestedSchema.
+// When targetArrowSchema is provided, an exact Arrow schema match can reuse the
+// input batch without walking the Iceberg schema or rebuilding any arrays.
+func toRequestedSchema(ctx context.Context, requested, fileSchema *iceberg.Schema, batch arrow.RecordBatch, opts SchemaOptions, targetArrowSchema *arrow.Schema) (arrow.RecordBatch, error) {
+	if targetArrowSchema != nil && arrowSchemaEqual(batch.Schema(), targetArrowSchema) {
```
laskoviymishka, line 1467: The optimization is sound as long as one invariant holds: the precomputed `targetArrowSchema` from `SchemaToArrowSchemaWithOptions` has to be byte-for-byte what `arrowProjectionVisitor` would have produced for a batch that needs no transformation. If those two ever drift (say `SchemaToArrowSchemaWithOptions` starts emitting a looser schema than the visitor), a batch matching the looser schema takes the fast path and skips a transformation it actually needed.

I traced both paths and they agree today, so this isn't a live bug. But nothing enforces the coupling, and the existing tests can't catch drift because they build the input batch from `SchemaToArrowSchemaWithOptions` too, so they're comparing that function against itself.

I'd add one cross-check test that breaks the circularity: push a batch through the slow path (`ToRequestedSchema` with a `nil` target), then assert the returned schema equals `SchemaToArrowSchemaWithOptions(requested, {IncludeFieldIDs: true})` for a non-trivial shape (a struct or list). That locks the equality-check contract to the real projection output. wdyt?

table/writer_schema_projection_test.go
```diff
@@ -0,0 +1,369 @@
...
+			}
+		},
+	}, metaBuilder, schema, 1024*1024)
+	require.NoError(t, err)
+	format := &captureWriteDataFileFormat{FileFormat: tblutils.GetFileFormat(iceberg.ParquetFile)}
+	factory.format = format
+
+	output := make(chan iceberg.DataFile, 1)
+	writer := factory.newRollingDataWriter(t.Context(), "", nil, output)
+	record := projectionTestRecord(t, memory.DefaultAllocator, arrowSchema, []byte(`{"id": 1}`))
```
laskoviymishka: This is the one test exercising the async stream path, the trickiest refcount route, where the fast-path `Retain` has to balance against the stream release, `writeConverted`, and the abort/cleanup defer. And it's the only one on `memory.DefaultAllocator`, which silently swallows an imbalance.

`TestDefaultDataFileWriterReusesExactBatch` already uses a checked allocator; I'd do the same here (`memory.NewCheckedAllocator(memory.NewGoAllocator())` with `defer mem.AssertSize(t, 0)`) so the test actually proves the balance it's meant to.

table/writer_schema_projection_test.go
```diff
@@ -0,0 +1,369 @@
...
+				iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64, Required: true},
+			),
+			batch: func(t *testing.T, mem memory.Allocator) arrow.RecordBatch {
+				schema := projectionTestSchema(t, iceberg.NewSchema(0,
+					iceberg.NestedField{ID: 1, Name: "id", Type: iceberg.PrimitiveTypes.Int64, Required: true},
+				))
+
+				return projectionTestRecord(t, mem, schema, []byte(`{"id": 1}`))
+			},
+			wantReuse: true,
```
laskoviymishka, line 105: This single-column Int64 case is the only one that actually takes the fast path; every other case is a rejection. If `arrowSchemaEqual` had a latent bug on multi-field or nested schemas we'd never see it here.

I'd add a multi-column exact-match and one nested (struct or list) exact-match to the `wantReuse: true` set. And a negative case worth having explicitly: a `Required: true` field where the incoming batch carries `Nullable: true`, which has to reject the fast path so full projection emits `Nullable: false` (required, not optional, in the written Parquet). Right now that rejection is untested.

table/writer_schema_projection_test.go
```diff
@@ -0,0 +1,369 @@
...
+	require.NoError(t, err)
+
+	format := &captureWriteDataFileFormat{FileFormat: tblutils.GetFileFormat(iceberg.ParquetFile)}
+	writer, err := newDataFileWriter(t.TempDir(), iceio.LocalFS{}, metaBuilder, iceberg.Properties{}, withFormat(format))
+	require.NoError(t, err)
+
+	arrowSchema := projectionTestSchema(t, schema)
+	record := projectionTestRecord(t, mem, arrowSchema, []byte(`{"id": 1}`))
+	defer record.Release()
+	record.Retain()
```
laskoviymishka, line 419: A one-line comment here would help. This `Retain` counterbalances the releases inside `writeFile` (the per-batch `rec.Release()` plus the `task.Batches` cleanup defer), and on the fast path `rec` is this same `record`. Without a note, someone tidying the test could delete it and get a double-release panic.

table/arrow_utils.go
```diff
@@ -1481,6 +1494,12 @@ func ToRequestedSchema(ctx context.Context, requested, fileSchema *iceberg.Schem
 	return out, nil
 }
 
+func arrowSchemaEqual(left, right *arrow.Schema) bool {
+	// Schema.Equal intentionally ignores top-level metadata, but projection
```
laskoviymishka: The comment has the direction backwards: it reads as if projection is what lacks the metadata, but the point is the other way. The input batch might carry top-level metadata that full projection would strip (`RecordFromStructArray` carries none), so we guard here to stop the fast path silently preserving it. Something like:

```go
// Schema.Equal ignores top-level metadata. Full projection strips it, so
// guard it here; otherwise the fast path would preserve metadata that
// projection would have dropped.
```

While we're here, both params get dereferenced immediately, so a one-line nil precondition (or guard) wouldn't hurt. It's unexported and every call site passes non-nil today, so this one's minor.

laskoviymishka (changes requested):
Nice optimization. Skipping the whole projection walk and array rebuild when the batch already matches the target schema is a clean win, and the equality check is thorough (types, nullability, field IDs, and top-level metadata are all covered).

I'd hold this before merging, though.

The bypass is only safe as long as the precomputed `targetArrowSchema` from `SchemaToArrowSchemaWithOptions` is byte-for-byte what the projection visitor would build for a no-op batch. I traced it and they agree today, so this isn't a live bug, but nothing enforces the coupling, and the current tests can't catch drift because they build their input batch from that same function, so they end up comparing it against itself. If the two ever diverge, a batch matching the looser schema would take the fast path and skip a transformation it actually needed.

Things I'd want before merge:

- A cross-check test that breaks the circularity: push a batch through the slow path and assert the result schema equals `SchemaToArrowSchemaWithOptions(requested, ...)` for a non-trivial shape (struct or list).
- A checked allocator on `TestRollingDataWriterReusesExactBatch`. It's the one test on the async stream path, exactly where a `Retain`/`Release` imbalance would hide, and it's currently on `DefaultAllocator`.
- A couple more `wantReuse: true` shapes (multi-column, nested), plus a negative case where a `Required` field arrives `Nullable: true`, so the required-vs-optional rejection is actually exercised.
- The gofumpt blank line in the benchmark (CI is red on it).

laskoviymishka (approved):

PR #1913, opened 2026-08-27
perf(table): classify manifest entries outside merge lock

fallintoplace:
## What changed

- **Classify manifest entries before taking `manifestEntries.mu`.**
- Keep only the bulk appends to the shared slices inside the lock.
- Reuse homogeneous batches and use exact-size buckets for mixed batches.
- Preserve the existing error behavior for unknown file content.

## Why

Large manifest batches were doing `DataFile()` and content checks while holding the shared mutex. This made concurrent manifest collection wait on work that does not touch shared state.

## Benchmark

Command:

```text
go test ./table -run '^$' -bench '^BenchmarkManifestEntryCollection/content=(data|deletes)/manifests=(8|64)/concurrency=(1|4|16)/entries=10000$' -benchtime=1s -count=3
```

Apple M1 Pro, Go 1.26.3. Median of 3 runs, 8 manifests with 10,000 entries each:

| Workload | Concurrency | Before | After |
| --- | ---: | ---: | ---: |
| Data | 4 | 1.952 ms/op | 1.403 ms/op |
| Data | 16 | 1.959 ms/op | 1.357 ms/op |
| Mixed deletes | 4 | 1.810 ms/op | 1.636 ms/op |
| Mixed deletes | 16 | 1.924 ms/op | 1.772 ms/op |

The 64-manifest cases stayed roughly flat at concurrency 16 while the lock covered less work. Data batches also dropped from 54 to 34 allocs/op in the 8-manifest cases.

## Tests

- `go test ./table -count=1`
- `go test -race ./table -run 'TestManifestEntries|TestIsDeletionVector' -count=1`
- `go vet ./table`

### DIFF
--- a/table/dv_scan_planning_test.go
+++ b/table/dv_scan_planning_test.go
@@ -141,6 +141,29 @@ func TestManifestEntries_DVClassification(t *testing.T) {
 	assert.Len(t, entries.equalityDeleteEntries, 1)
 }
 
+func TestManifestEntries_MergeRejectsUnknownContent(t *testing.T) {
+	snapshotID := int64(1)
+	validEntry := iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+		path:        "s3://bucket/data/data-001.parquet",
+		contentType: iceberg.EntryContentData,
+	})
+	invalidEntry := iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+		path:        "s3://bucket/data/unknown.bin",
+		contentType: iceberg.ManifestEntryContent(99),
+	})
+	entryAfterInvalid := iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+		path:        "s3://bucket/data/data-002.parquet",
+		contentType: iceberg.EntryContentData,
+	})
+
+	entries := newManifestEntries()
+	err := entries.merge([]iceberg.ManifestEntry{validEntry, invalidEntry, entryAfterInvalid})
+
+	assert.ErrorIs(t, err, ErrInvalidMetadata)
+	assert.ErrorContains(t, err, "unknown DataFileContent type")
+	assert.Equal(t, []iceberg.ManifestEntry{validEntry}, entries.dataEntries)
+}
+
 func TestManifestEntries_ConcurrentMerge(t *testing.T) {
 	snapshotID := int64(1)
 	batch := []iceberg.ManifestEntry{
--- a/table/dv_scan_planning_test.go
+++ b/table/dv_scan_planning_test.go
@@ -164,6 +164,41 @@ func TestManifestEntries_MergeRejectsUnknownContent(t *testing.T) {
 	assert.Equal(t, []iceberg.ManifestEntry{validEntry}, entries.dataEntries)
 }
 
+func TestManifestEntries_MergeRejectsUnknownContentAfterMixedEntries(t *testing.T) {
+	snapshotID := int64(1)
+	dataEntry := iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+		path:        "s3://bucket/data/data-001.parquet",
+		contentType: iceberg.EntryContentData,
+	})
+	positionalDeleteEntry := iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+		path:        "s3://bucket/data/pos-delete-001.parquet",
+		contentType: iceberg.EntryContentPosDeletes,
+	})
+	invalidEntry := iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+		path:        "s3://bucket/data/unknown.bin",
+		contentType: iceberg.ManifestEntryContent(99),
+	})
+	entryAfterInvalid := iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+		path:        "s3://bucket/data/data-002.parquet",
+		contentType: iceberg.EntryContentData,
+	})
+
+	entries := newManifestEntries()
+	err := entries.merge([]iceberg.ManifestEntry{
+		dataEntry,
+		positionalDeleteEntry,
+		invalidEntry,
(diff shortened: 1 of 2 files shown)


table/dv_scan_planning_test.go
```diff
@@ -141,6 +141,29 @@ func TestManifestEntries_DVClassification(t *testing.T) {
...
+		path:        "s3://bucket/data/unknown.bin",
+		contentType: iceberg.ManifestEntryContent(99),
+	})
+	entryAfterInvalid := iceberg.NewManifestEntry(iceberg.EntryStatusADDED, &snapshotID, nil, nil, &mockDataFile{
+		path:        "s3://bucket/data/data-002.parquet",
+		contentType: iceberg.EntryContentData,
+	})
+
+	entries := newManifestEntries()
+	err := entries.merge([]iceberg.ManifestEntry{validEntry, invalidEntry, entryAfterInvalid})
```
laskoviymishka, line 160: This case only exercises the `kinds == nil` path, since every valid entry here is data, so classification stays on the fast `classifiedManifestEntriesForKind` branch and the `kinds != nil` backfill loop never runs.

I'd add a mixed case like `[data, posDelete, invalid, data]` so the error hits with `kinds != nil` and the backfill path gets covered, asserting `dataEntries == [e0]` and `positionalDeleteEntries == [e1]`. While we're here, worth asserting the other three buckets stay empty too, so a bad entry bleeding into the wrong bucket can't pass silently.

table/dv_scan_planning_test.go
```diff
@@ -141,6 +141,29 @@ func TestManifestEntries_DVClassification(t *testing.T) {
...
+
+	assert.ErrorIs(t, err, ErrInvalidMetadata)
+	assert.ErrorContains(t, err, "unknown DataFileContent type")
+	assert.Equal(t, []iceberg.ManifestEntry{validEntry}, entries.dataEntries)
```
laskoviymishka, line 164: One thing worth calling out: this pins the partial-commit-on-error behavior as expected, but the real caller (`collectManifestEntriesWithSchema`) returns on the first error and discards these entries, so nothing actually depends on the partial result landing.

Not asking to change the behavior, just maybe a comment here or in `classifyManifestEntries` noting the partial commit is incidental, so this test doesn't get read as a hard requirement to preserve. wdyt?

table/scanner.go
```diff
@@ -147,30 +164,128 @@ func newManifestEntries() *manifestEntries {
...
+
+func newClassifiedManifestEntries(counts [manifestEntryKindCount]int) classifiedManifestEntries {
+	return classifiedManifestEntries{
+		dataEntries:             make([]iceberg.ManifestEntry, 0, counts[manifestEntryData]),
+		positionalDeleteEntries: make([]iceberg.ManifestEntry, 0, counts[manifestEntryPositionalDelete]),
+		equalityDeleteEntries:   make([]iceberg.ManifestEntry, 0, counts[manifestEntryEqualityDelete]),
+		dvEntries:               make([]iceberg.ManifestEntry, 0, counts[manifestEntryDV]),
+	}
+}
+
```
laskoviymishka, line 194: Minor: the homogeneous path assigns the caller's slice straight into the result (`classified.dataEntries = entries`) with no copy, so the returned struct aliases the `openManifest` slice that `merge` was handed. It's safe today since `merge` only reads it through `append`, but the aliasing is intentional and undocumented.

A one-line comment would save the next person who tries to mutate one of these buckets in place from a surprise.

table/scanner.go
```diff
@@ -147,30 +164,128 @@ func newManifestEntries() *manifestEntries {
...
+	}
+
+	classified := newClassifiedManifestEntries(counts)
+	for i, entry := range entries {
+		appendClassifiedManifestEntry(&classified, kinds[i], entry)
+	}
+
+	return classified, nil
+}
+
```
laskoviymishka, line 266: Neither this switch nor the one in `classifiedManifestEntriesForKind` has a default, so if a fifth `manifestEntryKind` ever gets added and someone forgets to wire it here, the entry is silently dropped with no error. `exhaustive` isn't enabled, so nothing catches that statically, and the non-data fast paths aren't covered by a test either, so a drop would be invisible.

Not blocking, but I'd add a `default: panic(fmt.Sprintf("unhandled manifestEntryKind %d", kind))` to both switches so a future miss fails loudly. wdyt?

table/scanner.go
```diff
@@ -147,30 +164,128 @@ func newManifestEntries() *manifestEntries {
...
+		classified.positionalDeleteEntries = append(classified.positionalDeleteEntries, entry)
+	case manifestEntryEqualityDelete:
+		classified.equalityDeleteEntries = append(classified.equalityDeleteEntries, entry)
+	case manifestEntryDV:
+		classified.dvEntries = append(classified.dvEntries, entry)
+	}
+}
+
+func (m *manifestEntries) merge(entries []iceberg.ManifestEntry) error {
+	classified, err := classifyManifestEntries(entries)
```
laskoviymishka, line 283: I'd restore `defer m.mu.Unlock()` here rather than unlocking explicitly at the end.

Realistically these four appends won't panic, so this isn't a live deadlock, but it's the one spot that dropped the defer idiom every other method on `m.mu` still uses, and if any append ever panics (OOM, or a future change) the mutex stays held and every subsequent `merge` deadlocks. It's free to keep the guarantee: `m.mu.Lock()` then `defer m.mu.Unlock()` right at the top.

laskoviymishka (changes requested):
This is a good change. Pulling the classification out of the critical section is exactly right, since the switch plus `IsDeletionVector` work is pure and read-only, so serializing every manifest goroutine on it was wasted contention. The zero-copy homogeneous fast path is a nice touch too.

I'd hold this for one thing before merge: the new `merge` dropped the `defer m.mu.Unlock()` and unlocks explicitly at the end instead. Realistically the four appends won't panic, so it's not a live deadlock, but it's the only method on `m.mu` that abandoned the defer idiom, and if any append ever panics the mutex stays held and every later `merge` deadlocks. It's free to keep the guarantee, so I'd just restore `defer m.mu.Unlock()` right after the `Lock()`.

The other thing I'd want is a test for the mixed-batch error path. The new test only covers the case where every valid entry shares one kind, so `kinds` stays nil and the backfill loop never runs. A case like `[data, posDelete, invalid, data]` would exercise the `kinds != nil` path that's currently untested.

The rest is non-blocking and noted inline: the two switches have no default so a future kind would drop silently, the slice aliasing could use a one-line comment, and the new test pins partial-commit-on-error as a contract when the real caller discards it.

Restore the defer and add the mixed-batch case and I'm happy to take another pass and approve.

laskoviymishka (approved):

PR #1914, opened 2026-08-27
perf(table): avoid per-file partition materialization in partitions metadata

fallintoplace:
## What changed

- **`Partitions()` now uses the borrowed partition map for aggregate-tree lookups.**
- **The coerced map and positional record are built only when a partition is first seen.**
- **Binary values are copied before they are retained by an aggregate.**
- Added coverage for binary, NaN, and binary ownership behavior.
- Added a before/after benchmark for repeated partitions.

## Why

`InspectTable.Partitions` was materializing several partition representations for every manifest entry. Most files share an existing partition, so that work was repeated and then discarded.

The new path keeps the existing partition evolution behavior. It also keeps the public `DataFile.Partition()` fallback for custom data file implementations.

## Benchmark

- **Command:** `go test ./table -run=^\\$ -bench=BenchmarkInspectPartitionAggregation -benchmem -benchtime=200ms -count=5`
- **Machine:** Apple M1 Pro, darwin/arm64
- **Scope:** aggregation loop only. Manifest decoding and Arrow output are outside the timed section.
- **Median of five samples:**
  - **10,000 files / 100 partitions / 1 int32 field:** `2.01 ms, 2.78 MB, 30.8k allocs` -> `0.395 ms, 59.0 KB, 911 allocs` (**80% lower time, 98% fewer bytes**)
  - **100,000 files / 100 partitions / 8 int32 fields:** `53.5 ms, 38.7 MB, 304k allocs` -> `25.8 ms, 295 KB, 4.11k allocs` (**52% lower time, 99% fewer bytes**)
  - **100,000 files / 100 partitions / 32 binary fields:** `610 ms, 732 MB, 13.6M allocs` -> `182 ms, 2.89 MB, 33.8k allocs` (**70% lower time, 99.6% fewer bytes**)

## Tests

- `go test ./... -count=1`
- `go vet ./table`

### DIFF
--- a/table/inspect_internal_test.go
+++ b/table/inspect_internal_test.go
@@ -2654,6 +2654,19 @@ func TestInspectPartitionAggregateTreeLookupPartition(t *testing.T) {
 		1001: []byte{1, 2, 4},
 		1002: math.NaN(),
 	}, partitionType))
+
+	// A file written with an older partition spec does not have the field
+	// added by a later spec. Both lookup paths must key the missing field as nil.
+	evolvedPartition := map[int]any{
+		1000: int32(7),
+		1002: math.NaN(),
+	}
+	evolvedRecord := newPartitionRecord(inspectCoercePartition(evolvedPartition, partitionType), partitionType)
+	evolvedAggregate := &inspectPartitionAggregate{specID: 2}
+	tree.insert(evolvedRecord, evolvedAggregate)
+
+	require.Same(t, evolvedAggregate, tree.lookupPartition(evolvedPartition, partitionType))
+	require.Same(t, evolvedAggregate, tree.lookup(evolvedRecord))
 }
 
 func TestCloneInspectPartitionClonesBinaryValues(t *testing.T) {
--- a/table/inspect_internal_test.go
+++ b/table/inspect_internal_test.go
@@ -2632,6 +2632,43 @@ func TestInspectPartitionAggregateTreeDistinguishesFloatWidthsAndSignedZero(t *t
 	require.Nil(t, tree.lookup(partitionRecord{float64(0)}))
 }
 
+func TestInspectPartitionAggregateTreeLookupPartition(t *testing.T) {
+	partitionType := &iceberg.StructType{FieldList: []iceberg.NestedField{
+		{ID: 1000, Name: "id", Type: iceberg.PrimitiveTypes.Int32},
+		{ID: 1001, Name: "payload", Type: iceberg.PrimitiveTypes.Binary},
+		{ID: 1002, Name: "score", Type: iceberg.PrimitiveTypes.Float64},
+	}}
+	partition := map[int]any{
+		1000: int32(7),
+		1001: []byte{1, 2, 3},
+		1002: math.NaN(),
+	}
+	record := newPartitionRecord(inspectCoercePartition(partition, partitionType), partitionType)
+	aggregate := &inspectPartitionAggregate{specID: 1}
+	tree := newInspectPartitionAggregateTree()
+	tree.insert(record, aggregate)
+
+	require.Same(t, aggregate, tree.lookupPartition(partition, partitionType))
+	require.Nil(t, tree.lookupPartition(map[int]any{
+		1000: int32(7),
+		1001: []byte{1, 2, 4},
+		1002: math.NaN(),
+	}, partitionType))
+}
+
+func TestCloneInspectPartitionClonesBinaryValues(t *testing.T) {
+	value := []byte{1, 2, 3}
+	cloned := cloneInspectPartition(map[int]any{1000: value})
+
+	value[0] = 4
+	require.Equal(t, []byte{1, 2, 3}, cloned[1000])
+
+	clonedValue := cloned[1000].([]byte)
+	clonedValue[1] = 5
+
+	require.Equal(t, []byte{4, 2, 3}, value)
+}
+
 func TestInspectPartitionsAggregatesDataAndDeletes(t *testing.T) {
 	const snapshotID = int64(1)
 	spec := partitionedSpec()
--- a/table/inspect_partitions.go
+++ b/table/inspect_partitions.go
@@ -43,7 +43,6 @@ type inspectPartitionAggregate struct {
 	equalityDeleteFiles int32
 	lastUpdatedAt       *int64
 	lastUpdatedSnapshot *int64
-	partitionRecord     partitionRecord
 	orderingKey         string
 }
 
@@ -76,10 +75,6 @@ func (t *inspectPartitionAggregateTree) lookupPartition(
 	partitionType *iceberg.StructType,
 ) *inspectPartitionAggregate {
 	node := t
-	if partitionType == nil {
-		return node.aggregate
-	}
-
 	for _, field := range partitionType.FieldList {
 		child, ok := node.children[comparablePartitionKey(partition[field.ID])]
 		if !ok {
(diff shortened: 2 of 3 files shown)


table/inspect_partitions.go
```diff
@@ -152,10 +173,11 @@ func (i InspectTable) partitionAggregates(ctx context.Context, partitionType *ic
...
-			record := newPartitionRecord(partition, partitionType)
-			aggregate := aggregateTree.lookup(record)
+			partitionValues := dataFilePartition(file)
+			aggregate := aggregateTree.lookupPartition(partitionValues, partitionType)
 			if aggregate == nil {
+				partition := inspectCoercePartition(partitionValues, partitionType)
+				record := newPartitionRecord(partition, partitionType)
 				aggregate = &inspectPartitionAggregate{
 					partition:       cloneInspectPartition(partition),
 					partitionRecord: record,
```
laskoviymishka: `partitionRecord` here still holds the borrowed `[]byte` slice headers from `dataFilePartition(file)`. `inspectCoercePartition` only shallow-copies the map, so those byte backing arrays stay aliased to the DataFile's Avro decode buffer and now escape across the whole aggregation.

The saving grace is that nothing reads `partitionRecord` anymore. `appendPartitionAggregate` goes through `aggregate.partition`, which `cloneInspectPartition` deep-copies (bytes included, now), so this field is dead weight that also pins memory and quietly re-opens the borrow-contract violation the rest of the PR is careful about.

I'd just drop the field from `inspectPartitionAggregate` and this struct literal. The trie key already owns its copy via `comparablePartitionKey`, and `lookupPartition` never touches it.

table/inspect_internal_test.go
```diff
@@ -2617,6 +2617,43 @@ func TestInspectPartitionAggregateTreeHandlesBinaryAndNaNValues(t *testing.T) {
...
+		1000: int32(7),
+		1001: []byte{1, 2, 3},
+		1002: math.NaN(),
+	}
+	record := newPartitionRecord(inspectCoercePartition(partition, partitionType), partitionType)
+	aggregate := &inspectPartitionAggregate{specID: 1}
+	tree := newInspectPartitionAggregateTree()
+	tree.insert(record, aggregate)
+
+	require.Same(t, aggregate, tree.lookupPartition(partition, partitionType))
```
laskoviymishka, line 2651: This is the equivalence the PR is really leaning on, that `lookupPartition` matches the old coerce+record+lookup path, and the partition-evolution case is the one I'd most want pinned down but it isn't covered here.

Could we add a sub-case where a file's partition map is missing one of the fields in `partitionType` (older spec, field added later)? `partition[field.ID]` returns nil and `inspectCoercePartition` skips it too, so both paths key on nil at that level and agree. That's exactly the kind of quiet invariant a future change to `inspectCoercePartition` could break without anything failing.

While we're here, the `partitionType == nil` early-return branch is also untested. wdyt?

table/inspect_partitions.go
```diff
@@ -70,6 +71,26 @@ func (t *inspectPartitionAggregateTree) lookup(record partitionRecord) *inspectP
 	return node.aggregate
 }
 
+func (t *inspectPartitionAggregateTree) lookupPartition(
+	partition map[int]any,
+	partitionType *iceberg.StructType,
+) *inspectPartitionAggregate {
+	node := t
+	if partitionType == nil {
```
laskoviymishka: This guard makes `lookupPartition` look like it handles a nil `partitionType`, but the aggregate-creation path after a nil lookup still calls `newPartitionRecord(nil, nil)` and panics dereferencing the nil `FieldList`. Same as before this PR, so not a regression, just newly misleading.

I'd either mirror the nil check where the aggregate gets built, or drop a one-line comment that nil is unreachable here and the guard is purely defensive. wdyt?

table/inspect_partitions.go
```diff
@@ -70,6 +71,26 @@ func (t *inspectPartitionAggregateTree) lookup(record partitionRecord) *inspectP
...
+		return node.aggregate
+	}
+
+	for _, field := range partitionType.FieldList {
+		child, ok := node.children[comparablePartitionKey(partition[field.ID])]
```
laskoviymishka, line 79: `insert` keys the trie positionally off the `partitionRecord` slice, while `lookupPartition` keys off `partition[field.ID]` walked in `FieldList` order. They only line up because `record` is always built from this same `partitionType.FieldList`, and that coupling is load-bearing now that `lookupPartition` is the only production reader.

I'd add a short comment on `insert` (or here) noting the two have to walk fields in the same order. `lookup` also only has test callers now, so it's worth deciding whether to keep it or move those tests onto `lookupPartition`.

table/inspect_partitions_bench_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/apache/iceberg-go"
+)
+
+var inspectPartitionAggregationBenchmarkSink int
+
+func BenchmarkInspectPartitionAggregation(b *testing.B) {
+	for _, benchmark := range []struct {
```
laskoviymishka: This struct is spelled out again in `benchmarkInspectPartitionFiles`'s signature, so adding a field later means matching the shape and order in both places. Could we lift it to a named type at package scope and use it in both spots?

table/inspect_partitions_bench_test.go
```diff
@@ -0,0 +1,160 @@
...
+func benchmarkInspectPartitionAggregation(
+	b *testing.B,
+	partitionType *iceberg.StructType,
+	files []iceberg.DataFile,
+	materialize bool,
+) {
+	b.Helper()
+	b.ReportAllocs()
+	b.ReportMetric(float64(len(files)), "files/op")
+	b.ResetTimer()
```
laskoviymishka: `b.Loop()` manages the timer itself (it resets and starts on the first call), and there's nothing expensive between `ReportMetric` and the loop, so this `ResetTimer` is redundant. I'd drop it, or add a comment if it's there for older Go.

laskoviymishka (changes requested):
Nice optimization, and the direction is right. Deferring per-file materialization and keying the trie off the borrowed partition map is a real win on the hot path, and the output stays equivalent to the old code.

The main thing is the `partitionRecord` field on `inspectPartitionAggregate`. Nothing reads it anymore (the output path goes through `aggregate.partition`, which is deep-cloned), but it still retains the borrowed `[]byte` slice headers from `dataFilePartition`, so it pins the DataFile's Avro decode buffers for the whole aggregation and quietly re-opens the borrow-contract violation the rest of the PR takes care to respect. I'd just drop the field.

The other thing is coverage. The partition-evolution case (a file from an older spec missing a field that was added later) is the exact equivalence this PR claims to preserve, and it's the one case the new test doesn't hit. I'd want a sub-case for it before merge.

- drop the dead `partitionRecord` field from `inspectPartitionAggregate`
- add a `lookupPartition` test for the missing-field / partition-evolution case
- a short comment tying `insert`'s positional keying to `lookupPartition`'s field-ID keying (they only agree by `FieldList` ordering)

The rest (the misleading nil-`partitionType` guard, a couple of benchmark tidy-ups) are minor and inline. Once the field's gone and the evolution case is covered, happy to take another pass and approve.

zeroshade (commented):
The updated code is clean from my re-review, and the earlier requested changes appear addressed:

- the dead `partitionRecord` retention is gone;
- retained binary partition values are cloned;
- the partition-evolution/missing-field case is covered;
- the insert/lookup field-order coupling is documented;
- the benchmark cleanup is complete.

I also verified repeated race tests, the full `table` package, vet, all benchmark paths, and latest-`main` integration.

Two addressed review threads remain unresolved, and the earlier change-request review still blocks the PR. Please have those threads resolved and the prior reviewer update their disposition before merge.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Iceberg Go maintainer. After the outstanding review disposition is updated, an Iceberg Go maintainer — a real person — will take the next look at the PR.*
>

laskoviymishka (approved):
LGTM now!

PR #1915, opened 2026-08-27
perf(table): borrow partitions in positional delete index

fallintoplace:
## What changed

- **Use `dataFilePartition`** when building the positional-delete index.
- **Use the borrowed view for data-file lookups** when matching partition-scoped deletes.
- **Keep the public fallback** for other `DataFile` implementations.
- **Add regression coverage** for the borrowed accessor.
- **Add a before/after benchmark** with 10,000 files and wide partitions.

## Benchmark

Command:

`go test ./table -bench=BenchmarkPositionalDeletePartitionKey -benchmem -benchtime=200ms -count=5`

Median of 5 runs:

| case | before | after | before B/op | after B/op | before allocs/op | after allocs/op |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 int32 field | 2.54 ms | 1.42 ms | 2.96 MB | 0.40 MB | 50,000 | 30,000 |
| 8 int32 fields | 5.99 ms | 4.37 ms | 6.62 MB | 4.06 MB | 90,000 | 70,000 |
| 32 binary fields | 62.45 ms | 37.25 ms | 90.88 MB | 59.28 MB | 790,000 | 110,000 |

## Tests

- `go test ./table -count=1`
- `go vet ./table`
- `go test ./... -count=1`

### DIFF
--- a/table/positional_delete_index.go
+++ b/table/positional_delete_index.go
@@ -47,7 +47,7 @@ func buildPositionalDeleteIndex(entries []iceberg.ManifestEntry) (*positionalDel
 			continue
 		}
 
-		partitionKey, err := canonicalPartitionKey(deleteFile.SpecID(), deleteFile.Partition())
+		partitionKey, err := canonicalPartitionKey(deleteFile.SpecID(), dataFilePartition(deleteFile))
 		if err != nil {
 			return nil, fmt.Errorf("indexing positional delete file %s: %w", deleteFile.FilePath(), err)
 		}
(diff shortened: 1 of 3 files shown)


laskoviymishka (commented):
Nice optimization!

laskoviymishka (approved):

zeroshade (approved):
Looks good. Both positional-delete index paths consume the borrowed partition map synchronously to construct an owned canonical key, without mutating or retaining borrowed metadata. Custom `DataFile` implementations continue through the public getter fallback.

I verified path/partition matching, sequence behavior, binary partition ownership, repeated race tests, the full `table` package, latest-main integration, and the benchmarked allocation reduction.

---

>

PR #1916, opened 2026-08-27
perf(table): avoid cloning schema history for snapshot scans

fallintoplace:
## What changed

- **Resolve snapshot schemas directly** for built-in table metadata.
- **Clone only the selected schema** instead of cloning the whole schema history.
- **Keep the public `Schemas()` fallback** for custom `Metadata` implementations.
- **Avoid the current-schema clone** when a historical snapshot already has a schema ID.
- Add regression coverage and a before/after benchmark.

## Why

`Schemas()` returns defensive copies of every schema. A historical scan only needs the schema referenced by its snapshot.

## Benchmark

Command:

`go test ./table -run ^$ -bench BenchmarkEffectiveSchemaSnapshot -benchmem -benchtime=200ms -count=5`

Median of 5 runs. Values are ns/op, B/op, allocs/op:

| schemas / fields | before | after |
| --- | ---: | ---: |
| 1 / 16 | 4,691 / 10,704 / 31 | 2,363 / 5,408 / 16 |
| 16 / 16 | 41,198 / 90,144 / 241 | 3,151 / 5,408 / 16 |
| 128 / 128 | 2,077,597 / 5,253,126 / 2,583 | 16,991 / 40,832 / 22 |
| 1,024 / 128 | 13,907,661 / 41,739,414 / 20,503 | 14,323 / 40,832 / 22 |

## Tests

- `go test ./table -count=1`
- `go vet ./...`
- `go test ./... -count=1`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2060,12 +2060,25 @@ func (c *commonMetadata) LastUpdatedMillis() int64   { return c.LastUpdatedMS }
 func (c *commonMetadata) LastColumnID() int          { return c.LastColumnId }
 func (c *commonMetadata) Schemas() []*iceberg.Schema { return cloneSchemas(c.SchemaList) }
 func (c *commonMetadata) CurrentSchema() *iceberg.Schema {
-	for _, s := range c.SchemaList {
-		if s.ID == c.CurrentSchemaID {
-			return cloneSchema(s)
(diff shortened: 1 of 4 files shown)


laskoviymishka (approved):
🚢

PR #1918, opened 2026-08-27
perf(table): skip inactive parquet stats columns

fallintoplace:
## What

- Skip `ColumnChunk` construction for variant-child, lineage, and `none` metrics columns.
- Keep the existing stats behavior for active columns.

## Why

`DataFileStatsFromMeta` visits every physical column in every row group. Inactive columns only added metadata wrapper work and allocations.

## Benchmark

Mostly inactive 32-row-group, 9-column case:

- 1,404 -> 636 allocs/op
- 202 KB -> 114 KB/op

## Checks

- `go test ./...`
- `go test -race ./table/internal ./table`

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -1483,13 +1483,13 @@ func (p parquetFormat) DataFileStatsFromMeta(meta Metadata, statsCols map[int]St
 			}
 
 			fieldID, statsCol := column.fieldID, column.statsCol
-			if _, invalid := invalidateCol[fieldID]; invalid {
-				continue
-			}
-
 			colChunk, err = rowGroup.ColumnChunk(pos)
 			if err != nil {
 				panic(err)
+			}
+
+			if _, invalid := invalidateCol[fieldID]; invalid {
+				continue
 			}
 
 			colSizes[fieldID] += colChunk.TotalCompressedSize()
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -1470,25 +1470,26 @@ func (p parquetFormat) DataFileStatsFromMeta(meta Metadata, statsCols map[int]St
 		}
 
 		for pos := range rowGroup.NumColumns() {
-			colChunk, err = rowGroup.ColumnChunk(pos)
-			if err != nil {
-				panic(err)
-			}
-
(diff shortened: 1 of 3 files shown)


tanmayrauth (approved):
LGTM. Nice, tight optimization — the reordering is output-identical: the set of columns that reach the colSizes/valueCounts accumulation is unchanged, splitOffsets still keys off ColumnChunk(0), and you preserved the resolveErr panic path (including fetching the chunk first so a doubly-broken column reports the ame error it did before). The none/skip/variant exclusion is already covered end-to-end by TestMetricsModeNone and TestColumnMetricsMode, both of which run through DataFileStatsFromMeta, so the perf win doesn't cost coverage.

The ColumnChunk(pos) fetch inside the resolveErr branch (parquet_files.go:1475) only exists to preserve which of two fatal panics fires. On a column that's both missing from the metrics plan and has undecodable chunk metadata, it surfaces the low-level parquet error instead of the more actionable "field id N not found in the metrics plan". Since both paths panic, you could just panic(column.resolveErr) directly — but it's behavior-preserving as written and this is an essentially unreachable path.

table/internal/parquet_files.go
```diff
@@ -1470,27 +1470,28 @@ func (p parquetFormat) DataFileStatsFromMeta(meta Metadata, statsCols map[int]St
...
-				continue
-			}
 			if _, invalid := invalidateCol[fieldID]; invalid {
 				continue
 			}
 
+			colChunk, err = rowGroup.ColumnChunk(pos)
+			if err != nil {
+				panic(err)
+			}
```
ferhatelmas, line 1489: I think this should be before `invalidateCol` check because otherwise there is a regression in metadata error reporting. 

For example, an active field is using full metrics and `StatsSet` errors below for row group 1, then it will be added into `invalidateCol` in line 1503 and metadata failure will be skipped for row group 2 due to early continue in line 1487

fallintoplace: Thanks, I think you are right, I will push the fix.

ferhatelmas: LGTM

ferhatelmas (approved):

laskoviymishka (approved):
LGTM

PR #1920, opened 2026-08-27
perf(table): avoid cloning schema history for position_deletes partition IDs

fallintoplace:
## What changed

- Start synthetic position-delete partition field IDs at `LastColumnID() + 1`.
- Keep the existing reserved-ID checks and partition-field remapping.
- Avoid cloning and indexing every historical schema on each call.

## Benchmark

- **Machine:** Apple M1 Pro, darwin/arm64
- **Go:** go1.26.3
- **Setup:** 64 fields per schema; median of 5 runs
- **Command:** `go test ./table -run '^$' -bench '^BenchmarkPositionDeletesPartitionType/schemas=(1|16|128|1024)/fields=64$' -benchmem -count=5 -benchtime=1s`

| schemas | ns/op | B/op | allocs/op |
| --- | ---: | ---: | ---: |
| 1 | 40,476 -> 18,736 | 100,368 -> 48,960 | 94 -> 56 |
| 16 | 347,383 -> 18,448 | 879,377 -> 48,960 | 570 -> 56 |
| 128 | 2,695,818 -> 18,115 | 6,692,085 -> 48,960 | 4,101 -> 56 |
| 1,024 | 20,795,189 -> 16,137 | 53,193,992 -> 48,960 | 32,328 -> 56 |

The optimized path is effectively flat as schema history grows.

## Tests

- Added coverage that the allocator does not request schema history.
- Added coverage for skipping reserved IDs after `LastColumnID()`.

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -346,8 +346,7 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 
 	type Alias Schema
 
-	aliasCopy := *(*Alias)(s)
-	aliasCopy.IdentifierFieldIDs = ids
+	aliasCopy := Alias{ID: s.ID, IdentifierFieldIDs: ids}
 
 	return json.Marshal(struct {
 		Type   string        `json:"type"`
--- a/schema_test.go
+++ b/schema_test.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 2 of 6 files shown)


zeroshade (changes requested):
The history-independent allocation is a substantial improvement, but the new starting-point calculation can produce duplicate or out-of-range field IDs.

### Blocking — Guard synthetic field-ID allocation (`table/inspect_position_deletes.go:817–831`)

`metadata.LastColumnID()+1` is used without validating it against the current schema or the signed 32-bit field-ID boundary.

I reproduced two regressions:

1. The metadata parser accepts a schema field with ID `2` and `last-column-id: 1`. This change assigns the partition field ID `2`, after which `PositionDeletesSchema` panics:

   ```text
   invalid schema: multiple fields for id 2: x and x
   ```

   The merge-base implementation assigns an unused ID and succeeds.

2. With `LastColumnID() == iceberg.MaxStructFieldID` and 193 active partition fields, the final generated ID is `2147483648`, outside Iceberg’s signed 32-bit field-ID range. This also wraps on 32-bit platforms.

Please keep the fast path but seed collision checks from the current schema and make allocation overflow-safe. If the range above `LastColumnID` is exhausted, fall back to low unused IDs or return a structured error. Regression coverage should include both cases above.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Iceberg Go maintainer. After you've addressed the points above and pushed an update, an Iceberg Go maintainer — a real person — will take the next look at the PR. The findings cite the project's review criteria; if you think one of them is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

zeroshade (approved):
The synthetic field-ID allocator now handles the collision and overflow cases I raised, and dropping the schema-history clone turns out not to expose any aliased mutable state. Approving.

For the record, my earlier review was on `29e119b`, which has since been force-pushed away — but the current head fixes the finding independently, so there's nothing left outstanding from it. That finding was:

> The history-independent allocation can produce duplicate or out-of-range field IDs ... `metadata.LastColumnID()+1` ... duplicate ... 2147483648 ... Please keep the fast path but seed collision checks from the current schema and make allocation overflow-safe.

### Allocator — fixed

`table/inspect_position_deletes.go:810-856` now seeds from `LastColumnID()+1` and checks each candidate against the base partition type, the reserved fixed position-delete IDs, and the current schema via `FindFieldByIDRef`. The boundary is an `int64` `maxFieldID = math.MaxInt32`, so the signed-32-bit overflow is gone; on exhaustion it wraps once into low IDs and then returns `ErrInvalidSchema` rather than emitting an out-of-range ID.

Regression coverage is in `table/inspect_position_deletes_regression_test.go`: current-schema collisions at `:226-255`, reserved IDs at `:258-282`, and the MaxInt32 limit fallback at `:284-326`. The reserved fixed IDs also still line up with the spec's position-deletes metadata table (`file_path` max-101, `pos` max-102, `row` max-103, partition max-5, spec max-4, `delete_file_path` max-1, content offset max-6, size max-7).

### Alias safety of removing the clone — verified

This was my main concern with the change, so spelling out what I checked:

- `Schema.MarshalJSON` now copies only the ID and identifier-ID fields (`schema.go:346`) and serializes fields read-only.
- `FindFieldByIDRef` documents its returned type graph as shared and read-only (`schema.go:466-474`), so the contract is explicit rather than incidental.
- `commonMetadata.CurrentSchema` still returns a defensive deep clone (`metadata.go:2110-2120` via `schemaByID`/`cloneSchema`).

So the helper only traverses shared state and no mutable schema internals escape. The defensive-copy contract callers rely on is intact.

### Evidence

`BenchmarkPositionDeletesPartitionType` is committed, covering 1/16/128/1024 schemas: 40,476 → 18,736; 347,383 → 18,448; 2,695,818 → 18,115; 20,795,189 → 16,137 ns/op, with allocations dropping from 94/570/4,101/32,328 to 56. That's a clear win and it scales the right way with schema-history length.

CI green (15/15).

---

>

PR #1921, opened 2026-08-27
perf(table): avoid heap scratch in canonical partition keys

fallintoplace:
## What changed

- Reuse small stack buffers in `canonicalPartitionKey`.
- Keep the existing heap fallback for wider partitions and longer keys.
- Keep sorting, value normalization, and key encoding unchanged.
- Add a focused benchmark and a test for the wide-partition fallback.

## Benchmark

Ran on an Apple M1 Pro (`darwin/arm64`) with Go `go1.26.3`.

```text
go test ./table -run '^$' -bench '^BenchmarkCanonicalPartitionKey$' -benchmem -count=5 -benchtime=1s
```

Median of 5 runs:

| Case | ns/op before | ns/op after | B/op before -> after | allocs/op before -> after |
| fields=1/int32 | 84.39 | 73.89 | 16 -> 8 | 2 -> 1 |
| fields=2/int32 | 179.8 | 94.01 | 40 -> 16 | 3 -> 1 |
| fields=4/int32 | 244.6 | 140.5 | 88 -> 32 | 4 -> 1 |
| fields=8/mixed | 435.8 | 247.0 | 392 -> 80 | 7 -> 1 |
| fields=4/binary-string | 254.8 | 157.2 | 184 -> 64 | 5 -> 1 |
| fields=16/mixed | 782.1 | 602.8 | 792 -> 288 | 8 -> 2 |

The common cases now use one allocation for the returned string. The wide case still exercises the heap fallback.

## Checks

- `go test ./...`
- `go vet ./...`
- `go test -race ./table`

### DIFF
--- a/table/conflict_validation.go
+++ b/table/conflict_validation.go
@@ -754,13 +754,18 @@ var filePathFieldID = func() int {
 // Java's DeleteFileIndex keys its PartitionMap by specId, and field IDs
 // are only meaningful within a single spec.
 func canonicalPartitionKey(specID int32, partition map[int]any) (string, error) {
-	ids := make([]int, 0, len(partition))
+	var idStorage [8]int
+	ids := idStorage[:0]
+	if len(partition) > len(idStorage) {
+		ids = make([]int, 0, len(partition))
+	}
 	for id := range partition {
 		ids = append(ids, id)
 	}
 	slices.Sort(ids)
 
-	buf := strconv.AppendInt(nil, int64(specID), 10)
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Looks good. The stack-backed scratch buffers preserve canonical partition-key encoding and safely retain heap fallback for wider partitions and longer keys.

I verified randomized output equivalence, fallback boundaries, returned-string ownership, repeated race tests, affected package tests, latest-main integration, and the allocation improvements.

---

>

PR #1923, opened 2026-08-27
perf(table): pre-size positional delete positions

fallintoplace:
## What changed

- Sum the lengths of the positional-delete Arrow chunks before creating the set.
- Use that total as the initial map capacity.
- Keep the existing validation and insertion logic unchanged.
- Keep nil chunks safe and add coverage for duplicates across chunks.
- Add a focused benchmark for 1K, 100K, and 1M positions.

## Benchmark

Ran on an Apple M1 Pro (`darwin/arm64`) with Go `go1.26.3`.

```text
go test ./table -run '^$' -bench '^BenchmarkCollectPosDeletePositions$' -benchmem -count=5 -benchtime=1s
```

Median of 5 runs:

| Input | ns/op before | ns/op after | B/op before -> after | allocs/op before -> after |
| 1K positions, 1 chunk | 39215 | 11898 | 74456 -> 36992 | 22 -> 6 |
| 100K positions, 4 chunks | 3427952 | 1500532 | 4729533 -> 2364594 | 532 -> 258 |
| 1M positions, 16 chunks | 58026680 | 45005787 | 75615856 -> 37832769 | 8210 -> 4098 |

## Checks

- `go test ./...`
- `go vet ./...`
- `go test -race ./table`

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -346,8 +346,7 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 
 	type Alias Schema
 
-	aliasCopy := *(*Alias)(s)
-	aliasCopy.IdentifierFieldIDs = ids
+	aliasCopy := Alias{ID: s.ID, IdentifierFieldIDs: ids}
 
 	return json.Marshal(struct {
 		Type   string        `json:"type"`
--- a/schema_test.go
+++ b/schema_test.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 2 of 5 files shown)


zeroshade (changes requested):
The unique-position benchmark improves, but sizing the set from raw row count creates severe over-allocation for valid duplicate-heavy delete files.

### Blocking — Bound preallocation by expected unique cardinality (`table/arrow_scanner.go:466–474`)

```go
totalPositions := 0
for _, chunk := range positionalDeletes {
    if chunk != nil {
        totalPositions += chunk.Len()
    }
}

deletes := make(set[int64], totalPositions) // raw rows, not unique positions
```

Position deletes are combined across files and may legitimately overlap. With 1,000,000 identical positions across 16 chunks, I measured:

- Previous growth strategy: cardinality 1, 0 B/op, roughly 4–6 ms
- This change: cardinality 1, approximately 37.8 MB/op, 4,098 allocations, roughly 7–20 ms

At larger scales or across concurrent scan tasks, valid overlapping deletes can reserve gigabytes despite producing a tiny set. The current benchmarks use only unique values, so they hide this regression. The capacity allocation also happens before type/null/negative validation.

Please retain cardinality-driven growth or use a benchmark-justified bounded hint such as `min(totalPositions, fixedCap)`. Add duplicate-heavy benchmark coverage alongside the unique cases.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Iceberg Go maintainer. After you've addressed the points above and pushed an update, an Iceberg Go maintainer — a real person — will take the next look at the PR. The findings cite the project's review criteria; if you think one of them is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

zeroshade (approved):
The over-allocation concern from my earlier review is fixed, and the sizing hint is safe. Approving.

My prior review was on `00ec6b4`, which is no longer in this branch's history after two force-pushes; the current head addresses the finding regardless.

### Sizing — fixed

`table/arrow_scanner.go:466-485` now saturates the hint at `64*1024`, stops summing before it can overflow `int`, and allocates with `make(map[...], min(total, cap))`. A wrong hint in either direction now only affects map allocation and growth behaviour — it cannot change the entries produced, the errors returned, or their ordering. Duplicate benchmark cases at `:261-263` cover it.

### On the memory-amplification angle — deliberately not flagging

I looked at whether a hostile capacity hint could be a memory-amplification vector, and it isn't, for two independent reasons.

Mechanically, the hint derives from actual Arrow chunk lengths rather than a self-declared record count, and it's bounded at 64K regardless.

It also wouldn't be a security finding here even if it were unbounded. `SECURITY-THREAT-MODEL.md` § *Parser hardening and malformed-input robustness* is explicit:

> Malformed-input crashes, runtime exceptions, and memory amplification are usually treated as robustness or hardening work rather than security issues in Iceberg Go itself.

and reaching it would require hostile catalog-supplied metadata, which § *Malicious catalog or external service scenarios* also places outside the primary boundary. Recording that reasoning so it doesn't get re-litigated as a vulnerability later.

CI green (15/15).

One incidental note, no action needed: the `schema.go:349` change in this PR has since landed on main via #1957, so expect that part to fall out on the next rebase.

---

>

PR #1930, opened 2026-08-28
perf(table): speed up bulk schema and partition spec removals

fallintoplace:
## What

- Build a set of requested IDs once in `RemovePartitionSpecs` and `RemoveSchemas`.
- Reuse it for protected-ID validation and list filtering.
- Keep removal order, update payloads, and no-op behavior unchanged.
- Add coverage for duplicates, requested-order differences, and atomic validation errors.
- Add benchmarks for small and historical metadata workloads.

## Why

Both methods used `slices.Contains` for every stored schema or partition spec. For `n` stored entries and `k` requested IDs, that makes the membership work O(n*k).

## Benchmark

Apple M1 Pro. Median of 3 runs:

```text
go test ./table -run '^$' -bench '^BenchmarkRemove(PartitionSpecs|Schemas)$' -benchmem -count=3 -benchtime=50ms
```

| Workload | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| Partition specs 128/64 | 4.65 us/op | 3.02 us/op | 1.5x |
| Partition specs 1,024/512 | 168 us/op | 21.0 us/op | 8.0x |
| Partition specs 8,192/4,096 | 8.69 ms/op | 183 us/op | 47x |
| Schemas 128/64 | 4.17 us/op | 3.52 us/op | 1.2x |
| Schemas 1,024/512 | 163 us/op | 18.6 us/op | 8.7x |
| Schemas 8,192/4,096 | 8.33 ms/op | 140 us/op | 60x |

The indexed path uses extra memory for the ID set, in exchange for avoiding repeated linear scans. The one-entry cases are not the target workload.

## Tests

- `go test ./table -count=1`
- `go vet ./table`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1548,27 +1548,30 @@ func (b *MetadataBuilder) reuseOrCreateNewSchemaID(newSchema *iceberg.Schema) in
 	return newSchemaID
 }
 
+func makeIntContainsFn(ints []int) func(int) bool {
+	// Scan small removal lists; reserve the index for bulk requests.
+	if len(ints) <= 16 {
+		return func(id int) bool { return slices.Contains(ints, id) }
+	}
+
+	ids := make(map[int]struct{}, len(ints))
+	for _, id := range ints {
+		ids[id] = struct{}{}
+	}
+
+	return func(id int) bool {
+		_, ok := ids[id]
+
+		return ok
+	}
+}
+
 func (b *MetadataBuilder) RemovePartitionSpecs(ints []int) error {
 	if len(ints) == 0 {
 		return nil
 	}
 
-	var removedIDs map[int]struct{}
-	if len(ints) > 1 {
-		removedIDs = make(map[int]struct{}, len(ints))
-		for _, id := range ints {
-			removedIDs[id] = struct{}{}
-		}
-	}
-	containsID := func(id int) bool {
-		if removedIDs == nil {
-			return id == ints[0]
-		}
-		_, ok := removedIDs[id]
-
-		return ok
-	}
-
+	containsID := makeIntContainsFn(ints)
 	if containsID(b.defaultSpecID) {
 		return fmt.Errorf("%w: can't remove default partition spec with id %d", iceberg.ErrInvalidArgument, b.defaultSpecID)
 	}
@@ -1598,22 +1601,7 @@ func (b *MetadataBuilder) RemoveSchemas(ints []int) error {
 		return nil
 	}
 
-	var removedIDs map[int]struct{}
-	if len(ints) > 1 {
-		removedIDs = make(map[int]struct{}, len(ints))
-		for _, id := range ints {
-			removedIDs[id] = struct{}{}
-		}
-	}
-	containsID := func(id int) bool {
-		if removedIDs == nil {
-			return id == ints[0]
-		}
-		_, ok := removedIDs[id]
-
-		return ok
-	}
-
+	containsID := makeIntContainsFn(ints)
 	if containsID(b.currentSchemaID) {
 		return fmt.Errorf("%w: can't remove current schema with id %d", iceberg.ErrInvalidArgument, b.currentSchemaID)
 	}
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -1553,14 +1553,19 @@ func (b *MetadataBuilder) RemovePartitionSpecs(ints []int) error {
 		return nil
 	}
 
-	if slices.Contains(ints, b.defaultSpecID) {
+	removedIDs := make(map[int]struct{}, len(ints))
+	for _, id := range ints {
+		removedIDs[id] = struct{}{}
+	}
+
+	if _, ok := removedIDs[b.defaultSpecID]; ok {
 		return fmt.Errorf("%w: can't remove default partition spec with id %d", iceberg.ErrInvalidArgument, b.defaultSpecID)
 	}
 
 	newSpecs := make([]iceberg.PartitionSpec, 0, len(b.specs))
 	removed := make([]int, 0, len(ints))
 	for _, spec := range b.specs {
-		if slices.Contains(ints, spec.ID()) {
+		if _, ok := removedIDs[spec.ID()]; ok {
 			removed = append(removed, spec.ID())
 
 			continue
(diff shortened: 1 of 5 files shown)


table/metadata.go
```diff
@@ -1506,14 +1506,30 @@ func (b *MetadataBuilder) RemovePartitionSpecs(ints []int) error {
 		return nil
 	}
 
-	if slices.Contains(ints, b.defaultSpecID) {
+	var removedIDs map[int]struct{}
+	if len(ints) > 1 {
```
laskoviymishka: The `> 1` threshold means we build the map as soon as there are two IDs to remove, and dropping a schema plus its rollback (or two specs) is a pretty common shape.

For two elements `slices.Contains` is two int comparisons and zero allocations, whereas the map is a heap allocation plus two hashes and inserts, so this path is strictly slower and allocates more than the code it replaces. In a PR that's specifically about cutting allocations that feels backwards. Break-even for map vs linear scan over ints usually lands somewhere around 8 to 16 elements.

The benchmark cases jump straight from removedCount 1 to 64, so the small-removal regression never shows up. I'd raise the threshold to something like `> 8` (pinned by a benchmark) and add a removedCount 2 or 4 case so the break-even is actually exercised. wdyt?

table/metadata.go
```diff
@@ -1506,14 +1506,30 @@ func (b *MetadataBuilder) RemovePartitionSpecs(ints []int) error {
...
+		removedIDs = make(map[int]struct{}, len(ints))
+		for _, id := range ints {
+			removedIDs[id] = struct{}{}
+		}
+	}
+	containsID := func(id int) bool {
```
laskoviymishka: This block is byte-for-byte identical in `RemoveSchemas` just below, so the threshold change above has to land in two places, and so does any future tweak to the fallback. I'd pull it into a small unexported helper:

```go
// makeIntContainsFn reports whether id is in ints. ints must be non-empty.
func makeIntContainsFn(ints []int) func(int) bool {
	if len(ints) <= 8 {
		return func(id int) bool { return slices.Contains(ints, id) }
	}
	set := make(map[int]struct{}, len(ints))
	for _, id := range ints {
		set[id] = struct{}{}
	}

	return func(id int) bool {
		_, ok := set[id]

		return ok
	}
}
```

Both methods collapse to `containsID := makeIntContainsFn(ints)`, and the `removedIDs == nil` / `ints[0]` dance goes away since the small case just scans the slice directly (which also drops the question of the closure capturing `ints[0]` while pinning the whole backing array). Thoughts?

table/metadata_builder_bench_test.go
```diff
@@ -42,6 +45,67 @@ var removeSnapshotsBenchmarkCases = []struct {
...
+			template := benchmarkPartitionSpecBuilder(tc.entryCount)
+			removed := benchmarkRemovedIDs(tc.removedCount)
+
+			b.ReportAllocs()
+			b.ReportMetric(float64(tc.entryCount), "spec_entries")
+			b.ReportMetric(float64(len(removed)), "removed_specs")
+			b.ResetTimer()
+
+			for range b.N {
+				builder := template
```
laskoviymishka, line 80: `BenchmarkRemoveSchemas` wraps its setup in StopTimer/StartTimer to keep the O(n) clone out of the measurement; this one doesn't, so the two benchmarks aren't measuring on the same terms. It happens to be correct today only because `RemovePartitionSpecs` allocates a fresh `newSpecs` instead of mutating in place. If it ever switched to `slices.DeleteFunc` like `RemoveSchemas`, this loop would quietly start timing shrinking input.

Also `builder.updates = nil` here is a no-op, since `builder := template` already gives you a nil `updates`.

I'd either mirror the StopTimer/StartTimer block, or drop a one-line comment noting the struct copy is O(1) and deliberately left in-window.

table/metadata_builder_internal_test.go
```diff
@@ -373,6 +373,37 @@ func TestRemovePartitionSpecsEmptyDoesNotUpdate(t *testing.T) {
 	require.Len(t, builder.specs, 1)
 }
 
+func TestRemovePartitionSpecsPreservesStoredOrderAndIgnoresDuplicates(t *testing.T) {
```
laskoviymishka, line 411: Every new test here passes three or more IDs, so they all drive the map path. The single-ID branch, which is the one bit of logic that's actually new in this PR, has no happy-path coverage, and the existing single-element tests only hit the error branch.

A refactor that broke the `removedIDs == nil` guard or the `ints[0]` compare wouldn't be caught by anything. I'd add two small cases: a one-element input that removes a matching non-default spec, and a one-element input that matches nothing, then the same pair for `RemoveSchemas`.

table/metadata_builder_internal_test.go
```diff
@@ -373,6 +373,37 @@ func TestRemovePartitionSpecsEmptyDoesNotUpdate(t *testing.T) {
...
+			iceberg.NewPartitionSpecID(7),
+		},
+		defaultSpecID: 99,
+	}
+
+	require.NoError(t, builder.RemovePartitionSpecs([]int{7, 1, 7, 123}))
+	require.Len(t, builder.specs, 1)
+	require.Equal(t, 4, builder.specs[0].ID())
+	require.Len(t, builder.updates, 1)
+	require.Equal(t, []int{1, 7}, builder.updates[0].(*removeSpecUpdate).SpecIds)
```
laskoviymishka: Worth flagging what this assertion pins: `removed` is built from the stored entries we actually find, so removing an absent ID (123 here) is silently dropped and the emitted `spec-ids` come out as `[1, 7]`. Java passes the caller's full requested set straight through, so it'd emit `[1, 7, 123]`, and PyIceberg raises on an unknown ID instead. None of this is introduced by your change, and in practice REST servers treat removing an absent spec as a no-op, so I'm not asking to change behavior.

The one thing I'd do is note it: this test now locks in the go-specific payload without saying so. A short comment (or a follow-up issue) marking the emitted IDs as intentionally the found subset would keep it from reading like an oversight later. Non-blocking.

laskoviymishka (changes requested):
This is a clean optimization and the O(n+k) rewrite is the right call for the bulk case. The added tests around order preservation, duplicate inputs, and pre-mutation validation are exactly what I'd want to see here. I'd hold it before merge though, and it's mostly about the fast path.

Main concern is the `len(ints) > 1` threshold: it allocates the map as soon as there are two IDs to remove, which is well below the point where a map beats a linear scan over ints. For the common small case, say dropping a schema and its rollback, this path now allocates and runs slower than the `slices.Contains` code it replaces, which is a bit backwards for a PR that's about cutting allocations. And because the benchmark cases jump straight from 1 to 64, the regression doesn't show up. I'd raise the threshold to something like 8 (pinned by a benchmark) and add a small-removal case.

The other thing I'd want before merge is a test for the single-ID scalar branch. It's the one path that's genuinely new here, and every new test currently drives three or more IDs, so it's uncovered.

- raise the map threshold past the break-even; add a removedCount 2/4 benchmark
- cover the single-ID happy path for both methods
- factor the duplicated `containsID` block into a shared helper
- align the two benchmarks' setup timing (and drop the no-op `updates = nil`)
- add a comment (or follow-up) on the found-only removal payload diverging from Java/PyIceberg

laskoviymishka (approved):
LGTM now.

PR #1931, opened 2026-08-28
perf(transforms): avoid regexp matching in ParseTransform

fallintoplace:
## What

- Avoid regexp matching in `ParseTransform`.
- Dispatch common lowercase transforms directly.
- Parse `bucket[...]` and `truncate[...]` with prefix, digit, and closing-bracket checks.
- Keep mixed-case support and the existing unknown/error behavior.
- Add edge-case coverage and a parser benchmark.

## Why

`ParseTransform` used to lowercase every input and run two regexp matches before checking the six simple transforms. This adds avoidable work to a common metadata parsing path.

## Benchmark

Apple M1 Pro. Median of 5 runs:

```text
go test . -run '^$' -bench '^BenchmarkParseTransform$' -benchmem -count=5 -benchtime=300ms
```

| Input | Before | After | Speedup | Allocations |
| `identity` | 16.2 ns/op, 0 B | 3.60 ns/op, 0 B | 4.5x | 0 -> 0 |
| `day` | 11.8 ns/op, 0 B | 3.03 ns/op, 0 B | 3.9x | 0 -> 0 |
| `bucket[16]` | 106 ns/op, 32 B | 14.0 ns/op, 0 B | 7.6x | 1 -> 0 |
| `truncate[8]` | 126 ns/op, 32 B | 16.0 ns/op, 0 B | 7.9x | 1 -> 0 |
| `BuCkEt[16]` | 146 ns/op, 48 B | 92.4 ns/op, 16 B | 1.6x | 2 -> 1 |
| `custom_transform[42]` | 89.1 ns/op, 16 B | 44.1 ns/op, 16 B | 2.0x | 1 -> 1 |

## Tests

- `go test . -count=1`
- `go vet .`

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"math"
 	"math/big"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -35,11 +34,6 @@ import (
 	"github.com/apache/iceberg-go/internal"
 	"github.com/google/uuid"
 	"github.com/twmb/murmur3"
-)
-
-var (
-	bucketTransformRegex   = regexp.MustCompile(`^bucket\[(\d+)\]$`)
-	truncateTransformRegex = regexp.MustCompile(`^truncate\[(\d+)\]$`)
 )
 
 // ParseTransform takes the string representation of a transform as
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM — the transform parser rewrite preserves canonical and ASCII mixed-case behavior, width validation, malformed-input handling, exact errors, and unknown-transform round-tripping while removing allocations from recognized transforms. Differential probes, repeated race tests, root tests, all-package compilation, vet, diagnostics, benchmarks, CI, and focused synthetic-main tests passed.

laskoviymishka (approved):

PR #1932, opened 2026-08-28
perf(table): avoid redundant MetadataBuilderFromBase copies

fallintoplace:
## What changed

- Added a small internal fast path for Iceberg's built-in metadata types.
- The builder now copies schemas, partition specs, snapshots, sort orders, properties, refs, statistics, partition statistics, and encryption keys directly from the internal representation.
- Custom `Metadata` implementations keep the existing getter path.
- Added ownership tests for nested mutable values.
- Added a benchmark for different metadata collection sizes.

## Why

The built-in metadata getters already clone their values. `MetadataBuilderFromBase` then cloned those results again. This removes that extra copy while keeping the builder's collections independent from the base metadata.

## Benchmark

Apple M1 Pro. Before is `upstream/main`. After is this branch. Values are the median of the runs.

| snapshots | before ns/op | after ns/op | before B/op | after B/op | before allocs/op | after allocs/op |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1,000 | 85,891 | 54,680 | 305,097 | 162,704 | 86 | 37 |
| 10,000 | 754,786 | 471,937 | 3,066,366 | 1,502,658 | 121 | 65 |

Command:

```text
go test ./table -run ^ -bench ^BenchmarkMetadataBuilderFromBase -benchmem -count=5
```

The new `BenchmarkMetadataBuilderFromBaseCollections` also covers 1, 16, 128, and 1,024 schemas; 1, 16, and 128 partition specs; 1,000 snapshots; 32 sort orders; and properties.

## Checks

- `go test ./...`
- `go vet ./table`

### DIFF
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -2329,6 +2329,10 @@ func cloneSnapshotRef(ref SnapshotRef) SnapshotRef {
 }
 
 func cloneSnapshotRefs(refs map[string]SnapshotRef) map[string]SnapshotRef {
+	if refs == nil {
+		return nil
+	}
+
 	clones := make(map[string]SnapshotRef, len(refs))
 	for name, ref := range refs {
 		clones[name] = cloneSnapshotRef(ref)
--- a/table/metadata.go
+++ b/table/metadata.go
@@ -310,6 +310,10 @@ func NewMetadataBuilder(formatVersion int) (*MetadataBuilder, error) {
 	}, nil
 }
 
+type metadataBuilderSource interface {
+	metadataBuilderCommon() *commonMetadata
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM — the built-in metadata fast path preserves v1/v2/v3 builder output, custom Metadata fallback behavior, nested ownership, snapshot/index state, version-specific fields, and update semantics while reducing redundant allocations. Differential probes, focused race tests, table suites, serial repository compilation, vet, diagnostics, benchmarks, CI, and focused synthetic-main tests passed.

PR #1933, opened 2026-08-28
perf(table): collapse metadata JSON preflight passes

fallintoplace:
## Summary

- **Reuse the top-level raw JSON map** in ParseMetadataBytes.
- Use the same map for format-version detection, legacy partition field ID normalization, and required spec-id checks.
- Keep the version-specific metadata decode and existing helper behavior unchanged.
- Added coverage for v1 legacy specs, v2 spec lists, and missing/null spec-id values. 🧪

## Benchmark

Local Apple M1 Pro. The fixture uses 1, 32, and 256 schemas/specs with 10, 1,000, and 10,000 snapshots.

| Case | Before ns/op | After ns/op | ns/op change | Before B/op | After B/op |
| --- | ---: | ---: | ---: | ---: | ---: |
| 1 schema, 1 spec, 10 snapshots | 151,089 | 107,116 | **-29.1%** | 39,381 | 32,994 |
| 32 schemas, 32 specs, 1,000 snapshots | 9,543,420 | 7,189,167 | **-24.7%** | 2,127,197 | 1,887,373 |
| 256 schemas, 256 specs, 10,000 snapshots | 95,382,604 | 69,582,812 | **-27.1%** | 22,700,280 | 20,409,368 |

**Commands:**

- go test ./table -run BenchmarkParseMetadataBytes -benchmem -benchtime=100ms -count=1
- go test ./table -count=1
- go test ./... -run no_such_test -count=1

🚀

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -346,8 +346,7 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 
 	type Alias Schema
 
-	aliasCopy := *(*Alias)(s)
-	aliasCopy.IdentifierFieldIDs = ids
+	aliasCopy := Alias{ID: s.ID, IdentifierFieldIDs: ids}
 
 	return json.Marshal(struct {
 		Type   string        `json:"type"`
--- a/schema_test.go
+++ b/schema_test.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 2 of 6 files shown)


zeroshade (changes requested):
The new exact-key preflight can disagree with the case-insensitive final struct decode. For metadata containing `"format-version": 1` followed by `"FORMAT-VERSION": 2`, preflight selects `*metadataV1`, but `json.Unmarshal` overwrites its embedded format version with 2. The parser then returns a v1 implementation reporting `Version() == 2` and bypasses required v2 validation such as `last-sequence-number`; the merge base rejects the same input. Please reject case-folded discriminator collisions or verify after decoding that the concrete type and decoded version match, and add a regression test.

zeroshade (approved):
This correctly closes the case-folded format-version hole, and the collapsed preflight does not reorder validation in any way that changes outcomes. Approving.

My prior review was on `a99f94f`, which is absent from the current four-commit history, so it was stale anyway — but the current diff fixes exactly what it described:

> The new exact-key preflight can disagree with the case-insensitive final struct decode. For metadata containing `"format-version": 1` followed by `"FORMAT-VERSION": 2`, preflight selects `*metadataV1`, but `json.Unmarshal` overwrites its embedded format version with 2. The parser then returns a v1 implementation reporting `Version() == 2` and bypasses required v2 validation such as `last-sequence-number`.

The fix is the exact raw-key map plus a post-decode consistency check comparing the decoded `ret.Version` against the preflight discriminator (`table/metadata.go` around `:1804`), with `TestParseMetadataBytesRejectsCaseFoldedFormatVersionCollision` pinning the collision.

### Validation-order audit

This is the part that actually gates a preflight-collapsing change, so here's what I verified. The resulting order is: top-level JSON map → exact `format-version` decode/switch → require `last-updated-ms` → partition-specs decode including field-map shape → spec-ID check → missing-partition-field-ID scan/normalization → version-specific unmarshal (type decode, version-scoped fields, `preValidate`/defaults, required UUID/location, common checks) → decoded-version consistency.

Against the old order, the only meaningful change is that the missing/null spec-ID check now precedes the last-partition-id/field-id scan. I found **no validation that can now be bypassed**.

Error identity is preserved: `ErrInvalidMetadata` for malformed/raw decode failures and `ErrInvalidMetadataFormatVersion` for unknown or mismatched versions. The only text difference is on the generic top-level non-object / wrong-`format-version`-type path, where the struct field path is no longer included, and no test or caller asserts on that string.

Using a pointer `*int` for the spec ID to distinguish missing from null is the right call — worth keeping in mind for similar preflight fields.

### Coverage and evidence

New malformed-metadata coverage: missing/null spec IDs, missing field IDs for v1 and v2, and the case-fold collision. The existing suite already covers `last-updated-ms` missing/null/wrong-type, required fields, and malformed schemas/specs/sort-orders/snapshots/logs/refs.

`metadata_preflight_bench_test.go` is committed with three sizes, and the body reports −24.7% to −29.1% ns/op with lower allocations.

CI green (15/15). `mergeStateStatus` reads BLOCKED only because this needed an approving review.

---

>

PR #1934, opened 2026-08-28
perf(table): avoid full rune allocation for truncated bounds

fallintoplace:
## Summary

- **Avoid a full-string rune scan and `[]rune` allocation** when truncating text bounds.
- Walk only to the requested UTF-8 rune boundary.
- Reuse the same path for variant string lower and upper bounds.
- Keep upper-bound carry behavior for the surrogate gap and maximal rune cases.
- Clone only the small truncated lower bound so it does not retain a large source string. ⚡

## Benchmark

Local Apple M1 Pro. `width=16`, `-benchmem`, `-benchtime=100ms`.

| Path | Before | After |
| --- | ---: | ---: |
| `TruncateUpperBoundText`, ASCII 64 KB | ~111,008 ns/op, 262,169 B/op | ~38 ns/op, 16 B/op |
| `TruncateUpperBoundText`, UTF-8 64 KB | ~375,620 ns/op, 139,312 B/op | ~73 ns/op, 32 B/op |
| Variant string bounds, ASCII 64 KB | ~285,109 ns/op, 524,432 B/op | ~169 ns/op, 128 B/op |
| Variant string bounds, UTF-8 64 KB | ~624,216 ns/op, 278,721 B/op | ~236 ns/op, 160 B/op |

The benchmark also covers 32 B and 1 KB values, Unicode boundaries, surrogate-gap carry, maximal-rune carry, and the all-maximal case.

## Tests

- `go test ./... -run "^$" -count=1`
- `go test ./table/internal -run "^$" -bench "^(BenchmarkTruncateUpperBoundText|BenchmarkTruncateVariantBoundString)$" -benchmem -benchtime=100ms -count=1`

### DIFF
--- a/table/internal/truncated_bounds_bench_test.go
+++ b/table/internal/truncated_bounds_bench_test.go
@@ -1,0 +1,83 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
LGTM — the byte-index truncation preserves Iceberg code-point semantics, scalar carry behavior, conservative lower/upper bounds, variant bound omission, and ownership while avoiding full rune allocation. Differential Unicode probes, focused race tests, full table suites, repository compilation, vet, diagnostics, benchmarks, CI, and synthetic-main tests all passed.

PR #1935, opened 2026-08-28
perf(table): honor row limits during local planning

fallintoplace:
## Summary

- **Use manifest-list row counts to stop local planning after enough rows are covered.**
- Keep the existing full planner for row filters, delete manifests, unknown counts, overflow, and limits that need the full list.
- Keep exact row-limit enforcement in the record reader.
- Add boundary tests, a manifest-open regression test, and a planning benchmark. ⚡

## Why

With no row filter and no delete manifests, all live rows in the selected data manifests are returned. The manifest list already has the live row counts, so local planning can stop opening manifests once the requested minimum is covered.

## Benchmark

**Machine:** Apple M1 Pro, arm64
**Setup:** in-memory v2 table, one data file and one row per manifest, max concurrency 1
**Command:** `go test ./table -run ^ -bench ^BenchmarkPlanFilesWithRowLimit -benchmem -benchtime=100ms -count=1`

`no-limit` is the full-planning control. `limit-1` means `UseRowLimit(1)`.

| Manifests | No limit | `UseRowLimit(1)` |
| ---: | ---: | ---: |
| 100 | 23.5 ms/op, 27.5 MB/op | 403 µs/op, 514 KB/op |
| 1,000 | 188.5 ms/op, 273.2 MB/op | 785 µs/op, 910 KB/op |
| 10,000 | 1.73 s/op, 2.73 GB/op | 4.71 ms/op, 5.19 MB/op |

The 10,000-manifest case is about **368x faster** with about **525x fewer allocated bytes**. 📈

## Tests

- `go test -race ./table -run 'Test(LimitManifestListByRows|CanLimitLocalPlanning|PlanFilesLocalUsesRowLimitToStopOpeningManifests)$' -count=1`
- `go vet ./table`

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -1103,6 +1103,13 @@ func (scan *Scan) planFilesLocal(ctx context.Context, acc *scanMetricsAccumulato
 	if err != nil || len(manifestList) == 0 {
 		return nil, err
 	}
+	if scan.canLimitLocalPlanning(acc) {
+		// The manifest counters above describe partition pruning. Keep them
+		// unchanged when a row limit narrows the manifest list afterwards.
+		if limitedManifests, limited := limitManifestListByRows(manifestList, scan.limit); limited {
+			manifestList = limitedManifests
+		}
+	}
 
 	// Step 2: Read manifest entries concurrently, accumulating data and positional deletes.
 	entries, err := scan.collectManifestEntriesWithSchema(ctx, manifestList, schema)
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM — the manifest-prefix optimization is safely gated to unfiltered, delete-free local scans; it falls back for unavailable or invalid counts and preserves exact reader-side row limiting. Focused race tests, full table suites, repository compilation, vet, diagnostics, benchmarks, CI, and synthetic-main tests all passed.

PR #1936, opened 2026-08-28
perf(table): cache compiled filter and pruning plans

fallintoplace:
## Summary

- **Cache Arrow filter setup per physical file schema during one scan.**
- Reuse column translation, binding, Substrait expression conversion, rewritten stats filters, and bloom predicates.
- Keep row-group stats evaluator creation per file because it owns mutable row-group maps.
- Build a fresh record processor for each task context.
- Use a low-allocation structural key that includes nested list/map field IDs.
- Keep residual task filters on the existing path.
- Keep the missing initial-default pruning fallback. Pruning plans are built only for Parquet files. ⚡

## Benchmark

Command:

```text
go test ./table -run '^$' -bench '^BenchmarkArrowScanFilterPlanSetup$' -benchmem -benchtime=1s -count=1
```

Apple M1 Pro, darwin/arm64:

| Case | ns/op | B/op | allocs/op |
| --- | ---: | ---: | ---: |
| recompile_each_file | 22,898 | 18,984 | 414 |
| cache_hit | 1,821 | 1,064 | 8 |

This benchmark isolates repeated filter and pruning setup from file I/O.

## Tests

- `go test ./...`
- `go test -race ./table -run '^TestCompiledFileFilterPlans|TestPhysicalSchemaKey|TestProcessRecords' -count=1`
- `go vet ./...`

### DIFF
--- a/table/arrow_filter_plan.go
+++ b/table/arrow_filter_plan.go
@@ -169,6 +169,7 @@ func (v *physicalSchemaKeyVisitor) Schema(_ *iceberg.Schema, _ struct{}) struct{
 func (v *physicalSchemaKeyVisitor) Struct(_ iceberg.StructType, _ []struct{}) struct{} {
 	return struct{}{}
 }
+
 func (v *physicalSchemaKeyVisitor) Field(_ iceberg.NestedField, _ struct{}) struct{} {
 	return struct{}{}
 }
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
LGTM — the per-physical-schema filter-plan cache preserves residual handling, schema-evolution safeguards, per-task mutable state, row-group pruning, row-position semantics, and delete behavior. Focused race tests, full table suites, vet, diagnostics, benchmarks, CI, and synthetic-main tests all passed.

PR #1939, opened 2026-08-28
perf(table): split parquet scan tasks by byte range

fallintoplace:
## What changed

- **Split large local Parquet files** at their recorded row-group offsets.
- Added `read.split.target-size` with a **128 MB default**.
- Parquet reads now keep only row groups inside each task range.
- Reuse immutable Parquet metadata, row-group range info, and physical schema projections across split tasks.
- Keep original file row positions for `_row_id`, positional deletes, and deletion vectors.
- Keep scan report file and delete metrics at the original data-file level.

## Why

A table with a few large Parquet files currently gives the scan one task per file. This lets the existing workers read independent row groups in parallel. Files without safe split offsets stay as one task.

Each split still owns its reader and decoded buffers for safe parallel reads. The repeated file metadata and schema setup is now shared within one scan.

## Benchmark 📈

Apple M1 Pro, 8 row groups, 32768 rows per group, a Parquet file read with 8 workers. Three runs with `-benchtime=2s -count=3`.

- One task: **7.07 to 8.41 ms/op**, 27.3 MB/op, about 2,322 allocs/op
- Row-group tasks: **4.14 to 4.24 ms/op**, 50.4 MB/op, about 3,995 allocs/op
- About **45% faster** in this workload.
- Splitting still uses more allocations because each task owns its reader and decoded buffers.
- The benchmark uses a local temporary file instead of `MemFS`. `MemFS.Open` copies the full file on every task open, which inflated the allocation comparison.

## Tests

- `go test ./table ./table/internal`
- `go test -race ./table ./table/internal`
- `go vet ./table ./table/internal`
- `go test ./table -run ^$ -bench BenchmarkArrowScanLargeParquetFileSplitTasks -benchmem -benchtime=2s -count=3`

### DIFF
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -127,7 +127,7 @@ func (parquetFormat) Open(ctx context.Context, fs iceio.IO, path string) (_ File
 		return nil, err
 	}
 
-	return wrapPqArrowReader{arrRdr}, nil
+	return wrapPqArrowReader{FileReader: arrRdr}, nil
 }
 
 func (parquetFormat) PathToIDMapping(sc *iceberg.Schema) (map[string]int, error) {
@@ -1822,6 +1822,7 @@ type RowGroupSpan struct {
 
 type wrapPqArrowReader struct {
 	*pqarrow.FileReader
+	rowGroupInfos []parquetRowGroupInfo
 }
 
 func (w wrapPqArrowReader) Metadata() Metadata {
@@ -1861,6 +1862,19 @@ func parquetRowGroupSplitOffset(rgMeta *metadata.RowGroupMetaData) int64 {
 	}
 
 	return rgMeta.FileOffset()
+}
+
+func parquetRowGroupInfos(rdr *file.Reader) []parquetRowGroupInfo {
+	result := make([]parquetRowGroupInfo, rdr.NumRowGroups())
+	for i := range result {
+		rgMeta := rdr.MetaData().RowGroup(i)
+		result[i] = parquetRowGroupInfo{
+			splitOffset: parquetRowGroupSplitOffset(rgMeta),
+			numRows:     rgMeta.NumRows(),
+		}
+	}
+
+	return result
 }
 
 func validateParquetRowGroupRange(fileSize, start, length int64) error {
@@ -1918,15 +1932,31 @@ func (w wrapPqArrowReader) GetRecords(ctx context.Context, cols []int, tester an
 		rgList = make([]int, 0)
 		var firstRowPos int64
 		for rg := range numRg {
-			rgMeta := fileMeta.RowGroup(rg)
-			numRows := rgMeta.NumRows()
 			pos := firstRowPos
+			var (
+				rgMeta  *metadata.RowGroupMetaData
+				numRows int64
+				offset  int64
+			)
+			if len(w.rowGroupInfos) == numRg {
+				info := w.rowGroupInfos[rg]
+				numRows = info.numRows
+				offset = info.splitOffset
+			} else {
+				rgMeta = fileMeta.RowGroup(rg)
+				numRows = rgMeta.NumRows()
+			}
 			firstRowPos += numRows
 
 			use := true
 			if rangeSet {
-				offset := parquetRowGroupSplitOffset(rgMeta)
+				if len(w.rowGroupInfos) != numRg {
+					offset = parquetRowGroupSplitOffset(rgMeta)
+				}
 				use = offset >= rangeStart && offset < rangeEnd
+			}
+			if use && (rowGroupTester.StatsFn != nil || bfReader != nil) && rgMeta == nil {
+				rgMeta = fileMeta.RowGroup(rg)
 			}
 			if use && rowGroupTester.StatsFn != nil {
 				var err error
@@ -2042,10 +2072,42 @@ func (pfs *ParquetFileSource) GetReader(ctx context.Context) (result FileReader,
 		}
 	}()
 
-	rdr, err := file.NewParquetReader(pf,
-		file.WithReadProps(parquet.NewReaderProperties(pfs.mem)))
-	if err != nil {
-		return nil, err
+	readProps := parquet.NewReaderProperties(pfs.mem)
+	var rdr *file.Reader
+	var rowGroupInfos []parquetRowGroupInfo
+	if cache := parquetMetadataCacheFromContext(ctx); cache != nil {
+		entry := cache.entry(parquetMetadataCacheKey{
+			path: pfs.file.FilePath(),
+			size: pfs.file.FileSizeBytes(),
+		})
+
+		var parsed *file.Reader
+		entry.once.Do(func() {
+			parsed, entry.err = file.NewParquetReader(pf, file.WithReadProps(readProps))
+			if entry.err == nil {
+				entry.fileMetadata = parsed.MetaData()
+				entry.rowGroupInfos = parquetRowGroupInfos(parsed)
+			}
+		})
+		if entry.err != nil {
+			return nil, entry.err
+		}
+		rowGroupInfos = entry.rowGroupInfos
+
+		if parsed != nil {
+			rdr = parsed
+		} else {
+			rdr, err = file.NewParquetReader(pf,
+				file.WithMetadata(entry.fileMetadata), file.WithReadProps(readProps))
+			if err != nil {
+				return nil, err
+			}
+		}
+	} else {
+		rdr, err = file.NewParquetReader(pf, file.WithReadProps(readProps))
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	props := TablePropertiesFromContext(ctx)
@@ -2064,7 +2126,7 @@ func (pfs *ParquetFileSource) GetReader(ctx context.Context) (result FileReader,
 		return nil, err
 	}
 
-	result = wrapPqArrowReader{fr}
+	result = wrapPqArrowReader{FileReader: fr, rowGroupInfos: rowGroupInfos}
 
 	return result, nil
 }
--- a/table/internal/parquet_files.go
+++ b/table/internal/parquet_files.go
@@ -1806,6 +1806,11 @@ type ParquetRowGroupTester struct {
 	// When RangeSet is true, Start and Length select row groups whose absolute
 	// Parquet row-group offset falls in [Start, Start+Length).
 	Start, Length int64
+	// PlanningSplitOffsets contains the split offsets used when the scan task
+	// was planned. When it contains one valid offset per row group, range
+	// selection uses these offsets instead of re-deriving them from the opened
+	// Parquet footer. Sparse split-offset lists fall back to the footer offsets.
+	PlanningSplitOffsets []int64
 	// Survivors, if non-nil, is reset and then filled with one span per row group
 	// that survives pruning, in file order, with positions relative to the full
 	// file. It lets callers reconstruct each emitted row's original position even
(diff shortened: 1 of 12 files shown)


table/scan_splits.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	for i, offset := range offsets {
+		if offset < 0 || offset >= fileSize || (i > 0 && offset <= offsets[i-1]) {
+			return nil, false
+		}
+	}
+
+	result := make([]FileScanTask, 0, len(offsets))
+	for i, start := range offsets {
```
zeroshade: **[P1] Sparse valid split offsets silently drop rows.** Iceberg requires `split_offsets` to be sorted, but does not require a complete list or inclusion of the first row group (all Parquet row-group offsets are only given as an example). I reproduced this with a real four-row-group file by supplying the valid sorted suffix `offsets[1:]`: `ReadTasks` returned 6 rows instead of 8. Because the first generated task starts at `offsets[0]`, no task owns an omitted leading row group. Please make the first generated range start at byte 0 (or otherwise prove completeness from the footer), and add regressions for omitted leading and interior offsets.

table/scan_splits.go
```diff
@@ -0,0 +1,70 @@
...
+// offsets stay unchanged. Keeping the original task in those cases preserves
+// the existing behavior for remote tasks and older files.
+func splitParquetScanTask(task FileScanTask, targetSize int64) ([]FileScanTask, bool) {
+	file := task.File
+	if file == nil || file.FileFormat() != iceberg.ParquetFile || targetSize <= 0 {
+		return nil, false
+	}
+
+	fileSize := file.FileSizeBytes()
+	if fileSize <= targetSize || task.Start != 0 || task.Length != fileSize {
```
zeroshade, line 39: **[P2] `read.split.target-size` is being used only as an enablement threshold, not as a target.** Once a file is one byte over the configured value, this function emits one task per recorded offset regardless of each task’s size. A file just above 128 MiB with thousands of small row groups can therefore create thousands of independently opened tasks. Please coalesce adjacent row-group ranges toward the configured target (allowing an indivisible row group to exceed it), or use and document a different threshold property rather than the standard target-size key.

table/internal/parquet_files.go
```diff
@@ -1853,6 +1883,9 @@ func (w wrapPqArrowReader) GetRecords(ctx context.Context, cols []int, tester an
 			*rowGroupTester.Survivors = (*rowGroupTester.Survivors)[:0]
 		}
 
+		rangeStart := rowGroupTester.Start
+		rangeEnd := rangeStart + rowGroupTester.Length
```
zeroshade, line 1953: **[P2] Validate partial-task ranges before adding them.** `Start + Length` can overflow, and zero or negative lengths bypass range selection in the caller and can read the whole file rather than fail. Please reject malformed ranges using subtraction-safe checks such as `Start >= 0`, `Length > 0`, `Start <= fileSize`, and `Length <= fileSize-Start`, returning `iceberg.ErrInvalidArgument` instead of silently selecting all or no row groups.

zeroshade (changes requested):
The normal split/read path, delete handling, row positions, metrics, and current-main integration otherwise look sound, and the focused race tests, affected suites, vet, diagnostics, CI, and parallel benchmark passed. One spec-valid metadata case still causes silent row loss; the two additional range-planning issues should also be addressed.

table/internal/parquet_files.go
```diff
@@ -1833,6 +1841,52 @@ func (w wrapPqArrowReader) PrunedSchema(projectedIDs map[int]struct{}, mapping i
 	return pruneParquetColumns(w.Manifest, projectedIDs, false, mapping)
 }
 
+// parquetRowGroupSplitOffset returns the same first-page offset used when
+// DataFileStatistics.SplitOffsets is built. RowGroup.file_offset is optional
+// in Parquet and Arrow returns zero when it is absent, so relying on it can
+// discard every row group in a split task from an otherwise valid file.
+func parquetRowGroupSplitOffset(rgMeta *metadata.RowGroupMetaData) int64 {
```
laskoviymishka, line 1853: This is probably the silent-row-loss case zeroshade already flagged, but let me pin the exact mechanism in case it helps: task boundaries come from `file.SplitOffsets()` (whatever the writer stored), while row-group membership here is re-derived from column 0's first page. For iceberg-go-written files those agree. For a file written by Java or iceberg-rust, the stored offsets come from `RowGroup.file_offset` / `getStartingPos()`, which can differ from column 0's first page when columns aren't laid out in field order. When they differ, a row group can fall outside every task's `[start, start+length)` and vanish from the scan with no error.

The durable fix is to stop re-deriving: carry the per-row-group offset that planning actually used (the stored `SplitOffsets` entry) into the task and filter against that same value in `GetRecords`, rather than recomputing from column 0 here. wdyt?

table/internal/parquet_files.go
```diff
@@ -1844,8 +1898,18 @@ func (w wrapPqArrowReader) GetRecords(ctx context.Context, cols []int, tester an
 		}
 	}
 
+	rangeSet := rowGroupTester != nil &&
```
laskoviymishka: The `RangeSet` field you added is exactly the right disambiguation, but this line then undercuts it: `RangeSet || Start != 0 || Length != 0` means any caller that sets `Length` without `RangeSet` silently turns on range filtering — the opposite of what the field's own doc comment promises ("the zero value keeps the historical full-file behavior"). `processRecordsWithPlans` only ever sets the fields together, so it's masked today, but `ParquetRowGroupTester` is exported, so this is a live footgun for the next caller.

I'd gate purely on the explicit flag:
```go
rangeSet := rowGroupTester != nil && rowGroupTester.RangeSet
```
wdyt?

table/scanner.go
```diff
@@ -1133,6 +1133,8 @@ func (scan *Scan) planFilesLocal(ctx context.Context, acc *scanMetricsAccumulato
 	}
 
 	results = make([]FileScanTask, 0, len(entries.dataEntries))
+	splitTargetSize := scan.metadata.Properties().GetInt64(
```
laskoviymishka, line 1165: Splitting only happens here in `planFilesLocal`, so REST-catalog scans (which go through `planFilesRemote` and hand back server-provided tasks) never split, even with `read.split.target-size` set. Reasonable scope for a first cut, but the property reads as global table config, so I'd call the local-only limitation out in the godoc and in `configuration.md` — otherwise someone benchmarks it on a local catalog and then wonders where the speedup went on REST. wdyt?

table/properties.go
```diff
@@ -34,6 +34,9 @@ const (
 	ObjectStoreEnabledKey                   = "write.object-storage.enabled"
 	ObjectStoreEnabledDefault               = false
 
+	ReadSplitTargetSizeKey     = "read.split.target-size"
```
laskoviymishka, line 40: `read.split.target-size` is one of a three-property group in Java — `read.split.planning-lookback` (10) and `read.split.open-file-cost` (4 MiB) go with it and shape the coalescing. Exporting only this one means a table carrying all three (written by a Java engine) has two silently ignored, with no discoverable constant for them. Even without wiring the behavior now, I'd reserve the two constants with their Java defaults so the group stays coherent and you avoid an API bump later.

(Tiny: the `// 128 MB` comment is really 128 MiB / 134217728 — the value's correct and matches Java's own loose comment, so only worth touching if you're already in here.)

table/internal/parquet_files.go
```diff
@@ -1844,8 +1898,18 @@ func (w wrapPqArrowReader) GetRecords(ctx context.Context, cols []int, tester an
...
+		(rowGroupTester.RangeSet || rowGroupTester.Start != 0 || rowGroupTester.Length != 0)
+	if rangeSet {
+		if err := validateParquetRowGroupRange(
+			w.SourceFileSize(), rowGroupTester.Start, rowGroupTester.Length); err != nil {
```
laskoviymishka, line 1923: One subtle mismatch: `splitParquetScanTask` builds task ranges from the manifest's `FileSizeBytes()`, but `validateParquetRowGroupRange` here checks against `w.SourceFileSize()` (the actual opened file). If those ever disagree — manifest written before the file was grown/truncated — the last split task's `length` can exceed `SourceFileSize() - start` and this returns a hard `ErrInvalidArgument` on otherwise-healthy metadata. Clamping the last range to `min(FileSizeBytes(), SourceFileSize())`, or validating locally-planned splits against `FileSizeBytes()`, avoids the surprise.

table/scan_splits_test.go
```diff
@@ -0,0 +1,466 @@
...
+
+func (f *prepareReadTrackingFile) Close() error {
+	f.fs.closed++
+
+	return f.File.Close()
+}
+
+func TestPrepareToReadClosesReaderOnSchemaError(t *testing.T) {
+	for _, cache := range []bool{false, true} {
+		t.Run(fmt.Sprintf("cache=%t", cache), func(t *testing.T) {
```
laskoviymishka: nit while you're here: `fmt.Sprintf("cache=%t", cache)` (and `field_ids=%t` below) trips `perfsprint` — `"cache=" + strconv.FormatBool(cache)` is the lint-clean form. Separately, in `TestSplitParquetScanTask` the `for _, split := range got` shadows the outer `split` bool; renaming the loop var reads cleaner and keeps `govet` shadow quiet.

laskoviymishka (commented):
Second pair of eyes on this — zeroshade's already got the main thread, so I'll keep to what's additive rather than reopen it.

First, +1 on the silent-row-loss concern: our read landed on the same spot. The mechanism, in case it helps pin it down, is that task boundaries come from the stored `SplitOffsets` while row-group membership is re-derived from column 0's first page, and those two can diverge for Java/rust-written files whose columns aren't in field order. Carrying the planning-time offset into the task and filtering against it (instead of recomputing) is the durable fix. Detail inline.

A few things I don't think are on the existing thread:

- The new `RangeSet` field is the right idea, but `GetRecords` still falls back to `Start != 0 || Length != 0`, which re-introduces the ambiguity the flag was added to remove — a footgun for any other caller of the exported tester.
- Splitting only fires on the local-planning path; REST-catalog scans silently don't split. Fine as scope, but worth documenting since the property looks global.
- Only one of Java's three split properties is exported.
- The last range's length is validated against the physical file size but built from the manifest size, so metadata drift becomes a hard error.

None of that is a rewrite, and the split / delete / row-id machinery itself reads well. Happy to re-look once the offset question and the `RangeSet` gate are settled.

zeroshade (approved):
The split implementation is safe and all three of my earlier findings are fixed. Approving.

My prior review was on `2654b58`, which is no longer in the current eight-commit history, so it was void after the force-push — but the current head addresses each finding on its merits:

| Prior finding | Now |
|---|---|
| Sparse `split_offsets` could leave leading/interior byte gaps and silently drop rows | `table/scan_splits.go` starts the first range at byte 0 and emits contiguous ranges through end-of-file; `TestPlanFilesSplitsLargeParquetFileAndReadsEachRowOnce` reads sparse offsets and asserts every row exactly once |
| `read.split.target-size` acted as a threshold, not a coalescing target | `splitParquetScanTask` coalesces adjacent row-group chunks before emitting; `TestSplitParquetScanTaskCoalescesRowGroups` covers it |
| Malformed/partial ranges were unsafe | `table/internal/parquet_files.go` validates ranges and offsets; `TestParquetRowGroupRangeSelectionRejectsInvalidRanges` covers negative, zero, out-of-file, and overflow |

### What I checked on the split math

Byte-range splitting has to partition rows exactly once, so this is where I concentrated:

- Ranges are contiguous and cover the complete half-open interval `[0, fileSize)`, so no leading, interior, or trailing gap — that closes the row-loss path from my earlier review.
- Row-group selection uses half-open offset membership and picks **whole row-group indexes**; the reader never byte-slices a row group, so no row can be bisected across two tasks. This is the property that makes the whole approach sound.
- Position spans preserve file-relative `_row_id`, positional-delete, and deletion-vector semantics. Worth calling out, since a file-relative position becoming split-relative would have been a silent correctness bug and it isn't.
- Edge shapes behave: partial, small, non-Parquet, and unsafe tasks are left unsplit; single-row-group, zero-row, and oversized-row-group files either stay unsplit or safely yield no rows.
- Planning metrics stay counted once per base data file rather than once per split.

Sparse or invalid planning metadata falls back to footer-derived offsets, which is the right conservative default.

Benchmark is committed, with the body reporting roughly 45% lower scan time (~7–8 ms/op → ~4.1–4.2 ms/op) at the cost of higher allocations — a reasonable trade for large-file scans.

No findings. CI green (15/15); `BLOCKED` was only the approval gate. @laskoviymishka's earlier comments were COMMENTED-only and their substance is addressed in the current diff.

---

>

PR #1941, opened 2026-08-28
perf(table): reuse staged schema in UpdateSpec

fallintoplace:
## Summary

- **Reuse one staged schema** throughout UpdateSpec.
- Avoid repeated defensive schema clones during partition field binding and final spec validation.
- Keep the public defensive-copy behavior unchanged.
- Add a focused benchmark for flat and nested large schemas.

## Benchmark

Command:

```
GOMAXPROCS=1 go test -run '^$' -bench '^BenchmarkUpdateSpecBuildUpdates$' -benchmem -benchtime=500ms -count=5 ./table
```

Apple M1 Pro, darwin/arm64. Results are medians of five runs.

| Case | Before | After | Speedup | Allocations |
| 100 fields, 8 partitions, 4 additions | 0.714 ms | 0.152 ms | **4.7x** | 805 KB -> 139 KB |
| 1,000 flat fields, 32 partitions, 16 additions | 29.73 ms | 1.46 ms | **20.3x** | 38.8 MB -> 1.75 MB |
| 1,000 nested fields, 32 partitions, 16 additions | 91.74 ms | 3.70 ms | **24.8x** | 102.5 MB -> 3.74 MB |

## Checks

- `go test` for all non-cloud packages
- `go vet ./...`

### DIFF
--- a/table/update_spec.go
+++ b/table/update_spec.go
@@ -43,6 +43,7 @@ type UpdateSpec struct {
 	// all schema-dependent operations in this update.
 	meta                  Metadata
 	schema                *iceberg.Schema
+	historicalFields      map[transformKey][]iceberg.PartitionField
 	err                   error
 	nameToField           map[string]iceberg.PartitionField
 	nameToAddedField      map[string]iceberg.PartitionField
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Nice improvement. Reusing one defensive staged-schema copy and caching historical partition fields preserves public copy semantics, canonical transform matching, version behavior, stored names, and lowest-spec-ID precedence while eliminating repeated deep clones and history scans. I re-ran the focused race tests repeatedly, the full table subtree, vet, the benchmark matrix against the merge base, and a synthetic merge with current main; all passed. The large nested case reduces memory from about 102.5 MB to 3.45 MB and allocations from 580,917 to 18,613.

PR #1942, opened 2026-08-28
perf(manifest): fuse partition conversion with Avro remapping

fallintoplace:
## Summary

- Fuse logical-type conversion into the final name-keyed Avro map.
- Remove the temporary ID-keyed converted map.
- Keep date, timestamp, decimal, primitive, and unknown-field handling.
- Add focused partition conversion benchmarks.

## Benchmark

Median of 5 runs on an Apple M1 Pro with `GOMAXPROCS=1`.

Command:

```text
GOMAXPROCS=1 go test -run '^$' -bench '^BenchmarkAvroEncodePartitionData$' -benchmem -benchtime=500ms -count=5 .
```

| Case | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| 4 primitive fields | 565 ns/op | 322 ns/op | 1.75x |
| 4 logical fields | 749 ns/op | 484 ns/op | 1.55x |
| 16 primitive fields | 2.80 us/op | 1.01 us/op | 2.76x |
| 16 logical fields | 3.70 us/op | 1.91 us/op | 1.94x |

For the 16 logical field case, memory dropped from 3,280 B/op and 35 allocs/op to 1,639 B/op and 28 allocs/op.

The existing end-to-end manifest writer benchmark also improved for 10,000 entries:

- 44.37 ms/op -> 41.00 ms/op, about 8% faster
- 22.33 MB/op -> 19.77 MB/op
- 120,463 allocs/op -> 100,460 allocs/op

Command:

```text
GOMAXPROCS=1 go test -run '^$' -bench '^BenchmarkManifestWriterPartitionSummaries$' -benchmem -benchtime=1s -count=5 .
```

## Checks

- `go test . -count=1`
- All non-cloud packages pass
- `go vet ./...`

### DIFF
--- a/data_file_codec.go
+++ b/data_file_codec.go
@@ -211,11 +211,7 @@ func cloneDataFileAvroFields(src *dataFile) *dataFile {
 // avro-friendly map the manifest-entry schema expects. Idempotent:
 // values already in primitive form pass through unchanged.
 func avroEncodePartitionData(idKeyed map[int]any, fields dataFileFieldMaps) (map[string]any, error) {
-	converted, err := avroPartitionData(idKeyed, fields.idToType, fields.idToFixedSize)
-	if err != nil {
-		return nil, err
-	}
-	out := make(map[string]any, len(converted))
+	out := make(map[string]any, len(fields.nameToID))
 	for name, id := range fields.nameToID {
 		if _, unknown := fields.unknownFieldIDs[id]; unknown {
 			// UnknownType has no value representation and is encoded as Avro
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Nice improvement. Fusing logical-type conversion into the final name-keyed map preserves primitive, logical, missing, unknown, decimal-validation, and error behavior while removing the temporary ID-keyed map. I re-ran focused race coverage, a direct old/new differential probe, the affected suites, vet, the benchmark matrix against the merge base, and a synthetic merge with current main; all passed. The 16-field logical case roughly halves runtime and bytes while reducing allocations.

PR #1944, opened 2026-08-28
perf(table): avoid schema clones in remote wildcard projection

fallintoplace:
## Summary

- Use `Schema.NumFields()` for the projected ID slice capacity.
- Avoid the extra defensive deep clone from `Schema.Fields()`.
- Keep the existing wildcard traversal and nested collection projection rules.
- Add a focused remote wildcard projection benchmark.

## Benchmark

Median of 5 runs on an Apple M1 Pro with `GOMAXPROCS=1`. The benchmark uses nested structs, lists, maps, and default values.

Command:

```text
GOMAXPROCS=1 go test -run '^$' -bench '^BenchmarkRemotePlanningSelectedFields$' -benchmem -benchtime=500ms -count=5 ./table
```

| Top-level fields | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| 100 | 152 us/op | 76.6 us/op | 1.99x |
| 1,000 | 1.43 ms/op | 750 us/op | 1.91x |
| 5,000 | 6.78 ms/op | 3.63 ms/op | 1.87x |

Memory dropped by about 47% and allocations dropped by about 50%:

- 5,000 fields: 8.52 MB/op and 60,005 allocs/op -> 4.49 MB/op and 30,005 allocs/op

## Checks

- `go test . -count=1`
- All non-cloud packages pass
- `go vet ./...`

### DIFF
--- a/table/scan_planning_remote.go
+++ b/table/scan_planning_remote.go
@@ -51,7 +51,7 @@ func remotePlanningSelectedFields(scan *Scan, schema *iceberg.Schema) ([]string,
 		return scan.remoteSelectedFields(schema), nil
 	}
 
-	ids := make([]int, 0, len(schema.Fields()))
+	ids := make([]int, 0, schema.NumFields())
 	for _, field := range schema.Fields() {
 		appendRemoteProjectedFieldIDs(&ids, field)
 	}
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Nice improvement. Using `NumFields()` for the capacity hint preserves wildcard traversal, ordering, nested collection handling, and defensive ownership while avoiding a redundant recursive schema clone. I re-ran the focused race tests repeatedly, the full table package, vet, the benchmark matrix against the merge base, and a synthetic merge with current main; all passed. The 5,000-field benchmark roughly halves runtime, bytes, and allocations.

PR #1945, opened 2026-08-28
perf(table): borrow source fields during DataFile finalization

fallintoplace:
## Summary

- Resolve each partition source field once during `DataFile` finalization.
- Use the borrowed source field type for partition bound conversion and transform result typing.
- Avoid defensive field clones in the normal `ToDataFile` path.
- Keep the existing behavior for missing statistics and invalid schemas.
- Add a focused finalization benchmark.

## Benchmark

Median of 5 runs on an Apple M1 Pro with `GOMAXPROCS=1`.

The fixture uses binary source fields with mutable defaults so the defensive field copies are visible in the allocation numbers.

Command:

```text
GOMAXPROCS=1 go test -run '^$' -bench '^BenchmarkDataFileStatisticsToDataFile$' -benchmem -benchtime=500ms -count=5 ./table/internal
```

| Partition fields | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| 1 | 1.80 us/op | 1.51 us/op | 1.19x |
| 8 | 8.33 us/op | 3.98 us/op | 2.09x |
| 32 | 39.6 us/op | 22.5 us/op | 1.76x |

For 32 partition fields:

- 29,088 B/op -> 18,336 B/op
- 676 allocs/op -> 292 allocs/op

## Checks

- `go test . -count=1`
- All non-cloud packages pass
- `go vet ./...`

### DIFF
--- a/table/internal/utils.go
+++ b/table/internal/utils.go
@@ -36,6 +36,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/decimal"
 	"github.com/apache/arrow-go/v18/arrow/decimal128"
 	"github.com/apache/iceberg-go"
+	iceberginternal "github.com/apache/iceberg-go/internal"
 	"github.com/twmb/avro/atype"
 	"golang.org/x/sync/errgroup"
 )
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Nice improvement. Borrowing each source field once during DataFile finalization preserves the existing partition inference and logical-type behavior while avoiding repeated defensive clones. I re-ran the focused race tests repeatedly, the full table subtree, vet, the benchmark comparison, and a synthetic merge with current main; all passed. The longer benchmark samples confirmed both the allocation reduction and speedup across multi-field specs.

PR #1947, opened 2026-08-28
perf(table): reuse schema fields when resolving sort keys

fallintoplace:
## What

- Reuse the cloned top-level schema fields while resolving sort keys.
- Call Schema.Fields once per sort-order resolution instead of once per sort field.
- Add before and after benchmark coverage for wide and nested schemas.

## Why

Schema.Fields deep-clones fields and nested types. Sort key resolution was repeating that full clone for every sort key. This keeps the lookup behavior and output unchanged while removing the repeated copy work.

## Benchmark

Command: go test ./table -run=^$ -bench=^BenchmarkResolveSortKeys$ -benchmem -benchtime=1s -count=5

Apple M1 Pro, Go 1.26.3, darwin/arm64. Values below are representative samples from the five runs.

| Case | Before | After |
| --- | ---: | ---: |
| 32 fields / 4 sort keys | 12,864 B/op, 5 allocs/op | 3,264 B/op, 2 allocs/op |
| 32 fields / 16 sort keys | 51,456 B/op, 17 allocs/op | 3,456 B/op, 2 allocs/op |
| 256 fields / 8 sort keys | 218,240 B/op, 9 allocs/op | 27,392 B/op, 2 allocs/op |
| 256 fields / 32 sort keys | 872,960 B/op, 33 allocs/op | 27,776 B/op, 2 allocs/op |
| 2,048 fields / 32 sort keys | 6,291,972 B/op, 33 allocs/op | 197,120 B/op, 2 allocs/op |
| 128 nested fields / 16 sort keys | 1,052,929 B/op, 4,113 allocs/op | 66,048 B/op, 258 allocs/op |

The 2,048-field, 32-key case uses about 96.9% fewer bytes per operation.

## Tests

- go test ./table -count=1
- go test -race ./table -count=1
- go vet ./table
- golangci-lint run --timeout=10m (0 issues)

Signed-off-by: Minh Vu <vuhoangminh97@gmail.com>

### DIFF
--- a/table/sort_keys.go
+++ b/table/sort_keys.go
@@ -47,9 +47,10 @@ func resolveSortKeys(order SortOrder, fileSchema *iceberg.Schema) ([]compute.Sor
 		return nil, nil
 	}
 
+	fields := fileSchema.Fields()
 	keys := make([]compute.SortKey, 0, order.Len())
 	for _, field := range order.fields {
-		idx, ok := topLevelFieldIndex(fileSchema, field.SourceID())
+		idx, ok := topLevelFieldIndex(fields, field.SourceID())
 		if !ok {
 			return nil, fmt.Errorf("sort order %d: source id %d is not a top-level column in schema",
 				order.OrderID(), field.SourceID())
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Nice improvement. Reusing a single defensive `Schema.Fields()` clone preserves the resolver’s ownership and top-level lookup semantics while removing the clone-per-sort-key cost. I re-ran the resolver race tests repeatedly, the full table suite, vet, and the complete benchmark matrix against current main; all passed. The wide and nested cases confirm that allocations now remain constant as the number of sort keys increases.

PR #1948, opened 2026-08-28
perf(table): collect fanout partitions into one slice

fallintoplace:
## What changed

- **Collect partition leaves into one output slice.**
- Pre-size it from the root partition count.
- Remove per-node leaf count bookkeeping.
- Add multi-field coverage and a high-cardinality benchmark.

## Why

The old collector built a temporary result slice at every nested map level and appended those slices into the parent. This created extra allocations for multi-field partition specs.

The new traversal passes one destination slice through the tree. It keeps the same partition set and map iteration behavior.

## Checks

- `go test ./... -count=1`
- `go vet ./table`
- Collection benchmark reports `1 allocs/op`, including 16,384 partitions.

### DIFF
--- a/table/partition_extraction_bench_test.go
+++ b/table/partition_extraction_bench_test.go
@@ -82,6 +82,66 @@ func BenchmarkPartitionExtraction(b *testing.B) {
 	}
 }
 
+func BenchmarkPartitionMapCollection(b *testing.B) {
+	for _, test := range []struct {
+		fieldCount     int
+		partitionCount int
+	}{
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Nice improvement. Passing one destination slice through the partition tree preserves the complete partition set and existing arbitrary map traversal while eliminating intermediate collection buffers. I re-ran the focused race tests repeatedly, the full table package, vet, the head/base benchmark comparison, and a synthetic merge with current main; all passed. The 16,384-partition cases now collect with one allocation.

PR #1949, opened 2026-08-28
perf(table): reuse Parquet schema metadata across file writers

fallintoplace:
## What changed

- **Precompute Parquet column mappings once per writer lifecycle.**
- Reuse variant field IDs too.
- Pass both maps through the default and rolling writer paths.
- Keep the existing fallback for direct file-writer callers.
- Add correctness coverage and a writer setup benchmark.

## Why

Each physical Parquet file rebuilt the same schema mappings even though the schema is invariant for the writer. The maps are read-only, so they can be shared safely across files.

## Checks

- `go test ./... -count=1`
- `go test -race ./table/internal -count=1`
- `go vet ./table ./table/internal`

The focused benchmark shows about 12-14% lower writer setup time and 33-790 fewer allocations per file for 8-256 fields.

### DIFF
--- a/table/internal/interfaces.go
+++ b/table/internal/interfaces.go
@@ -119,6 +119,8 @@ type WriteFileInfo struct {
 	Spec             iceberg.PartitionSpec
 	FileName         string
 	StatsCols        map[int]StatisticsCollector
+	ColMapping       map[string]int
+	VariantFieldIDs  map[int]struct{}
 	WriteProps       any
 	RowGroupBytes    int64
 	Content          iceberg.ManifestEntryContent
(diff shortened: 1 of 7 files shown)


zeroshade (approved):
Nice improvement. Precomputing the Parquet path mapping and variant field IDs after the final writer schema is selected preserves direct-call fallback behavior while safely sharing read-only metadata across physical file writers. I re-ran focused race coverage, the full affected packages, vet, high-risk shredded-variant, geometry, and row-lineage writer tests, the benchmark matrix, and a synthetic merge with current main; all passed. The benchmark confirms lower setup allocations at every schema size.

PR #1950, opened 2026-08-28
perf(table): lazily materialize dictionary columns

fallintoplace:
## Summary

- **Delay the column and field copies until a dictionary is found.**
- **Keep the common no-dictionary path allocation-free.**
- Add coverage for unchanged batches and dictionaries after an earlier column.
- Add a benchmark for plain records with 1, 8, and 64 columns.

`materializeDictionaryColumns` is used for partial fanout batches. Most batches do not contain dictionary arrays, so the old eager copies were unnecessary.

## Tests

- `go test ./... -count=1`
- `go test ./table -count=1`
- `go test ./table -run no_tests -bench BenchmarkMaterializeDictionaryColumns -benchmem -count=1`
- Benchmark result: `0 B/op`, `0 allocs/op` on the no-dictionary cases.

### DIFF
--- a/table/partition_batch_bench_test.go
+++ b/table/partition_batch_bench_test.go
@@ -71,6 +71,28 @@ func BenchmarkPartitionBatchByKey(b *testing.B) {
 	}
 }
 
+func BenchmarkMaterializeDictionaryColumns(b *testing.B) {
+	const rows = 1024
+
+	for _, columns := range []int{1, 8, 64} {
+		record := newPartitionBatchBenchmarkRecord(columns, rows)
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Nice improvement. Deferring the column, field, and materialized-array copies until the first dictionary is found preserves record identity, metadata, ownership, and nested dictionary behavior while making the common plain-record path allocation-free. I re-ran the focused race tests repeatedly, the full table package, vet, the benchmark matrix against the merge base, and a synthetic merge with current main; all passed. The benchmarks confirm zero allocations for plain records at every tested schema width.

PR #1952, opened 2026-08-28
fix(table): resolve dropped equality fields from schema history

### LINKED ISSUE #1951: Equality-delete scans fail after dropping the delete key column
### Description

Scanning a format-v2 table fails after an equality-delete key column is dropped from the current schema, even when the scan selects only surviving columns.

Equality delete files store stable table field IDs, and Iceberg retains historical schemas after a field is dropped. The read path currently resolves equality field IDs only through `metadata.CurrentSchema()`, so a live delete file becomes unreadable as soon as its key is absent from that schema.

### Expected behavior

The scan should resolve the equality field ID from schema history, apply the delete, and return only the requested current-schema columns.

This matches the existing rewrite invariant: `validateRewriteEqualityFieldIDs` accepts equality fields retained only in schema history, and `TestReplaceFilesWithDeleteFilesAllowsDroppedEqualityField` permits rewriting such a delete file.

### Actual behavior

Consuming the Arrow record iterator fails with:

```text
equality delete field ID 1 not found in table schema for .../data/delete.parquet
```

The failure occurs before equality-delete filtering can run.

### Reproduction

1. Create a format-v2 table with `id: long` (field ID 1) and `data: string`.
2. Append rows `(1, "keep")` and `(2, "delete")`.
3. Commit an equality-delete file keyed by field ID 1 that deletes `id = 2`.
4. Confirm the scan returns only `(1, "keep")`.
5. Drop `id` with `UpdateSchema(...).DeleteColumn([]string{"id"})` and commit.
6. Scan only `data` and consume the record iterator.

The iterator returns the error above instead of one row containing `"keep"`.

I reproduced this twice on current `main` at `0bf8092c8586baa279a5378afddb678baafed4a8` with Go 1.26.0 on macOS arm64. The same current-schema lookup is present in v0.6.0.

### Root cause

`readEqualityDeleteFile` calls `tableSchema.FindColumnName(fid)`, while `arrowScanner` supplies only `metadata.CurrentSchema()`. Unlike rewrite validation, this path cannot fall back to the table's historical schemas.

### Suggested fix

Resolve each equality field ID from current or historical table schemas for physical delete evaluation while preserving the current schema and requested projection for returned records. Add an end-to-end regression test that drops the key and scans a surviving column.

mattfaltyn:
Fixes #1951.

## Summary

- resolve equality-delete fields against the newest table schema that contains the complete key when those fields are absent from the current schema
- load schema history only for scans whose equality keys are missing from the current schema
- extend the equality-delete round trip to drop the key and verify a surviving-column projection still applies the delete

## Why

Equality delete files retain stable table field IDs after schema evolution. The reader previously resolved those IDs only against the current schema, so dropping a live equality key made affected scans fail before filtering.

Using the newest historical schema that contains the delete file's complete key preserves field identity and keeps the current result projection unchanged.

## Testing

- `go test ./table -run '^TestEqualityDeleteReadRoundTrip$' -count=2 -v`
- `go test ./table -run 'EqualityDelete' -count=1`
- `go test -race ./table -run '^TestEqualityDeleteReadRoundTrip$' -count=1`
- `go vet ./table`
- `make test`
- `make lint`
- `make test-race`
- `make test-assert`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -2189,8 +2189,22 @@ func (as *arrowScan) GetRecords(ctx context.Context, tasks []FileScanTask) (*arr
 		return nil, nil, err
 	}
 
+	var tableSchemas []*iceberg.Schema
+
+loadSchemaHistory:
+	for _, task := range tasks {
+		for _, deleteFile := range task.EqualityDeleteFiles {
+			for _, fieldID := range deleteFile.EqualityFieldIDs() {
+				if _, found := invariants.tableSchema.FindFieldByID(fieldID); !found {
+					tableSchemas = as.metadata.Schemas()
+
+					break loadSchemaHistory
+				}
+			}
+		}
+	}
 	equalityDeleteLoader, err := newLazyEqualityDeleteLoader(
-		as.fs, invariants.tableSchema, invariants.nameMapping, tasks)
+		as.fs, invariants.tableSchema, tableSchemas, invariants.nameMapping, tasks)
 	if err != nil {
 		releasePerFilePosDeletes(deletesPerFile)
 
--- a/table/equality_delete_reader.go
+++ b/table/equality_delete_reader.go
@@ -302,9 +302,35 @@ func newEqualityDeleteFileSet(id int, deleteSet *equalityDeleteSet) *equalityDel
 	}
 }
 
+func schemaForEqualityFields(current *iceberg.Schema, schemas []*iceberg.Schema, fieldIDs []int) *iceberg.Schema {
+	hasAllFields := func(schema *iceberg.Schema) bool {
+		for _, fieldID := range fieldIDs {
+			if _, ok := schema.FindFieldByID(fieldID); !ok {
+				return false
+			}
+		}
+
+		return true
+	}
+
+	if hasAllFields(current) {
+		return current
+	}
+	// Scan tasks do not retain the equality delete's sequence number, so use
+	// the newest schema that can resolve the file's complete equality key.
+	for i := len(schemas) - 1; i >= 0; i-- {
+		if hasAllFields(schemas[i]) {
+			return schemas[i]
+		}
+	}
+
+	return current
+}
+
 type lazyEqualityDeleteLoader struct {
 	fs           iceio.IO
 	tableSchema  *iceberg.Schema
+	tableSchemas []*iceberg.Schema
 	nameMapping  iceberg.NameMapping
 	files        map[string]*lazyEqualityDeleteFile
 	combinations sync.Map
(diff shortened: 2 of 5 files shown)


zeroshade (commented):
Thanks for this — the fix is correct and it's a real bug worth closing. Resolving equality-delete field IDs from schema history is the right approach, and I checked the two things that most often go wrong here:

- **Fields are matched by ID, not name.** This is the important one: names can be reused across different field IDs, so name-matching would silently apply deletes to the wrong column. Matching by ID avoids that.
- **A field absent from every historical schema produces a clean error** (propagated from `FindColumnName`) rather than a silent skip. A silent skip would drop the delete and resurrect rows, so failing loudly is right.

Two things before this can land, one mechanical and one small.

### Please rebase

`mergeStateStatus` is `DIRTY`. Thirteen PRs from an unrelated stack merged earlier today, so this just needs a rebase onto current `main`.

### Minor — add a not-found regression test

The clean-error path above isn't pinned by a test. Since the alternative behaviour (silently dropping a delete) is a data-correctness failure, it's worth locking in.

### Follow-up observation — pre-existing, explicitly not caused by this PR

While reviewing I checked whether type evolution interacts badly here, and there is an issue — but it is not yours. If an equality key evolved `int32` → `int64`, the delete file's values are encoded at the old physical width while the data file uses the new one, and `makeColEncoder` tags `Int32` as `u32` and `Int64` as `u64`, so the comparison silently fails to match and deletes don't apply.

I confirmed this already happens on `main` for **live** (non-dropped) evolved fields: `readEqualityDeleteFile` uses the table schema only for name mapping, and `processEqualityDeletesColumnarForFile` runs *before* `ToRequestedSchema` performs any widening. Your change simply routes the narrow dropped-field case into the same pre-existing path. Fixing it means adding coercion (or an explicit rejection) on that comparison, which belongs in its own change — it may deserve a separate issue.

### One design question, not a blocker

The selection rule is "newest schema containing the IDs". That's reasonable, but it's ambiguous if several historical schemas contain the same field ID with different types. Would selecting by the delete file's sequence number be more principled, or is the necessary metadata not available at that point? Genuinely asking — if newest-containing is the best option available, a short comment saying so would help the next reader.

CI green, DCO signed.

---

>

mattfaltyn:
> Please rebase
>
> Minor — add a not-found regression test
>
> If newest-containing is the best option available, a short comment saying so would help the next reader.

Thanks for the thoughtful review! Addressed all three: rebased onto current `main`, added the not-found regression test, and documented that `FileScanTask` does not retain the equality delete’s sequence number, so the newest schema containing the complete key is the best available choice. All local and GitHub checks are green.

laskoviymishka (approved):
LGTM!

PR #1953, opened 2026-08-28
perf(internal): pack from end without cloning input

fallintoplace:
## Summary

- `PackEnd` now iterates over the input slice backwards instead of cloning and reversing it.
- It reuses the packed outer slice instead of allocating a second `[][]T`.
- The input slice remains unchanged.
- Empty input keeps returning a non-nil empty result.

## Benchmarks

Ran on an Apple M1 Pro with Go 1.26.3. Medians from five runs:

- int64, 128 items: 1.27 us, 3,160 B/op, 12 allocs/op -> 0.96 us, 2,112 B/op, 10 allocs/op
- 64-byte item, 16,384 items: 604 us, 3.16 MB/op, 1,162 allocs/op -> 461 us, 2.11 MB/op, 1,160 allocs/op
- 64-byte item, 131,072 items: 4.41 ms, 25.27 MB/op, 9,229 allocs/op -> 4.04 ms, 16.85 MB/op, 9,227 allocs/op

## Tests

- `go test -timeout=5m $(go list ./... | rg -v "/io/gocloud$")`
- `go test -race -timeout=5m ./internal ./table/...`

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -178,18 +178,24 @@ func (s *SlicePacker[T]) Pack(items []T, weightFunc func(T) int64) [][]T {
 }
 
 func (s *SlicePacker[T]) PackEnd(items []T, weightFunc func(T) int64) [][]T {
-	items = slices.Clone(items)
-	slices.Reverse(items)
-	packed := s.Pack(items, weightFunc)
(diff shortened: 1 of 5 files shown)


laskoviymishka (approved):
LGTM

PR #1955, opened 2026-08-28
perf(table): skip metadata builds when Transaction.apply has no new requirements

fallintoplace:
## Summary

- **Lazy metadata builds:** `Transaction.apply` now waits to build immutable metadata until a new requirement needs validation.
- **No-op requirement paths:** Empty requirement lists and duplicate requirements skip `MetadataBuilder.Build()` while keeping conflict checks.
- **Atomic staging:** Updates are applied to cloned builders and published only after all updates succeed.
- **Staged metadata logs:** Repeated applies with no new requirements no longer add duplicate previous metadata-log entries when `StagedTable()` builds metadata. The commit path is unchanged: the catalog replays staged updates and validates the committed metadata.

## Benchmark

Apple M1 Pro, Go 1.26.3, 10,000 snapshots:

- **No requirements:** about 1.50 MB / 65 allocs to 1.21 MB / 14 allocs
- **Duplicate requirement:** about 1.50 MB / 70 allocs to 1.21 MB / 19 allocs
- **New requirement:** about 1.51 MB / 77 allocs before and after

## Tests

- `go test -timeout=5m` across all packages except `io/gocloud`
- `go test -race -timeout=5m . ./table/...`

### DIFF
--- a/table/metadata_builder_bench_test.go
+++ b/table/metadata_builder_bench_test.go
@@ -206,6 +206,7 @@ func BenchmarkTransactionApplyNoRequirements(b *testing.B) {
 			txn := &Transaction{meta: template.clone()}
 
 			b.ReportAllocs()
+			b.ReportMetric(float64(snapshotCount), "snapshot_entries")
 			b.ResetTimer()
 
 			for b.Loop() {
@@ -214,7 +215,6 @@ func BenchmarkTransactionApplyNoRequirements(b *testing.B) {
 				}
 				metadataBuilderBenchmarkSink = len(txn.meta.snapshotList)
 			}
-			b.ReportMetric(float64(snapshotCount), "snapshot_entries/op")
 		})
 	}
 }
@@ -231,6 +231,7 @@ func BenchmarkTransactionApplyDuplicateRequirement(b *testing.B) {
 			}
 
 			b.ReportAllocs()
+			b.ReportMetric(float64(snapshotCount), "snapshot_entries")
 			b.ResetTimer()
 
 			for b.Loop() {
@@ -239,7 +240,6 @@ func BenchmarkTransactionApplyDuplicateRequirement(b *testing.B) {
 				}
 				metadataBuilderBenchmarkSink = len(txn.meta.snapshotList)
 			}
-			b.ReportMetric(float64(snapshotCount), "snapshot_entries/op")
 		})
 	}
 }
@@ -253,6 +253,7 @@ func BenchmarkTransactionApplyNewRequirement(b *testing.B) {
 			requirement := AssertCurrentSchemaID(0)
 
 			b.ReportAllocs()
+			b.ReportMetric(float64(snapshotCount), "snapshot_entries")
 			b.ResetTimer()
 
 			for b.Loop() {
@@ -262,7 +263,6 @@ func BenchmarkTransactionApplyNewRequirement(b *testing.B) {
 				}
 				metadataBuilderBenchmarkSink = len(txn.meta.snapshotList)
 			}
-			b.ReportMetric(float64(snapshotCount), "snapshot_entries/op")
 		})
 	}
 }
--- a/table/metadata_builder_bench_test.go
+++ b/table/metadata_builder_bench_test.go
@@ -198,6 +198,75 @@ func BenchmarkMetadataBuilderCloneAndBuild(b *testing.B) {
 	}
 }
 
+func BenchmarkTransactionApplyNoRequirements(b *testing.B) {
+	for _, snapshotCount := range []int{1, 100, 1_000, 10_000} {
+		b.Run(fmt.Sprintf("snapshots=%d", snapshotCount), func(b *testing.B) {
+			template := benchmarkSnapshotBuilder(snapshotCount)
+			template.ensureSnapshotIndex()
+			txn := &Transaction{meta: template.clone()}
+
+			b.ReportAllocs()
+			b.ResetTimer()
+
+			for b.Loop() {
+				if err := txn.apply(nil, nil); err != nil {
+					b.Fatal(err)
+				}
+				metadataBuilderBenchmarkSink = len(txn.meta.snapshotList)
+			}
+			b.ReportMetric(float64(snapshotCount), "snapshot_entries/op")
+		})
+	}
+}
+
+func BenchmarkTransactionApplyDuplicateRequirement(b *testing.B) {
+	for _, snapshotCount := range []int{1, 100, 1_000, 10_000} {
+		b.Run(fmt.Sprintf("snapshots=%d", snapshotCount), func(b *testing.B) {
+			template := benchmarkSnapshotBuilder(snapshotCount)
+			template.ensureSnapshotIndex()
+			requirement := AssertCurrentSchemaID(0)
+			txn := &Transaction{
+				meta: template.clone(),
+				reqs: []Requirement{requirement},
+			}
+
+			b.ReportAllocs()
+			b.ResetTimer()
+
+			for b.Loop() {
+				if err := txn.apply(nil, []Requirement{requirement}); err != nil {
+					b.Fatal(err)
+				}
+				metadataBuilderBenchmarkSink = len(txn.meta.snapshotList)
+			}
+			b.ReportMetric(float64(snapshotCount), "snapshot_entries/op")
+		})
+	}
+}
+
+func BenchmarkTransactionApplyNewRequirement(b *testing.B) {
+	for _, snapshotCount := range []int{1, 100, 1_000, 10_000} {
+		b.Run(fmt.Sprintf("snapshots=%d", snapshotCount), func(b *testing.B) {
+			template := benchmarkSnapshotBuilder(snapshotCount)
+			template.ensureSnapshotIndex()
+			txn := &Transaction{meta: template.clone()}
+			requirement := AssertCurrentSchemaID(0)
+
+			b.ReportAllocs()
+			b.ResetTimer()
+
+			for b.Loop() {
+				txn.reqs = nil
+				if err := txn.apply(nil, []Requirement{requirement}); err != nil {
+					b.Fatal(err)
+				}
+				metadataBuilderBenchmarkSink = len(txn.meta.snapshotList)
+			}
+			b.ReportMetric(float64(snapshotCount), "snapshot_entries/op")
+		})
+	}
+}
+
 func BenchmarkMetadataBuilderFromBase(b *testing.B) {
 	for _, snapshotCount := range []int{1_000, 10_000} {
 		b.Run(fmt.Sprintf("snapshots=%d", snapshotCount), func(b *testing.B) {
--- a/table/transaction.go
+++ b/table/transaction.go
@@ -183,43 +183,50 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 	// staged metadata: a deduplicated twin re-asserts base state the
 	// staged metadata has intentionally moved past, and its kept twin
 	// was validated when first added.
-	existing := make(map[string]Requirement, len(t.reqs))
-	stagedReqs := make([]Requirement, 0, len(t.reqs)+len(reqs))
-	for _, r := range t.reqs {
-		key, err := requirementSemanticKey(r)
-		if err != nil {
-			return err
-		}
-
-		existing[key] = r
-		stagedReqs = append(stagedReqs, r)
-	}
-
-	for _, r := range reqs {
-		key, err := requirementSemanticKey(r)
-		if err != nil {
-			return err
-		}
-
-		if prev, ok := existing[key]; ok {
-			if err := checkRequirementConflict(prev, r); err != nil {
-				return err
-			}
-
-			continue
-		}
-		if current == nil {
-			built, err := stagedMeta.Build()
+	stagedReqs := t.reqs
+	if len(reqs) > 0 {
+		existing := make(map[string]Requirement, len(t.reqs))
+		for _, r := range t.reqs {
+			key, err := requirementSemanticKey(r)
 			if err != nil {
 				return err
 			}
-			current = built
+
+			existing[key] = r
 		}
-		if err := r.Validate(current); err != nil {
-			return err
+
+		copiedReqs := false
+		for _, r := range reqs {
+			key, err := requirementSemanticKey(r)
+			if err != nil {
+				return err
+			}
+
+			if prev, ok := existing[key]; ok {
+				if err := checkRequirementConflict(prev, r); err != nil {
+					return err
+				}
+
+				continue
+			}
+			if current == nil {
+				built, err := stagedMeta.Build()
+				if err != nil {
+					return err
+				}
+				current = built
+			}
+			if err := r.Validate(current); err != nil {
+				return err
+			}
+			if !copiedReqs {
+				stagedReqs = make([]Requirement, len(t.reqs), len(t.reqs)+len(reqs))
+				copy(stagedReqs, t.reqs)
+				copiedReqs = true
+			}
+			existing[key] = r
+			stagedReqs = append(stagedReqs, r)
 		}
-		existing[key] = r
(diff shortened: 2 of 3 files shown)


zeroshade (changes requested):
The fast path itself looks correct — I specifically checked that it can't commit stale metadata — but nothing in the test suite pins which branch was taken, and that's the part I'd like fixed before this lands.

### The safety question, answered

The obvious hazard is an update-only transaction skipping the rebuild and committing stale metadata. It doesn't: `Transaction.Commit` only calls `doCommit` when there are updates, and the catalog's `CommitTable` applies each update and then builds and validates the metadata itself. `StagedTable` and `Scan` also build. I found no path where `Build` is skipped entirely. Recording that explicitly so it doesn't have to be re-derived later.

Requirement dedup keys off `requirementSemanticKey` — canonical JSON for the seven assert types, and type+ref for `assert-ref` — with duplicates taking the conflict check and continuing, and `t.reqs` only appended after validation. That reads correct.

### Major — no test pins the skip vs no-skip branch (`table/transaction.go:172`)

This PR's whole behaviour is "sometimes rebuild, sometimes don't", and the added tests are a schema-concurrency test plus three benchmarks. Benchmarks exercise both paths but assert nothing, so nothing currently fails if a future change makes the fast path swallow a needed rebuild — exactly the regression this optimisation invites.

Please add unit tests that assert:

1. A no-new-requirement or duplicate-requirement apply does **not** rebuild metadata (`:172` is the branch to pin).
2. A genuinely new requirement **does** rebuild (`:208`).
3. A failing requirement still produces the same error, identically on both paths.

### Minor — deferred validation is an observable change

On a no-new-requirement apply, `Build` validation and its side effects (`updateSnapshotLog`, initialising `lastUpdatedMS`) are now deferred rather than run eagerly. Nothing is skipped — they still happen at `Commit`/`StagedTable`/`Scan` — but a caller that previously got an invalid-staged-metadata error out of `apply` will now see it later, from `Commit`. Worth a doc note on the exported method so that relocation is intentional and discoverable rather than a surprise.

Benchmark evidence in the body is good (1.50 MB/65 allocs → 1.21 MB/14 for the no-requirement path; 70 → 19 for duplicates; unchanged for genuinely new requirements). The win is real, which is why I'd rather see it protected by tests.

CI green (15/15).

---

>

table/transaction.go
```diff
@@ -208,6 +208,12 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 
 			continue
 		}
+		if current == nil {
+			current, err = stagedMeta.Build()
```
laskoviymishka: Still not a bug, since you check `err` on the next line, but this assignment lands in two different scopes: `current` is the function-level cache, while `err` binds to the loop-body `err` from `key, err := requirementSemanticKey(r)` a few lines up, not the function-level one. A future reorder or rename could quietly have it fall through to the outer `err`. A local keeps the intent obvious:

```go
if current == nil {
    built, err := stagedMeta.Build()
    if err != nil {
        return err
    }
    current = built
}
```

Cheap to do while you're in here; fine to leave if you'd rather not churn it.

table/metadata_builder_bench_test.go
```diff
@@ -198,6 +198,75 @@ func BenchmarkMetadataBuilderCloneAndBuild(b *testing.B) {
...
+			b.ReportAllocs()
+			b.ResetTimer()
+
+			for b.Loop() {
+				if err := txn.apply(nil, nil); err != nil {
+					b.Fatal(err)
+				}
+				metadataBuilderBenchmarkSink = len(txn.meta.snapshotList)
+			}
+			b.ReportMetric(float64(snapshotCount), "snapshot_entries/op")
```
laskoviymishka: Small consistency thing across all three new benchmarks: the others in this file call `ReportMetric` before `ResetTimer` and drop the `/op` suffix (the framework already divides by iteration count). Matching that would keep the numbers lined up with the rest of the file. Pure nit.

laskoviymishka (changes requested):
Thanks, this addresses what I was holding on. Splitting the `MarshalJSON` race fix out to its own commit was the right call, and the new `transaction_internal_test.go` cases lock exactly the contract I was worried about: the no-op and duplicate-requirement paths defer validation (the latent `ErrInvalidMetadata` stays latent through `apply`), a new requirement forces the `Build()` and propagates its error, and the "new requirement after a duplicate" case still surfaces the validation error. The `Commit` doc comment naming Commit/StagedTable/Scan as the validation points makes the deferral a decided contract rather than a side effect. That's what I wanted to see.

I'm happy to approve. Two small things left, both optional and neither blocking. The `current, err = stagedMeta.Build()` assignment still lands `err` in the loop-body scope rather than the function scope (a local error var reads cleaner), and the three new benchmarks call `ReportMetric` after `b.Loop()` with a `/op` suffix while the rest of the file does it before `ResetTimer` without the suffix. Tidy them if you're touching the file again; otherwise this is good to land.

table/transaction.go
laskoviymishka: Still not a bug, since you check `err` on the next line, but this assignment lands in two different scopes: `current` is the function-level cache, while `err` binds to the loop-body `err` from `key, err := requirementSemanticKey(r)` a few lines up, not the function-level one. A future reorder or rename could quietly have it fall through to the outer `err`. A local keeps the intent obvious:

```go
if current == nil {
    built, err := stagedMeta.Build()
    if err != nil {
        return err
    }
    current = built
}
```

Cheap to do while you're in here; fine to leave if you'd rather not churn it.

table/metadata_builder_bench_test.go
laskoviymishka: Small consistency thing across all three new benchmarks: the others in this file call `ReportMetric` before `ResetTimer` and drop the `/op` suffix (the framework already divides by iteration count). Matching that would keep the numbers lined up with the rest of the file. Pure nit.

laskoviymishka (approved):
Thanks, this addresses what I was holding on. Splitting the `MarshalJSON` race fix out to its own commit was the right call, and the new `transaction_internal_test.go` cases lock exactly the contract I was worried about: the no-op and duplicate-requirement paths defer validation (the latent `ErrInvalidMetadata` stays latent through `apply`), a new requirement forces the `Build()` and propagates its error, and the "new requirement after a duplicate" case still surfaces the validation error. The `Commit` doc comment naming Commit/StagedTable/Scan as the validation points makes the deferral a decided contract rather than a side effect. That's what I wanted to see.

I'm happy to approve. Two small things left, both optional and neither blocking. The `current, err = stagedMeta.Build()` assignment still lands `err` in the loop-body scope rather than the function scope (a local error var reads cleaner), and the three new benchmarks call `ReportMetric` after `b.Loop()` with a `/op` suffix while the rest of the file does it before `ResetTimer` without the suffix. Tidy them if you're touching the file again; otherwise this is good to land.

zeroshade (approved):
Both of my previous points are addressed, and I traced the skip predicate's soundness properly this round rather than reasoning about it from the diff. Approving.

## What I verified

**The skip predicate is sound.** `current` is read at exactly one place — `r.Validate(current)` at `table/transaction.go:218` — and never escapes `apply`. The risk isn't the read, it's `Build()`'s side effects on the builder, so I enumerated each:

| `Build()` side effect | consequence of deferring | verdict |
|---|---|---|
| `ensureSchemaIndex`/`ensureSnapshotIndex` + `shared = true` | none — `clone()` already sets `shared=true` on both parent and child (`metadata.go:546-564`) | safe; probe confirms `shared=true` after a skipped apply, so no COW aliasing hazard |
| `updateSnapshotLog()` prunes intermediate entries | runs once later instead of progressively | result-equivalent: a 3-append transaction ends with `SnapshotLogs()` == 1 on both head and reverted |
| `lastUpdatedMS = now()` when 0 | left at 0 until the eventual Build | no external difference; `buildCommonMetadata` stamps it before `common.validate()` |
| `AppendMetadataLog` when `HasChanges()` | fewer duplicate entries | real behaviour change — see below |

No update can be dropped: `t.meta = stagedMeta` at `:242` is unconditional and independent of whether Build ran. No stale metadata can be committed: `Commit` never calls `Build`; it hands `meta.updates` to `doCommit` and the catalog rebuilds from a fresh base via `MetadataBuilderFromBase` + replay + `Build()`.

**Divergence cases I enumerated:** reqs empty / updates non-empty → skip (`SetProperties`, `RemoveProperties`, `UpgradeFormatVersion`, `RollbackToSnapshot` on a transaction-staged branch); reqs non-empty but all duplicates → skip (the common real case — the 2nd+ snapshot producer on one branch, since `requirementSemanticKey` keys `assertRefSnapshotID` by type+ref only); reqs with ≥1 new → build once and reuse.

**Tests are load-bearing, not decorative.** Reverting only the non-test hunk fails `TestTransactionApplyDefersMetadataBuildForNoopRequirements` on both subtests. Mutating head to never build fails `TestTransactionApplyBuildsMetadataForNewRequirement` and both subtests of `TestTransactionApplyPreservesRequirementValidationErrorAfterDuplicate`. Both branches are genuinely pinned, and the `AssertCurrentSchemaID(0)` → `{AssertCurrentSchemaID(0), AssertCurrentSchemaID(1)}` case covers the dangerous shape (same-type-different-value not swallowed as a duplicate).

**Benchmark table is not stale** — every figure reproduces to within exactly +1 alloc on linux/amd64 go1.26.7 vs your darwin/M1 go1.26.3 (claimed 65→14 / 70→19 / 77 unchanged; measured 66→15 / 71→20 / 78 unchanged; byte figures match). The consistent one-alloc offset is a toolchain artifact.

`go test ./table/... -count=1` all pass; `-race` clean; `golangci-lint` 0 issues; CI 15/15.

## Two notes, neither blocking

**1. `table/transaction.go:3267-3269` — the new doc comment misstates the trigger.** It says the build is skipped when an operation "only adds no-op updates or duplicate requirements", but the skip is driven purely by *no new requirements*; whether the updates are no-ops is irrelevant. `SetProperties` passes `nil` reqs and skips the build even when the property change is entirely material. Suggest: "…when the operation adds no requirement that is not already staged."

Also "deferred until Commit, StagedTable, or Scan" isn't exhaustive — `classifyFilesForFilteredDeletions` (`:2630`), `rewriteSingleFile` (`:2883`) and `makePositionDeleteRecordsForFilter` (`:3129`) also build — and `Commit` itself never calls `Build`; validation there happens inside the catalog's `UpdateTableMetadata`.

**2. Undocumented (beneficial) change to `StagedTable()`/`Scan()` metadata.** Because the old code built once per `apply`, a multi-operation transaction accumulated one duplicate `metadata-log` previous-file entry per apply. Measured: 3× `SetProperties` in one transaction gives `PreviousFiles()` length 3 before, **1** after; same for 3× fast append. The new count is the correct one, so this PR incidentally masks a latent pre-existing bug. Committed metadata is unaffected. Worth either a line in the description or, if it should count as a fix, a test asserting `PreviousFiles()` length is independent of the number of applies.

## One thing worth stating for the record

The headline "−77% allocs / −20% bytes" is the microbenchmark in isolation. Measured end-to-end across `NewTransaction → SetProperties → Commit` against a catalog that performs the real `UpdateTableMetadata` replay, on a 10,000-snapshot table, it's **−16.5% to −26.5% allocs and −5% to −8% bytes** (and the 51 allocs saved per skipped apply matches the microbenchmark's 66→15 delta exactly, which is a good cross-check). That's with an in-process fake catalog and zero data files; a real append also writes manifests and does a network round trip, shrinking the share further. Still free, just smaller in context — worth stating so expectations are set correctly.

Also worth a caveat for whoever measures this next: my first *sequential* A/B run showed an apparent +25–33% regression on the new-requirement path at p ≤ 0.041 with byte-identical allocations. Under interleaved measurement it vanished completely. That was machine load, not a real regression — don't re-derive it.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Iceberg Go maintainer.*

table/transaction.go
```diff
@@ -172,47 +172,61 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
 		return errors.New("cannot apply updates to nil metadata")
 	}
 
-	current, err := stagedMeta.Build()
-	if err != nil {
-		return err
-	}
+	// Only new requirement validation needs the immutable metadata view.
+	// Updates can be applied directly to the staged builder, and duplicate
```
zeroshade, line 176: **minor** — StagedTable() metadata-log output changed; PR body claims publish behavior is unchanged

Removing the eager Build() also removes its non-idempotent AppendMetadataLog side effect (table/metadata.go:1378-1384). Three SetProperties calls previously yielded three duplicate previous-file entries from StagedTable(); now they yield one. This is a fix, but it is an observable output change that the PR body's 'the existing cloned builder and publish behavior stay unchanged' denies. Commit() is unaffected because it ships meta.updates to the catalog rather than built metadata.

table/transaction_internal_test.go
```diff
@@ -1266,6 +1267,70 @@ func TestTransactionApplyKeepsRequirementsUnchangedOnUpdateFailure(t *testing.T)
...
+		// after the first apply so this path cannot accidentally rebuild metadata.
+		txn.meta.defaultSpecID = 999
+		require.NoError(t, txn.apply(nil, []Requirement{requirement}))
+		_, err := txn.meta.Build()
+		require.ErrorIs(t, err, ErrInvalidMetadata)
+	})
+}
+
+func TestTransactionStagedTableBuildsOnePreviousMetadataLogEntry(t *testing.T) {
+	txn, _ := createTestTransactionWithMemIO(t, *iceberg.UnpartitionedSpec)
```
zeroshade, line 1295: **minor** — TestTransactionStagedTableBuildsOnePreviousMetadataLogEntry overstates the invariant it pins

The test name asserts a general 'one previous metadata log entry' property, but the underlying non-idempotency of MetadataBuilder.Build() (AppendMetadataLog has no dedupe guard) is untouched. The duplication is only avoided on the no-new-requirement path. A transaction that adds a distinct requirement per operation, or that calls StagedTable() more than once, still accumulates duplicates. Consider renaming to reflect the no-requirement path, or fixing AppendMetadataLog to be idempotent.

table/transaction.go
```diff
@@ -172,47 +172,61 @@ func (t *Transaction) apply(updates []Update, reqs []Requirement) error {
...
-		key, err := requirementSemanticKey(r)
-		if err != nil {
-			return err
-		}
-
-		existing[key] = r
-		stagedReqs = append(stagedReqs, r)
-	}
+	stagedReqs := t.reqs
+	if len(reqs) > 0 {
```
zeroshade, line 187: **nit** — len(reqs)==0 now skips semantic-key computation over existing t.reqs

Previously every apply re-ran requirementSemanticKey over t.reqs, so a requirement that failed to marshal would error even on a no-requirement apply. That early error is now skipped. Unreachable in practice since every requirement in t.reqs had its key computed successfully when it was inserted, so this is behavior-preserving; noting only because it is a silent removal of a check.

zeroshade (approved):
The macOS CI failure is the known-flaky TestWaitForPlanBoundsSlowServerSideCancel in catalog/rest, which this diff does not touch; the deferral itself is behavior-preserving for the commit path and its four new tests all fail under mutation.

Re-review verification: 2 of 2 prior findings confirmed fixed at `3ea41b9` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust).

```text
go build ./... (OK); go vet ./table/ (clean); go test ./table/ -run TestTransaction -count=1 (ok 0.609s); go test -race ./table/ -run TestTransaction -count=1 (ok 2.045s); go test ./table/ -bench BenchmarkTransactionApply -benchtime=10x (all pass; NoRequirements/snapshots=10000 = 15 allocs/op vs NewRequirement path 45-48 allocs/op, confirming the claimed reduction); mutation A (eager Build restored) -> 2 new tests red; mutation B (Build block deleted from new-requirement path) -> 2 new tests red; throwaway probe table/pr1955_probe_test.go (removed)
```

</details>

---

>

PR #1956, opened 2026-08-28
perf(rest): lazy-decode snapshots in commitTableResponse when snapshot-loading-mode=refs

### LINKED ISSUE #1946: perf(rest): lazy-decode snapshots in commitTableResponse when snapshot-loading-mode=refs
### Proposed Change

When a client commits changes via `Transaction.Commit()`, the REST server returns a `commitTableResponse` containing the full table metadata, including all snapshots. Today, iceberg-go eagerly materializes every snapshot even though common commit paths either discard the returned table or only inspect the snapshot referenced by the current or target branch.

For tables with long snapshot histories (1000+ snapshots), parsing and allocating the complete snapshot history can add material latency and memory pressure to each commit. The exact improvement should be established with repository benchmarks rather than assumed.

**Proposed change:**

When `snapshot-loading-mode=refs` is configured, add a parse-time deferred-snapshot path for REST commit responses:

- Intercept the `snapshots` array before the normal metadata parser materializes every `Snapshot`; applying laziness after a full `ParseMetadataBytes` call does not achieve the optimization.
- Retain the original raw snapshot JSON for later materialization.
- Eagerly materialize every snapshot targeted by `refs` (branches and tags), not only `current-snapshot-id`. This keeps `CurrentSnapshot()` and `SnapshotByName()` fast, including commits to non-main branches.
- Synchronously validate and index the deferred snapshot data sufficiently to preserve the current parse-time error contract. Deferred decoding must not turn malformed metadata into a silent "snapshot not found" result.
- On `SnapshotByID(id)`, consult the eager referenced-snapshot index first and materialize full history only on a miss.
- Materialize full history when `Snapshots()` is called.
- Use pointer-owned, thread-safe lazy state so metadata copies do not copy a used lock or `sync.Once` value.

The implementation should remain internal to metadata parsing. REST code should select the parsing mode for the raw commit response rather than mutate an already parsed metadata object.

**Correctness requirements:**

- `CurrentSnapshot()` and lookups of any named branch or tag must not materialize unreferenced history.
- `Snapshots()` and historical `SnapshotByID()` lookups must return the complete snapshot history from the commit response.
- Concurrent access must materialize history at most once and must be race-free.
- JSON serialization must preserve the complete snapshot list even before materialization; it must never serialize only the eager subset.
- Metadata equality, cloning, builders, format-version upgrades, and other operations that require complete metadata must preserve their existing logical behavior.
- Existing snapshot validation must remain synchronous, or the design must provide an explicit error-returning path. Lazy getters must not silently swallow deferred decode errors.
- Default behavior remains unchanged when `snapshot-loading-mode` is unset or set to `all`.

**Performance validation:**

Add benchmarks using realistic snapshots with non-empty summaries at multiple history sizes (for example 10, 100, 1,000, and 10,000 snapshots). At minimum, measure time and allocations for:

- the existing eager commit-response parse;
- deferred parse without snapshot access;
- `CurrentSnapshot()` and named-ref access;
- the first historical `SnapshotByID()` lookup;
- the first `Snapshots()` call;
- repeated access after materialization.

The intended result is that commit-and-exit and referenced-snapshot-only paths avoid constructing historical `Snapshot` objects. This remains an `O(response size)` client-side scan and does not reduce server serialization or network transfer; eliminating those costs would require a REST protocol change.

**Backward compatibility:**

- No breaking public API changes are required.
- Full snapshot history remains available on demand; metadata must never expose silent truncation.
- The optimization is initially gated by `snapshot-loading-mode=refs` for compatibility and rollout.

**Related:**

- #1748 — added the `snapshot-loading-mode` catalog property
- #1749 — implemented `?snapshots=refs` on `LoadTable` (merged)
- #1792 — lazy-load unreferenced snapshots omitted from `LoadTable(...?snapshots=refs)` responses

#1792 is related but distinct: it requires a follow-up network request and context-aware catalog behavior, while this issue concerns deferred local materialization of snapshot bytes already present in a commit response.

**Spec reference:**

No spec change is required for this client-side optimization. A future protocol-level commit-response projection could additionally reduce response size and server serialization, but is outside this issue.

**Willingness to contribute**

- [x] I can contribute this improvement/feature independently
- [ ] I would be willing to contribute this improvement/feature with guidance from the Iceberg community
- [ ] I cannot contribute this improvement/feature at this time

### Proposal document

_Not applicable — client-side optimization._

### Specifications

- [x] Table
- [ ] View
- [x] REST
- [ ] Puffin
- [ ] Encryption
- [ ] Other

technicolorbeat:
## Summary

Avoid materializing unreferenced snapshot history while parsing REST commit responses when the catalog is configured with `snapshot-loading-mode=refs`.

Fixes #1946.

## Motivation

The REST update-table endpoint returns complete table metadata. `Transaction.Commit()` parses that response before returning the committed table, and the existing path decodes every entry in `snapshots` even when `snapshot-loading-mode=refs` means the caller normally needs only the current snapshot and named-ref targets.

Each historical snapshot decode constructs the snapshot, summary map, strings, and—for v1—manifest-location slice. For tables with long retained history, this makes client-side commit-response CPU and allocation cost scale with snapshots the caller may never read.

This change makes `refs` effective on that path. It does not reduce response size, network transfer, snapshot-log decoding, or server work; it avoids constructing unused historical snapshot object graphs.

## Design

`Catalog.parseCommitMetadata` selects the deferred parser only for REST commit responses in `snapshot-loading-mode=refs`. Load-table parsing and default/`all` mode remain eager.

The deferred parser:

1. Runs the normal metadata normalization and version-specific decode while retaining `snapshots` as `json.RawMessage`.
2. Scans the snapshot array once, recording element byte ranges and snapshot IDs.
3. Synchronously validates every snapshot so malformed historical metadata remains a parse-time error. The lightweight validators preserve the eager decoder's acceptance semantics, including JSON `null` and last-key-wins duplicate fields.
4. Materializes the current snapshot and named-ref targets. Other snapshots remain indexed raw JSON.

Access behavior is:

| Access | Behavior |
|---|---|
| Referenced snapshot | Returned from the eager subset |
| Historical `SnapshotByID` | Decodes that indexed entry once |
| `Snapshots`, serialization, equality, conversion, or `MetadataBuilderFromBase` | Materializes the complete array once |

Full materialization clears the raw array, entry boundaries, and deferred ID map, avoiding retention of both representations. Per-entry and full-array `sync.Once` values prevent duplicate decoding; an `RWMutex` protects the transition and prevents historical lookup from racing with index release.

The public `Metadata` API and metadata semantics are unchanged. The custom marshal projection places the top-level `snapshots` key after the embedded metadata fields, so serialized JSON key order changes; JSON content and round-trip behavior are preserved.

## Benchmarks

Measured on an Apple M1 Max after rebasing implementation commit `876b8d8` onto `aa76a28`, the latest `origin/main` commit at measurement time. The eager baseline was run from an isolated worktree at `aa76a28`; the deferred path was run from `876b8d8`. The current branch adds only the empty CI-retry commit `9e7e1bd` on top. Both benchmark sides use the same payload generator and loop. Values are medians of five runs.

```text
# clean aa76a28 worktree, using the same test-only harness
go test ./catalog/rest -run '^$' \
  -bench 'BenchmarkBaseline(FullParse|ParseThenMetadataBuilder)$' \
  -benchtime=1s -benchmem -count=5

# rebased feature implementation 876b8d8
go test ./catalog/rest -run '^$' \
  -bench 'BenchmarkStage(DeferredParse|ParseThenMetadataBuilder)$' \
  -benchtime=1s -benchmem -count=5
```

### Commit-response parse path

| Metadata | Eager parse | Deferred/ref parse | Time | Allocations | Allocated bytes |
|---:|---:|---:|---:|---:|---:|
| 7,056 B | 0.305 ms | 0.222 ms | -27.2% | -41.1% | -14.3% |
| 59,078 B | 2.362 ms | 1.535 ms | -35.0% | -71.1% | -25.1% |
| 582,880 B | 23.270 ms | 14.684 ms | -36.9% | -75.9% | -25.9% |
| 1,999,830 B | 79.229 ms | 49.747 ms | -37.2% | -76.3% | -29.9% |
| 5,856,882 B | 229.716 ms | 145.963 ms | -36.5% | -76.4% | -30.6% |

Small-payload wall time is noisier; allocation counts were stable across repetitions. At approximately 2 MB, deferred parsing uses `41,407 allocs/op` and `7,649,741 B/op`, versus `174,653 allocs/op` and `10,913,526 B/op` for eager parsing.

### Immediate subsequent transaction

`MetadataBuilderFromBase` requires complete history today. Creating another transaction immediately after parsing therefore pays both deferred indexing and full decoding, making this path worse than eager parsing:

| Metadata | Eager parse + builder | Deferred parse + builder | Time | Allocations | Allocated bytes |
|---:|---:|---:|---:|---:|---:|
| 1,999,830 B | 79.397 ms | 97.964 ms | +23.4% | +20.7% | +38.9% |
| 5,856,882 B | 231.899 ms | 287.486 ms | +24.0% | +20.7% | +38.5% |

This PR optimizes the common `refs` commit-response path when the returned table is consumed without immediately starting another write. It is not a strict improvement for repeated writes on the same returned `Table`; carrying deferred state into `MetadataBuilder` is a separate follow-up opportunity.

The retained-heap benchmark also verifies that full materialization releases deferred indexing state:

| Metadata | Retaining deferred index | Releasing deferred index | Reduction |
|---:|---:|---:|---:|
| 7,056 B | 7,648 B | 5,736 B | 25.0% |
| 59,078 B | 68,048 B | 49,272 B | 27.6% |
| 582,880 B | 667,032 B | 474,392 B | 28.9% |
| 1,999,830 B | 2,277,304 B | 1,679,080 B | 26.3% |
| 5,856,882 B | 6,700,568 B | 4,881,240 B | 27.2% |

## Testing

- `go test -race ./...`
- v1/v2/v3 parsing, value/pointer serialization, and round trips
- eager/deferred validation parity for historical snapshots
- selective lookup and complete materialization
- defensive copies, deferred-state release, and concurrent lookup/materialization

### DIFF
--- a/catalog/rest/lazy_snapshot_bench_test.go
+++ b/catalog/rest/lazy_snapshot_bench_test.go
@@ -140,6 +140,44 @@ func BenchmarkStageDeferredAllSnapshots(b *testing.B) {
 	}
 }
 
+// BenchmarkStageParseThenMetadataBuilder measures the path taken when a caller
+// starts another transaction from the Table returned by Commit. Builder
+// creation requires complete snapshot history, so this exposes the cost of
+// indexing deferred snapshots and then fully materializing them.
+func BenchmarkStageParseThenMetadataBuilder(b *testing.B) {
+	for _, n := range benchSnapshotCounts {
+		body := makeTableResponseWithSnapshots(n)
+		rawMeta := extractMetadata(b, body)
+		for _, mode := range []struct {
+			name  string
+			parse func([]byte) (table.Metadata, error)
+		}{
+			{name: "eager", parse: table.ParseMetadataBytes},
+			{name: "deferred", parse: table.ParseMetadataBytesDeferredSnapshots},
+		} {
+			b.Run(metadataBenchmarkName(rawMeta)+"/"+mode.name, func(b *testing.B) {
+				b.ResetTimer()
+				b.ReportAllocs()
+				b.ReportMetric(float64(len(rawMeta)), "metadata-bytes")
+				b.ReportMetric(float64(n), "snapshots")
+				for range b.N {
+					meta, err := mode.parse(rawMeta)
+					if err != nil {
+						b.Fatal(err)
+					}
+					builder, err := table.MetadataBuilderFromBase(meta, "")
+					if err != nil {
+						b.Fatal(err)
+					}
+					if builder == nil {
+						b.Fatal("missing metadata builder")
+					}
+				}
+			})
+		}
+	}
+}
+
 // BenchmarkStageDeferredRetainedHeap measures the live heap held by one
 // deferred metadata object before and after full snapshot materialization.
 // Run with -benchtime=1x: retained-heap measurements describe one object and
--- a/catalog/rest/lazy_snapshot_bench_test.go
+++ b/catalog/rest/lazy_snapshot_bench_test.go
@@ -1,0 +1,191 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package rest
+
+import (
+	"encoding/json"
+	"fmt"
+	"runtime"
+	"testing"
+
+	"github.com/apache/iceberg-go/table"
+	"github.com/stretchr/testify/require"
+)
+
+// These counts generate metadata payloads of approximately 7 KiB, 58 KiB,
+// 569 KiB, 2 MB, and 5.6 MiB. Benchmark names and metrics report actual bytes.
+var benchSnapshotCounts = []int64{10, 100, 1000, 3418, 10000}
+
+func extractMetadata(tb testing.TB, body []byte) json.RawMessage {
+	tb.Helper()
+	var top map[string]json.RawMessage
+	require.NoError(tb, json.Unmarshal(body, &top))
+
+	return top["metadata"]
+}
+
+func metadataBenchmarkName(raw json.RawMessage) string {
+	return fmt.Sprintf("metadata=%d-bytes", len(raw))
+}
+
+// BenchmarkStageFullParse measures table.ParseMetadataBytes, the call the REST client
+// actually makes on both LoadTable and CommitTable responses.
+func BenchmarkStageFullParse(b *testing.B) {
+	for _, n := range benchSnapshotCounts {
+		body := makeTableResponseWithSnapshots(n)
+		rawMeta := extractMetadata(b, body)
+		b.Run(metadataBenchmarkName(rawMeta), func(b *testing.B) {
+			b.ResetTimer()
+			b.ReportAllocs()
+			b.ReportMetric(float64(len(rawMeta)), "metadata-bytes")
+			b.ReportMetric(float64(n), "snapshots")
+			for range b.N {
+				meta, err := table.ParseMetadataBytes(rawMeta)
+				if err != nil {
+					b.Fatal(err)
+				}
+				if meta.CurrentSnapshot() == nil {
+					b.Fatal("missing current snapshot")
+				}
+			}
+		})
+	}
+}
+
+// BenchmarkStageDeferredParse measures the parse-time lazy materialization
+// path used for CommitTable responses when snapshot-loading-mode is refs.
+func BenchmarkStageDeferredParse(b *testing.B) {
+	for _, n := range benchSnapshotCounts {
+		body := makeTableResponseWithSnapshots(n)
+		rawMeta := extractMetadata(b, body)
+		b.Run(metadataBenchmarkName(rawMeta), func(b *testing.B) {
+			b.ResetTimer()
+			b.ReportAllocs()
+			b.ReportMetric(float64(len(rawMeta)), "metadata-bytes")
+			b.ReportMetric(float64(n), "snapshots")
+			for range b.N {
+				meta, err := table.ParseMetadataBytesDeferredSnapshots(rawMeta)
+				if err != nil {
+					b.Fatal(err)
+				}
+				if meta.CurrentSnapshot() == nil {
+					b.Fatal("missing current snapshot")
+				}
+			}
+		})
+	}
+}
+
+// BenchmarkStageDeferredHistoricalLookup measures parsing followed by a lookup
+// of one unreferenced historical snapshot. The indexed lazy path should decode
+// only that snapshot rather than materializing the complete history.
+func BenchmarkStageDeferredHistoricalLookup(b *testing.B) {
+	for _, n := range benchSnapshotCounts {
+		body := makeTableResponseWithSnapshots(n)
+		rawMeta := extractMetadata(b, body)
+		b.Run(metadataBenchmarkName(rawMeta), func(b *testing.B) {
+			b.ResetTimer()
+			b.ReportAllocs()
+			b.ReportMetric(float64(len(rawMeta)), "metadata-bytes")
+			b.ReportMetric(float64(n), "snapshots")
+			for range b.N {
+				meta, err := table.ParseMetadataBytesDeferredSnapshots(rawMeta)
+				if err != nil {
+					b.Fatal(err)
+				}
+				if meta.SnapshotByID(0) == nil {
+					b.Fatal("missing historical snapshot")
+				}
+			}
+		})
+	}
+}
+
+// BenchmarkStageDeferredAllSnapshots measures the intentional fallback that
+// parses and then materializes the complete snapshot history.
+func BenchmarkStageDeferredAllSnapshots(b *testing.B) {
+	for _, n := range benchSnapshotCounts {
+		body := makeTableResponseWithSnapshots(n)
+		rawMeta := extractMetadata(b, body)
+		b.Run(metadataBenchmarkName(rawMeta), func(b *testing.B) {
+			b.ResetTimer()
+			b.ReportAllocs()
+			b.ReportMetric(float64(len(rawMeta)), "metadata-bytes")
+			b.ReportMetric(float64(n), "snapshots")
+			for range b.N {
+				meta, err := table.ParseMetadataBytesDeferredSnapshots(rawMeta)
+				if err != nil {
+					b.Fatal(err)
+				}
+				if len(meta.Snapshots()) != int(n) {
+					b.Fatal("incomplete snapshot history")
+				}
+			}
+		})
+	}
+}
+
+// BenchmarkStageDeferredRetainedHeap measures the live heap held by one
+// deferred metadata object before and after full snapshot materialization.
+// Run with -benchtime=1x: retained-heap measurements describe one object and
+// cannot be meaningfully averaged across b.N independent lifecycles.
+func BenchmarkStageDeferredRetainedHeap(b *testing.B) {
+	if b.N != 1 {
+		b.Skip("retained-heap benchmark requires -benchtime=1x")
+	}
+
+	for _, n := range benchSnapshotCounts {
+		body := makeTableResponseWithSnapshots(n)
+		rawMeta := extractMetadata(b, body)
+		b.Run(metadataBenchmarkName(rawMeta), func(b *testing.B) {
+			meta, err := table.ParseMetadataBytesDeferredSnapshots(rawMeta)
+			if err != nil {
+				b.Fatal(err)
+			}
+
+			runtime.GC()
+			var before runtime.MemStats
+			runtime.ReadMemStats(&before)
+
+			materializeSnapshots(b, meta, n)
+			runtime.GC()
+			var after runtime.MemStats
+			runtime.ReadMemStats(&after)
+
+			b.ReportMetric(float64(len(rawMeta)), "metadata-bytes")
+			b.ReportMetric(float64(before.HeapAlloc), "heap-before-bytes")
+			b.ReportMetric(float64(after.HeapAlloc), "heap-after-bytes")
+			delta := int64(after.HeapAlloc) - int64(before.HeapAlloc)
+			b.ReportMetric(float64(delta), "heap-delta-bytes")
+			b.ReportMetric(float64(delta)/float64(len(rawMeta)), "heap-growth-ratio")
+			b.ReportMetric(0, "ns/op")
+			runtime.KeepAlive(meta)
+			runtime.KeepAlive(rawMeta)
+			runtime.KeepAlive(body)
+		})
+	}
+}
+
+// Keep the defensive copy returned by Snapshots out of the post-call heap
+// sample so it measures metadata-owned state rather than caller-owned memory.
+func materializeSnapshots(b *testing.B, meta table.Metadata, expected int64) {
+	b.Helper()
+	if len(meta.Snapshots()) != int(expected) {
+		b.Fatal("incomplete snapshot history")
+	}
+}
--- a/catalog/rest/rest.go
+++ b/catalog/rest/rest.go
@@ -179,18 +179,6 @@ type identifier struct {
 type commitTableResponse struct {
 	MetadataLoc string          `json:"metadata-location"`
 	RawMetadata json.RawMessage `json:"metadata"`
-	Metadata    table.Metadata  `json:"-"`
-}
-
-func (t *commitTableResponse) UnmarshalJSON(b []byte) (err error) {
-	type Alias commitTableResponse
-	if err = json.Unmarshal(b, (*Alias)(t)); err != nil {
-		return err
-	}
-
-	t.Metadata, err = table.ParseMetadataBytes(t.RawMetadata)
-
-	return err
 }
 
 // StorageCredential carries REST-vended storage credentials scoped to matching
@@ -1645,8 +1633,20 @@ func (r *Catalog) CommitTable(ctx context.Context, ident table.Identifier, requi
 	if err != nil {
 		return nil, "", err
 	}
+	metadata, err := r.parseCommitMetadata(ret.RawMetadata)
+	if err != nil {
+		return nil, "", err
+	}
 
-	return ret.Metadata, ret.MetadataLoc, nil
+	return metadata, ret.MetadataLoc, nil
+}
+
+func (r *Catalog) parseCommitMetadata(raw json.RawMessage) (table.Metadata, error) {
+	if r.snapshotMode == SnapshotModeRefs {
+		return table.ParseMetadataBytesDeferredSnapshots(raw)
+	}
+
+	return table.ParseMetadataBytes(raw)
 }
 
 // CommitTransaction atomically commits changes to multiple tables in a
@@ -1853,11 +1853,15 @@ func (r *Catalog) UpdateTable(ctx context.Context, ident table.Identifier, requi
 	if err != nil {
 		return nil, err
 	}
+	metadata, err := r.parseCommitMetadata(ret.RawMetadata)
+	if err != nil {
+		return nil, err
+	}
 
 	config := maps.Clone(r.props)
-	maps.Copy(config, ret.Metadata.Properties())
+	maps.Copy(config, metadata.Properties())
 
-	return r.tableFromResponse(ctx, ident, ret.Metadata, ret.MetadataLoc, config, config, false)
+	return r.tableFromResponse(ctx, ident, metadata, ret.MetadataLoc, config, config, false)
 }
 
 func (r *Catalog) DropTable(ctx context.Context, identifier table.Identifier) error {
--- a/table/deferred_snapshots.go
+++ b/table/deferred_snapshots.go
@@ -101,11 +101,13 @@ func (s *deferredSnapshotState) snapshotByID(id int64) (*Snapshot, error) {
 		return snapshot, nil
 	}
 	entryIndex, ok := s.byID[id]
-	s.mu.RUnlock()
 	if !ok {
+		s.mu.RUnlock()
+
 		return nil, nil
 	}
 	entry := &s.entries[entryIndex]
+	s.mu.RUnlock()
 
 	entry.once.Do(func() {
 		s.mu.RLock()
--- a/table/deferred_snapshots.go
+++ b/table/deferred_snapshots.go
@@ -168,10 +168,15 @@ func validateStringArray(raw json.RawMessage) error {
 		if i == len(raw)-1 {
 			return nil
 		}
-		if raw[i] != '"' {
+		switch {
+		case raw[i] == '"':
+			i = scanJSONString(raw, i)
+		case hasJSONNullAt(raw, i):
+			// Match encoding/json's []string behavior: null decodes to "".
+			i += len("null")
+		default:
 			return errors.New("cannot unmarshal manifest location into string")
 		}
-		i = scanJSONString(raw, i)
 		i = skipJSONSpace(raw, i)
 		if i < len(raw)-1 && raw[i] == ',' {
 			i++
@@ -194,10 +199,15 @@ func validateSummaryObject(raw json.RawMessage) error {
 	if raw[0] != '{' || raw[len(raw)-1] != '}' {
 		return errors.New("cannot unmarshal summary into map[string]string")
 	}
+	// Summary.UnmarshalJSON decodes through map[string]string. Preserve its
+	// acceptance contract here: null becomes "", and duplicate keys use the
+	// last value. In particular, only the final operation value determines
+	// whether ErrInvalidOperation is returned.
+	operationSeen, operationEmpty := false, false
 	for i := 1; i < len(raw)-1; {
 		i = skipJSONSpace(raw, i)
 		if i == len(raw)-1 {
-			return nil
+			break
 		}
 		if raw[i] != '"' {
 			return errors.New("invalid summary key")
@@ -210,13 +220,27 @@ func validateSummaryObject(raw json.RawMessage) error {
 			return errors.New("invalid summary object")
 		}
 		i = skipJSONSpace(raw, i+1)
-		if i >= len(raw)-1 || raw[i] != '"' {
+		if i >= len(raw)-1 {
 			return errors.New("cannot unmarshal summary value into string")
 		}
-		valueStart := i
-		i = scanJSONString(raw, i)
-		if summaryKeyIsOperation(key) && i == valueStart+2 {
-			return fmt.Errorf("%w: found empty operation", ErrInvalidOperation)
+
+		valueEmpty := false
+		switch {
+		case raw[i] == '"':
+			valueStart := i
+			i = scanJSONString(raw, i)
+			valueEmpty = i == valueStart+2
+		case hasJSONNullAt(raw, i):
+			// Match encoding/json's map[string]string behavior: null decodes
+			// to "". A final null operation is therefore invalid.
+			i += len("null")
+			valueEmpty = true
+		default:
+			return errors.New("cannot unmarshal summary value into string")
+		}
+		if summaryKeyIsOperation(key) {
+			operationSeen = true
+			operationEmpty = valueEmpty
 		}
 		i = skipJSONSpace(raw, i)
 		if i < len(raw)-1 && raw[i] == ',' {
@@ -227,9 +251,18 @@ func validateSummaryObject(raw json.RawMessage) error {
 		if i != len(raw)-1 {
 			return errors.New("invalid summary object")
 		}
+
+		break
+	}
+	if operationSeen && operationEmpty {
+		return fmt.Errorf("%w: found empty operation", ErrInvalidOperation)
 	}
 
 	return nil
+}
+
+func hasJSONNullAt(raw []byte, i int) bool {
+	return i+len("null") <= len(raw) && bytes.Equal(raw[i:i+len("null")], []byte("null"))
 }
 
 func summaryKeyIsOperation(raw []byte) bool {
@@ -276,14 +309,14 @@ type rawJSONSpan struct {
 }
 
 func splitJSONArray(raw []byte) ([]rawJSONSpan, error) {
-	raw = bytes.TrimSpace(raw)
-	if len(raw) < 2 || raw[0] != '[' || raw[len(raw)-1] != ']' {
+	arrayStart, arrayEnd := trimJSONSpan(raw, 0, len(raw))
+	if arrayEnd-arrayStart < 2 || raw[arrayStart] != '[' || raw[arrayEnd-1] != ']' {
 		return nil, errors.New("expected JSON array")
 	}
 
 	elements := make([]rawJSONSpan, 0)
-	start, depth := 1, 0
-	for i := 1; i < len(raw)-1; i++ {
+	start, depth := arrayStart+1, 0
+	for i := start; i < arrayEnd-1; i++ {
 		switch raw[i] {
 		case '"':
 			i = scanJSONString(raw, i) - 1
@@ -299,7 +332,7 @@ func splitJSONArray(raw []byte) ([]rawJSONSpan, error) {
 			}
 		}
 	}
-	elementStart, elementEnd := trimJSONSpan(raw, start, len(raw)-1)
+	elementStart, elementEnd := trimJSONSpan(raw, start, arrayEnd-1)
 	if elementStart < elementEnd {
 		elements = append(elements, rawJSONSpan{start: elementStart, end: elementEnd})
 	}
@@ -329,6 +362,10 @@ func parseNormalizedMetadataBytesDeferredSnapshots(normalized []byte, formatVers
 		}
 
 		return nil, fmt.Errorf("%w: %w", ErrInvalidMetadata, err)
+	}
+	if metadata.Version() != formatVersion {
+		return nil, fmt.Errorf("%w: preflight selected version %d, decoded version %d",
+			ErrInvalidMetadataFormatVersion, formatVersion, metadata.Version())
 	}
 
 	common := commonMetadataOf(metadata)
--- a/table/deferred_snapshots.go
+++ b/table/deferred_snapshots.go
@@ -1,0 +1,493 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+	"sync"
+)
+
+// deferredSnapshotState owns the raw snapshot array and the materialized
+// collection. It is pointer-owned because commonMetadata is copied by format
+// upgrades and builders; copying a used sync.Once would be unsafe.
+type deferredSnapshotState struct {
+	raw     json.RawMessage
+	entries []deferredSnapshotEntry
+	byID    map[int64]int
+
+	allOnce   sync.Once
+	mu        sync.RWMutex
+	snapshots []Snapshot
+	index     *snapshotIndexData
+	err       error
+}
+
+type deferredSnapshotEntry struct {
+	start int
+	end   int
+
+	once     sync.Once
+	snapshot Snapshot
+	err      error
+}
+
+func (s *deferredSnapshotState) load() ([]Snapshot, *snapshotIndexData, error) {
+	s.allOnce.Do(func() {
+		s.mu.RLock()
+		raw := s.raw
+		s.mu.RUnlock()
+
+		var snapshots []Snapshot
+		if err := json.Unmarshal(raw, &snapshots); err != nil {
+			s.mu.Lock()
+			s.err = fmt.Errorf("%w: deferred snapshots: %w", ErrInvalidMetadata, err)
+			s.mu.Unlock()
+
+			return
+		}
+
+		s.mu.Lock()
+		s.snapshots = snapshots
+		s.index = buildSnapshotIndex(snapshots)
+		s.raw = nil
+		s.entries = nil
+		s.byID = nil
+		s.mu.Unlock()
+	})
+
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return s.snapshots, s.index, s.err
+}
+
+func (s *deferredSnapshotState) snapshotByID(id int64) (*Snapshot, error) {
+	s.mu.RLock()
+	if s.err != nil {
+		err := s.err
+		s.mu.RUnlock()
+
+		return nil, err
+	}
+	if s.snapshots != nil {
+		i, ok := snapshotIndexPosition(s.index, s.snapshots, id)
+		if !ok {
+			s.mu.RUnlock()
+
+			return nil, nil
+		}
+		snapshot := cloneSnapshotPtr(&s.snapshots[i])
+		s.mu.RUnlock()
+
+		return snapshot, nil
+	}
+	entryIndex, ok := s.byID[id]
+	s.mu.RUnlock()
+	if !ok {
+		return nil, nil
+	}
+	entry := &s.entries[entryIndex]
+
+	entry.once.Do(func() {
+		s.mu.RLock()
+		if s.snapshots != nil {
+			i, found := snapshotIndexPosition(s.index, s.snapshots, id)
+			if found {
+				entry.snapshot = s.snapshots[i]
+			}
+			s.mu.RUnlock()
+
+			return
+		}
+		raw := s.raw[entry.start:entry.end]
+		s.mu.RUnlock()
+
+		if err := json.Unmarshal(raw, &entry.snapshot); err != nil {
+			entry.err = fmt.Errorf("%w: deferred snapshot %d: %w", ErrInvalidMetadata, id, err)
+		}
+	})
+	if entry.err != nil {
+		return nil, entry.err
+	}
+
+	return cloneSnapshotPtr(&entry.snapshot), nil
+}
+
+type deferredSnapshotFields struct {
+	SnapshotID        *int64          `json:"snapshot-id"`
+	ParentSnapshotID  *int64          `json:"parent-snapshot-id,omitempty"`
+	SequenceNumber    int64           `json:"sequence-number"`
+	TimestampMs       *int64          `json:"timestamp-ms"`
+	ManifestList      string          `json:"manifest-list,omitempty"`
+	ManifestLocations json.RawMessage `json:"manifests,omitempty"`
+	Summary           json.RawMessage `json:"summary,omitempty"`
+	SchemaID          *int            `json:"schema-id,omitempty"`
+	FirstRowID        *int64          `json:"first-row-id,omitempty"`
+	AddedRows         *int64          `json:"added-rows,omitempty"`
+}
+
+func (s deferredSnapshotFields) validateHeavyFields() error {
+	if err := validateStringArray(s.ManifestLocations); err != nil {
+		return err
+	}
+
+	return validateSummaryObject(s.Summary)
+}
+
+func validateStringArray(raw json.RawMessage) error {
+	raw = bytes.TrimSpace(raw)
+	if len(raw) == 0 || bytes.Equal(raw, []byte("null")) {
+		return nil
+	}
+	if raw[0] != '[' || raw[len(raw)-1] != ']' {
+		return errors.New("cannot unmarshal manifests into []string")
+	}
+	for i := 1; i < len(raw)-1; {
+		i = skipJSONSpace(raw, i)
+		if i == len(raw)-1 {
+			return nil
+		}
+		if raw[i] != '"' {
+			return errors.New("cannot unmarshal manifest location into string")
+		}
+		i = scanJSONString(raw, i)
+		i = skipJSONSpace(raw, i)
+		if i < len(raw)-1 && raw[i] == ',' {
+			i++
+
+			continue
+		}
+		if i != len(raw)-1 {
+			return errors.New("invalid manifests array")
+		}
+	}
+
+	return nil
+}
+
+func validateSummaryObject(raw json.RawMessage) error {
+	raw = bytes.TrimSpace(raw)
+	if len(raw) == 0 || bytes.Equal(raw, []byte("null")) {
+		return nil
+	}
+	if raw[0] != '{' || raw[len(raw)-1] != '}' {
+		return errors.New("cannot unmarshal summary into map[string]string")
+	}
+	for i := 1; i < len(raw)-1; {
+		i = skipJSONSpace(raw, i)
+		if i == len(raw)-1 {
+			return nil
+		}
+		if raw[i] != '"' {
+			return errors.New("invalid summary key")
+		}
+		keyStart := i
+		i = scanJSONString(raw, i)
+		key := raw[keyStart:i]
+		i = skipJSONSpace(raw, i)
+		if i >= len(raw)-1 || raw[i] != ':' {
+			return errors.New("invalid summary object")
+		}
+		i = skipJSONSpace(raw, i+1)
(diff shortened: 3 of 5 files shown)


zeroshade (changes requested):
The lazy-decode design here is sound — I have no findings against the mechanism itself. What's blocking is mechanical: the branch has conflicts and most commits are missing DCO sign-off.

### The substance is clean

Taking the risky parts in turn:

- **Mode resolution.** Absent property means empty/eager, `all` means eager, `refs` means deferred, and unknown values are rejected rather than silently treated as one or the other. The defaults → client → server override chain resolves contradictory inputs deterministically. That's the right shape.
- **A deferred snapshot later needed** resolves through a per-entry `sync.Once` in `SnapshotByID` with no nil-deref path, so a late access gets the snapshot or a clean error, never a panic.
- **No observable loss.** `Snapshots()` fully materialises every snapshot present in the response, so callers iterating it don't silently see fewer. Snapshots the server omits in refs mode are absent by design, which is the point of the mode.
- **Unknown snapshot fields** keep the existing typed-decode behaviour, so round-tripping isn't degraded.
- Concurrency handling and the benchmark evidence both look good.

I also went looking for an aliasing bug — retained raw bytes pointing at the caller's buffer — and there isn't one: `json.RawMessage.UnmarshalJSON` copies (`*m = append((*m)[0:0], data...)`), so the deferred state owns its bytes. Mentioning it because it's the natural worry with deferred decoding and it's genuinely handled.

### Major — DCO sign-off missing on most commits

`CONTRIBUTING.md` requires a `Signed-off-by` trailer on every commit. Current state:

| Commit | Sign-off |
|---|---|
| `5f31081` perf(rest): defer unreferenced commit snapshots | yes |
| `763d567` fix(table): synchronize deferred snapshot index cleanup | **no** |
| `d6986d2` merge commit | **no** |
| `714bbc4` fix(table): preserve deferred snapshots in metadata builder | **no** |

Worth knowing: this repo has no DCO check among its 15 status checks, so all-green CI does **not** mean sign-off is satisfied. It's an ASF requirement regardless and will hold up the merge.

### Minor — merge commit, and the branch is conflicted

`mergeStateStatus` is `DIRTY` — real conflicts with base, since main has moved a fair bit. The branch also carries merge commit `d6986d2` rather than being rebased.

**All three of the above collapse into one action:** rebase onto current `main`, dropping the merge commit, and add `-s` to each commit as you go (`git rebase -i --signoff origin/main` handles sign-off in the same pass). Then push and let CI re-run.

Thanks for this one — the refs-mode deferral is a nice win and the careful `sync.Once` handling on late access is exactly right.

---

>

table/metadata.go
```diff
@@ -2184,10 +2214,37 @@ func (c *commonMetadata) SnapshotByID(id int64) *Snapshot {
...
+		}
+
+		return snapshot
+	}
 
 	return nil
 }
 
+func (c *commonMetadata) allSnapshots() []Snapshot {
+	snapshots, err := c.snapshotsForMarshal()
```
laskoviymishka: I'd surface this error rather than collapse it to nil. `allSnapshots()` throws away the `snapshotsForMarshal()` error and returns nil, and that feeds both `Snapshots()` and `Equals()`. So a snapshot that clears the light `validateHeavyFields` check at parse time but fails a full unmarshal on first access would make `Snapshots()` report an empty history, and `Equals()` can return true between two metadata objects that both failed to load (`SliceEqualHelper(nil, nil)`). In a commit/OCC retry loop that false positive could suppress a re-read we actually needed.

`SnapshotByID` a few lines up has the milder version of the same thing: a decode error becomes nil, i.e. "not found", so a time-travel or ancestry walk truncates and looks successful.

I'd either propagate the error (change `allSnapshots` to return `([]Snapshot, error)` and thread it through `Equals`/`Snapshots`), or panic with a clear message if we're treating a post-parse decode failure as programmer error given validation already passed. At minimum I don't think we can drop it silently. wdyt?

table/deferred_snapshots.go
```diff
@@ -0,0 +1,502 @@
...
+		return nil, nil
+	}
+	entry := &s.entries[entryIndex]
+	s.mu.RUnlock()
+
+	entry.once.Do(func() {
+		s.mu.RLock()
+		if s.snapshots != nil {
+			i, found := snapshotIndexPosition(s.index, s.snapshots, id)
+			if found {
```
laskoviymishka: If `s.snapshots != nil` but `snapshotIndexPosition` returns `found == false`, we fall out of the `once.Do` with `entry.snapshot` still zero and `entry.err` nil, so the caller gets a non-nil `*Snapshot{SnapshotID: 0}` back instead of nil. `SnapshotByID`'s contract is nil-means-not-found, so `if snap := meta.SnapshotByID(id); snap != nil` would read a zero-value snapshot as real.

It's unreachable today since `load()` and `prepareDeferredSnapshots` work off the same bytes, but that invariant is load-bearing and easy to break. I'd add an `else` that sets `entry.err`, or track a `found` bool on the entry and return `nil, nil` when it's false.

table/metadata.go
```diff
@@ -346,7 +346,11 @@ func MetadataBuilderFromBase(metadata Metadata, currentFileLocation string) (*Me
 		if b.props == nil {
 			b.props = iceberg.Properties{}
 		}
-		b.snapshotList = cloneSnapshots(common.SnapshotList)
+		snapshots, err := common.snapshotsForMarshal()
+		if err != nil {
+			return nil, err
+		}
+		b.snapshotList = cloneSnapshots(snapshots)
```
laskoviymishka: One to flag a few lines down (outside this hunk): after `snapshotsForMarshal()` materializes everything and we rebuild `b.snapshotList`/`b.snapshotIndex`, the `currentSnapshotID` lookup still resolves against `common.snapshotIndex`/`common.SnapshotList`, which is the pre-materialization eager subset. It works today only because `prepareDeferredSnapshots` always adds `CurrentSnapshotID` to the ref set. If that invariant ever slips, `b.currentSnapshotID` gets silently left nil even though the snapshot is present. I'd point that lookup at `b.snapshotIndex`/`b.snapshotList` so it reads the full set we just built.

table/deferred_snapshots.go
```diff
@@ -0,0 +1,502 @@
...
+	metadata, rawSnapshots, err := decodeMetadataWithRawSnapshots(normalized)
+	if err != nil {
+		if errors.Is(err, ErrInvalidMetadata) {
+			return nil, err
+		}
+
+		return nil, fmt.Errorf("%w: %w", ErrInvalidMetadata, err)
+	}
+
+	common := commonMetadataOf(metadata)
```
laskoviymishka: `commonMetadataOf` returns nil for anything outside the V1/V2/V3 switch, and we use it right away (`common.SnapshotList = eager`, `common.deferredSnapshots = deferred`), so an unexpected concrete type nil-panics here rather than erroring. Unreachable now, but a new format version would trip it. I'd add a guard: `if common == nil { return nil, fmt.Errorf("%w: unsupported metadata implementation %T", ErrInvalidMetadata, metadata) }`.

table/deferred_snapshots.go
```diff
@@ -0,0 +1,502 @@
...
+	deferred, eager, err := prepareDeferredSnapshots(rawSnapshots, common, metadata)
+	if err != nil {
+		return nil, err
+	}
+	common.SnapshotList = eager
```
laskoviymishka: Since `common.SnapshotList` is only the ref subset here, when `finishDeferredMetadataUnmarshal` later runs `validate()` -> `checkLastSequenceNumber(m, m.SnapshotList)` it's checking a partial list. `prepareDeferredSnapshots` already validated sequence numbers across all snapshots, so no correctness gap today, but the missing-`last-sequence-number` error message diverges from the eager path (you get "seq > -1" rather than "last-sequence-number is required"), and any future check added to `checkLastSequenceNumber` would silently cover only the eager subset in this path. Worth a comment, or passing the full validated list explicitly.

table/deferred_snapshots_test.go
```diff
@@ -0,0 +1,261 @@
...
+			require.NoError(t, err)
+			reparsed, err := ParseMetadataBytes(serialized)
+			require.NoError(t, err)
+			assert.True(t, eager.Equals(reparsed))
+		})
+	}
+}
+
+func TestDeferredSnapshotsValidateUnreferencedHistory(t *testing.T) {
+	var fields map[string]json.RawMessage
```
laskoviymishka: This exercises the summary-value path, but `prepareDeferredSnapshots` added five more branches with no coverage: missing `snapshot-id`, missing `timestamp-ms`, duplicate IDs, v2 sequence number over `last-sequence-number`, and v3 negative `first-row-id`. These are new and structurally distinct enough that I'd want each one hit. The existing setup table-drives cleanly into subtests.

catalog/rest/lazy_snapshot_bench_test.go
```diff
@@ -0,0 +1,191 @@
...
+			b.ResetTimer()
+			b.ReportAllocs()
+			b.ReportMetric(float64(len(rawMeta)), "metadata-bytes")
+			b.ReportMetric(float64(n), "snapshots")
+			for range b.N {
+				meta, err := table.ParseMetadataBytesDeferredSnapshots(rawMeta)
+				if err != nil {
+					b.Fatal(err)
+				}
+				if meta.SnapshotByID(0) == nil {
```
laskoviymishka: Hard-coding `SnapshotByID(0)` couples this to the generator emitting sequential IDs from 0. If it ever moves to random/UUID IDs the benchmark silently stops measuring a hit (the `b.Fatal` fires, or it mismeasures) instead of failing to compile. I'd capture a known historical ID from the generated metadata before the loop and look that up.

catalog/rest/lazy_snapshot_bench_test.go
```diff
@@ -0,0 +1,191 @@
...
+		})
+	}
+}
+
+// BenchmarkStageDeferredRetainedHeap measures the live heap held by one
+// deferred metadata object before and after full snapshot materialization.
+// Run with -benchtime=1x: retained-heap measurements describe one object and
+// cannot be meaningfully averaged across b.N independent lifecycles.
+func BenchmarkStageDeferredRetainedHeap(b *testing.B) {
+	if b.N != 1 {
```
laskoviymishka: This `b.N != 1` guard is on the outer benchmark's `b`, but each sub-benchmark under `b.Run` gets its own `b` and its own `b.N`, so they still iterate freely. The heap delta then gets sampled after several materializations rather than the single controlled lifecycle the comment describes, which is exactly the averaging problem the comment says it's avoiding. I'd move the guard inside the `b.Run` closure and check the sub-benchmark's `b.N`.

table/deferred_snapshots.go
```diff
@@ -0,0 +1,502 @@
...
+
+func (s deferredSnapshotFields) validateHeavyFields() error {
+	if err := validateStringArray(s.ManifestLocations); err != nil {
+		return err
+	}
+
+	return validateSummaryObject(s.Summary)
+}
+
+func validateStringArray(raw json.RawMessage) error {
```
laskoviymishka: Not a blocker, but this hand-rolled structural validator (~160 lines across the scan/split/validate helpers) has to be kept in sync with the JSON layout by hand, and it's now a third site alongside `deferredSnapshotFields` and `Snapshot.UnmarshalJSON` that tracks any snapshot-schema change. If it's deliberate for the parse-time win, a short comment saying so would help the next maintainer; if not, `json.Decoder.Token()` gets most of the same safety.

While we're here: `validateStringArray` and `validateSummaryObject` both accept a trailing comma (`["a",]`, `{"k":"v",}`) because the loop exits right after consuming the comma. Harmless against well-formed server JSON, but looser than the spec if we keep the hand-roll.

catalog/rest/rest.go
```diff
@@ -1645,8 +1633,20 @@ func (r *Catalog) CommitTable(ctx context.Context, ident table.Identifier, requi
...
 	}
+	metadata, err := r.parseCommitMetadata(ret.RawMetadata)
+	if err != nil {
+		return nil, "", err
+	}
+
+	return metadata, ret.MetadataLoc, nil
+}
+
+func (r *Catalog) parseCommitMetadata(raw json.RawMessage) (table.Metadata, error) {
```
laskoviymishka: Worth a comment that this is intentionally asymmetric: `LoadTable` in refs mode sends `?snapshots=refs` and parses eagerly, while the commit/update path receives a possibly-full history and parses it lazily. So a `*table.Table` from `LoadTable` has fully materialized history and one from `UpdateTable` may not, and the first `Snapshots()`/`Equals()` on a commit-returned metadata pays an allocation the load path already paid. Fine to diverge, but a line explaining why saves the next reader the double-take.

zeroshade (changes requested):
All three of my prior blockers are fixed (DCO on every commit, merge commit gone, branch is `MERGEABLE`), and the mechanism itself is sound. I want to call out the test suite specifically: I reverted each of the eight defensive changes in turn and **every one was caught by a named test** — no inert assertions, no tautologies. That is the strongest test suite I've reviewed in this batch.

Two substantive problems remain.

## Blocking

**1. `table/deferred_snapshots.go:213-215` and `:171-173` — the deferred parser rejects metadata the eager parser accepts.**

`validateSummaryObject` and `validateStringArray` are hand-rolled stand-ins for what `encoding/json` does when decoding into `map[string]string` / `[]string`, and they diverge on JSON `null`. Go's decoder stores `null` into a string as `""` without error; both validators require the value token to start with `"`. Minimal reproducers — both parse clean today, both fail with `snapshot-loading-mode=refs`:

```
"summary":   {"operation":"append","extra":null}
  eager=<nil>   defer=invalid metadata: snapshot …: cannot unmarshal summary value into string

"manifests": ["a.avro", null]
  eager=<nil>   defer=invalid metadata: snapshot …: cannot unmarshal manifest location into string
```

Why this matters more than a normal parse-strictness difference: it fires on the *commit response*, i.e. **after the server has already applied the commit**. `Transaction.Commit` would return `ErrInvalidMetadata` for a transaction that actually succeeded, and the natural user reaction — retry — risks a double-apply. It also directly contradicts the PR's claim that "the public `Metadata` API and v1/v2/v3 behavior are unchanged": enabling a *performance* mode must not change what parses.

Fix: in `validateSummaryObject`, accept a `null` value token (`bytes.HasPrefix(raw[i:], []byte("null"))` → advance 4 and continue), and likewise for elements in `validateStringArray`; add a parity test asserting both documents above still parse under the deferred path. A lower-risk alternative is to validate by decoding into `map[string]string`/`[]string` for unreferenced entries — you lose part of the allocation win but get parity for free.

(A third, much more obscure instance of the same class: `"summary":{"operation":"","operation":"append"}` — Go's last-key-wins map decode accepts it, the scanner rejects on the first key. Not worth fixing on its own, but the null fix should come with a comment stating the intended acceptance contract.)

**2. `table/metadata.go:425-429` + `table/table.go:210` — the realistic post-commit path is an unmeasured ~21% allocation regression.**

`Table.NewTransaction*` calls `MetadataBuilderFromBase`, which calls `snapshotsForMarshal()` and fully materialises. So the *second* write on a committed table pays the deferred index build **and** the full decode. Measured (parse + `MetadataBuilderFromBase`, `-count=3`):

| metadata bytes | eager allocs/op | deferred allocs/op | Δ | eager B/op | deferred B/op | Δ |
|---:|---:|---:|---:|---:|---:|---:|
| 7,056 | 906 | 1,096 | **+21.0%** | 58,952 | 77,272 | **+31.1%** |
| 59,078 | 6,132 | 7,405 | **+20.8%** | 419,130 | 582,586 | **+39.0%** |
| 582,880 | 58,345 | 70,424 | **+20.7%** | 3,953,508 | 5,547,678 | **+40.3%** |
| 1,999,830 | 198,613 | 239,713 | **+20.7%** | 13,850,244 | 19,212,749 | **+38.7%** |
| 5,856,882 | 580,444 | 700,558 | **+20.7%** | 41,413,480 | 57,358,112 | **+38.5%** |

At 5.86 MB the deferred side was also consistently slower in wall time (256/278/293 ms vs 213/227/239 ms).

Your own `BenchmarkStageDeferredAllSnapshots` already contains this evidence — at ~2 MB it reports 239,675 allocs / 19.07 MB versus `BenchmarkStageFullParse`'s 174,658 allocs / 10.43 MB — but it's only described as "the intentional fallback … pay the deferred full-decoding cost once" and is never put next to the eager baseline. That reads as "you pay it once" rather than "it is strictly worse than not deferring".

Please either (a) add this parse→`NewTransaction` pair to the benchmark set and state the trade-off in the description and in the doc comment on `ParseMetadataBytesDeferredSnapshots`, or (b) carry the deferred state into `MetadataBuilder` so a builder that never touches history doesn't materialise. I don't think (b) is required for this PR, but (a) is — as written, a reader would conclude this is a strict improvement for REST commit traffic, and for any table that gets more than one write per `Table` object it isn't.

## Major

**3. `table/metadata.go:3071`,`:3169`,`:3289` — marshalling a metadata *value* silently drops unreferenced snapshots.** The three `MarshalJSON` methods that project `snapshotsForMarshal()` have **pointer receivers**, so `json.Marshal` on a non-addressable `metadataV2` value skips them and falls back to reflection over `commonMetadata.SnapshotList` — which under deferral holds only the ref-reachable subset. Measured on the example v2 document with one unreferenced snapshot: `json.Marshal(ptr)` → **2** snapshots, `json.Marshal(*ptr)` → **1**, no error.

Every production write path is safe today (`catalog/internal/utils.go:72` encodes the interface, which holds a pointer; `metadataV1.ToV2()` returns a value but clears `deferredSnapshots` first), so this is latent — I could not find a live call site that trips it. But this PR is what turns the consequence into *silent snapshot loss in a written metadata file* rather than a missing `current-snapshot-id`, and the type does return by value (`ToV2`). Fix is three characters: value receivers (`func (m metadataV2) MarshalJSON()`), which puts the method in both method sets — `Snapshot.MarshalJSON` at `snapshots.go:306` already uses that form. Please add a regression test asserting `json.Marshal(*meta.(*metadataV2))` preserves the full history.

**4. `table/metadata.go:2384-2390`,`:2372-2378` — a deferred decode error is swallowed and presented as "no snapshots".** `allSnapshots()` discards the error and returns `nil`, so `Snapshots()` reports an empty history and `Equals()` compares two empty lists as equal; `SnapshotByID` returns `nil`, indistinguishable from absent. The two callers that *can* return an error (`MarshalJSON`, `MetadataBuilderFromBase`) do propagate it, which is the right half.

I tried hard to reach the swallowed path and couldn't — `deferredSnapshotFields` is field-for-field type-identical to `Snapshot` except `manifests`/`summary`, the validators are equal-or-stricter than `encoding/json` (20,000-case fuzz found 0 eager-stricter), and `encoding/json` has syntax-checked the document before `splitJSONArray` runs. So it isn't currently reachable. But it's precisely the "not yet decoded treated as absent" failure mode, and the invariant keeping it unreachable is exactly the one Blocking 1 shows isn't being maintained deliberately. At minimum add a comment at `:2384` recording that `snapshotsForMarshal` cannot fail for state produced by `prepareDeferredSnapshots`, and why.

## What checks out

- **8/8 mutations killed** — reverting the head commit, moving `s.mu.RUnlock()` before `&s.entries[i]` (real data race at `deferred_snapshots.go:108` under `-race -count=5`), deleting each `MarshalJSON`, stripping `ToV2` materialisation, `Equals` → `c.SnapshotList`, `Snapshots()` → `c.SnapshotList`, stripping the v3 `snapshots` projection, stripping the deferred branch from `SnapshotByID`. Each caught by a named test.
- **Differential fuzz, 20,000 generated v2 documents**: 3,697 divergences, **all in the deferred-stricter direction, 0 eager-stricter**. For all 16,303 mutually-accepted documents, deferred and eager marshalled **byte-identically** while still deferred, and `SnapshotByID` on an unreferenced snapshot returned a value `reflect.DeepEqual` to the eager one.
- **Enumerated every non-test reader** of `commonMetadata.SnapshotList` between response decode and final metadata (`metadata.go:451`, `:1403`, `:2364-2370`, `:2736`, `:2868`, `:3187`, `:3314`, `:3338`, `:3090`). All are either safe by construction — the current snapshot and every ref target are eagerly decoded, and `deferredSnapshots` is assigned *after* `finishDeferredMetadataUnmarshal`, so `checkRefsExist` can't see a half-populated state — or duplicated synchronously in `prepareDeferredSnapshots`. Every public reader in the repo goes through `Snapshots()`/`SnapshotByID`, which materialise. One hole found (Major 3).
- **Benchmarks reproduce.** At ~2 MB I measure 41,409 vs 174,658 allocs/op against your 41,503 vs 174,749 — a match to <0.1%. Historical-lookup delta 41,460 vs 41,407 (+53) against your +52.
- `golangci-lint` 0 issues; `-race` clean on both touched packages; CI 15/15.

## Minor

- The new `MarshalJSON` wrappers move the top-level `snapshots` key to the **end** of the serialized object for v1, v2 *and* v3, in **all** snapshot-loading modes, because Go orders fields by index sequence and the explicit `SnapshotList` field follows the embedded `*Alias`. Semantically irrelevant, but it changes every metadata file this client writes, and "v1/v2/v3 behavior are unchanged" doesn't cover it.
- `metadata.go:1953-1974` — `ParseMetadataBytesDeferredSnapshots` duplicates the preflight of `ParseMetadataBytes` almost verbatim; factor out the shared step. It also drops the `ret.Version() != formatVersion` cross-check the eager path does at `:1942-1945`. I couldn't construct an input where they disagree, so this is a consistency nit.
- `metadata.go:450-453` — the current-snapshot presence check uses `common.snapshotIndex, common.SnapshotList` (the eager subset) even though `b.snapshotIndex`/`b.snapshotList` were built two lines earlier. Correct today only because the current snapshot is always eagerly decoded; using `b.*` removes the dependency.
- `deferred_snapshots.go:279` — `splitJSONArray` re-trims a slice its only caller already trimmed, and returns offsets the caller applies to *its* slice. Works only because the second trim is a no-op; a future caller passing untrimmed bytes gets silently wrong spans. (I stress-tested the scanner on braces/commas/brackets inside strings, escaped quotes and backslashes, nested structures, empty and whitespace-only arrays, and a 501-element array — all correct.)
- Commit hygiene: the builder assertion at `deferred_snapshots_test.go:87-100` was present from the first commit, and reverting only the head commit's single hunk makes it fail — so `go test ./table/` was **red at both `d704277d` and `187c473b`**. Bisect-hostile; please squash or fix in place.
- ASF headers present on both new files; range-over-int used correctly (the index loops in the scanners are genuine index arithmetic and correctly not converted).

## Prior items

- DCO missing on 3 of 4 commits → **Fixed**, all three commits signed off.
- Merge commit `d6986d2` → **Fixed**, history is linear.
- `mergeStateStatus: DIRTY` → **Fixed**, now `MERGEABLE`.
- Mode resolution / unknown values rejected → **still correct, N.A.**
- Per-entry `sync.Once`, never a panic → **Confirmed and now stronger**: reverting `187c473b`'s lock ordering produces a real race the suite catches.
- "No observable loss; `Snapshots()` fully materialises" → **Partially fixed.** True for the interface path, and the head commit closed the `MetadataBuilderFromBase` leak my prior review missed. But the value-marshal path does observably lose snapshots (Major 3), and `Snapshots()` returns nil rather than an error on a deferred decode failure (Major 4).
- Unknown snapshot fields round-trip → **Confirmed.** Fuzz cases carrying unknown nested objects round-trip byte-identically on both paths.
- Aliasing / `json.RawMessage` ownership → **Confirmed, unchanged.**
- "The lazy-decode design is sound, no findings against the mechanism" → **partially superseded.** The mechanism is sound; that round didn't examine validation parity between the two parsers, nor benchmark the parse→`NewTransaction` sequence. Those are Blocking 1 and 2.

## Description

- **"The public `Metadata` API and v1/v2/v3 behavior are unchanged" is inaccurate twice** — the deferred path rejects documents the eager path accepts, and serialized top-level key order changes for all three versions in all modes.
- "Synchronously validate … summary values. Historical corruption therefore remains a parse-time error" — true, but the validation isn't *equivalent* to the eager path's; it's strictly stricter on `null`.
- "Callers that request all snapshots pay the deferred full-decoding cost once" — understated. Full materialisation costs **+37% allocations and +83% bytes** versus never deferring, from your own benchmarks, and it's triggered by `Table.NewTransaction`, not only by explicit history access.
- Benchmark revisions are honestly disclosed as stale relative to head. Allocation figures still reproduce to <0.1%, so the substance holds. Two don't: the claimed **38.2% improvement at 7,056 B** isn't statistically distinguishable on my box (290.0 µs vs 218.5 µs, p=0.123 over 10 interleaved rounds) — allocations at that size are unambiguously better (−41%), so I'd lead with allocations at the small end rather than a wall-clock percentage. And **−25.2% B/op at ~2 MB** now measures **−29.9%**, because main's baseline moved. Worth re-measuring on the rebased head.

---

technicolorbeat:
> All three of my prior blockers are fixed (DCO on every commit, merge commit gone, branch is `MERGEABLE`), and the mechanism itself is sound. I want to call out the test suite specifically: I reverted each of the eight defensive changes in turn and **every one was caught by a named test** — no inert assertions, no tautologies. That is the strongest test suite I've reviewed in this batch.
> 
> Two substantive problems remain.
> 
> ## Blocking
> **1. `table/deferred_snapshots.go:213-215` and `:171-173` — the deferred parser rejects metadata the eager parser accepts.**
> 
> `validateSummaryObject` and `validateStringArray` are hand-rolled stand-ins for what `encoding/json` does when decoding into `map[string]string` / `[]string`, and they diverge on JSON `null`. Go's decoder stores `null` into a string as `""` without error; both validators require the value token to start with `"`. Minimal reproducers — both parse clean today, both fail with `snapshot-loading-mode=refs`:
> 
> ```
> "summary":   {"operation":"append","extra":null}
>   eager=<nil>   defer=invalid metadata: snapshot …: cannot unmarshal summary value into string
> 
> "manifests": ["a.avro", null]
>   eager=<nil>   defer=invalid metadata: snapshot …: cannot unmarshal manifest location into string
> ```
> 
> Why this matters more than a normal parse-strictness difference: it fires on the _commit response_, i.e. **after the server has already applied the commit**. `Transaction.Commit` would return `ErrInvalidMetadata` for a transaction that actually succeeded, and the natural user reaction — retry — risks a double-apply. It also directly contradicts the PR's claim that "the public `Metadata` API and v1/v2/v3 behavior are unchanged": enabling a _performance_ mode must not change what parses.
> 
> Fix: in `validateSummaryObject`, accept a `null` value token (`bytes.HasPrefix(raw[i:], []byte("null"))` → advance 4 and continue), and likewise for elements in `validateStringArray`; add a parity test asserting both documents above still parse under the deferred path. A lower-risk alternative is to validate by decoding into `map[string]string`/`[]string` for unreferenced entries — you lose part of the allocation win but get parity for free.
> 
> (A third, much more obscure instance of the same class: `"summary":{"operation":"","operation":"append"}` — Go's last-key-wins map decode accepts it, the scanner rejects on the first key. Not worth fixing on its own, but the null fix should come with a comment stating the intended acceptance contract.)
> 
> **2. `table/metadata.go:425-429` + `table/table.go:210` — the realistic post-commit path is an unmeasured ~21% allocation regression.**
> 
> `Table.NewTransaction*` calls `MetadataBuilderFromBase`, which calls `snapshotsForMarshal()` and fully materialises. So the _second_ write on a committed table pays the deferred index build **and** the full decode. Measured (parse + `MetadataBuilderFromBase`, `-count=3`):
> 
> metadata bytes	eager allocs/op	deferred allocs/op	Δ	eager B/op	deferred B/op	Δ
> 7,056	906	1,096	**+21.0%**	58,952	77,272	**+31.1%**
> 59,078	6,132	7,405	**+20.8%**	419,130	582,586	**+39.0%**
> 582,880	58,345	70,424	**+20.7%**	3,953,508	5,547,678	**+40.3%**
> 1,999,830	198,613	239,713	**+20.7%**	13,850,244	19,212,749	**+38.7%**
> 5,856,882	580,444	700,558	**+20.7%**	41,413,480	57,358,112	**+38.5%**
> At 5.86 MB the deferred side was also consistently slower in wall time (256/278/293 ms vs 213/227/239 ms).
> 
> Your own `BenchmarkStageDeferredAllSnapshots` already contains this evidence — at ~2 MB it reports 239,675 allocs / 19.07 MB versus `BenchmarkStageFullParse`'s 174,658 allocs / 10.43 MB — but it's only described as "the intentional fallback … pay the deferred full-decoding cost once" and is never put next to the eager baseline. That reads as "you pay it once" rather than "it is strictly worse than not deferring".
> 
> Please either (a) add this parse→`NewTransaction` pair to the benchmark set and state the trade-off in the description and in the doc comment on `ParseMetadataBytesDeferredSnapshots`, or (b) carry the deferred state into `MetadataBuilder` so a builder that never touches history doesn't materialise. I don't think (b) is required for this PR, but (a) is — as written, a reader would conclude this is a strict improvement for REST commit traffic, and for any table that gets more than one write per `Table` object it isn't.
> 
> ## Major
> **3. `table/metadata.go:3071`,`:3169`,`:3289` — marshalling a metadata _value_ silently drops unreferenced snapshots.** The three `MarshalJSON` methods that project `snapshotsForMarshal()` have **pointer receivers**, so `json.Marshal` on a non-addressable `metadataV2` value skips them and falls back to reflection over `commonMetadata.SnapshotList` — which under deferral holds only the ref-reachable subset. Measured on the example v2 document with one unreferenced snapshot: `json.Marshal(ptr)` → **2** snapshots, `json.Marshal(*ptr)` → **1**, no error.
> 
> Every production write path is safe today (`catalog/internal/utils.go:72` encodes the interface, which holds a pointer; `metadataV1.ToV2()` returns a value but clears `deferredSnapshots` first), so this is latent — I could not find a live call site that trips it. But this PR is what turns the consequence into _silent snapshot loss in a written metadata file_ rather than a missing `current-snapshot-id`, and the type does return by value (`ToV2`). Fix is three characters: value receivers (`func (m metadataV2) MarshalJSON()`), which puts the method in both method sets — `Snapshot.MarshalJSON` at `snapshots.go:306` already uses that form. Please add a regression test asserting `json.Marshal(*meta.(*metadataV2))` preserves the full history.
> 
> **4. `table/metadata.go:2384-2390`,`:2372-2378` — a deferred decode error is swallowed and presented as "no snapshots".** `allSnapshots()` discards the error and returns `nil`, so `Snapshots()` reports an empty history and `Equals()` compares two empty lists as equal; `SnapshotByID` returns `nil`, indistinguishable from absent. The two callers that _can_ return an error (`MarshalJSON`, `MetadataBuilderFromBase`) do propagate it, which is the right half.
> 
> I tried hard to reach the swallowed path and couldn't — `deferredSnapshotFields` is field-for-field type-identical to `Snapshot` except `manifests`/`summary`, the validators are equal-or-stricter than `encoding/json` (20,000-case fuzz found 0 eager-stricter), and `encoding/json` has syntax-checked the document before `splitJSONArray` runs. So it isn't currently reachable. But it's precisely the "not yet decoded treated as absent" failure mode, and the invariant keeping it unreachable is exactly the one Blocking 1 shows isn't being maintained deliberately. At minimum add a comment at `:2384` recording that `snapshotsForMarshal` cannot fail for state produced by `prepareDeferredSnapshots`, and why.
> 
> ## What checks out
> * **8/8 mutations killed** — reverting the head commit, moving `s.mu.RUnlock()` before `&s.entries[i]` (real data race at `deferred_snapshots.go:108` under `-race -count=5`), deleting each `MarshalJSON`, stripping `ToV2` materialisation, `Equals` → `c.SnapshotList`, `Snapshots()` → `c.SnapshotList`, stripping the v3 `snapshots` projection, stripping the deferred branch from `SnapshotByID`. Each caught by a named test.
> * **Differential fuzz, 20,000 generated v2 documents**: 3,697 divergences, **all in the deferred-stricter direction, 0 eager-stricter**. For all 16,303 mutually-accepted documents, deferred and eager marshalled **byte-identically** while still deferred, and `SnapshotByID` on an unreferenced snapshot returned a value `reflect.DeepEqual` to the eager one.
> * **Enumerated every non-test reader** of `commonMetadata.SnapshotList` between response decode and final metadata (`metadata.go:451`, `:1403`, `:2364-2370`, `:2736`, `:2868`, `:3187`, `:3314`, `:3338`, `:3090`). All are either safe by construction — the current snapshot and every ref target are eagerly decoded, and `deferredSnapshots` is assigned _after_ `finishDeferredMetadataUnmarshal`, so `checkRefsExist` can't see a half-populated state — or duplicated synchronously in `prepareDeferredSnapshots`. Every public reader in the repo goes through `Snapshots()`/`SnapshotByID`, which materialise. One hole found (Major 3).
> * **Benchmarks reproduce.** At ~2 MB I measure 41,409 vs 174,658 allocs/op against your 41,503 vs 174,749 — a match to <0.1%. Historical-lookup delta 41,460 vs 41,407 (+53) against your +52.
> * `golangci-lint` 0 issues; `-race` clean on both touched packages; CI 15/15.
> 
> ## Minor
> * The new `MarshalJSON` wrappers move the top-level `snapshots` key to the **end** of the serialized object for v1, v2 _and_ v3, in **all** snapshot-loading modes, because Go orders fields by index sequence and the explicit `SnapshotList` field follows the embedded `*Alias`. Semantically irrelevant, but it changes every metadata file this client writes, and "v1/v2/v3 behavior are unchanged" doesn't cover it.
> * `metadata.go:1953-1974` — `ParseMetadataBytesDeferredSnapshots` duplicates the preflight of `ParseMetadataBytes` almost verbatim; factor out the shared step. It also drops the `ret.Version() != formatVersion` cross-check the eager path does at `:1942-1945`. I couldn't construct an input where they disagree, so this is a consistency nit.
> * `metadata.go:450-453` — the current-snapshot presence check uses `common.snapshotIndex, common.SnapshotList` (the eager subset) even though `b.snapshotIndex`/`b.snapshotList` were built two lines earlier. Correct today only because the current snapshot is always eagerly decoded; using `b.*` removes the dependency.
> * `deferred_snapshots.go:279` — `splitJSONArray` re-trims a slice its only caller already trimmed, and returns offsets the caller applies to _its_ slice. Works only because the second trim is a no-op; a future caller passing untrimmed bytes gets silently wrong spans. (I stress-tested the scanner on braces/commas/brackets inside strings, escaped quotes and backslashes, nested structures, empty and whitespace-only arrays, and a 501-element array — all correct.)
> * Commit hygiene: the builder assertion at `deferred_snapshots_test.go:87-100` was present from the first commit, and reverting only the head commit's single hunk makes it fail — so `go test ./table/` was **red at both `d704277d` and `187c473b`**. Bisect-hostile; please squash or fix in place.
> * ASF headers present on both new files; range-over-int used correctly (the index loops in the scanners are genuine index arithmetic and correctly not converted).
> 
> ## Prior items
> * DCO missing on 3 of 4 commits → **Fixed**, all three commits signed off.
> * Merge commit `d6986d2` → **Fixed**, history is linear.
> * `mergeStateStatus: DIRTY` → **Fixed**, now `MERGEABLE`.
> * Mode resolution / unknown values rejected → **still correct, N.A.**
> * Per-entry `sync.Once`, never a panic → **Confirmed and now stronger**: reverting `187c473b`'s lock ordering produces a real race the suite catches.
> * "No observable loss; `Snapshots()` fully materialises" → **Partially fixed.** True for the interface path, and the head commit closed the `MetadataBuilderFromBase` leak my prior review missed. But the value-marshal path does observably lose snapshots (Major 3), and `Snapshots()` returns nil rather than an error on a deferred decode failure (Major 4).
> * Unknown snapshot fields round-trip → **Confirmed.** Fuzz cases carrying unknown nested objects round-trip byte-identically on both paths.
> * Aliasing / `json.RawMessage` ownership → **Confirmed, unchanged.**
> * "The lazy-decode design is sound, no findings against the mechanism" → **partially superseded.** The mechanism is sound; that round didn't examine validation parity between the two parsers, nor benchmark the parse→`NewTransaction` sequence. Those are Blocking 1 and 2.
> 
> ## Description
> * **"The public `Metadata` API and v1/v2/v3 behavior are unchanged" is inaccurate twice** — the deferred path rejects documents the eager path accepts, and serialized top-level key order changes for all three versions in all modes.
> * "Synchronously validate … summary values. Historical corruption therefore remains a parse-time error" — true, but the validation isn't _equivalent_ to the eager path's; it's strictly stricter on `null`.
> * "Callers that request all snapshots pay the deferred full-decoding cost once" — understated. Full materialisation costs **+37% allocations and +83% bytes** versus never deferring, from your own benchmarks, and it's triggered by `Table.NewTransaction`, not only by explicit history access.
> * Benchmark revisions are honestly disclosed as stale relative to head. Allocation figures still reproduce to <0.1%, so the substance holds. Two don't: the claimed **38.2% improvement at 7,056 B** isn't statistically distinguishable on my box (290.0 µs vs 218.5 µs, p=0.123 over 10 interleaved rounds) — allocations at that size are unambiguously better (−41%), so I'd lead with allocations at the small end rather than a wall-clock percentage. And **−25.2% B/op at ~2 MB** now measures **−29.9%**, because main's baseline moved. Worth re-measuring on the rebased head.
> 
> > _This review was drafted by an AI-assisted tool and confirmed by an Iceberg Go maintainer. The findings cite the project's review criteria; if you think one is mis-applied, please reply and a maintainer will weigh in._



Thanks for the detailed review and for validating the test suite so thoroughly. I reproduced the findings and pushed the follow-up changes.

At a high level, I aligned deferred parsing with eager JSON semantics, ensured full snapshot history is preserved across serialization, documented the deferred-state invariants, and added regression and parse-to-builder benchmark coverage.

The updated benchmarks make the trade-off explicit. At ~2 MB, deferred commit-response parsing improves from **78.420 ms to 50.326 ms**, with **76.3% fewer allocations** and **29.9% fewer allocated bytes**. However, immediately creating a `MetadataBuilder` forces full materialization: parse-plus-builder increases from **80.611 ms to 98.314 ms**, with **20.7% more allocations** and **38.9% more allocated bytes**.

I agree this is not a strict improvement for repeated writes on the same returned `Table`. This PR optimizes the `snapshot-loading-mode=refs` commit-response path when full history is not immediately required. Carrying deferred state into `MetadataBuilder` would be a separate follow-up - #1986

table/deferred_snapshots.go
```diff
@@ -0,0 +1,532 @@
...
+		seen[*fields.SnapshotID] = struct{}{}
+		state.entries[i] = deferredSnapshotEntry{start: span.start, end: span.end}
+		state.byID[*fields.SnapshotID] = i
+
+		if err := fields.validateHeavyFields(); err != nil {
+			return nil, nil, fmt.Errorf("%w: snapshot %d: %w", ErrInvalidMetadata, *fields.SnapshotID, err)
+		}
+		if metadata.Version() > 1 && fields.SequenceNumber > metadata.LastSequenceNumber() {
+			return nil, nil, fmt.Errorf("%w: snapshot %d has sequence number %d which is greater than last-sequence-number %d",
+				ErrInvalidMetadata, *fields.SnapshotID, fields.SequenceNumber, metadata.LastSequenceNumber())
```
zeroshade, line 509: **nit** — Divergent (but still rejecting) error message when v2 metadata omits last-sequence-number

prepareDeferredSnapshots runs its sequence-number check before finishUnmarshal, so a v2 document with no last-sequence-number reports the per-snapshot message against the -1 sentinel instead of the eager path's 'last-sequence-number is required for format versions greater than 1'. Both paths reject with ErrInvalidMetadata, so acceptance parity holds; only the diagnostic differs.

table/metadata.go
```diff
@@ -2969,42 +3036,69 @@ func (m *metadataV1) UnmarshalJSON(b []byte) error {
...
+	}{
+		Alias:        (*Alias)(&m),
+		SnapshotList: snapshots,
+	})
+}
+
 func (m *metadataV1) ToV2() metadataV2 {
 	commonOut := m.commonMetadata
+	commonOut.SnapshotList = m.allSnapshots()
+	commonOut.snapshotIndex = buildSnapshotIndex(commonOut.SnapshotList)
```
zeroshade, line 3100: **nit** — ToV2() result aliases the source's snapshot backing array (pre-existing, not a regression)

commonOut.SnapshotList = m.allSnapshots() returns the internal slice rather than a clone, so mutating the returned metadataV2's SnapshotList mutates what the source metadata reports. I verified this is NOT introduced by this PR: the eager path has the identical aliasing on main because commonOut := m.commonMetadata shallow-copies the slice header. Noting only so it is not mistaken for new behaviour.

catalog/rest/lazy_snapshot_bench_test.go
```diff
@@ -0,0 +1,229 @@
...
+		b.Skip("retained-heap benchmark requires -benchtime=1x")
```
zeroshade, line 187: **nit** — Retained-heap benchmark skips unless -benchtime=1x

BenchmarkStageDeferredRetainedHeap b.Skip()s under a default `go test -bench` invocation and reports a synthetic ReportMetric(0, "ns/op"). It is legitimate (it produced the retained-heap table in the description) but is an unusual benchmark shape that will read as dead to anyone running the suite normally; a one-line comment already explains the -benchtime=1x requirement.

zeroshade (approved):
All four prior findings (2 blocking, 2 major) are fixed at head and each fix is pinned by a test that fails when I mutate the mechanism; a 3,000-case differential fuzz found zero eager/deferred acceptance divergences.

Re-review verification: 7 of 8 prior findings confirmed fixed at `9e7e1bd` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *partially fixed* — Minor: splitJSONArray re-trims a slice its only caller already trimmed and returns offsets into the caller's slice

```text
go build ./... (OK); go vet ./table/... ./catalog/rest/... (clean); go test -race -count=1 ./table/ (ok, 18.8s); go test -race -count=1 ./catalog/rest/ (ok, 7.5s); go vet ./catalog/hadoop/ on both PR head and main (both clean - the stale harness 'setup failed' is a docker/minio environment issue, not a PR regression). Throwaway probes (deleted): acceptance-parity over 26 targeted summary/manifests documents; 3000-case differential fuzz; value-receiver marshal history check; 17 snapshots-array shape parity cases incl. null/[]/[null]/[1]/duplicate-id/whitespace/embedded-delimiter strings; ToV2 aliasing on both eager and deferred. Mutations: neutered hasJSONNullAt (3 subtests red); reverted all three MarshalJSON to pointer receivers (3 subtests red); both restored via git checkout.
```

</details>

---

>

table/deferred_snapshots.go
technicolorbeat: Deferred v2/v3 parsing now reports the same missing last-sequence-number validation error as eager parsing and added tests covering original-buffer offsets, missing/null/empty-array behavior, and eager/deferred validation-error parity

table/deferred_snapshots.go
```diff
@@ -0,0 +1,539 @@
...
+	case *metadataV3:
+		return typed.finishUnmarshal()
+	default:
+		return fmt.Errorf("%w: unsupported metadata implementation %T", ErrInvalidMetadata, metadata)
+	}
+}
+
+func prepareDeferredSnapshots(rawSnapshots json.RawMessage, common *commonMetadata, metadata Metadata) (*deferredSnapshotState, []Snapshot, error) {
+	if metadata.Version() > 1 {
+		if err := requireLastSequenceNumber(metadata); err != nil {
```
zeroshade, line 454: **nit** — Required-field pre-check reorders diagnostics ahead of eager's checkRequiredFields, diverging on double-fault documents

requireLastSequenceNumber now runs before every other deferred validation, whereas the eager path reaches checkLastSequenceNumber only inside validate(), after preValidate() and checkRequiredFields(). A document with BOTH a missing last-sequence-number and another fault therefore reports a different (still-rejecting) message on each path. Acceptance parity is unaffected — both reject with ErrInvalidMetadata in all cases. A verified alternative gives exact parity: delete the :453-457 pre-check and instead guard the per-snapshot comparison at :479 with metadata.LastSequenceNumber() != -1, letting finishUnmarshal's checkLastSequenceNumber produce the error in eager order.

zeroshade, line 454: **nit** — metadata.Version() > 1 guard is redundant

metadataV1.LastSequenceNumber() returns the constant 0 (table/metadata.go:2979), never the -1 sentinel, so requireLastSequenceNumber can never fire for v1 and the version guard is unreachable-by-construction. Harmless and arguably self-documenting (it mirrors the fact that metadataV1.validate() never calls checkLastSequenceNumber), so this is purely informational — no change requested.

zeroshade (approved):
The delta correctly fixes the missing-last-sequence-number diagnostic divergence with a mutation-verified test and incidentally resolves the prior splitJSONArray double-trim minor, with zero acceptance divergences across a 36-case eager/deferred probe.

Re-review verification: 2 of 4 prior findings confirmed fixed at `f2c0b27` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *not fixed* — nit — Retained-heap benchmark skips unless -benchtime=1x

```text
go build ./... (OK); go vet ./table/... ./catalog/rest/... (clean); go test -race -count=1 ./table/ (ok, 19.062s); go test -race -count=1 ./catalog/rest/ (ok, 7.395s); targeted go test -run 'TestDeferredSnapshotsMissingLastSequenceNumberErrorParity|TestSplitJSONArray|TestDeferredSnapshotsValidationMatchesEagerJSONSemantics' -v (all PASS). Mutations (all restored via git checkout, git status --porcelain empty): (1) deleted the requireLastSequenceNumber pre-check -> parity test red on v2+v3; (2a) deleted splitJSONArray's null/empty early return -> TestSplitJSONArrayPreservesMissingNullAndEmptyDistinction red with 'expected JSON array'; (2b) returned []rawJSONSpan{} instead of nil for null/missing -> same test red ('Expected nil, but got: []table.rawJSONSpan{}') plus the wider ./table/ suite red, confirming the nil-vs-empty distinction is load-bearing; (3) applied the zero-divergence alternative fix -> 0/36 probe divergences. Throwaway probe table/pr1956_probe_test.go (36 eager-vs-deferred parity cases) written, run, and deleted.
```

</details>

---

>

PR #1957, opened 2026-08-28
perf(table): resolve only requested equality-delete field paths

fallintoplace:
## Summary

- **Resolve less:** Equality-delete setup now resolves only field IDs from the active delete sets.
- **Reuse traversal state:** Internal schema fields are borrowed for the read-only walk, and only matching paths are copied.
- **Keep validation:** Existing nested paths and ambiguous-column errors remain covered.

## Benchmark

Apple M1 Pro, Go 1.26.3, 2,048-field schema:

- **1 equality field:** about 586 KB / 4,128 allocs to 672 B / 7 allocs
- **8 equality fields:** about 586 KB / 4,128 allocs to 896 B / 21 allocs
- **1-field setup time:** about 280 us to 10 us

## Tests

- `go test -timeout=5m` across all packages except `io/gocloud`
- `go test -race -timeout=5m . ./table/...`

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -352,8 +352,7 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 
 	type Alias Schema
 
-	aliasCopy := *(*Alias)(s)
-	aliasCopy.IdentifierFieldIDs = ids
+	aliasCopy := Alias{ID: s.ID, IdentifierFieldIDs: ids}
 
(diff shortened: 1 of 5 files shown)


laskoviymishka (commented):
🌞

laskoviymishka (approved):

PR #1962, opened 2026-08-28
perf(codec): precompute DataFile Avro field indexes

fallintoplace:
## What changed

- Precompute the indexes of `dataFile` fields carrying an `avro` tag once during package initialization.
- Reuse those indexes when cloning a DataFile for `MarshalAvroEntry`.
- Keep tag discovery reflection-based so `dataFile` remains the single source of truth for the wire shape.
- Add a coverage test and focused benchmarks.

## Why

`MarshalAvroEntry` needs a shallow copy because `dataFile` contains synchronization state. The old clone helper scanned every struct field and looked up its Avro tag for every encoded manifest entry. The new helper keeps the same copy semantics and removes that repeated metadata lookup.

## Benchmark

Apple M1 Pro, Go 1.26.3, `-cpu=1`, median of 8 runs.

| Benchmark | Before | After | Change |
| --- | ---: | ---: | ---: |
| CloneDataFileAvroFields | 2.00 us | 0.46 us | -77% |
| MarshalAvroEntry/v1 | 18.02 us | 15.88 us | -12% |
| MarshalAvroEntry/v2 | 17.53 us | 15.92 us | -9% |
| MarshalAvroEntry/v3 | 18.51 us | 15.77 us | -15% |

Allocation counts and bytes per operation are unchanged:

- Clone: 320 B/op, 1 alloc/op
- Marshal v1: 13,376 B/op, 144 allocs/op
- Marshal v2/v3: 13,360 B/op, 144 allocs/op

Command:

```bash
go test . -run '^$' -bench '^(BenchmarkCloneDataFileAvroFields|BenchmarkMarshalAvroEntry)$' -benchmem -benchtime=500ms -count=8 -cpu=1
```

## Testing

- `go test .`
- `go test -race .`
- `go vet .`
- All packages other than the metadata-endpoint-dependent `io/gocloud` package pass with `go test`.

### DIFF
--- a/data_file_codec.go
+++ b/data_file_codec.go
@@ -175,16 +175,29 @@ func newDecodeEntry(version int) (any, *dataFile) {
 	return &manifestEntry{Data: df}, df
 }
 
+var dataFileAvroFieldIndexes = avroFieldIndexes(reflect.TypeOf(dataFile{}))
+
+func avroFieldIndexes(t reflect.Type) []int {
+	indexes := make([]int, 0, t.NumField())
+	for i := range t.NumField() {
+		if _, hasAvroTag := t.Field(i).Tag.Lookup("avro"); hasAvroTag {
+			indexes = append(indexes, i)
+		}
+	}
+
+	return indexes
+}
+
 // cloneDataFileAvroFields returns a fresh *dataFile populated with src's
 // avro-tagged fields. Internal state (sync.Once, lazy-init caches,
 // specID, the field-id lookup maps) is intentionally left at zero
 // values because the avro encoder reads only the avro-tagged fields.
 //
-// Using reflection over the tag set means a new avro-tagged field
-// upstream is auto-copied without an update here — the dataFile struct
-// remains the single source of truth for the wire shape. It also
-// sidesteps the go-vet copies-lock warning that would fire on a
-// struct-literal copy of *dataFile (it embeds sync.Once).
+// The avro-tagged field indexes are discovered once when the package is
+// initialized. This keeps dataFile as the single source of truth for the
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
Clean change — this removes repeated reflection metadata lookup from DataFile cloning without introducing any schema-dependent state. Approving.

The risk I looked for was a precomputed index going stale or being applied to the wrong schema, and it can't happen here: `dataFileAvroFieldIndexes` is computed from `reflect.TypeOf(dataFile{})` and used only to copy fields between `*dataFile` values (`data_file_codec.go:175-220`). These are **Go struct field positions, not Avro schema offsets**, so decoding v1, v2, and v3 manifests cannot apply an index derived from one schema to another — differing manifest versions only change which tagged fields the Avro schema reads, not the clone mapping. That also rules out the wrong-neighbour read that would silently corrupt decoded values.

Missing, optional, and union field handling stays where it was, in Avro's existing tag/name-based encoding — untouched by this change. The new tag-coverage test (`data_file_codec_test.go:146-180`) derives expectations from the same struct and asserts every current `avro`-tagged field is copied, which guards future field additions from silently falling out of the clone.

The `schema.go:337-357` change is a good catch alongside it: constructing the JSON alias with only `ID` and `IdentifierFieldIDs` avoids copying the lazy atomic caches, and the concurrent marshal/lookup test covers the race it fixes.

Benchmarks in the body: `cloneDataFileAvroFields` 2.00 µs → 0.46 µs (−77%), `MarshalAvroEntry` −12% v1, −9% v2, −15% v3, allocations unchanged.

No findings. CI green (15/15), both commits signed off.

---

>

PR #1967, opened 2026-08-29
perf(table): index schemas by ID

fallintoplace:
## Summary

- **Added schema ID indexes** to metadata and `MetadataBuilder`.
- Keeps indexes current when schemas are added or removed.
- Uses copy-on-write for builder clones and built metadata.
- **Added regression tests and benchmarks** for lookup behavior, stale slices, cloning, and concurrent reads.

## Benchmark

Command: `go test ./table -run ^ -bench "Benchmark(CommonMetadataSchemaByID|MetadataBuilderGetSchemaByID)" -benchtime=50ms -count=1`

- **8,192 schemas, builder last:** about 9 ns/op indexed vs 7.4 us/op linear.
- **8,192 schemas, builder miss:** about 150 ns/op indexed vs 7.8 us/op linear.
- **8,192 schemas, metadata last:** about 200 ns/op indexed vs 7.7 us/op linear.
- Metadata lookups still return defensive schema copies.

## Tests

- `go test ./table -count=1`
- `go test -race ./table -run "SchemaIndex|SchemaByID|MetadataBuilderCloneCoversAllFields" -count=1`
- `go vet ./table`

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -346,8 +346,7 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 
 	type Alias Schema
 
-	aliasCopy := *(*Alias)(s)
-	aliasCopy.IdentifierFieldIDs = ids
+	aliasCopy := Alias{ID: s.ID, IdentifierFieldIDs: ids}
 
 	return json.Marshal(struct {
 		Type   string        `json:"type"`
--- a/schema_test.go
+++ b/schema_test.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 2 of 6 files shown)


zeroshade (approved):
This safely adds schema-ID indexing without changing the public metadata API or the existing defensive-copy contract. Approving.

The sibling change #1910 (partition specs by ID) had a genuine duplicate-ID race — a read-only path rebuilt a shared index and wrote it back. I checked for that same class here and it isn't present:

- `commonMetadata.schemaByID` (`table/metadata.go:2230-2241`) rebuilds a stale index **into a local** and never assigns it back to shared metadata, so concurrent read-only lookups can't race.
- The builder shares its index only as immutable state (`:539-555`, `:611-619`) and does copy-on-write before `AddSchema` mutates it (`:693-697`); `RemoveSchemas` rebuilds (`:1700-1713`).
- Unknown builder schema IDs return `ErrInvalidArgument` (`:1461-1470`) rather than a zero-value schema. Metadata's missing-ID path stays nil, as before.
- `Schemas()` and `CurrentSchema()`/indexed lookups keep deep defensive cloning (`:2223-2234`, `:2360-2388`).
- Only private derived fields and helpers are added — the exported `Metadata` and `MetadataBuilder` API is unchanged.

One difference from #1910 worth recording, since the two PRs look like siblings and the defect class does *not* carry over: duplicate schema IDs are already rejected on base main by `checkSchemas` (`:2642-2661`), and every production route reaches it — `ParseMetadataBytes` via each version's `UnmarshalJSON`, and `MetadataBuilder.Build` before returning metadata. In #1910 the duplicate rejection was introduced *by* that PR and duplicates were reachable in the index path; here they aren't. The `sourceCount` guard additionally tolerates duplicate-ID state in direct in-package fixtures, so a concurrent duplicate-ID test would be optional defence-in-depth rather than a coverage gap.

Committed benchmarks span 1–8,192 schemas with first/middle/last/miss lookups on both metadata and builder paths; the body reports roughly 9 ns/op indexed vs 7.4 µs/op linear for builder-last, 150 ns/op vs 7.8 µs/op for a miss, and ~200 ns/op vs 7.7 µs/op for metadata-last — while indexed metadata lookups still return defensive copies.

No findings. CI green (15/15), all 3 commits signed off.

---

>

PR #1968, opened 2026-08-29
perf(table): load deletion-vector Puffin groups lazily

fallintoplace:
## Summary

- Index deletion-vector ownership without opening Puffin files during scan setup.
- Load each Puffin group on first use and cache the decoded bitmaps with single-flight behavior.
- Keep the existing same-file grouping, range coalescing, and duplicate-DV validation.
- Start scan workers when the iterator is consumed, so an unread iterator does not load DV files.

## Benchmark

Apple M1 Pro. 100 Puffin groups with 10 DV blobs per group.

| case | Puffin opens/op | ns/op | B/op |
| --- | ---: | ---: | ---: |
| eager all groups | 100 | 2.75 ms | 3.41 MB |
| lazy unread | 0 | 0.23 ms | 0.33 MB |
| lazy first group | 1 | 0.30 ms | 0.36 MB |
| lazy ten groups | 10 | 0.62 ms | 0.64 MB |
| lazy full scan | 100 | 2.71 ms | 3.40 MB |

## Tests

- `go test -race ./table/... -count=1`
- `go vet ./...`


zeroshade (changes requested):
The Puffin lazy-loading machinery is solid, but there's one validation gap in the new code path that can silently return deleted rows. That needs fixing before this lands.

### Blocking — empty `referenced_data_file` silently drops a deletion vector (`table/arrow_scanner.go:234-263`)

`collectUniqueDeletionVectors` rejects a nil ref but permits an empty one:

```go
_, _, ref, contentOffset, contentSize := iceinternal.BorrowedDataFilePointers(d)
if ref == nil {
    return nil, fmt.Errorf("deletion vector %s missing referenced_data_file", d.FilePath())
}
```

With `*ref == ""` this falls through to `uniqueDVs[*ref] = d`, indexing the DV under the empty key. The worker then looks it up by the task's actual data-file path (`:1937`), which is never `""`, so the DV is never found and the scan proceeds as though that data file had no deletion vector — returning rows that have been deleted.

What makes this blocking rather than a hardening nit is that the path it replaces already rejected this. `table/dv/deletion_vector.go:339-341` on `main`:

```go
if manifestReferencedDataFile == nil || *manifestReferencedDataFile == "" {
    return fmt.Errorf("%w: DV file %s missing or empty %s property", ErrInvalidDeletionVector, ...)
}
```

Note the `|| *... == ""`. So this is a regression in failure mode: an input the codebase deliberately fails closed on now silently produces wrong query results.

This reads as a simple oversight rather than a design decision — the adjacent `contentOffset`/`contentSize` check in the same function has a careful comment about surfacing the real cause instead of a misleading dedup error, so the validation intent is clearly there.

Two ways to fix it:

1. Minimal: `if ref == nil || *ref == ""` at `:239`, ideally wrapping `ErrInvalidDeletionVector` so the error identity matches the eager path it replaced.
2. Preferred: validate every DV entry up front through the existing shared validator, so the lazy and eager paths cannot drift apart again. The bug exists precisely because a second validation site was introduced with slightly weaker rules.

Please also add a regression test with a DV entry whose `referenced_data_file` is empty, asserting an error rather than a successful scan with deleted rows present.

### Everything else checks out

To be clear about scope — the rest of the lazy DV work looks right:

- **Puffin integrity is preserved.** The lazy load calls the unchanged `dv.ReadDVs` (`arrow_scanner.go:199-220`), which still goes through `puffin.NewReader` header/footer magic validation, footer metadata index and range checks, and `DeserializeDV` magic/length/CRC/bitmap plus manifest-cardinality checks, and defers `file.Close`.
- **Absent vs failed is distinguished.** `ReadDVs` returns `nil, error` on any partial group failure, so a failed read cannot masquerade as "no deletes". A genuinely empty bitmap is non-nil and correctly means no deletes.
- **Concurrency.** The group `sync.Once` is per Puffin path and caches `group.err`, so every waiter observes the same error and each Puffin file is read once. Shared-file, concurrent, and cached-open-error tests cover it.
- **No leak on abandonment.** An unread iterator opens no Puffin reader, and the open-count test pins that.

### Minor — lazy-path corruption and abandonment coverage

Tests cover valid shared-file on-demand loading (`lazy_deletion_vector_test.go:35-65`), concurrent single-flight (`:68-97`), cached open errors (`:113-139`), and unread-iterator open count (`:158-181`), but none drives a corrupt or truncated Puffin *through* `lazyDeletionVectorLoader`, or abandons a live iterator mid-read. The lower-level `table/dv` tests cover corruption at the `ReadDV`/`Deserialize` level, so this is a gap in the new layer rather than an absence of coverage overall.

### Not a finding, noted for completeness

Context is checked before and after the synchronous `ReadDVs`, but `File` has no context-aware `ReadAt`, so cancellation can't interrupt an in-flight read. That limitation is inherited from the eager path, not introduced here.

CI green (15/15), all 3 commits signed off.

---

>

table/arrow_scanner.go
```diff
@@ -144,31 +144,109 @@ func readAllDeleteFiles(ctx context.Context, fs iceio.IO, tasks []FileScanTask,
...
+	}
+
+	group.once.Do(func() {
+		bitmaps, err := dv.ReadDVs(l.fs, group.files)
+		if err != nil {
+			group.err = fmt.Errorf("read deletion vectors from %s: %w", group.puffinPath, err)
+
+			return
+		}
+		if err := ctx.Err(); err != nil {
```
laskoviymishka: This one survived the rebase, and I'd still pull it out of the `once.Do` body. It can turn a successful read into a permanent failure for the whole group.

Once `ReadDVs` has returned cleanly the bitmaps are correct and safe to cache. But the goroutine that wins the `Once` runs on `scanCtx`, which any other worker can cancel. So if worker C hits an I/O error and calls `cancel(err)` just after worker A finished `ReadDVs`, A sees the cancelled ctx here, writes `group.err = context.Canceled`, and drops the bitmaps it just read. Worker B, waiting on the same `Once`, then unblocks and gets a cancellation that originated in a completely unrelated file.

It also bites on plain double-iteration: range the returned `iter.Seq2`, `break` early (which fires `cancel(nil)`), then range it again. `once` is already spent, so every group touched in the first pass returns the stale cancellation even though the fresh pass has a live context.

The post-`Do` block just below already does `if err := ctx.Err(); err != nil { return nil, err }` against the current caller's ctx, which is the right place to surface cancellation. I'd delete this in-`Once` check entirely and let that handle it. wdyt?

table/arrow_scanner.go
```diff
@@ -144,31 +144,109 @@ func readAllDeleteFiles(ctx context.Context, fs iceio.IO, tasks []FileScanTask,
...
+			group.err = err
+
+			return
+		}
+		if len(bitmaps) != len(group.referencedDataFiles) {
```
laskoviymishka: Small one: `ReadDVs` either returns `len(dvFiles)` bitmaps or an error, never a short slice on success, so this branch can't fire, and if it somehow did, the index wiring just below would panic out-of-bounds first anyway. I'd either drop it or leave a `//nolint` note that it's a belt-and-suspenders assertion so it doesn't read as load-bearing.

table/arrow_scanner.go
```diff
@@ -2177,12 +2291,10 @@ func (as *arrowScan) GetRecords(ctx context.Context, tasks []FileScanTask) (*arr
...
-	// DV bitmaps stay in their native form rather than being materialized
-	// into int64 positions and merged with the Parquet pos-delete map.
-	// filterByDeletionVector applies the bitmap to each batch via a Boolean
-	// keep-mask + compute.Filter — O(1) Contains lookups, vectorized Filter,
-	// no intermediate position set.
-	dvBitmaps, err := readAllDeletionVectors(ctx, as.fs, tasks, as.concurrency)
+	// Index DV ownership up front, but defer Puffin reads until a task using a
+	// referenced data file enters the iterator. The loader keeps each shared
+	// Puffin group cached after its first read.
+	dvLoader, err := newLazyDeletionVectorLoader(as.fs, tasks)
```
laskoviymishka, line 2196: The new commit lands the eager-structural-validation half of what I was after here: `newLazyDeletionVectorLoader` runs `collectUniqueDeletionVectors` synchronously, so missing/empty-ref and duplicate-DV violations still fail `GetRecords` up front (and now with a nice `ErrInvalidDeletionVector` sentinel). That's the right split.

What's left is the Puffin I/O errors, which used to surface from `GetRecords` and now only surface on first iteration. Two consequences worth a line of docs: a caller using `err` from `GetRecords` as a gate (`if err != nil { return }; use(iter)`) silently stops seeing DV read errors, and if it never iterates the error is dropped entirely; and if a filter or row-limit prunes away every task referencing a corrupt DV, the read error never surfaces at all, which diverges from Java, where `DeleteFileIndex` validates every DV before emitting rows. I don't think lazy-by-default is wrong, but I'd document the changed timing in the `GetRecords` godoc so callers know the DV-read error now rides on the iterator. wdyt?

table/lazy_deletion_vector_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (f *failingLazyDeletionVectorIO) Open(string) (iceio.File, error) {
+	f.opens.Add(1)
+
+	return nil, f.err
+}
+
+func (f *failingLazyDeletionVectorIO) Remove(string) error { return nil }
+
+func TestLazyDeletionVectorLoaderCachesGroupErrors(t *testing.T) {
```
laskoviymishka, line 141: Nice to see the new `TestLazyDeletionVectorLoaderRejectsEmptyReferencedDataFile` covering the empty-ref path. Once the `ctx.Err()`-inside-`once.Do` issue is fixed, I'd add a sibling test right around here that pins it: load a group through an already-cancelled context (or cancel right after `ReadDVs` returns), then load the same group again with a fresh context and assert the second call returns the real bitmaps rather than a cached cancellation. The singleflight and error-caching tests are good, but neither exercises the succeed-then-poisoned path, so that bug would regress silently.

laskoviymishka (changes requested):
Really like where this landed after the rebase. Splitting the `MarshalJSON` atomic-copy fix out to its own change keeps this PR focused on the lazy DV loader, and the follow-up commit hardening `collectUniqueDeletionVectors` (sentinel-wrapped errors plus rejecting an empty `referenced_data_file`) is a good touch that keeps structural validation eager and fail-fast.

I'd hold it before merging on one correctness bug that survived the rebase. The `ctx.Err()` check inside `once.Do` stores a context cancellation into `group.err` for good. Once `ReadDVs` has returned, the bitmaps are valid and safe to cache, but the winning goroutine runs on the shared `scanCtx`, so an unrelated worker's failure (or a plain early `break` and re-range of the iterator) can cancel it, drop the bitmaps we just read, and hand every waiter on that group a cancellation that came from somewhere else. The post-`Do` check already surfaces caller cancellation the right way, so this is a delete-the-inner-check fix plus a regression test.

- Drop the `ctx.Err()` check inside `once.Do`, and add a test for the succeed-then-cancelled / double-iteration path.
- Document the new error timing. Structural DV errors still fail `GetRecords` up front (nice), but the actual Puffin read errors now surface on first iteration, and get skipped entirely if the referencing tasks are pruned, which diverges from Java. At minimum a godoc note on `GetRecords`.

The rest are minors and I've left them inline. Once the cache bug is sorted, happy to take another pass.

zeroshade (changes requested):
The empty-`referenced_data_file` blocker from last round is genuinely fixed and I verified it end to end. One new blocker, which @laskoviymishka also spotted after your head commit — I reproduced it deterministically, and it's a five-line deletion.

## Blocking — `table/arrow_scanner.go:209-213`: the `ctx.Err()` check inside `once.Do` poisons the cache

```go
group.once.Do(func() {
    bitmaps, err := dv.ReadDVs(l.fs, group.files)
    if err != nil { group.err = …; return }
    if err := ctx.Err(); err != nil {   // :209  <-- delete
        group.err = err
        return
    }
```

By the time control reaches `:209` the Puffin has been opened, its footer decoded, every blob range read, and every bitmap CRC- and cardinality-checked. The bitmaps are correct and cheap to keep. Instead they're dropped and `group.err` is set to a cancellation that has nothing to do with this file — permanently, because `once` is spent.

The winner of the `Once` runs on `scanCtx`, which any worker can cancel (`:2199`, `:2206`, `:2216`) and which `createIterator` cancels unconditionally on exit (`:2115` `defer cancel(nil)`), so the trigger isn't exotic.

Reproduced with an `iceio.IO` whose `Open` cancels the ctx as a side effect of succeeding: `dv.ReadDVs` completes cleanly, then `:209` stores `context.Canceled` — `group.err = context canceled`, `group.bitmaps = 0 entries`, `opens = 1`. A second `load` on `context.Background()` returns `context canceled` instead of the bitmaps that were successfully decoded.

Stating the blast radius precisely so it isn't oversold: within a single iteration the poison only becomes reachable once `scanCtx` is already dead, so the damage there is diagnostic — worker A can emit a spurious `context.Canceled` record at `:2198` that races the real cause into the sequencer and masks it. Across a *second* pass over the same `iter.Seq2` it can fail a healthy re-read, but I could not make that variant fail deterministically (pass1 4 rows, pass2 4 rows, both clean) because the poison needs the cancel to land in the window between `ReadDVs` returning and `:209`.

It's blocking anyway because the cache is knowingly storing a wrong permanent state, the fix is deleting five lines, and the post-`Do` check at `:227-229` already surfaces the caller's cancellation against the caller's own context — which is the correct place and the correct context. Please also add the regression test @laskoviymishka asked for: load a group, cancel after `ReadDVs` returns, load the same group on a fresh context, assert real bitmaps.

## Major

**1. `table/arrow_scanner.go:307` — `readAllDeletionVectors` has no production caller.** `grep -rn readAllDeletionVectors --include=*.go . | grep -v _test.go` returns only the definition at `:307` and comments at `:67`, `:291`, `:382`, `:389`. It survives solely because `dv_scanner_read_test.go` (including a compile-time signature assertion at `:146`) and `lazy_deletion_vector_bench_test.go:127` call it. `golangci-lint` doesn't flag it because test usage counts as usage. That leaves ~70 lines of production-shaped code — its own errgroup fan-out, its own result plumbing — shipping in the binary with nothing exercising it outside tests. This is exactly the drift vector called out last round: the empty-ref regression existed *because* a second validation site was introduced. Either delete it and re-point `dv_scanner_read_test.go` at `lazyDeletionVectorLoader`, or keep it deliberately and say so.

**2. `table/scanner.go:1644-1647` — the public error contract moved but the godoc didn't.** Structural DV violations still fail up front, but Puffin *read* errors now ride on the iterator, and a row limit can drop them entirely. Measured with 65 tasks and a corrupt DV on the last one: unlimited surfaces the error; `UseRowLimit(1)`, `(10)`, `(50)` return rows and **never surface it**; `(100)` surfaces it. Under the eager path `GetRecords` failed regardless of the limit, and Java's `DeleteFileIndex` validates every DV before emitting rows. A caller using `err` from `ReadTasks`/`ToArrowRecords` as a gate silently stops seeing DV read errors. Needs a godoc sentence on `ReadTasks` and ideally `ToArrowRecords`.

## What checks out

- **Deferral is real, in bytes not just opens.** Wrapping `ReadAt`/`Read` on `*.puffin` handles: after `ReadTasks` returned, `opens=0, bytes=0`. After full iteration of 8 tasks sharing one Puffin: `opens=1, bytes=2658`, 32 rows.
- **Concurrent demand is clean.** 256 goroutines against one shared group holding 8 blobs, `-race -count=5`: `opens=1`; every caller gets the *identical* `*dv.RoaringPositionBitmap` (`require.Same`); ref *i* owns positions *i* and *i+8*; no cross-wiring, no races. No partially-initialized read is possible — every reader passes through `once.Do` before touching `bitmaps`/`err`.
- **Resource lifetime clean.** `opens == closes` in the loader-direct case, the full pipeline, and the abandoned-iterator case (0/0).
- **Legitimately-empty DVs are not rejected** — a zero-cardinality Puffin blob loads fine and returns non-nil with `Cardinality()==0`, correctly distinct from "no DV". `""` and nil refs are rejected; `" "` and `"\t"` are accepted, matching `main`'s `validateDVFile`, so no divergence.
- **No deadlock on the error path** — 128 tasks each with an unopenable DV terminates cleanly.
- **Benchmark table is not stale.** `puffin-opens/op` (100/0/1/10/100) and `B/op` (3.41MB/0.33MB/0.36MB/0.64MB/3.40MB) reproduce exactly on Intel Ultra 7 155H; reverse-applying the final commit's non-test hunk leaves `lazy_unread` unchanged.
- No pre-existing test file was modified (`git diff` vs merge-base shows only the two new files, +400/-0), so nothing was silently weakened.
- `-race -count=5` clean (170s); `golangci-lint` 0 issues; CI 15/15.

**Perf, interleaved A/B** (10 alternating samples, benchstat) — setup-only improves 88–95%, and the shared-Puffin full-read shape shows **no significant regression** (p = 0.529 / 0.912 / 0.684). Per-file-Puffin `limit_1` improves 15–44%. Worth recording that my first non-interleaved pass showed shared_puffin full_read at +36%/+23%; that was machine load from sibling agents, and interleaving refutes it. Don't re-derive that number.

## Minor

- `:214-219` — dead branch. `dv.ReadDVs` returns `make([]*RoaringPositionBitmap, len(dvFiles))` or an error, never a short slice, and `groupDeletionVectors` appends `referencedDataFiles` and `files` in lockstep. Raised last round, still open.
- `:154` — `lazyDeletionVectorLoader.groups` is never read in production; it exists for test assertions. Comment it as test-facing or assert on `byDataFile` instead.
- `:257` — the `content_offset/content_size_in_bytes` error is the only one of the three checks in `collectUniqueDeletionVectors` not wrapped with `dv.ErrInvalidDeletionVector`, now that its neighbours are.
- Eager structural validation is only partial: `dv.validateDVFile`'s format and `contentSize` checks still run inside `ReadDVs` at demand time, so those surface on the iterator.
- `lazy_deletion_vector_test.go:68-78` pins the empty-ref rejection at the constructor, not through a scan. I verified the scan-level behaviour is correct, but nothing in the repo pins it.
- Still no test drives a *corrupt or truncated* Puffin through the lazy loader, and none abandons a live iterator mid-read. Raised last round, still open.

## Prior items

- **Blocking: empty `referenced_data_file` silently drops a DV → Fixed.** `:247-250` adds the check in the *shared* `collectUniqueDeletionVectors`, so the eager path inherits it. Verified two ways: constructor returns `ErrInvalidDeletionVector`, and end-to-end a well-formed Puffin deleting positions 1 and 3 with an empty ref makes `ReadTasks` fail closed with a nil iterator rather than returning 10 rows.
- **Preferred single-shared-validator fix → Partially fixed.** ref/offset/dup checks are shared; `dv.validateDVFile` remains a separate later-running validator, and the eager path survives as dead code (Major 1).
- **Regression test for the empty ref → Partially fixed** (constructor-level, not "a successful scan with deleted rows present").
- **Corruption / mid-read-abandonment coverage → Still open.**
- @laskoviymishka's four items (all filed *after* your head commit, so none are addressed): `ctx.Err()` poisoning → **Still open**, reproduced; succeed-then-cancelled test → **Still open**; dead `len(bitmaps)` branch → **Still open**; document the error timing → **Still open**.

## Description

- Bullet 4 ("Start scan workers when the iterator is consumed, so an unread iterator does not load DV files") describes structure that already exists at the merge base — `git show 87789102:table/arrow_scanner.go` already wraps the worker spawn in `return func(yield …)`. The DV *consequence* is new and real (0 opens, 0 bytes at setup), but this PR doesn't change when workers start.
- Omits that `readAllDeletionVectors` is left with no production caller.
- Omits the error-timing change (Puffin read errors move from `GetRecords` to the iterator and can be skipped entirely under a row limit).

---

table/scanner.go
```diff
@@ -1872,6 +1872,10 @@ type FileScanTask struct {
 // If an error is encountered, during the planning and setup then this will return the
 // error directly. If the error occurs while iterating the records, it will be returned
 // by the iterator.
+// Deletion-vector references are validated during setup. Deletion-vector Puffin files
+// are loaded lazily, so a Puffin read error is returned by the iterator when a task
```
zeroshade, line 1876: **minor** — 'validated during setup' overstates what is checked eagerly

collectUniqueDeletionVectors validates ref presence/non-emptiness, content_offset/content_size presence, and duplicate-DV conflicts. It does NOT check file_format or the content_size range — those live in dv.validateDVFile, which now only runs inside ReadDVs at demand time. So a structurally invalid DV entry (e.g. file_format=PARQUET) passes setup and, if a filter or row limit prunes away every task referencing it, is never validated at all, where main failed GetRecords unconditionally. Suggest softening to something like 'Deletion-vector manifest references are validated during setup; format and blob-level validation happens with the Puffin read.' Same text at :1894 on ReadTasks.

table/lazy_deletion_vector_test.go
```diff
@@ -0,0 +1,294 @@
...
+		f.opens.Add(1)
+	}
+
+	return file, err
+}
+
+func (f *countingPuffinOpenIO) Remove(name string) error { return f.base.Remove(name) }
+
+func TestReadTasksDoesNotLoadDeletionVectorsBeforeIteration(t *testing.T) {
+	baseFS := iceio.LocalFS{}
```
zeroshade, line 239: **minor** — No test abandons a partially-consumed iterator

TestReadTasksDoesNotLoadDeletionVectorsBeforeIteration covers the never-iterated case (opens==0) and the fully-drained case (opens==1), but nothing breaks out of an iterator that has already forced a Puffin load. That is the shape where a worker can be parked inside once.Do while the consumer walks away, so it is worth pinning. A Close-counting iceio.IO asserting opens==closes after an early break would cover it in ~20 lines.

table/arrow_scanner_test.go
```diff
@@ -430,7 +430,7 @@ func TestReleasePerFilePosDeletes(t *testing.T) {
 	t.Run("nil chunk in slice does not panic", func(t *testing.T) {
 		// Defensive: production code paths never insert a nil *arrow.Chunked
 		// into the map. This subtest pins the guard so a future caller (the
-		// in-flight readAllDeletionVectors merger, or a refactor of readDeletes)
+		// in-flight delete-file merger, or a refactor of readDeletes)
 		// can't silently NPE the cleanup path.
```
zeroshade, line 434: **nit** — Comment points at a plan that no longer exists

The comment was rewritten from 'the in-flight readAllDeletionVectors merger' to 'the in-flight delete-file merger' when the eager loader was deleted. There is no in-flight merger any more; the sentence now names nothing. Suggest just describing the guard ('a future caller, or a refactor of readDeletes').

table/lazy_deletion_vector_bench_test.go
```diff
@@ -0,0 +1,157 @@
...
+				DeletionVectorFiles: []iceberg.DataFile{file},
+			})
+		}
+	}
+
+	return fixture
+}
+
+func BenchmarkLazyDeletionVectorLoading(b *testing.B) {
+	fixture := newLazyDVBenchmarkFixture(b)
```
zeroshade, line 100: **nit** — Description's benchmark table cites an 'eager all groups' row that is no longer reproducible

BenchmarkLazyDeletionVectorLoading only defines lazy_unread / lazy_first_group / lazy_ten_groups / lazy_full_scan. The 'eager all groups' baseline came from BenchmarkReadAllDeletionVectorsSharedPuffin, deleted in 17f771a. Worth a note in the PR body that the baseline row was measured against the pre-PR tree, so a reviewer does not go looking for the benchmark.

zeroshade (approved):
Both prior blockers (once.Do cache poisoning, empty referenced_data_file) are fixed and each is pinned by a test that goes red under mutation; the dead eager loader is gone, the error-timing godoc landed, and only doc-precision and test-coverage minors remain.

Re-review verification: 8 of 11 prior findings confirmed fixed at `17f771a` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *not fixed* — Minor: no test abandons a live iterator mid-read
- *not fixed* — Minor: eager structural validation is only partial — dv.validateDVFile's format and contentSize checks still run at demand time
- *not fixed* — Description bullet 4 ('Start scan workers when the iterator is consumed') describes structure already present at the merge base

```text
In /Users/zeroshade/Projects/iceberg-go/.pi-worktrees/pr1968 @ 17f771a: `go build ./table/...` PASS; `go vet ./table/...` PASS; `go vet ./catalog/` PASS (refutes the automated '[setup failed]' reports, which used package paths outside the worktree module). `go test ./table/ ./table/dv/ -count=1` PASS (7.9s / 0.3s). `go test -race ./table/ ./table/dv/ -count=1` PASS (20.1s / 2.4s). `go test -race ./table/ ./table/dv/ -run 'DeletionVector|DVScan|LazyDeletion|ReadTasks|DV' -count=2` PASS. Mutation A (re-add ctx.Err() inside once.Do) -> TestLazyDeletionVectorLoaderDoesNotCacheCallerCancellation FAIL, reverted. Mutation B (drop the *ref=="" check) -> TestLazyDeletionVectorLoaderRejectsEmptyReferencedDataFile + TestReadTasksRejectsEmptyReferencedDataFile FAIL, reverted. Mutation C (rotate ref->bitmap wiring) -> TestLazyDeletionVectorLoaderLoadsSharedPuffinGroupsOnDemand FAIL, reverted. Three throwaway probes (error fidelity under shared corrupt Puffin; open/close accounting on abandon-mid-read and full scan; non-PUFFIN format validation timing) written, run, and deleted. `git status --short` empty at HEAD 17f771a.
```

</details>

---

>

laskoviymishka (approved):
:shipit:

laskoviymishka:
@fallintoplace can you fix a conflicts? I think this is good to merge, thanks!

zeroshade (approved):
APPROVE. Since the approval at `17f771a`, the only new commit is merge `0ba1822e`; `git range-diff` keeps all three PR commits identical, and merge reconstruction shows manual resolution was confined to `table/arrow_scanner.go` and `table/scanner.go`.

I re-reviewed both resolutions. The worker now composes main's lazy positional-delete loader with this PR's lazy DV loader (`table/arrow_scanner.go:2225-2324`). DV state is still published through `sync.Once` only after `dv.ReadDVs` fully succeeds (`table/arrow_scanner.go:318-351`); caller cancellation remains outside `once.Do`, the Puffin handle closes before return (`table/dv/deletion_vector.go:250-327`), and positional-delete chunks are released only after all workers stop (`table/arrow_scanner.go:2159-2171`). I found no race, double-close, use-after-release, or half-initialized cache path.

The prior blockers remain addressed: the `once.Do` cache poisoning stays fixed and its regression passes; empty `referenced_data_file` still fails closed at `table/arrow_scanner.go:360-367`. The prior validation-timing doc minor is now addressed at `table/scanner.go:2008-2035`. Targeted `-race` runs for the merged lazy-loader/DV scan paths pass, and current-head CI is 15/15 green.

Non-blocking, and not worth another round: `table/arrow_scanner.go:2328-2334` retained main's pre-merge comment saying deletion-vector errors return before iterator creation. Puffin read errors are now lazy, so that sentence contradicts both the implementation and the corrected exported docs. Worth fixing next time you touch the file.

PR #1969, opened 2026-08-29
perf(table): fetch orphan-cleanup manifest lists concurrently

fallintoplace:
## Summary

- **Reads snapshot manifest lists concurrently** using the existing cleanup concurrency limit.
- Stores each snapshot result separately, then folds results in snapshot order before deduplicating manifests.
- Calls `metadata.Snapshots()` once and adds regression and benchmark coverage.

## Benchmark

Representative `-benchtime=1x` results with a 1 ms artificial `Open` delay on an Apple M1 Pro:

| Snapshots | Concurrency 1 | Concurrency 16 |
| --- | ---: | ---: |
| 16 | 22.9 ms | 3.4 ms |
| 128 | 171.2 ms | 13.3 ms |
| 1,024 | 1.40 s | 99.6 ms |

## Tests

- `go test -race ./table/... -count=1`
- `go vet ./...`

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -346,8 +346,7 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 
 	type Alias Schema
 
-	aliasCopy := *(*Alias)(s)
-	aliasCopy.IdentifierFieldIDs = ids
+	aliasCopy := Alias{ID: s.ID, IdentifierFieldIDs: ids}
 
 	return json.Marshal(struct {
 		Type   string        `json:"type"`
--- a/schema_test.go
+++ b/schema_test.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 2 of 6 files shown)


zeroshade (approved):
The safety-critical ordering is correct, which is what matters most here given this path deletes files: any manifest-list fetch error is returned by `listGroup.Wait()` (`table/orphan_cleanup.go:531`) **before** the per-snapshot results are folded into `referenced`. So a partial reference set can never reach orphan planning or file deletion. Approving, with two minor follow-ups.

Also verified:

- Fan-out is bounded by `SetLimit(max(1, min(maxConcurrency, len(snapshots))))`.
- Each worker writes only its own result slot, so no race on the accumulator.
- The fold preserves snapshot order, and reference membership stays complete and order-independent.
- Context is checked before and after each `Manifests(fs)` call. (Caveat, inherited rather than introduced: an in-flight `Open` can't be interrupted through the current IO API.)

### Minor — nondeterministic error selection among simultaneous failures (`table/orphan_cleanup.go:531`)

`errgroup.Wait` returns the first non-nil error, so when several manifest-list fetches fail you get whichever completed first. `Snapshot.Manifests` wraps backend open errors (a `fs.ErrNotExist` remains discoverable through the wrap) and Avro/decode errors, then this function adds the snapshot ID.

I'm calling this `minor` rather than `major` deliberately: no in-repo caller dispatches on a manifest-fetch sentinel and the docs promise no specific error, so the practical impact is nondeterministic diagnostics rather than changed control flow. That's the distinction from #1959, where the affected errors are catalog sentinels that callers genuinely branch on.

Suggested fix, cheap either way: capture errors by snapshot index and return a deterministic snapshot-order selection or `errors.Join`, excluding sibling `context.Canceled` errors caused by the group's own cancellation.

### Minor — no partial-failure test on a destructive path (`table/orphan_cleanup_test.go:2262`)

The new regression test covers successful concurrent reads and the worker bound, but not a partial manifest-list failure. To be unambiguous: **this is a coverage gap, not evidence that references are under-reported** — I verified the abort-before-fold ordering directly above. Still, given the downstream action is deleting files, a test that fails one list open among several and asserts an error with no plan and no deletion is worth having as a permanent guard.

Benchmark evidence is provided (`-benchtime=1x` with a 1 ms artificial `Open` delay: 16 snapshots 22.9 → 3.4 ms; 128 171.2 → 13.3 ms; 1,024 1.40 s → 99.6 ms).

The one failing check is the known pre-existing SQLite lock-contention flake tracked in #1793; neither changed production file touches `catalog/sql`, so it's unrelated and needs nothing from you. Both commits signed off.

---

>

PR #1970, opened 2026-08-29
perf(table): cache snapshot manifests across scans

fallintoplace:
## What

- **Reuse decoded manifest lists** across local scans, incremental append scans, manifest/file/entry inspectors, and `AllManifests`.
- **Share concurrent reads.** IO setup and reads keep the caller's cancellation. Healthy waiters retry with their own IO if that caller cancels.
- **Bound retention** to 64 snapshots and 32,768 manifest descriptors per table handle. Memory also depends on paths, partition summaries, and key metadata.
- **Allow opt-out** with `read.manifest-list-cache.enabled=false`. Refresh replaces the cache and picks up property changes. Transaction scans use a separate cache and respect staged properties.
- Keep the data-manifest group ready for incremental append planning. Filters and projections are still applied per scan.

## Benchmark

The benchmark compares caching disabled and enabled on the same decoding path. Each iteration starts with a cold cache and performs 10 scans. With one manifest, manifest-list opens drop from **10 to 1**. Each scan still reads the manifest entries.

## Checks

- Full test suite.
- Race detector on table tests.
- Lint.
- Regression tests cover canceled IO setup and reads, healthy waiters, early `AllManifests` stop, cache opt-out, and descriptor limits.

### DIFF
--- a/table/inspect_files.go
+++ b/table/inspect_files.go
@@ -131,13 +131,11 @@ func (i InspectTable) manifestEntryReader(
 		return nil, errors.New("table file IO is not configured")
 	}
 
-	fs, err := i.tbl.fsF(ctx)
+	manifestSet, err := i.tbl.manifestSetWithFSF(ctx, *snapshot, sharedSnapshotManifestFSF(i.tbl.fsF))
 	if err != nil {
 		return nil, err
 	}
-	manifestSet, err := i.tbl.manifestSetWithFSF(ctx, *snapshot, func(context.Context) (iceio.IO, error) {
-		return fs, nil
-	})
+	fs, err := i.tbl.fsF(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -176,7 +174,7 @@ func (i InspectTable) allManifestEntryReader(
 	if err != nil {
 		return nil, err
 	}
-	manifestFS := func(context.Context) (iceio.IO, error) { return fs, nil }
+	manifestFS := sharedSnapshotManifestFSF(i.tbl.fsF)
 
 	return i.manifestEntryReaderFromManifestSource(
 		ctx, arrowSchema, fs, discardDeleted, includeManifest, appendEntry,
--- a/table/inspect_files.go
+++ b/table/inspect_files.go
@@ -135,7 +135,9 @@ func (i InspectTable) manifestEntryReader(
 	if err != nil {
 		return nil, err
 	}
-	manifestSet, err := i.tbl.manifestCache.get(ctx, *snapshot, fs)
+	manifestSet, err := i.tbl.manifestSetWithFSF(ctx, *snapshot, func(context.Context) (iceio.IO, error) {
+		return fs, nil
+	})
 	if err != nil {
 		return nil, err
 	}
@@ -174,6 +176,7 @@ func (i InspectTable) allManifestEntryReader(
 	if err != nil {
 		return nil, err
 	}
+	manifestFS := func(context.Context) (iceio.IO, error) { return fs, nil }
 
 	return i.manifestEntryReaderFromManifestSource(
 		ctx, arrowSchema, fs, discardDeleted, includeManifest, appendEntry,
@@ -185,7 +188,7 @@ func (i InspectTable) allManifestEntryReader(
 
 					return
 				}
-				manifestSet, err := i.tbl.manifestCache.get(ctx, snapshot, fs)
+				manifestSet, err := i.tbl.manifestSetWithFSF(ctx, snapshot, manifestFS)
 				if err != nil {
 					yield(nil, fmt.Errorf("read snapshot %d manifests: %w", snapshot.SnapshotID, err))
 
--- a/table/inspect_files.go
+++ b/table/inspect_files.go
@@ -135,10 +135,11 @@ func (i InspectTable) manifestEntryReader(
 	if err != nil {
 		return nil, err
 	}
-	manifests, err := snapshot.Manifests(fs)
+	manifestSet, err := i.tbl.manifestCache.get(ctx, *snapshot, fs)
 	if err != nil {
 		return nil, err
 	}
+	manifests := manifestSet.allManifests()
 
 	return i.manifestEntryReaderFromManifestSource(
 		ctx, arrowSchema, fs, discardDeleted, includeManifest, appendEntry,
@@ -184,13 +185,13 @@ func (i InspectTable) allManifestEntryReader(
 
 					return
 				}
-				snapshotManifests, err := snapshot.Manifests(fs)
+				manifestSet, err := i.tbl.manifestCache.get(ctx, snapshot, fs)
 				if err != nil {
 					yield(nil, fmt.Errorf("read snapshot %d manifests: %w", snapshot.SnapshotID, err))
 
 					return
 				}
-				for _, manifest := range snapshotManifests {
+				for _, manifest := range manifestSet.allManifests() {
 					if _, ok := seen[manifest.FilePath()]; ok {
 						continue
 					}
--- a/table/inspect_files.go
+++ b/table/inspect_files.go
@@ -131,11 +131,12 @@ func (i InspectTable) manifestEntryReader(
 		return nil, errors.New("table file IO is not configured")
 	}
 
-	manifestSet, err := i.tbl.manifestSetWithFSF(ctx, *snapshot, sharedSnapshotManifestFSF(i.tbl.fsF))
+	manifestFS := sharedSnapshotManifestFSF(i.tbl.fsF)
+	manifestSet, err := i.tbl.manifestSetWithFSF(ctx, *snapshot, manifestFS)
 	if err != nil {
 		return nil, err
 	}
-	fs, err := i.tbl.fsF(ctx)
+	fs, err := manifestFS(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -170,11 +171,11 @@ func (i InspectTable) allManifestEntryReader(
 		return nil, errors.New("table file IO is not configured")
 	}
 
-	fs, err := i.tbl.fsF(ctx)
+	manifestFS := sharedSnapshotManifestFSF(i.tbl.fsF)
+	fs, err := manifestFS(ctx)
 	if err != nil {
 		return nil, err
 	}
-	manifestFS := sharedSnapshotManifestFSF(i.tbl.fsF)
 
 	return i.manifestEntryReaderFromManifestSource(
 		ctx, arrowSchema, fs, discardDeleted, includeManifest, appendEntry,
--- a/table/snapshot_manifest_cache.go
+++ b/table/snapshot_manifest_cache.go
@@ -1,0 +1,161 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+import (
+	"context"
+	"slices"
+	"strings"
+	"sync"
+
+	"github.com/apache/iceberg-go"
+	iceio "github.com/apache/iceberg-go/io"
+)
+
+// snapshotManifestSet keeps the complete manifest list and its content
+// partitions together. Manifest descriptors are immutable for an Iceberg
+// snapshot, so a table can safely share this decoded result across scans.
+type snapshotManifestSet struct {
+	all     []iceberg.ManifestFile
+	data    []iceberg.ManifestFile
+	deletes []iceberg.ManifestFile
+}
+
+func newSnapshotManifestSet(manifests []iceberg.ManifestFile) snapshotManifestSet {
+	set := snapshotManifestSet{all: slices.Clone(manifests)}
+	if len(manifests) == 0 {
+		return set
+	}
+
+	set.data = make([]iceberg.ManifestFile, 0, len(manifests))
+	set.deletes = make([]iceberg.ManifestFile, 0, len(manifests))
+	for _, manifest := range set.all {
+		if manifest.ManifestContent() == iceberg.ManifestContentDeletes {
+			set.deletes = append(set.deletes, manifest)
+		} else {
+			set.data = append(set.data, manifest)
+		}
+	}
+
+	return set
+}
+
+func (s snapshotManifestSet) allManifests() []iceberg.ManifestFile {
+	return slices.Clone(s.all)
+}
+
+func (s snapshotManifestSet) dataManifests() []iceberg.ManifestFile {
+	return slices.Clone(s.data)
+}
+
+func (s snapshotManifestSet) deleteManifests() []iceberg.ManifestFile {
+	return slices.Clone(s.deletes)
+}
+
+type snapshotManifestCacheKey struct {
+	snapshotID              int64
+	manifestList            string
+	hasEmbeddedSources      bool
+	embeddedManifestSources string
+}
+
+func snapshotManifestCacheKeyFor(snapshot Snapshot) snapshotManifestCacheKey {
+	var (
+		hasEmbeddedSources      bool
+		embeddedManifestSources string
+	)
+	if snapshot.ManifestList == "" {
+		hasEmbeddedSources = snapshot.ManifestLocations != nil
+		embeddedManifestSources = strings.Join(snapshot.ManifestLocations, "\x00")
+	}
+
+	return snapshotManifestCacheKey{
+		snapshotID:              snapshot.SnapshotID,
+		manifestList:            snapshot.ManifestList,
+		hasEmbeddedSources:      hasEmbeddedSources,
+		embeddedManifestSources: embeddedManifestSources,
+	}
+}
+
+type snapshotManifestCacheEntry struct {
+	ready     chan struct{}
+	manifests snapshotManifestSet
+	err       error
+}
+
+// snapshotManifestCache memoizes successful manifest-list reads and shares an
+// in-flight read with concurrent scans. Failed reads are removed so a
+// transient object-store error does not poison the cache.
+type snapshotManifestCache struct {
+	mu      sync.Mutex
+	entries map[snapshotManifestCacheKey]*snapshotManifestCacheEntry
+}
+
+func newSnapshotManifestCache() *snapshotManifestCache {
+	return &snapshotManifestCache{
+		entries: make(map[snapshotManifestCacheKey]*snapshotManifestCacheEntry),
+	}
+}
+
+func (c *snapshotManifestCache) get(
+	ctx context.Context,
+	snapshot Snapshot,
+	fio iceio.IO,
+) (snapshotManifestSet, error) {
+	if c == nil {
+		manifests, err := snapshot.Manifests(fio)
+
+		return newSnapshotManifestSet(manifests), err
+	}
+
+	key := snapshotManifestCacheKeyFor(snapshot)
+	c.mu.Lock()
+	if entry, ok := c.entries[key]; ok {
+		c.mu.Unlock()
+
+		select {
+		case <-entry.ready:
+			return entry.manifests, entry.err
+		default:
+			select {
+			case <-entry.ready:
+				return entry.manifests, entry.err
+			case <-ctx.Done():
+				return snapshotManifestSet{}, ctx.Err()
+			}
+		}
+	}
+
+	entry := &snapshotManifestCacheEntry{ready: make(chan struct{})}
+	c.entries[key] = entry
+	c.mu.Unlock()
+
+	manifests, err := snapshot.Manifests(fio)
+	value := newSnapshotManifestSet(manifests)
+
+	c.mu.Lock()
+	entry.manifests = value
+	entry.err = err
+	if err != nil {
+		delete(c.entries, key)
+	}
+	close(entry.ready)
+	c.mu.Unlock()
+
+	return value, err
+}
--- a/table/snapshot_manifest_cache.go
+++ b/table/snapshot_manifest_cache.go
@@ -19,6 +19,7 @@ package table
 
 import (
 	"context"
+	"fmt"
 	"slices"
 	"strings"
 	"sync"
@@ -51,10 +52,11 @@ func newSnapshotManifestSet(manifests []iceberg.ManifestFile) snapshotManifestSe
 	set.data = make([]iceberg.ManifestFile, 0, len(manifests))
 	set.deletes = make([]iceberg.ManifestFile, 0, len(manifests))
 	for _, manifest := range set.all {
-		if manifest.ManifestContent() == iceberg.ManifestContentDeletes {
-			set.deletes = append(set.deletes, manifest)
-		} else {
+		switch manifest.ManifestContent() {
+		case iceberg.ManifestContentData:
 			set.data = append(set.data, manifest)
+		case iceberg.ManifestContentDeletes:
+			set.deletes = append(set.deletes, manifest)
 		}
 	}
 
@@ -87,6 +89,8 @@ func snapshotManifestCacheKeyFor(snapshot Snapshot) snapshotManifestCacheKey {
 	)
 	if snapshot.ManifestList == "" {
 		hasEmbeddedSources = snapshot.ManifestLocations != nil
+		// Valid file locations cannot contain a literal NUL, so this join is
+		// collision-free while preserving nil versus empty locations.
 		embeddedManifestSources = strings.Join(snapshot.ManifestLocations, "\x00")
 	}
 
@@ -103,6 +107,8 @@ type snapshotManifestCacheEntry struct {
 	manifests snapshotManifestSet
 	err       error
 }
+
+type snapshotManifestLoader func(context.Context) (snapshotManifestSet, error)
 
 // snapshotManifestCache memoizes successful manifest-list reads and shares an
 // in-flight read with concurrent scans. Completed reads use a bounded LRU so a
@@ -130,12 +136,10 @@ func newSnapshotManifestCache() *snapshotManifestCache {
 func (c *snapshotManifestCache) get(
 	ctx context.Context,
 	snapshot Snapshot,
-	fio iceio.IO,
+	load snapshotManifestLoader,
 ) (snapshotManifestSet, error) {
 	if c == nil {
-		manifests, err := snapshot.Manifests(fio)
-
-		return newSnapshotManifestSet(manifests), err
+		return load(ctx)
 	}
 
 	key := snapshotManifestCacheKeyFor(snapshot)
@@ -165,10 +169,38 @@ func (c *snapshotManifestCache) get(
 	c.entries[key] = entry
 	c.mu.Unlock()
 
-	manifests, err := snapshot.Manifests(fio)
-	value := newSnapshotManifestSet(manifests)
+	var (
+		value snapshotManifestSet
+		err   error
+	)
+	defer func() {
+		if recovered := recover(); recovered != nil {
+			c.finish(key, entry, snapshotManifestSet{}, fmt.Errorf(
+				"panic while reading snapshot %d manifest list: %v", snapshot.SnapshotID, recovered))
+			panic(recovered)
+		}
 
+		c.finish(key, entry, value, err)
+	}()
+
+	value, err = load(ctx)
+
+	return value, err
+}
+
+func (c *snapshotManifestCache) finish(
+	key snapshotManifestCacheKey,
+	entry *snapshotManifestCacheEntry,
+	value snapshotManifestSet,
+	err error,
+) {
 	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if current, ok := c.entries[key]; !ok || current != entry {
+		return
+	}
+
 	delete(c.entries, key)
 	entry.manifests = value
 	entry.err = err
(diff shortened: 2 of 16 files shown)


zeroshade (changes requested):
The cross-scan caching is correctly keyed and the cached value is safe to reuse — I checked that carefully, because a cache outliving a single scan is where this kind of change usually goes wrong. The problem is that it never evicts.

### What's right — the reuse itself is sound

- **Key is not just the snapshot ID.** It's snapshot ID plus manifest-list location, with V1 embedded locations NUL-joined (`snapshot_manifest_cache.go:70-91`), so two different tables' snapshots can't collide.
- **Cached value is pre-filter.** Raw `ManifestFile` descriptors are cached, with filter, schema, and spec applied afterwards (`scanner.go:881-891`). Had anything post-filter or post-projection been cached, reuse across scans with different filters would have been silently wrong; it isn't.
- **Refresh invalidates.** The cache is table-owned and `Refresh` swaps in a fresh one (`:239-241`), so a stale entry can't survive a refresh to new metadata.
- **Snapshots are immutable** in Iceberg, which is what legitimises snapshot-keyed caching in the first place.
- **Concurrency is handled.** The map plus ready-channel handoff is race-free and all waiters observe the same error. `ManifestFile` accessors defensive-copy.

### Major — the cache grows without bound (`snapshot_manifest_cache.go:106`, `:155`)

The map is insert-only. `get` at `:155` deletes only *failed* entries; successful entries are never evicted and there's no size or byte cap.

The concrete exposure: `AllManifests` (`table.go:396`) and the historical Inspect/Incremental paths insert one full descriptor set **per snapshot**, and each set clones the all/data/delete slices. A single sweep over a long snapshot history on a long-lived `Table` handle therefore retains manifest metadata for every snapshot in that history simultaneously, for the lifetime of the handle.

In fairness, `Refresh` resets the cache, so growth is bounded by the snapshots touched between refreshes — this is unbounded retention under a specific access pattern rather than an unconditional leak. But bulk historical sweeps are exactly that pattern, and they're one-shot: they gain nothing from caching while paying its full retention cost.

Options, in order of preference:

1. Add a bounded LRU or an entry/byte cap.
2. Exclude the bulk historical paths (`AllManifests` and the Inspect/Incremental sweeps) from populating the cache, since a single pass never reuses its entries.
3. Scope the cache to an operation rather than the `Table` lifetime — closest to how the other caches in this stack behave.

A cap is the smallest change that makes the worst case predictable, so I'd start there.

Benchmark evidence is good (10 → 1 manifest-list opens, 4.26 → 2.36 ms), and the win is real — I'd like to see it land with the retention bounded.

CI green (15/15), all 3 commits signed off.

---

>

table/snapshot_manifest_cache.go
```diff
@@ -0,0 +1,182 @@
...
+				return snapshotManifestSet{}, ctx.Err()
+			}
+		}
+	}
+
+	entry := &snapshotManifestCacheEntry{ready: make(chan struct{})}
+	c.entries[key] = entry
+	c.mu.Unlock()
+
+	manifests, err := snapshot.Manifests(fio)
```
laskoviymishka: The bounding commit plumbed `ctx` into `get`, which lets a waiter bail out of its wait, but the producer still reads through a context-less `snapshot.Manifests(fio)` and nothing cleans up on an early exit. So both failure shapes are still here: if that read stalls, the in-flight `entries[key]` sits there until it returns, and every retrying waiter with a fresh ctx re-blocks against it; if it panics, the `delete` and `close` at the bottom never run and the entry is a zombie with `ready` never closed. Before this PR each caller read independently and timed out on its own; now one stuck read serializes everyone behind it until `Refresh`.

A deferred guard that deletes the entry and closes `ready` on any early return covers the panic and error cases; the stall itself needs ctx threaded into the read. I'd still want a test for the retry-after-stuck-producer path, since the canceled-waiter test covers the cancel but not the case where a waiter escapes, retries, and resolves only once the producer finishes. wdyt?

table/snapshot_manifest_cache.go
```diff
@@ -0,0 +1,182 @@
...
+
+func newSnapshotManifestSet(manifests []iceberg.ManifestFile) snapshotManifestSet {
+	set := snapshotManifestSet{all: slices.Clone(manifests)}
+	if len(manifests) == 0 {
+		return set
+	}
+
+	set.data = make([]iceberg.ManifestFile, 0, len(manifests))
+	set.deletes = make([]iceberg.ManifestFile, 0, len(manifests))
+	for _, manifest := range set.all {
```
laskoviymishka: This partitions negatively, so anything that isn't `ManifestContentDeletes` falls into `data`, where the old append-scan loop was a positive `== ManifestContentData` check. Identical with today's two content values, but the forward-compat posture flips: if a future spec rev adds a third content type, the old code kept it out of the append scan and this quietly treats it as data. Java's `ManifestContent.fromId` hard-errors on anything but 0/1. Not blocking, but a positive check for the data partition would preserve the old semantics. wdyt?

table/transaction.go
```diff
@@ -3210,6 +3210,7 @@ func (t *Transaction) Scan(opts ...ScanOption) (*Scan, error) {
 		metadata:         updatedMeta,
 		metadataLocation: t.tbl.metadataLocation,
 		ioF:              t.tbl.fsF,
+		manifestCache:    t.tbl.manifestCache,
```
laskoviymishka: Transaction scans plan over `updatedMeta` (which includes the staged snapshot) but share the committed table's cache, so a staged snapshot's manifest set lands in the committed cache under its staged ID. The bounding commit takes most of the sting out of this: those orphaned entries now age out of the LRU instead of lingering until `Refresh`, so it's smaller than it was last round. It's still a little surprising that a committed table's cache picks up snapshots that may never commit; a fresh `newSnapshotManifestCache()` for transaction scans would keep the two views cleanly separate. Not blocking. Thoughts?

table/snapshot_manifest_cache_internal_test.go
```diff
@@ -0,0 +1,286 @@
...
+			errs <- err
+		}()
+	}
+
+	select {
+	case <-fs.started:
+	case <-time.After(time.Second):
+		t.Fatal("manifest-list read did not start")
+	}
+	close(fs.release)
```
laskoviymishka: If this `t.Fatal` fires, the `close(fs.release)` never runs and all 16 goroutines are stuck: the owner inside `Open`, the 15 waiters on `entry.ready`. Under `-race` a leaked goroutine tends to surface as a false-positive race in whatever test runs next. The canceled-waiter test you added this round already has the fix, a `sync.Once` plus `t.Cleanup` that closes `release` unconditionally. Worth mirroring here, and using `t.Context()` in the goroutines while you're at it.

table/snapshot_manifest_cache.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func snapshotManifestCacheKeyFor(snapshot Snapshot) snapshotManifestCacheKey {
+	var (
+		hasEmbeddedSources      bool
+		embeddedManifestSources string
+	)
+	if snapshot.ManifestList == "" {
+		hasEmbeddedSources = snapshot.ManifestLocations != nil
+		embeddedManifestSources = strings.Join(snapshot.ManifestLocations, "\x00")
```
laskoviymishka, line 100: A one-liner noting that `\x00` can't appear in a valid file path or URI (so the join is collision-free) would make this obviously safe rather than something the next reader has to reason through.

laskoviymishka (changes requested):
This is close. The single-flight cache is the right shape, resetting it on `Refresh` is the correct lifecycle anchor, and the bounding commit landed the piece I was most concerned about last round: completed reads now live in a 64-entry LRU, so a history traversal like `AllManifests` can't retain every snapshot for the life of the table.

The one thing I'd still hold on before merging is the producer path. It inserts its entry into the in-flight map and then reads through a context-less `snapshot.Manifests`, and nothing cleans up on an early exit. So if that read stalls, waiters can abandon their own wait but the entry stays in-flight and every retry re-blocks against it until `Refresh`; if it panics, the entry is a zombie with `ready` never closed. Before this PR each caller could time out independently, so this is a real shift in failure behavior. The `ctx` that got threaded into `get` this round helps the waiters but doesn't reach the read itself.

- Guard the producer so a stalled or panicking read can't strand retrying waiters (deferred cleanup for the panic/error case, ctx into the read for the stall), plus a test for the retry-after-stuck-producer path.
- The append-scan content-type filter now partitions negatively (not-deletes becomes data) where it used to be a positive data-only check; worth restoring the old posture for forward-compat.

The rest is small: the transaction scan sharing the committed cache (the LRU already takes most of the sting out of that), a test-hygiene leak, and a comment nit, all inline. Once the producer path is settled I'm happy to approve.

zeroshade (changes requested):
The cache is correctly keyed, per-`Table`-scoped, bounded, and race-clean, and the benchmark numbers reproduce. Both my prior items and all seven of @laskoviymishka's are fixed. But the round-3 producer-context plumbing introduced a regression: a shared in-flight read is now owned by whichever caller won the race, so an unrelated scan's cancellation fails healthy waiters.

## Blocking — `table/snapshot_manifest_cache.go:186`, `:157`/`:161`, `:235`

The shared read runs under the *first* caller's context and its error is handed verbatim to every waiter, so a waiter whose own context is perfectly alive fails because an unrelated caller went away.

Direct probe on the cache:
```
producer err = context canceled
waiter   err = context canceled   (waiter ctx was context.Background())
```

End-to-end through the public API, two independent triggers:
```
scan#1 (canceled ctx)   err = context canceled
scan#2 (Background ctx) err = context canceled
AllManifests err        = context canceled
concurrent Scan err     = context canceled
```

Same setup against the physically-reverted base tree:
```
BASE scan#1 (canceled ctx)   err = context canceled
BASE scan#2 (Background ctx) err = <nil>
```

Why it's new: round 3 moved IO resolution inside the producer (`readSnapshotManifestSet:227`) precisely so a ctx-bound factory would abort a stalled read. That part works — `blobfs.FileIO` captures ctx at construction (`io/gocloud/blobfs/blob.go:389`) and uses `bfs.ctx` for every `Open`/`NewReader`. But it also makes the shared read's lifetime owned by one arbitrary caller. Pre-PR each scan resolved its own IO with its own ctx (`scanner.go:1492` at base) and was independent.

Trigger (b) is not exotic: `Table.AllManifests` does `defer cancel()` on its iterator (`table/table.go:459`), so any early consumer break — or simply a finished sweep — cancels `groupCtx` and fails a concurrent `Scan` waiting on the same snapshot.

`readSnapshotManifestSet:235` compounds it: a **fully decoded** manifest list is discarded when the producer's ctx expired after the read completed — `late-cancel: len(all)=0 err=context canceled cacheLen=0`.

Suggested fix: run the shared producer under a context detached from its initiator — `context.WithoutCancel(ctx)`, plus a cache-owned deadline if you want one — so no single caller owns the shared read; and/or on producer error let a waiter whose own ctx is still live re-drive the load rather than inheriting a foreign `ctx.Err()`. Drop the post-read `ctx.Err()` check so a completed decode is never thrown away. `TestSnapshotManifestCacheCanceledWaiterDoesNotCancelSharedRead` covers waiter→producer; the producer→waiter direction has no test.

## Major

**1. `snapshot_manifest_cache.go:53`, `:58-60`, `:74-76` — `set.deletes` has no production reader.** `deleteManifests()` is called only from `snapshot_manifest_cache_internal_test.go:99,105,451`. `dataManifests()` has exactly one caller (`incremental_append_scan.go:176`), and the dominant local path re-derives its own split from the *post-filter* list (`scanner.go:1530 splitManifestList`), so it structurally cannot use the cached partition. Both partitions are allocated at `cap = len(manifests)` and never clipped:

```
n=1000   all(len=1000 cap=1023)  data(1000)  deletes(0/1000)   waste=16,368 B
n=10000  all(len=10000 cap=10240) data(10000) deletes(0/10000) waste=163,840 B
```

160 KB retained per wide entry, ~10.5 MB across a full LRU, holding zero information. Drop `deletes`/`deleteManifests()` until something reads them, and `slices.Clip` what's kept. The description's "Keep data and delete manifest groups ready for planning paths that need one type" is true only for the data group.

**2. `snapshot_manifest_cache.go:35` — `snapshotManifestCacheSize = 64` bounds entries, not bytes.** Measured worst-case live retention per `Table` handle at 10,000 manifests/snapshot: **147.2 MiB** (≈2.3 MiB per 1,000 manifests per snapshot). This does satisfy my "make the worst case predictable" ask — the predictable worst case is just large. Steady-state working set is 1 entry; time travel, a handful. Either make the bound size-aware, or lower the default and say in the doc what the per-entry cost scales with.

## What checks out

- **Key is complete.** `{id:1, list:"v1.avro"}` vs `{id:1, list:"v2.avro"}` return the right thing; `ManifestLocations: nil` vs `[]string{}` produce different keys (that's what `hasEmbeddedSources` buys); a `ManifestList`-bearing snapshot never collides with an embedded-locations snapshot naming the same path. When both are set the key ignores `ManifestLocations` — correct, because `Snapshot.Manifests` dispatches the same way.
- **Scope is per-`Table` handle**, never package-global or per-catalog. `New` (`table.go:1416`) is the only constructor and every catalog goes through it, so two `LoadTable` calls never share. No cross-tenant hazard.
- **Concurrency:** 64 goroutines × 200 iterations, mixed same-key/200-distinct-key, under `-race`, mutating every handed-out slice → 0 wrong-data, 0 errors.
- **`Refresh` invalidates, functionally verified** (not just pointer-compared): `openCount[manifestList]` goes 1 → 1 (hit) → **2** after `Refresh` + rescan.
- **Failed loads are not cached** — a loader failing twice then succeeding invokes the loader 3 times.
- **Benchmark table reproduces on the final commit.** At `manifests=1/scans=10` vs the reverted base: allocs **37.08k → 23.57k** (you said 37,048 → 23,519), bytes **4.530 → 2.655 MiB** (you said 4.74 → 2.78 MB). `manifest-list-opens` 10 → 1 confirmed. Your "with 10,000 manifests total planning time stays close" caveat is honest and if anything understated — I see no significant change at 1,000 or 10,000 (time p ≥ 0.39, bytes/allocs within ±0.2%).
- `-race -count=5` clean; `golangci-lint` 0 issues; CI 15/15.

## Minor

- `scanner.go:1506-1508` — orphaned comment; it documented the `fs, err := scan.ioF(ctx)` this PR deleted and now sits above a blank line. Related and worth a word: the manifest-list `ioF` call now happens only on a cache miss, so a credential-renewing factory sees one fewer invocation per scan.
- `snapshot_manifest_cache.go:200-202` — `finish`'s early return exits without `close(entry.ready)`; any waiter blocks until its own ctx fires. Currently unreachable (I forced it to confirm), but close `ready` on every path so the invariant is structural rather than incidental.
- `snapshot_manifest_cache.go:47` — `slices.Clone(manifests)` is redundant; the single production caller passes a freshly decoded, exclusively owned slice. It also over-allocates (`cap 10240` for `len 10000`).
- `table.go:239-244` — `manifestCache` joins the `Table` fields `Refresh` mutates unsynchronized. Pre-existing race class, not introduced here, and harmless: the key includes the manifest-list location and Iceberg manifest lists are write-once, so a "stale" hit is byte-identical. That does mean the `Refresh` reset is a *memory* measure, not a staleness measure — the doc and the description's "so a new metadata snapshot never reuses stale descriptors" oversell it.
- `snapshot_manifest_cache_internal_test.go:376`,`:403` — the new panic test's `release` channel has no unconditional `t.Cleanup` close; a `t.Fatal` at `:403` strands the producer goroutine. Same hygiene class @laskoviymishka flagged on the older test.
- `snapshot_manifest_cache_internal_test.go:139-153` — `TestSnapshotManifestCacheSharesInFlightRead` can pass without single-flight (late arrivals hit the LRU and `openCount` stays 1). Not wrong, just weaker than its name.
- `transaction.go:3213` — a fresh 64-slot LRU per `Transaction.Scan()` means repeated `txn.Scan()` gets no reuse and each call allocates a pre-sized 64-bucket map.
- `>64`-snapshot sweeps pay an unconditional tax (measured +0.50% allocs, +0.31% B/op, p=0.002 at 128 snapshots) with zero hit rate — my option 2 would remove it. I did check whether such a sweep evicts the hot current-snapshot entry: it does not, because `metadata.Snapshots()` visits the current snapshot last, so it ends as MRU.

## Prior items

**Mine:** unbounded growth → **Fixed** (64-entry LRU; the eviction test is genuine, not a tautology — it re-gets `[0]` to touch it and shows the true LRU was evicted). Option 2 (exclude bulk historical paths) → **not adopted**, residual cost measured above. Option 3 (scope to an operation) → **adopted for transactions**. Key correctness, pre-filter caching, `Refresh` invalidation, defensive-copy accessors → **all re-verified independently.**

**@laskoviymishka:** producer zombie on panic → **Fixed** (defer `finish`, panic branch closes with a synthetic error). Error case → **Fixed and structural.** ctx threaded into the read → **Partially fixed** — it reaches ctx-bound backends, but a non-ctx-bound IO (`LocalFS`, `MemFS`, custom) still strands everyone, and this is what created the blocking inversion above. Retry-after-stuck-producer test → **Fixed.** Negative content-type partition → **Fixed** (positive `switch` + a test feeding `ManifestContent(2)`). Transaction cache isolation → **Fixed.** Test-hygiene leak → **Fixed** (new instance introduced at `:376`). NUL-join comment → **Fixed.**

Worth confirming one posture: the retry-re-blocks behaviour is now *codified as intended* at `snapshot_manifest_cache_internal_test.go:358-362` (`t.Fatalf` if the retry returns before the producer finishes). Given the blocking item, you may want to revisit whether that's the semantic you want.

## Description

- **Unstated benchmark methodology.** The "without-cache" arm is `tbl.manifestCache = nil` (`snapshot_manifest_cache_bench_test.go:53`), which still runs `newSnapshotManifestSet` + `allManifests()`. That isn't the pre-PR baseline, so the table understates the new machinery's cost. Also `newRowLimitPlanningBenchmarkTable` is built outside `b.ResetTimer()`, so the cache persists across `b.N` — "10 repeated scans per iteration" is effectively warm-cache.
- **"Apply the cache to … table inspection" is overstated** — only `inspect_files.go:136,191`. Still uncached: `inspect.go:459`, `inspect_all_manifests.go:51`, `inspect_partitions.go:153`, `inspect_position_deletes.go:87`, `orphan_cleanup.go:518`, `pos_delete_collect.go:51`, `rewrite_manifests.go:162`, `compaction/eq_delete_collect.go:85`.
- **"Reset the cache on Refresh so a new metadata snapshot never reuses stale descriptors"** — the reset is real, but the stated reason isn't the operative one (see Minor); it bounds memory.
- Not mentioned: `sharedSnapshotManifestFSF`, a new `sync.Once` IO-resolution helper that also memoizes the *failure* of the first `fsF(ctx)` for that call's lifetime.

---

table/snapshot_manifest_cache.go
```diff
@@ -0,0 +1,286 @@
...
+			panic(recovered)
+		}
+
+		c.finish(key, entry, value, err)
+	}()
+
+	// The read is shared with callers whose contexts may outlive this one. Do
+	// not let the producer's cancellation abort a healthy waiter's read.
+	value, err = load(context.WithoutCancel(ctx))
+
```
zeroshade, line 233: **major** — context.WithoutCancel makes the IO-factory resolution uncancellable, so scans can hang past their deadline

The producer runs load(context.WithoutCancel(ctx)), and readSnapshotManifestSet:254 resolves fsF(ctx) inside that detached context. FSysF is not a cheap accessor: io.LoadFSFunc -> LoadFS -> the S3 backend calls config.LoadDefaultConfig(ctx) (io/gocloud/s3/s3.go:114), which performs ctx-bound IMDS/STS network I/O for credential resolution. Before this PR, planFilesLocal called scan.ioF(ctx) with the caller's context, so a deadline bounded that call. Now nothing can abort it: the caller's cancellation is stripped, no timeout is imposed, and the goroutine plus its IO are pinned indefinitely. Waiters can bail on their own ctx, but the producer cannot. Suggested fix: resolve the IO with the caller's context before entering the single-flight, and keep WithoutCancel only around the shared manifest-list read (the part whose result is memoized). No test pins the lost guarantee.

```text
Probe (deleted after use) installing tbl.fsF = func(ctx){<-ctx.Done(); return nil, ctx.Err()} and calling tbl.Scan().PlanFiles(ctx) with a 200ms timeout. At merge base 52593109: 'PROBE1 PlanFiles returned within deadline: context deadline exceeded' --- PASS (0.20s). At head 2feda53: 'PROBE1 FAIL: PlanFiles hung past its context deadline; the IO factory call is uncancellable' --- FAIL (3.01s).
```

</details>

table/snapshot_manifest_cache.go
```diff
@@ -0,0 +1,286 @@
...
+	entry.readyOnce.Do(func() { close(entry.ready) })
+}
+
+func readSnapshotManifestSet(
+	ctx context.Context,
+	snapshot Snapshot,
+	fsF FSysF,
+) (snapshotManifestSet, error) {
+	if err := ctx.Err(); err != nil {
+		return snapshotManifestSet{}, err
```
zeroshade, line 270: **minor** — Unreachable ctx.Err() guard in readSnapshotManifestSet, pinned only by a test of an impossible path

Every production call reaches readSnapshotManifestSet through get(), which passes context.WithoutCancel(ctx); that context's Err() is always nil, and get() already checked ctx.Err() at :158. The guard can never fire outside tests. TestReadSnapshotManifestSetKeepsCompletedReadAfterCancellation exercises this by calling readSnapshotManifestSet directly with a raw cancellable ctx, which no caller does. Either drop the guard and the test, or make it meaningful by not detaching the factory call (see the major finding).

table/snapshot_manifest_cache.go
```diff
@@ -0,0 +1,286 @@
...
+	err error,
+) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if current, ok := c.entries[key]; ok && current == entry {
+		delete(c.entries, key)
+		if err == nil {
+			if c.complete.Contains(key) {
+				c.complete.Remove(key)
```
zeroshade: **minor** — Unreachable defensive Contains/Remove before Add in finish

A producer is only created on a complete-cache miss (get:164-168), and entries[key] serializes producers for a key, so no second finish for the same key can run concurrently. complete therefore cannot contain key when finish reaches :222. The branch is dead; if it is intended as insurance against a future refactor, that intent should be a comment, otherwise it should be removed along with the double-accounting it guards against.

table/snapshot_manifest_cache.go
```diff
@@ -0,0 +1,286 @@
...
+func (s snapshotManifestSet) allManifests() []iceberg.ManifestFile {
+	return slices.Clone(s.all)
+}
+
+func (s snapshotManifestSet) dataManifests() []iceberg.ManifestFile {
+	return slices.Clone(s.data)
+}
+
+func snapshotManifestSetSize(set snapshotManifestSet) int {
+	return len(set.all) + len(set.data)
```
zeroshade: **minor** — snapshotManifestSetSize double-counts data manifests, halving the documented descriptor bound

size = len(all) + len(data), but data holds the same descriptors already counted in all. For an all-data snapshot (the common case) every manifest counts twice, so snapshotManifestCacheManifestLimit = 32*1024 admits only ~16384 distinct manifests, not the '32 * 1024 manifest descriptors' the comment at :37-40 claims. Either count len(all) only, or restate the constant's meaning.

table/table.go
```diff
@@ -1398,6 +1413,7 @@ func New(ident Identifier, meta Metadata, metadataLocation string, fsF FSysF, ca
 		metadata:         meta,
 		metadataLocation: metadataLocation,
 		fsF:              fsF,
+		manifestCache:    newSnapshotManifestCache(),
 		cat:              cat,
```
zeroshade, line 1416: **minor** — Cross-scan manifest caching is always on with no way to disable or bound it by bytes

New() unconditionally installs a cache, so every Table now retains up to 64 decoded manifest lists (bounded only by a descriptor count, not bytes) for its lifetime, with no table property or Option to turn it off. That is a meaningful change in memory posture for long-lived Table handles over wide snapshot histories, and it diverges from Java, where manifest caching is opt-in via io.manifest.cache-enabled (default false) with byte-based bounds (io.manifest.cache.max-total-bytes, io.manifest.cache.max-content-length). Worth at least a property to disable it.

table/table.go
```diff
@@ -376,6 +388,7 @@ func (t Table) AllManifests(ctx context.Context) iter.Seq2[iceberg.ManifestFile,
 	ch := make(chan list, allManifestsWorkerCount(n))
 	workers := allManifestsWorkerCount(n)
 	g, groupCtx := errgroup.WithContext(workCtx)
+	manifestFS := sharedSnapshotManifestFSF(t.fsF)
```
zeroshade, line 391: **nit** — AllManifests no longer surfaces an IO-factory error for a table with zero snapshots

Previously fs, err := t.fsF(ctx) ran eagerly and a failure was yielded as an error even when there were no snapshots. sharedSnapshotManifestFSF is lazy and is never invoked when the jobs channel yields nothing, so an empty table with a broken IO factory now returns an empty sequence instead of an error. Probably the better behavior, but it is an undocumented, untested change.

table/inspect_files.go
```diff
@@ -131,14 +131,15 @@ func (i InspectTable) manifestEntryReader(
 		return nil, errors.New("table file IO is not configured")
 	}
 
-	fs, err := i.tbl.fsF(ctx)
+	manifestSet, err := i.tbl.manifestSetWithFSF(ctx, *snapshot, sharedSnapshotManifestFSF(i.tbl.fsF))
 	if err != nil {
```
zeroshade, line 136: **nit** — sharedSnapshotManifestFSF's sync.Once buys nothing on single-use paths, and inspect.go hand-rolls a fourth variant

At inspect_files.go:134 the shared wrapper is constructed inline for a single manifestSetWithFSF call, so the memoization can never be exercised. Meanwhile inspect.go:455-461 wraps i.tbl.fsF in a bespoke closure that only adds a 'get file IO: %w' prefix instead of using the same helper. Four call sites, three different shapes, for one job.

zeroshade (commented):
Cache key, isolation and concurrency are sound and every prior finding is fixed and mutation-pinned, but context.WithoutCancel at snapshot_manifest_cache.go:205 makes the IO-factory resolution uncancellable, so PlanFiles now hangs past a caller's deadline where the merge base returned context deadline exceeded.

Re-review verification: 5 of 6 prior findings confirmed fixed at `2feda53` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *not fixed* — A stalled manifest-list read cannot be aborted; retrying waiters with fresh ctx re-block against it

```text
go build ./... (OK); go vet ./table/... (OK); go test ./table/... -count=1 -race -timeout 1800s -> ok table 19.913s, table/compaction 9.970s, table/dv 1.813s, table/internal 3.298s, table/substrait 3.046s. Mutation runs M1-M10 (each restored with git checkout --). Throwaway probes in table/pr1970_probe_test.go: PROBE1 deadline (FAIL at head, PASS at merge base 52593109), PROBE2 concurrent shared descriptors under -race (PASS), PROBE3 stuck-producer coupling (informational; not a regression -- pre-PR the later caller would block forever since Snapshot.Manifests takes no ctx), PROBE4 cached-vs-uncached differential over PlanFiles + AllManifests (byte-identical). Probe file and /tmp/pr1970base removed; git status --porcelain empty.
```

</details>

---

>

table/snapshot_manifest_cache.go
```diff
@@ -0,0 +1,310 @@
...
+	entry.readyOnce.Do(func() { close(entry.ready) })
+}
+
+func readSnapshotManifestSet(
+	ctx context.Context,
+	snapshot Snapshot,
+	fsF FSysF,
+) (snapshotManifestSet, error) {
+	if err := ctx.Err(); err != nil {
+		return snapshotManifestSet{}, err
```
zeroshade, line 270: **nit** — ctx.Err() early-out in readSnapshotManifestSet is covered by no test

Deleting the guard at lines 269-271 leaves the full table package green. It is no longer dead code (a genuine window exists between get's ctx.Err() check at :169 and the load call at :232), so it is defensible as a fast-path early-out, but nothing pins it. Either drop it and rely on the factory/IO honoring ctx, or add a direct unit test on readSnapshotManifestSet with an already-canceled context.

table/snapshot_manifest_cache_internal_test.go
```diff
@@ -0,0 +1,921 @@
...
+		}, nil)
+	for manifest, err := range tbl.AllManifests(t.Context()) {
+		t.Fatalf("empty table yielded %v, %v", manifest, err)
+	}
+}
+
+type contextBoundSnapshotManifestIO struct {
+	iceio.IO
+	ctx         context.Context
+	started     chan struct{}
```
zeroshade, line 721: **nit** — contextBoundSnapshotManifestIO.Open unconditionally closes a shared started channel

Open does close(fs.started) on a channel captured from the enclosing test scope, while the factory hands out a fresh IO instance per call. This is safe today only because at most one Open reaches the blocking branch per test (guarded by blockedPath in the early-stop test, and by producerContextKey in the retry test). A second blocked Open would panic with 'close of closed channel' rather than failing cleanly. A sync.Once around the close would make the helper robust to future test additions.

zeroshade (approved):
Every prior finding is genuinely fixed and pinned by tests that go red under mutation; the WithoutCancel removal is correctly paired with a race-free entry.canceled retry mechanism, leaving only two nits.

Re-review verification: 6 of 7 prior findings confirmed fixed at `4e14f94` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *partially fixed* — minor: unreachable ctx.Err() guard in readSnapshotManifestSet pinned only by a test of an impossible path

```text
go build ./... (clean); go vet ./table/ (clean); go test ./table/ -race -timeout=600s -count=1 => ok 20.916s (full package, unmutated); go test ./table/ -race -count=5 -run 'SnapshotManifest|PlanFilesRetries|PlanFilesSnapshot|AllManifests|InspectFiles|ManifestListCache' => ok 5.469s (no flakiness). Six mutation experiments: (1) restore context.WithoutCancel => 5 tests RED; (2) force entry.canceled=false => 3 tests RED; (3) restore len(all)+len(data) => TestSnapshotManifestCacheCountsDataDescriptorsOnce RED; (4) ignore the cache property => 3 tests RED; (5) revert inspect_files IO reuse => TestInspectFilesReuseManifestListIO RED; (6) delete readSnapshotManifestSet ctx.Err() guard => full package GREEN (finding kept open as a nit). Worktree restored after every mutation; 'git status --porcelain' empty at 4e14f94.
```

</details>

---

>

laskoviymishka (approved):
LGTM

PR #1974, opened 2026-08-29
perf(table): stream data manifest entries into tasks

fallintoplace:
## Summary

- **Streams data manifest entries directly into scan tasks** instead of keeping a global data-entry slice.
- Reads delete manifests first and builds positional, equality, and deletion-vector indexes.
- Reads data manifests concurrently while keeping task order by manifest order.
- Uses one shared task buffer when manifest counts are known. Unknown counts use a safe fallback.
- Keeps the existing collector for incremental append scans.

## Tests

- `go test ./...`
- `go test -race ./table`
- `go vet ./table`

## Benchmark

Command:
`go test ./table -run '^$' -bench '^BenchmarkPlanDataManifestTasks/manifests=8/entries=10000/(buffered|streamed)$' -benchmem -benchtime=1s`

Apple M1 Pro, 8 manifests x 10,000 entries:

| Path | ns/op | B/op | allocs/op |
| --- | ---: | ---: | ---: |
| buffered | 28,409,791 | 86,939,909 | 905,562 |
| streamed | 23,727,901 | 80,683,642 | 905,561 |

The streamed path is about **16% faster** and uses about **7% fewer allocated bytes** on this workload.

### DIFF
--- a/table/scan_metrics.go
+++ b/table/scan_metrics.go
@@ -32,16 +32,14 @@ import (
 // scanMetricsAccumulator gathers scan-planning counts. Every field is written
 // from a single goroutine — the manifest counts in
 // fetchPartitionSpecFilteredManifestsWithSchema and the result/delete counts in
-// planFilesLocal after collectManifestEntriesWithSchema's concurrency barrier —
-// so plain integers are race-free; no field is touched from the concurrent
-// manifest workers.
+// planFilesLocal after its concurrent manifest/task barrier — so plain
+// integers are race-free; no field is touched from the concurrent workers.
 //
-// The scanned/skipped manifest counts measure manifest selection before opening
-// manifests, including partition-spec pruning and known-empty manifests. The
-// per-entry skipped-data-files / skipped-delete-files counts (which would be
-// incremented inside the concurrent openManifest loop, and would use atomics)
-// and indexed-delete-files are left for a follow-up and omitted rather than
-// reported as zero.
+// The scanned/skipped manifest counts measure partition-spec pruning (the
+// filter applied while listing manifests). The per-entry skipped-data-files /
+// skipped-delete-files counts (which would be incremented inside the concurrent
+// manifest-entry loop, and would use atomics) and indexed-delete-files are left
+// for a follow-up and omitted rather than reported as zero.
 type scanMetricsAccumulator struct {
 	totalDataManifests     int64
 	totalDeleteManifests   int64
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -1098,7 +1098,7 @@ func (scan *Scan) collectManifestEntriesWithSchema(
 	partitionFilters *keyDefaultMapErr[int, iceberg.BooleanExpression],
 ) (*manifestEntries, error) {
 	return scan.collectManifestEntriesWithSchemaMinSequenceNum(
-		ctx, manifestList, schema, scan.partitionFiltersForSchema(schema), minSequenceNum(manifestList))
+		ctx, manifestList, schema, partitionFilters, minSequenceNum(manifestList))
 }
 
 func (scan *Scan) collectManifestEntriesWithSchemaMinSequenceNum(
@@ -1245,6 +1245,7 @@ func (scan *Scan) planDataManifestTasks(
 	}
 
 	concurrencyLimit := min(scan.concurrency, len(manifestList))
+	manifestIO := newManifestIOBatch(scan.ioF, concurrencyLimit)
 	g, gctx := errgroup.WithContext(ctx)
 	g.SetLimit(concurrencyLimit)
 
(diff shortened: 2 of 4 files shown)


zeroshade (approved):
Streaming planning preserves one task per matching live data entry, deterministic manifest order, and complete delete attachment. Approving, with two minor follow-ups.

### What I verified

The failure modes streaming most plausibly introduces, and why each is handled:

- **No silently-truncated result.** `streamManifest` visits each iterator pair once, applies the same partition and metrics predicates as the previous materialised loop, and propagates iterator, predicate, and callback errors. `g.Wait()` is checked and `planFilesLocal` returns `nil, err`, so a partial batch can never come back looking like a complete task list. This was my main worry and it's the right shape.
- **Every task gets its complete delete set.** `planFilesLocal` (`table/scanner.go:1374-1419`) reads delete manifests and finishes `buildPositionalDeleteIndex`, `buildDVIndex`, and `buildEqualityDeleteIndex` **before** calling `planDataManifestTasks`; each callback then runs `fileScanTaskForDataEntry`. So no task is emitted before its position deletes, equality deletes, and deletion vectors are resolved — a task emitted early would have returned deleted rows.
- **No reader leak.** `ManifestFile.Entries` and `iterManifest` defer-close the file and manifest reader on EOF, on read error, and on callback early-stop. The new planner consumes the helper synchronously, so there's no externally abandonable iterator here.
- **Deterministic ordering.** Per-manifest batches are flattened in `manifestList` order after the errgroup barrier, so task order still matches the previous materialised behaviour — including the empty and inaccurate-count fallback paths.
- **Race safety.** Known counts write into disjoint regions of one preallocated buffer, batches are written and read only across the errgroup barrier, and evaluator caches are locked.

### Minor — cancellation responsiveness (`table/scanner.go:1154`, helper at `:359-363`)

`planDataManifestTasks` passes the errgroup context to `ioF`, but `streamManifest` itself takes no context and doesn't check between entries. With context-oblivious IO or expensive filters, a cancellation after `Open` isn't observed until EOF and planning keeps going.

Not a correctness issue, and worth saying why: `File` has no context-aware `ReadAt` anywhere in this codebase, so IO-layer propagation is the established posture rather than a gap unique to this PR. Passing `gctx` into the helper and checking `ctx.Err()` between entries would just make it more responsive.

### Minor — planner-level edge-case tests (`table/scanner_internal_test.go:638`, `:1952`)

New tests cover manifest order and inaccurate counts. The lower-level `manifest_test.go` already verifies reader close on early break and mid-stream read errors, so coverage isn't absent overall — but the new planner path has no targeted test injecting a mid-stream read/filter/task-construction error and asserting `(nil, error)`, nor one for cancellation or an empty data manifest.

Benchmark is committed (`BenchmarkPlanDataManifestTasks`, buffered vs streamed); body reports M1 Pro 8×10,000 at 23.73 ms vs 28.41 ms (~16% faster) and 80.68 MB vs 86.94 MB (~7% fewer bytes).

CI green, all 4 commits signed off.

---

>

zeroshade:
My approval above stands, but this went conflicted before I could merge it.

Same cause as #1965: the duplicated `schema.go` `MarshalJSON` lazy-cache hunk shared across this stack has now landed on `main`, so the identical hunk here conflicts. A rebase onto current `main` should drop it and leave just the streaming-planner change.

Once rebased and green I'll merge — no need to re-request review. The two minor follow-ups from my review (context check between entries, planner-level edge-case tests) are fine to pick up here or separately.

zeroshade:
Update: I rebased #1965 and #1976 onto current `main` and merged both. I stopped short on this one, because its rebase is not mechanical and I did not want to guess at your design.

The `schema.go` hunk drops out cleanly (already upstream). The real conflict is that `PlanFiles` has changed underneath this PR in three ways that interact with the streaming restructure:

1. **#1939 splits one entry into N tasks.** `splitParquetScanTask(task, splitTargetSize)` can emit multiple tasks per manifest entry. `planDataManifestTasks` sizes its `directBuffer` via `manifestTaskCapacity` at exactly one task per live entry, so that invariant no longer holds.
2. **#1971 computes per-task residuals.** Main keeps `residualEvaluators` as a plain `map[int]*partitionResidualEvaluator` populated per entry, which is fine sequentially but would race inside your `g.Go` workers.
3. **Result metrics accumulate per task.** `acc.resultDataFiles++` and `acc.addResultDeleteMetrics(task)` are unsynchronized for the same reason.

So reapplying this needs decisions that are yours to make: whether residuals are computed inside each worker (needing a concurrency-safe per-spec cache) or in a sequential pass after flattening; whether splitting happens inside the worker (giving up the exact-capacity buffer) or afterwards (costing a second pass, which partly offsets what the PR is optimizing); and how metrics accumulate safely.

My review and approval above still stand on the design as reviewed — this is purely about how it lands on top of what has since merged. Also note `openManifest` no longer exists, so the `scan_metrics.go` comment referencing it needs updating to match `streamManifest`.

Related: #1976 is now merged and reworks `collectManifestEntriesWithSchema` into per-manifest result slots, which your delete-manifest path still uses. Worth rebasing on top of that rather than around it.

Happy to review again once rebased. If you'd rather I take a pass at the merge and you review my choices, say so and I will.

laskoviymishka (approved):

PR #1975, opened 2026-08-29
perf(table): retain only required delete stats

fallintoplace:
## Summary

- **Keep only the stats needed by each delete index.**
- Positional deletes retain `file_path` metrics for partition-scoped pruning.
- Equality deletes retain metrics for `EqualityFieldIDs()`.
- Deletion vectors retain no stats, but keep the metadata needed to read the Puffin range.
- Store only the compact delete file and sequence number in the indexes.
- Release the original delete manifest entries before data task planning continues.
- Keep the existing fallback for external or malformed `DataFile` implementations.

## Benchmark

Command:
`go test ./table -run "^$" -bench "^BenchmarkDeleteIndexRetainsOnlyRequiredStats$" -benchmem -benchtime=1s`

Apple M1 Pro, 100 delete files, 512 stat fields:

| index | ns/op | B/op | allocs/op | source stats/file | retained stats/file |
| --- | ---: | ---: | ---: | ---: | ---: |
| positional | 227,435 | 346,987 | 4,211 | 2,565 | 5 |
| equality | 185,194 | 238,189 | 3,731 | 2,560 | 10 |

The source and retained columns count entries across value, null, NaN, lower-bound, and upper-bound maps.

## Tests

- `go test ./...`
- `go test -race ./table -run "TestCompactDeleteFileForIndex|TestDeleteIndexesRetainOnlyRequiredStats|Test(PositionalDeleteIndex|EqualityDeleteIndex|BuildDVIndex|MatchDV|PlanFiles)" -count=1`
- `go vet ./table`

### DIFF
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -299,7 +299,7 @@ func readAllDeletionVectors(ctx context.Context, fs iceio.IO, tasks []FileScanTa
 	for _, t := range tasks {
 		for _, d := range t.DeletionVectorFiles {
 			_, _, ref, contentOffset, contentSize := iceinternal.BorrowedDataFilePointers(d)
-			if ref == nil {
+			if ref == nil || *ref == "" {
 				return nil, fmt.Errorf("deletion vector %s missing referenced_data_file", d.FilePath())
 			}
 			if contentOffset == nil || contentSize == nil {
--- a/table/arrow_scanner.go
+++ b/table/arrow_scanner.go
@@ -402,9 +402,9 @@ func readAllDeletionVectors(ctx context.Context, fs iceio.IO, tasks []FileScanTa
 }
 
 // sameDVBlob reports whether two DV manifest entries point at the same puffin
-// blob — identical puffin file path and content offset. Different of either
-// means two distinct DVs for the same data file, the over-deletion case
-// readAllDeletionVectors rejects.
+// blob — identical puffin file path, content offset, and content size.
+// Different metadata means two distinct DVs for the same data file, the
+// over-deletion case readAllDeletionVectors rejects.
 //
 // Java's DeleteFileIndex is stricter: any second DV for the same data file is
 // rejected with ValidationException("Can't index multiple DVs for %s"), even
@@ -417,10 +417,10 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
 		return false
 	}
 
-	_, _, _, aOffset, _ := iceinternal.BorrowedDataFilePointers(a)
-	_, _, _, bOffset, _ := iceinternal.BorrowedDataFilePointers(b)
+	_, _, _, aOffset, aSize := iceinternal.BorrowedDataFilePointers(a)
+	_, _, _, bOffset, bSize := iceinternal.BorrowedDataFilePointers(b)
 
-	return *aOffset == *bOffset
+	return *aOffset == *bOffset && *aSize == *bSize
 }
 
 func filePathValueType(dt arrow.DataType) arrow.DataType {
--- a/table/compaction/analyze_test.go
+++ b/table/compaction/analyze_test.go
@@ -167,6 +167,56 @@ func TestAnalyze_AllOptimal(t *testing.T) {
 	assert.Empty(t, plan.Groups)
 }
 
+func TestPlanCompaction_BoundsScopedPositionalDeleteRatioFromScan(t *testing.T) {
+	tbl := newAnalyzeTestTable(t)
+	dataPath := tbl.Location() + "/data/data.parquet"
+
+	dataBuilder, err := iceberg.NewDataFileBuilder(
+		*iceberg.UnpartitionedSpec, iceberg.EntryContentData,
+		dataPath, iceberg.ParquetFile, nil, nil, nil, 100, 100)
+	require.NoError(t, err)
+	tx := tbl.NewTransaction()
+	require.NoError(t, tx.NewRowDelta(nil).AddRows(dataBuilder.Build()).Commit(t.Context()))
+	tbl, err = tx.Commit(t.Context())
+	require.NoError(t, err)
+
+	deleteBuilder, err := iceberg.NewDataFileBuilder(
+		*iceberg.UnpartitionedSpec, iceberg.EntryContentPosDeletes,
+		tbl.Location()+"/data/delete.parquet", iceberg.ParquetFile, nil, nil, nil, 40, 100)
+	require.NoError(t, err)
+	filePathField, ok := iceberg.PositionalDeleteSchema.FindFieldByName("file_path")
+	require.True(t, ok)
+	deleteFile := deleteBuilder.
+		LowerBoundValues(map[int][]byte{filePathField.ID: []byte(dataPath)}).
+		UpperBoundValues(map[int][]byte{filePathField.ID: []byte(dataPath)}).
+		Build()
+	tx = tbl.NewTransaction()
+	require.NoError(t, tx.NewRowDelta(nil).AddDeletes(deleteFile).Commit(t.Context()))
+	tbl, err = tx.Commit(t.Context())
+	require.NoError(t, err)
+
+	tasks, err := tbl.Scan().PlanFiles(t.Context())
+	require.NoError(t, err)
+	require.Len(t, tasks, 1)
+	require.Len(t, tasks[0].DeleteFiles, 1)
+	require.NotNil(t, tasks[0].DeleteFiles[0].ReferencedDataFile())
+	assert.Equal(t, dataPath, *tasks[0].DeleteFiles[0].ReferencedDataFile())
+
+	cfg := compaction.Config{
+		TargetFileSizeBytes:  100,
+		MinFileSizeBytes:     50,
+		MaxFileSizeBytes:     150,
+		MinInputFiles:        1,
+		DeleteFileThreshold:  2,
+		DeleteRatioThreshold: 0.3,
+		PackingLookback:      compaction.DefaultPackingLookback,
+	}
+	plan, err := cfg.PlanCompaction(tasks)
+	require.NoError(t, err)
+	require.Len(t, plan.Groups, 1)
+	assert.Len(t, plan.Groups[0].Tasks, 1)
+}
+
 func TestAnalyze_EmptyTable(t *testing.T) {
 	tbl := newAnalyzeTestTable(t)
 
--- a/table/delete_file_index.go
+++ b/table/delete_file_index.go
@@ -18,10 +18,6 @@
 package table
 
 import (
-	"cmp"
-	"fmt"
-	"slices"
-
 	"github.com/apache/iceberg-go"
 	iceberginternal "github.com/apache/iceberg-go/internal"
 )
@@ -56,7 +52,12 @@ func compactDeleteFileForIndexWithReference(
 	statFieldIDs []int,
 	referencedDataFile *string,
 ) iceberg.DataFile {
-	partitionSpec := syntheticPartitionSpec(file.SpecID(), partition)
+	// DataFile does not expose its original partition spec. The compact copy
+	// only needs the spec ID and the field-ID keyed partition values; codecs
+	// receive the actual spec separately when encoding the file. Keeping an
+	// ID-only spec avoids allocating synthetic field metadata that no caller
+	// can use for schema resolution.
+	partitionSpec := iceberg.NewPartitionSpecID(int(file.SpecID()))
 	builder, err := iceberg.NewDataFileBuilder(
 		partitionSpec,
 		file.ContentType(),
@@ -96,9 +97,7 @@ func compactDeleteFileForIndexWithReference(
 	// Read these small non-statistics fields directly. The borrowed collection
 	// helper also exposes column sizes and the built-in implementation lazily
 	// initializes every statistics map while preparing that view.
-	keyMetadata := file.KeyMetadata()
-	splitOffsets := file.SplitOffsets()
-	equalityFieldIDs := file.EqualityFieldIDs()
+	_, keyMetadata, splitOffsets, equalityFieldIDs := dataFileCollections(file)
 	if keyMetadata != nil {
 		builder.KeyMetadata(keyMetadata)
 	}
@@ -129,30 +128,6 @@ func compactDeleteFileForIndexWithReference(
 	}
 
 	return builder.Build()
-}
-
-// syntheticPartitionSpec gives the built-in DataFile implementation enough
-// field metadata to expose the copied partition map and to remain usable by
-// the DataFile codec. Partition values are already transformed values, so the
-// identity transforms here are only a local storage description. SourceIDs
-// are synthetic because DataFile exposes the partition values and spec ID,
-// but not the original partition spec; callers must not use this local spec
-// for schema resolution.
-func syntheticPartitionSpec(specID int32, partition map[int]any) iceberg.PartitionSpec {
-	fields := make([]iceberg.PartitionField, 0, len(partition))
-	for fieldID := range partition {
-		fields = append(fields, iceberg.PartitionField{
-			SourceIDs: []int{fieldID},
-			FieldID:   fieldID,
-			Name:      fmt.Sprintf("partition_%d", fieldID),
-			Transform: iceberg.IdentityTransform{},
-		})
-	}
-	slices.SortFunc(fields, func(a, b iceberg.PartitionField) int {
-		return cmp.Compare(a.FieldID, b.FieldID)
-	})
-
-	return iceberg.NewPartitionSpecID(int(specID), fields...)
 }
 
 func dataFileStatsForFields(
@@ -209,7 +184,7 @@ func selectByteStats(source map[int][]byte, fields map[int]struct{}) map[int][]b
 	selected := make(map[int][]byte, len(fields))
 	for fieldID := range fields {
 		if value, ok := source[fieldID]; ok {
-			selected[fieldID] = slices.Clone(value)
+			selected[fieldID] = value
 		}
 	}
 	if len(selected) == 0 {
--- a/table/delete_file_index.go
+++ b/table/delete_file_index.go
@@ -47,6 +47,15 @@ func compactDeleteFileForIndex(
 	partition map[int]any,
 	statFieldIDs []int,
 ) (iceberg.DataFile, error) {
+	return compactDeleteFileForIndexWithReference(file, partition, statFieldIDs, nil)
+}
+
+func compactDeleteFileForIndexWithReference(
+	file iceberg.DataFile,
+	partition map[int]any,
+	statFieldIDs []int,
+	referencedDataFile *string,
+) (iceberg.DataFile, error) {
 	partitionSpec := syntheticPartitionSpec(file.SpecID(), partition)
 	builder, err := iceberg.NewDataFileBuilder(
 		partitionSpec,
@@ -100,7 +109,7 @@ func compactDeleteFileForIndex(
 		builder.EqualityFieldIDs(equalityFieldIDs)
 	}
 
-	sortOrderID, firstRowID, referencedDataFile, contentOffset, contentSize := iceberginternal.BorrowedDataFilePointers(file)
+	sortOrderID, firstRowID, manifestReferencedDataFile, contentOffset, contentSize := iceberginternal.BorrowedDataFilePointers(file)
 	if sortOrderID != nil {
 		builder.SortOrderID(*sortOrderID)
 	}
@@ -109,6 +118,8 @@ func compactDeleteFileForIndex(
 	}
 	if referencedDataFile != nil {
 		builder.ReferencedDataFile(*referencedDataFile)
+	} else if manifestReferencedDataFile != nil {
+		builder.ReferencedDataFile(*manifestReferencedDataFile)
 	}
 	if contentOffset != nil {
 		builder.ContentOffset(*contentOffset)
--- a/table/delete_file_index.go
+++ b/table/delete_file_index.go
@@ -100,8 +100,7 @@ func compactDeleteFileForIndex(
 		builder.EqualityFieldIDs(equalityFieldIDs)
 	}
 
-	sortOrderID, firstRowID, referencedDataFile, contentOffset, contentSize :=
-		iceberginternal.BorrowedDataFilePointers(file)
+	sortOrderID, firstRowID, referencedDataFile, contentOffset, contentSize := iceberginternal.BorrowedDataFilePointers(file)
 	if sortOrderID != nil {
 		builder.SortOrderID(*sortOrderID)
 	}
--- a/table/delete_file_index.go
+++ b/table/delete_file_index.go
@@ -46,7 +46,7 @@ func compactDeleteFileForIndex(
 	file iceberg.DataFile,
 	partition map[int]any,
 	statFieldIDs []int,
-) (iceberg.DataFile, error) {
+) iceberg.DataFile {
 	return compactDeleteFileForIndexWithReference(file, partition, statFieldIDs, nil)
 }
 
@@ -55,7 +55,7 @@ func compactDeleteFileForIndexWithReference(
 	partition map[int]any,
 	statFieldIDs []int,
 	referencedDataFile *string,
-) (iceberg.DataFile, error) {
+) iceberg.DataFile {
 	partitionSpec := syntheticPartitionSpec(file.SpecID(), partition)
 	builder, err := iceberg.NewDataFileBuilder(
 		partitionSpec,
@@ -73,7 +73,7 @@ func compactDeleteFileForIndexWithReference(
 		// construction historically accepted such DataFile implementations;
 		// valid manifest files take the compact path above without changing the
 		// error behavior of the surrounding planner.
-		return file, nil
+		return file
 	}
 
 	valueCounts, nullCounts, nanCounts, lowerBounds, upperBounds := dataFileStatsForFields(file, statFieldIDs)
@@ -128,13 +128,16 @@ func compactDeleteFileForIndexWithReference(
 		builder.ContentSizeInBytes(*contentSize)
 	}
 
-	return builder.Build(), nil
+	return builder.Build()
 }
 
 // syntheticPartitionSpec gives the built-in DataFile implementation enough
 // field metadata to expose the copied partition map and to remain usable by
 // the DataFile codec. Partition values are already transformed values, so the
-// identity transforms here are only a local storage description.
+// identity transforms here are only a local storage description. SourceIDs
+// are synthetic because DataFile exposes the partition values and spec ID,
+// but not the original partition spec; callers must not use this local spec
+// for schema resolution.
 func syntheticPartitionSpec(specID int32, partition map[int]any) iceberg.PartitionSpec {
 	fields := make([]iceberg.PartitionField, 0, len(partition))
 	for fieldID := range partition {
--- a/table/delete_file_index.go
+++ b/table/delete_file_index.go
@@ -1,0 +1,207 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package table
+
+import (
+	"cmp"
+	"fmt"
+	"slices"
+
+	"github.com/apache/iceberg-go"
+	iceberginternal "github.com/apache/iceberg-go/internal"
+)
+
+// deleteFileIndexEntry is the part of a manifest entry used after delete
+// manifests have been read. Keeping the sequence number separate lets the
+// index release the full ManifestEntry, including its unused metadata maps.
+type deleteFileIndexEntry struct {
+	file        iceberg.DataFile
+	sequenceNum int64
+}
+
+// compactDeleteFileForIndex copies a delete file without retaining the wide
+// metadata maps that are not needed by scan planning or delete-file readers.
+// The returned file still carries all non-statistics metadata exposed by the
+// DataFile interface, plus only the selected statistics fields.
+//
+// Positional-delete indexes select file_pathFieldID because partition-scoped
+// position deletes use those bounds for candidate pruning. Equality-delete
+// indexes select their equality field IDs. Deletion vectors select no stats.
+func compactDeleteFileForIndex(
+	file iceberg.DataFile,
+	partition map[int]any,
+	statFieldIDs []int,
+) (iceberg.DataFile, error) {
+	partitionSpec := syntheticPartitionSpec(file.SpecID(), partition)
+	builder, err := iceberg.NewDataFileBuilder(
+		partitionSpec,
+		file.ContentType(),
+		file.FilePath(),
+		file.FileFormat(),
+		partition,
+		nil,
+		nil,
+		file.Count(),
+		file.FileSizeBytes(),
+	)
+	if err != nil {
+		// Keep the existing file for malformed or external metadata. Index
+		// construction historically accepted such DataFile implementations;
+		// valid manifest files take the compact path above without changing the
+		// error behavior of the surrounding planner.
+		return file, nil
+	}
+
+	valueCounts, nullCounts, nanCounts, lowerBounds, upperBounds := dataFileStatsForFields(file, statFieldIDs)
+	if valueCounts != nil {
+		builder.ValueCounts(valueCounts)
+	}
+	if nullCounts != nil {
+		builder.NullValueCounts(nullCounts)
+	}
+	if nanCounts != nil {
+		builder.NaNValueCounts(nanCounts)
+	}
+	if lowerBounds != nil {
+		builder.LowerBoundValues(lowerBounds)
+	}
+	if upperBounds != nil {
+		builder.UpperBoundValues(upperBounds)
+	}
+
+	// Read these small non-statistics fields directly. The borrowed collection
+	// helper also exposes column sizes and the built-in implementation lazily
+	// initializes every statistics map while preparing that view.
+	keyMetadata := file.KeyMetadata()
+	splitOffsets := file.SplitOffsets()
+	equalityFieldIDs := file.EqualityFieldIDs()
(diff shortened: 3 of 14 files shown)


zeroshade (changes requested):
The compact delete-file representation keeps what the delete readers need, but it drops the metadata compaction relies on to recognise a file-scoped positional delete. That plus a conflicted branch is what's holding this up.

### Major — inferred path scope is lost before compaction sees it (`table/positional_delete_index.go:41`)

The path-scoped branch calls `compactDeleteFileForIndex(deleteFile, partition, nil)`. For a v2 position delete with no `ReferencedDataFile`, the target is inferred from equal `file_path` lower/upper bounds — and this call drops those bounds from the compact copy.

Downstream, `table/compaction.isFileScoped` can no longer infer the reference, `referencedDataFilePath` returns `""`, and `fileScopedDeletedRows` therefore omits that delete's `Count`. Concretely: 40 deleted rows out of 100 on a right-sized file can fall below `DeleteFileThreshold` and miss the configured `DeleteRatioThreshold`, so compaction that should run is skipped.

**Impact is maintenance, not data loss** — scans still return the correct rows. What's lost is space reclamation, silently.

**Why green CI doesn't contradict this**, since it otherwise looks like it should: `TestPlanCompaction_BoundsScopedPositionalDeleteRatioForcesCompaction` does document the intended behaviour, but it lives in package `compaction_test`, constructs `testDataFile` and its tasks directly with bounds set, and calls `cfg.PlanCompaction` without ever passing through scanner index construction. It cannot observe what the scanner now hands to compaction, so the regression sits in a seam no test crosses.

Fix either way works: preserve the `file_path` bounds when the reference is inferred, or populate `ReferencedDataFile` explicitly on the compacted copy. I'd lean toward the explicit reference — it removes the dependency on bounds-inference entirely rather than keeping a second thing that must stay in sync.

### Minor — add a test across that seam (`table/delete_file_index_test.go:227-254`)

The current path-scoped test uses an explicit `ReferencedDataFile` and deliberately expects stats to be nil. Please add a bounds-only case that routes a positional delete through scanner index construction into `compaction.PlanCompaction` and asserts the ratio-triggered group, so this metadata contract is actually covered rather than assumed.

### Please rebase

`mergeStateStatus` is `DIRTY`. Twelve PRs from this stack merged in the last couple of hours, and the duplicated `schema.go` `MarshalJSON` hunk this branch carries has already landed on `main` — that's the likely conflict, and it should simply vanish on rebase.

### What I checked on the other consumers

This was the interaction I was most concerned about, and it's handled: equality deletes retain all five selected metrics for `EqualityFieldIDs` **including NaN counts**, which is what keeps the just-merged #1960 pruning working correctly. Equality readers use equality IDs and path; position readers use path and file metadata; DV readers use the preserved Puffin reference plus offset and size. All intact.

One note on the benchmark: it reports absolute figures (positional 227,435 ns/op / 346,987 B/op with 2,565 source stats reduced to 5; equality 185,194 ns/op / 238,189 B/op with 2,560 reduced to 10) but no before/after baseline. The retained-stat counts make the intent clear, so this isn't a finding — a baseline comparison would just make the win easier to see.

CI green, all 3 commits signed off.

---

>

table/scanner.go
```diff
@@ -872,14 +872,21 @@ func partitionsMatch(a, b map[int]any) bool {
 // buildDVIndex indexes deletion vectors by the data file path they reference.
 // The spec requires at most one DV per data file; a second entry for the same
 // path is rejected with an error.
-func buildDVIndex(dvEntries []iceberg.ManifestEntry) (map[string]iceberg.ManifestEntry, error) {
-	dvIndex := make(map[string]iceberg.ManifestEntry, len(dvEntries))
+func buildDVIndex(dvEntries []iceberg.ManifestEntry) (map[string]deleteFileIndexEntry, error) {
```
laskoviymishka, line 898: This is the same base-staleness issue from last round, and I was able to pin it down concretely this time. The branch sits on 8778910, one commit before the streaming scanner (#1974) that's now main's head, and both sides rewrote planFilesLocal.

buildDVIndex here returns map[string]deleteFileIndexEntry and matchDVToData takes it, but on current main planDataManifestTasks and fileScanTaskForDataEntry still thread dvIndex map[string]iceberg.ManifestEntry into matchDVToData. A merge-tree of this branch into origin/main conflicts in scanner.go, and once that's resolved the DV type won't line up with the streaming path.

I'd rebase onto current main, thread deleteFileIndexEntry through planDataManifestTasks and fileScanTaskForDataEntry, and land the entries.*Entries = nil release just before the planDataManifestTasks call. Worth building the merge result locally, since the branch checks stay green regardless.

table/delete_file_index.go
```diff
@@ -0,0 +1,217 @@
...
+
+// compactDeleteFileForIndex copies a delete file without retaining the wide
+// metadata maps that are not needed by scan planning or delete-file readers.
+// The returned file still carries all non-statistics metadata exposed by the
+// DataFile interface, plus only the selected statistics fields.
+//
+// Positional-delete indexes select file_pathFieldID because partition-scoped
+// position deletes use those bounds for candidate pruning. Equality-delete
+// indexes select their equality field IDs. Deletion vectors select no stats.
+func compactDeleteFileForIndex(
```
laskoviymishka, line 41: Still the always-nil error return, and the split into compactDeleteFileForIndex / compactDeleteFileForIndexWithReference doubled it: both declare (iceberg.DataFile, error), the NewDataFileBuilder failure path returns (file, nil) as the intended fallback, and builder.Build() has no error. So the dead `if err != nil` branch is now in four call sites (equality_delete_index.go:481, positional_delete_index.go:62 and its by-path sibling, scanner.go:883).

I'd drop the error from both signatures and simplify the callers, keeping the fallback comment so the "hold onto the original file for external/malformed metadata" intent stays documented. wdyt?

table/delete_file_index.go
```diff
@@ -0,0 +1,217 @@
...
+
+// syntheticPartitionSpec gives the built-in DataFile implementation enough
+// field metadata to expose the copied partition map and to remain usable by
+// the DataFile codec. Partition values are already transformed values, so the
+// identity transforms here are only a local storage description.
+func syntheticPartitionSpec(specID int32, partition map[int]any) iceberg.PartitionSpec {
+	fields := make([]iceberg.PartitionField, 0, len(partition))
+	for fieldID := range partition {
+		fields = append(fields, iceberg.PartitionField{
+			SourceIDs: []int{fieldID},
```
laskoviymishka: Unchanged from last round. partition is keyed by partition-field IDs, so SourceIDs: []int{fieldID} and FieldID: fieldID end up equal, which claims the source column is field 1000 rather than the real schema column.

No impact on the current lookups, since they take (specID, partition map) directly and never interpret this synthetic spec. But the compacted DataFile rides along in the FileScanTask, and now that 11d180e is deliberately making these compact copies survive into compaction planning, a consumer that resolves compacted.SpecID() to reason about partitioning would read a source-id that doesn't match the schema. I'd rather preserve the original spec than hand out one with wrong source IDs, or at least drop a comment saying the source IDs are synthetic.

table/compaction/analyze_test.go
```diff
@@ -167,6 +167,56 @@ func TestAnalyze_AllOptimal(t *testing.T) {
 	assert.Empty(t, plan.Groups)
 }
 
+func TestPlanCompaction_BoundsScopedPositionalDeleteRatioFromScan(t *testing.T) {
```
laskoviymishka, line 170: This new test is the right shape, it proves a bounds-scoped positional delete keeps its ReferencedDataFile all the way through scan and into compaction grouping, which is exactly the kind of "compaction dropped a needed field" regression I was worried about last round.

The gap that's left is the general one: nothing asserts a compact index returns the same delete-file matches as a full-stat index across the equality and DV paths too. This test covers the positional referenced-file case specifically. A small case that builds both a full-stat and a compact index from the same entries and asserts forDataFile returns the same set would lock the whole trimming contract, not just this slice. Not blocking, and I'd be fine taking it as a follow-up given the merge rework this needs anyway.

laskoviymishka (changes requested):
Catch-up pass on the three commits now on the branch (c504683, 31c21fe, 11d180e). I'm holding this rather than approving, and the one gating item is a merge/rebase problem, not the code itself.

The branch is based one commit short of the streaming scanner (#1974, now main's head). Both sides rewrote planFilesLocal, and this PR changes the DV index type, so a 3-way merge into current main conflicts in table/scanner.go and, once resolved, collides on the DV type (buildDVIndex now returns map[string]deleteFileIndexEntry while main's planDataManifestTasks / fileScanTaskForDataEntry still take map[string]iceberg.ManifestEntry). I confirmed the conflict with a merge-tree; the branch checks stay green because they don't cover the merge result. A rebase onto current main, threading the new type through the streaming path, is what unblocks this.

The substantive fix this round (11d180e) is good: preserving the inferred positional-delete scope so the compact copy still carries its referenced data file into compaction, with a regression test. The equality-pruning and schema concerns from my earlier look are resolved now that main moved forward. What's left in the inline notes is cleanup: an always-nil error return that's now spread across two functions, a synthetic partition spec that sets source IDs to partition-field IDs, and a gap in the trimming-equivalence tests. None of those block on their own.

Leaving this pending so it can be finalized.

zeroshade (changes requested):
The trimming contract itself is safe — I verified it end to end, including through a real Puffin DV — and all four of my prior items plus three of @laskoviymishka's four are fixed. But the parity test added in the head commit specifically to close the remaining gap cannot fail, and the DV path pays the full compaction cost for a measured zero benefit.

## Blocking — `table/delete_file_index_test.go:309`: `TestDeleteIndexesRetainMatchingParity` is a tautology

This test was added in `aa230a5c` in response to @laskoviymishka's *"nothing asserts a compact index returns the same delete-file matches as a full-stat index"*. It does not close that gap. I built four separate mutations that each break the trimming contract and ran only this test against each:

| mutation | breaks the contract? | this test |
|---|---|---|
| `dataFileStatsForFields` → all-nil (drop every stat) | yes — 6 other tests fail | **PASS** |
| `byPath` compaction drops the explicit `&path` ref (reverts `5fafb17f`) | yes — `TestPlanCompaction_BoundsScopedPositionalDeleteRatioFromScan` fails | **PASS** |
| drop `EqualityFieldIDs` from the compact copy | yes — 24 tests fail | **PASS** |
| drop `content_offset`/`content_size_in_bytes` | yes — 1 test fails | **PASS** |

Per subtest:
- **`:310` positional** — the delete file has *equal* `file_path` bounds, so it routes to `byPath`, and `appendPositionalDeletesFromSequence` (`positional_delete_index.go:165`) does no stat-based filtering at all. `fullIndex` is hand-built in `byPath` too. The assertion reduces to `["position-delete.parquet"] == ["position-delete.parquet"]`.
- **`:354` equality** — delete bounds `[10,20]`, data bounds `[0,30]`: **overlapping**. Both indexes return the file whether the bounds are present, absent, or garbage, because absent bounds are conservative.
- **`:399` deletion vector** — `matchDVToData` uses only path + sequence number. No stats involved by construction.

Fix: pick scenarios where the two indexes must **diverge** if trimming is wrong.
- Equality: use a **disjoint** delete range (delete `[100,200]`, data `[0,30]`) and assert the compact index still prunes it — that fails the moment equality-field bounds stop being retained.
- Positional: use **unequal** `file_path` bounds so the file lands in `byPartition`, and assert the compact index still prunes a data path outside `[lower, upper]`.
- DV: drop the subtest, or assert the Puffin range end-to-end (`ref`/`offset`/`size` non-nil after `PlanFiles`) — as written it asserts nothing.

Supporting datapoint: dropping the `[]int{filePathFieldID}` retention at `positional_delete_index.go:59` fails **only** `TestDeleteIndexesRetainOnlyRequiredStats`, a white-box assertion that the map keys exist. Nothing anywhere exercises `filePathMayMatch` on the retained bounds behaviourally. My probe does; it should live in the tree.

## Major — `table/scanner.go:906`: the DV compact copy is pure overhead

```go
indexedFile := compactDeleteFileForIndex(deleteFile, dataFilePartition(deleteFile), nil)
```

A Puffin DV manifest entry carries no column statistics, so there is nothing to trim. Measured on 100 DVs:

- retained heap: 53,464 B → 53,256 B — **208 bytes total, 2 B per DV, 0.39%**
- construction: 7.16 µs → 67.6 µs (**+845%**, p=0.008), 5.3 KiB → 54.6 KiB, **4 → 804 allocations**

Every byte the PR set out to reclaim here is already reclaimed by the `deleteFileIndexEntry` struct itself, which drops the `ManifestEntry`. Replacing the call with `file: deleteFile` keeps that benefit, removes 800 allocations per 100 DVs, and removes the synthetic partition spec from the DV path entirely. The `statFieldIDs == nil` argument is itself the signal that there's nothing to select.

The same reasoning partly applies to the `byPath` positional branch (`positional_delete_index.go:44`, also `statFieldIDs = nil`), but there the explicit `&path` materialisation is load-bearing for `compaction.isFileScoped`, so that copy has to stay.

## What checks out — the trimming is safe

**End-to-end probe**, v2 table, two data files, four delete files covering every path the trimmed index touches (partition-scoped positional with unequal bounds → `byPartition`; bounds-inferred file-scoped → `byPath`; overlapping equality; disjoint equality). Rows `[2 4 8]`, disjoint equality correctly pruned. **Identical output against a build with compaction physically disabled** — same rows, same delete→task attachment.

Separately drove a **real Puffin DV** (`dv.NewDVWriter`, positions {1,3}) through actual scan planning on a v3 table: `PlanFiles` attached it, the compact copy carried `ref/offset=4/size=44/count=2`, rows came back `[1 3]`. Identical on both builds. Worth knowing: the existing `TestDVScanEndToEnd` hand-builds `[]FileScanTask` and never crosses `buildDVIndex`, so it does not cover this.

**Retained-heap A/B, 100 delete files** — the memory win is real and holds at narrow widths too:

| index | stat fields | base | HEAD | delta |
|---|---:|---:|---:|---:|
| positional | 8 | 510,856 B | 151,640 B | −70.3% |
| positional | 512 | 22,070,840 B | 146,064 B | **−99.34%** |
| equality | 8 | 308,128 B | 96,960 B | −68.5% |
| equality | 512 | 21,616,160 B | 96,960 B | **−99.55%** |
| DV | 0 | 53,464 B | 53,256 B | −0.39% |

`selectByteStats`' `slices.Clone` genuinely severs the source backing arrays — the 22 MB really is released, not aliased.

**Consumer census.** I checked every reader of delete-file metadata downstream of planning against the retained set — `filePathMayMatch`, equality field metrics, `compaction.referencedDataFilePath`, `readAllDeleteFiles`, DV read + `sameDVBlob`, `equality_delete_reader.go:520`, `scan_metrics.go:96`, `codec/file_scan_task.go:71`. All satisfied. Notably the delete file's **`pos`-column bounds are read by nothing** (`arrow_scanner.go:705` reads the `pos` column *data*, not its statistics), so dropping them is safe.

**Every failure mode is conservative, not optimistic.** Forcing `dataFileStatsForFields` to return all-nil changed the probe from `eq=1` to `eq=2` — *more* delete files attached, never fewer — with byte-identical rows. `equalityDeleteCanContainData` continues on absent bounds; `filePathMayMatch` skips its checks on absent counts. **No silent-data-corruption path exists in this PR.**

`golangci-lint` 0 issues; `-race -count=5` clean; CI 15/15; `git merge-tree upstream/main aa230a5c` clean.

## Minor

- `delete_file_index.go:213` — `selectByteStats` does `slices.Clone(value)`, then the only consumer (`builder.LowerBoundValues`/`UpperBoundValues`) calls `mapToAvroColMapClonedBytes`, which clones again. Drop the first.
- `delete_file_index.go:99-101` + `equality_delete_index.go:481` — equality field IDs are cloned **three** times (call site, `:101`, then `builder.EqualityFieldIDs`); same double-clone for `KeyMetadata` and `SplitOffsets`. ~6 allocations per delete file, material against the measured 42/file.
- `delete_file_index.go:59` — the synthetic spec's only product is dead weight. `fieldNameToID` is read only by `initPartitionData`, which is a guaranteed no-op here (the builder pre-sets `fieldIDToPartitionData` at the same length), and `PartitionData` is unconditionally overwritten by `MarshalAvroEntry` from `DataFilePartitionRef` + the caller's real spec. So every compact copy retains two maps keyed by `partition_<id>` names that nothing reads: 1,516 B/file partitioned vs 970 B/file unpartitioned, and roughly two-thirds of that 546 B delta is provably dead. This is a memory PR and the fix is in scope.
- `scanner.go:900-902` — `buildDVIndex` tests `ref != nil` but not `*ref != ""`, so a DV with an empty-string `referenced_data_file` is indexed under `""` and `compaction.referencedDataFilePath` (stricter) can no longer fall back to `file_path` bounds. Robustness only — catalog metadata is trusted per the threat model — but worth aligning the two predicates.
- `delete_file_index.go:42-44` — the doc says positional indexes select `filePathFieldID`; true only for `byPartition`. `byPath` selects nothing and relies on the materialised reference. That exact distinction is what broke last round, so it's worth a sentence.
- Intermediate commit `5fafb17f` doesn't build its test binary (type errors in `scanner_plan_tasks_bench_test.go` / `scanner_internal_test.go`); `aa230a5c` fixes it. Harmless under squash-merge, noted for `git bisect`.
- `scanner.go:1551-1553` — the release is genuinely effective, not cosmetic: `newManifestEntries()` returns a `*manifestEntries`, so nilling the fields mutates the shared heap object and nothing else retains it. Good change, accurate comment.

## Prior items

**Mine (CHANGES_REQUESTED 2026-09-01):**
1. Inferred path scope lost before compaction sees it → **Fixed.** `5fafb17f` passes `&path` into `compactDeleteFileForIndexWithReference`. Mutation-verified: reverting that single argument makes `TestPlanCompaction_BoundsScopedPositionalDeleteRatioFromScan` fail and nothing else. Probe confirms `ReferencedDataFile()` is now `data-1.parquet` where the manifest had `nil`. You took the option I recommended.
2. Test across that seam → **Fixed**, and confirmed a live guard rather than decoration.
3. Rebase → **Fixed.**
4. Benchmark reports absolutes with no baseline → **Still open.** I measured it: construction is **+1295% to +2876% slower** and **+1158% to +3225% more bytes**, in exchange for 99.3–99.6% less retained statistics (wide) / 50–70% (narrow). See below.
5. Equality NaN counts for #1960 → **re-confirmed**; dropping them fails `TestEqualityDeleteIndexPrunesNullOnlyMetricRanges` and `TestEqualityDeleteIndexDoesNotUseUncertainFloatBounds`.

**@laskoviymishka:** base one commit short of #1974 → **Fixed.** Always-nil error return → **Fixed**, four dead branches gone. Synthetic spec source IDs → **Partially fixed** (documentation option taken). I verified why that's defensible: `SpecID()` returns the real ID, `Partition()` returns real id-keyed values, the synthetic names are unreachable, and a full `EncodeFileScanTask`→`DecodeFileScanTask` round-trip of a compacted *partitioned* positional delete preserves partition values, path, count and the retained `file_path` lower bound. The leftover is wasted retention (Minor), not incorrect source IDs escaping. Compact-vs-full parity across equality and DV → **Still open** — present in form, inert in substance (Blocking).

## Description

- **Omits a public-surface behaviour change.** The compact copy now *materialises* an inferred `referenced_data_file` on path-scoped positional deletes: `FileScanTask.DeleteFiles[i].ReferencedDataFile()` was `nil` before and is now the inferred data path, while the `file_path` bounds that used to be there are gone. That's the whole point of `5fafb17f` and it belongs in the summary — a downstream consumer round-tripping a scan-planned delete file into a manifest will now write a field the source manifest didn't carry.
- **Omits `table/compaction/analyze_test.go`** and any mention of compaction. A reader can't tell the trimming interacts with `compaction.isFileScoped` at all — which is precisely where it broke.
- **Benchmark table has no baseline and reads as a win table.** The two deterministic columns reproduce exactly, and B/op and allocs/op are within run-to-run variance of your M1 figures, so it's **not stale**. But these absolutes are ~13× the pre-change cost; the win lives entirely in the two custom metrics. Please add the base column and label ns/op and B/op as the cost side.
- **Test list is stale** — omits both tests added in the last two commits.
- `BenchmarkDeleteIndexRetainsOnlyRequiredStats` passes a `nil` schema to `buildEqualityDeleteIndex` (`delete_file_index_bench_test.go:54`), short-circuiting `newEqualityDeleteIndexEntry` before any field-metric decoding — so the equality row measures stat selection only, not the metric path #1960 added. Worth a footnote.
- `source_stats_fields/file` is asserted, not measured (`:37-40` computes `5*fieldCount [+5]` as a constant). It matches the fixture today but will silently drift; `countDataFileStats` on a source file would make it self-checking, as the retained column already is.

---

table/scanner.go
```diff
@@ -923,13 +929,18 @@ func buildDVIndex(dvEntries []iceberg.ManifestEntry) (map[string]iceberg.Manifes
 // would never satisfy dataSeq <= -1 and would silently drop the DV,
 // resurfacing deleted rows; an unset data sequence likewise satisfies
 // -1 <= dvSeq for any known DV sequence.
-func matchDVToData(dataEntry iceberg.ManifestEntry, dvIndex map[string]iceberg.ManifestEntry) []iceberg.DataFile {
-	dvEntry, ok := dvIndex[dataEntry.DataFile().FilePath()]
+func matchDVToData(dataEntry iceberg.ManifestEntry, dvIndex map[string]deleteFileIndexEntry) []iceberg.DataFile {
+	dataFile := dataEntry.DataFile()
+	dvEntry, ok := dvIndex[dataFile.FilePath()]
 	if !ok {
```
zeroshade, line 935: **major** — New DV spec/partition gate silently drops deletion vectors instead of erroring

matchDVToData now returns nil when dvEntry.file.SpecID() != dataFile.SpecID() or partitions differ. origin/main has no such check (verified: `git show origin/main:table/scanner.go` sed 646-656 matches purely on referenced_data_file + sequence number). Two problems. (a) Failure mode: when matchDVToData returns nil, fileScanTaskForDataEntry (scanner.go:1376-1382) falls back to posDeleteIndex.forDataFile; on a v3 DV-only table there are no positional deletes, so the data file is planned with NO deletes and previously-deleted rows resurface with no error and no log. A gate that can only ever reject should surface an error, not a silent nil. (b) Inconsistency: the equality-delete path 50 lines earlier (scanner.go:877-882) deliberately guards `len(delPartition) > 0 && len(dataPartition) > 0` so empty partitions apply globally; the new DV gate has no such guard, so a nil-partition DV is dropped where a nil-partition equality delete is kept. This is also unrelated to the PR's stated purpose (retaining only required stats). Calibration: I could not construct a mismatch from iceberg-go's own writer (dv_writer.go:250 derives the DV spec/partition from the data file), so I am NOT claiming proven live corruption - the finding is the wrong failure mode plus unrequested scope in a perf PR.

```text
Probe TestProbe_DVSpecEvolutionDropsDeletionVector: 'DV explicitly referencing s3://bucket/data/data-001.parquet -> matched 0 delete files' ... 'Error: "[]" should have 1 item(s), but has 0'. Probe TestProbe_DVNilPartitionVsPartitionedData: 'nil-partition DV -> matched 0 delete files'. Both DVs carry referenced_data_file naming the exact data file.
```

</details>

table/delete_file_index.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// compactDeleteFileForIndex copies a delete file without retaining the wide
+// metadata maps that are not needed by scan planning or delete-file readers.
+// The returned file still carries all non-statistics metadata exposed by the
+// DataFile interface, plus only the selected statistics fields.
+//
+// Positional-delete indexes select file_pathFieldID because partition-scoped
+// position deletes use those bounds for candidate pruning. Equality-delete
+// indexes select their equality field IDs. Deletion vectors select no stats.
```
zeroshade, line 40: **minor** — PR body and source comment claim DV stat trimming that the code does not implement

The comment at delete_file_index.go:39 states 'Deletion vectors select no stats' and the PR body states 'Deletion vectors retain no stats, but keep the metadata needed to read the Puffin range'. Neither is true: `grep -rn compactDeleteFileForIndex --include=*.go` shows only three call sites (positional_delete_index.go:44,59 and equality_delete_index.go:481) - none for DVs. buildDVIndex (scanner.go:909) stores the original unmodified iceberg.DataFile. Consequently the `deleteEntries.dvEntries = nil` release at scanner.go:1558 frees only the ManifestEntry wrappers while every DV's wide stat maps stay live for the rest of planning, so the claimed DV saving does not exist. Either wire DVs through compactDeleteFileForIndex or correct the comment and the PR description.

table/equality_delete_index.go
```diff
@@ -474,13 +478,14 @@ func buildEqualityDeleteIndex(
 			isUnpartitioned = spec.IsUnpartitioned()
 			unpartitionedBySpecID[df.SpecID()] = isUnpartitioned
 		}
+		indexedFile := compactDeleteFileForIndex(df, partition, df.EqualityFieldIDs())
+		indexedEntry := newEqualityDeleteIndexEntry(entry, indexedFile, schema)
```
zeroshade, line 482: **minor** — Allocating public getter used in the hot indexing loop of an allocation-reduction PR

`compactDeleteFileForIndex(df, partition, df.EqualityFieldIDs())` calls the public getter, which does `slices.Clone(*d.EqualityIDs)` (manifest.go:2901-2907), allocating a fresh slice per delete file. compactDeleteFileForIndex then immediately re-reads the same equality field IDs through the borrowed, non-allocating accessor `dataFileCollections(file)` (delete_file_index.go:101). Using the borrowed accessor at the call site as well would drop one allocation per delete file, which is squarely the goal of this PR.

table/arrow_scanner.go
```diff
@@ -285,10 +285,10 @@ func sameDVBlob(a, b iceberg.DataFile) bool {
...
 	}
 
-	_, _, _, aOffset, _ := iceinternal.BorrowedDataFilePointers(a)
-	_, _, _, bOffset, _ := iceinternal.BorrowedDataFilePointers(b)
+	_, _, _, aOffset, aSize := iceinternal.BorrowedDataFilePointers(a)
+	_, _, _, bOffset, bSize := iceinternal.BorrowedDataFilePointers(b)
 
-	return *aOffset == *bOffset
+	return *aOffset == *bOffset && *aSize == *bSize
 }
```
zeroshade, line 292: **minor** — Unrelated behavior changes bundled into a perf PR

Four observable-behavior changes ride along with the stat-trimming work: sameDVBlob now also compares content_size (arrow_scanner.go:288-291); buildDVIndex now hard-errors on a missing/empty referenced_data_file where main silently skipped the entry (scanner.go:906); filePathMayMatch now skips pruning entirely when bounds are inverted (positional_delete_index.go:155-163); and appendDeletionVectorRows / readAllDeletionVectors now reject empty-string refs. Each is individually defensible - the inverted-bounds change is conservative and the empty-ref checks close a real hole - but none is stat retention, and buildDVIndex's new error turns a table that previously scanned into a hard scan failure. Splitting these into their own PR would let them be reviewed on their own merits and keep this diff's perf claim auditable. I verified sameDVBlob cannot nil-deref: readAllDeletionVectors nil-checks contentOffset/contentSize (arrow_scanner.go:173) before the uniqueDVs lookup that reaches sameDVBlob.

zeroshade (commented):
Prior findings [0],[1],[2] are genuinely fixed and stat trimming is proven behavior-preserving by a 400-input differential probe, but the PR bundles a new DV spec/partition gate that silently drops deletion vectors and ships a DV stat-trimming claim the code does not implement.

Re-review verification: 3 of 4 prior findings confirmed fixed at `4d0b89c` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *partially fixed* — Nothing asserts a compact index returns the same delete-file matches as a full-stat index across equality and DV paths

```text
go build ./... (OK); go vet ./table/... (clean); go test ./table ./table/compaction -count=1 (ok 5.9s / 2.1s); go test -race ./table -run 'TestCompactDeleteFileForIndex|TestDeleteIndexesRetain|TestPositionalDeleteIndex|TestEqualityDeleteIndex|TestBuildDVIndex|TestMatchDV|TestPlanFiles|TestFilePathMayMatch|TestReadAllDeletionVectors' (ok 2.3s); go test ./catalog/glue ./catalog/rest (ok, refuting the pi-lens setup-failure reports). Mutation: forced dataFileStatsForFields to return all-nil -> 7 tests red (TestCompactDeleteFileForIndexRetainsSelectedStats, TestDeleteIndexesRetainOnlyRequiredStats, TestDeleteIndexesRetainMatchingParity/{positional,equality}, TestEqualityDeleteIndexPrunesDisjointMetricRanges, TestEqualityDeleteIndexPrunesNullOnlyMetricRanges, TestEqualityDeleteIndexDoesNotUseUncertainFloatBounds, TestEqualityDeleteReadPrunesNonOverlappingDeleteFiles); restored via git checkout. Throwaway probe table/pr1975_probe_test.go (4 probes) run then deleted; git status --porcelain empty at finish.
```

</details>

---

>

table/scanner.go
```diff
@@ -923,13 +929,18 @@ func buildDVIndex(dvEntries []iceberg.ManifestEntry) (map[string]iceberg.Manifes
...
 		return nil
 	}
-	if dvSeq := dvEntry.SequenceNum(); dvSeq < 0 || dataEntry.SequenceNum() <= dvSeq {
-		return []iceberg.DataFile{dvEntry.DataFile()}
+	if dvEntry.file.SpecID() != dataFile.SpecID() ||
```
laskoviymishka: Agreed with the earlier point that this gate drops the DV silently on a mismatch, where Java errors out via validatePartitionMatch. Our metadata is trusted so I don't expect it to fire, but a silent skip is the one failure mode that resurfaces deleted rows, so I'd either propagate an error like Java or leave a short comment that the skip is deliberate. Happy to keep the details on that thread.

table/delete_file_index_test.go
```diff
@@ -0,0 +1,403 @@
...
+			newDeleteIndexStatsFile(t, spec, iceberg.EntryContentData, iceberg.ParquetFile,
+				dataPath, map[int]any{1000: "us"}, 1, 100, nil, nil, nil, nil, nil, nil),
+		)
+
+		fullMatches, err := fullIndex.forDataFile(dataEntry)
+		require.NoError(t, err)
+		compactMatches, err := compactIndex.forDataFile(dataEntry)
+		require.NoError(t, err)
+		assert.Empty(t, fullMatches)
+		assert.Empty(t, compactMatches)
```
laskoviymishka, line 363: This closes the equivalence gap I flagged last round nicely, the compact and full indexes genuinely diverge on the prune case for both positional and equality now.

The one piece still missing is the include case: an assertion that a delete which should still match is retained by both indexes, not just that a non-matching one is pruned. If the compact bounds were ever subtly narrower, these prune-only subtests wouldn't catch it. Could we add an include subtest with a data path inside the bounds asserting both return the same non-empty match?

table/delete_file_index_test.go
```diff
@@ -0,0 +1,403 @@
...
+		count:       1,
+		filesize:    1,
+		valueCounts: map[int]int64{1: 1},
+	}
+
+	compacted := compactDeleteFileForIndex(file, file.partition, []int{1})
+	assert.Same(t, file, compacted)
+}
+
+func TestDeleteIndexesRetainOnlyRequiredStats(t *testing.T) {
```
laskoviymishka, line 181: This one runs the four index types as flat sequential blocks, so a require failure in the positional section hides the equality and DV sections below it. TestDeleteIndexesRetainMatchingParity right below already uses t.Run per type; I'd split this the same way so a failure points straight at the index that broke.

laskoviymishka (approved):
Almost there. The rebase landed and CI is green, so the merge-type problem I gated on last round is fully resolved, and everything else I asked for is in too.

The one substantive thing still open for me is the DV spec/partition gate that came up in the earlier thread: on a mismatch, matchDVToData returns nil and drops the deletion vector silently rather than erroring the way Java does. Our metadata is trusted so I don't expect it to fire, but a silent skip is the failure mode that resurfaces deleted rows, so I'd like us to either propagate the error like Java or leave a comment that the skip is deliberate. I'm aligned with that concern, happy to keep the details on that thread.

On my own note from last round, the one follow-on I'd add is an include-case assertion in the parity test, so we also pin that a delete which should still match is retained by both indexes, not just that a non-matching one is pruned. I left the specifics inline.

Everything I flagged in review #1 is in:
- rebase done, the scanner.go merge conflict and DV-type collision are gone, all checks green
- the inferred positional-delete scope fix landed with its regression test, and the compaction-from-scan test exercises the bounds-to-referenced_data_file path end to end
- the trimming-equivalence test gap is closed, the parity test is divergent now rather than a tautology

Settle the DV skip question and add the include-case assertion and I'm happy to approve.

table/scanner.go
```diff
@@ -942,16 +950,27 @@ func buildDVIndex(dvEntries []iceberg.ManifestEntry) (map[string]iceberg.Manifes
...
+	}
+	if dvEntry.file.SpecID() != dataFile.SpecID() {
+		return nil, fmt.Errorf("%w: deletion vector %s does not match data file %s partition spec",
+			ErrInvalidMetadata, dvEntry.file.FilePath(), dataFile.FilePath())
 	}
-	if dvSeq := dvEntry.SequenceNum(); dvSeq < 0 || dataEntry.SequenceNum() <= dvSeq {
-		return []iceberg.DataFile{dvEntry.DataFile()}
+	dvPartition := dataFilePartition(dvEntry.file)
+	dataPartition := dataFilePartition(dataFile)
+	if len(dvPartition) > 0 && len(dataPartition) > 0 && !partitionsMatch(dvPartition, dataPartition) {
```
zeroshade: This is the nil-partition case from the earlier round, changed from silently dropping the DV to silently applying invalid metadata.

```go
if len(dvPartition) > 0 && len(dataPartition) > 0 && !partitionsMatch(dvPartition, dataPartition) {
```

The length guard skips the equality check whenever either side is empty, so same-spec metadata `{}` against `{1000: "us"}` — in either direction — falls through to the sequence check and the DV is attached rather than returning `ErrInvalidMetadata`. I reproduced both directions; both returned a nil error.

Iceberg requires the spec *and* the partition values to match for deletion vectors; the unpartitioned/global exception applies to equality deletes only. Nil and empty maps already compare equal under `partitionsMatch`, so the guard buys nothing for genuinely unpartitioned files:

```go
if !partitionsMatch(dvPartition, dataPartition) {
```

Please add both one-sided-empty directions to `TestMatchDVToData_RequiresMatchingPartition`.

zeroshade (changes requested):
REQUEST_CHANGES — the 09-09 fix closes the spec-mismatch half of the earlier DV thread but leaves the nil/empty partition case open, now failing in the opposite direction. Detail inline.

The stat-retention contract itself checks out: partition-scoped position deletes keep all five `file_path` metric maps, file-scoped deletes materialize the reference compaction needs, equality deletes keep all five metric maps for every equality field, and DVs are left unmodified. I found no nil-map or nil-slice dereference on the pruned paths, and the targeted race and end-to-end tests pass. The earlier positional-delete compaction regression is fixed end to end, and the hot-path equality-ID allocation is gone.

Non-blocking: the PR summary still says deletion vectors "retain no stats", but `buildDVIndex` deliberately stores the original `DataFile` — only the `ManifestEntry` wrapper is dropped. The source comment is now accurate; please bring the description in line with it.

PR #1976, opened 2026-08-29
perf(table): flatten manifest entries after parallel reads

fallintoplace:
## Summary

- **Keep one classified result per manifest** while reads run in parallel.
- **Flatten after all reads finish** into exact-sized data and delete slices.
- Preserve manifest-list order and keep `manifestEntries.merge` for existing callers.
- Add focused tests for mixed buckets, empty result slots, and exact capacities.
- Add a merge-versus-flatten benchmark.

## Benchmark

Apple M1 Pro. 64 manifests, 10,000 entries per manifest, concurrency 16.

- **Data manifests:** 18.4 ms/op and 71.9 MB/op with merge, versus 6.4 ms/op and 20.7 MB/op with flatten.
- **Delete manifests:** 12.9 ms/op and 74.6 MB/op with merge, versus 6.3 ms/op and 32.4 MB/op with flatten.

## Tests

- `go test ./... -count=1`
- `go vet ./table`
- Targeted `go test -race ./table` pass.

### DIFF
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -1093,7 +1093,6 @@ func (scan *Scan) collectManifestEntriesWithSchema(
 			continue
 		}
 
-		manifestIndex, mf := manifestIndex, mf
 		g.Go(func() error {
 			fs, err := manifestIO.acquire(gctx)
 			if err != nil {
--- a/table/scanner.go
+++ b/table/scanner.go
@@ -315,6 +315,41 @@ func (m *manifestEntries) merge(entries []iceberg.ManifestEntry) error {
 	return err
 }
 
+func flattenClassifiedManifestEntries(results []classifiedManifestEntries) *manifestEntries {
+	var counts [manifestEntryKindCount]int
+	for _, result := range results {
+		counts[manifestEntryData] += len(result.dataEntries)
+		counts[manifestEntryPositionalDelete] += len(result.positionalDeleteEntries)
+		counts[manifestEntryEqualityDelete] += len(result.equalityDeleteEntries)
+		counts[manifestEntryDV] += len(result.dvEntries)
+	}
+
+	flattened := &manifestEntries{
+		dataEntries:             make([]iceberg.ManifestEntry, counts[manifestEntryData]),
+		positionalDeleteEntries: make([]iceberg.ManifestEntry, counts[manifestEntryPositionalDelete]),
+		equalityDeleteEntries:   make([]iceberg.ManifestEntry, counts[manifestEntryEqualityDelete]),
+		dvEntries:               make([]iceberg.ManifestEntry, counts[manifestEntryDV]),
+	}
+
+	var offsets [manifestEntryKindCount]int
+	for _, result := range results {
+		offsets[manifestEntryData] += copy(
+			flattened.dataEntries[offsets[manifestEntryData]:], result.dataEntries,
+		)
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
This keeps manifest-entry collection complete while removing merge-time append growth. Approving.

Workers classify into disjoint per-manifest slots, `g.Wait()` forms the barrier, and the flatten then copies slots in manifest-list order. What I verified:

- **Completeness.** `flattenClassifiedManifestEntries` (`table/scanner.go:318-348`) counts each category, allocates exact lengths, and copies every per-manifest batch exactly once — nothing dropped at batch boundaries, nothing duplicated. Skipped sequence-number manifests leave empty slots rather than shifting anything.
- **Deterministic order.** Parallel reads complete out of order, so this matters: `collectManifestEntriesWithSchema` (`:1005-1046`) writes results to disjoint manifest-index slots and flattens only after the barrier, so output is in manifest-list order. Worth noting this is actually *more* deterministic than before — the previous merge appended under a mutex in completion order, so the new behaviour is the intended planning order rather than an accidental one. Entry order within each manifest is preserved.
- **No partial result on error.** Any open or classification failure returns `nil, error` before the flatten runs, so a truncated result can never be exposed as complete.
- **Race safety.** Workers touch only their own index; the accumulator isn't written by workers at all, and reads happen only after `errgroup.Wait`. Go 1.25 per-iteration loop variables make the closure captures safe.
- **Edge cases.** Empty input yields non-nil zero-length slices; single-manifest and mixed-kind cases are covered with exact-capacity and order assertions in `table/scanner_manifest_entries_test.go`.

I checked that no consumer depends on completion order — `PlanFiles` indexes the delete categories and then iterates `dataEntries` — so the ordering change is safe.

**Relationship to #1974:** these are complementary rather than duplicate, which was my main question given both restructure planning. #1974 streams data manifests directly into task batches, bypassing this collector; this PR improves the collector that #1974 still uses for delete manifests, plus the legacy and incremental paths. They do overlap in `scanner.go`, so whichever lands second will need a rebase.

One honest trade-off: this retains per-manifest batches plus the flattened arrays until return, so peak live heap is higher than a streaming approach. The benchmark shows that's a good deal here — 64×10k at concurrency 16 on M1: data merge 18.4 ms/71.9 MB → flatten 6.4 ms/20.7 MB; deletes 12.9 ms/74.6 MB → 6.3 ms/32.4 MB.

The `schema.go` hunk is the known lazy-cache fix already merged on main; it should disappear on rebase.

No findings. CI green, all 3 commits signed off.

---

>

zeroshade:
My approval above stands, but this went conflicted before I could merge it.

#1781 (`reuse FileIO during local scan planning`) landed first and also edits `table/scanner.go` — specifically `collectManifestEntriesWithSchema`, which this PR restructures. That's the overlap I noted in the review. The duplicated `schema.go` lazy-cache hunk is also on `main` now and should drop out.

A rebase onto current `main` should resolve both. The one thing to check while rebasing: #1781 introduced `manifestIOBatch` acquisition inside the same collector loop this PR converts to per-manifest result slots, so please confirm the batch `acquire` call still sits inside the worker (and still sees the errgroup context) once the slots refactor is reapplied.

Also worth sequencing against #1974, which is approved and awaiting its own rebase and touches the same collector from the streaming side. Once rebased and green I'll merge — no need to re-request review.

PR #1981, opened 2026-09-01
perf(table): skip metadata normalization on the common path

### LINKED ISSUE #1979: perf(table): eliminate repeated full-document scans in ParseMetadataBytes
### Description

`table.ParseMetadataBytes` currently makes multiple complete passes over the metadata JSON before returning a concrete metadata object:

1. `assignMissingPartitionFieldIDs` decodes the full document into raw-message maps to normalize legacy partition field IDs.
2. `requirePartitionSpecIDs` decodes the full document again to validate partition spec IDs.
3. Format-version detection via `json.Unmarshal` scans the full input even though it needs one top-level field.
4. The version-specific metadata decoder scans the full input again.

The deferred-snapshot work discussed in #1946 additionally captures and validates raw snapshots. Avoiding full historical snapshot materialization reduces allocations, but repeated full-document scans and copies can still consume substantial CPU for large metadata.

This issue proposes a general `ParseMetadataBytes` optimization, separate from #1946, because it should benefit eager parsing and all callers rather than only REST commit responses.

### Proposed direction

Introduce a single top-level structural scan that can:

- identify and validate `format-version`;
- establish that required fields such as `last-updated-ms` are present with the expected JSON shape;
- record byte ranges for `partition-spec`, `partition-specs`, `snapshots`, and `snapshot-log`;
- detect the partition-spec fields that require legacy field-ID normalization or spec-ID validation;
- allow the ordinary version-specific decoder to process a document with a large snapshot array removed or replaced by an empty array when deferred snapshot parsing is requested;
- validate and normalize only the relevant partition-spec subdocuments instead of decoding the complete metadata document into raw-message maps repeatedly; and
- retain raw ranges safely by copying owned bytes rather than keeping aliases into caller-owned input.

The exact implementation does not need to be a hand-written general JSON decoder. A small structural scanner combined with the standard library decoder may be sufficient and easier to maintain.

### Correctness requirements

- Preserve support and validation semantics for metadata format versions v1, v2, and v3.
- Preserve legacy missing partition field-ID assignment and required partition spec-ID behavior.
- Preserve precise, synchronous errors for malformed JSON, missing required fields, invalid field types, overflow, duplicates, and version-specific constraints.
- Preserve unknown-field compatibility and current `null`/absent-field behavior.
- Preserve input ownership: any JSON retained after parsing must be copied before `ParseMetadataBytes` returns.
- Preserve serialization, equality, builders, version upgrades, and concurrent-read behavior.
- Do not change public API behavior or move existing parse-time errors to later accessors.
- Prefer the Go standard library unless a dependency has a clearly demonstrated, compatible, and safe advantage.

### Benchmarks and profiling

Add benchmarks organized by actual serialized metadata byte size, not only snapshot count. Suggested size points include approximately 100 KiB, 500 KiB, 1 MiB, 2 MiB, and 5 MiB, with the exact generated byte count reported.

For each relevant case, compare:

- current eager parsing;
- optimized eager parsing;
- deferred-snapshot parsing from #1946, when available; and
- deferred parsing plus current-snapshot access versus a historical lookup/full materialization where useful.

Report `ns/op`, `B/op`, and `allocs/op`, and collect CPU/allocation profiles to establish which repeated passes materially affect runtime. Fixtures should distinguish snapshot-heavy metadata from metadata whose size comes from schemas, partition specs, properties, or snapshot logs.

PR #1654 added a reusable load-table response JSON benchmark and evaluated `goccy/go-json`. Its reported results were mixed across snapshot counts, and the proposed replacement also raised encoder compatibility and concurrency-safety concerns, so a decoder replacement was not merged. A future standard-library JSON implementation or safe decoder evaluation is orthogonal to this issue: a faster decoder may reduce the cost of each pass, while this issue aims to eliminate unnecessary passes altogether.

Related: #1946
Related benchmark work: #1654

### Are you willing to contribute this feature?

- [x] Yes, I am willing to contribute this feature.

technicolorbeat:
Closes #1979.

## Motivation

`table.ParseMetadataBytes` is on the critical path for table loads, refreshes, and metadata returned by catalogs. Before this change, every metadata document paid for several full-document operations:

1. decode `format-version`;
2. decode the entire document into `json.RawMessage` maps to normalize legacy partition field IDs;
3. decode it again to validate partition spec IDs; and
4. decode the concrete v1/v2/v3 metadata value.

The normalization pass copies top-level JSON values and re-marshals the document even though current metadata already contains partition field IDs. Large metadata files—especially tables with long snapshot histories or large properties maps—therefore incur avoidable latency, transient memory, and garbage-collection work on every load.

The optimization should not require Iceberg to own a second JSON parser. This change keeps `encoding/json` as the only JSON implementation and makes the expensive compatibility work conditional.

## Design

The first decode targets a small header containing:

- `format-version`;
- `last-updated-ms`;
- `partition-spec`; and
- `partition-specs`.

Partition values are retained as `json.RawMessage`. The header inspection:

- selects the concrete metadata version;
- preserves the required `last-updated-ms` check;
- validates required partition `spec-id` values; and
- detects partition fields that omit `field-id`.

For normal metadata, the original bytes then go directly to the existing concrete v1/v2/v3 decoder. This avoids the full-document map construction, normalization marshal, and redundant spec-ID pass.

For legacy metadata with missing partition field IDs, the code deliberately falls back to the existing normalization and validation functions. This keeps compatibility behavior isolated and unchanged while optimizing the overwhelmingly common case.

All JSON syntax, escaping, duplicate-key, and concrete metadata semantics remain governed by `encoding/json`. There is no custom tokenizer or structural scanner.

## Benchmarks

The benchmark uses actual serialized document size and covers three shapes:

- `snapshot-history`: large arrays of nested snapshot objects;
- `properties`: document size concentrated in a top-level property map;
- `legacy-field-id-fallback`: v1 metadata that requires compatibility normalization.

Command:

```console
go test ./table -run '^$' -bench '^BenchmarkParseMetadataBytes$' -benchmem -benchtime=3x -count=3
```

Apple M1 Max, darwin/arm64. Values below are medians of three runs.

| Profile | Size | Before ns/op | After ns/op | Time | Before B/op | After B/op | Bytes |
|---|---:|---:|---:|---:|---:|---:|---:|
| snapshot-history | 100 KiB | 4,887,292 | 3,499,764 | -28.4% | 793,128 | 571,304 | -28.0% |
| snapshot-history | 1 MiB | 49,876,389 | 35,710,208 | -28.4% | 7,702,250 | 5,595,794 | -27.3% |
| snapshot-history | 5 MiB | 247,908,708 | 177,362,611 | -28.5% | 39,873,421 | 29,378,352 | -26.3% |
| properties | 100 KiB | 3,083,320 | 1,893,958 | -38.6% | 389,528 | 166,552 | -57.2% |
| properties | 1 MiB | 30,861,861 | 18,542,583 | -39.9% | 3,873,040 | 1,765,352 | -54.4% |
| properties | 5 MiB | 154,029,917 | 93,257,097 | -39.5% | 18,591,010 | 8,094,805 | -56.5% |
| legacy fallback | 100 KiB | 3,497,750 | 3,554,583 | +1.6% | 513,488 | 478,840 | -6.7% |
| legacy fallback | 1 MiB | 35,734,861 | 36,039,153 | +0.9% | 5,575,933 | 5,604,341 | +0.5% |
| legacy fallback | 5 MiB | 179,071,792 | 180,047,445 | +0.5% | 27,312,224 | 29,061,160 | +6.4% |

Across all measured sizes, normal metadata is approximately 28–40% faster and allocates 26–57% fewer bytes. The rare compatibility fallback remains within roughly 0–2% CPU of the previous implementation.

The benchmark reports `metadata-bytes` after `ResetTimer`, ensuring the custom metric is retained in benchmark output.

## Correctness coverage

Tests cover:

- fast-path selection for v1, v2, and v3 metadata;
- fallback selection for both legacy `partition-spec` and `partition-specs` forms;
- missing partition field-ID assignment across specs;
- missing and null partition spec IDs;
- malformed and trailing JSON;
- escaped top-level keys and nested unknown values;
- duplicate-key last-value behavior;
- top-level `null` error compatibility; and
- all existing version-specific metadata constraints through the repository suite.

## Validation

```console
go test -p 2 ./...
golangci-lint run ./...
```

Both pass locally. `-p 2` was used only to limit peak linker resource usage.


technicolorbeat:
Resolved by #1933, which reuses the top-level raw JSON map across metadata preflight checks and eliminates the redundant full-document decoding passes targeted here.

PR #1984, opened 2026-09-02
fix(manifest): normalize file_format casing when reading manifest entries

### LINKED ISSUE #1983: Manifest decoding does not support lower-case values for `file_format`, making data files from some writers unreadable
### Apache Iceberg version

main (development)

## Version

Observed on `main` at `87789102577c` (2026-09-01) using DuckDB `v1.5.5` as writer.

## Summary

Reading a table whose Avro manifests specify a lowercase `file_format` for data files (e.g. `parquet`) fails on the first data file with the following error:
```
not implemented: only parquet format is implemented, got parquet
```

This happens via the ordinary read path: `LoadTable` followed by `tbl.Scan()`.

The Iceberg spec [uses lowercase strings as illustrative example values](https://iceberg.apache.org/spec/#data-file-fields) for `file_format`, so so I don't think this is reasonably characterized as a DuckDB quirk/bug.

Meanwhile, unlike `iceberg-go`, both `pyiceberg` and the Java reference implementation appear to be case-agnostic with respect to `file_format` (Pyiceberg has a [test case](https://github.com/apache/iceberg-python/blob/d12f68042a459e5d4732bd6ee5a188e891b0dd44/tests/catalog/test_scan_planning_models.py#L208) specifically intended to ensure case-insensitivity when scanning). `iceberg-go` ought to behave similarly to ensure portability of tables written by other writers.

<details><summary><b>Steps to reproduce</b></summary>
<br/>

**1. Start a REST catalog over a local directory.**

```bash
mkdir -p /tmp/icewh
docker run -d --name ice-rest -p 8181:8181 \
  -v /tmp/icewh:/tmp/icewh \
  -e CATALOG_WAREHOUSE=file:///tmp/icewh \
  apache/iceberg-rest-fixture:latest
```

**2. Create and populate a table with DuckDB.**

```sql
INSTALL iceberg; LOAD iceberg;
ATTACH 'file:///tmp/icewh' AS cat (
    TYPE ICEBERG, ENDPOINT 'http://localhost:8181', AUTHORIZATION_TYPE 'none'
);
CREATE SCHEMA cat.ns;
CREATE TABLE cat.ns.t (id BIGINT);
INSERT INTO cat.ns.t VALUES (1);
```

**3. Confirm the manifest records a lowercase `file_format`**

```sql
SELECT file_format FROM iceberg_metadata(cat.ns.t); -- "parquet" expected
```

**4. Scan the table with iceberg-go**:

```go
cat, _ := rest.NewCatalog(ctx, "cat", "http://localhost:8181",
    rest.WithWarehouseLocation("file:///tmp/icewh"))

tbl, err := cat.LoadTable(ctx, table.Identifier{"ns", "t"})
// succeeds: table metadata and manifest lists parse fine

_, recs, err := tbl.Scan().ToArrowRecords(ctx)
for rec, err := range recs {
    if err != nil {
        log.Fatal(err) // not implemented: only parquet format is implemented, got parquet
    }
    _ = rec
}
```
</details> 

<details><summary><b>Details on Java/Python behavior</b></summary>
<br/>
Both the Java and Python reference implementations avoid problems here by normalizing `file_format` casing on decode. `pyiceberg` has a test case specifically to ensure case-insensitivity.

**Java** — `BaseFile.internalSet` routes the raw Avro value through `FileFormat.fromString`:

```java
// core/src/main/java/org/apache/iceberg/BaseFile.java
case 2:
  this.format = FileFormat.fromString(value.toString());
  return;
```

```java
// api/src/main/java/org/apache/iceberg/FileFormat.java
public static FileFormat fromString(String fileFormat) {
  Preconditions.checkArgument(null != fileFormat, "Invalid file format: null");
  try {
    return FileFormat.valueOf(fileFormat.toUpperCase(Locale.ROOT));
  } catch (IllegalArgumentException e) {
    throw new IllegalArgumentException(
      String.format("Invalid file format: %s", fileFormat), e);
  }
}
```

**Python** — `FileFormat` is a `str` enum with a `_missing_` hook that uppercases before matching:

```python
# pyiceberg/manifest.py
class FileFormat(str, Enum):
    AVRO = "AVRO"
    PARQUET = "PARQUET"
    ORC = "ORC"
    PUFFIN = "PUFFIN"

    @classmethod
    def _missing_(cls, value: object) -> None | str:
        for member in cls:
            if member.value == str(value).upper():
                return member
        return None
```
</details>

wroever:
Closes #1983, which has the full write-up and reproduction.

## Problem

The Avro manifest decoders assign the wire `file_format` string directly into the typed `FileFormat` field, whose constants are uppercase:

```go
// manifest.go
Format FileFormat `avro:"file_format"`
```

`GetFile` (`table/internal/interfaces.go`) then matches the decoded value against `iceberg.ParquetFile` (`"PARQUET"`) to pick a reader. A manifest written with lowercase `parquet` never matches, so the scan fails on the first data file — with a message that renders both spellings identically:

```
not implemented: only parquet format is implemented, got parquet
```

Two decode paths are affected: `ManifestReader.ReadEntry`, and `unmarshalAvroDataFileEntry` (`data_file_codec.go`), which is reachable publicly through `codec.DecodeDataFile`. Beyond `GetFile`, the un-normalized value also mis-compares at `table/scan_splits.go`, `table/scanner.go` (`IsDeletionVector`), and `table/dv/deletion_vector.go`.

Nothing unusual is needed to hit this: `LoadTable` followed by a bare `tbl.Scan()`. DuckDB's `iceberg` extension writes lowercase, so no table it has written is readable. It also isn't limited to explicit scans — `Transaction.Delete` runs a copy-on-write rewrite through the same path, so appends and deletes against such a table fail too.

## Fix

Factor the normalization into `dataFile.normalizeFormat`, which routes the decoded value through the existing, already case-insensitive `FileFormatFromString`, and call it from both decode paths — `ManifestReader.ReadEntry`, before the existing status/content validation and covering the normal and fallback entry paths (they share the same `*dataFile`), and `unmarshalAvroDataFileEntry`.

This mirrors what the other implementations do on this exact path — Java's `BaseFile.internalSet` → `FileFormat.fromString`, and PyIceberg's `FileFormat._missing_` — so it brings Go in line rather than introducing new leniency.

It also makes the manifest reader consistent with the rest of this library, which already normalizes everywhere else this field crosses the wire:

- `catalog/rest/scan_task_decoder.go` — the REST scan-planning decoder calls `FileFormatFromString` before building the data file.
- `table/writer.go` and `table/rolling_data_writer.go` — both writers normalize.

~~The Avro manifest reader was the only place treating wire casing as authoritative.~~

**Correction:** this was wrong, as @zeroshade found in review. `unmarshalAvroDataFileEntry` (`data_file_codec.go`), the other manifest-entry Avro decoder and the one behind the public `codec.DecodeDataFile`, had the identical bug.

## Behavior changes

Two user-visible consequences of normalizing on read, both flagged by @zeroshade in review:

1. **Metadata tables report the canonical spelling.** `table/inspect_files.go` appends `string(file.FileFormat())` directly, so `files`, `data_files`, `delete_files` and the `all_*` variants now show `PARQUET` for a manifest written with lowercase `parquet`.
2. **Rewriting a manifest persists the normalized spelling.** Reading a DuckDB-written table and running a delete writes `PARQUET` back to disk. Normalization is idempotent, so this is a one-time upgrade and not manifest churn.

## Note for reviewers

An unrecognized `file_format` is now a hard error at decode time, naming the offending file path, rather than surfacing later as a confusing reader-selection failure. This matches Java, where `FileFormat.fromString` throws. The lenient alternative — pass unknown values through untouched and let file-open fail — is a one-line change if reviewers prefer it.

The blast radius of that hard error is wider than scan. Because it fires at decode time, it also reaches paths that never open the data file — `inspect_files`, expire-snapshots and orphan cleanup — and it aborts the entire manifest read even for entries `iterManifest(discardDeleted=true)` would have discarded. That is unreachable with any format this library writes, and Java fails identically, but it is worth stating explicitly.

## Testing

`TestManifestReaderNormalizesFileFormat` writes a manifest through the Avro writer and reads it back, asserting the canonical constant comes out: lowercase/mixed/canonical `parquet`, lowercase `avro` and `orc`, `puffin` in a delete manifest, and the v1 fallback entry path; unrecognized `csv` and an empty `file_format` error.

`TestUnmarshalAvroDataFileEntryNormalizesFileFormat` is the equivalent table for the codec decoder, round-tripping through `MarshalAvroEntry` → `unmarshalAvroDataFileEntry`.

`make test` and `make lint` are clean.

### DIFF
--- a/data_file_codec_test.go
+++ b/data_file_codec_test.go
@@ -93,6 +93,46 @@ func TestDataFileCodecWithDroppedPartitionSource(t *testing.T) {
 	require.Equal(t, map[int]any{1000: nil, 1001: int32(3)}, decoded.Partition())
 }
 
+func TestUnmarshalAvroDataFileEntryNormalizesFileFormat(t *testing.T) {
+	spec := NewPartitionSpec()
+	schema := NewSchema(0)
+
+	tests := []struct {
+		name          string
+		written       FileFormat
+		expected      FileFormat
+		errorContains string
+	}{
+		{name: "spec lowercase", written: "parquet", expected: ParquetFile},
+		{name: "mixed case", written: "Parquet", expected: ParquetFile},
+		{name: "uppercase", written: ParquetFile, expected: ParquetFile},
+		{name: "lowercase orc", written: "orc", expected: OrcFile},
+		{name: "unknown format", written: "csv", errorContains: "unknown file format: csv"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			builder, err := NewDataFileBuilder(spec, EntryContentData,
+				"s3://bucket/table/data.parquet", ParquetFile, nil, nil, nil, 1, 1024)
+			require.NoError(t, err)
+			source := builder.Build().(*dataFile)
+			source.Format = tt.written
+
+			encoded, err := source.MarshalAvroEntry(spec, schema, 2)
+			require.NoError(t, err)
+
+			decoded, err := unmarshalAvroDataFileEntry(encoded, spec, schema, 2)
+			if tt.errorContains != "" {
+				require.ErrorContains(t, err, tt.errorContains)
+
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, tt.expected, decoded.FileFormat())
+		})
+	}
+}
+
 // TestManifestEntrySchemaForDistinguishesSameSpecIDDifferentPartitionTypes
 // guards against the cache-key collision that would occur if the cache
 // were keyed only by spec.ID(). Two tables can both use
--- a/data_file_codec_test.go
+++ b/data_file_codec_test.go
@@ -108,6 +108,7 @@ func TestUnmarshalAvroDataFileEntryNormalizesFileFormat(t *testing.T) {
 		{name: "uppercase", written: ParquetFile, expected: ParquetFile},
 		{name: "lowercase orc", written: "orc", expected: OrcFile},
 		{name: "unknown format", written: "csv", errorContains: "unknown file format: csv"},
+		{name: "empty format", written: "", errorContains: "unknown file format: "},
 	}
 
 	for _, tt := range tests {
--- a/manifest.go
+++ b/manifest.go
@@ -918,11 +918,9 @@ func (c *ManifestReader) ReadEntry() (ManifestEntry, error) {
 		tmp = tmp.(*fallbackManifestEntry).toEntry()
 	}
 	if df, ok := tmp.DataFile().(*dataFile); ok {
-		format, err := FileFormatFromString(string(df.Format))
-		if err != nil {
-			return nil, fmt.Errorf("manifest entry for %q has invalid file format: %w", df.Path, err)
+		if err := df.normalizeFormat(); err != nil {
+			return nil, err
 		}
-		df.Format = format
 	}
 	switch tmp.Status() {
 	case EntryStatusEXISTING, EntryStatusADDED, EntryStatusDELETED:
@@ -2776,6 +2774,20 @@ func (d *dataFile) setFieldIDToLogicalTypeMap(m map[int]string) {
 
 func (d *dataFile) setFieldIDToDecimalScaleMap(m map[int]int) {
 	d.fieldIDToDecimalScale = m
+}
+
+// normalizeFormat sets d.Format to the FileFormat constant matching its
+// decoded spelling, or returns an error if it names no known format.
+// Every Avro decode path must call it before the value is compared or
+// exposed.
+func (d *dataFile) normalizeFormat() error {
+	format, err := FileFormatFromString(string(d.Format))
+	if err != nil {
+		return fmt.Errorf("data file %q has invalid file format: %w", d.FilePath(), err)
+	}
+	d.Format = format
+
+	return nil
 }
 
 func (d *dataFile) ContentType() ManifestEntryContent { return d.Content }
--- a/manifest.go
+++ b/manifest.go
@@ -917,6 +917,13 @@ func (c *ManifestReader) ReadEntry() (ManifestEntry, error) {
 	if c.isFallback {
 		tmp = tmp.(*fallbackManifestEntry).toEntry()
 	}
+	if df, ok := tmp.DataFile().(*dataFile); ok {
+		format, err := FileFormatFromString(string(df.Format))
+		if err != nil {
+			return nil, fmt.Errorf("manifest entry for %q has invalid file format: %w", df.Path, err)
+		}
+		df.Format = format
+	}
 	switch tmp.Status() {
 	case EntryStatusEXISTING, EntryStatusADDED, EntryStatusDELETED:
 	default:
--- a/manifest.go
+++ b/manifest.go
@@ -2783,7 +2783,7 @@ func (d *dataFile) setFieldIDToDecimalScaleMap(m map[int]int) {
 func (d *dataFile) normalizeFormat() error {
 	format, err := FileFormatFromString(string(d.Format))
 	if err != nil {
-		return fmt.Errorf("data file %q has invalid file format: %w", d.FilePath(), err)
+		return fmt.Errorf("data file %q: %w", d.FilePath(), err)
 	}
 	d.Format = format
 
--- a/data_file_codec.go
+++ b/data_file_codec.go
@@ -141,6 +141,9 @@ func unmarshalAvroDataFileEntry(data []byte, spec PartitionSpec, schema *Schema,
 	if _, err := s.Decode(data, entry); err != nil {
 		return nil, fmt.Errorf("iceberg: unmarshalAvroDataFileEntry: %w", err)
 	}
+	if err := df.normalizeFormat(); err != nil {
+		return nil, fmt.Errorf("iceberg: unmarshalAvroDataFileEntry: %w", err)
+	}
 	df.specID = int32(spec.ID())
 	df.fieldNameToID = maps.nameToID
 	df.fieldIDToLogicalType = maps.idToType
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -2426,6 +2426,63 @@ func (m *ManifestTestSuite) TestManifestReaderRejectsInvalidEntries() {
 	}
 }
 
+func (m *ManifestTestSuite) TestManifestReaderNormalizesFileFormat() {
+	partitionSpec := NewPartitionSpecID(1,
+		PartitionField{FieldID: 1000, SourceIDs: []int{1}, Name: "VendorID", Transform: IdentityTransform{}},
+		PartitionField{FieldID: 1001, SourceIDs: []int{2}, Name: "tpep_pickup_datetime", Transform: IdentityTransform{}})
+	partitionSchema, err := partitionTypeToAvroSchema(partitionSpec.PartitionType(testSchema))
+	m.Require().NoError(err)
+	entrySchema, err := internal.NewManifestEntrySchema(partitionSchema, 2)
+	m.Require().NoError(err)
+
+	tests := []struct {
+		name          string
+		written       FileFormat
+		expected      FileFormat
+		errorContains string
+	}{
+		{name: "spec lowercase", written: "parquet", expected: ParquetFile},
+		{name: "mixed case", written: "Parquet", expected: ParquetFile},
+		{name: "uppercase", written: ParquetFile, expected: ParquetFile},
+		{name: "lowercase avro", written: "avro", expected: AvroFile},
+		{name: "unknown format", written: "csv", errorContains: "unknown file format: csv"},
+	}
+
+	for _, tt := range tests {
+		m.Run(tt.name, func() {
+			entry := *manifestEntryV2Records[0]
+			file := cloneDataFileAvroFields(entry.Data.(*dataFile))
+			file.Format = tt.written
+			entry.Data = file
+
+			mw := ManifestWriter{version: 2, spec: partitionSpec, schema: testSchema, content: ManifestContentData}
+			metadata, err := mw.meta()
+			m.Require().NoError(err)
+
+			var buf bytes.Buffer
+			writer, err := ocf.NewWriter(&buf, entrySchema,
+				ocf.WithSchema(entrySchema.String()), ocf.WithMetadata(metadata))
+			m.Require().NoError(err)
+			m.Require().NoError(writer.Encode(&entry))
+			m.Require().NoError(writer.Close())
+
+			manifest := &manifestFile{version: 2, SpecID: 1, Content: ManifestContentData}
+			reader, err := NewManifestReader(manifest, bytes.NewReader(buf.Bytes()))
+			m.Require().NoError(err)
+			defer reader.Close()
+
+			got, err := reader.ReadEntry()
+			if tt.errorContains != "" {
+				m.ErrorContains(err, tt.errorContains)
+
+				return
+			}
+			m.Require().NoError(err)
+			m.Equal(tt.expected, got.DataFile().FileFormat())
+		})
+	}
+}
+
 func (m *ManifestTestSuite) TestManifestEntryBuilder() {
 	dataFileBuilder, err := NewDataFileBuilder(
 		NewPartitionSpec(),
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -2432,11 +2432,12 @@ func (m *ManifestTestSuite) TestManifestReaderNormalizesFileFormat() {
 		PartitionField{FieldID: 1001, SourceIDs: []int{2}, Name: "tpep_pickup_datetime", Transform: IdentityTransform{}})
 	partitionSchema, err := partitionTypeToAvroSchema(partitionSpec.PartitionType(testSchema))
 	m.Require().NoError(err)
-	entrySchema, err := internal.NewManifestEntrySchema(partitionSchema, 2)
-	m.Require().NoError(err)
 
 	tests := []struct {
 		name          string
+		version       int
+		content       ManifestContent
+		entryContent  ManifestEntryContent
 		written       FileFormat
 		expected      FileFormat
 		errorContains string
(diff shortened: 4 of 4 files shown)


zeroshade (commented):
Thanks for this — the fix is correct, minimal, and genuinely spec-justified. I verified it rather than taking the description's word for it, and the substance holds up. One real gap and a couple of description corrections before merge.

## What I verified

- Reverting only the 7-line `manifest.go` hunk makes 4 of the 5 new subcases fail (`spec_lowercase` got `"parquet"`, `mixed_case` got `"Parquet"`, `lowercase_avro` got `"avro"`, `unknown_format` "An error is expected but got nil"). It is a real regression test, not a tautology.
- Spec grounding: `format/spec.md` names the values in **lowercase** (`avro`, `orc`, `parquet`, `puffin`) while Java writes uppercase via `FileFormat.name()`. Case-insensitive read is the only portable behaviour, so the framing in #1983 is right and this isn't a DuckDB quirk.
- Java parity by reading the source, not assuming: `FileFormat.fromString` is `valueOf(toUpperCase(Locale.ROOT))` and throws on unknown, and `BaseFile` calls it. Both the normalization *and* the new hard error match.
- Probes: the v1 fallback decode path normalizes; lowercase `orc` → `ORC`; lowercase `puffin` in a delete manifest → `PUFFIN`; and a round-trip through `ManifestWriter.Existing` followed by a raw non-normalizing decode shows `"PARQUET"` on disk — normalization is idempotent and doesn't churn manifests.
- Perf is a non-issue: `FileFormatFromString("PARQUET")` is 10–19 ns/op with **0 allocs** (`strings.ToUpper` returns all-uppercase ASCII unchanged). A 5000-entry `ReadEntry` loop measures 295,588 allocs/op both with and without the fix — identical.
- `go test . -count=1` → 1396 subtests pass; `-race` clean; `golangci-lint run ./...` → 0 issues; CI 15/15.

## Major — the sibling decoder has the identical bug

`data_file_codec.go:132-150` — `unmarshalAvroDataFileEntry`, the *other* manifest-entry Avro decoder (public entry point `codec.DecodeDataFile`), still treats wire casing as authoritative. I built a `dataFile` with `Format = "parquet"`, ran `MarshalAvroEntry` then `unmarshalAvroDataFileEntry`, and got `file_format = "parquet"` back. Every downstream comparison then mis-fires exactly as #1983 describes — `table/internal/interfaces.go:38` (`GetFile`), `table/scan_splits.go:34`, `table/scanner.go:433` (`IsDeletionVector`), `table/dv/deletion_vector.go:331`.

This matters because the `codec` package doc frames raw manifest bytes as in-contract input: *"The bytes it produces are the same Avro bytes a manifest carries for the corresponding value"* and *"Legacy manifests that already carry the field on the wire still decode correctly through `DecodeDataFile`."*

Fix: the same four lines after the `s.Decode` at `data_file_codec.go:141`, or better, factor the normalization into one helper both `ReadEntry` and `unmarshalAvroDataFileEntry` call — the shared `newDecodeEntry`/`*dataFile` plumbing already exists.

This also makes the description's claim *"The Avro manifest reader was the only place treating wire casing as authoritative"* false, which is worth correcting either way.

## Description omissions

Two real, user-visible behaviour changes aren't mentioned. Both are improvements — I'm asking for disclosure, not changes:

1. `table/inspect_files.go:665` does `b.fileFormat.Append(string(file.FileFormat()))`, so the `files` / `data_files` / `delete_files` / `all_*` metadata tables now report `PARQUET` for a lowercase-written manifest.
2. Rewriting a manifest persists the normalized spelling, so reading a DuckDB-written table and running a delete silently upgrades the casing on disk.

## Minor

- `manifest.go:920-926` — an unrecognized format now aborts the **entire** manifest read even for entries `iterManifest(discardDeleted=true)` would discard (probe: `vortex` on a `DELETED` entry → 0 entries, hard error; previously it decoded and was skipped). Unreachable today and Java hard-fails identically, so I'm not asking for the lenient variant — but the blast radius is wider than scan (`inspect_files`, expire-snapshots, orphan cleanup never open the data file yet would now fail). Worth a sentence.
- `manifest.go:923` uses `df.Path` where the adjacent first-row-id block at `:960` uses `df.FilePath()`. Pick one.
- Pre-existing asymmetry this PR makes more visible: the reader is now lenient while `NewDataFileBuilder` (`manifest.go:3082`) and `table/transaction.go:1099` stay exact-case — `NewDataFileBuilder(..., FileFormat("parquet"), ...)` still errors. Routing the builder through `FileFormatFromString` would be a coherent follow-up.
- Test coverage gaps (all four verified working by probe, so coverage not correctness): the v1 fallback path, `orc`, `puffin`-in-a-delete-manifest, and empty `file_format`. The v1 case is the one I'd add since the description makes an explicit claim about it — swapping in `internal.NewManifestEntrySchema(partSchema, 1)` and encoding a `fallbackManifestEntry` is about 3 lines in the existing table.

Not blocking on any of the above except that I'd like the `data_file_codec.go` decoder either fixed here or explicitly deferred to a follow-up issue, since it's the same bug the PR exists to fix.

---

wroever:
@zeroshade Thanks for the review. Have addressed the previously-unhandled codec path in [9a12014](https://github.com/apache/iceberg-go/pull/1984/commits/9a12014ca8884bdadcc3f02b68d6cbbac8e0defa), happy to make further edits as needed.

manifest.go
```diff
@@ -2771,6 +2776,20 @@ func (d *dataFile) setFieldIDToDecimalScaleMap(m map[int]int) {
...
 
+// normalizeFormat sets d.Format to the FileFormat constant matching its
+// decoded spelling, or returns an error if it names no known format.
+// Every Avro decode path must call it before the value is compared or
+// exposed.
+func (d *dataFile) normalizeFormat() error {
+	format, err := FileFormatFromString(string(d.Format))
+	if err != nil {
+		return fmt.Errorf("data file %q has invalid file format: %w", d.FilePath(), err)
+	}
```
zeroshade, line 2787: **nit** — Redundant error nesting: 'has invalid file format: unknown file format: csv'

normalizeFormat wraps FileFormatFromString's message, which already says 'unknown file format: <s>'. The combined string reads 'data file "..." has invalid file format: unknown file format: csv'. Something like 'data file %q: %w' would read cleanly while keeping the path context that motivated the wrap. Purely cosmetic.

data_file_codec_test.go
```diff
@@ -93,6 +93,46 @@ func TestDataFileCodecWithDroppedPartitionSource(t *testing.T) {
...
+		written       FileFormat
+		expected      FileFormat
+		errorContains string
+	}{
+		{name: "spec lowercase", written: "parquet", expected: ParquetFile},
+		{name: "mixed case", written: "Parquet", expected: ParquetFile},
+		{name: "uppercase", written: ParquetFile, expected: ParquetFile},
+		{name: "lowercase orc", written: "orc", expected: OrcFile},
+		{name: "unknown format", written: "csv", errorContains: "unknown file format: csv"},
+	}
```
zeroshade, line 111: **nit** — Codec test table is a strict subset of the manifest one

The manifest table covers avro, puffin and empty-format; the codec table stops at parquet/orc/csv. Both call the same normalizeFormat helper so this is not a correctness gap, but since the two tables are otherwise parallel, mirroring the empty-format row would make the asymmetry deliberate rather than incidental. Not worth a round-trip on its own.

zeroshade (approved):
The prior major (unnormalized sibling decoder in unmarshalAvroDataFileEntry) is genuinely fixed and pinned by a mutation-verified test, all four requested coverage gaps are now real non-vacuous subcases, and no un-normalized Avro decode path remains.

Re-review verification: 3 of 7 prior findings confirmed fixed at `9a12014` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *not fixed* — description omission: table/inspect_files.go:665 means the files/data_files/delete_files/all_* metadata tables now report PARQUET for a lowercase-written manifest
- *not fixed* — description omission: rewriting a manifest persists the normalized spelling, silently upgrading casing on disk
- *partially fixed* — minor: an unrecognized format now aborts the entire manifest read even for entries iterManifest(discardDeleted=true) would discard; blast radius wider than scan — worth a sentence

```text
go build ./... (clean); go vet . ./codec (clean); go test . ./codec -count=1 (ok 0.481s / 0.436s); go test ./table/... -count=1 (all 5 packages ok); go test . -race -count=1 -run 'TestManifests|TestUnmarshal|TestDataFileCodec|TestEncodeDecode' (ok, race-clean); gofmt -l on all 4 changed files (clean). Mutation runs: removed manifest.go:917-922 -> 8/9 subcases FAIL; removed data_file_codec.go:143-145 -> 4/5 subcases FAIL. Throwaway probe pr1984_probe_test.go (isFallback assertion + DELETED-entry hard-error) written, run, and deleted. Local golangci-lint v2.12.2 panics under Go 1.26 (goanalysis loader crash, environment issue) — deferred to the green CI lint job.
```

</details>

---

>

data_file_codec_test.go
wroever: 👍 Addressed in [a2867e3](https://github.com/apache/iceberg-go/pull/1984/commits/a2867e35660ea6728b2cbcd2da62da225594f077).

manifest.go
wroever: 👍 Addressed in [a2867e3](https://github.com/apache/iceberg-go/pull/1984/commits/a2867e35660ea6728b2cbcd2da62da225594f077).

manifest.go
```diff
@@ -2771,6 +2776,20 @@ func (d *dataFile) setFieldIDToDecimalScaleMap(m map[int]int) {
...
 
+// normalizeFormat sets d.Format to the FileFormat constant matching its
+// decoded spelling, or returns an error if it names no known format.
+// Every Avro decode path must call it before the value is compared or
+// exposed.
+func (d *dataFile) normalizeFormat() error {
+	format, err := FileFormatFromString(string(d.Format))
+	if err != nil {
+		return fmt.Errorf("data file %q: %w", d.FilePath(), err)
+	}
```
zeroshade, line 2787: **nit** — Corrected error message is pinned by no assertion

The delta's whole payload is the wording of this format string, but no test asserts it. Reintroducing the redundant 'has invalid file format:' prefix keeps the suite green, so the improvement can silently regress. Tightening one row per table to assert the full 'data file "...": unknown file format: csv' would lock it in. Cosmetic-only, so not worth a round-trip on its own.

data_file_codec_test.go
```diff
@@ -93,6 +93,47 @@ func TestDataFileCodecWithDroppedPartitionSource(t *testing.T) {
...
+		expected      FileFormat
+		errorContains string
+	}{
+		{name: "spec lowercase", written: "parquet", expected: ParquetFile},
+		{name: "mixed case", written: "Parquet", expected: ParquetFile},
+		{name: "uppercase", written: ParquetFile, expected: ParquetFile},
+		{name: "lowercase orc", written: "orc", expected: OrcFile},
+		{name: "unknown format", written: "csv", errorContains: "unknown file format: csv"},
+		{name: "empty format", written: "", errorContains: "unknown file format: "},
+	}
```
zeroshade, line 112: **nit** — empty-format row asserts a prefix common to every unknown-format error

errorContains "unknown file format: " is a substring of "unknown file format: csv" and of every other unknown-format message, so the row cannot distinguish the empty value from any other rejected spelling. It still verifies that an empty file_format is not silently accepted, which is the point of the row, so this is purely about assertion precision. Same applies to manifest_test.go:2456.

zeroshade (approved):
The redundant error nesting is genuinely fixed at manifest.go:2786 (probe-confirmed message, unwrap chain intact, no sentinel ever existed to break), both normalization call sites are independently pinned by mutation-verified tests, and the only residue is that the corrected message itself is guarded by no assertion.

Re-review verification: 5 of 7 prior findings confirmed fixed at `a2867e3` (each verified by mutating the fix and observing the suite go red, not by taking the claim on trust). Still open:
- *not fixed* — a test should pin the corrected message

```text
In worktree .pi-worktrees/pr1984 at a2867e3: go build ./... (OK); go vet . (OK); go test . -race -run 'TestManifests|TestUnmarshalAvroDataFileEntry|TestDataFileCodec' -timeout=300s -> ok 1.652s; go test ./catalog/glue/ -timeout=120s -> ok 0.867s (refuting a stale pi-lens 'setup failed' captured during my mutation window). Four mutation probes plus one throwaway pr1984_probe_test.go, all reverted; final 'git status --porcelain' empty.
```

</details>

---

>

manifest.go
```diff
@@ -917,6 +917,11 @@ func (c *ManifestReader) ReadEntry() (ManifestEntry, error) {
 	if c.isFallback {
 		tmp = tmp.(*fallbackManifestEntry).toEntry()
 	}
+	if df, ok := tmp.DataFile().(*dataFile); ok {
```
laskoviymishka, line 920: the `ok` guard here silently skips normalization if `DataFile()` ever returns something other than `*dataFile`. It's dead today (both the v2/v3 path and the v1 fallback populate `&dataFile{}`), but the invariant is load-bearing: a future entry variant that slipped through un-normalized would quietly recreate the original lowercase-comparison bug with no test failure.

Since `ReadEntry` owns the allocation, I'd either assert unconditionally, or keep the guard and return an error in the else. wdyt?

manifest.go
```diff
@@ -917,6 +917,11 @@ func (c *ManifestReader) ReadEntry() (ManifestEntry, error) {
...
+		if err := df.normalizeFormat(); err != nil {
+			return nil, err
```
laskoviymishka, line 922: the sibling decode path in `unmarshalAvroDataFileEntry` wraps this same error with `iceberg: unmarshalAvroDataFileEntry: %w`, but here we return it bare, so a failure surfacing from `ReadManifest`/`Entries` carries the file path with no call-site attribution.

I'd match it: `return nil, fmt.Errorf("iceberg: ReadEntry: %w", err)`, consistent with the status and content-type checks just below that already prefix their errors.

data_file_codec_test.go
```diff
@@ -93,6 +93,47 @@ func TestDataFileCodecWithDroppedPartitionSource(t *testing.T) {
...
+		written       FileFormat
+		expected      FileFormat
+		errorContains string
+	}{
+		{name: "spec lowercase", written: "parquet", expected: ParquetFile},
+		{name: "mixed case", written: "Parquet", expected: ParquetFile},
+		{name: "uppercase", written: ParquetFile, expected: ParquetFile},
+		{name: "lowercase orc", written: "orc", expected: OrcFile},
+		{name: "unknown format", written: "csv", errorContains: "unknown file format: csv"},
+		{name: "empty format", written: "", errorContains: "unknown file format: "},
```
laskoviymishka, line 111: the manifest-reader test covers all four formats but this codec table only hits parquet variants and orc. Since the codec path is independently callable, I'd round it out with lowercase avro and puffin so the two tests stay symmetric.

(puffin needs `source.Content = EntryContentPosDeletes` before encoding to get past the builder validation.)

laskoviymishka (approved):
Took an independent pass over this and it looks solid. The case normalization on the manifest read path is the right call, and I like that the tests pin the normalized constant rather than just asserting on the error string, since that's the part that survives a refactor.

Most of the substantive stuff was already caught and fixed in the earlier rounds, so I don't have anything to re-open, and the earlier error-message pinning thread looks covered by the existing `require.Equal` + `ErrorContains` assertions. A few small things left, none blocking, all inline: the bare `return nil, err` in `ReadEntry` where the codec sibling wraps with context, the `ok`-guarded type assertion that'll quietly skip normalization if that invariant ever shifts, and the codec test that could cover avro/puffin to stay symmetric with the manifest-reader test.

Separately and not for this PR: the write side still emits uppercase format strings against the spec's lowercase canonical (Java and rust do the same, so it's interop-safe and pre-existing). Its own follow-up if we want it.

Nothing here needs to hold up the merge. Fix or defer those and I'm happy.

PR #1989, opened 2026-09-03
[dnm] lazy create manifest creator to avoid empty man fest create error

michaelchendd:


michaelchendd:
sorry mistakenly opened here, closing

